PR #880, opened 2019-11-14
Update driving model

pd0wm:


ErichMoraga:
Hi Willem.  Can you provide a high level summary of this model change?

ErichMoraga:
I see the model was trained 4x longer, and is more stable per...
https://discordapp.com/channels/469524606043160576/621451618122072123/644670803115245621

PR #884, opened 2019-11-22
Remove ZMQ publish socket assert() from messaging layer

jyoung8607:
# Bug Fix
This PR fixes the new messaging library to accommodate unlogger playing back data in conjunction with controlsd and other ZMQ publishers, by making pub socket binding failures less fatal.

## Description
In a development environment, unlogger might run with other ZMQ publishers started beforehand. The original (if ugly) behavior was for the first publisher to win, unlogger would fail to bind those sockets, but would soldier on with the rest of its job. After the message library refactoring, failure to bind a ZMQ publisher socket is treated as unconditionally fatal.

In a perfect world, the bind result would be propagated upward and the "real" publishers would grow some higher-layer exception handling, so they could react to pub socket binding failures while unlogger could continue mostly ignoring them. But that's a lot of touches in a lot of places for tonight. This one-liner gets us back to where we were.

## Testing
Passes Docker testing here, and controlsd+unlogger can be used together.


pd0wm:
See my comment on the other PR. In the meantime I'd recommend using the `--disable` flag in unlogger to not bind certain sockets. eg. `--disable sendcan,controlsState,carState,carControl,carEvents,carParams` if you want to run controlsd.

jyoung8607:
Aha! Good pointer on the --disable flag, I didn't realize that existed. I was working off [an old Medium article](https://medium.com/@comma_ai/open-sourcing-openpilot-development-tools-a5bc427867b6) that just said to start controls before unlogger so it won the race. That'll work!

PR #893, opened 2019-11-26
Smart dsu

wocsor:
# Feature
This pull adds support for Smart DSU, a conditional CAN filter for 0x343 on Toyotas based on the Panda

## Description
Allows openpilot to detect the presence of the Smart DSU, send 343, and not send any static messages if ID 0x2FF is detected on the bus.
DSU will take over longitudinal control if 

## Testing
Tested the custom harness on a 2016 RAV4 Hybrid, a 2018 Corolla LE Eco, (both using grey pandas and giraffe setup) and ongoing tests with the same harness on a 2017 Toyota Prius (black panda + car harness)


rbiasini:
@wocsor so if I understand correctly, this PR needs a panda/car harness to do the forwarding of all the messages and just filter 0x343, correct?  
If so, we are generally supportive of this. Long term, it should not be allowed to disconnect the DSU entirely, but rather just filtering 0x343.

Specific to this PR: shouldn't you prevent  prevent all the DSU specific messages to be sent? see https://github.com/commaai/openpilot/pull/893/files#diff-551957d790a89965f4f76d8276a24c41L40

wocsor:
yep, that's actually exactly how this works. if 343 is detected on bus 0, it will not forward it from bus 2. my firmware implementation is a little rough and currently cannot be merged into panda, but it does work as expected. 

is the detection in carcontroller not enough to prevent all static messages? i know the camera ones were removed already

andrewcopenpilot:
@rbiasini @wocsor could we come up a name that is more generic than SmartDSU? I'm building a 'SmartASCM' for the GM side and a parameter with this function would be super useful to that project.

Looks like if I understand interface.py across different makes 'is_ecu_disconnected' references the control unit that does longitudinal control. Keeping in line with 'GasInterceptor' would 'ECUInterceptor' be a reasonable cross make name?

A further extension would be that for some makes (at least GM) safety control module may operate on multiple buses. So perhaps ECUInterceptorBus1, ECUInterceptorBus2? In the case of GM, if ECUInterceptorBus1=True and ECUInterceptorBus2 = False, than OP would operate in lateral only mode, but if ECUInterceptorBus1=True and ECUInterceptorBus2 = True then it would operate with full lat+lon control.

Thoughts?

rbiasini:
ECUInterceptor is ok. In the case of GM: instead of 1,2... I would use the bus name (e.g. ECUInterceptorBusPt)

andrewcopenpilot:
@rbiasini how standard are bus names across makes? I'm really used to thinking in terms of pt/chas/obj bus, but I don't know if those are all GM specific buses.

Was hoping we could come up with something generic enough to upstream.

rbiasini:
For GM, we can just use the GM specific bus names. No need to pick standard bus names that work across all brands

geohot:

PR #894, opened 2019-11-27
Fix logic for installing /data/pythonpath symlink

jyoung8607:
# Bug Fix
This PR fixes a minor but slightly annoying error message at openpilot startup.

## Old Behavior
An installed openpilot instance, one that's been run a few times, will emit a symlink creation error to stderr on startup. It's mostly harmless but I'm tired of looking at it.

```
root@localhost:/data/openpilot$ ./launch_openpilot.sh
ln: failed to create symbolic link '/data/pythonpath/openpilot': File exists
make: Nothing to be done for 'all'.
Result: Parcel(00000000 00000001   '........')
dinfo 1920x1080
...
```
The first time OP is launched, the offending `ln -s` in `launch_chffrplus.sh` will do its job as expected. The second time, `ln` will follow the existing /data/pythonpath symlink and drop a garbage symlink at /data/openpilot/openpilot. The third and subsequent times OP is launched, `ln` will follow the symlink and fail because of the new garbage symlink.

## Fixed Behavior
Use `ln -sfn` to replace any symlink found at /data/pythonpath without following it. Any leftover garbage symlink will eventually get deleted with `git reset --hard` at the next OP update .

## Testing
LGTM in Docker here, although I don't think Docker would catch issues here. Anyway, it's a pretty simple and well-understood change.


pd0wm:
This fix was already done internally. Will ship with next release.

Issue #890, opened 2019-11-25
Camera radar vs steering angle


arne182:
At high steering angles 100-400° detected objects should be ignored at a distance where the car will safely pass the object.

**How to reproduce or log data**
Drive at low speed and turn into a road with parked traffic on both sides.

**Expected behavior**
If the object is inside the view of the camera. Reduce the distance of detectable objects with steering angle. I.e. if making a U-turn with steering angle at 400°. Do not brake for an object 10m away as this object will not be reached at that steering angle.

** Device/Version information (please complete the following information):**
 - Device: EON/EON Gold
 - Version: 0.6.6
 - Car make/model: all with low speed engagable and longitudinal control.

HaraldSchafer:
This is one of the many problems inherent to ACC type longitudinal control. Will be fixed as e2e longitudinal control gets phase in

PR #896, opened 2019-11-29
Do not brake if no collision possible

arne182:
To fix issue:
https://github.com/commaai/openpilot/issues/890
If the steering angle is above 100° then do a check for the turning radius of the car assuming that the edge of the car is 1m away from the camera. Find the center of the rotation and add an extra meter for safety. if the radar point lies within the turning radius then brake, if it is outside the radius do not consider it for braking as we will not be driving into it. The math is taken from here but is pretty simple to understand.

https://www.quora.com/What-is-the-method-to-calculate-turning-radius-if-only-the-steering-ratio-and-vehicles-basic-dimensions-are-known

https://math.stackexchange.com/questions/198764/how-to-know-if-a-point-is-inside-a-circle

### DIFF
--- a/selfdrive/controls/lib/planner.py
+++ b/selfdrive/controls/lib/planner.py
@@ -88,12 +88,23 @@ class Planner():
 
     self.params = Params()
 
-  def choose_solution(self, v_cruise_setpoint, enabled):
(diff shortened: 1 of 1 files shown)


geohot:

PR #466, opened 2018-12-15
GM: soft disengage on PCM fault

vntarasov:
Happens when
- first attempt to engage is done with "resume" instead lf "set" button.
- attempt to engage outside of PCM's speed limits (shouldn't happen unless people override min allowed engage speed).
- pre-enable doesn't work on 2018 Volt.

In all those cases, showing openpilot as engages is confusing, because car ignores gas commands.

PR #908, opened 2019-12-11
GM fix 'Controls Failed' on resuming after cycling main cruise switch

ghost:
# Feature
GM fix 'Controls Failed' on resuming after cycling main cruise switch

## Description
The GM ECM independently tracks two cruise states, 'main on' and 'speed is set'.
Both must be true when resuming or the ECM generates a cruise fault. If
the ECM is driven to the main off state by the main cruise switch,
the set speed state is reset to false.

The OEM GM flow is if main cruise is cycled, the ASCM resets its speed
to 0, and the ECM goes into the no set speed state. 'Resume' presses will be
ignored (as the ASCM has a resume speed of 0). The 'Set' Button must be
pressed while meeting the minimum speed requirements to drive the ECM
and ASCM into the correct states to allow for ACC to operate.

Currently in OP controlsd remembers previously set speeds when cycling
the main cruise switch. This causes controlsd and the GM ECM's
respective state machines to get out of sync.

For example if we have a set speed of 55 mph, we press main cruise button once,
controlsd goes into a disabled state, the GM ECM goes into the 'main
off' and 'no speed set' state. We press the main cruise button a second time,
controlsd goes into the enabled state with the remembered speed of 55
mph, while the GM ECM goes into a 'main on' and 'no speed set' state. The UI
presents the user with the controlsd remembered 55 mph, so it is logical to
press 'Resume' if that is desired, however due to the GM ECM still being
in the 'no speed set' state, the ECM generates a cruise fault. Currently this
results in a frightening 'Controls failed' message on the Eon.

This pull request adds a new controlsd event RESET_V_CRUISE
which allows the gm interface to keep controlsd in sync with the GM ECM.
On 'main on' presses, the gm interface triggers a reset of v_cruise_kph
back to 0. On pressing 'Resume' the GM interface now also checks to make
sure that v_cruise_kph has been set before sending a ENABLE event on Resume
presses.

## Testing
Tested on 2018 Cadillac ATS, Cycled main switch and attempted resumes. No 'Controls Failed' messages were able to be generated. Due to a one off giraffe being used, along with other custom code, this should also be tested by Volt users.

### DIFF
--- a/cereal/car.capnp
+++ b/cereal/car.capnp
@@ -19,6 +19,7 @@ struct CarEvent @0x9b1657f34caf3ad3 {
   immediateDisable @6 :Bool;
   preEnable @7 :Bool;
   permanent @8 :Bool;
+  resetVCruise @9 :Bool;
 
   enum EventName @0xbaa8c5d505f727de {
     # TODO: copy from error list
--- a/selfdrive/car/gm/interface.py
+++ b/selfdrive/car/gm/interface.py
@@ -301,15 +301,15 @@ class CarInterface(CarInterfaceBase):
       # handle button presses
       for b in ret.buttonEvents:
         # do enable on both accel and decel buttons
-        # The ECM will fault if resume triggers an enable while speed is set to 0
-        if b.type == ButtonType.accelCruise and c.hudControl.setSpeed > 0 and not b.pressed:
+        # The ECM will fault if resume triggers an enable while speed is unset (unset is greater than 70 m/s)
+        if b.type == ButtonType.accelCruise and c.hudControl.setSpeed < 70 and not b.pressed:
           events.append(create_event('buttonEnable', [ET.ENABLE]))
         if b.type == ButtonType.decelCruise and not b.pressed:
           events.append(create_event('buttonEnable', [ET.ENABLE]))
         # do disable on button down
         if b.type == ButtonType.cancel and b.pressed:
           events.append(create_event('buttonCancel', [ET.USER_DISABLE]))
-        # The ECM independently tracks a ‘speed is set’ state that is reset on main off.
+        # The ECM independently tracks if speed has been set previously. That state is reset on main off.
         # To keep controlsd in sync with the ECM state, generate a RESET_V_CRUISE event on main cruise presses.
         if b.type == ButtonType.altButton3 and b.pressed:
           events.append(create_event('buttonCancel', [ET.RESET_V_CRUISE, ET.USER_DISABLE]))
--- a/selfdrive/car/gm/interface.py
+++ b/selfdrive/car/gm/interface.py
@@ -301,11 +301,18 @@ class CarInterface(CarInterfaceBase):
       # handle button presses
       for b in ret.buttonEvents:
         # do enable on both accel and decel buttons
-        if b.type in [ButtonType.accelCruise, ButtonType.decelCruise] and not b.pressed:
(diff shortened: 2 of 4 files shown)


kishkaru:
I tested this change on my Volt and it works as expected. When the main cruise on/off switch is toggled, the previously set cruise speed is "forgotten" on both the car side and EON, so pressing RES (+) on the wheel is a noop (as opposed to previously an error). Furthermore, pressing RES (+) on first ignition is now also a noop. This now mimics the behaviour of the stock GM ACC system.

rbiasini:
nice. Can this be done without adding a new event type? That's a pretty invasive change.

andrewcopenpilot:
@rbiasini I've attempted to solve the problem within the GM specific code, however I think the best we can do there is what Vasily has already done here: https://github.com/commaai/openpilot/pull/466

Sadly the questions in #gm have gone from "why doesn't op accelerate?" to "My op is totally broken I'm getting 'Controls Failed' all the time." 

I understand that this is a change in the shared logic for the entirety of openpilot. My hope was GM isn't the only car that has the CC UI of cycling main off/on causes a previously set speed to be forgotten. I believe this is how most cars function as a hold over from days when cars didn't have LCD screens to communicate detailed state to the user. Possibly this could be an issue for other brands as their communities try to get full lat+long support?

Any thoughts on an alternative that might have a better chance of being upstreamed, or any improvements to this strategy that would make it more palatable to upstream?

andrewcopenpilot:
@rbiasini I think I might want to try a little more around this. I still think we need to be able to reset the v_cruise_kph in controlsd whenever main switch is toggled, but I've discovered this isn't the only case that results in a Controls Failed message.

Another case is if you set OP while the gas is depressed. The ECM also sets an ACC state of 3 in this case too, which we've labeled as ACC Fault. I don't think its actually a fault state. I think its actually more of a signal to the ASCM that the ECM is overriding its request for acceleration. Ie in the OEM system if you have ACC enabled, and you depress the accelerator, you don't get a disengage or an an error, you get a message that says that ACC braking is unavailable during manual acceleration. The ACC state on the ECM is set to 3, but as soon as the accelerator pedal is released, the message goes away ACC takes over again and the ECM ACC state goes back to 1 (enabled).

I think OP should also mimic the ASCM in this case as well. If we simply ignore the ACC state of 3, the system doesn't recover automatically like the OEM system. I need to do some more digging into the OEM can logs to figure out how the ASCM and ECM work together to handle recovery of ACC control when the driver presses and releases the gas.

andrewcopenpilot:
@rbiasini I don't know if it is a no go to add a new event type, but I cleaned things up a bit and fixed a bug on first startup.

I realized that v_cruise_kph's initial state is 255 not 0. So I updated the reset event to set it back to its initial value of 255.

I thought more about if there are any other options that don't involve creating a new event type. We could have a state machine in the GM code that tracks this and overrides what we display on the dash/hud, but then the OP UI could end up displaying one speed while the dash would display no set speed, so it would be a bit confusing.

geohot:

kishkaru:
You didn't need to close it. It's possible to change the target branch in an existing open PR. We can re-open it and change the target merge branch. 

--
Kishan Karunaratne

sshane:
Would anyone like to re-open this? It seems like it may be causing a large portion of ACC faults with GMs still.

PR #909, opened 2019-12-11
Safe updater

pd0wm:


selfdrive/updated.py
```diff
@@ -1,57 +1,267 @@
...
+  if not os.geteuid() == 0:
+    cloudlog.error("updated must be launched as root!")
+    exit(1)
+
+  ov_lock_fd = open(OVERLAY_LOCK, 'r')
+  try:
+    fcntl.flock(ov_lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+    init_ovfs()
+  except IOError:
+    print("couldn't get overlay lock; is another updated running? skipping init and continuing in update-only mode...")
```
pd0wm: Should we continue here? Maybe it's better to not run updater at all.

selfdrive/updated.py
```diff
@@ -1,57 +1,267 @@
...
+# **** main loop ****
+
+def main(gctx=None):
+  if not os.geteuid() == 0:
+    cloudlog.error("updated must be launched as root!")
+    exit(1)
+
+  ov_lock_fd = open(OVERLAY_LOCK, 'r')
+  try:
+    fcntl.flock(ov_lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
```
pd0wm: Don't forget to clear the lock on boot/start_chffrplus.sh in case updated crashed/shut down while performing an update.

pd0wm: You're probably aware, but `/tmp` is persistent across reboots on NEOS

selfdrive/updated.py
```diff
@@ -1,57 +1,267 @@
...
       r = r[:r.find(b'\n\n')] # Slice latest release notes
       params.put("ReleaseNotes", r + b"\n")
-    except:
+    except Exception:
       params.put("ReleaseNotes", "")
+    params.put("UpdateAvailable", "1")
+  else:
+    params.put("UpdateAvailable", "0")
 
+  if with_date:
```
pd0wm: This should probably moved inside the `.update_succeeded` check.

pd0wm: Personal preference, I would rename the parameter to something like `write_last_update_time`.

selfdrive/updated.py
```diff
@@ -1,57 +1,267 @@
...
+  upd_lock_fd = open(UPDATE_LOCK, 'r')
+  try:
+    fcntl.flock(upd_lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+  except OSError:
+    print("couldn't get update lock; is another updated running? going back to sleep...")
+    upd_lock_fd.close()
+    return
+
+  cloudlog.info("attempting git update inside staging overlay")
+  os.chdir(OVERLAY_MERGED)
```
pd0wm: Can we do all these things without calling `os.chdir`? That changes some global state that I don't think is needed.

selfdrive/updated.py
```diff
@@ -1,57 +1,267 @@
...
+  if os.path.isfile(os.path.join(BASEDIR, ".update_succeeded")):
+    os.remove(os.path.join(BASEDIR, ".update_succeeded"))
+  if os.path.isdir(STAGING_ROOT):
+    shutil.rmtree(STAGING_ROOT)
+
+  for dirname in [STAGING_ROOT, OVERLAY_UPPER, OVERLAY_METADATA, OVERLAY_MERGED, FINALIZED]:
+    os.mkdir(dirname, 0o755)
+  if not os.lstat(BASEDIR).st_dev == os.lstat(STAGING_ROOT).st_dev:
+    cloudlog.error("base and staging directories are on different filesystems, not valid for overlay FS!")
+    exit(1)
```
pd0wm: Maybe throw a RunTimeError instead of cloudlog + exit. If I look at other uses of cloudlog.error they are all recoverable errors.

selfdrive/updated.py
```diff
@@ -1,57 +1,267 @@
...
+  r = run(NICE_LOW_PRIORITY + ["git", "fetch"], OVERLAY_MERGED)
+  cloudlog.info("git fetch success: %s", r)
+
+  cur_hash = run(["git", "rev-parse", "HEAD"], OVERLAY_MERGED).rstrip()
+  upstream_hash = run(["git", "rev-parse", "@{u}"], OVERLAY_MERGED).rstrip()
+
+  cloudlog.info("comparing %s to %s" % (cur_hash, upstream_hash))
+  if cur_hash != upstream_hash:
+    cloudlog.info("git reset in progress")
+    r = [run(NICE_LOW_PRIORITY + ["git", "reset", "--hard", "@{u}"], OVERLAY_MERGED),
```
pd0wm: should we check the exit codes of these calls?

selfdrive/updated.py
```diff
@@ -1,57 +1,267 @@
...
+def run(cmd, cwd=None):
+  return subprocess.check_output(cmd, cwd=cwd, stderr=subprocess.STDOUT, encoding='utf8')
+
+# **** meaty functions ****
+
+def init_ovfs():
+  if os.path.ismount(OVERLAY_MERGED):
+    # We had to have the overlay lock to get this far, so this must be a stale
+    # mount from a terminated updated process.
+    run(["umount", OVERLAY_MERGED])
```
pd0wm: Let's add a `cloudlog.error()` here since this should never happen in practice.

selfdrive/updated.py
```diff
@@ -1,57 +1,267 @@
...
+      os.symlink(os.readlink(source_obj), target_obj)
+      os.chmod(target_obj, S_IMODE(st[ST_MODE]), follow_symlinks=False)
+    else:
+      # Ran into a FIFO, socket, etc. Should not happen in OP install dir.
+      # Ignore without copying for the time being, revisit if necessary.
+      pass
+    os.chown(target_obj, st[ST_UID], st[ST_GID], follow_symlinks=False)
+    os.utime(target_obj, (st[ST_ATIME], st[ST_MTIME]), follow_symlinks=False)
+
+def finalize_from_ovfs():
```
pd0wm: Where is this function called?

selfdrive/updated.py
```diff
@@ -1,57 +1,267 @@
...
+  # Given a search root, produce a dictionary mapping of inodes to relative
+  # pathnames of regular files (no directories, symlinks, or special files).
+  inode_map = {}
+  os.chdir(search_dir)
+  for root, dirs, files in os.walk('.', topdown=True):
+    for file_name in files:
+      full_name = os.path.join(search_dir, root, file_name)
+      st = os.stat(full_name)
+      if S_ISREG(st[ST_MODE]):
+        inode_map.update({st[ST_INO]: full_name})
```
pd0wm: why use dict.update with a single item?

selfdrive/updated.py
jyoung8607: At some point (I can't find it now) a Comma employee on Discord recommended someone run updated manually to force an immediate update attempt. The idea here (which is not yet completed) was to preserve that ability by running in a one-shot mode if it can't get the lock. Doing it safely will require wrapping attempt_update() in a separate lock. I can certainly throw that overboard if you want to make a design statement that updated should not be invoked outside of manager.

selfdrive/updated.py
jyoung8607: The file itself is persistent, but my understanding is the advisory lock will disappear on reboot (or even death of the locking process). I could have it garbage-collected on reboot, but I'd be recreating it immediately...

selfdrive/updated.py
jyoung8607: There were a couple of design reasons I did so, which have mostly been discarded at this point. I'll go back and tweak the directory walk and object copy operations to work with absolute paths instead of relative, and get rid of the chdirs.

selfdrive/updated.py
jyoung8607: Can do.

selfdrive/updated.py
jyoung8607: Generally yes; I intend to do so. I have some experimentation to do, deliberately corrupting the Git repo by various means, to see how I can best detect that result from Git during updated runtime. I may or may not be able to get that from numeric exit code, I may need to parse its output. For example, I may not be able to distinguish git fetch failure due to poor Internet connectivity (recoverable) from failure due to Git repo corruption (requires active measures to reinit Git).

selfdrive/updated.py
jyoung8607: Agreed.

selfdrive/updated.py
pd0wm: We can throw that overboard if we write a logfile somewhere when the update fails. As long as we have an easy way to inspect the last updater failure it's fine.

selfdrive/updated.py
pd0wm: Oh didn't know that it get's unlocked when the process dies. Think we just have to verify that that also works when we pull power.

selfdrive/updated.py
jyoung8607: You caught an interim version. I'm making a minor design change.

I removed it from the main update loop because Git turns out to be overly clever, and somehow notices if the hard-link count increases. So after I finalized, Git would decide everything in OVERLAY_MERGED had been touched and zap it during `git reset --hard`, which broke all the hard-links.

New approach will be to finalize exactly once, at bootup, assuming the success flag is present. I'll do that by having continue.sh call the merged version of updated with a new command-line switch. Finalizing only takes about a second so it won't be a huge contributor to boot time.

selfdrive/updated.py
pd0wm: Well, we never have to nuke the git repo, right? Can't we always assume the git repo in the BASEDIR is correct? We just have to start with a new overlay if any of these commands fail.

selfdrive/updated.py
jyoung8607: Valid observation, I'll simplify that.

selfdrive/updated.py
jyoung8607: Okay, I will reject all attempts to start updated if the overlay lock is already held, and log something accordingly.

selfdrive/updated.py
jyoung8607: If I recall correctly, for non-NFS local file scenarios, advisory flock()ing is a purely in-memory live-kernel-state object which never hits disk, so we don't much care about the file persisting after sudden crash/power loss scenarios. The file can be thought of as just an identifier for the lock and a proxy for access permissions.

jyoung8607: That said, it'll do no harm to clean it up and recreate it for safety's sake, so I'll add that.

selfdrive/updated.py
jyoung8607: > Can't we always assume the git repo in the BASEDIR is correct?

It _should_ be, but geohot had a strong opinion about keeping the ability to develop directly on-EON, and if that's happening:
- BASEDIR .git is now re-exposed to the corruption scenarios updated is designed to protect it from
- I have to watch and plan for BASEDIR .git to be messed with while updated is running (have not added the check for this in my code yet)

jyoung8607:
Can be closed, superseded by #923.

geohot:

PR #917, opened 2019-12-15
Increase reaction time of planner

arne182:
If a lead crosses our path open pilot will brake late and even when the lead is gone card on braking for the next few moments. This fix should increase the reaction time and not leave you braking when there is no lead.


geohot:

PR #922, opened 2019-12-17
Added flicker handling for blinkers in Subarus

michaelhonan:
# Description
Testing the unreleased ALC feature revealed a flaw in the CarState values for the blinkers. The values for the blinkersare mapped to the Dashlights in the DBC, and this means that when the blinkers are activated, they flicker between True and False at the same rate that the blinkers "blink". 

This issue causes ALC to act strangely, having the lines disappear and reappear in accordance to the value of the blinkers in CarState.

My solution uses a counter that counts up to a certain threshold (50 atm, little longer than the blink rate, which is ~32) every frame with the counter resetting to 0 after every blinker light state change. After the threshold is hit, it then sets the indicator status to False. 

A few other cars may require a similar change if they do not have a constant value when the blinkers are activated.

# Testing
I've been running ALC for the past few hours, and it's been working. Print tests in tmux show good consistent results as well.

### DIFF
--- a/selfdrive/car/subaru/carstate.py
+++ b/selfdrive/car/subaru/carstate.py
@@ -86,12 +86,18 @@ class CarState():
   def __init__(self, CP):
     # initialize can parser
     self.CP = CP
-
     self.car_fingerprint = CP.carFingerprint
+
     self.left_blinker_on = False
     self.prev_left_blinker_on = False
+    self.prev_left_blinker_light_on = False
+    self.left_blinker_interval = 0
+
     self.right_blinker_on = False
     self.prev_right_blinker_on = False
+    self.prev_right_blinker_light_on = False
+    self.right_blinker_interval = 0
+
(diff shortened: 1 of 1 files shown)


pd0wm:
Can we find the blinker stalk in the CAN data? I would prefer a change to the DBC file.

mchonan1996:
100% with you there, and still actively looking for it. I refuse to believe there isn't data on CAN regarding this.

ShaneSmiskol:
> 100% with you there, and still actively looking for it. I refuse to believe there isn't data on CAN regarding this.

I think @arne182 has a module to find differences in CAN messages. You turn the blinker on, hit enter, turn it off, hit enter, etc. and it narrows down which possible signals it could be!

mchonan1996:
I'll check it out, thanks!

geohot:

PR #923, opened 2019-12-17
Safe Updater: Part Deux

jyoung8607:
# Feature
This is a revised and tested version of the safe updater I've been working on with @rbiasini and @pd0wm. It should be pretty robust in the face of typical crash or power failure situations, most stupid user tricks, and even a couple smart user tricks.

I believe this incorporates all the feedback from Willem on the last round #909. I was able to solve the finalize problem without pushing it off to next bootup.

# Validation
* Works for me
* Does not break the Docker build
* Absolutely noCamelCaseWhatSoEver

I don't know that I can perform overlay mounts in Docker containers and cloud CI services, but if you want, I can write up some unit tests that exercise the rest of the tree copy-by-hardlink operation.

Per our discussion a couple weeks ago, Comma can and should test this using their own EONs to explore failure scenarios I don't have available to test here. I also expect you'll need to tweak the continue.sh part for your internal needs.

# Extra Requirements
* NEOS 14 with OverlayFS in the kernel, commaai/android_kernel_comma_msm8996#2
* Update to continue.sh, wherever that comes from... the installer? NEOS userland? Pasted below.

```
#!/usr/bin/env bash

BASEDIR="/data/openpilot"
STAGING_ROOT="/data/safe_staging"

# Check to see if there's a valid overlay-based update available. Conditions
# are as follows:
#
# 1. The BASEDIR init file has to exist, with a newer modtime than anything in
#    the BASEDIR Git repo. This checks for local development work or the user
#    switching branches/forks, which should not be overwritten.
# 2. The FINALIZED consistent file has to exist, indicating there's an update
#    that completed successfully and synced to disk.

if [ -f "${BASEDIR}/.overlay_init" ]; then
  find ${BASEDIR}/.git -newer ${BASEDIR}/.overlay_init | grep -q '.' 2> /dev/null
  if [ $? -eq 0 ]; then
    echo "${BASEDIR} has been modified, skipping overlay update installation"
  else
    if [ -f "${STAGING_ROOT}/finalized/.overlay_consistent" ]; then
      echo "Valid overlay update found, installing"
      mv $BASEDIR /data/safe_staging/old_openpilot
      mv "${STAGING_ROOT}/finalized" $BASEDIR
    else
      echo "No valid overlay update found, skipping overlay update installation"
    fi
  fi
else
  echo "No update performed or ${BASEDIR} has been modified, skipping overlay update installation"
fi

cd /data/openpilot
exec ./launch_openpilot.sh
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,8 @@ venv/
 .tags
 .ipynb_checkpoints
 .idea
+.overlay_canary
+.overlay_consistent
 .sconsign.dblite
 .vscode
 model2.png
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,7 @@ venv/
 .tags
 .ipynb_checkpoints
 .idea
-.overlay_canary
+.overlay_init
 .overlay_consistent
 .sconsign.dblite
 .vscode
--- a/launch_chffrplus.sh
+++ b/launch_chffrplus.sh
@@ -14,14 +14,6 @@ function launch {
   # Wifi scan
   wpa_cli IFNAME=wlan0 SCAN
 
-  # apply update
-  if [ "$(git rev-parse HEAD)" != "$(git rev-parse @{u})" ]; then
-    git reset --hard @{u} &&
-    git clean -xdf &&
-
-    exec "${BASH_SOURCE[0]}"
-  fi
-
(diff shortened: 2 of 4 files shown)


geohot:

PR #931, opened 2019-12-29
Fix CIVIC_BOSCH fingerprint for car harness users

csouers:
## Description
Adds the following unique IDs from OBD2 to the combo fingerprint. `423: 2, 507: 1, 815: 8, 825: 4, 846: 8, 881: 8, 882: 4, 888: 8, 918: 7, 983: 8, 1064: 7, 1092: 1, 1115: 2, 1125: 8, 1127: 2, 1296: 8` These were previously unseen and a current workaround is to not use comma power on these cars. This fixes this issue.

## Testing
Docker tests passed. Works fine on my vehicle.


rbiasini:
Not sure I understand: why is the fingerprint from OBD2 relevant? fingerprint should come from the CAN trace seen at the stock camera intercept.

csouers:
The fingerprinting process must grab these frames before it completes the vin query and toggles the relay. Depending on the speed of openpilot starting and finally toggling the relay, it will sometimes see all or only some of these messages (lower priority messages sometimes aren't captured, but I've included them to prevent errors by exclusion).

Cabana for review with some of the obd2 messages making it through before the relay toggle. https://my.comma.ai/cabana/?route=0a78dfbacc8504ef%7C2020-01-05--13-19-45&exp=1609899171&sig=dSz4rJq3JqkAEMu2q5lnh6YcdKOZnmA2iV8%2BRWPxJq4%3D&max=14&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2F0a78dfbacc8504ef%2Fbc8b4357d60d9c90359de8a61b4763d1_2020-01-05--13-19-45

gregjhogan:
CR-V has the same problem, seems to be messages only on OBD port leaking into fingerprinting.  I think what we need to do is clear the panda RX buffer after VIN query and before fingerprinting.

csouers:
I have to correct my prior comment. I was mistaken about the relay. I didn’t know at the time that the switch for the obd2 to stock camera f-can flip was being done electronically. I hope that makes more sense, Riccardo.

gregjhogan:
https://github.com/commaai/panda/issues/422 will resolve this issue by moving OBD port CAN messages to a different bus ID

geohot:

PR #935, opened 2020-01-01
Add support for 2018 Lexus ES 300h + Safety Improvements

TheInventorMan:
#2 Car support
This pull request adds support for Lexus ES 300h 2018, Standard. This should also work with 2016 and 2017 models as well, since they all use the same Lexus Safety System+ ECUs according to 
[this document.](https://www.lexus.com/documents/safety/Lexus-Safety-System-Plus-062017.pdf)

This is an explorer link to a drive with openpilot system enabled: [https://my.comma.ai/dc7629e7a6909176/1577918213342/1577919136565](https://my.comma.ai/dc7629e7a6909176/1577918213342/1577919136565)

~~ACC is handled by stock since the DSU is quite difficult to reach 😭. As a result, OP can only be engaged above 28mph.~~

Managed to disconnect the DSU and vehicle is now full stop-and-go capable. 
Explorer link: [https://my.comma.ai/dc7629e7a6909176/1578078720843/1578081292648](https://my.comma.ai/dc7629e7a6909176/1578078720843/1578081292648)

Also, there was an issue with disengaging on brake pedal press, and it seems that this car was broadcasting the signal on the BRAKE_MODULE2 message instead of BRAKE_MODULE. Since it's possible that this issue could arise with future Toyota vehicles, I tweaked the carstate.py code to OR the two signals.


TheInventorMan:
Added a block in the panda Toyota safety code to check if the brake pedal has been pressed.

TheInventorMan:
Discovered that this car model broadcasts its gas pedal state on a different message, so I added a check for this in the panda safety code

geohot:

PR #942, opened 2020-01-06
ECU Interceptor filtering the GM ASCM

andrewcopenpilot:
# Feature
This pull requests adds support for an ECU Interceptor filtering the GM ASCM

## Description
First crack at implementing a system for allowing OP to support an ECU Interceptor while still functioning correctly if the ASCM has been completely disconnected.

GasRegen/Friction Brake/ASCMStatus commands have all had proxy messages on unused IDs added to the respective DBC files. If OP detects an ECU Interceptor, rather than emulating the ASCM and sending these messages directly, OP will send the desired inputs to these proxy messages. The ECU Interceptor handles overriding the ASCM messages while implementing its own safety checks that mirror safety_gm.h in the Panda code.

This strategy is used so that while the Eon is booting (or driving without the Eon) OEM ACC is available and there are no dash board warnings. As soon as the Eon is ready and connected, the ECU interceptor overrides with the Eon messages while merging the rolling counters so that there are no dropped messages at the transition point.

## Testing
Limited testing with Interceptor if the PR gets to a point where upstream is viable, testing will need to include the ASCM disconnected install path as well.


geohot:

PR #950, opened 2020-01-16
fix for 0.7.1

arne182:
Not needed anymore


arne182:
This does not fix the issue

PR #964, opened 2020-01-18
Add orange alert if you are not running a devel or release branch

geohot:
Untested, who wants to try?


geohot:
Now tested in the simulator!

PR #966, opened 2020-01-18
add 2017 Civic Hatchback LX fp

cfranyota:
for support with Black Panda.


cfranhonda:
Why? This is same one that passed on 7.0

geohot:
I restarted the build?

PR #912, opened 2019-12-14
Hyundai, Kia and Genesis (HKG)

xx979xx:
8 new cars
12 additional fingerprints
SCC Radar support
Lane departure warning support
Visible lane support
Fix auto resumes after car stopped
support for car without SCC
Fix hard fault for Genesis
modernise the code
update supported cars info

PR #970, opened 2020-01-19
Hyundai-Kia-Genesis (HKG)

xx979xx:
8 new cars
12 additional fingerprints
SCC Radar support
Lane departure warning support
Visible lane support
Fix auto resumes after car stopped
Fix hard fault for Genesis
modernise the code
update supported cars info
check prev PR for more #912


xx979xx:
this PR has to accompany by:
[openddbc](https://github.com/commaai/opendbc/pull/211)
[panda](https://github.com/commaai/panda/pull/427)

selfdrive/car/hyundai/carstate.py
```diff
@@ -181,70 +201,85 @@ def update(self, cp, cp_cam):
...
     self.left_blinker_flash = cp.vl["CGW1"]['CF_Gway_TurnSigLh']
     self.right_blinker_on = cp.vl["CGW1"]['CF_Gway_TSigRHSw']
     self.right_blinker_flash = cp.vl["CGW1"]['CF_Gway_TurnSigRh']
-    self.steer_override = abs(cp.vl["MDPS11"]['CR_Mdps_DrvTq']) > STEER_THRESHOLD
+    self.steer_override = abs(cp.vl["MDPS12"]['CR_Mdps_StrColTq']) > STEER_THRESHOLD
     self.steer_state = cp.vl["MDPS12"]['CF_Mdps_ToiActive'] #0 NOT ACTIVE, 1 ACTIVE
     self.steer_error = cp.vl["MDPS12"]['CF_Mdps_ToiUnavail']
     self.brake_error = 0
-    self.steer_torque_driver = cp.vl["MDPS11"]['CR_Mdps_DrvTq']
+    self.steer_torque_driver = cp.vl["MDPS12"]['CR_Mdps_StrColTq']
```
geohot: Why did this change? Shouldn't this be only the driver torque, not the total torque on the column?

selfdrive/car/hyundai/carstate.py
```diff
@@ -145,6 +163,7 @@ def __init__(self, CP):
     self.left_blinker_flash = 0
     self.right_blinker_on = 0
     self.right_blinker_flash = 0
+    self.no_radar = self.CP.carFingerprint in FEATURES["non_scc"]
```
geohot: We won't upstream this, either a car uses stock adaptive cruise control or openpilot's, but no ACC isn't an option. It violates the fully on fully off nature of openpilot.

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -1,47 +1,101 @@
...
+              left_line, right_line, left_lane_depart, right_lane_depart):
 
     ### Steering Torque
     new_steer = actuators.steer * SteerLimitParams.STEER_MAX
     apply_steer = apply_std_steer_torque_limits(new_steer, self.apply_steer_last, CS.steer_torque_driver, SteerLimitParams)
     self.steer_rate_limited = new_steer != apply_steer
 
-    if not enabled:
+    # disable if steer angle reach 90 deg, otherwise mdps fault in some models
+    lkas_active = enabled and abs(CS.angle_steers) < 90.
```
geohot: Is there a better way to detect this fault? It should throw a steering unavailable error if this happens.

selfdrive/car/hyundai/values.py
```diff
@@ -57,33 +69,84 @@ class Buttons:
...
-  }
-  ],
+    67: 8, 127: 8, 304: 8, 320: 8, 339: 8, 356: 4, 544: 8, 593: 8, 608: 8, 688: 6, 764: 8, 809: 8, 854: 7, 870: 7, 871: 8, 872: 8, 897: 8, 902: 8, 903: 8, 905: 8, 909: 8, 916: 8, 1040: 8, 1042: 8, 1056: 8, 1057: 8, 1064: 8, 1078: 4, 1107: 5, 1136: 8, 1151: 6, 1155: 8, 1162: 8, 1164: 8, 1168: 7, 1170: 8, 1173: 8, 1180: 8, 1183: 8, 1186: 2, 1227: 8, 1265: 4, 1280: 1, 1287: 4, 1290: 8, 1292: 8, 1294: 8, 1312: 8, 1322: 8, 1345: 8, 1348: 8, 1363: 8, 1369: 8, 1371: 8, 1378: 8, 1384: 8, 1407: 8, 1414: 3, 1419: 8, 1427: 6, 1456: 4, 1470: 8, 1988: 8, 2000: 8, 2004: 8, 2008: 8, 2012: 8
+  }],
+  CAR.SANTA_FE_1: [{
```
geohot: What is a SANTA_FE_1 and how is it different from a SANTA_FE?

selfdrive/car/hyundai/carstate.py
xx979xx: Some cars do not have CR_Mdps_DrvTq signal, as it is for Smart Parking Assist System, also it is reverse (right is +) .
` SG_ CR_Mdps_DrvTq : 11|12@1+ (1.0,-2048.0) [-2048.0|2046.0] ""  SPAS`
The right signal is CR_Mdps_StrColTq which output identical pattern, it is the signals OEM LKAS listen to and match LKAS CR_Lkas_StrToqReq signal perfectly.
` SG_ CR_Mdps_StrColTq : 0|11@1+ (0.0078125,-8.0) [-8.0|7.9921875] "Nm"  LDWS_LKAS`
There is another signal for total torque which is CR_Mdps_OutTq
` SG_ CR_Mdps_OutTq : 52|12@1+ (0.1,-204.8) [-204.8|204.7] ""  ESC,LDWS_LKAS`

selfdrive/car/hyundai/carstate.py
xx979xx: Ok I will remove all non_scc cars then

selfdrive/car/hyundai/carcontroller.py
xx979xx: it is a hard fault. if it happened, MDPS will not accept any cmd until car restart
I can do it based on car model as I did with Genesis:
`    # Fix for sharp turns mdps fault and Genesis hard fault at low speed
    if CS.v_ego < 16.7 and self.car_fingerprint == CAR.GENESIS and not CS.mdps_bus:
      lkas_active = 0`

selfdrive/car/hyundai/carstate.py
xx979xx: removed

selfdrive/car/hyundai/values.py
xx979xx: removed

selfdrive/car/hyundai/carstate.py
xx979xx: please see discussion in [panda PR](https://github.com/commaai/panda/pull/427)

selfdrive/car/hyundai/carcontroller.py
geohot: Better, but I still don't like the hard fault at 90. At 90, there's probably a lot of torque on the wheel, and this will just drop it. Can the fault be detected?

selfdrive/car/hyundai/carstate.py
```diff
@@ -187,64 +195,78 @@ def update(self, cp, cp_cam):
...
+    self.steer_torque_driver = cp.vl["MDPS12"]['CR_Mdps_StrColTq']
     self.steer_torque_motor = cp.vl["MDPS12"]['CR_Mdps_OutTq']
     self.stopped = cp.vl["SCC11"]['SCCInfoDisplay'] == 4.
+    self.lead_distance = cp.vl["SCC11"]['ACC_ObjDist']
 
     self.user_brake = 0
 
     self.brake_pressed = cp.vl["TCS13"]['DriverBraking']
-    self.brake_lights = bool(self.brake_pressed)
+    self.brake_lights = bool(cp.vl["TCS13"]['BrakeLight'] or self.brake_pressed)
```
geohot: Ahh, so this used to be wrong...can it just be cp.vl["TCS13"]['BrakeLight']?

geohot:
Once we get panda merged, this all looks pretty straightforward.

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -1,47 +1,101 @@
...
-      can_sends.append(create_clu11(self.packer, CS.clu11, Buttons.RES_ACCEL))
+      can_sends.append(create_clu11(self.packer, CS.clu11, Buttons.CANCEL, clu11_cnt))
+
+    if CS.stopped:
+      # run only first time when the car stopped
+      if self.last_lead_distance == 0:
+        # get the lead distance from the Radar
+        self.last_lead_distance = CS.lead_distance
+        self.resume_cnt = 0
+      # when lead car starts moving, create 6 RES msgs
```
geohot: Confirming this doesn't actually raise the speed?

selfdrive/car/hyundai/carstate.py
```diff
@@ -45,63 +45,72 @@ def get_can_parser(CP):
...
-    ("SAS_Angle", "SAS11", 0),
-    ("SAS_Speed", "SAS11", 0),
   ]
 
   checks = [
     # address, frequency
     ("MDPS12", 50),
-    ("MDPS11", 100),
     ("TCS15", 10),
-    ("TCS13", 50),
```
geohot: There are still things depending on this signal (TCS13). You can't remove the check

selfdrive/car/hyundai/carstate.py
```diff
@@ -45,63 +45,72 @@ def get_can_parser(CP):
...
     ("CLU11", 50),
     ("ESP12", 100),
-    ("EMS12", 100),
```
geohot: Same for this one

selfdrive/car/hyundai/carstate.py
xx979xx: this signal is only for ACC brake light, are you sure?

selfdrive/car/hyundai/carcontroller.py
xx979xx: yes since the condition `if CS.stopped` only true when SCC not active and the driver prompted to press Res button (after 3 sec)
only press resume when lead distance changes, spamming the button all time is annoying as  prompted message flashes with audible alert

selfdrive/car/hyundai/carstate.py
xx979xx: this is HKG stuff, some car do not have this signal. I can go back to see what car was 
it is not my car otherwise, I would have tried to find the alternative msg

selfdrive/car/hyundai/carstate.py
xx979xx: Same as prev

selfdrive/car/hyundai/carstate.py
geohot: But the TCS13 signal is used above to determine things. See line 166 for example. If you want car state to depend on the signal, you need to check to make sure it's valid, otherwise car state is wrong!

selfdrive/car/hyundai/carstate.py
geohot: See line 219

selfdrive/car/hyundai/carstate.py
xx979xx: hyundai HYUNDAI SANTA FE without SCC does not have this message
I will revert this change and this car will no longer supported

selfdrive/car/hyundai/carstate.py
xx979xx: these cars do not have this message: KIA OPTIMA HYBRID and HYUNDAI IONIQ ELECTRIC /HYBRID. which make sense, they do not have engine ECU signal.
also, they are use the new gear signal
we can do something like this:
`  if CP.carFingerprint in FEATURES["use_elect_gears"]:
    checks.remove(("EMS12", 100))`

geohot: This isn't the problem, the problem is that the message is being used in an ungated way. What I'm hearing here is that those cars don't have a complete carstate.

selfdrive/car/hyundai/carstate.py
geohot: We don't support cars without SCC

selfdrive/car/hyundai/carcontroller.py
jpancotti: If there is a limitation at 90, why not have a warning and prevent additional torque from being requested as angle approaches 90? You could ramp down torque request in the high 80s. Disengagement is not an option

selfdrive/car/hyundai/carstate.py
xx979xx: fixed

selfdrive/car/hyundai/carstate.py
xx979xx: glad you mention that since all new signals have not added to checks
all signals and checks are gated or removed

selfdrive/car/hyundai/carcontroller.py
SiGmAX666: It feels like ramping is probably the best option, I don't think we have a specific CAN signal discovered for it, otherwise I'd assume it would be in use.

AskAlice:
🏃‍♀ ✨  gimmie a few weeks lmao

README.md
```diff
@@ -84,10 +86,15 @@ Supported Cars
...
 | Hyundai   | Elantra 2017-19<sup>1</sup>   | SCC + LKAS        | Stock            | 19mph              | 34mph        |
+| Hyundai   | Elantra GT 2017-19<sup>1</sup>| All               | Stock            | 0mph               | 30mph        |
 | Hyundai   | Genesis 2018<sup>1</sup>      | All               | Stock            | 19mph              | 34mph        |
+| Hyundai   | Ioniq 2017<sup>1</sup>        | All               | Stock            | 0mph               | 34mph        |
+| Hyundai   | Kona 2017-19<sup>1</sup>      | LDWS              | Stock            | 22mph              | 0mph         |
 | Hyundai   | Santa Fe 2019<sup>1</sup>     | All               | Stock            | 0mph               | 0mph         |
 | Jeep      | Grand Cherokee 2016-18        | Adaptive Cruise   | Stock            | 0mph               | 9mph         |
 | Jeep      | Grand Cherokee 2019           | Adaptive Cruise   | Stock            | 0mph               | 39mph        |
+| Kia       | Forte 2018<sup>1</sup>        | LKAS              | Stock            | 0mph               | 0mph         |
```
geohot: Is SCC default on this?

selfdrive/car/hyundai/carcontroller.py
geohot: Agree with @jpancotti

README.md
```diff
@@ -84,10 +86,15 @@ Supported Cars
 | Honda     | Pilot 2019                    | All               | openpilot        | 25mph<sup>5</sup>  | 12mph        |
 | Honda     | Ridgeline 2017-19             | Honda Sensing     | openpilot        | 25mph<sup>5</sup>  | 12mph        |
 | Hyundai   | Elantra 2017-19<sup>1</sup>   | SCC + LKAS        | Stock            | 19mph              | 34mph        |
+| Hyundai   | Elantra GT 2017-19<sup>1</sup>| All               | Stock            | 0mph               | 30mph        |
 | Hyundai   | Genesis 2018<sup>1</sup>      | All               | Stock            | 19mph              | 34mph        |
+| Hyundai   | Ioniq 2017<sup>1</sup>        | All               | Stock            | 0mph               | 34mph        |
```
SiGmAX666: Base Ioniq in the USA doesn't have SCC, requires Limited trim. I think all have LKAS. Probably should be changed to `SCC + LKAS`

Also need to add the line for `Ioniq 2019 Electric`, with `SCC + LKAS`

README.md
```diff
@@ -84,10 +86,15 @@ Supported Cars
 | Honda     | Pilot 2019                    | All               | openpilot        | 25mph<sup>5</sup>  | 12mph        |
 | Honda     | Ridgeline 2017-19             | Honda Sensing     | openpilot        | 25mph<sup>5</sup>  | 12mph        |
 | Hyundai   | Elantra 2017-19<sup>1</sup>   | SCC + LKAS        | Stock            | 19mph              | 34mph        |
+| Hyundai   | Elantra GT 2017-19<sup>1</sup>| All               | Stock            | 0mph               | 30mph        |
```
SiGmAX666: Base GT in the USA doesn't have SCC, requires Tech Package. I think all have LKAS. Probably should be changed to `SCC + LKAS`

README.md
SiGmAX666: The 2020 base model can have SCC, but it's a package. I can't imagine it was default on a 2018 if it isn't on a 2020. I think it needs to be updated to `SCC + LKAS`.

SiGmAX666 (changes requested):
Readme model clarity

README.md
AskAlice: I have a 2019 kia forte that I bought used, it was was originally a rental car, and it does not have SCC. So I would say it is not 'default' unless there is a particular trim for rental cars that i am unaware of

AskAlice: Another thing to note with this is that the 2018 kia forte and the 2019 kia forte have completely different specs/tuning (ie steering ratio, mass, etc), but the fingerprints are identical

selfdrive/car/hyundai/carcontroller.py
xx979xx: > If there is a limitation at 90, why not have a warning and prevent additional torque from being requested as angle approaches 90? You could ramp down torque request in the high 80s to prevent going over. Disengagement is not an option

@geohot @jpancotti, the issue is not about openpilot requesting additional torque that exceed 90 deg as MDPS firmware will deal with that and prevent any off-limit steering .
**The issue only arise when the driver override and steering angle exceed the limit and Openpilot keep sending steer request.** 
any changes to steer_req will have no effect whatsoever

README.md
xx979xx: merge

README.md
xx979xx: merge

selfdrive/car/hyundai/carcontroller.py
xx979xx: well, that won't work, as it will mdps will fault as soon as driver let go steering wheel while steering angle above the limit

xx979xx (commented):
everything is ok for me. still need a confirmation about `CS.brake_lights ` [here](https://github.com/commaai/openpilot/pull/970#discussion_r370715245) and LKAS steering angle limit [here](https://github.com/commaai/openpilot/pull/970#discussion_r379714128)

flamethrower225:
Any update on this?

AskAlice:
I've noticed that openpilot does not disengage when the doors open.
https://github.com/xx979xx/openpilot/blob/332cb82886d22a0dac1dad45117a15d137cf720f/selfdrive/car/hyundai/carstate.py#L302

PR #972, opened 2020-01-19
Fix for steer unavailable for RAV4H and Corolla

arne182:
https://github.com/commaai/openpilot/pull/856

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -134,7 +134,9 @@ class CarController():
       self.last_fault_frame = frame
 
     # Cut steering for 2s after fault
-    if (frame - self.last_fault_frame < 200) or abs(CS.angle_steers) > 100 or abs(CS.angle_steers_rate) > 100:
+    if (frame - self.last_fault_frame < 200) or \
+       (abs(CS.angle_steers_rate) > 100 and CS.CP.carFingerprint in (CAR.RAV4H, CAR.COROLLA)) or \
+       (CS.CP.carFingerprint == CAR.RAV4H and abs(CS.angle_steers) > 100):
       new_steer = 0
       apply_steer_req = 0
     else:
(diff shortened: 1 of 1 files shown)


geohot:
I don't understand the implications of this change, can it be written in a more minimal way? Like does the place where apply_toyota_steer_torque_limits is called need to change?

ShaneSmiskol:
I don't think we need the steer angle part, I've tested the code with just the rate limiter and it's worked wonderfully

arne182:
Yes, limits should be applied after the fault not before. Applying it before causing steering faults as the torque rate limit is not held. The only reason to jump down to 0 in one frame is if it is not enabled anymore.

arne182:
Maybe for the Corolla it is only rate that needs to be limited but I (Rav4h) definitely have situations where the rate is much lower but the steering angle goes over 100 and then faults.

geohot:
What does the change look like with "just the rate limiter"?

arne182:
how do I get the ./update_refs.py to run so that this check does not fail.

pd0wm:
We can take care of that before we merge the PR.

selfdrive/car/toyota/carcontroller.py
```diff
@@ -128,20 +128,27 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, hud_alert,
...
       self.last_fault_frame = frame
 
     # Cut steering for 2s after fault
-    if not enabled or (frame - self.last_fault_frame < 200):
-      apply_steer = 0
+    if (frame - self.last_fault_frame < 200) or \
+       (abs(CS.angle_steers_rate) > 100 and CS.CP.carFingerprint in (CAR.RAV4H, CAR.COROLLA)) or \
+       (CS.CP.carFingerprint == CAR.RAV4H and abs(CS.angle_steers) > 100):
+      new_steer = 0
       apply_steer_req = 0
```
pd0wm: Do you want the steer_request bit to be zero while you're winding down?

arne182: I have tried it without switching it off and then there are still some occurrences where the steering faults. Having it switch to 0 steering request is the right solution.

PR #974, opened 2020-01-20
Fix blinkers in Subaru CarState

michaelhonan:
**FROM OLD PR:**
Testing the unreleased ALC feature revealed a flaw in the CarState values for the blinkers. The values for the blinkersare mapped to the Dashlights in the DBC, and this means that when the blinkers are activated, they flicker between True and False at the same rate that the blinkers "blink".

This issue causes ALC to act strangely, having the lines disappear and reappear in accordance to the value of the blinkers in CarState.

My solution uses a counter that counts up to a certain threshold (50 atm, little longer than the blink rate, which is ~32) every frame with the counter resetting to 0 after every blinker light state change. After the threshold is hit, it then sets the indicator status to False.

A few other cars may require a similar change if they do not have a constant value when the blinkers are activated.

**NEW NOTES**
A few of us (Bugsy, C-ton, and myself) in the #subaru channel have come to the conclusion that there isn't a stalk signal on CAN for the indicators. Meaning that Subaru only has the indicators available via the oscillating signal.

It's presumed that Subaru do a similar method of "timing" the indicator oscillations because you can signal, cancel, cross a lane line and you don’t get a Lane Departure Warning even though the stalk is in the resting position.

### DIFF
--- a/selfdrive/car/subaru/carstate.py
+++ b/selfdrive/car/subaru/carstate.py
@@ -88,10 +88,17 @@ class CarState():
     self.CP = CP
 
     self.car_fingerprint = CP.carFingerprint
+
     self.left_blinker_on = False
     self.prev_left_blinker_on = False
+    self.prev_left_blinker_light_on = False
+    self.left_blinker_interval = 0
+
     self.right_blinker_on = False
     self.prev_right_blinker_on = False
+    self.prev_right_blinker_light_on = False
+    self.right_blinker_interval = 0
+
     self.steer_torque_driver = 0
     self.steer_not_allowed = False
     self.main_on = False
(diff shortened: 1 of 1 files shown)


geohot:
I'd want to be totally sure there's no simpler signal before merging this. Are you sure that the stock system times it and there isn't some status bit hidden somewhere?

pd0wm:
If there is no better signal, this code can also be much shorter. I see a lot of duplicate lines. I feel like this could be added like:

```python
self.left_blinker_cnt = 50 if cp.vl["Dashlights"]['LEFT_BLINKER'] else max(self.left_blinker_cnt - 1, 0)
self.left_blinker_on = self.left_blinker_cnt > 0
```

mchonan1996:
@G : Going to go over the drives I have again and make 100% sure there's nothing that could act as a timing bit or something.

@p : Thanks for the feedback, I'll definitely swap it over to use your snippet. Appreciate it.

Issue #946, opened 2020-01-14
Lower STEER_MAX for Honda vehicles with dead zone above LKAS max request value


gregjhogan:
Civic and CR-V steering request values above what stock LKAS sends do not result in additional torque.  This causes openpilot steering control to have a dead zone that in certain situations decreasing steering performance.

Since it seems likely this would apply to most or all Honda vehicles.  We need to collect stock drives for Honda vehicles, compare estimated torque output of openpilot max request to stock max request, and lower STEER_MAX/kp/ki/kf for each vehicle accordingly.

Known stock LKAS max request values:
* `CIVIC` = 0xA00
* `CIVIC_BOSCH` = 0xA00
* `CRV_5G` = 0xF00

gregjhogan:
It turns out that for some vehicles the LKAS camera max request value is not the max torque (e.g. 2017 CR-V), so you can't just look at the max request value of the stock camera to determine where the dead zone starts.  You have to look at the estimated torque output (but this is very noisy so not very accurate) or ideally, the firmware.

PR #976, opened 2020-01-20
Increase STEER_MAX with modified EPS firmware and Civic Bosch

cfranyota:
If eps fw is modded then add civic_bosch


energee:
I'm not sure if this is what you think it is, have you changed the max_steer value allowed on 0xE4?

pd0wm:
Everything above 0x1000 ( or maybe even 0xA00, https://github.com/commaai/openpilot/issues/946). Is a deadzone. This will not give you more torque.

PR #977, opened 2020-01-20
speed up car model test

adeebshihadeh:
Running locally with logs cached, this shaves off about ~3.5m and now runs in ~3m. Should see similar time savings in CI.

The sleep after it starts the processes unnecessary because controlsd [blocks](https://github.com/commaai/openpilot/blob/master/selfdrive/controls/controlsd.py#L497) on can anyway. The sleep after everything's killed is also unnecessary because manager ensures all the processes are killed. Most of the time savings are from the 4s saved each time a route's tested.

### DIFF
--- a/selfdrive/test/test_car_models.py
+++ b/selfdrive/test/test_car_models.py
@@ -27,19 +27,17 @@ from selfdrive.car.mock.values import CAR as MOCK
 os.environ['NOCRASH'] = '1'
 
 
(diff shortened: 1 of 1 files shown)


quillford:
Looks like almost 5 min are saved in Travis

geohot:
I'm not sure this has the same failure behavior, it's more than the processes not starting that triggers a failure. Like controls_state_result

PR #985, opened 2020-01-21
[Memory leak]uniq_ptr for tmsg

gaurav1086:
Use uniq_ptr for tmsg


geohot:
Did you actually get this code running? If so, how?

We are working on making it compatible with the public API, but this isn't done yet.

gaurav1086:
@geohot not yet. I was skimming over the code and came across that.

PR #986, opened 2020-01-21
GitHub Actions for CI

adeebshihadeh:
Improvements over Travis:
* Run tests in parallel (without external docker registry)
* Debug artifacts from process replay and longitudinal test
* [Nicer interface](https://github.com/quillford/openpilot/actions) and output isn't one giant block of text
* Possible docker layer caching built-in, still looking into this
* Seems to run automatically as long as the config file is in the repo, unlike Travis, so forks would automatically have CI

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -46,14 +46,14 @@ jobs:
         run: docker load -i tmppilot.tar.gz/tmppilot.tar.gz
       - name: Run unit tests
         run: |
-          docker run --shm-size 1G --rm tmppilot /bin/sh -c "python -m unittest discover common"
-          docker run --shm-size 1G --rm tmppilot /bin/sh -c "python -m unittest discover opendbc/can"
-          docker run --shm-size 1G --rm tmppilot /bin/sh -c "python -m unittest discover selfdrive/boardd"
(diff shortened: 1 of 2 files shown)


pd0wm:
Cool! How does it save the docker container after the build to run stuff in parallel?

I was looking into moving everything to azure-pipelines, but that still requires a registry to push things to.

quillford:
Instead of pushing to a registry, it’s just building the image fresh each time and uploading it as an artifact for the test stages to download.

pd0wm:
Ah, I see, you're using `docker save` and gzipping them.

quillford:
Seems like docker layer caching isn't implemented in GitHub Actions yet, but it can still be done with a docker registry. Other than that, this should be good to merge.

pd0wm:
- Can we remove the travis file? We don't need that anymore.
 - The caching solution you implemented sounds fine.
 - Where can I see the test outputs, "checks" doesn't show anything for me. Do I need to enable it on the repo?

quillford:
I think it needs to be merged in to a branch on `commaai/openpilot` before it'll run.

pd0wm:
Let's see what happens!

Thanks for this PR :)

PR #987, opened 2020-01-21
Subsock no longer exsists

arne182:
use the new SubSocket as SubSock has been removed


arne182:
Thanks, worth keeping compatibility for the maps code.

PR #762, opened 2019-07-26
Bounty: Mazda

Jafaral:
~There is still work to be done but I think a PR gives more visibility to this port with more potential help from members of the community~. I have been working on this port  since late last year with help from a few members in the community (extra thanks to @bugsy924).

~~Unfortunately, Mazda has the same steer lockout issue as ford~~ 
[update Nov 2019: The lockout in Mazda is different from Ford. Ford lockout happens at regular intervals and there is no solution for that. Mazda on the other hand just forces the driver to keep hands on the steering all the time with a grace period of only 5 seconds with no driver torque. If driver torque is detected, the LKAS system functions normally with no issue. Applying a weight to the steering fixes the issue completely. ]

If the hands are off the steering wheel for more than 5 seconds you get a warning and the steer commands eventually get ignored. So far I have not found a solution for that,  maybe we will have better luck with new models next year.   My solution until now is to use a weight on the steering wheel. 

Torque is modest but enough to handle most curves on a freeway (up to 8-10 degrees). I've driven 10s of hours so far with no major issues. A few users in the community have been using my port and they reported positive results.

~I need to work more on the safety side of things, this will be done in the next few weeks. I have put all of my energy so far to find a programmatic solution for the lock out issue.~ [DONE]

The Mazda giraffe is available here:
https://github.com/Jafaral/neo/tree/mazda/giraffe/mazda

~I will submit a PR against the neo repo, but I need to move the lateral control to CAN 2, in the current design that is on CAN 1.~

[Update Nov 2019] : The giraffe now has lateral control on CAN 2, although it doesn't matter with the new black panda.  DBC files and panda safety were all upstreamed!

Comments are welcome!
Jafar/Jacar

PR #988, opened 2020-01-21
Bounty: Mazda

Jafaral:
Migrating PR #762 to master


geohot:
Does this still have the 10 second steering lockout?

bugsy924:
Yes, alleviated with a weight on the steering

On Jan 24, 2020 6:44 AM, George Hotz <notifications@github.com> wrote:

Does this still have the 10 second steering lockout?

Jafaral:
@geohot yes, and it is 5 seconds actually.  It is different from the ford lockout as I explained in #762. As @Bugsy#7983 the problem is mitigated with a weight on the steering wheel.  In my case I use a steering wheel cover with some weight hidden along the curve.  Didn't have issues since.

geohot:
So this doesn't work for something we can ship. If we want to merge this, we'd have to detect and alert the user whenever they are locked out. Can you add an alert for this?

Though I hear good news for 2019 Mazda 3, that there's no lockout?

Jafaral:
Fair, I can easily add an alert and maybe issue a soft disable, would that work ?

correct, Mazda3 doesn't seem to have the issue, but it still needs some work porting. I have been focusing on getting the current Gen Mazda cars officially supported before I put more effort into Mazda3.

I will the lockout alert and update the PR.

Jafaral:
I have worked on adding alerts for steer lockout and low speed and working on improving various aspects of the port. Will be updating this PR in the new few days.

Jafaral:
@geohot I have pushed a few updates to this PR with a few improvements and also to address several situations including  your comment above. Notable new changes:
     
      * Do not engage if LKAS is not available at low speeds. This happens anytime  the speed dips below 28mph and doesn't get enabled until the speed hits 32mpg. If the user tried to engage they get "OpenPilot Unavailable, Speed Two Low" 

      * If OP is engaged and the speeds dips below 28mpg where LKAs becomes unavailable the user see a continuous "TAKE CONTROL: Steer Unavailable Below 28MPH"  warning. OP doesn't disengage in this case but the warning will be shown as long as the LKAs is not available. Once the speed requirement is met the warning goes away and OP resumes normal operation. 

    * If the user have their hands off the steering wheel and the steer lockout timeout is triggered,  a Warning is issued "TAKE CONTROL: Steering is Temporarily Unavailable". We can create a specific warning for cars with lockout timeout if we want to make it more obvious way it is not available. The stock car system itself also issues a HUD warning and an audible alert in that situation, so the user is very aware of what  is going on.

   * Disengage on gas press which is required by Comma AI safety guidelines. break/seatbelt/door disengagement were already there. 

   * Refactored a lot of the car params/constants and moved them to values.py

   * Improved the handling of cruise buttons presses to give a better user feedback
   
   * Added CX-2019 GT fingerprint as a couple of new Mazda users have that car and use it with Comma 2

   The only remaining issue is that sometimes I get steer limit warnings even when driving on straight  lines. More tuning is needed for that.

Please let me know if you have any comments or questions!

Jafaral:
I did write a guide for Mazda user covering all aspects of this port.  I will update the guide with screen shots of the various warnings as examples when I get a chance:

https://medium.com/@to.jafar/how-to-setup-openpilot-on-mazda-3eb54c62fdc5

Jafaral:
I update the Mazda guide above with example warnings. Will add one for hands off warning when I get one.

geohot:
The expectation from every car in openpilot is that it be able perform on the highway without many disengagements ( referenced in #1060 ), and unfortunately with a 10 second lockout that's unlikely. I want to be clear on this policy going forward, sorry for being indecisive.

If some cars don't have the lockout, we can upstream them. I still think the warning is a good idea, and we'll keep this PR open in hopes future Mazda's don't have the lockout, or a way is found to bypass it.

Jafaral:
@geohot, I understand.  I'd like to point out that on the highway I hardly get any disengagement for long stretches, 10+ minutes, with a neatly hidden weight under a steering wheel cover. All Mazda users are using this trick.

geohot:
Let's merge this as a dashcam only port for now.

Jafaral:
@geohot Thanks for the update!

@pd0wm aside from updating the PR to address the merge conflicts, please let me know what other things you want added/dropped to make it work in dashcam mode only.

geohot:
Can you fix test car models?

pd0wm:
Just set `    ret.dashcamOnly = True` in interface.py. (https://github.com/commaai/openpilot/blob/master/selfdrive/car/ford/interface.py#L22). Also make sure the route gets marked as such in test_car_models. Just copy what ford does.

Also there have been some refactors lately in carstate/carinterface. Can you make sure you follow the same conventionts:
 - Build carState struct in carstate.py, not in interface. That save a lot of copies
 - Use the `create_common_events` method to avoid some duplication in the regular events like door, seatbelt, etc.
 - Use the kalman filter from CarStateBase
 - I see the gear is not parsed. Make sure it disengages when going in reverse.

Jafaral:
@geohot  @pd0wm will do!  I will update the PR and test in the next few days.

Jafaral:
Finally got some time to move a bit on this, still more work to do. Will update when I get more time to work on it.

Jafaral:
@geohot
  test is working now

 @pd0wm 

  * Code was refactored following the new convention.
  * Gear is parsed
  * gas/brake events including pressure are all added 
  * Fingerprints were added to cover many of the Mazda cars on the road now
  * refactored the code to prepare for future updates for the new gen Mazda  such as Mazda 3

Remaining items:
    * Force dashmode: I'm keeping this as the last step after everything else is working/approved
    * Disengage on gas-press , this was working before, missing something in my refactoring
    * Fix process_replay test, the drive is synced by you but I still get this error:
```AssertionError: Failed to download /tmp/openpilot/selfdrive/test/process_replay/32a319f057902bb3|2020-04-27--15-18-58--2_controlsd_b33259e2dae9852c58b0507dbb98fb01d8b884b8.bz2```

If you have any comments please let me know.

File: selfdrive/car/mazda/carstate.py [hunk 1]
```diff
@@ -0,0 +1,214 @@
...
+    if gear >= 1 & gear <= 5:
+      ret.gearShifter = GearShifter.drive
+    elif gear == 0:
+      ret.gearShifter = GearShifter.park
+    elif gear == 6:
+      ret.gearShifter = GearShifter.reverse
+    elif gear == 7:
+      ret.gearShifter = GearShifter.neutral
+    else:
+      ret.gearShifter = GearShifter.unknown
```
quillford, line 52: Why the shift for the gear value? Also seems like a pretty big range for drive. If you define these values in the DBC, you can use [parse_gear_shifter](https://github.com/commaai/openpilot/blob/master/selfdrive/car/interfaces.py#L158) like the other cars.

selfdrive/car/mazda/interface.py
```diff
@@ -0,0 +1,114 @@
...
+
+    events = self.create_common_events(ret)
+
+    if ret.cruiseState.enabled and self.CS.lkas_speed_lock:
+      self.low_speed_alert = True
+    else:
+      self.low_speed_alert = False
+
+    # events
+    events = self.create_common_events(ret)
```
quillford, line 93: this is already called in line 85

selfdrive/car/mazda/values.py
```diff
@@ -0,0 +1,69 @@
...
+#  ],
+
+}
+
+class ECU:
+  CAM = 0
+
+ECU_FINGERPRINT = {
+  ECU.CAM: [579],   # steer torque cmd
+}
```
quillford, line 63: don't need to make a new class. you can use car.CarParams.Ecu from cereal like this https://github.com/commaai/openpilot/blob/master/selfdrive/car/toyota/values.py#L64

selfdrive/car/mazda/interface.py
```diff
@@ -0,0 +1,114 @@
...
+
+    if self.CS.low_speed_lockout:
+      events.append(create_event('speedTooLow', [ET.NO_ENTRY]))
+      if ret.cruiseState.enabled:
+        ret.cruiseState.enabled = False
+
+    if self.low_speed_alert:
+      events.append(create_event('belowSteerSpeed', [ET.WARNING]))
+
+    if self.CS.steer_lkas.handsoff:
```
quillford, line 103: If you use `self.steer_warning` in carstate.py, `create_common_events` will create the event for you.

selfdrive/car/mazda/mazdacan.py
```diff
@@ -0,0 +1,57 @@
...
+  if ahi == 1:
+    csum = csum - 15
+
+  if csum < 0:
+    if csum < -256:
+      csum = csum + 512
+    else:
+      csum = csum + 256
+
+  csum = csum % 256
```
quillford, line 41: If the checksum isn't specific to this message, maybe give it its own function

quillford:
@Jafaral process replay needs both the drive synced and the reference logs it compares against. That error means the reference logs are missing from the CI bucket

selfdrive/car/mazda/carstate.py
Jafaral: Shift because I found that the low bit is not used, i.e shouldn't be part of the signal. I already opened a PR to fix that (https://github.com/commaai/opendbc/pull/254) once that is merged the shift can be dropped. 1 through 5 are gear positions for speeds  1 to 5.  I did see this that function, do I fill in letters instead and leave the mapping to the function?

selfdrive/car/mazda/interface.py
Jafaral: good catch!

selfdrive/car/mazda/carstate.py
quillford: You have to define the gear values in the DBC like this https://github.com/commaai/opendbc/blob/master/toyota_prius_2017_pt_generated.dbc#L401

selfdrive/car/mazda/mazdacan.py
Jafaral: I doubt the checksum will work in other messages. It is all over the place and  specific to the signals of this message!

selfdrive/car/mazda/interface.py
Jafaral: noted!

quillford: sorry, I actually just changed this. set carState.steerWarning instead  https://github.com/commaai/openpilot/pull/1444

Jafaral:
@quillford The drive 32a319f057902bb3|2020-04-27--15-18-58 is used with test_car_models as well and I was getting an error there until @pd0wm synced it. After the sync, the test_Cat_models went away but the the error for the segment within that drive here persist, not sure what that means.

selfdrive/car/mazda/values.py
Jafaral: fixed!

selfdrive/car/mazda/carstate.py
Jafaral: Do I edit the file manually or can this be done via cabana?

selfdrive/car/mazda/interface.py
Jafaral: double noted!

selfdrive/car/mazda/carstate.py
quillford: I don't think cabana can do this. I'd just do it manually

Jafaral: here is what I came up with: 
```
VAL_ 552 GEAR 0 "P" 1 "D" 2 "D" 3 "D" 4 "D" 5 "D" 6 "R" 7 "N";
```
There might be a better signal to  track the "stick" state, but I haven't found that one if it does exist.

Jafaral: Does this work?

quillford: Yeah, that should work. What do you mean by stick state? If you mean the actual gear position, I'd just put it in a comment

Jafaral:
@quillford Aside from parsing the gear, all other comments are addressed. Will test drive this tomorrow. Will wait for the dbc updates to refactor gear parsing.

Thanks for taking the  time to look at this!

selfdrive/car/mazda/carcontroller.py
```diff
@@ -0,0 +1,35 @@
+from selfdrive.car.mazda import mazdacan
+from selfdrive.car.mazda.values import DBC, SteerLimitParams
+from opendbc.can.packer import CANPacker
+from selfdrive.car import apply_std_steer_torque_limits
+
+class CarController():
+  def __init__(self, dbc_name, CP, VM):
+    self.steer_idx = 0
```
quillford, line 8: unused

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,219 @@
...
+
+    if ret.cruiseState.enabled:
+      ret.cruiseState.speed = self.cruise_speed
+      if self.lkas_speed_lock:
+        ret.steerWarning = True
+    else:
+      ret.cruiseState.speed = 0
+
+
+    self.brake_error = False
```
quillford, line 123: unused

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,217 @@
...
+    # On if no driver torque the last 5 seconds
+    self.steer_lkas.handsoff = cp.vl["STEER_RATE"]['HANDS_OFF_5_SECONDS']
+
+    # LKAS is enabled at 50kph going up and disabled at 45kph going down
+    if self.speed_kph > LKAS_LIMITS.ENABLE_SPEED and self.low_speed_lockout:
+      self.low_speed_lockout = False
+    elif self.speed_kph < LKAS_LIMITS.DISABLE_SPEED and not self.low_speed_lockout:
+      self.low_speed_lockout = True
+
+    if (self.low_speed_lockout or self.steer_lkas.block) and self.speed_kph < LKAS_LIMITS.DISABLE_SPEED:
```
quillford, line 88: self.steer_lkas block can just be local to the function. don't see it used anywhere else

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,217 @@
...
+from selfdrive.car.interfaces import CarStateBase
+from selfdrive.car.mazda.values import DBC, LKAS_LIMITS, CAR
+
+GearShifter = car.CarState.GearShifter
+
+class STEER_LKAS():
+  def __init__(self):
+    self.block = 1
+    self.track = 1
+    self.handsoff = 0
```
quillford, line 13: doesn't seem like this is needed. block should be local to CS.update, tack is unused, and handsoff will be replaced by steerWarning

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,217 @@
...
+      ret.gearShifter = GearShifter.unknown
+
+    ret.leftBlinker = cp.vl["BLINK_INFO"]['LEFT_BLINK'] == 1
+    ret.rightBlinker = cp.vl["BLINK_INFO"]['RIGHT_BLINK'] == 1
+
+    ret.steeringAngle =  cp.vl["STEER"]['STEER_ANGLE']
+    ret.steeringTorque = cp.vl["STEER_TORQUE"]['STEER_TORQUE_SENSOR']
+    ret.steeringPressed = abs(ret.steeringTorque) > LKAS_LIMITS.STEER_THRESHOLD
+
+    self.steer_torque_motor = cp.vl["STEER_TORQUE"]['STEER_TORQUE_MOTOR']
```
quillford, line 61: `ret.steeringTorqueEps`

selfdrive/car/mazda/interface.py
```diff
@@ -0,0 +1,103 @@
...
+    ret = self.CS.update(self.cp, self.cp_cam)
+    ret.canValid = self.cp.can_valid and self.cp_cam.can_valid
+
+    # TODO: button presses
+    ret.buttonEvents = []
+
+    if ret.cruiseState.enabled and self.CS.lkas_speed_lock:
+      self.low_speed_alert = True
+    else:
+      self.low_speed_alert = False
```
quillford, line 82: unused

selfdrive/car/mazda/interface.py
```diff
@@ -0,0 +1,103 @@
...
+from selfdrive.car import STD_CARGO_KG, scale_rot_inertia, scale_tire_stiffness, gen_empty_fingerprint, is_ecu_disconnected
+from selfdrive.car.interfaces import CarInterfaceBase
+
+ButtonType = car.CarState.ButtonEvent.Type
+
+class CarInterface(CarInterfaceBase):
+  def __init__(self, CP, CarController, CarState):
+    super().__init__(CP, CarController, CarState)
+
+    self.low_speed_alert = False
```
quillford, line 16: `self.low_speed_alert` is unused. this can all be deleted

quillford:
@Jafaral I merged your opendbc PR, so you can fix the gear parsing now.

Before this gets merged, the mazda panda safety should have tests. You can use the [toyota safety test](https://github.com/commaai/panda/blob/master/tests/safety/test_toyota.py) as an example. At a minimum, it should include all the tests from `common.PandaSafetyTest`.

Jafaral:
Thanks @quillford , it also needs to be synced to the OP's git submodules before it actually works, right?

I will address your latest comment and prepare for gear parsing. Will do panda tests as well.

One of my changes last night caused controlsd crash this morning. I will track that down. Can  you please find out why  process_replay is crashing? That  should help me catch runtime errors before I roll them on to the car. I can provide a new segment ID if that fixes the issue.

Thanks Again!

Jafaral:
Just pushed a Dashcam only commit as requested to finalize this PR.

selfdrive/car/mazda/values.py
```diff
@@ -0,0 +1,69 @@
...
+  STEER_MAX = 600                 # max_steer 2048
+  STEER_STEP = 1                  # how often we update the steer cmd
+  STEER_DELTA_UP = 10             # torque increase per refresh
+  STEER_DELTA_DOWN = 20           # torque decrease per refresh
+  STEER_DRIVER_ALLOWANCE = 15     # allowed driver torque before start limiting
+  STEER_DRIVER_MULTIPLIER = 1     # weight driver torque heavily
+  STEER_DRIVER_FACTOR = 1         # from dbc
+
+class CAR:
+#  CX3 = "Mazda CX-3"
```
adeebshihadeh, line 18: Seems like these are supported, but just treated as a CX5. Either break them out into separate cars, or delete these commented lines.

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,194 @@
...
+      self.lkas_on = True
+    elif speed_kph < LKAS_LIMITS.DISABLE_SPEED and block:
+      self.lkas_on = False
+
+    # if any of the cruize buttons is pressed force state update
+    if any([cp.vl["CRZ_BTNS"]['RES'],
+                cp.vl["CRZ_BTNS"]['SET_P'],
+                cp.vl["CRZ_BTNS"]['SET_M']]):
+      self.acc_active = True
+      self.cruise_speed = ret.vEgoRaw
```
adeebshihadeh, line 76: There should be a signal for set speed in the can data. Every other car has this.

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,194 @@
...
+
+    ret.cruiseState.available = cp.vl["CRZ_CTRL"]['CRZ_ACTIVE'] == 1
+    if not ret.cruiseState.available:
+      self.acc_active = False
```
adeebshihadeh, line 80: doesn't need a condition

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,194 @@
...
+          self.acc_active = False
+          self.low_speed_lockout = True
+        else:
+          self.low_speed_alert = True
+      else:
+        self.low_speed_lockout = False
+        self.low_speed_alert = False
+
+      if handsoff:
+        ret.steerWarning = True
```
adeebshihadeh, line 97: doesn't need a condition

adeebshihadeh:
Can you add Mazda to the README? Thanks for all the work on this port, it's looking pretty good now.

selfdrive/car/mazda/carstate.py
Jafaral: Why not ?

selfdrive/car/mazda/carstate.py
Jafaral: why not ?

selfdrive/car/mazda/carstate.py
Jafaral: I'm sure there is! no long control, so it wasn't as important to dig out, will try to find it.

selfdrive/car/mazda/carstate.py
adeebshihadeh: `ret.steerWarning = handsoff`

selfdrive/car/mazda/values.py
Jafaral: They all have identical CAN buses and control, and even overlapping fingerprints. Many of these were added recently as more Mazda cars came online. I was planning on adding blocks for each car but hasn't been a priority so far

selfdrive/car/mazda/carstate.py
Jafaral: ah yes

selfdrive/car/mazda/carstate.py
Jafaral: Only want to sync if cruiseState.available is false. acc_active can be already false but available is true

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,190 @@
...
+    super().__init__(CP)
+
+    can_define = CANDefine(DBC[CP.carFingerprint]['pt'])
+    self.shifter_values = can_define.dv["GEAR"]['GEAR']
+
+    self.cruise_speed = 0
+    self.acc_active_last = False
+    self.low_speed_lockout = True
+    self.low_speed_alert = False
+    self.lkas_on = False
```
adeebshihadeh, line 19: Let's do the same rename we did in panda: `lkas_allowed`

selfdrive/car/mazda/carcontroller.py
```diff
@@ -0,0 +1,44 @@
...
+from opendbc.can.packer import CANPacker
+from selfdrive.car import apply_std_steer_torque_limits
+
+class CarController():
+  def __init__(self, dbc_name, CP, VM):
+    self.apply_steer_last = 0
+    self.packer = CANPacker(dbc_name)
+    self.steer_rate_limited = False
+
+  def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert,
```
adeebshihadeh, line 12: Remove unused args

Jafaral: I just added those last night hoping to build on top of them as we go!

selfdrive/car/mazda/values.py
```diff
@@ -0,0 +1,63 @@
...
+  SET_MINUS = 2
+  RESUME = 3
+  CANCEL = 4
+
+FINGERPRINTS = {
+  CAR.CX5: [
+    # CX-5 2017 GT
+    {
+      64: 8, 70: 8, 80: 8, 117: 8, 118: 8, 120: 8, 121: 8, 130: 8, 134: 8, 145: 8, 154: 8, 155: 8, 157: 8, 158: 8, 159: 8, 253: 8, 304: 8, 305: 8, 357: 8, 358: 8, 359: 8, 512: 8, 514: 8, 515: 8, 529: 8, 533: 8, 535: 8, 539: 8, 540: 8, 541: 8, 542: 8, 543: 8, 552: 8, 576: 8, 577: 8, 578: 8, 579: 8, 580: 8, 581: 8, 582: 8, 605: 8, 606: 8, 607: 8, 608: 8, 628: 8, 832: 8, 836: 8, 863: 8, 865: 8, 866: 8, 867: 8, 868: 8, 869: 8, 870: 8, 976: 8, 977: 8, 978: 8, 1034: 8, 1045: 8, 1056: 8, 1061: 8, 1067: 8, 1070: 8, 1078: 8, 1080: 8, 1085: 8, 1086: 8, 1088: 8, 1093: 8, 1108: 8, 1114: 8, 1115: 8, 1116: 8, 1139: 8, 1143: 8, 1147: 8, 1154: 8, 1157: 8, 1160: 8, 1163: 8, 1166: 8, 1177: 8, 1178: 8, 1179: 8, 1180: 8, 1183: 8, 1233: 8, 1236: 8, 1237: 8, 1238: 8, 1239: 8, 1241: 8, 1242: 8, 1243: 8, 1244: 8, 1264: 8, 1266: 8, 1267: 8, 1269: 8, 1270: 8, 1271: 8, 1272: 8, 1274: 8, 1275: 8, 1277: 8, 1278: 8, 1409: 8, 1416: 8, 1425: 8, 1430: 8, 1435: 8, 1440: 8, 1446: 8, 1456: 8, 1479: 8
+  },
```
adeebshihadeh, line 37: indent

selfdrive/car/mazda/carcontroller.py
adeebshihadeh: I'd prefer to only include what's needed now. Easy enough to add later.

selfdrive/car/mazda/carcontroller.py
```diff
@@ -0,0 +1,44 @@
...
+      apply_steer = apply_std_steer_torque_limits(new_steer, self.apply_steer_last,
+                                                  CS.out.steeringTorque, SteerLimitParams)
+      self.steer_rate_limited = new_steer != apply_steer
+
+      if CS.out.standstill and frame % 50 == 0:
+        # Mazda Stop and Go requires a RES button (or gas) press if the car stops more than 3 seconds
+        # Send Resume button at 2hz if we're engaged at standstill to support full stop and go!
+        # TODO: improve the resume trigger logic by looking at actual radar data
+        can_sends.append(mazdacan.create_cancel_acc(self.packer, CS.CP.carFingerprint, Buttons.RESUME))
+    else:
```
adeebshihadeh, line 32: This should be like the other cars where we cancel on `pcm_cancel_cmd`

selfdrive/car/mazda/carcontroller.py
```diff
@@ -0,0 +1,44 @@
...
+    can_sends = []
+
+    ### STEER ###
+
+    if enabled:
+      # calculate steer and also set limits due to driver torque
+      new_steer = int(round(actuators.steer * SteerLimitParams.STEER_MAX))
+      apply_steer = apply_std_steer_torque_limits(new_steer, self.apply_steer_last,
+                                                  CS.out.steeringTorque, SteerLimitParams)
+      self.steer_rate_limited = new_steer != apply_steer
```
adeebshihadeh, line 25: This isn't updated when not enabled

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,190 @@
...
+      if not self.lkas_on:
+        if not self.acc_active_last:
+          self.low_speed_lockout = True
+        else:
+          self.low_speed_alert = True
+      else:
+        self.low_speed_lockout = False
+        self.low_speed_alert = False
+
+      ret.steerWarning = handsoff
```
adeebshihadeh, line 92: handsoff isn't used anywhere else, just directly set ret.steerWarning

selfdrive/car/mazda/carstate.py
```diff
@@ -0,0 +1,190 @@
...
+
+    ret.seatbeltUnlatched = cp.vl["SEATBELT"]['DRIVER_SEATBELT'] == 0
+    ret.doorOpen = any([cp.vl["DOORS"]['FL'], cp.vl["DOORS"]['FR'],
+                        cp.vl["DOORS"]['BL'], cp.vl["DOORS"]['BR']])
+
+    ret.gas = cp.vl["ENGINE_DATA"]['PEDAL_GAS']
+    ret.gasPressed = ret.gas > 0
+
+    # No steer if block signal is on
+    # block = cp.vl["STEER_RATE"]['LKAS_BLOCK'] == 1
```
adeebshihadeh, line 60: Unused, remove

selfdrive/car/mazda/carcontroller.py
Jafaral: Sometimes we may need to send a couple of msgs to get acc cancel, would switching to pcm_cancel_cmd cause an issue there?

adeebshihadeh: No, it should behave the same in that regard

selfdrive/car/mazda/mazdacan.py
```diff
@@ -0,0 +1,104 @@
...
+
+  if button == Buttons.CANCEL:
+    can = 1
+    res = 0
+  elif button == Buttons.RESUME:
+    can = 0
+    res = 1
+  else:
+    can = 0
+    res = 0
```
adeebshihadeh, line 70:
```
can = button == Buttons.CANCEL
res = button == Buttons.RESUME
```

PR #991, opened 2020-01-22
WIP: Add Arm64 CI with Travis

EwoutH:
Runs the docker tests on Aarch64 platform

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,9 @@
+sudo: required
+language: minimal
+arch: arm64
+
+services:
+  - docker
+
+script:
+  - ./run_docker_tests.sh
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ services:
 
 env:
  - SETUP="cd /tmp/openpilot && "
- - RUN="docker run --shm-size 1G --rm tmppilot /bin/sh -c"
+   RUN="docker run --shm-size 1G --rm tmppilot /bin/sh -c"
 
 install:
  - docker build -t tmppilot -f Dockerfile.openpilot .
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,5 +5,25 @@ arch: arm64
 services:
   - docker
 
+env:
+ - SETUP="cd /tmp/openpilot && "
+ - RUN="docker run --shm-size 1G --rm tmppilot /bin/sh -c"
+
+install:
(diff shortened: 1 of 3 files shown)


EwoutH:
Now that the main CI switched to GitHub Actions, here's an attempt use the [multiple CPU architecture](https://docs.travis-ci.com/user/multi-cpu-architectures/) possibility of Travis by trying to compile and run the docker tests on Arm64. Testing on Arm64 (ARMv8 Aarch64 to be precise) should more closely mirror running on the target hardware (like _comma two_).

An alternative approach for CI on Arm64 would be [self-hosted runners](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/hosting-your-own-runners) with GitHub Actions. In that case we need to add `runs-on: [self-hosted, linux, ARM64]` to the Actions build matrix. We could even run the tests on actual _comma two_ hardware, if they were to be [added as self-hosted hardware](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/adding-self-hosted-runners).

If there's any preference for either one of the two methods please let me know.

EwoutH:
Seems for now to [be stuck](https://travis-ci.org/commaai/openpilot/builds/640438811#L3275) on [Dockerfile.openpilot#L61](https://github.com/commaai/openpilot/blob/master/Dockerfile.openpilot#L61)

pd0wm:
Hey Ewout!

Just saw that Github has binaries for ARM64 self-hosted runners. We just need to set up an separate network/DMZ on our side to connect them to, but then we could add some comma two runners. We're pretty busy right now, so that could take a couple more weeks.

I think having both methods is nice. The travis one can run on "untrusted" PRs, the local hardware can run on everything that is vetted by us and merged into the master branch.

Is there a prebuilt wheel you can install to get around the numpy issue?

EwoutH:
Wereld is klein!

At this moment I don't think Android is supported, at least [not officially](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/about-self-hosted-runners#supported-operating-systems-for-self-hosted-runners). To start we could just use any Linux distribution however on any Arm64 hardware (a Pi 4 or VIM3 for example) to start testing.

I will dive deeper into the numpy issue.

pd0wm:
Just merged on-premise building on an EON!

pd0wm:
Closing this PR for now. Feel free to reopen when you want to continue working on this.

PR #659, opened 2019-05-22
Have Wheel Rotate

arne182:
Take the steering wheel angle and have the wheel in the top right rotate in that direction.

PR #993, opened 2020-01-23
Rotate the steering wheel on the ui.

rav4kumar:
small details matters!

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -645,6 +645,7 @@ static void ui_draw_vision_event(UIState *s) {
     const int img_wheel_size = bg_wheel_size*1.5;
     const int img_wheel_x = bg_wheel_x-(img_wheel_size/2);
     const int img_wheel_y = bg_wheel_y-25;
+    const float img_rotation = s->scene.angleSteers/180*3.141592;
     float img_wheel_alpha = 0.1f;
     bool is_engaged = (s->status == STATUS_ENGAGED);
     bool is_warning = (s->status == STATUS_WARNING);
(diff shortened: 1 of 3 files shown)


VirtuallyChris:
I've always thought this should be in the UI, it would be an extra troubleshooting method for people to know if their car fingerprinted correctly while also looking extremely cool and clean.

mchonan1996:
Love this detail, nice PR.

jyoung8607:
It's a pretty nice "polish" feature to have, and would match the Explorer behavior. I'd love to see it. That said, if this is integrated, be prepared to get some questions or problem reports about the displayed steering wheel not matching up with the real steering wheel.

The EON wheel is too small to visually spot fine differences like alignment and torsion/flex and straight-ahead tracking learned corrections, but you could absolutely see differences from folks with variable ratio steering racks. I can tell you right now you'll see it on VW MQB, because we drive on a signal from the car that is corrected for the above factors.

**Steering wheel angle vs true rack angle:**
Mfr specification: variable ratio 14.1 (center) to 9.5:1 (outside)
Steering from 0 to 45, 90, 180, full lock (340-ish)

lwi-vs-eps

dibble9012:
I'd love this feature, but for some reason when I apply it, all my UI goes blank with the exception of the green border.

arne182:
https://github.com/commaai/openpilot/pull/659
Let us see if @geohot has a change of heart

valish:
Hi @rav4kumar, this is a fun enhancement to the UI for playful reasons. Thanks for taking the openpilot UI and making it more fun to use in your car. However, I don't believe this feature should be merged into openpilot master as this confuses the original intent of the openpilot status icon and may increase unneeded attention to the device screen. The openpilot icon serves as a way to communicate the state of the openpilot system as either engaged or disengaged. Rotating the icon with the steering wheel is novel and can be fun, but the motive with the openpilot UI is to communicate critical operating states and relevant road features that may impact the performance of openpilot.

mchonan1996:
Maybe the real solution here is to have a different icon or something to what Explorer has. Like maybe a Comma logo in the circle instead?

PR #998, opened 2020-01-24
Add 2019 Civic Hatchback Touring

IppSec:
Added fingerprint for 2019 Civic Hatchback Touring.


vanillagorillaa:
This fingerprint was taken from the wrong bus. Needs to be taken from bus 1, not 0

IppSec:
> This fingerprint was taken from the wrong bus. Needs to be taken from bus 1, not 0

Can you link me somewhere that explains how to pull from a different bus?  Had used: PYTHONPATH=/data/openpilot PREPAREONLY=1 /data/openpilot/selfdrive/debug/get_fingerprint.py to pull it.  Once I updated the fingerprint, my EON recognized the car so assumed I had done it correctly.  

The only issue I'm having is the eon doesn't recognize the car if it is powered on prior to the car being on.  If I turn the car on, then reboot the EON it will detect but was under the impression that this was a 7.1 bug.

IppSec:
Closing pull request. Method of fingerprinting changed to firmware.

PR #1000, opened 2020-01-24
Subaru continuous blinker signals for assisted lane change

martinl:
I rebased to devel and minimised the lines of code as recommended in https://github.com/commaai/openpilot/pull/974

Tested on two 2018 Crosstreks, assisted lane change is working after the change

Before
alc_oscillating_blinker

After
alc_continuous_blinker


geohot:
I think this is good once we fix the tests.

quillford:
No, update_refs can only be done internally right now. Maybe a small public CI bucket could be made that anyone can push to? That could be a problem if people overwrite existing reference logs though. Another solution might be to write a github bot that updates the refs when it's tagged by anyone in the organization.

pd0wm:
I'll update the ref on Monday and merge the PR.

A Github bot would be pretty cool. Not sure if it's worth the time writing it. I'll write some helpers scripts in the meantime to make it easier.

PR #1011, opened 2020-01-26
dump fingerprints from all CAN buses into /data/my_car/my_fingerprint…

che220:
Currently new users needs to get fingerprints with get_fingerprint.py when EON is connected but OEM ACC/LKAS is enabled. I found this process cumbersome and unreliable, based on my own experience and frustrations expressed on Discord. This change is to dump out fingerprints the OP receives when OP is in enabled. The output file is /data/my_car/my_fingerprints.txt. This file gets created every time when OP is enabled, whether the car can be recognized or not. Benefit:

1. No longer need a separate process to get fingerprints
2. much easier for new users and support on fingerprint-related issues

I ran tests with my 2019 Honda Accord LX (Bosch). All were successful. No impact on any other aspects of OP.

### DIFF
--- a/selfdrive/car/car_helpers.py
+++ b/selfdrive/car/car_helpers.py
@@ -117,7 +117,7 @@ def fingerprint(logcan, sendcan, has_relay):
   return car_fingerprint, finger, vin, car_fw
 
 
-def dump_fingerprints(fingerprints):
+def dump_fingerprints(candidate, fingerprints):
   """
   dump all fingerprints from various CAN buses into /data/my_car/my_fingerprints.txt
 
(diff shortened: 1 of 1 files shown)


che220:
I forgot one thing. If it can dump out the candidate car, it can help people see if there is a possible fingerprint conflict.

geohot:
We are moving away from fingerprinting 1.0 and aren't going to support this for much longer. Fingerprinting 2.0 I believe is already doing this, but using a param instead of a new directory.

We'll have a much more clear fingerprinting guide and strategy once 2.0 is live. In fact, we should even be able to do it remotely when we can correlate with VIN from future match.

PR #1013, opened 2020-01-27
Combine and clean up fingerprints

arne182:
To have less collisions I suggest combinding fingerprints and eventuel splitting Lexus_IS into two as they are 
Mismatches: {921: Unsupported: 7, this: 8; 467: Unsupported: 8, this: 5}
This would allow the inclusion of other fingerprints without having doubles.


arne182:
One collision 
{581: 5, 295: 8, 810: 2, 1166: 8, 166: 8, 170: 8, 466: 8, 865: 8, 608: 8, 467: 8, 560: 7, 1020: 8, 610: 8, 296: 8, 552: 4, 550: 8, 950: 8, 37: 8, 1167: 8, 36: 8, 180: 8, 1595: 8, 1779: 8, 562: 6, 1059: 1, 452: 8, 713: 8, 1017: 8, 812: 8, 1114: 8, 1002: 8, 975: 5, 824: 8, 426: 6, 800: 8, 971: 7, 933: 8, 863: 8, 945: 8, 898: 8, 951: 8, 1132: 8, 1085: 8, 764: 8, 1057: 8, 1071: 8, 829: 2, 905: 8, 1541: 8, 956: 8, 1552: 8, 1568: 8}

Exact fingerprint match!
TOYOTA RAV4 HYBRID 2019
84 less key/value pairs

Exact fingerprint match!
LEXUS ES 300H 2019
79 less key/value pairs

Here we still have 79 keys that we could check to determine the difference.

geohot:
Fingerprinting 1.0 is about to be deprecated, plan is to remove it entirely after 2.0 is working. This pull request is really hard to test, and I'm concerned it'll break what's already working.

PR #1015, opened 2020-01-27
Honda Pedal Fixes

VirtuallyChris:
This pull request fixes a long-standing issue with Honda Pedal users. When using a pedal currently, the cruise cancel command is still sent which causes the brake to oscillate between values rapidly causing a [noticeable grinding noise](https://youtu.be/CrcZB59LmAc) at low speeds. This is fixed by preventing the cancel command to be sent if a pedal is present.

I've also copied the code to limit max gas velocity that the Toyota interface file uses which greatly prevents lurching at low speeds.

Lastly, I've added some pedal tuning for the Civic as it has the worst tune out of all Hondas thanks to the higher value longitudinal tune. Without the tune it would floor it, slam brakes, and repeat until reaching around 35mph.


VirtuallyChris:
I removed the Civic specific tuning as I'm not sure this is the best way to go about tuning comma pedal. If accel overshoot is prioritized, the braking force is too weak to stop the car in time, so I was left with a middle where it accels pretty abruptly but brakes too much then creeps forward to meet the lead car in stop-and-go. It could need a gas delay like steer actuator delay since the car takes a second to respond most likely causing the loop to overshoot, but this will need to be a separate pull request after some more digging.

[Here's a route with the changes in this PR](https://my.comma.ai/cabana/?route=ea70677b54a1bca1%7C2020-02-03--16-33-06&exp=1612402151&sig=VIqdPmzaC2OEOkoFbkE1w9mkEmY6W6XjMGHhHiLaAXE%3D&max=10&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2Fea70677b54a1bca1%2Feaf5c9985824e237e4955f32c90ac025_2020-02-03--16-33-06) - I do believe both of the changes in here will make the pedal experience much better for Honda owners (getting rid of the grinding is a huge win!)

Let me know if you need anything else from me to get this PR merged.

selfdrive/car/honda/carcontroller.py
```diff
@@ -178,6 +178,9 @@ def update(self, enabled, CS, frame, actuators, \
         idx = frame // 2
         ts = frame * DT_CTRL
         pump_on, self.last_pump_ts = brake_pump_hysteresis(apply_brake, self.apply_brake_last, self.last_pump_ts, ts)
+        # If pedal detected, don't send cancel command to prevent brake grinding.
```
geohot: I don't think this is the right place for this, it's way too close to the output. Why is cancel being generated, and is there a better place to block it if that's what we want.

VirtuallyChris:
Thanks for the comment George, that was a really hacky fix now that took a closer look at it.

The root of the problem was pcm_cancel_cmd being set to true every update in carcontroller because of ret.enableCruise being set to false in interface when a gas interceptor was detected. With this being true always, the carcontroller operates like normal thus the brakes don't think they are being constantly canceled by cruise, so they no longer grind.

[Cabana with pedal](https://my.comma.ai/cabana/?route=ea70677b54a1bca1%7C2020-02-14--20-50-07&exp=1613329618&sig=i4941vINqJ6BrQPRgIrMnVqSUWyTylUekm4p0wJ4N8c%3D&max=20&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2Fea70677b54a1bca1%2F55bb95766bd13d378d0d5f1b1d5abb4b_2020-02-14--20-50-07) - no more brake grinding and pcm_cancel_cmd operating normal again

[Cabana without pedal](https://my.comma.ai/cabana/?route=ea70677b54a1bca1%7C2020-02-14--21-17-48&exp=1613330376&sig=EqOU6pMZa%2B6G%2BHC0JIFOuq5q%2Bisc8LgfZTg1uiVrlAk%3D&max=9&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2Fea70677b54a1bca1%2Fb34d543bad89f204755063536782eb83_2020-02-14--21-17-48) - still functions as expected, no issues

PR #1016, opened 2020-01-28
dmonitoringd

ZwX1616:
* DM is no longer a part of controlsd
* reduce wheel touch timeout to 1min
* improve model as in 0.7.2 description


selfdrive/controls/controlsd.py
```diff
@@ -77,6 +75,8 @@ def data_sample(CI, CC, sm, can_sock, driver_status, state, mismatch_counter, ca
   sm.update(0)
 
   events = list(CS.events)
+  if sm.updated['monitorState']:
```
pd0wm: Don't you always want to add the events?

selfdrive/controls/monitord.py
```diff
@@ -0,0 +1,111 @@
...
+
+  # wait until carState live
+  CS_updated = False
+  while(not CS_updated):
+    sm.update(0)
+    CS_updated = sm.updated['carState']
+
+  # 10Hz <- monitoringd
+  while(True):
+    start_time = sec_since_boot()
```
pd0wm: Is this used?

selfdrive/controls/monitord.py
```diff
@@ -0,0 +1,111 @@
...
+    sm.update(0)
```
pd0wm: Don't update with zero timeout. Use the default timeout if there is no specific reason to change it.

selfdrive/controls/monitord.py
```diff
@@ -0,0 +1,111 @@
...
+    cal_rpy = [0,0,0]
+    if cal_status == Calibration.CALIBRATED:
+      rpy = sm['liveCalibration'].rpyCalib
+      if len(rpy) == 3:
+        cal_rpy = rpy
+
+    driver_status.get_pose(sm['driverMonitoring'], cal_rpy, CS.vEgo, CS.cruiseState.enabled)
+
+    # Get model meta
+    if sm.updated['model']:
```
pd0wm: Because you continue the loop if `driverMonitoring` is not updated, this will only be true if they were updated at the same time. I would structure the code like:

```python
if sm.updated['liveCalibration']:
  # update calibration
if sm.updated['carState']:
  # update v_cruise + driver_engaged
if sm.updated['gpsLocation'] and not driver_status.is_rhd_region_checked:
 # gps stuff 
if sm.updated['driverMonitoring']:
  # driver monitoring stuff
  # send out packet

File: selfdrive/test/process_replay/process_replay.py [hunk 5]
```diff
@@ -228,6 +228,16 @@ def calibration_rcv_callback(msg, CP, cfg, fsm):
     init_callback=get_car_params,
     should_recv_callback=calibration_rcv_callback,
   ),
+  ProcessConfig(
+    proc_name="monitord",
+    pub_sub={
+      "monitorState": [],
```
pd0wm: You're using the wrong syntax here. It's:

```python
pub_sub = {
 "socket that your process subscribes to": ["packet that is send as a response"]
}
```

So should be something like:
```python
"driverMonitoring": ["monitoringState"], "liveCalibration": [], "carState": [], "model": [], "gpsLocation": [],
```

selfdrive/controls/controlsd.py
```diff
@@ -387,8 +357,7 @@ def data_send(sm, pm, CS, CI, CP, VM, state, events, actuators, v_cruise_kph, rk
     "alertBlinkingRate": AM.alert_rate,
     "alertType": AM.alert_type,
     "alertSound": AM.audible_alert,
-    "awarenessStatus": max(driver_status.awareness, -0.1) if isEnabled(state) else 1.0,
-    "driverMonitoringOn": bool(driver_status.face_detected),
+    "driverMonitoringOn": bool(sm['monitorState'].driverState.faceDetected),
```
pd0wm: don't need the bool wrapper anymore. This is already a bool from the capnp structure.

selfdrive/controls/monitord.py
```diff
@@ -0,0 +1,111 @@
...
+  v_cruise_last = 0
+
+  # wait until carState live
+  CS_updated = False
+  while(not CS_updated):
+    sm.update(0)
+    CS_updated = sm.updated['carState']
+
+  # 10Hz <- monitoringd
+  while(True):
```
pd0wm:
```suggestion
  while True:
```

selfdrive/controls/monitord.py
```diff
@@ -0,0 +1,111 @@
...
+  driver_status = DriverStatus()
+  is_rhd = params.get("IsRHD")
+  if is_rhd is not None:
+    driver_status.is_rhd = bool(int(is_rhd))
+
+  sm['liveCalibration'].calStatus = Calibration.INVALID
+  v_cruise_last = 0
+
+  # wait until carState live
+  CS_updated = False
```
pd0wm: Is this really needed? Can't you set up carState with some sane default values? See how plannerd sets up the `liveParameters`.

selfdrive/controls/monitord.py
```diff
@@ -0,0 +1,111 @@
...
+    # Get data from monitoringd
+    if not sm.updated['driverMonitoring']:
+      continue
+
+    CS = sm['carState']
+    events = []
+
+    # GPS coords RHD parsing, once every restart
+    if not driver_status.is_rhd_region_checked and sm.updated['gpsLocation']:
+      is_rhd = is_rhd_region(sm['gpsLocation'].latitude, sm['gpsLocation'].longitude)
```
pd0wm: Move this code to the driver_status class?

selfdrive/controls/controlsd.py
ZwX1616: fixed. experienced this bug on my test drive too

selfdrive/controls/monitord.py
```diff
@@ -0,0 +1,109 @@
...
+                        v_cruise != v_cruise_last or \
+                        sm['carState'].steeringPressed
+      v_cruise_last = v_cruise
+
+    # Get model meta
+    if sm.updated['model']:
+      driver_status.set_policy(sm['model'])
+
+    # Get data from monitoringd
+    if not sm.updated['driverMonitoring']:
```
pd0wm: Invert condition and move code below into the if

pd0wm:
Did you see any change to cpu usage after the refactor? Interested in controlsd usage before and after this change.

ZwX1616:
> Did you see any change to cpu usage after the refactor? Interested in controlsd usage before and after this change.

will do another test drive today. we shall see

ZwX1616:
> Did you see any change to cpu usage after the refactor? Interested in controlsd usage before and after this change.

controlsd: ~51% at normal cruise, ~56% with distraction event; 
monitord: 5.3%-5.6% at all times

pd0wm:
How much was controlsd before the change?

pd0wm:
Can you rebase to latest master and make sure the release build also works? You probably need to add `selfdrive/controlsd/monitoringd.py` to `release/files_common`.

Do you want this to ship with 0.7.2?

ZwX1616:
> How much was controlsd before the change?

around 60%

ZwX1616:
> Can you rebase to latest master and make sure the release build also works? You probably need to add `selfdrive/controlsd/monitoringd.py` to `release/files_common`.
> 
> Do you want this to ship with 0.7.2?

Yes and yes. The added log struct will be super useful for future troubleshooting. 
Adding the file now...

PR #1022, opened 2020-01-30
Support for Lexus NX 300h

kengggg:


kengggg:
Should I test for both DSU-plugged and unplugged?

kengggg:
Hm. It seems my segment could not be found hence the test_car_model fails. Did I do anything wrong? @vanillagorillaa

selfdrive/test/test_car_models.py
```diff
@@ -313,6 +313,15 @@ def get_route_logs(route_name):
     'carFingerprint': TOYOTA.LEXUS_RXH,
     'enableCamera': True,
     'enableDsu': True,
+  },
```
pd0wm: Think something went wrong solving the merge conflict here. You're adding the RX_TSS2 again, which was added in another PR.

kengggg: I was confused while rebasing, sorry.

kengggg:
This PR is a mess now. Will start over.

PR #765, opened 2019-07-27
Add support for Toyota Corolla LE Hybrid 2020

hdoublearp:
Now working! Big shout out to @wocsor, couldn't have done it without you.

Will provide cabana logs for stock vs op in a few hours.

PR #1023, opened 2020-01-30
Add support for 2020 Corolla Hybrid LE - Rebase on master

hdoublearp:
Previously based on 0.6.1, rebased on master. Refer to #765 for cabana logs.

Waiting for fingerprinting 2.0 before merging.


pd0wm:
- Why did you change the wheelbase?
 - Why do we need to wait till fingerprinting 2.0? It looks like there are no overlapping fingerprints

arne182:
I have had now a few corollas 2020 on my fork and they do not complete the fingerprint process as there fingerprint a subset of the RAV4 2019 and the code has 2 candidates left. The rav4's seem to be fine because they have a longer fingerprint but the Corolla has no mismatching signals and then just gets kicked out at an unknown car

pd0wm:
Then it doesn't make much sense to merge this PR. 0.7.2. will be able to fingerprint them based on FW versions.

thesoundofom:
@pd0wm 
- Indeed, the wheel base number I got was from Toyota's website that I mentioned in the commit description. I corroborated that number from other sources as well. It appears to me that when CAR.COROLLAH_TSS2 was created, the wheelbase was copied over from Corolla Hatch, which is slightly shorter in length at 103.9 inches (2.63906), whereas the Hybrid has a longer wheelbase at 106.3 inches (2.70002 meters). Ultimately they are not same car, one being a hatchback and the other being a sedan. The Altis Hybrid and Corolla Hybrid however share the same unibody and wheelbase.

@arne182 @pd0wm 
- As far as the fingerprint conflict goes, during testing under 0.6.1 the car was identifying as the Camry Hybrid and therefore using the wrong ADAS configuration. It turned out, that by removing the Camry Hybrid allowed the Corolla Hybrid to function correctly. I believe the Corolla Hybrid fingerprint also broke compatability with the Camry Hybrid in the same fashion. @geohot had asked me to rebase this on the master branch, but since it had been so long since I'd pulled any changes from upstream I decided to rebase my fork and republish the changes. This was the result of many weeks of testing and trial and error. @wocsor I believe ran into the same issue with another car port.
- However, since 0.7.2. will be able to fingerprint them based on FW (is this the VIN fingerprinting that @geohot mentioned?), it doesn't make sense to merge this PR. At the end of the day, I submitted it to have a record of the work put into porting the car. I'm not sure if folks using the EON with the original panda will be able to use VIN fingerprinting, so my fork will remain available for those folks for the time being if needed.

Thanks,
Christian

pd0wm:
@thesoundofom 

FW based query is now merged into master. Asked a few people with a Corolla LE, and their FW versions seem to be in the database now.

thesoundofom:
Corolla LE or Corolla Hybrid LE?

PR #1028, opened 2020-01-31
honda steering improvements for stock and modded eps

jpancotti:
Taken from Gregs PR, and updated with breakpoint values for civic, civic_bosch, and crv_5g. I also added breakpoints for stock firmware for better control and linear response on each car.
steer_max is now more specific for each car to eliminate the dead zone in civic.
TO DO: define these by eps fw version number from new fingerprint system


ffhspa:
Any Update on this? Many Honda users would love to have a better Steering experience with OP. 
Thanks a lot!

PR #1030, opened 2020-01-31
Add network_type to thermald

valish:


selfdrive/thermald.py
```diff
@@ -135,6 +137,36 @@ def handle_fan_uno(max_cpu_temp, bat_temp, fan_speed, ignition):
 
   return new_speed
 
+def get_network_type():
+  wifi_check = android.parse_service_call_string(android.service_call(["connectivity", "2"]))
+  cell_check = android.parse_service_call_unpack(android.service_call(['phone', '59']), ">q")
```
pd0wm: maybe move the service call inside the if so we don't call it if not needed?

selfdrive/thermald.py
```diff
@@ -135,6 +137,36 @@ def handle_fan_uno(max_cpu_temp, bat_temp, fan_speed, ignition):
...
+  cell_networks = {
```
pd0wm: This looks very android specific. Maybe move the service call + parsing into the android library?

pd0wm:
LGTM!

PR #1036, opened 2020-02-02
Multipath supercombo

haraschax:
New model
- predicts validity in meters of path and lanelines
- trained with multipath, so paths should not suffer from being somewhere in between viable paths
- gaussian loss lead means for less critical lead speed errors
- path trained with supercombo features, should improve path all round, most noticeably smooth high speed lane changes


selfdrive/modeld/models/driving.cc
```diff
@@ -133,13 +133,19 @@ void fill_path(cereal::ModelData::PathData::Builder path, const float * data, bo
   float poly_arr[POLYFIT_DEGREE];
   float std;
   float prob;
+  float valid_len;
 
+  valid_len =  data[MODEL_PATH_DISTANCE*2];
   for (int i=0; i<MODEL_PATH_DISTANCE; i++) {
     points_arr[i] = data[i] + offset;
-    stds_arr[i] = softplus(data[MODEL_PATH_DISTANCE + i]);
+    if (i < valid_len) {
```
geohot: Is there a minimum lower sanity bound on this? Concerned it'll be <= 4, and fit a crazy third order polynomial.

HaraldSchafer: Hmmmm, during training the minimum for path is 5. However there may not be a minimum for lanelines. I'll add a minimum.

geohot: Always good to enforce the minimum on the model outputs here as well, it's always possible to get weird outputs from metaparams like this.

pd0wm:
Can you add the validity to the path/lane information output from the model? Then we can use in the ui and planner.

selfdrive/modeld/models/driving.cc
HaraldSchafer: yeah I'll add that

HaraldSchafer:
Yeah I can add it to the msgs, but it doesn't really mean anything actually. It's just kind of a hack for missing data during training. It doesn't say anything about how far the model can see. The stds should still be used for that

PR #1037, opened 2020-02-02
manager runs on Mac, and other openpilot for PC fixes

geohot:


geohot:
Hmm, so this is actually broken. The C params API is using a different directory, guess it doesn't matter for the tests?

geohot:
Test car models is timing out still!

PR #1045, opened 2020-02-03
GM: Resume Required Alert should be triggered off of ECM state

andrewcopenpilot:
On GM cars the ECM goes into cruise state 4 (standstill) and in order
for the ECM to accept cruise inputs, the resume button must be pressed
to drive it to state 1 (active).

Previously rather than testing for if the ECM was in state 4, the wheel
speed sensors were checked that the vehicle was going less than 0.5 m/s.

In some cases this could cause the resume required alert to be displayed
before the car actually came to a stop as far as the ECM was concerned.
If the alert is triggered before state 4 and OP demands acceleration,
its possible for OP to acelerate while the Resume Required Alert is
displayed created a very confusing UX.

This PR updates the alert to be triggered to fire when the ECM reports that it is in state 4.

### DIFF
--- a/selfdrive/car/gm/gmcan.py
+++ b/selfdrive/car/gm/gmcan.py
@@ -67,8 +67,8 @@ def create_friction_brake_command(packer, bus, apply_brake, idx, near_stop, at_f
   else:
     mode = 0xa
 
-    if at_full_stop:
-      mode = 0xd
+  if at_full_stop:
+    mode = 0xd
     # TODO: this is to have GM bringing the car to complete stop,
     # but currently it conflicts with OP controls, so turned off.
     #elif near_stop:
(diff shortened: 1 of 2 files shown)


pd0wm:
This makes sense! What models was this tested on? Do we want to make sure it works on all models before we merge the PR?

andrewcopenpilot:
@pd0wm I tested on my 2018 ATS, no issues, I've asked someone in the community to verify on a Volt as well.

andrewcopenpilot:
@pd0wm after more review the actual root cause appears to be that the mode for `at_full_stop` can't be set if the braking pressure is 0. I'm assuming there is some latency between brake commands and the actuator responding with pressure, or I suppose going up a hill. In rare cases it seems to be possible to cross the threshold of when the full stop mode should be set when the brake demand is 0.

The current if/else structure prevents full stop from ever being set when there is 0 brake pressure.

Thanks Kish for pointing out that I had confused two different standstill states and the first commit of this PR was actually a no-op.

pd0wm:
Thanks for verifying.

What cars was this new code tested on? If we have a few different models I'll merge it in.

kishkaru:
Tested on Volt, working as expected. FWIW, a popular community fork had this fix for over a year now.

geohot:
Wow, this just looks like a typo

PR #1050, opened 2020-02-04
Cleanup simulator and add usage instructions

geohot:


tools/sim/README
```diff
@@ -0,0 +1,26 @@
...
+# Add export PYTHONPATH=$HOME/openpilot to your bashrc
+
+== Tensorflow (optional, for fast models) ==
+
+cd ~/openpilot/external/tensorflow
+./download.sh
+
+== Install (in tab 1) ==
+
+./start_carla.sh  # install CARLA 0.9.7 and start the server
```
pd0wm: cd ~/openpilot/tools/sim ?

tools/sim/README
```diff
@@ -0,0 +1,26 @@
...
+
+== openpilot (in tab 2) ==
+
+cd ~/openpilot/selfdrive/
+NOBOARD=1 ./manager.py
```
pd0wm: For me it complained about Passive not being set. Should we add an env variable, e.g. `PASSIVE=0 ./manager.py` that writes the param?

tools/sim/start_carla.sh
```diff
@@ -0,0 +1,18 @@
...
+
+FILE=CARLA_0.9.7.tar.gz
+if [ ! -f $FILE ]; then
+  curl -O http://carla-assets-internal.s3.amazonaws.com/Releases/Linux/$FILE
+fi
+if [ ! -d carla ]; then
+  mkdir -p carla_tmp
+  cd carla_tmp
+  tar xvf ../$FILE
+  easy_install PythonAPI/carla/dist/carla-0.9.7-py3.5-linux-x86_64.egg
```
pd0wm: This needs sudo depending on how python3 is installed.

tools/sim/start_carla.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -e
+
+FILE=CARLA_0.9.7.tar.gz
+if [ ! -f $FILE ]; then
+  curl -O http://carla-assets-internal.s3.amazonaws.com/Releases/Linux/$FILE
+fi
+if [ ! -d carla ]; then
+  mkdir -p carla_tmp
```
pd0wm: clean up the old tmp dir?

tools/sim/start_carla.sh
pd0wm: Also got this when running with sudo:
```Processing carla-0.9.7-py3.5-linux-x86_64.egg
creating /usr/local/pyenv/versions/3.7.3/lib/python3.7/site-packages/carla-0.9.7-py3.5-linux-x86_64.egg
Extracting carla-0.9.7-py3.5-linux-x86_64.egg to /usr/local/pyenv/versions/3.7.3/lib/python3.7/site-packages
Adding carla 0.9.7 to easy-install.pth file

Installed /usr/local/pyenv/versions/3.7.3/lib/python3.7/site-packages/carla-0.9.7-py3.5-linux-x86_64.egg
Processing dependencies for carla==0.9.7
Searching for carla==0.9.7
Reading https://pypi.org/simple/carla/
No local packages or working download links found for carla==0.9.7
error: Could not find suitable distribution for Requirement.parse('carla==0.9.7')```

geohot:
@HaraldSchafer Merged in. This should work when the keras model matches the dlc, since it was already broken in master I just merged it.

tools/sim/start_carla.sh
geohot: Hmm, I don't want to put sudo in there, that's bad python if that's how it's set up

PR #1053, opened 2020-02-04
WIP: honda bosch longitudinal using vision radar

gregjhogan:
No longer requires unplugging radar!

panda safety from https://github.com/commaai/panda/pull/268

credit goes to @csouers for reverse engineering most of the signals

TODO:
- [ ] doesn't always decel fast enough (vision radar problem? mpc problem?)
- [ ] can be jerky going into and out of standstill
- [ ] doesn't always accel fast enough
- [ ] sometimes accel is too fast
- [ ] how should `GAS_COMMAND` be set relative to `ACCEL_COMMAND` when positive?
- [ ] correct to switch most things over to using `openpilotLongitudinalControl` instead of `radarOffCan`?
- [ ] auto-detect when this should be active
- [ ] fix setting `gasMaxV`/`brakeMaxV`
- [x] fix pressing gas while pressing speed increase button repeatedly causing error
- [ ] sometimes there is a auto high beam error on start

### DIFF
--- a/common/params.py
+++ b/common/params.py
@@ -85,6 +85,7 @@ keys = {
   "LiveParameters": [TxType.PERSISTENT],
   "LongitudinalControl": [TxType.PERSISTENT],
   "OpenpilotEnabledToggle": [TxType.PERSISTENT],
+  "VisionRadarToggle": [TxType.PERSISTENT],
   "LaneChangeEnabled": [TxType.PERSISTENT],
   "PandaFirmware": [TxType.CLEAR_ON_MANAGER_START, TxType.CLEAR_ON_PANDA_DISCONNECT],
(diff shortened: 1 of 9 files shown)


jpancotti:
Thanks for putting this together Greg. 
Wanted to share my notes as well, as I have done some testing to tune out the decel issue. I noticed that the lead indicator is right on target at all times, even going off screen when you are about to collide, so it seems that perception is not the problem.
So far I have tried:
Increase Kp and Ki to the point where acceleration is very abrupt - no effect on weak decel.
Disable max_accel_v to allow full accel/decel at 20mph and below - no effect on weak decel.
Reduce min_accel - causes cruise faults, no effect on weak decel.

Positives: 
creeping is smooth as butter, better than stock acc
stopped car and cut in detection seems to be an improvement over stock acc as well, aside from the weak decel issue

PR #1055, opened 2020-02-05
test car models: fix random hanging

### LINKED ISSUE #1043: Test car models CI stage gets stuck sometimes

adeebshihadeh:
Running locally, I reproduced a similar failure to the one seen in CI recently. It would get stuck downloading the front camera video, which was unnecessary, and was probably only done to make unlogger happy. Not having to download those videos each time should speed up the test a bit. Also added a timeout to the request so it doesn't just wait forever.

Fixes #1043


pd0wm:
Nice! This should also significantly cut down on bandwidth cost to our commadataci bucket :)

PR #1056, opened 2020-02-06
use services.h in loggerd

### LINKED ISSUE #1039: Refactor loggerd to use services.h
Right now, it's reading the yaml file directly. See the code in cereal for how to use services.h

Do this, and we can also remove the yaml-cpp stuff from phonelibs I believe.

adeebshihadeh:
fixes #1039


pd0wm:
That's a nice negative diff!

PR #1059, opened 2020-02-07
Focusss

ZwX1616:
improve autofocus stability


jpancotti:
Does this prevent cam from focusing in to the windshield? I've had that happen a few times if there is water on the windshield

ZwX1616:
> Does this prevent cam from focusing in to the windshield? I've had that happen a few times if there is water on the windshield

Yeah it fixes some unexpected wrong focus including accidentally focusing to somewhere very close

PR #1060, opened 2020-02-07
Lateral-only car port for Chevy Bolt EV, Chevy Equinox

JMPZ11:
Basic modifications to implement support for the Chevy Bolt EV Premier (all years) and the 2020 Chevy Equinox. It is possible this code will work with minimal modification for any other non-ASCM GM vehicles.

A few changes were required due to the lack of an ASCM and longitudinal controls.

No version of the Bolt EV supports ACC, so for the moment this relies on manual cruise control for longitudinal.

(Work is in progress to make use of the AEB commands and a Comma Pedal to implement camera-based ACC.)

This references the PR on the Panda side - kind of depends on it.


selfdrive/car/gm/carcontroller.py
```diff
@@ -137,17 +143,17 @@ def update(self, enabled, CS, frame, actuators, \
       else:
         apply_gas = int(round(interp(final_pedal, P.GAS_LOOKUP_BP, P.GAS_LOOKUP_V)))
         apply_brake = int(round(interp(final_pedal, P.BRAKE_LOOKUP_BP, P.BRAKE_LOOKUP_V)))
+      if not self.car_fingerprint in NO_ASCM_CARS:
```
andrewcopenpilot: This check does the same thing is the `if self.car_fingerprint not in SUPERCRUISE_CARS:` at line 130.

At the very least I think this if statement should go away and we should do something like.

`if self.car_fingerprint not in SUPERCRUISE_CARS and self.car_fingerprint not in NO_ASCM_CARS:`

Long term its probably better to have new car parameter 'LongitudinalControl' and just check that in the if statement for if we are going to do gas/brake commands.

selfdrive/car/gm/values.py
```diff
@@ -10,9 +10,13 @@ class CAR:
   MALIBU = "CHEVROLET MALIBU PREMIER 2017"
   ACADIA = "GMC ACADIA DENALI 2018"
   BUICK_REGAL = "BUICK REGAL ESSENCE 2018"
+  BOLT = "CHEVROLET BOLT"
+  EQUINOX = "CHEVROLET EQUINOX"
 
 SUPERCRUISE_CARS = [CAR.CADILLAC_CT6]
 
+NO_ASCM_CARS = [CAR.BOLT, CAR.EQUINOX]
```
andrewcopenpilot: I don't think categorizing NO_ASCM_CARS is the best representation by car is the best representation.

LKA_ONLY or NO_ACC is probably a general categorization. Since variable is just used to prevent OP from generating gas/brake commands for cars that can't use them.

This makes sense as written for the Bolt as there is no ACC version, however the Volt comes in both an ACC+LKA version as well as a LKA only version. So a Volt could fit in either category depending on how its optioned.

As I stated in my comment in carcontroller.py I think the answer here is to have a 'LongitudinalControl' parameter, now that Toyota/Honda has firmware based fingerprinting, I think we should look towards implementing something similar in GM and with firmware version we could probably tell the difference between a LKA only Volt and an ACC+LKA Volt.

andrewcopenpilot (commented):
Nice work! Left a few comments.

.gitmodules
```diff
@@ -1,16 +1,16 @@
 [submodule "panda"]
 	path = panda
-	url = ../../commaai/panda.git
```
geohot: Please remove this for PR

geohot:
Thank you for this pull request!

So our current policy has been no openpilot on cars without ACC, though it may be worth it for the Bolt and have it in community supported. Does the Equinox have ACC?

JasonJShuler:
The Equinox we've tested it on doesn't have ACC

.gitmodules
JasonJShuler: I had planned to, but my copy wouldn't build with the commaai references

JasonJShuler:
Readme updated, gitmodules fixed

geohot:
For now, we are going to keep the ACC requirement for upstreaming. The expectation from every car in openpilot is that it perform on the highway without many disengagements, and unfortunately without ACC that's unlikely.

Let's keep this pull request open, hoping the gas/brake can be controlled soon!

JasonJShuler:
I am closing this PR because I am redoing the implementation fresh in the latest Comma code, and I've moved it to a different repository. Additionally I should have used a feature Branch rather than issuing my PR directly out of my fork of Master so that was a mistake

PR #1068, opened 2020-02-09
Add support for Lexus NX 300h 2018

kengggg:
Tested on RHD Lexus NX 300h 2018.


geohot:
Tests are currently failing, can you investigate?

kengggg:
@geohot fixed. The test passes.

selfdrive/car/toyota/interface.py
```diff
@@ -253,6 +253,16 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), has_relay=False,
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.3], [0.05]]
       ret.lateralTuning.pid.kf = 0.00007
 
+    elif candidate == CAR.LEXUS_NXH:
+      stop_and_go = True
+      ret.safetyParam = 100
```
pd0wm: This does not match the 73 from the STEER_TORQUE_EPS message in the DBC file. Which one is correct? If you plot it on top LKAS request they should match.

kengggg: It should be 73.

PR #1075, opened 2020-02-11
Push container to dockerhub

pd0wm:


.github/workflows/test.yaml
```diff
@@ -26,6 +26,24 @@ jobs:
         name: tmppilot.tar.gz
         path: tmppilot.tar.gz
 
+  push:
+    name: push
+    runs-on: ubuntu-16.04
+    needs: build
+    if: github.ref == 'master' && github.event_name != 'pull_request'
```
quillford: This should also check `github.repository == 'commaai/openpilot'` so it doesn't run on forks

pd0wm: I'll add it in case someone turns on actions on their repo. By default they won't run though.

PR #1077, opened 2020-02-11
Add support for Lexus NX 300h 2018

kengggg:
The previous PR contains multiple unrelated commits. This should work.


geohot:
Please add it to README as well. And looks like the car models test is passing!

kengggg:
@geohot added. thanks!

PR #1082, opened 2020-02-12
Add support for Toyota smartDSU interceptor

wocsor:
- Changed DSU fingerprint to 0x283 so DSU can be detected while OP does long control and still allow stock AEB (this also allows carcontroller to not send fake DSU messages while the smartDSU is installed
- Run all of the logic in interface so there's no need to add a bool to car.capnp
- Add smartDSU CAN ID to fingerprints


selfdrive/car/toyota/interface.py
```diff
@@ -297,10 +297,11 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), has_relay=False,
 
     ret.enableCamera = is_ecu_disconnected(fingerprint[0], FINGERPRINTS, ECU_FINGERPRINT, candidate, Ecu.fwdCamera) or has_relay
     # In TSS2 cars the camera does long control
+    smartDsu = 0x2FF in fingerprint[0]
     ret.enableDsu = is_ecu_disconnected(fingerprint[0], FINGERPRINTS, ECU_FINGERPRINT, candidate, Ecu.dsu) and candidate not in TSS2_CAR
     ret.enableApgs = False  # is_ecu_disconnected(fingerprint[0], FINGERPRINTS, ECU_FINGERPRINT, candidate, Ecu.apgs)
     ret.enableGasInterceptor = 0x201 in fingerprint[0]
-    ret.openpilotLongitudinalControl = ret.enableCamera and (ret.enableDsu or candidate in TSS2_CAR)
+    ret.openpilotLongitudinalControl = ret.enableCamera and (smartDsu or (ret.enableDsu or candidate in TSS2_CAR))
```
pd0wm: don't need extra brackets around this or

pd0wm:
Can you also set `ret.communityFeature` in interface.py when the smartDsu is detected?

pd0wm:
Can you add some comments to the code what the smartDSU does? 

As I understand it it forwards everything from the DSU, except for 0x343 so that openpilot can send it. The smartDSU sends 0x343 for a few seconds to make the car happy, and then stops. How do you prevent openpilot and smart dsu sending at the same time?

wocsor:
that's correct. to prevent double-sending 343, the smartDSU firmware reads bus 1 before filtering/forwarding from bus 2, so if it sees 0x343 on bus 1 it will not forward it from bus 2. it's a little rough, but here is the part in panda firmware where that happens: [filter forward](https://github.com/wocsor/panda/blob/5a7f327a6ff285e3930535a7e33346a029af7113/board/safety/safety_defaults.h#L47)

selfdrive/car/toyota/interface.py
```diff
@@ -311,7 +314,8 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), has_relay=False,
     ret.minEnableSpeed = -1. if (stop_and_go or ret.enableGasInterceptor) else 19. * CV.MPH_TO_MS
 
     # removing the DSU disables AEB and it's considered a community maintained feature
-    ret.communityFeature = ret.enableGasInterceptor or ret.enableDsu
+    # intercepting the DSU is a community feature since it requires unofficial hardware
+    ret.communityFeature = ret.enableGasInterceptor or ret.enableDsu or ret.smartDsu
```
pd0wm: typo. No `ret` before smartDsu

pd0wm:
Can you not remove the newlines at the end of the files? See the diff.

wocsor:
my bad! added them back.

PR #1091, opened 2020-02-14
Add support for 2019 Civic Sedan 1.6L Diesel with CVT

csouers:
Unlike other 2019 civics, this car doesn't steer down to 2mph. It's 12mph like the others. Differences in GEARBOX and GAS_PEDAL_2 frames like CRV HYBRID and some other obscure cars like Odyssey and Pilot.

Public cabana route: https://my.comma.ai/cabana/?route=d83f36766f8012a5%7C2020-02-05--18-42-21&exp=1612469711&sig=wgaItnHLgnNBZ1FDm3QS2JdO%2F%2BwtQTIs6illZcfsvik%3D&max=12&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2Fd83f36766f8012a5%2Fa640feac185d79812f37e7f7a2553597_2020-02-05--18-42-21

Todo:
- [ ]  Verify firmware fingerprint is successful on this car. Car belongs to @ffhspa

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,90 +62,91 @@ At the moment, openpilot supports the [EON DevKit](https://comma.ai/shop/product
 Supported Cars
 ------
 
-| Make      | Model (US Market Reference)   | Supported Package | ACC              | No ACC accel below | No ALC below |
-| ----------| ------------------------------| ------------------| -----------------| -------------------| -------------|
-| Acura     | ILX 2016-18                   | AcuraWatch Plus   | openpilot        | 25mph<sup>5</sup>  | 25mph        |
-| Acura     | RDX 2016-18                   | AcuraWatch Plus   | openpilot        | 25mph<sup>5</sup>  | 12mph        |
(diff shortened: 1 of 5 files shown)


pd0wm:
A final thing, as we don't add new Honda/Toyota fingerprints we need to make sure it only fingerprints using FW query. Add a `IGNORED_FINGERPRINTS` list to values, like for the RAVH_TSS2: https://github.com/commaai/openpilot/blob/master/selfdrive/car/toyota/values.py#L270

To make sure the test still works, set 'fingerprintSource' to 'fixed' in the test_car_models.py route definition. Like: https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_car_models.py#L292

csouers:
Thank you. Will revise this and also the Euro CRV PR with this in mind.

csouers:
@pd0wm Do you want me to remove the can-based fingerprint from this as well?

csouers:
@pd0wm I think this is good to go. I'll squash and force push again right now.

PR #1092, opened 2020-02-14
classify honda ecus

jpancotti:
classify Honda ecus according to Honda nomenclature

### DIFF
--- a/selfdrive/car/fw_versions.py
+++ b/selfdrive/car/fw_versions.py
@@ -85,7 +85,7 @@ def match_fw_to_car(fw_versions):
       ecu_type = ecu[0]
       addr = ecu[1:]
       found_version = fw_versions_dict.get(addr, None)
-
+      ESSENTIAL_ECUS = [Ecu.engine, Ecu.eps, Ecu.esp, Ecu.fwdRadar, Ecu.fwdCamera, Ecu.vsa, Ecu.electricBrakeBooster]
       if ecu_type == Ecu.esp and candidate in [TOYOTA.RAV4, TOYOTA.COROLLA, TOYOTA.HIGHLANDER] and found_version is None:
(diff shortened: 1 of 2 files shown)


pd0wm:
Nice!

We also need to make a change to fw_versions.py to make sure we ignore ECUs that are not important like we did when they were still unknown.

https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py#L97
Let's change that to `if ecu_type not in ESSENTIAL_ECUS`, where `ESSENTIAL_ECUS = [engine, eps, esp, vsa, fwRadar, fwdCamera, brakeBooster]`. Basically anything that has to do with ACC, steering, gas and brakes.

jpancotti:
> Nice!
> 
> We also need to make a change to fw_versions.py to make sure we ignore ECUs that are not important like we did when they were still unknown.
> 
> https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py#L97
> Let's change that to `if ecu_type not in ESSENTIAL_ECUS`, where `ESSENTIAL_ECUS = [engine, eps, esp, vsa, fwRadar, fwdCamera, brakeBooster]`. Basically anything that has to do with ACC, steering, gas and brakes.

done

pd0wm:
Just merged your cereal PR and updated master. Can you rebase your branch to latest master?

jpancotti:
> Just merged your cereal PR and updated master. Can you rebase your branch to latest master?

done

pd0wm:
The code is broken. See linter.

Also I also see changes to the submodules. Did you rebase properly?

pd0wm:
Looks good! 

I'll run a fingerprint test on all the versions that we got back to make sure there are no regressions, and then I'll merge it.

PR #774, opened 2019-08-01
Fix auto resume for Honda Bosch, less spam

cfranyota:
fix for Auto resuming on Honda Bosch radar. Spamming resume all the time causing jerking for no reason in heavy traffic, this change keeps car in a "Stopped" state. Car resumes like normal once the lead car actually starts driving forward. changes made to carcontroller.py and carstate.py.
I'm not the author of these fixes, but i always put them in my fork because it's too jerky without it.

PR #1095, opened 2020-02-14
Resume improvement for Bosch Civic's & CRVs

cfranyota:
-I am not the author nor did I write this improvement. Not sure who created this.

Anyway, it keeps the car in a "Stopped" state when stopped, then presses the resume button once when the lead car starts moving. Instead of the hurky jerky spam resume making you inch forward every two seconds.

-This is sort of a temp fix until Bosch gets long control.


gregjhogan:
similar discussion from 6 months ago (similar code in last comment)
https://github.com/commaai/openpilot/pull/774

PR #1096, opened 2020-02-14
Increase days of no Internet connectivity to 30

ErichMoraga:
Making good on @geohot's promise...
https://discordapp.com/channels/469524606043160576/524328425415245827/664532003382755349


geohot:
Instead, we made it so you can reboot to reset this. Fair?

ErichMoraga:
Rebooting is a fair compromise, and should encourage basic users to get their devices online regularly.  Next time someone complains about the warning, I'll advise they reboot, and if that doesn't do the trick, we'll investigate accordingly.

geohot:
Sounds good

PR #1108, opened 2020-02-16
save cpu in thermald

adeebshihadeh:
In offroad, cpu usage goes from 4.01% to 1.7%


pd0wm:
I'm fine with lowering to every 10 seconds too. How much does that save?

PR #1110, opened 2020-02-17
Chrysler: carstate returns capnp struct directly 

rbiasini:
incremental wrt https://github.com/commaai/openpilot/pull/1107

process replay test failure is expected and due to:
- float32 vs float64 calc propagation
- steeringTorqueEps was mistakenly no logged before

@pd0wm, could you please regen the test ref? Next PR will be for all cars, just wanted to prototype the change for only one car for now.


selfdrive/car/chrysler/interface.py
```diff
@@ -220,8 +179,13 @@ def update(self, c, can_strings):
...
     self.brake_pressed_prev = ret.brakePressed
     self.cruise_enabled_prev = ret.cruiseState.enabled
+    self.left_blinker_prev = ret.leftBlinker
+    self.right_blinker_prev = ret.rightBlinker
+
+    # copy back carState packet to CS
+    self.CS.out = ret.as_reader()
 
-    return ret.as_reader()
+    return copy.copy(self.CS.out)
```
pd0wm: Why do we need this copy?

selfdrive/car/chrysler/interface.py
```diff
@@ -113,78 +116,34 @@ def update(self, c, can_strings):
...
-    ret.cruiseState.available = self.CS.main_on
-    ret.cruiseState.speedOffset = 0.
-    ret.cruiseState.standstill = False
+    ret.steeringRateLimited = self.CC.steer_rate_limited if self.CC is not None else False
 
     # TODO: button presses
     buttonEvents = []
 
-    if self.CS.left_blinker_on != self.CS.prev_left_blinker_on:
+    if ret.leftBlinker != self.left_blinker_prev:
```
pd0wm: Do we need button events for the blinker? We don't use them anywhere? Maybe in a different PR remove them for all cars?

rbiasini: yeah, noticed that too. Will remove in another PR.

selfdrive/car/chrysler/interface.py
rbiasini: `CS.out` is used by carcontroller and wanted to avoid unexpected mods that might happen if the output of this method is modified.

pd0wm: Can you even make mods to a capnp reader (you call as_reader above it)?

rbiasini: yeah, you can't. I thought the as_reader was just to get rid of the traversal limit. Ok, I'll remove the copy.

PR #1112, opened 2020-02-17
cat process replay diff

### LINKED ISSUE #1111: diff.txt not uploaded in process_replay test in CI

adeebshihadeh:
Fixes #1111


pd0wm:
Thanks! That was quick :)

PR #1114, opened 2020-02-17
0.7.3 Release

pd0wm:


selfdrive/manager.py
```diff
@@ -289,6 +289,15 @@ def prepare_managed_process(p):
       subprocess.check_call(["make", "clean"], cwd=os.path.join(BASEDIR, proc[0]))
       subprocess.check_call(["make", "-j4"], cwd=os.path.join(BASEDIR, proc[0]))
 
+
+def join_process(process, timeout):
+  # Process().join(timeout) will hang due to a python 3 bug: https://bugs.python.org/issue28382
+  # We have to poll the exitcode instead
+  t = time.time()
+  while time.time() - t < timeout and process.exitcode is None:
+    time.sleep(0.001)
```
geohot: Should it actually join the process after this? Otherwise it's an orphan PID or something.

selfdrive/thermald.py
```diff
@@ -192,10 +192,12 @@ def thermald_thread():
     if health is not None:
       usb_power = health.health.usbPowerMode != log.HealthData.UsbPowerMode.client
 
-    try:
-      network_type = get_network_type()
-    except subprocess.CalledProcessError:
-      pass
+    # get_network_type is an expensive call. update every 10s
```
geohot: This really shouldn't live in thermald if it's slow. Mods like this are a bad idea.

geohot:
Why does the commit have the name "openpilot v0.7.2 release"?

selfdrive/thermald.py
pd0wm: the timing of thermald is not that important. It's also not that slow, we were talking a few percent cpu, but that's just unnecessary when offroad.

selfdrive/manager.py
pd0wm: After calling this function manager calls `os.kill` on the process, so it won't be orphaned. Or did you want to move that into the helper function?

pd0wm:
Fixing commit name!

andyh2 (approved):
LGTM

selfdrive/thermald.py
geohot: Oh, it's an offroad thing for power. Makes sense, but I still dislike mods like this.

PR #1117, opened 2020-02-18
Disable Power Down option for desk devices

geohot:
Non UI parameter.

cd /data/openpilot/common
'./params.py DisablePowerDown 1' to use

### DIFF
--- a/common/params.py
+++ b/common/params.py
@@ -22,6 +22,7 @@ file in place without messing with <params_dir>/d.
 """
 import time
 import os
+import string
 import binascii
 import errno
 import sys
@@ -409,7 +410,7 @@ if __name__ == "__main__":
       pp = params.get(k)
       if pp is None:
         print("%s is None" % k)
-      elif all(c < 128 and c >= 32 for c in pp):
+      elif all(chr(c) in string.printable for c in pp):
         print("%s = %s" % (k, pp))
       else:
         print("%s = %s" % (k, binascii.hexlify(pp)))
--- a/common/params.py
+++ b/common/params.py
@@ -22,6 +22,7 @@ file in place without messing with <params_dir>/d.
 """
 import time
 import os
+import binascii
 import errno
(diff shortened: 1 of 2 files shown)


pd0wm (approved):

common/params.py
```diff
@@ -407,10 +409,10 @@ def f(key, val):
       pp = params.get(k)
       if pp is None:
         print("%s is None" % k)
-      elif all(ord(c) < 128 and ord(c) >= 32 for c in pp):
+      elif all(c < 128 and c >= 32 for c in pp):
```
pd0wm: could use `chr(c) in string.printable`

selfdrive/boardd/boardd.cc
```diff
@@ -392,12 +392,21 @@ void can_health(PubSocket *publisher) {
...
+      // reset the counter to not continually poll the parameter
+      no_ignition_cnt = 0;
+    } else {
+      printf("TURN OFF CHARGING!\n");
+      pthread_mutex_lock(&usb_lock);
+      libusb_control_transfer(dev_handle, 0xc0, 0xe6, (uint16_t)(cereal::HealthData::UsbPowerMode::CLIENT), 0, NULL, 0, TIMEOUT);
+      pthread_mutex_unlock(&usb_lock);
+      printf("POWER DOWN DEVICE\n");
+      system("service call power 17 i32 0 i32 1");
+    }
```
pd0wm: Where do you free `disbale_power_down`?

geohot:
Fixed free, in general, we should move code like that to C++. Was stupidly thinking it would go out of scope.

Also, will power down right away if the ignition timer is expired and you clear the param. The polling shouldn't be expensive.

common/params.py
damianzim: `string.printable` returns some (white) ASCII characters with value under 32

PR #1122, opened 2020-02-18
WIP: add steerConfigIndex to interface for Honda EPS config

jpancotti:
Honda EPS fw uses tables with 7 rows to define torque input and output. Deadzones variy depending on row choice. Row choice is available for parsing via STEER_CONFIG_INDEX in dbc.

TODO:
Figure out how to get STEER_CONFIG_INDEX into get_params(). I don’t think we need to pass the entire CarState, since this is only something that needs to be read once. It might make more sense to read eps config at boot and pass it in as a subversion of the eps fw.

Added steerConfigIndex
Added values for CR-V
Recombined Civic and Civic Bosch (don’t need them separate with the scaled back interp)
Added values for Insight to prepare for upstreaming with fp 2.0

### DIFF
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -2,7 +2,7 @@ from collections import namedtuple
 from cereal import car
 from common.realtime import DT_CTRL
 from selfdrive.controls.lib.drive_helpers import rate_limit
-from common.numpy_fast import clip
+from common.numpy_fast import clip, interp
 from selfdrive.car import create_gas_command
 from selfdrive.car.honda import hondacan
 from selfdrive.car.honda.values import CruiseButtons, CAR, VISUAL_HUD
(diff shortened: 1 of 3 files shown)


gregjhogan:
what trims of which vehicles will benefit from this change? I have seen trims using different indexes, but the max was always the same.

jpancotti:
> what trims of which vehicles will benefit from this change? I have seen trims using different indexes, but the max was always the same.

The steer max is the same for most, it is the deadzone in stock fw that varies. Not sure which trims use which indexes, id have to look at fleet config index data. Worst one is insight, 3 diff deadzones depending on config index.

gregjhogan:
Yeah by max I meant the start of the dead zone.

So it is possible these changes have zero net effect (because the rows that you are accounting for are never used).  To make a call on whether it is worth increasing the complexity we need more information (routes showing the rows with different values being used).

jpancotti:
> Yeah by max I meant the start of the dead zone.
> 
> So it is possible these changes have zero net effect (because the rows that you are accounting for are never used). To make a call on whether it is worth increasing the complexity we need more information (routes showing the rows with different values being used).

ok, I will close this for now. Once we upstream insight maybe you can check the fleet data on your side for variation in config index. I will save the data for dead zone by config index for all the fw versions I have so far.

PR #1125, opened 2020-02-19
cancel lane change if driver took over

xx979xx:


pd0wm:
In 0.7.4 the lane change will be handled completely by the model. The lane change will be started by a single pulse to the model. The model will decide when the lane change is done, so there will be no more way to cancel it while in progress. Therefore merging this change now wouldn't make sense.

PR #1129, opened 2020-02-19
Abstract common events + event cleanup

adeebshihadeh:
- remove unnecessary junk/dead code
- `LOW_LOWEST` -> `LOWER`
- abstract common events

not ready to merge. meant to open as a draft


quillford:
process_replay is only failing because of the events order. it passes with some local changes that sort carEvents and carState.events

```
diff --git a/selfdrive/test/process_replay/compare_logs.py b/selfdrive/test/process_replay/compare_logs.py
index 82e701a..0c73914 100755
--- a/selfdrive/test/process_replay/compare_logs.py
+++ b/selfdrive/test/process_replay/compare_logs.py
@@ -47,6 +47,23 @@ def compare_logs(log1, log2, ignore=[]):
       print(msg1, msg2)
       assert False, "msgs not aligned between logs"
 
+    if msg1.which() == "carState":
+      msg1 = msg1.as_builder()
+      msg1.carState.events = sorted(msg1.carState.events, key=lambda evt: evt.name)
+      msg1 = msg1.as_reader()
+      msg2 = msg2.as_builder()
+      msg2.carState.events = sorted(msg2.carState.events, key=lambda evt: evt.name)
+      msg2 = msg2.as_reader()
+
+
+    if msg1.which() == "carEvents":
+      msg1 = msg1.as_builder()
+      msg1.carEvents = sorted(msg1.carEvents, key=lambda evt: evt.name)
+      msg1 = msg1.as_reader()
+      msg2 = msg2.as_builder()
+      msg2.carEvents = sorted(msg2.carEvents, key=lambda evt: evt.name)
+      msg2 = msg2.as_reader()
+
     msg1_bytes = remove_ignored_fields(msg1, ignore).as_builder().to_bytes()
     msg2_bytes = remove_ignored_fields(msg2, ignore).as_builder().to_bytes()
```

selfdrive/car/interfaces.py
```diff
@@ -33,6 +34,16 @@ def update(self, c, can_strings):
   def apply(self, c):
     raise NotImplementedError
 
+  def create_common_events(self, c, cs_out):
```
rbiasini: really like this abstraction. A couple of things:
- since does not depend on self, a staticmethod could be more appropriate.
- I think a lot more standard events can be abstracted: espDisable, wrongMode, wrongGear, reverseGear. 
- With some more work, some other events can be commonized too (I can do this). 

what do you think?

quillford: That's why this was supposed to be a draft :)

There's definitely more that can be abstracted here. I went back and did a few quick ones. The other events have a couple minor differences between the interfaces

quillford:
replay failures:
- subaru didn't check for `wrongCarMode` before. difference is on first frame where the default value is used for cruiseState.available until it gets the msg on can
- order of events has changed. lists are ordered in python, so it doesn't make sense to change process_replay to ignore that, so I've been running the test locally using the diff above just for this

I've been running process_replay locally as I go and everything's the same, but VW doesn't have a route yet so it should get some extra scrutiny

rbiasini:
@quillford , looks good to me. Can you fix the linter?

jyoung8607:
> I've been running process_replay locally as I go and everything's the same, but VW doesn't have a route yet so it should get some extra scrutiny

What behaviors do you want to see in the way of route data to play under testing? I can supply from my Golf R, I already have to drive a new route for the Panda safety work.

quillford:
linter should be happy now

rbiasini:
@jyoung8607 , just supplied a VW segment myself. @pd0wm will include it in the regression test.
Willem, can you please also regen the test ref? Then this PR is good to go.

pd0wm:
@quillford just pushed VW tests to master. Can you rebase? Then I'll generate a new reference file for this PR and we can merge it.

selfdrive/ui/ui.cc
```diff
@@ -6,7 +6,6 @@
 #include <sys/mman.h>
 #include <sys/resource.h>
 
-#include <json.h>
```
pd0wm: Can we also remove json form the libs section in the SConscript?

selfdrive/car/interfaces.py
```diff
@@ -48,6 +49,30 @@ def update(self, c, can_strings):
...
+      events.append(create_event('reverseGear', [ET.NO_ENTRY, ET.IMMEDIATE_DISABLE]))
+    if not cs_out.cruiseState.available:
+      events.append(create_event('wrongCarMode', [ET.NO_ENTRY, ET.USER_DISABLE]))
+
+    # TODO: move this stuff to the capnp strut
+    if getattr(self.CS, "steer_error", False):
+      events.append(create_event('steerUnavailable', [ET.NO_ENTRY, ET.IMMEDIATE_DISABLE, ET.PERMANENT]))
+    elif getattr(self.CS, "steer_warning", False):
+      events.append(create_event('steerTempUnavailable', [ET.NO_ENTRY, ET.WARNING]))
+    if getattr(self.CS, "esp_disabled", False):
```
pd0wm: in VW this is called `espDisabled`. Am I missing some conversion?

pd0wm (changes requested):

pd0wm (approved):

selfdrive/car/interfaces.py
quillford: No, I just missed this one. I think if we're standardizing these names in the CarState class then we should probably just add them to the struct

Also, I think process_replay is fast enough now to run on whole routes, and it should catch stuff like these if given interesting enough routes.

pd0wm:
`446c1fd961444b88cceecb962ca336aaf92e3ee4` can we used as `ref_commit` to fix the tests.

pd0wm:
Ah, https://github.com/commaai/openpilot/pull/1138 was not merged into your branch. I'll generate a new test ref.

pd0wm:
`0491cad238025a44b7fc636cf03efc27d80dcdae`

pd0wm:
```Traceback (most recent call last):
  File "/usr/local/pyenv/versions/3.7.3/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/usr/local/pyenv/versions/3.7.3/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/home/batman/openpilot/selfdrive/controls/controlsd.py", line 591, in main
    controlsd_thread(sm, pm, logcan)
  File "/home/batman/openpilot/selfdrive/controls/controlsd.py", line 532, in controlsd_thread
    CS, events, cal_perc, mismatch_counter, can_error_counter = data_sample(CI, CC, sm, can_sock, state, mismatch_counter, can_error_counter, params)
  File "/home/batman/openpilot/selfdrive/controls/controlsd.py", line 73, in data_sample
    CS = CI.update(CC, can_strs)
  File "/home/batman/openpilot/selfdrive/car/subaru/interface.py", line 66, in update
    be.type = ButtonType.accelCruise
NameError: name 'ButtonType' is not defined
```

PR #1131, opened 2020-02-19
Abstract CarInterface init

adeebshihadeh:


selfdrive/car/volkswagen/interface.py
```diff
@@ -10,32 +10,17 @@
...
 
-    # *** init the major players ***
-    self.CS = CarState(CP, CANBUS)
-    self.VM = VehicleModel(CP)
-    self.pt_cp = get_mqb_pt_can_parser(CP, CANBUS)
-    self.cam_cp = get_mqb_cam_can_parser(CP, CANBUS)
-
+    self.CC = None
     # sending if read only is False
     if CarController is not None:
```
rbiasini: we can probably abstract this if too.

selfdrive/car/gm/interface.py
```diff
@@ -2,42 +2,20 @@
...
-from selfdrive.car.gm.values import DBC, CAR, Ecu, ECU_FINGERPRINT, \
+from selfdrive.car.gm.values import CAR, Ecu, ECU_FINGERPRINT, \
                                     SUPERCRUISE_CARS, AccState, FINGERPRINTS
 from selfdrive.car.gm.carstate import CarState, CruiseButtons, get_powertrain_can_parser
 from selfdrive.car import STD_CARGO_KG, scale_rot_inertia, scale_tire_stiffness, is_ecu_disconnected, gen_empty_fingerprint
 from selfdrive.car.interfaces import CarInterfaceBase
 
 ButtonType = car.CarState.ButtonEvent.Type
 
-class CanBus(CarInterfaceBase):
```
rbiasini: oh woah, this inheritance made no sense.

quillford: yeah, then instances were made and copied all over the place

selfdrive/car/interfaces.py
```diff
@@ -4,14 +4,30 @@
...
+from selfdrive.controls.lib.vehicle_model import VehicleModel
 
 GearShifter = car.CarState.GearShifter
 
 # generic car and radar interfaces
 
 class CarInterfaceBase():
-  def __init__(self, CP, CarController):
-    pass
+  def __init__(self, CP, CarController, CarState, get_can_parser, get_cam_can_parser=None):
```
rbiasini: get_can_parser and get_cam_can_parser can just be static methods of the CarState class.
No need to do it now, but a good clean up for later.

rbiasini: Indeed, it would be better to have the parent class having the same init API as the subclasses, so we can avoid
https://github.com/commaai/openpilot/pull/1131/files#diff-93874011152ca529debcfce500178461R16

rbiasini:
Even though the VW precess replay regr test is still missing, I'm merging this since the VW changes are very minimal. 
Note that runtime errors would still be caught by the test_car_models regr test, which has VW routes.

PR #1132, opened 2020-02-19
Ford: add models. Placeholders for Fingerprints and DBC. 

roxasthenobody98:
Added models to values.py. Made placeholders for fingerprints and dbc's


selfdrive/car/ford/values.py
```diff
@@ -5,18 +5,113 @@
...
+  CAR.MKT: [{
+  }], 
+  CAR.MKX: [{
+  }], 
+  CAR.MKZ: [{
+  }], 
+  CAR.NAVIGATOR: [{
+  }], 
+  CAR.NAUTILUS: [{
+  }],
```
roxasthenobody98: Uncomment to add fingerprint. Failing validation.

roxasthenobody98:
Commented out blank lines to pass validation. Data is still present, so fingerprints and DBC's can be added.

pd0wm:
There is no point adding unsupported cars now, since it will add unused lines of code that need to be maintained. The cars can be added one by one when there is a port.

PR #1133, opened 2020-02-20
LeadTwo chevron

arne182:
Draw also second car indicator from LeadTwo Data

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -109,7 +109,6 @@ static void draw_chevron(UIState *s, float x_in, float y_in, float sz,
 }
 
 static void draw_lead(UIState *s, float d_rel, float v_rel, float y_rel){
-    const UIScene *scene = &s->scene;
     // Draw lead car indicator
     float fillAlpha = 0;
     float speedBuff = 10.;
@@ -126,8 +125,6 @@ static void draw_lead(UIState *s, float d_rel, float v_rel, float y_rel){
 }
 
 static void ui_draw_lane_line(UIState *s, const model_path_vertices_data *pvd, NVGcolor color) {
-  const UIScene *scene = &s->scene;
-
   nvgSave(s->vg);
   nvgTranslate(s->vg, 240.0f, 0.0); // rgb-box space
   nvgTranslate(s->vg, -1440.0f / 2, -1080.0f / 2); // zoom 2x
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -409,7 +409,7 @@ static void ui_draw_world(UIState *s) {
   if (scene->lead_status) {
     draw_lead(s,scene->lead_d_rel,scene->lead_v_rel,scene->lead_y_rel);
   }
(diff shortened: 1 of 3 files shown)


arne182:
@pd0wm 
do you really want 
  if ((scene->lead_status2) &&(fabs(scene->lead_d_rel - scene->lead_d_rel2)>0.1)||(fabs(scene->lead_y_rel - scene->lead_y_rel2)>0.1))) {
as condition?

selfdrive/ui/ui.cc
```diff
@@ -357,11 +357,17 @@ void handle_message(UIState *s, Message * msg) {
     struct cereal_RadarState datad;
     cereal_read_RadarState(&datad, eventd.radarState);
     struct cereal_RadarState_LeadData leaddatad;
+    struct cereal_RadarState_LeadData leaddatae;
     cereal_read_RadarState_LeadData(&leaddatad, datad.leadOne);
+    cereal_read_RadarState_LeadData(&leaddatae, datad.leadTwo);
```
pd0wm: Can you reuse the same struct? Just do the `cereal_read_RadarState_LeadData(&leaddatae, datad.leadTwo);` after you're done with the first lead.

selfdrive/ui/paint.cc
```diff
@@ -404,6 +404,21 @@ static void ui_draw_world(UIState *s) {
     draw_chevron(s, scene->lead_d_rel+2.7, scene->lead_y_rel, 25,
                   nvgRGBA(201, 34, 49, fillAlpha), nvgRGBA(218, 202, 37, 255));
   }
+  if (scene->lead_status2) {
+    // Draw lead2 car indicator
```
pd0wm: Can you refactor this duplicated code into a `draw_lead(scene, d_rel, v_rel, y_rel)` function?

pd0wm (changes requested):
Yes. That condition sounds fine. Otherwise the leads are not distinct.

selfdrive/ui/ui.cc
arne182: Done. Even though you have a spelling mistake in your comment

selfdrive/ui/paint.cc
arne182: let us see if the tests come back with a tick

selfdrive/ui/paint.cc
```diff
@@ -108,6 +108,23 @@ static void draw_chevron(UIState *s, float x_in, float y_in, float sz,
   nvgRestore(s->vg);
 }
 
+static void draw_lead(UIState *s, float d_rel, float v_rel, float y_rel){
+    const UIScene *scene = &s->scene;
```
pd0wm: scene is unused in this function.

HaraldSchafer:
> @pd0wm
> do you really want
> if ((scene->lead_status2) &&(fabs(scene->lead_d_rel - scene->lead_d_rel2)>0.1)||(fabs(scene->lead_y_rel - scene->lead_y_rel2)>0.1))) {
> as condition?

There needs to be some check that they are different otherwise might be a lot of almost overlapping chevrons. Could also just do distance maybe (fabs(scene->lead_d_rel - scene->lead_d_rel2)>3)

selfdrive/ui/paint.cc
arne182: didn't know if this is needed. The other functions had it. Thanks for the explaination

pd0wm (approved):

PR #1136, opened 2020-02-20
get_can_parser and get_cam_can_parser are now standard static methods

rbiasini:


rbiasini:
@quillford , this is in reference to https://github.com/commaai/openpilot/pull/1131#discussion_r381737582

Makes sense?

quillford:
Yeah, this is definitely cleaner and the way it should have been

PR #1139, opened 2020-02-20
Fix artifact upload in CI when step fails

adeebshihadeh:


rbiasini:
will this fix the missing diff.txt from process replay test?

quillford:
Yeah, it should always copy the artifacts now. I think it wasn't copying them before due to the step failing which wouldn't let it run subsequent commands in the step (i.e. the copying)

.github/workflows/test.yaml
```diff
@@ -97,6 +97,9 @@ jobs:
         run: |
           CONTAINER_NAME="tmppilot_${GITHUB_SHA}"
           docker run --shm-size 1G --name ${CONTAINER_NAME} tmppilot /bin/sh -c "cd /tmp/openpilot/selfdrive/test/process_replay && CI=1 ./test_processes.py"
+      - name: Copy diff
+        if: always()
+        run: |
```
pd0wm: CONTAINER_NAME is not set

PR #1141, opened 2020-02-20
See if HONDA.ACCORDH is fixed

arne182:


arne182:
Drive needs to be updated. Anyone with a ACCORDH drive?

pd0wm:
Seems like that route was never copied over into the new data bucket since it was commented out. Think we have to find a new route.

Weird why this was commented out, since we never added support for the Insight anyway.

K-Shay:
@arne182 are you still looking for a long ACCORDH drive?  Will be driving home from Dallas to San Diego next week.  around 1100ish miles over two days stopping in El Paso.

PR #1147, opened 2020-02-21
Updates to Volkswagen port

jyoung8607:
Join the negative LoC PR party and remove unreachable code from the Volkswagen port.

Under lateral-only/stock ACC, this never did anything because the exact same status variable drives OP engagement. If a radar fault already existed when the driver hit Set, OP would never see an engage attempt. If a radar fault happened while engaged, OP would disengage along with stock ACC. Under future OP longitudinal control, we won't be taking input from the stock ACC radar as it doesn't output points data that we're aware of.


rbiasini:
makes sense. Will wait for tests to finish and then I'll merge when you get out of "draft" mode.

jyoung8607:
I'm unclear as to why the replay test is failing, this should be a null-effect change. I think this is something you recently added, any insight?

rbiasini:
not exactly null effect: at the very first cycle, ACC_STATUS is zero, which before was triggering the radar fault event (in case a ACC_06 update wasn't immediately available).

jyoung8607:
Aha, nicely spotted!

Sit on this PR until tomorrow then? I'm going to do a test drive tomorrow anyway with the torque and ACC_06->TSK_06 changes we discussed, I can include this change as well.

pd0wm:
Yeah, let me know if it drives fine. If it's good I'll update the test reference.

jyoung8607:
This PR now includes a few more changes since I did one test-drive with all of them included:

- The aforementioned extra radar fault is removed
- ACC state is now tracked from TSK_06 instead of ACC_06 for install location neutrality
- Torque cap is raised from 2.5Nm to 3.0Nm per discussion with @rbiasini 

This PR is designed to be merged together with commaai/panda#462.

rbiasini (approved):

PR #1148, opened 2020-02-21
only run system() command when vol has changed

eFiniLan:
reduce the frequency of calling "service call" command.


geohot:
Can you fix the indentation? Should be 2 spaces.

PR #1153, opened 2020-02-22
Process replay enhancements

### LINKED ISSUE #1116: test_processes.py should fail if a car make is not tested
**Describe the bug**
realized VW port isn't tested by process replay regression test. Would be nice if the test would automatically fail if a car port is missing from the test.

adeebshihadeh:
- [x] args to blacklist procs/cars 
- [x] args for whitelisting procs/cars
- [x] arg for ignoring a packet or field
- [x] check to make sure each car brand is tested
- [x] check for enable on each route

fixes #1116


rbiasini:
great enhancements

quillford:
`0375fdf7b1ce594d|2019-06-13--08-32-25--3` (honda bosch) fails the enable check so it needs to be replaced with a new route

rbiasini:
Can we turn off the enable check so that we can merge these changes? We should change routes, but it'll probably take some time.

selfdrive/test/process_replay/test_processes.py
```diff
@@ -1,117 +1,164 @@
...
+  parser.add_argument("--whitelist-procs", type=str, nargs="*", default=[],
+                        help="Whitelist given processes from the test (e.g. controlsd)")
+  parser.add_argument("--whitelist-cars", type=str, nargs="*", default=[],
+                        help="Whitelist given cars from the test (e.g. HONDA)")
+  parser.add_argument("--blacklist-procs", type=str, nargs="*", default=[],
+                        help="Blacklist given processes from the test (e.g. controlsd)")
+  parser.add_argument("--blacklist-cars", type=str, nargs="*", default=[],
+                        help="Blacklist given cars from the test (e.g. HONDA)")
+  parser.add_argument("--ignore-fields", type=str, nargs="*", default=[],
+                        help="Extra fields or msgs to ignore (e.g. carState.events)")
```
rbiasini: Can I also ignore a certain log type entirely? eg: CarEvents.

quillford: I thought the ignore fields would work for this but I guess it doesn't for carEvents when the list isn't initialized or the length is 0. I added an `--ignore-msgs` arg for this

PR #1158, opened 2020-02-24
simplify chrysler checksum

rbiasini:
very minor change. cleanup before porting it to panda safety


adhintz:
If you do want to change the chryslercan.calc_checksum argument to not have the last byte, then don't forget to also update its unit test at `selfdrive/car/chrysler/test_chryslercan.py`

rbiasini:
@adhintz , yes thanks, but process replay was failing too, so I just reverted. Have to look into it.

PR #1163, opened 2020-02-25
Add brake press disengagement in Chrysler

rbiasini:
Refs can be updated:
diffs below, only due to newly present brakePressed event
```
***** results for segment b6e1317e1bfbefa6|2019-07-06--04-05-26--5 *****
        controlsd
                ['carState', 'events', 2, 'name']: 732
                ['carState', 'events', 2, 'noEntry']: 214
                ['carState', 'events', 2, 'userDisable']: 732
                ['carState', 'events', 2, 'warning']: 732
                ['carState', 'events', 3, 'name']: 214
                ['carState', 'events', 3, 'noEntry']: 214
                carState.events: 732
        radard
        plannerd
        calibrationd
        dmonitoringd
```


adhintz:
Looks good.

I hadn't added this because on Chrysler pressing the brake cancels the stock ACC. So on brake press, ret.cruiseState.enabled goes to false which disengages due to
```
    elif not ret.cruiseState.enabled:
      events.append(create_event('pcmDisable', [ET.USER_DISABLE]))
```

rbiasini:
@adhintz , correct, that was the assumption for all other "lateral only" cars as well. However, we are standardizing entry and exit conditions and we decided to not rely on the cancellation from the stock system.

PR #417, opened 2018-11-02
Add Honda Insight

energee:
We should be able to test this next week

PR #1166, opened 2020-02-27
Add 2019 Honda Insight

theantihero:
Related to closed PR #417 due to fp issue. Was able to get fw ids from the query script.


pd0wm:
Do you have a route we can use for the regression tests? Please put one in test_car_models with a fixed fingerprint: like https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_car_models.py#L293

pd0wm:
And you also need to add the fingerprint to the list of fingerprints since that is still used in the `is_ecu_disconnected` fingerprint. To make sure it doesn't interfere with other cars fingerprinting add a list with `IGNORED_FINGERPRINTS` like https://github.com/commaai/openpilot/blob/master/selfdrive/car/toyota/values.py#L269

theantihero:
@pd0wm Hey, thanks for the additional comments, I added the actual fp and added it to ignored list.
I will provide a route later today. Does it have to be a stock repo drive or can a dirty fork be used?

pd0wm:
The route doesn’t matter

pd0wm:
Pushed a fix to the fingerprint test to master:https://github.com/commaai/openpilot/commit/6eca88e9bcc4c6d32304444c6565ca6be0d77bbb

Can you rebase?

selfdrive/test/test_car_models.py
```diff
@@ -191,6 +191,10 @@ def get_route_log(route_name):
   "690c4c9f9f2354c7|2018-09-15--17-36-05": {
     'carFingerprint': HONDA.ACCORDH,
     'enableCamera': True,
+  },
+    "fb51d190ddfd8a90|2020-02-25--14-43-43": {
+    'carFingerprint': HONDA.INSIGHT,
+    'enableCamera': True,
```
pd0wm: Add `'fingerprintSource': 'fixed'`

theantihero: Will do

theantihero (commented):
> Submodule cereal updated from 20c7fd to 25eaf9
> Submodule panda updated from dad585 to d7f119

I think these are the correct changes, please correct me if I'm wrong. If I checked correctly those are latest on master.

selfdrive/car/honda/values.py
```diff
@@ -651,6 +673,7 @@ class CAR:
   CAR.PILOT: dbc_dict('honda_pilot_touring_2017_can_generated', 'acura_ilx_2016_nidec'),
   CAR.PILOT_2019: dbc_dict('honda_pilot_touring_2017_can_generated', 'acura_ilx_2016_nidec'),
   CAR.RIDGELINE: dbc_dict('honda_ridgeline_black_edition_2017_can_generated', 'acura_ilx_2016_nidec'),
+  CAR.INSIGHT: dbc_dict('honda_insight_ex_2019_can_generated.dbc', None),
```
pd0wm: Don't add .dbc exension. This is why the test hangs.

theantihero: Ah.. Happy little accident, my mistake.

PR #1170, opened 2020-02-28
better text on UI before max speed is set

berno22:
user-friendly text


valish:
@berno22 Thank you for suggesting an edit here. I agree that the "N/A" max speed status can be updated to be more user friendly, however I'm not certain that "SET" is the correct term here. "SET" can be interpreted as a state itself, as in the max speed is set. Also, I believe some may interpret it as an action to "SET" by touching the device interface itself. A more direct and appropriate state would be "not set", although that does not fit the current UI. This topic should remain open for discussion, but will be closing this pull request for now.

PR #1180, opened 2020-02-29
Add missing deps for ubuntu_setup.sh

littlemountainman:


tools/ubuntu_setup.sh
```diff
@@ -40,7 +40,15 @@ sudo apt-get update && sudo apt-get install -y \
...
     sudo \
     vim \
-    wget
+    wget \
+    gcc-arm-none-eabi \
+    libgles2-mesa-dev \
+    libbz2-dev \
+    scons 
+    
+# Install scons for py2 and py3, Rest can be handled with pyenv
```
geohot: Why can't scons be installed by pipenv?

tools/ubuntu_setup.sh
```diff
@@ -61,7 +69,9 @@ fi
 source ~/.bashrc
 if [ -z "$OPENPILOT_ENV" ]; then
   echo "source $HOME/openpilot/tools/openpilot_env.sh" >> ~/.bashrc
+  
   source ~/.bashrc
+  exec bash
```
geohot: I shouldn't need to do this, the source should handle it.

tools/ubuntu_setup.sh
```diff
@@ -40,7 +40,15 @@ sudo apt-get update && sudo apt-get install -y \
     screen \
     sudo \
     vim \
-    wget
+    wget \
+    gcc-arm-none-eabi \
+    libgles2-mesa-dev \
```
geohot: These are already in there

PR #1182, opened 2020-03-01
Adding error checking to only use the cached fp if they're valid

suspi:
Caused issues trying to pull carFw and carVin due to having older versions of OP running that had all these values cached.

### DIFF
--- a/selfdrive/car/car_helpers.py
+++ b/selfdrive/car/car_helpers.py
@@ -76,7 +76,7 @@ def fingerprint(logcan, sendcan, has_relay):
       if cached_params.carName == "mock":
         cached_params = None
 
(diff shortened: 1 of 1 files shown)


geohot (approved):

geohot:
I def support the firmware length one, @pd0wm, does the VIN at least return something on all cars?

pd0wm:
The param is `[TxType.CLEAR_ON_MANAGER_START, TxType.CLEAR_ON_PANDA_DISCONNECT],` So should clean on reboot and panda unplug. So not sure what happened.

Still think this is a good change in case the fingerprinting fails on the first start of the car, restarting the car should fix problems.

PR #1184, opened 2020-03-01
Restore openpilot backup if manager fails to run

suspi:
In the case where we install a bad build of openpilot, we are stuck with a black screen until we go in and fix it with ssh. This change swaps out the build with a backup kept in openpilot.bak so that we can get back to a running state. Workbench executes the following command by default when Install Openpilot is selected.

`cd /data;mv ./openpilot ./openpilot.bak;git clone %git_url% openpilot;cd openpilot;git checkout %git_branch%`

So newer users can rely on having a openpilot.bak available to failover into.

Also added redirecting the manager.py output to a log file to keep with the old build to diagnose what broke.

### DIFF
--- a/launch_chffrplus.sh
+++ b/launch_chffrplus.sh
@@ -86,10 +86,13 @@ function launch {
 
   # start manager
   cd selfdrive
-  ./manager.py
+  ./manager.py > runtime.log
 
-  # if broken, keep on screen error
-  while true; do sleep 1; done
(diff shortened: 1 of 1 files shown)


geohot:
This changes behavior in 2 ways that may not be desirable to everyone
1) I can't tmux attach to see manager
2) If I control c manager, it'll say it failed.
For these reasons, I'm not going to merge.

PR #1185, opened 2020-03-01
Make Nui tests work. Framereader moved to tools/clib

littlemountainman:
* Adjusted .pro script to find the .hpp file.

* Changed path in TestFramereader to find Framereader in tools/clib


tools/ubuntu_setup.sh
```diff
@@ -42,8 +42,7 @@ sudo apt-get update && sudo apt-get install -y \
     sudo \
     vim \
     wget \
-    gcc-arm-none-eabi
-
+    gcc-arm-none-eabi
```
geohot: Null change, please remove

tools/sim/start_carla.sh
```diff
@@ -15,5 +15,4 @@ if [ ! -d carla ]; then
 fi
 
 cd carla
-./CarlaUE4.sh Town04
-
+./CarlaUE4.sh
```
geohot: Why change this?

geohot:
Other then those 2 comments, LGTM

tools/sim/start_carla.sh
littlemountainman: It's set via the client, Town04 changes nothing, town gets loaded in bridge.py in line 89.

littlemountainman:
fixed first comment, explained second why Town04 was removed.

littlemountainman:
Can be merged now. imo

PR #1191, opened 2020-03-03
Ford: Disable Radar Interface

roxasthenobody98:
Testing confirms on non Stop/Go Fords, radar is connected directly to HS2 and cannot be intercepted. Stop/Go Ford's can be intercepted, but the ACC is able to go to 0 and resume anyway.


rbiasini:
why deleting the whole interface? a conditional case based on the presence of the radar on CAN would be better. See Honda.

roxasthenobody98:
There isnt much improvement to be made over stock.

pd0wm:
If openpilot can do longitudinal control we want to use that. In the future openpilot longitudinal will be significantly better than the stock one in detecting cut-ins, slowing down for turns, etc...

PR #1192, opened 2020-03-04
Refactor frame

valish:


selfdrive/ui/ui.cc
```diff
@@ -888,8 +970,15 @@ int main(int argc, char* argv[]) {
       set_awake(s, false);
     }
 
-    // Don't waste resources on drawing in case screen is off or car is not started.
-    if (s->awake && s->vision_connected) {
+    // manage hardware disconnect
```
pd0wm: We should handle this in boardd. It should not stop sending health packets (and therefore thermal) packets when you unplug a panda. I'll let you know when you can remove this.

pd0wm:
I see a lot of repeater color definitions in the code. Can we replace those by a define with a name of the color?

pd0wm:
Why did the jenkins test not run? Can you rebase to latest master so we see CPU usage in CI?

selfdrive/ui/sidebar.cc
```diff
@@ -0,0 +1,255 @@
+#include <stdio.h>
+#include <string.h>
+#include <math.h>
+#include "ui.hpp"
+
+static void ui_draw_sidebar_background(UIState *s) {
```
pd0wm: Can hasSidebar be passed in as an argument to all these functions? Then you only need to do the lookup once.

selfdrive/ui/sidebar.cc
```diff
@@ -0,0 +1,255 @@
...
+    home_btn_w, home_btn_h, 0, s->img_button_home, 1.0f);
+  nvgRect(s->vg, home_btn_xr, home_btn_y, home_btn_w, home_btn_h);
+  nvgFillPaint(s->vg, imgPaint);
+  nvgFill(s->vg);
+}
+
+static void ui_draw_sidebar_network_strength(UIState *s) {
+  const UIScene *scene = &s->scene;
+  bool hasSidebar = !s->scene.uilayout_sidebarcollapsed;
+  const int network_img_h = 27;
```
pd0wm: Use `nvgImageSize` instead of hardcoding image size?

selfdrive/ui/sidebar.cc
```diff
@@ -0,0 +1,255 @@
...
+static void ui_draw_sidebar_network_type(UIState *s) {
+  const UIScene *scene = &s->scene;
+  bool hasSidebar = !s->scene.uilayout_sidebarcollapsed;
+  const int network_x = hasSidebar ? 50 : -(sbr_w);
+  const int network_y = 273;
+  const int network_w = 100;
+  const int network_h = 100;
+  char network_type_str[32];
+
+  if (s->scene.networkType == NETWORKTYPE_NONE) {
```
pd0wm: Something along the lines of?
```c++
string names[] = {"--", "WiFi", "2G", "3G", "4G", "5G"};
if ( s->s->scene.networkType <= 5){
  snprintf(network_type_str, sizeof(network_type_str), "%s", names[idx]);
}
```

selfdrive/ui/ui.cc
```diff
@@ -403,35 +452,82 @@ void handle_message(UIState *s, Message * msg) {
...
-    }
   } else if (eventd.which == cereal_Event_liveMapData) {
     struct cereal_LiveMapData datad;
     cereal_read_LiveMapData(&datad, eventd.liveMapData);
     s->scene.map_valid = datad.mapValid;
+  } else if (eventd.which == cereal_Event_thermal) {
+    struct cereal_ThermalData datad;
+    cereal_read_ThermalData(&datad, eventd.thermal);
+
+    if (datad.networkType == cereal_ThermalData_NetworkType_none) {
```
pd0wm: Why this conversion? Can't you use the cereal struct values everywhere?

selfdrive/ui/ui.cc
```diff
@@ -403,35 +452,82 @@ void handle_message(UIState *s, Message * msg) {
...
+    auto polls = s->poller->poll(0);
+
+    if (polls.size() == 0)
+      break;
+
+    for (auto sock : polls){
+      Message * msg = sock->receive();
+      if (msg == NULL) continue;
+
+      if (s->vision_connected) {
```
pd0wm: Why is this here?

pd0wm: Maybe have the check_message function return if something was received, and use that to set awake?

selfdrive/ui/ui.hpp
```diff
@@ -38,6 +38,28 @@
 #define ALERTSIZE_MID 2
 #define ALERTSIZE_FULL 3
 
+#define NETWORKTYPE_NONE 0
```
pd0wm: Use capnp enums for all of this

PR #1193, opened 2020-03-04
removing ipas steering control option from toyota

rbiasini:


rbiasini:
Talked to Willem, good idea to merge this in.

PR #1194, opened 2020-03-04
Add input parameter to SubMaster.update() to wait for specific services

sshane:
I couldn't find the `messaging/__init__.py` file on master, so I just opened this temporary PR for now.

This adds the ability to specify a list of services to wait for when updating with SubMaster. I took out the poller simply because I didn't know if it was possible to identify which socket/service it is from the poller. I was debating if it was better or not to send the `wait_for` list when we initialize the SubMaster, or for better modularity, keep it inside the update function. So let me know if there's anything I can improve

### DIFF
--- a/cereal/messaging/__init__.py
+++ b/cereal/messaging/__init__.py
@@ -162,14 +162,16 @@ class SubMaster():
   def __getitem__(self, s):
     return self.data[s]
 
-  def update(self, timeout=1000):
+  def update(self, wait_for=None):
     msgs = []
-    for sock in self.poller.poll(timeout):
-      msgs.append(recv_one_or_none(sock))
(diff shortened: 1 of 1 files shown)


pd0wm:
This PR should be made against the `cereal` repo, that's where the code lives.

Taking out the poller would result in higher cpu usage. If you want this feature you only need to add the sockets that you want to block on to the poller.

ShaneSmiskol:
@pd0wm Thanks! I didn't know that repo existed. I saw the comment under `update_msgs()`: `# TODO: add optional input that specify the service to wait for` so I decided to try to tackle this todo. I first noticed it because of the use of the thermal sock in manager.py not using the SubMaster

PR #1215, opened 2020-03-08
LCA toggle live update

xx979xx:


pd0wm:
We try to have people not touch their comma 2 / eon while driving, so I don't want to add this.

PR #1218, opened 2020-03-09
abstract pedalPressed event

rbiasini:


selfdrive/car/gm/interface.py
```diff
@@ -173,10 +173,6 @@ def update(self, c, can_strings):
         events.append(create_event('speedTooLow', [ET.NO_ENTRY]))
       if self.CS.park_brake:
         events.append(create_event('parkBrake', [ET.NO_ENTRY, ET.USER_DISABLE]))
-      # disable on pedals rising edge or when brake is pressed and speed isn't zero
-      if (ret.gasPressed and not self.gas_pressed_prev) or \
-        (ret.brakePressed): # and (not self.brake_pressed_prev or ret.vEgo > 0.001)):
```
rbiasini: @pd0wm not sure why the second part of this line was commented out. This abstraction "uncomments" it and it's the only functional change. What do you think?

rbiasini: @vntarasov, @Jamezz, any recollection on why we chose to comment out part of that line?

vntarasov: @rbiasini This is for resume from a full stop, while holding a brake pedal, right?
At least for my volt, after touching the brake, PCM won't let openpilot to control gas until car is moving above ~18mph.
Back when this GM support was upstreamed, openpilot didn't detect PCM faults, the UI was still showing green, car slowly creeps forward, but never picks up speed above 3mph - very annoying experience.
But now that https://github.com/commaai/openpilot/pull/466 is merged, openpilot would show engagement fault in UI, so it's fine to land this change.

rbiasini: @vntarasov, Thanks. 
> This is for resume from a full stop, while holding a brake pedal, right?
Correct, it's for allowing to engage from full stop with brake pressed.
Would not engage prevent the PCM fault? If so, then maybe not allowing to engage would still be better than engaging and showing the fault.

vntarasov: With current code, openpilot would still engage, but then immediately disengage, right?
But I agree showing "faulted" does seem less user friendly.

rbiasini: currently, if you have the brake pressed, openpilot won't engage because a NO_ENTRY event is present. It would engage if you press the brake, release it, then press set/resume. In this case, would a fault be present until speed reaches 18mph?

vntarasov: @rbiasini  I just realized the whole thing is only a concern for Acadia and CT6, since all other GM cars have minimum speed to engage set to 18mph (`minEnableSpeed`).

PCM sends "faulted" status as long as openpilot sends ACC commands in an "improper" state.
- With PR 466, openpilot stops ACC commands on "faulted", and PCM stops sending "faulted" status back, immediately.
- This "faulted" status doesn't persist in any other way. That is, users can engage openpilot as soon as they go above 18mph regardless if they tried to engage below 18mph or not (even if openpilot didn't prevent this by `minEnableSpeed`).

rbiasini: ah, good catch! Ok, then the proposed change seems a no brainer. And the Acadia missing minEnableSpeed is probably a bug. Thanks for the explanation!

PR #1224, opened 2020-03-09
revise wifi signalstrength dumpsys query

valish:


common/android.py
```diff
@@ -235,9 +235,12 @@ def get_cdma_level(cdmadbm, cdmaecio):
...
     network_strength = NetworkStrength.unknown
     for line in out.split('\n'):
-      if "SignalStrength" in line:
-        arr = line.split(' ')
-        lvl = int(arr[53][:-2])
+      signal_str = "SignalStrength: "
+      if signal_str in line:
+        arr = list(line)
+        lvl_idx_start = int(line.find(signal_str) + len(signal_str))
+        lvl_idx_end = int(line.rfind(']}'))
```
pd0wm: Can you assume the order of the list you're parsing is always the same? I would look for the first `]` after the Signalstrength. Something like

```python
lvl_idx_end = line.find(']', lvl_idx_start)
```

Also, why the casts to int?

PR #1226, opened 2020-03-10
Skoda Superb B8 fingerprint

netadvanced:
Here's the latest Skoda Superb B8 fingerprint, as part of the Volkswagen MQB platform.
I took the liberty to slightly update the code to fit the current fingerprinting method and add custom adaptations for the Superb.
Jyoung suggested to use the new car interface parameters as defined in vw-openpilot-dev.


rbiasini:
I would not mix car support and new tuning in the same PR. 
Can we just limit this PR to new car support? Also, can you please add a SKODA drive to `test_car_models.py`? It'll be made public for regression tests.

netadvanced:
Thanks, will work on reverting changes to fingerprint only and add a test drive.
Can you please point me in the right direction concerning a fingerprint validation error failing the unit test, I seem to have to deal with 840:8 in the fingerprint, not sure where to start looking.

jyoung8607:
I know this PR is closed pending resubmission, but I have some commentary in the meantime:

> I would not mix car support and new tuning in the same PR.

Agreed. That said, @rbiasini I'd appreciate your feedback on the tune if you have a chance to run it on your GTI. It's been a fight to find something that doesn't oscillate at low speeds (hence all the crazy breakpoints in the old code) and this is working for most people now.

> Can we just limit this PR to new car support?

@netadvanced submitted this with my blessing (not that he needs it) but be advised Comma should not take very many traditional fingerprints for VW MQB, as they have extremely high overlap with each other, they essentially won't work to disambiguate between MQB vehicles.

The core code itself supports any MQB, and there are a LOT of those, but proper identification of other vehicles require some other mechanism like the UDS v2 fingerprinting. Unfortunately I've tried it and it's looking like it may take some effort because I think we may have to implement Bosch CAN TP 2.0 as a substitute transport layer for ISO-TP. I have at least two other alternative approaches but I think Comma would probably be less interested in upstreaming them.

> Also, can you please add a SKODA drive to `test_car_models.py`? It'll be made public for regression tests.

Another question for @rbiasini - what's your guidance for when you'd like to see test drive data? For the vast majority of MQB vehicles, we're exercising no conditionals in the code, just setting a wheelbase and mass and PIF value set and they Just Work. Exceptions are manual trans vehicles and the all-electric e-Golf, both of which just take some special cases for current gear detection, and there will be a tiny tweak for vehicles with no factory LKAS.

Happy to submit test drives for any and all supported vehicles as they get sent upstream, for the sake of completeness and standardization if that's what you'd like, just be aware that most of them will behave totally identically.

netadvanced:
@jyoung8607  @rbiasini 
I can't run the tests properly due to an error with fingerprint: 840 :8, and I still don't know where to exclude this value in the tests as @jyoung8607 suggested.  Can you please help me with that ?
I will push the fingerprint independently from my code changes once I get this test resolved.
Also, what kind of test drive do you need, is highway only enough ?

PR #1227, opened 2020-03-11
stricter calibration requirements

haraschax:
This rejects about 20% of the worst measurements for calibration once the device has initial calibration. Also changes the calibration max window of measurements used from 100s to 250s.


geohot:
I think we should start predicting the "calibration offset" directly in the network and using that.

Software 2.0 is eating the world!

PR #1243, opened 2020-03-16
Update to README.md

BRose98:
The footnotes 5 and 6 are now moved up to be below the supported car table, where they are actually mentioned, instead of below the community maintained cars and features table.


geohot:
Those are "community maintained" features, hence why they are there.

PR #1246, opened 2020-03-16
fade ll out and in

haraschax:


selfdrive/controls/lib/pathplanner.py
```diff
@@ -110,19 +115,31 @@ def update(self, sm, pm, CP, VM):
...
           self.lane_change_state = LaneChangeState.off
         elif torque_applied:
+          self.lane_change_state = LaneChangeState.laneChangeStarting
+
+      # starting
+      elif self.lane_change_state == LaneChangeState.laneChangeStarting:
+        # fade out lanelines over 1s
+        self.lane_change_ll_prob = max(self.lane_change_ll_prob - DT_MDL, 0.0)
+        # 99% certainty
+        if lane_change_prob < 0.01:
```
pd0wm: How do you prevent it from not immediately going to the next state if lane_change_prob is still low because it hasn't started?

PR #1251, opened 2020-03-18
Spinner updates: show current status and update TOTAL_SCONS_NODES

sshane:
- With this PR I do not count non-python preimports as progress updates so the progress bar gets closer to the end than previously (we can also move `locationd` closer to the beginning, or right at the end)
- This code should allow for backwards compatibility, it falls back to normal behavior if `spinner_status` is not set in `spinner.update()` (both with text and integer strings)
- Update `TOTAL_SCONS_NODES` to correlate with what my EON reports as finished
- Made the scons build finished progress 70% so it feels like it's working quicker (and this process takes up the bulk of the start up process if not compiled)

What it changes: gives the user more information about what their EON/C2 is doing when booting up and not compiled

What it looks like: https://streamable.com/j443p

I can change any of the text (changing to percentage for ex.) or add/remove messages, just let me know if that's something that's desired!


ShaneSmiskol:
Working on adding an ETA below the status so the user can expect when the compilation will complete

ShaneSmiskol:
Having trouble getting an accurate estimated time remaining, anyone have any ideas? Currently trying to use the nodes per second multiplied by the current node count/progress but since the nodes don't exactly correlate with the number of compiled files the nodes per second varies widely causing very inaccurate times. Also tried to smooth the output estimated time remaining value but still not ideal

pd0wm:
Thanks for working on this.

I like the change of 70% of spinner real estate for the compilation. However, I'm not a fan of all the text. I don't think it provides any useful information, even to a developer.

Something that would be greatly appreciated though, is something like the spinner that shows the error when it fails to compile or start manager. There would be two cases here, a scons compilation error, or something else in manager (e.g. registration) fails.

ShaneSmiskol:
Interesting, that would be useful for errors for end users who can't easily ssh in. Would the only situation for printing an error be an scons compilation error (because of git corruption?) or module import error (because of git corruption or the user has changed a file with incorrect syntax?)?

Instead of printing the current status, would you be interested at all in an accurate ETA printing instead? I'm having trouble getting that working so if that's not desired I can quit working on it and try to add your error printing with a button to reset to origin, or git stash like you suggested

pd0wm:
Think the ETA is not super useful. release2 is precompiled, and devel users will only see it the first time since after that it's cached.

ShaneSmiskol:
Okay, so far it only shows C compilation errors, single line (the first erroring line/message that returns) but is this closer to what you want? https://streamable.com/n59ya

I've never been too great with C or designing UI but if we want to implement a button to reset/stash I'm thinking we would need to spawn a thread in manager once an error occurs to listen to messages from the spinner process if the user touches a 'reboot' or 'reset' button. Once spinner receives this error string like in the video, it would remain in a loop waiting for a button touch (maybe timing out at some set duration). If that sounds okay I can add that.

geohot:
I think the new video looks much better. Text for errors, no text for no errors. I'd be fine with merging a small PR that does this.

ShaneSmiskol:
Sounds good then. How about the normal clean when scons fails and a `git reset --hard` to origin when Python files fail to pre-import? If the latter fails more me it's usually because of a shutdown from low battery causing some Python files to get corrupted with null bytes, resetting with git usually fixes those errors for me

ShaneSmiskol:
Okay, here's the new behavior if the scons build fails: https://streamable.com/wdwwg

And here's what it does if something in the preparation loop fails calling `prepare_mangaged_process()`: https://streamable.com/e1j9j

If something fails to import or is otherwise missing (like locationd for example), the second loop will fetch and then hard reset to origin/`@{u}`. I deleted an entire selfdrive folder (locationd), then started up openpilot. It failed to find the module so then it fetched and hard reset. On rebooting it started up successfully. Will change my PR here to reflect all these changes soon.

I also made some changes in spinner to differentiate between extra messages and extra error messages (to use smaller font sizes, which is also picked based on the length of the message).

ShaneSmiskol:
The only I'm having is if there's a big chunk of text with no spaces, it will go off the screen. Is there an easy way to get the number of rows nanovg is going to split the text into?

selfdrive/common/spinner.c
```diff
@@ -166,11 +181,29 @@ int spin(int argc, char** argv) {
...
       nvgTextAlign(vg, NVG_ALIGN_CENTER | NVG_ALIGN_TOP);
       nvgFontSize(vg, 96.0f);
       nvgText(vg, fb_w/2, (fb_h*2/3)+24, spintext, NULL);
+    } else if (has_extra) {
+      nvgTextAlign(vg, NVG_ALIGN_CENTER | NVG_ALIGN_TOP);
+      if (err_msg) {
+        int break_row_width = 1300;
+        int y_offset = strlen(spinerr) > 160 ? 76 : 96;
+        // need smaller font for longer error msg
+        int fontsize = strlen(spinerr) > 50 ? 68.0f : 72.0f;
```
ShaneSmiskol: Could need some tuning. If there's an error with a large single chunk (no spaces) it will go off screen, I currently have no way to know how many rows nanovg is splitting text into

selfdrive/manager.py
```diff
@@ -97,20 +105,30 @@ def unblock_stdout():
...
           if spinner is not None:
-            spinner.update("%d" % (50.0 * (i / TOTAL_SCONS_NODES)))
-        elif len(line):
-          print(line.decode('utf8'))
+            spinner.update("%d" % progress)
+
+        if len(line) and not line.startswith(prefix):
+          line = line.decode('utf8')
+          print(line)
+          if any([err in line for err in ['error: ', 'not found, needed by target']]):
```
ShaneSmiskol: Any other specific types of messages to add as an error to display?

selfdrive/manager.py
```diff
@@ -294,6 +315,7 @@ def prepare_managed_process(p):
       cloudlog.warning("building %s failed, make clean" % (proc, ))
```
ShaneSmiskol: Might want to add another try except here and return the exception. I don't ever run into these types of errors however

selfdrive/manager.py
```diff
@@ -478,12 +500,40 @@ def manager_prepare(spinner=None):
...
+      e = prepare_managed_process(p)
+      progress = (100.0 - total) + total * (i + 1) / len(managed_processes)
+      if spinner is not None:
+        if e is None:
+          spinner.update("%d" % progress)
+        else:
+          prep_failed = True
+          for _ in range(10):
+            spinner.update("%d" % progress, format_spinner_error(str(e)))
+            time.sleep(1)
```
ShaneSmiskol: I chose to write it this way because of the way spinner receives messages. If this is ran quickly after a normal update, it completely skips showing the error message

selfdrive/manager.py
```diff
@@ -478,12 +500,40 @@ def manager_prepare(spinner=None):
...
+      if retry:
+        print("preparation failed, hard resetting in")
+        for i in range(5):
+          print(5 - i)
+          spinner.update("%d" % progress, "preparation failed, hard resetting in {}...".format(5 - i))
+          time.sleep(1)
+
+        subprocess.check_output(["git", "fetch"], cwd="/data/openpilot", stderr=subprocess.STDOUT, encoding='utf8')
+        r = subprocess.check_output(["git", "reset", "--hard", "@{u}"], cwd="/data/openpilot", stderr=subprocess.STDOUT,
+                                    encoding='utf8')
```
ShaneSmiskol: Should more be here to let the user know they should connect to the internet before it fetches, and wait for that? Or should the reset be good enough? `updated` isn't running here so if a fork's owner writes bad code it would fail to start and pull the fix by itself.

selfdrive/common/spinner.c
pd0wm: You can split the lines in python. If you put in `\n` nanovg will correctly render them as linebreaks.

selfdrive/manager.py
```diff
@@ -61,6 +61,12 @@ def unblock_stdout():
     exit_status = os.wait()[1] >> 8
     os._exit(exit_status)
 
+def format_spinner_error(err):
+  if len(err) > 192 - 8:
+    err = err[:184].strip() + '...'
+  print(err)
```
pd0wm: Now the error is printed twice

selfdrive/manager.py
pd0wm: Let's remove the git reset for now. Not sure yet how I feel about doing that automatically.

selfdrive/manager.py
```diff
@@ -97,20 +105,30 @@ def unblock_stdout():
...
-        elif len(line):
-          print(line.decode('utf8'))
+            spinner.update("%d" % progress)
+
+        if len(line) and not line.startswith(prefix):
+          line = line.decode('utf8')
+          print(line)
+          if any([err in line for err in ['error: ', 'not found, needed by target']]):
+            build_error = True
+            for _ in range(10):
```
pd0wm: Why not a time.sleep(10)?

selfdrive/manager.py
ShaneSmiskol: So just show the error for 10 seconds and raise exception for now?

pd0wm: yes

selfdrive/manager.py
ShaneSmiskol: Since I experienced the bug where it wouldn't show the error if the last spinner update is sent almost immediately for the prepare loop I wanted to ensure that never happened for the scons error. Though I never experienced it for this loop so it was just me adding it just in case it happens to happen. I can change to a time.sleep(10) if you think that it will be good

selfdrive/manager.py
ShaneSmiskol: Whoops, that was left in from debugging

selfdrive/common/spinner.c
ShaneSmiskol: I tried that but I think in spinner.c, this line breaks it: `spintext[strcspn(spintext, "\n")] = 0;` It renders `\n` weirdly, flashing the two lines when a new update is received. Is that `\n` only meant for the last one?

pd0wm: That is to make sure the last `\n` is cut off.

ShaneSmiskol: I wrote some code to cut off the last char (` = '0';`) if it's a `\n` in spinner.c however it still does not play nicely with newlines added in from Python, flashes on the screen. With `nvgTextBox` at least. I'll try `nvgText` and manually create new lines

ShaneSmiskol: Got it working by replacing `\n` with the ASCII unit separator and then converting it back inside spinner to newlines! Now getting the line splitting logic ready. I also noticed I forgot to add `line = 'ERR,' + line` inside the formatter function so I'll add that when I get this done

ShaneSmiskol:
How long error messages were handled before:



How they're handled now:



Thanks for the suggestion to move the logic into Python, much better

selfdrive/manager.py
```diff
@@ -61,6 +61,23 @@ def unblock_stdout():
     exit_status = os.wait()[1] >> 8
     os._exit(exit_status)
 
+def format_spinner_error(err):
+  err_list = [[]]
```
pd0wm: This should only be a few lines. Use the built in textwrap module, which will deal with all the edge cases (such as no spaces to break on).

selfdrive/manager.py
pd0wm: Can you figure out why this happens? This should now be solved with a hacky sleep.

ShaneSmiskol: Sure, I'll check it out tomorrow

selfdrive/manager.py
ShaneSmiskol: Got it, moved it to textwrap. After upping the max length from 50 to 65 the output from textwrap is looking a lot more similar to the function I had. At 50 it was very close to nanovg so I chose to write my own with a little more leniency. But it looks good now.

selfdrive/manager.py
```diff
@@ -97,20 +109,29 @@ def unblock_stdout():
...
+            spinner.update("%d" % progress)
+
+        if len(line) and not line.startswith(prefix):
+          line = line.decode('utf8')
+          print(line)
+          if any([err in line for err in ['error: ', 'not found, needed by target']]):
+            build_error = True
+            spinner.update("%d" % progress, format_spinner_error(line))
+            time.sleep(10)
+            break
```
pd0wm: You should also wait for the scons command to finish. Otherwise you might start the clean command while it's still building. Because it's building in multiple threads an error can show up, and it could still be finishing things in another thread.

selfdrive/manager.py
```diff
@@ -478,12 +503,25 @@ def manager_prepare(spinner=None):
...
+      if e is None:
+        spinner.update("%d" % progress)
+      else:
+        prep_failed = True
+        for _ in range(10):
+          spinner.update("%d" % progress, format_spinner_error(str(e)))
+          time.sleep(1)
+        break
+
+  if prep_failed:
```
pd0wm: Can you reraise the original error when it fails?

ShaneSmiskol: Done!

ShaneSmiskol:
Closing PR in favor of using Willem's new text binary for displaying errors

ShaneSmiskol:
Closing PR in favor of using Willem's new text binary for displaying errors

PR #1255, opened 2020-03-20
Preparing for the efficientnet

geohot:
Using SNPE 1.35.0.698, sadly currently taking 65% of a core and over 50ms to run!


geohot:
@HaraldSchafer what's the ETA on a B2? Would be nice to merge this for tici work.

HaraldSchafer:
A shippable B2, would take 5 more days I think :/, though I can merge this for current model?

geohot:
That's a good idea, I put the old model back, let's see if tests pass

PR #1256, opened 2020-03-20
Fix for steer unavailable

arne182:
https://github.com/commaai/openpilot/pull/972
Redone

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -82,20 +82,28 @@ class CarController():
 
     # steer torque
     new_steer = int(round(actuators.steer * SteerLimitParams.STEER_MAX))
-    apply_steer = apply_toyota_steer_torque_limits(new_steer, self.last_steer, CS.out.steeringTorqueEps, SteerLimitParams)
-    self.steer_rate_limited = new_steer != apply_steer
+
 
     # only cut torque when steer state is a known fault
     if CS.steer_state in [9, 25]:
       self.last_fault_frame = frame
 
     # Cut steering for 2s after fault
-    if not enabled or (frame - self.last_fault_frame < 200):
+    if (frame - self.last_fault_frame < 200) or \
+       (abs(CS.out.angle_steers_rate) > 100 or abs(CS.out.angle_steers) > 100):
+      new_steer = 0
       apply_steer = 0
       apply_steer_req = 0
     else:
       apply_steer_req = 1
-
+      
+    apply_steer = apply_toyota_steer_torque_limits(new_steer, self.last_steer, CS.out.steeringTorqueEps, SteerLimitParams)
+    self.steer_rate_limited = new_steer != apply_steer
+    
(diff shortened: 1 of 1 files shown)


ShaneSmiskol:
I think this is a very important PR for us with Toyotas, any sudden movement in the wheel (which is common at low speeds with tight curves) triggers steering faults and the car would drive off the road without taking over or this code. However I don't agree with limiting the angle to 100 degrees, I think the rate limit by itself is adequate.

Of course this is just a fix for side effect of bad steering tuning, all of the situations that the faults occur is because openpilot jerks the wheel to try to follow the path because it was either understeering or oversteering in the first place. So either this PR or the upcoming improved tuning would solve this problem I think.

arne182:
I have seen this on the tssp rav4 as well in fast movements that is why I have taken the limitation out for only rav4 and corrolla

VirtuallyChris:
If it fails at low speeds when not touching the wheel I agree that's bad! This might be intentional behavior though when you are manually taking over besides minor corrections. On all Honda's (including my Civic) they will throw a steer unavailable alert when I jerk the wheel to take a large correction and I just assumed this was intentional to not fight the driver. Can anyone chime in on that?

pd0wm:
I don't like the cutout after a 100 degrees steering angle. Can you remove that?

So the changes are:
1) Cutout steering when steering rate is over 100 deg/s
2) Move rate limiter after cutout, so now it ramps down slowly.

By doing 2 we change existing behavior when a fault occurs. Instead of going to 0 immediately we now ramp down slowly.

If that works I'd like to propose the following:
1) In case of a hard fault, drop torque to 0 immediately like we used to do
2) In case of a steering rate over 100 deg/s go to 0 through the rate limiter.

pd0wm:
@arne182 Are you still working on this?

ShaneSmiskol:
Untested, but how does this look? https://github.com/ShaneSmiskol/openpilot/commit/2dd28419913954346d271e62aa4dad41d126a087

Going out to test now

arne182:
I know that it faults if it is not 0. Even 10 is not small enough. If steering request and steer apply is not 0 as quickly as possible it will fault.

pd0wm:
I don't understand. In this PR you make the new_steer steer go through the rate limiter, you're just settings `apply_steer_req` to 0 with a non zero torque now. Why did you move the rate limiter to below the fault check?

ShaneSmiskol:
@pd0wm Is there any version of this PR you would think about merging? I would really appreciate this upstreamed

ShaneSmiskol:
This is the minimum required, I think: https://github.com/ShaneSmiskol/openpilot/commit/8b3fcab28541ae0760371e9e2b152381e8b26019

ShaneSmiskol:
@pd0wm Here is an example of what it's like for us on stock openpilot on about 40% of sharp curves: https://www.youtube.com/watch?v=EwBIkGsywbc Would appreciate your input, if this PR can't be merged, then are there any other solutions?

PR #1261, opened 2020-03-22
paint.cc uses color defined in ui.hpp

eFiniLan:


selfdrive/ui/ui.hpp
```diff
@@ -40,9 +40,13 @@
 
 #define COLOR_BLACK_ALPHA nvgRGBA(0, 0, 0, 85)
 #define COLOR_WHITE nvgRGBA(255, 255, 255, 255)
-#define COLOR_WHITE_ALPHA nvgRGBA(255, 255, 255, 85)
+#define COLOR_WHITE_ALPHA_20 nvgRGBA(255, 255, 255, 20)
```
geohot: Can you use a macro for this?

aka #define COLOR_WHITE_ALPHA(x) ...

eFiniLan: done!

PR #1264, opened 2020-03-22
@Wei Prius FP

arne182:
dongle_id = 68f0cfd614f0ddf7


richardtaur:
I have a 2017 Toyota Prius Two

PR #1269, opened 2020-03-23
Power monitoring fix

robbederks:


pd0wm (approved):

pd0wm:
Looks good now. Can you do some final testing on all panda combinations? Then I'll ship to devel.

PR #1272, opened 2020-03-24
Adds 2020 Ridgeline + pedal support for 2019+2020

Davidss2:
This change adds the 2020 Ridgeline and pedal support for both the 2019 and 2020. I have tested this on my own 2020 Ridgeline.

I asked for help a few times on discord but no one was willing so I studied up and did this myself. Apologies if I screwed something up.


vanillagorillaa:
So this is really two pull requests. 

We will merge the additional pedal support for the 2019 Ridgeline. 

For the 2020 Ridgeline we will only merge V2 fingerprints (fingerprints based off of the ECU FW Versions). See this script, https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py, for getting ECU FW versions. This will require V2 hardware (Black Panda or a comma two). 

If you have the V2 hardware you can also post the dongleID (settings -> Device) for your device and we can pull the FW versions that way.

Davidss2:
Here is the dongle ID 5a14cea88398e3ab.

pd0wm:
FW versions and pedal messages are added on master.

PR #1280, opened 2020-03-26
Move all non Honda+Toyota to community supported

geohot:
Since comma.ai doesn't have these cars, they are maintained and supported by the community.


geohot:
I think this is good to merge if we update the process replay. Website is already updated.

pd0wm (approved):

PR #1282, opened 2020-03-27
not run these processes with webcam

ZwX1616:
- do we need uploader to run in webcam mode?
- don't need ubloxd since no gps

### DIFF
--- a/selfdrive/manager.py
+++ b/selfdrive/manager.py
@@ -189,8 +189,12 @@ persistent_processes = [
   'thermald',
   'logmessaged',
   'ui',
-  'uploader',
 ]
+
+if not WEBCAM:
+  persistent_processes += [
+    'uploader',
+  ]
 
 if ANDROID:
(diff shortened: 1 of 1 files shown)


geohot:
Why no GPS, should totally work with black panda?

And sure, why not re: uploader and logger. I don't think the camera logging works on PC, but the qlog/rlog should, and there's no reason to not get them.

ZwX1616:
> Why no GPS, should totally work with black panda?
> 
> And sure, why not re: uploader and logger. I don't think the camera logging works on PC, but the qlog/rlog should, and there's no reason to not get them.
> 
> Also, the less we diverge from OP on comma two, the better.

Oh yeah, I was thinking webcam even includes working with simulation; but since we run monitoring too with webcam so let's assume it is in a real car. Sorry for the confusion

Let's also keep the logs then; they will be interesting for comparing performance!

PR #1283, opened 2020-03-27
Removed self-assignments

BRose98:
Removed self-assignments from manager.py and test_lateral_mpc.py


pd0wm:
Can you make sure the submodules are not changed?

geohot:
Submodules are still changed.

PR #1284, opened 2020-03-27
Added constant to prevent literal duplication

bartziengs:


geohot:
Doesn't seem more readable to me.

PR #1287, opened 2020-03-28
Split nested conditional into multiple lines

pevdh:


pevdh:
Oops, let me fix these errors.

geohot:
Why change this? The nested one is readable I think, and less likely to be wrong.

PR #1288, opened 2020-03-28
use put_nonblocking when possible

eFiniLan:
use put_nonblocking instead of params.put when possible.

### DIFF
--- a/selfdrive/camerad/snapshot/snapshot.py
+++ b/selfdrive/camerad/snapshot/snapshot.py
@@ -6,7 +6,7 @@ import subprocess
 import time
 from PIL import Image
 from common.basedir import BASEDIR
-from common.params import Params, put_nonblocking
+from common.params import Params
 from selfdrive.camerad.snapshot.visionipc import VisionIPC
 
 with open(BASEDIR + "/selfdrive/controls/lib/alerts_offroad.json") as json_file:
@@ -33,7 +33,7 @@ def snapshot():
   ps = subprocess.Popen("ps | grep camerad", shell=True, stdout=subprocess.PIPE)
   ret = list(filter(lambda x: 'grep ' not in x, ps.communicate()[0].decode('utf-8').strip().split("\n")))
   if len(ret) > 0:
-    put_nonblocking("IsTakingSnapshot", "0")
+    params.put("IsTakingSnapshot", "0")
     params.delete("Offroad_IsTakingSnapshot")
     return None
 
@@ -63,7 +63,7 @@ def snapshot():
   proc.send_signal(signal.SIGINT)
   proc.communicate()
 
-  put_nonblocking("IsTakingSnapshot", "0")
+  params.put("IsTakingSnapshot", "0")
   params.delete("Offroad_IsTakingSnapshot")
   return ret
 
--- a/selfdrive/camerad/snapshot/snapshot.py
+++ b/selfdrive/camerad/snapshot/snapshot.py
@@ -6,7 +6,7 @@ import subprocess
 import time
 from PIL import Image
 from common.basedir import BASEDIR
-from common.params import Params
+from common.params import Params, put_nonblocking
 from selfdrive.camerad.snapshot.visionipc import VisionIPC
 
 with open(BASEDIR + "/selfdrive/controls/lib/alerts_offroad.json") as json_file:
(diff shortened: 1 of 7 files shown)


selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -33,7 +33,7 @@ def snapshot():
   ps = subprocess.Popen("ps | grep camerad", shell=True, stdout=subprocess.PIPE)
   ret = list(filter(lambda x: 'grep ' not in x, ps.communicate()[0].decode('utf-8').strip().split("\n")))
   if len(ret) > 0:
-    params.put("IsTakingSnapshot", "0")
+    put_nonblocking("IsTakingSnapshot", "0")
```
pd0wm: You don't want the asynchronous version in this file since that could cause race conditions.

selfdrive/car/car_helpers.py
```diff
@@ -91,7 +91,7 @@ def fingerprint(logcan, sendcan, has_relay):
     fw_candidates, car_fw = set(), []
 
   cloudlog.warning("VIN %s", vin)
-  Params().put("CarVin", vin)
+  put_nonblocking("CarVin", vin)
```
pd0wm: We can't use the asynchronous version here, since boardd is waiting for the vin to be set and change the safety mode.

selfdrive/controls/controlsd.py
```diff
@@ -486,9 +486,9 @@ def controlsd_thread(sm=None, pm=None, can_sock=None):
 
   # Write CarParams for radard and boardd safety mode
   cp_bytes = CP.to_bytes()
-  params.put("CarParams", cp_bytes)
-  params.put("CarParamsCache", cp_bytes)
-  params.put("LongitudinalControl", "1" if CP.openpilotLongitudinalControl else "0")
+  put_nonblocking("CarParams", cp_bytes)
```
pd0wm: Other services are waiting for carParams to start. Use synchronous version here.

selfdrive/registration.py
```diff
@@ -6,20 +6,19 @@
...
-  params.put("TrainingVersion", training_version)
+  put_nonblocking("Version", version)
+  put_nonblocking("TermsVersion", terms_version)
+  put_nonblocking("TrainingVersion", training_version)
 
-  params.put("GitCommit", get_git_commit(default=""))
-  params.put("GitBranch", get_git_branch(default=""))
-  params.put("GitRemote", get_git_remote(default=""))
-  params.put("SubscriberInfo", get_subscriber_info())
+  put_nonblocking("GitCommit", get_git_commit(default=""))
```
pd0wm: Don't use asynchronous version in this file. It could cause a race condition where loggerd starts without the params set

selfdrive/updated.py
```diff
@@ -110,20 +110,18 @@ def set_consistent_flag():
 
 
 def set_update_available_params(new_version=False):
-  params = Params()
-
   t = datetime.datetime.utcnow().isoformat()
-  params.put("LastUpdateTime", t.encode('utf8'))
+  put_nonblocking("LastUpdateTime", t.encode('utf8'))
```
pd0wm: Don't use asynchronous version here. There could be a race condition where the UpdateAvailableParam is set, but the release notes aren't there yet.

Updated is also no in a critical loop, so timeouts don't matter.

pd0wm (changes requested):

selfdrive/camerad/snapshot/snapshot.py
pd0wm: I don't think we need the asynchronous version here since it's not in a timing critical loop.

eFiniLan:
@pd0wm thanks for the feedback :)
regarding to registration process, I have seen a few cases where their devices failed to authenticate due to modify persist partition (e.g. key deleted and regenerated).

The current solution is either send a request to unregister or modify codes to "skip" the process, would it be possible not to ```raise Exception("server registration failed")```? or have a way for users to re-register their devices?

cheers!

pd0wm:
If the private key is corrupt they need to contact comma. We need to manually remove the public from our database, otherwise you can't register the device with a different key. Otherwise you could just reregister with a new private key and take over someones account.

PR #1289, opened 2020-03-28
Nissan set speed

avolmensky:
Added cruiseState.speed to Nissan X-Trail.


selfdrive/car/nissan/carstate.py
```diff
@@ -48,12 +48,16 @@ def update(self, cp, cp_adas, cp_cam):
     elif self.CP.carFingerprint == CAR.LEAF:
       ret.cruiseState.available = bool(cp.vl["CRUISE_THROTTLE"]["CRUISE_AVAILABLE"])
 
-    # TODO: Find mph/kph bit on XTRAIL
-    if self.CP.carFingerprint == CAR.LEAF:
+    # TODO: Find mph/kph bit on XTRAIL until then, assume xtrail is kph.
+    # Unable to change kph to mph on the xtrail, need a rogue to test it on
+    if self.CP.carFingerprint == CAR.LEAF or self.CP.carFingerprint == CAR.XTRAIL:
```
pd0wm: You can remove this check.

pd0wm:
If you replace the contents of `openpilot/selfdrive/test/process_replay/ref_commit` by `64da2a519cfff9f74bd07190dd96425224a8ad1f` the tests should pass.

selfdrive/car/nissan/carcontroller.py
```diff
@@ -76,7 +76,7 @@ def update(self, enabled, CS, frame, actuators, cruise_cancel, hud_alert,
         can_sends.append(nissancan.create_cancel_msg(self.packer, CS.cancel_msg, cruise_cancel))
 
     can_sends.append(nissancan.create_steering_control(
-        self.packer, self.car_fingerprint, apply_angle, frame, acc_active, self.lkas_max_torque))
+        self.packer, self.car_fingerprint, apply_angle, frame, enabled, self.lkas_max_torque))
```
sshane, line 79: @avolmensky do you know any reason why we set the apply bit with the cruise state? Can we just set it with the steering state (`CC.latActive`)? The only difference would be at a standstill, the bit would now fall, and for any steering faults (which we do not log right now, so no difference there).

Issue #1270, opened 2020-03-24
[TSS2 Rav4] OpenPilot (engaged) sporadically creeps forward slowly at stopped traffic while at a red light 


dpanych:
**Describe the bug**
While OP is engaged, when coming to a complete stop (like a red light), OP will first stop at a normal distance, but then it will continue to creep toward the stopped car (very slowly). 

**How to reproduce or log data**
I am unable to create this as it seems to be random. For futures sake, I will save cabana links when it occurs again.
https://my.comma.ai/cabana/?route=7e34a988419b5307%7C2020-03-19--19-19-14&exp=1616186697&sig=GOzZLgLrNbOUJ9FTgnAQF0HTbhe34bN28UfWwi6UJo4%3D&max=16&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2F7e34a988419b5307%2F1dac8843e1a4c769857cd329d8baad59_2020-03-19--19-19-14 (Look at 466.405)

**Expected behavior**
The car should stop and stay stopped until traffic starts moving again.

**Device/Version information (please complete the following information):**
 - Device: EON GOLD
 - Version: 0.7.4 staging-devel
 - Car make/model: 2020 RAV4 Hybrid XLE

arne182:
Have seen this on my fork users. 
I have increased this line from longcontrol form 0.5 to
BRAKE_STOPPING_TARGET = 0.75  # apply at least this amount of brake to maintain the vehicle stationary
Giving 25% more braking at standstill.

arne182:
I still think we should include lead car distance into longcontrol to avoid this having both the rate and the amount dynamically change based on lead car behaviour.

pd0wm:
How much braking does the stock system send at standstill?

We send an acceleration to the car, 0.5 braking maps to -1.5 m/s^2. Theoretically everything smaller than zero should keep the car stopped, since they have a lower level controller that should keep the brakes applied.

arne182:
Yes 0.5 -1.5m/s^2 should keep the car at a standstill but it does not for the TSS2 RAV4 and hybrid. Having -2.0 to -2.5m/s^2, does keep the car from creeping. The funny part is that vEgo is 0 but lead distance is decreasing.

arne182:
![image](https://user-images.githubusercontent.com/844102/77742902-84a22900-7017-11ea-86f8-fa4e8e284dda.png)

dpanych:
[...]

PR #1291, opened 2020-03-30
Stop tss2_cars from creeping

arne182:
This needs to be tested but it looks like it could fix.
https://github.com/commaai/openpilot/issues/1270


pd0wm:
Does this require a button press to resume?

arne182:
Yes. But we could always just fix it with increasing the brake power to 75%

pd0wm:
I think the brake power to 75% might be a better solution then. We would need to parametrize the stopping brake, rate up and rate down in the carParams so we can keep the behavior on other cars the same.

arne182:
Agreed. I at least know that this works. I wonder if this is why my car is going into waiting mode.

PR #1296, opened 2020-03-31
Static 0xe5 on Bosch

csouers:
Block 0xe5 (Bosch only) and send only static values. Wheel wobble functionality is disabled on vehicles that support it. This important safety change is crucial for Bosch cars because the wobble overrides any torque command on 0xe4 sent by OP (or stock LKAS).

Panda PR for safety. 0xe5 controls wheel wobble on Bosch and also has a bit for braking request during RDM. https://github.com/commaai/panda/pull/515

Opendbc changes are a requirement too https://github.com/commaai/opendbc/pull/250

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,11 @@
 [submodule "panda"]
 	path = panda
-	url = ../../commaai/panda.git
+	url = ../../csouers/panda.git
+	branch = no_e5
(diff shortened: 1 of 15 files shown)


csouers:
Splitting e5 and RDM stuff. Closing this for now.

PR #1299, opened 2020-03-31
RHD support for driver monitoring

ZwX1616:
TODO:
- <s>(push comma two model)</s>
- <s>better geo parse</s>
- <s>front view in UI</s>


selfdrive/manager.py
```diff
@@ -481,6 +482,11 @@ def manager_thread():
       logger_dead = False
       for p in reversed(car_started_processes):
         kill_managed_process(p)
+      # this is ugly
```
ZwX1616, line 485: is there a better way to manage this?

selfdrive/thermald/thermald.py
```diff
@@ -329,6 +331,9 @@ def thermald_thread():
     # check if system time is valid
     should_start = should_start and time_valid
 
+    # check if is in preview dcam mode
+    should_start = should_start and (not is_viewing_driver)
```
pd0wm, line 335: move into if statement below

Issue #1130, opened 2020-02-19
Sensors Init Failure -- $1000 Bounty


geohot:
After some reboots, the gyro does not initialize correctly, forcing the user to reboot to get the gyro back. Seems to affect the LeEco more than the OnePlus.

The alert this will fix is "No Data from Device Sensors" "Reboot your Device"

$1000 for a description of the root cause and merged pull request that fixes this issue.

danishtamboli123:
Hi, I think this has to do with a faulty Qualcomm firmware on the LeEco (Le Pro 3, the one in the EON), When the Gyro doesn't initialize are there instances where other sensors don't initialize aswell? (such as the accelerometer), This might just require you to flash a recent firmware via Qfil.
I could be completely wrong on this as I'm still a freshman at uni, but it could be worth a try. JUst an avid fan of what you guys are doing.

danishtamboli123:
Hi, I think this has to do with a faulty Qualcomm firmware on the LeEco (Le Pro 3, the one in the EON) (faulty firmware on the SD 821 to be specific), When the Gyro doesn't initialize are there instances where other sensors don't initialize aswell? (such as the accelerometer), This might just require you to flash a recent firmware via Qfil. I could be completely wrong on this as I'm still a freshman at uni, but it could be worth a try.  😅

geohot:
Do you have more information on this? Which firmware version is broken? Where is the new firmware?

danishtamboli123:
A lot have used the Method below to "Un-brick" their phones, But many forums also suggest that having done this has fixed their Sensor Issues as well.

Link to the QFIL image file and QFIL itself:
https://forum.xda-developers.com/le-pro3/how-to/unbrick-leeco-phones-leeco-variants-t3847220
https://forum.xda-developers.com/le-pro3/leeco-le-pro-3-elite-model-x722-guides-development--and-mods/qfil-x722-t3846553
https://forum.xda-developers.com/le-pro3/help/leeco-pro3-x720-sensors-problem-t3753165

Some have also suggested modifications in the Persists folders with regards to Sensors:
https://forum.xda-developers.com/oneplus-3t/how-to/fix-sensors-stopped-t3630854
https://forum.xda-developers.com/xiaomi-redmi-5a/how-to/guide-restoring-persist-partition-to-t3906424
https://drive.google.com/file/d/1ZgaWydW4c7Eo5vuRw_Vhc6EVmBFThNV2/view

[...]

PR #1303, opened 2020-04-01
Boot-loop testing script for EON/C2 device sensors

jyoung8607:
This PR adds a script to boot-loop an EON/C2, exercising the device sensors each time, pursuant to #1130. It uses @pd0wm's new text display library, fixing a minor issue along the way. It's not pretty but it's enough to repro the bug and accumulate statistics on how often it occurs.

This script could be trivially modified to halt for additional forensics when a failure is noticed, or to just invoke sensord repeatedly without the reboots to exercise another potential failure case that was recently mentioned, or test/burn-in other hardware as done in `leeco_selftest.sh`.

**Starting:** Running the script will enter a continuous test/display 10 seconds/reboot cycle until stopped.

**Stopping:** Touch the Exit button while results are being displayed and it'll reboot into normal OP.


pd0wm:
Can you make sure the panda submodule is not changed?

jyoung8607:
> Can you make sure the panda submodule is not changed?

Whoops, sorry about that, got some work-in-progress trees mixed up. I see you handled it already though, thanks!

PR #1306, opened 2020-04-02
Direct logentries logging from manager

pd0wm:


selfdrive/manager.py
```diff
@@ -88,25 +89,33 @@ def unblock_stdout():
...
         elif len(line):
-          print(line.decode('utf8'))
+          compile_output.append(line)
+          print(line.decode('utf8', 'replace'))
       except Exception:
         pass
 
     if scons.returncode != 0:
+      # Read remaining output
+      line = scons.stderr.readline()
```
pd0wm, line 115: Use while loop. There can be more than 1 line.

PR #1309, opened 2020-04-02
Toyota Tss2 Blindspot signals

arne182:
Adding these signals are easy, they are on the can bus.


geohot:
LGTM, @pd0wm can we write instructions for regenerating the process replay?

quillford:
* check out branch
* run `selfdrive/test/process_replay/update_refs.py`
* push updated ref_commit

I can add this to the process replay readme if you want

pd0wm:
Unfortunately it's a little annoying on github to checkout someones branch in such a way that you can push to it.

@arne182 Can you change `selfdrive/test/process_replay/ref_commit` to 
`63ee3eaab23b45821a3a04888ae7506ec27dedb3` then the tests should pass

arne182:
@quillford the problem then is to have an environment where it can run. I usually use termux on android to say update the generator of the opendbc things but for this command to work you need to have a whole bunch of things installed first. The last time I tried it on the eon it was also missing at least one dependancy.

PR #1311, opened 2020-04-03
Add angle based steer saturation timer

avolmensky:
Add a timer to the steer saturation warning. Was getting a lot of false positives when grabbing the wheel, or when line markings were unclear.

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -278,7 +278,7 @@ def state_control(frame, rcv_frame, plan, path_plan, CS, CP, state, events, v_cr
   saturated_count = saturated_count + 1 if angle_control_saturated and not CS.steeringPressed and active else 0
 
   # Send a "steering required alert" if saturation count has reached the limit
-  if (saturated_count > STEER_ANGLE_SATURATION_TIMEOUT) and (lac_log.saturated or angle_control_saturated) and not CS.steeringPressed:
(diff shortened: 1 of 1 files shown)


selfdrive/controls/controlsd.py
```diff
@@ -277,9 +275,10 @@ def state_control(frame, rcv_frame, plan, path_plan, CS, CP, state, events, v_cr
   angle_control_saturated = CP.steerControlType == car.CarParams.SteerControlType.angle and \
     abs(actuators.steerAngle - CS.steeringAngle) > STEER_ANGLE_SATURATION_THRESHOLD
 
+  saturated_count = saturated_count + 1 if angle_control_saturated and not CS.steeringPressed and active else 0
+
   # Send a "steering required alert" if saturation count has reached the limit
-  # TODO: add a minimum duration, now a warning will show every time the path changes briefly
-  if (active_count > SATURATION_TIMEOUT_ON_ENGAGE) and (lac_log.saturated or angle_control_saturated) and not CS.steeringPressed:
+  if (saturated_count > STEER_ANGLE_SATURATION_TIMEOUT) and (lac_log.saturated or angle_control_saturated) and not CS.steeringPressed:
```
pd0wm, line 281: This now needs to be an `or`, otherwise it will never trigger.

```suggestion
  if (lac_log.saturated or angle_control_saturated or (saturated_count > STEER_ANGLE_SATURATION_TIMEOUT)) and not CS.steeringPressed:
```

PR #1314, opened 2020-04-05
Fix unlogger read from files

SippieCup:
When reading from .hvec files downloaded from cabana unlogger adds the unknown -format option to ffprobe and then tries to convert an already correct format video file

### DIFF
--- a/tools/replay/unlogger.py
+++ b/tools/replay/unlogger.py
@@ -114,7 +114,6 @@ class UnloggerWorker(object):
           print("FRAME(%d) LAG -- %.2f ms" % (frame_id, fr_time*1000.0))
 
         if img is not None:
-          img = img[:, :, ::-1] # Convert RGB to BGR, which is what the camera outputs
           img = img.flatten()
           smsg.frame.image = img.tobytes()
 
--- a/tools/replay/unlogger.py
+++ b/tools/replay/unlogger.py
@@ -114,6 +114,7 @@ class UnloggerWorker(object):
           print("FRAME(%d) LAG -- %.2f ms" % (frame_id, fr_time*1000.0))
 
         if img is not None:
+          img = img[:, :, ::-1] # Convert RGB to BGR, which is what the camera outputs
           img = img.flatten()
           smsg.frame.image = img.tobytes()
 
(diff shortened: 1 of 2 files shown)


tools/replay/unlogger.py
```diff
@@ -114,7 +114,6 @@ def _send_logs(self, data_socket):
           print("FRAME(%d) LAG -- %.2f ms" % (frame_id, fr_time*1000.0))
 
         if img is not None:
-          img = img[:, :, ::-1] # Convert RGB to BGR, which is what the camera outputs
```
pd0wm, line 117: Why change this?

If I unlog a route using ./unlogger "<route name", straight from the API the colors are messed up.

pd0wm (changes requested):

tools/replay/unlogger.py
SippieCup: Ah, unfortunately my Panda dongle is "banned" from my.comma.ai for broken data uploads and i never got around to unbanning it.

As for why I removed that code, downloaded files are already the correct color palette. I'll make a check to see if its coming from my.comma.ai directly, and do the conversion rather than outright removal.

Who should I ping for removal of my dongle so I can test it via my.comma.ai before pushing the commit?

SippieCup: Reverted that change, Was really more due to the new pipeline we are using with webcams and me testing it on that branch. sorry about that!

pd0wm:
If we don't do anything with the argument, can we also remove it from the function arguments?

SippieCup:
```ffprobe --version
ffprobe version 3.4.6-0ubuntu0.18.04.1 Copyright (c) 2007-2019 the FFmpeg developers
```

pd0wm:
I'm very reluctant to change any of the frame reader code. It has been working for a long time, and it's used a lot in our training backend.

Does it work for your version if you use the `-f` flag instead of the undocumented `-format`?

pd0wm:
Fixed by https://github.com/commaai/openpilot/commit/cd79cee41a3120f867b7d89b8dd7d9b65073d275

SippieCup:
Ha! never realized i didnt push those changes for you. but yeah so minimal im glad you just did it. thanks!

PR #1316, opened 2020-04-05
Attempt to fetch FW versions with grey panda

KyleBoyer:
Issue: FW versions are unavailable to detect if the EPS is modded. These FW versions are unavailable because the editted `if` statement will not return anything if you do not `has_relay`(have a black panda).

Credit to: @wirelessnet2


geohot:
Sadly, without the OBD-II CAN bus, this is not possible on all cars. As such, this feature will not be supported in master for white/grey pandas, we recommend you upgrade to a black panda or comma two.

PR #1325, opened 2020-04-07
replace get_events with has_event in if-else statement

deanlee:
it's faster and more memory efficient.

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -13,8 +13,9 @@ from selfdrive.boardd.boardd import can_list_to_can_capnp
 from selfdrive.car.car_helpers import get_car, get_startup_alert
 from selfdrive.controls.lib.lane_planner import CAMERA_OFFSET
 from selfdrive.controls.lib.drive_helpers import get_events, \
-                                                 has_event, \
+                                                 get_events_type_set, \
                                                  create_event, \
+                                                 has_event, \
                                                  EventTypes as ET, \
                                                  update_v_cruise, \
                                                  initialize_v_cruise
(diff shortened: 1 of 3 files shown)


pd0wm:
How much cpu does this save?

deanlee:
it is negligible.
just want to avoiding unnecessary memory allocation in a long time looping.

selfdrive/controls/controlsd.py
```diff
@@ -168,37 +169,37 @@ def state_transition(frame, CS, CP, state, events, soft_disable_timer, v_cruise_
...
   elif state == State.enabled:
-    if get_events(events, [ET.USER_DISABLE]):
+    if has_event(events, [ET.USER_DISABLE]):
       state = State.disabled
       AM.add(frame, "disable", enabled)
 
-    elif get_events(events, [ET.IMMEDIATE_DISABLE]):
+    elif has_event(events, [ET.IMMEDIATE_DISABLE]):
       state = State.disabled
       for e in get_events(events, [ET.IMMEDIATE_DISABLE]):
```
pd0wm, line 178: The same lookup is done twice here. Can we combine them in a smart way? Maybe convert the events into a dict first of event type to list of events of that type.

pd0wm: In general there is a lot of repeated code, is it possible to abstract some of it into a function.

deanlee: I have added a set to do this. dict will create lists for types everytime we might not need.

geohot:
This adds complexity and 31 lines($1550), and I'm not sure what the gain is.

deanlee:
Oh I See

PR #1327, opened 2020-04-07
CAR.COROLLAH_TSS2 x018821F3301100\x00\x00\x00\x00

ErichMoraga:
@capthawker#8604 DongleID a362a5e990fe49e9


ErichMoraga:
I've implemented the change from my PR, locally on @capthawker#8604's C2, and confirmed it works.

PR #1329, opened 2020-04-07
cache file checksum for apks

deanlee:
Avoid calculating the checksum for apks on every reboot. make eon boot faster

### DIFF
--- a/common/apk.py
+++ b/common/apk.py
@@ -5,7 +5,7 @@ import hashlib
 import shutil
 import json
 from common.basedir import BASEDIR
-from common.params import Params, put_nonblocking
+from common.params import Params
 from selfdrive.swaglog import cloudlog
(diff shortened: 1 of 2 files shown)


pd0wm:
How much does it save of the boot time?

Adding this code also adds complexity, especially if we have to debug any installation issues in the future. The boot time savings need to be quite significant to be worth the complexity.

deanlee:
yes you r right. I just finished the test.it's only save 0.1 second :smile: 

before:
[0.5840582809996704, 0.5710070309996809, 0.5750553640000362, 0.5740222910003467, 0.5575822389996574]

after:
[0.45924635399978797, 0.4641881249999642, 0.4534554169999865, 0.5125296870000966, 0.4449574999998731]

PR #1332, opened 2020-04-08
improve error handling

deanlee:
BZ2_bzWriteClose is called to clean up after an error

### DIFF
--- a/selfdrive/loggerd/logger.cc
+++ b/selfdrive/loggerd/logger.cc
@@ -134,6 +134,8 @@ fail:
   LOGE("logger failed to open files");
   if (h->qlog_file) fclose(h->qlog_file);
   if (h->log_file) fclose(h->log_file);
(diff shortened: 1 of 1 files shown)


selfdrive/loggerd/logger.cc
```diff
@@ -132,8 +132,22 @@ static LoggerHandle* logger_open(LoggerState *s, const char* root_path) {
...
+  if (h->qlog_file) {
+    fclose(h->qlog_file);
+    h->qlog_file = NULL;
+  }
+  if (h->log_file) {
+    fclose(h->log_file);
+    h->log_file = NULL;
+  }
+  if (h->bz_file) {
+    BZ2_bzWriteClose(&bzerror, h->bz_file, 0, NULL, NULL);
```
pd0wm, line 144: Shouldn't the fcloses be moved to after the bzwrite? Otherwise it will write to a closed file?

deanlee: yes,my fault

PR #1335, opened 2020-04-08
Toyota: use universal gas pressed bit to match panda safety

adeebshihadeh:
Process replay shows two differences in gas pressed probably due to toyota using a slightly different threshold. These differences lead to differences in other packets, like carEvents as a pedalPressed event is added.


pd0wm:
Updated ref commit `9b4b08487380a6741646e22e191fa5bd1615f2d3`.

PR #1336, opened 2020-04-08
Show alert when camerad stops unexpectedly

pd0wm:


selfdrive/ui/ui.cc
```diff
@@ -1034,14 +1064,23 @@ int main(int argc, char* argv[]) {
...
         play_alert_sound(s->alert_sound);
+        s->alert_override = true;
       }
+
       s->alert_sound_timeout--;
       s->controls_seen = false;
     }
 
+    // stop playing alert override sound
+    if ((!s->started || (s->started && s->alert_sound_timeout == 0)) &&
```
pd0wm, line 1077: Add check if alert is playing

geohot:
Why do this in UI? Can't controls do it?

pd0wm:
Depends on what you want. Controls already has an indirectly check for the model (through pathPlanner). But that wouldn't trigger if the visionipc connection between ui and camerad crashes.

pd0wm:
All the changes from `vision_connected` to started need to be made anyway, otherwise it immediately goes offroad when camerad dies and you can't show the alert.

pd0wm:
I'll merge https://github.com/commaai/openpilot/pull/1351 first, then we can decide what to do about the alert.

Issue #1323, opened 2020-04-06
UI sidebar overlays on top of the github id keyboard input screen in settings


csouers:
**Describe the bug**
Sidebar doesn't hide when the keyboard dialog comes up. Likely related to https://github.com/commaai/openpilot/issues/1293

**How to reproduce or log data**
Open settings and click on text box to bring up the keyboard.

**Expected behavior**
Sidebar hides when keyboard is present

**Device/Version information (please complete the following information):**
 - Device: two
 - Version: master branch - f9257fc75f68c673f9e433985fbe739f23310bb4
 - Car make/model: All

PR #1337, opened 2020-04-08
offroad/ui: Transparent layout mode and msgq-based layout control

### LINKED ISSUE #1293: UI bug when going onroad while in advanced wifi settings
screenshot

andyh2:
* Reimplements ui->offroad layout control, using `offroadLayout` socket instead of android broadcasts. 
* Adds UiLayoutState.App.None, with which offroad will draw a transparent/invisible activity window instead of its normal content.  
* Fixes regression in onboarding where the green engaged border did not appear after pressing the "SET" button on screen. It had previously relied upon Android broadcasts, so this bug appeared with the frame refactor. Now we communicate a field in UiLayoutState. 
* New APK (https://github.com/commaai/openpilot-apks/pull/47) supports the above changes and also resolves https://github.com/commaai/openpilot/issues/1293, https://github.com/commaai/openpilot/issues/1323.


geohot:
Testing now

selfdrive/ui/ui.cc
```diff
@@ -964,19 +1008,24 @@ int main(int argc, char* argv[]) {
       enable_event_processing(true);
       if (s->status != STATUS_STOPPED) {
         update_status(s, STATUS_STOPPED);
+        s->active_app = cereal_UiLayoutState_App_home;
+        update_offroad_layout_state(s);
```
geohot, line 1012: I'm wondering if we should just send this every frame instead of on update.

geohot:
Hmm, why does settings still seem to be slow when I press the button?

geohot:
LGTM, nice job! Will merge this when the tests pass and fix up the bg thread PR

geohot (approved):

andyh2:
Offroad is responsible for the remaining latency in opening settings. I'm looking into what the bottleneck is there. We should end up with at most the same perceived latency as pre-frame-refactor

geohot:
Yea it actually seems a bit better.

PR #1339, opened 2020-04-09
add Jeep GC 2019 USA fingerprint from al_nyc

adhintz:
This adds these messages that are not in the existing Jeep GC 2019 fingerprint:
`678: 8, 729: 5, 766: 8, 800: 8, 810: 8, 863: 8`
If you'd prefer, could instead add them to the existing Jeep GC 2019 fingerprint.


pd0wm:
Yeah, let's just add them to the existing one.

adhintz:
Sounds good, done.

PR #1340, opened 2020-04-09
use map() to do type conversions

deanlee:


geohot:
I don't this this is more readable or faster, why should we merge it?

arne182:
@deanlee I gladly accept PR's
My github is now:
https://github.com/arne182/ArnePilot
You can find the updated version of mapd there. One that actually works as expected.

deanlee:
> I don't this this is more readable or faster, why should we merge it?
it's more pythonic and recommended method to perform this particular task, also  faster than using list comprehension.nearly 2x faster in my testing(10-1000 elements in list).

deanlee:
> I don't this this is more readable or faster, why should we merge it?

t's more pythonic and recommended method to perform this particular task, also faster than using list comprehension.nearly 2x faster in my testing(10-1000 elements in list).

deanlee:
> @deanlee I gladly accept PR's
> My github is now:
> https://github.com/arne182/ArnePilot
> You can find the updated version of mapd there. One that actually works as expected.

@arne182 Ok i will take a look

geohot:
So in microbenchmarks it may be faster, but is it faster in openpilot?

To merge a perf related PR, we should be able to measure an improvement in the system. To merge a refactor PR, it should decrease complexity.

PR #1342, opened 2020-04-09
do not duplicate end start nodes

brianczapiga:
If we already have pnts. way.points_in_car_frame will give us a duplicate pnt because it gives us the start of the new line where the node of the last line had that point already. We can safely delete the first point to avoid high curvature values with two points being on top of one another.


pd0wm:
mapd is currently not used/maitained. Closing this PR.

PR #1343, opened 2020-04-09
Add tool for splitting keras model for faster loading

SippieCup:
By splitting the Keras model, initialization and first frame output is reduced from over 2 minutes to under 20 seconds on Jetson TX2 hardware. The split model and original files are an exact binary match when loaded and does not change its output in any way.

This PR also provides the tool to split the models with the command:

```python ./tools/keras/split_keras.py ./models/supercombo.keras```


HaraldSchafer:
Lemme see if I can just use the json for everything. Don't want too many copies.

SippieCup:
They aren't copies, it is separating the model and the weights into two files.

splitting it into two files is what makes it load faster, putting everything into a single file will just make it take 3 minutes to load again.

ShaneSmiskol:
@SippieCup It takes just a few seconds for me. Which version of TF are you on?

Issue #1294, opened 2020-03-30
Add support for Hyundai radar -- $500 bounty


geohot:
Your goal is to implement the RadarInterface class, test it, and get it upstreamed: https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/radar_interface.py

Fine if it's only on cars with SCC. Must be from a physical radar, not from a camera or a fusion. Must work on comma's 2020 Sonata.

You'll have to figure out how to get points out of the radar. It's unclear how easy this is.

Currently we only train on cars with radars, this will add Hyundai to the training set.

Olshansk:
I saw this and looked at a few of the other RadarInterface implementations. There are a few commonalities, but the business logic seems to really vary.

I honestly don't think I have enough context to solve this, but I'm curious if you could at least give a quick pointer of how one would go about doing this? Do I need raw data to analyze as a test set?

TK211X:
Can the reward be a Comma 2 instead of $200?

geohot:
Increased bounty due to difficultly of getting radar points. I don't obviously see them on bus 1, the camera points don't count.

TK211X:
Inspection of a disconnect event from MFC demonstrates that 3 messages continue to be created by the Radar ECU.

These are `0x5ED` `0x5EE` `0x5EF`, they appear to serve the purpose of signaling the presence of the Radar ECU on LCAN and only contain a Counter and Checksum on all three messages.

The fallout of the messages on LCAN during MFC disconnect is as follows

43 messages ceased transmission @ .003 Mark / 0.030 FREQ

2 messages ceased transmission @ .013 Mark / 0.030 FREQ

7 messages ceased transmission @ .023 Mark / 0.030 FREQ

3 messages continued to transmit. / 0.050 FREQ

Next course of action is to attempt to trigger the Radar ECU into spitting out radar data.

Most LCAN status messages are very simple and just have a counter. It's very possible you could simply set up a script that goes through every ID on LCAN and generates a message for a few seconds containing a counter in each byte. Several could trigger a set of radar tracks to be output.

TK211X:
I can't seem to understand the difference in models that output Radar tracks onto LCAN and the Sonata where they they decided not to do this for whatever reason.

J-Rojas:
[...]

PR #1346, opened 2020-04-10
add radar interface for Hyundai/Kia/Genesis

AskAlice:
Co-authored-by: xx979xx <40252818+xx979xx@users.noreply.github.com>

I haven't tested this, but this should work.

RE: #1294 Pay xx979xx. He probably could use it more than I do atm considering the situation he's in.

If you'd like to do something for me for jumping through the hoops to make this happen (ie: translating the community code to the point where the CI checks pass), send me a dang replacement OTG dongle and bottom case part (with a higher infill if possible 😬 so easy to break... also would be happy with an STL for that lol) for the eon without making me send it in. I'm happy to pay for the parts and shipping, I just wish you'd let me buy those parts

### DIFF
--- a/selfdrive/car/hyundai/interface.py
+++ b/selfdrive/car/hyundai/interface.py
@@ -18,7 +18,9 @@ class CarInterface(CarInterfaceBase):
 
     ret.carName = "hyundai"
     ret.safetyModel = car.CarParams.SafetyModel.hyundai
-    ret.radarOffCan = True
+    ret.sccBus = 0 if 1056 in fingerprint[0] else 1 if 1056 in fingerprint[1] and 1296 not in fingerprint[1] \
+                                                                     else 2 if 1056 in fingerprint[2] else -1   ret.radarOffCan = ret.sccBus == -1
+
 
     # Hyundai port is a community feature, since we don't own one to test
     ret.communityFeature = True
--- a/selfdrive/car/hyundai/interface.py
+++ b/selfdrive/car/hyundai/interface.py
@@ -18,12 +18,7 @@ class CarInterface(CarInterfaceBase):
 
     ret.carName = "hyundai"
     ret.safetyModel = car.CarParams.SafetyModel.hyundai
-    #not in cereal/car.capnp but I don't wanna fork cereal so
-    #ret.sccBus = 0 if 1056 in fingerprint[0] else 1 if 1056 in fingerprint[1] and 1296 not in fingerprint[1] \
-    #                                                                 else 2 if 1056 in fingerprint[2] else - 1
-    ret.radarOffCan = False if 1056 in fingerprint[0] else False if 1056 in fingerprint[1] and 1296 not in fingerprint[1] else False if 1056 in fingerprint[2] else True
-    # ret.sccBus == -1
-
+    ret.radarOffCan = not (1056 in fingerprint[0])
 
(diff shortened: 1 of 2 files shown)


AskAlice:
Not sure why the planner would error, I didn't change the planner at all. Why is the MPC cost too high? it's not `mpc_nans` because I think it would have logged that in cloudlog.

Interested in seeing someone test this in-vehicle.

selfdrive/car/hyundai/interface.py
```diff
@@ -18,7 +18,12 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), has_relay=False,
 
     ret.carName = "hyundai"
     ret.safetyModel = car.CarParams.SafetyModel.hyundai
-    ret.radarOffCan = True
+    #not in cereal/car.capnp but I don't wanna fork cereal so
+    #ret.sccBus = 0 if 1056 in fingerprint[0] else 1 if 1056 in fingerprint[1] and 1296 not in fingerprint[1] \
+    #                                                                 else 2 if 1056 in fingerprint[2] else - 1
+    ret.radarOffCan = False if 1056 in fingerprint[0] else False if 1056 in fingerprint[1] and 1296 not in fingerprint[1] else False if 1056 in fingerprint[2] else True
```
pd0wm, line 24: can you use the `is_ecu_disconnected` helper function here? (selfdrive/car/__init__.py)

pd0wm: See https://github.com/commaai/openpilot/blob/master/selfdrive/car/toyota/interface.py#L246

selfdrive/car/hyundai/interface.py
```diff
@@ -18,7 +18,12 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), has_relay=False,
...
+    # ret.sccBus == -1
```
pd0wm, line 25: remove comments?

pd0wm:
Thanks for the PR!

Do you have a recent cabana route I can use to test? Unfortunately our Sonata isn't wired up yet.

selfdrive/car/hyundai/interface.py
pd0wm: You can also only check on bus 0 for now.

AskAlice:
@pd0wm I am unable to test this unfortunately, as my vehicle does not have SCC, but I've been working on hacking together SCC emulation for longitudional on all HKG vehicles with LKAS, which has gotten me familiar enough with this code to make such a PR.

selfdrive/car/hyundai/interface.py
AskAlice: i'm assuming this would be without `or has_relay` because some vehicles might not have SCC, and if we do ever implement SCC emulation then radar data wouldn't be a part of that.

AskAlice: resolved.

selfdrive/car/hyundai/interface.py
AskAlice: resolved with 264eb318171524d76cfaf864740d2c59524e7d0f

pd0wm:
This will go in with the main hkg PR, so I'm closing this.

Regarding the bounty, I think we were looking for a parser that parses the whole point cloud from the radar. I just realized this only gets one point. But I'm not sure on what that data will be present.

PR #1347, opened 2020-04-10
Remove bg_thread from ui

geohot:
Now that offroad can become invisible, this should work. Red diff!

Two minor issues remain:
1) If you start camerad, start ui, kill ui, kill camerad, start ui. offroad stays invisible.
2) If you go to settings onroad, then exit onroad, it stays in settings.


geohot:
@andyh2 The bug for 2 seems to be in offroad, it's staying on settings when transitioning from 4 -> 0

PR #1350, opened 2020-04-10
Abstract pcmEnable and pcmDisable events

adeebshihadeh:
also includes some pruning of `selfdrive/car/*`

### DIFF
--- a/selfdrive/manager.py
+++ b/selfdrive/manager.py
@@ -526,6 +526,9 @@ def uninstall():
 def main():
   os.environ['PARAMS_PATH'] = PARAMS
 
+  # the flippening!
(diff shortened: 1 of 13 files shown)


selfdrive/manager.py
```diff
@@ -526,9 +526,6 @@ def uninstall():
 def main():
   os.environ['PARAMS_PATH'] = PARAMS
 
-  # the flippening!
-  os.system('LD_LIBRARY_PATH="" content insert --uri content://settings/system --bind name:s:user_rotation --bind value:i:1')
```
quillford, line 530: Seems like this is done in NEOS now?

pd0wm: I'd have to check that. Can you leave it in for now?

quillford: done

pd0wm (approved):

PR #1356, opened 2020-04-13
Subaru: calculate checksum in can packer

adeebshihadeh:
Need to bump opendbc after merging https://github.com/commaai/opendbc/pull/241

I got no differences when running process replay locally with the opendbc changes.

### DIFF
--- a/selfdrive/car/subaru/subarucan.py
+++ b/selfdrive/car/subaru/subarucan.py
@@ -4,41 +4,28 @@ from selfdrive.car.subaru.values import CAR
 
 VisualAlert = car.CarControl.HUDControl.VisualAlert
 
-def subaru_checksum(packer, values, addr):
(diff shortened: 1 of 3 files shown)


selfdrive/car/subaru/subarucan.py
```diff
@@ -4,42 +4,29 @@
...
+    "LKAS_Output": apply_steer,
+    "LKAS_Request": 1 if apply_steer != 0 else 0,
+    "SET_1": 1
+  }
 
   return packer.make_can_msg("ES_LKAS", 0, values)
 
 def create_steering_status(packer, car_fingerprint, apply_steer, frame, steer_step):
-
-  if car_fingerprint == CAR.IMPREZA:
```
quillford, line 29: Is there a reason this was done? Impreza is the only supported subaru

PR #1363, opened 2020-04-14
rely on carstate to be sure

haraschax:


selfdrive/test/process_replay/process_replay.py
```diff
@@ -186,7 +187,14 @@ def radar_rcv_callback(msg, CP, cfg, fsm):
...
-  recv_socks = ["liveCalibration"] if (fsm.frame + 1) % 5 == 0 else []
+  recv_socks = []
+  if msg.which() == 'carState':
+    if not hasattr(fsm, 'cs_counter'):
+      fsm.cs_counter = 0
+    if (fsm.cs_counter % 25) == 0:
+      recv_socks = ["liveCalibration"]
+    fsm.cs_counter += 1
+  print(msg.which(), fsm.cs_counter, recv_socks)
   return recv_socks, True
```
quillford, line 197: I think if you only

HaraldSchafer:
@pd0wm Pretty hacky way to get test to work, is there a better way to do this?

selfdrive/test/process_replay/process_replay.py
```diff
@@ -186,7 +187,13 @@ def radar_rcv_callback(msg, CP, cfg, fsm):
...
   # should_recv always true to increment frame
-  recv_socks = ["liveCalibration"] if (fsm.frame + 1) % 5 == 0 else []
+  recv_socks = []
+  if msg.which() == 'carState':
+    if not hasattr(fsm, 'cs_counter'):
+      fsm.cs_counter = 0
+    if (fsm.cs_counter % 25) == 0:
+      recv_socks = ["liveCalibration"]
+    fsm.cs_counter += 1
   return recv_socks, True
```
quillford, line 197: I think if you only return True for should_recv when it's a carState msg, you can just use fsm.frame instead of adding this counter.

Something like:
```
def calibration_rcv_callback(msg, CP, cfg, fsm):
  recv_socks = ["liveCalibration"] if (fsm.frame + 1) % 25 == 0 else []
  return recv_socks, msg.which() == 'carState'
```

HaraldSchafer: ah great, thanks!

quillford:
looks like the differences are in the subaru checksum byte, so updating opendbc then updating the refs should fix this.

HaraldSchafer:
Thanks for the help @quillford !

PR #1368, opened 2020-04-15
release resources when thread is done

deanlee:
useless but good practice


pd0wm:
I need to do some testing with messaging in ZMQ mode before I can merge this. If you close a ZMQ context, but forget to close on of the sockets first it hangs forever.

PR #1369, opened 2020-04-15
fix event_handler:unlock a mutex that has not been locked

deanlee:
From the [document](https://linux.die.net/man/3/pthread_mutex_unlock) about pthread_mutex_unlock:

>  If a thread attempts to unlock a mutex that it has not locked or a mutex which is unlocked, undefined behavior results.


pd0wm:
Good catch. That looks like something that could cause issues.

PR #1376, opened 2020-04-16
Updater read git config first

pd0wm:
This should reduce the number of unneeded changes to `.git/config` reducing the chances of corruption.

Also does the change in the overlay instead of in the current copy.


jyoung8607:
LGTM!

PR #1377, opened 2020-04-16
Update values.py

cowanhm:
Request to add HR-V fingerprints to /honda/values.py
Generated from /fingerprint_helper.py

### DIFF
--- a/selfdrive/car/honda/values.py
+++ b/selfdrive/car/honda/values.py
@@ -34,6 +34,7 @@ class CAR:
   CRV_EU = "HONDA CR-V 2016 EXECUTIVE"
   CRV_HYBRID = "HONDA CR-V 2019 HYBRID"
   FIT = "HONDA FIT 2018 EX"
+  HRV = "HONDA HRV 2019 TOURING"
   ODYSSEY = "HONDA ODYSSEY 2018 EX-L"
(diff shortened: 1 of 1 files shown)


vanillagorillaa:

Also, the fingerprint which you have is the "V1" fingerprint. We are only merging "V2" fingerprinting which is based off of ECU firmware versions. See https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py

selfdrive/car/honda/values.py
```diff
@@ -695,6 +699,7 @@ class CAR:
   CAR.CRV_EU: dbc_dict('honda_crv_executive_2016_can_generated', 'acura_ilx_2016_nidec'),
   CAR.CRV_HYBRID: dbc_dict('honda_crv_hybrid_2019_can_generated', None),
   CAR.FIT: dbc_dict('honda_fit_ex_2018_can_generated', 'acura_ilx_2016_nidec'),
+  CAR.HRV: dbc_dict('honda_fit_ex_2018_can_generated', 'acura_ilx_2016_nidec'),
```
vanillagorillaa, line 702: You will want an HRV specific DBC for this. Pull requests can be made to the opendbc repo, https://github.com/commaai/opendbc

PR #1378, opened 2020-04-16
Update carstate.py

cowanhm:
Updated for HRV setting.
Door signal has not been discovered in F-CAN.
Safety verified using seatbeltUnlatched for disengagement


vanillagorillaa:

PR #1379, opened 2020-04-16
Update interface.py

cowanhm:
Update for Honda HRV 2019 Touring (AWD)
Tuning based on actual vehicle specifications, and real-world driving.


vanillagorillaa:

PR #1380, opened 2020-04-16
Honda HRV

cowanhm:
PR / Update for "Honda HRV 2019 Touring"
1. values.py - incl. Class, FP, FW_VERSIONS,  DBC Identical to CAR.FIT, SPEED_FACTOR 1.025
2. carstate.py - incl. carFingerprint shared with CAR.FIT, doorOpen OVERRIDE (signal unknown)
      Safety test concludes that removal of drivers seatbelt IMMEDIATELY disengages OP
3. interface.py - incl. HR-V vehicle parameters, tuning (tested for about 100 miles)
      Longitudinal tuning parameters unchanged from CAR.FIT

Thanks, Cowan Moore 330-701-3922

### DIFF
--- a/selfdrive/car/honda/values.py
+++ b/selfdrive/car/honda/values.py
@@ -34,6 +34,7 @@ class CAR:
   CRV_EU = "HONDA CR-V 2016 EXECUTIVE"
   CRV_HYBRID = "HONDA CR-V 2019 HYBRID"
   FIT = "HONDA FIT 2018 EX"
+  HRV = "HONDA HRV 2019 TOURING"
   ODYSSEY = "HONDA ODYSSEY 2018 EX-L"
   ODYSSEY_CHN = "HONDA ODYSSEY 2019 EXCLUSIVE CHN"
   ACURA_RDX = "ACURA RDX 2018 ACURAWATCH PLUS"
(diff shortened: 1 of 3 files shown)


selfdrive/car/honda/values.py
```diff
@@ -93,6 +94,9 @@ class CAR:
   CAR.FIT: [{
     57: 3, 145: 8, 228: 5, 304: 8, 342: 6, 344: 8, 380: 8, 399: 7, 401: 8, 420: 8, 422: 8, 427: 3, 428: 8, 432: 7, 464: 8, 487: 4, 490: 8, 506: 8, 597: 8, 660: 8, 661: 4, 773: 7, 777: 8, 780: 8, 800: 8, 804: 8, 808: 8, 829: 5, 862: 8, 884: 7, 892: 8, 929: 8, 985: 3, 1024: 5, 1027: 5, 1029: 8, 1036: 8, 1039: 8, 1108: 8, 1322: 5, 1361: 5, 1365: 5, 1424: 5, 1600: 5, 1601: 8
   }],
+  CAR.HRV: [{
```
vanillagorillaa, line 97: This section can actually get removed.

cowanhm: Yes sir, I realize that.  If it's possible to leave as a reference for other hrv'ers, as there is still some code that helps people discover their fp's.  I'm not sure the differences between HRV's at this point. 

Can we remove after further hrv users review?  Thanks

vanillagorillaa: Actually looks like we do need it for other stuff. So leaving it is fine

cowanhm:
Planning to conduct further testing with other commaai/openpilot branches

PR #1382, opened 2020-04-17
Car Support - 2020 Hyundai Palisade

scottadair:


selfdrive/car/hyundai/carcontroller.py
```diff
@@ -96,8 +96,7 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert,
 
 
     # 20 Hz LFA MFA message
-    if frame % 5 == 0 and self.car_fingerprint == CAR.SONATA:
+    if frame % 5 == 0 and (self.car_fingerprint == CAR.SONATA or self.car_fingerprint == CAR.PALISADE) :
```
pd0wm, line 99: can you change this to:
```suggestion
    if frame % 5 == 0 and self.car_fingerprint in [CAR.SONATA, CAR.PALISADE]:
```

That makes it cleaner to all cars

selfdrive/car/hyundai/hyundaican.py
```diff
@@ -19,7 +19,7 @@ def create_lkas11(packer, frame, car_fingerprint, apply_steer, steer_req,
   values["CF_Lkas_MsgCount"] = frame % 0x10
   values["CF_Lkas_Chksum"] = 0
 
-  if car_fingerprint == CAR.SONATA:
+  if (car_fingerprint == CAR.SONATA) or (car_fingerprint == CAR.PALISADE):
```
pd0wm, line 22:
```suggestion
  if car_fingerprint in [CAR.SONATA, CAR.PALISADE]:
```

pd0wm:
Do you have a cabana route we can use for the test_car_models test?

scottadair:
The changes make sense, I've accepted and committed them. For the cabana route, what do you need for that? Shall I just go for a drive, enable OP and then send you the cabana link?

pd0wm:
It can be any drive, just not too long (max 20 segments).

PR #1385, opened 2020-04-17
Fix UnicodeDecodeError in get_network_strength

deanlee:
subprocess.check_output will throw an UnicodeDecodeError  if the result contain non-ASCII characters, such as chinese or japense.

for example,the result from China Telecom Card:
>   Phone Id=0
  mCallState=0
  mCallIncomingNumber=
  mServiceState=0 0 voice home data home 中国电信 CT 46011 中国电信 CT 46011  1xRTT LTE_CA CSS not supported 302 14137 RoamInd=1 DefRoamInd=0 EmergOnly=false IsDataRoamingFromRegistration=false
  mSignalStrength=SignalStrength: 99 0 -68 -70 -120 -1 -1 26 -90 -9 228 2147483647 2147483647 gsm|lte

### DIFF
--- a/common/android.py
+++ b/common/android.py
@@ -232,7 +232,7 @@ def get_network_strength(network_type):
   if network_type == NetworkType.none:
     return network_strength
   if network_type == NetworkType.wifi:
-    out = subprocess.check_output('dumpsys connectivity', shell=True).decode('ascii')
+    out = subprocess.check_output('dumpsys connectivity', shell=True).decode('utf-8')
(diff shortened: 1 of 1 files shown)


jyoung8607:
This is definitely a good thing to fix. "dp" on Discord was hitting this same thing just a couple days ago, choking on `dumpsys connectivity` output, with a Unicode non-breaking space in his WiFi SSID instead of the carrier name.

PR #1388, opened 2020-04-17
0.7.5 release

pd0wm:


andyh2 (approved):
LGTM. Panda safety tests are so much more readable 👍

pd0wm:
All thanks to @quillford !

README.md
```diff
@@ -57,106 +57,123 @@ openpilot should preserve all other vehicle's stock features, including, but are
...
-| Kia       | Sorento 2018<sup>1</sup>      | All               | Stock            | 0mph               | 0mph              |
-| Kia       | Stinger 2018<sup>1</sup>      | SCC + LKAS        | Stock            | 0mph               | 0mph              |
-| Lexus     | CT Hybrid 2017-18             | All               | Stock<sup>5</sup>| 0mph               | 0mph              |
+| Honda     | Odyssey 2018-20               | Honda Sensing     | openpilot        | 25mph<sup>1</sup>  | 0mph              |
+| Honda     | Passport 2019                 | All               | openpilot        | 25mph<sup>1</sup>  | 12mph             |
+| Honda     | Pilot 2016-18                 | Honda Sensing     | openpilot        | 25mph<sup>1</sup>  | 12mph             |
+| Honda     | Pilot 2019                    | All               | openpilot        | 25mph<sup>1</sup>  | 12mph             |
+| Honda     | Ridgeline 2017-20             | Honda Sensing     | openpilot        | 25mph<sup>1</sup>  | 12mph             |
+| Hyundai   | Sonata 2020                   | All               | Stock            | 0mph               | 0mph              |
```
geohot, line 84: These shouldn't be here until we sell a harness for them.

pd0wm: Oops, though those were put in the community maintained section. We also need to do FW queries, and some final tuning before we promote them to official support.

Will fix before shipping

pd0wm: Fixed

geohot:
This should fix the LGTM analysis: https://github.com/commaai/openpilot/commit/bb9ca1ebd67f3f0d737f1171231cdbfd39c10b51

selfdrive/ui/sound.cc
```diff
@@ -13,28 +13,34 @@ extern "C"{
 #include "slplay.h"
 }
 
+int last_volume = 0;
+
 void set_volume(int volume) {
-  char volume_change_cmd[64];
-  sprintf(volume_change_cmd, "service call audio 3 i32 3 i32 %d i32 1 &", volume);
+  if (last_volume != volume) {
```
adeebshihadeh, line 19: are we sure we're the only ones changing the volume? do the volume buttons do anything?

adeebshihadeh: Tested on an EON, volume buttons have no effect

geohot:
Can we disable locationd on this branch? It isn't used yet (confirm?), and we don't have the CPU for shadow mode.

Currently at 81% looking at the corolla, w/o locationd we'd be back at 73%. I don't like releases above 75%.

pd0wm:
Updated PR with those changes

@geohot LGTM is still failing

geohot:
Huh, now LGTM fails because Devel can't be built on x86.

geohot:
I'm still working on the LGTM thing. The JavaScript X is a bug, and you seem to have to contact them to disable JavaScript for your project.

The C++ X is because we don't include the external/bin dir with capnp. Moving to capnp 0.7.0 will fix this, and we should be able to use the system capnp with capnpc-c gone.

PR #1389, opened 2020-04-17
use  test_bit to determining if a device is touchscreen

deanlee:
current implementation may cause assert in touch_init?(not sure):
https://github.com/commaai/openpilot/blob/09a342c52ff77205ecc5a3753babd8d2c27af791/selfdrive/common/touch.c#L31-L36

### DIFF
--- a/selfdrive/common/touch.c
+++ b/selfdrive/common/touch.c
@@ -10,6 +10,13 @@
 
 #include "touch.h"
 
+/* this macro is used to tell if "bit" is set in "array"
+ * it selects a byte from the array, and does a boolean AND
+ * operation with a byte that only has the relevant bit set.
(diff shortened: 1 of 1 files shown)


pd0wm:
Don't think this is worth it since it's used only once.

What do you mean it can cause an assert?

deanlee:
the bit mask generated by (ABS_MT_POSITION_X - x_key) is incorrect.

> #define ABS_MT_POSITION_X	0x35	/* Center X touch position */
> #define ABS_MT_POSITION_Y	0x36	/* Center Y touch position */
> 

ABS_MT_POSITION_X - x_key = 0b101111 
ABS_MT_POSITION_Y - y_key = 0b110000
1<<(ABS_MT_POSITION_X%8) = 0b100000
1<<(ABS_MT_POSITION_Y%8) = ob1000000

PR #1410, opened 2020-04-22
dommonmodel.c: unmap a previously mapped buff

### LINKED ISSUE #1397: clEnqueueMapBuffer without clEnqueueUnmapMemObject
clEnqueueMapBuffer is getting called repeatedly without clEnqueueUnmapMemObject 

https://github.com/commaai/openpilot/blob/28d8b4abac1a97905a74ea126d275125d2e93cd0/selfdrive/modeld/models/commonmodel.c#L36-L54

https://github.com/commaai/openpilot/blob/28d8b4abac1a97905a74ea126d275125d2e93cd0/selfdrive/modeld/models/driving.cc#L103

from the [document](https://www.khronos.org/registry/OpenCL/sdk/2.0/docs/man/xhtml/clEnqueueMapBuffer.html) of clEnqueueMapBuffer:

>if a memory object is currently mapped for reading, the application must ensure that the memory object is unmapped before any enqueued kernels or commands that write to this memory object or any of its associated memory objects (sub-buffer or 1D image buffer objects) or its parent object (if the memory object is a sub-buffer or 1D image buffer object) begin execution; otherwise the behavior is undefined.

deanlee:
resolve #1397

### DIFF
--- a/selfdrive/modeld/models/driving.cc
+++ b/selfdrive/modeld/models/driving.cc
@@ -112,6 +112,8 @@ ModelDataRaw model_eval_frame(ModelState* s, cl_command_queue q,
     assert(1==2);
   #endif
 
+  clEnqueueUnmapMemObject(q, s->frame.net_input, (void*)new_frame_buf, 0, NULL, NULL);
+
   // net outputs
   ModelDataRaw net_outputs;
   net_outputs.path = &s->output[PATH_IDX];


geohot:
oh wow, it was just this one that was broken. the others seem to do it right

geohot:
oh wow, it was just this one that was broken. the others seem to do it right

PR #1411, opened 2020-04-22
Update README.md w/ HRV in Supported Car List

cowanhm:
Adding Honda HRV to Supported Car List in root of openpilot (README.md)


pd0wm:
The official spelling is HR-V. Can you add the dash?

cowanhm:
Changes made per request.  Expanded to 2017-29 for vehicles with Honda Sensing, which based on literature will include LKAS & ACC

pd0wm:
As far I can tell sensing was only an option form 2019 and onwards. Can you change the years back to 2019.

cowanhm:
Per request, updated to narrow scope to model year 2019- equipped with Honda Sensing

PR #1418, opened 2020-04-24
Bypass alignment related copy whenever possible

deanlee:
The memory returned by msg->getData() is most likely already be aligned. add class AlignedArray to bypass alignment related copy whenever possible.


geohot:
Hmm this is a nice cleanup.

Instead of a class, can it be a function that returns a kj::heapArray<capnp::word>?

geohot:
In fact, I'm reading this pattern. Could it return a "cereal::Event::Reader" and that would save even more lines?

deanlee:
we still need a class to free allocated memory in destructor.currently, I can't figure out how to return cereal::Event::Reader from a single function.

geohot:
Hmm, then maybe a class is okay, and we put the cereal::Event::Reader functionality into it.

deanlee:
how about this?exactly the same lines - what a coincidence lol.

deanlee:
The FlatArrayMessageReader is the owner of the message content,and only can be initialized from constructor. 
if  we put cereal::Event::Reader into class, we have to create FlatArrayMessageReader in heap that may make things more complicated.

I've modified the previous code to reduce lines.

selfdrive/locationd/paramsd.cc
```diff
@@ -105,17 +105,11 @@ int main(int argc, char *argv[]) {
...
-      auto amsg = kj::heapArray<capnp::word>((msg->getSize() / sizeof(capnp::word)) + 1);
-      memcpy(amsg.begin(), msg->getData(), msg->getSize());
-
-      capnp::FlatArrayMessageReader capnp_msg(amsg);
-      cereal::Event::Reader event = capnp_msg.getRoot<cereal::Event>();
-
-      localizer.handle_log(event);
-
-      auto which = event.which();
+      AlignedMessage amsg = s->receive();
```
deanlee, line 108: (implicit)default copy constructor be called

selfdrive/common/alignedmessage.h
```diff
@@ -0,0 +1,47 @@
...
+
+    if (size % sizeof(capnp::word) != 0) buf_size = (size / sizeof(capnp::word) + 1) * sizeof(capnp::word);
+
+    if (((reinterpret_cast<uintptr_t>(message->getData())) % sizeof(capnp::word) != 0) || buf_size != size) {
+      // malloc is defined as being word aligned
+      buf = (char *)malloc(buf_size);
+      memcpy(buf, message->getData(), size);
+      LOGD("message is not aligned");
+    }
+    msg_reader = new capnp::FlatArrayMessageReader(kj::ArrayPtr<const capnp::word>(reinterpret_cast<capnp::word *>(buf ? buf : msg->getData()), buf_size / sizeof(capnp::word)));
```
deanlee, line 26: should figure out how to avoid alloc FlatArrayMessageReader in heap

deanlee: I'm still working on this.

deanlee:
still not good.need to figure out a better pattern.

selfdrive/loggerd/loggerd.cc
```diff
@@ -654,21 +656,14 @@ int main(int argc, char** argv) {
   while (!do_exit) {
     for (auto sock : poller->poll(100 * 1000)){
       while (true) {
-        Message * msg = sock->receive(true);
-        if (msg == NULL){
+        AlignedMessage amsg(sock->receive(true));
```
deanlee, line 659: only data from frame_sock  need to be parsed.alignedmesage is not suitable for this situation

PR #1421, opened 2020-04-24
angles instead of vp

haraschax:


pd0wm:
Do you want to handle migrating from existing calibration file?

HaraldSchafer:
Is that worth it? Only takes 2 minutes to re calibrate, seems ok to me.

pd0wm:
Let's do it cleanly. We can remove the code the release after.

HaraldSchafer:
No don't want that. I will just handle the migration then.

PR #1425, opened 2020-04-26
Add HIGHLANDERH_TSS2 Support

ErichMoraga:


pd0wm:
Do you have a route we can add to test_car_models?

Can you also add the car to the list of IGNORED_FINGERPRINTS in values.py so it only fingerprints with fingeprint 2.0?

ErichMoraga (commented):
Added to IGNORED_FINGERPRINTS

Issue #1238, opened 2020-03-13
All "Supported Cars" need to move to fingerprinting 2.0


geohot:
Move fingerprinting 1.0, and subsequently white/grey panda, to "Community Maintained." Honda and Toyota are already good.

- [ ] Chrysler/Jeep
- [ ] Hyundai/Kia
- [ ] Subaru
- [ ] Volkswagen

After these 4 are migrated, we can move all of fingerprinting 1.0 to community maintained.

pd0wm:
Do you want to try fingerprinting 2.0 through a giraffe? Then we can remove the code completely.

arne182:
To maintain this. Which I don't mind doing I would like to have access our at least get the sentry unrecognised car errors. The partial can messages that are captured give a good enough indication of the car type and possible missing signals

jyoung8607:
> Move fingerprinting 1.0, and subsequently white/grey panda, to "Community Maintained." Honda and Toyota are already good.
>
> * [ ]  Volkswagen

I am actively looking at this, because fingerprinting 1.0 doesn't really work for VW anyway. The current OP and Panda example UDS query code tries to work, and it elicits partial responses from a few addresses, but it doesn't fully succeed on anything at this time.

This week I'll be snooping diagnostic comms from VCDS to see what the difference is. I have a hypothesis that some things are fine with UDS over ISO-TP but some things may need UDS over VW/Bosch CAN TP 1.6 or 2.0, because German Engineering™. To be verified.

Once we get comms working to the steering rack, I'm confident we can use it for high quality fingerprinting. The racks are widely shared between vehicles, but can be queried for a four-digit number that identifies what assist characteristic map it was loaded with, which lets us discriminate between harder cases like the 16,384 VW Golf variants, etc.

martinl:
I have done some testing with FW fingerprinting on Subaru Crosstrek 2018

Current state is:
- fw_versions.py script hangs at VIN query when running on OBD2 port (can1) 
- fw_versions.py runs on car main can0 and returns VIN but no FW version results for Honda and Toyota queries

I was able to extract responding ECU addresses from fw_versions.py debug output and ran some ISO-TP queries using panda
- All discovered ECU-s respond to following Subaru vendor-specific queries
```
SSMID (22 F1 00) - 3 byte ECU identifier
ROMID (22 F1 82) - 4-5 byte ECU ROM identifier
System String (22 F1 97) - ECU description string (eg Power Steering System)
```
[...]

PR #1427, opened 2020-04-27
Discussion-only PR for Volkswagen FPv2

jyoung8607:
Do not merge, discussion only. CI expected to fail.

This is an example of minimally viable FPv2 for Volkswagen MQB. It adds support for different rx_addr offsets to the UDS and ISO-TP code in a way that tries to avoid affecting queries for other cars. Identification is done by querying the steering characteristic parameters from the EPS rack. A version of this does work on my development vehicle.

Further discussion: https://github.com/commaai/openpilot/issues/1238#issuecomment-620142101

**TL;DR:** I am not in favor of the steering-rack parameter ONLY approach to identification, but I'll implement it if that's what Comma wants. I still strongly advocate for using the VIN, backed up by UDS validation of versions, as the most correct and future-proof solution. It could be done in a way that only affects car port(s) that want to use the VIN as part of the identity mix.

Thoughts requested from @gregjhogan and @pd0wm based on our DM discussions.

### DIFF
--- a/selfdrive/car/fw_versions.py
+++ b/selfdrive/car/fw_versions.py
@@ -3,7 +3,7 @@ import traceback
 import struct
 from tqdm import tqdm
 
-from selfdrive.car.isotp_parallel_query import IsoTpParallelQuery
+from selfdrive.car.isotp_parallel_query import IsoTpParallelQuery, RX_OFFSET
 from selfdrive.swaglog import cloudlog
 from selfdrive.car.fingerprints import FW_VERSIONS
 from selfdrive.car.toyota.values import CAR as TOYOTA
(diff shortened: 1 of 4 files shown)


jyoung8607:
This would also include a change to the Panda UDS library at this location:

https://github.com/commaai/panda/blob/ba59ada0e95df66bcca28d858fbf2daf6a2a7e40/python/uds.py#L471

To this:

```python
def get_rx_addr_for_tx_addr(tx_addr, rx_offset=0x8):
  if tx_addr in FUNCTIONAL_ADDRS:
    return None

  if tx_addr < 0xFFF8:
    # Standard 11 bit response addr (add 8), unless overridden
    # Some manufacturers use different offsets
    return tx_addr + rx_offset
```

Just didn't want to carry around a second Panda PR for this discussion.

pd0wm:
As discussed:

All MQB cars have the same ecu fw versions, and they speak the same API. For openpilot there will be only one car. CAR.VW_MQB. The only place we need to distinguish between car models is in the get_params to get to correct wheelbase and tuning params. For now we'll use VIN there, if in the future that doesn't work we can very easily switch to something else (e.g. the EPS has an dataflash id). By solving it this way the FW query is the same as all other cars, and the VIN code is very contained.

jyoung8607:
Withdrawing for now, per our discussion. Will take another run at this in late May/early June.

PR #1429, opened 2020-04-28
Ambient light calibration parameters on persist partition

robbederks:
This allows us to fine-tune the light sensor calibration in production in a persistent way

### DIFF
--- a/selfdrive/common/params.cc
+++ b/selfdrive/common/params.cc
@@ -92,13 +92,11 @@ int write_db_value(const char* key, const char* value, size_t value_size, bool p
   // Make sure params path exists
   result = ensure_dir_exists(params_path);
   if (result < 0) {
-    printf("1\n");
     goto cleanup;
   }
 
   result = snprintf(path, sizeof(path), "%s/d", params_path);
   if (result < 0) {
-    printf("2\n");
     goto cleanup;
   }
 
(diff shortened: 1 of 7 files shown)


pd0wm:
This will break the python implementation of the params. /data/params/d is not a folder, but a symlink to a folder containing the params. This alows atomic swaps of the params folder.

You want to create a temp folder `/data/params/.tmp<random string>` and then create `/data/params/d` as a symlink to that.

selfdrive/ui/ui.cc
```diff
@@ -938,7 +945,11 @@ int is_leon() {
   return strstr(str, "letv") != NULL;
 }
 
-
+#ifdef QCOM
```
pd0wm, line 948: can we handle the persistent params path in `params.cc`? Maybe pass in a flag to `read_db_value` to tell if it's a persistent param. It's all in C++, so you can add an argument with a default value, so the other code using it wouldn't need to change.

selfdrive/common/params.cc
```diff
@@ -57,8 +63,16 @@ static int fsync_dir(const char* path){
   }
 }
 
-int write_db_value(const char* params_path, const char* key, const char* value,
-                   size_t value_size) {
+static int ensure_dir_exists(const char* path) {
+  struct stat st;
+  if (stat(path, &st) == -1) {
+    printf("Creating folder: %s\n", path);
```
pd0wm, line 69: Remove debug print

PR #1430, opened 2020-04-28
Update Elantra Values one last time

TK211X:
I got the params learner values from EON. These are the values it settled on after driving over a week.

### DIFF
--- a/selfdrive/car/hyundai/interface.py
+++ b/selfdrive/car/hyundai/interface.py
@@ -65,7 +65,7 @@ class CarInterface(CarInterfaceBase):
       ret.lateralTuning.pid.kf = 0.00006
       ret.mass = 1275. + STD_CARGO_KG
       ret.wheelbase = 2.7
-      ret.steerRatio = 13.73   #Spec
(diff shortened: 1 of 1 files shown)


selfdrive/car/hyundai/interface.py
```diff
@@ -65,8 +65,8 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), has_relay=False,
       ret.lateralTuning.pid.kf = 0.00006
       ret.mass = 1275. + STD_CARGO_KG
       ret.wheelbase = 2.7
-      ret.steerRatio = 13.73   #Spec
-      tire_stiffness_factor = 0.385
+      ret.steerRatio = 15.4         # 14 is Stock | Settled Params Learner values are steerRatio: 15.401566348670535 /stiffnessFactor: 1.0081302973865127 for 2017 Elantra
+      tire_stiffness_factor = 1.0
```
pd0wm, line 69: Did the stiffness factor from the learner settle on 1.0? That's a multiplier on the stiffness_factor in the car params, so that means the 0.385 was correct.

PR #1432, opened 2020-04-28
better ui_draw()

deanlee:


pd0wm:
What exactly did you change here? Why is it better?

deanlee:
less code and function calls to do the same thing with better order. remove the following "empty" code block that looks like a patch.
https://github.com/commaai/openpilot/blob/ae37626ecdf7ed08e7f31af364b7f4fb22600946/selfdrive/ui/paint.cc#L940-L949


you'r welcome! it's my pleasure.

pd0wm:
This PR changes the look of the ui. Can you make sure it looks the same?

Before:
Image from iOS


After:
Image from iOS (1)

deanlee:
fixed. 

sorry for my mistake.  the left border is drawn by ui_draw_sidebar(), which I had not called if sidebar is collapsed.

the  ui_draw_sidebar_background() is the only necessary function when sidebarcollapsed is true,so I added a conditional return in ui_draw_sidebar.

PR #1436, opened 2020-04-29
DLC models sanity check

ZwX1616:
TODO: 
- add driving
- run flag somewhere in test


adeebshihadeh:
Closing as we have a modeld replay test now, which will get a dmonitoringmodeld replay soon.

PR #1441, opened 2020-04-30
Palisade Settings Tweak

scottadair:
Tweek the steerRatio and set the minSteerSpeed to 0


pd0wm:
That value is automatically learned by the params learner, this value only defines the initial value. Such a small change should not make a difference.

I cleaned up the min steer speed on master.

PR #1446, opened 2020-05-01
simplify drawing text and rect

deanlee:


geohot:
So one of the issues with merging this is we don't have tests for the UI. Ideally, since this shouldn't make any pixel changes to the UI, we should write a test framework which renders it and writes the output to a png which is then diffed.

Reading it I don't see any issues. If you feel you've tested it well, I will merge.

Thanks again for submitting these PRs, I feel the code getting better!

deanlee:
Yes u r right, it's hard to do this type of testing.I have checked codes many times to make sure the UI will be the same as before.I think it is ready to be merged. 

I learned a lot from this great opensource project,thank you too.

PR #1452, opened 2020-05-02
Using lgtm.com and fixing found alerts

geohot:
Let's improve code quality!

Many unused imports, issues in class inits, and duplicate lines.


geohot:
@pd0wm do you know why this is failing? I see "tmppilot.tar.gz" with 0 bytes in the "push", but it's okay in the "pull_request"

PR #1455, opened 2020-05-02
UI: move c-capnp to c++

deanlee:
move to capnproto

all function and logic is strictly the same as C version.

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -737,7 +737,7 @@ static void ui_draw_vision(UIState *s) {
   if (scene->alert_size != ALERTSIZE_NONE) {
     // Controls Alerts
     ui_draw_vision_alert(s, scene->alert_size, s->status,
-                            scene->alert_text1, scene->alert_text2);
+                            scene->alert_text1.c_str(), scene->alert_text2.c_str());
   } else {
     if (!scene->frontview){ui_draw_vision_footer(s);}
   }
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -1,5 +1,5 @@
-#include <assert.h>
 #include "ui.hpp"
+#include <assert.h>
 
 #include "common/util.h"
 
@@ -759,7 +759,7 @@ void ui_draw(UIState *s) {
   glViewport(0, 0, s->fb_w, s->fb_h);
   nvgBeginFrame(s->vg, s->fb_w, s->fb_h, 1.0f);
   ui_draw_sidebar(s);
-  if (s->started && s->active_app == cereal_UiLayoutState_App_none && s->status != STATUS_STOPPED && s->vision_seen) {
+  if (s->started && s->active_app == cereal::UiLayoutState::App::NONE && s->status != STATUS_STOPPED && s->vision_seen) {
       ui_draw_vision(s);
   } 
   nvgEndFrame(s->vg);
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -1,6 +1,6 @@
 #include "ui.hpp"
 #include <assert.h>
-
+#include <map>
 #include "common/util.h"
 
 #define NANOVG_GLES3_IMPLEMENTATION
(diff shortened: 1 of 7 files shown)


deanlee:
Done, tested for a few miles.

selfdrive/ui/ui.cc
```diff
@@ -387,207 +364,146 @@ static void update_status(UIState *s, int status) {
...
-      s->scene.satelliteCount = reportdatad.numMeas;
+  } else if (which == cereal::Event::LIVE_MAP_DATA) {
+    scene.map_valid = event.getLiveMapData().getMapValid();
+  } else if (which == cereal::Event::THERMAL) {
+    auto data = event.getThermal();
+    
+    scene.networkType = data.getNetworkType();
+    scene.networkStrength = data.getNetworkStrength();
+    scene.batteryPercent = data.getBatteryPercent();
+    scene.batteryCharging = data.getBatteryStatus() == "Charging" ? true : false;
```
geohot, line 482: ? true : false doesn't do anything

selfdrive/ui/ui.cc
```diff
@@ -387,207 +364,146 @@ static void update_status(UIState *s, int status) {
...
+        s->alert_type = data.getAlertType();
+      }
+      s->alert_sound = alert_sound;
     }
-
-    if (datad.alertStatus == cereal_ControlsState_AlertStatus_userPrompt) {
+    scene.alert_text1 = data.getAlertText1();
+    scene.alert_text2 = data.getAlertText2();
+    scene.alert_ts = event.getLogMonoTime();
+    scene.alert_size = (uint8_t)data.getAlertSize();
```
geohot, line 405: Don't cast this, it's making an assumption of the numbers of the capnp enum. Instead, we should remove the ALERTSIZE_NONE... defines and use the cereal ones everywhere.

selfdrive/ui/sidebar.cc
```diff
@@ -56,8 +55,8 @@ static void ui_draw_sidebar_network_type(UIState *s) {
   const char *network_types[6] = {"--", "WiFi", "2G", "3G", "4G", "5G"};
   char network_type_str[32];
 
-  if (s->scene.networkType <= 5) {
-    snprintf(network_type_str, sizeof(network_type_str), "%s", network_types[s->scene.networkType]);
+  if ((int)s->scene.networkType <= 5) {
```
geohot, line 58: Similar comment to below on the casts if there's a nice way to remove this one. We shouldn't make assumptions like "5"

deanlee: Seriously， we should not use this enum value as index of array.

from the capnproto's document:

> Keep in mind when writing switch blocks that an enum read off the wire may have a numeric value that is not listed in its definition. This may be the case if the sender is using a newer version of the protocol, or if the message is corrupt or malicious.

deanlee:
use switch to handle enum values now.

selfdrive/ui/sidebar.cc
quillford: can we use a std::map?

deanlee: sure we can. but I think switch is clear enough and is the "standard" way to do this type of thing.
use std::map can redule a few lines, but it  alloc memory from heap,and do a "hash table  search"  everytime to get the value from key.I really don't like that.

selfdrive/ui/paint.cc
```diff
@@ -670,18 +663,33 @@ static void ui_draw_vision_footer(UIState *s) {
...
+void ui_draw_vision_alert(UIState *s, cereal::ControlsState::AlertSize va_size, int va_color,
                           const char* va_text1, const char* va_text2) {
   const UIScene *scene = &s->scene;
   const bool hasSidebar = !scene->uilayout_sidebarcollapsed;
   const bool mapEnabled = scene->uilayout_mapenabled;
   bool longAlert1 = strlen(va_text1) > 15;
 
   const uint8_t *color = alert_colors[va_color];
-  const int alr_s = alert_sizes[va_size];
+  int alr_s = 0;
```
geohot, line 674: I find this much less readable. Does

const int alert_sizes[] = {		
   [cereal::ControlsState::AlertSize::NONE] = 0,		
   [cereal::ControlsState::AlertSize::SMALL] = 241,

work?

selfdrive/ui/sidebar.cc
geohot: Actually from a perf perspective I'm not sure std::map would be worse. switch statement = branch mispredict, idk if map would.

selfdrive/ui/paint.cc
deanlee: no, it didn't work in this case:
error: integral constant expression must have integral or unscoped enumeration type, not 'capnp::schemas::ThermalStatus_98c15adf4549151f'

selfdrive/ui/sidebar.cc
deanlee: @quillford you'r right.std::map is much clearer than switch.

selfdrive/ui/sidebar.cc
```diff
@@ -9,28 +10,32 @@ static void ui_draw_sidebar_background(UIState *s) {
...
 static void ui_draw_sidebar_home_button(UIState *s) {
-  bool homeActive = s->active_app == cereal_UiLayoutState_App_home;
+  bool homeActive = s->active_app == cereal::UiLayoutState::App::HOME;
   const int home_btn_xr = !s->scene.uilayout_sidebarcollapsed ? home_btn_x : -(sbr_w);
 
   ui_draw_image(s->vg, home_btn_xr, home_btn_y, home_btn_w, home_btn_h, s->img_button_home, homeActive ? 1.0f : 0.65f);
 }
 
 static void ui_draw_sidebar_network_strength(UIState *s) {
+  static std::map<cereal::ThermalData::NetworkStrength, int> network_type_map = {
```
geohot, line 27: network_strength_map?

geohot:
Aside from that variable name, LGTM

selfdrive/ui/sidebar.cc
deanlee: fixed -_-b

deanlee:
I checked again and there should be no problem

selfdrive/ui/ui.cc
```diff
@@ -387,207 +363,150 @@ static void update_status(UIState *s, int status) {
...
-  cereal_read_Event(&eventd, eventp);
-
-  if (eventd.which == cereal_Event_controlsState && s->started) {
-    struct cereal_ControlsState datad;
-    cereal_read_ControlsState(&datad, eventd.controlsState);
+void handle_message(UIState *s,  Message* msg) {
+  auto amsg = kj::heapArray<capnp::word>((msg->getSize() / sizeof(capnp::word)) + 1);
+  memcpy(amsg.begin(), msg->getData(), msg->getSize());
+  capnp::FlatArrayMessageReader cmsg(amsg);
+  cereal::Event::Reader event = cmsg.getRoot<cereal::Event>();
```
deanlee, line 370: use FlatArrayMessageReader instead of custom MessageReader

deanlee:
I'm done with this

geohot:
Sweet, LGTM

PR #1458, opened 2020-05-03
pre-reqs for honda bosch longitudinal control

gregjhogan:
updates to support longitudinal control without radar on honda bosch
* switch `radarOffCan` checks to be either fingerprint check or `openpilotLongitudinalControl` check
* gas/brake control interface (implemented but not used)
* additional car state signals

this is just the first step and should not change how openpilot functions in any way

### DIFF
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -5,7 +5,7 @@ from selfdrive.controls.lib.drive_helpers import rate_limit
 from common.numpy_fast import clip, interp
 from selfdrive.car import create_gas_command
 from selfdrive.car.honda import hondacan
-from selfdrive.car.honda.values import CruiseButtons, CAR, VISUAL_HUD
+from selfdrive.car.honda.values import CruiseButtons, CAR, VISUAL_HUD, HONDA_BOSCH
 from opendbc.can.packer import CANPacker
 
 VisualAlert = car.CarControl.HUDControl.VisualAlert
@@ -135,8 +135,6 @@ class CarController():
     # **** process the car messages ****
 
     # steer torque is converted back to CAN reference (positive when steering right)
-    apply_gas = clip(actuators.gas, 0., 1.)
-    apply_brake = int(clip(self.brake_last * P.BRAKE_MAX, 0, P.BRAKE_MAX - 1))
     apply_steer = int(interp(-actuators.steer * P.STEER_MAX, P.STEER_LOOKUP_BP, P.STEER_LOOKUP_V))
 
     lkas_active = enabled and not CS.steer_not_allowed
@@ -147,14 +145,14 @@ class CarController():
     # Send steering command.
     idx = frame % 4
     can_sends.append(hondacan.create_steering_control(self.packer, apply_steer,
-      lkas_active, CS.CP.carFingerprint, idx, CS.CP.isPandaBlack))
+      lkas_active, CS.CP.carFingerprint, idx, CS.CP.isPandaBlack, CS.CP.openpilotLongitudinalControl))
 
     # Send dashboard UI commands.
     if (frame % 10) == 0:
       idx = (frame//10) % 4
-      can_sends.extend(hondacan.create_ui_commands(self.packer, pcm_speed, hud, CS.CP.carFingerprint, CS.is_metric, idx, CS.CP.isPandaBlack, CS.stock_hud))
+      can_sends.extend(hondacan.create_ui_commands(self.packer, pcm_speed, hud, CS.CP.carFingerprint, CS.is_metric, idx, CS.CP.isPandaBlack, CS.CP.openpilotLongitudinalControl, CS.stock_hud))
 
-    if CS.CP.radarOffCan:
+    if not CS.CP.openpilotLongitudinalControl:
       if (frame % 2) == 0:
         idx = frame // 2
         can_sends.append(hondacan.create_bosch_supplemental_1(self.packer, CS.CP.carFingerprint, idx, CS.CP.isPandaBlack))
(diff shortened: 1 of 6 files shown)


gregjhogan:
process replay failure is due to chrysler outputting `radarState.leadOne` and `radarState.leadTwo` (and I am guessing showing the lead vehicle indicator in the UI) even though OP does not perform longitudinal control.  we do get radar points from the chrysler radar, and I am guessing this is an unintended side effect?  my radard change caused this, is it working correctly now?

selfdrive/car/honda/hondacan.py
```diff
@@ -6,7 +6,9 @@ def get_pt_bus(car_fingerprint, has_relay):
   return 1 if car_fingerprint in HONDA_BOSCH and has_relay else 0
 
 
-def get_lkas_cmd_bus(car_fingerprint, has_relay):
+def get_lkas_cmd_bus(car_fingerprint, has_relay, openpilot_longitudinal_control=False):
```
pd0wm, line 9: Maybe rename argument to `radar_present` since that is was actually matters right? Since it forwards the commands?

selfdrive/car/honda/carcontroller.py
```diff
@@ -134,9 +134,13 @@ def update(self, enabled, CS, frame, actuators, \
 
     # **** process the car messages ****
 
+    if CS.CP.carFingerprint in HONDA_BOSCH:
```
pd0wm, line 137: No need to but behind if statement

selfdrive/car/honda/hondacan.py
pd0wm: Also add some comments about what all the buses are and why they are different for all the different checks in here. I always forget, so this might be a good place to put an explanation.

selfdrive/car/honda/carcontroller.py
gregjhogan: are you saying get rid of the `if`/`else` until we actually have more bosch longitudinal code? I was trying to make the changes in a way that merge conflicts are minimized (by inserting the conditional logic now)

selfdrive/car/honda/hondacan.py
gregjhogan: How about `radar_disabled`?
Should I change all functions I added the parameter `openpilot_longitudinal_control` to be `radar_disabled`?

selfdrive/car/honda/carcontroller.py
gregjhogan: https://github.com/gregjhogan/openpilot/blob/f76ad2eac0cc63264e255b68944b849511f94ef1/selfdrive/car/honda/carcontroller.py#L142-L156

gregjhogan:
updated based on feedback, chrysler replay still failing due to my controlsd change (see previous comment)

selfdrive/controls/radard.py
```diff
@@ -166,7 +166,7 @@ def update(self, frame, sm, rr, has_radar):
     dat.radarState.radarErrors = list(rr.errors)
     dat.radarState.controlsStateMonoTime = sm.logMonoTime['controlsState']
 
-    if has_radar:
+    if enable_lead:
```
adeebshihadeh, line 169: I think we want the radar data in the logs even if we're not going to show the lead

gregjhogan: since leads don't depend on having a radar at all, would you like `leadOne`/`leadTwo` always in the logs and use something new to determine if leads are displayed in UI?

gregjhogan: added `radarOffCan` back in as an `or` condition to preserve existing behavior (even if wrong)

gregjhogan:
updated based on feedback, all tests passing now

techasit14607:
แก้ปัญหาโดย เตชสิทธ์ ล้วนทร

selfdrive/controls/radard.py
techasit14607: แก้ปัญหาโดย เตชสิทธ์ ล้วนทร

selfdrive/car/honda/carcontroller.py
gregjhogan: I cleaned this up

gregjhogan:
@adeebshihadeh rebased and updated again if you want to review

selfdrive/car/honda/hondacan.py
```diff
@@ -76,6 +136,7 @@ def create_ui_commands(packer, pcm_speed, hud, car_fingerprint, is_metric, idx,
       "FCM_PROBLEM": stock_hud["FCM_PROBLEM"],
       "ICONS": stock_hud["ICONS"],
     }
+  if radar_disabled:
```
adeebshihadeh, line 139: maybe pass in CarParams instead? this name is confusing, and CarParams has the fingerprint

selfdrive/car/honda/hondacan.py
```diff
@@ -1,12 +1,26 @@
 from selfdrive.config import Conversions as CV
 from selfdrive.car.honda.values import HONDA_BOSCH
 
+# CAN bus layout with relay
+# 0 = ACC-CAN - radar side
+# 1 = F-CAN B - powertrain
+# 2 = ACC-CAN - camera side
+# 3 = F-CAN A - OBDII port
+
+# CAN bus layout with giraffe
```
adeebshihadeh, line 10: we can leave the giraffe out

selfdrive/car/honda/carcontroller.py
```diff
@@ -169,14 +167,19 @@ def update(self, enabled, CS, frame, actuators,
...
-        can_sends.append(hondacan.create_brake_command(self.packer, apply_brake, pump_on,
-          pcm_override, pcm_cancel_cmd, hud.fcw, idx, CS.CP.carFingerprint, CS.CP.isPandaBlack, CS.stock_brake))
-        self.apply_brake_last = apply_brake
-
-        if CS.CP.enableGasInterceptor:
-          # send exactly zero if apply_gas is zero. Interceptor will send the max between read value and apply_gas.
-          # This prevents unexpected pedal range rescaling
-          can_sends.append(create_gas_command(self.packer, apply_gas, idx))
+        if CS.CP.carFingerprint in HONDA_BOSCH:
+          raise NotImplementedError()
```
adeebshihadeh, line 171: this shouldn't ever be hit, but pass with a TODO?

selfdrive/car/honda/hondacan.py
gregjhogan: What is confusing about `radar_disabled`? It literally means the radar has been disabled and is not outputting any messages.

gregjhogan: also, the fingerprint by itself doesn't provide enough info (we don't want to output `ACC_HUD` if the radar is also outputting it)

gregjhogan: oh I see now, this isn't whether the radar is disabled, it is whether openpilot is handling longitudinal control, I will fix the name

selfdrive/car/honda/carcontroller.py
gregjhogan: done

selfdrive/car/honda/hondacan.py
gregjhogan: fixed

PR #1459, opened 2020-05-03
honda bosch disable radar

gregjhogan:
the following steps disable the radar:
* enter extended diagnostic session
* send communication control message for disable tx and rx with suppress response bit
* send tester present with suppress response bit periodically to keep radar silent
  (radar wakes back up if it doesn't get a tester present message for a few seconds)

disabling an ecu in this manner is not unique to honda

### DIFF
--- a/selfdrive/car/disable_ecu.py
+++ b/selfdrive/car/disable_ecu.py
@@ -1,0 +1,38 @@
+#!/usr/bin/env python3
+import traceback
+
+import cereal.messaging as messaging
+from selfdrive.car.isotp_parallel_query import IsoTpParallelQuery
+from selfdrive.swaglog import cloudlog
+
+EXT_DIAG_REQUEST = b'\x10\x03'
+EXT_DIAG_RESPONSE = b'\x50\x03'
+COM_CONT_REQUEST = b'\x28\x83\x03'
+COM_CONT_RESPONSE = b''
+
+def disable_ecu(ecu_addr, logcan, sendcan, bus, timeout=0.1, retry=5, debug=False):
+  print(f"ecu disable {hex(ecu_addr)} ...")
+  for i in range(retry):
+    try:
+      # enter extended diagnostic session
+      query = IsoTpParallelQuery(sendcan, logcan, bus, [ecu_addr], [EXT_DIAG_REQUEST], [EXT_DIAG_RESPONSE], debug=debug)
+      for addr, dat in query.get_data(timeout).items():
+        print(f"ecu communication control disable tx/rx ...")
+        # communication control disable tx and rx
+        query = IsoTpParallelQuery(sendcan, logcan, bus, [ecu_addr], [COM_CONT_REQUEST], [COM_CONT_RESPONSE], debug=debug)
+        query.get_data(0)
+        return True
+      print(f"ecu disable retry ({i+1}) ...")
+    except Exception:
+      cloudlog.warning(f"ecu disable exception: {traceback.format_exc()}")
+
+  return False
+
+
+if __name__ == "__main__":
+  import time
+  sendcan = messaging.pub_sock('sendcan')
+  logcan = messaging.sub_sock('can')
+  time.sleep(1)
+  disabled = disable_ecu(0x18DAB0F1, logcan, sendcan, 1, debug=False)
+  print(f"disabled: {disabled}")
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -5,7 +5,7 @@ from selfdrive.controls.lib.drive_helpers import rate_limit
 from common.numpy_fast import clip, interp
 from selfdrive.car import create_gas_command
 from selfdrive.car.honda import hondacan
-from selfdrive.car.honda.values import CruiseButtons, CAR, VISUAL_HUD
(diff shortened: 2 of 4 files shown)


gregjhogan:
looking for feedback ... I split this up between controlsd and carcontroller because:
* the initial disabling is similar to reading the VIN (request/response)
* the periodic message to keep it disabled is much simpler (fire and forget) and I would think we want as much as possible in the car specific interface

this doesn't feel great, is there a better approach? once the approach is agreed upon we can clean up things like the hard-coded CAN addr, etc...

gregjhogan:
we probably also need something that detects if the radar wakes back up which would disable openpilot

csouers:
> we probably also need something that detects if the radar wakes back up which would disable openpilot
one way to see if radar is back up is to check the counter on the radar's status frame (0x400 on bus 0 of black panda/uno). if it is different than the last time we saw it, then it must be alive. would need to setup a new can parser for that though. maybe camera outputs a connection state on the private can that we can check? is the radar's private can output totally silent with the disable?

csouers:
> we probably also need something that detects if the radar wakes back up which would disable openpilot

one way to see if radar is back up is to check the counter on the radar's status frame (0x400 on bus 0 of black panda/uno). if it is different than the last time we saw it, then it must be alive. would need to setup a new can parser for that though. maybe camera outputs a connection state on the private can that we can check? is the radar's private can output totally silent with the disable?

csouers:
bit 21 on 0x374 (f-can) goes high exclusively with the radar disconnected. I had an acc fault even after the reconnection and the bit went back low again after the reconnect. just a status bit for the control frames? is this bit low with op long?

gregjhogan:
we should be able to pick any radar message addr on the ACC bus that comes from the radar and ensure it is not present (ignoring the data)

I wish this all fit better into radard/radar_interface, but that is a lot more complicated if you want to prevent openpilot from sending any messages that the radar sends until after the radar is shut off

selfdrive/car/disable_ecu.py
```diff
@@ -0,0 +1,38 @@
...
+from selfdrive.car.isotp_parallel_query import IsoTpParallelQuery
+from selfdrive.swaglog import cloudlog
+
+EXT_DIAG_REQUEST = b'\x10\x03'
+EXT_DIAG_RESPONSE = b'\x50\x03'
+COM_CONT_REQUEST = b'\x28\x83\x03'
+COM_CONT_RESPONSE = b''
+
+def disable_ecu(ecu_addr, logcan, sendcan, bus, timeout=0.1, retry=5, debug=False):
+  print(f"ecu disable {hex(ecu_addr)} ...")
```
pd0wm, line 14: change prints to cloudlog or remove them if only needed for testing

selfdrive/car/honda/carcontroller.py
```diff
@@ -144,6 +144,12 @@ def update(self, enabled, CS, frame, actuators, \
     # Send CAN commands.
     can_sends = []
 
+    if CS.CP.carFingerprint in HONDA_BOSCH and CS.CP.openpilotLongitudinalControl:
+      # TODO: radar disable hacked together to see if it works
+      if (frame % 10) == 0:
+        # tester present - w/ no response (keeps radar disabled)
+        can_sends.append([0x18DAB0F1, 0, b"\x02\x3E\x80\x00\x00\x00\x00\x00", 1 if CS.CP.isPandaBlack else 0])
```
pd0wm, line 151: Put radar debug addr in values.py

selfdrive/car/disable_ecu.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
+import traceback
+
+import cereal.messaging as messaging
+from selfdrive.car.isotp_parallel_query import IsoTpParallelQuery
+from selfdrive.swaglog import cloudlog
+
+EXT_DIAG_REQUEST = b'\x10\x03'
```
pd0wm, line 8: Use constants from panda.uds

selfdrive/car/honda/carcontroller.py
pd0wm: Use constants from panda.uds

selfdrive/controls/controlsd.py
```diff
@@ -492,6 +493,9 @@ def controlsd_thread(sm=None, pm=None, can_sock=None):
   put_nonblocking("CarParamsCache", cp_bytes)
   put_nonblocking("LongitudinalControl", "1" if CP.openpilotLongitudinalControl else "0")
 
+  if CP.openpilotLongitudinalControl and CP.safetyModel in [car.CarParams.SafetyModel.hondaBoschGiraffe, car.CarParams.SafetyModel.hondaBoschHarness]:
```
pd0wm, line 496: why not check fingerprint in HONDA_BOSCH? Instead of safety mode?

pd0wm (changes requested):

martinl:
Would it make sense to make the ISO-TP queries more generic, for example create single and recurring query functions with car specific commands and query rate configurable from values.py 

I'm asking because we plan to query one OBDII PID value for some Subaru models after fingerprinting (related to country/market specific speed unit values) and I could not find a clean way to implement this query within existing framework (https://github.com/martinl/openpilot/blob/subaru-PR-test/selfdrive/car/obd_query.py)

adeebshihadeh:
Merged in the ECU disable script in 98fedff6b1f286c28a5beb12fd40b392b1d4ce65

PR #1460, opened 2020-05-03
simplify code: use vector to store SubSockets

deanlee:


geohot:
Ideally, we want a copy of the SubMaster class in C++ for things like this.

deanlee:
cool. I'll try that

PR #1461, opened 2020-05-03
fix wrong order of cleanup calls

deanlee:
delete Context before publisher can potentially hang forever

> Killing old publisher: androidLog
Killing old publisher: androidLog
Killing old publisher: androidLog
Killing old publisher: androidLog
Killing old publisher: androidLog
Killing old publisher: androidLog


geohot:
Hmm, we should fix this in cereal.

PR #1462, opened 2020-05-03
Discussion only: A high performance MessageBuilder.

deanlee:
main features:
1. return first segment from stack in allocateSegment().
only alloc  from heap if message is too big.the default stack  size is 256 words(128 is ok too), fits almost all packages,except proclogd(>15000 words).
2. "zero overhead" when building message to array.
if there is only one segment and it's already in the stack（Most of the time）,skip alloc & copy & build message to HeapArray. all we need to do is just to set a one word size head at the begin of the stack.

### DIFF
--- a/selfdrive/common/messagehelp.h
+++ b/selfdrive/common/messagehelp.h
@@ -1,6 +1,6 @@
 #pragma once
 #include <capnp/serialize.h>
-
+#include <kj/vector.h>
 #include "cereal/gen/cpp/log.capnp.h"
 #include "messaging.hpp"
 class MessageReader {
--- a/selfdrive/common/messagehelp.h
+++ b/selfdrive/common/messagehelp.h
@@ -87,10 +87,10 @@ class MessageBuilder : public capnp::MessageBuilder {
   }
 
  protected:
+  kj::Array<capnp::word> flatArray;
+  kj::Vector<capnp::word *> moreSegments;
   bool returnedFirstSegment;
   uint nextSize;
   // the first words of stackSement is used internally to set the head table.
-  capnp::word stackSegment[STACK_SEGEMENT_BUF_SIZE + 1];
-  kj::Array<capnp::word> flatArray;
-  kj::Vector<capnp::word *> moreSegments;
+  alignas(8) capnp::word stackSegment[STACK_SEGEMENT_BUF_SIZE + 1];
 };
--- a/selfdrive/common/messagehelp.h
+++ b/selfdrive/common/messagehelp.h
@@ -70,7 +70,7 @@ class MessageBuilder : public capnp::MessageBuilder {
   kj::ArrayPtr<kj::byte> toBytes() {
     auto segments = getSegmentsForOutput();
     if (segments.size() == 1 && segments[0].begin() == stackSegment + 1) {
-      cosnt size_t segment_size = segments[0].size();
+      const size_t segment_size = segments[0].size();
       uint32_t *table = (uint32_t *)stackSegment;
       table[0] = 0;
       table[1] = segment_size;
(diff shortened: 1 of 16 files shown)


selfdrive/loggerd/loggerd.cc
```diff
@@ -558,7 +558,7 @@ static void bootlog() {
     std::string lastPmsg = util::read_file("/sys/fs/pstore/pmsg-ramoops-0");
     boot.setLastPmsg(capnp::Data::Reader((const kj::byte*)lastPmsg.data(), lastPmsg.size()));
 
-    auto words = capnp::messageToFlatArray(msg);
+    auto words = msg.toFlatArrayPtr();
```
geohot, line 561: I don't think we ever need the "words" ptr, maybe return after asBytes?

selfdrive/common/messagehelp.h
```diff
@@ -40,3 +40,56 @@ class MessageReader {
...
+    }
+    uint size = kj::max(minimumSize, nextSize);
+    capnp::word *result = (capnp::word *)calloc(size, sizeof(capnp::word));
+    moreSegments.add(result);
+    nextSize += size;
+    return kj::ArrayPtr<capnp::word>(result, size);
+  }
+
+  kj::ArrayPtr<capnp::word> toFlatArrayPtr() {
+    kj::ArrayPtr<const kj::ArrayPtr<const capnp::word>> segments = getSegmentsForOutput();
```
geohot, line 71: Hmm, does "word>>" compile? I think it used to have to be "word> >"

deanlee: >> is ok. an "auto"  maybe more clearly.

selfdrive/loggerd/loggerd.cc
deanlee: just want to keep the parten with capnproto at that moment:return words arrary,let user choose howto convert it.

it looks unnecessary now.

deanlee:
Done. all messages except proclog will be built from stack with this custom class. zero malloc & copy.

deanlee:
The result of my test is, only two messages cannot be buit from stack, one is proclogd(>1500words),anther is 500+words every 2 or 3 sencods. so I think the default stack size 256 words(2048 bytes) is big enough.

selfdrive/common/messagehelp.h
```diff
@@ -40,3 +40,57 @@ class MessageReader {
   capnp::FlatArrayMessageReader *msg_reader;
   cereal::Event::Reader event;
 };
+
+#define STACK_SEGEMENT_BUF_SIZE 256
+class MessageBuilder : public capnp::MessageBuilder {
+ public:
+  MessageBuilder(uint firstMallocSize = 2048) : returnedFirstSegment(false), nextSize(firstMallocSize), stackSegment{} {};
```
geohot, line 47: What is returnedFirstSegment? Is it ever true? Maybe add a comment

deanlee: this variable is used to make sure stack will only be returned once at first segment.

geohot: and why does there have to be a firstMallocSize?

I'm not sure these perf improvements are worth the hit to readability. They can always be made later in one place if a benchmark shows they are good.

deanlee: firstMallocSize is used to avoid too many alloc calls & segments for large message such as proclogd.for example ,MallocMessageBuilder's  default behaviors is alloc 1024 words everytime .so ,it need's alloc many segments to perpare the prologd's message.

deanlee: yes you'r right.I'm not sure isn't it worth too.so I tuned this PR to draft,close it later.

deanlee: not only 1 malloc.  save 2 malloc and one copy for a single message.

geohot: I like the lines this PR removes from the programs. I still think it's worth it, just with the simplest implementation in the class.

PR #1466, opened 2020-05-04
Alerts + Events refactor

adeebshihadeh:
changes
* better defined relationship between alerts and events
* remove duplicated code in alert definitions
* remove unused alerts
  * geofence
  * dataNeeded ("Calibration Needs Data. Upload Drive, Try Again")

process replay differences
* `controlsState.alertType`
  * alerts no longer have their own name. event name from car.capnp is now used
* `carState.events`, `constrolsState.alertSize`, `constrolsState.alertText1`, `controlsState.alertText2`
  * there's a bug in the current code that shows the startup alert for one too few frames. fixed by adding the startup alert after updating SubMaster

Will clean up things on the controlsd side in the controlsd refactor.


geohot:
Should we have "events.append(create_event(.." everywhere or what about making events a class? like "events.add_new_event(..."

selfdrive/controls/lib/events.py
```diff
@@ -0,0 +1,524 @@
...
+                    AlertStatus.normal, AlertSize.none,
+                    Priority.MID, VisualAlert.none, AudibleAlert.chimeEngage, .2, 0., 0.)
+
+disable_alert = Alert(
+                    "",
+                    "",
+                    AlertStatus.normal, AlertSize.none,
+                    Priority.MID, VisualAlert.none, AudibleAlert.chimeDisengage, .2, 0., 0.)
+
+EVENTS = {
```
deanlee, line 98: Would it be better to put EVENTS to a seperate file? it may be very helpful for multilanguage support

quillford: I'd prefer it'd stay here for now. When multilanguage gets added, it'll be easy to pull out

quillford:
@geohot What do you think? There's more to clean up in controlsd, but this PR is already pretty big. I think it can go in the controlsd refactor

geohot:
@pd0wm would be the one to review this code really, I'm not that familiar with it. But I'm reading now.

selfdrive/controls/lib/driver_monitor.py
```diff
@@ -1,10 +1,13 @@
 from common.numpy_fast import interp
 from math import atan2, sqrt
 from common.realtime import DT_DMON
-from selfdrive.controls.lib.drive_helpers import create_event, EventTypes as ET
 from common.filter_simple import FirstOrderFilter
 from common.stat_live import RunningStatFilter
 
+from cereal import car
+
+EventName = car.CarEvent.EventName
```
geohot, line 9: You use EN in some places and EventName here, any reason to not always use EN?

selfdrive/car/honda/interface.py
```diff
@@ -5,7 +5,7 @@
 from common.realtime import DT_CTRL
 from selfdrive.swaglog import cloudlog
 from selfdrive.config import Conversions as CV
-from selfdrive.controls.lib.drive_helpers import create_event, EventTypes as ET, get_events
+from selfdrive.controls.lib.events import ET
```
geohot, line 8: If we are importing this I'm thinking there's a way this can be refactored. Why does only Honda import this?

selfdrive/car/nissan/interface.py
```diff
@@ -77,9 +76,9 @@ def update(self, c, can_strings):
     events = self.create_common_events(ret)
 
     if self.CS.lkas_enabled:
-      events.append(create_event('invalidLkasSetting', [ET.PERMANENT]))
+      events.add(car.CarEvent.EventName.invalidLkasSetting)
 
-    ret.events = events
+    events.to_msg(ret)
```
geohot, line 81: I don't like this pattern everywhere, you really shouldn't modify what you pass in.

ret.events = events.to_msg() is a lot more readable

geohot:
I like the overall refactor with the event types being separate from the events themselves. 

Why is there still an ALERTS? Why aren't these EVENTS?

selfdrive/car/nissan/interface.py
quillford: I thought it'd be faster to avoid the python list and keep everything as a capnp list, especially since this is a high frequency call. I profiled it though, and turns out it doesn't make much of a difference

quillford:
I kept ALERTS around as we show some alerts that don't meet the definition of an event, things that cause controls state machine transition. However, we have a bunch of events that only have the warning type, which means they don't satisfy that definition. We can either open up the definition of an event or move those warning only events to alerts. I'm leaning towards moving those alerts to warning only events, but I'll think about it some more.

selfdrive/car/honda/interface.py
adeebshihadeh: The code that uses this is pretty old and can probably just be deleted, but I don't want to make this PR bigger and harder to test. Will fix after merging

selfdrive/controls/lib/events.py
```diff
@@ -0,0 +1,694 @@
...
+    if static:
+      self.static_events.append(event_name)
+    self.events.append(event_name)
+
+  def clear(self):
+    self.events = self.static_events.copy()
+
+  def any(self, event_type):
+    for e in self.events:
+      if event_type in EVENTS.get(e, {}).keys():
```
pd0wm, line 58: Should we build a cache for this on startup?

pd0wm (approved):

PR #1468, opened 2020-05-05
add class SocketMaster

deanlee:
Manage Context,Poller, PubSocket and Subsocket  in one class. delete them in the right order.

### DIFF
--- a/selfdrive/common/socketmaster.h
+++ b/selfdrive/common/socketmaster.h
@@ -45,7 +45,7 @@ class SocketMaster {
         break;
       }
       for (auto sock : polls) {
-        Message *msg = sock->receive();
+        Message *msg = sock->receive(true);
         if (msg) {
           messages.push_back(msg);
         }
--- a/selfdrive/common/socketmaster.h
+++ b/selfdrive/common/socketmaster.h
@@ -1,3 +1,5 @@
+#pragma once
+
 #include <assert.h>
 #include <map>
 #include "messaging.hpp"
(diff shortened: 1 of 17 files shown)


selfdrive/common/socketmaster.h
```diff
@@ -0,0 +1,65 @@
...
+  std::vector<Message *> pollMessages(int timeout) {
+    assert(poller != NULL);
+    std::vector<Message *> messages;
+    while (true) {
+      auto polls = poller->poll(timeout);
+      if (polls.size() == 0) {
+        break;
+      }
+      for (auto sock : polls) {
+        Message *msg = sock->receive();
```
quillford, line 48: I think this should be a non blocking, we're already waiting on the poller timeout

quillford:
I think we should call this SubMaster like the python version, and this should go with the rest of messaging in cereal/messaging.

deanlee:
I thought about that before. but it looks like that there are no much thing to do with PubSocket,we just need to  guaranteed it will be deleted. And it will introduce the complexity of managing context if we 
split into two classes.

pd0wm:
I also prefer a separate SubMaster/PubMaster class. Could be helpful when doing dependency injection for testing.

Can you submit a PR for cereal? I'd also like the API to SubMaster to be comparable to the python version (with the alive and valid checks), we should really have the same checks in the c code.

deanlee:
ok i'll try

PR #1469, opened 2020-05-05
Minimal supercombo implementation with frame reader

littlemountainman:
* You can read a hevc from your latest drive and see how the supercombo model would have performed. 

* Loads the model files, predicts on the yuv converted video, parses the output and displays it to you.


geohot:
There's a ton of duplicate code in here. This should be using common/transformations, not going to merge a +1,328 line PR.

geohot:
Going to close for now, resubmit with a PR that doesn't recommit the libraries and we can review.

PR #1472, opened 2020-05-06
camerad: move c-capnp to c++

deanlee:
1. rename camera_qcom.c to camera_qcom.cc
2. fix c++ build errors
3. convert c-capnp to c++


deanlee:
Done.I think we may be able to delete c-capnp completely after this.

selfdrive/camerad/cameras/camera_qcom.cc
```diff
@@ -177,12 +177,12 @@ static int imx298_apply_exposure(CameraState *s, int gain, int integ_lines, int
     {0x3002,0x0,0}, // long autoexposure off
 
     // FRM_LENGTH
-    {0x340, frame_length >> 8, 0}, {0x341, frame_length & 0xff, 0},
+    {0x340, (unsigned short)(frame_length >> 8), 0}, {0x341, (unsigned short)(frame_length & 0xff), 0},
```
geohot, line 180: Include stdint and do uint16_t instead of unsigned short?

selfdrive/camerad/cameras/camera_qcom.cc
```diff
@@ -563,37 +565,37 @@ static void sensors_init(DualCameraState *s) {
       .power_setting_array = {
         .power_setting_a = {
           {
-            .seq_type = 1,
+            .seq_type = SENSOR_GPIO,
```
geohot, line 568: Cool! Much more readable.

selfdrive/camerad/cameras/camera_qcom.cc
```diff
@@ -1268,7 +1282,7 @@ static void camera_open(CameraState *s, bool rear) {
           .reg_addr = 240,
           .hw_shift = 0,
           .data_type = 10,
-          .addr_type = 4,
+          .addr_type = MSM_CAMERA_I2C_ADDR_TYPE_MAX,
```
geohot, line 1285: I don't think this one is right, the MAX one shouldn't be used.

deanlee: MSM_CAMERA_I2C_ADDR_TYPE_MAX is equal 4. same with previous.

> enum msm_camera_i2c_reg_addr_type {
	MSM_CAMERA_I2C_BYTE_ADDR = 1,
	MSM_CAMERA_I2C_WORD_ADDR,
	MSM_CAMERA_I2C_3B_ADDR,
	MSM_CAMERA_I2C_ADDR_TYPE_MAX,
};

selfdrive/camerad/cameras/camera_qcom.cc
deanlee: ok

selfdrive/camerad/cameras/camera_qcom.cc
geohot: it is 4, but I'm saying that's the wrong enum. there's probably another 4 somewhere

deanlee: i see.sorry for my bad reading.

deanlee: can't figure out where the correct enum is. so I switched them back to 4.

PR #1476, opened 2020-05-08
use PubMaster & SubMaster

deanlee:
use PubMaster & SubMaster added in cereal https://github.com/commaai/cereal/pull/42

### DIFF
--- a/selfdrive/boardd/SConscript
+++ b/selfdrive/boardd/SConscript
@@ -1,6 +1,6 @@
-Import('env', 'common', 'messaging')
+Import('env', 'common', 'cereal', 'messaging')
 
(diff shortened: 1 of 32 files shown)


deanlee:
I have tested it and it seems to work fine.

deanlee:
since the copy between event:Reader is very cheap (just set the internal pointer to others), SubMaster::poll return std::vector<event::Reader>  to make it more easy to use.

deanlee:
updated to new Sub&Pub AP I:https://github.com/commaai/cereal/pull/42#issuecomment-628481128

pd0wm:
Moved to a new branch so I can get all the CI to run.

PR #1477, opened 2020-05-08
Shebang Fix

iHelp101:
Per George's "I'd merge a PR replacing the 4 python3.7 with python3" on Discord. George updated Python to 3.8.2 for webcam tools, which broke manager from launching nicely. Instead of chasing the exact version, it seems more reasonable to fall back to 3.


adeebshihadeh:
Can you fix the rest of the python3.7 references?

iHelp101:
Sure, let me hunt them down.

PR #1480, opened 2020-05-09
LGTM fixes

adeebshihadeh:


geohot:
I wouldn't just comment these things out, most of them can be deleted (like the CANPacker import)

adeebshihadeh:
You're right, even the commented code that needed the import is dead. Fixed in master.

PR #1499, opened 2020-05-13
CAR.GENESIS_G80 has incorrect parameters

xps-genesis:
Hyundai Genesis 2015-2016 is similar to Genesis G80. The Genesis G90 is the larger vehicle. These are grouped wrongly.


pd0wm:
The steer ratio is the most important parameter for controls and is significantly different. Do you know which on is correct for the G80? You can also create a separate entry for the G80.

xps-genesis:
unfortunately I do not have that info and i am not able to find anything on the internet. all 3 vehicles have Variable Gear Ratio racks.
Hyundai Genesis and Genesis G80 are identical chassis with only brand change + some extra features.
Genesis G90 is a larger vehicle and may have a different ratio. 

The proposed change kind of puts the G80 close to where it should be.

I also doubt the 16.5 ratio on the Hyundai genesis to be accurate.

PR #1507, opened 2020-05-14
Add code to steer using TSS2 LTA message

pd0wm:
This approach was rejected a while ago, but there was a mistake in the signal definition. Maybe it will work now.


pd0wm:
Experiments showed that it now works as expected, but still worse than torque mode. Putting the code as comments in case we want to experiment more.

PR #1511, opened 2020-05-15
Use Conversions in calibrationd

qiubit:
Hi, newbie here, trying to make some small commit to get started.

I noticed that `calibrationd` is the only process that is not using `Conversions` as defined in `selfdrive.config`, so this small commit provides consistency.


geohot:
Sweet! That CI failure isn't your fault, merging.

Thank you for the contribution, and welcome to being an openpilot developer

PR #1512, opened 2020-05-15
thneed saves 45% of a core

geohot:
I now have 45 CPU offset credits to spend!

The code in debug/ isn't used, but I like having it around for future investigation. If we don't use it in a while we can delete it.

thneed itself is about 500 lines, $25,000 for the CPU savings, though it's mostly debug prints and relatively abstracted, so I'd charge more like 10k, I think worth it for the CPU savings. (happy to pay 80k for 0 CPU usage!)

There is also a dumb copy being made for the frame on input. model runs in 22ms, 25 ms in modeld, and 30ms total with ui running. For bigmodel, we are going to have to do better.


selfdrive/modeld/thneed/thneed
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+export LD_LIBRARY_PATH="/data/pythonpath/phonelibs/snpe/aarch64/:$HOME/openpilot/phonelibs/snpe/larch64:$HOME/openpilot/phonelibs/snpe/x86_64-linux-clang:$LD_LIBRARY_PATH"
```
pd0wm, line 2: openpilot is now always in `/data/openpilot`

geohot: oh this file isn't used, let me move to debug and fix

geohot:
This has bugs still, please wait before using.

PR #1516, opened 2020-05-15
White panda is deprecated

pd0wm:
We want to start using the localizer soon!

### DIFF
--- a/selfdrive/controls/lib/events.py
+++ b/selfdrive/controls/lib/events.py
@@ -244,6 +244,7 @@ EVENTS = {
       "Upgrade to comma two or black panda",
       AlertStatus.normal, AlertSize.mid,
       Priority.LOWER, VisualAlert.none, AudibleAlert.none, 0., 0., .2),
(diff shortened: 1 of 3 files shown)


selfdrive/controls/lib/events.py
```diff
@@ -198,6 +198,14 @@ def calibration_incomplete_alert(CP, sm, metric):
       Priority.LOWER, VisualAlert.none, AudibleAlert.none, 0., 0., 15.),
   },
 
+  EventName.startupWhitePanda: {
+    ET.PERMANENT: Alert(
+      "WARNING: White panda is deprecated",
```
adeebshihadeh, line 203: The other alert capitalizes panda

pd0wm: I see we broke the rule to not capitalize product names in the events. I'll change it back to uppercase for consistency

pd0wm: The other startup alerts are not capitalized, so I'll leave this as is to have consistency there.

PR #1521, opened 2020-05-16
use constant for speedTooHigh event

eFiniLan:


adeebshihadeh:
The 90mph limit is actually already defined in `selfdrive/controls/drive_helpers.py`. Let's use that instead of duplicating the constant.

eFiniLan:
updated :)

PR #1522, opened 2020-05-16
Bugfix: Reset saturated_count if angle no longer saturated

avolmensky:
saturated_count wasn't being reset when steering was no longer saturated.


adeebshihadeh:
Nice catch, looks like this was introduced in the refactor.

PR #1529, opened 2020-05-17
fix executing python code info

xx979xx:


geohot:
This shouldn't be needed, why doesn't the dot slash work?

PR #1530, opened 2020-05-17
Pip: Move from tensorflow-gpu to tensorflow package

EwoutH:
From tensorflow 2.1 onward the regular tensorflow package contains GPU support and replaces the tensorflow-gpu package, which will be deprecated at some point.


EwoutH:
Some issue with the podfile.lock, will resolve it in a bit

EwoutH:
Pipfile.lock is updated and CI passes. For some reason, when updating the Pipfile.lock, the Python markers weren't generated, while I used the `--requirements` parameter with `pipenv lock`. The Pipfile.lock was generated in [this run](https://github.com/EwoutH/openpilot/actions/runs/109729585).

A few minor dependencies also got updated in the process.

PR #1542, opened 2020-05-19
Add workflow to update Pipfile.lock weekly

EwoutH:
This is quite an experiment, so let me know what you think!

When working on #1530 I noticed that I had to manually update the Pipfile.lock, which gave some issues with platforms and versions at first. So when I got it working, I decided to try to automate it for consistency and efficiency.

In this PR I added a workflow with GitHub actions that opens a PR to update the Pipfile.lock on one of two conditions:
 - Weekly (currently Monday 08:00 UTC)
 - When the Pipfile is modified

The weekly PR will be opened directly against the master branch, while when the Pipfile is modified the PR would open against the branch in which the Pipfile is modified. This way dependencies won't change in an unexpected manner, but will be updated regularly.

The PR will be opened in the `update-pipfile` branch. If it doesn't exist it will be created, if last weeks PR is still open it will be updated and when it's merged is will be deleted.

Currently only the major versions of dependencies that were previously fully locked are now locked. This PR also move the tensorflow-gpu to tensorflow package (which are at this point identical) and thus can replace #1530.

An example run of the workflow can be seen [here](https://github.com/EwoutH/openpilot/runs/690823425) and an example of the PR here: https://github.com/EwoutH/openpilot/pull/3

If you like the general idea, please also comment on specifics (used versions, the schedule, naming, etc.)


pd0wm:
Cool! Let's try it and see if we like it.

EwoutH:
@pd0wm and @adeebshihadeh I saw that with #2178 this workflow was removed. Could you give me some feedback about what you liked and didn't like about the CI configuration that updated the pip dependencies?

PR #1552, opened 2020-05-22
Fix hyundai EV gas signal

adeebshihadeh:


selfdrive/car/hyundai/carstate.py
```diff
@@ -54,9 +54,12 @@ def update(self, cp, cp_cam):
     # TODO: Check this
     ret.brakeLights = bool(cp.vl["TCS13"]['BrakeLight'] or ret.brakePressed)
 
-    #TODO: find pedal signal for EV/HYBRID Cars
-    ret.gas = cp.vl["EMS12"]['PV_AV_CAN'] / 100
-    ret.gasPressed = bool(cp.vl["EMS16"]["CF_Ems_AclAct"])
+    if self.CP.carFingerprint in EV_HYBRID:
+      ret.gas = cp.vl["E_EMS11"]['Accel_Pedal_Pos'] / 256.
+      ret.gasPressed = ret.gas > 0
```
xx979xx, line 59: Hi, I already implement this and tested it [here ](https://github.com/xx979xx/openpilot/commit/bf3cb45791a098e4335d3de54a2e23b28b0e05a3)
I think the threshold should be about 5. to prevent false positive
also, we can add new feature instead of new class

TK211X:
There was some use in the Discord that mentioned this code worked for him BTW. We have some other users waiting on this.

adeebshihadeh:
superseded by #1694

PR #1554, opened 2020-05-22
resizable tools/ui window

xx979xx:
not perfect and I did not test its impact on performance
have some bugs and need some works
still better than fixed window


pd0wm:
I'm not sure this is better than a fixed window since the ui doesn't actually scale. It also became completely unusable on i3, for some reason it either looks compressed or draws things twice.

PR #1562, opened 2020-05-25
GitHub Actions: Fix Update Pipfile.lock workflow

EwoutH:
Small patch after #1542.

Weekly [run ran](https://github.com/commaai/openpilot/actions/runs/114523851), but didn't update the Pipfile.lock and therefore didn't open an PR.

Since we don't use a requirements.txt the `--requirements` can be dropped. It also modifies the workflow to run when changing the _Update Pipfile.lock_ workflow.

The PR from github-actions now merged into this branch will be opened next week against the master branch.

### DIFF
--- a/.github/workflows/update-pipfile.yml
+++ b/.github/workflows/update-pipfile.yml
@@ -24,3 +24,4 @@ jobs:
       with:
         title: "Update Pipfile.lock (dependencies)"
         branch: update-pipfile
+        commit-message: "[Bot] Update Pipfile.lock dependencies"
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Can you revert the changes to Pipfile.lock?

EwoutH:
Done!

adeebshihadeh:
Thanks, I'll change it to 8:00 pacific so we can test it out today.

EwoutH:
Now an PR is also opened when the `update-pipfile.yml` is changed, see #1563

PR #1574, opened 2020-05-27
cython capnp wrapper

pd0wm:
cereal part is done. Let's do some benchmarks to see the effect in openpilot.


pd0wm:
Only 6% of a core :( Not worth the complexity.

PR #1575, opened 2020-05-27
Added pre-commit hook for reorder-python-imports.

dapperfu:
Tool for automatically reordering python imports. Similar to isort but uses static analysis more.

From https://github.com/asottile/reorder_python_imports

> ## why this style?
> The style chosen by reorder-python-imports has a single aim: reduce merge conflicts.
> By having a single import per line, multiple contributors can add / remove imports from a single module without resulting in a conflict.


pd0wm:
Sorry, I'm not a fan. This PR adds a huge number of lines, and would take a significant time to review. Don't think the style is that useful, and a commit hook for this sounds overkill.

The `commit-hook` project sounds interesting though. I would definitely merge something that runs flake8 and pylint on the modified files!

dapperfu:
It passes all tests. It improves code quality and makes future merges easier.

adeebshihadeh:
I’m not a fan of the style either. It really doesn’t make merging easier. I don’t remember the last time I had a merge conflict because of import lines.

PR #1577, opened 2020-05-27
"The commit-hook project sounds interesting though. I would definitel…

dapperfu:
…y merge something that runs flake8 and pylint on the modified files!"

-  pd0wm, https://github.com/commaai/openpilot/pull/1575#issuecomment-634974344

image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,16 @@
+repos:
+-   repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: master
+    hooks:
+    -   id: check-ast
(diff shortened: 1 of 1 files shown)


pd0wm:
pylint needed some more config options and work to pass. See https://github.com/commaai/openpilot/pull/1580

Specifically it couldn't be run in an isolated virtualenv because it uses dynamic analysis and checks all the imports. Once I get that to replace our current pylint CI step I'll look at flake8. 

Currently we have a whole list of violations we don't check for (see .pylintrc), we should make small PRs where we gradually remove those and fix them everywhere in the codebase.

PR #1585, opened 2020-05-28
ui.cc: remove variables from UIScene, use event::Reader to get data 

deanlee:
SubMaster already has a event::reader for each incoming message, we can get data from it directly without copy in handle_message.

this PR can save lines of code and avoid  unnecessary reading of data that may not used later.for example, most data of thermal are used only when the sidebar is visible.


adeebshihadeh:
Your new struct is just a subset of the existing LeadData struct from capnp. Can we just use the one from capnp?

selfdrive/ui/paint.cc
```diff
@@ -154,7 +154,7 @@ static void update_track_data(UIState *s, bool is_mpc, track_vertices_data *pvd)
 
   bool started = false;
   float off = is_mpc?0.3:0.5;
-  float lead_d = scene->lead_d_rel*2.;
+  float lead_d = (*s->sm)["radarState"].getRadarState().getLeadOne().getDRel()*2.;
```
deanlee, line 157: average read time:30 ns.it's acceptable.

deanlee:
save reader to UIScene for better performance.
from the document of capnproto:

> Here are some tips for using the C++ Cap’n Proto runtime most effectively:
>    Accessor methods for primitive (non-pointer) fields are fast and inline. They should be just as fast as accessing a struct field through a pointer.
>    Accessor methods for pointer fields, on the other hand, are not inline, as they need to validate the pointer. If you intend to access the same pointer multiple times, it is a good idea to save the value to a local variable to avoid repeating this work.

    Example:

    // BAD
    frob(foo.getBar().getBaz(),
         foo.getBar().getQux(),
         foo.getBar().getCorge());

    // GOOD
    auto bar = foo.getBar();
    frob(bar.getBaz(), bar.getQux(), bar.getCorge());

selfdrive/ui/ui.hpp
```diff
@@ -266,16 +239,14 @@ typedef struct UIState {
...
+  bool preview_started;
   bool vision_seen;
 
   std::atomic<float> light_sensor;
 
   int touch_fd;
 
   // Hints for re-calculations and redrawing
-  bool model_changed;
-  bool livempc_or_radarstate_changed;
```
adeebshihadeh, line 278: remove irrelevant comment

deanlee: done

selfdrive/ui/sidebar.cc
```diff
@@ -34,7 +34,7 @@ static void ui_draw_sidebar_network_strength(UIState *s) {
   const int network_img_w = 176;
   const int network_img_x = !s->scene.uilayout_sidebarcollapsed ? 58 : -(sbr_w);
   const int network_img_y = 196;
-  const int img_idx = s->scene.networkType == cereal::ThermalData::NetworkType::NONE ? 0 : network_strength_map[s->scene.networkStrength];
+  const int img_idx = s->scene.thermal.getNetworkType() == cereal::ThermalData::NetworkType::NONE ? 0 : network_strength_map[s->scene.thermal.getNetworkStrength()];
```
adeebshihadeh, line 37: Why is this in a ternary instead of the map?

selfdrive/ui/ui.cc
```diff
@@ -471,9 +426,7 @@ void handle_message(UIState *s, SubMaster &sm) {
 }
 
 static void check_messages(UIState *s) {
-  while (true) {
-    if (s->sm->update(0) == 0)
-      break;
+  if (s->sm->update(0) > 0){
```
adeebshihadeh, line 429: while

selfdrive/ui/sidebar.cc
deanlee: they are not the same type.I copied this ternary from c-capnproto version.maybe we can make it better next time.

adeebshihadeh: ah, I see

selfdrive/ui/ui.cc
deanlee: sm->update() will set all updated to false before polling new message. if we use while here,we can't use sm->updated() out of handle_message again. (currently,sm.updated("model") and sm.updated("radarState") is used in paint.cc)

the sockets in SubMaster are always conflates, so maybe we don't have do use while here ?

adeebshihadeh: The while is just to preserve the old behavior, but it should be fine with the if.

deanlee: old version's SubSocket are not conflates. I think we can remove while now without any problems

PR #1586, opened 2020-05-28
UI: better read_model

deanlee:
update model data in-place instead of deep copy.


pd0wm:
Thanks! Can you also update the `update_all_lane_lines_data` function to pass the `PathData` by reference in `paint.cc`? According to LGTM that one copies a 1500 byte struct each call.

deanlee:
done

PR #1592, opened 2020-05-29
0.7.6 release

pd0wm:
```
Version 0.7.6 (2020-06-05)
========================
* White panda is deprecated, upgrade to comma two or black panda
* 2017 Nissan X-Trail and 2018 Nissan Leaf support thanks to avolmensky!
* 2017 Mazda CX-5 support in dashcam mode thanks to Jafaral!
* Huge CPU savings in modeld by using thneed!
* Lots of code cleanup and refactors
```


ayokings202:
Thanks for all the hard work guys :)

RELEASES.md
```diff
@@ -1,11 +1,18 @@
...
+* Lots of code cleanup and refactors
+
 Version 0.7.5 (2020-05-13)
 ========================
 * Right-Hand Drive support for both driving and driver monitoring!
 * New driving model: improved at sharp turns and lead speed estimation
 * New driver monitoring model: overall improvement on comma two
 * Driver camera preview in settings to improve mounting position
 * Added support for many Hyundai, Kia, Genesis models thanks to xx979xx!
-* 2019 Nissan X-Trail and 2018 Nissan Leaf support thanks to avolmensky!
```
geohot, line 8: Why did this change?

pd0wm: I forgot to actually put the files in previous release, so the release was moved to the 0.7.6 release notes.

cereal/log.capnp
```diff
@@ -855,6 +855,9 @@ struct LiveLocationKalman {
   gpsTimeOfWeek @14 :Float64;
   status @15 :Status;
   unixTimestampMillis @16 :Int64;
+  inputsOK @17 :Bool = true;
```
geohot, line 858: Hmm, why do these have initializers that set OK to true? If you forget to set it, it should default to bad. Does this matter on old data or something?

pd0wm: yes it's for old data. Otherwise controlsd will think the sensors are offline on old liveLocationData in the process replay test. Once we have a better infrastructure to regenerate logs for testing we can take it out.

PR #1595, opened 2020-05-30
Added 2020 Santa Fe fingerprint and updated interface values

Wykillin:
Added my 2020 Santa Fe's fingerprint as well as changed the values of the car in the interface.py.

I also changed the default color of the LKAS icon to orange. This is how the stock systems act when LKAS is not engaged. This just seems more normal to me.


TK211X:
Nothing = Off

White = Ready

Green = LKA

Orange = Fault

From my experience.

TK211X:
Can you get and add the FW print?

selfdrive/car/hyundai/values.py
```diff
@@ -25,7 +25,7 @@ class CAR:
   KIA_STINGER = "KIA STINGER GT2 2018"
   KONA = "HYUNDAI KONA 2019"
   KONA_EV = "HYUNDAI KONA ELECTRIC 2019"
-  SANTA_FE = "HYUNDAI SANTA FE LIMITED 2019"
+  SANTA_FE = "HYUNDAI SANTA FE 2019-2020"
```
pd0wm, line 28: No need to change the year in the car name.

selfdrive/car/hyundai/hyundaican.py
```diff
@@ -29,7 +29,7 @@ def create_lkas11(packer, frame, car_fingerprint, apply_steer, steer_req,
     # FcwOpt_USM 2 = Green car + lanes
     # FcwOpt_USM 1 = White car + lanes
     # FcwOpt_USM 0 = No car + lanes
-    values["CF_Lkas_FcwOpt_USM"] = 2 if enabled else 1
+    values["CF_Lkas_FcwOpt_USM"] = 2 if enabled else 4
```
pd0wm, line 32: Let's keep this the white icon. I'm pretty sure orange means something is faulted.

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -96,7 +96,7 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert,
 
 
     # 20 Hz LFA MFA message
-    if frame % 5 == 0 and self.car_fingerprint in [CAR.SONATA, CAR.PALISADE]:
+    if frame % 5 == 0 and self.car_fingerprint in [CAR.SONATA, CAR.PALISADE, CAR.SANTA_FE]:
```
pd0wm, line 99: Would this be a problem for the 2019 Santa FE that doesn't have LFA? Otherwise we need to split them into two different cars. Maybe you can find someone on discord with a 2019 one and have them try this PR.

selfdrive/car/hyundai/hyundaican.py
```diff
@@ -19,7 +19,7 @@ def create_lkas11(packer, frame, car_fingerprint, apply_steer, steer_req,
   values["CF_Lkas_MsgCount"] = frame % 0x10
   values["CF_Lkas_Chksum"] = 0
 
-  if car_fingerprint in [CAR.SONATA, CAR.PALISADE]:
+  if car_fingerprint in [CAR.SONATA, CAR.PALISADE, CAR.SANTA_FE]:
```
pd0wm, line 22: Same comment as the lfa_mfa message.

selfdrive/car/hyundai/carcontroller.py
TK211X: 2019 and 2020 are same platform

TK211X: Any vehicle that steers down to 0 has LFA supported.

Wykillin:
I've changed the icon back to white. I didn't realize that it was causing faults.

selfdrive/car/hyundai/carcontroller.py
Wykillin: I test drove a 2019 Santa Fe and it has the same displays. I believe the only thing that is different between the 2019 and 2020 is the fingerprint.

selfdrive/car/hyundai/hyundaican.py
Wykillin: I have changed back to white + lanes.

pd0wm:
Can you share a cabana route with the stock system of your car? I'd like to check if's correct that we're now sending the `mfa_lfa` message.

adeebshihadeh:
Feel free to reopen after sharing a route.

PR #1596, opened 2020-05-30
[GM] Show FCW from OP on dashboard UI / LED

kishkaru:
This change adds functionality to show OpenPilot's FCW warning to the car's OEM FCW system.
Tested on '18 Volt.

### DIFF
--- a/selfdrive/car/gm/carcontroller.py
+++ b/selfdrive/car/gm/carcontroller.py
@@ -60,6 +60,7 @@ class CarController():
     self.apply_steer_last = 0
     self.lka_icon_status_last = (False, False)
     self.steer_rate_limited = False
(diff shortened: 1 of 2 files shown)


selfdrive/car/gm/carcontroller.py
```diff
@@ -74,6 +75,10 @@ def update(self, enabled, CS, frame, actuators, \
     # Send CAN commands.
     can_sends = []
 
+    # FCW: trigger FCWAlert for 100 frames (4 seconds)
+    if hud_alert == VisualAlert.fcw:
+      self.fcw_frames = 100
```
kishkaru, line 80: This can be tweaked as needed. 4 seconds may be too long?

PR #1599, opened 2020-05-30
dmonitoring : use memory cache to avoid malloc/free on every frame

deanlee:


selfdrive/modeld/models/dmonitoring.h
```diff
@@ -31,6 +31,10 @@ typedef struct DMonitoringModelState {
   bool is_rhd;
   bool is_rhd_checked;
   float output[OUTPUT_SIZE];
+  kj::Array<uint8_t> resized_buf;
```
pd0wm, line 34: Any specific reason for using kj::Array instead of std::array?  If we don't have to rely on the capnp dependency that is always nice.

Also, aren't als these sizes static anyway, do wee need the logic to increase sizes?

deanlee: if we initialize the buffer in function dmonitoring_eval_frame (it's more simple and easy to understand), we still need an if statement to make sure the buffer is allocated.we can think this as an initialization logic.

PR #1600, opened 2020-05-30
driving.cc: Pass ModelDataRaw by reference

deanlee:


deanlee:
pass NVGpaint by refer too.

PR #1606, opened 2020-05-31
phonelibs cleanup

adeebshihadeh:
removes openmax from phonelibs, and boringssl and curl from release


pd0wm:
The NEOS updater uses curl + boringssl.

pd0wm:
If you want you can remove it from the release files since we ship the updater as a binary.

PR #1609, opened 2020-05-31
Nissan mph bit

avolmensky:
Added MPH signal for Rogue and X-Trail
Renamed SEATBELT message to HUD

Once opendbc changes have been merged I'll update the PR with the opendbc commit


selfdrive/car/nissan/carstate.py
```diff
@@ -11,6 +11,7 @@ class CarState(CarStateBase):
   def __init__(self, CP):
     super().__init__(CP)
     can_define = CANDefine(DBC[CP.carFingerprint]['pt'])
+    self.seatbeltUnlatched = True
```
pd0wm, line 14: Why add this?

PR #1611, opened 2020-05-31
Update RELEASES.md

AskAlice:


pd0wm:
We usually don't put every change to supported cars in the readme, they would get rather long otherwise. It's mostly used to thank people for writing new car ports, or for cars that are significantly different than others and required lot of work.

PR #1615, opened 2020-06-01
Mazda: Slower rate for ACC cancel msg to avoid cruise disable

Jafaral:
Sending ACC cancel message at 10hz may end up disabling
main cruise state. That is because stoc ACC use the same
signal for both functions. If cruise is already enabled
the signal canceles it, otherwise cruise is disabled.
10hz seems to be still fast in some cases not allowing
the state to sync up between OP and the Stock state.
5hz avoids that issue. Also, the resume message was also
updated to 5hz as that seems to work better as well.

### DIFF
--- a/selfdrive/car/mazda/carcontroller.py
+++ b/selfdrive/car/mazda/carcontroller.py
@@ -23,17 +23,17 @@ class CarController():
                                                   CS.out.steeringTorque, SteerLimitParams)
       self.steer_rate_limited = new_steer != apply_steer
 
-      if CS.out.standstill and frame % 50 == 0:
+      if CS.out.standstill and frame % 20 == 0:
         # Mazda Stop and Go requires a RES button (or gas) press if the car stops more than 3 seconds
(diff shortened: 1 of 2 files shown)


selfdrive/car/mazda/carcontroller.py
```diff
@@ -23,17 +23,17 @@ def update(self, enabled, CS, frame, actuators):
                                                   CS.out.steeringTorque, SteerLimitParams)
       self.steer_rate_limited = new_steer != apply_steer
 
-      if CS.out.standstill and frame % 50 == 0:
+      if CS.out.standstill and frame % 20 == 0:
```
adeebshihadeh, line 26: Can you use DT_CTRL instead?

Jafaral: not familiar with it pointer?

Jafaral: I'm going to leave this as is for now @adeebshihadeh as it works fine now. DT_CTRL will require more code changes and more test drives which delay things if we want to get these fixes in 0.7.6.

PR #1619, opened 2020-06-02
update-pipfile.yml: Remove Pipfile.lock before locking

EwoutH:
Small bugfix to the Update Pipfile.lock workflow. Should fix [this failed run](https://github.com/commaai/openpilot/runs/727743464?check_suite_focus=true).

### DIFF
--- a/.github/workflows/update-pipfile.yml
+++ b/.github/workflows/update-pipfile.yml
@@ -11,6 +11,7 @@ jobs:
     - uses: actions/setup-python@v2
     - run: pip install wheel
     - run: pip install pipenv
+    - run: rm Pipfile.lock
     - run: pipenv lock
     - uses: actions/upload-artifact@v2
       with:
--- a/.github/workflows/update-pipfile.yml
+++ b/.github/workflows/update-pipfile.yml
@@ -10,8 +10,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
     - run: pip install wheel
-    - run: pip install pipenv
-    - run: rm Pipfile.lock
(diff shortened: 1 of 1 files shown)


EwoutH:
@pd0wm Small comment, could you please add a short explanation about why a change in CI configuration is made, in [commits like this](https://github.com/commaai/openpilot/commit/db8f3aee1189533b3cd733b8723c9b01e0c157dc)? A single line is enough most of the times but help a lot for tracking and understanding configuration changes. I assume it's because of [this issue](https://github.com/commaai/openpilot/pull/1563#issuecomment-635644608), but's it's nice to have it documented.

I also noticed #1580 contained a Pipfile.lock containing python_version markers. Are those desired, and if so what configuration (and pipenv version) do you use to generate the Pipfile.lock? Then I will adjust the CI to also generate those.

EwoutH:
Unfortunately this doesn't fix the issue. It probably is an regression with the major [Pipenv release](https://github.com/pypa/pipenv/releases).

EwoutH:
Fixed the pipenv version to 2018.11.26 on a [test branch](https://github.com/EwoutH/openpilot/tree/ci-test) and this resolved the issue. 2020.5.28 (and the following bugfix 2020.6.2) introduced huge changes, so a regression was likely.

I suggest merging this and trying next month to unlock the pipenv version again.

pd0wm:
I disable running on branches because it always uses the same branch to make a PR from. This wouldn't work if there were two branches open at the same time that need a pipfile update.

I'm actually thinking of switching to dependabot. Seems like they already do all of this, and we don't have to maintain it.

PR #1626, opened 2020-06-02
Make pylint more strict

pd0wm:


selfdrive/athena/athenad.py
```diff
@@ -29,7 +29,7 @@
 from selfdrive.swaglog import cloudlog
 
 ATHENA_HOST = os.getenv('ATHENA_HOST', 'wss://athena.comma.ai')
-HANDLER_THREADS = os.getenv('HANDLER_THREADS', 4)
+HANDLER_THREADS = int(os.getenv('HANDLER_THREADS', "4"))
```
adeebshihadeh, line 32: why make the default a string?

pd0wm: because getenv returns a string. This would have broken if you actually passed in an env variable.

PR #1636, opened 2020-06-04
Don't auto-resume Honda Bosch ACC if we have an active steerRequired alert

csouers:
This would cover all DM events like driver terminally distracted, low driver detection accuracy, and driver distracted but not yet terminal (because we're stopped). Also, it would prevent auto resuming if the driver has violated the safety model as the lead car has started to move but we haven't yet disengaged because of it (unlatched seatbelt, for example).

Once caveat is that it could also not resume if the vision model output is uncertain, but that would probably be a rare occurrence. Also, this doesn't enforce DM driver attentiveness for allowing auto resume.

Todo: Expand this to all makes and control styles.

### DIFF
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -161,7 +161,7 @@ class CarController():
       # If using stock ACC, spam cancel command to kill gas when OP disengages.
       if pcm_cancel_cmd:
         can_sends.append(hondacan.spam_buttons_command(self.packer, CruiseButtons.CANCEL, idx, CS.CP.carFingerprint, CS.CP.isPandaBlack))
-      elif CS.out.cruiseState.standstill:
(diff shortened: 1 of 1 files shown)


pd0wm:
Interesting thought.

I'm not against this change, but I feel like if we decide to change this we need to do it across all cars where have some control over longitudinal. E.g. it shouldn't start moving when you're not paying attention. I'll discuss it internally.

csouers:
I totally agree with a global change. This would make the push for using DM even stronger. It’s good enough these days and is working for RHD cars now. There isn’t really a valid reason not to use it anymore. I suppose a secondary method would to be require a wheel touch to initiate the resume if the user isn’t using DM or the face detection is poor. 

For transparency, this isn’t entirely my original idea. Someone from the Discord complained about it and that got the conversation going. George seemed on board with it too. I’ll close this PR for now.

PR #1638, opened 2020-06-05
Reduce scheduler latency for realtime processes

jyoung8607:
Under certain conditions, the Qualcomm RIL and friends like to spend tens or hundreds of milliseconds doing non-preempt-able work of some sort, which doesn't play well with the 100Hz loops in `boardd` and `controlsd`. Since they can't play nice together, we'll have them play on different playgrounds. 

Changes:

* Pin low priority interrupts (modem, wifi, flash storage) to cpu 1, the second Kryo efficiency core
* Pin the highest-demand real-time tasks to cpu 3, the second Kryo performance core
* Prevent all Android stuff from running on cpu 3, but allow OP processes to overflow there
* Tweak all OP real-time priorities to sit clearly above Android mediaserver and system_server

General thinking is there's a nice USB->boardd->controlsd and reverse flow that shouldn't result in much contention or preemption among the three.

With these changes, `boardd` no longer takes missed cycles under any realistic conditions I've been able to find. It now handles SIM eject and reinsert cleanly, which it didn't before. It does a much better, but not completely perfect job of surviving modem subsystem restart, but that's a more difficult scenario and not representative of a real-world event.

### DIFF
--- a/launch_chffrplus.sh
+++ b/launch_chffrplus.sh
@@ -59,7 +59,7 @@ function launch {
   # NEOS installed app processes can run anywhere
   echo 0-2 > /dev/cpuset/background/cpus
   echo 0-2 > /dev/cpuset/system-background/cpus
-  echo 0-2 > /dev/cpuset/foreground/boost/cpus
+  [ -d "/dev/cpuset/foreground/boost/cpus" ] && echo 0-2 > /dev/cpuset/foreground/boost/cpus  # Not present in < NEOS 15
   echo 0-2 > /dev/cpuset/foreground/cpus
   echo 0-2 > /dev/cpuset/android/cpus
   echo 0-3 > /dev/cpuset/app/cpus
--- a/launch_chffrplus.sh
+++ b/launch_chffrplus.sh
@@ -55,18 +55,22 @@ function launch {
     fi
   fi
 
-  # no cpu rationing for now
-  echo 0-3 > /dev/cpuset/background/cpus
-  echo 0-3 > /dev/cpuset/system-background/cpus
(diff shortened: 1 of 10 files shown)


launch_chffrplus.sh
```diff
@@ -55,18 +55,22 @@ function launch {
...
+  echo 0-2 > /dev/cpuset/android/cpus
+  echo 0-3 > /dev/cpuset/app/cpus
+
+  # Collect RIL and other possibly long-running I/O interrupts onto CPU 1
+  echo 1 > /proc/irq/78/smp_affinity_list # qcom,smd-modem (LTE radio)
+  echo 1 > /proc/irq/33/smp_affinity_list # ufshcd (flash storage)
+  echo 1 > /proc/irq/35/smp_affinity_list # wifi (wlan_pci)
+  # USB traffic needs realtime handling on cpu 3
+  [ -d "/proc/irq/733" ] && echo 3 > /proc/irq/733/smp_affinity_list # USB for LeEco
+  [ -d "/proc/irq/736" ] && echo 3 > /proc/irq/736/smp_affinity_list # USB for OP3T
```
adeebshihadeh, line 73: Maybe we should wrap these in an if statement so it doesn't just silently fail if for some reason both of these don't exist?

jyoung8607: Short term, this script really only works on NEOS, which will definitely have one of those two interrupts. Long term, I'd like to have at-launch platform detection and then have individual platform setup routines, such that we can automatically startup on Ubuntu, OSX, NEOS/EON, NEOS/LEON, NEOS/C2, any future platform, etc. That's becoming more important every day. For those scripts, we can and should fail loudly if platform mandatory stuff is missing.

PR #1645, opened 2020-06-05
boardd: no conflate on sendcan

### LINKED ISSUE #1641: Intermittent temp steering faults for Toyota
Some users are seeing intermittent temp steering faults on sharp turns for Toyota. A solution in openpilot was attempted in #1256 but not yet accepted.

It's been reported that driving in all-output safety mode makes the problem go away. This is not yet fully verified, and obviously that's not a good daily-drive configuration, but it's of diagnostic interest. It tells us the Panda safety mode may be discarding messages for torque level or ramp-rate violations, and the temp steering faults from the steering rack could be a consequence of those missing messages.

If this can be verified, it means we need to find and fix two issues:

1) We have airspace between what OP can send and what Panda safety is willing to pass
2) We have airspace between what OP can send and Panda regression testing is exercising

If you're experiencing this issue, please add a comment with a drive segment log, and I'll take a look early next week.

@wocsor -  fbd011384db5e669|2020-06-04--15-46-08--11
@ShaneSmiskol - ?

pd0wm:
Boardd was dropping messages if sent too fast

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -589,16 +589,28 @@ void can_send(cereal::Event::Reader &event) {
 
 void *can_send_thread(void *crap) {
   LOGD("start send thread");
-  SubMaster sm({"sendcan"});
 
-  // drain sendcan to delete any stale messages from previous runs
-  sm.drain();
+  Context * context = Context::create();
+  SubSocket * subscriber = SubSocket::create(context, "sendcan");
+  assert(subscriber != NULL);
+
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Should resolve #1641

deanlee:
@pd0wm should we add the "bool conflate" back to SubMaster's constructor?

pd0wm:
Submaster is really designed to have the latest available packet. I think it's fine to have a raw sub socket in this case.

PR #1650, opened 2020-06-05
Getting PC openpilot working again

geohot:
@adeebshihadeh Can we run the modeld tests on PC as well?


adeebshihadeh:
The existing test should just run on PC. Let me try adding it to GitHub Actions.

geohot:
Oh and a great benefit of this is that we can generate the golden truth on the computer then!

PR #1651, opened 2020-06-06
Potential fix for steer unavailable after fast override

pd0wm:


ShaneSmiskol:
Are there any updates on this?

pd0wm:
We're working on it. Reproduced it today on our Rav4.

I still need to find some more time to look into more data. I'm a little reluctant to merge this as is, since I'm not sure when false positives happen. I'll need to search through driving data and look how often the steer rate goes over a 100. I think we also want to add a check to see if the torque is in the same direction as the steer rate and only cut if they are in the same direction. \

Another concern I have is that this violates the rate limit in the panda safety code, so instead of sending 0 torque, the message will just get blocked by the panda. Can you share a cabana drive with this triggering? You'll probably see that it stops sending instead of sending 0.

ShaneSmiskol:
Sure, do you want a drive  on stock so it faults or with this change to cut torque when rate goes above 100? I don't think it's too common where the rate reaches this limit and we *aren't* on a severe curve, for me it only faults when openpilot has trouble on a sharp curve where it vastly oversteers first, and then when it releases torque, it releases too much.

While this PR would fix a really annoying bug and safety issue, I think it's just treating the side-effect of bad steering tuning or even possibly an unstable model that predicts a curve is sharper than it is. (in my video I uploaded on the old issue you could see it tried to bring me into the oncoming side of the road when I had no lane lines, thus when it tried to center the wheel again it caused the fault). So either a better steering controller (for example, how does Tesla control the wheel with such precision?) or better models in the future that don't have its predictions jumping around could fix this and remove the need for this semi-hacky bodge.

Because even with this change it doesn't make the steering around curves any better, it just fixes it not steering for 2 to 4 seconds after; where sometimes it's able to recover and sometimes not.

adeebshihadeh:
@ShaneSmiskol a drive with this change where it triggers

ShaneSmiskol:
Took openpilot on some back roads where it activated multiple times, uploading the drive now

adeebshihadeh:
Shane's been looking into this more, and the faults don't seem to be caused by the steering rate.

PR #1652, opened 2020-06-06
paint.cc: add inline keyword to function lerp

deanlee:


pd0wm:
Any measurable improvement? I think manually adding inline is frowned upon, since the compiler usually does a better job at deciding to inline or not.

deanlee:
just want to make sure it's inlined.this function is called in update_track_data,and I think it's really designed as inline.

pd0wm:
If it's faster the compiler will already do it. Don't we need to care about this kind of micro-optimizations.

PR #1653, opened 2020-06-06
paint.cc : Faster lane line drawing

deanlee:
optimize ghost lane line drawing, save many loops.(save 4x rcount calls to car_space_to_full_frame and nvgLineTo)

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -213,20 +213,12 @@ static void update_all_track_data(UIState *s) {
 
 
 static void ui_draw_track(UIState *s, bool is_mpc, track_vertices_data *pvd) {
+  if (pvd->cnt == 0) return;
+
   nvgBeginPath(s->vg);
-  bool started = false;
-  for(int i = 0;i < pvd->cnt;i++) {
-    float x = pvd->v[i].x;
-    float y = pvd->v[i].y;
-    if (x < 0 || y < 0) {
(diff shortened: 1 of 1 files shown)


selfdrive/ui/paint.cc
```diff
@@ -127,25 +127,23 @@ static void draw_lead(UIState *s, const cereal::RadarState::LeadData::Reader &le
   draw_chevron(s, d_rel, lead.getYRel(), 25, nvgRGBA(201, 34, 49, fillAlpha), COLOR_YELLOW);
 }
 
-static void ui_draw_lane_line(UIState *s, const model_path_vertices_data *pvd, NVGcolor color) {
+static void ui_draw_lane_line(UIState *s, const model_path_vertices_data *pvd, NVGcolor color, bool is_ghost) {
```
adeebshihadeh, line 130: You were able to remove lines by assuming the order of lane line drawing?

adeebshihadeh: If that's true, I'm not a fan of this change. Sounds like a time-consuming bug waiting to be written the next time someone touches this code.

deanlee: nothing about the order of drawing.
the ghost line's width is always 1px.we don't need to draw it twice(draw back from line end to start), that can save 196 calls to car_space_to_full_frame and  lineto if path.validLen is 98

PR #1654, opened 2020-06-06
Fixups for Mac, and a bunch for PC

geohot:


geohot:
Shouldn't have changed much, a bunch of things that only apply to PC. Three things apply to phone:
1. Using ipc_bind in camerad, I think it's the same code.
2. Removing CL_MEM_SVM_FINE_GRAIN_BUFFER
3. Replace map, memcpy, unmap with clEnqueueWriteBuffer

selfdrive/camerad/main.cc
```diff
@@ -1215,10 +1210,12 @@ void party(VisionState *s) {
 
   zsock_signal(s->terminate_pub, 0);
 
+#ifndef __APPLE__
 #if !defined(QCOM2) && !defined(QCOM_REPLAY)
```
adeebshihadeh, line 1214: why a separate condition?

selfdrive/camerad/main.cc
```diff
@@ -1194,12 +1186,15 @@ void party(VisionState *s) {
                        processing_thread, s);
   assert(err == 0);
 
+#ifndef __APPLE__
+  // TODO: should this be for Linux too?
 #ifndef QCOM2
```
adeebshihadeh, line 1191: same as above

PR #1659, opened 2020-06-08
add set_brightness in framebuffer.h

deanlee:
there are three places  (framebuffer.cc, text.cc, paint.cc)  call fopen("/sys/class/leds/lcd-backlight/brightness") to set the brightness.

combine them into one set_brightness function.

### DIFF
--- a/selfdrive/ui/linux.cc
+++ b/selfdrive/ui/linux.cc
@@ -64,6 +64,8 @@ void framebuffer_swap(FramebufferState *s) {
   glfwPollEvents();
 }
 
+bool set_brightness(int brightness) { return true; }
+
 void touch_init(TouchState *s) {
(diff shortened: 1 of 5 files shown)


selfdrive/ui/linux.cc
```diff
@@ -64,6 +64,8 @@ void framebuffer_swap(FramebufferState *s) {
   glfwPollEvents();
 }
 
+bool set_brightness(int brightness) { return true; }
```
pd0wm, line 67: Why do we need this implementation? The other one will just not do anything on a computer.

deanlee: framebuffer.cc will not be built on computer, this dummy implementation is used to avoid the link error.

PR #1660, opened 2020-06-08
UI: remove function is_leon

deanlee:
use util::read_file to check if it's letv. (util::read_file will return an empty string on fail.)


adeebshihadeh:
Nice cleanup, thanks for the PR!

PR #1661, opened 2020-06-08
remove unnecessarily cancel spamming

xx979xx:
this causes HKG cars to send cancel continuously when change enableCruise param
I don't see why OP should  spams the car with cruise cancel button when Cruise control is off


pd0wm:
This should not change anything for Hyundai. `not self.CP.enableCruise` is True so this change doesn't change anything.

PR #1663, opened 2020-06-08
Support for NEOS 15

jyoung8607:
Requirements:

- [x] Incorporate NEOS 15 from commaai/eon-neos-builder#26
  - [x] Pointer for expected NEOS version
  - [x] Pointer for OTA updates
- [x] UI to take over framebuffer power control from NEOS
- [x] Fix issue with libSNPE linking of libsymphony-cpu
- [x] OTA in-place upgrading
- [x] OTA in-place downgrading
- [x] Fix Android "unused DT entry" linker complaints

Moved to separate projects:

* Background updater, will work in concert with OP background updater and overlay
* Deeper power save mode using accelerometer for touch hinting, standalone project


jyoung8607:
This runs and drives without apparent drama. I'm still testing out some OTA self-upgrade and self-downgrade scenarios. Notes:

* Upgrading in-place needs a `scons --clean`. Just switching branches from `master` to `neos15updates` won't rebuild already-compiled Cython `.so` objects which are linked to Python 3.7 libraries. There may be some way to have `scons` notice and handle this for us.
* Downgrading needs manual help. `scons` fails with "ValueError : unsupported pickle protocol: 5" until you `rm /data/openpilot/.sconsign.dblite`. Just running `scons --clean` isn't enough. We may want to do both automatically whenever a NEOS update is completed.
* Seeing a few `boardd` missed cycles creep back in. Unsure why, because worst-case scheduler latency on core 3 still looks really good. Looking into it.

jyoung8607:
The NEOS updater is sometimes saying it failed to download the recovery image, instantly after I try. I think the real problem is I'm trying within a couple seconds of the dialog coming up and WiFi hasn't finished connecting yet. I may add a short sleep somewhere to avoid that.

SConstruct
```diff
@@ -105,6 +102,11 @@ ldflags_asan = ["-fsanitize=address"] if GetOption('asan') else []
 # change pythonpath to this
 lenv["PYTHONPATH"] = Dir("#").path
 
+try:
+  rpath
+except:
+  rpath=[]
```
adeebshihadeh, line 108: Seems simpler to write:
```
if rpath is None:
  rpath = []
```

jyoung8607: Agreed, this is done.

AskAlice:
I'll still be writing to /system if `/home/.zshrc` / (aka `/data/data/com.termux/files/home/.zshrc`) doesn't exist.

jyoung8607:
> My eon wasn't working for a few days, just stuck in qualcomm mode unable to get to fastboot, but that was definitely unrelated as I hadn't checked out this branch yet. Finally got it installed on it and I got this:

I have seen that here as well, working on figuring out when it crept in. I'm going to be closing this PR shortly and opening another as github has gotten confused about the history after rebasing, but consider that reported and replicated.

PR #1664, opened 2020-06-08
Palisade INDI 

scottadair:
Found that INDI works much better with the Palisade. Test drove for about 400 kms over the weekend, mix of city and highway, and this config performed better than PID.


pd0wm:
When I have some time I'll try it on the Sonata. Did you also try to improve the PID tuning first?

scottadair:
Yep I did. PID worked pretty good, but found that this takes corners better.

scottadair:
Strange, it handles great on the highway but had lots of issues with some city driving. Going to go back to the drawing board. Will close for now.

PR #1665, opened 2020-06-08
recover EON/C2 AF

ZwX1616:
goal: fix cameras in some devices sometimes not focusing

triggers (any should trigger a recovery attempt):
1. lens stuck due to wrong sensor output
2. focused position may not on the road


pd0wm:
Can you add a cloudlog for this so we can track how often it happens and look at footage?

selfdrive/camerad/main.cc
```diff
@@ -470,6 +470,18 @@ void* processing_thread(void *arg) {
     /*t11 = millis_since_boot();
     printf("process time: %f ms\n ----- \n", t11 - t10);
     t10 = millis_since_boot();*/
+
+    // setup self recover
+    if (is_blur(&s->lapres[0]) && (s->cameras.rear.lens_true_pos < (DEVICE_LP3? LP3_AF_DAC_DOWN:OP3T_AF_DAC_DOWN)+1||s->cameras.rear.lens_true_pos > (DEVICE_LP3? LP3_AF_DAC_UP:OP3T_AF_DAC_UP)-1) && s->cameras.rear.self_recover < 2) {
+      s->cameras.rear.self_recover -= 1;
+      if (s->cameras.rear.self_recover < -FOCUS_RECOVER_PATIENCE) {
+        LOG("rear camera bad state detected. attempting recovery from %.1f", s->cameras.rear.lens_true_pos);
```
pd0wm, line 478: Why no LOGW or LOGE?

selfdrive/camerad/cameras/camera_qcom.cc
```diff
@@ -2039,10 +2047,13 @@ static void* ops_thread(void* arg) {
...
   assert(sensor_sock);
 
+  zsock_t *controls_sock = zsock_new_sub(">tcp://127.0.0.1:8007", "");
+  assert(controls_sock);
+
   zsock_t *terminate = zsock_new_sub(">inproc://terminate", "");
   assert(terminate);
 
-  zpoller_t *poller = zpoller_new(cameraops, sensor_sock, terminate, NULL);
+  zpoller_t *poller = zpoller_new(cameraops, sensor_sock, controls_sock, terminate, NULL);
```
pd0wm, line 2056: Need to use a msgq sock for controlsState

selfdrive/controls/controlsd.py
```diff
@@ -208,6 +208,9 @@ def update_events(self, CS):
       self.events.add(EventName.vehicleModelInvalid)
     if not self.sm['liveLocationKalman'].posenetOK:
       self.events.add(EventName.posenetInvalid)
+    if not self.sm['frame'].recoverState < 2:
```
pd0wm, line 211: use capnpn enum for the field

ZwX1616: this is more like just a counter number other than having any discrete meaning. I would say just keep it this way?

selfdrive/camerad/cameras/camera_qcom.cc
ZwX1616: lol. fixed

PR #1685, opened 2020-06-12
prevent cloudlog spam in tmux a unless manager isn't doing its job

AskAlice:
so much dev friction with this one honestly.

May I suggest, if you don't like the way this works, log when a process dies only, and prevent spam entirely?

### DIFF
--- a/selfdrive/manager.py
+++ b/selfdrive/manager.py
@@ -496,9 +496,15 @@ def manager_thread():
       elif "driverview" in running and params.get("IsDriverViewEnabled") == b"0":
         kill_managed_process("driverview")
 
-    # check the status of all processes, did any of them die?
-    running_list = ["%s%s\u001b[0m" % ("\u001b[32m" if running[p].is_alive() else "\u001b[31m", p) for p in running]
-    cloudlog.debug(' '.join(running_list))
+    # check the status of all processes, did any of them die?'
+    process_dead = False
+    for p in running:
+      if not running[p].is_alive():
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
What problem does this solve?

AskAlice:
@adeebshihadeh if you're developing openpilot, and you get a traceback, the tmux buffer probably fills up 100x slower in comparison, which is great when you get a full screen disengage alert on the highway and feel like pulling over to send the logs back, or just general sanity when you're testing code within a vehicle. I've lost a few juicy tracebacks just because the buffer fills up with these useless messages

pd0wm:
Should we just increase the tmux traceback in neos? @jyoung8607 can we do something like 7200 llines? That should be enough for an hour.

jyoung8607:
@AskAlice, right now we value the heartbeat in the logs for a few reasons, but I'm going to increase the tmux scrollback buffer, and @pd0wm and I are talking about making it more convenient for community developers to capture tmux console output to disk. Stay tuned.

jyoung8607:
> Should we just increase the tmux traceback in neos? @jyoung8607 can we do something like 7200 llines? That should be enough for an hour.

Done: https://github.com/commaai/eon-neos-builder/pull/26/commits/1aea85e47001740aa65c7018a45488233a1e13a7

AskAlice:
beautiful. thanks. I knew you guys had a better solution, i just wanted you to have it on your mind as you're thinking about updating neos

PR #1688, opened 2020-06-12
Cool down device before going onroad

pd0wm:


pd0wm:
2020-06-12-144547_2023x1131_scrot

offroad cpu temperatures over the last 14 days. 75 degrees looks like an acceptable threshold.  245 / 224196 measurements.

PR #1699, opened 2020-06-14
Fix resource leak in visionbuf_free

deanlee:


selfdrive/common/visionbuf_ion.c
```diff
@@ -137,6 +138,9 @@ void visionbuf_sync(const VisionBuf* buf, int dir) {
 }
 
 void visionbuf_free(const VisionBuf* buf) {
+  clReleaseMemObject(buf->buf_cl);
+  munmap(buf->addr, buf->len + PADDING_CL);
```
pd0wm, line 142: What do you think of saving the actual length in the buf struct as well? If we ever change the padding computation this will cause bugs.

PR #1716, opened 2020-06-16
use SubMessage to handle non conflate sockets

deanlee:
https://github.com/commaai/cereal/pull/57

updated


adeebshihadeh:
Can you keep PRs that aren't ready to merge as drafts, and only mark them as ready for review when CI passes?

deanlee:
These PRS are ready for review.they can't pass CI test because they depend on another PR in cereal

pd0wm:
Not worth the complexity

PR #1729, opened 2020-06-17
GM cleanup + ignition fix

adeebshihadeh:


selfdrive/car/gm/interface.py
```diff
@@ -156,6 +159,8 @@ def update(self, c, can_strings):
       events.add(EventName.resumeRequired)
     if self.CS.pcm_acc_status == AccState.FAULTED:
       events.add(EventName.controlsFailed)
+    if ret.vEgo < self.CP.minSteerSpeed:
+      events.add(car.CarEvent.EventName.belowSteerSpeed)
```
kishkaru, line 163: Is this necessary? Due to:
```
    if ret.vEgo < self.CP.minEnableSpeed:
```
We won't be able to engage below `minEnableSpeed` anyway (which is currently set to 18mph for Volt).

Change `minEnableSpeed` to 7mph instead? (This is true for Volt, but I'm sure about other GM vehicles).
FWIW, 2018's can engage at 0mph, but 2017s and earlier cannot.

adeebshihadeh: This covers the case where you're already engaged and you go below the minSteerSpeed.

kishkaru: Ok, gotcha. I think a warning when < 7mph is fair (although arguably distracting in stop-n-go traffic).

selfdrive/car/gm/interface.py
```diff
@@ -48,6 +49,8 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), has_relay=False,
       ret.steerRatio = 15.7
       ret.steerRatioRear = 0.
       ret.centerToFront = ret.wheelbase * 0.4  # wild guess
+      ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0., 15.0], [0., 15.0]]
+      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.2, 0.4], [0.02, 0.08]]
```
kishkaru, line 53: I tried these PID tuning values in 076 and it was arguably worse than the earlier values. It was jerking the wheel back and forth more during sharper curves, at lower speeds. Does the tuning need to be on 08 to work properly?

adeebshihadeh: Reverted the tuning as it was worse at some speeds

PR #1730, opened 2020-06-17
Update wrongCarMode alert

AskAlice:
Co-authored-by: xx979xx <40252818+xx979xx@users.noreply.github.com>

see: https://discordapp.com/channels/469524606043160576/524611978208215070/722770077526523984


AskAlice:
I'd update the EventName in the code to also be more verbose, but I'd rather not PR cereal just to change some display text

pd0wm:
Can you add a callback function to this alert to change the text based on car type? On honda nidec there is a button labeled "main", so this alert makes sense there. For other car brands we probably want the more generic text.

adeebshihadeh:
Done in cab96374c5c511adcb83882c52502924a2d08089

PR #1735, opened 2020-06-18
updated.py : unlock lockfile before exit

deanlee:


pd0wm:
Is this solving a problem you obeserved? flock should automatically release the lock when the fd is closed, which happens when the process exits.

deanlee:
I'm sorry to take up your time,I  made a mistake: I must have pressed ctrl+c twice ,so the update.py is still running that's why I got this error. I tested this PR after reboot with single ctrl+c and thought  I have "fixed" it lol.
sorry again for trouble you @pd0wm

PR #1738, opened 2020-06-18
Fix gyro orientation offset for EONs with OP3T mainboards

jyoung8607:
The localizer in 0.8 exposed a problem with some devices reporting orientation with an strange offset. Only OP3T-based devices seem affected, and only a subset of those. Resetting the sensor HAL registry and letting the device rebuild it seems to clear the issue. The origin of the bad sensor registry data isn't clear. No other devices seem to be affected.

As of this change, all OP3T will have their sensor registry reset and rebuilt, exactly one time. The old sensor registry is kept as a backup (in case this has unforeseen fallout) and it serves as a flag that we've already performed the reset on this device.

### DIFF
--- a/launch_chffrplus.sh
+++ b/launch_chffrplus.sh
@@ -101,8 +101,8 @@ function launch {
 
   # One-time fix for a subset of OP3T with gyro orientation offsets.
   # Remove and regenerate qcom sensor registry. Only done on OP3T mainboards.
-  # Performed exactly once. Existing registry is preserved just-in-case, and
-  # doubles as a flag denoting we've already done the reset once.
+  # Performed exactly once. The old registry is preserved just-in-case, and
+  # doubles as a flag denoting we've already done the reset.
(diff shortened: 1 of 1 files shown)


TK211X:
Wouldn’t it be better to periodically calibrate the 0 position of the sensor against vision data so that it accounts for any variables in installation position. I’m sure every sensor has some variance in how it’s installed on the board. This would make it so that it wouldn’t matter what position the sensor is in.


This is how an AHRS unit works and I think implementation of this logic would be universal to any previous or future product.

pd0wm:
We already do this in the backend, which is how we found the issue with the op3ts. We just want to reset them once to make sure the consistent offset is gone.

jyoung8607:
Checks out with repeated loop testing for sensor count, and limited drive testing.

jyoung8607:
Merged per offline discussion with @pd0wm.

PR #1764, opened 2020-06-23
Add firmware for 2020 RAV4 Hybrid TSS2

Taik:
Tested and verified on my car.


selfdrive/car/toyota/values.py
```diff
@@ -925,12 +925,14 @@ class CAR:
   },
   CAR.RAV4H_TSS2: {
     (Ecu.engine, 0x700, None): [
+      b'\x02896634A23001\x00\x00\x00\x00897CF1203001\x00\x00\x00\x00',
```
adeebshihadeh, line 928: duplicate

selfdrive/car/toyota/values.py
```diff
@@ -939,18 +941,21 @@ class CAR:
       b'F152642541\x00\x00\x00\x00\x00\x00',
     ],
     (Ecu.eps, 0x7a1, None): [
+      b'8965B42171\x00\x00\x00\x00\x00\x00',
```
adeebshihadeh, line 944: duplicate

selfdrive/car/toyota/values.py
```diff
@@ -939,18 +941,21 @@ class CAR:
...
       b'8965B42170\x00\x00\x00\x00\x00\x00',
       b'8965B42171\x00\x00\x00\x00\x00\x00',
       b'8965B42181\x00\x00\x00\x00\x00\x00',
       b'\x028965B0R01200\x00\x00\x00\x008965B0R02200\x00\x00\x00\x00',
       b'\x028965B0R01300\x00\x00\x00\x008965B0R02300\x00\x00\x00\x00',
     ],
     (Ecu.fwdRadar, 0x750, 0xf): [
+      b'\x018821F3301400\x00\x00\x00\x00',
```
adeebshihadeh, line 952: duplicate

adeebshihadeh (changes requested):
Once the duplicates are fixed, we can merge this. Added a test to catch this in the future b5efaef98afb812846a5f1c8df07970a9c73e7be

selfdrive/car/toyota/values.py
Taik: Just double checked and this looks like it's the only unique value.

PR #1766, opened 2020-06-23
Correct a typo in compat.py

felixonmars:


adeebshihadeh:
Seems like this unused, deleted in master

PR #1771, opened 2020-06-24
bridge.py optimized to run in low resources machines (without GPU)

sjdavalle:
added simple parameter to launch the bridge with autopilot=ON --> bridge.py -a
refactor in go function to avoid Carla simluator to crash...
- get available world instead of forcing one.
- get available car and pick one randomly since tesla is not available
- avoid creating a VehicleControl object on every loop --> this optimizes the memory usage.


tools/sim/bridge.py
```diff
@@ -80,138 +84,164 @@ def go(q):
...
+          if throttle_out > 0.3:
+            cruise_button = CruiseButtons.CANCEL
+            is_openpilot_engaged = False
+        if m[0] == "brake":
+          brake_out = float(m[1]) / 100.
+          if brake_out > 0.3:
+            cruise_button = CruiseButtons.CANCEL
+            is_openpilot_engaged = False
+        if m[0] == "reverse":
+          in_reverse = not in_reverse
```
ZwX1616, line 201: seems that VehicleControl is not picking up this?

tools/sim/bridge.py
```diff
@@ -80,138 +84,164 @@ def go(q):
...
-        is_openpilot_engaged = False
-      if m[0] == "cruise":
-        if m[1] == "down":
-          cruise_button = CruiseButtons.DECEL_SET
-          is_openpilot_engaged = True
-        if m[1] == "up":
-          cruise_button = CruiseButtons.RES_ACCEL
-          is_openpilot_engaged = True
-        if m[1] == "cancel":
+  try:
```
ZwX1616, line 91: I suggest at least print the exception

ZwX1616:
I added some miscellaneous comments. other than that lgtm.
can you rebase on the latest master and make sure it still works? so that I can safely merge it, thanks

tools/sim/bridge.py
sjdavalle: If you mean... reverse is not being handle by the keyboard_ctrl... you are right
image
only cruise up/down/cancel are being capture

tools/sim/bridge.py
sjdavalle: will do

adeebshihadeh:
Does the whole thing (openilot + carla) run ok without a GPU?

sjdavalle:
@adeebshihadeh yes, it works... after some minutes you get a segmentation fault and the simulator stops... (my system runs out of memory)...
I've tried all the examples included in Carla, and all of them works perfectly... I only see the segmentation fault when running the whole thing (openpilot + carla) after some minutes... (I think we can continue working on this optimization)

adeebshihadeh:
There's a bunch of irrelevant changes here. Can you clean this up and we'll get this in with #2011?

PR #1773, opened 2020-06-24
Hyundai Ioniq EV port

baldwalker:
adding fingerprint for Ioniq EV 2020 Premium SE
also addition to car controller to stop LFA dash error


adeebshihadeh:
It seems like these are two different models. The fingerprints look very different, and the message that you're now sending isn't on the existing Ioniq. Can you break them out into two different cars?

PR #1774, opened 2020-06-24
DSU wiki link, and SDSU info.

ErichMoraga:
SDSU will allow OP longitudinal control, while maintaining stock Toyota AEB for safety.


pd0wm:
We will not make the claim that a smart DSU brings back AEB, since we don't actually verify or maintain that.

PR #1785, opened 2020-06-26
Support for NEOS 15

jyoung8607:
Requirements:

- [x] Incorporate NEOS 15 from commaai/eon-neos-builder#26
  - [x] Pointer for expected NEOS version
  - [x] Pointer for OTA updates
- [x] UI to take over framebuffer power control from NEOS
- [x] Fix issue with libSNPE linking of libsymphony-cpu
- [x] OTA in-place upgrading
- [x] OTA in-place downgrading
- [x] Fix Android "unused DT entry" linker complaints
- [x] Background updater, working in concert with OP background updater and overlay

Moved to separate project:

* Deeper power save mode using accelerometer for touch hinting, standalone project


launch_chffrplus.sh
```diff
@@ -6,6 +6,10 @@ export NUMEXPR_NUM_THREADS=1
 export OPENBLAS_NUM_THREADS=1
 export VECLIB_MAXIMUM_THREADS=1
 
+if [ -z "$NEOS_VERSION" ]; then
+  export NEOS_VERSION="15-RC2"
```
pd0wm, line 10: Keep 14 here, and set 15-rc2 for in launch.sh for internal use

launch_chffrplus.sh
```diff
@@ -64,49 +66,60 @@ function launch {
...
-      cp "$DIR/scripts/continue.sh" "/data/data/com.termux/files/continue.sh"
+  if [ "$(< /VERSION)" != "$NEOS_VERSION" ]; then
+    if [ -f "$BASEDIR/scripts/continue.sh" ]; then
+      cp "$BASEDIR/scripts/continue.sh" "/data/data/com.termux/files/continue.sh"
     fi
 
     if [ ! -f "$BASEDIR/prebuilt" ]; then
       echo "Clearing build products and resetting scons state prior to NEOS update"
-      cd $BASEDIR && scons --clean
+      git clean -xdf
```
pd0wm, line 100: Why no more scons clean? Git clean -xdf doesn't clean the submodules.

launch_chffrplus.sh
jyoung8607: Will do.

launch_chffrplus.sh
jyoung8607: This bit of script was about downgrades, not upgrades (#1677). I had streamlined it a bit since scons clean isn't effective in that case anyway, and it added several seconds of unusual black-screen UI between the bootup spinner and the NEOS update dialog. I took out the explicit database rm since git clean -xdf does it.

I haven't yet noticed an issue with scons needing help to handle upgrades correctly, even without cleanup. However, your point is well taken about submodules. Even though I think scons rebuilds everything anyway after the database zap, there's benefit to a complete clean. I'll do so after the database zap so it works in both upgrade and downgrade cases.

pd0wm: You can also add a `git submodule foreach --recursive git clean -xdf` to clean the submodules.

pd0wm:
opendbc is changed in this PR? I assume that's not intentional

launch_chffrplus.sh
jyoung8607: That seems preferable to the heavyweight scons clean then.

jyoung8607:
> opendbc is changed in this PR? I assume that's not intentional

Correct, it's just behind-the-times since my last rebase. I'll be rebasing again before I make the last anticipated RC this weekend.

PR #1788, opened 2020-06-26
tuning of Honda CR-V Hybrid lateral values

Martint1980:


pd0wm:
Can you add some explanation to this PR? How did you verify the shortened range? Why is the tuning different from the regular CRV?

And for a tuning PR we also need a few other people to verify it. They can just comment on this PR saying if it's better or not.

gregjhogan:
on non-hybrid cr-v 5th gen vehicles the max value corresponds to the wheel shake torque request sent by the radar that occurs during lane departure mitigation

Martint1980:
I tested the range of the standard 5th gen CR-V with the hybrid. It feels better in a curve. The warning comes now, when the wheel can not turn anymore. Before there was a gap between warning and doesn't turn anymore. It was sometimes a bit "dangerous" with the delayed warning, because it drifted already over the lines.
It seems it have the same EPS torque curves/values. @gregjhogan which can message I need to review for the lane departure migration? I could check that. 

The tuning is trial and error. I found a good solution to make tigher turns, but don't overshoot  so much.

gregjhogan:
@Martint1980 you can't judge when openpilot hits the max request value based on when the alert occurs vs. when the wheel stops turning

The road departure mitigation wheel shake is an oscillating min/max torque request output by the radar on 1:e4
image

PR #1789, opened 2020-06-26
some enhancements to get openpilot compiled with use_webcam=1

sjdavalle:
Platform: -ubuntu 16:04

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -116,6 +116,8 @@ env = Environment(
     "-Werror",
     "-Wno-deprecated-register",
     "-Wno-inconsistent-missing-override",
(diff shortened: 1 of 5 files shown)


SConstruct
```diff
@@ -17,6 +17,8 @@ if platform.system() == "Darwin":
 if arch == "aarch64" and not os.path.isdir("/system"):
   arch = "larch64"
 
+is_ubuntu = 'ubuntu' in os.getenv('DESKTOP_SESSION', 'unknown')
```
adeebshihadeh, line 20: This isn't a reliable way to detect the OS, returns `i3` (window manager) on my machine

PR #1792, opened 2020-06-27
Build openpilot on mac in CI

adeebshihadeh:
This is good to merge, but installing the dependencies takes forever. Need to implement dependency caching before merging.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -58,8 +58,10 @@ jobs:
         submodules: true
     - name: Install dependencies
       run: ./tools/mac_setup.sh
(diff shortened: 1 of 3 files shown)


kronner:
I'm also very interested in this!

Is there an instruction?

adeebshihadeh:
@kronner mac support isn’t a priority for us, this is simply so we know if we’re breaking the mac build. The new setup file ‘tools/mac_setup.sh’ should be a good starting point, but probably won’t be complete for running openpilot. We strongly welcome contributions to improve quality on mac (general support, documentation, setup, etc)!

PR #1794, opened 2020-06-28
CAR.COROLLA_TSS2 engine and ESP f/w

ErichMoraga:
@stevheng#6662's 2019 Australian Corolla HB  DongleID/route 0a9995ae54acb892|2020-06-28--07-24-06


ErichMoraga:
Confirmed working!

PR #1795, opened 2020-06-28
Small changes for dos compatibility

robbederks:


robbederks:
Merged, but github is being weird

PR #1808, opened 2020-07-02
Call update_offroad_layout_state in one place

deanlee:
- Call update_offroad_layout_state in one place
- reduce unnecessary offroadLayout messages.

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -65,35 +65,26 @@ static void set_awake(UIState *s, bool awake) {
 }
 
 static void update_offroad_layout_state(UIState *s) {
-  capnp::MallocMessageBuilder msg;
-  auto event = msg.initRoot<cereal::Event>();
-  event.setLogMonoTime(nanos_since_boot());
-  auto layout = event.initUiLayoutState();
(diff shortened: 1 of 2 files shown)


selfdrive/ui/ui.cc
```diff
@@ -64,69 +64,53 @@ static void set_awake(UIState *s, bool awake) {
...
-#endif
-}
-
-static void navigate_to_home(UIState *s) {
-#ifdef QCOM
-  if (s->started) {
-    s->active_app = cereal::UiLayoutState::App::NONE;
-  } else {
-    s->active_app = cereal::UiLayoutState::App::HOME;
+static void update_offroad_layout_state(UIState *s, int* timeout) {
```
adeebshihadeh, line 67: why not make the timeout a static variable inside the function?

adeebshihadeh:
Nice cleanup. This changes the behavior for PC, can you gate sending the layout packet on `QCOM`?

adeebshihadeh:
We really appreciate all the refactors and clean-ups. In the future, it's easiest to merge PRs like this where it's straightforward to test and immediately obvious if something's wrong.

PR #1812, opened 2020-07-03
ui: use struct for rear&front params

deanlee:


adeebshihadeh:
Not sure this front view stuff is relevant anymore, it accomplishes the same thing as driver view.

deanlee:
It looks like that driverview.py just send controlstate message to active front view in ui.

PR #1815, opened 2020-07-04
UI: const global variables inside namespace

deanlee:


adeebshihadeh:
What's the benefit here? It just seems more verbose.

deanlee:
try to avoiding confusion between the local and global variables.

adeebshihadeh:
I don't think that's really a problem, and we don't use this convention anywhere else.

PR #1823, opened 2020-07-06
Alert when modeld is lagging by more than a frame

adeebshihadeh:


selfdrive/modeld/models/driving.cc
```diff
@@ -290,13 +291,13 @@ void model_publish(PubMaster &pm, uint32_t vipc_frame_id, uint32_t frame_id,
 
   auto meta = framed.initMeta();
   fill_meta(meta, net_outputs.meta);
-  event.setValid(frame_id < vipc_frame_id + MAX_FRAME_AGE);
+  event.setValid((frame_id < vipc_frame_id + MAX_FRAME_AGE) && sm_alive_valid);
```
pd0wm, line 294: I would compute frame age only once, and store in a local variable. Same computation is done on line 257.

pd0wm:
LGTM

pd0wm (approved):

PR #1825, opened 2020-07-06
sending to master. interface support kfBP and kfV

birdman6450:
i accidentally PRed to release2 and i am i trying to send it to master

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -200,8 +200,10 @@ class CarInterface(CarInterfaceBase):
       ret.steerRatio = 13.9
       tire_stiffness_factor = 0.444  # not optimized yet
       ret.mass = 3060. * CV.LB_TO_KG + STD_CARGO_KG
-      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.6], [0.1]]
-      ret.lateralTuning.pid.kf = 0.00007818594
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
I closed your other two PRs as it should be one PR.

xps-genesis:
this is a big change. every selfdrive/car/xxx/interface.py should switch to List[] Kf. This change error will fault out every other car variant that doesn't provide kfBp and KfV values.

majorwolf:
> 
> 
> this is a big change. every selfdrive/car/xxx/interface.py should switch to List[] Kf. This change will fault out every other car variant that doesn't provide kfBp and KfV values.

Should someone put in the work and go through and change all the interface files to match?

xps-genesis:
I am not sure who decides that? what if you put in all the work and comma decides having Kf speed dependent doesn't make sense and reject he PR? I have Hyundai interface.py ready on my fork and I can copy paste and create a PR. but what about other brands? If someone from comma agrees to this change, some of us can contribute to this change.

pd0wm:
What is the benefit of this change?

It doesn't really make sense, kF is already scaled with the square of the speed (https://github.com/commaai/openpilot/blob/master/selfdrive/controls/lib/latcontrol_pid.py#L36).

adeebshihadeh:
Closing due to inactivity. Feel free to re-open.

PR #1830, opened 2020-07-06
Mazda BSM

Jafaral:
Plus a couple of new fingerprints


selfdrive/car/mazda/carstate.py
```diff
@@ -139,6 +142,8 @@ def get_can_parser(CP):
         ("SET_P", "CRZ_BTNS", 0),
         ("SET_M", "CRZ_BTNS", 0),
         ("CTR", "CRZ_BTNS", 0),
+        ("LEFT_BS1", "BSM", 0),
+        ("RIGHT_BS1", "BSM", 0),
```
adeebshihadeh, line 146: I see there's a bunch of BSM signals defined in the DBC, but it's not clear what they are. Can you add comment lines to the DBC?

PR #1833, opened 2020-07-06
Nissan: Increase steer threshold

avolmensky:
Occationally when going around a corner, a bump in the road would trigger the steer threshold, reducing the torque applied by LKAS and the car would run wide.  A value of 1.75 is roughly what the stock system uses for the `STEERING_PRESSED` signal.


pd0wm:
You can update `ref_commit` to `520cf8bbbc7e37d3a54ce8d443f52eb232758be5`

PR #1847, opened 2020-07-09
ui: conditionally check messages

deanlee:
- check ubloxGnss,health only when uilayout_sidebarcollapsed is false
- check driverState only when frontview is true

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -172,8 +172,7 @@ static void ui_init(UIState *s) {
                                     , "liveMapData"
 #endif
   });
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
I don’t really see the point of this. Eventually you’re just gonna end up with raw sockets wrapped in a different API called sub master.

deanlee:
how about this?: combine them into extra_sm

we don't need to waste resources to check and parse these  messages when onroading.

adeebshihadeh:
This still introduces 10 new lines and something called “extra_sm”. What’s the benefit of this extra complexity? What resources are we wasting and how much of them?

PR #1848, opened 2020-07-09
NEOS background download

jyoung8607:
Fetches any pending NEOS update in the background before informing the user it's time to switch to the upgraded version.


selfdrive/updated.py
```diff
@@ -126,6 +130,11 @@ def set_update_available_params(new_version=False):
     params.put("UpdateAvailable", "1")
 
 
+def set_neos_download_param(downloading=False):
```
adeebshihadeh, line 133: This doesn't need a function
```
Params().put("Offroad_NeosUpdate", "1" if downloading else "0")
```

selfdrive/updated.py
```diff
@@ -302,11 +311,35 @@ def attempt_update():
 
     finalize_from_ovfs_copy()
 
+    # If a NEOS update is required, download it in the background
+    with open("/VERSION", "r") as current_neos_file:
+      current_neos_version = current_neos_file.read().replace("\n", "")
+    required_neos_version = run(["bash", "-c", r"source launch_env.sh && echo -n $REQUIRED_NEOS_VERSION"], FINALIZED)
+    if current_neos_version != required_neos_version or DEBUG_FORCE_UPDATE:
+      print(f"Beginning background download for NEOS {required_neos_version}")
+      try:
```
adeebshihadeh, line 320: why not `os.path.isdir`?

adeebshihadeh:
What do you think of building the standalone binaries like `updater` in CI and pushing them out with master-ci instead of checking them into master?

installer/updater/updater.cc
```diff
@@ -301,11 +304,13 @@ struct Updater {
   void set_progress(std::string text) {
     std::lock_guard<std::mutex> guard(lock);
     progress_text = text;
+    printf("%s\n", text.c_str());
   }
 
   void set_error(std::string text) {
     std::lock_guard<std::mutex> guard(lock);
     error_text = text;
+    printf("%s\n", text.c_str());
```
adeebshihadeh, line 313: `std::cout`?

launch_chffrplus.sh
```diff
@@ -1,20 +1,9 @@
...
 if [ -z "$BASEDIR" ]; then
   BASEDIR="/data/openpilot"
 fi
-
-if [ -z "$PASSIVE" ]; then
-  export PASSIVE="1"
-fi
-
-STAGING_ROOT="/data/safe_staging"
+source "$BASEDIR/launch_env.sh"
```
adeebshihadeh, line 6: What's the point of breaking up the launch script? It seems like nothing else uses `launch_env.sh`

jyoung8607: Don't want to track the openpilot required NEOS version in two places, can't source `launch_chffrplus.sh` as it exists. Chatted with Willem about maybe moving OP version to a bare standalone file like the EON `/VERSION` but we settled on this.

selfdrive/updated.py
jyoung8607: I don't follow?

selfdrive/updated.py
jyoung8607: Done.

selfdrive/updated.py
```diff
@@ -302,11 +311,35 @@ def attempt_update():
...
+        shutil.rmtree("/data/neoupdate")
+      except FileNotFoundError:
+        pass
+      update_json = f'file:///{FINALIZED}/installer/updater/update.json'
+      while True:
+        set_neos_download_param(True)
+        run(NICE_LOW_PRIORITY + ["installer/updater/updater", "bgcache", update_json], FINALIZED)
+        set_neos_download_param(False)
+        if os.path.isfile("/data/neoupdate/cache_success_marker"):
```
adeebshihadeh, line 324: I'm not sure I understand the point of the cache success marker. Can't you just check if the update file exists then check the hash?

selfdrive/updated.py
adeebshihadeh:
```
if os.path.isdir(shutil.rmtree("/data/neosupdate")):
  shutil.rmtree("/data/neosupdate")
```

selfdrive/updated.py
jyoung8607: Diff minimization. The existing code rather freely mixes functional tasks and UI flows and updates in the same functions; I would have had to break it into pieces that can run headless and be called under varying circumstances. This method made maximal reuse of the existing tested paths.

adeebshihadeh: Seems like it would only take a couple more lines, no? You're already passing the update file to the updater which has the hash and check the actual update file instead of this cache success file?

selfdrive/updated.py
jyoung8607: Ah, didn't understand from the snippet github showed me of your question. Sure, change made.

jyoung8607:
> What do you think of building the standalone binaries like `updater` in CI and pushing them out with master-ci instead of checking them into master?

If we can do it, sure. I thought there were issues with cross-building and we didn't want public CI to build on EONs due to trust issues.

installer/updater/updater.cc
jyoung8607: When in Rome. There's a lot of other C-isms in updater.cc if we're going to go there.

selfdrive/updated.py
```diff
@@ -72,15 +76,14 @@ def graceful_shutdown(self, signum, frame):
 
   def update_now(self, signum, frame):
     cloudlog.info("caught SIGHUP, running update check immediately")
+    global DEBUG_FORCE_UPDATE
+    DEBUG_FORCE_UPDATE = True
```
jyoung8607, line 80: FIXME: testing-only configuration. Before merging into master, this line should be commented out, and commits 103a244 and 7f8cca6 should be reverted as well.

selfdrive/updated.py
jyoung8607: Per our discussion, I've moved things around to do without the cache semaphore file.

selfdrive/updated.py
```diff
@@ -49,6 +50,8 @@
 NICE_LOW_PRIORITY = ["nice", "-n", "19"]
 SHORT = os.getenv("SHORT") is not None
```
pd0wm, line 51: SHORT is no longer used?

selfdrive/updated.py
```diff
@@ -303,11 +305,36 @@ def attempt_update():
 
     finalize_from_ovfs_copy()
 
+    # If a NEOS update is required, download it in the background
+    with open("/VERSION", "r") as current_neos_file:
+      current_neos_version = current_neos_file.read().replace("\n", "")
```
pd0wm, line 310: use strip()

selfdrive/updated.py
```diff
@@ -303,11 +305,36 @@ def attempt_update():
...
+    required_neos_version = run(["bash", "-c", 
+               r"unset REQUIRED_NEOS_VERSION && source launch_env.sh && echo -n $REQUIRED_NEOS_VERSION"],
+               FINALIZED)
+    print(f"NEOS version update check: {current_neos_version} current, {required_neos_version} in update")
```
pd0wm, line 314: replace all these prints by cloudlogs

selfdrive/updated.py
```diff
@@ -303,11 +305,36 @@ def attempt_update():
...
+      while True:
+        Params().put("Offroad_NeosUpdate", "1")
+        try:
+          run(NICE_LOW_PRIORITY + ["installer/updater/updater", "bgcache", update_json], FINALIZED)
+          Params().put("Offroad_NeosUpdate", "0")
+          print("NEOS background download successful!")
+          break
+        except subprocess.CalledProcessError:
+          print("NEOS background download failed, will retry at next wait interval")
+          Params().put("Offroad_NeosUpdate", "0")
```
pd0wm, line 327: why clear the param? It's still trying to download

selfdrive/updated.py
```diff
@@ -303,11 +305,36 @@ def attempt_update():
...
+    with open("/VERSION", "r") as current_neos_file:
+      current_neos_version = current_neos_file.read().replace("\n", "")
+    required_neos_version = run(["bash", "-c", 
+               r"unset REQUIRED_NEOS_VERSION && source launch_env.sh && echo -n $REQUIRED_NEOS_VERSION"],
+               FINALIZED)
+    print(f"NEOS version update check: {current_neos_version} current, {required_neos_version} in update")
+    if current_neos_version != required_neos_version or DEBUG_FORCE_UPDATE:
+      print(f"Beginning background download for NEOS {required_neos_version}")
+      update_json = f'file:///{FINALIZED}/installer/updater/update.json'
+      while True:
```
pd0wm, line 318: I would add a maximum number of attempts here

installer/updater/updater.cc
```diff
@@ -37,6 +38,8 @@
 #define MANIFEST_URL_EON_LOCAL "http://192.168.5.1:8000/neosupdate/update.local.json"
 #define MANIFEST_URL_EON "https://github.com/commaai/eon-neos/raw/master/update.json"
 const char *manifest_url = MANIFEST_URL_EON;
+bool background_cache = false;
```
pd0wm, line 41: why global and used as argument?

selfdrive/updated.py
```diff
@@ -303,11 +305,36 @@ def attempt_update():
...
+        Params().put("Offroad_NeosUpdate", "1")
+        try:
+          run(NICE_LOW_PRIORITY + ["installer/updater/updater", "bgcache", update_json], FINALIZED)
+          Params().put("Offroad_NeosUpdate", "0")
+          print("NEOS background download successful!")
+          break
+        except subprocess.CalledProcessError:
```
pd0wm, line 325: What happens if the file is corrupted? Running the updater again is not going to fix it. The corrupted file will only get cleared on reboot.

selfdrive/updated.py
jyoung8607: It's from way back, before the version of the updater I built in December 2019 with the SIGHUP handler. I'll garbage-collect it. Short-cycling the updater can potentially cause strange behavior I don't want to bring it back.

selfdrive/updated.py
jyoung8607: Ok.

selfdrive/updated.py
jyoung8607: Ok.

selfdrive/updated.py
jyoung8607: UX judgment call, could go either way. For not clearing we'll get "why does it show I'm downloading when I'm not even connected to the Internet", for clearing we'll get "why did it say I was getting an update and then disappear". Depends on the wording we want the APK to show the user.

selfdrive/updated.py
jyoung8607: If the user leaves Internet coverage, the maximum number of attempts at 10 minutes per would have to be really high to avoid unnecessarily restarting the entire process.

installer/updater/updater.cc
jyoung8607: `run_stages()` is called from two places; only one of them varies that argument based on `background_cache`. The meaning of that argument is distinct from `background_cache`, and `run_stages()` needs to know its actual value either way it's called.

selfdrive/updated.py
jyoung8607: The NEOS updater will unlink() a file if it fails the signature check:

https://github.com/commaai/openpilot/blob/c43df81214d34d90b318defbd7f43e240b0c8f45/installer/updater/updater.cc#L332-L339

Historically the user gets to reboot and try again at that point. In the background cache mode, the NEOS updater still bails out, but the openpilot updater will call it again later and get a fresh try at it.

pd0wm: Ah, missed the `unlink`

selfdrive/updated.py
pd0wm: Think we still want to bail out after like a day / 144 attempts in case we mess something up and ship a broken manifest or something.

installer/updater/updater.cc
pd0wm: I would still pass it around using arguments instead of using a global.

installer/updater/updater.cc
```diff
@@ -192,7 +195,7 @@ struct Updater {
 
   CURL *curl = NULL;
 
-  Updater() {
+  void ui_init() {
```
adeebshihadeh, line 198: This is confusing. We have a `ui_init` and a `start_ui`

selfdrive/updated.py
```diff
@@ -73,15 +75,14 @@ def graceful_shutdown(self, signum, frame):
...
     self.ready_event.set()
 
 
 def wait_between_updates(ready_event):
   ready_event.clear()
-  if SHORT:
-    ready_event.wait(timeout=10)
-  else:
-    ready_event.wait(timeout=60 * 10)
+  ready_event.wait(timeout=WAIT_BETWEEN_ATTEMPTS)
```
adeebshihadeh, line 85: this function is only called in one place, don't think it makes sense to make this a function

PR #1849, opened 2020-07-10
Hyundai: use SCC12 for stock ADAS signals on cars that don't have FCA11

### LINKED ISSUE #1813: Some older HKG vehicles do not have FCA11msg. CAN Error: Check Connections Alert.
Tested on 2016 Hyundai Genesis. Due to FCA11 check in Hyundai\carstate.py, the cp.can_valid fails. FCA11 check needs to be vehicle specific.

adeebshihadeh:
resolves #1813


xps-genesis (approved):
looks good

PR #1853, opened 2020-07-10
Make updated more resilient

jyoung8607:
Sleep 30 seconds before kicking off the first `updated` run cycle. Actually makes the first check faster, since much of the time we were "winning" a race with IsOffroad not yet being set, resulting in a 10 minute wait before the first successful check.

On bootup, after swapping in a finalized overlay upgrade of openpilot, do a `git reset --hard` of all files. This is a mitigation (not a fix) of the intermittent filesystem damage we've been seeing, wherein random file contents are replaced with the name of a symlink target somewhere else in the openpilot tree. It won't save us from corrupted files in `.git`, but if that's shown to be a problem, we can explore some repo integrity checks before the overlay swap-in.

### DIFF
--- a/launch_chffrplus.sh
+++ b/launch_chffrplus.sh
@@ -44,7 +44,7 @@ function launch {
           cd $BASEDIR
 
           # Partial mitigation for symlink-related filesystem corruption
-          # Make sure all files match the repo versions after update
+          # Ensure all files match the repo versions after update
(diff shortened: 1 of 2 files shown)


qadmus:
I saw a `cp -r` become a symlink in /data/, if that helps reproduce the intermittent filesystem damage.

PR #1856, opened 2020-07-10
boardd: Add new functions usb_read, usb_write

deanlee:
it seems that there are many inappropriate flag 0xc0 used for write operation. although it does not affect the result, their value should be 0x40

Add functions usb_read and usb_write to fix this problem and make the code clearer.


pd0wm:
Indeed, `bmRequestType` is technically wrong, but not checked in the panda.

PR #1865, opened 2020-07-11
Nissan: Tweaking steeringPressed/LKAS_MAX_TORQUE

avolmensky:
I was occationally getting false positives for steeringPressed when hitting bumps in the road, this wouldn't normally be a problem on the freeway, but if going around a tighter corner it could cause OP to scale LKAS torque to zero, causing the car to run wide. I have added a simple moving average filter to fix the false positives. 

I have changed STEER_THRESHOLD back down to 1.0 as the filtering has stopped the false positives, and 1.75 felt a bit high for smaller steering corrections.

I have also tweaked the scaling of LKAS_MAX_TORQUE, I think it is dangerous to scale LKAS_MAX_TORQUE to zero (especially if the driver doesn't have their hands on the steering wheel), so I now scale it down to 50%, also tweaked the scaling to make it feel better.

A quick pyplot showing a bump in the road, the red line is the filtered driver torque value.
image


pd0wm:
Can you make sure the panda repo is not touched?

PR #1869, opened 2020-07-12
Driver view refactor

adeebshihadeh:
Seems like there's some thresholds for face detection in the UI code. The driver view functionality should really be a part of the driver monitoring daemon, but for the time being, this reduces the foot print of driver view across the codebase.

Edit: The driver view functionality has been moved to dmonitoringd, and the face detection thresholds have been removed from UI.


adeebshihadeh:
@ZwX1616 the `rhdChecked` field doesn't seem to be used anywhere. do you check for it during training or somewhere else? if not, I'd like to deprecate it

PR #1872, opened 2020-07-12
Chevy Volt PI tuning, 0.7.7 localizer live params

hewers:
kp 0.12 and ki 0.05 have legacy with kegman custom fork, adding a 5 m/s breakpoint with 0.07 kp reduces jerky oversteering-and-correction on high curvature low speed side roads.

Steer ratio and tire stiffness are from 0.7.7 master-ci this weekend, with localizer live params, one block from the exit ramp after 20 highway miles. If that is invalid, I can repeat on the road.


pd0wm:
Our policy on tuning PRs is that a few people using the new tuning must comment on this PR saying if it's better or not.

qadmus:
Got two good reviews and one bad review - will look into the bad review. Closing pull request because I'm favoring a INDI tuning now that isn't yet complete.

PR #1877, opened 2020-07-13
Revert modeld vipc conflate behavior and alert on dropped frame percent

adeebshihadeh:
new cereal field
```
struct ModelData {
  frameDropPerc @13 :Float32;
```


pd0wm:
Made some small changes, but think the PR is good now.

I assume you did some temperature testing to verify the percentages. Did you use similar percentages in your tests @HaraldSchafer ?

PR #1879, opened 2020-07-15
Add 2020 Insight FW

csouers:
This car is the same as the 2019, just with newer firmware IDs.

Route ID: d952837313c3e269|2020-07-13--18-53-09
User is using OP 0.7.6.1 due to Camera Refocus alerts from a recent commit of master.

This change will break Insight users still on giraffe due to not being able to query all now necessary (i.e. Ecu.eps) firmware IDs from the car. <Thanks to @theantihero for that info.


csouers:
Padding due having to add "-20" to the Model line.

csouers:
> Why is the diff on the readme so large?

My bad. Fixed.

PR #1881, opened 2020-07-15
Minor change for better readability.

ghost:
This may just be my OCD but I wasn't to big of a fan of the { on the same line as the else. Very minor I know.


adeebshihadeh:
That's the convention used in all the other C/C++ code.

PR #1894, opened 2020-07-19
Disable raw upload while onroad

adeebshihadeh:
Saves about 10% of a core if a user is connected to wifi


pd0wm:
Think it's very unlikely to be onroad and in wifi range, but it's a simple change.

roxasthenobody98:
I feel like this should be a UI toggle rather than off. Some people might have it connected to in vehicle hotspots (OnStar, Uconnect, SYNC Connect, Entune)

PR #1912, opened 2020-07-23
add Kia Sorento to legacy safety check

xps-genesis:
confirmed kia sorento to use legacy safety checks


adeebshihadeh:
I'd like to get a test route before merging this in. The new test in #1903 would catch this if we had a test route.

maueddu03:
5875672fc1d4bf57|2020-07-23--21-33-28--2

adeebshihadeh:
Thanks for the route @maueddu03, added in 4bc90c578aa1e229c18b62aaec6668dee9b49a84

PR #1915, opened 2020-07-23
Add Subaru XV 2018 and Subaru Impreza Premium 2019 FPv1

martinl:
Adding two new Crosstrek/Impreza fingerprints


martinl:
2019 Impreza Premium will require FPv2 because it overlaps with Ascent, closing

PR #1916, opened 2020-07-24
Add argument to ui executable that forces the drawing of the UI

sshane:
This allows developers to start the UI by typing `./ui debug` so you can see your design without plugging the C2/EON into your car and starting it up. Everything works as intended when starting ui without debug argument.

If there's a better way to draw the UI in paint.cc without setting all these variables every loop, let me know. It seems to work fine though as this PR stands.

### DIFF
--- a/selfdrive/ui/ui
+++ b/selfdrive/ui/ui
@@ -1,4 +1,3 @@
 #!/bin/sh
 export LD_LIBRARY_PATH="/system/lib64:$LD_LIBRARY_PATH"
-exec ./_ui
-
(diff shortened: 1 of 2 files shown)


deanlee:
good idea but I think there might be a better way to do this than changing the ui code.for example,we may be able to  use the same pattern as driverview.py to achieve this.

ShaneSmiskol:
Oh that's a good idea too, I didn't know driverview was a thing. I'll start on something like uiview.py and see if I can get it working just like driverview

PR #1917, opened 2020-07-24
refactor driverview.py to display UI/front camera with -f

sshane:
- [x] update `TOTAL_SCONS_NODES` with latest
- [x] catches SIGINT (ctrl-c) as well

How to run it:
- `cd /data/openpilot/selfdrive/monitoring; python driverview.py --front`
- or use the alias:
  - `cd /data/openpilot/selfdrive/ui; ./ui_debug`


ShaneSmiskol:
@deanlee how does this look?

pd0wm:
Why not add a `--front` flag to driverview.py? No need to duplicate all the code.

ShaneSmiskol:
i'll move the driverview file to selfdrive/debug then and unify both the files tomorrow!

pd0wm:
You have to leave the file in it's place, since it's started by offroad when you press the driver view button.

ShaneSmiskol:
One of the issues with following the comment to use manager's start and stop functions is that it takes about 5 seconds for it to load since manager runs scons and a few other lines of code before it starts. Is this a concern and should I revert it back to not use manager?

ShaneSmiskol:
Think it should be ready again to look at @pd0wm

ShaneSmiskol:
it crashes when visiting the driver preview in the offroad UI, checking it out

ShaneSmiskol:
fixed, made the main args optional. also added an alias in selfdrive/ui since it's a shorter directory and command to type!

adeebshihadeh:
This is a lot of extra code (net +37 lines) for something that should be a few lines. I think it makes more sense to make a standalone script you can keep in your fork.

ShaneSmiskol:
Well let me see if I can clean it up enough to be able to get merged, it would be really handy to put on the development or custom forks wiki so new users can test their UIs easily. Keep in mind that 6 lines of the additions is the `send_thermal_packet` function, so it's at least going to be 10 or 20 or so. Need the logic to only start what's needed and only kill what's needed as well

ShaneSmiskol:
This might be unrelated to this specific PR, but we could even remove some of the existing lines in driverview.py related to the todo comment that was there by moving all top level code from manager.py to `if __name__ == "__main__"`, that way we can import the start and kill functions from manager without taking any time at all, and it would be a good refactor overall in my opinion. Either that, or move the start and kill and process lists to a manager helper function

adeebshihadeh:
I'm refactoring driver view in #1869. When I have some time, I'm planning on just integrating it's functionality into dmonitoringd.

ShaneSmiskol:
So then would a simple uiview file be accepted in selfdrive/ui? it would definitely be much smaller than driverview here

adeebshihadeh:
Closing this because driverview is going away in the refactor. Feel free to open a new PR with a short, minimal script in selfdrive/debug.

PR #1919, opened 2020-07-24
Panda abstraction layer in boardd

pd0wm:


gregjhogan:
Too bad this doesn't use libusb if it ends up extremely similar
https://github.com/commaai/panda/blob/master/drivers/windows/panda_shared/panda.cpp

selfdrive/boardd/panda.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include "cereal/gen/cpp/car.capnp.h"
+#include "cereal/gen/cpp/log.capnp.h"
+
+// double the FIFO size
+#define RECV_SIZE (0x1000)
+#define TIMEOUT 0
+
+// copied from panda/board/main.c
+struct __attribute__((packed)) health_t {
```
adeebshihadeh, line 17: maybe we should make this a standalone header panda and include it here?

adeebshihadeh: or we can wait until we move this abstraction into panda

pd0wm: Let's do it when we move it to panda. I also want to create an enum for all the different commands instead of the hardcoded hex values in main.c and panda.cc

selfdrive/boardd/panda.cc
```diff
@@ -0,0 +1,263 @@
...
+
+  err = libusb_set_configuration(dev_handle, 1);
+  if (err != 0) { goto fail; }
+
+  err = libusb_claim_interface(dev_handle, 0);
+  if (err != 0) { goto fail; }
+
+  hw_type = get_hw_type();
+  is_pigeon =
+    (hw_type == cereal::HealthData::HwType::GREY_PANDA) ||
```
adeebshihadeh, line 36: think it makes sense to add a USB command for things like this, but in this instance, everything should be a pigeon now.

selfdrive/boardd/boardd.cc
```diff
@@ -644,48 +323,165 @@ void *can_recv_thread(void *crap) {
...
+
+    bool ignition = ((health.ignition_line != 0) || (health.ignition_can != 0));
+
+    if (ignition) {
+      no_ignition_cnt = 0;
+    } else {
+      no_ignition_cnt += 1;
+    }
+
+  #ifndef __x86_64__
```
adeebshihadeh, line 371:
```suggestion
  #ifdef QCOM
```

This shouldn't apply to QCOM2 right?

selfdrive/boardd/panda.cc
pd0wm: You mean change the logic? So if not WHITE, PEDAL, UNKOWN, it's pigeon?

selfdrive/boardd/boardd.cc
pd0wm: I think the power modes don't apply, but the safety model when ignition is off does apply. I'll add two conditions

selfdrive/boardd/panda.h
adeebshihadeh: We can do this later, but ideally there's a USB command to query board features. there's lists like this in a few different places in openpilot that should just be part of a panda abstraction

selfdrive/boardd/panda.cc
adeebshihadeh: We can do this later, but ideally there's a USB command to query board features. there's lists like this in a few different places in openpilot that should just be part of a panda abstraction

selfdrive/boardd/panda.cc
```diff
@@ -0,0 +1,263 @@
...
+    (hw_type == cereal::HealthData::HwType::BLACK_PANDA) ||
+    (hw_type == cereal::HealthData::HwType::UNO);
+  has_rtc = (hw_type == cereal::HealthData::HwType::UNO);
+
+  return;
+
+ fail:
```
adeebshihadeh, line 43: minor: extra space

selfdrive/boardd/panda.cc
```diff
@@ -0,0 +1,263 @@
...
+    libusb_release_interface(dev_handle, 0);
+    libusb_close(dev_handle);
+  }
+
+  if (ctx) {
+    libusb_exit(ctx);
+  }
+
+  throw std::runtime_error("Error connecting to panda");
+}
```
adeebshihadeh, line 54: move this into a function?

selfdrive/boardd/panda.cc
```diff
@@ -0,0 +1,263 @@
...
+
+ fail:
+  if (dev_handle){
+    libusb_release_interface(dev_handle, 0);
+    libusb_close(dev_handle);
+  }
+
+  if (ctx) {
+    libusb_exit(ctx);
+  }
```
adeebshihadeh, line 51: move this into a function?

selfdrive/boardd/boardd.cc
```diff
@@ -884,60 +650,24 @@ int main() {
...
   }
 
-  if (getenv("BOARDD_LOOPBACK")){
-    loopback_can = true;
-  }
+  std::thread health_thread(can_health_thread);
 
-  err = pthread_mutex_init(&usb_lock, NULL);
-  assert(err == 0);
+  while (!do_exit){
```
adeebshihadeh, line 655: Why the while loop? We do the version check in pandad, what if openpilot starts -> pandad upates panda -> pandad launches boardd -> panda is disconnected -> different panda with old firmware is connected?

selfdrive/boardd/boardd.cc
```diff
@@ -148,175 +112,94 @@ void *safety_setter_thread(void *s) {
...
+  LOGW("setting safety model: %d with param %d", (int)safety_model, safety_param);
 
-  usb_write(dev_handle, 0xdc, safety_model, safety_param);
+  panda->set_safety_model(safety_model, safety_param);
 
-  pthread_mutex_unlock(&usb_lock);
-
-  return NULL;
+  safety_setter_thread_running = false;
+  return;
```
adeebshihadeh, line 123: this has no effect

selfdrive/boardd/boardd.cc
```diff
@@ -148,175 +112,94 @@ void *safety_setter_thread(void *s) {
...
-  err = libusb_set_configuration(dev_handle, 1);
-  if (err != 0) { goto fail; }
-
-  err = libusb_claim_interface(dev_handle, 0);
-  if (err != 0) { goto fail; }
+  try {
+    assert(panda == NULL);
+    panda = new Panda();
+  } catch (std::exception &e) {
+    return false;
```
adeebshihadeh, line 132: seems like this can get into an infinity loop if panda isn't null for some reason?

selfdrive/boardd/boardd.cc
```diff
@@ -351,205 +234,22 @@ void can_recv(PubMaster &pm) {
...
-  }
-  // send to health
-  pm.send("health", msg);
-
-  // send heartbeat back to panda
-  usb_write(dev_handle, 0xf3, 1, 0, &usb_lock);
-}
-
-
 void can_send(cereal::Event::Reader &event) {
```
adeebshihadeh, line 237: can send and recv should probably be in the panda class

selfdrive/boardd/boardd.cc
```diff
@@ -644,48 +323,165 @@ void *can_recv_thread(void *crap) {
...
+      struct tm sys_time = get_time();
+      if (time_valid(sys_time)){
+        panda->set_rtc(sys_time);
+      }
+    }
+
+    ignition_last = ignition;
+    uint16_t fan_speed_rpm = panda->get_fan_speed();
+
+    // set fields
```
adeebshihadeh, line 426: build health packet in panda class like the python library?

adeebshihadeh: actually not sure about this once since it'll create a dependency on cereal. python lib gets away with it because pycapnp can take a python dict

pd0wm: there is already a dependency on cereal. But the health packet actually contains some other logic as well. It needs the ignition spoofing and the fan speed needs to be put in.

selfdrive/boardd/boardd.cc
pd0wm: what do you mean? The assert won't throw an exception. It's just to verify that the previous connection was closed correctly.

selfdrive/boardd/boardd.cc
pd0wm: This is what we've always done. I agree that maybe just exiting and having pandad restart boardd might be more logical. I'll leave behavior the same as before the refactor, but we can revisit this later.

adeebshihadeh: Once this is a more complete panda library, we should be able to move all the pandad functionality into boardd.

PR #1921, opened 2020-07-24
Block LCA if trailer is connected to vehicle. 

roxasthenobody98:
Referencing cereal#65

LCA should be blocked if a trailer is connected. LCA could potentially cause the trailer to sway if the change is too abrupt, and the driver may lose control of the vehicle in that instance. The driver should manually make the lane change. 

Not all vehicles will have a can message for the 4/7 pin Trailer connector, but those that do should automatically block LCA when connected. 

Code fails test due to unmerged cereal pr.

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -181,7 +181,7 @@ class Controls:
         self.events.add(EventName.calibrationInvalid)
 
     # Handle lane change
-    if not CS.trailerConnected:
+    if not CS.trailerConnected: 
       if self.sm['pathPlan'].laneChangeState == LaneChangeState.preLaneChange:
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
This is a fairly unusual use case. Can't you just disable lane changes via the settings toggle?

roxasthenobody98:
I mean, yes. But if we have the ability to do so automatically, we should.

roxasthenobody98:
SUV's like the Highlander, Pilot, Ascent, Palisade, etc should have a can message that would be for trailer connected.

adeebshihadeh:
Closing as this is a fairly unusual use case, but it sounds like a nice feature for a fork!

PR #1923, opened 2020-07-25
Changing to legacy safety mode for Stinger

tabdallah:
Stinger faults out with new safety code introduced in 0.7.7


adeebshihadeh:
Can you include a test route in test_car_models.py? We have tests that would have caught this is we had a test route.

adeebshihadeh:
Thanks for the PR, made the changes in 360cd7d8a86460fcc339ead89c0c6a439417eb66 along with the added test route.

PR #1926, opened 2020-07-26
boardd: use std::atomic for safety_setter_thread_initialized

deanlee:
the read of safety_setter_thread_initialized is not protected by mutex.use std::atomic to fix this issue

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -10,6 +10,7 @@
 #include <sys/types.h>
 #include <sys/time.h>
 #include <sys/resource.h>
+#include <atomic>
 
 #include <assert.h>
 #include <pthread.h>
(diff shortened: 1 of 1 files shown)


pd0wm:
I don't think the mutex was for the race on the variable itself. Now the variable can be set to false, kicking of another instance of the thread. By doing it in the mutex the other thread can't do anything yet.

This is al very messy. I'm working on a refactor that can hopefully clean up boardd a lot.

jameszow2:
Hello Everyone my name is James Zow, I come from China. We have a djl community. Are you interested in joining us? We want to do an open source project




| |
jameszow
|
|
邮箱：jameszow@163.com
|

Signature is customized by Netease Mail Master

On 07/27/2020 17:33, Willem Melching wrote:

I don't think the mutex was for the race on the variable itself. Now the variable can be set to false, kicking of another instance of the thread. By doing it in the mutex the other thread can't do anything yet.

This is al very messy. I'm working on a refactor that can hopefully clean up boardd a lot.

PR #1932, opened 2020-07-27
Custom message builder

deanlee:
use a simple custom MessageBuilder from  https://github.com/commaai/cereal/pull/72


adeebshihadeh:
I couldn't push to your branch for some reason, so I pushed a new branch based off this one. #2161

PR #1935, opened 2020-07-27
SubMaster refactor

adeebshihadeh:


adeebshihadeh:
Doesn't save much CPU, so not worth it, otherwise we'll have to maintain a separate SubMaster that only runs in the tests.

PR #1937, opened 2020-07-27
Add a minimal debugging tool to show the UI while device not in car

sshane:
From the previous PR (#1917), minimizes that


adeebshihadeh:
Let's remove ui_debug. Can you make this smaller? This should be like 10 lines.

selfdrive/debug/uiview.py
```diff
@@ -0,0 +1,41 @@
...
+import cereal.messaging as messaging
+from common.basedir import BASEDIR
+
+
+def send_packets(pm, idx=0):
+  while True:
+    dat = messaging.new_message('controlsState')
+    dat.controlsState.rearViewCam = False
+    pm.send('controlsState', dat)
+    if idx % 50 == 0:  # thermal runs at 2 hz
```
adeebshihadeh, line 16: For the purposes of this tool, I don't think the freq matters too much, but these are defined in `common.realtime`.

ShaneSmiskol: Cool, didn't think it did either but wanted to be correct more than anything

ShaneSmiskol:
I'm really not sure how you want it to be 10 lines, doesn't each message sent require at least three lines? I could use semi-colons and combine multiple lines into one haha 😬

selfdrive/debug/uiview.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env python3
+import time
+import subprocess
+import cereal.messaging as messaging
+from common.basedir import BASEDIR
+
+pm = messaging.PubMaster(['controlsState', 'thermal'])
+subprocess.Popen(BASEDIR + "/selfdrive/camerad/camerad", cwd=BASEDIR + "/selfdrive/camerad")
+subprocess.Popen(BASEDIR + "/selfdrive/ui/ui", cwd=BASEDIR + "/selfdrive/ui")
```
adeebshihadeh, line 8: use manager helper functions

adeebshihadeh: and explicitly kill them on exit

ShaneSmiskol: The only problem with that is importing anything from manager calls scons which takes upwards of 15 to 20 seconds before you're allowed to see the UI. I tried that with the driverview PR but that wouldn't be good since it's supposed to be callable in the UI

ShaneSmiskol: I'd be happy to open a PR to move the scons code (from `# Run scons`) into the main() function so that it's possible to import manager helper functions and dicts/lists without a large delay though, just let me know.

adeebshihadeh: It doesn't really matter since this is a debug tool. Manager building at import time also slows down some of our tests, I'll refactor it soon.

selfdrive/debug/uiview.py
```diff
@@ -0,0 +1,20 @@
...
+  dat = messaging.new_message('controlsState')
+  dat.controlsState.rearViewCam = False
+  pm.send('controlsState', dat)
+  dat = messaging.new_message('thermal')
+  dat.thermal.started = True
+  pm.send('thermal', dat)
+  time.sleep(1 / 100.)
+
+proc_cam.send_signal(signal.SIGINT)
+proc_ui.send_signal(signal.SIGINT)
```
adeebshihadeh, line 20: This will never run since there's no way to break out of your while loop. I'd suggest wrapping the while loop in a try block and catching `KeyboardInterrupt`

selfdrive/debug/uiview.py
ShaneSmiskol: Ah okay, good to hear. I'll use those then. My reasoning was it can be annoying when all someone wants to do is see the UI, especially if you're actively adding UI elements and you just want to pull, compile the one file and see the updated UI quickly

selfdrive/debug/uiview.py
ShaneSmiskol: Thanks for the catch, I thought it would just exit out of the loop and not the whole process, but I was forgetting fundamental interrupts

ShaneSmiskol:
At the expense of immediate readability, it's at 15 lines with manager's functions and a few list comprehensions (I keep forgetting they exist for some reason)

adeebshihadeh:
I didn't literally mean 10 lines, so if you want to make it readable and concise, I'll hold off on merging.

ShaneSmiskol:
Okay, what do you think about that last commit? didn't realize you could just re-use messages, so i think having fewer stuff in the loop makes it read easier. Yeah I actually thought you meant ~10 lines haha. it should be a good balance now

adeebshihadeh:
I would've skipped the list comprehension for the start/kill, but otherwise looks good.

PR #1942, opened 2020-07-28
Process replay test fixed fingerprint support

martinl:
This PR adds fixed fingerprint support to process replay tests, similar to test_car_models.py
It's mostly useful for testing car models which have been added to IGNORED_FINGERPRINTS  or only have FPv2 fingerprint

### DIFF
--- a/selfdrive/test/process_replay/compare_logs.py
+++ b/selfdrive/test/process_replay/compare_logs.py
@@ -3,6 +3,7 @@ import bz2
 import os
 import sys
 import numbers
+import capnp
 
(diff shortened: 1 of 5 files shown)


adeebshihadeh:
This is a much bigger change than necessary. The `fingerprint` function in process replay has all of the log messages and should handle this.

martinl:
ok, I will simplify

adeebshihadeh:
Closing this for now, feel free to re-open or make a new PR after simplifying this.

PR #1947, opened 2020-07-30
modeld: read frame_id if sm.update(0)>0

deanlee:


selfdrive/modeld/modeld.cc
```diff
@@ -184,6 +184,7 @@ int main(int argc, char **argv) {
     uint32_t last_vipc_frame_id = 0;
     double last = 0;
     int desire = -1;
+    uint32_t frame_id = 0;
```
adeebshihadeh, line 187: can you move this to the same line as last_vipc_frame_id?

PR #1948, opened 2020-07-30
add function cl_get_device_id

deanlee:


pd0wm:
Can you add the error handling for the case where no platform is found (like in modeld.cc)? Now it returns an uninitialized `device_id`

selfdrive/common/clutil.c
```diff
@@ -36,6 +36,39 @@ void clu_init(void) {
...
+    cl_uint num_devices;
+    err = clGetDeviceIDs(platform_ids[i], device_type, 0, NULL, &num_devices);
+    if (err != 0 || !num_devices) {
+      continue;
+    }
+    // Get first device
+    err = clGetDeviceIDs(platform_ids[i], device_type, 1, &device_id, NULL);
+    assert(err == 0);
+    cl_print_info(platform_ids[i], device_id);
+    printf("\n");
```
pd0wm, line 64: Why newline here and not in line 53?

PR #1949, opened 2020-07-30
Optimize capnp

adeebshihadeh:
Follow [capnp best practices](https://capnproto.org/cxx.html#tips-and-best-practices) in hopes of saving a bit of CPU in each process.


deanlee:
x.y is 3x-4x faster than reader.x.y in my tests

adeebshihadeh:
@deanlee do you see any other best practices we’re currently violating? pycapnp seems to make up a decent amount of the CPU usage in our python processes

adeebshihadeh:
Not merging as-is. This doesn't have a significant impact on performance, and it's less readable.

deanlee:
maybe we should find or create a capnp's compiler plugin to compile  schemas to cython.

adeebshihadeh:
This still makes very little difference. I think the way to go is optimizing pycapnp for cheap encoding (currently it makes a few copies) or writing something in Sub/PubMaster to serialize capnp bytes to python dicts for cheap read/write, because currently we're paying for read/write and serialization.

PR #1951, opened 2020-07-30
loggerd refactor

deanlee:
- rewrite LoggerState&LoggerHandle in C++
- using shared_ptr instead of handles array.


adeebshihadeh:
I’m hesitant to consider this PR. We don’t have a lot of testing around loggerd right now, and it’s a critical part of openpilot that’s currently working fine. Does it save any CPU? If the only benefit is the code cleanup, maybe we leave it open until we get some time to write loggerd tests?

deanlee:
Yes, that would be better.

Issue #1876, opened 2020-07-13
openpilot can enable at startup while thermal status too high


Hubblesphere:
**Description of bug**
At initial on-road from startup or after cooling from thermal on-road protection open pilot can enable for first ~25-30 seconds after going on road. (tested in master-ci) It will then alert and disable after 20-30 seconds of being engaged. 

**How to reproduce or log data**

Enable openpilot within ~25-30 seconds of going on-road while temps in yellow or above.

**Expected behavior**

openpilot should not allow enable

**Device/Version information (please complete the following information):**
 - Device: Comma Two
 - Dongle ID: f91c796ac0bf5eb6
 - Route: f91c796ac0bf5eb6|2020-07-13--12-02-21--0
 - Version: commit 7ac0040
 - Car make/model: 2019 Toyota Corolla Hatchback

**Additional context**

Alternate Route example: f91c796ac0bf5eb6|2020-07-08--12-08-59--0

techasit14607:
Fix bug.

TK211X:
Could we also close the switch so that vehicle can be stock of temp is too high. Currently it lights up the dash and is very annoying.

adeebshihadeh:
This should be fixed now, especially with the recent thermal improvements due to the CPU usage reduction and downclock.

PR #1952, opened 2020-07-30
Improved longetivity and fixed issue #1876

nkitan:
changing threshold values led to reduced hot-times
resolves issue #1876

### DIFF
--- a/selfdrive/thermald/thermald.py
+++ b/selfdrive/thermald/thermald.py
@@ -273,10 +273,10 @@ def thermald_thread():
     # If device is offroad we want to cool down before going onroad
     # since going onroad increases load and can make temps go over 107
     # We only do this if there is a relay that prevents the car from faulting
-    if max_cpu_temp > 107. or bat_temp >= 63. or (has_relay and (started_ts is None) and max_cpu_temp > 70.0):
+    if max_cpu_temp > 107.0 or bat_temp > 62.0 or (has_relay and (started_ts is None) and (max_cpu_temp > 65.0 or bat_temp > 60.)):
       # onroad not allowed
       thermal_status = ThermalStatus.danger
-    elif max_comp_temp > 96.0 or bat_temp > 60.:
+    elif max_comp_temp > 96.0 or bat_temp > 60.0:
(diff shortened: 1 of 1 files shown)


pd0wm:
None of these changes fix the issue in #1876. Also the added if conditions don't make sense.

selfdrive/thermald/thermald.py
```diff
@@ -365,15 +365,15 @@ def thermald_thread():
...
     # controls will warn with CPU above 95 or battery above 60
     if thermal_status >= ThermalStatus.danger:
       should_start = False
-      if thermal_status_prev < ThermalStatus.danger:
+      if thermal_status_prev >= ThermalStatus.red:              # show warning if tempeature gets too high
         put_nonblocking("Offroad_TemperatureTooHigh", json.dumps(OFFROAD_ALERTS["Offroad_TemperatureTooHigh"]))
     else:
-      if thermal_status_prev >= ThermalStatus.danger:
-        params.delete("Offroad_TemperatureTooHigh")
+      if thermal_status_prev >= thermal_status and thermal_status < ThermalStatus.danger:
```
pd0wm, line 375: - The thermal status is already < danger, otherwise this branch wouldn't be entered.
 - The comment mentions decreasing, but it's always decreasing in this case. The param is only set on danger, and only cleared when no longer danger -> so it must be decreasing.

selfdrive/thermald/thermald.py
```diff
@@ -273,10 +273,10 @@ def thermald_thread():
     # If device is offroad we want to cool down before going onroad
     # since going onroad increases load and can make temps go over 107
     # We only do this if there is a relay that prevents the car from faulting
-    if max_cpu_temp > 107. or bat_temp >= 63. or (has_relay and (started_ts is None) and max_cpu_temp > 70.0):
+    if max_cpu_temp > 107.0 or bat_temp > 62.0 or (has_relay and (started_ts is None) and (max_cpu_temp > 65.0 or bat_temp > 60.)):
```
pd0wm, line 276: I don't think the 2 degree difference in battery temperature here will have any effect.

PR #1953, opened 2020-07-31
Fix OMX error on loggerd rotation when using multiple cameras

ZwX1616:
- from my observation the omx crash is due to a race condition in accessing its buffer, so the strategy is to wait for others done
- also refactored camera indexing


selfdrive/loggerd/loggerd.cc
```diff
@@ -111,17 +143,19 @@ void encoder_thread(bool is_streaming, bool raw_clips, bool front) {
   int encoder_segment = -1;
   int cnt = 0;
 
-  PubSocket *idx_sock = PubSocket::create(s.ctx, front ? "frontEncodeIdx" : "encodeIdx");
+  PubSocket *idx_sock = PubSocket::create(s.ctx, cam_idx == CAM_IDX_DCAM ? "frontEncodeIdx" : (cam_idx == CAM_IDX_ECAM ? "wideEncodeIdx" : "encodeIdx"));
```
adeebshihadeh, line 146: think it would be cleaner to set  the frame packet with the thread name above

selfdrive/loggerd/loggerd.cc
```diff
@@ -51,9 +52,26 @@
...
+#define CAM_IDX_ECAM 2
+
 #define CAMERA_FPS 20
 #define SEGMENT_LENGTH 60
+
+#define MAIN_BITRATE 5000000
+#ifndef QCOM2
+#define DCAM_BITRATE 2500000
+#else
+#define DCAM_BITRATE 5000000
```
adeebshihadeh, line 66:
```suggestion
#define DCAM_BITRATE MAIN_BITRATE
```

selfdrive/loggerd/loggerd.cc
```diff
@@ -51,9 +52,26 @@
...
+#endif
+
+#ifndef QCOM2
 #define LOG_ROOT "/data/media/0/realdata"
+#else
+#define LOG_ROOT "/userdata"
```
adeebshihadeh, line 72: think this is currently `/data`

selfdrive/loggerd/loggerd.cc
```diff
@@ -84,21 +102,35 @@ struct LoggerdState {
...
-
+#endif
     set_thread_name("FrontCameraEncoder");
-  } else {
+  } else if (cam_idx == CAM_IDX_FCAM) {
     set_thread_name("RearCameraEncoder");
+  } else if (cam_idx == CAM_IDX_ECAM) {
+    set_thread_name("WideCameraEncoder");
+  } else {
+    printf("unexpected camera index provided");
```
adeebshihadeh, line 132:
```suggestion
    LOGE("unexpected camera index provided");
```

PR #1955, opened 2020-08-01
Add Hyundai Veloster 2019

xps-genesis:
Added Hyundai Veloster FP1 and FP2

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -148,6 +148,9 @@ ECU_FINGERPRINT = {
   Ecu.fwdCamera: [832, 1156, 1191, 1342]
 }
 
+# Don't use these fingerprints for fingerprinting, they are still used for ECU detection
+IGNORED_FINGERPRINTS = [CAR.VELOSTER]
+
 FW_VERSIONS = {
   CAR.SONATA: {
     (Ecu.fwdRadar, 0x7d0, None): [
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -34,6 +34,7 @@ class CAR:
   SONATA = "HYUNDAI SONATA 2020"
   SONATA_2019 = "HYUNDAI SONATA 2019"
   PALISADE = "HYUNDAI PALISADE 2020"
+  VELOSTER = "HYUNDAI VELOSTER 2019"
 
 
 class Buttons:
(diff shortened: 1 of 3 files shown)


xps-genesis:
@CruiseBrantley

xps-genesis:
d624b3d19adce635|2020-07-31--22-39-17

selfdrive/car/hyundai/values.py
```diff
@@ -138,6 +139,9 @@ class Buttons:
   CAR.PALISADE: [{
     67: 8, 127: 8, 304: 8, 320: 8, 339: 8, 356: 4, 544: 8, 549: 8, 576: 8, 593: 8, 608: 8, 688: 6, 809: 8, 832: 8, 854: 7, 870: 7, 871: 8, 872: 8, 897: 8, 902: 8, 903: 8, 905: 8, 909: 8, 913: 8, 916: 8, 1040: 8, 1042: 8, 1056: 8, 1057: 8, 1064: 8, 1078: 4, 1107: 5, 1123: 8, 1136: 8, 1151: 6, 1155: 8, 1156: 8, 1157: 4, 1162: 8, 1164: 8, 1168: 7, 1170: 8, 1173: 8, 1180: 8, 1186: 2, 1191: 2, 1193: 8, 1210: 8, 1225: 8, 1227: 8, 1265: 4, 1280: 8, 1287: 4, 1290: 8, 1292: 8, 1294: 8, 1312: 8, 1322: 8, 1342: 6, 1345: 8, 1348: 8, 1363: 8, 1369: 8, 1371: 8, 1378: 8, 1384: 8, 1407: 8, 1419: 8, 1427: 6, 1456: 4, 1470: 8, 2000: 8, 2004: 8, 2005: 8, 2008: 8, 2012: 8
   }],
+  CAR.VELOSTER: [{
```
adeebshihadeh, line 142: add `CAR.VELOSTER` to the ignored fingerprints

adeebshihadeh:
Looks like this should be using the `hyundaiLegacy` safety mode

xps-genesis:
there were no errors on the vehicle without legacy check. the test failed even after putting it inl
 legacy check

selfdrive/car/hyundai/values.py
xps-genesis: why?

adeebshihadeh: new car ports should only use FW fingerprinting

xps-genesis: so if i add veloster to ignored fingerprint will it still use the FW fobgerprint?

adeebshihadeh:
The panda safety test is passing. The can valid check is the one that’s failing, which means one of the checks [here](https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/carstate.py#L136) is wrong for this car model. It could be the frequency, but it’s more likely checking a message that doesn’t exist on this model.

adeebshihadeh:
Can you replace the route with a longer one with openpilot engaged for at least a minute?

selfdrive/car/hyundai/values.py
adeebshihadeh: yes

adeebshihadeh:
Thanks for the car port, @xps-genesis!

PR #1957, opened 2020-08-01
Modeld: inline functions softplus&sigmoid

deanlee:


adeebshihadeh:
Does this actually have an impact on performance? As far as I know, the compiler should handle optimizations like this already.

deanlee:
yes you're right,I'm still stuck in the old C++ time,lol

PR #1966, opened 2020-08-02
camerad: protect concurrent access to shared variables and avoid race conditions

deanlee:


selfdrive/camerad/cameras/camera_qcom.cc
```diff
@@ -1732,7 +1734,9 @@ void actuator_move(CameraState *s, uint16_t target) {
   //LOGD("actuator move focus: %d", err);
 
   s->cur_step_pos = dest_step_pos;
+  pthread_mutex_lock(&s->frame_info_lock);
   s->cur_lens_pos = actuator_cfg_data.cfg.move.curr_lens_pos;
```
ZwX1616, line 1738: is this necessary?

deanlee: it looks like that this value will be written in ops_thread(by calling do_autofocus) and read in the main thread.did I get that right?

ZwX1616: the one constantly being written is called true_len_pos? I think the cur_lens_pos should be just a read-only number returned from the device.

selfdrive/camerad/main.cc
```diff
@@ -1,6 +1,7 @@
 #include <stdio.h>
 #include <signal.h>
 #include <cassert>
+#include <mutex>
```
adeebshihadeh, line 4: looks like this is only used in camera_qcom?

adeebshihadeh:
Can you rebase, merging one of your other PRs caused a merge conflict?

deanlee:
Fixed merge conflict

PR #1970, opened 2020-08-03
Add Volvo V40 support and dashcam support for Volvo EUCD-platform

danielzmod:
Hi! 
Been working a long time on a port for the older Volvo vehicles.  
Volvo V40 is on the so called C1MCA platform. These vehicles have a EPS that supports steering just by sending a angle request. The other vehicles is on the EUCD platform (V70, XC70, S60, V60, XC60) and to todays date i have not yet found a way to bypass the built in restrictions of the software. It works but not very good, since it only accepts steering in one direction at a time. When switching steering direction it needs a pause before accepting a new steering command.

Anyway the V40 works good and can support drivers with highway driving without intervention for long periods of time >10min.  Hope you like it 👍


roxasthenobody98:
Interesting. 
C1MCA is Ford Global C1 Platform. 

Does yours also have a steering lockout?

EUCD is also a Ford Platform.

danielzmod:
> Interesting.
> C1MCA is a subset of the Ford Global C1 Platform.
> 
> Does yours also have a steering lockout?
> 
> EUCD is also a Ford Platform.

No steering lockout for any of them.  =)

adeebshihadeh:
Sorry it's taken so long to respond. We can't merge this as-is. The branch includes lot of extraneous changes, including code for a platform that isn't fully working. Although it's unlikely we'd merge since the platform is no longer in production, you can open a clean, minimal PR for Volvo V40 support, and we'd consider it.

PR #1974, opened 2020-08-04
Updater tests

adeebshihadeh:
Runs the updater in various cases and scenarios. We should refactor the final swapping at some point, so we can test that too. Looks like we'll need to run this on our Jenkins instance since this needs a privileged docker container.
* [x] no update for 50 cycles
* [x] no update, then update
* [x] overlay initialized, then basedir touched
* [x] multiple updated instances
* [x] loop 10 times and update each time

### DIFF
--- a/selfdrive/test/test_updated.py
+++ b/selfdrive/test/test_updated.py
@@ -66,13 +66,16 @@ class TestUpdater(unittest.TestCase):
     for c in cmd:
       subprocess.check_output(c, cwd=cwd, shell=True)
 
-  def _start_updater(self):
+  def _get_updated_proc(self):
     os.environ["PYTHONPATH"] = self.basedir
     os.environ["UPDATER_TESTING"] = "1"
     os.environ["UPDATER_LOCK_FILE"] = os.path.join(self.tmp_dir.name, "updater.lock")
     os.environ["UPDATER_STAGING_ROOT"] = self.staging_dir
     updated_path = os.path.join(self.basedir, "selfdrive/updated.py")
-    self.updated_proc = subprocess.Popen(updated_path, env=os.environ)
+    return subprocess.Popen(updated_path, env=os.environ)
+   
+  def _start_updater(self):
+    self.updated_proc = self._get_updated_proc()
 
   def _update_now(self):
     self.updated_proc.send_signal(signal.SIGHUP)
@@ -155,7 +158,7 @@ class TestUpdater(unittest.TestCase):
   def test_overlay_reinit(self):
     self.params.put("IsOffroad", "1")
     self._start_updater()
-  
+
     time.sleep(2)
 
     overlay_init_fn = os.path.join(self.basedir, ".overlay_init")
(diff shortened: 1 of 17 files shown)


selfdrive/updated.py
```diff
@@ -96,10 +70,8 @@ def run(cmd, cwd=None):
 def remove_consistent_flag():
   os.system("sync")
   consistent_file = Path(os.path.join(FINALIZED, ".overlay_consistent"))
-  try:
+  if os.path.isfile(consistent_file):
```
pd0wm, line 73: `if consistent_file.exists()`?

Once we switch to python 3.8 we can use `consistent_file.unlink(missing_ok=True)` :)

selfdrive/updated.py
```diff
@@ -26,66 +26,40 @@
...
 from common.basedir import BASEDIR
 from common.params import Params
 from selfdrive.swaglog import cloudlog
 
+TEST_IP = "8.8.8.8"
+LOCK_FILE = "/tmp/safe_staging_overlay.lock"
 STAGING_ROOT = "/data/safe_staging"
+if os.getenv("UPDATER_TESTING") is not None:
+  TEST_IP = "localhost"
+  LOCK_FILE = os.getenv("UPDATER_LOCK_FILE", LOCK_FILE)
```
pd0wm, line 48: Why not always get this from the env variable with the default?

selfdrive/updated.py
```diff
@@ -315,46 +287,63 @@ def attempt_update():
...
   try:
     fcntl.flock(ov_lock_fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
   except IOError:
     raise RuntimeError("couldn't get overlay lock; is another updated running?")
 
-  # Wait a short time before our first update attempt
-  # Avoids race with IsOffroad not being set, reduces manager startup load
-  time.sleep(30)
-  wait_helper = WaitTimeHelper()
+  # Setup a signal handler to immediately trigger an update
```
pd0wm, line 309: What was wrong with the class? Now there is all this random code in the middle of the logic.

selfdrive/test/test_updated.py
```diff
@@ -0,0 +1,205 @@
...
+    while self.params.get("LastUpdateTime") is None:
+      if time.monotonic() - start_time > timeout:
+        raise Exception("timed out waiting for update to complate")
+      time.sleep(0.05)
+
+  def _make_commit(self):
+    self._run([
+      "git config user.email tester@testing.com",
+      "git config user.name Testy Tester",
+      "git commit --allow-empty -m 'an update'",
```
pd0wm, line 93: Maybe don't make an empty commit? Test for:
 - Creating a new file
 - Deleting an existing file
 - Making a change to a file

And verify that all three basic operations work correctly?

selfdrive/updated.py
adeebshihadeh: though the class was pretty confusing when first reading updated since it was basically a wrapper around threading.Event, can revert this

selfdrive/updated.py
adeebshihadeh: Cleaned up, was worried about these getting set outside of testing

selfdrive/updated.py
adeebshihadeh: reverted

selfdrive/test/test_updated.py
adeebshihadeh: those operations and more are now done in the commit

PR #1979, opened 2020-08-04
camerad: cache rgb_roi_buf&conv_result

deanlee:


selfdrive/camerad/main.cc
```diff
@@ -144,6 +144,11 @@ struct VisionState {
 
   PubMaster *pm;
 
+#if defined(QCOM) && !defined(QCOM_REPLAY)
+  std::unique_ptr<uint8_t[]> rgb_roi_buf;
+  std::unique_ptr<int16_t[]> conv_result;
+#endif
```
adeebshihadeh, line 150: why add these to VisionState? looks like they're only used in the processing thread

PR #1983, opened 2020-08-05
UI refactor : move driverview out of ui

deanlee:
run driverview in a separate process.

- move common functions to uicommon
- using VisionStream instead of raw vipc functions
- handle offroadLayout message in  ui to make sure driverview is shown correctly

CPU usage:

|  |  ui   | driverview  |
| ---- |  ----  | ----  |
| driver preview  | 0.3%  | 2.3% |
| on road  | 6.5%  | - |
| off road  | 10.5%  | - |
| sleep  | 0.3%  | - |

### DIFF
--- a/release/files_common
+++ b/release/files_common
@@ -324,6 +324,7 @@ selfdrive/ui/SConscript
 selfdrive/ui/*.cc
 selfdrive/ui/*.hpp
 selfdrive/ui/ui
+selfdrive/ui/driverview
 selfdrive/ui/spinner/Makefile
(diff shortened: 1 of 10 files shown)


adeebshihadeh:
This is a pretty big refactor with the potential for lots of subtle bugs, but the CPU savings are nice. Do you think we can get similar CPU savings with minimal changes to UI? Also, can you test the GPU performance (especially while running modeld)?

deanlee:
ok I'll try

PR #1985, opened 2020-08-05
Reduce Kp/Ki by 30% to eliminate steer oscillation on 2017+ CRV

energee:
Wasn’t sure what caused this but wheel wobble is really bad on released version, have done thorough testing, several hours of driving on these values and it’s a big improvement.


gregjhogan:
seems like something changed from 0.7.6 to 0.7.7 that caused this, below is before/after back-to-back drives on same corner at same speed

0.7.6
image

0.7.7
image

energee:
Anything required to merge?

PR #1989, opened 2020-08-07
Add Sonata Hybrid port

xps-genesis:


xps-genesis:
route shows the hybrid models have issues with DBC. @adeebshihadeh  see this route which I was talking about earlier where I recommended DBC change for EV messages. The OP engaged but never steers because the vehicle speed is coming in as gas pedal info.

8ad3857087181f80|2020-08-06--19-01-03

xps-genesis:
Thanks for helping with the port of SONATA HYBRID 2020
@realistic69

adeebshihadeh:
It does look like the signal is different for this car. Before merging, we'll need a new signal. Ideally, the new one should also work on the currently supported and EVs.

xps-genesis:
> It does look like the signal is different for this car. Before merging, we'll need a new signal that ideally works with the currently supported EVs.

https://github.com/commaai/openpilot/pull/1990

PR #1990, opened 2020-08-07
Separate Hyundai hybrid from Electric

xps-genesis:


xps-genesis:
https://github.com/commaai/opendbc/pull/292

xps-genesis:
https://github.com/commaai/panda/pull/582

adeebshihadeh:
All these changes are unnecessary. Let's do any necessary changes in your sonata hybrid PR. It should use the same DBC, and we'll move to a common gas pressed signal.

PR #1991, opened 2020-08-07
Hyundai: use common gas pressed signal for all cars

adeebshihadeh:
@xps-genesis There's 116 differences in `carState.gasPressed` over a single segment on the sonata.

### DIFF
--- a/selfdrive/car/hyundai/carstate.py
+++ b/selfdrive/car/hyundai/carstate.py
@@ -174,6 +174,7 @@ class CarState(CarStateBase):
       ("ACCEnable", "TCS13", 0),
       ("BrakeLight", "TCS13", 0),
       ("DriverBraking", "TCS13", 0),
+      ("DriverOverride", "TCS13", 0),
 
       ("ESC_Off_Step", "TCS15", 0),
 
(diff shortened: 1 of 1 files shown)


xps-genesis:
2 bytes
DriverOverride state-
0- none
1-gas pressed
2-Brake pressed
3.- NA

I think there is a loop or 2 difference between when the DriverOverride shows up vs CF_Ems_AclAct, this could be just because of where the information is originating and where it is getting relayed. EMS16 seems the actual originator, TCS13 is relaying information. See the snap attached. Similarly we see few loops delay with actual gas pedal position and the override information. Is there a way to compare the difference by ignoring the few loops before and after carState.gasPressed?

image


image

xps-genesis:
@adeebshihadeh - does my explanation makes sense? what more info do you need?

adeebshihadeh:
@xps-genesis I tested this in our Sonata, and it isn't exactly a gas pressed signal. The "brake pressed" and "gas and brake pressed" states aren't distinguishable in this signal, meaning gasPressed doesn't get set when also pressing the brake.

xps-genesis:
isn't that even better?  you have a wholesome signal containing both information and correctly reporting the driver override?

brake pressed info trumps over gas pressed+brake pressed. for panda safety or what ever the use case is for this signal, it seems to safety correctly that we may ise just this signal for both brake pressed and gas pressed info in panda safety code

Issue #1882, opened 2020-07-15
Improve car battery management


VirtuallyChris:
Currently the device will stay on for 30 hours, then shut off. This mostly solved battery drain issues, but for select drivers who exclusively do short drives, this causes issues.

If a driver repeatedly has a daily 5 minute commute, their comma two will consistently drain battery, and the user will not have driven enough to charge the battery between drives.

A proposed solution is to have a counter that conservatively tracks how much the car battery gets charged during the drive, and count down as the car is off. This allows the comma two to remain on for long periods if the user has just completed a long drive, but will allow it to shut off quickly if the user barely drove enough to charge their battery.

Reducing the 30 hours is also a solution, but will limit comma connect features even if the car battery is full.

cheta48:
Agreed with you.

PR #1994, opened 2020-08-07
Car power integrator + power management refactor

robbederks:
Issue #1882 

### Charging:
- If on for < 30h
    and voltage > VBAT_START_CHARGING
    and no CDP_mode

### Shutdown + stop charging:
- If on for 30h with no ignition
- If on for >10m and power integration is low
- If voltage < VBAT_PAUSE_CHARGING 
    and CDP_mode 
    and no ignition
    and no DisablePowerDown

### Shutdown:
- If batt_perc < BATT_PERC_OFF
    and battery is discharging
    and started_seen
    and ignition gone for > 1m (check off_ts)


gregjhogan:
should the integrator only count up if the car voltage is above 13.5 volts or something that indicates the alternator is running?  Lots of cars turn the alternator off to improve fuel economy.

robbederks:
> should the integrator only count up if the car voltage is above 13.5 volts or something that indicates the alternator is running? Lots of cars turn the alternator off to improve fuel economy.

Wouldn't they only turn it off if the battery is nearly full anyways? The 30Wh max drain and taking ~45min to recharge that I think are pretty conservative anyways, no?

robbederks:
@pd0wm It's writing a param about every half second. Should that be decimated, or is that speed fine?

selfdrive/boardd/boardd.cc
```diff
@@ -428,18 +400,35 @@ void hardware_control_thread() {
...
       uint16_t fan_speed = sm["thermal"].getThermal().getFanSpeed();
       if (fan_speed != prev_fan_speed || cnt % 100 == 0){
         panda->set_fan_speed(fan_speed);
         prev_fan_speed = fan_speed;
       }
+
+      // Charging mode
+      bool charging_disabled = sm["thermal"].getThermal().getChargingDisabled();
+      if (charging_disabled != prev_charging_disabled){
+#ifdef QCOM
```
adeebshihadeh, line 421: This seems confusing. Why isn't the whole thing gated on QCOM?

selfdrive/thermald/thermald.py
```diff
@@ -187,7 +184,7 @@ def thermald_thread():
   has_relay = False
 
   params = Params()
-  pm = PowerMonitoring()
+  pm = PowerMonitoring(params)
```
adeebshihadeh, line 187: No need to pass the params in. The convention is just to create a new params instance on class init.

selfdrive/thermald/power_monitoring.py
```diff
@@ -59,92 +67,120 @@ def panda_current_to_actual_current(panda_current):
...
           self.power_used_uWh = 0
         return
 
+      # Low-pass battery voltage
+      self.car_voltage_mV = ((health.health.voltage * CAR_VOLTAGE_LOW_PASS_K) + (self.car_voltage_mV * (1 -  CAR_VOLTAGE_LOW_PASS_K)))
+
+      # Cap the car battery power and save it in a param
+      self.car_battery_power_uWh = max(self.car_battery_power_uWh, 0)
+      self.car_battery_power_uWh = min(self.car_battery_power_uWh, CAR_BATTERY_CAPACITY_uWh)
+      self.params.put("CarBatteryPower", str(int(self.car_battery_power_uWh)))
```
adeebshihadeh, line 110: Let's decimate this to every second or two and make it a non blocking write (`put_nonblocking`)

adeebshihadeh:
This all seems nicely contained in thermald, so maybe we should write some unit tests for this? Especially considering that it can just shut devices off.

selfdrive/thermald/thermald.py
```diff
@@ -393,16 +390,17 @@ def thermald_thread():
...
-
     # Offroad power monitoring
     pm.calculate(health)
     msg.thermal.offroadPowerUsage = pm.get_power_used()
 
+    # Check if we need to disable charging (handled by boardd)
+    msg.thermal.chargingDisabled = pm.should_disable_charging(health, off_ts)
+
+    # Check if we need to shut down
+    if pm.should_shutdown(health, off_ts, started_seen, LEON):
```
adeebshihadeh, line 401: pass in LEON on the class init?

selfdrive/thermald/power_monitoring.py
```diff
@@ -157,10 +193,42 @@ def _perform_integration(self, t, current_power):
...
+      return False
+
+    now = sec_since_boot()
+    disable_charging = False
+    disable_charging |= (now - offroad_timestamp) > MAX_TIME_OFFROAD_S
+    disable_charging |= (self.car_voltage_mV < (VBATT_PAUSE_CHARGING * 1e3))
+    disable_charging |= (self.car_battery_power_uWh <= 0)
+    disable_charging &= (not health.health.ignitionLine and not health.health.ignitionCan)
+    disable_charging &= (self.params.get("DisablePowerDown") != b"1")
+    return disable_charging
```
adeebshihadeh, line 217: Hmm, why not use logical operators? mixing these bools and bitwise operators doesn't seem great

gregjhogan:
> Wouldn't they only turn it off if the battery is nearly full anyways? The 30Wh max drain and taking ~45min to recharge that I think are pretty conservative anyways, no?

Yeah could be, I guess someone would have to look at the data.  There are some other situations to consider, though, like I drive someone somewhere, wait for them with the car off and the ignition on to use the radio, and it will think the battery is charging.  New cars seem to draw a lot of current in accessory mode (my CR-V would only last less than 2 hours with climate control, etc...).  This would probably be much less of a concern if the charging cut-off was higher than 11V.

selfdrive/thermald/power_monitoring.py
```diff
@@ -59,92 +67,120 @@ def panda_current_to_actual_current(panda_current):
...
+    
+
   # Calculation tick
   def calculate(self, health):
     try:
       now = sec_since_boot()
 
       # Check that time is valid
-      if datetime.datetime.fromtimestamp(now).year < 2019:
+      if datetime.datetime.now().year < 2019:
```
adeebshihadeh, line 93: Why do we care if the time is valid here? Do we need anything besides seconds since boot?

selfdrive/boardd/boardd.cc
robbederks: It was at first, but then it complains about prev_charging_disabled being unused in the non-qcom case

selfdrive/thermald/thermald.py
robbederks: Don't think that'll always work, as LEON is set the first time the fan is controlled iirc

selfdrive/thermald/power_monitoring.py
robbederks: Is there a short way for doing that? Want to avoid should_start = should_start and <>

selfdrive/thermald/power_monitoring.py
robbederks: Now that we switched to sec since boot, that should indeed be fine I guess.
That doesn't jump when the system time is updated I assume?

robbederks: Now that we switched to sec since boot, that should indeed be fine I guess.
That doesn't jump when the system time is updated I assume?

robbederks: Now that we switched to sec since boot, that should indeed be fine I guess.
That doesn't jump when the system time is updated I assume?

robbederks:
> > Wouldn't they only turn it off if the battery is nearly full anyways? The 30Wh max drain and taking ~45min to recharge that I think are pretty conservative anyways, no?
> 
> Yeah could be, I guess someone would have to look at the data.  There are some other situations to consider, though, like I drive someone somewhere, wait for them with the car off and the ignition on to use the radio, and it will think the battery is charging.  New cars seem to draw a lot of current in accessory mode (my CR-V would only last less than 2 hours with climate control, etc...).  This would probably be much less of a concern if the charging cut-off was higher than 11V.

Sure, guess that would be possible, although that's also a case where the user probably is aware that stuff in the car is drawing power, although that would still throw off the efficacy counter somewhat.
I'm just reluctant to do any logic based on the battery voltage, since that's a signal which varies quite a bit per car, and with a terrible signal to noise ratio

selfdrive/thermald/power_monitoring.py
adeebshihadeh: No, it shouldn't

selfdrive/boardd/boardd.cc
adeebshihadeh: I'd gate prev_charging_disabled on QCOM too then. We could also define an UNUSED macro like we do in panda

selfdrive/thermald/power_monitoring.py
adeebshihadeh: I don't think so, usually it's done like this in python
```python
disable_charging = (condition and
                    second_condition and
                    third condition)
```

selfdrive/thermald/power_monitoring.py
pd0wm: If you do a non blocking put, keep in mind the writes can happen in the wrong order. Is that fine for the rest of the implementation?

robbederks: Should be finem yes. The param is only used on init (after reboot for example). It's not critical that it's always the exact value, it's an approximation anyways

selfdrive/thermald/power_monitoring.py
robbederks: Hmm, that would become this:
```return ((((now - offroad_timestamp) > MAX_TIME_OFFROAD_S) or
            (self.car_voltage_mV < (VBATT_PAUSE_CHARGING * 1e3)) or
            (self.car_battery_power_uWh <= 0)) and 
            (not health.health.ignitionLine and not health.health.ignitionCan) and
            (self.params.get("DisablePowerDown") != b"1"))```
Not sure if that's clearer or worse

selfdrive/thermald/power_monitoring.py
robbederks: Ok, removed!

PR #1999, opened 2020-08-08
Fix SIGSEGV bug 

deanlee:
if do_exit is true while trying reconnecting , visionstream_destroy will be called incorrectly after exiting loop


selfdrive/modeld/dmonitoringmodeld.cc
```diff
@@ -84,11 +83,9 @@ int main(int argc, char **argv) {
       LOGD("dmonitoring process: %.2fms, from last %.2fms", t2-t1, t1-last);
       last = t1;
     }
-
+    visionstream_destroy(&stream);
   }
 
-  visionstream_destroy(&stream);
```
deanlee, line 90: if do_exit is true while reconnecting, visionstream_destroy will be called twice here

PR #2004, opened 2020-08-09
ublox_msg.cc: simplify map operations in gen_nav_data

deanlee:


geohot:
Is this the same thing? It looks like if subframeId is 1, it cleared the map in the past.

deanlee:
@geohot i think they are exactly the same now.

adeebshihadeh:
Can you rebase?

deanlee:
Done

Issue #1041, opened 2020-02-02
Refactor params to use the same underlying implementation -- $200 bounty


geohot:
Currently there are two (or maybe three in Java):

* selfdrive/common/params.cc
* common/params.py

$200 for a merged PR that follows the spec below.

pd0wm:
Think the python version has the most safe implementation. The C++ implementation needs to be written like the Python one, and then the Python one can become a Cython wrapper around the C++ implementation.

LoganxDev:
Started working on this one.

ahowe444:
Just submitted a PR for this.

PR #2005, opened 2020-08-09
Params refactor

whowe444:
This PR addresses issue #1041 . I've basically done what @pd0wm suggested. 

1) I first built the C++ implementation and replicated the tests in openpilot/common/tests/test_params.py within the main method of openpilot/common/params_helper.cpp . I then built the cython wrapper using the new C++ params implementation and was able to pass all the python unittests. 

2) I symlinked my C++ implementation into selfdrive/common/params.cc and redid the implementation of params.cc using the new underlying implementation. This passes all tests as far as I can tell. I'm not sure if the test_params.cc file is used currently though...

Some sketchy things I had to do
------------------------------------------
1) the pylint checks fail on cython-derived extensions (as far as I can tell this is a bug in pylint) so I had to ignore these checks anytime anyone imports Params `# pylint: disable=no-name-in-module, import-error`

2) In order to pass the blocking unittest in common/tests/test_params.py which starts a thread, I have to release the GIL in the cython wrapper params.get blocking function. This allows the blocking variant of the get function to work while the non-blocking version does not with a delayed thread.  

3) I hope someone can more closely review DBWriter::exit() in params_helper.cpp. I tried to make the implementation fairly similar to the python in structure, but the params.py file had some weird exception handling in its exit() function. 

So, I'm passing the tests, but I worry that they are a low bar...


ahowe444:
The first check fails only because there is no longer a params.py file, the extension is built in SConscript with cython.

pd0wm:
Thanks for doing the refactor! This quite a big PR, so will take some time to review. Stay tuned!

adeebshihadeh:
This is in the milestones for 0.7.10. Can you rebase on master?

ahowe444:
> This is in the milestones for 0.7.10. Can you rebase on master?

yes, will do.

pd0wm:
@ahowe444 Did you receive our email regarding the bounty?

PR #2006, opened 2020-08-10
hyundai - always hda mode when engaged

gregjhogan:
removes the need to spam resume, depends on DBC update from HKG community
maybe also adds slow down for corners?

### DIFF
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -1,5 +1,5 @@
 from cereal import car
-from common.realtime import DT_CTRL
+#from common.realtime import DT_CTRL
 from selfdrive.car import apply_std_steer_torque_limits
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
Since this only delays the need to spam to resume, is there any benefit to this over fixing resume button spamming?

gregjhogan:
I believe the only other benefit might be adjusting the stock longitudinal speed (if openpilot ever outputs a desired speed, or automatically adjusting speed based on speed limit from cars internal maps)

Button spamming is unreliable at least on the palisade. Somewhere in the range of 1% to 10% of the time the car will not accept the resume command from the openpilot for 5+ seconds (which would be 25 attempts since it is currently sent at 5 Hz).  As long as this issue exists, only requiring a resume command every 30 seconds makes a big difference (since it is very rare to stop for more than 30 seconds).

I will open a separate issue for the button spamming problem.

PR #2010, opened 2020-08-11
Car Port: 2020 Hyundai Kona

xps-genesis:


adeebshihadeh:
LGTM, but the test route hasn't been uploaded. Make sure the route fully uploads or use a new one that has uploaded.

selfdrive/car/hyundai/values.py
```diff
@@ -28,6 +28,7 @@ class CAR:
   KIA_OPTIMA_H = "KIA OPTIMA HYBRID 2017 & SPORTS 2019"
   KIA_SORENTO = "KIA SORENTO GT LINE 2018"
   KIA_STINGER = "KIA STINGER GT2 2018"
+  KONA = "HYUNDAI KONA ELECTRIC 2020"
```
adeebshihadeh, line 31: why does KONA_EV have the same name?

xps-genesis: oops my mistake! fixed it.

PR #2012, opened 2020-08-11
Setup script improvements to MacOS / Ubuntu

michaelhonan:
**Description**: Only install brew if required on MacOS + setup script improvements to MacOS / Ubuntu

**MacOS install script:**
MacOS setup script was installing Homebrew even if it was already installed. Now it'll only install if required, and will run `brew update` if an installation is found. This script will also auto detect zsh/bash shell and pick the appropriate RC file to install to.

**bashrc/zshrc changes:**
I've also modified the bashrc/zshrc install to detect the repository root directory. So it won't matter whether your working directory is `.` or `./tools` when you run the scripts.

**Verification**: I was able to simply run the script, then instantly run `scons` and openpilot built on MacOS 10.15.3.

Edit: 2020-08-12 - Just ran up a brand new install of MacOS 10.15.6 and installed. openpilot built after my recent `libusb` commit.

### DIFF
--- a/tools/mac_setup.sh
+++ b/tools/mac_setup.sh
@@ -16,6 +16,7 @@ brew install capnp \
              ffmpeg \
              glfw \
              libarchive \
+             libusb \
              libtool \
              llvm \
              pyenv \
--- a/tools/mac_setup.sh
+++ b/tools/mac_setup.sh
@@ -1,7 +1,13 @@
 #!/bin/bash -e
 
-# install brew
-/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+# Install brew if required.
+if [[ $(command -v brew) == "" ]]; then
+    echo "Installing Hombrew"
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+else
+    echo "Updating Homebrew"
+    brew update
+fi
 
 brew install capnp \
              czmq \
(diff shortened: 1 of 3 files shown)


tools/mac_setup.sh
```diff
@@ -21,11 +21,19 @@ brew install capnp \
              pyenv \
              zeromq
 
+# install to .zshrc
+source ~/.zshrc
```
adeebshihadeh, line 25: can we detect their shell instead of assuming it's zsh?

tools/mac_setup.sh
```diff
@@ -15,12 +21,31 @@ brew install capnp \
...
+  RC_FILE="$HOME/.bash_profile"
+else
+  echo "-------------------------------------------------------------"
+  echo "Unsupported shell: \"$SHELL\", cannot install to RC file."
+  echo "Please run: echo \"source $OP_DIR/tools/openpilot_env.sh\" >> %YOUR SHELL's RC file%"
+  echo "-------------------------------------------------------------"
+fi
+
+# Install to RC file.
+source ~/.zshrc
```
adeebshihadeh, line 37: this is unnecessary

tools/mac_setup.sh
```diff
@@ -15,12 +21,31 @@ brew install capnp \
...
+if [ -z "$OPENPILOT_ENV" ] && [ -n "$RC_FILE" ]; then
+  OP_DIR=$(git rev-parse --show-toplevel)
+  echo "source $OP_DIR/tools/openpilot_env.sh" >> $RC_FILE
+  source ~/.zshrc
```
adeebshihadeh, line 41:
```suggestion
  source $RC_FILE
```

tools/mac_setup.sh
michaelhonan: This was present on the ubuntu_setup.sh, should it also be removed there?

PR #2014, opened 2020-08-11
Don't check dirty on prebuilt release

### LINKED ISSUE #973: Importing selfdrive.version writes to .git
Importing selfdrive.version writes to the .git directory. Probably due to:
`        subprocess.check_call(["git", "update-index", "--refresh"])`

Writing to disk during the startup process could cause disk corruption on the comma two if were to unplug it during this write operation.

Writing to the `.git` directory also causes the updater to reinitalize the overlay.

pd0wm:
fixes: #973

### DIFF
--- a/selfdrive/version.py
+++ b/selfdrive/version.py
@@ -68,15 +68,18 @@ if (origin is not None) and (branch is not None):
         subprocess.check_call(["git", "update-index", "--refresh"])
       except subprocess.CalledProcessError:
         pass
-      dirty = dirty or (subprocess.call(["git", "diff-index", "--quiet", branch, "--"]) != 0)
+      dirty = (subprocess.call(["git", "diff-index", "--quiet", branch, "--"]) != 0)
 
       # Log dirty files
       if dirty:
-        dirty_files = subprocess.check_output(["git", "diff-index", branch, "--"], encoding='utf8')
-        commit = subprocess.check_output(["git", "rev-parse", "--verify", "HEAD"], encoding='utf8').rstrip()
-        origin_commit = subprocess.check_output(["git", "rev-parse", "--verify", branch], encoding='utf8').rstrip()
-        cloudlog.event("dirty comma branch", version=version, dirty=dirty, origin=origin, branch=branch,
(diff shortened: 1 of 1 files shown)


selfdrive/version.py
```diff
@@ -1,47 +1,47 @@
 #!/usr/bin/env python3
 import os
 import subprocess
+from common.basedir import BASEDIR
 from selfdrive.swaglog import cloudlog
 
 
-def get_git_commit(default=None):
+def run_cmd(cmd):
+    return subprocess.check_output(cmd, encoding='utf8').strip()
```
adeebshihadeh, line 9: we should probably abstract this somewhere else with an optional default

selfdrive/version.py
```diff
@@ -51,35 +51,42 @@ def get_git_remote(default=None):
...
-    subprocess.check_call(["git", "update-index", "--refresh"])
-  except subprocess.CalledProcessError:
-    pass
-
-  if (origin is not None) and (branch is not None):
     comma_remote = origin.startswith('git@github.com:commaai') or origin.startswith('https://github.com/commaai')
     tested_branch = get_git_branch() in ['devel', 'release2-staging', 'dashcam-staging', 'release2', 'dashcam']
 
-    dirty = not comma_remote
+    dirty = False
```
adeebshihadeh, line 59:
```suggestion
    dirty = not prebuilt
```

selfdrive/version.py
```diff
@@ -1,47 +1,47 @@
 #!/usr/bin/env python3
 import os
 import subprocess
+from common.basedir import BASEDIR
 from selfdrive.swaglog import cloudlog
 
 
-def get_git_commit(default=None):
+def run_cmd(cmd):
```
adeebshihadeh, line 8: take a string and split by whitespace?

pd0wm: Like the list better.

selfdrive/version.py
pd0wm: I don't think that will matter. If it's not prebuilt the dirty variable will overwritten a few lines below. If the git command fails it will be marked dirty anyway.

PR #2017, opened 2020-08-12
update optima fingerprint and Fw Fp

xps-genesis:
Add optima FW fingerprint.
Clean up optima multiple fingerprint strings


xps-genesis:
98d9b2466a864999|2020-08-11--21-00-36

adeebshihadeh:
We only ignore fingerprints for new car ports.

PR #2030, opened 2020-08-16
Fw FP should only look for MDPS part number

xps-genesis:
It is noticed that the Fw FP can get out of proportion as vehicle ECUs can have multiple part number depending on variants. There is no need for looking and capturing ECU part numbers other than MDPS. MDPS alone should give us unique part number per vehicle model. This reduces work load in adding multiple FwFP for each variant of a vehicle model


adeebshihadeh:
MDPS fw version doesn’t uniquely identify a vehicle. The Optima and Santa Fe have the same MDPS fw version.

PR #2034, opened 2020-08-16
Hyundai to look for MDPS ecu Fw for FP2.0, if identical between 2 vehicles, look for Engine ecu

xps-genesis:
Hyundai to look for MDPS Fw for FP2.0, if identical between 2 vehicles, look for Engine, And only add more ECU checks if there are more identical Fw part numbers between 2 or more vehicles

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -151,105 +151,34 @@ IGNORED_FINGERPRINTS = [CAR.VELOSTER, CAR.GENESIS_G70, CAR.KONA]
 
 FW_VERSIONS = {
   CAR.SONATA: {
-    (Ecu.fwdRadar, 0x7d0, None): [
-      b'\xf1\x00DN8_ SCC FHCUP      1.00 1.00 99110-L0000         ',
-      b'\xf1\x00DN8_ SCC F-CU-      1.00 1.00 99110-L0000         ',
-    ],
-    (Ecu.esp, 0x7d1, None): [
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
What’s the benefit to this?

xps-genesis:
I expect the current setup will cause a never ending request to add more FW because vehicle ECUS will have hundreds of Fw part number every year. that multiplied by the number of vehicles under HKG brand. we are talking about 100s or may be even 1000s of Fw FPs. 

I just saw Toyota values.py, that file has blown to so many lines of code with information that is essentially only used to determine which vehicle. if we apply this rule we can cut down that file to 1/10th of its current size.

Since Hyundai is only starting to add Fw FPs, we may start a never ending addition of Fw FPs. 

Minimum data to figure out which vehicle makes most sense.

one example is MY20 and MY21 palisade has all the Fw FPs different expect for MDPS. so essentially now we need to record all other ecus for MY21 palisade which we could have avoided if we only recorded MDPS and still got the same result.

We chose MDPS ecu over other because it is seen so far that MDPS seems to have the least variability in a vehicle model, yet almost different between vehicles.

pd0wm:
We use FW query for more than just determining the make/model. As we move towards a consumer product we need to make sure that the messages on the CAN bus are exactly as we expect them. If one of the important modules that we process CAN messages from gets updated, we want to make sure openpilot still works before we add its FW version.

I do want to keep building the full database of possible firmware versions. If the number becomes too large and it changes too quickly we will move the database outside of values.py and ship it separate from openpilot so it can be updated more frequently.

PR #2046, opened 2020-08-19
Improved updater robustness

adeebshihadeh:
* [x] replace ping with git ls-remote
* [x] improve responsiveness, get remote heads every minute and git fetch every 10


adeebshihadeh:
I think the main advantage to keeping the overlay is how cheap it is to start over if either the finalized update or the merged overlay get into a bad state.

PR #2037, opened 2020-08-17
ui: refactor draw model

deanlee:
1. refactor model drawing.
2.  remove unused "liveMpc" related funtions.(I'll restore them if you think it's necessary)

PR #2063, opened 2020-08-23
UI: refactor vision

deanlee:
using a single VisionStream instance for both rear and front camera.

### DIFF
--- a/selfdrive/ui/android_ui.cc
+++ b/selfdrive/ui/android_ui.cc
@@ -329,7 +329,7 @@ int main(int argc, char* argv[]) {
     // safe to do this outside the lock?
     if (should_swap) {
       double u2 = millis_since_boot();
-      if (u2-u1 > 66) {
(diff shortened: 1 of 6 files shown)


adeebshihadeh:
Does this have the same CPU savings as #1983? Any improvement with the GPU?

deanlee:
I think it does.this PR does not subscribe and process front camera data when on-roading. the performance might be better if combine with these PR: https://github.com/commaai/openpilot/pull/2025 https://github.com/commaai/openpilot/pull/2037 .but I need more testing to confirm that

adeebshihadeh:
Did a quick check, and this seems to have the same CPU savings. Can you rebase and I'll review tomorrow?

deanlee:
Ok i'll do it in a while

adeebshihadeh:
Thanks for the PR, fixed up some bugs and did some more refactoring in #2115.

PR #2065, opened 2020-08-23
camerad: close file descriptors

deanlee:
close all file descriptors with unique_fd

### DIFF
--- a/selfdrive/common/utilpp.h
+++ b/selfdrive/common/utilpp.h
@@ -63,9 +63,9 @@ inline std::string readlink(std::string path) {
 }
 
 struct unique_fd {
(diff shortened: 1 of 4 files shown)


selfdrive/common/utilpp.h
```diff
@@ -63,4 +62,10 @@ inline std::string readlink(std::string path) {
 
 }
 
-#endif
+struct unique_fd {
+  unique_fd(int fd=-1) : fd_(fd) {}
+  unique_fd& operator=(unique_fd&& uf) { fd_ = uf.fd_; uf.fd_ = -1; return *this; }
```
adeebshihadeh, line 67: add some new lines here

deanlee: done

PR #2067, opened 2020-08-24
swaglog.cc: Remove unnecessary string copy

deanlee:


pd0wm:
Are we sure there wasn't a reason why this was done? Maybe to solve alignment issues?

deanlee:
I don't know why they are here. I dived into json11's source and pretty sure the memory returned by dump is already aligned. The the possible reason(maybe) is want to be consistent with previous c version in this pr:https://github.com/commaai/openpilot/commit/46f3665279222ce53333a8c001def1b2a8e32926. @adeebshihadeh  did I get that right?

pd0wm:
That makes sense. I can imaging how this came into existence. If @adeebshihadeh doesn't remember anything special about this code I'll merge it.

adeebshihadeh:
No, don't remember anything special about this.

PR #2068, opened 2020-08-24
modeld: adding checks for sm.updated

deanlee:


adeebshihadeh:
Does this actually have any impact on performance? Both should be updated every time anyway.

deanlee:
You are right, it really doesn't make sense.
another question: can we move them into live_thread?

adeebshihadeh:
I don't really see the point of that. It just creates more globals

PR #2076, opened 2020-08-24
Tune TSS1 Corolla + add derivative for better control

sshane:
**Description** Added derivative to pid.py for better lateral control. I've added a tune for my 2017 TSS1 Corolla and set all other cars to 0, rendering no change in control. The steering feels much more solid with derivative added, and it loses control *much* less often in tight corners and with curves requiring quick response. It also almost completely removes oscillation on slight curves when the curve is easing back into a straight.

I can provide some routes or data comparing the change, just let me know!

[A change to cereal is also required, here's that PR](https://github.com/commaai/cereal/pull/83)


selfdrive/controls/lib/pid.py
```diff
@@ -51,22 +56,24 @@ def reset(self):
...
     self.p = error * self.k_p
     self.f = feedforward * self.k_f
+    d = self.k_d * (error - self.last_error)
 
     if override:
       self.i -= self.i_unwind_rate * float(np.sign(self.i))
     else:
       i = self.i + error * self.k_i * self.i_rate
-      control = self.p + self.f + i
+      control = self.p + self.f + i + d
```
ShaneSmiskol, line 76: I assume we should add derivative here, but not sure

selfdrive/controls/lib/pid.py
```diff
@@ -51,22 +56,24 @@ def reset(self):
...
-    self.control = 0
+    self.control = 0.0
 
   def update(self, setpoint, measurement, speed=0.0, check_saturation=True, override=False, feedforward=0., deadzone=0., freeze_integrator=False):
     self.speed = speed
 
     error = float(apply_deadzone(setpoint - measurement, deadzone))
     self.p = error * self.k_p
     self.f = feedforward * self.k_f
+    d = self.k_d * (error - self.last_error)
```
ShaneSmiskol, line 70: Should I add this to the pid log?

ShaneSmiskol:
I'm beside myself how good steering is with the safetyParam you guys just merged plus full derivative. My car can take curves and quick changes in curvature it never could before, and with much less oscillation on straights. Is there anyway I can help get this merged?

ShaneSmiskol:
I will upload a graph of the data gathered using that debugging script vs. stock

adeebshihadeh:
Closing this as I don't think we want to add a derivative term to the controller.

ShaneSmiskol:
Do you mind if I ask why you don't want derivative? It really has helped smoothed out and upped the level of control openpilot has over the wheel in turns.

adeebshihadeh:
Willem and I discussed it a bit, and we think INDI should be used if you want more parameters to play with. We're also looking into the Toyota issue more.

PR #2077, opened 2020-08-25
Loggerd rotation test

ZwX1616:
this only tests rotations and its main goal is to verify #1953


pd0wm:
Can we do this without actually starting camerad? Can we create a fake camerad that just outputs frames?

ZwX1616:
What is the benefit of using a fake camerad except from the real one occasionally not closing? I think that adds more code because it needs to be device specific and such.

ZwX1616:
> Can we do this without actually starting camerad? Can we create a fake camerad that just outputs frames?

What is the benefit of using a fake camerad except from the real one occasionally not closing? I think that adds more code because it needs to be device specific and such.

PR #2080, opened 2020-08-25
Hardware abstraction class

pd0wm:
Only functional change should be that if reading IMEI on android fails, it now won't return a random string. I don't think that was intended anyway.

### DIFF
--- a/selfdrive/manager.py
+++ b/selfdrive/manager.py
@@ -14,7 +14,7 @@ from selfdrive.swaglog import cloudlog, add_logentries_handler
 
 
 from common.basedir import BASEDIR, PARAMS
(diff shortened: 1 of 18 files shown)


selfdrive/manager.py
```diff
@@ -218,7 +217,8 @@ def get_running():
   'uploader',
 ]
 
-if ANDROID:
+# TODO: if not PC:
+if not ANDROID:
```
adeebshihadeh, line 221: why did this change? seems like a debugging artifact?

pd0wm: Oh, my bad. That's wrong.

PR #2081, opened 2020-08-25
UI: refactor the way read paramaters

deanlee:


selfdrive/ui/android_ui.cc
```diff
@@ -150,6 +150,12 @@ static void update_offroad_layout_state(UIState *s) {
   if (timeout > 0) {
     timeout--;
   }
+  if (prev_app != s->active_app && s->active_app == cereal::UiLayoutState::App::NONE) {
+    read_param(&s->is_metric, "IsMetric");
```
pd0wm, line 154: Can you put this in a function in ui.cc so we can also call it from qt?

adeebshihadeh:
UI doesn't really need to be reading most of these params, except LastAthenaPingTime and IsMetric. Cleaning this up in #2091.

adeebshihadeh:
Merged in #2091, so this shouldn't be necessary anymore

PR #2088, opened 2020-08-26
Ford: Add FPv2 Support

roxasthenobody98:
Added support for FPv2 on Ford. 

Requests have to be sent on both bus 0 and bus 1.


adeebshihadeh:
Nice, how about we do this in a big Ford PR once a port's done without the steering lockout?

adeebshihadeh:
Closing, let's just merge this into #1899

PR #2108, opened 2020-08-31
Reduce plannerd and dmonitoringd CPU usage

adeebshihadeh:
Saves around ~10% of a core on comma two. There's more CPU to be saved in the locationd processes (probably another ~15%) in another PR.

plannerd 18% -> 12%
dmonitoringd 6% -> 1.8%


pd0wm:
Nice CPU savings!

I don't think this is correct for locationd though. For things like dmonitoringd it doesn't matter what the intermediate messages for controlsState were. However, for locationd all the intermediate measurements need to be put into the filter.

PR #2109, opened 2020-08-31
Refactor Python code to follow PEP8 guidelines and increase consistency

sshane:
Some changes I wanted to make to follow some PEP8 guidelines as well as make the coding style more consistent across openpilot
- Remove redundant parentheses (for if checks as well as class declarations with no bases)
- Correct spacing
  - For comments
  - Between functions
  - Between classes
  - For global variables under imports
  - For if checks

### DIFF
--- a/common/api/__init__.py
+++ b/common/api/__init__.py
@@ -4,10 +4,11 @@ from datetime import datetime, timedelta
 from common.basedir import PERSIST
 from selfdrive.version import version
 
-class Api():
+
+class Api:
   def __init__(self, dongle_id):
     self.dongle_id = dongle_id
-    with open(PERSIST+'/comma/id_rsa') as f:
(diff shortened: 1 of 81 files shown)


pd0wm:
@ShaneSmiskol We've gotten PRs like this in the past and decided it's not worth the massive effort to review. If you'd fix all of the minor PEP8 formatting stuff it would be tens of thousands of lines.

We put in a linter to enforce some of the sane PEP8 rules but not on formatting, since we don't want to increase the burden on contributors. Feel free to fix up a file if you're making PR for it anyway, but it's not worth it to start fixing everything.

ShaneSmiskol:
Sounds good!

ShaneSmiskol:
Sounds good

PR #2111, opened 2020-08-31
sensord cleanup

adeebshihadeh:


selfdrive/sensord/sensors_qcom.cc
```diff
@@ -70,30 +69,33 @@ void sensor_loop() {
...
+
+    std::map<int, int64_t> sensors = {
+      {SENSOR_GYRO_UNCALIBRATED, ms2ns(10)},
+      {SENSOR_MAGNETOMETER_UNCALIBRATED, ms2ns(100)},
+      {SENSOR_ACCELEROMETER, ms2ns(10)},
+      {SENSOR_GYRO, ms2ns(10)},
+      {SENSOR_MAGNETOMETER, ms2ns(100)},
+      {SENSOR_PROXIMITY, ms2ns(100)},
+      {SENSOR_LIGHT, ms2ns(100)}
+    };
```
deanlee, line 91: It might be simpler to use a sensor struct?

adeebshihadeh: think I'll just merge as is since it'll need to be revisited at some point. right now, we have the qualcomm daemon and our daemon running (4% and 6% CPU respectively), and ideally we should be able to only use one daemon with far less CPU

deanlee: oh I get it.

PR #2114, opened 2020-08-31
TextWindow Enhancements

sshane:
- [x] More detailed error messages (previously only shows first three lines and last). Now it really shows last 3 lines


adeebshihadeh:
All of the git stuff is intentionally contained in `updated`. I'm not sure any of this stuff is useful outside of forks. All of our CI built branches only get pushed if they build, so the failed to build window is really only a last resort.

ShaneSmiskol:
Yes, I had forks in mind so users had a way to easily pull and reset to origin if a fork owner made a mistake causing updated to not start. But it could be useful on stock if a user was making local changes and he or she messed up some code causing it to again not start. I think the font and transparency changes are really nice though, so I can leave out the git code and button if that's what you want.

The way I did it was send a different exit code from text if the user taps on the reset button, then manager would handle the git code. I could import `attempt_update` from updated and call that function from manager maybe?

adeebshihadeh:
Hmm, I think the transparency is a bit much, but can you post a before/after of the font size change? Calling `attempt_update` won't work, and that's pretty dirty. The extra git stuff can easily live inside a fork.

ShaneSmiskol:
Sure! Once I get the submodules to work on my EON I'll get some screenshots of before/after

ShaneSmiskol:
Here's some comparisons: https://imgur.com/a/rXbVWlL (the order might be incorrect, it's original, original with more verbose error, and new textwindow). Ignore the failing check for now, that's just to get the TextWindow to show up

adeebshihadeh:
I think the point of this is largely just to indicate some kind of build failure, which really shouldn't happen on a stock branch. If you're making local changes or using a fork, you should confirm at the very least that it builds before running the code. The current text window is informative and large enough to see while mounted which I think is more important than cramming a bunch of lines on screen.

ShaneSmiskol:
Okay but what about it showing the first three lines + the last exception? It doesn't say where the error is coming from or the actual line

adeebshihadeh:
Sure, using the last 3 lines sounds fine. That should just be `traceback.format_exc(3)	` -> `traceback.format_exc(-3)`. I'm not a fan of the transparency, really shouldn't be depending on this text window.

ShaneSmiskol:
Sounds good, I'll just move the text over and the button down a bit so the lines don't clip out too much, I'll leave the text size the same

adeebshihadeh:
Can you use the original x offset? Other than that, LGTM.

selfdrive/manager.py
```diff
@@ -624,9 +624,8 @@ def main():
     cloudlog.exception("Manager failed to start")
 
     # Show last 3 lines of traceback
-    error = traceback.format_exc(3)
-
-    error = "Manager failed to start\n \n" + error
+    error = [line.replace('/data/openpilot', '') for line in traceback.format_exc(-3).splitlines()]
```
adeebshihadeh, line 627: This should just be `traceback.format_exc(-3)`, no need to hide the full path.

selfdrive/controls/lib/events.py
```diff
@@ -331,7 +331,7 @@ def wrong_car_mode_alert(CP, sm, metric):
       "TAKE CONTROL",
       "Lane Departure Detected",
       AlertStatus.userPrompt, AlertSize.mid,
-      Priority.LOW, VisualAlert.steerRequired, AudibleAlert.chimePrompt, 1., 2., 3.),
+      Priority.LOW, VisualAlert.steerRequired, AudibleAlert.chimePrompt1, 1., 2., 3.),
```
adeebshihadeh, line 334: why did this change?

ShaneSmiskol: I had that just to trigger an exception, that will be removed

ShaneSmiskol:
My EON isn't turning on so I can't test the new changes, give me a bit and I'll update the PR

ShaneSmiskol:
Here's without any textview changes: https://i.imgur.com/Oodkep9.jpg You can see there's a bit of extra space on the left where we could move the text over to see more of the error, if you're fine with keeping that

adeebshihadeh:
I think the margin is good, but is there a simple way to wrap the text?

ShaneSmiskol:
I think I tried looking into that with nanovg a while ago but couldn't find anything too straight forward, I'll take a look again though soon

PR #2116, opened 2020-08-31
Subtract standard camera offset from CAMERA_OFFSET with p_poly

sshane:
Just so if a user changes their camera offset offline or with a fork to compensate for their mount, it doesn't introduce oscillation when the lane lines are fading in and out, or if they are low confidence. Since they will be mismatched if not offsetting pPoly with the difference of the offset already in the path from the model.

This should have no effect on stock openpilot


adeebshihadeh:
I really don't think adding null-effect random constants is the solution here.

ShaneSmiskol:
How about an if check with a STANDARD_CAMERA_OFFSET variable and a changeable CAMERA_OFFSET variable? Then throw an if check to offset for the path poly if they're different, from the user. Think it might be a good idea to include this, especially since Torq_boi has said that changing the camera offset introduces weird behavior here: https://discord.com/channels/469524606043160576/538741329799413760/695014354428362868

PR #2117, opened 2020-09-01
Hyundai SCC standstill resume logic

xps-genesis:
The current implementation of cruise standstill resume logic is not good enough! There is a constant hit resume display on cluster, and constant cycling until the vehicle start to move. Also the frequency to send clu11 is unnecessary. we can send the clu11 msg at the loop rate the logic is running. 

the logic sends resume when the vehicle in front moves, just like how human driver would hit physical resume switch.

This works very well and works every single time without any errors( atleast no errors in the past 2 days of driving)


xps-genesis:
reopen, accidental closure

PR #2125, opened 2020-09-02
Refactor camerad

deanlee:
1.Add class CamraBuf  to handle buffers. remove a lot of redundant code.
2.Move all camera-related data to the corresponding files under /cameras.
3.Each camera uses its own callback function to process the frame, function name start with the prefix camera_process_.
4.The main.cc  only handles client connections.
5.Slightly modified visionbuf to meet the needs of refactor,and fixed the memory leak caused by visionbuf_to_cl

tested on QCOM


ZwX1616:
Great PR! Is it possible that we remove all the unnecessary #ifdefs and move them to the individual files in cameras/ ? Ideally more generic main.cc the better

deanlee:
That would be better.but it will cause a very large PR.
can we simply move the camera buffs into class in this PR?then create another PR to solve the rest.how do you think about this idea?

deanlee:
ok i'll try

deanlee:
It's basically done, @ZwX1616  how about this version?

ZwX1616:
> It's basically done, only tested on QCOM. @ZwX1616 how about this version?

Looking good. I'll test it and let you know. Thanks!

deanlee:
Continue to refactor based on @ZwX1616 's suggestions. thank you @ZwX1616 for your advice and encouragement on WeChat!

deanlee:
Although there are still many things can be optimized, I think this PR has already done what I originally wanted to do.

selfdrive/camerad/cameras/camera_webcam.cc
```diff
@@ -221,56 +213,72 @@ CameraInfo cameras_supported[CAMERA_ID_MAX] = {
...
 
-void cameras_run(MultiCameraState *s) {
-  set_thread_name("webcam_thread");
 
-  int err;
-  pthread_t rear_thread_handle;
-  err = pthread_create(&rear_thread_handle, NULL,
-                        rear_thread, &s->rear);
-  assert(err == 0);
+void camera_process_front(MultiCameraState *s, CameraState *c, int cnt, ) {
```
ihsakashi, line 254: tiny cleanup

adeebshihadeh:
Continued in #2310

PR #2126, opened 2020-09-03
Tici updated

adeebshihadeh:


pd0wm:
overlayfs is now enabled in the kernel. We just need to figure out how to allow mounting as non root

PR #2129, opened 2020-09-04
UI: refactor light sensor

adeebshihadeh:
In preparation of #2147.

### DIFF
--- a/selfdrive/ui/android/ui.cc
+++ b/selfdrive/ui/android/ui.cc
@@ -3,6 +3,8 @@
 #include <math.h>
 #include <sys/resource.h>
 
+#include <algorithm>
+
 #include "common/util.h"
 #include "common/utilpp.h"
 #include "common/params.h"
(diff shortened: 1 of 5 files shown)


deanlee:
I think it might not be as good as tthread.light sensor event is generated only if the value has changed,the performance of the threading will be much better than sensorEvents

adeebshihadeh:
We're planning on implementing double tap to wake with the accelerometer soon, so this should be the same since the accelerometer updates faster than UI.

selfdrive/ui/android/ui.cc
```diff
@@ -93,24 +41,35 @@ static void enable_event_processing(bool yes) {
...
+  static int display_mode = HWC_POWER_MODE_OFF;
+  static int display_timeout = 0;
+
+  // determine desired state
+  int desired_mode = display_mode;
+  if (user_input || s->ignition || s->started) {
+    desired_mode = HWC_POWER_MODE_NORMAL;
+    display_timeout = 30*UI_FREQ;
+  } else {
+    display_timeout = std::min(display_timeout-1, 0);
```
deanlee, line 56: should be std::max

pd0wm:
Branch became impossible to rebase. Squashed all changes into single commit.

PR #2135, opened 2020-09-05
improve tici camerad robustness

ZwX1616:


selfdrive/camerad/cameras/camera_qcom2.c
```diff
@@ -950,31 +962,34 @@ void cameras_run(MultiCameraState *s) {
...
         if (event_data->session_hdl == s->rear.req_mgr_session_info.session_hdl) {
-          if (event_data->u.frame_msg.request_id > 0) {s->rear.first = false;}
+          if (event_data->u.frame_msg.request_id > -1) {s->rear.first = false;}
           s->rear.frame_id++;
+          //s->rear.n_skipped = 0;
           //printf("rear %d\n", s->rear.frame_id);
-          int buf_idx = s->rear.frame_id % FRAME_BUF_COUNT;
+          int buf_idx = (s->rear.frame_id - 1) % FRAME_BUF_COUNT;
+          //printf("frame_id/buf_idx 0 = %d / %d\n", s->rear.frame_id, buf_idx);
           s->rear.camera_bufs_metadata[buf_idx].frame_id = s->rear.frame_id;
```
robbederks, line 973: Aren't the frame_id's you put in wrong here since they're already incremented?

selfdrive/camerad/cameras/camera_qcom2.c
```diff
@@ -484,10 +484,10 @@ void enqueue_buffer(struct CameraState *s, int i) {
   if (s->buf_handle[i]) {
```
robbederks, line 484: Is there no better source for the `request_id` here? Seems kinda hacky to get it from the metadata...

robbederks:
I think the implementation still looks very hacky overall, I'm sure if we clean up the logic a bit it would be so much easier to follow and debug

ZwX1616:
> I think the implementation still looks very hacky overall, I'm sure if we clean up the logic a bit it would be so much easier to follow and debug

still testing some ideas - will do a final cleanup once everything works perfectly!

selfdrive/camerad/cameras/camera_qcom2.c
```diff
@@ -479,15 +479,15 @@ void config_isp(struct CameraState *s, int io_mem_handle, int fence, int request
...
     release(s->video0_fd, s->buf_handle[i]);
     // wait
-    // struct cam_sync_wait sync_wait = {0};
-    // sync_wait.sync_obj = s->sync_objs[i];
-    // sync_wait.timeout_ms = 100;
-    // ret = cam_control(s->video1_fd, CAM_SYNC_WAIT, &sync_wait, sizeof(sync_wait));
+    struct cam_sync_wait sync_wait = {0};
+    sync_wait.sync_obj = s->sync_objs[i];
+    sync_wait.timeout_ms = 50;
```
robbederks, line 489: Isn't 50ms too close? Since we're running at 20Hz?
Or do we want to skip the frame completely if it doesn't come in right in time?

ZwX1616: it's 200ms really, since we have 4 buffers each. 50 is kind of arbitrary here

ZwX1616: no wait is fine too, because by the time we call this it should be ready anyways (from the event).

ZwX1616:
there seems to be no more random frame skips with the performance cpu governor

PR #2136, opened 2020-09-06
Add ZSS support to openpilot (for Prius)

sshane:
Supporting PRs:
- cereal: https://github.com/commaai/cereal/pull/85
- opendbc: https://github.com/commaai/opendbc/pull/295

This adds support for ZSS without changing behavior for cars without it


selfdrive/car/toyota/carstate.py
```diff
@@ -51,12 +51,15 @@ def update(self, cp, cp_cam):
...
+    if self.accurate_steer_angle_seen or self.CP.hasZss:
+      if self.CP.hasZss:
+        ret.steeringAngle = cp.vl["SECONDARY_STEER_ANGLE"]['ZORRO_STEER'] - self.angle_offset
+      else:
+        ret.steeringAngle = cp.vl["STEER_TORQUE_SENSOR"]['STEER_ANGLE'] - self.angle_offset
 
       if self.needs_angle_offset:
         angle_wheel = cp.vl["STEER_ANGLE_SENSOR"]['STEER_ANGLE'] + cp.vl["STEER_ANGLE_SENSOR"]['STEER_FRACTION']
-        if abs(angle_wheel) > 1e-3 and abs(ret.steeringAngle) > 1e-3:
+        if (abs(angle_wheel) > 1e-3 and abs(ret.steeringAngle) > 1e-3) or self.CP.hasZss:
```
ShaneSmiskol, line 62: Not sure if this is required to specify `or hasZss`. My thinking is what if the user is on an older Prius without the more accurate sensor? We still want to offset the ZSS sensor

selfdrive/car/toyota/interface.py
```diff
@@ -38,7 +39,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), has_relay=False,
       ret.lateralTuning.init('indi')
       ret.lateralTuning.indi.innerLoopGain = 4.0
       ret.lateralTuning.indi.outerLoopGain = 3.0
-      ret.lateralTuning.indi.timeConstant = 1.0
+      ret.lateralTuning.indi.timeConstant = 0.1 if ret.hasZss else 1.0
```
ShaneSmiskol, line 42: People often change timeConstant and steerRateCost when ZSS is installed

pd0wm:
I don't want to endorse this sensor. Both the code and hardware are not even close to ASIL B, which I would expect for a sensor measuring steering angle. If the sensor fails openpilot will basically apply max torque left or right.

ShaneSmiskol:
Fair enough. Though is there a possibility for some code that could check the integrity of the ZSS angle and switch to the standard sensor if it's misaligned after initial offset? Something that once a discrepancy is detected, don't use ZSS for the rest of the drive.

pd0wm:
I would definitely add that to your own fork if you're using one of these sensors. I won't upstream that to openpilot, that sounds like a bunch of complexity (e.g. risk for bugs) for a very small amount of users.

ShaneSmiskol:
I understand the safety concerns, I just wanted to merge this for other users who had a ZSS. I'll close the PRs

PR #2151, opened 2020-09-10
Auto reset bad calibration

### LINKED ISSUE #1933: Add check for bad calibration
If a device suddenly moves the calibration will wrong. This is not checked and will cause bad driving behavior.

haraschax:
Auto reset calibration if a large spread is detected. Still needs to interact with controls correctly.

Also refactor param saving and increase save rate 5x to about once every minute.

### DIFF
--- a/selfdrive/locationd/calibrationd.py
+++ b/selfdrive/locationd/calibrationd.py
@@ -22,13 +22,13 @@ MIN_SPEED_FILTER = 15 * CV.MPH_TO_MS
 MAX_VEL_ANGLE_STD = np.radians(0.25)
 MAX_YAW_RATE_FILTER = np.radians(2)  # per second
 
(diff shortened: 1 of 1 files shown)


selfdrive/locationd/calibrationd.py
```diff
@@ -24,9 +24,11 @@
 MAX_YAW_RATE_FILTER = np.radians(2)  # per second
 
 # This is all 20Hz, blocks needed for efficiency
-BLOCK_SIZE = 100
+SMOOTH_CYCLES = 400  # 20 seconds
```
adeebshihadeh, line 27: use DT_MDL from common.realtime instead of hardcoding 20Hz

HaraldSchafer: Doesn't really depend on time as much as model cycles. Just removed references to time.

Issue #2154, opened 2020-09-11
Add Documentation to Readme for CARLA Docker


geohot:
Would be nice to have a very simple getting started with openpilot + CARLA on PC in the main README, with a link from the top of it.

I think this will help bring a lot of people into the codebase who don't have a car or comma hardware.

(looks like we already passed Apollo in stars, but more can't hurt)

pd0wm:
@grekiki Can you expand the `Testing on PC` section of the main readme with a simple example on how to start the two docker containers?

PR #2156, opened 2020-09-11
Update README.md

ghost:
Fixing https://github.com/commaai/openpilot/issues/2154

### DIFF
--- a/README.md
+++ b/README.md
@@ -285,7 +285,7 @@ curl https://raw.githubusercontent.com/commaai/openpilot/master/tools/sim/start_
 ```
 2) Start openpilot on second terminal
 ```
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Can we add two lines to copy/paste that run the start scripts without having to clone the repo? Ideally, someone should be able to stumble across openpilot, copy and paste two lines from the README, and see openpilot running on their computer.

grekiki:
> Can we add two lines to copy/paste that run the start scripts without having to clone the repo? Ideally, someone should be able to stumble across openpilot, copy and paste two lines from the README, and see openpilot running on their computer.

Hm like a docker container that would have the repository already preinstalled? User then just has to start the container

adeebshihadeh:
We already have that. We just need two lines that curl each script and run them instead of needing to clone openpilot.

grekiki:
I think that should work

PR #2162, opened 2020-09-13
more messagebuilder

deanlee:
worked a little more after https://github.com/commaai/openpilot/pull/2161

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -189,7 +189,8 @@ void usb_retry_connect() {
 void can_recv(PubMaster &pm) {
   // create message
   MessageBuilder msg;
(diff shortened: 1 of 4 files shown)


selfdrive/boardd/boardd.cc
```diff
@@ -187,12 +187,9 @@ void usb_retry_connect() {
...
 
 void can_recv(PubMaster &pm) {
-  uint64_t start_time = nanos_since_boot();
-
   // create message
   MessageBuilder msg;
   auto event = msg.initEvent();
-  event.setLogMonoTime(start_time);
 
   int recv = panda->can_receive(event);
```
adeebshihadeh, line 194: might as well make this:
int recv = panda->can_receive(msg.initEvent());

PR #2164, opened 2020-09-13
Missing fingerprints should have higher priority than canError

xps-genesis:
When mock car is detected, the canError can show up along with Dashcam mode. This makes troubleshooting with users very difficult. Dashcam mode, Car unrecognized should be shown even if there is a can error resulting of the missing fingerprint.


pd0wm:
Isn't the correct solution to make sure `CS.canValid` is True for the mock car? Then you still get the error in case there is no CAN data at all.

adeebshihadeh:
Fixed in ea06941d4d023b82eae8f9d342f972c4744353a1

PR #2166, opened 2020-09-14
Improve realtime performance on NEOS

adeebshihadeh:
After reproducing the boardd missed cycles while running ftrace, it seems the CPU 3 jitter is caused by unbound kernel workers doing work that disables preemption being scheduled onto CPU 3 between boardd cycles. Fixing `set_core_affinity` reduces the likelihood of boardd missed cycles since CPU 3 now spends at least 75% of its time in realtime processes.

We still need a way to shield CPU 3 from dynamically created unbound kworkers. Newer kernels expose a config option to do this in sysfs, `/sys/devices/virtual/workqueue/cpumask`.

As a test since the NEOS kernel doesn't have this option, a small waste process running as SCHED_FIFO with minimum realtime priority and pinned to CPU 3 prevents any boardd missed cycles. Boardd and controlsd will preempt the waste process, but the SCHED_OTHER kworker can't get scheduled onto CPU3.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -1,5 +1,6 @@
 Version 0.7.9 (2020-XX-XX)
 ========================
+ * NEOS update: improve realtime performance
  * Improved car battery power management
 
 Version 0.7.8 (2020-08-19)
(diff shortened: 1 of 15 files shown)


adeebshihadeh:
Merging the fix for setting affinity in python and the cpuset changes. Workqueue for CPU 3 will come with NEOS 15.

PR #2168, opened 2020-09-14
Refactor VisionStream

deanlee:
1. Convert VisionStream from struct to class 
2. Replace the double loop pattern with one function call: stream.acquire(), to complete all the work.


selfdrive/common/visionipc.c
```diff
@@ -119,19 +106,17 @@ int visionstream_init(VisionStream *s, VisionStreamType type, bool tbuffer, Visi
...
-      .d = { .stream_rel = {
-        .type = s->last_type,
-        .idx = s->last_idx,
-      }}
-    };
-    err = vipc_send(s->ipc_fd, &rep);
-    s->last_idx = -1;
-  }
+static int visionstream_release(VisionStream *s) {
+  VisionPacket rep = {
```
adeebshihadeh, line 110: why remove the check on last_idx here instead of the places where visionstream_release is called?

PR #2170, opened 2020-09-14
FileReader: cache files locally

ghost:
Cache downloaded files for faster acces next time

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,8 @@ selfdrive/modeld/_modeld
 selfdrive/modeld/_dmonitoringmodeld
 /src/
 
+tools/replay/tmp/
+
 one
 openpilot
 notebooks
--- a/.gitignore
+++ b/.gitignore
@@ -49,7 +49,7 @@ selfdrive/modeld/_modeld
 selfdrive/modeld/_dmonitoringmodeld
 /src/
 
-tools/replay/tmp/
+tools/replay/tmp/*
 
 one
 openpilot
--- a/.gitignore
+++ b/.gitignore
@@ -49,9 +49,6 @@ selfdrive/modeld/_modeld
 selfdrive/modeld/_dmonitoringmodeld
 /src/
 
-tools/replay/tmp/*
-!tools/replay/tmp/.gitkeep
-
 one
 openpilot
 notebooks
--- a/.gitignore
+++ b/.gitignore
@@ -50,6 +50,7 @@ selfdrive/modeld/_dmonitoringmodeld
 /src/
 
 tools/replay/tmp/*
+!tools/replay/tmp/.gitkeep
 
 one
 openpilot
--- a/tools/lib/tests/test_caching.py
+++ b/tools/lib/tests/test_caching.py
@@ -7,43 +7,55 @@ class TestFileDownload(unittest.TestCase):
 
   def test_downloads(self):
     small_file_url = "https://raw.githubusercontent.com/commaai/openpilot/master/SAFETY.md"
-
-    #Load the end 100 bytes of both files
-    file_small_cached2 = URLFile(small_file_url)
-    file_small_download2 = URLFile(small_file_url, download=True)
-    self.assertEqual(file_small_cached2.get_length(),file_small_download2.get_length())
-    length = file_small_download2.get_length()
-    #print(length)
-    file_small_cached2.seek(length-100)
-    file_small_download2.seek(length-100)
-    self.assertEqual(file_small_cached2.read(),file_small_download2.read())
     
     #Load full small file
     file_small_cached = URLFile(small_file_url)
-    file_small_download = URLFile(small_file_url, download=True)
-    self.assertEqual(file_small_cached.get_length(),file_small_download.get_length())
-    self.assertEqual(file_small_cached.read(),file_small_download.read())
+    file_small_download = URLFile(small_file_url, cache=False)
+    self.assertEqual(file_small_cached.get_length(), file_small_download.get_length())
+    self.assertEqual(file_small_cached.read(), file_small_download.read())
     large_file_url = "https://commadataci.blob.core.windows.net/openpilotci/0375fdf7b1ce594d/2019-06-13--08-32-25/3/fcamera.hevc"
-    
+
+    #Load the end 100 bytes of both files
+    file_small_cached2 = URLFile(small_file_url)
+    file_small_download2 = URLFile(small_file_url, cache=False)
+    self.assertEqual(file_small_cached2.get_length(), file_small_download2.get_length())
+    length = file_small_download2.get_length()
+    #print(length)
+    file_small_cached2.seek(length - 100)
+    file_small_download2.seek(length - 100)
+    response_cached=file_small_cached2.read()
+    response_download=file_small_download2.read()
+    print(response_cached)
+    print(response_download)
+    self.assertEqual(response_cached, response_download)
+
+    #Load small file 100 bytes at a time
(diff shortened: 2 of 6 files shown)


.gitignore
```diff
@@ -49,6 +49,9 @@ selfdrive/modeld/_modeld
 selfdrive/modeld/_dmonitoringmodeld
 /src/
 
+tools/replay/tmp/*
```
pd0wm, line 52: Use `/tmp`

pd0wm:
As mentioned. This needs to be cleanup up, and it needs to be really solid since it's used by all offline tools. Also needs tests.

tools/lib/tests/test_caching.py
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env python3
+
+import unittest
+from tools.lib.url_file import URLFile
+
+class TestFileDownload(unittest.TestCase):
+
+  def test_downloads(self):
+    small_file_url = "https://raw.githubusercontent.com/commaai/openpilot/master/SAFETY.md"
```
adeebshihadeh, line 9: ideally the test shouldn't need internet. you can make a temp file, fill it with junk bytes, and refer to it like this `file:///path/to/the/tmp/file`

pd0wm: The tests also need to check the partial download functionality using range headers. If we want to run locally we need something like flask. However we already ran into some issues with azure implementing the range headers in a non standard way.

grekiki: Yeah not sure there is any benefit from not needing internet, there are some bugs in previous version that are imo not reproducible in local testing, for example getting the whole file if the file is large enough and the end of range is not specified

grekiki:
Despite all checks working I maybe still perhaps want to run the branch on a device, just to make sure the master is not broken in some hard to find way?

adeebshihadeh:
These tools aren't meant to run on devices. Can you revert the changes to the submodules?

grekiki:
Yup as soon as I figure out how :)

tools/lib/tests/test_caching.py
gregjhogan: can we change it to a qlog file to make the egress bandwidth as small as possible?
also azure storage account emulator (that runs in docker): https://github.com/Azure/Azurite

pd0wm (changes requested):
General comments:
 - If we want to download more than a chunk (say the whole file), it still gets downloaded in smaller chunks. Which could mean a lot of round trips to the server. If we request a range larger than a chunk size, just round it up, download the whole thing and then write it out to multiple cache files.
  - Only use the threadlocal curl. The get_curl function was for my multithreaded implementation. That can be removed.
 - Remove commented out print statements. In general keep the code clean. This code will be read by a lot of people in the future.
 - Fix submodules.

tools/lib/url_file.py
```diff
@@ -33,10 +39,67 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+      file_name = sha256((self._url.split("?")[0] + " " + str(increment)).encode('utf-8')).hexdigest()
+      full_path = "/tmp/" + str(file_name)
+      #If we don't have a file, download it
+      if not os.path.exists(full_path):
+        #print("Downloading")
+        data = self.read_aux(ll = CACHE_SIZE)
+        with open(full_path, "wb") as new_cached_file:
+          new_cached_file.write(data)
+
+      with open(full_path, "rb") as cached_file:
```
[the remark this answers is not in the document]
grekiki: I thought that would be nicer code, will change

tools/lib/url_file.py
```diff
@@ -34,9 +47,65 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    if self._length is not None:
+      return self._length
+    path_to_name = PATH + LENGTHS_SUBFOLDER + str(sha256((self._url.split("?")[0]).encode('utf-8')).hexdigest()) + "_length"
+    if os.path.exists(path_to_name):
+      with open(path_to_name, "r") as file_length:
+          content = file_length.read()
+          self._length = int(content)
+          return self._length
+
+    c = pycurl.Curl()
```
pd0wm, line 60: Reuse `self._curl`

tools/lib/url_file.py
```diff
@@ -34,9 +47,65 @@ def __exit__(self, exc_type, exc_value, traceback):
       self._local_file = None
 
   @retry(wait=wait_random_exponential(multiplier=1, max=5), stop=stop_after_attempt(3), reraise=True)
+  def get_length(self):
+    if self._length is not None:
+      return self._length
+    path_to_name = PATH + LENGTHS_SUBFOLDER + str(sha256((self._url.split("?")[0]).encode('utf-8')).hexdigest()) + "_length"
```
pd0wm, line 53: - Don't put lengths in a subfolder
 - Use `os.path.join` to add subpath folders together.

tools/lib/url_file.py
```diff
@@ -6,23 +6,36 @@
...
+K=1000
+CHUNK_SIZE=1000 * K
 
+PATH="/tmp/comma_download_cache/"
+LENGTHS_SUBFOLDER="lengths/" 
 class URLFile(object):
   _tlocal = threading.local()
 
-  def __init__(self, url, debug=False):
+  def __init__(self, url, debug=False, cache=True):
```
pd0wm, line 22: - Cache needs to be off by default
 - Also have env variable to turn on caching, overriding this parameter

tools/lib/url_file.py
pd0wm: Abstract url hashing into method. It's repeated below

tools/lib/url_file.py
```diff
@@ -34,9 +47,65 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    while True:
+      self._pos = position
+      chunk_number = self._pos / CHUNK_SIZE
+      file_name = str(sha256((self._url.split("?")[0]).encode('utf-8')).hexdigest()) + "_" + str(chunk_number)
+      full_path = PATH + str(file_name)
+      data = None
+      #If we don't have a file, download it
+      if not os.path.exists(full_path):
+        data = self.read_aux(ll=CHUNK_SIZE)
+        with open(full_path, "wb") as new_cached_file:
```
pd0wm, line 92: use `atomic_write_in_dir`

tools/lib/url_file.py
```diff
@@ -34,9 +47,65 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+    c.setopt(pycurl.NOSIGNAL, 1)
+    c.setopt(pycurl.TIMEOUT_MS, 500000)
+    c.setopt(pycurl.FOLLOWLOCATION, True)
+    c.setopt(pycurl.URL, self._url)
+    c.setopt(c.NOBODY, 1)
+    c.perform()
+
+    length = int(c.getinfo(c.CONTENT_LENGTH_DOWNLOAD))
+    self._length = length
+    with open(path_to_name, "w") as file_length:
```
pd0wm, line 70: use `atomic_write_in_dir`

tools/lib/url_file.py
```diff
@@ -6,23 +6,36 @@
...
     self._local_file = None
     self._debug = debug
-
+    #We download if we cannot cache 
+    self._force_download = not cache
     try:
       self._curl = self._tlocal.curl
     except AttributeError:
       self._curl = self._tlocal.curl = pycurl.Curl()
+    if not os.path.isdir(PATH):
```
pd0wm, line 34: use `mkdirs_exists_ok`

PR #2173, opened 2020-09-14
UI: Release resources before exiting

deanlee:


selfdrive/ui/paint.cc
```diff
@@ -821,3 +820,26 @@ void ui_nvg_init(UIState *s) {
     s->priv_hnds[i] = NULL;
   }
 }
+
+void ui_nvg_destroy(UIState *s) {
+  nvgDeleteImage(s->vg, s->img_wheel);
```
adeebshihadeh, line 825: let's put all the images in a map or something

PR #2179, opened 2020-09-16
fix memory leak in visionbuf_to_cl

deanlee:
1.buf_cl is not assigned in visionbuf_to_cl(), that will cause memory leak.
2.clReleaseMemObject() should be called only if needed


deanlee:
fixed in https://github.com/commaai/openpilot/pull/2310

PR #2183, opened 2020-09-16
Tici hardware abstraction layer

pd0wm:


common/hardware.py
```diff
@@ -50,5 +51,7 @@ def get_network_strength(self, network_type):
 
 if EON:
   HARDWARE = cast(HardwareBase, Android())
```
adeebshihadeh, line 53: elif, otherwise android is broken

common/hardware.py
```diff
@@ -50,5 +51,7 @@ def get_network_strength(self, network_type):
...
+if TICI:
```
adeebshihadeh, line 54: elif, otherwise android is broke

pd0wm: oops

common/hardware_tici.py
```diff
@@ -0,0 +1,51 @@
...
+from cereal import log
+
+
+NetworkType = log.ThermalData.NetworkType
+NetworkStrength = log.ThermalData.NetworkStrength
+
+
+class Tici(HardwareBase):
+  @staticmethod
+  def get_cmdline():
```
adeebshihadeh, line 12: not specific to tici, maybe move to hardware base class?

PR #2194, opened 2020-09-17
Realtime shield

adeebshihadeh:


geohot:
Isn't this going to just waste CPU?

yufeng66:
I am tired of the model lag error so I switched to the master-ci branch after this pull request is merged. I haven't experience any more model lag error since. Also, after 0.7.7, I sometime felt comma c2 turns the car slightly late. That got fixed with master-ci as well. Not so sure if it is related to this real time issue as well

PR #2197, opened 2020-09-18
modeld: eliminate the memory copy to GPU

deanlee:


pd0wm:
Is it not required to sync the buffers? What device did you test this on?

I think it might work when it's backed by an ion buffer (but no guarantees because the caches are not properly invalidated), or buffers allocated with `clSVMAlloc`. If you'd try on a device with openCL 1.2 it probably won't work. Did I miss something?

deanlee:
I think the buffer is already synchronized. tested on PC, C1&C2(LeEco Le Pro3) with openCL 1.2&2.0.have driven thousands of kilometers,But I don’t have enough tools to do a very complete test.

pd0wm:
It probably works because when using ion buffers/cpu buffers we don't touch the buffer from the cpu side so it doesn't matter that we clean the cache.

We also explicity ask for a CPU device for openCL, therefore we don't notice that the buffer is not copied to the GPU if needed. This is probably also why it doesn't work right now if you try to use the GPU.

I suggest we add a `visionbuf_sync(buf, VISIONBUF_SYNC_TO_DEVICE)` after the call to `visionstream_get`.

pd0wm:
I'm seeing now that adding the sync is not trivial, since we don't have a VisionBuf object on the receiving side and it's not entirely trivial to construct it again. We can do without the sync for now, and I'll include it in the visionipc rewrite.

PR #2198, opened 2020-09-18
commonmodel: use clEnqueueWriteBuffer

deanlee:
use clEnqueueWriteBuffer instead of clEnqueueMapBuffer+memcpy+clEnqueueUnmapMemObject

I think this may be simpler and better performance.(not very sure)


adeebshihadeh:
This will be addressed in #19521

PR #2203, opened 2020-09-19
driving.cc: Use circular buffer to reduce memory copy

deanlee:
Save 768kb of memory copy per cycle


deanlee:
It works.but the code is pretty ugly, I can't figure out how to simplify it.:stuck_out_tongue_closed_eyes: any suggestions are welcome

PR #2204, opened 2020-09-20
uploader: Use cache to find the next file

deanlee:
Uploader will search all directories to get the next file to be uploaded,that's not cheap.


grekiki:
On some custom testing with every file being already uploaded I got 
```
  File "uploader.py", line 256, in uploader_fn
    idx, d = uploader.next_file_to_upload(with_raw=allow_raw_upload and should_upload)
TypeError: cannot unpack non-iterable NoneType object

```

grekiki:
Also there might be issues when the device changes the timezone

deanlee:
My bad...thanks for your review and time! 
I'll close this PR and rethink  issues you mentioned

PR #2207, opened 2020-09-20
Params: use a multiple-reader / single-writer flock to improve concurrency

deanlee:


pd0wm:
Is it necessary to take the lock before reading? The python implementation of the params doesn't do this (only when reading all the values, I assume to get a consistent view).

deanlee:
yes,we should prevent reading of the file while it's being modified or deleted. I'll update this PR to fix that too.

deanlee:
My bad, I did not read the code carefully.the current implementation does not require locks for reading

pd0wm:
I think the lock on line 288 in params.cc can be removed completely. The rest looks fine!

PR #2214, opened 2020-09-21
Uploader speedup

ghost:
Did some benchmarking and improved the uploader. I think it should be about 3 times faster.


selfdrive/loggerd/tools/mark_all_uploaded.py
```diff
@@ -0,0 +1,9 @@
+import os
+from common.xattr import setxattr
+from selfdrive.loggerd.uploader import UPLOAD_ATTR_NAME, UPLOAD_ATTR_VALUE
+
+PATH = "/data/media/0/realdata/"
+for folder in os.walk(PATH):
+    for file1 in folder[2]:
```
adeebshihadeh, line 7: two spaces

selfdrive/loggerd/tools/mark_all_uploaded.py
```diff
@@ -0,0 +1,9 @@
+import os
+from common.xattr import setxattr
+from selfdrive.loggerd.uploader import UPLOAD_ATTR_NAME, UPLOAD_ATTR_VALUE
+
+PATH = "/data/media/0/realdata/"
+for folder in os.walk(PATH):
```
adeebshihadeh, line 6:
```suggestion
from selfdrive.loggerd.config import ROOT
for folder in os.walk(ROOT):
```

selfdrive/loggerd/uploader.py
```diff
@@ -236,14 +237,19 @@ def uploader_fn(exit_event):
...
   backoff = 0.1
+  counter = 0
+  should_upload = False
   while not exit_event.is_set():
     offroad = params.get("IsOffroad") == b'1'
     allow_raw_upload = (params.get("IsUploadRawEnabled") != b"0") and offroad
-    on_hotspot = is_on_hotspot()
-    on_wifi = is_on_wifi()
-    should_upload = on_wifi and not on_hotspot
+    check_network = (counter % 12 == 0 if offroad else True)
```
gregjhogan, line 245: If I am on cellular and only have raw files to upload, then I get on wifi it can take 12 minutes now for it to start uploading raw files (when offroad)?  Seems like this counter shouldn't be necessary when sleeping for 60 seconds?

Omeryl: @gregjhogan Yeah, I noticed this too, I thought it was being a bit weird until I saw this PR. Is this intentional?

grekiki: I think this part of the code is updated on master, so it shouldn't be an issue.

Omeryl: @grekiki Ah, has it yet to hit a release? I'll go take a look then.

PR #2216, opened 2020-09-22
fix dmonitoringmodeld random high CPU usage

### LINKED ISSUE #2132: dmonitoringmodeld high CPU usage
Normally, dmonitoringmodeld uses 3% of a core, but occasionally it uses up to 25% of a core http://openpilot-ci.internal:8080/blue/organizations/jenkins/openpilot/detail/master/193/pipeline

ZwX1616:
to fix #2132 
see comments

### DIFF
--- a/selfdrive/manager.py
+++ b/selfdrive/manager.py
@@ -293,6 +293,8 @@ def nativelauncher(pargs, cwd):
 def start_managed_process(name):
   if name in running or name not in managed_processes:
     return
+  if name == 'dmonitoringmodeld':
+    time.sleep(10)
   proc = managed_processes[name]
   if isinstance(proc, str):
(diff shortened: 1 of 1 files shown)


selfdrive/test/test_cpu_usage.py
```diff
@@ -101,6 +101,8 @@ def test_cpu_usage():
 
   passed = False
   try:
-    passed = test_cpu_usage()
+    # TODO: remove loop
+    for _ in range(200):
+      passed = test_cpu_usage()
```
adeebshihadeh, line 106: this doesn't do what you want. you want to exit immediately on failure. the exit status will only reflect the last test, and you'll have to scroll through 200 tests to verify they all passed

PR #2219, opened 2020-09-23
Implement skeleton for logging capabilities using athena

Phat3:
Add skeleton code to support logging through Athena instead of using LogEntries

### DIFF
--- a/selfdrive/athena/athenad.py
+++ b/selfdrive/athena/athenad.py
@@ -29,13 +29,9 @@ from common.realtime import sec_since_boot
 from selfdrive.loggerd.config import ROOT
 from selfdrive.swaglog import cloudlog
 
-# FIXME: For local development
-# ATHENA_HOST = os.getenv('ATHENA_HOST', 'wss://athena.comma.ai:8765')
-ATHENA_HOST = os.getenv('ATHENA_HOST', 'ws://athena.comma.ai:8765')
+ATHENA_HOST = os.getenv('ATHENA_HOST', 'wss://athena.comma.ai:8765')
 HANDLER_THREADS = int(os.getenv('HANDLER_THREADS', "4"))
 LOCAL_PORT_WHITELIST = set([8022])
-# TODO: Decice the proper queue size (usually the highest number of logs
-# created on openpilot is around 1300 a minute)
 MAX_LOG_QUEUE_SIZE = 100000
 LOG_RETHREIVE_UPLOAD_THRESHOLD = 3
 BATCH_LOG_SIZE = 100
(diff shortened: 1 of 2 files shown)


selfdrive/athena/athenad.py
```diff
@@ -317,6 +358,12 @@ def main():
   api = Api(dongle_id)
 
   conn_retries = 0
+
+  # Set up zmq socket
+  ctx = zmq.Context().instance()
+  sock = ctx.socket(zmq.PULL)
+  sock.bind("ipc:///tmp/logmessage")
```
gregjhogan, line 365: can zmq get disconnected (and we need to reconnect)?

selfdrive/athena/athenad.py
```diff
@@ -101,6 +109,11 @@ def _do_upload(upload_item):
                         headers={**upload_item.headers, 'Content-Length': str(size)},
                         timeout=10)
 
+def _clear_log_batch_retry():
+  global current_num_retrieve_upload_log_batch
+  global last_uploaded_log_batch
```
gregjhogan, line 114: is this necessary?

selfdrive/athena/athenad.py
```diff
@@ -138,6 +151,20 @@ def do_reboot():
 
   return {"success": 1}
 
+@dispatcher.add_method
+def pullLogs(prev_success):
+  # if the upload of the previous batch didn't finish correcly then we
+  # reupload the same batch
+  global current_num_retrieve_upload_log_batch
+  global last_uploaded_log_batch
```
gregjhogan, line 159: is this necessary?

selfdrive/athena/athenad.py
```diff
@@ -28,25 +29,32 @@
 from selfdrive.loggerd.config import ROOT
 from selfdrive.swaglog import cloudlog
 
-ATHENA_HOST = os.getenv('ATHENA_HOST', 'wss://athena.comma.ai')
+ATHENA_HOST = os.getenv('ATHENA_HOST', 'wss://athena.comma.ai:8765')
```
gregjhogan, line 32: should not have changed

selfdrive/athena/athenad.py
Phat3: If we want to keep using those 2 global variables then we need the `global` keyword otherwise python will just create 2 local variables within the scope of the function therefore any modification to them would not be reflected on the global variables

selfdrive/athena/athenad.py
Phat3: Same as above

selfdrive/athena/athenad.py
Phat3: Ops, it's a leftover of a fix that I introduce to allow local development. I don't have push access anymore so I cannot push the fix on this branch

selfdrive/athena/athenad.py
Phat3: True, it might be possible. For the same reason, we should also fix this problem in `selfdrive/swaglog,py#22` (as it is now if the socket disconnects then all the log messages will be dropped silently)

PR #2220, opened 2020-09-23
longcontrol.py cleanup, remove unused globals, use self.reset()

hewers:
**Description** Unused file scope globals, and use longcontrol.reset(velocity) to reset PID and set target

**Verification** CI success, I've used this longcontrol for ~1000 miles.


qadmus:
Can remove the longcontrol.reset() change if preferred!

PR #2221, opened 2020-09-23
add delay to acc pedal alert

cydia2020:
So that the screen wouldn't appear to "flicker" when openpilot is engaged while the acc pedal is being released


pd0wm:
I think this would mess with the state machine in controlsd. The creation delay would cause it to go into the enable immediately instead of the pre enabled state.

@adeebshihadeh am I correct? Is there an elegant way we can make this change without a second alert? I support this change since it was bothering me too.

cydia2020:
Yeah tried it w/ my fork and it doesn't work 100% of the time, sometimes the alert still gets created immediately

adeebshihadeh:
No, creation_delay only delays the alert. The events (what controlsd looks at) are always there still.

PR #2222, opened 2020-09-23
tici: fix ui rotation

pd0wm:


ZwX1616:
iirc the frame looked stretched due to some hand coded aspect ratio assumption somewhere in paint.cc. can you also fix that?

pd0wm:
Yes, will check that out too.

PR #2225, opened 2020-09-25
Fix grade force in test_long plant

hewers:
**Description** Use grade percent, where 1.0 == 100% == 45 degrees uphill, "rise over run".
Want actual percent grade units?
Close PR if we want no trig functions, simple "effective grade".
https://en.wikipedia.org/wiki/Grade_(slope)

**Verification** Output of longitudinal tests with non-zero grades looks good.


qadmus:
F_grade = - grade * mass
image

F_grade = - sin(arctan(grade) * mass * gravity
image

Can close if (-grade * mass) is a fine approximation!

pd0wm:
Not sure why we specified grade as `grade * g`. Thanks for fixing.

PR #2236, opened 2020-09-28
Merging pandad into boardd

### LINKED ISSUE #2102: Merge pandad into boardd

LoganxDev:
**Description**
Same as title

**Verification**
Built and ran on Comma2, I don't own an external panda for testing, no functionality was changed though, only moved to boardd.

Closes #2102


adeebshihadeh:
Thanks for the PR, but this isn’t the actual board daemon that runs with openpilot. This was supposed to extend the functionality of the C++ panda library so the pandad functionality could be moved to boardd.cc.

LoganxDev:
Oh yikes that's embarrassing! I see now though.

PR #2242, opened 2020-09-29
updated: log git corruption

adeebshihadeh:


selfdrive/updated.py
```diff
@@ -201,6 +201,11 @@ def finalize_update() -> None:
     shutil.rmtree(FINALIZED)
   shutil.copytree(OVERLAY_MERGED, FINALIZED, symlinks=True)
 
+  # Log git repo corruption
+  fsck = run(["git", "fsck", "--no-progress"], FINALIZED).rstrip()
```
pd0wm, line 205: Won't git fsck throw a CalledProcessError if it returns with a non zero return code?

adeebshihadeh: I haven’t seen it exit with something non-zero, and the docs don’t specify any other return codes. We can keep an eye on this and see if it ever happens while it’s in master.

PR #2245, opened 2020-09-29
CI: Add valgrind

ghost:
Will try to add some valgrind tests to CI


selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,102 @@
...
+else:
+  from tqdm import tqdm   # type: ignore
+
+import cereal.messaging as messaging
+from collections import namedtuple
+from tools.lib.logreader import LogReader
+
+ProcessConfig = namedtuple('ProcessConfig', ['proc_name', 'pub_sub', 'ignore', 'command', 'path'])
+
+BASE_URL = "https://commadataci.blob.core.windows.net/openpilotci/"
```
pd0wm, line 19: Import this from `sefldrive/test/openpilotci`

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,102 @@
...
+    pub_sub={
+      "ubloxRaw": ["ubloxGnss", "gpsLocationExternal"],
+    },
+    ignore=[],
+    command="./ubloxd & sleep 20; kill $!",
+    path="../locationd",
+  ),
+]
+
+class SimplePubMaster():
```
pd0wm, line 33: This is exactly the same as the real pubmaster. Just use that.

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,102 @@
...
+BASE_URL = "https://commadataci.blob.core.windows.net/openpilotci/"
+
+CONFIGS = [
+  ProcessConfig(
+    proc_name="ubloxd",
+    pub_sub={
+      "ubloxRaw": ["ubloxGnss", "gpsLocationExternal"],
+    },
+    ignore=[],
+    command="./ubloxd & sleep 20; kill $!",
```
pd0wm, line 28: Let the thing that starts the process deal with the killing

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,102 @@
...
+  def valgrindlauncher(self, arg, cwd):
+    os.chdir(cwd)
+
+    # Run valgrind on a process
+    command = "valgrind --leak-check=full " + arg
+
+    p = subprocess.Popen(command, stdout=subprocess.PIPE, stderr=subprocess.PIPE, shell=True)
+    _, err = p.communicate()
+    error_msg = str(err, encoding='utf-8')
+    err_lost1 = error_msg.split("definitely lost: ")[1]
```
pd0wm, line 54: Isn't a regex cleaner to parse the result?

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,102 @@
...
+    definitely_lost_amount = int(err_lost1.split(" ")[0])
+    indirectly_lost_amount = int(err_lost2.split(" ")[0])
+    possibly_lost_amount = int(err_lost3.split(" ")[0])
+    if max(definitely_lost_amount, indirectly_lost_amount, possibly_lost_amount) > 0:
+      self.leak = True
+      return
+    self.leak = False
+
+  def replay_process(self, cfg, lr):
+    pub_sockets = [s for s in cfg.pub_sub.keys() if s != 'can']  # We dump data from logs here
```
pd0wm, line 66: Why do we treat `can` special here?

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,102 @@
...
+    print("Sorting logs")
+    all_msgs = sorted(lr, key=lambda msg: msg.logMonoTime)
+    pub_msgs = [msg for msg in all_msgs if msg.which() in list(cfg.pub_sub.keys())]
+
+    thread = threading.Thread(target=self.valgrindlauncher, args=(cfg.command, cfg.path))
+    thread.daemon = True
+    thread.start()
+    time.sleep(5)  # We give the process time to start
+    for msg in tqdm(pub_msgs):
+      pm.send(msg.which(), msg.as_builder())
```
pd0wm, line 78: You have to wait for a result here. This will run *way* faster than the receiving process can handle the message. The receiving often uses `conflate` which means it has a receive buffer of size 1 that only contains the latest message.

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,102 @@
...
+
+  def get_segment(self, segment_name, original=True):
```
pd0wm, line 80: Don't copy paste code. Import from test_processes.py

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,102 @@
...
+      process_replay_dir = os.path.dirname(os.path.abspath(__file__))
+      model_ref_commit = open(os.path.join(process_replay_dir, "model_ref_commit")).read().strip()
+      rlog_url = BASE_URL + "%s/%s/rlog_%s.bz2" % (route_name.replace("|", "/"), segment_num, model_ref_commit)
+
+    return rlog_url
+
+  def test_config_0(self):
+    cfg = CONFIGS[0]
+
+    URL = self.get_segment("0375fdf7b1ce594d|2019-06-13--08-32-25--3")
```
pd0wm, line 94: Define segment to use in the process config.

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,102 @@
...
+    lr = LogReader(URL)
+    self.replay_process(cfg, lr)
+    # Wait for the replay to complete
+    time.sleep(30)
+    assert not self.leak
```
pd0wm, line 99: use unittest asserts (`self.assertFalse()`)

pd0wm (dismissed):
- After sending messages you need to wait for a result from the process under test to verify that it's done with that cycle
 - Don't change submodules.

grekiki:
Finally learnt to deal with subrepositories...

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,90 @@
...
+
+CONFIGS = [
+  ProcessConfig(
+    proc_name="ubloxd",
+    pub_sub={
+      "ubloxRaw": ["ubloxGnss", "gpsLocationExternal"],
+    },
+    ignore=[],
+    command="./ubloxd",
+    path="../locationd",
```
pd0wm, line 27: with reference to `BASEDIR`

selfdrive/test/test_valgrind_replay.py
```diff
@@ -0,0 +1,90 @@
...
+    segment="0375fdf7b1ce594d|2019-06-13--08-32-25--3",
+    wait_for_response=True
+  ),
+]
+
+class TestValgrind(unittest.TestCase):
+
+  def extract_leak_sizes(self, log):
```
adeebshihadeh, line 35: this seems more complicated than it needs to be. can't we just check the valgrind exit code and use it's report if it fails?

grekiki: I get the sizes of leaks here, and also what type of leak they are. The file is intended to also be usable locally, not just in CI

grekiki:
I think it should be good for merging now

PR #2250, opened 2020-10-01
Tune steer actuator delay for TSS1 Corolla

sshane:
Measured this in Cabana, I noticed that the wheel turns slightly too late for curves but this makes it turn more in time.

https://my.comma.ai/cabana/?route=e010b634f3d65cdb%7C2020-09-30--22-49-10&exp=1633072042&sig=2G03Ce9JH%2B03peCFT%2FeMgFLwH13ADkJBNNGJicJVn5o%3D&max=20&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2Fe010b634f3d65cdb%2Fff11d8e209b975497c959445840c8f6e_2020-09-30--22-49-10


pd0wm:
I strongly doubt you can notice a 20 ms delay. You can try the same drive a couple times with this script running: https://github.com/commaai/openpilot/blob/master/selfdrive/debug/internal/measure_steering_accuracy.py That should tell you if there is a statistically significant difference between the two.

ShaneSmiskol:
I ran the file with both delays and found it's a bit better at higher angles 
Screenshot_120

gregjhogan:
make sure you drive the same road in the same conditions (same direction, lane, speed, wind, etc...) for each data set and you probably need thousands of data points for each angle to have meaningful numbers

that script could also be improved quite a bit, for example better statistical analysis and once you hit max torque you will have high error, but that data isn't currently excluded

ShaneSmiskol:
I did, same exact route of about 5 miles. There was some traffic so I was going slower on one of the routes for about 30 seconds. This isn't what I felt, but what I measured in Cabana, but definitely felt a slight improvement in turning in time with the curve

gregjhogan:
I have personally had a lot of feelings that turned out to be wrong, so I would be careful drawing conclusions from feelings.

The data you collected looks like no improvement from 0 - 2 degrees, and there are not enough samples to draw conclusions from higher angles

ShaneSmiskol:
I understand what you're saying. So would 5 miles equivalent at each angle (at least 2 or 3 thousand) be enough data for steering tuning changes?

gregjhogan:
My statistical analysis skills are weak, maybe MSE would be better?  I will defer to others on ways to better analyze the data.  Also note that this script was set up to run against replays of drives, too.

PR #2252, opened 2020-10-01
openpilot v0.7.9 release

pd0wm:
Version 0.7.9 (2020-10-09)
========================
 * Improved car battery power management
 * Improved updater robustness
 * Improved realtime performance
 * Reduced UI and modeld lags
 * Increased torque on 2020 Hyundai Sonata and Palisade


SConstruct
```diff
@@ -176,6 +183,44 @@ env = Environment(
   ]
 )
 
+qt_env = None
```
adeebshihadeh, line 186: @pd0wm any reason this is here instead of `selfdrive/ui/SConscript`?

pd0wm: I assume we will have more qt binaries for things like spinner, installer, etc..

PR #2258, opened 2020-10-03
Present alert if fan isn't spinning

LoganxDev:
When a fan is spinning at 0 RPM but requested to spin at 10+% power, display a warning.

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -61,8 +61,8 @@ class Controls:
       self.can_sock = messaging.sub_sock('can', timeout=can_timeout)
 
     # wait for one health and one CAN packet
-    hw_type = messaging.recv_one(self.sm.sock['health']).health.hwType
-    has_relay = hw_type in [HwType.blackPanda, HwType.uno, HwType.dos]
+    self.hw_type = messaging.recv_one(self.sm.sock['health']).health.hwType
+    has_relay = self.hw_type in [HwType.blackPanda, HwType.uno, HwType.dos]
     print("Waiting for CAN messages...")
     get_one_can(self.can_sock)
 
@@ -123,6 +123,7 @@ class Controls:
     self.distance_traveled = 0
     self.events_prev = []
     self.current_alert_types = [ET.PERMANENT]
+    self.last_desired_fan_speed = 0
 
     self.sm['liveCalibration'].calStatus = Calibration.CALIBRATED
     self.sm['thermal'].freeSpace = 1.
@@ -140,7 +141,7 @@ class Controls:
       self.events.add(EventName.communityFeatureDisallowed, static=True)
     if not car_recognized:
       self.events.add(EventName.carUnrecognized, static=True)
-    if hw_type == HwType.whitePanda:
+    if self.hw_type == HwType.whitePanda:
       self.events.add(EventName.whitePandaUnsupported, static=True)
 
     # controlsd is driven by can recv, expected at 100Hz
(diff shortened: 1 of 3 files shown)


pd0wm:
- fanSpeedRpm is always zero on EON. This alert can only be implemented on comma twos.
 - controlsd is already subscribing to both health and thermal, I think the logic can be implemented solely in controlsd
 - I would choose the threshold to 50% desired fan speed. If the fan is broken it will quickly reach that since the device will heat up. This will prevent false positives due to stuff like low car battery voltage (the fan is driven by the 12V input).
 - The event should be a green permanent alert (like the no gps alert).

LoganxDev:
I'm guessing `HwType::dos` is the comma two, is it safe to ignore `HwType::uno`?

I don't see where to set an alert color, I've set it to `ET.PERMANENT` though.

pd0wm:
Both `uno`, and `dos` have fan speed measurement, so you can trigger on both.

selfdrive/controls/controlsd.py
```diff
@@ -171,6 +172,11 @@ def update_events(self, CS):
     if self.sm['thermal'].memUsedPercent > 90:
       self.events.add(EventName.lowMemory)
 
+    if self.hw_type in [HwType.uno, HwType.dos]:
+      if self.sm['health'].fanSpeedRpm == 0 and self.last_desired_fan_speed > 50:
+        self.events.add(EventName.fanMalfunction)
+      self.last_desired_fan_speed = self.sm['thermal'].fanSpeed
```
pd0wm, line 178: Why store last_desired fan speed? Controlsd runs at 100 Hz, so that wouldn't be enough time for the fan to spin up + the health packet being received (500 ms).

You probably want to make sure that fan speed is above 50% for at least a second.

selfdrive/controls/lib/events.py
```diff
@@ -469,6 +469,14 @@ def wrong_car_mode_alert(CP: car.CarParams, sm: messaging.SubMaster, metric: boo
       Priority.LOW, VisualAlert.steerRequired, AudibleAlert.chimePrompt, 1., 2., 3.),
   },
 
+  EventName.fanMalfunction: {
+    ET.PERMANENT: Alert(
+      "Device Fan Malfunction",
+      "Risk of overheating",
```
adeebshihadeh, line 475:
```suggestion
      "Fan Malfunction",
      "Reboot your Device",
```

selfdrive/controls/controlsd.py
```diff
@@ -61,8 +61,8 @@ def __init__(self, sm=None, pm=None, can_sock=None):
       self.can_sock = messaging.sub_sock('can', timeout=can_timeout)
 
     # wait for one health and one CAN packet
-    hw_type = messaging.recv_one(self.sm.sock['health']).health.hwType
-    has_relay = hw_type in [HwType.blackPanda, HwType.uno, HwType.dos]
+    self.hw_type = messaging.recv_one(self.sm.sock['health']).health.hwType
```
adeebshihadeh, line 64: not necessary, you're already reading the health packet with the other logic

selfdrive/controls/controlsd.py
LoganxDev: Ah okay, I'll store the last time the fan was seen spinning, and make sure that is over 5 seconds. I'll also switch to the current thermal fanSpeed for the desired fan speed check.

selfdrive/controls/controlsd.py
LoganxDev: Gotcha, will switch to checking the health packet.

pd0wm:
cereal is updated on master to include the alert. If you rebase it should solve the conflicts.

I don't think rebooting will fix the fan. The second line can say "Contact Support"

PR #2260, opened 2020-10-03
tbuffer: unlock mutex before sending signal

deanlee:
We don't need to hold the lock when sending signal,doing so is a pessimization,since the notified thread would immediately block again, waiting for the notifying thread to release the lock.


pd0wm:
I'm not going to merge this PR. I'm worried about very subtle bugs. I'm currently working on a rewrite of visionipc that hopefully won't have any locking at all.

PR #2261, opened 2020-10-03
fix calibration invalid flashing on startup

adeebshihadeh:


pd0wm:
Duplicate of https://github.com/commaai/openpilot/pull/2270

PR #2269, opened 2020-10-05
add caching to CI tests

### LINKED ISSUE #1902: Cache routes used in CI
Use the [cache action](https://github.com/actions/cache) to cache routes for process replay and test_car_models. This should speed up CI and reduce our bandwidth costs. This should be implemented as a generic cache for URLFile/LogReader like FrameReader already has.

ghost:
Attempting to fix https://github.com/commaai/openpilot/issues/1902


adeebshihadeh:
We have a limited cache from github, so we need to make sure the cache we use for the mac build and this aren’t greater than the limit. Otherwise we effectively won’t have a cache since they’ll just overwrite each other.

grekiki:
Well I don't expect videos to take up more than let's say 500MB. Is there a way to see what is in a cache to kind of monitor it? Or at least see what it's size is?

pd0wm:
macos cache is `3484 MB` according to the cache download step. Sounds like we have space for both.

.github/workflows/test.yaml
```diff
@@ -251,11 +267,18 @@ jobs:
...
+            /tmp/comma_download_cache
+          key: ${{ hashFiles('selfdrive/test/test_car_models.py') }}
       - name: Build Docker image
         run: eval "$BUILD"
       - name: Test car models
         run: |
-          $PERSIST "cd /tmp/openpilot && \
+          $PERSIST_WITH_CACHE "cd /tmp/openpilot && \
                     scons -j$(nproc) && \
                     coverage run --parallel-mode -m nose --processes=4 --process-timeout=60 \
```
pd0wm, line 283: You forgot `FILEREADER_CACHE=1`

```suggestion
                    FILEREADER_CACHE=1 coverage run --parallel-mode -m nose --processes=4 --process-timeout=60 \v
```

.github/workflows/test.yaml
```diff
@@ -136,11 +138,18 @@ jobs:
...
+            /tmp/comma_download_cache
+          key: ${{ hashFiles('selfdrive/test/test_valgrind_replay.py') }}
       - name: Build Docker image
         run: eval "$BUILD"
       - name: Run valgrind
         run: |
-          $PERSIST "cd /tmp/openpilot && \
+          $PERSIST_WITH_CACHE "cd /tmp/openpilot && \
                     scons -j$(nproc) && \
                     python selfdrive/test/test_valgrind_replay.py"
```
pd0wm, line 154: You forgot FILEREADER_CACHE=1

```suggestion
                    FILEREADER_CACHE=1 python selfdrive/test/test_valgrind_replay.py"
```

Issue #2262, opened 2020-10-03
Delay controls start until system is ready


adeebshihadeh:
Controls should wait for a message from each process and all the carState signals to be updated before starting up. If this also delayed setting the safety mode, it would fix #1623. This should also properly fix the random alerts quickly flashing on startup before the actual startup alert.

LoganxDev:
Tried to delay based on SubMaster all_valid and CarState canValid, but the camera can doesn't come up as valid. Is that a bad route to take?

PR #2270, opened 2020-10-05
Fix calibration invalid alert on startup

pd0wm:


geohot:
Hmm, isn't calibration actually invalid though since we haven't gotten a packet? Perhaps add a "Calibration.UNINITIALIZED" instead of saying it's calibrated?

adeebshihadeh:
If controlsd doesn't get a packet before trying to engage, it'll throw a commIssue. The real fix for this is #2262.

PR #2274, opened 2020-10-05
Alert cleanup

adeebshihadeh:


selfdrive/controls/lib/events.py
```diff
@@ -588,27 +581,27 @@ def wrong_car_mode_alert(CP: car.CarParams, sm: messaging.SubMaster, metric: boo
   EventName.calibrationInvalid: {
     ET.PERMANENT: Alert(
       "Calibration Invalid",
-      "Reposition Device and Recalibrate",
+      "Remount Device and Recalibrate",
       AlertStatus.normal, AlertSize.mid,
       Priority.LOWER, VisualAlert.none, AudibleAlert.none, 0., 0., .2),
-    ET.SOFT_DISABLE: SoftDisableAlert("Calibration Invalid: Reposition Device & Recalibrate"),
-    ET.NO_ENTRY: NoEntryAlert("Calibration Invalid: Reposition Device & Recalibrate"),
+    ET.SOFT_DISABLE: SoftDisableAlert("Calibration Invalid: Remount Device & Recalibrate"),
```
adeebshihadeh, line 587: @HaraldSchafer with the new calibration can we just make this "Calibration Invalid: Remount Device"?

HaraldSchafer: Hmmm, if they remount with less than 2deg difference it won't trigger a recalibration. So not sure.

adeebshihadeh: let's leave it then

VirtuallyChris: I think "Calibration Invalid: Reposition Mount" might work here

Issue #2277, opened 2020-10-06
Loggerd for PC


adeebshihadeh:
All of loggerd should build and work on PC, including the encoder.

adeebshihadeh:
@deanlee now that you're familiar with loggerd, interested in this one?

deanlee:
yes, I'll try it out.

PR #2278, opened 2020-10-06
Loggerd on PC

deanlee:
issue https://github.com/commaai/openpilot/issues/2277
- [x] c++ encoder
- [ ] encoder for pc


adeebshihadeh:
That was fast :)

Can you make sure encoder.c is properly git moved to encoder.cc? It'll make it much easier to review the diff.

PR #2281, opened 2020-10-06
The correct (factory) way of displaying barriers on toyotas

cydia2020:
**Description**
Currently, when departing lanes, only the left barrier is displayed on Toyota's MFD, this PR makes the MFD display the left hand side barrier when departing left lane and display the right hand side barrier when departing the right lane. (essentially partially mimicking the factory system's behaviour.)

**Verification**
I intentionally departed lanes and verified this works as expected and did not create any problems.

The image below shows what the MFD would display when the barrier signal is set to 0, 1, 2, and 3.
barrier

### DIFF
--- a/selfdrive/car/toyota/toyotacan.py
+++ b/selfdrive/car/toyota/toyotacan.py
@@ -69,7 +69,7 @@ def create_ui_command(packer, steer, chime, left_line, right_line, left_lane_dep
   values = {
     "RIGHT_LINE": 3 if right_lane_depart else 1 if right_line else 2,
     "LEFT_LINE": 3 if left_lane_depart else 1 if left_line else 2,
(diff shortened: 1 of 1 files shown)


pd0wm:
Are you sure this correct?  I would assume that the signal has two bits, where each bit is a side. Still good to show the correct side though instead of both.

@adeebshihadeh when you have some time, can you confirm this one of our cars?

cydia2020:
Hi, with the original `3 if left_lane_depart or right_lane_depart else 0`

Only the left barrier is displayed when departing lanes, I've tested it with my own fork (including just hard set the signal to 0, 1, 2, and 3 in openpilot instead of actually going on road) and this works correctly

adeebshihadeh:
Tested on our Prius, and it works as expected.

PR #2288, opened 2020-10-07
Process replay for C++ daemons

ghost:
Checks consistency of c++ services in comparison to older references.


selfdrive/test/process_replay/process_replay.py
```diff
@@ -206,6 +211,24 @@ def calibration_rcv_callback(msg, CP, cfg, fsm):
     recv_socks = ["liveCalibration"]
   return recv_socks, fsm.frame == 0 or msg.which() == 'cameraOdometry'
 
+def ublox_rcv_callback(msg):
+  byte_list = msg.as_builder().ubloxRaw
+  msg_class = byte_list[2]
+  msg_id = byte_list[3]
+  # print(msg_class, msg_id)
```
pd0wm, line 218: `return (msg_class, msg_id) in ((1, 7 * 16 + 0), (2, 1 * 16 + 5), ...)`

selfdrive/test/process_replay/process_replay.py
```diff
@@ -18,11 +21,12 @@ def tqdm(x):
 from common.params import Params
 from cereal.services import service_list
 from collections import namedtuple
+from common.basedir import BASEDIR
 
 # Numpy gives different results based on CPU features after version 19
 NUMPY_TOLERANCE = 1e-7
 
-ProcessConfig = namedtuple('ProcessConfig', ['proc_name', 'pub_sub', 'ignore', 'init_callback', 'should_recv_callback', 'tolerance'])
+ProcessConfig = namedtuple('ProcessConfig', ['lang', 'proc_name', 'pub_sub', 'ignore', 'command', 'path', 'init_callback', 'should_recv_callback', 'tolerance'])
```
adeebshihadeh, line 29: we don't need any of this extra stuff, it's already in manager

adeebshihadeh: https://github.com/commaai/openpilot/blob/master/selfdrive/manager.py#L288

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +408,39 @@ def replay_process(cfg, lr):
...
+  sub_sockets = [s for _, sub in config.pub_sub.items() for s in sub]  # We get responses here
+  pm = messaging.PubMaster(pub_sockets)
+  sm = messaging.SubMaster(sub_sockets)
+
+  print("Sorting logs")
+  all_msgs = sorted(logreader, key=lambda msg: msg.logMonoTime)
+  pub_msgs = [msg for msg in all_msgs if msg.which() in list(config.pub_sub.keys())]
+
+  os.chdir(os.path.join(BASEDIR, config.path))
+  p = subprocess.Popen(config.command, stderr=subprocess.PIPE)
```
adeebshihadeh, line 423: `manager.start_managed_process`

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +408,39 @@ def replay_process(cfg, lr):
...
+          sent_package = True
+          # Create an event
+          m = messaging.new_message(s)
+          setattr(m, s, sm.data[s])
+          log_msgs.append(m.as_reader())
+      if not sent_package:
+        print("Package was not received")
+        print(msg)
+    
+  os.kill(p.pid, signal.SIGINT)
```
adeebshihadeh, line 445: `manager.kill_managed_process`

selfdrive/test/process_replay/process_replay.py
grekiki: I deal with sockets differently depending on the language

adeebshihadeh: you can borrow the same logic manager uses instead of adding this - https://github.com/commaai/openpilot/blob/master/selfdrive/manager.py#L292

grekiki: Hm ok I can try refactoring

selfdrive/test/process_replay/process_replay.py
```diff
@@ -206,88 +210,139 @@ def calibration_rcv_callback(msg, CP, cfg, fsm):
...
+    proc_name="ubloxd",
+    pub_sub={
+      "ubloxRaw": ["ubloxGnss", "gpsLocationExternal"],
+    },
+    ignore=["logMonoTime"],
+    command="./ubloxd",
+    path="selfdrive/locationd/",
+    init_callback=None,
+    should_recv_callback=ublox_rcv_callback,
+    tolerance=NUMPY_TOLERANCE,
```
adeebshihadeh, line 331: ubloxd doesn't use numpy :)

selfdrive/test/process_replay/process_replay.py
grekiki: Will keep lang, because the parsing of received packages is quite different, but will remove command and path

selfdrive/test/process_replay/process_replay.py
```diff
@@ -206,6 +206,13 @@ def calibration_rcv_callback(msg, CP, cfg, fsm):
     recv_socks = ["liveCalibration"]
   return recv_socks, fsm.frame == 0 or msg.which() == 'cameraOdometry'
 
+def ublox_rcv_callback(msg):
+  byte_list = msg.as_builder().ubloxRaw
```
adeebshihadeh, line 210:
```suggestion
  byte_list = msg.ubloxRaw
```

selfdrive/test/process_replay/process_replay.py
```diff
@@ -206,6 +206,13 @@ def calibration_rcv_callback(msg, CP, cfg, fsm):
...
+  msg_class = byte_list[2]
+  msg_id = byte_list[3]
```
adeebshihadeh:
```suggestion
  msg_class, msg_id = msg.ubloxRaw[2:4]
```

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +380,39 @@ def replay_process(cfg, lr):
...
+  time.sleep(5)  # We give the process time to start
+
+  log_msgs = []
+  for s in sub_sockets:
+    messaging.recv_one_or_none(sockets[s])
+
+  time.sleep(1)
+  for msg in tqdm(pub_msgs):
+    pm.send(msg.which(), msg.as_builder())
+    for s in config.should_recv_callback(msg):
```
pd0wm, line 404: What if there is no recv_callback?

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +380,39 @@ def replay_process(cfg, lr):
 
         recv_cnt -= m.which() in recv_socks
   return log_msgs
+
+def cpp_replay_process(config, logreader):
+  pub_sockets = [s for s in config.pub_sub.keys()]  # We dump data from logs here
+  sub_sockets = [s for _, sub in config.pub_sub.items() for s in sub]  # We get responses here
+  pm = messaging.PubMaster(pub_sockets)
+  sockets = {s : messaging.sub_sock(s) for s in sub_sockets}
+  print("Sorting logs")
```
pd0wm, line 389: remove print

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +380,39 @@ def replay_process(cfg, lr):
...
+  for s in sub_sockets:
+    messaging.recv_one_or_none(sockets[s])
+
+  time.sleep(1)
+  for msg in tqdm(pub_msgs):
+    pm.send(msg.which(), msg.as_builder())
+    for s in config.should_recv_callback(msg):
+      ctime = time.time()
+      response = None
+      while response is None and time.time() - ctime < 2.5:
```
pd0wm, line 407: use timeout argument on socket creation instead of manually implementing this. This uses 100% of a core while waiting.

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +380,39 @@ def replay_process(cfg, lr):
...
+        response = messaging.recv_one_or_none(sockets[s])
+      if response is None:
```
pd0wm, line 409: You can remove the printing

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +380,33 @@ def replay_process(cfg, lr):
 
         recv_cnt -= m.which() in recv_socks
   return log_msgs
+
+def cpp_replay_process(config, logreader):
```
adeebshihadeh, line 384: s/logreader/lr/g to match the other replay function

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +380,33 @@ def replay_process(cfg, lr):
...
+  pub_sockets = [s for s in config.pub_sub.keys()]  # We dump data from logs here
+  sub_sockets = [s for _, sub in config.pub_sub.items() for s in sub]  # We get responses here
+  pm = messaging.PubMaster(pub_sockets)
```
adeebshihadeh, line 387:
```suggestion
  sub_sockets = [s for _, sub in config.pub_sub.items() for s in sub]  # We get responses here
  pm = messaging.PubMaster(config.pub_sub.keys())
```

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +380,33 @@ def replay_process(cfg, lr):
...
+  sockets = {s : messaging.sub_sock(s, timeout=1000) for s in sub_sockets}
+
+  all_msgs = sorted(logreader, key=lambda msg: msg.logMonoTime)
+  pub_msgs = [msg for msg in all_msgs if msg.which() in list(config.pub_sub.keys())]
+
+  manager.start_managed_process(config.proc_name)
```
adeebshihadeh, line 392: currently manager builds on import, but let's not rely on that behavior
```suggestion
  manager.prepare_managed_process(config.proc_name)
  manager.start_managed_process(config.proc_name)
```

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +380,33 @@ def replay_process(cfg, lr):
...
+
+  time.sleep(1)  # We give the process time to start
+
+  log_msgs = []
+  for s in sub_sockets:
+    messaging.recv_one_or_none(sockets[s])
+
+  for msg in tqdm(pub_msgs):
+    pm.send(msg.which(), msg.as_builder())
+    resp_sockets = config.should_recv_callback(msg) if config.should_recv_callback is not None else sub_sockets
```
adeebshihadeh, line 403: order needs to be reversed, this will break when a callback isn't defined

selfdrive/test/process_replay/process_replay.py
```diff
@@ -353,3 +380,33 @@ def replay_process(cfg, lr):
...
+    for s in resp_sockets:
+      response = messaging.recv_one(sockets[s])
+      if response is not None:
+        m = messaging.new_message(s)
+        setattr(m, s, getattr(response,s))
+        log_msgs.append(m.as_reader())
```
adeebshihadeh, line 409: why do we need this?

grekiki: Something to do with capnp read-only and writable packets. Will see what happens without it, but it was a bug fix before the refactoring. @pd0wm

pd0wm: Otherwise it's not an event struct, just the inner sctruct. This causes them to not be readable by logreader.

adeebshihadeh: it must be an event or recv_one wouldn't work. it's setting the same attr it's getting just on a new message.

pd0wm: It used to use submaster before, which is why it's necessary. You're right, is not needed anymore.

PR #2291, opened 2020-10-07
Fix loggerd rotate if camera frame id starts at 0

pd0wm:


ZwX1616 (approved):
lgtm. 
we will need to refactor the loggerd unittest to emulate openpilot startup conditions better though.

PR #2292, opened 2020-10-07
Add more cars to max torq increase

xps-genesis:


pd0wm:
Is there a reliable way to tell which cars have more torque available? Not sure how we're supposed to check this when we review the PR.

KorayZ (changes requested):
add CAR.IONIQ? 2020 EV can handle 384 in my testing.

SiGmAX666:
> add CAR.IONIQ? 2020 EV can handle 384 in my testing.

And CAR.IONIQ_EV_LTD as it can handle 409 no problem.

PR #2298, opened 2020-10-08
Params refactor

pd0wm:
Continuation of https://github.com/commaai/openpilot/pull/2005. Credit goes to @ahowe444


ahowe444:
Cool. Thanks for fixing things up.

pd0wm:
I'm entertaining the thought of dropping the transaction support since we don't really rely on that behavior. That should lead to a lot less code (see https://github.com/commaai/openpilot/pull/2300). I feel a little bad about all the work you put in this, but if we're not actually going to use the code there is no point in maintaining it.

ahowe444:
That's fine. Trying to recreate the original python in C++ got complicated rather quickly. This is simpler.

PR #2300, opened 2020-10-08
Params refactor, simplified

### LINKED ISSUE #1041: Refactor params to use the same underlying implementation -- $200 bounty

pd0wm:
Simplified version of:  https://github.com/commaai/openpilot/pull/2298. Will see if can get away by simplifying the python api to what the c++ version already supports.

TODO:
- [x] Check on completely clean device and make sure folders are create sucessfully
- [x] Handle interrupt in get_db_value_blocking


common/params_pyx_setup.py
```diff
@@ -0,0 +1,21 @@
+import os
+from distutils.core import Extension, setup
+from Cython.Build import cythonize
+from common.cython_hacks import BuildExtWithoutPlatformSuffix
+from common.basedir import BASEDIR
+
+sourcefiles = ['params_wrapper.pyx']
+extra_compile_args = ["-std=c++17"]
```
adeebshihadeh, line 8: why is this 17?

pd0wm: Was done in the original params PR. Will change it what we use everywhere else. We should switch to 17 at some point though, it has some nicer for loops over maps.

adeebshihadeh:
Let's add a test for TODO #1

common/params_pxd.pxd
```diff
@@ -0,0 +1,17 @@
...
+  pass
+
+# Declare the class with cdef
+cdef extern from "selfdrive/common/params.h":
+  cdef cppclass Params:
+    Params(bool)
+    Params(string)
+    string get(string, bool) nogil
+    int delete_db_value(const char*);
+    int write_db_value(string, string);
```
adeebshihadeh, line 17:
```suggestion
    int delete_db_value(const char*)
    int write_db_value(string, string)
```

adeebshihadeh: why does delete take a const char and everything else uses a string?

common/params_pyx.pyx
```diff
@@ -0,0 +1,157 @@
...
+from libcpp cimport bool
+from libcpp.string cimport string
+from params_pxd cimport Params as c_Params
+
+cdef enum TxType:
+  PERSISTENT = 1
+  CLEAR_ON_MANAGER_START = 2
+  CLEAR_ON_PANDA_DISCONNECT = 3
+
+keys = {
```
adeebshihadeh, line 13: why are we only checking this in python?

common/params_pyx.pyx
```diff
@@ -0,0 +1,157 @@
...
+
+  def __dealloc__(self):
+    del self.p
+
+  def _clear_keys_with_type(self, tx_type):
+    for key in keys:
+      if tx_type in keys[key]:
+        self.delete(key)
+
+  def manager_start(self):
```
adeebshihadeh, line 91: we're missing a clear all function now. maybe we should just make a clear_keys function that takes a list of tx types and the default is all?

common/SConscript
```diff
@@ -4,3 +4,11 @@ Import('env', 'cython_dependencies')
 env.Command(['common_pyx.so', 'clock.cpp'],
             cython_dependencies + ['common_pyx_setup.py', 'clock.pyx'],
             "cd common && python3 common_pyx_setup.py build_ext --inplace")
+
+# Build cython params module
+env.Command(['params_pyx.so', 'params_wrapper.cpp'],
```
adeebshihadeh, line 9:
```suggestion
env.Command(['params_pyx.so', 'params_pyx.cpp'],
```

common/params_pyx.pyx
```diff
@@ -0,0 +1,157 @@
...
+  b"Offroad_UpdateFailed": [TxType.CLEAR_ON_MANAGER_START],
+}
+
+class UnknownKeyName(Exception):
+  pass
+
+cdef class Params:
+  cdef c_Params* p
+
+  def __cinit__(self, d=None):
```
adeebshihadeh, line 77: might as well add an arg for persistent params to be consistent with the c++ api

common/params_pxd.pxd
pd0wm: Fixed. Think we should move everything in c++ over to the new std::string versions at some point and retire the pointer based ones.

common/params_pyx.pyx
pd0wm: Do we really care? I could move it to C++ but not sure if it's worth the complexity.

pd0wm:
Folders need to be created recursively. If we move openpilot to c++17 we can use std::filesystem

PR #2307, opened 2020-10-09
boardd always send can packet

pd0wm:
Previously if no messages were seen on the bus the message would not be sent. This makes timing analysis a lot harder.


selfdrive/boardd/panda.cc
```diff
@@ -323,6 +323,7 @@ int Panda::can_receive(cereal::Event::Builder &event){
 
   // return if length is 0
   if (recv <= 0) {
+    event.initCan(0);
```
adeebshihadeh, line 326: if we're committed to this, we can just clip recv at 0 and get rid of the condition

PR #2308, opened 2020-10-09
white/grey/black pandas still have power ctrl

eFiniLan:
white/grey/black panda should still be able to turn off USB power like what we had prior 0.7.9

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -384,9 +384,6 @@ void hardware_control_thread() {
   LOGD("start hardware control thread");
   SubMaster sm({"thermal", "frontFrame"});
 
-  // Other pandas don't have hardware to control
-  if (panda->hw_type != cereal::HealthData::HwType::UNO && panda->hw_type != cereal::HealthData::HwType::DOS) return;
-
   uint64_t last_front_frame_t = 0;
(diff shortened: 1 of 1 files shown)


selfdrive/boardd/boardd.cc
```diff
@@ -423,6 +422,10 @@ void hardware_control_thread() {
       }
 #endif
     }
+
+    // Other pandas don't have IR hardware to control
+    if (panda->hw_type != cereal::HealthData::HwType::UNO && panda->hw_type != cereal::HealthData::HwType::DOS) return;
```
pd0wm, line 427: `continue`?

pd0wm:
You can move the charging code to above the fan speed section. There is no specific order. That should save you the extra if condition.

eFiniLan:
> You can move the charging code to above the fan speed section. There is no specific order. That should save you the extra if condition.

thanks! just refactor it a bit, please let me know how it goes :)

pd0wm:
Looks good except for the 'return'. Think that needs to be a 'continue'. Otherwise it exits the thread instead of starting at the top of the loop.

eFiniLan:
> Looks good except for the 'return'. Think that needs to be a 'continue'. Otherwise it exits the thread instead of starting at the top of the loop.

updated, cheers!

PR #2309, opened 2020-10-09
Startup speedup: scons decider

ghost:
Findings

It takes about 4s for thermald to start and around 15s for locationd to start. Thermald might be slightly fixable

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -7,7 +7,7 @@ import sys
 import platform
 
 TICI = os.path.isfile('/TICI')
-Decider('make')
+Decider('MD5-timestamp')
 
 AddOption('--test',
           action='store_true',
--- a/SConstruct
+++ b/SConstruct
@@ -7,6 +7,7 @@ import sys
 import platform
 
 TICI = os.path.isfile('/TICI')
+Decider('make')
 
 AddOption('--test',
           action='store_true',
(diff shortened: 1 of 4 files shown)


common/apk.py
```diff
@@ -28,18 +29,38 @@ def install_apk(path):
...
+  perms3 = [p.replace(" ","") for p in perms2]
+  permsfiltered = [p.split(":")[0].split("android.permission.")[1] for p in perms3 if len(p) > 20]
+  return permsfiltered
+
 
 def set_package_permissions():
-  pm_grant("ai.comma.plus.offroad", "android.permission.ACCESS_FINE_LOCATION")
-  pm_grant("ai.comma.plus.offroad", "android.permission.READ_PHONE_STATE")
-  pm_grant("ai.comma.plus.offroad", "android.permission.READ_EXTERNAL_STORAGE")
+  out = subprocess.Popen(['dumpsys', 'package', 'ai.comma.plus.offroad'], stdout=subprocess.PIPE, stderr=subprocess.STDOUT)
```
adeebshihadeh, line 44: do we really need all this extra apk code? is it any faster than the thread?

SConstruct
```diff
@@ -7,6 +7,7 @@ import sys
 import platform
 
 TICI = os.path.isfile('/TICI')
+Decider('make')
```
adeebshihadeh, line 10: can you try `MD5-timestamp`? It's safer and should still be much faster than just MD5

pd0wm:
Please remove ui change, that has nothing to do with this PR.

PR #2218, opened 2020-09-22
Added fingerprint for 2020 Nissan Leaf SV Plus

mkfischer:
Added fingerprint for 2020 Nissan Leaf SV Plus

PR #2311, opened 2020-10-10
Add Nissan Leaf 2020

VirtuallyChris:
Continuation of https://github.com/commaai/openpilot/pull/2218


pd0wm:
I see there are messages with the same id but different lengths. In that case the fingerprint *does* need to be split into two.

VirtuallyChris:
Yeah I was talking to Adeeb about this. I'll split them back up then merge this 👍

PR #2314, opened 2020-10-10
make hyundai resume reliable

### LINKED ISSUE #2142: Hyundai SCC automatic resume is unreliable

gregjhogan:
fixes #2142

I found that sending 100 resume commands all at once gives 100% reliability accepting first attempt to resume by spamming resume button over 10+ minutes

sending 100 resume commands all at once might not be a good idea due to the amount of time it takes to send them (since nothing with a higher CAN addr can send messages during that time) so I send 25 every 0.1 seconds

before this change I can sit behind a parked car for less than a minute and have a resume command not be accepted in less than a second (sometimes it will even take 10+ seconds before a resume command is accepted)

after this change I can sit behind a parked car for more than 10 minutes and the longest I saw it take to accept a resume command was 3 attempts (0.3 seconds)

### DIFF
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -77,9 +77,10 @@ class CarController():
     if pcm_cancel_cmd:
       can_sends.append(create_clu11(self.packer, frame, CS.clu11, Buttons.CANCEL))
     elif CS.out.cruiseState.standstill:
-      # send resume at a max freq of 5Hz
-      if (frame - self.last_resume_frame)*DT_CTRL > 0.2:
-        can_sends.append(create_clu11(self.packer, frame, CS.clu11, Buttons.RES_ACCEL))
+      # send resume at a max freq of 10Hz
+      if (frame - self.last_resume_frame)*DT_CTRL > 0.1:
+        # send 25 messages at a time to increases the likelyhood of resume being accepted
+        can_sends.extend([create_clu11(self.packer, frame, CS.clu11, Buttons.RES_ACCEL)] * 25)
         self.last_resume_frame = frame
 
     # 20 Hz LFA MFA message


pd0wm:
If my calculations are correct (assuming 128 bits/packet due headers an bit stuffing) it will take less than 10 ms to send out those 25 packets. Probably fine.

gregjhogan:
I think since it is a 4 byte message it should be no more than 90 bits worst case? When I did the math I got less than 5ms.

pd0wm:
Oh, just assumed it was 8 bytes. Yeah, with 4 bytes it's probably less than 5 ms.

adeebshihadeh:
Let me do some tests in the sonata, then I’ll merge it.

adeebshihadeh:
merged in #2342

PR #2315, opened 2020-10-10
There should not be gl configuration in between nvgBeginFrame and nvgEndFrame

sohamm17:
***** Template: Bug fix *****

**Description**
The UI stops in Intel iGPUs when it runs in a x86 machine with the Carla simulator. This happens because in the ui_draw_vision the OpenGL code is mixed with nvgBeginFrame and nvgEndFrame.

**Verification**
I run Carla, bridge.py and launch_openpilot.sh in tools/sim and the UI stopped working after I started the bridge.py


adeebshihadeh:
Should be fixed in b9c14c92a7603e2be08cb519cedfb19e563caeb9, just implemented this in a different way. Tested on an Intel iGPU.

Issue #2316, opened 2020-10-11
Needs a better way of detecting hotspots


cydia2020:
On my Pixel 4 XL running Android 11, the hotspot's IP address starts with 192.168. and ends with 143; the only variable in this ip address is the third part of it, breaking the hotspot detection function in uploader. openpilot should probably have a better way of detecting hotspots.

eg: list wi-fi name as a hotspot when the car is above 50 km/h and the said SSID is connected.

adeebshihadeh:
I’m not sure hotspot detection is useful anymore since we don’t upload raw logs while onroad. What’s your use case for this?

cydia2020:
Hi, sometimes I do a bit of tinkering in the car while I am connected to my hotspot. However this might be an isolated case and if it doesn't have widespread affect I will just make the changes in my fork and close the issue. Thank you.

PR #2319, opened 2020-10-12
remove hotspot exceptions in uploader

adeebshihadeh:
This shouldn't be necessary anymore since we don't upload raw logs while onroad.

### DIFF
--- a/selfdrive/loggerd/uploader.py
+++ b/selfdrive/loggerd/uploader.py
@@ -229,11 +229,11 @@ def uploader_fn(exit_event):
   while not exit_event.is_set():
     offroad = params.get("IsOffroad") == b'1'
     allow_raw_upload = (params.get("IsUploadRawEnabled") != b"0") and offroad
-    if not offroad or counter % 12 == 0:
+    if offroad and counter % 12 == 0:
       on_wifi = is_on_wifi()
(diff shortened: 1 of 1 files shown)


pd0wm:
Why do we still run the network check at all while onroad? Shouldn't we just only upload qlogs when onroad?

xps-genesis:
Right now the upload will wait till it can connect to home wifi if I forget to disconnect the hotspot after a drive.

will this change enable the upload on the hotspot when offroad? sometimes my phones hotspot is still connected to comma while the car is in the garage, don't want it upload on the hotspot.

adeebshihadeh:
@xps-genesis yes, this change remove the exceptions for hotspots. what's your use case for connecting the hotspot?

xps-genesis:
> @xps-genesis yes, this change remove the exceptions for hotspots. what's your use case for connecting the hotspot?

mainly live tuning, try code change while on a drive, and debugging. Is there a reason to remove this check? I thought this check is very smart and stopped uploading huge data files over mobile hotpot.

pd0wm:
The problem is issue like this https://github.com/commaai/openpilot/issues/2316. There is always going to be another os/version that has a different IP range.

PR #2324, opened 2020-10-12
Fix lane change when blinker is turned off early

pd0wm:
It was introduced in https://github.com/commaai/openpilot/commit/105b95ffd712e462149ab1df2de8b3f1d2307850#diff-b4bfd678cbcd3f58a51d69911247c437L126, but not sure anymore why.

### DIFF
--- a/selfdrive/controls/lib/pathplanner.py
+++ b/selfdrive/controls/lib/pathplanner.py
@@ -106,7 +106,7 @@ class PathPlanner():
     elif sm['carState'].rightBlinker:
       self.lane_change_direction = LaneChangeDirection.right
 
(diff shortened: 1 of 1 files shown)


VirtuallyChris:
I remember this was changed because sometimes the lane change overshoots. This let the user decide when the lane change ended. Has this improved with newer models?

pd0wm:
It was added in 0.7.2 when we didn't have the 1 pulse desire yet. So it's probably improved a lot by now. 

Now I just get a lot of inconsistent lane changes because I'm turn the blinker off too early (which causes a very aggressive centering with some overshoot).

PR #2348, opened 2020-10-16
np.matrix is no longer recommended and may be removed in the future

sshane:
This PR changes all active `np.matrix` variables to `np.array` in `latcontrol_indi.py` ([np.matrix docs message about deprecation](https://numpy.org/doc/stable/reference/generated/numpy.matrix.html)). Since the arrays are all no larger than 2D, no change in output should occur.

Here's the comparison of the dot products of K and C both as matrices and as arrays:

```python
>>> np.allclose(np.dot(K, C), np.dot(np.array(K), np.array(C)))
True
```

### DIFF
--- a/selfdrive/controls/lib/latcontrol_indi.py
+++ b/selfdrive/controls/lib/latcontrol_indi.py
@@ -13,24 +13,24 @@ class LatControlINDI():
   def __init__(self, CP):
     self.angle_steers_des = 0.
 
-    A = np.matrix([[1.0, DT_CTRL, 0.0],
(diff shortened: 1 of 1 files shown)


pd0wm:
I see this is not tested in process replay. Let's get a RAV4 and a Prius in there so both LQR and INDI are verified.

adeebshihadeh:
Added Prius and RAV4 routes to process replay in 516da927f9a1f04be4cbc281bc1b3c73092a4a36. Can you rebase?

ShaneSmiskol:
Let me try

PR #2350, opened 2020-10-16
refactor omx encoder,save 6% of a core

deanlee:
- [x] refactor encoder
  - [x] downscale on GPU
  - [x] 2-4x faster write
  - [x] reuse messagebuilder
  - [x] convert to c++


deanlee:
Because camera_common.h is now a C++ header file, it will cause compilation error in encoder.c.so i'm going to convert encoder.c to c++ in this PR.

pd0wm:
You can use `git mv` to rename the file. That should keep the history.

deanlee:
I have used `git mv`,the history log is correct:
```
commit 3e342caf1ec519d9daa95c532b1c114959ca9cac
Author: deanlee <deanlee3@gmail.com>
Date:   Sun Oct 18 16:54:20 2020 +0000

    renamed  encoder.c -> encoder.cc

 selfdrive/loggerd/{encoder.c => encoder.cc} | 0
 1 file changed, 0 insertions(+), 0 deletions(-)
```
But it seems that if it differs too much, the review tool will consider it's a new file.

deanlee:
looks like I overdid it.It is difficult to review and may introduce new issues。I will temporarily close this PR,think about how to do this appropriately.

PR #2354, opened 2020-10-17
fixed typo , removed unwanted spaces and added punctuation marks .

lakshit24:


adeebshihadeh:
The third change isn't correct, and I think the rest is fine as is.

PR #2357, opened 2020-10-17
Fingerprint for 2021 RAV4H XLE

mattbsea:
Using this fingerprint, I was able to run OpenPilot on a 2021 RAV4 XLE.  It's not my car or comma2, but I can get you the dongle ID if you'd like to verify the data in Cabana.


selfdrive/car/toyota/values.py
```diff
@@ -259,7 +259,10 @@ class CAR:
...
   },
-  ],
+  # 2021 XLE
```
adeebshihadeh, line 262: can you remove this? we're only merging FW versions for toyota

mattbsea: Removed v1 fingerprint.

PR #2365, opened 2020-10-20
fix build warning

deanlee:
`util.c:55:12: warning: comparison of integer expressions of different signedness: ‘ssize_t’ {aka ‘long int’} and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]`

### DIFF
--- a/selfdrive/common/util.c
+++ b/selfdrive/common/util.c
@@ -52,7 +52,7 @@ int write_file(const char* path, const void* data, size_t size) {
   }
   ssize_t n = write(fd, data, size);
   close(fd);
-  return n == (ssize_t)size ? 0 : -1;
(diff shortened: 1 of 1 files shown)


selfdrive/common/util.c
```diff
@@ -52,7 +52,7 @@ int write_file(const char* path, const void* data, size_t size) {
   }
   ssize_t n = write(fd, data, size);
   close(fd);
-  return n == size ? 0 : -1;
+  return n == (ssize_t)size ? 0 : -1;
```
adeebshihadeh, line 55:
```suggestion
  return (n >= 0 && (size_t)n == size) ? 0 : -1;
```
think this is safer

PR #2366, opened 2020-10-20
prevent snapshot while onroad

### LINKED ISSUE #2364: Snapshot request when onroad causes openpilot to go offroad
**Describe the bug**

Requesting a snapshot while on onroad causes openpilot to go temporarily offroad. I see the offroad snapshot alert and then eventually dash errors (a bit of delay on Honda Bosch). Onroad does come back after the snapshot sequence finishes.

Note: It looks like the process check in snapshot.py is returning an empty array/list when onroad. Is something killing camerad?

1. Go onroad.
2. Request a snapshot via the API.
3. Device will go offroad and show the snapshot alert. CAN sending appears to stop during the sequence.
4. The snapshots intermittently make it back through via the API. Other times, they don't.

**Expected behavior**

Nothing. Return None and don't pull snapshots when onroad.

 - Device: comma two
 - Dongle ID: 95e637a323d6e9fc
 - Routes: These are consecutive. My house requests snapshots every 5 minutes or so, hence the timed nature. 2020-10-19--17-39-48, 2020-10-19--17-43-06, 2020-10-19--17-48-07
 - Timestamp: N/A
 - Version: master-ci 29a96263ea27c7998753a234bcd8c10261ec7fef (10/19/2020)
 - Car make/model: 2017 Civic Hatchback

adeebshihadeh:
fixes #2364

### DIFF
--- a/selfdrive/camerad/snapshot/snapshot.py
+++ b/selfdrive/camerad/snapshot/snapshot.py
@@ -19,7 +19,7 @@ def snapshot():
   params = Params()
   front_camera_allowed = int(params.get("RecordFront"))
 
-  if params.get("IsTakingSnapshot") == b"1":
+  if params.get("IsOffroad") is not None or params.get("IsTakingSnapshot") == b"1":
     return None
 
   params.put("IsTakingSnapshot", "1")
--- a/selfdrive/camerad/snapshot/snapshot.py
+++ b/selfdrive/camerad/snapshot/snapshot.py
@@ -19,7 +19,7 @@ def snapshot():
   params = Params()
   front_camera_allowed = int(params.get("RecordFront"))
 
-  if params.get("IsOffroad") is not None or params.get("IsTakingSnapshot") == b"1":
+  if params.get("IsOffroad") != b"1" or params.get("IsTakingSnapshot") == b"1":
     return None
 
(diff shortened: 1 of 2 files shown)


csouers:
The issue still occurs with this branch. It's as if the `ps | grep camerad` is returning as if camerad is not running.

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -19,7 +19,7 @@ def snapshot():
   params = Params()
   front_camera_allowed = int(params.get("RecordFront"))
 
-  if params.get("IsTakingSnapshot") == b"1":
+  if params.get("IsOffroad") is not None or params.get("IsTakingSnapshot") == b"1":
```
csouers, line 22: With a change to `  if params.get("IsOffroad") is None or params.get("IsTakingSnapshot") == b"1":
`, this works as expected

adeebshihadeh: oops, should be fixed now

csouers: Tested fine. Thank you!

csouers:
Testing fine now. Thanks!

PR #2367, opened 2020-10-20
Refactor LoggerState, use ZSTD for compression

deanlee:
Combine https://github.com/commaai/openpilot/pull/1098  https://github.com/commaai/openpilot/pull/1951 into one.

- [ ] C++ LoggerState
  - [x] use shared_ptr instead of handles array
  - [x] use ZSTD for compression
- [ ] test routine for LoggerState


adeebshihadeh:
We've thought about moving to zstd, but it's not something we're prepared to spend time on for the work that needs to be done outside of openpilot. However, we're thinking of offloading the compression to the DSP if you want to look into that.

PR #2370, opened 2020-10-20
Fix UI typo

ghost:


pd0wm:
Kept the same as current offroad: https://github.com/commaai/openpilot-apks/blob/master/offroad/js/components/Settings/Settings.js#L338

PR #2374, opened 2020-10-20
improve tici cameras timing

ZwX1616:
some say cici is better


ZwX1616:
this was a mess. opening a new alt one

PR #2379, opened 2020-10-21
openpilot v0.7.10 release

adeebshihadeh:
```
Version 0.7.10 (2020-10-26)
========================
 * NEOS update: update to Python 3.8.2 and lower CPU frequency
 * Improved thermals due to reduced CPU frequency
 * Update SNPE to 1.41.0
 * Reduced offroad power consumption
 * Various system stability improvements
 * Acura RDX 2020 support thanks to csouers!
```


README.md
```diff
@@ -309,7 +310,7 @@ Directory Structure
     ├── phonelibs           # Libraries used on NEOS devices
     ├── pyextra             # Libraries used on NEOS devices
     └── selfdrive           # Code needed to drive the car
-        ├── assets          # Fonts, images, and sounds for UI
```
geohot, line 312: lol where'd the comma go? (no need to actually change this)

geohot:
This code is getting a ton better. Gave it a quick read through, there's a huge diff in visiond that's hard to read

selfdrive/thermald/thermald.py
```diff
@@ -403,8 +377,10 @@ def thermald_thread():
         started_seen = True
         os.system('echo performance > /sys/class/devfreq/soc:qcom,cpubw/governor')
     else:
+      if startup_conditions["ignition"] and (startup_conditions != startup_conditions_prev):
+        cloudlog.event("Startup blocked", startup_conditions=startup_conditions)
       if should_start_prev or (count == 0):
-        put_nonblocking("IsOffroad", "1")
+        params.put("IsOffroad", "1")
```
geohot, line 383: Are you sure this is okay to do nonblocking? I agree it's offroad, but params.put has been the source of so many issues. (perhaps nonblocking should be the default in later versions)

ZwX1616:
> This code is getting a ton better. Gave it a quick read through, there's a huge diff in visiond that's hard to read

basically it's just less duplicate code and a lot less ifdefs in main.

selfdrive/thermald/thermald.py
adeebshihadeh: It shouldn't be a problem since it's only done once on the onroad -> offroad transition. It's done to prevent an unlikely race condition with things like snapshot, which could start a second camerad.

PR #2405, opened 2020-10-24
Hyundai Genesis tuning

Donfyffe:
2015 Genesis 3.8 L. tuning to stop extreme ping pong on cornering and on straights.


adeebshihadeh:
Our policy for tuning PRs is that a few users have to confirm that it's better, but I don't think this is a particularly popular car, so confirmation from one user should be fine.

xps-genesis:
I can attest this works very well.

Donfyffe:
This would make me able to use OP main on my genesis, right now I have to using other forks that might get me banned.

Please note: message attached

PR #2407, opened 2020-10-25
Hyundai Veloster fingerprint

CruiseBrantley:
**Description** At times the Veloster is not detected on Comma Two boot, which requires power cycle of the Comma Two until it works. I believe the issue is that the carFw is detecting transmission differently each time mock is shown in user admin, so I've added the alternate fw detected there to the array.
Also noticed that Veloster was not included with 384 STEER_MAX, I have tested this extensively and never had an issue.

**Verification** Have tested a few starts since and the issue has not occurred again. 384 STEER_MAX has also never caused an issue for my vehicle with months of testing.

**Route**
Route: d624b3d19adce635


xps-genesis:
good catch, probably my fork only looked at mdps fw this issue was not present. hope this will put Veloster into good position to use stock 👍

adeebshihadeh:
Removed the increased torque limit for now. We'll write a script to verify the increased torque range is usable, and you'll just have to give us a route.

PR #2428, opened 2020-10-27
make a smarter delay

ghost:
The 0.01s delay is sometimes too low in CI tests. We increase it by 10 times, but if the message arrives, we send it as soon as possible.

### DIFF
--- a/selfdrive/controls/tests/test_startup.py
+++ b/selfdrive/controls/tests/test_startup.py
@@ -66,7 +66,7 @@ class TestStartup(unittest.TestCase):
       pm.send('can', can_list_to_can_capnp(msgs))
       curr_time = time.time()
       msgs = []
-      while len(msgs)==0 and time.time()-curr_time < 0.01:
+      while len(msgs)==0 and time.time()-curr_time < 0.1:
         msgs = messaging.drain_sock(controls_sock)
       if len(msgs):
         event_name = msgs[0].controlsState.alertType.split("/")[0]
--- a/selfdrive/controls/tests/test_startup.py
+++ b/selfdrive/controls/tests/test_startup.py
@@ -38,10 +38,7 @@ class TestStartup(unittest.TestCase):
     (EventName.startupNoCar, None, False),
   ])
   def test_startup_alert(self, expected_event, car, toggle_enabled):
-    if car is not None:
-      os.environ["FINGERPRINT"] = car
-    else:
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
I don't think I've seen this test fail randomly. Are you sure this is needed?

selfdrive/controls/tests/test_startup.py
```diff
@@ -36,9 +37,12 @@ class TestStartup(unittest.TestCase):
...
     (EventName.startupNoCar, None, False),
   ])
-  @with_processes(['controlsd'])
   def test_startup_alert(self, expected_event, car, toggle_enabled):
-
+    if car is not None:
+      os.environ["FINGERPRINT"] = car
+    else:
+      os.environ["FINGERPRINT"] = ""
+    start_managed_process('controlsd')
```
adeebshihadeh, line 45: why did this change?

selfdrive/controls/tests/test_startup.py
```diff
@@ -64,16 +68,18 @@ def test_startup_alert(self, expected_event, car, toggle_enabled):
     for _ in range(500):
       msgs = [[addr, 0, b'\x00'*length, 0] for addr, length in finger.items()]
       pm.send('can', can_list_to_can_capnp(msgs))
-
-      time.sleep(0.01)
-      msgs = messaging.drain_sock(controls_sock)
+      curr_time = time.time()
+      msgs = []
+      while len(msgs)==0 and time.time()-curr_time < 0.1:
+        msgs = messaging.drain_sock(controls_sock)
```
adeebshihadeh, line 74:
```suggestion
      time.sleep(0.1)
      msgs = messaging.drain_sock(controls_sock)
```

selfdrive/controls/tests/test_startup.py
```diff
@@ -36,9 +37,12 @@ class TestStartup(unittest.TestCase):
     (EventName.startupNoCar, None, True),
     (EventName.startupNoCar, None, False),
   ])
-  @with_processes(['controlsd'])
   def test_startup_alert(self, expected_event, car, toggle_enabled):
-
+    if car is not None:
+      os.environ["FINGERPRINT"] = car
+    else:
+      os.environ["FINGERPRINT"] = ""
```
adeebshihadeh, line 44:
```suggestion
    os.environ["FINGERPRINT"] = "" if car is None else car
```

PR #2431, opened 2020-10-27
Abstract set_unsafe_mode function to Panda

sshane:
Provides an easy function for forkers to use to set various unsafe modes without modifying panda fw or using a raw USB command.

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -107,6 +107,8 @@ void safety_setter_thread() {
   cereal::CarParams::Reader car_params = cmsg.getRoot<cereal::CarParams>();
   cereal::CarParams::SafetyModel safety_model = car_params.getSafetyModel();
 
+  panda->set_unsafe_mode(0);  // see safety_declarations.h for allowed values
+
   auto safety_param = car_params.getSafetyParam();
   LOGW("setting safety model: %d with param %d", (int)safety_model, safety_param);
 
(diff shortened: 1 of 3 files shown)


selfdrive/boardd/panda.cc
```diff
@@ -188,6 +188,10 @@ void Panda::set_safety_model(cereal::CarParams::SafetyModel safety_model, int sa
   usb_write(0xdc, (uint16_t)safety_model, safety_param);
 }
 
+void Panda::set_unsafe_mode(int unsafe_mode) {
+  panda->usb_write(0xdf, unsafe_mode, 0);
```
ShaneSmiskol, line 192: Did I get this right? The second argument is the `wValue` which panda/main.cc uses to set unsafe_mode

selfdrive/boardd/boardd.cc
```diff
@@ -107,6 +107,8 @@ void safety_setter_thread() {
   cereal::CarParams::Reader car_params = cmsg.getRoot<cereal::CarParams>();
   cereal::CarParams::SafetyModel safety_model = car_params.getSafetyModel();
 
+  panda->set_unsafe_mode(0);  // see safety_declarations.h for allowed values
```
ShaneSmiskol, line 110: Do you want it to send a zero by default or just provide the function and remove this?

selfdrive/boardd/panda.cc
adeebshihadeh, line 192:
```suggestion
void Panda::set_unsafe_mode(uint16_t unsafe_mode) {
  usb_write(0xdf, unsafe_mode, 0);
```

selfdrive/boardd/boardd.cc
adeebshihadeh: yeah, this is what we want

selfdrive/boardd/panda.cc
ShaneSmiskol: Just saw that, fixed

PR #2447, opened 2020-10-30
loggerd on PC

deanlee:
issue #2277


adeebshihadeh:
Thanks for the PR, but the raw logger isn't really used. Ideally, we want all the logic in loggerd.cc and encoder_thread to run on a similar encoder on PC as to what runs on device, meaning no ifdefs for different encoder types. Let me know if you want further clarification.

ZwX1616:
> Thanks for the PR, but the raw logger isn't really used. Ideally, we want all the logic in loggerd.cc and encoder_thread to run on a similar encoder on PC as to what runs on device, meaning no ifdefs for different encoder types. Let me know if you want further clarification.

some generalization similar to camerad would be nice!

adeebshihadeh:
Actually, I think I have a nice way to do this in a few lines. Let me try that first.

PR #2467, opened 2020-11-01
switch build flags to use opengl framework rather than gl library for OSX

christomitov:
Modified Sconscripts to use OpenGL framework for OSX systems.


selfdrive/ui/SConscript
```diff
@@ -20,11 +20,10 @@ else:
 
   if arch == "larch64":
     qt_libs += ["GLESv2", "wayland-client"]
-  else:
+  elif != "Darwin":
```
christomitov, line 23: forget `arch`

adeebshihadeh: oops, fixed

PR #2483, opened 2020-11-04
Ioniq Ev 2019 fingerprint

NinoFurger:
Adds fingerprint for Hyundai Ioniq EV 2019


adeebshihadeh:
The VIN/FW query seems incomplete on some HKG models, including your Ioniq. Until we fix it, let's just add any missing messages to your model's CAN fingerprint.

PR #2487, opened 2020-11-05
Car Port: Sonata Hybrid 2020

adeebshihadeh:


realistic69:
I get a CAN error: Check Connections when I start the car on this branch

PR #2410, opened 2020-10-25
Genesis G70 fingerprint and tuning

ohber:
**Description**
- Update values.py to match entries for Genesis G70 from xps fork (stock does not work without these changes)
- Update tuning values in interface.py
- Add G70 to car list for 384 torque

**Verification**
- 384 torque and values.py changes have been present in xps' fork, and I have encountered no issues (used for 2+ months)
- Tuning is slightly changed from xps' fork but verified working well on various roads and speeds

**Route**
70c5bec28ec8e345|2020-10-25--11-42-28

PR #2491, opened 2020-11-06
Genesis G70 fingerprint and tuning

ohber:
**Description**
Update values.py to match entries for Genesis G70 from xps fork (stock does not work without these changes)
Update tuning values in interface.py
Add G70 to car list for 384 torque

**Verification**
384 torque and values.py changes have been present in xps' fork, and I have encountered no issues (used for 2+ months)
Tuning is slightly changed from xps' fork but verified working well on various roads and speeds

**Route**
70c5bec28ec8e345|2020-10-25--11-42-28

### DIFF
--- a/selfdrive/car/hyundai/interface.py
+++ b/selfdrive/car/hyundai/interface.py
@@ -155,12 +155,15 @@ class CarInterface(CarInterfaceBase):
 
     # Genesis
     elif candidate == CAR.GENESIS_G70:
-      ret.lateralTuning.pid.kf = 0.00005
-      ret.mass = 1640. + STD_CARGO_KG
+      ret.lateralTuning.init('indi')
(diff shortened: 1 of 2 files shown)


ohber:
Ran into issues rebasing my branch - addressed all issues from PR #2410 and resubmitted under a new branch and this new PR.
- Removing car from 'ignored fingerprints' appears to be necessary; I've verified the FP and FW versions already present in Stock match exactly what my car produces

VirtuallyChris:
You're right that the FW versions match exactly but as seen in `70c5bec28ec8e345|2020-11-04--14-06-06` it doesn't identify correctly. It still is trying to identify via CAN. Strange things happening here..

adeebshihadeh:
There was FW for an extra ECU that was probably added by mistake. Your car should now fingerprint correctly.

PR #2497, opened 2020-11-07
logger.cc: remove function mkpath

deanlee:


selfdrive/common/params.cc
```diff
@@ -96,7 +96,7 @@ static int mkdir_p(std::string path) {
   return 0;
 }
 
-static int ensure_dir_exists(std::string path) {
+int ensure_dir_exists(std::string path) {
```
adeebshihadeh, line 99: can we just replace this with std::file system::create_directories now that we’re on C++17?

deanlee: I tried, but it seems that the library is missing on EON
`undefined reference to `std::__ndk1::__fs::filesystem::__create_directories(std::__ndk1::__fs::filesystem::path const&, std::__ndk1::error_code*)'
`

pd0wm: > LLVM implementation prior to LLVM 9.0 requires linking with -lc++fs

You can link that library globally in the sconstruct if needed.

adeebshihadeh:
Closing this as I think it's pretty dirty. This should ideally use `std::filesystem::create_directories`, but if not, move mkpath to our utils file.

PR #2500, opened 2020-11-07
Update tensorflow and cryptography in Pipfile to fix vulnerabilities

martinl:
This PR fixes 2 vulnerabilities found by github dependabot:

- cryptography https://github.com/advisories/GHSA-hggm-jpg3-v476
- tensorflow https://github.com/advisories/GHSA-hx2x-85gr-wrpq


pd0wm:
The vulnerable code in both packages is not used by openpilot, so it's not an issue.

PR #2502, opened 2020-11-08
Update DM awareness times

sshane:
- This tunes driver monitoring awareness times, previously 2 seconds were given before the alert made noise, now it's up to 6 seconds (9 seconds originally before change in October).
- Also I moved the `hi_std_alert` to only be reset when red terminal status, it would end up displaying every ~20 seconds causing me to take extra time taking my eyes off the road to see what the device is alerting me of. Having it reset when red should have it displaying only once per drive usually (since we don't need to constantly be reminded of the same issue).

### DIFF
--- a/selfdrive/monitoring/driver_monitor.py
+++ b/selfdrive/monitoring/driver_monitor.py
@@ -15,8 +15,8 @@ EventName = car.CarEvent.EventName
 # ******************************************************************************************
 
 _AWARENESS_TIME = 35.  # passive wheel touch total timeout
-_AWARENESS_PRE_TIME_TILL_TERMINAL = 7.
-_AWARENESS_PROMPT_TIME_TILL_TERMINAL = 5.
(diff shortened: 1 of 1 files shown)


geohot:
As discussed, we aren't raising the limit, but we will make the alert show earlier. Line 17 shouldn't be changed.

ShaneSmiskol:
Did you still want me to revert it?

ZwX1616:
Yeah I think we should keep the time from nothing to orange around 30s. Tradeoffs can be made by sliding the green one.

35/10(9)/5 or 35/12/6 would be fine.

ShaneSmiskol:
Let me go test these changes and make sure that this totally fixes the alert, I'll let you know in a few hours if it's good!

ShaneSmiskol:
Yes, these changes work great! That unnecessary uncertain alert no longer shows until it goes into terminal status

PR #2508, opened 2020-11-09
Try to fix TSS2 creep with higher stopping target

pd0wm:
If these tuning values fix the TSS2 creep, and still provide smooth stopping I can put them in carParams so they can be tuned on a vehicle by vehicle basis.

Original discussion: https://github.com/commaai/openpilot/issues/1270

I think we need to increase start/stopping speed as well to maintain the same resume speed from a stop. If these values work we also need to make sure we only change them for the vehicles that need it and keep other cars the same.

### DIFF
--- a/selfdrive/controls/lib/longcontrol.py
+++ b/selfdrive/controls/lib/longcontrol.py
@@ -10,9 +10,9 @@ STOPPING_TARGET_SPEED = MIN_CAN_SPEED + 0.01
 STARTING_TARGET_SPEED = 0.5
 BRAKE_THRESHOLD_TO_PID = 0.2
 
-STOPPING_BRAKE_RATE = 0.2  # brake_travel/s while trying to stop
-STARTING_BRAKE_RATE = 0.8  # brake_travel/s while releasing on restart
-BRAKE_STOPPING_TARGET = 0.5  # apply at least this amount of brake to maintain the vehicle stationary
+STOPPING_BRAKE_RATE = 2 * 0.2  # brake_travel/s while trying to stop
+STARTING_BRAKE_RATE = 2 * 0.8  # brake_travel/s while releasing on restart
+BRAKE_STOPPING_TARGET = 2 * 0.5  # apply at least this amount of brake to maintain the vehicle stationary
 
 RATE = 100.0
 


zorrobyte:
I would argue that, off the cuff, these values differ per vehicle; especially between hybrid/non-hybrid braking systems.
It's been a long time since I've touched the state machine but if I remember correctly it could use a rewrite/refactor (such as releasing brakes 100% when creeping, then entering in the stopping condition again, etc).

You want to modulate the brake pedal and allow the car to creep forward under own power without completely releasing the pedal, something that is impossible with current state machine. Maybe interp to lead distance would be the best way to handle this.

briskspirit:
This workaround won't help imho as I tested it. 
In my opinion problem is that vEgo gets negative so "system"(longcontrol.py?) releases brakes and applies them again when vEgo = 0. Sorry if I am wrong, just trying to fix the same issue with my Corolla and started from longitudinal tuning, then found that MIN_CAN_SPEED is wrong for TSS2 Corolla (should be .375) and was checking this graph after each tune:

Capture

pd0wm:
I don't think it has to do with the speed going negative.

What's interesting though is that the system briefly goes into CRUISE_STATE 11 when the wheel speeds hit zero. This in turn causes ACCEL_NET to jump up. Seems like they have some internal processing of our accel command, and going into certain states makes the actual braking lower than we command.

briskspirit:
Stopping brake rate 0.1:
f66032c2b5aa18ac 2020-12-03--16-55-23 340sec
f66032c2b5aa18ac_2020-12-03--16-55-23_340sec_STOPPING_BRAKE_RATE01

Stopping brake rate 0.2(on old long tune):
f66032c2b5aa18ac 2020-11-21--13-02-06 507sec
f66032c2b5aa18ac_2020-11-21--13-02-06_507sec_STOPPING_BRAKE_RATE02_oldtune

Stopping brake rate 0.4:
f66032c2b5aa18ac 2020-12-03--17-31-01 182sec
f66032c2b5aa18ac_2020-12-03--17-31-01_182sec_STOPPING_BRAKE_RATE04

briskspirit:
Tried stopping brake rate 0.3 - feels good, will post graphs later

pd0wm:
Did you also increase the stopping target? If changing the value is a good fix for this problem we should indeed parameterize it.

briskspirit:
No, I change parameters one by one, so stopping target is the same. Will check graphs in the evening and parametrize it.

briskspirit:
Confirm that stopping brake rake 0.3 solves(or hides) the problem.
f66032c2b5aa18ac 2020-12-04--17-55-48 2293sec
@pd0wm what you say? If you agree I will parametrize this cost and add to every TSS2 car + to defaults as 0.2 

f66032c2b5aa18ac_2020-12-04--17-55-48_2293sec_STOPPING_BRAKE_RATE03_

briskspirit:
Ok, so I figured out the fix, will PR today.
f66032c2b5aa18ac 2020-12-05--14-33-16_549sec
STOPPING_BRAKE_RATE = 0.1 ! So now it stops fine and smooth.

f66032c2b5aa18ac_2020-12-05--14-33-16_549sec_STOPPING_BRAKE_RATE01

PR #2512, opened 2020-11-10
Cleaned up Prius entries a bit

zorrobyte:
Priuses should go together, some spelling errors; also, why is safetyParam 73 on TSS2 Prius when safetyParam is 66 in Prius? The conversion factor for STEER_TORQUE_EPS is the same across these models and they use the same EPS, rack, etc.


zorrobyte:
Ah,
"that said, comma has acknowledged that the tss2 safety param is incorrect so they're writing a script to get the true scale and update it" https://discordapp.com/channels/469524606043160576/574796986822295569/768674843087994911

Makes sense

zorrobyte:
Closing, found correct safetyParam

PR #2513, opened 2020-11-10
Cleaned up TSS2 Prius entry a bit, added correct safetyParam

zorrobyte:
As it says on the tin. Also, factory sR is 13.4.

safetyParam of 66 is correct (from "old" Prius), they didn't update the EPS, rack, etc.

Screen Shot 2020-11-10 at 1 22 33 AM
66 safetyParam

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -31,7 +31,7 @@ class CarInterface(CarInterfaceBase):
       stop_and_go = True
       ret.safetyParam = 66  # see conversion factor for STEER_TORQUE_EPS in dbc file
       ret.wheelbase = 2.70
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Can you verify the EPS factor with `selfdrive/debug/toyota_eps_factor.py`? If it needs to be changed, the DBC also needs to be updated.

zorrobyte:
@rav4kumar reports safetyparam of 55 is most accurate

Screenshot_114

Let me do some work and I'll verify the EPS factor with the script on both years

zorrobyte:
This may not be needed, pending https://github.com/commaai/openpilot/issues/2514

adeebshihadeh:
I'd like to verify all the toyota safety params and test them in CI, so this will get done at that point.

PR #2515, opened 2020-11-10
Add fingerprint for european/nordic Hyundai Kona EV 2019

Neptos:
Adds fingerprint for european/nordic Hyundai Kona EV 2019


adeebshihadeh:
What's your dongle ID?

Neptos:
@adeebshihadeh My dongle ID is 2fcb04d077ee7d28

PR #2516, opened 2020-11-10
UI: read LastAthenaPingTime in ui_draw_sidebar_connectivity

deanlee:


adeebshihadeh:
I think this is fine the way it is. Ideally, we should have one thread for updating state and another one handling the drawing.

deanlee:
yes u r right

PR #2518, opened 2020-11-10
use enabled flag from controlsState instead of carState

ZwX1616:


selfdrive/monitoring/dmonitoringd.py
```diff
@@ -22,12 +22,12 @@ def dmonitoringd_thread(sm=None, pm=None):
   sm['liveCalibration'].calStatus = Calibration.INVALID
   sm['liveCalibration'].rpyCalib = [0, 0, 0]
   sm['carState'].vEgo = 0.
-  sm['carState'].cruiseState.enabled = False
   sm['carState'].cruiseState.speed = 0.
   sm['carState'].buttonEvents = []
   sm['carState'].steeringPressed = False
   sm['carState'].gasPressed = False
   sm['carState'].standstill = True
+  sm['controlsState'].enabled = False
```
adeebshihadeh, line 30: this is already the default

ZwX1616: kk. can you fix the force decel in long test?

PR #2523, opened 2020-11-11
Intern mockup

kkuehlz:


grekiki:
Will take some code from here, but otherwise closing.

PR #2525, opened 2020-11-11
UI Refactor: update state and draw in separate threads

deanlee:
1.mv all states to class UIScene
2.update states from another thread

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -16,7 +16,6 @@
 #include "ui.hpp"
 #include "paint.hpp"
 
-
 int write_param_float(float param, const char* param_name, bool persistent_param) {
   char s[16];
   int size = snprintf(s, sizeof(s), "%f", param);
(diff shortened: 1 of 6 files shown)


adeebshihadeh:
This seems more complicated than it needs to be. If we can atomically read from a UI state struct in the drawing thread, then we can make this simpler.

deanlee:
Did u mean passing a state struct by value?

deanlee:
It’s still too complicated. I’ll think about how to make it better.

adeebshihadeh:
I think if we do this right, it’ll be easier to read than before :)

deanlee:
Done,I am satisfied with it personally.

deanlee:
the state thread processes and copies data extremely fast.now the main bottleneck is in function visionstream_get().most of the "slow frame" messages are caused by it.I'm confused why is it laggy

deanlee:
> the state thread processes and copies data extremely fast.now the main bottleneck is in function visionstream_get().most of the "slow frame" messages are caused by it.I'm confused why is it laggy

I now know why this happens, because of the wrong flag, poll did not work as expected. I submitted a PR to fix this: https://github.com/commaai/openpilot/pull/2550

adeebshihadeh:
Nice work! Let's merge #2550 into this PR. I'll have some time to review this after 0.8 is released.

selfdrive/ui/ui.cc
```diff
@@ -157,6 +157,22 @@ static void update_model(UIScene &scene, const cereal::ModelDataV2::Reader &mode
   update_line_data(scene, model.getPosition(), 0.5, 0, &scene.track_vertices, path_length);
 }
 
+static void update_lead(const cereal::RadarState::LeadData::Reader &lead,
```
deanlee, line 160: @HaraldSchafer: This function is still in the draft stage, do you think it is correct?

HaraldSchafer: Looks correct to me!

deanlee:
refactor alert   https://github.com/commaai/openpilot/pull/2751/files

deanlee:
convert to draft,the code of the qt part has been changing

adeebshihadeh:
I like the direction of this, but it's a big PR. Can we break it up into smaller PRs and build up to it?

deanlee:
How about using a PR to put all the data that needs to be synchronized into the struct UIScene? Then use another PR to complete the synchronization

adeebshihadeh:
Hmm, that still sounds like big PRs. If this takes big PRs, I think it will sit around for a while, since the UI is stable and not a high priority.

deanlee:
What about using variables instead of cereal::read to get data? If we want to update in the background, we can no longer use cereal::read

PR #2529, opened 2020-11-12
Torch Model: follow capnp best practices

deanlee:


deanlee:
coverted to draft ,I think https://github.com/commaai/openpilot/pull/2525 is much better than this one

PR #2530, opened 2020-11-12
ui_draw_line: remove std::clamp

deanlee:


pd0wm:
Do we even need this check? Won't the nvgScissor make sure nothing outside of the video gets drawn?

deanlee:
I think it might be better to guarantee the validity of the data by ourself

PR #2541, opened 2020-11-13
Simplifying code on numpy_fast

JonanOribe:
Making if statement easier to read


adeebshihadeh:
This isn't really more readable, and the tests are failing.

PR #2543, opened 2020-11-13
AcuraWatch is standard on 2019+ Acura RDX

csouers:
**Description**: Fix the supported car list in README. AcuraWatch is standard on the 3rd generation Acura RDX.

**Verification**: Acura's sales documentation


VirtuallyChris:
Whoops, already did this in master :) I'll close this now

PR #2548, opened 2020-11-14
Remove reference to nonexistent file

ghost:


adeebshihadeh:
We shouldn't entirely remove the section. Let's update it for the new tests that are run.

pd0wm:
Let's create a new script to run the tests locally. Is there a script that can automatically run github actions locally?

PR #2564, opened 2020-11-17
tici front/wide encodeIdx msgs

ZwX1616:
since they are all full HEVCs on tici, we need to distinguish them in the logs


ZwX1616:
also use lh_log instead of #2562

PR #2566, opened 2020-11-17
Remove local work queue from C2 debayer

geohot:
According to the profiler, this uses way less GPU. And considering the debayer10 kernel doesn't use any local memory, it should be fine.


ZwX1616:
lol don't know why it was there and don't understand how it is using more GPU, we don't even allocate any local memory for that... weird.

anyways LGTM. shouldn't have any concern removing that.

geohot:
Huh, the CPU usage test fails @adeebshihadeh

ZwX1616 (approved):

PR #2567, opened 2020-11-18
send frame in packet on demand

ZwX1616:
e.g. `ZMQ=1 SEND_REAR=1 XMIN=388 XMAX=776 YMIN=200 YMAX=500 SCALE=4 ./camerad`

### DIFF
--- a/selfdrive/camerad/cameras/camera_common.cc
+++ b/selfdrive/camerad/cameras/camera_common.cc
@@ -29,7 +29,6 @@ const int env_xmax = getenv("XMAX") ? atoi(getenv("XMAX")) : -1;
 const int env_ymin = getenv("YMIN") ? atoi(getenv("YMIN")) : 0;
 const int env_ymax = getenv("YMAX") ? atoi(getenv("YMAX")) : -1;
 const int env_scale = getenv("SCALE") ? atoi(getenv("SCALE")) : 1;
-const char *env_send_front = getenv("SEND_FRONT");
 
 static cl_program build_debayer_program(cl_device_id device_id, cl_context context, const CameraInfo *ci, const CameraBuf *b) {
   char args[4096];
--- a/selfdrive/camerad/cameras/camera_common.cc
+++ b/selfdrive/camerad/cameras/camera_common.cc
@@ -24,6 +24,13 @@
 #include "common/util.h"
 #include "imgproc/utils.h"
 
+const int env_xmin = getenv("XMIN") ? atoi(getenv("XMIN")) : 0;
+const int env_xmax = getenv("XMAX") ? atoi(getenv("XMAX")) : -1;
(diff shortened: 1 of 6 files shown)


selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -220,6 +220,30 @@ void fill_frame_data(cereal::FrameData::Builder &framed, const FrameMetadata &fr
   framed.setGainFrac(frame_data.gain_frac);
 }
 
+void fill_frame_image(cereal::FrameData::Builder &framed, uint8_t *dat, int w, int h, int stride) {
+  if (dat != nullptr) {
+    int scale = 1;
+    int x_min = 0; int y_min = 0; int x_max = w-1; int y_max = h-1;
+    if (getenv("XMIN")) x_min = atoi(getenv("XMIN"));
```
adeebshihadeh, line 227: check this only once

selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -24,6 +24,13 @@
 #include "common/util.h"
 #include "imgproc/utils.h"
 
+const int env_xmin = getenv("XMIN") ? atoi(getenv("XMIN")) : 0;
+const int env_xmax = getenv("XMAX") ? atoi(getenv("XMAX")) : -1;
+const int env_ymin = getenv("YMIN") ? atoi(getenv("YMIN")) : 0;
+const int env_ymax = getenv("YMAX") ? atoi(getenv("YMAX")) : -1;
+const int env_scale = getenv("SCALE") ? atoi(getenv("SCALE")) : 1;
+const char *env_send_front = getenv("SEND_FRONT");
```
adeebshihadeh, line 32:
```suggestion
const bool env_send_front = getenv("SEND_FRONT") != NULL;
```

selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -34,6 +34,8 @@
 
 
 extern volatile sig_atomic_t do_exit;
+const char *env_send_rear = getenv("SEND_REAR");
+const char *env_send_wide = getenv("SEND_WIDE");
```
adeebshihadeh, line 38: I'd put all these together in camera common.

pd0wm:
Is this what we want? Won't loggerd completely choke on this?

Why not write a visionipc to zmq bridge?

ZwX1616:
> Is this what we want? Won't loggerd completely choke on this?
> 
> Why not write a visionipc to zmq bridge?

From what I understand loggerd won't care if we use the ZMQ flag. The only downside is that the frame packets are diverted to somewhere else and thus don't get logged. But we will be debugging the camera(s) anyways, and can handle the msgs with the tool of choice (e.g. replay ui or fl ui).

PR #2578, opened 2020-11-19
fix frame drop percentage

haraschax:


selfdrive/modeld/modeld.cc
```diff
@@ -193,17 +193,18 @@ int main(int argc, char **argv) {
             model_eval_frame(&model, q, yuv_ion.buf_cl, buf_info.width, buf_info.height,
                              model_transform, NULL, vec_desire);
         mt2 = millis_since_boot();
+        float model_execution_time = (mt2 - mt1) / 1000.0;
```
adeebshihadeh, line 196: any reason this is in seconds?

HaraldSchafer: Yes, want as much as possible in SI units.

selfdrive/modeld/models/driving.cc
```diff
@@ -461,7 +466,9 @@ void model_publish(PubMaster &pm, uint32_t vipc_frame_id, uint32_t frame_id,
 }
 
 void posenet_publish(PubMaster &pm, uint32_t vipc_frame_id, uint32_t frame_id,
-                     uint32_t vipc_dropped_frames, float frame_drop, const ModelDataRaw &net_outputs, uint64_t timestamp_eof) {
+                     uint32_t vipc_dropped_frames, float frame_drop,
+                     const ModelDataRaw &net_outputs, uint64_t timestamp_eof,
+                     float model_execution_time) {
```
adeebshihadeh, line 471: this is unused

HaraldSchafer: fixed

PR #2581, opened 2020-11-20
TSS2 INDI tuning

birdman6450:
i dont expect this to be merged. i just wanted to call attention to some of the values used and encourage others to resubmit this tune in a more complete format. in order for this to be merged it may need to have dbc edited for safetyparam 53 (probably needs to be lived tuned as it varies on average per trim and even changes within the same car per minute), this tune also requires outerBP V support which i did not add yet and hope someone else will resubmit this tune with it, this tune also requires at least 3 routes from 3 seperate individuals which i do not have yet. for this tune to be complete it also requires steerup15 in panda/values however it does work well without it and it could be merged without it for now. i am submitting this pr, knowing it wont be merged, just to get the values seriously considered as i work on, doing injection test of steerup15, getting routes, adding support in PR for outerBP, and finding out if dbc needs to be edited or if interface alone can be edited  to change safetyparam. about 7 people have tested this and like it and anyone who approves of it can feel free to submit a new, complete PR, and take credit for it. i only want to help others have a better tune. if anyone wants to test it, it is for corollatss2 only, it is on jamcar23 fork r2++(full tune) and shanes SA(with no steerup15). i look forward to a more complete PR in the future. thanks for your consideration.

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -195,13 +195,19 @@ class CarInterface(CarInterfaceBase):
 
     elif candidate in [CAR.COROLLA_TSS2, CAR.COROLLAH_TSS2]:
       stop_and_go = True
-      ret.safetyParam = 73
       ret.wheelbase = 2.63906
-      ret.steerRatio = 13.9
-      tire_stiffness_factor = 0.444  # not optimized yet
       ret.mass = 3060. * CV.LB_TO_KG + STD_CARGO_KG
-      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.6], [0.1]]
-      ret.lateralTuning.pid.kf = 0.00007818594
+      ret.safetyParam = 53
+      ret.steerActuatorDelay = 0.60
+      ret.steerRatio = 15.33
+      tire_stiffness_factor = 0.996  # not optimized yet
+      ret.lateralTuning.init('indi')
+      ret.lateralTuning.indi.innerLoopGain = 9.0
+      ret.lateralTuning.indi.outerLoopGainBP = [20, 21, 25 ,26]
+      ret.lateralTuning.indi.outerLoopGainV = [5.0, 9.0, 9.0, 15.0]
+      ret.lateralTuning.indi.timeConstant = 5.5
+      ret.lateralTuning.indi.actuatorEffectiveness = 9.0
+      
 
     elif candidate in [CAR.LEXUS_ES_TSS2, CAR.LEXUS_ESH_TSS2]:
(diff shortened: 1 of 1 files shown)


pd0wm:
Just a quick note. The outer loop controller should always be slower than the inner loop (so the gain must be lower), otherwise you end up with a potentially unstable system. To be upstreamed at least the tuning needs to have this property.

birdman6450:
> The outer loop controller should always be slower than the inner loop (so the gain must be lower),

i was advised the same thing last time and i took a step back to retest. the advise was only valuable in that lowering outer below inner below 45mph eliminated ping pong on slower curves, and that produced the current outerBP profile. lowering outer below inner above 60mph is not advantagious in practice and causes lazy lane centering on the freeway. in effect, outer is lane centering strictness, my outerBP profile has never ping pong, never sway, and with this tune it feels tight like a guitar string. i do 4k of the full spectrum of rigorious road tests per month and i have found no other way to produce accurate rubber on road placement with no ping pong and no sway. my tune produces the highest probability of accurate of tire placement. as indi applies to cars, a new rule book may be needed based on my tests and i think my tune should serve as a template especially as it relate to inner vs. outer. inner is curvature gain, outer is lane centering gain, time is smoothness and roundness, and effectiveness is the smart use of that gain. there may be more than one tune that can produce the same results but i have yet to find more than one. again, my tests have concluded the idea "outer always needs to be lower than inner" is false, but i hope i am wrong and i hope that someone produces a better tune than mine as it can always improve. i have shared my reasons for my values based on tested, please share your reason why inner should be higher than outer, what tests have you done with this particular system (tss2 indi) or is it based off a paper(please send link) ?

birdman6450:
i may have found a tune that allows inner to be higher than outer. i maintains performance but reduces hyper occilation that was tight like a guitar string. if it continues to pass my tests i will modify pr or submit a new one. thanks willem. also i would like to understand how you came to that concluaion out of curiousity.

birdman6450:
> Just a quick note. The outer loop controller should always be slower than the inner loop (so the gain must be lower), otherwise you end up with a potentially unstable system. To be upstreamed at least the tuning needs to have this property.

i updated the tune so that inner is always higher than outer so that it can be seriously considered for upstream. this new tune passed all my initial tests on corner case turns, no ping pong, no sway, improved cornering, improved lane centering reaction time, improved smoothness. thanks willem. i usually like to test a new tune for several hundred miles before reporting on it but i can sense already that this is excellent and i will report back if i ever feel any errors.

birdman6450:
Have out several thousand miles on this tune now just to confirm consistancy. I will submit routes in a few weeks.

PR #2586, opened 2020-11-20
health is not None was checked twice

sshane:


adeebshihadeh:
This will now show the hardware unsupported alert when there's no health packet.

Issue #2585, opened 2020-11-20
Model lagging after update to revision a6e8778ae6ee6ef4a8ea6805a5457a827b8c0355


briskspirit:
**Describe the bug**

Latest master-ci (revision a6e8778ae6ee6ef4a8ea6805a5457a827b8c0355) totally broke 0.8.0 . Constantly getting model lagging error. 
Updated from revision 3328cf50b49a9ba429f249394190728fdbb640ab

**How to reproduce or log data**

Update to git revision a6e8778ae6ee6ef4a8ea6805a5457a827b8c0355

**Expected behavior**

No lagging

**Device/Version information (please complete the following information):**
 - Device: Comma Two
 - Dongle ID: f66032c2b5aa18ac
 - Route: f66032c2b5aa18ac
 - Timestamp: 2020-11-20--09-27-38, 2020-11-20--09-23-54, 2020-11-20--09-22-33
 - Version: a6e8778ae6ee6ef4a8ea6805a5457a827b8c0355
 - Car make/model: Toyota Corolla 2021 SE

geohot:
Tracking in #2546, there was an issue in the lag detector that was fixed recently.

briskspirit:
Model lagging unfortunately was not fixed with your commit...

Device: f66032c2b5aa18ac
Route: f66032c2b5aa18ac|2020-11-20--17-39-47
Exact segments:
f66032c2b5aa18ac|2020-11-20--17-39-47--4
f66032c2b5aa18ac|2020-11-20--17-39-47--5
f66032c2b5aa18ac|2020-11-20--17-39-47--6
f66032c2b5aa18ac|2020-11-20--17-39-47--14

Now it was definitely better than in the morning (when I had panics after 5-10 sec of driving)
Commit I was on : f10783b57501e123857502de8358f04a20095bf5

geohot:
Can confirm model lags on "f66032c2b5aa18ac/2020-11-20--17-39-47/4"

But also, dmonitoring at 14ms. So that commit seemed to fix that issue, but not enough for the main model lags. Perhaps the governors will fix it.

briskspirit:
Could you also check f66032c2b5aa18ac|2020-11-19--17-35-53 ? My point is that on previous commit there was no lagging panics. Maybe something got broken by another commit? Just to compare on the same device/car/etc.

geohot:
The lag detector was tightened yesterday. The lag was always there, you just didn't know about it.

[...]

PR #2590, opened 2020-11-21
New settings for governors, +30% memory bandwidth

geohot:
Fix model lag?


geohot:
Merge if tests pass?

I also added an ignore for the first 10 frames while modeld gets set up.

geohot:
Something is still going on. I killed ui and surfaceflinger, still dropping frame with waste. It seems like just modeld, camerad, and waste alone don't drop frames.

```
root@localhost:/data/openpilot$ lsof | grep kgsl
3210    /data/openpilot/selfdrive/camerad/camerad       /dev/kgsl-3d0
5195    /data/openpilot/selfdrive/modeld/_modeld        /dev/kgsl-3d0
```

Working on #2585

geohot:
Reverting loggerd priority fixes model lags with waste running. Reverted dmonitoringd priority for good measure too.

PR #2592, opened 2020-11-21
update frame drop filter

adeebshihadeh:


geohot:
Merge this? I think it's reasonable.

PR #2600, opened 2020-11-22
agnos updater

adeebshihadeh:


selfdrive/updated.py
```diff
@@ -219,6 +223,58 @@ def finalize_update() -> None:
...
+
+  current_slot = subprocess.check_output(["abctl", "--boot_slot"], encoding='utf-8').strip()
+  target_slot = "_b" if current_slot == "_a" else "_a"
+  target_slot_number = "0" if target_slot == "_a" else "1"
+
+  # set target slot as unbootable
+  #os.system(f"abctl --set_unbootable {target_slot_number}")
+
+  for partition in update:
+    img_path = f"/tmp/{partition['name']}_{partition['hash']}.img"
```
robbederks, line 264: Isn't `/tmp` too small to hold the system partition image?
Also, the uncompressed image is still sparse, which doesn't seem to be handled. Does this work?
We might need to stream it directly into the partition, not save the image first

selfdrive/updated.py
```diff
@@ -219,6 +223,58 @@ def finalize_update() -> None:
...
+        decompressed_chunk = decompressor.decompress(chunk)
+        sha256.update(decompressed_chunk)
+        f.write(decompressed_chunk)
+
+  if sha256.hexdigest().lower() != img_hash.lower():
+    raise Exception(f"'{filename}' hash mismatch, got {sha256.hexdigest().lower()}")
+
+
+def handle_agnos_update(wait_helper):
+  with open(NEOS_VERSION, "r") as f:
```
robbederks, line 242: Maybe rename `NEOS_VERSION` as well?

adeebshihadeh: moved to hw abstraction layer

launch_chffrplus.sh
```diff
@@ -104,6 +104,55 @@ function two_init {
...
+
+  # Check if AGNOS update is required
+  if [ $(< /VERSION) != "$AGNOS_VERSION" ]; then
+    # Get number of slot to switch to
+    CUR_SLOT=$(abctl --boot_slot)
+    if [[ "$CUR_SLOT" == "_a" ]]; then
+      OTHER_SLOT="_b"
+      OTHER_SLOT_NUMBER="1"
+    else
+      OTHER_SLOT="_b"
```
adeebshihadeh, line 119: wrong slot

PR #2605, opened 2020-11-23
modeld: eliminate the memory copy to GPU 

pd0wm:
Continuation of https://github.com/commaai/openpilot/pull/2197

### DIFF
--- a/selfdrive/modeld/modeld.cc
+++ b/selfdrive/modeld/modeld.cc
@@ -148,8 +148,7 @@ int main(int argc, char **argv) {
     const float frame_filter_k = (dt / ts) / (1. + dt / ts);
     float frames_dropped = 0;
 
-    // one frame in memory
-    VisionBuf yuv_ion = visionbuf_allocate_cl(buf_info.buf_len, device_id, context);
+    std::map<void*, cl_mem> cl_buffers;
 
     uint32_t frame_id = 0, last_vipc_frame_id = 0;
     double last = 0;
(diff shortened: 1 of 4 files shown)


pd0wm:
Doesn't pass the model replay test. Need to check why the model outputs change.

selfdrive/modeld/modeld.cc
```diff
@@ -210,7 +212,9 @@ int main(int argc, char **argv) {
       }
 
     }
-    visionbuf_free(&yuv_ion);
+    for(auto &v : cl_buffers) {
+      clReleaseMemObject(v.second);
```
mitchellgoffpc, line 216: is clReleaseMemObject all we need here? visionbuf_free also calls munmap and closes the buffer's file descriptor

mitchellgoffpc:
> Doesn't pass the model replay test. Need to check why the model outputs change.

geohot:
The reason the model output change I believe is because the test is broken.

If using the ION implementation for visionbuf_allocate_cl, the memcpy is okay since that's allocated with the UNCACHED flag. But it doesn't work for the CL implementation.

So I believe we've been testing with 0s the whole time.

selfdrive/modeld/modeld.cc
pd0wm: On the receiving end we don't have a fd to the ion buffer handle to close

pd0wm:
I made another PR to add the sync: https://github.com/commaai/openpilot/pull/2675. Let's see if that also breaks the tests.

PR #2606, opened 2020-11-23
Prerotate dmonitoringmodeld input

geohot:
This should fix the cache issues for good.

https://chromium.googlesource.com/libyuv/libyuv/+/HEAD/docs/rotation.md uses an 8x8 kernel, the proper way to rotate.

Despite having an extra buffer, this actually reduces runtime. I do not expect it to be variable anymore.

Also fixes a bug in input order!

### DIFF
--- a/selfdrive/modeld/models/dmonitoring.cc
+++ b/selfdrive/modeld/models/dmonitoring.cc
@@ -114,7 +114,45 @@ DMonitoringResult dmonitoring_eval_frame(DMonitoringModelState* s, void* stream_
                     resized_width, resized_height,
                     mode);
 
+  // new, with prerotate
+
+  uint8_t *resized_buf_rot = get_buffer(s->resized_buf, resized_width*resized_height*3/2);
+  uint8_t *resized_y_buf_rot = resized_buf;
+  uint8_t *resized_u_buf_rot = resized_y_buf + (resized_width * resized_height);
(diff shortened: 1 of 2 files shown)


selfdrive/modeld/models/dmonitoring.cc
```diff
@@ -114,24 +114,39 @@ DMonitoringResult dmonitoring_eval_frame(DMonitoringModelState* s, void* stream_
...
   // one shot conversion, O(n) anyway
   // yuvframe2tensor, normalize
   for (int c = 0; c < MODEL_WIDTH/2; c++) {
     for (int r = 0; r < MODEL_HEIGHT/2; r++) {
       // Y_ul
-      net_input_buf[(c*MODEL_HEIGHT/2) + r] = input_lambda(resized_buf[(2*r*resized_width) + (2*c)]);
-      // Y_ur
-      net_input_buf[(c*MODEL_HEIGHT/2) + r + (2*(MODEL_WIDTH/2)*(MODEL_HEIGHT/2))] = input_lambda(resized_buf[(2*r*resized_width) + (2*c+1)]);
+      net_input_buf[(c*MODEL_HEIGHT/2) + r + (0*(MODEL_WIDTH/2)*(MODEL_HEIGHT/2))] = input_lambda(resized_buf_rot[(2*r) + (2*c)*resized_height]);
       // Y_dl
```
ZwX1616, line 140: ur and dl should be swapped back since it's rotated.
besides that everything lgtm.

PR #2611, opened 2020-11-23
raw pred api

ZwX1616:


selfdrive/modeld/dmonitoringmodeld.cc
```diff
@@ -66,7 +66,11 @@ int main(int argc, char **argv) {
       double t2 = millis_since_boot();
 
       // send dm packet
-      dmonitoring_publish(pm, extra.frame_id, res, (t2-t1)/1000.0);
+      const float* raw_pred_ptr = nullptr;
+        if (env_send_raw_dmonitor) {
```
geohot, line 70: spacing.

selfdrive/modeld/models/dmonitoring.cc
```diff
@@ -186,6 +186,9 @@ void dmonitoring_publish(PubMaster &pm, uint32_t frame_id, const DMonitoringResu
   framed.setLeftBlinkProb(res.left_blink_prob);
   framed.setRightBlinkProb(res.right_blink_prob);
   framed.setSgProb(res.sg_prob);
+  if (env_send_raw_dmonitor) {
```
geohot, line 189: if (raw_pred) {

Don't check env in both places.

selfdrive/modeld/models/dmonitoring.h
```diff
@@ -11,6 +11,8 @@ extern "C" {
 
 #define OUTPUT_SIZE 34
 
+const bool env_send_raw_dmonitor = getenv("SEND_RAW_DMONITOR") != NULL;
```
geohot, line 14: Wait does this work in C?

pd0wm: Think this works in C++ only. Functions like this get called before main. If you create objects their destructors are called when main returns.

selfdrive/modeld/modeld.cc
```diff
@@ -204,8 +204,12 @@ int main(int argc, char **argv) {
         if (run_count < 10) frames_dropped = 0;  // let frame drops warm up
         float frame_drop_ratio = frames_dropped / (1 + frames_dropped);
 
-        model_publish(pm, extra.frame_id, frame_id,  vipc_dropped_frames, frame_drop_ratio, model_buf, extra.timestamp_eof, model_execution_time);
-        model_publish_v2(pm, extra.frame_id, frame_id,  vipc_dropped_frames, frame_drop_ratio, model_buf, extra.timestamp_eof, model_execution_time);
+        const float* raw_pred_ptr = nullptr;
+        if (env_send_raw_driving) {
+          raw_pred_ptr = (const float *)model.output;
```
stevencleary, line 209: Same thing with inlining here:
`const float* raw_pred_ptr = env_send_raw_driving ? (const float *)model.output : nullptr;`

PR #2613, opened 2020-11-24
tici timing improvements

ZwX1616:


selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -195,11 +195,12 @@ bool CameraBuf::acquire() {
   pool_acquire(&yuv_pool, cur_yuv_idx);
   pool_push(&yuv_pool, cur_yuv_idx);
 
+  tbuffer_dispatch(&ui_tb, cur_rgb_idx);
```
geohot, line 198: What does this change? Why can this be here?

ZwX1616: this change skips the callbacks which are just autoexposures and logging.

*should* make ui feel a bit faster if anything

PR #2618, opened 2020-11-24
Wifi improvements

ghost:


grekiki:
We now detect wrong passwrods and request the correct password from the user. We also write "CONNECTING" if we are actively connecting to a network. 

For a tiny usability upgrade, we now display "Refreshing networks" text, while loading the wifi networks available around us.

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -259,3 +295,12 @@ QString WifiManager::get_adapter(){
 
   return adapter_path;
 }
+
+void WifiManager::change(unsigned int a,unsigned int b,unsigned int c){
+  raw_adapter_state = a;
+  if(a==60 && c==8){
+    wrongPassword(connecting_to_network);
```
pd0wm, line 302: use `emit` for signals

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -259,3 +295,12 @@ QString WifiManager::get_adapter(){
 
   return adapter_path;
 }
+
+void WifiManager::change(unsigned int a,unsigned int b,unsigned int c){
+  raw_adapter_state = a;
+  if(a==60 && c==8){
```
pd0wm, line 301: Make defines for these magic values at the top of the file.

pd0wm:
Remove all qDebugs()

PR #2621, opened 2020-11-25
Make the DSP work everywhere

geohot:
Doesn't work yet...but it should

### DIFF
--- a/common/realtime.py
+++ b/common/realtime.py
@@ -4,15 +4,20 @@ import os
 import time
 import multiprocessing
 
-from common.hardware import PC
+from common.hardware import PC, TICI
 from common.clock import sec_since_boot  # pylint: disable=no-name-in-module, import-error
 
 
 # time step for each process
 DT_CTRL = 0.01  # controlsd
 DT_MDL = 0.05  # model
-DT_DMON = 0.1  # driver monitoring
 DT_TRML = 0.5  # thermald and manager
+
+# driver monitoring
+if TICI:
+  DT_DMON = 0.05
+else:
+  DT_DMON = 0.1
 
 
 class Priority:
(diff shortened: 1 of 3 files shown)


launch_chffrplus.sh
```diff
@@ -11,6 +11,11 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 function tici_init {
   sudo su -c 'echo "performance" > /sys/class/devfreq/soc:qcom,memlat-cpu0/governor'
   sudo su -c 'echo "performance" > /sys/class/devfreq/soc:qcom,memlat-cpu4/governor'
+
+  # init the DSP
+  sudo systemctl start adsprpcd
+  sudo systemctl start cdsp
+  sudo chmod 666 /dev/adsprpc-smd
```
adeebshihadeh, line 18: this should be done in a boot script

pd0wm: Shouldn’t it just be a udev rule?

common/realtime.py
```diff
@@ -4,16 +4,21 @@
...
 
 
 # time step for each process
 DT_CTRL = 0.01  # controlsd
 DT_MDL = 0.05  # model
-DT_DMON = 0.1  # driver monitoring
 DT_TRML = 0.5  # thermald and manager
 
+# driver monitoring
+if TICI:
```
adeebshihadeh, line 17: we need to handle this in service_list at some point too. our comm issue check is wrong for frontFrame/driverState/dMonitoringState for example

pd0wm: Agree! Ideally these values should also be pulled from the service list.

PR #2625, opened 2020-11-25
add legacy reset line support

robbederks:
On old tici, run `touch /persist/LEGACY_PANDA_RESET`


geohot:
Hmm, the panda should be on be default without the SOM on. Is this still true?

robbederks:
It is now, didn't use to be the case

PR #2626, opened 2020-11-25
pad vlayout

ghost:
We pad vlayout on wifi settings to ensure widgets don't grow too large.


grekiki:
Tested on device, merging

PR #2630, opened 2020-11-26
Move thermald hardware calls into HW abstraction layer

robbederks:

### DIFF
--- a/common/hardware.py
+++ b/common/hardware.py
@@ -52,19 +52,19 @@ class Pc(HardwareBase):
     return 100
 
   def get_battery_status(self):
-    raise NotImplementedError()
+    return ""
 
   def get_battery_current(self):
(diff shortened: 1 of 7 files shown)


common/hardware.py
```diff
@@ -48,6 +48,30 @@ def get_sim_info(self):
...
+    return 0
+
+  def get_battery_voltage(self):
+    return 0
+
+  def get_battery_charging(self):
+    return True
+
+  def set_battery_charging(self, on):
+    print(f"SET BATTERY CHARGING: {on}")
```
adeebshihadeh, line 67: why print this on PC?

robbederks: meh, not necessary, was based on the fact that reboot also had a print (which in hindsight is way less common, so I'll remove it)

PR #2634, opened 2020-11-27
manager cleanup

adeebshihadeh:


selfdrive/manager.py
```diff
@@ -333,12 +317,12 @@ def prepare_managed_process(p):
...
+      subprocess.check_call(["scons", "u", "-j4", "."], cwd=os.path.join(BASEDIR, proc[0]))
     except subprocess.CalledProcessError:
-      # make clean if the build failed
-      cloudlog.warning("building %s failed, make clean" % (proc, ))
-      subprocess.check_call(["make", "clean"], cwd=os.path.join(BASEDIR, proc[0]))
-      subprocess.check_call(["make", "-j4"], cwd=os.path.join(BASEDIR, proc[0]))
+      # clean and retry if the build failed
+      cloudlog.warning("building %s failed, cleaning and retrying" % (proc, ))
+      subprocess.check_call(["scons", "-u", "-c", "."], cwd=os.path.join(BASEDIR, proc[0]))
+      subprocess.check_call(["scons", "-u", "-j4", "."], cwd=os.path.join(BASEDIR, proc[0]))
```
adeebshihadeh, line 325: @pd0wm thoughts on removing this? it hasn't done anything for several versions and fixing it will noticeably slow startup

adeebshihadeh: hmm, the building individual processes is nice for things like tests. maybe we add an arg to build and don't for manager's purposes?

pd0wm: That should have been removed after the scons migration. It can go completely.

PR #2639, opened 2020-11-27
Qt ui: improve toggle buttons

ghost:
Make nicer toggle buttons


grekiki:
UI tested on device

grekiki:
Params also tested on device.

PR #2640, opened 2020-11-27
Add new speed polynomial for feedforward

sshane:
Fitted poly using current TSS1 Corolla `kf` of `0.00003`, so changing `kf` shouldn't be necessary. (I'm unsure how this will change other cars though, let me know how you'd like me to handle this if this PR is wanted)

I really recommend checking out the differences of stock feedforward and this new fitted feedforward poly: https://imgur.com/a/uCKZFxB

### DIFF
--- a/selfdrive/controls/lib/latcontrol_pid.py
+++ b/selfdrive/controls/lib/latcontrol_pid.py
@@ -34,7 +34,8 @@ class LatControlPID():
       if CP.steerControlType == car.CarParams.SteerControlType.torque:
         # TODO: feedforward something based on path_plan.rateSteers
         steer_feedforward -= path_plan.angleOffset   # subtract the offset, since it does not contribute to resistive torque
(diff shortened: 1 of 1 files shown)


pd0wm:
This PR is not mergable since it changes the behavior of other cars.

Before I'm considering all these new tuning values to PID (feedforward, D term, etc). I want to see how well INDI lends itself to automatic tuning. If lateral control would be more auto-tuned I feel like we could gain a lot of performance on a lot of cars (especially the ones with bad tuning that we might not even know about).

ShaneSmiskol:
Okay, but I hope you can find a solution to INDI's biggest problem of not applying torque when the wheel is oscillating from bumps in the road. INDI wasn't able to correct for this and just let the car slowly leave the path and eventually the road if the ground is too bumpy. PID handles this perfectly, maybe the integral or feedforward portion of it helps?

PR #2644, opened 2020-11-28
refactor params

deanlee:
1. remove functions write_param_float, read_param, read_db_bool, read_db_bytes
2. add template functions get(),put()
3. rewrite read_db_value to to eliminate memory copy


adeebshihadeh:
Nice, we've been planning to do this since the initial params refactor. We also wanted to make the python and C API the same. Can we make the C API match the python one?

selfdrive/common/params.h
```diff
@@ -1,46 +1,45 @@
...
-  int read_db_value_blocking(const char* key, char** value, size_t* value_sz);
-
-  int read_db_all(std::map<std::string, std::string> *params);
-  std::vector<char> read_db_bytes(const char* param_name);
-  bool read_db_bool(const char* param_name);
-
-  std::string get(std::string key, bool block=false);
+  Params(std::string_view path);
+  
+  bool delete_value(std::string key);
```
adeebshihadeh, line 17: let’s make this just delete to match the python

selfdrive/loggerd/loggerd.cc
```diff
@@ -514,28 +514,20 @@ kj::Array<capnp::word> gen_init_data() {
   if (!clean) {
     init.setDirty(true);
   }
-  Params params = Params();
 
-  std::vector<char> git_commit = params.read_db_bytes("GitCommit");
-  if (git_commit.size() > 0) {
-    init.setGitCommit(capnp::Text::Reader(git_commit.data(), git_commit.size()));
+  Params params = Params();
+  if (std::string v = params.get("GitCommit"); v.size() > 0) {
```
adeebshihadeh, line 519: do we still need this check?

selfdrive/loggerd/loggerd.cc
```diff
@@ -613,7 +605,7 @@ int main(int argc, char** argv) {
   }
   bool record_front = true;
 #ifndef QCOM2
-  record_front = Params().read_db_bool("RecordFront");
+  record_front = Params().get<bool>("RecordFront").value_or(false);
```
adeebshihadeh, line 608: bool params are pretty common. maybe make a helper?

selfdrive/common/params.h
deanlee: delete is a reserved C++ keyword.we can't use it as function name.

selfdrive/loggerd/loggerd.cc
deanlee: how about add a new function getBool()?

deanlee:
don't know why the test always failed

deanlee:
Need to rearrange the usage of string_view

adeebshihadeh:
It looks like the tests are timing out in CI, probably due to bug that's causing params to block in the refactor.

deanlee:
> It looks like the tests are timing out in CI, probably due to a bug that's causing params to block in the refactor.

u r right,there are bugs in ensure_symlink,which always return false

deanlee:
Delete obsolete test_params.c

selfdrive/boardd/boardd.cc
```diff
@@ -73,12 +73,10 @@ void safety_setter_thread() {
       return;
     };
 
-    std::vector<char> value_vin = Params().read_db_bytes("CarVin");
-    if (value_vin.size() > 0) {
+    if (std::string vin = Params().get("CarVin"); vin.size() > 0) {
```
adeebshihadeh, line 76: use two lines here

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -169,16 +168,15 @@ QWidget * developer_panel() {
...
-    {"Panda Firmware", params.get("PandaFirmwareHex", false)},
+    {"Version", brand + " v" + params.get("Version")},
+    {"Git Branch", params.get("GitBranch")},
+    {"Git Commit", params.get("GitCommit").substr(0, 10)},
+    {"Panda Firmware", params.get("PandaFirmwareHex")},
   };
 
-  std::string os_version = util::read_file("/VERSION");
-  if (os_version.size()) {
+  if (std::string os_version = util::read_file("/VERSION"); os_version.size()) {
```
adeebshihadeh, line 179: use two lines

selfdrive/loggerd/loggerd.cc
```diff
@@ -469,28 +460,14 @@ kj::Array<capnp::word> gen_init_data() {
...
-    // log params
-    std::map<std::string, std::string> params_map;
-    params.read_db_all(&params_map);
+  Params params = Params();
+  init.setGitCommit(params.get("GitCommit"));
+  init.setGitBranch(params.get("GitBranch"));
+  init.setGitRemote(params.get("GitRemote"));
+  init.setPassive(params.getBool("Passive"));
+  // log params
+  if (std::map<std::string, std::string> params_map; params.read_all(params_map)) {
```
adeebshihadeh, line 470: use two lines here

selfdrive/ui/qt/offroad/settings.cc
adeebshihadeh, line 179: use two lines here

selfdrive/ui/qt/offroad/onboarding.cc
```diff
@@ -35,7 +35,7 @@ QWidget * OnboardingWindow::terms_screen() {
   QPushButton *accept_btn = new QPushButton("Accept");
   main_layout->addWidget(accept_btn, 2, 1);
   QObject::connect(accept_btn, &QPushButton::released, [=]() {
-    Params().write_db_value("HasAcceptedTerms", LATEST_TERMS_VERSION);
+    Params().put("HasAcceptedTerms", LATEST_TERMS_VERSION, strlen(LATEST_TERMS_VERSION));
```
adeebshihadeh, line 38: why does this need strlen? it should be passed in as a string

selfdrive/common/params.h
```diff
@@ -1,46 +1,47 @@
...
+  std::optional<T> get(const char *param_name, bool block = false) {
+    if (auto data = read_value(param_name, block)) {
+      std::istringstream iss(*data);
+      T value{};
+      iss >> value;
+      if (!iss.fail()) return value;
+    }
+    return std::nullopt;
+  }
+  inline bool getBool(const char *param_name, bool block = false) {
```
adeebshihadeh, line 34: the bool helper should take a string, otherwise we'll have `.c_str()` everywhere

selfdrive/common/params.h
```diff
@@ -1,46 +1,47 @@
...
-
-  std::string get(std::string key, bool block=false);
+  bool delete_value(std::string key);
+  bool read_all(std::map<std::string, std::string> &params);
+
+  inline std::string get(std::string key, bool block = false) {
+    return read_value(key.c_str(), block).value_or("");
+  }
+  template <class T>
+  std::optional<T> get(const char *param_name, bool block = false) {
```
adeebshihadeh, line 25: why do we have functions defined in the header?

deanlee: we use it to read other types such as float,double. for example:
`float brightness_b = params.get<float>("BRIGHTNESS_B").value_or(-1.)`

adeebshihadeh:
Can you rebase?

deanlee:
temporarily convert to draft due to a compile error

deanlee:
ready for review

adeebshihadeh:
Can you rebase?

deanlee:
Converted to draft to continue adjust, the master has changed too much

deanlee:
reddy for review

deanlee:
Need an interface to read vector.

selfdrive/boardd/boardd.cc
```diff
@@ -146,13 +145,13 @@ bool usb_connect() {
...
 
-    params.write_db_value("PandaFirmwareHex", fw_sig_hex_buf, 16);
+    params.put("PandaFirmwareHex", fw_sig_hex_buf, 16);
     LOGW("fw signature: %.*s", 16, fw_sig_hex_buf);
   } else { return false; }
 
   // get panda serial
-  if (auto serial = tmp_panda->get_serial(); serial) {
-    params.write_db_value("PandaDongleId", serial->c_str(), serial->length());
+  if (auto serial = panda->get_serial(); serial) {
```
pd0wm, line 153:
```suggestion
  if (auto serial = tmp_panda->get_serial(); serial) {
```

Mistake from rebasing?

pd0wm:
Like the changes using optional and the templates.

I'd prefer to not touch the implementations or only make minimal changes in `write_db_value`,`delete_db_value` and `read_all` (using the get method is fine though instead of reimplementing the read) unless something is actually broken. The current code might not be the shortest, but it's well commented and reasonable clear.

deanlee:
OIC

deanlee:
How about this one? replaced with a very python-like interface, and use  `util:read_file `to read file : https://github.com/commaai/openpilot/pull/20506 
It is difficult to continue to split into smaller PRs.

pd0wm:
That looks manageable. Thanks!

PR #2647, opened 2020-11-29
Fix toyota_eps_factor.py script

sshane:
Previously the script was using samples where the user was touching the wheel, when that happens the EPS torque starts to behave similarly to `driverTorque`, not ideal. This now skips those samples.

What happens when `steeringPressed` is true:

Screenshot_27

I'm on my Windows laptop which doesn't have Docker to test this, anyone want to see how this changes the results? They should be much more consistent. Previously, I would get widely varied results from the same car from different routes.

### DIFF
--- a/selfdrive/debug/toyota_eps_factor.py
+++ b/selfdrive/debug/toyota_eps_factor.py
@@ -18,14 +18,12 @@ def to_signed(n, bits):
 
 
 def get_eps_factor(lr, plot=False):
-  all_msgs = sorted(lr, key=lambda msg: msg.logMonoTime)
-
   engaged = False
   steering_pressed = False
   torque_cmd, eps_torque = None, None
   cmds, eps = [], []
 
-  for msg in all_msgs:
+  for msg in lr:
     if msg.which() != 'can':
       continue
 
--- a/selfdrive/debug/toyota_eps_factor.py
+++ b/selfdrive/debug/toyota_eps_factor.py
@@ -2,26 +2,32 @@
 import sys
 import numpy as np
 import matplotlib.pyplot as plt
-from sklearn import linear_model # pylint: disable=import-error
+from sklearn import linear_model  # pylint: disable=import-error
 
 from tools.lib.route import Route
 from tools.lib.logreader import MultiLogIterator
 
+MIN_SAMPLES = 30 * 100
 
-MIN_SAMPLES = 30*100
 
 def to_signed(n, bits):
   if n >= (1 << max((bits - 1), 0)):
     n = n - (1 << max(bits, 0))
   return n
 
+
 def get_eps_factor(lr, plot=False):
+  all_msgs = sorted(lr, key=lambda msg: msg.logMonoTime)
 
   engaged = False
+  steering_pressed = False
   torque_cmd, eps_torque = None, None
   cmds, eps = [], []
 
-  for msg in lr:
+  for msg in all_msgs:
+    if msg.which() == 'carState':
+      steering_pressed = msg.carState.steeringpressed
+
     if msg.which() != 'can':
       continue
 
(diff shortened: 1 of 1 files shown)


selfdrive/debug/toyota_eps_factor.py
```diff
@@ -2,26 +2,32 @@
...
 
   engaged = False
+  steering_pressed = False
   torque_cmd, eps_torque = None, None
   cmds, eps = [], []
 
-  for msg in lr:
+  for msg in all_msgs:
+    if msg.which() == 'carState':
+      steering_pressed = msg.carState.steeringpressed
```
adeebshihadeh, line 29:
```suggestion
      steering_pressed = msg.carState.steeringPressed
```

selfdrive/debug/toyota_eps_factor.py
```diff
@@ -2,26 +2,32 @@
...
+
```
ShaneSmiskol, line 30: Is this the correct way to get the `steeringPressed` message when we're at a CAN message? Basically cache them into a variable and just reference it every time we're in a CAN message?

selfdrive/debug/toyota_eps_factor.py
ShaneSmiskol: thanks! fixed

selfdrive/debug/toyota_eps_factor.py
adeebshihadeh: ideally, all of this should just use the can parser, but then you have to account for the factor of the DBC you use since you want the raw signal. i'd just parse out steeringPressed from the can msg instead of using carState.

ShaneSmiskol: How would I get that info from the can signal? if `STEER_TORQUE_DRIVER` is > 100?

adeebshihadeh: yeah, that's how we do it in carstate

ShaneSmiskol: Okay, I was just concerned that the way it's looping through all the messages one by one and not as frames that the last carState message might be out of sync when we're at the CAN data. I'm not really sure how to parse the binary CAN data but I'll try to at least start on getting it from the CAN message

selfdrive/debug/toyota_eps_factor.py
```diff
@@ -37,6 +35,7 @@ def get_eps_factor(lr, plot=False):
         torque_cmd = to_signed((m.dat[1] << 8) | m.dat[2], 16)
       elif m.address == 0x260 and m.src == 0:
         eps_torque = to_signed((m.dat[5] << 8) | m.dat[6], 16)
+        steering_pressed = abs(to_signed((m.dat[1] << 8) | m.dat[2], 16)) > STEER_THRESHOLD
```
ShaneSmiskol, line 38: Is this right?

selfdrive/debug/toyota_eps_factor.py
```diff
@@ -2,26 +2,30 @@
...
-MIN_SAMPLES = 30*100
 
 def to_signed(n, bits):
   if n >= (1 << max((bits - 1), 0)):
     n = n - (1 << max(bits, 0))
   return n
 
+
 def get_eps_factor(lr, plot=False):
+  all_msgs = sorted(lr, key=lambda msg: msg.logMonoTime)
```
adeebshihadeh, line 21: revert this, it's not necessary

selfdrive/debug/toyota_eps_factor.py
adeebshihadeh: yeah, that looks right

selfdrive/debug/toyota_eps_factor.py
ShaneSmiskol: 👍

selfdrive/debug/toyota_eps_factor.py
ShaneSmiskol: awesome! do you mind running it on a route of mine? curious to know how much more accurate it is now: 14431dbeedbf3558%7C2020-11-27--13-03-23

Should be 0.88 to 0.9

PR #2648, opened 2020-11-29
Add ability to manually set token if running in docker/without GUI

sshane:
I had to start Python and import the `set_token` function manually without this change, this makes it a bit easier


gregjhogan:
I would try mounting the token file into the container as a volume

adeebshihadeh:
Yeah, I would either mount as a volume or write the file manually. Not sure how this is easier than writing the file.

PR #2679, opened 2020-12-03
Added model_replay and fixed some bugs in camera_replay

mitchellgoffpc:
This PR shouldn't be merged until #2675 has been merged, since it depends on that PR's changes. Also, FrameReader still takes about 1.5 minutes per segment on device, so we should probably look into building ffmpeg with omx.


selfdrive/modeld/modeld.cc
```diff
@@ -114,11 +114,7 @@ int main(int argc, char **argv) {
   PubMaster pm({"modelV2", "model", "cameraOdometry"});
   SubMaster sm({"pathPlan", "frame"});
 
-#if defined(QCOM) || defined(QCOM2)
   cl_device_type device_type = CL_DEVICE_TYPE_DEFAULT;
```
adeebshihadeh, line 117: we can remove this now and pass it directly in `cl_get_device_id`

PR #2680, opened 2020-12-03
Update values.py

cdkes564:


cdkes564:
Added 2020 Hyundai Kona fingerprint

adeebshihadeh:
All PRs should be made against master.

Issue #2657, opened 2020-11-30
If no road camera, show camera malfunction alert instead of blank screen


VirtuallyChris:
**Describe the bug**

If a comma two has a broken road camera, an alert never shows. The user will get a blank blue screen since onroad fails to start. Would be great to show some sort of "Camera Malfunction: Contact Support" error here.

![image](https://user-images.githubusercontent.com/37757984/100668430-c0ad2c80-3310-11eb-84ac-3dc21b179ee5.png)

PR #2697, opened 2020-12-04
road camera malfunction alert

ZwX1616:
#2657


selfdrive/ui/paint.cc
```diff
@@ -540,7 +540,7 @@ void ui_draw(UIState *s) {
   }
 
   const bool draw_vision = s->started && s->active_app == cereal::UiLayoutState::App::NONE &&
-                           s->status != STATUS_OFFROAD && s->vision_connected;
+                           s->status != STATUS_OFFROAD && (s->vision_connected || s->camera_unresponsive);
```
adeebshihadeh, line 543: this shouldn't be needed

PR #2695, opened 2020-12-04
Please delete this PR

alancyau:
dongle ID: 88628232a4259218

This is for an EU 2019 Honda Civic hatchback, which was probably why the US Civic fingerprints didn't work. 

Thanks to Discord user dave237.

PR #2698, opened 2020-12-04
Add fingerprint for EU 2019 Civic Hatch

alancyau:
Please close/delete PR #2695 and use this PR instead.

dongle ID: 88628232a4259218

This is for an EU 2019 Honda Civic hatchback, which was probably why the US Civic fingerprints didn't work.

Thanks to Discord user dave237.

### DIFF
--- a/selfdrive/car/honda/values.py
+++ b/selfdrive/car/honda/values.py
@@ -447,6 +447,7 @@ FW_VERSIONS = {
       b'37805-5BB-C630\x00\x00',
       b'37805-5BB-L540\x00\x00',
       b'37805-5BB-L640\x00\x00',
+      b'37805-5AZ-G740\x00\x00',
     ],
     (Ecu.transmission, 0x18da1ef1, None): [
       b'28101-5CG-A920\x00\x00',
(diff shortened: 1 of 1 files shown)


selfdrive/car/honda/values.py
```diff
@@ -598,6 +602,7 @@ class CAR:
       b'46114-TLA-A050\x00\x00',
       b'46114-TLA-A930\x00\x00',
       b'46114-TMC-U020\x00\x00',
+      b'39494-TGL-G030\x00\x00',
```
VirtuallyChris, line 605: Revert this change. On some cars, we don't include all the possible fw queries, and you put this one is in the Civic_diesel

VirtuallyChris (changes requested):

selfdrive/car/honda/values.py
alancyau: Removed `39494-TGL-G030\x00\x00` from the `electricBrakeBooster`.

selfdrive/car/honda/values.py
```diff
@@ -597,7 +600,7 @@ class CAR:
       b'46114-TLA-A040\x00\x00',
       b'46114-TLA-A050\x00\x00',
       b'46114-TLA-A930\x00\x00',
-      b'46114-TMC-U020\x00\x00',
+      b'46114-TMC-U020\x00\x00',
```
VirtuallyChris, line 603: Just remove this extra space and we're good to go!

VirtuallyChris (changes requested):

selfdrive/car/honda/values.py
alancyau: Good catch! Extra space removed.

PR #2700, opened 2020-12-05
make toyota accelation gentler at higher speed. It has better gas mileage as well

che220:
**Description** [] Toyota's acceleration max is set at 1.5 m/s2 as a constant. This fix is to lower the accel_max as the speed increases. The car will driver smoother and gas mileage will be better

**Verification** []I have been using this fix since March. You can read the code and test it.

**Route**
Route: highway, local ...


pd0wm:
We're not merging this PR. Everyone has different preferences for acceleration profiles. Feel free to run your own fork with different tuning.

PR #2701, opened 2020-12-05
using simple array instead of static std::map

deanlee:
this will be simpler, faster and more safe than static std::map https://google.github.io/styleguide/cppguide.html#Static_and_Global_Variables


pd0wm:
I'm not sure if this PR makes the code more readable. I understand the concerns of the authors of the style guide, but I'm not sure if it's a trade off worth making here.

deanlee:
I agree

PR #2703, opened 2020-12-05
fill_lead_v2 should not be called  in inner loop

deanlee:


HaraldSchafer:
Oh that's a bug. Nice find!

PR #2704, opened 2020-12-05
refactor modeldstate

deanlee:
extract out the duplicate code into functions.make the code easy to read.


HaraldSchafer:
We're still working on tests for modeld, will merge it in after those tests work and this PR passes them

adeebshihadeh:
Continued in #2720.

PR #2705, opened 2020-12-05
Add STEER_TORQUE to car kalman filter

geohot:
Some cars have bad angle sensors. If we switch the PID loop to use this, they will be improved.

I even wonder if this will make the angle offset more accurate across the board, since it feels the torque. But I don't really know.

Need to fix the hardcoded constants and do large scale stability tests.

### DIFF
--- a/selfdrive/locationd/models/car_kf.py
+++ b/selfdrive/locationd/models/car_kf.py
@@ -32,6 +32,8 @@ class States():
   YAW_RATE = _slice(1)  # [rad/s]
   STEER_ANGLE = _slice(1)  # [rad]
 
+  STEER_TORQUE = _slice(1)   # [torques?]
+
 
 class CarKalman(KalmanFilter):
   name = 'car'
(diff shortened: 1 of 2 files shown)


geohot:
Hmm, this model isn't really right at all. It needs to be velocity dependent.

geohot:
Code to run the filter for testing:

https://gist.github.com/geohot/7c387246949e343c1ce21315de844fd2

geohot:
I think the deeper question is:

What units are STEER_TORQUE_EPS and STEER_TORQUE_CMD in (they are the same)? At low speeds, the torques are way too low to be correct.

And this would correspond with observed behavior. The same torque command at lower speeds turns the wheel more than that command at higher speeds.

geohot:
Proposing the torque unification project.

We have all these controllers that output "torque", but what does that mean. Let's unify it, and write a function that understands what the torques mean per car.

ShaneSmiskol:
After measuring steer delay from some ZSS data I'm getting 0.6 seconds as mean, even though it should be 0.12 right? I'm measuring from steer torque cmd to change in zss angle

PR #2707, opened 2020-12-06
Fix random encoder lockups

ZwX1616:
Ticis seem to be happy finally. Output files on cici also have exactly the same hash as old ones.


selfdrive/loggerd/encoder.c
```diff
@@ -610,9 +611,8 @@ void encoder_close(EncoderState *s) {
 
       OMX_BUFFERHEADERTYPE* in_buf = queue_pop(&s->free_in);
       in_buf->nFilledLen = 0;
-      in_buf->nOffset = 0;
```
geohot, line 613: Hmm, is this uninitted now?

ZwX1616: Oh it is initted when first called by encoder_encode_frame. If a segment is somehow shorter than the amount of buffers then pBuffer will be nullptr anyways. Shouldn't matter either way?

geohot: Unless you have to remove, I'd keep setting it. That timestamp thing seems very plausibly like the issue, I remember something like this when I did the qcameras.

ZwX1616: Sounds good. I will just add it back.
Added some more changes (replaced a hack with something more reasonable). It seems qcom/qcom2 don't work in the exact same way :|

PR #2709, opened 2020-12-07
GM: Fix to allow brakes to release on resume press

andrewcopenpilot:
**Description**
Fixes issue where GM vehicles can't resume from a stop by pressing the resume button.

A previous PR better matched the OEM brake mode transitions, however it created a regression.

Once mode 0xd (at full stop) has been entered, even sending a request for 0 brakes will not cause the brakes to be released.

This change updates the logic such that when a request for 0 brakes is sent, the mode is forced to mode 0x1. Which allows the brakes to release on resume press.

**Verification**
In the stopped state, verified the resume on button press functions as expected.

**Route**
Route: 160b5ba4a44512f2|2020-12-05--17-48-53--7


andrewcopenpilot:
ErikN confirms via gm discord this fix works as intended.

PR #2710, opened 2020-12-07
UI style improvements with improved keyboard

ghost:


selfdrive/ui/qt/widgets/keyboard.cc
```diff
@@ -7,31 +7,31 @@
...
-    for(QString p : s){
+    for (QString p : s){
       QPushButton* btn = new QPushButton(p);
       btn->setFixedHeight(120);
-
+      btn_group->addButton(btn);
+      hlayout->addSpacing(10);
       if (p == QString("  ")){
-        btn->setFixedWidth(1024);
+        hlayout->addWidget(btn, 3);
```
grekiki, line 30: Do we want a comment for what the 3 and 1 do?

pd0wm: It's fine.

selfdrive/ui/qt/widgets/keyboard.cc
```diff
@@ -7,31 +7,31 @@
...
+      }else{
```
pd0wm, line 31: spaces

selfdrive/ui/qt/widgets/keyboard.cc
adeebshihadeh: don’t think that’s necessary, but maybe move them to constants on top for the different stretch values

selfdrive/ui/qt/widgets/keyboard.cc
```diff
@@ -7,35 +7,37 @@
 
 #include "keyboard.hpp"
 
+const int DEFAULT_WIDTH = 1;
+const int SPACEBAR_WIDTH = 3;
```
adeebshihadeh, line 11: stretch isn't really width. why did we offset the default by 1? shouldn't 0 work the same?

grekiki: Oh I thought it was "weight". I'll check 0

grekiki: Hm 0 doesn't really work(https://imgur.com/a/TPJrTtZ). I call it width since we are in a HBoxLayout.

PR #2711, opened 2020-12-07
Carla runner

ghost:
Let's make openpilot work nicely with carla.


grekiki:
I'll close for now, will reopen if more time comes up.

Issue #2473, opened 2020-11-02
Toyota TSS2: Double Brake when Approaching Stopped Vehicle


C0mpl3t3N00b:
**Describe the bug**

With OpenPilot enabled, when approaching a stopped car at a traffic light or heavy traffic situation, OpenPilot will slow to a stop behind the stopped car, but will then release the brakes and move forward slightly and stop again. This is with the lead car not moving at all.

**How to reproduce or log data**

Approaching a stopped vehicle will exhibit this behavior about 50% of the time.

**Expected behavior**

When approaching a stopped car, OpenPilot should slow to a stop at an appropriate distance from the lead car and only move forward if the lead car moves forward.

**Device/Version information (please complete the following information):**
 - Device: Comma Two
 - Dongle ID: bdad50b8a1f9fa8e|2020-11-01--16-47-58--5
 - Route: bdad50b8a1f9fa8e|2020-11-01--16-47-58--5
 - Timestamp: https://my.comma.ai/cabana/?route=bdad50b8a1f9fa8e%7C2020-11-01--16-47-58&exp=1635827347&sig=huumgNlfxiHDVminmiuY9uGtn%2FL3BO%2FrcSWvX5BBpg4%3D&max=18&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2Fbdad50b8a1f9fa8e%2F6cd92beecc597e5fb5f53b068f0c3f7a_2020-11-01--16-47-58
 - Version: master-ci
 - Car make/model: 2020 Rav4 Hybrid LE

**Additional context**

cydia2020:
Same

kozuch:
Can this be because humans sometimes do this to and the model learned it from data?

grekiki:
No, longitudinal policy is not learnt.

zorrobyte:
This is old news, has been an issue since at least 0.5. This is why they are working on E2E long control.

timgudex:
I get this as well but it looks like it will touch the vehicle rather then stopping again.

pd0wm:
Your case seems related to https://github.com/commaai/openpilot/issues/1270. 

Does this PR fix it? Also let me know if stops are too rough now, then we need to come up with a different solution. https://github.com/commaai/openpilot/pull/2508

yufeng66:
[...]

PR #2716, opened 2020-12-08
Fix TSS2 creep issue, no need to send standstill request

briskspirit:
TSS2 vehicles don't need for standstill request to be sent. If we send it PCM will go from status 8 to status 11 upon full stop(that where issue is from as we remove standstill request as soon as status != 8). Status 11 is a 3 seconds timer after which PCM goes into state 7 - standstill and asks to press resume or accelerator. After PCM is entered state 7 - standstill_req = False will not switch it back to cruise (8).

Was tested on Corolla TSS2 and Corolla Hybrid TSS2

Associated discussions: #2508 , #1270 , #2473

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -84,7 +84,7 @@ class CarController():
       pcm_cancel_cmd = 1
 
     # on entering standstill, send standstill request
-    if CS.out.standstill and not self.last_standstill:
+    if CS.out.standstill and not self.last_standstill and CS.CP.carFingerprint not in TSS2_CAR:
       self.standstill_req = True
     if CS.pcm_acc_status != 8:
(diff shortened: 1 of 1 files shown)


briskspirit:
Route : f66032c2b5aa18ac 2020-12-08--08-35-47
Car: 2021 Corolla TSS2
BRAKING_STOP_RATE = 0.1 (That caused most notable creep issue)

no_need_for_standstill

briskspirit:
Changed set to look in from TSS2_CAR to NO_STOP_TIMER_CAR as it is more appropriate for that fix and related to timer

briskspirit:
Will also add that this fix as of now was tested on: Corolla, CorollaH, Rav4H, HighlanderH.

adeebshihadeh:
LGTM, tested on our corolla.

PR #2717, opened 2020-12-08
Dynamically colored alert widget. 

ghost:
Make the alerts widget blue, if there is an update available.

### DIFF
--- a/selfdrive/ui/qt/home.cc
+++ b/selfdrive/ui/qt/home.cc
@@ -99,16 +99,31 @@ void OffroadHome::refresh() {
   }
 
   alert_notification->setVisible(true);
-  alert_notification->setStyleSheet(QString(R"(
-    padding: 15px;
-    padding-left: 30px;
-    padding-right: 30px;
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/home.cc
```diff
@@ -99,16 +99,31 @@ void OffroadHome::refresh() {
...
-    border-radius: 5px;
-    font-size: 40px;
-    font-weight: bold;
-    background-color: red;
-  )"));
+
+  // Red background for alerts, blue for update available
+  if (alerts_widget->updateAvailable){
+    alert_notification->setStyleSheet(QString(R"(
+      padding: 15px;
```
adeebshihadeh, line 106: most of these lines are duplicate

grekiki: That's true, I'll check if there is a way to keep most of it the same.

grekiki: Solved

PR #2720, opened 2020-12-08
Test model refactor

haraschax:

### DIFF
--- a/selfdrive/common/modeldata.h
+++ b/selfdrive/common/modeldata.h
@@ -1,10 +1,10 @@
 #pragma once
 
-#define MODEL_PATH_DISTANCE 192
-#define TRAJECTORY_SIZE 33
-#define MIN_DRAW_DISTANCE 10.0
-#define MAX_DRAW_DISTANCE 100.0
-#define POLYFIT_DEGREE 4
-#define SPEED_PERCENTILES 10
-#define DESIRE_PRED_SIZE 32
-#define OTHER_META_SIZE 4
+constexpr int MODEL_PATH_DISTANCE = 192;
+constexpr int  TRAJECTORY_SIZE = 33;
+constexpr float MIN_DRAW_DISTANCE = 10.0;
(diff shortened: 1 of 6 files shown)


selfdrive/modeld/models/driving.cc
```diff
@@ -8,25 +8,42 @@
...
+
+constexpr int LEAD_MHP_N = 5;
+constexpr int LEAD_MHP_VALS = 4;
+constexpr int LEAD_MHP_SELECTION = 3;
+constexpr int LEAD_MHP_GROUP_SIZE = (2*LEAD_MHP_VALS + LEAD_MHP_SELECTION);
+
+constexpr int POSE_SIZE = 12;
+
+constexpr int MIN_VALID_LEN = 10.0;
+constexpr int TRAJECTORY_TIME = 10.0;
```
adeebshihadeh, line 31:
```suggestion
constexpr int MIN_VALID_LEN = 10;
constexpr int TRAJECTORY_TIME = 10;
```

HaraldSchafer:
modelV2 packet needs to be tested before this can be merged

selfdrive/common/modeldata.h
```diff
@@ -1,10 +1,10 @@
...
 
-#define MODEL_PATH_DISTANCE 192
-#define TRAJECTORY_SIZE 33
-#define MIN_DRAW_DISTANCE 10.0
-#define MAX_DRAW_DISTANCE 100.0
-#define POLYFIT_DEGREE 4
-#define SPEED_PERCENTILES 10
-#define DESIRE_PRED_SIZE 32
-#define OTHER_META_SIZE 4
+constexpr int MODEL_PATH_DISTANCE = 192;
```
adeebshihadeh, line 3: const should be the same here and it's shorter

deanlee: no special reason, just a quick copy and paste

PR #2721, opened 2020-12-08
Car Port: 2021 Toyota Camry (TSS 2.5)

adeebshihadeh:


briskspirit:
Could you consider making separate lateral tune section in interface.py for TSS2 Camry? I assume the difference will be as huge as between TSS1 Corolla and TSS2 Corolla. My 2 cents, thank you.

adeebshihadeh:
This is just a POC first port for TSS2.5. Since we don't have these cars, it's difficult to tune them. We'd certainly merged improved tuning, including separating the TSS1 and 2 models, if camry users opened a PR for it.

PR #2727, opened 2020-12-09
hardware.py: get network info over dbus

pd0wm:


common/hardware_tici.py
```diff
@@ -1,62 +1,134 @@
...
+    primary_connection = self.nm.Get(NM, 'PrimaryConnection', dbus_interface=DBUS_PROPS)
+    primary_connection = self.bus.get_object(NM, primary_connection)
+    tp = primary_connection.Get(NM_CON_ACT, 'Type', dbus_interface=DBUS_PROPS)
+
+    if tp in ['802-3-ethernet', '802-11-wireless']:
+      return NetworkType.wifi
+    elif tp in ['gsm']:
+      modem = self.get_modem()
+      access_t = modem.Get(MM_MODEM, 'AccessTechnologies', dbus_interface=DBUS_PROPS)
+      if access_t >= MM_MODEM_ACCESS_TECHNOLOGY_LTE:
```
aniva, line 53: I am testing with LTE SIM (verified) and it registers with 3G only (UMTS), I see this recent commit and I belive it does not detect modem properly as LTE modem, so that it is effectively 3G only

PR #2730, opened 2020-12-10
Retune MPC

haraschax:
Laneless branch with MPC work WIP. Currently stil cuts turns in simulation and the real world.


HaraldSchafer:
First experience report on corolla
- Didn't try to kill me
- Still cuts turns like a race car driver
- Not rock solid on straight highway

Issue #2726, opened 2020-12-09
adaptive cruise will be forcefully disengaged after 3 minutes of standstill on Toyota Hybrids


cydia2020:
**Describe the bug**
With openpilot's current standstill, ACC will be forcefully disengaged by the car after exactly 3 minutes if the front car doesn't move. In this situation, the car will be shifted to park. This is a potential problem for comma pedal users as comma pedal users tend to not move the cruise control lever when the car is stopped, if the front car doesn't move due to heavy traffic, the car will be shifted to P

**How to reproduce or log data**
Stop behind a completely stationary car in a Toyota hybrid vehicle, after 3 minutes with no movements, ACC will be forcefully disengaged and the car shifted to P. 

**Expected behavior**
Enter the "Waiting" mode (seems to be related to `CRUISE_CONTROL_STATE` 11 in `PCM_CRUISE_SM`) instead of the current  "operate cruise lever or accelerator pedal to resume" mode.

**Device/Version information (please complete the following information):**
 - Device: comma two
 - Dongle ID: 399b1cca765e80b4
 - Route: 399b1cca765e80b4|2020-12-09--11-54-59
 - Timestamp: the final disengagement in this route
 - Version: 0.8.1
 - Car make/model: Toyota Prius 2018

**Additional context**
With openpilot's standstill logic, the car's dash will immediately display a "operate cruise lever or accelerator pedal to resume" message upon entering standstill, while the stock system will display the "waiting" message and only display "operate cruise lever or accelerator pedal to resume" after the front car has moved, which doesn't seem to trigger this lock-out.

cydia2020:
> TIL that if you sit with cruise control engaged in stand still traffic, eventually a 2017 Prius Prime will put itself in park. This causes OP to freak out and alarm.

https://discord.com/channels/469524606043160576/524327905937850394/761394246493929482
This has been documented before on the discord server.

briskspirit:
Waiting mode is 7, mode 11 is a 3 seconds timer after which PCM enters mode 7. 
Same will happen if you will wait too long in the mode 7 (Tested on Corolla Hybrid 2020).
What I've discovered for TSS2 models: OP don't need to send standstill request. Could be true for all Toyotas, need to check.
If you would like to participate in the tests - DM me on the discord.

cydia2020:
> Waiting mode is 7, mode 11 is a 3 seconds timer after which PCM enters mode 7.
[...]

PR #2732, opened 2020-12-10
alert user when car is about to disengage adaptive cruise

cydia2020:
Must be used together with https://github.com/commaai/cereal/pull/105

Minimise the effect of #2726

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -340,6 +340,10 @@ class CarInterface(CarInterfaceBase):
         # while in standstill, send a user alert
         events.add(EventName.manualRestart)
 
+    if ret.vEgo < 0.001 and self.CS.pcm_acc_status == 7:
+      # alert user when the car is about to disengage adaptive cruise
+      events.add(EventName.standstillTimer)
+
     ret.events = events.to_msg()
 
     self.CS.out = ret.as_reader()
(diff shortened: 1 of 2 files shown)


grekiki:
Nothing wrong with the PR idea, but please mark as ready for review at least after passing the CI tests.

cydia2020:
> Nothing wrong with the PR idea, but please mark as ready for review at least after passing the CI tests.

Hi, I am not sure but I think it failed because of commaai/cereal#105 . Is there a way to incorporate that PR when doing the checks?

grekiki:
If you cd to openpilot/cereal you will get into another repository, where you can then checkout to your PR. Then if you go back to openpilot and run git status, you should see that the submodule was updated. Then you can commit and push, and it should hopefully pass the tests.

cydia2020:
> If you cd to openpilot/cereal you will get into another repository, where you can then checkout to your PR. Then if you go back to openpilot and run git status, you should see that the submodule was updated. Then you can commit and push, and it should hopefully pass the tests.

Ah, thanks

selfdrive/car/toyota/interface.py
```diff
@@ -340,6 +340,10 @@ def update(self, c, can_strings):
         # while in standstill, send a user alert
         events.add(EventName.manualRestart)
 
+    if ret.vEgo < 0.001 and self.CS.pcm_acc_status == 7:
```
adeebshihadeh, line 343: doesn't the car show an alert? I don't like this alert creation delay hack

cydia2020: No, the car only alerts when it reaches 180 seconds (and shift to p), in which case openpilot will panic as well for wrong gear.

cydia2020:
closing until i find a less janky way of counting this.

PR #2734, opened 2020-12-10
Carla runner on docker

marschr:
This should enable carla to run on docker along with openpilot with onnxruntime-gpu on a nvidia GPU.
Users should build the container themselves til a proper way to ship cudnn libraries is found.
There's an updated the readme file with instructions on how to build the container.


marschr:
Closing as it doesn't seem like the best solution. Needs to be revisited.

PR #2742, opened 2020-12-11
Nissan altima

avolmensky:
PR for Nissan Altima.

Will get the routes submitted soon

**Checklist**
- [X] added to README
- [X] test route added to [test_car_models](../../selfdrive/test/test_car_models.py)
- [X] route with openpilot: `b72d3ec617c0a90f|2020-12-11--15-38-17`
- [X] route with stock system: `b72d3ec617c0a90f|2020-12-13--11-05-33`


pd0wm:
Merged panda PR and updated submodule on master

selfdrive/car/nissan/carstate.py
```diff
@@ -82,15 +97,23 @@ def update(self, cp, cp_adas, cp_cam):
...
 
-    self.lkas_hud_msg = copy.copy(cp_adas.vl["PROPILOT_HUD"])
-    self.lkas_hud_info_msg = copy.copy(cp_adas.vl["PROPILOT_HUD_INFO_MSG"])
+    if self.CP.carFingerprint == CAR.ALTIMA:
+      # Altima does not support LKAS HUD
+      self.lkas_hud_msg = None
+      self.lkas_hud_info_msg = None
+    else:
+      self.lkas_hud_msg = copy.copy(cp_adas.vl["PROPILOT_HUD"])
+      self.lkas_hud_info_msg = copy.copy(cp_adas.vl["PROPILOT_HUD_INFO_MSG"])
```
adeebshihadeh, line 116:
```suggestion
    if self.CP.carFingerprint != CAR.ALTIMA:
      self.lkas_hud_msg = copy.copy(cp_adas.vl["PROPILOT_HUD"])
      self.lkas_hud_info_msg = copy.copy(cp_adas.vl["PROPILOT_HUD_INFO_MSG"])
```

adeebshihadeh: and set both to None on CarState init

selfdrive/car/nissan/carstate.py
```diff
@@ -179,108 +199,150 @@ def get_can_parser(CP):
...
+        ("SET_0x80_2", "LKAS", 0),
+        ("MAX_TORQUE", "LKAS", 0),
+        ("SET_0x80", "LKAS", 0),
+        ("COUNTER", "LKAS", 0),
+        ("LKA_ACTIVE", "LKAS", 0),
+        ("LKAS_ENABLED", "LKAS_SETTINGS", 0),
+
+        ("CRUISE_ON", "PRO_PILOT", 0),
+      ]
+      checks = []
```
adeebshihadeh, line 238: add all the msgs you're reading as checks

selfdrive/car/nissan/carstate.py
```diff
@@ -179,108 +199,150 @@ def get_can_parser(CP):
...
-    checks = [
-    ]
+    elif CP.carFingerprint == CAR.ALTIMA:
+      signals += [
+        ("STEER_TORQUE_DRIVER", "STEER_TORQUE_SENSOR", 0),
+      ]
+      checks += [
+        ("STEER_TORQUE_SENSOR", 100),
+      ]
+      return CANParser(DBC[CP.carFingerprint]['pt'], signals, checks, 0)
```
adeebshihadeh, line 345: just use the return below and add logic to use the right bus

adeebshihadeh:
The port looks good, except we need to switch the powertrain and camera can buses. We've like to maintain the existing bus assignments, instead of swapping the PT and cam buses on the altima.

avolmensky:
@adeebshihadeh as discussed. I don't think maintaining bus names in this case makes sense. We are physically swapping the bus on the harness for the Altima. We'll also need to duplicate a lot of code to keep the bus assignments as they are on the rogue/leaf.

adeebshihadeh:
Thanks for the port!

Issue #2739, opened 2020-12-10
camerad memory leak


ZwX1616:
**Describe the bug**

camerad uses 2-3% more memory per day if run for days

**How to reproduce or log data**

run camerad for extended amount of time

**Expected behavior**

constant memory usage

**Device/Version information (please complete the following information):**
 - Device: EON/comma two
 - <s>Dongle ID:</s>
 - <s>Route:</s>
 - <s>Timestamp:</s>
 - Version: current master
 - <s>Car make/model:</s>

**Additional context**

ZwX1616:
this might have been in camerad for ever. are you interested in looking into it? @deanlee

deanlee:
I would love to try

pd0wm:
Merged the PR. @ZwX1616 Can you check if the leak is gone?

ZwX1616:
> Merged the PR. @ZwX1616 Can you check if the leak is gone?

yeah the testing closet devices have been updated. we should be able to see the difference easily

adeebshihadeh:
Seeing no leak after two days, thanks @deanlee :)

deanlee:
I'm glad the problem was finally solved

PR #2743, opened 2020-12-11
camerad: fix memory leak in create_thumbnail

deanlee:
Related issue: https://github.com/commaai/openpilot/issues/2739

I'm not sure this issue can be completely resolved after this fix.need to continue to look for possible memory leaks elsewhere

### DIFF
--- a/selfdrive/camerad/cameras/camera_common.cc
+++ b/selfdrive/camerad/cameras/camera_common.cc
@@ -294,15 +294,17 @@ void create_thumbnail(MultiCameraState *s, CameraState *c, uint8_t *bgr_ptr) {
     row_pointer[0] = row;
     jpeg_write_scanlines(&cinfo, row_pointer, 1);
   }
-  free(row);
   jpeg_finish_compress(&cinfo);
   jpeg_destroy_compress(&cinfo);
+  free(row);
 
   MessageBuilder msg;
   auto thumbnaild = msg.initEvent().initThumbnail();
   thumbnaild.setFrameId(b->cur_frame_data.frame_id);
   thumbnaild.setTimestampEof(b->cur_frame_data.timestamp_eof);
   thumbnaild.setThumbnail(kj::arrayPtr((const uint8_t*)thumbnail_buffer, thumbnail_len));
+
+  free(thumbnail_buffer);
 
   if (s->pm != NULL) {
     s->pm->send("thumbnail", msg);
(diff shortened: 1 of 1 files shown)


deanlee:
the memory returned by jpeg_mem_dest should be freed too

selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -294,15 +294,18 @@ void create_thumbnail(MultiCameraState *s, CameraState *c, uint8_t *bgr_ptr) {
...
+  jpeg_destroy_compress(&cinfo);
+  free(row);
 
   MessageBuilder msg;
   auto thumbnaild = msg.initEvent().initThumbnail();
   thumbnaild.setFrameId(b->cur_frame_data.frame_id);
   thumbnaild.setTimestampEof(b->cur_frame_data.timestamp_eof);
   thumbnaild.setThumbnail(kj::arrayPtr((const uint8_t*)thumbnail_buffer, thumbnail_len));
 
+  free(thumbnail_buffer);
```
pd0wm, line 307: Can we free before sending the message? Does the `setThumbnail` call make a copy or just reference the `thumbnail_buffer`?

deanlee: my bad

deanlee:
my mind went blank.I should free thumbnail_buffer after send.

pd0wm:
No problem. Thanks for finding the leak!

I'll add the jpeg processing to `camera_frame_stream` so we can catch this on the computer as well. Also confirmed the existence of this (obvious) leak using valgrind just as a sanity check and made sure it was gone after your changes.

```
==1759538== 73,884 (1,176 direct, 72,708 indirect) bytes in 7 blocks are definitely lost in loss record 189 of 192
==1759538==    at 0x483B7F3: malloc (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==1759538==    by 0x4A11D1B: jinit_memory_mgr (in /usr/lib/x86_64-linux-gnu/libjpeg.so.8.2.2)
==1759538==    by 0x49E56AC: jpeg_CreateCompress (in /usr/lib/x86_64-linux-gnu/libjpeg.so.8.2.2)
==1759538==    by 0x407F30: create_thumbnail(MultiCameraState*, CameraState*, unsigned char*) (camera_common.cc:262)
==1759538==    by 0x408730: processing_thread(MultiCameraState*, char const*, CameraState*, int, void (*)(MultiCameraState*, CameraState*, int)) (camera_common.cc:360)
==1759538==    by 0x4F9FD83: ??? (in /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28)
==1759538==    by 0x49C7608: start_thread (pthread_create.c:477)
==1759538==    by 0x51E9292: clone (clone.S:95)
```

PR #2748, opened 2020-12-11
oneplus deprecation

adeebshihadeh:


selfdrive/car/car_helpers.py
```diff
@@ -23,6 +24,8 @@ def get_startup_event(car_recognized, controller_available):
     event = EventName.startupNoCar
   elif car_recognized and not controller_available:
     event = EventName.startupNoControl
+  elif ANDROID and "letv" not in open("/proc/cmdline").read():
```
geohot, line 27: why can't this use not LEON?

adeebshihadeh: leon is defined a line above the if in updated, not in the hw abstraction layer. we could define in the hw abstraction layer, but would prefer to keep these temp changes as local as possible

selfdrive/controls/lib/events.py
```diff
@@ -255,6 +255,14 @@ def wrong_car_mode_alert(CP: car.CarParams, sm: messaging.SubMaster, metric: boo
       Priority.LOWER, VisualAlert.none, AudibleAlert.none, 0., 0., 15.),
   },
 
+  EventName.startupOneplus: {
+    ET.PERMANENT: Alert(
+      "WARNING: Original EON deprecated",
+      "Upgrade to comma two",
```
geohot, line 261: Maybe instead of "Update to comma two" say

"Device will no longer update"

PR #2751, opened 2020-12-12
cleanup camera malfunction alert

deanlee:


selfdrive/ui/ui.cc
```diff
@@ -283,17 +283,16 @@ void ui_update(UIState *s) {
...
+    const uint64_t frame_delayed = (s->sm)->frame - frame_pkt;
+    if ((frame_pkt > s->started_frame && frame_delayed > 5*UI_FREQ) ||
+        frame_delayed > 35*UI_FREQ) {
       // controls is fine, but rear camera is lagging or died
       s->scene.alert_text1 = "Camera Malfunction";
       s->scene.alert_text2 = "Contact Support";
       s->scene.alert_size = cereal::ControlsState::AlertSize::FULL;
       s->status = STATUS_DISENGAGED;
       s->sound->stop();
-      ui_draw_vision_alert(s, s->scene.alert_size, s->status,
```
deanlee, line 295: we don't need to call ui_draw_vision_alert here

adeebshihadeh: it's gated on vision_connected. https://github.com/commaai/openpilot/blob/master/selfdrive/ui/paint.cc#L541

deanlee: oic

deanlee:
Move ui_draw_vision_alert() out of  ui_draw_vision(),put it before nvgEndFrame(), ensure that the alert will always be displayed.

selfdrive/ui/ui.cc
deanlee: We should not call ui_draw_vision_alert outside of beginframe/endframe.

ZwX1616:
looks fine to me. works also.
waiting for approval from @adeebshihadeh

PR #2758, opened 2020-12-12
MPC rework

haraschax:
Do not penalize turning steering wheel if model says you should.


HaraldSchafer:
Continue from https://github.com/commaai/openpilot/pull/2730

SConstruct
```diff
@@ -136,9 +136,8 @@ env = Environment(
     "-fPIC",
     "-O2",
     "-Wunused",
-    "-Werror",
     "-Wno-unknown-warning-option",
-    "-Wno-deprecated-register",
+    "-Wno-deprecated",
```
adeebshihadeh, line 140: let's fix this in the MPC SConscript, so we can rebuild MPC without these hacks in the main SConsctruct

HaraldSchafer: sounds good

PR #2759, opened 2020-12-12
Discuss: don't dispatch tbuffer in CameraBuf::accquire

deanlee:
I don’t think we can call tbuffer_dispatch in CameraBuf::acquire. we still use this buffer(cur_rgb_buffer) at somewhere else such as  camera_process_frame(),dispatch it earyly may cause problems(Maybe the UI will also hang).

@ZwX1616  am I right?

related PR: https://github.com/commaai/openpilot/pull/2613


adeebshihadeh:
This is going away in #2668.

PR #2761, opened 2020-12-12
camerad: add class JpegThumbnail

deanlee:
it seems that create_thumbnail will cause ui/model laggy. mv it into class , cache resources used by libjpeg


deanlee:
It may not be worth.the thumb is only created every 100 frames.

PR #2763, opened 2020-12-12
Add Forester 2020 fingerprint

jcak:
Added v1 fingerprint for 2020 Subaru Forester.
Same "Subaru global platform" as the 2019 Forester, so no need for new vehicle


selfdrive/car/subaru/values.py
```diff
@@ -7,7 +7,7 @@
 class CAR:
   ASCENT = "SUBARU ASCENT LIMITED 2019"
   IMPREZA = "SUBARU IMPREZA LIMITED 2019"
-  FORESTER = "SUBARU FORESTER 2019"
+  FORESTER = "SUBARU FORESTER 2019 - 2020"
```
pd0wm, line 10: Don't change vehicle name

```suggestion
  FORESTER = "SUBARU FORESTER 2019"
```

pd0wm:
What is your dongle id?

No need to update the year in the vehicle name since the car behaves the same. Can you add the year in README.md?

pd0wm:
The fingerprint you added also overlaps with the existing one. Can you just combine them? (e.g. make a new fingerprint which is a superset of both fingerprints).

jcak:
made the change as requested, my dongle ID is 4ee121621f388597

jcak:
Merged in both the 2019 forester touring fingerprint messages from #19499 and merged the existing 2019 forester sport messages, all the same fingerprint.

pd0wm:
Looks good. For later reference `961: 8` and `984: 8` were added.

PR #2764, opened 2020-12-12
camera_qcom: read SensorEvents in op thread 

deanlee:
Since the sensorsEvents is only used by do_autofocus(), it is not efficient to read it in the processing_thread. putting it in do_autofocus will make the code clearer and simpler


deanlee:
do_autofocus() runs every 50ms,so  the last_sag_ts is no longer needed to control the reading frequency

ZwX1616:
> do_autofocus() runs every 50ms,so the last_sag_ts is no longer needed to control the reading frequency

it's actually 150ms. but you are right about the ts, and the code now looks way more straightforward!

PR #19494, opened 2020-12-14
Move MCLK freq on tici

robbederks:
GPS really dislikes the 24MHz clock. This moves it out of the band.

### DIFF
--- a/selfdrive/camerad/cameras/camera_qcom2.cc
+++ b/selfdrive/camerad/cameras/camera_qcom2.cc
@@ -287,7 +287,7 @@ void sensors_init(int video0_fd, int sensor_fd, int camera_num) {
   power->count = 1;
   power->cmd_type = CAMERA_SENSOR_CMD_TYPE_PWR_UP;
   power->power_settings[0].power_seq_type = 0;
-  power->power_settings[0].config_val_low = 28000000; //Hz
+  power->power_settings[0].config_val_low = 24000000; //Hz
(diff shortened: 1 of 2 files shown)


geohot:
Do we also have to adjust the PLL in the sensor? I think it's overclocked now.

robbederks:
> Do we also have to adjust the PLL in the sensor? I think it's overclocked now.

Was just looking into this, yes, we'll have to change the config a bit for the predivision/multiplication to keep pixclk the same

robbederks:
Ok, that PLL just brought it back. Makes sense, since with this div and multiplication fVCO is 528MHz, exactly 1/3 of the noise I'm seeing.
Looking to see if it's ok to change fVCO to another value which doesn't have a direct harmonic at GPS freqs

robbederks:
Turns out it's mainly noise from the serial clock, so not really dependent on the extclk.
Running with these params now:
```
Ext clk: 24.0MHz
fVCO: 480.0MHz
Pix clk: 80.0MHz
Clk OP: 48.0MHz
fSerial: 480.0MHz
Freq around GNSS: 1440000000.0MHz
Harmonic at: 960.0MHz
Harmonic at: 1440.0MHz
Harmonic at: 1920.0MHz
Harmonic at: 2400.0MHz
```
Old params:
```
Ext clk: 24.0MHz
fVCO: 528.0MHz
Pix clk: 88.0MHz
Clk OP: 52.8MHz
fSerial: 528.0MHz
Freq around GNSS: 1584000000.0MHz
Harmonic at: 1056.0MHz
Harmonic at: 1584.0MHz
```

geohot:
What's the serial clock?

robbederks:
The CSI clock

PR #19520, opened 2020-12-16
Toyota: don't cut steering for 2 seconds after fault recovery

sshane:
When a fault occurs, LKA_STATE is 25 on rising edge, then 9 for ~200 frames. Since the code in carcontroller kept resetting the `last_fault_frame` variable, it only started the 2 second timer once the fault was over and steering was possible. Now instead it will start steering immediately after state returns to normal, so 4 seconds to 2 seconds of loss of control after a fault.

You can see here that after the first 2 seconds, the state is 1 (not applying torque but ready) for another 2 seconds when we could be steering again. I'm wondering if this breaks anything for other Toyotas, though It worked perfectly for my 17 Corolla.

Screenshot_137

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -38,7 +38,6 @@ class CarController():
     self.last_standstill = False
     self.standstill_req = False
 
-    self.last_fault_frame = -200
     self.steer_rate_limited = False
 
     self.fake_ecus = set()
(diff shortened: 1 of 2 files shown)


pd0wm:
We added this somewhere in 2018 when we thought it would help recover faster from the fault after the rate is too high, but don't think we ever had any concrete proof that it recovered any faster.

So probably fine to merge this. But we might want to get some metrics first to track these steering faults and their recovery.

ShaneSmiskol:
Yeah from my poking around in Cabana, every single fault I looked at was in a fault state for 2 seconds, then was reporting it was okay to send torque again. Never saw it fault for more than 2 seconds concurrently.

And I just went on a drive with this, after it faults it waits for the fault to clear then steers immediately after it was okay to do so. Do you want the route? I sort of went crazy, triggered lots of faults to test it.

pd0wm:
Yeah, would be great if you can post a cabana screenshot from that second drive.

ShaneSmiskol:
Here's three, then a close up of the last.

https://imgur.com/a/X54Cz5K

pd0wm:
Yeah looks good!

ShaneSmiskol:
It most often faults for me when it's near crossing zero degrees with high torque when the wheel is moving relatively quickly, so like when it goes from a sharp angle down to 0 and then in the opposite direction. It feels like a strange safety check to me, so that the wheel doesn't travel too quickly from one direction across center to the opposite direction.

It can handle giving torque at high angle rates, I've seen it go to 400 deg/s, then once it drops down to 300 deg/s it faults; so it's not purely angle rate.

adeebshihadeh:
LGTM

PR #19521, opened 2020-12-16
Refactor modeld to keep input in CL

geohot:
Still WIP. Hopefully all model_eval_frame can end up in thneed

### DIFF
--- a/selfdrive/modeld/models/commonmodel.cc
+++ b/selfdrive/modeld/models/commonmodel.cc
@@ -5,51 +5,6 @@
 #include "common/mat.h"
 #include "common/timing.h"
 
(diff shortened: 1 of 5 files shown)


pd0wm:
The memcpy (https://github.com/commaai/openpilot/blob/master/selfdrive/modeld/modeld.cc#L185) is removed in the VisionIPC 2.0 PR. So no need to look into that part of the processing.

geohot:
I'm working after that. My goal is to get the transforms into thneed, remove the clEnqueueMap, and remove these memmoves. CL the whole way.

https://github.com/commaai/openpilot/blob/master/selfdrive/modeld/models/driving.cc#L109

PR #19523, opened 2020-12-16
2020 RDX fw [56a8b30d8b604750]

VirtuallyChris:


pd0wm:
@VirtuallyChris You can use `./test_fw_query_on_routes.py --car "ACURA RDX 2020 TECH" "56a8b30d8b604750|2020-12-16--12-54-49"` (in selfdrive/debug) to check if it fingerprints correctly. In this case it works fine!

PR #19526, opened 2020-12-16
Only register when needed

pd0wm:

### DIFF
--- a/selfdrive/manager.py
+++ b/selfdrive/manager.py
@@ -398,7 +398,7 @@ def manager_init():
     print("WARNING: failed to make /dev/shm")
 
   # set dongle id
-  reg_res = register()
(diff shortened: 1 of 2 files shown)


selfdrive/manager.py
```diff
@@ -406,7 +406,7 @@ def send_managed_process_signal(name, sig):
 
 def manager_init(should_register=True):
   if should_register:
-    reg_res = register()
+    reg_res = register(spinner)
```
adeebshihadeh, line 409: can't we just set the spinner here instead of passing it in? should be fast enough to not flicker if it doesn't need to register

pd0wm: You definitely see it. At least on tici

PR #19549, opened 2020-12-20
UI:  fix buggy poll

deanlee:


adeebshihadeh:
What issue does this fix?

deanlee:
safe_poll is used to skip read visionstream if there is no data during the timeout period to avoid ui laggy or freezing.

deanlee:
possible fix for: #19579

dingyf12593:
I tried this plan tonight, but the problem still exists.

deanlee:
@dingyf12593 Has the problem improved or gotten worse?

PR #19554, opened 2020-12-20
camerad: refactor exposure

deanlee:


selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -1203,10 +1158,8 @@ void cameras_run(MultiCameraState *s) {
 
   LOG(" ************** STOPPING **************");
 
-  err = pthread_join(ae_thread_handle, NULL);
-  assert(err == 0);
+  for (auto &t : threads)
```
deanlee, line 1161: wait all threads exit before call cameras_close

selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -1203,10 +1158,8 @@ void cameras_run(MultiCameraState *s) {
...
+    t.join();
 
   cameras_close(s);
```
deanlee, line 1164: wait all threads exit before call cameras_close

deanlee: submitted another PR to fix this problem https://github.com/commaai/openpilot/pull/19555

selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -361,6 +364,19 @@ void *processing_thread(MultiCameraState *cameras, const char *tname,
 
     callback(cameras, cs, cnt);
 
+#if defined(QCOM) || defined(QCOM2)
```
adeebshihadeh, line 367: if we start putting ifdefs like this everywhere, camera_common will just become what main.cc used to be

deanlee: yes you're right,I'll reconsider it.

deanlee: yes you're right,I'll reconsider it.

deanlee:
use condition_variable to trigger camera_settings thread.it's more efficient and real-time as compared to sleep 50ms

PR #19555, opened 2020-12-20
FIX: It is not safe to call cameras_close before all threads are done.

deanlee:
call cameras_close after Thread().join()


deanlee:
should call CameraBuf::stop the wakeup CameraBuf::acquire

deanlee:
related issue : https://github.com/commaai/openpilot/issues/19724

PR #19560, opened 2020-12-20
Refactor camera_qcom: use different classes for rear and front cameras

deanlee:


deanlee:
ready for review.we can do further refactoring base on this PR.

PR #19562, opened 2020-12-21
Added pyenv-virtualenv to ubuntu install script

michaelhonan:
**Description**

~~After a fresh install on Ubuntu 20.04. `pyenv-virtualenv` was not installed, and caused an error when running `openpilot_env.sh`. Updated the install script to include `pyenv-virtualenv`.~~

This was in error. Openpilot was not cloned to `~/openpilot` and so that caused installation problems. Probably will work out a way to improve the script to prevent this in future...

**Verification**

~~I ran the exact command I added into the install script, and was successfully able to run `openpilot_env.sh`.~~


michaelhonan:
Is `scons` not installed by this script as well?

PR #19563, opened 2020-12-21
Add speed limit indicator

mayfieldiv:
Made a new service `speedlimitd` that uses gps long/lat from `gpsLocationExternal`, looks up the recent road segments in Mapbox's map matching API, and publishes `liveMapData`, which is previously unused as far as I can tell.

Here's a showcase of it from a drive earlier this evening: https://youtube.com/watch?v=GkIRIIrR5ik

The biggest downside of the feature is that right now, it needs a Mapbox access token to work (located at a new path `/persist/mapbox/access_token`). So either each person wanting to use the feature needs to get their own Mapbox account and access token, or it could be a comma prime feature where y'all could expose a new endpoint backed by Mapbox.

This branch also includes a couple other minor fixes/improvements

Looking forward to the feedback!

### DIFF
--- a/release/files_common
+++ b/release/files_common
@@ -299,6 +299,8 @@ selfdrive/locationd/models/constants.py
 
 selfdrive/locationd/calibrationd.py
 
+selfdrive/locationd/speedlimitd.py
+
 selfdrive/logcatd/SConscript
 selfdrive/logcatd/logcatd_android.cc
 selfdrive/logcatd/logcatd_systemd.cc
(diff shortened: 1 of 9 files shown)


julianz-20 (approved):

adeebshihadeh:
Nice feature and thanks for the PR! We used to have something similar to this called mapd, but it was too unreliable. We won't merge this for the same reasons, but it might be nice to put this on the wiki for others to use.

mayfieldiv:
Gotcha, I guess this feature will need to wait until there's more navigation capability built into openpilot. I'll open another PR with the minor unrelated changes from this PR in the meantime.

PR #19565, opened 2020-12-21
UI: add function ui_read_params

deanlee:
Read all paramaters in  `ui_read_params()`.

because `frame` starts from 0, all parameters will be read at the first time call of `ui_read_params()`. so we don’t need to read them in `ui_init`.

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -29,7 +29,6 @@ void ui_init(UIState *s) {
   s->started = false;
   s->status = STATUS_OFFROAD;
   s->scene.satelliteCount = -1;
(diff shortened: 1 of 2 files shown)


selfdrive/ui/ui.cc
```diff
@@ -238,8 +237,28 @@ void update_sockets(UIState *s) {
   s->started = scene.thermal.getStarted() || scene.frontview;
 }
 
-void ui_update(UIState *s) {
+static void ui_read_params(UIState *s) {
+  static uint64_t frame = 0;
```
adeebshihadeh, line 241: why are we making a new frame? this should use the submaster frame

PR #19566, opened 2020-12-21
UI: refactor alerts

deanlee:
- update alert in function `update_alert` .
  - new struct `Alert`.
  - add three new alerts for UI: `ControlsUnseen`, `ControlsUnresponsive`, `CameraMalfunction`.
-  update status in `update_status`
- move alert blinking stuff into paint.cc.

### DIFF
--- a/selfdrive/ui/ui.hpp
+++ b/selfdrive/ui/ui.hpp
@@ -114,14 +114,6 @@ typedef struct UIScene {
   // responsive layout
   Rect viz_rect;
 
-<<<<<<< HEAD
-  std::string alert_text1;
-  std::string alert_text2;
-  std::string alert_type;
-  float alert_blinking_rate;
-  cereal::ControlsState::AlertSize alert_size;
-
-=======
   struct Alert {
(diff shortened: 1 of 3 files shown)


deanlee:
update status in one place

selfdrive/ui/ui.hpp
```diff
@@ -112,12 +112,14 @@ typedef struct UIScene {
...
-  float alert_blinking_rate;
-  cereal::ControlsState::AlertSize alert_size;
-
+  struct Alert {
+    std::string type, text1, text2;
+    cereal::ControlsState::AlertSize size;
+    cereal::ControlsState::AlertStatus status;
+    AudibleAlert sound;
+    float blinking_rate = 0.;
+  } alert;
```
adeebshihadeh, line 121: why do we need this? we already save the controlsState packet

deanlee: It's is not read-only, UI will update this data structure when constrls&camera timeout, I think using controlstate can be more complicated than a struct

PR #19567, opened 2020-12-21
Fix interp to be consistent with np.interp

jamcar23:
**Description** 
`interp` in `numpy_fast` doesn't match `np.interp` in the event the arguments differ in length. 

**Verification** 
Wrote unit tests to test current function with varied length arrays, tests failed. Update function to be consistent with numpy, tests pass.

**Warning**
Potentially breaking change in when interp's `xp` (which is usually `BP`) is shorter than `fp` (which is usually `V`). This change would would make the fast interp consistent with numpy but it's different than before. See the below unit test result, notice how numpy skips `-0.67` and jumps straight to `-0.3`:

```
======================================================================
FAIL: test_short_bp_long_v (tests.test_numpy_fast.InterpTest)
----------------------------------------------------------------------
  File "/tmp/openpilot/common/tests/test_numpy_fast.py", line 34, in test_short_bp_long_v
    np.testing.assert_equal(actual, expected)
  File "/root/.pyenv/versions/3.8.2/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 342, in assert_equal
    return assert_array_equal(actual, desired, err_msg, verbose)
  File "/root/.pyenv/versions/3.8.2/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 930, in assert_array_equal
    assert_array_compare(operator.__eq__, x, y, err_msg=err_msg,
  File "/root/.pyenv/versions/3.8.2/lib/python3.8/site-packages/numpy/testing/_private/utils.py", line 840, in assert_array_compare
    raise AssertionError(msg)
AssertionError: 
Arrays are not equal

Mismatched elements: 1 / 16 (6.25%)
Max absolute difference: 0.37
Max relative difference: 1.23333333
 x: array([-1.   , -1.   , -1.   , -0.84 , -0.8  , -0.774, -0.748, -0.67 ,
       -0.3  , -0.3  , -0.3  , -0.3  , -0.3  , -0.3  , -0.3  , -0.3  ])
 y: array([-1.   , -1.   , -1.   , -0.84 , -0.8  , -0.774, -0.748, -0.3  ,
       -0.3  , -0.3  , -0.3  , -0.3  , -0.3  , -0.3  , -0.3  , -0.3  ])
```

### DIFF
--- a/common/numpy_fast.py
+++ b/common/numpy_fast.py
@@ -6,13 +6,14 @@ def clip(x, lo, hi):
 
 def interp(x, xp, fp):
   N = len(xp)
+  T = len(fp)
 
   def get_interp(xv):
     hi = 0
     while hi < N and xv > xp[hi]:
       hi += 1
     low = hi - 1
-    return fp[-1] if hi == N and xv > xp[low] else (
+    return fp[-1] if hi >= T or (hi == N and xv > xp[low]) or (N < T and xv == xp[-1]) else (
       fp[0] if hi == 0 else
       (xv - xp[low]) * (fp[hi] - fp[low]) / (xp[hi] - xp[low]) + fp[low])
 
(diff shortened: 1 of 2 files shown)


pd0wm:
How can xp and fp differ in length? When I try to use numpy interp I get a value error:
```>>> np.interp(12, [0, 10], [0, 10, 5])
  File "<stdin>", line 1, in <module>
  File "<__array_function__ internals>", line 5, in interp
  File "/home/batman/.pyenv/versions/3.8.2/lib/python3.8/site-packages/numpy/lib/function_base.py", line 1423, in interp
    return interp_func(x, xp, fp, left, right)
ValueError: fp and xp are not of the same length.
```
I'm fine with the numpy_fast version throwing one as well.

PR #19569, opened 2020-12-21
remove sidebar.hpp

deanlee:


adeebshihadeh:
I think this is fine the way it is.

PR #19575, opened 2020-12-22
UI: reducing CPU usage from 13% to 4% when offroad

deanlee:
`use std::sleep_for `to accurately control the UI frequency.

`usleep` will be interrupted by a signal at any time, that will cause the UI freq to be very unstable and high. it will also cause the screen to go to sleep too quickly(less than 10s which is expected to 30s)


deanlee:
Added function `sleep_for(milliseconds)`  to replace `usleep` in other places later.

PR #19580, opened 2020-12-23
camera_qcom: refactor focus

deanlee:
Consistent with the original code, only the following refactor was performed:
- [x] Move focusing related code into class
- [x] Use a single mutex to guarantee that the focusing procedure are real atomic.(The code before refactoring is not atomic,std::atomic can only guarantee that a single variable is atomic,not the whole procedure )


ZwX1616:
#20064 looks better. It's not worth the complexity to move focus to a separate class and file. Keeping them all in qcom/qcom2 is fine. We can just improve the clarity inside those files?

deanlee:
ok I'll try

PR #19583, opened 2020-12-23
Refactor alert blinking

deanlee:


deanlee:
reset alpha to 1.0 if blink_rate is 0,to make sure the blinking is always start from 1.0.

pd0wm:
I don't like this hidden state in this function. Why not pass in the current alpha by reference so the function can update it?

deanlee:
How is it now

pd0wm:
I still don't like the hidden state in the static variable. Can't we compute the blink alpha using a `sin` and just pass in the current time/submaster frame?

deanlee:
A static variable is still used to ensure that alpha starts from 1.0. If this is not important, we can use current time instead of frame

deanlee:
Has also been merged into https://github.com/commaai/openpilot/pull/19566

adeebshihadeh:
Let's keep doing these small PRs when possible. They're much easier to review than the big ones. As far as the static var, I don't think we care about starting from 1.

deanlee:
use time now.

selfdrive/ui/paint.cc
```diff
@@ -356,6 +357,11 @@ static void ui_draw_vision_footer(UIState *s) {
   ui_draw_vision_face(s);
 }
 
+static float get_alert_alpha(float blink_rate) {
+  const float tm = millis_since_boot() / 1000;
+  return blink_rate > 0. ? (0.375 * sin(tm * 2 * M_PI * blink_rate) + 0.625) : 1.0;
```
adeebshihadeh, line 362: if we use cos, we can remove the ternary

PR #19587, opened 2020-12-24
Replace hardcoded ACCEL_SCALE

briskspirit:
Replace hardcoded accel scale value with calculated from car controller


pd0wm:
We can't import from carcontroller anywhere, since those files will get removed when we make the dascham version of openpilot. I like the change, so we should probably move all the constants into `values.py` and import from there.

briskspirit:
I started from moving to values.py but thought it won't be accepted ) So will do, move all constants into values.py

Issue #19579, opened 2020-12-23
Video frame dropping


dingyf12593:
Device: [EON Gold]
Dongle ID: [3f84146622df5f39]
Version: [0.8.1]
Car make/model: [ golf 7.5 2018]

Video frame dropping ，reinstall  twice，problem alwaysl

https://user-images.githubusercontent.com/6504134/102968705-55313600-452f-11eb-8183-00ac2b81f9d7.mov

nickgeurnop:
**Issue:** 
Comma Two display is dropping frames or appears "laggy." However, ALC and ACC operation is unaffected. Version 0.8.1 performs exceptional.

**Expected Behavior:** 
Comma Two display should relay real time driving footage uninterrupted and smoothly.   

Device: Comma Two
Dongle ID: 1615388ccbe3edc9
Version:  version 0.8.1 release2
Car make/model: 2019 Toyota Corolla Hatchback XSE (TSS2.0)

**Addition Notes**
Behavior occured as soon as I updated from 0.8.0 release2 --> 0.8.1 release2. I tried uninstalling and reinstalling once with behavior unchanged. 

https://user-images.githubusercontent.com/74438392/103042622-37bc9600-4548-11eb-905e-f328d936f84f.mov

davetapley:
Same as ⬆️ , can post more info if needed?

adeebshihadeh:
How consistently is this happening? @dingyf12593 your routes aren't fully uploaded so I couldn't look into it. However, I looked at a segment from one of @nickgeurnop's latest routes which was normal. I suspect this is isolated to UI, have any of you felt any degradation in driving performance when it happens? I can look into this more with specific routes and segments that are fully uploaded.

dingyf12593:
After upgrading to 0.8.1, the total driving distance is expected to be 40 km. This situation has been happening all the time, and the driving performance has not decreased @adeebshihadeh

dingyf12593:
le pro3 6+64GB ，Is it possible that the memory version is different ，If most people are normal

nickgeurnop:
@adeebshihadeh it occurs all the time, when openpilot is engaged or not. 

Driving performance is unaffected as far as I know. In fact, 0.8.1 is a great improvement over 0.8.0 in my experience. 

My only real concern would be video playback quality if I needed dashcam footage. Other than that it just looks weird since you can't turn off the screen. Hope it is an easy fix! Thank you for looking into this!

adeebshihadeh:
[...]

PR #19589, opened 2020-12-25
camerad: revert tbuffer dispatch change

adeebshihadeh:
possible fix for #19579


deanlee:
It shouldn't be here anyway

PR #19594, opened 2020-12-25
UI: simplify ui_draw_vision_speed

deanlee:
Delete unnecessary function calls `nvgBeginPath` and `nvgRect` ,`nvgText`  will center itself based on the x coordinate

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -329,10 +329,9 @@ static void ui_draw_vision_maxspeed(UIState *s) {
 
 static void ui_draw_vision_speed(UIState *s) {
   const float speed = s->scene.controls_state.getVEgo() * (s->is_metric ? 3.6 : 2.2369363);
-  char speed_str[32];
-  snprintf(speed_str, sizeof(speed_str), "%d", (int)std::nearbyint(speed));
+  const std::string speed_str = std::to_string((int)std::nearbyint(speed));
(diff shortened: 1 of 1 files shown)


selfdrive/ui/paint.cc
```diff
@@ -328,23 +328,12 @@ static void ui_draw_vision_maxspeed(UIState *s) {
...
-    speed = v_ego * 3.6 + 0.5;
-  }
-  const int viz_speed_w = 280;
-  const int viz_speed_x = viz_rect.centerX() - viz_speed_w/2;
+  const float speed = s->scene.controls_state.getVEgo() * (s->is_metric ? 3.6 : 2.2369363);
   char speed_str[32];
-
-  nvgBeginPath(s->vg);
-  nvgRect(s->vg, viz_speed_x, viz_rect.y, viz_speed_w, header_h);
+  snprintf(speed_str, sizeof(speed_str), "%d", (int)std::rint(speed));
```
pd0wm, line 333: Maybe use `std::nearbyint` since that one doesn't raise an exception? It's not entirely clear to me in what cases rint raised `FE_INEXACT`.

Maybe also use `std::format` instead of snprintf?

deanlee: std::format is in C++ 20,we can use std:to_string in c++11. but it has one more memory allocation than snprintf.

pd0wm: Let's use std::to_string. Those few allocated bytes are negligible at 20 fps.

PR #19595, opened 2020-12-25
manager tests + make all processes exit cleanly

adeebshihadeh:


selfdrive/logcatd/logcatd_systemd.cc
```diff
@@ -18,16 +31,18 @@ int main(int argc, char *argv[]) {
   assert(sd_journal_seek_tail(journal) >= 0);
 
   int r;
-  while (true) {
+  while (!do_exit) {
```
deanlee, line 34: I think one loop should be enough：
```
while (!do_exit) {
    int r = sd_journal_next(journal);
    assert(r >= 0);

    // Wait for new message if we didn't receive anything
    if (r == 0){
      r = sd_journal_wait(journal, 1000 * 1000);
      assert(r >= 0);
      continue;
    }
    ...
  }

```

adeebshihadeh: will leave this for now, want to keep the scope of this PR minimal

PR #19599, opened 2020-12-26
2021 Corolla Hybrid TS 1.8 (EU) engine fingerprint

mysliwiec-tech:
Adding the value for engine fingerprint, so my car is now working with openpilot. The rest of the values for radar, camera etc. were already there.


selfdrive/car/toyota/values.py
```diff
@@ -636,6 +636,7 @@ class CAR:
       b'\x018966342M5000\x00\x00\x00\x00',
       b'\x02896630ZQ3000\x00\x00\x00\x008966A4703000\x00\x00\x00\x00',
       b'\x02896630ZR2000\x00\x00\x00\x008966A4703000\x00\x00\x00\x00',
+      b'\x02896630ZT8000\x00\x00\x00\x008966A4703000\x00\x00\x00\x00,
```
adeebshihadeh, line 639: missing a closing quote

mysliwiec-tech: Thank you for suggestion! Have fixed that.

PR #19603, opened 2020-12-27
Added fingerprint for Ecu.programmedFuelInjection in Honda Civic Bosch

abrarmatin:


abrarmatin:
Added fingerprint for Honda Civic Ecu.programmedFuelInjection

PR #19615, opened 2020-12-29
ui_fill_rect

deanlee:
Split ui_draw_rect into two functions ui_draw_rect and ui_fill_rect. this makes the semantics clearer.


pd0wm:
I don't think the new code is more readable. Is there a way to simplify and also have `ui_fill_rect` and `ui_draw_rect` share some code?

deanlee:
I'll rethink how to do it better

PR #19619, opened 2020-12-29
Car Port: Nissan Leaf with ADAS ECU in alt location

razem-io:
Adds 2 Fingerprints for the Leaf 2nd Zero Edition (01/18 - 02/18) edition and  Tekna (02/18 - 01/19) which both have the ADAS module behind the [instrument cluster](https://github.com/commaai/openpilot/wiki/Nissan#instrument-cluster). 
Introduces a custom car class because the seatbelt signal is inverted.

Tested with 2.ZERO edition leaf for more then 400 km by me.
Tested with Tekna edition leaf by @Cargomatch.


selfdrive/car/nissan/values.py
```diff
@@ -7,6 +7,7 @@
 class CAR:
   XTRAIL = "NISSAN X-TRAIL 2017"
   LEAF = "NISSAN LEAF 2018"
+  LEAF_IC = "NISSAN LEAF 2018"
```
adeebshihadeh, line 10: this needs to be unique

razem-io: IC is short for "instrument cluster" as this version of the leaf does not have the adapter behind the glovebox but behind the instrument cluster. I am not yet sure if a differentiation between the glovebox and the IC version is even needed. If I comment one line at least the IC version works as expected.

razem-io: @adeebshihadeh done, also added a route which can not yet be found by ci

selfdrive/car/nissan/carstate.py
```diff
@@ -49,13 +49,16 @@ def update(self, cp, cp_adas, cp_cam):
     if self.CP.carFingerprint in [CAR.ROGUE, CAR.XTRAIL]:
       ret.seatbeltUnlatched = cp.vl["HUD"]["SEATBELT_DRIVER_LATCHED"] == 0
       ret.cruiseState.available = bool(cp_cam.vl["PRO_PILOT"]["CRUISE_ON"])
-    elif self.CP.carFingerprint == CAR.LEAF:
-      ret.seatbeltUnlatched = cp.vl["SEATBELT"]["SEATBELT_DRIVER_LATCHED"] == 0
+    elif self.CP.carFingerprint in [CAR.LEAF, CAR.LEAF_2_ZERO]:
+      if self.CP.carFingerprint == CAR.LEAF:
+        ret.seatbeltUnlatched = cp.vl["SEATBELT"]["SEATBELT_DRIVER_LATCHED"] == 0
+      # TODO: fix this for LEAF_2_ZERO
```
pd0wm, line 55: Did you find the seatbelt signal for this car?

pd0wm:
Uploaded the route to CI, so that should work now. Before we can merge we need to find the seatbelt signal.

selfdrive/car/nissan/carstate.py
razem-io: Yes, just yesterday. We checked every bit for 0/1 correlation and only found 0:640 and 130:640 to be good. Everything else was far off. Just pushed it.

razem-io: Also if this helps, the last bit of the route contains a clip where the car is in the garage. Everytime I cover the camera with my hand, the seatbelt is latched. I repeat this 5-8 times. This is how we found this signal to be good.

adeebshihadeh:
Hmm, I'm not sure we need to add another car for this model. Do you know if the new seatbelt signal works for the current Leaf? Why was it already defined in the DBC and not used?

razem-io:
> Hmm, I'm not sure we need to add another car for this model. Do you know if the new seatbelt signal works for the current Leaf? Why was it already defined in the DBC and not used?

Yes, my thoughts exactly. On the other hand, why do the other signals not work for me? That's why I made the differentiation. To not interfere with the already working ones. I can't be sure, if it works since I only have my leaf to test it.

adeebshihadeh:
I'd rather wait and have an existing Leaf user test the new signal than unnecessarily introduce a new car model.

razem-io:
Okay, will do that.

adeebshihadeh:
Any update on this? Maybe @avolmensky can provide some background

razem-io:
Yes, @willemcvu provided logs from his leaf in the Nissan discord channel yesterday. Cancel Seatbelt Message is also working but seems to be inverted with the zero edition. Not sure on how to proceed. Maybe another car class is needed after all. 

Let me check it again next week to make sure. Hopefully I will find sometime to work on it soon. I became a father today 😊. So my schedule is a bit mixed up for the next few days.

razem-io:
Okay, after some inspection I am pretty sure that the 2.ZERO edition inverts the seatbelt signal. Making the new car class nesseary, if I am not mistaken.

@willemcvu made a screenshot of his leafs seatbelt signals



Based on this screenshot one can see that the signals for ``SEATBELT_DRIVER_UNLATCHED` and `CANCEL_SEATBELT` are the same for the "normal" leaf.  "Normal" leaf uses 0 "2.Zero Edition" (my leaf) uses 1 as signal.

razem-io:
A few days ago, another instrument cluster leaf poped up in discord. Turns out, it has the same issues as the 2.ZERO edition. I added the fingerprint here and confirmed with @cargomatch that the inverted seatbelt signal also works for him.

He opened a PR here with his fingerprint: https://github.com/commaai/openpilot/pull/19923

@adeebshihadeh how should I proceed. It might make sense now, to rename the car class back to `LEAF_IC`? Everything else looks fine for me now.

razem-io:
Due to lack of feedback I implemented it the way I think it is best. If there are still ideas on how to improve please tell me. Will mark this ready for review.

Notes on the changes:
- Since there are now mutiple cars with an instrument cluster ADAS it makes sense to rename it back to LEAF_IC.
- I also renamed the glovebox car class from LEAF to LEAF_GB to keep the changes consistent.

adeebshihadeh:
Let's do this:
* Revert the changes to the existing Leaf
* Rename the new Leaf to the name of trim level instead of IC
* Add a comment explaining that the two models are the same except for the seatbelt signal. Hopefully this can be cleaned up in the future

razem-io:
* _Revert the changes to the existing Leaf_ (ok)
* _Rename the new Leaf to the name of trim level instead of IC_ (question)
The problem is 2.zero and Tekna edition are different trim levels. Should I use 2.zero or tekna as trim level? Could you give a conrecte example for the naming. I will just use this then.
* _Add a comment ..._ (ok)
The comment about the seatbelt is already there. I will add that everything else is the same except for the seatbelt signal.

adeebshihadeh:
Hmm, guess this is really like LEAF_ALT but LEAF_IC is a bit better. Let's just keep it.

selfdrive/car/nissan/values.py
```diff
@@ -28,6 +30,17 @@ class CAR:
       2: 5, 42: 8, 264: 3, 361: 8, 372: 8, 384: 8, 389: 8, 403: 8, 459: 7, 460: 4, 470: 8, 520: 1, 569: 8, 581: 8, 634: 7, 640: 8, 643: 5, 644: 8, 645: 8, 646: 5, 658: 8, 682: 8, 683: 8, 689: 8, 724: 6, 758: 3, 761: 2, 772: 8, 773: 6, 774: 7, 775: 8, 776: 6, 777: 7, 778: 6, 783: 3, 852: 8, 853: 8, 856: 8, 861: 8, 943: 8, 944: 1, 976: 6, 1008: 7, 1009: 8, 1010: 8, 1011: 7, 1012: 8, 1013: 8, 1019: 8, 1020: 8, 1021: 8, 1022: 8, 1057: 3, 1227: 8, 1228: 8, 1261: 5, 1342: 1, 1354: 8, 1361: 8, 1402: 8, 1459: 8, 1477: 8, 1497: 3, 1549: 8, 1573: 6, 1821: 8, 1837: 8
     },
   ],
+  # Leaf with ADAS @ instrument cluster / inverted seatbelt signal
+  CAR.LEAF_IC: [
+    # 2nd Zero Edition (01/18 - 02/18)
+    {
+      2: 5, 42: 6, 264: 3, 361: 8, 372: 8, 384: 8, 389: 8, 403: 8, 459: 7, 460: 4, 470: 8, 520: 1, 569: 8, 581: 8, 634: 7, 640: 8, 643: 5, 644: 8, 645: 8, 646: 5, 658: 8, 682: 8, 683: 8, 689: 8, 724: 6, 758: 3, 761: 2, 783: 3, 852: 8, 853: 8, 856: 8, 861: 8, 943: 8, 944: 1, 976: 6, 1001: 6, 1008: 7, 1011: 7, 1057: 3, 1227: 8, 1228: 8, 1229: 8, 1261: 5, 1342: 1, 1354: 8, 1361: 8, 1402: 8, 1459: 8, 1477: 8, 1497: 3, 1514: 6, 1549: 8, 1573: 6
+    },
+    # Tekna (02/18 - 01/19)
```
adeebshihadeh, line 39: remove these comments

selfdrive/car/nissan/values.py
```diff
@@ -6,7 +6,8 @@
 
 class CAR:
   XTRAIL = "NISSAN X-TRAIL 2017"
-  LEAF = "NISSAN LEAF 2018"
+  LEAF_GB = "NISSAN LEAF 2018" # Glovebox
+  LEAF_IC = "NISSAN LEAF 2018 Instrument Cluster"
```
adeebshihadeh, line 10: put the explanatory comment above the LEAF_IC line

selfdrive/car/nissan/values.py
razem-io: Done

selfdrive/car/nissan/values.py
razem-io: Done

selfdrive/car/nissan/carstate.py
```diff
@@ -49,13 +49,16 @@ def update(self, cp, cp_adas, cp_cam):
     if self.CP.carFingerprint in [CAR.ROGUE, CAR.XTRAIL]:
       ret.seatbeltUnlatched = cp.vl["HUD"]["SEATBELT_DRIVER_LATCHED"] == 0
       ret.cruiseState.available = bool(cp_cam.vl["PRO_PILOT"]["CRUISE_ON"])
-    elif self.CP.carFingerprint == CAR.LEAF:
-      ret.seatbeltUnlatched = cp.vl["SEATBELT"]["SEATBELT_DRIVER_LATCHED"] == 0
+    elif self.CP.carFingerprint in [CAR.LEAF, CAR.LEAF_IC]:
```
adeebshihadeh, line 52: this is unnecessary, just make an elif for each leaf

razem-io: I would then have to duplicate line 57 for each leaf. If you prefer this, I will implement it like that.

adeebshihadeh: missed that

adeebshihadeh:
Not sure why the static analysis checks are timing out. Can you try rebasing?

razem-io:
Done. There where too many commits anyway.

razem-io:
Not sure why this is happening. The logs actually look fine. It just hangs after flake8 check.

```2021-02-02T09:19:17.6302401Z [INFO] Installing environment for https://github.com/PyCQA/flake8.
2021-02-02T09:19:17.6306811Z [INFO] Once installed this environment will be reused.
2021-02-02T09:19:17.6307389Z [INFO] This may take a few minutes...
2021-02-02T09:19:20.3061749Z Check python ast.........................................................Passed
2021-02-02T09:19:20.3600052Z Check JSON...............................................................Passed
2021-02-02T09:19:20.3637010Z Check Xml............................................(no files to check)Skipped
2021-02-02T09:19:20.4937134Z Check Yaml...............................................................Passed
2021-02-02T09:19:20.5886997Z Check for merge conflicts................................................Passed
2021-02-02T09:19:20.6380361Z Check for broken symlinks................................................Passed
2021-02-02T09:19:28.5752552Z mypy.....................................................................Passed
2021-02-02T09:20:39.6926749Z flake8...................................................................Passed
2021-02-02T10:06:48.3635960Z ##[error]The operation was canceled.```

File: selfdrive/test/test_car_models.py [hunk 6]
```diff
@@ -472,11 +472,15 @@ def get_route_log(route_name):
   "fbbfa6af821552b9|2020-03-03--08-09-43": {
     'carFingerprint': NISSAN.XTRAIL,
     'enableCamera': True,
-  },
+  },
```
adeebshihadeh, line 475: revert this

razem-io: Ah, already wondered if this might be the issue. Done.

PR #19628, opened 2020-12-31
hyundai: update steer warning

gregjhogan:
current steer unavailable warning for hyundai does not include:
* LKAS11 CF_Lkas_ToiFlt == 1
* LKAS11 CF_Lkas_Chksum invalid
* LKAS11 CF_Lkas_MsgCount invalid
* LKAS11 CF_Lkas_ActToi == 0 and LKAS11 CR_Lkas_StrToqReq != 0
* LKAS11 CR_Lkas_StrToqReq > 2.8 Nm for more than a long period of time (105s?)
* LKAS11 CR_Lkas_StrToqReq > 3.2 Nm for more than a short period of time (0.1s?)
* LKAS11 CR_Lkas_StrToqReq changes by more than 0.078125 Nm (10 request units) per time step
* MDPS11 CR_Mdps_DrvTq > 6.0 Nm
* CLU11 CF_Clu_Vanz > 128 MPH
* SAS11 SAS_Angle > 90 degrees
(values based on palisade power steering firmware, could be different for other vehicles)

I believe you recover from these errors by sending zero torque for a short period of time

another very similar signal is CF_Mdps_ToiFlt which looks the same as above except it excludes:
* LKAS11 CF_Lkas_ToiFlt == 1

also, if setting LKAS11 CF_Lkas_ToiFlt = 1 when openpilot transitions from enabled to disabled causes a controlled wind-down of torque instead of the immediate drop to zero that currently happens, do we want to start setting it?

### DIFF
--- a/selfdrive/car/hyundai/carstate.py
+++ b/selfdrive/car/hyundai/carstate.py
@@ -34,7 +34,7 @@ class CarState(CarStateBase):
     ret.steeringTorque = cp.vl["MDPS12"]['CR_Mdps_StrColTq']
     ret.steeringTorqueEps = cp.vl["MDPS12"]['CR_Mdps_OutTq']
     ret.steeringPressed = abs(ret.steeringTorque) > STEER_THRESHOLD
-    ret.steerWarning = cp.vl["MDPS12"]['CF_Mdps_ToiUnavail'] != 0 or cp.vl["MDPS12"]['CF_Mdps_FailStat'] != 0
+    ret.steerWarning = cp.vl["MDPS12"]['CF_Mdps_ToiUnavail'] != 0 or cp.vl["MDPS12"]['CF_Mdps_ToiFlt'] != 0
 
     # cruise state
     if self.CP.openpilotLongitudinalControl:
@@ -194,7 +194,7 @@ class CarState(CarStateBase):
       ("CR_Mdps_StrColTq", "MDPS12", 0),
       ("CF_Mdps_ToiActive", "MDPS12", 0),
       ("CF_Mdps_ToiUnavail", "MDPS12", 0),
-      ("CF_Mdps_FailStat", "MDPS12", 0),
+      ("CF_Mdps_ToiFlt", "MDPS12", 0),
       ("CR_Mdps_OutTq", "MDPS12", 0),
 
       ("SAS_Angle", "SAS11", 0),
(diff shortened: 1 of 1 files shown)


pd0wm:
Then we should probably implement similar logic to the toyotas: https://github.com/commaai/openpilot/blob/master/selfdrive/car/toyota/carcontroller.py#L76

gregjhogan:
I thought this would be a good idea because we don't know all the criteria for most cars, and we have seen evidence of differences between vehicles, and a new car could come out with different limits, and this would give the user a heads-up that something is going on.

Related, I don't really like that openpilot cuts torque at 90 deg without a warning (feels very similar to reaching torque limit which has warning).  If we didn't cut torque at 90 deg my vehicle would hit the above limit in the firmware and the wheel slowly let the wheel back toward zero.  Instead, my wheel bounces back and forth at 90 deg as openpilot stops and starts sending torque, and perhaps setting `CF_Lkas_ToiFlt` will make it release slowly from 90 deg.  Or even simpler, maybe let it fault with this PR change (openpilot may just needs to send zero torque until `CF_Mdps_FailStat` clears)?

gregjhogan:
I see now steer state for hyundai is:
```
self.steer_state = cp.vl["MDPS12"]['CF_Mdps_ToiActive']  # 0 NOT ACTIVE, 1 ACTIVE
```

This is probably the best signal to use as it appears to indiacte that torque is being requested via LKAS11 CF_Lkas_ActToi and it is actively steering (no fault).  So now I am thinking change the `steerWarning` to be that signal, too, and add the logic you mentioned from toyota to send zero torque when in this state

selfdrive/car/hyundai/carstate.py
```diff
@@ -34,7 +34,7 @@ def update(self, cp, cp_cam):
     ret.steeringTorque = cp.vl["MDPS12"]['CR_Mdps_StrColTq']
     ret.steeringTorqueEps = cp.vl["MDPS12"]['CR_Mdps_OutTq']
     ret.steeringPressed = abs(ret.steeringTorque) > STEER_THRESHOLD
-    ret.steerWarning = cp.vl["MDPS12"]['CF_Mdps_ToiUnavail'] != 0
+    ret.steerWarning = cp.vl["MDPS12"]['CF_Mdps_ToiUnavail'] != 0 or cp.vl["MDPS12"]['CF_Mdps_ToiFlt'] != 0
```
adeebshihadeh, line 37: need to add this signal to the can parser

gregjhogan: yeah I noticed that in the build failure, will do

adeebshihadeh:
Is this good to merge or do you want to use `CF_Mdps_ToiActive`?

gregjhogan:
Since we use fault status everywhere else, I would say let's use `CF_Mdps_ToiFlt`
I think this is good now, but haven't had a chance to go out and drive on this PR

gregjhogan:
tested and this seems good now

this updated warning status should allow us to eliminate the steer angle limit check (covering it with a more broad and generic check), I will open a separate PR for that

PR #19629, opened 2020-12-31
UI: fix video frame drop

deanlee:
possible fix for: https://github.com/commaai/openpilot/issues/19579

`poll` did not work as expected due to inappropriate flag: `POLLOUT`(should be `POLLIN`) ,it also return -1 when `EINTR` occurred.that may cause some necessary `visionstream_get` calls be skipped,causing frame drop. 
I don't understand why there was no such issue before. May be related to this PR : https://github.com/commaai/openpilot/pull/19589 the combination of these two problematic codes caused  frame drop.

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -83,18 +83,8 @@ void ui_update_vision(UIState *s) {
     }
   }
 
-  if (s->vision_connected) {
-    if (!s->started) goto destroy;
-
-    // poll for a new frame
-    struct pollfd fds[1] = {{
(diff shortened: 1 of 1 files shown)


deanlee:
I think https://github.com/commaai/openpilot/pull/19549 is a more complete solution

deanlee:
create_thumbnail may be another cause of ui laggy

pd0wm:
Let's hold off on these changes for a little bit. Hopefully the new visionipc will be done soon, making all these fixes obsolete.

PR #19632, opened 2021-01-03
Make macos ci homebrew dep-caching fast

### LINKED ISSUE #19627: Make macOS CI fast
Currently, the mac CI tests take ~11 minutes fully cached. Only 4.5 minutes of that is building openpilot, and the rest is dependency installation and cache handling. Should be able to get this well under 10 minutes.

mayfieldiv:
Resolves #19627

1. Instead of caching the tar.gz downloads (that are then installed into Cellar during `brew install`), the Cellar directories themselves are now cached and `brew link`-ed on cache restore
2. Since the github action macos environment already has a bunch of homebrew things installed, they're now excluded from our cache to save several GBs of cache space/bandwidth
3. Use `restore-keys` to have fast dep installation when `mac_setup.sh` changes - still updates cache with new deps when cache key isn't an exact match - see [these docs](https://docs.github.com/en/free-pro-team@latest/actions/guides/caching-dependencies-to-speed-up-workflows#matching-a-cache-key) for more info
4. A fully cached build took [less than 6 minutes](https://github.com/mayfieldiv/openpilot/runs/1637997427) in my fork, and partial caching (from `restore-keys`) will speed up other builds as well, though I didn't test the timing for that

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -67,11 +67,8 @@ jobs:
         # key: ${{ hashFiles('tools/mac_setup.sh') }}
         key: asdfasdf4
     - name: Determine pre-existing Homebrew packages
-      if: steps.cache.outputs.cache-hit != 'true'
+      if: steps.dependency-cache.outputs.cache-hit != 'true'
       run: |
-        echo 'EXISTING_DOWNLOADS<<EOF' >> $GITHUB_ENV
-        ls -1 ~/Library/Caches/Homebrew/downloads >> $GITHUB_ENV
-        echo 'EOF' >> $GITHUB_ENV
         echo 'EXISTING_CELLAR<<EOF' >> $GITHUB_ENV
         ls -1 /usr/local/Cellar >> $GITHUB_ENV
         echo 'EOF' >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


.github/workflows/test.yaml
```diff
@@ -62,14 +68,36 @@ jobs:
...
     - name: Install dependencies
       run: ./tools/mac_setup.sh
     - name: Build openpilot
       run: eval "$(pyenv init -)" && scons -j$(nproc)
-    - name: Brew cleanup
-      run: brew cleanup || true # keeps our cache small
+    - name: Remove pre-existing Homebrew packages for caching
+      if: steps.dependency-cache.outputs.cache-hit != 'true'
+      run: |
+        new_cellar=$(ls -1 /usr/local/Cellar)
```
adeebshihadeh, line 88: cd into /usr/local/Cellar so we don't have to write it everywhere

.github/workflows/test.yaml
```diff
@@ -62,14 +68,36 @@ jobs:
...
           ~/Library/Caches/pip
           ~/Library/Caches/pipenv
-          ~/Library/Caches/Homebrew
-        key: ${{ hashFiles('tools/mac_setup.sh') }}
+          /usr/local/Cellar
+          ~/github_brew_cache_entries.txt
+        key: macos-deps-${{ hashFiles('tools/mac_setup.sh') }}
+        restore-keys: macos-deps-
+    - name: Brew link restored dependencies
+      if: steps.dependency-cache.outputs.cache-hit == 'true'
```
adeebshihadeh, line 76: is this if necessary? should do nothing if ~/github_brew_cache_entries.txt doesn't exist

adeebshihadeh:
Nice work!

PR #19635, opened 2021-01-03
hyundai: update LFAHDA_MFC generation

gregjhogan:
~needs testing on non-HDA vehicle - guessing that setting `HDA_USM` and `LFA_USM` is OK everywhere~ no need to set these to make HDA work

requires https://github.com/commaai/opendbc/pull/338 and https://github.com/commaai/opendbc/pull/342

### DIFF
--- a/selfdrive/car/hyundai/hyundaican.py
+++ b/selfdrive/car/hyundai/hyundaican.py
@@ -67,21 +67,11 @@ def create_clu11(packer, frame, clu11, button):
   return packer.make_can_msg("CLU11", 0, values)
 
 
-def create_lfa_mfa(packer, frame, enabled):
+def create_lfahda_mfc(packer, enabled, hda_set_speed=0):
   values = {
-    "ACTIVE": enabled,
+    "LFA_Icon_State": 2 if enabled else 0,
+    "HDA_Active": 1 if hda_set_speed else 0,
+    "HDA_Icon_State": 2 if hda_set_speed else 0,
+    "HDA_VSetReq": hda_set_speed,
   }
-
-  # ACTIVE 1 = Green steering wheel icon
-
-  # LFA_USM 2 & 3 = LFA cancelled, fast loud beeping
-  # LFA_USM 0 & 1 = No mesage
-
-  # LFA_SysWarning 1 = "Switching to HDA", short beep
-  # LFA_SysWarning 2 = "Switching to Smart Cruise control", short beep
-  # LFA_SysWarning 3 =  LFA error
-
-  # ACTIVE2: nothing
-  # HDA_USM: nothing
-
   return packer.make_can_msg("LFAHDA_MFC", 0, values)
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -1,7 +1,7 @@
 from cereal import car
 from common.realtime import DT_CTRL
 from selfdrive.car import apply_std_steer_torque_limits
-from selfdrive.car.hyundai.hyundaican import create_lkas11, create_clu11, create_lfa_mfa
+from selfdrive.car.hyundai.hyundaican import create_lkas11, create_clu11, create_lfahda_mfc
 from selfdrive.car.hyundai.values import Buttons, CarControllerParams, CAR
 from opendbc.can.packer import CANPacker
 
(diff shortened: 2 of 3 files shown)


gregjhogan:
vehicles with HDA have speed limit on CAN (map based, it is pretty good) so with this someone can automatically adjust the ACC set speed up and down based on the speed limit

selfdrive/car/hyundai/hyundaican.py
```diff
@@ -67,21 +67,26 @@ def create_clu11(packer, frame, clu11, button):
   return packer.make_can_msg("CLU11", 0, values)
 
 
-def create_lfa_mfa(packer, frame, enabled):
+def create_lfahda_mfc(packer, enabled, hda_active=0, hda_set_speed=0):
   values = {
-    "ACTIVE": enabled,
+    # HDA_C_State:
```
adeebshihadeh, line 72: I think this kind of stuff should really live in the DBC. I could easily see it getting lost in a cleanup or refactor. What happens when this gets out of sync with the DBC defined comments/vals?

gregjhogan: yeah, i was just maintaining what was already there

gregjhogan: value definitions moved to https://github.com/commaai/opendbc/pull/338

gregjhogan:
updated to copy the USM signals from the camera, which ensures we don't break any non-HDA vehicles

hopefully ready to merge now (once replay output is updated, opendbc changes are merged, and opendbc ref is updated)

selfdrive/car/hyundai/hyundaican.py
```diff
@@ -67,21 +67,10 @@ def create_clu11(packer, frame, clu11, button):
...
-
-  # LFA_SysWarning 1 = "Switching to HDA", short beep
-  # LFA_SysWarning 2 = "Switching to Smart Cruise control", short beep
-  # LFA_SysWarning 3 =  LFA error
-
-  # ACTIVE2: nothing
-  # HDA_USM: nothing
-
+def create_lfahda_mfc(packer, enabled, lfahda_mfc, hda_active=0, hda_set_speed=0):
+  values = lfahda_mfc
```
adeebshihadeh, line 71: remove this

gregjhogan: what do you mean? the goal is to copy fields like `HDA_USM` and `LFA_USM` being sent by camera so I copied what we do for `CLU11`

adeebshihadeh: I just meant that line is unnecessary. why not just use lfahda_mfc?

gregjhogan: I made it consistent with both `create_clu11` and `create_lkas11` do you want me to change it and make it different?

If we want to update these I really think we should change them to not modify the param passed in because you would never expect these functions to modify the parameters (which can lead to bugs in the future).  I think that it is best to treat params as immutable unless it is obvious (e.g. `memcpy`) or special cases (e.g. not wanting to make a copy for super high performance)

adeebshihadeh: I'm not a fan of the pattern we have right now of copying in car state then modifying in car controller. One of the problems is it's not clear what fields are copied, and it relies on the can parser using a dict. I think it would be cleaner to just make a new dict in these car controller functions and explicitly copy out every field you want.

gregjhogan: that is similar to how it started, for example:
https://github.com/commaai/openpilot/blob/6e388948e3ecbb6b9ffa7a05fcd6ce3966ac5f27/selfdrive/car/hyundai/hyundaican.py#L47-L63

so what do you want to do for this PR?

gregjhogan:
Removed copying `LFA_USM` and  `HDA_USM` because the camera faults when it notices it is not requesting torque, but the power steering says control is active, and then these signals have a value that seems to indicate a fault, and you don't have to set these signals to get HDA to work.

selfdrive/car/hyundai/hyundaican.py
gregjhogan: no longer relevant (not copying values from LFAHDA_MFC)

gregjhogan:
for anyone in the future who might be interested in building good HDA as a feature (read speed limit on CAN, adjust set speed with this message) the speed limit information is in 0x225.  It isn't simple, though, because the messages 0x222-0x225 fire based on events like taking a turn, it sends each message 3 times (even counter is the same in the set), and the information in the message seems to change based on a few identification bytes (they aren't all speed limit messages).  Once you find the speed limit in the message, you will discover it isn't the current speed limit, but the speed limit at some point in the future (seems like navigation type information).

Here is an example that works to get the camera to enable HDA mode (openpilot not even in the loop) and to change the speed limit to whatever you desire.  This requires a panda between the camera and C-CAN forwarding all addrs except 0x222-0x225 (not sure if 0x222-0x224 need to be blocked) so you can inject 0x225.
https://github.com/gregjhogan/panda/blob/examples/examples/hda_test.py

PR #19636, opened 2021-01-03
ui.py: white text on graphs

ghost:
* improved readability through white text
* automatic size selection based on window size

### DIFF
--- a/tools/replay/ui.py
+++ b/tools/replay/ui.py
@@ -18,7 +18,7 @@ from selfdrive.controls.lib.vehicle_model import VehicleModel
 import cereal.messaging as messaging
 from tools.replay.lib.ui_helpers import (_BB_TO_FULL_FRAME, _FULL_FRAME_SIZE, _INTRINSICS,
                                          BLACK, BLUE, GREEN,
-                                         YELLOW, RED,
+                                         YELLOW, RED, WHITE,
                                          CalibrationTransformsForWarpMatrix,
                                          draw_lead_car, draw_lead_on, draw_mpc,
                                          extract_model_data,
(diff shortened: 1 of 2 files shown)


pd0wm:
Can you fix the static analysis checks? By writing to `HOR` you have to mark it as global explicitly. You can probably clean it up so it's only set once the window is available.

tools/replay/ui.py
```diff
@@ -45,7 +41,14 @@ def ui_thread(addr, frame_address):
   pygame.font.init()
   assert pygame_modules_have_loaded()
 
-  if HOR:
+  disp_info = pygame.display.Info()
+  max_height = disp_info.current_h
+
+  hor_mode = os.getenv("HORIZONTAL") is not None
+  hor_mode = True if max_height < 960+300 else hor_mode
+  angle_scale = 5.0
```
pd0wm, line 49: No need to change angle_scale

tools/replay/ui.py
```diff
@@ -234,8 +237,8 @@ def ui_thread(addr, frame_address):
     SPACING = 25
 
     lines = [
-      info_font.render("ENABLED", True, GREEN if sm['controlsState'].enabled else BLACK),
-      info_font.render("BRAKE LIGHTS", True, RED if sm['carState'].brakeLights else BLACK),
+      info_font.render("ENABLED", True, GREEN if sm['controlsState'].enabled else WHITE),
```
pd0wm, line 240: Let's leave the colors for enabled and brake. Now it's not immediately clear if white enable means enabled or not.

tools/replay/ui.py
BraunPhilipp: Appreciate the quick feedback 😅

pd0wm:
The plots are much more readable! We should have done this earlier :)

BraunPhilipp:
Noticed the readability issue and thought it would be a nice first PR 😉

pd0wm:
You can ignore the macOS failure. Not related to your PR.

PR #19639, opened 2021-01-04
Refactor UIScene,prepare for update data in background thread

deanlee:
Because we need to update and populate data  in the background thread later,the main thread is only responsible for the drawing routines.so we need to rearrange the data structure and put all the data that needs to be updated by background thread in the structure `UIScene`, and then we can copy it to the UI thread.
This PR modifies the data structure of `UIScene` and prepares for future background updates.

- [x] Put the data that needs to be updated in the background into `UIScene`
- [x] Put the data that needs to be updated by both threads into `UIState`(they should be protected by std::atomic later)
- [x] Use variables instead of ceal::read to save data for thread safety. (Except model, to make this one easier to review)

### DIFF
--- a/selfdrive/ui/android/ui.cc
+++ b/selfdrive/ui/android/ui.cc
@@ -35,14 +35,14 @@ static void handle_display_state(UIState *s, bool user_input) {
   constexpr float accel_samples = 5*UI_FREQ;
   static float accel_prev = 0., gyro_prev = 0.;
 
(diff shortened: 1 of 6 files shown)


deanlee:
Set as a draft for now, I am not sure if it is worth updating in the background thread,need to do more testing to determine if it is necessary to introduce additional complexity

PR #19641, opened 2021-01-04
VisionIPC 2.0

pd0wm:
- [x] Handling reconnects when camerad stops/starts
- [x] Sending frame metadata together with the buffer id over msgq (e.g. frame id, timestamps)
- [x] Handling aligned RGB buffers, on QCOM there is some padding to make the buffer more cache friendly
- [x] Changing camera_qcom and camera_qcom2 to use the new API
- [x] Changing loggerd and modeld to use the new API
- [x] General cleanup, e.g. making sure everything shuts down nicely, implement timeouts on all the recive calls
- [x] Fix snapshot.py
- [x] Add support for zmq as a transport layer
- [x] Writing tests + testing in testing closet


selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -79,37 +87,20 @@ void CameraBuf::init(cl_device_id device_id, cl_context context, CameraState *s,
...
+  rgb_stride = vipc_server->get_buffer(rgb_type)->stride;
+
+  vipc_server->create_buffers(yuv_type, YUV_COUNT, false, rgb_width, rgb_height);
 
   if (ci->bayer) {
     cl_program prg_debayer = build_debayer_program(device_id, context, ci, this);
     krnl_debayer = CL_CHECK_ERR(clCreateKernel(prg_debayer, "debayer10", &err));
     CL_CHECK(clReleaseProgram(prg_debayer));
+  } else {
+    krnl_debayer = nullptr;
```
deanlee, line 100: `krnl_debayer` has already been initialized as nullptr by `CameraBuf() = default;`  .maybe we can remove the `else` here

selfdrive/manager.py
```diff
@@ -440,7 +440,7 @@ def manager_thread():
   cloudlog.info({"environ": os.environ})
 
   # save boot log
-  subprocess.call(["./loggerd", "--bootlog"], cwd=os.path.join(BASEDIR, "selfdrive/loggerd"))
+  # subprocess.call(["./loggerd", "--bootlog"], cwd=os.path.join(BASEDIR, "selfdrive/loggerd"))
```
adeebshihadeh, line 443: why is this commented out?

SConstruct
```diff
@@ -314,19 +315,20 @@ if SHARED:
 else:
   cereal = [File('#cereal/libcereal.a')]
   messaging = [File('#cereal/libmessaging.a')]
+  visionipc = [File('#cereal/libvisionipc.a')]
```
adeebshihadeh, line 318: why don't we import these from the cereal SConscript?

selfdrive/ui/ui.hpp
```diff
@@ -19,12 +19,13 @@
...
 
 #include "common/mat.h"
-#include "common/visionipc.h"
 #include "common/visionimg.h"
 #include "common/framebuffer.h"
 #include "common/modeldata.h"
 #include "common/params.h"
 #include "sound.hpp"
+#include "visionipc.h"
+#include "visionipc_client.h"
```
adeebshihadeh, line 28: we should probably standardize our include path convention. just reading this header file, I have no idea where to find visionipc*.h

selfdrive/manager.py
pd0wm: oh, yes. That was when loggerd didn't build yet. I'll turn that back on.

SConstruct
pd0wm: Not sure. Decided to do it the same as messaging, but can look into it after this PR.

selfdrive/ui/ui.hpp
pd0wm: We did the same for messaging, but I agree.

PR #19643, opened 2021-01-04
Replace usleep with util::sleep_for

deanlee:
Related PR: https://github.com/commaai/openpilot/pull/19575

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -261,8 +261,7 @@ void can_recv_thread() {
     uint64_t cur_time = nanos_since_boot();
     int64_t remaining = next_frame_time - cur_time;
     if (remaining > 0){
-      int sleep = remaining / 1000;
-      std::this_thread::sleep_for(std::chrono::microseconds(sleep));
+      std::this_thread::sleep_for(std::chrono::nanoseconds(remaining));
     } else {
       if (ignition){
         LOGW("missed cycles (%d) %lld", (int)-1*remaining/dt, remaining);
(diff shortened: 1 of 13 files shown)


selfdrive/boardd/boardd.cc
```diff
@@ -260,8 +261,8 @@ void can_recv_thread() {
     uint64_t cur_time = nanos_since_boot();
     int64_t remaining = next_frame_time - cur_time;
     if (remaining > 0){
-      useconds_t sleep = remaining / 1000;
-      usleep(sleep);
+      int sleep = remaining / 1000 / 1000;
+      util::sleep_for(sleep);
```
pd0wm, line 265: Can we add a more precise sleep here? I'd like to have higher accuracy that 1ms. This is what drives the whole controls stack.

selfdrive/boardd/boardd.cc
```diff
@@ -260,8 +261,8 @@ void can_recv_thread() {
     uint64_t cur_time = nanos_since_boot();
     int64_t remaining = next_frame_time - cur_time;
     if (remaining > 0){
-      useconds_t sleep = remaining / 1000;
-      usleep(sleep);
+      int sleep = remaining / 1000;
+      std::this_thread::sleep_for(std::chrono::microseconds(sleep));
```
pd0wm, line 265: If we're using std::chrono here we can also get rid of `sleep` and do `      std::this_thread::sleep_for(std::chrono::nanoseconds(remaining));`

pd0wm:
Yeah, macOS builds seem to be broken. I'll merge.

PR #19644, opened 2021-01-04
Toyota Highlander FW versions

slavka89:
added new ecu.engine. for highlander_tss2


pd0wm:
What is your dongle id

slavka89:
dongle is f114d8080c260ba3

PR #19645, opened 2021-01-04
2018 Lexus NX Hybrid FW versions

rickyb82:
New Ecu Engine and Ecu Esp for NX Hybrid My18 European Edition, I'm From Italy


pd0wm:
What is your dongle id?

rickyb82:
f889f2b77ff9b903

I've tested fork installation but i've black screen after reboot, why?

Please update values because every reboot, if i've internet connection, i lost my mod...

Thankyou

PR #19650, opened 2021-01-04
cleanup unused pip packages

adeebshihadeh:


adeebshihadeh:
these are used outside of openpilot

PR #19651, opened 2021-01-05
Prius Prime 2021

VirtuallyChris:
Same platform as the TSS2 Prius

Dongle ID 654b68829e15b1b6


VirtuallyChris:
Confirmed working!

PR #19655, opened 2021-01-05
cleanup draw alert

deanlee:
Remove unnecessary conditional statements: `alert_size` cannot be` AlertSize::NONE`


pd0wm:
Should we also remove it from the size map then?

PR #19656, opened 2021-01-05
refactor ui_draw_image

deanlee:
pass `const Rect &rect` as function parameter


pd0wm:
Should we have `ui_draw_rect` also take in a rect? I assume you can also construct it in place, so no need to create temporary variables if we don't want to.

deanlee:
ok,I'll do it later

PR #19658, opened 2021-01-05
UI: refactor transform

deanlee:
1. Only calculate transform once at initialization (`ui_nvg_init`)
2. Transform to screen coordinates in function car_space_to_full_frame.
3. Make sure the lead car indicator  is always displayed.

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -65,10 +65,11 @@ static void ui_draw_text(NVGcontext *vg, float x, float y, const char* string, f
 static void draw_chevron(UIState *s, float x_in, float y_in, float sz,
                           NVGcolor fillColor, NVGcolor glowColor) {
   vertex_data out;
-  if (!car_space_to_full_frame(s, x_in, y_in, 0.0, &out)) return;
+  car_space_to_full_frame(s, x_in, y_in, 0.0, &out);
 
   auto [x, y] = out;
   sz = std::clamp((sz * 30) / (x_in / 3 + 30), 15.0f, 30.0f);
+  y = std::min(s->scene.viz_rect.bottom() - sz / 2,  y);
 
   // glow
   float g_xo = sz/5;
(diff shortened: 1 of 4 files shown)


selfdrive/ui/paint.cc
```diff
@@ -70,10 +71,12 @@ static void draw_chevron(UIState *s, float x_in, float y_in, float sz,
                           NVGcolor fillColor, NVGcolor glowColor) {
   float x, y;
   if (!car_space_to_full_frame(s, x_in, y_in, 0.0, &x, &y)) {
-    return;
+    if (x < 0 || y < 0) 
+      return;
   }
 
   sz = std::clamp((sz * 30) / (x_in / 3 + 30), 15.0f, 30.0f);
+  y = std::min(s->scene.viz_rect.bottom() - sz / 2,  y);
```
deanlee, line 79: since y is already a screen coordinate, we can make sure it's always displayed here

deanlee:
looks ok on my device, but not tested on the road,I'm not sure that the real environment will also work.

we already applied room by `nvgScale(s->vg, zoom, zoom);` can this solve the problem you said?

pd0wm:
With the default zoom level you don't see the issue. Try setting the zoomlevel to something like 1 and make sure the lane lines run all the way to the edge of the screen.

deanlee:
Yes, I've seen the problem.any further suggestions?

deanlee:
It seems that we need the nearest point outside the screen

pd0wm:
> It seems that we need the nearest point outside the screen

Yes. Just keeping one point outside the screen would also fix it.

deanlee:
the problem fixed now

deanlee:
Still use margins to simplify the problem

pd0wm:
I understand the margin is not the cleanest solution, but keeping the first point outside the screen needed a lot of non trivial code. So agree with reverting back.

deanlee:
I think the best part of this PR is that the indicator of the lead car can always be displayed.It will not disappear if the lead car is too close like the previous version.It’s clearer and safer for users

PR #19660, opened 2021-01-05
Mpc rework2

haraschax:
1.  Removes polyfits, this increases accuracy of MPC and cleans up code
2. Different lag compensation
3. Increased estimated lag for all cars, by .2s to account for other effects
4. Remove lanelines from MPC
5. Use model heading instead of computing heading from points


selfdrive/test/process_replay/model_replay_ref_commit
```diff
@@ -1 +1 @@
-852c79998828975cce184114537b0067b80bc608
+4d71a89ccbfd351cbe58fcf217ee2eefa48eee2d
```
adeebshihadeh, line 1: why did this change?

selfdrive/test/longitudinal_maneuvers/plant.py
```diff
@@ -113,6 +113,7 @@ def __init__(self, lead_relevancy=False, rate=100, speed=0.0, distance_lead=2.0)
       Plant.logcan = messaging.pub_sock('can')
       Plant.sendcan = messaging.sub_sock('sendcan')
       Plant.model = messaging.pub_sock('model')
+      Plant.modelV2 = messaging.pub_sock('modelV2')
```
adeebshihadeh, line 116: rebase and  just add this to Plant.pm

selfdrive/controls/lib/lateral_mpc/lib_mpc_export/acado_qpoases_interface.cpp
```diff
@@ -40,7 +40,7 @@ int acado_solve( void )
 {
 	acado_nWSR = QPOASES_NWSRMAX;
 
-	QProblem qp(24, 40);
+	QProblem qp(20, 32);
```
adeebshihadeh, line 43:
```suggestion
	QProblem qp(QPOASES_NVMAX, QPOASES_NCMAX);
```

selfdrive/test/longitudinal_maneuvers/plant.py
HaraldSchafer: fixed

selfdrive/test/process_replay/model_replay_ref_commit
HaraldSchafer: The time index constants of the model were changed so they are unified with MPC. Caused a rounding error.

PR #19661, opened 2021-01-05
common signal handle

deanlee:
- handle `SIGINT&SIGTERM` in class `SignalState`
- `use std::atomic `instead of `volatile sig_atomic_t`

### DIFF
--- a/selfdrive/common/utilpp.h
+++ b/selfdrive/common/utilpp.h
@@ -88,8 +88,8 @@ public:
     std::signal(SIGTERM, (sighandler_t)set_do_exit);
   };
   inline operator bool() { return do_exit; }
-  inline ExitHandler& operator=(bool exit) {
-    do_exit = exit;
+  inline ExitHandler& operator=(bool v) {
+    do_exit = v;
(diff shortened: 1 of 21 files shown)


selfdrive/common/utilpp.h
```diff
@@ -2,13 +2,17 @@
...
-
+#include <csignal>
 #include <string>
 #include <memory>
+#include <atomic>
 #include <sstream>
 #include <fstream>
 #include <thread>
 #include <chrono>
+#ifndef sighandler_t
```
adeebshihadeh, line 13: add a space above

selfdrive/common/utilpp.h
```diff
@@ -76,6 +80,16 @@ inline void sleep_for(const int milliseconds) {
 }
 }
 
+class SignalState {
```
adeebshihadeh, line 83: maybe something like ExitHandler would be a better name?

adeebshihadeh:
Nice cleanup. Ran the tests to check everything exits cleanly still and all good!

PR #19663, opened 2021-01-05
Move all CarControllerParams into values.py

pd0wm:


selfdrive/controls/lib/latcontrol_indi.py
```diff
@@ -4,7 +4,7 @@
 from cereal import log
 from common.realtime import DT_CTRL
 from common.numpy_fast import clip
-from selfdrive.car.toyota.values import SteerLimitParams
```
adeebshihadeh, line 7: why is this importing from toyota? we have hyundai models that use INDI

pd0wm: Because INDI needs to be aware of the rate limits. We should probably fix it such that it used the proper rate limits.

Issue #2545, opened 2020-11-13
UI - Yellow lead indicator isn't in 3D 


ShaneSmiskol:
Whenever I'm going up or down a hill with a lead, the yellow indicator doesn't follow the car and stays in its current y position.

VirtuallyChris:
I noticed this and was about to report it. Nice job 😆

HaraldSchafer:
Ah yes, I can fix that

HaraldSchafer:
@deanlee Interested in looking into this? The lead prediction of the model only predicts and x and y, no z (height). But we could use the height of predicted path at that point to draw the lead triangle.

deanlee:
@HaraldSchafer  good idea. I'd like to have a try in pull request https://github.com/commaai/openpilot/pull/2525

adeebshihadeh:
@deanlee can we bring this out of the big UI PR and make a small PR for just this?

deanlee:
@adeebshihadeh added 20 lines of code in this new PR: https://github.com/commaai/openpilot/pull/19793

PR #19674, opened 2021-01-06
UI: Refactor lead_data

deanlee:
Releated issue : https://github.com/commaai/openpilot/issues/2545

### DIFF
--- a/selfdrive/ui/ui.hpp
+++ b/selfdrive/ui/ui.hpp
@@ -138,7 +138,8 @@ typedef struct UIScene {
 
   // radarState
   struct LeadData{
-    cereal::RadarState::LeadData::Reader data;
+    bool status;
+    float d_rel, v_rel, y_rel;
     vertex_data vd;
   } lead[2] = {};
 } UIScene;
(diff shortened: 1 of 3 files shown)


Torresangl925 (approved):

selfdrive/ui/ui.hpp
```diff
@@ -131,6 +130,13 @@ typedef struct UIScene {
   track_vertices_data track_vertices;
   line_vertices_data lane_line_vertices[4];
   line_vertices_data road_edge_vertices[2];
+
+  // radarState
+  struct LeadData{
+    bool status;
+    float d_rel, v_rel, y_rel;
+    vertex_data vd;
+  } lead[2] = {};
```
adeebshihadeh, line 139: why create this struct that's just a subset of LeadData from cereal?

deanlee:
@adeebshihadeh  Do you think this would be better?

deanlee:
revert "use cereal::reader". I still think a struct will make the code simpler

deanlee:
merged https://github.com/commaai/openpilot/pull/19705

PR #19675, opened 2021-01-06
Qt offroad: pairing

ghost:
Pair the app with the device in Qt


grekiki:
Still some crashes from qt network management and the library for getting the rsa signature.

grekiki:
Current commit seems a lot more friendly to the file system

grekiki:
Could use some testing with bad network connection or reconnecting after a while.

grekiki:
No more event loop on networking, should be stable now hopefully.

PR #19689, opened 2021-01-07
logcatd bugfix + tests

adeebshihadeh:


adeebshihadeh:
If we add the ANDROID_LOG_NONBLOCK flag, subsequent reads after the first always return empty. Without the flag, the blocking read can make logcatd not exit cleanly if no messages come in. @deanlee interested in looking into this?

deanlee:
I'm working on it : https://github.com/commaai/openpilot/pull/19994

PR #19693, opened 2021-01-08
Fix loggerd to handle binary params properly part 2 of 3

sobomax:
**Description** This issue has been discovered while investigating failure of the tools/lib/logreader.py to decode latest rlog with the following exception:

```
  root@localhost:/data/openpilot$ python tools/lib/logreader.py /tmp/xxx_2021-01-04--21-40-41--0--rlog
    File "tools/lib/logreader.py", line 147, in <module>
      print(msg)
    File "capnp/lib/capnp.pyx", line 1091, in capnp.lib.capnp._DynamicStructReader.str
  UnicodeDecodeError: 'utf-8' codec can't decode byte 0xd5 in position 30118: invalid continuation byte
```
Further investigation and [discussion](https://github.com/capnproto/capnproto/pull/1143) revealed that the loggerd is abusing log.capn schema by storing binary blob into a value declared as Text:
```
        ( key = "PandaFirmware",
          value = "\x11\x01[\x16|GSe\xd5\x1a+\x94\xff\\f|" ),
```
while key/value declared as:
```
struct InitData {
[...]
  params @17 :Map(Text, Text);
```
Not only that, the binary blob also gets truncated at '\0'. The real value is as follows:
```
        ( key = "PandaFirmware",
          value = (
            bin = "\021\001[\026|GSe\325\032+\224\377\\f|\000\241\321\366j\360\247\354E?:\274\205\301\035\241\241T\370X\224\031\367P\257\350*eY\223rt\246\204=\223\301u\207\277\027xE\tn\365?J\234$\032\375\226\300a\215MB\326\342 }\r\366]5\250\236\006\312\000\025\353>Y\v+U\340C\321\017M\231\0307\220\236\303\005\327\'gJ\302Kv\270V\301F8\246\030\206[\344\2239T\003\275" ) ),
```

This is part two of the change, there is also matching change for the [cereal to convert Text into Union{Text, Data}](https://github.com/commaai/cereal/pull/109) and also bug in the capnproto to pretty-print [binary data properly in python3](https://github.com/capnproto/capnproto/pull/1143). 

**Verification** Log has been collected with both cereal and this change applied and then decoded on a machine with patched pycapnp. Data is decoded correctly without any exceptions.

**Note** The automated test(s) would be failing until https://github.com/commaai/cereal/pull/109 is merged and submodule is updated. The test completes after that as can bee seen in my working branch here: https://github.com/sobomax/openpilot/actions/runs/470280650

### DIFF
--- a/selfdrive/loggerd/loggerd.cc
+++ b/selfdrive/loggerd/loggerd.cc
@@ -458,7 +458,7 @@ kj::Array<capnp::word> gen_init_data() {
     for (auto& kv : params_map) {
       auto lentry = lparams[i];
       lentry.setKey(kv.first);
-      lentry.setValue(kv.second);
+      lentry.setValue(capnp::Data::Reader((const kj::byte*)kv.second.data(), kv.second.size()));
       i++;
     }
   }


pd0wm:
Why don't we change the field to be binary instead of text?

sobomax:
@pd0wm hmm, good question. However other parameters (vast majority of them) are actually legit text. Also I am in discussion with @kentonv https://github.com/capnproto/capnproto/pull/1143 to possibly changing pretty-print format for Data blobs to be `dataField = 0x"abcd1234"`, which if implemented would make that information more obscure, at least when pyton tools are used.

pd0wm:
What is the outcome of the capnp PR? Do I understand correctly data blobs will still be printed as text? In that case we should probably just change the field to data.

sobomax:
@pd0wm well yes, it's been sorted out by escaping upper half of the code space. As such change to binary would do. I understand there is a reluctancy to change log structure too much? Do you want me to re-work patch towards that end?

kentonv:
Note that if you change a `Text` field to `Data`, you'll find all the messages written before the change have an extra zero byte on the end when parsed after the change, because `Text` always adds a NUL terminator. If you can deal with that, then the change is otherwise backwards-compatible.

pd0wm:
I think having the union adds complexity in both writing and parsing code, and I'm also pretty sure it's not backwards compatible. We always try to avoid making backwards compatible changes, since we do a lot of processing on older logs.

@kentonv The extra null terminator is acceptable. Thanks for letting us know. Also thanks for building capnproto! As you can see we use it for all our inter process communication in openpilot, and we've never had any issues running it in production on a pretty large scale for both real-time communication and off-line processing.

kentonv:
> I think having the union adds complexity in both writing and parsing code, and I'm also pretty sure it's not backwards compatible.

FWIW, retroactive unionization _is_ backwards-compatible (new programs can read old messages), although not forwards-compatible (old programs will get confused by new messages that use the new union member; they won't know to ignore it, because they don't know there's a union tag specifying which field is set). This is approximately equivalent to the compatibility story of changing `Text` to `Data` (old programs won't be able to read messages created by new programs that are missing the NUL terminator). But the union definitely creates annoying code complexity, so if dealing with the trailing NUL byte is easier, it makes sense to go that way.

sobomax:
> > I think having the union adds complexity in both writing and parsing code, and I'm also pretty sure it's not backwards compatible.
> 
> FWIW, retroactive unionization _is_ backwards-compatible (new programs can read old messages), although not forwards-compatible (old programs will get confused by new messages that use the new union member; they won't know to ignore it, because they don't know there's a union tag specifying which field is set). This is approximately equivalent to the compatibility story of changing `Text` to `Data` (old programs won't be able to read messages created by new programs that are missing the NUL terminator). But the union definitely creates annoying code complexity, so if dealing with the trailing NUL byte is easier, it makes sense to go that way.

@kentonv that was definitely something I've seen here. Trying to parse new "unionized" log using old plain schema ended up in exception saying something like "structure is found where Text is expected". This is I think unavoidable given union not being first-class citizen. :( Technically speaking the parser could have figured it out if it were not the case IMHO. 

Consider this hypothetical change:

```
+union PValue {
+  txt @0 :Text;
+  bin @1 :Data;
+}
+
struct InitData {
  dirty @9 :Bool;
  passive @12 :Bool;
-  param @17 :Text;
+  param @17 :PValue;

  enum DeviceType {
    unknown @0;
```

Even given the old schema and new data in this case parser could have figured that anyone accessing InitData.param should be really looking at the InitData.param.txt. Or could it?

sobomax:
> Oh, sorry, that's not the right way to declare a retroactive union. You should do it like this:
> 
> ```diff
>  struct InitData {
>    dirty @9 :Bool;
>    passive @12 :Bool;
> -  param @17 :Text;
> +  param :union {
> +    text @17 :Text;    # (use the old field number of `param`)
> +    data @18 :Data;    # (use the next available field number)
> +  }
> 
>    enum DeviceType {
>      unknown @0;
> ```
> 
> If you declare the `union` as a separate type declaration (as in your example), that's equivalent to declaring a new `struct` that contains just a single `union`. Replacing `Text` with a struct type is definitely not backwards-compatible. It's generally better to declare unions inline, as this allows the union tag and contents to be laid out efficiently within the parent struct, avoids the need for a pointer indirection, and gives you backwards-compatibility when unionizing retroactively.

Thanks, @kentonv for a detailed explanation! However, keep in mind that I've dumbed down the change in question for illustration purposes, actual schema in this case is:
```
struct Map(Key, Value) {
  entries @0 :List(Entry);
  struct Entry {
    key @0 :Key;
    value @1 :Value;
  }
}

struct InitData {
  kernelArgs @0 :List(Text);
[...]
  params @17 :Map(Text, Text);
```
I am not 100% sure how one would use anonymous union in this scenario (any hints?), which is why I opted for embedding it into struct. Keep in mind we have other instances of Map used where Text/Text is totally fine:

```
  androidProperties @16 :Map(Text, Text);
```

sobomax:
@kentonv you were right wrt old programs not being able to read new data correctly, but that's exactly what @pd0wm is trying to prevent here. :( Would it help perhaps if the new code would add extra `\0' when pushing "text" into "Data"?

New logs, new schema:
```
        ( key = "OpenpilotEnabledToggle",
          value = "1" ),
        ( key = "PandaDongleId",
          value = "d9db085b0fa4aa55" ),
        ( key = "PandaFirmware",
          value = "\021\001[\026|GSe\325\032+\224\377\\f|\000\241\321\366j\360\247\354E?:\274\205\301\035\241\241T\370X\224\031\367P\257\350*eY\223rt\246\204=\223\301u\207\277\027xE\tn\365?J\234$\032\375\226\300a\215MB\326\342 }\r\366]5\250\236\006\312\000\025\353>Y\v+U\340C\321\017M\231\0307\220\236\303\005\327\'gJ\302Kv\270V\301F8\246\030\206[\344\2239T\003\275" ),
        ( key = "PandaFirmwareHex",
          value = "11015b167c475365" ),
```
Old schema, new logs:
```
  File "./tools/lib/logreader.py", line 147, in <module>
    print(msg)
  File "capnp/lib/capnp.pyx", line 1092, in capnp.lib.capnp._DynamicStructReader.__str__
capnp.lib.capnp.KjException: capnp/layout.c++:2445: failed: expected cptr[unboundedSize] == '\0'; Message contains text that is not NUL-terminated.
stack: 7f8a9b922a97 7f8a9b92b9ea 7f8a9b92b374 7f8a9b92ba19 7f8a9b92ba19 7f8a9b92bf4e 7f8a9b92c670 7f8a9b7f3d2f 53b1b2 5d66de 5d6027 511da4 50944d 507964 5d3086 5f1bb4 5ecfd2 5f5baa 5f5643 5ea3d8 5c802c 7f8a9cc5fcb1 5c7f2d
```

kentonv:
Ahh... Yes, the `Map` complicates things. Assuming you don't want to modify the definition of `Map`, then the only reasonable thing you could do with a union here is something like:

```capnp
params :union {
  text @17 :Map(Text, Text);
  data @18 :Map(Text, Data);
}
```

But that's... probably not what anyone wants. I think simply changing the map to `Map(Text, Data)` seems like the best bet.

> Would it help perhaps if the new code would add extra `\0' when pushing genuine "text" into "Data"?

Yes, if you manually append a NUL byte to every data value, then old code that expects `Text` should be able to read it.

pd0wm:
Regarding backwards compatibility. We only care about new code being able to read old logs. Our processing code can always use the latest schema definition. So we don't need to insert null bytes.

Thanks for looking into this so extensively. I guess the only change needed is a one line change to schema then. @sobomax do you want to make that PR?

sobomax:
> Regarding backwards compatibility. We only care about new code being able to read old logs. Our processing code can always use the latest schema definition. So we don't need to insert null bytes.
> 
> Thanks for looking into this so extensively. I guess the only change needed is a one line change to schema then. @sobomax do you want to make that PR?

Thanks @pd0wm for a constructive review and communication. I've updated both PRs to do as discussed.

I really liked the product you've built so far, so it's a great pleasure to make it a tiny bit better. I'm also pretty excited to learn @kentonv's capnproto, I'm definitely looking forward to using it more in some of my own real-time projects!

pd0wm:
I'll need to make a PR in our backend code to strip the extra null byte, then I'll merge all PRs

pd0wm:
Merged: https://github.com/commaai/openpilot/commits/master

PR #19695, opened 2021-01-08
loggerd: fix use after free and memory leaks

adeebshihadeh:


selfdrive/loggerd/encoder.cc
```diff
@@ -282,26 +277,26 @@ void encoder_init(EncoderState *s, const char* filename, int width, int height,
...
-      OMX_INDEXTYPE index_type = OMX_IndexParamVideoProfileLevelCurrent;
-    #endif
-      hecv_type.nSize = sizeof(hecv_type);
-      hecv_type.nPortIndex = (OMX_U32) PORT_INDEX_OUT;
-      err = OMX_GetParameter(s->handle, index_type,
-                             (OMX_PTR) &hecv_type);
-      assert(err == OMX_ErrorNone);
+    // setup HEVC
+  #ifndef QCOM2
+    OMX_VIDEO_PARAM_HEVCTYPE hecv_type = {0};
```
ZwX1616, line 282: this is a typo of hevc_type. can you rename it?

PR #19698, opened 2021-01-08
Dockerfile.sim: start from nvidia docker container

pd0wm:
TODO: make onnxruntime_gpu use cuda 11.1


gregjhogan:
It isn't difficult to install CUDA/CUDNN if you want to keep openpilot as base image
https://gitlab.com/nvidia/container-images/cuda/-/blob/master/dist/11.1/ubuntu20.04-x86_64/base/Dockerfile
https://gitlab.com/nvidia/container-images/cuda/-/blob/master/dist/11.1/ubuntu20.04-x86_64/runtime/Dockerfile

PR #19700, opened 2021-01-08
Thneed load/save

geohot:
Will be able to compile the dlc to thneed files.

This halves the size of the model in release. This also starts modeld faster.

### DIFF
--- a/selfdrive/modeld/SConscript
+++ b/selfdrive/modeld/SConscript
@@ -17,7 +17,7 @@ thneed_src = [
 ]
 
 if arch == "aarch64" or arch == "larch64":
-  libs += ['gsl', 'CB', 'gnustl_shared']
+  libs += ['gsl', 'CB']
   libs += ['gnustl_shared'] if arch == "aarch64" else ['pthread', 'dl']
 
   common_src += thneed_src
--- a/selfdrive/modeld/SConscript
+++ b/selfdrive/modeld/SConscript
@@ -12,7 +12,7 @@ common_src = [
 
 thneed_src = [
   "thneed/thneed.cc",
-  "thneed/serialize.cc"
+  "thneed/serialize.cc",
   "runners/thneedmodel.cc",
 ]
 
@@ -47,6 +47,12 @@ else:
     del common_src[common_src.index('runners/snpemodel.cc')]
 
 common = lenv.Object(common_src)
+
+if arch == "aarch64" or arch == "larch64":
+  lenv.Program('thneed/compile', [
+      "thneed/compile.cc"
+    ]+common, LIBS=libs)
+  # LD_LIBRARY_PATH="/data/pythonpath/phonelibs/snpe/aarch64:$LD_LIBRARY_PATH" selfdrive/modeld/thneed/compile
 
 lenv.Program('_dmonitoringmodeld', [
     "dmonitoringmodeld.cc",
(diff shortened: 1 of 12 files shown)


geohot:
Hmm, the model replay test is not expected to fail.

geohot:
@adeebshihadeh Can you look into why tests are failing here? I expect the model to be exactly identical.

Could it be a race now that modeld starts up a lot faster?

adeebshihadeh:
Test passes locally, but I can't reach jenkins remotely for some reason. Will check when in the office.

geohot:
I'd been driving around with this, good to merge?

selfdrive/modeld/SConscript
```diff
@@ -42,6 +48,14 @@ else:
 
 common = lenv.Object(common_src)
 
+# build thneed model
+if arch == "aarch64" or arch == "larch64":
+  compiler = lenv.Program('thneed/compile', ["thneed/compile.cc" ]+common, LIBS=libs)
```
adeebshihadeh, line 53: this common shadows an imported common

selfdrive/modeld/runners/thneedmodel.h
```diff
@@ -0,0 +1,33 @@
+#ifndef THNEEDMODEL_H
+#define THNEEDMODEL_H
+
+#include "runmodel.h"
+
+#define USE_CPU_RUNTIME 0
```
adeebshihadeh, line 6: these are also defined in snpemodel.h

selfdrive/modeld/runners/thneedmodel.h
```diff
@@ -0,0 +1,33 @@
+#ifndef THNEEDMODEL_H
+#define THNEEDMODEL_H
```
adeebshihadeh, line 2: we've been replacing this style header guard with `#pragma once` across the codebase

adeebshihadeh:
Still not sure why the outputs are different. Faster startup shouldn't matter since we only send each frame once we've got a model packet back. I generated new refs and ran the test a few times, but the output isn't deterministic.

selfdrive/modeld/thneed/compile.cc
```diff
@@ -0,0 +1,34 @@
...
+#include "../runners/snpemodel.h"
+
+#define TEMPORAL_SIZE 512
+#define DESIRE_LEN 8
+#define TRAFFIC_CONVENTION_LEN 2
+
+// TODO: This should probably use SNPE directly.
+int main(int argc, char* argv[]) {
+  #define OUTPUT_SIZE 0x10000
+  float *output = (float*)calloc(OUTPUT_SIZE, sizeof(float));
```
adeebshihadeh, line 12: do we need to add TEMPORAL_SIZE here?

adeebshihadeh: oh, we're using a separate buffer but it's part of the output buffer in driving.cc

geohot:
Can you check for nondeterminism again? There was a missing memset 0 for the recurrent state, maybe that was it?

geohot:
@adeebshihadeh model replay test is passing, just "openpilot tests / longitudinal (push)" isn't

selfdrive/modeld/thneed/compile.cc
geohot: nah, this is a junk buffer

adeebshihadeh:
If you rebase the push and pull request tests should match.

geohot:
Done

PR #19705, opened 2021-01-09
ui.cc: refactor update_model

deanlee:
Simplify update_model. we only need to limit the length of the path when there is a lead car.

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -112,6 +112,7 @@ static void update_model(UIState *s, const cereal::ModelDataV2::Reader &model) {
   // update path
   int max_idx =  TRAJECTORY_SIZE - 1;
   if (scene.lead_data[0].getStatus()) {
+    // limit the max_idx if lead car detected
     const float lead_d = scene.lead_data[0].getDRel() * 2.;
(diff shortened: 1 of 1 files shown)


selfdrive/ui/ui.cc
```diff
@@ -101,28 +108,30 @@ static void update_line_data(const UIState *s, const cereal::ModelDataV2::XYZTDa
...
   const auto road_edge_stds = model.getRoadEdgeStds();
   for (int i = 0; i < std::size(scene.road_edge_vertices); i++) {
     scene.road_edge_stds[i] = road_edge_stds[i];
-    update_line_data(s, road_edges[i], 0.025, 1.22, &scene.road_edge_vertices[i], max_distance);
+    update_line_data(s, road_edges[i], 0.025, 1.22, &scene.road_edge_vertices[i]);
   }
 
   // update path
-  const float lead_d = scene.lead_data[0].getStatus() ? scene.lead_data[0].getDRel() * 2. : MAX_DRAW_DISTANCE;
-  float path_length = (lead_d > 0.) ? lead_d - fmin(lead_d * 0.35, 10.) : MAX_DRAW_DISTANCE;
```
deanlee, line 123: By the way, I don't know why path_length is calculated like this. looks a bit complicated?

PR #19708, opened 2021-01-10
A minor update

ghost:
Hello,
There is still alot of more work to be done, but will commit to this project once i have more free time.
Also where are we installing two things that do the same thing??
`wget` and `curl` pretty much do the same thing and one is enough, so there we are installing both of them??
- Prajwal

### DIFF
--- a/update_requirements.sh
+++ b/update_requirements.sh
@@ -17,7 +17,6 @@ function dist-check() {
         # shellcheck disable=SC1091
         source /etc/os-release
         DISTRO=$ID
-        DISTRO_VERSION=$VERSION_ID
     fi
 }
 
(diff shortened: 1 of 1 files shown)


pd0wm:
What exactly is the purpose of this PR? We don't want to official support any other operating system than 20.04 since we don't have the capacity to make sure everything keeps working.

Prajwal-Koirala:
I'll add a if statement to allow 20.04 only.

The purpose of this is to inform the users during installation what's allowed and what is not.

adeebshihadeh:
We're certainly interested in making it easier to get setup with openpilot, but I don't think this really improves things. If you're looking to improve things, take a look at the setup scripts and README in tools/. Also feel free to start a GitHub Discussion to discuss improving the getting started experience.

PR #19711, opened 2021-01-10
Add merge chance badge to README.md

brunotoic:
Awesome stream!


brunotoic:
I have just realized that I messed up the badge name, should be 'Merge chance' or something similar, but I left it as 'Custom badge'. I am now considering moving to the woods and living my life in solitude with my eternal shame as my only companion.

The words of a famous poet sum up my doings quite well:
_I just came to f*** a b****
  and skip over the dinner date._

I was trying to be first so bad, I messed up. But since the PR after mine fixes my mistake, you could make the day of two people instead of one and keep your Merge chance high.

The streams are really dope, keep 'em coming, we'll keep watching. 
Thanks, bye!

pd0wm:
See https://github.com/commaai/openpilot/pull/19713#issuecomment-757874285

PR #19715, opened 2021-01-11
ui/android/text: convert text.c to text.cc

deanlee:


adeebshihadeh:
Unless there's a reason for this, I don't want to merge changes to any of the android UI stuff as we'll be moving to Qt soon.

deanlee:
just want to convert touch.c into class TouchState. If it is not necessary, I will close these two PRs

adeebshihadeh:
Once we’re using Qt, we shouldn’t need to handle touch events ourselves.

deanlee:
OIC

PR #19720, opened 2021-01-11
convert glutil.c to gltuil.cc & add class GLShader

deanlee:
I used `git mv` to rename file. but it differs too much,the difftools consider it as delete/new


selfdrive/common/glutil.cc
```diff
@@ -0,0 +1,57 @@
...
+  if (!status) {
+    glGetProgramiv(prog, GL_INFO_LOG_LENGTH, &len);
+    if (len) {
+      std::string msg(len, '\0');
+      glGetProgramInfoLog(prog, len, NULL, msg.data());
+      fprintf(stderr, "error linking program:\n%s\n", msg.c_str());
+    }
+    assert(0);
+  }
+  texture_loc = glGetUniformLocation(prog, "uTexture");
```
pd0wm, line 49: Think these locations/springs are too specific. Maybe you can add a helper function to the GLShader class to get locations, and have the user of the class call that. Cache them in an std::map to prevent stalls.

PR #19732, opened 2021-01-12
util.h: extern C for write_file

deanlee:
write_file is still called by C function such as set_brightness in framebuffer.cc(which is used by text.c too)


pd0wm:
Isn't it better to just move those .c sources over to .cc?

deanlee:
Yes it makes more sense,I'll do it in https://github.com/commaai/openpilot/pull/19734

PR #19736, opened 2021-01-12
fill_frame_image: reduce function parameters

deanlee:


pd0wm:
Maybe add an assert for `b->rgb` to make sure it's not a YUV buffer?

PR #19746, opened 2021-01-13
added Hyundai Santafe limited 2020 FP in hyundai values and updated description for SANTA_FE

newstackdevelopment:


selfdrive/car/hyundai/values.py
```diff
@@ -28,7 +28,7 @@ class CAR:
   IONIQ_EV_2020 = "HYUNDAI IONIQ ELECTRIC 2020"
   KONA = "HYUNDAI KONA 2020"
   KONA_EV = "HYUNDAI KONA ELECTRIC 2019"
-  SANTA_FE = "HYUNDAI SANTA FE LIMITED 2019"
+  SANTA_FE = "HYUNDAI SANTA FE LIMITED 2019-2020"
```
adeebshihadeh, line 31: revert this, no need to update it

newstackdevelopment: has been reverted back to only 2019

newstackdevelopment (commented):
reverted the description change back to only 2019

adeebshihadeh:
What's your Dongle ID?

VirtuallyChris:
Which harness type does this car use? Still Hyundai D?

newstackdevelopment:
Yes still D but I pushed it to my comma and am getting a unable to import error

newstackdevelopment:
My error was not related to the pull request. I pushed this branch into my device and this branch is missint a steeringlimits class and it was replace with carparams class. So my pull request is still good to go

selfdrive/car/hyundai/values.py
newstackdevelopment: Any news on this. It has fixed my issue locally. Am I missing anything else to get this approved?

PR #19748, opened 2021-01-13
Toyota: Only use gas interceptor under 19 mph

sshane:
Above 19 mph, the min ACC speed for most older Corollas, we switch from sending a gas command (which isn't what the longcontrol loop is set up to do) to sending acceleration, just like if we didn't have a comma pedal installed.

This fixes:

- oscillations that occurred above 19 mph due to the longcontrol loop being set up to control acceleration, but was actually being sent as a gas percentage
- late braking when lead slows down quickly, or with large speed differences (sometimes causing fcw). previously different pedal tuning was used which affected braking when it shouldn't have.

Also declared `stop_and_go` to start as true which removes some duplicate lines in interface

What still needs to be done:

- [x] when engaging at a stop with no lead (or even with a lead), gas is jerky and uncomfortable.
  - I used to think that it was hitting the planner acceleration limit, but now I think it's caused by the controller being set up to output acceleration instead of gas. For example, we start by commanding some small amount of acceleration from a stop, which a sng car would reach quite quickly and smoothly, but since it's actually a gas value, the output isn't enough to reach the speed/accel quick enough. Then either integral or the MPC commands a burst of "acceleration" to get us up to speed causing overshoot and then braking. I still need to download the logs and look at the data to know for sure, since you can't see any of that in Cabana

This is what accelerating from a stop with no lead looks like: https://i.imgur.com/LAxVm5h.png

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -17,12 +17,12 @@ def compute_gb_gas_interceptor(accel, speed):
   # This converts desired positive acceleration at a speed to gas percentage
   # It's only accurate up to MIN_ACC_SPEED (19 mph) for now since the function was fitted on data up to that speed
   # Once we reach that speed, we switch to sending acceleration anyway so this isn't a problem
-  def accel_to_gas(_speed, _accel):  # converts accel to gas using current speed
-    _c1, _c2, _c3, _c4 = [0.04412016647510183, 0.018224465923095633, 0.09983653162564889, 0.08837909527049172]
-    return (_accel * _c1 + (_c4 * (_speed * _c2 + 1))) * (_speed * _c3 + 1)
-
   if accel > -0.1 and speed <= MIN_ACC_SPEED:  # todo: -0.1 is smooth but in data user was giving gas when a_ego was down to -0.5
-    return accel_to_gas(speed, accel)
+    _c1, _c2, _c3, _c4 = [0.04412016647510183, 0.018224465923095633, 0.09983653162564889, 0.08837909527049172]
+    # fixme: this function is arbitrary, plot how accel and speed relate to gas and then make a cleaner function
+    # my guess is speed isn't linear but accel is?
+    return (accel * _c1 + (_c4 * (speed * _c2 + 1))) * (speed * _c3 + 1)
+
   return float(accel) / CarControllerParams.ACCEL_SCALE
 
 
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -320,10 +320,8 @@ class CarInterface(CarInterfaceBase):
     ret.communityFeature = ret.enableGasInterceptor or ret.enableDsu or smartDsu
 
     if ret.enableGasInterceptor:
-      ret.gasMaxBP = [0., 9., 35]
-      ret.gasMaxV = [0.2, 0.5, 0.7]
-      ret.longitudinalTuning.kpV = [1.2, 0.8, 0.5]
-      ret.longitudinalTuning.kiV = [0.18, 0.12]
+      ret.gasMaxBP = [0., MIN_ACC_SPEED]
+      ret.gasMaxV = [0.2, 0.5]
 
     return ret
 
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -23,7 +23,6 @@ class CarInterface(CarInterfaceBase):
 
     ret.steerActuatorDelay = 0.12  # Default delay, Prius has larger delay
     ret.steerLimitTimer = 0.4
-    stop_and_go = True
 
     # Default longitudinal tune
     ret.longitudinalTuning.deadzoneBP = [0., 9.]
@@ -32,6 +31,7 @@ class CarInterface(CarInterfaceBase):
     ret.longitudinalTuning.kiBP = [0., 35.]
     ret.longitudinalTuning.kpV = [3.6, 2.4, 1.5]
     ret.longitudinalTuning.kiV = [0.54, 0.36]
+    stop_and_go = True
 
     if candidate not in [CAR.PRIUS, CAR.RAV4, CAR.RAV4H]:  # These cars use LQR/INDI
       ret.lateralTuning.init('pid')
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python3
 from cereal import car
 from selfdrive.config import Conversions as CV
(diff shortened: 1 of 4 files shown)


selfdrive/car/toyota/interface.py
```diff
@@ -313,17 +296,15 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
...
     # intercepting the DSU is a community feature since it requires unofficial hardware
     ret.communityFeature = ret.enableGasInterceptor or ret.enableDsu or smartDsu
 
     if ret.enableGasInterceptor:
-      ret.gasMaxBP = [0., 9., 35]
-      ret.gasMaxV = [0.2, 0.5, 0.7]
-      ret.longitudinalTuning.kpV = [1.2, 0.8, 0.5]
-      ret.longitudinalTuning.kiV = [0.18, 0.12]
+      ret.gasMaxBP = [0., MIN_ACC_SPEED]
+      ret.gasMaxV = [0.2, 0.5]
```
ShaneSmiskol, line 307: For now though, we can probably lower these gas max values, still need to tune them

ShaneSmiskol:
Yep, I was wrong initially when I thought it was hitting the planner acceleration limit! It's commanding a constant speed and acceleration with no oscillation: https://imgur.com/a/4zdmih4

Meanwhile here's proportional and integral over the same period, prop seems to be causing most of the jerking: https://i.imgur.com/j4O2n6h.png

VirtuallyChris:
I think it would be great if you can fix pedal jerk universally for both Honda and Toyota in general instead of having it switch to accel above 19mph on Toyotas.

On Nidec Hondas, it's preferable to use the pedal for accel at all speeds due to it being direct control versus sending the PCM a set speed and desired accel (causes lots of jerk/inaccuracies!) I see how for Toyota the easier thing would be to stop using pedal above a certain speed since the Toyota accel command is nice when it can be used, but if we want one universal pedal fix; getting rid of the oscillation in the controller seems like a better thing to fix to me.

ShaneSmiskol:
With the Honda Nidec, I see in carcontroller it says it's sending a brake command, is that actually the brake in terms of 0 to 1 being 0 to 100% brake, or is that desired deceleration? Also without a pedal, Honda Nidec can brake but not accelerate? I don't know too much about how the Honda systems work or their intricacies yet.

In order to have openpilot produce a gas value instead of an acceleration, I think it would require mimicking how the Toyota's ACC system works where you give it a desired acceleration (negative or positive) and there's another control mechanism that keeps the car at or matches that acceleration. But since the longcontrol pif loop is set up to output acceleration instead of gas and brake it inherently introduces oscillation as there's no feedforward component to constantly apply gas or brake to keep us at the desired speed.

Immediately I'm thinking we'd need a second controller that takes the acceleration from longcontrol and converts it into gas and brakes, with some sort of function that can produce a gas value given acceleration, speed, and maybe incline of the road. So for ex. if des. acceleration is 0, speed is 20 mph, inclination is 0, the output would need to be some constant gas value instead of 0.

But then since Toyota can't send brake directly (as far as I can see from toyota/carcontroller) and only deceleration, it would need to be able to switch seamlessly from using the secondary gas loop to the original longcontrol loop.

This is all just brainstorming, I first just want to understand how the Toyota system works with acceleration, hills, wind drag, etc. so I'll go and try to test some things when I get some time here

VirtuallyChris:
Yes, we have control of braking as deceleration on Nidec, but not positive accel. pedal works well at highway speeds but has crazy oscillation at low speeds probably the same as Toyota due to the reasons you mentioned above.

ShaneSmiskol:
I imagine you have the slow oscillation at high speeds as well due to no feedforward adding constant gas to keep those speeds. So it's the same as Toyota? Acceleration is gas commands but brake is (de)acceleration commands? That at least unifies the problem a bit.

The tricky part is switching between those two inherently different types of control. For example, acceleration output from longcontrol could be positive, but we might want to brake in certain situations, and vice versa.

ShaneSmiskol:
Can we merge this PR to get rid of the slow oscillation at high speeds for pedal users in the next release? Then I can take a look at outputting gas values later in another PR. I almost think it would need to be either model-specific, or make-specific in the least, since we're converting from acceleration to a raw gas percentage applied to the pedal. Previously we relied on the car to come up with this, so I feel like each car or make would need slightly different tuning

ShaneSmiskol:
I fit a simple function based on me manually driving around disengaged, I made sure the user's gas is the same value as what it shows when openpilot is in control. Anyway, I need to correct for the delay from pressing gas to reaching the correct acceleration, like an actuator delay, but this is some comparison data: https://imgur.com/a/CFqRlI0

It didn't make sense to just transform feedforward, as I realized kp and ki are tuned for acceleration in mind as well, so once I put the entirety of the PIF loop into the accel to gas function, sng is much better than before.

selfdrive/car/toyota/interface.py
```diff
@@ -9,10 +9,34 @@
...
+
+def compute_gb_gas_interceptor(accel, speed):
+  # This converts desired positive acceleration at a speed to gas percentage
+  # It's only accurate up to MIN_ACC_SPEED (19 mph) for now since the function was fitted on data up to that speed
+  # Once we reach that speed, we switch to sending acceleration anyway so this isn't a problem
+  def accel_to_gas(_speed, _accel):  # converts accel to gas using current speed
+    _c1, _c2, _c3, _c4 = [0.04412016647510183, 0.018224465923095633, 0.09983653162564889, 0.08837909527049172]
+    return (_accel * _c1 + (_c4 * (_speed * _c2 + 1))) * (_speed * _c3 + 1)
+
+  if accel > -0.1 and speed <= MIN_ACC_SPEED:  # todo: -0.1 is smooth but in data user was giving gas when a_ego was down to -0.5
```
ShaneSmiskol, line 24: Also I'm not sure what the best way to handle the transition would be. If it was `accel > 0`, 0 accel with the function is some amount of gas, so we don't want a quick hard switch if accel is hovering around a small negative value to 0.

ShaneSmiskol: Todo: Test what happens if you request a very small amount of negative acceleration while going up a hill under 19 mph. (Ideally) Does the car apply gas to keep us from undershooting the desired acceleration even though we can't apply positive acceleration at that speed? If so, then we can just do accel > 0

ShaneSmiskol: I'm going to try to use a function I fit on data that gives us the deceleration of the car given a speed under 19 mph, then we check if accel is greater than that to use the converted gas, else use the regular compute_gb function of dividing it by 3.

ShaneSmiskol:
Some of the stuff I'm doing here is wrong (like converting accel to gas percentage before it gets clipped in longcontrol assuming it's accel rather than gas), I just took a break from working on this. I'll fix and clean it up soon once I get a clean way of converting accel to gas that works for most cars with pedal.

PR #19762, opened 2021-01-15
UI: handle alert in function update_alert

deanlee:


pd0wm:
For some reason the camera view doesn't start for me, but don't see an obvious reason. Let's rebase and see if the problem is still there.

pd0wm:
Still doesn't show the camera view

deanlee:
My test is ok, I really don’t understand what’s going on

selfdrive/ui/ui.cc
```diff
@@ -202,6 +178,57 @@ static void update_sockets(UIState *s) {
   s->started = scene.thermal.getStarted() || scene.frontview;
 }
 
+static void update_alert(UIState *s) {
+  if (!s->started || s->scene.frontview) return;
+
+  UIScene &scene = s->scene;
+  const uint64_t cs_frame = s->sm->rcv_frame("controlsState");
+  if (cs_frame == s->sm->frame) {
```
adeebshihadeh, line 186: use SubMaster's updated function

selfdrive/ui/ui.cc
```diff
@@ -202,6 +178,57 @@ static void update_sockets(UIState *s) {
   s->started = scene.thermal.getStarted() || scene.frontview;
 }
 
+static void update_alert(UIState *s) {
+  if (!s->started || s->scene.frontview) return;
```
adeebshihadeh, line 182: why does this check frontview? controlsState shouldn't be updated anyway

adeebshihadeh:
I'm also not seeing the camera view.

selfdrive/ui/ui.cc
deanlee: checking the control times also requires this condition

deanlee:
Can you tell me which script you used to run the camera view?

PR #19764, opened 2021-01-15
CameraBuf::acquire : using a scoped lock

deanlee:


pd0wm:
If you're interested you can also look into replacing our old thread safe queue implementation: https://github.com/commaai/openpilot/blob/master/selfdrive/common/cqueue.c

We probably want to abstract the C++ version from CameraBuf into a class, and then replace the usage of cqueue everywhere.

deanlee:
OK, I will see how to do it

deanlee:
https://github.com/commaai/openpilot/pull/19774

PR #19768, opened 2021-01-15
use std::map for images

deanlee:
There is not much code reduction, but the uistate structure looks clearer and it is convenient to add/remove/free resources in the future

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -56,11 +56,11 @@ bool car_space_to_full_frame(const UIState *s, float in_x, float in_y, float in_
 }
 
 
-static void ui_draw_text(NVGcontext *vg, float x, float y, const char* string, float size, NVGcolor color, int font){
-  nvgFontFaceId(vg, font);
-  nvgFontSize(vg, size);
-  nvgFillColor(vg, color);
-  nvgText(vg, x, y, string, NULL);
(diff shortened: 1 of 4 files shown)


selfdrive/ui/ui.hpp
```diff
@@ -151,18 +154,8 @@ typedef struct UIState {
...
-  int img_wheel;
-  int img_turn;
-  int img_face;
-  int img_button_settings;
-  int img_button_home;
-  int img_battery;
-  int img_battery_charging;
-  int img_network[6];
+  // images
+  std::map<const char *, int, cmp_str> images;
```
adeebshihadeh, line 158: can we use std::string here? then we don't need cmp_str

selfdrive/ui/paint.cc
```diff
@@ -525,36 +525,39 @@ void ui_nvg_init(UIState *s) {
...
+      {"network_1", "../assets/images/network_1.png"},
+      {"network_2", "../assets/images/network_2.png"},
+      {"network_3", "../assets/images/network_3.png"},
+      {"network_4", "../assets/images/network_4.png"},
+      {"network_5", "../assets/images/network_5.png"},
+  };
+  for (auto [name, file] : images) {
+    int img_id = nvgCreateImage(s->vg, file, 1);
+    assert(img_id != 0);
+    s->images[name] = img_id;
```
adeebshihadeh, line 560:
```suggestion
    s->images[name] = nvgCreateImage(s->vg, file, 1);
    assert(s->images[name] != 0);
```

PR #19769, opened 2021-01-15
remove old model

haraschax:


HaraldSchafer:
@adeebshihadeh This works. Not sure how you want to deal with testing. I ignored the controlsd  proc_replay tests to get this to pass.

adeebshihadeh:
I’d like to wait until we get the missing routes. If it takes too long, I’ll write something to use private routes.

PR #19772, opened 2021-01-15
UI: conditional update satelliteCount

deanlee:


pd0wm:
Think this will only lead to confusion if we later use the satellite count outside of the sidebar. Not worth the minimal performance gain.

deanlee:
yes u r right

PR #19774, opened 2021-01-15
remove cqueue, use class SafeQueue

deanlee:


pd0wm:
All these blocking pops make me a little concerned. That's just asking for stuff to block on shutdown. Can we easily turn that into non-blocking pops? Maybe best to do that in a separate PR though.

deanlee:
Yes, other places may have similar problems, need to find a general way to deal with it.  another PR may be better.

pd0wm:
If we check the exit code in pop we still need to trigger the condition variable. Something like a `close` method on SafeQueue that gets called on shutdown and triggers a `cv.notify_all()` so all the encoders can wake up and close.

selfdrive/loggerd/omx_encoder.cc
```diff
@@ -448,7 +445,7 @@ int OmxEncoder::encode_frame(const uint8_t *y_ptr, const uint8_t *u_ptr, const u
   // this sometimes freezes... put it outside the encoder lock so we can still trigger rotates...
   // THIS IS A REALLY BAD IDEA, but apparently the race has to happen 30 times to trigger this
   //pthread_mutex_unlock(&this->lock);
-  OMX_BUFFERHEADERTYPE* in_buf = (OMX_BUFFERHEADERTYPE *)queue_pop(&this->free_in);
+  OMX_BUFFERHEADERTYPE* in_buf = this->free_in.pop();
```
deanlee, line 448: Or add a timeout parameter to try_pop,and call try_pop(timeout) here?

deanlee:
It seems that we have to change the loggerd's code too, maybe another PR is more suitable

selfdrive/loggerd/omx_encoder.cc
pd0wm: Then we still need some early returns, which the caller of encode_frame needs to handle.

pd0wm:
Think a separate PR is most appropriate. Interested in your thoughts.

This PR looks good, but will do some testing on Monday.

deanlee:
replaced blocked `pop` with `try_pop(timeout) `in encoder

selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -121,14 +121,7 @@ CameraBuf::~CameraBuf() {
...
 bool CameraBuf::acquire() {
-  {
-    std::unique_lock<std::mutex> lk(frame_queue_mutex);
-    bool got_frame = frame_queue_cv.wait_for(lk, std::chrono::milliseconds(1), [this]{ return !frame_queue.empty(); });
-    if (!got_frame) return false;
-
-    cur_buf_idx = frame_queue.front();
-    frame_queue.pop();
-  }
+  if (!safe_queue.try_pop(cur_buf_idx, 20)) return false;
```
pd0wm, line 124:
```suggestion
  if (!safe_queue.try_pop(cur_buf_idx, 1)) return false;
```

deanlee:
rebased to resolve conflict

selfdrive/loggerd/loggerd.cc
```diff
@@ -271,11 +271,16 @@ void encoder_thread(int cam_idx) {
...
                                                &out_segment, extra.timestamp_eof);
+        if (out_id == -1) {
+          continue;
+        }
         if (encoders.size() > 1) {
           int out_segment_alt = -1;
-          encoders[1]->encode_frame(buf->y, buf->u, buf->v,
-                                    buf->width, buf->height,
-                                    &out_segment_alt, extra.timestamp_eof);
+          if (-1 == encoders[1]->encode_frame(buf->y, buf->u, buf->v,
```
pd0wm, line 279: Maybe we should ignore a failure of the alternate encoders? Now we can successfully encode the normal HEVC, but not send out a corresponding encodeIdx message.

deanlee: Maybe a for loop is better. can solve these two problems

deanlee: If you think this is not good, I will revert back

selfdrive/loggerd/loggerd.cc
```diff
@@ -266,38 +266,34 @@ void encoder_thread(int cam_idx) {
...
-        eidx.setType(cam_idx == LOG_CAMERA_ID_DCAMERA ? cereal::EncodeIndex::Type::FRONT : cereal::EncodeIndex::Type::FULL_H_E_V_C);
-  #endif
-        eidx.setEncodeId(cnt);
-        eidx.setSegmentNum(out_segment);
-        eidx.setSegmentId(out_id);
-
-        if (lh) {
-          auto bytes = msg.toBytes();
-          lh_log(lh, bytes.begin(), bytes.size(), false);
+        if (!out_id) break;
```
pd0wm, line 274: Why break on failure? Why not replace the if condition by `if (i == 0 & out_id != 1)`

As a side note `-1` evaluates as true (since it's non zero). I've been bit by that in the past.

deanlee: My bad,I was too hasty.

PR #19784, opened 2021-01-16
OmxEncoder: use std::recursive_mutex

deanlee:


deanlee:
I think we don't need this mutex anymore : https://github.com/commaai/openpilot/pull/19948

PR #19786, opened 2021-01-16
OmxEncoder: use c++ mutex&condition_variable

deanlee:


adeebshihadeh:
Nice cleanup, and thanks for breaking up these changes into multiple PRs :)

selfdrive/loggerd/omx_encoder.h
```diff
@@ -3,7 +3,8 @@
 #include <stdio.h>
 #include <stdint.h>
 #include <stdbool.h>
-
+#include <mutex>
+#include <condition_variable>
```
adeebshihadeh, line 7: cleanup the includes we don't need anymore

selfdrive/loggerd/omx_encoder.cc
```diff
@@ -34,11 +34,8 @@ extern ExitHandler do_exit;
 // ***** OMX callback functions *****
 
 void OmxEncoder::wait_for_state(OMX_STATETYPE state) {
-  pthread_mutex_lock(&this->state_lock);
-  while (this->state != state) {
-    pthread_cond_wait(&this->state_cv, &this->state_lock);
-  }
-  pthread_mutex_unlock(&this->state_lock);
+  std::unique_lock lk(this->state_lock);
+  this->state_cv.wait(lk, [&]{return this->state == state;});
```
adeebshihadeh, line 38: this is cute, but I think the while loop is cleaner

PR #19744, opened 2021-01-13
hyundai: increase steering angle limit for custom eps fw

gregjhogan:
requires https://github.com/commaai/cereal/pull/110

PR #19789, opened 2021-01-16
hyundai: add fpv2 for palisade with power steering fw mod

gregjhogan:
forgot to add this in https://github.com/commaai/openpilot/pull/19744

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -275,7 +275,10 @@ FW_VERSIONS = {
       b'\xf1\x81640J0051\x00\x00\x00\x00\x00\x00\x00\x00',
       b'\xf1\x81640K0051\x00\x00\x00\x00\x00\x00\x00\x00',
     ],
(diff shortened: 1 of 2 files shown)


selfdrive/car/hyundai/values.py
```diff
@@ -275,7 +275,10 @@ class Buttons:
       b'\xf1\x81640J0051\x00\x00\x00\x00\x00\x00\x00\x00',
       b'\xf1\x81640K0051\x00\x00\x00\x00\x00\x00\x00\x00',
     ],
-    (Ecu.eps, 0x7d4, None): [b'\xf1\x00LX2 MDPS C 1.00 1.03 56310-S8020 4LXDC103',],
+    (Ecu.eps, 0x7d4, None): [
+      b'\xf1\x00LX2 MDPS C 1,00 1,03 56310-S8020 4LXDC103',
```
adeebshihadeh, line 279: can we add a comment next to it?

gregjhogan: before I opened this PR, I looked at honda to see if there was anything like that, and there is not, I will add it to both I guess

PR #19794, opened 2021-01-17
loggerd: use unique_ptr

deanlee:


deanlee:
merged into another PR: https://github.com/commaai/openpilot/pull/19802

PR #19797, opened 2021-01-17
loggerd: Refactor QlogState

deanlee:
- rename to SocketState
- add member variable fpkt_id
- get it out of the while loop
- remove std::vector<SubSocket*> socks;


deanlee:
merged into another PR: https://github.com/commaai/openpilot/pull/19802

PR #19800, opened 2021-01-17
convert framebuffer to class 

deanlee:


adeebshihadeh:
LGTM once the bootlog binary checked in is removed

pd0wm:
@deanlee Can you remove that binary?

deanlee:
done

PR #19802, opened 2021-01-17
refactor loggerd

deanlee:
- [x] convert struct LoggerdState to class
- [x] split main into  functions to reduces the overall complexity


deanlee:
split the main function into  member functions of LoggerdState. the logic is exactly the same as before.

deanlee:
reverted "fix msgs from the first poll getting dropped".keep the functions consistent with the original version.

adeebshihadeh:
I agree loggerd is in need of some refactoring, but let's keep it small. I think a good first step is to break bootlog off from loggerd.

deanlee:
I'll split them to small PRS

PR #19804, opened 2021-01-17
prevent openpilot from engaging with no lead at zero speed on toyotas

### LINKED ISSUE #19801: [Toyota car bug] openpilot can be engaged at zero speed with no lead
**Describe the bug**

openpilot can be engaged at zero speed with no lead by releasing the brakes and pressing the cruise control stalk.

1. stop on hill without lead
2. release brakes (car will not move forward as it's on a hill, so zero speed)
3. engage adaptive cruise
4. openpilot engages and immediately goes into resume mode (with comma pedal it presses on the pedal and the car goes out of standstill)

**Expected behavior**

openpilot should display "no close lead" and force acc to disengage

 - Device: comma two
 - Dongle ID: 399b1cca765e80b4
 - Version: all versions seem to be affected
 - Car make/model: Toyota Prius 2018

**Additional context**

This has been a problem since I firstly started to use openpilot

cydia2020:
**Description**
fixes #19801

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -107,7 +107,7 @@ class CarController():
 
     # we can spam can to cancel the system even if we are using lat only control
     if (frame % 3 == 0 and CS.CP.openpilotLongitudinalControl) or (pcm_cancel_cmd and Ecu.fwdCamera in self.fake_ecus):
-      lead = lead or CS.out.vEgo < 12.    # at low speed we always assume the lead is present do ACC can be engaged
+      lead = lead or (CS.out.vEgo < 12. and CS.out.vEgo > 0.001)    # at low speed we always assume the lead is present so ACC can be engaged
 
       # Lexus IS uses a different cancellation message
       if pcm_cancel_cmd and CS.CP.carFingerprint == CAR.LEXUS_IS:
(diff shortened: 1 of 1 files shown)


selfdrive/car/toyota/carcontroller.py
```diff
@@ -107,7 +107,7 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, hud_alert,
 
     # we can spam can to cancel the system even if we are using lat only control
     if (frame % 3 == 0 and CS.CP.openpilotLongitudinalControl) or (pcm_cancel_cmd and Ecu.fwdCamera in self.fake_ecus):
-      lead = lead or CS.out.vEgo < 12.    # at low speed we always assume the lead is present do ACC can be engaged
+      lead = lead or (CS.out.vEgo < 12. and CS.out.vEgo > 0.001)    # at low speed we always assume the lead is present so ACC can be engaged
```
pd0wm, line 110:
```suggestion
      lead = lead or (CS.out.vEgo < 12. and (not CS.out.standstill))    # at low speed we always assume the lead is present so ACC can be engaged
```

pd0wm: The standstill field is probably better suited. Does that also work correctly?

cydia2020: One sec, let me test this

cydia2020: This doesn't work, op can still engage at zero speed

pd0wm: Strange. The threshold for standstill is the same (0.001). Only thing is it uses raw wheel speed instead of the Kalman filered one.

Can you make it work with the vEgoRaw, but with a larger threshold? The Kalman filter can behave a little erratic at very low speed causing vEgo to not be correct.

cydia2020: It might be because openpilot mistaking my bushes as a lead, which happened before. I can try again on road this morning

cydia2020: OK this does work.

PR #19817, opened 2021-01-18
swaglog: use std::mutex & remove extern "C"

deanlee:


pd0wm:
Thanks! Another improvement that swaglog could use is a replacement with a C++ class with a proper constructor and destructor. If you're running valgrind on a process using swaglog you'll notice it always leaks memory due to the lack of proper cleanup.

deanlee:
OIC

pd0wm:
Thanks! We already have plenty of PRs to review, so no rush :)

deanlee:
Had a hard weekend LOL :P

PR #19818, opened 2021-01-18
cameras_qcom2: more readable camera_process_frame

deanlee:
Too many ternary, use one "if else" to make it more readable.


ZwX1616:
for all the different packet names, adeeb says we should use the capnp dynamic api to handle this more cleanly
for the send conditions, maybe we should just add env_send_xxx to each camera_state?

PR #19819, opened 2021-01-18
camerad: remove duplicated set_realtime_priority call

deanlee:


selfdrive/camerad/main.cc
```diff
@@ -47,7 +46,6 @@ void party(cl_device_id device_id, cl_context context) {
 #endif
 
 int main(int argc, char *argv[]) {
-  set_realtime_priority(53);
```
adeebshihadeh, line 50: keep this one and remove the one in party

adeebshihadeh:
Done in 5cb6e69a78c9dadf0b12102e049fdb1feefa0295

PR #19820, opened 2021-01-18
Qcom2 camera serial clock change

robbederks:
Fix reverted PR #19494.  We also needed to change the readout timing param, since that was tuned for 20Hz in freerunning mode with the higher serial clock.
Reducing this timing also allowed for some slack in slave mode, since it now gets 24Hz in freerunning mode.


robbederks:
@ZwX1616: it doesn't work with the minimum timing requirements for 2-exposure HDR, so I used the minimum for 3-exposure. Any problems with that?

ZwX1616:
> @ZwX1616: it doesn't work with the minimum timing requirements for 2-exposure HDR, so I used the minimum for 3-exposure. Any problems with that?

Oh I think it's fine. 3 > 2 so the current setting should still work. Will test

PR #19824, opened 2021-01-18
Carla fixes

### LINKED ISSUE #2501: Carla sim "Openpilot unavailable Communication issue between processes"
I have ubuntu 18.04 and have run the two given scripts for running openpilot with Carla. I get a warning saying "Grey Panda is deprecated" followed by  "Openpilot unavailable Communication issue between processes" when I engage by pressing 1 on the terminal.
Operating system: [e.g. Ubuntu 18.04]

marschr:
Added some fixes for known carla issues along with other patches:
- fixed issue with carla window not showing when CARLA is ran from docker
- added WASD keys for gas/brake/steering when running bridge.py
- comm issues on controldsd.py, this perhaps fixes #2501 
- CARLA camera provides a frame number https://carla.readthedocs.io/en/0.9.7/cameras_and_sensors/


pd0wm:
Is it necessary to patch out the communication issue? What packet is it exactly triggering on?

selfdrive/controls/controlsd.py
```diff
@@ -198,14 +198,15 @@ def update_events(self, CS):
...
-    elif not self.sm.all_alive_and_valid():
-      self.events.add(EventName.commIssue)
+    if not SIMULATION:
+      if (self.sm['health'].safetyModel != self.CP.safetyModel and self.sm.frame > 2 / DT_CTRL) or \
+        self.mismatch_counter >= 200:
+        self.events.add(EventName.controlsMismatch)
+      if not self.sm.alive['plan'] and self.sm.alive['pathPlan']:
+        # only plan not being received: radar not communicating
+        self.events.add(EventName.radarCommIssue)
+      elif not self.sm.all_alive_and_valid():
```
marschr, line 208: this `all_alive_and_valid()` call fails with `ubloxRaw` and `frontFrame`, triggering the communication issue. Both are being addressed later on about lines 230~240 to then prevent reading from them when on simulation.

adeebshihadeh: we can conditionally ignore whether frontFrame and ubloxRaw are alive based on the SIMULATION flag. it's a kw arg to SubMaster

selfdrive/controls/controlsd.py
```diff
@@ -55,6 +55,8 @@ def __init__(self, sm=None, pm=None, can_sock=None):
       self.sm = messaging.SubMaster(['thermal', 'health', 'model', 'liveCalibration', 'ubloxRaw',
                                      'dMonitoringState', 'plan', 'pathPlan', 'liveLocationKalman',
                                      'frame', 'frontFrame'])
+      if SIMULATION:
+        self.sm.ignore_alive = ['ubloxRaw','frontFrame']
```
marschr, line 59: Setting it right on the `__init__` like so? Worked as expected here, saved doing an `if` on every call to `update_events()`

adeebshihadeh: pass it in on instantiation

adeebshihadeh, line 59:
```suggestion
      ignore = ['ubloxRaw', 'frontFrame'] if SIMULATION else []
      self.sm = messaging.SubMaster(['thermal', 'health', 'model', 'liveCalibration', 'ubloxRaw',
                                     'dMonitoringState', 'plan', 'pathPlan', 'liveLocationKalman',
                                     'frame', 'frontFrame'], ignore_alive=ignore)
```

marschr:
Tested again. All good.

adeebshihadeh:
The CARLA window not being shown was intentional. Why do we want to show it? Can we put it behind an env var flag and default to not shown?

marschr:
Oh, I thought the other way around from people on Discord expecting to see something when they run CARLA.
But yes, an env var flag would be nice and we default it to windowless. We can add the instructions for those wanting the window on the Simulation Wiki and README later on.

adeebshihadeh:
You should see the openpilot UI when starting CARLA. Does that not happen?

marschr:
As far as I know one would have to run `start_carla.sh` first and then `launch_openpilot.sh` separately, that goes for both running bare metal or the docker version of openpilot (which has it's own current issues with the nvidia cuda+cudnn deps etc).

adeebshihadeh:
Yeah, that's what I meant. Do you have a use case for having both displays? Either way, I'm fine with having the flag to enable that as long as it defaults off.

marschr:
Mostly for assisting someone else on Discord, checking if CARLA is running fine to begin with, although it's useful to have for debugging or inspecting stuff on CARLA sometimes, like the drawing distance/anti-aliasing issues we were discussing on Discord few days ago.
But yeah, other than that and also thinking on any future automation/CI related stuff with CARLA, defaulting to off is the way.

adeebshihadeh:
Makes sense, thanks for the fixes :)

PR #19831, opened 2021-01-19
loggerd: split bootlog to a separate program

deanlee:


selfdrive/loggerd/logger.cc
```diff
@@ -12,12 +12,102 @@
...
 #include "logger.h"
 
+void append_property(const char* key, const char* value, void *cookie) {
+  std::vector<std::pair<std::string, std::string> > *properties =
+    (std::vector<std::pair<std::string, std::string> > *)cookie;
+
+  properties->push_back(std::make_pair(std::string(key), std::string(value)));
+}
+
+static bool file_exists(const std::string& fn) {
```
adeebshihadeh, line 36: let's move this to util

selfdrive/loggerd/SConscript
```diff
@@ -1,7 +1,9 @@
 Import('env', 'arch', 'cereal', 'messaging', 'common', 'visionipc', 'gpucommon')
 
-src = ['loggerd.cc', 'logger.cc']
-libs = ['zmq', 'capnp', 'kj', 'z',
+
+logger_lib = env.Library('logger', ["logger.cc"])
+src = ['loggerd.cc']
```
adeebshihadeh, line 5: just put this directly in the env.Program call

selfdrive/loggerd/logger.cc
```diff
@@ -12,12 +12,97 @@
...
 #include "logger.h"
 
+void append_property(const char* key, const char* value, void *cookie) {
+  std::vector<std::pair<std::string, std::string> > *properties =
+    (std::vector<std::pair<std::string, std::string> > *)cookie;
+
+  properties->push_back(std::make_pair(std::string(key), std::string(value)));
+}
+
+kj::Array<capnp::word> gen_init_data() {
```
adeebshihadeh, line 36: let's make this log_gen_init_data to match the pattern we use for sentinel

selfdrive/loggerd/SConscript
deanlee: no,we can't

adeebshihadeh:
Nice, this should make loggerd.cc a lot more readable! Can you add bootlog to the gitignore in the openpilot root? We should have a CI test for that.

selfdrive/loggerd/SConscript
adeebshihadeh: why not?

selfdrive/loggerd/logger.cc
deanlee: call `log_init_data` in `logger_init` now.Is this what you want

selfdrive/loggerd/SConscript
deanlee: `src` is modified by the following conditional statement

adeebshihadeh: ah, didn't see it in the diff view :)

selfdrive/loggerd/logger.h
```diff
@@ -5,7 +5,13 @@
 #include <stdint.h>
 #include <pthread.h>
 #include <bzlib.h>
-
+#include <kj/array.h>
+#include <capnp/serialize.h>
+#if defined(QCOM) || defined(QCOM2)
```
adeebshihadeh, line 10: add some whitespace here

adeebshihadeh: let's also just make logger.c into a cc file so we can remove the extern c stuff

selfdrive/loggerd/loggerd.cc
```diff
@@ -495,11 +362,7 @@ int main(int argc, char** argv) {
   }
 
   // init logger
-  {
-    auto words = gen_init_data();
-    auto bytes = words.asBytes();
-    logger_init(&s.logger, "rlog", bytes.begin(), bytes.size(), true);
-  }
+  logger_init(&s.logger, "rlog", true);
```
adeebshihadeh, line 365: this is so much nicer :)

selfdrive/loggerd/logger.cc
adeebshihadeh: yes

selfdrive/loggerd/loggerd.cc
deanlee: I think so :laughing:

PR #19833, opened 2021-01-19
loggerd: new function drain_socket

deanlee:


adeebshihadeh:
I don't think this is any cleaner. This just moves the ugliness into a function.

deanlee:
I think it’s still a bit better than putting it in the `main()`. and keep `fpkt_id` in `SocketState` is better than kepp `fpkt_sock` in `RotateState` too.

deanlee:
closed. I'm thinking of a complete solution

PR #19816, opened 2021-01-18
LoggerState: refactor to class

deanlee:
1. convert structs to classes.
2. use std::shared_ptr to encapsulate Logger handles.
3. get `segment_path`&`part` from LoggerHandle to ensure thread-safety

PR #19839, opened 2021-01-19
Refactor loggerd rotations

deanlee:
1. Remove class `RotateState`, only use one mutex `rotate_lock` to control all rotations.
2. Log frame socket in the encoder thread, sync frame id is no longer required.
3. Minimize the use of locks


deanlee:
rebased master,passed the tests of `test_encoder` and `test_loggerd`

ZwX1616:
see comment in #19816

PR #19843, opened 2021-01-19
2018 Sonata FPv1 Fingerprint Support

Jon8RFC:
2018 Hyundai Sonata Limited Ultimate (SCC+LKAS)

**Checklist**
- [Y] added to README
- [N] test route added to [test_car_models](../../selfdrive/test/test_car_models.py)
- [Y] route with openpilot: 03690f0d837307af|2021-01-19--15-57-30
- [Y] route with stock system: 03690f0d837307af|2021-01-13--21-36-53

1. Update to the Sonata 2019 FPv1 support already there because of a minor fingerprint difference.  As far as I could tell, things worked fine on the 2019 fingerprint, but it should be updated.
2. Renamed "Sonata_2019" to "Sonata_LF" (LF = 2014-2019 Sonata) because 2016-2017 Sonata Limited Ultimate vehicles with the 2.4L engine also have SCC+LKAS and can eventually be supported.  The key factor is the Ultimate package on those year-models to have both SCC and LKAS.
I don't know about the 2016-2017 Sonata Limited Ultimate 2.0T


adeebshihadeh:
I'd prefer not to merge new FP v1 values for hyundai. Can you remove the CAN fingerprint update and add your FW versions instead?

Jon8RFC:
> I'd prefer not to merge new FP v1 values for hyundai. Can you remove the CAN fingerprint update and add your FW versions instead?

Unfortunately, I cannot properly do that.  OP is not retrieving the EPS / MDPS FW information and I have no idea how that all plays in.  There are multiple versions for transmission and camera FW on my car which OP retrieves as well, which I also don't know how to properly interpret.

Jon8RFC:
Obviously, it's not as simple as I had hoped with failures all over.  Bummer.
I'm not a software developer, so I have no idea how to find dependencies and work backwards from making a change to properly update everything.

The 2018 Sonata does work without any known issues and there are both FPv1 and FPv2 differences, but I apparently can't make this work without checks failing all over.  Sorry for wasting your time.

PR #19845, opened 2021-01-20
Update cpu_usage_stat.py

ulriknaess:
Minor changes.

Checking if the condition is met, then do rather than checking if the condition is not met then don't do.


adeebshihadeh:
I think this is a bit more readable the way it is.

PR #19846, opened 2021-01-20
2021 Pacifica Hybrid fingerprint

VirtuallyChris:
The 2021 Pacifica Hybrid would sometimes identify as mock - this should fix.

The fingerprint is a sub-set of this existing one, merged them. Dongle ID b66bf349eecf2fda


selfdrive/car/chrysler/values.py
```diff
@@ -62,7 +62,7 @@ class CAR:
...
     },
-    # Based on "8190c7275a24557b|2020-02-24--09-57-23"
+    # Based on "8190c7275a24557b|2020-02-24--09-57-23" and "b66bf349eecf2fda|2021-01-19--22-47-32"
```
adeebshihadeh, line 65: I'd just remove this comment

VirtuallyChris: Sounds good.

PR #19848, opened 2021-01-20
logger: cache init_data

deanlee:


selfdrive/loggerd/logger.cc
```diff
@@ -253,7 +253,8 @@ int logger_next(LoggerState *s, const char* root_path,
   pthread_mutex_unlock(&s->lock);
 
   // write beggining of log metadata
-  log_init_data(s);
+  auto bytes = s->init_data.asBytes();
```
adeebshihadeh, line 256: if we save the bytes, we don't have to do this conversion every time

deanlee: The kj library has no obvious way to do this,so I used std::vector to make a copy

deanlee: Maybe the original version will be better.adding a conversion will not have any impact.

deanlee: There is a function `releaseAsBytes` can be used

deanlee: sorry,I'm wrong,we can't call releaseAsBytes on kj::Array<capnp::word>:
` error: static_assert failed due to requirement 'sizeof(capnp::word) == sizeof(unsigned char)' "releaseAsBytes() only possible on arrays with byte-size elements (e.g. chars)."`

deanlee:
closed,merged to https://github.com/commaai/openpilot/pull/19942

PR #19849, opened 2021-01-20
vision-only radar toggle

gregjhogan:
toggle for enabling/disabling stock longitudinal control


davetapley:
Am I correct in thinking this is https://github.com/commaai/openpilot/discussions/2552 ?

PR #19853, opened 2021-01-20
loggerd: run clear_locks() asynchronously

deanlee:


pd0wm:
Is there a clear benefit? Is this operation slowing down loggerd start significantly?

deanlee:
I'm not sure,still testing.I have encountered it running for 0.5 seconds

pd0wm:
Why do we want to clear the locks up in the first place? In case loggerd didn't shut down cleanly?

deanlee:
this seems to be the reason.

adeebshihadeh:
Closing this for now. I think we re-evaluate whether we really want to clear the locks. If anything, the uploader should probably do this.

PR #19857, opened 2021-01-20
loggerd : log frame socket in the encoder thread

deanlee:
split https://github.com/commaai/openpilot/pull/19839 into smaller PRS


adeebshihadeh:
I like the cleanup of the main thread, but not sure this is much cleaner.

deanlee:
I think it is more reasonable to log the frame packet with the encoded frame in the encoder thread，which can save a lot of code to synchronize the frame_id with the main thread.
I'll try to keep making the code clearer.

PR #19858, opened 2021-01-20
INDI: Time constant is used based on breakpoints

sshane:
Previously only was created on class initialization which caused speed to always be zero for it, now we update it every loop so the breakpoints are used now


pd0wm:
If it's only used in the update function it doesn't have to be a class member anymore

ShaneSmiskol:
I thought about that but didn't know if we needed it for inspecting from outside the class, pushed

PR #19860, opened 2021-01-20
Split planner and pathplanner publishing into separate 'publish' methods

mitchellgoffpc:
This PR will allow us to use pathplanner in the simulator without using a fake PubMaster.

### DIFF
--- a/selfdrive/controls/lib/long_mpc.py
+++ b/selfdrive/controls/lib/long_mpc.py
@@ -28,20 +28,21 @@ class LongitudinalMpc():
     self.n_its = 0
     self.duration = 0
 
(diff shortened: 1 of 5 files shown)


selfdrive/controls/lib/long_mpc.py
```diff
@@ -120,3 +119,7 @@ def update(self, pm, CS, lead):
       self.v_mpc = v_ego
       self.a_mpc = CS.aEgo
       self.prev_lead_status = False
+
+  def publish(self, pm):
+    if LOG_MPC:
+      self.send_mpc_solution(pm, self.n_its, self.duration)
```
adeebshihadeh, line 125: i'd just rename send_mpc_solution to publish and move the condition

PR #19869, opened 2021-01-21
Set parents on all Qt objects

pd0wm:


grekiki:
Probably not a good idea

PR #19870, opened 2021-01-21
Qt parents3

ghost:
Reduce the amount of parents we pass in the qt widgets. Turns out it is unnecessary.


grekiki:
Actually, we are also going to imrprove the UI look in this commit

PR #19876, opened 2021-01-22
Wifi-widget onroad lag reduction

ghost:
Fix UI onroad lag by not refreshing the wifi widget if it is not visible


marschr:
I wonder if qt state machine APIs wouldn't help managing these things as the whole offroad interface start to grow.

grekiki:
> I wonder if qt state machine APIs wouldn't help managing these things as the whole offroad interface start to grow.

Haven`t even heard of this feature yet... 
I'll test them here https://github.com/commaai/openpilot/pull/19879

PR #19880, opened 2021-01-22
manager.py broadcast process states

pd0wm:

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -29,6 +29,7 @@ STEER_ANGLE_SATURATION_THRESHOLD = 2.5  # Degrees
 
 SIMULATION = "SIMULATION" in os.environ
 NOSENSOR = "NOSENSOR" in os.environ
+IGNORE_PROCESSES = set(["rtshield", "uploader", "deleter", "loggerd", "logmessaged", "tombstoned", "logcatd", "proclogd", "clocksd", "gpsd", "updated"])
 
 ThermalStatus = log.ThermalData.ThermalStatus
 State = log.ControlsState.OpenpilotState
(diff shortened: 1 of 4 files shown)


selfdrive/controls/lib/events.py
```diff
@@ -615,6 +615,11 @@ def wrong_car_mode_alert(CP: car.CarParams, sm: messaging.SubMaster, metric: boo
                               audible_alert=AudibleAlert.chimeDisengage),
   },
 
+  EventName.processNotRunning: {
+    ET.NO_ENTRY: NoEntryAlert("Not All Processes Are Running",
```
adeebshihadeh, line 619: what about something like system malfunction? I think this has too much information for a user and can cause confusion. what's a process? will they start up in a minute if I try to reenagage?

pd0wm: Something like: "System Malfunction. Reboot Device"?

pd0wm:
We also need a way for this check to not trigger during tests in CI. We can do two things:
 - Change all tests to also broadcast a managerState packet
 - Only trigger the check if we have seen at least 1 managerState packet.

I'd like to do the latter.

selfdrive/controls/controlsd.py
```diff
@@ -166,6 +167,11 @@ def update_events(self, CS):
     if self.sm['thermal'].memUsedPercent > 90:
       self.events.add(EventName.lowMemory)
 
+    # Check if all manager processes are running
+    not_running = set(p.name for p in self.sm['managerState'].processes if not p.running)
+    if self.sm.rcv_frame['managerState'] and (not_running - IGNORE_PROCESSES):
```
adeebshihadeh, line 172: hmm, why don't we add a timeout to rcv_frame? it's possible for this check to not work if the packet is never published

adeebshihadeh: guess the alive check will catch this

PR #19888, opened 2021-01-23
Fix modeld: use dangling pointers to build message

deanlee:


deanlee:
My bad the list will not have this problem

deanlee:
My bad,the list does not have this problem

PR #19892, opened 2021-01-23
Fix panda:  dangling pointer problem in can_receive

deanlee:
1. Fix dangling pointer problem, we're returning a dangling pointer here:
    https://github.com/commaai/openpilot/blob/c9679222aebfcbf0dcce7ac00d2e03d19101a81f/selfdrive/boardd/panda.cc#L361
    dat is defined as `@2 :Data;`, the `capnp::Data::Reader` doesn't make a copy, it just reference to the `data` .
2. Do not publish empty "can" messages when there is no data

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -192,14 +192,6 @@ void usb_retry_connect() {
   LOGW("connected to board");
 }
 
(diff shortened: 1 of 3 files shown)


pd0wm:
Good catch!

We still want to send messages even if there is no data, otherwise we can't do proper timing analysis on the messages from boardd.

PR #19894, opened 2021-01-23
Use refactored Submaster::updated

deanlee:
related PR in cereal: https://github.com/commaai/cereal/pull/112

the new `SubMaster::updated` function returns `std::optional<cereal::Event::Reader>`,merge the two calls `sm::updated(name)`,` sm[name`] into one


adeebshihadeh:
I don’t think this is cleaner, and it breaks from the python SubMaster’s pattern.

PR #19897, opened 2021-01-23
Panda: add check for msg_count in function can_send

deanlee:


pd0wm:
Not sure we should do it. I think we want to keep the traffic to the panda at a 100Hz, even if there are no messages to prevent weird changes in timing.

deanlee:
oic

PR #19904, opened 2021-01-24
OmxEncoder: do not call handle_out_buf if flag is EOS

deanlee:


adeebshihadeh:
I don't think we can do this because https://github.com/commaai/openpilot/blob/master/selfdrive/loggerd/omx_encoder.cc#L398.

PR #19907, opened 2021-01-24
hyundai: WHL_SPD11 checksum

gregjhogan:
let's see if all the tests pass after https://github.com/commaai/panda/pull/622


gregjhogan:
first test was with WHL_SPD11 added to legacy checks, which didn't work
once panda PR gets merged will rebase and this should turn green

adeebshihadeh:
We can just bump on master once the panda PR is merged.

PR #19912, opened 2021-01-25
can_list_to_can_capnp_cpp: write message directly to the output string

deanlee:


pd0wm:
Can the code be cleaner if it just returns the `std::string`? I don't immediately see a reason why it has to be passed by reference.

deanlee:
return string will cause a link error
cython use C-linkage:
__PYX_EXTERN_C DL_IMPORT(std::string) can_list_to_can_capnp_cpp

PR #19913, opened 2021-01-25
UI Refactor: move vision releated code into class UIVision

deanlee:
1. move vision releated code into class `UIVision`
2. create and delete `UIVision` on demand.

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -108,30 +108,6 @@ static void ui_draw_line(UIState *s, const line_vertices_data &vd, NVGcolor *col
   nvgFill(s->vg);
 }
 
-void UIVision::draw() {
-  glBindVertexArray(frame_vao);
-  glActiveTexture(GL_TEXTURE0);
-
-  if (last_frame) {
-    glBindTexture(GL_TEXTURE_2D, texture[last_frame->idx]->frame_tex);
-#ifndef QCOM
-    // this is handled in ion on QCOM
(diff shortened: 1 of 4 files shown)


pd0wm:
Sorry I left this unreviewed for so long. Merged your PR with moving the intrinsics matrix to modeldata, so we'd need to integrate that here.

Would be possible to make the ui adapt to different intrinsics matrices on the fly? Currently they are hardcoded by QCOM/QCOM2 ifdefs, but that makes it impossible to review data on the computer. You should be able to use the image size reported by VisionIPC to choose the correct one. This would also require delaying some of the init code until the first frame comes in.

deanlee:
ok I'll try

deanlee:
Put the code of UIVision in paint.cc to make it easier for review. can be moved to ui.cc next time.

only inited `road_view_transform` from `VisionBuf`. I can't figure out a clear way to init driver_view_transform now. it might be better to do it next time.

adeebshihadeh:
I'm not sure this refactor makes much sense with the new UI. I think it makes more sense to move the onroad UI to Qt from NVG and make a nice camera view widget. @pd0wm what do you think?

adeebshihadeh:
So we want to move the onroad UI to Qt, but I think #20516 is a necessary pre-requisite. We'll get that merged then we can slowly migrate all the NanoVG stuff to Qt. As for this PR, I think it's best to refactor it a bit to make it a generic camera view widget.

PR #19915, opened 2021-01-25
2018 Hyundai Sonata

Jon8RFC:
2018 Hyundai Sonata Limited Ultimate (SCC+LKAS)

**Checklist**
- [Y] added to README
- [Y&N] test route added to [test_car_models](../../selfdrive/test/test_car_models.py)
- (the above already has a route; my car worked without FPv1 modifications and it was slightly different, but should still be updated to FPv2 anyway)
- [Y] route with openpilot: 03690f0d837307af|2021-01-25--07-06-52
- [Y] route with openpilot: 03690f0d837307af|2021-01-20--15-00-10
- [Y] route with stock system: 03690f0d837307af|2021-01-12--11-25-29

1. Updated the 2019 Sonata FPv1 to include 2018 Sonata FPv2 (includes all firmware variants I've seen in my logs).
2. Refactored "Sonata_2019" to "Sonata_LF" (LF = 2015-2019 Sonatas) because the 2016-2018 Sonata Limited Ultimate vehicles with the 2.4L engine also have SCC+LKAS and could eventually be supported.  The key factor is the Ultimate package on those year-models to have both SCC and LKAS.


Jon8RFC:
I'd like to fix any failures in the tests first.

selfdrive/car/hyundai/values.py
```diff
@@ -30,7 +30,7 @@ class CAR:
   KONA_EV = "HYUNDAI KONA ELECTRIC 2019"
   SANTA_FE = "HYUNDAI SANTA FE LIMITED 2019"
   SONATA = "HYUNDAI SONATA 2020"
-  SONATA_2019 = "HYUNDAI SONATA 2019"
+  SONATA_LF = "HYUNDAI SONATA 2018-2019"
```
pd0wm, line 33:
```suggestion
  SONATA_LF = "HYUNDAI SONATA 2019"
```

We usually don't change the car name after it's been in there for a while. This string ends up in our backend.

VirtuallyChris:
What harness do you use with this car?

Jon8RFC:
> What harness do you use with this car?

Hyundai E

PR #19916, opened 2021-01-25
dmonitoring_publish: use kj::ArrayPtr<const float> raw_pred as paramater instead of float *

deanlee:


adeebshihadeh:
In the future, I'd make one combined PR with #19917.

PR #19920, opened 2021-01-26
camera_common: rename create_thumbnail to publish_thumbnail

deanlee:
rename create_thumbnail to publish_thumbnail, and send` PubMaster*` to it instead of `MultiCameraState*`

### DIFF
--- a/selfdrive/camerad/cameras/camera_common.cc
+++ b/selfdrive/camerad/cameras/camera_common.cc
@@ -237,7 +237,7 @@ void fill_frame_image(cereal::FrameData::Builder &framed, const CameraBuf *b) {
   delete[] resized_dat;
 }
 
-static void create_thumbnail(MultiCameraState *s, const CameraBuf *b) {
+static void publish_thumbnail(PubMaster *pm, const CameraBuf *b) {
   uint8_t* thumbnail_buffer = NULL;
   unsigned long thumbnail_len = 0;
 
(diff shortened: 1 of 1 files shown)


selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -295,9 +295,7 @@ static void create_thumbnail(MultiCameraState *s, const CameraBuf *b) {
   thumbnaild.setTimestampEof(b->cur_frame_data.timestamp_eof);
   thumbnaild.setThumbnail(kj::arrayPtr((const uint8_t*)thumbnail_buffer, thumbnail_len));
 
-  if (s->pm != NULL) {
```
pd0wm, line 298: Why is the null check no longer necessary? Is the pubmaster initialized on all camera types?

deanlee: yes,except for the camera_frame_stream that does not call `processing_thread`

pd0wm: Shall we just leave the null check just for extra safety? Don't want to have unexpected surprises if we refactor the code later.

deanlee: how about check it before call publish_thumbnail

pd0wm: that's fine too!

PR #19922, opened 2021-01-26
rgb_yuv_init: remove memset

deanlee:
remove unnecessary memset, all three member variables:`width`, `height`,`rgb_to_yuv_krnl`  have been initialized in rgb_yuv_init


adeebshihadeh:
Is there a point to this besides saving a line? This only happens to be ok because we set all three fields, but what if we added another one we didn't initialize in rgb_yuv_init?

deanlee:
I personally think use the default initialization x = {} or default member initializer is better, memset on object only if it is really required。

pd0wm:
I can't find the default initialization in the code, so don't see how we can around the memset.

PR #19935, opened 2021-01-27
CameraBuf::acquire : release on failure

deanlee:


pd0wm:
I only see two instances of the cloudlog over the last two weeks, so this seems to be hit only very rarely.

PR #19945, opened 2021-01-28
camerad frame lag/drop jenkins tests

ZwX1616:
better separates camera test and logger test


selfdrive/camerad/test/test_camerad.py
```diff
@@ -74,27 +72,28 @@ def _is_exposure_okay(self, i, med_ex=np.array([0.2,0.4]), mean_ex=np.array([0.2
...
-      # check again for consistency
-      pic, fpic = get_snapshots()
-      self.assertTrue(self._is_really_sharp(pic))
-      self.assertTrue(self._is_exposure_okay(pic))
-      self.assertTrue(self._is_exposure_okay(fpic))
-    elif TICI:
-      raise unittest.SkipTest # TBD
-    else:
-      raise unittest.SkipTest
+  # @with_processes(['camerad'])
```
adeebshihadeh, line 75: use the skip test decorator from the unittest module

PR #19947, opened 2021-01-28
fix each segment in a route having a different initial logMonoTime

gregjhogan:
fix every log having a different `logMonoTime` in `initData` packet

### DIFF
--- a/selfdrive/loggerd/logger.h
+++ b/selfdrive/loggerd/logger.h
@@ -31,6 +31,7 @@ typedef struct LoggerHandle {
 
 typedef struct LoggerState {
   pthread_mutex_t lock;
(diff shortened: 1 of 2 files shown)


selfdrive/loggerd/logger.h
```diff
@@ -31,6 +31,7 @@ typedef struct LoggerHandle {
 
 typedef struct LoggerState {
   pthread_mutex_t lock;
+  kj::ArrayPtr<capnp::byte> init_data;
```
deanlee, line 34: dangling pointer.
I have submitted a PR  https://github.com/commaai/openpilot/pull/19942 and can use that as a cache

PR #19948, opened 2021-01-28
OMXEncoder: remove pthread_mutex_t lock  from member variables

deanlee:
I don’t see any reason to use a re-entrant mutex in QMXEncoder.
It may be used to rotate encoder in another thread (main thread) before. now we rotate it in the encoder thread, there is no multi-threading problem, I think we don't need it anymore,did I miss something?


selfdrive/loggerd/omx_encoder.cc
```diff
@@ -409,18 +408,11 @@ int OmxEncoder::encode_frame(const uint8_t *y_ptr, const uint8_t *u_ptr, const u
...
                              int *frame_segment, uint64_t ts) {
   int err;
-  pthread_mutex_lock(&this->lock);
-
   if (!this->is_open) {
-    pthread_mutex_unlock(&this->lock);
     return -1;
   }
 
-  // this sometimes freezes... put it outside the encoder lock so we can still trigger rotates...
```
deanlee, line 419: It seems to be used for rotation

PR #19949, opened 2021-01-28
cloudlog once on commIssue

pd0wm:


adeebshihadeh:
I think this should be somewhere in log.Event. What if some process controlsd subscribes to is invalid? Do we add a cloudlog to that process too? Then, we'd also get them in the qlogs.

pd0wm:
You mean handle this more generic in SubMaster?

We could make an optional argument to the SubMaster constructor to pass in a logger object. If it exists we can use that to log some interesting events.

adeebshihadeh:
Let's merge this for now, but it would be nice to get this in a struct in cereal.

PR #19951, opened 2021-01-28
Pigeon::receive:  reserve 4kb+64b for std::string

deanlee:


selfdrive/boardd/pigeon.cc
```diff
@@ -105,9 +105,9 @@ void PandaPigeon::send(const std::string &s) {
 
 std::string PandaPigeon::receive() {
   std::string r;
-
+  r.reserve(0x1000 + 0x40);
+  unsigned char dat[0x40];
   while (true){
-    unsigned char dat[0x40];
     int len = panda->usb_read(0xe0, 1, 0, dat, sizeof(dat));
     if (len <= 0 || r.length() > 0x1000) break;
```
deanlee, line 112: If the string length exceeds 0x1000, the message that has been read will be dropped.is that ok?

pd0wm: That looks like a bug. Probably doesn't happen in practice much, but let's fix it.

Issue #19776, opened 2021-01-16
PyJWT 2.0.0 does not return bytes for encode()


gregjhogan:
calling `jwt.encode(...).decode('utf-8')` on pyjwt 2.0.0 and higher throws an error because `jwt.encode(...)` returns a `str` now instead of `bytes` (and Pipfile was updated to pyjwt 2.0.0)
https://github.com/commaai/openpilot/blob/def24ae27ad4cd5948731770745d5039f20ffcf2/Pipfile.lock#L613-L620
if tests didn't catch this, we should probably add one!
https://github.com/commaai/openpilot/blob/def24ae27ad4cd5948731770745d5039f20ffcf2/common/api/__init__.py#L30

ShadyD45:
can we fix this by converting the return value of `encode()` to bytes using `bytes()` instead of decode and then return the resulting value which is in `bytes` ? 
Something like this:  `return bytes(jwt.encode(payload, self.private_key, algorithm='RS256'), 'utf8')` 
Should I open a PR?

PR #19958, opened 2021-01-29
Fix jwt.encode return type (#19776)

ShadyD45:


pd0wm:
Get token is supposed to return a string, so the decode can just be removed. However, the code needs to be compatible with both pyjwt before and after 2.0. Can you check the return type of jwt.encode and decode if necessary?

Can you also write a test to make sure we can successfully generate a token and use it make an api request?

ShadyD45:
We can just store the encoded value returned from jwt.encode in a variable and check its type as string or byte. If string then we return the variable as it is and if byte then we use decode on it and return the result. 

Have edited the code, can you please review and see if everything's right?

pd0wm:
You can write a new test in this folder https://github.com/commaai/openpilot/tree/master/common/tests

I would to the following:
 - Make sure a private key is generated. See selfdrive/registration.py. Extract line 40-44 in a function so you can call it from your tests. This is a prerequisite to create a jwt token.
 - Create a token, make sure it's a string

ShadyD45:
I have a question. while writing test for get_token can i provide a dummy dongle_id to Api class for creating its object

pd0wm:
Yes, that's fine

common/api/__init__.py
```diff
@@ -27,7 +27,12 @@ def get_token(self):
       'iat': now,
       'exp': now + timedelta(hours=1)
     }
-    return jwt.encode(payload, self.private_key, algorithm='RS256').decode('utf8')
+    token = jwt.encode(payload, self.private_key, algorithm='RS256')
+    if isinstance(token, str):
+      return token
+    elif isinstance(token, bytes):
+      return token.decode('utf8')
```
adeebshihadeh, line 34:
```suggestion
    if isinstance(token, bytes):
      token = token.decode('utf8')
    return token
```

ShadyD45:
> Yes, that's fine

Have opened a PR for the test file. Can you review?

pd0wm:
Can you add the test to this PR?

ShadyD45:
`import unittest
import os

from common.api import Api
from common.file_helpers import mkdirs_exists_ok
from common.basedir import PERSIST

class TestApiGetToken(unittest.TestCase):
	def is_private_key_generated(self):
		mkdirs_exists_ok(PERSIST+"/comma")
		assert os.system("openssl genrsa -out "+PERSIST+"/comma/id_rsa.tmp 2048") == 0
		assert os.system("openssl rsa -in "+PERSIST+"/comma/id_rsa.tmp -pubout -out "+PERSIST+"/comma/id_rsa.tmp.pub") == 0
		os.rename(PERSIST+"/comma/id_rsa.tmp", PERSIST+"/comma/id_rsa")
		os.rename(PERSIST+"/comma/id_rsa.tmp.pub", PERSIST+"/comma/id_rsa.pub")
	
	def test_get_token(self):
		# This is prerequisite to create jwt token
		self.is_private_key_generated()
		
		# Create Api class object with dummy DongleId
		test_obj = Api("cb38263377b873ee")
		# Obtain a jwt token
		test_token = test_obj.get_token()	
		
		# Check if the token is a string
		self.assertEqual(isinstance(test_token, str), True)
		
if __name__ == "__main__":
	unittest.main()`

please check if its alright?

ShadyD45:
`
import unittest
import os

from common.api import Api
from common.file_helpers import mkdirs_exists_ok
from common.basedir import PERSIST

class TestApiGetToken(unittest.TestCase):
	def is_private_key_generated(self):
		mkdirs_exists_ok(PERSIST+"/comma")
		assert os.system("openssl genrsa -out "+PERSIST+"/comma/id_rsa.tmp 2048") == 0
		assert os.system("openssl rsa -in "+PERSIST+"/comma/id_rsa.tmp -pubout -out "+PERSIST+"/comma/id_rsa.tmp.pub") == 0
		os.rename(PERSIST+"/comma/id_rsa.tmp", PERSIST+"/comma/id_rsa")
		os.rename(PERSIST+"/comma/id_rsa.tmp.pub", PERSIST+"/comma/id_rsa.pub")
	
	def test_get_token(self):
		# This is prerequisite to create jwt token
		self.is_private_key_generated()
		
		# Create Api class object with dummy DongleId
		test_obj = Api("cb38263377b873ee")
		# Obtain a jwt token
		test_token = test_obj.get_token()	
		
		# Check if the token is a string
		self.assertEqual(isinstance(test_token, str), True)
		
if __name__ == "__main__":
	unittest.main()

`

ShadyD45:
```
import unittest
import os

from common.api import Api
from common.file_helpers import mkdirs_exists_ok
from common.basedir import PERSIST

class TestApiGetToken(unittest.TestCase):
	def is_private_key_generated(self):
		mkdirs_exists_ok(PERSIST+"/comma")
		assert os.system("openssl genrsa -out "+PERSIST+"/comma/id_rsa.tmp 2048") == 0
		assert os.system("openssl rsa -in "+PERSIST+"/comma/id_rsa.tmp -pubout -out "+PERSIST+"/comma/id_rsa.tmp.pub") == 0
		os.rename(PERSIST+"/comma/id_rsa.tmp", PERSIST+"/comma/id_rsa")
		os.rename(PERSIST+"/comma/id_rsa.tmp.pub", PERSIST+"/comma/id_rsa.pub")
	
	def test_get_token(self):
		# This is prerequisite to create jwt token
		self.is_private_key_generated()
		
		# Create Api class object with dummy DongleId
		test_obj = Api("cb38263377b873ee")
		# Obtain a jwt token
		test_token = test_obj.get_token()	
		
		# Check if the token is a string
		self.assertEqual(isinstance(test_token, str), True)
		
if __name__ == "__main__":
	unittest.main()

```
please check if its alright

common/tests/test_get_token.py
```diff
@@ -0,0 +1,32 @@
+import unittest
+import os
+
+from common.api import Api
+from common.file_helpers import mkdirs_exists_ok
+from common.basedir import PERSIST
+
+class TestApiGetToken(unittest.TestCase):
+    def is_private_key_generated(self):
```
adeebshihadeh, line 9: wrong indentation, use 2 spaces

adeebshihadeh, line 9: don't copy this code, just abstract it and use it in the tests

ShadyD45: Should I just check if the private key is generated by checking if id_rsa exists, and not actually try to run commands to generate it?

pd0wm: yes, don't generate if already present. The logic for this should all live in registration.py though

ShadyD45: Okay, Something like this in check private key function :

`assertEqual(os.path.exists(path_to_id_rsa), True)`

And `path_to_id_rsa` will be the path extracted from registration.py

Or use If-else instead of assert 

What to do if private key is not present? Raise an exception and stop?

pd0wm: Let me just write the test, that's probably easier. I'll merge the changes to the api

PR #19959, opened 2021-01-29
Logger: new class BZFile

deanlee:
add a small class `BZFile` to encapsulate the writing.

### DIFF
--- a/selfdrive/loggerd/bootlog.cc
+++ b/selfdrive/loggerd/bootlog.cc
@@ -1,12 +1,6 @@
 #include <assert.h>
 #include <string>
-#include <cstdio>
-
-#include <bzlib.h>
-
 #include "common/swaglog.h"
-#include "common/util.h"
-#include "messaging.hpp"
 #include "logger.h"
 
 int main(int argc, char** argv) {
--- a/selfdrive/loggerd/bootlog.cc
+++ b/selfdrive/loggerd/bootlog.cc
@@ -26,30 +26,13 @@ int main(int argc, char** argv) {
   int r = logger_mkpath((char*)path.c_str());
   assert(r == 0);
 
-  FILE * file = fopen(path.c_str(), "wb");
-  assert(file != nullptr);
-
-  // Open as bz2
-  int bzerror;
-  BZFILE* bz_file = BZ2_bzWriteOpen(&bzerror, file, 9, 0, 30);
-  assert(bzerror == BZ_OK);
+  std::unique_ptr<BZFile> bz_file = std::make_unique<BZFile>(path.c_str());
 
   // Write initdata
-  kj::Array<capnp::word> init_msg = logger_build_init_data();
-  auto bytes = init_msg.asBytes();
-  BZ2_bzWrite(&bzerror, bz_file, bytes.begin(), bytes.size());
-  assert(bzerror == BZ_OK);
+  bz_file->write(logger_build_init_data().asBytes());
 
   // Write bootlog
-  kj::Array<capnp::word> boot_msg = logger_build_boot();
-  bytes = boot_msg.asBytes();
-  BZ2_bzWrite(&bzerror, bz_file, bytes.begin(), bytes.size());
-  assert(bzerror == BZ_OK);
+  bz_file->write(logger_build_boot().asBytes());
 
-  // Close bz2 and file
-  BZ2_bzWriteClose(&bzerror, bz_file, 0, NULL, NULL);
-  assert(bzerror == BZ_OK);
-
-  fclose(file);
   return 0;
 }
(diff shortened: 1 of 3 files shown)


selfdrive/loggerd/logger.h
```diff
@@ -15,18 +17,39 @@ const std::string LOG_ROOT = util::getenv_default("HOME", "/.comma/media/0/reald
...
+  }
+  ~BZFile() {
+    int bzerror;
+    BZ2_bzWriteClose(&bzerror, bz_file, 0, nullptr, nullptr);
+    fclose(file);
+  }
+  inline void write(void* data, size_t size) {
+    int bzerror;
+    BZ2_bzWrite(&bzerror, bz_file, data, size);
+    assert(bzerror == BZ_OK);
```
pd0wm, line 36: For loggerd we probably want a more graceful handling of write errors + some cloudlog.

selfdrive/loggerd/bootlog.cc
```diff
@@ -26,30 +20,13 @@ int main(int argc, char** argv) {
...
   assert(r == 0);
 
-  FILE * file = fopen(path.c_str(), "wb");
-  assert(file != nullptr);
-
-  // Open as bz2
-  int bzerror;
-  BZFILE* bz_file = BZ2_bzWriteOpen(&bzerror, file, 9, 0, 30);
-  assert(bzerror == BZ_OK);
+  std::unique_ptr<BZFile> bz_file = std::make_unique<BZFile>(path.c_str());
```
pd0wm, line 23: Why a pointer an not just on the stack?

selfdrive/loggerd/logger.h
deanlee: Only log the first error per segment,is that ok?

pd0wm: Sounds good!

selfdrive/loggerd/logger.h
```diff
@@ -15,18 +17,44 @@ const std::string LOG_ROOT = util::getenv_default("HOME", "/.comma/media/0/reald
...
+    int bzerror;
+    bz_file = BZ2_bzWriteOpen(&bzerror, file, 9, 0, 30);
+    assert(bzerror == BZ_OK);
+  }
+  ~BZFile() {
+    int bzerror;
+    BZ2_bzWriteClose(&bzerror, bz_file, 0, nullptr, nullptr);
+    fclose(file);
+  }
+  inline void write(void* data, size_t size) {
```
pd0wm, line 33: Why is `write` declared as `inline`? This doesn't lead to any performance benefits and should only be done if required for correct linkage.

selfdrive/loggerd/logger.h
```diff
@@ -15,18 +17,43 @@ const std::string LOG_ROOT = util::getenv_default("HOME", "/.comma/media/0/reald
...
+    BZ2_bzWrite(&bzerror, bz_file, data, size);
+    if (bzerror != BZ_OK && log_err) {
+      LOGE("BZ2_bzWrite error, bzerror=%d", bzerror);
+      log_err = false;
+    }
+  }
+  inline void write(kj::ArrayPtr<capnp::byte> array) { write(array.begin(), array.size()); }
+
+ private:
+  bool log_err = true;
```
adeebshihadeh, line 44: let's give this a better name, it's pretty confusing

selfdrive/loggerd/logger.h
```diff
@@ -15,18 +17,43 @@ const std::string LOG_ROOT = util::getenv_default("HOME", "/.comma/media/0/reald
 
 #define LOGGER_MAX_HANDLES 16
 
+class BZFile {
+ public:
+  BZFile(const char* path) {
+    assert(file = fopen(path, "wb"));
```
pd0wm, line 23: Never put code inside an assert statement. If you compile with `-DNDEBUG` it would not open the file anymore.

selfdrive/loggerd/logger.h
```diff
@@ -15,18 +17,43 @@ const std::string LOG_ROOT = util::getenv_default("HOME", "/.comma/media/0/reald
...
+    int bzerror;
+    bz_file = BZ2_bzWriteOpen(&bzerror, file, 9, 0, 30);
+    assert(bzerror == BZ_OK);
+  }
+  ~BZFile() {
+    int bzerror;
+    BZ2_bzWriteClose(&bzerror, bz_file, 0, nullptr, nullptr);
```
pd0wm, line 30: Check error?

deanlee: assert or loge?

pd0wm: LOGE is fine

selfdrive/loggerd/logger.h
```diff
@@ -15,18 +17,44 @@ const std::string LOG_ROOT = util::getenv_default("HOME", "/.comma/media/0/reald
...
+    assert(bzerror == BZ_OK);
+    assert(fclose(file) == 0);
```
adeebshihadeh, line 32: same as above comment, move the fclose out of the assert

PR #19960, opened 2021-01-29
Timezoned

pd0wm:


selfdrive/timezoned.py
```diff
@@ -0,0 +1,57 @@
+#!/usr/bin/env python3
+import time
+import json
+import subprocess
+from common.params import Params
+from selfdrive.swaglog import cloudlog
+
+from selfdrive.hardware import TICI
+
+if TICI:
```
adeebshihadeh, line 10: I think as much as possible should be not EON, we could just get this in manager for PC

selfdrive/timezoned.py
```diff
@@ -0,0 +1,57 @@
...
+  from timezonefinder import TimezoneFinder # pylint: disable=import-error
+
+
+def main():
+  if not TICI:
+    return
```
adeebshihadeh, line 16: same as above, should be handled in manager

selfdrive/timezoned.py
pd0wm: The problem is that the import will fail on NEOS. Until we ship a new NEOS I'd like to gate the import.

PR #19964, opened 2021-01-29
boardd: delete params "CarVin"&"CarParams" in safety_setter_thread

deanlee:
According to the current logic,  it may be better to delete them in the `safety_setter_thread()`  to ensure the integrity of the operation


pd0wm:
Why could they not be good right now?

It's important that the fields are cleared before manager detects ignition. Clearing them in the safety setter thread could lead to a race condition where manager starts services before the params are cleared.

deanlee:
i see

PR #19965, opened 2021-01-29
Panda: remove lock in destructor

deanlee:
panda is deleted after` threads.join(),` we don't need lock in destructor.


adeebshihadeh:
It's not worth making assumptions about how the Panda class is used in order to save a couple of lines.

PR #19973, opened 2021-01-30
Openpilot build readme

iejMac:


grekiki:
@adeebshihadeh So is this getting merged? I'd kind of want to link to it from the documentations I'll be making on the wiki.

PR #19982, opened 2021-01-31
get_frame_image:  moved the global variables into function static

deanlee:


Tammychmarney:
Business permanently closed in ORANGEVIILE ONTARIO CANADA

selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -207,14 +201,20 @@ void fill_frame_data(cereal::FrameData::Builder &framed, const FrameMetadata &fr
...
 kj::Array<uint8_t> get_frame_image(const CameraBuf *b) {
+  static const int x_min = getenv("XMIN") ? atoi(getenv("XMIN")) : 0;
+  static const int y_min = getenv("YMIN") ? atoi(getenv("YMIN")) : 0;
+  static const int env_xmax = getenv("XMAX") ? atoi(getenv("XMAX")) : -1;
+  static const int env_ymax = getenv("YMAX") ? atoi(getenv("YMAX")) : -1;
+  static const int scale = getenv("SCALE") ? atoi(getenv("SCALE")) : 1;
+
   assert(b->cur_rgb_buf);
+
+  const int x_max = env_xmax != -1 ? env_xmax : b->rgb_width - 1;
```
adeebshihadeh, line 212: we have the CameraBuf, can we get rid of env_xmax and make it static like the min?

deanlee: It may not be possible to do this, the width and height of each camera are different.

PR #19987, opened 2021-02-01
Fork ssh toggle

ghost:
Run toggleSSH system calls on a non UI thread.


grekiki:
Have a better version ready.

PR #19994, opened 2021-02-02
logcatd:  fix bug: subsequent reads after the first always return empty

deanlee:
retry the `-EAGAIN` caused this bug.  the `errno` is `ENOENT` when `-EAGAIN` return, continuing to call `android_logger_list_read`  will only return the same error again.

we need to reopen&read log when `-EAGAIN` was returned in non blocking mode.

### DIFF
--- a/selfdrive/logcatd/logcatd_android.cc
+++ b/selfdrive/logcatd/logcatd_android.cc
@@ -5,57 +5,53 @@
 #include "common/util.h"
 #include "messaging.hpp"
 
-ExitHandler do_exit;
-log_time last_log_time = {};
-
-static void publish_log(PubMaster &pm) {
-  // setup android logging
-  logger_list *loggers =
-      last_log_time.tv_sec == 0
-          ? android_logger_list_alloc(ANDROID_LOG_RDONLY | ANDROID_LOG_NONBLOCK, 0, 0)
-          : android_logger_list_alloc_time(ANDROID_LOG_RDONLY | ANDROID_LOG_NONBLOCK, last_log_time, 0);
-  assert(loggers);
-
-  const log_id_t log_ids[] = {LOG_ID_MAIN, LOG_ID_RADIO, LOG_ID_SYSTEM, LOG_ID_CRASH, (log_id_t)5};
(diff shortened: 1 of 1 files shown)


selfdrive/logcatd/logcatd_android.cc
```diff
@@ -1,68 +1,62 @@
...
 #include "common/util.h"
 #include "messaging.hpp"
 
+ExitHandler do_exit;
+log_time last_log_time = {};
 
-int main() {
-  ExitHandler do_exit;
-
+static void publish_log(PubMaster &pm) {
```
adeebshihadeh, line 11: let's just keep this in the main function, should be simpler and we don't need the global

selfdrive/logcatd/logcatd_android.cc
```diff
@@ -1,68 +1,62 @@
...
-  assert(kernel_logger);
-
-  PubMaster pm({"androidLog"});
+  logger_list *loggers =
+      last_log_time.tv_sec == 0
+          ? android_logger_list_alloc(ANDROID_LOG_RDONLY | ANDROID_LOG_NONBLOCK, 0, 0)
+          : android_logger_list_alloc_time(ANDROID_LOG_RDONLY | ANDROID_LOG_NONBLOCK, last_log_time, 0);
+  assert(loggers);
+
+  const log_id_t log_ids[] = {LOG_ID_MAIN, LOG_ID_RADIO, LOG_ID_SYSTEM, LOG_ID_CRASH, (log_id_t)5};
```
adeebshihadeh, line 19: let's define LOG_ID_KERNEL

selfdrive/logcatd/logcatd_android.cc
```diff
@@ -1,68 +1,62 @@
...
+  android_logger_list_free(loggers);
+}
 
-    pm.send("androidLog", msg);
-  }
+int main() {
+  PubMaster pm({"androidLog"});
+  while (!do_exit) {
+    publish_log(pm);
+    if (do_exit) break;
```
adeebshihadeh, line 57: we can wait the 500ms

PR #19997, opened 2021-02-02
Don't call function in assert, prepreq for NDEBUG

deanlee:


pd0wm:
Thanks! I don't think we'll turn that flag on any time soon, but at least we don't get any unexpected surprises.

PR #19998, opened 2021-02-02
Qt add Terms & Conditions

ghost:
Add T&C


selfdrive/version.py
```diff
@@ -45,7 +45,7 @@ def get_git_remote(default: Optional[str] = None) -> Optional[str]:
 prebuilt = os.path.exists(os.path.join(BASEDIR, 'prebuilt'))
 
 training_version: bytes = b"0.2.0"
-terms_version: bytes = b"2"
+terms_version: bytes = b"3"
```
pd0wm, line 48: For testing? Remove so we don't accidentally merge it in.

grekiki: Well it is a new version :). Ok I'll remove it

selfdrive/ui/qt/offroad/onboarding.cc
```diff
@@ -121,7 +121,16 @@ void OnboardingWindow::updateActiveScreen() {
   }
 }
 
+void enableBacklight(){
```
adeebshihadeh, line 124: this shouldn't go in onboarding. ideally we want a nice HW abstraction in cpp like we have in python

adeebshihadeh: it should also be gated on QCOM2

pd0wm: It's also only a problem for testing when you keep restarting the ui. On boot the backlight is already set to max.

selfdrive/ui/qt/offroad/onboarding.cc
```diff
@@ -67,31 +72,26 @@ TrainingGuide::TrainingGuide(QWidget* parent) {
...
-    font-size: 75px;
-    border-radius: 10px;
-    background-color: #292929;
-  )");
-  main_layout->addWidget(terms, 1, 0, 1, -1);
-  main_layout->setRowStretch(1, 1);
-
+  QVBoxLayout *main_layout = new QVBoxLayout;
+  main_layout->setContentsMargins(40, 0, 40, 0);
+  QWebEngineView *view = new QWebEngineView(this);
```
pd0wm, line 77: Think this slows ui start a lot. We should only load chromium when the terms actually have to be displayed.

pd0wm: Just make sure the widget is never created if we're not going to show it.

grekiki: Fixed

selfdrive/ui/qt/offroad/onboarding.hpp
```diff
@@ -37,10 +40,15 @@ class OnboardingWindow : public QStackedWidget {
...
+  QPushButton* accept_btn;
 
 signals:
   void onboardingDone();
 
 public slots:
   void updateActiveScreen();
+
+private slots:
+  void scrollPosition(QPointF position);
```
adeebshihadeh, line 53: can we give this a more descriptive name?

pd0wm, line 53:
```suggestion
  void scrollPositionChanged(QPointF position);
```

selfdrive/ui/qt/offroad/onboarding.cc
```diff
@@ -145,10 +154,21 @@ OnboardingWindow::OnboardingWindow(QWidget *parent) : QStackedWidget(parent) {
...
+    QPushButton:disabled {
+      color: #777777;
+      background-color: #222222;
+    }
   )");
 
   updateActiveScreen();
 }
+
+void OnboardingWindow::scrollPosition(QPointF position){
```
pd0wm, line 169:
```suggestion
void OnboardingWindow::scrollPositionChanged(QPointF position){
```

selfdrive/ui/qt/offroad/onboarding.cc
```diff
@@ -70,31 +73,31 @@ TrainingGuide::TrainingGuide(QWidget* parent) {
...
+  accept_btn = new QPushButton("Scroll to accept");
+  accept_btn->setEnabled(false);
+  buttons->addWidget(accept_btn);
   QObject::connect(accept_btn, &QPushButton::released, [=]() {
     Params().write_db_value("HasAcceptedTerms", current_terms_version);
     updateActiveScreen();
   });
 
-  main_layout->addWidget(new QPushButton("Decline"), 2, 0);
+  QObject::connect(view->page(), SIGNAL(scrollPositionChanged(QPointF)), this, SLOT(scrollPosition(QPointF)));
```
pd0wm, line 96:
```suggestion
  QObject::connect(view->page(), SIGNAL(scrollPositionChanged(QPointF)), this, SLOT(scrollPositionChanged(QPointF)));
```

PR #20000, opened 2021-02-02
Zookeeper support

robbederks:
To test in CI that we didn't screw up any of the low power modes openpilot is supposed to use


robbederks:
it's talking to the zookeeper now, but ssh is not set up on that c2 yet

PR #20006, opened 2021-02-03
camerad: new struct CameraServerCtx

deanlee:
~make vipc_server global, don't passing it over a long chain of calls~
Added struct CameraServerCtx,passing it over functions

### DIFF
--- a/selfdrive/camerad/cameras/camera_frame_stream.cc
+++ b/selfdrive/camerad/cameras/camera_frame_stream.cc
@@ -75,9 +75,9 @@ CameraInfo cameras_supported[CAMERA_ID_MAX] = {
   },
 };
 
-void cameras_init(const CameraServerCtx &ctx) {
-  camera_init(ctx, &ctx.cameras->rear, CAMERA_ID_IMX298, 20);
-  camera_init(ctx, &ctx.cameras->front, CAMERA_ID_OV8865, 10);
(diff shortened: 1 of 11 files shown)


adeebshihadeh:
Hmm, I don't think we really want to introduce a global here. Maybe we should make a struct for the common things passed around in camerad and put it there like we do for UI and loggerd?

deanlee:
good idea

selfdrive/camerad/cameras/camera_frame_stream.cc
```diff
@@ -73,11 +73,9 @@ CameraInfo cameras_supported[CAMERA_ID_MAX] = {
   },
 };
 
-void cameras_init(VisionIpcServer *v, MultiCameraState *s, cl_device_id device_id, cl_context ctx) {
-  camera_init(v, &s->rear, CAMERA_ID_IMX298, 20, device_id, ctx,
-              VISION_STREAM_RGB_BACK, VISION_STREAM_YUV_BACK);
-  camera_init(v, &s->front, CAMERA_ID_OV8865, 10, device_id, ctx,
-              VISION_STREAM_RGB_FRONT, VISION_STREAM_YUV_FRONT);
+void cameras_init(const CameraServerCtx &ctx, MultiCameraState *s) {
```
adeebshihadeh, line 76: the multi camera state is already part of ctx

PR #20008, opened 2021-02-03
panda: check do_exit in handle_usb_issue

deanlee:


pd0wm:
Will integrate this into: https://github.com/commaai/openpilot/pull/20061

PR #20020, opened 2021-02-04
draw_panda_metric: use ternary operator

deanlee:


adeebshihadeh:
I think the if statement is cleaner than these multi line ternaries.

PR #20021, opened 2021-02-04
Toyota interface.py: Create separate section for custom long tune

briskspirit:
Moving custom long tune to separate section, so it will be easier to add more models to that custom tune.


pd0wm:
Put the other longitudinal tuning in the else. Also call it something like "new" or "improved" tuning.

briskspirit:
Done! Thank you for teaching me how to do the right way!

PR #20034, opened 2021-02-07
loggerd: remove the segment variable from encoders

deanlee:
encoder's segment is always  `rotate_state.cur_seg,` no need to save a `segment` parameter in the encoder class.
In addition,`RawLogger:: encode_frame` does not assign a value to the parameter `*frame_segment`. that should be a bug.


adeebshihadeh:
Nice find :)

PR #20037, opened 2021-02-08
hyundai: set start accel

gregjhogan:
desired accel was zero when longitudinal mpc is in starting state (no effect before hyundai has longitudinal support)


pd0wm:
Fixed by https://github.com/commaai/openpilot/commit/7dc4b22dae50c727e8c270fe03344c6e3af2ab29

PR #20047, opened 2021-02-09
loggerd: removed variable std::vector<SubSocket*> socks from main()

deanlee:
`SubSocket* `  is already saved in map `qlog_states` as key

### DIFF
--- a/selfdrive/loggerd/loggerd.cc
+++ b/selfdrive/loggerd/loggerd.cc
@@ -332,7 +332,7 @@ int main(int argc, char** argv) {
   typedef struct QlogState {
     int counter, freq;
   } QlogState;
-  std::map<SubSocket*, QlogState> qlog_states;
+  std::map<SubSocket*, QlogState> sock_to_qs_map;
 
   s.ctx = Context::create();
   Poller * poller = Poller::create();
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
I like this, but let's rename the qlog state struct to something more appropriate given its new use.

selfdrive/loggerd/loggerd.cc
```diff
@@ -332,11 +332,10 @@ int main(int argc, char** argv) {
   typedef struct QlogState {
```
adeebshihadeh, line 332: I meant rename this struct to something like SocketState

selfdrive/loggerd/loggerd.cc
```diff
@@ -332,11 +332,10 @@ int main(int argc, char** argv) {
...
     int counter, freq;
```
adeebshihadeh, line 333: s/freq/qlog_freq

selfdrive/loggerd/loggerd.cc
```diff
@@ -332,11 +332,10 @@ int main(int argc, char** argv) {
...
   } QlogState;
-  std::map<SubSocket*, QlogState> qlog_states;
+  std::map<SubSocket*, QlogState> sock_to_qs_map;
```
adeebshihadeh, line 335: we can just call this sockets

selfdrive/loggerd/loggerd.cc
deanlee: keep a Subcket* s in struct too?

PR #20062, opened 2021-02-11
Camerad fixes

pd0wm:

### DIFF
--- a/selfdrive/camerad/cameras/camera_qcom2.cc
+++ b/selfdrive/camerad/cameras/camera_qcom2.cc
@@ -732,11 +732,14 @@ static void camera_open(CameraState *s) {
   // config csiphy
   LOG("-- Config CSI PHY");
   {
+
     uint32_t cam_packet_handle = 0;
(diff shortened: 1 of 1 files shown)


selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -177,10 +176,10 @@ void sensors_i2c(struct CameraState *s, struct i2c_random_wr_payload* dat, int l
   pkt->header.op_code = op_code;
   struct cam_cmd_buf_desc *buf_desc = (struct cam_cmd_buf_desc *)&pkt->payload;
 
-  buf_desc[0].size = buf_desc[0].length = sizeof(struct cam_cmd_i2c_random_wr) + (len-1)*sizeof(struct i2c_random_wr_payload);
+  buf_desc[0].size = buf_desc[0].length = sizeof(struct i2c_rdwr_header) + len*sizeof(struct i2c_random_wr_payload);
```
pd0wm, line 179: This is actually the same, but think it's cleaner this way.

PR #20064, opened 2021-02-11
cameras_qcom: refactor focus

deanlee:
Consistent with the original code, the following refactor was performed:
-  ~Move focusing related code into class `RearCameraFocus`~
-  Use a single std::mutex to guarantee that the focusing procedure are real atomic and thread safe.(The code before refactoring is not,std::atomic can only guarantee that a single variable is atomic,not the whole procedure )

### DIFF
--- a/selfdrive/camerad/cameras/camera_qcom.cc
+++ b/selfdrive/camerad/cameras/camera_qcom.cc
@@ -842,6 +842,8 @@ static void parse_autofocus(CameraState *s, uint8_t *d) {
     printf("%2.2X ", d[i]);
   }*/
 
+  std::lock_guard lk(s->focus_lock);
+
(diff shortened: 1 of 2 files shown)


deanlee:
For the convenience of review, does not encapsulate the focus-related code into class in this PR.

deanlee:
@ZwX1616 : It works great on my device, is there anything else that needs to be improved in this PR?

deanlee:
For making the code clearer, we may need to use a class to encapsulate related operations

ZwX1616:
We currently don't have any solid regression test for AF, so this is very hard to test and risky to merge. I doubt if we ever would add test for AF in the future. I suggest we just keep the AF code as is, thoughts? @adeebshihadeh

PR #20071, opened 2021-02-13
Fix division by zero in lateral_planner.py

briskspirit:
lateral_planner.py:191: RuntimeWarning: invalid value encountered in double_scalars

When v_ego = 0 and OP activated we are getting division by 0 which sets value to nan, so param angleSteers becomes set to nan and could potentially break other code.

How to reproduce: TSS2 Corolla, shift D, enable brake hold, enable ACC, enable OpenPilot. tmux will throw an error and ZMQ will publish lateralPlan with param angleSteers: nan while car is not moving.


briskspirit:
Or != 0 would be better? Sometimes we are getting negative v_ego

HaraldSchafer:
Hmmm, I don't really like that logic. Since if v_ego is super small it can potentially output weird things. 

I think the best is to replace v_ego with max(v_ego, 1e-1). Can you do that?

briskspirit:
Done

briskspirit:
Think I need to add a new reference

PR #20074, opened 2021-02-13
Tesla AP2 port

robbederks:
Port for AP2 model S

Works well, but will be merged as dashcam only due to not having user steer torque blending.

**Checklist**
- [x] added to README
- [x] test route added to [test_routes](../../selfdrive/test/test_routes.py)

### DIFF
--- a/release/files_common
+++ b/release/files_common
@@ -170,6 +170,13 @@ selfdrive/car/mazda/radar_interface.py
 selfdrive/car/mazda/values.py
 selfdrive/car/mazda/carcontroller.py
 selfdrive/car/mazda/mazdacan.py
+selfdrive/car/tesla/__init__.py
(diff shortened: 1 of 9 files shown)


adeebshihadeh:
Should we get this in as dashcam only for now?

robbederks:
> Should we get this in as dashcam only for now?

Yes, it's pretty stable now. I very rarely still fault my cruise control with disengage on gas (like once every 100+ times, pretty rare), haven't figured out why that happens yet.
For a dashcam port, do we also need the routes for in CI?

adeebshihadeh:
I usually get them to make sure we have a preserved route for the car as a reference, but not necessary in this case.

robbederks:
Ok, will get this cleaned up and merged.

PR #20077, opened 2021-02-14
agnos 0.6

adeebshihadeh:


selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -267,14 +268,12 @@ void AdvancedNetworking::toggleTethering(int enable) {
 }
 
 void enableSSH(Toggle* toggle_switch_SSH){
-  system("sudo systemctl enable ssh");
-  system("sudo systemctl start ssh");
+  Params().write_db_value("SshEnabled", "1");
```
pd0wm, line 271: Shouldn't we keep the `systemctl start`? Do we really need a param for this? Now there is state about SSH in two places (the param and the systemd symlink) which can get out of sync.

adeebshihadeh: the symlink doesn’t work on a ro rootfs, so we have a systemd service that triggers on the ssh param changing

pd0wm: If there the param is the only source of truth for ssh being on it's fine!

Issue #20075, opened 2021-02-13
Define locations for file storage in a configuration file, instead of using python's __file__ path


wolterhv:
Currently, there are about 17 lines spread among different files in the openpilot codebase where something similar to the following occurs:

```python
SOME_DIR = os.path.dirname(os.path.abspath(__file__))
```

which is merely a definition of a path, probably for storing files or reading previously stored ones. The benefit of this is probably being able to define paths dynamically, without breaking the system if something is moved from one place to another. 

There is however a cleaner way to do this, which is defining these locations in a configuration file instead, like an INI file or a YAML file. Then all the algorithms which need access to these paths will be able to get them by reading the configuration file. There are plenty of lightweight configuration-file-reading libraries out there which could be used to accomplish this.

I'm interested in hearing thoughts on this enhancement request.

adeebshihadeh:
This doesn't sound cleaner, but maybe I'm misunderstanding what you mean. Open a PR with an example if you'd like to discuss further.

wolterhv:
Well yes, I guess "cleaner" is also a matter of opinion. I have created a very rough PR to better illustrate what I propose. Let me know what you think.

PR #20078, opened 2021-02-14
Showcase prototype implementation of enhancement #20075 to get paths from configuration file

wolterhv:
**Description** 

This PR is merely for showcasing what I mean with the enhancement request #20075, it is not meant to be actually merged to master.

### DIFF
--- a/common/configreader.py
+++ b/common/configreader.py
@@ -1,0 +1,74 @@
+#!/bin/env python3
+"""
+This is a demo script which is supposed to be called by any module to retrieve
+a particular path from a config.ini file which defines a set of relevant paths.
+
+Not huge care has been taken to make it work properly since it's just a demo
+(albeit not a working one).
+"""
+
+import os
+import sys
+import configparser 
+
+
+_THISPATH = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
+
+class ConfigReader:
+
+  def __init__(self):
+    # SEARCHPATH is the set of paths where the config.ini file may be found
+    SEARCHPATH = []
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
What problem does this solve? It’s adding 170 lines for something that’s more brittle.

wolterhv:
Well, to be fair, it's not 170 lines of new brittle code. 70 of them belong to a hurriedly coded, with much room for improvement, config file reading class, and 86 to a heavily-commented config file.

What problem does it solve? Mainly it just centralizes location definitions. Currently, paths to relevant files are spread all around the code. In my opinion, this could become a problem in the future, specially since some of the paths to a single object are defined in multiple places. I guess this is where my opinion of that not being clean and yours of it being better than what I propose diverge.

Secondarily, it allows defining locations for data which are separate from the application source code. I tend to think that keeping logic, configuration and generated data separate is a good idea.

I understand if you think this is not necessarily or really helpful at the current state of things, but just wanted to share the idea. Thanks for looking at it anyway!

adeebshihadeh:
> What problem does it solve? Mainly it just centralizes location definitions.

For many of these, the definition is __file__. This really isn't cleaner.

If you'd like to contribute, I'd suggest looking at issues tagged "enhancement" or "good first issue".

wolterhv:
Fair enough. Thanks for your suggestion about contributing. 

How do you currently define paths in CPP sources, where in python you would use the `__file__` attribute? From what I could scan, mostly absolute paths are defined, either pointing to the user's comma config at `~/.comma` or at locations like `/data/` and `/persist/`.

PR #20080, opened 2021-02-15
INDI: only allow wind down on user override

pd0wm:


VirtuallyChris:
Tested on the loop - Big improvements to user override. no regressions.

PR #20081, opened 2021-02-15
Add fingerprint for 2018 Nissan Leaf G (Japan)

robin-reckmann:
Adds the fingerprint for the 2018 Nissan Leaf G (japanese version), which has the ADAS ECU behind the instrument cluster.


pd0wm:
What's your dongle id?

robin-reckmann:
fd3b0ef05a984a10

pd0wm:
I see a whole bunch of messages in a drive from your device that are not in a fingerprint. They are probably low frequency, but maybe good to add them all so it also fingerprints if the car is already started:

Output from `compare_fingerprints.py` comparing to the current fingerprint:
```
1792 not in f2
1872 not in f2
1937 not in f2
1953 not in f2
1968 not in f2
1988 not in f2
2000 not in f2
2001 not in f2
2004 not in f2
2005 not in f2
2015 not in f2
2016 not in f2
2017 not in f2
```

PR #20088, opened 2021-02-16
added Toyota Corolla Touring Sport 2021 engine fwVersion

ErikBjare:
I have a new Toyota Corolla Touring Sport 2021 (Hybrid) with the 2L engine (manufactured in UK) that wasn't recognized. I followed https://github.com/commaai/openpilot/wiki/Fingerprinting and here's the result.

I'm going to test it now, so stand by as I do (should be done in an hour or two).

Edit: Working great, thanks for an amazing piece of hardware & software!


pd0wm:
What's your dongle id?

ErikBjare:
@pd0wm 9c5057e6d155baaa

pd0wm:
Verified!

PR #20093, opened 2021-02-16
up tici hevc bitrate

ZwX1616:


selfdrive/loggerd/tests/test_encoder.py
```diff
@@ -27,9 +27,9 @@
   ]
 else:
   CAMERAS = [
-    ("fcamera.hevc", 20, FULL_SIZE),
-    ("dcamera.hevc", 20, FULL_SIZE),
-    ("ecamera.hevc", 20, FULL_SIZE),
+    ("fcamera.hevc", 20, FULL_SIZE*2),
+    ("dcamera.hevc", 20, FULL_SIZE*2),
+    ("ecamera.hevc", 20, FULL_SIZE*2),
```
adeebshihadeh, line 32: this is confusing, i'd just define separate eon and tici full size

PR #20102, opened 2021-02-18
Toyota: simplify angle offsetting

sshane:
We can save some lines and make the code a bit simpler if we combine the two steer angle > 1e-3 checks into one


selfdrive/car/toyota/carstate.py
```diff
@@ -53,10 +51,9 @@ def update(self, cp, cp_cam):
 
     if self.accurate_steer_angle_seen:
       ret.steeringAngleDeg = cp.vl["STEER_TORQUE_SENSOR"]['STEER_ANGLE'] - self.angle_offset
-
       if self.needs_angle_offset:
         angle_wheel = cp.vl["STEER_ANGLE_SENSOR"]['STEER_ANGLE'] + cp.vl["STEER_ANGLE_SENSOR"]['STEER_FRACTION']
-        if abs(angle_wheel) > 1e-3 and abs(ret.steeringAngleDeg) > 1e-3:
+        if abs(angle_wheel) > 1e-3:
```
ShaneSmiskol, line 56: Could even remove this line too since this signal is usually always available immediately, or at least before `STEER_TORQUE_SENSOR` is, and we now wait for that.

pd0wm: That's not guaranteed right? Just depends on when you turn on openpilot in the message cycle.

ShaneSmiskol: https://imgur.com/a/UgwpzxQ It seems like data from STEER_ANGLE_SENSOR is available from the very start, plus the torque message always takes a while so it's probably pretty safe if we want to remove it.

pd0wm: Not worth removing for the edge case where it might cause issues.

PR #20106, opened 2021-02-18
loggerd: added funtion to initXXXEncoderIdx

deanlee:


adeebshihadeh:
This should just use capnproto's dynamic API to init. Include the encode idx event in LogCameraInfo and use that to init.

adeebshihadeh:
Closing this since it's small and needs to be rewritten to reduce the number of open PRs

PR #20114, opened 2021-02-19
qt-UI close settings on onroad transition

ghost:
Should close setting when going onroad


grekiki:
Should fix https://github.com/commaai/agnos-builder/issues/101

PR #20117, opened 2021-02-19
Qt-UI Close settings when screen turns off 

ghost:


selfdrive/ui/qt/home.cc
```diff
@@ -143,6 +143,7 @@ HomeWindow::HomeWindow(QWidget* parent) : QWidget(parent) {
   home = new OffroadHome();
   layout->addWidget(home, 0, 0);
   QObject::connect(glWindow, SIGNAL(offroadTransition(bool)), this, SLOT(setVisibility(bool)));
+  QObject::connect(glWindow, &GLWindow::screen_shutoff, [=](){emit closeSettings();});
```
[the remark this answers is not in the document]
grekiki: Should be close enough, I think.

selfdrive/boardd/panda.cc
```diff
@@ -17,12 +17,12 @@ void panda_set_power(bool power){
   err += gpio_init(GPIO_STM_RST_N, true);
   err += gpio_init(GPIO_STM_BOOT0, true);
 
-  err += gpio_set(GPIO_STM_RST_N, true);
+  err += gpio_set(GPIO_STM_RST_N, false);
```
adeebshihadeh, line 20: why are there panda changes in a UI PR?

grekiki: That was an oversight... My bad will fix.

PR #20129, opened 2021-02-22
Initialize lateralTuning

ErichMoraga:
Present in all other `interface.py` except this one.


pd0wm:
As far as I can tell only toyota explicitly sets `pid`

PR #20132, opened 2021-02-22
Fix tici powerdown and add support for forcing

robbederks:
Should've been 2 PRs, but the history is interwoven now


robbederks:
note that forcing only works if the device has gone onroad at least once

PR #20133, opened 2021-02-22
Always linearize, better way of compensating for lag

haraschax:


HaraldSchafer:
I tested it in batch in sim and it performs better, for cars with low and high lag. I also looked at individual segments and it looked fine. Improved some problematic segments.

PR #20148, opened 2021-02-24
Qt-UI add driver monitoring view in offroad

ghost:

### DIFF
--- a/selfdrive/ui/qt/home.cc
+++ b/selfdrive/ui/qt/home.cc
@@ -160,7 +160,13 @@ void HomeWindow::setVisibility(bool offroad) {
 
 void HomeWindow::mousePressEvent(QMouseEvent* e) {
   UIState* ui_state = &glWindow->ui_state;
-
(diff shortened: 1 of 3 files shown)


selfdrive/ui/qt/home.cc
```diff
@@ -160,7 +160,13 @@ void HomeWindow::setVisibility(bool offroad) {
 
 void HomeWindow::mousePressEvent(QMouseEvent* e) {
   UIState* ui_state = &glWindow->ui_state;
-
+  if (GLWindow::ui_state.started && GLWindow::ui_state.scene.frontview) {
+    if(Params().get("IsDriverViewEnabled") == "1"){ // Not needed right now, might avoid some bugs in the
```
adeebshihadeh, line 164: comment isn't complete

adeebshihadeh: though if frontview is true, I'd just always write the param

PR #20149, opened 2021-02-24
Mention 2021 prius prime support on 0.8.2 release notes

aronasorman:
[This PR adds the fingerprint for the 2021 Prius Prime](https://github.com/commaai/openpilot/pull/19651/files).

But said PR did not include the corresponding change to release notes. I've been fielding questions from friends asking when support for this would come in. This change mentions that support has been added on 0.8.2!

Also this is a minor change that doesn't seem to match any of the templates. Let me know if I need to change the description.


pd0wm:
0.8.2 is already too far in the release process. It's already added to the readme and the website, so that should be enough indication that it's supported.

PR #20154, opened 2021-02-25
Add `, Shift+S key bind to save tmux scrollback

sshane:
The current `capture-pane` line only saved up to then, it wouldn't ever write what happened after openpilot/manager actually started. This makes it a bit easier for the user to save the log to a file whenever they're trying to get help with an issue from community members by pressing the prefix (`) then Shift+S, then the Enter key.

### DIFF
--- a/launch_chffrplus.sh
+++ b/launch_chffrplus.sh
@@ -228,8 +228,8 @@ function launch {
     tici_init
   fi
 
-  # write tmux scrollback to a file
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
I don't want to put any configuration like this in the launch script. You can find all the cloudlogged messages in the rlogs. If you want still want this, make a PR against eon-neos-builder to the tmux configuration file.

PR #20152, opened 2021-02-25
Update TSS1 Corolla minCanSpeed and steerRatio

sshane:
openpilot has settled at around 17.4 (`{"carFingerprint": "TOYOTA COROLLA 2017", "steerRatio": 17.431434631347656, "stiffnessFactor": 0.999917209148407, "angleOffsetAverageDeg": 0.20000874996185303}`) for me, which is closer to the factory spec of 17.8 (found [here](https://www.socalpreowned.net/vehicle-details/2017-toyota-corolla-l-sedan-5e07d0f3d980184aa8c519977a67e0a9) and [here](https://www.anmautosales.com/vehicle-details/2017-toyota-corolla-se-sedan-a3c926baf2fd8349a2ed8423971c5429)) compared to what's there now.

For `minSpeedCan` I found the smallest value reported by any wheel before it drops to 0:

Screenshot_185
Screenshot_186

PR #20156, opened 2021-02-25
Toyota: Only use gas interceptor under 19 mph, convert accel to gas when needed

sshane:
- steerRatio and minSpeedCan changes explanation: #20152

Previously I changed the compute_gb function in the Toyota interface dynamically if a pedal was detected, but I realized that the clip in longcontrol was for accel not gas, so if it clipped it to 0.5, it's not 50% of max accel, it's 50% max gas percentage which is wrong. Now, accel is clipped as usual to 50%, and the gas percentage is clipped from 0 to 100%.

How the transition works from accel to gas: I have a simple function that's based off data of coasting at various speeds from 0 to 19 mph, and it switches between sending accel and gas based off if the desired accel is greater than what the coast acceleration would be at that speed.
So for example since coast accel is positive from 0 to ~2.5 m/s, if des accel is less than that (0.2 m/s/s is less than 0.5 m/s/s let's say) we send accel letting the car coast up, applying a small amount of brake to keep us lower than the positive coast accel at those speeds. At above 2.5 m/s it's the opposite with negative coast accel.
It might be car specific though, since I know some hybrids slow down much quicker when you're coasting, so I'm not sure how to handle that.

Todos:
- [x] Fix braking momentarily when engaging at low speeds.
- [ ] Test briskspirit's new long tune (braking is still a bit jerky even though we're using acc and stock non-pedal tuning)


selfdrive/car/toyota/carcontroller.py
```diff
@@ -49,15 +61,15 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, hud_alert,
...
-      # send only negative accel if interceptor is detected. otherwise, send the regular value
+    if CS.CP.enableGasInterceptor and enabled and CS.out.vEgo < MIN_ACC_SPEED and \
+            apply_accel * CarControllerParams.ACCEL_SCALE > coast_accel(CS.out.vEgo):
       # +0.06 offset to reduce ABS pump usage when OP is engaged
-      apply_accel = 0.06 - actuators.brake
-    else:
-      apply_accel = actuators.gas - actuators.brake
+      # converts desired acceleration to gas percentage for pedal
+      apply_gas = clip(compute_gb_pedal(apply_accel * CarControllerParams.ACCEL_SCALE, CS.out.vEgo), 0., 1.)
+      apply_accel = 0.06
```
ShaneSmiskol, line 72: Do we need to send 0.06 * 3? Or could it be even a higher number? Coast accel under 5 mph is about 0.5 m/s/s so it could brake when we're sending 0% gas. Setting it above `.5/3` might fix it from braking once the gas pedal is released and we want to truly coast

selfdrive/car/toyota/carcontroller.py
```diff
@@ -49,15 +61,15 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, hud_alert,
...
+    apply_accel = actuators.gas - actuators.brake
 
-    apply_gas = clip(actuators.gas, 0., 1.)
-
-    if CS.CP.enableGasInterceptor:
-      # send only negative accel if interceptor is detected. otherwise, send the regular value
+    if CS.CP.enableGasInterceptor and enabled and CS.out.vEgo < MIN_ACC_SPEED and \
+            apply_accel * CarControllerParams.ACCEL_SCALE > coast_accel(CS.out.vEgo):
       # +0.06 offset to reduce ABS pump usage when OP is engaged
-      apply_accel = 0.06 - actuators.brake
```
ShaneSmiskol, line 58: We might need to still apply 0.06 offset only when applying brake, since it's a bit jerky for me. I'll test and update if it's better

selfdrive/car/toyota/interface.py
```diff
@@ -90,7 +90,8 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       stop_and_go = False
       ret.safetyParam = 88
       ret.wheelbase = 2.70
-      ret.steerRatio = 18.27
+      ret.steerRatio = 17.43
```
pd0wm, line 93: Can we put the steerRatio in a different PR? Doesn't have to do with longitudinal.

ShaneSmiskol: You got it!

selfdrive/car/toyota/carcontroller.py
pd0wm, line 72: Why not 0?

pd0wm: And then put the next two lines in the else?

ShaneSmiskol: The hysteresis and clipping lines in the else? And I was confused about the original reason it was added for offsetting the brake, you think we don't need it?

The only thing is, if the function wants to apply 0% gas to coast up at a low speed and desired accel is 0, the car might brake since the coast accel at that low speed (0-2.5m/s) is positive.

pd0wm: Can you check if it runs the pump when you send exactly 0? If not then we should just use that.

ShaneSmiskol: I think it changes based on speed, but will do! For example if we're at less than 1 or 2 mph, but still moving, sending 0 accel will make the car brake to a stop for some reason. I'll make sure though. Also how does the new change look? I thought we should be sending the function what a TSS2 Toyota would get, the accel after it goes through the hysteresis and clipping steps.

pd0wm: At some speed the PCM will decide that's time to stop and will just brake no matter what you do (unless you send very high accel). So 0 or 0.06 is probably not going to make a difference for that.

pd0wm: New change looks good. Think in this case we just want to make sure it sends zero, otherwise the hysteresis could send some non zero value and you could have gas/brake fighting.

pd0wm:
Let me know when some people have tested this. 

Also I can't easily check the values for `coast_accel`, but think it's a little weird that the highest speed has less decel than the second to highest speed.

ShaneSmiskol:
@pd0wm Here's the data I gathered where I just let it coast at various speeds (this is offsetting a_ego, bringing it forward in time by about half a second to account for delay). Low speed could be a bit higher, I don't know why it's not in the middle.

coasting plot

ShaneSmiskol:
New function is smoother at low speed but incorrectly ramps down gas as accel raises above ~1m/s/s or so, making 5-20 mph times way too slow. I'll try to fix that with the final function

ShaneSmiskol:
Latest coast function and the logic to always send offsetted brake under the min acc speed almost completely fixes it braking for a second when engaging at low speed for me

selfdrive/car/toyota/carcontroller.py
```diff
@@ -10,33 +10,33 @@
...
+  return interp(speed, *zip(*points))
 
-  return accel, accel_steady
 
+def compute_gb_pedal(accel, speed, braking):
+  def accel_to_gas(a_ego, v_ego):
+    _a3, _a4, _a5, _a6, _a7, _a8, _s1, _s2, _s3, _s4, _offset = [-0.0034109221270790142, -0.02989942810035373, 0.002005326552420498, 0.1356381902353583, 0.0014222019070588158, 0.008436894892099946,
+                                                                 0.0009439048033890968, -0.0017786568461504919, 0.002986433642380856, 0.021810785976030644, -0.007020501995388009]
+    speed_part = (_s1 * a_ego + _s2) * v_ego ** 2 + (_s3 * a_ego + _s4) * v_ego
+    accel_part = (_a7 * v_ego + _a8) * a_ego ** 4 + (_a3 * v_ego + _a4) * a_ego ** 3 + _a5 * a_ego ** 2 + _a6 * a_ego
```
ShaneSmiskol, line 25: I'll merge the values into the calculations before merging

selfdrive/car/toyota/carcontroller.py
```diff
@@ -10,33 +10,33 @@
...
-            [3.443, -0.249], [MIN_ACC_SPEED, -0.145]]
-  return interp(speed, *zip(*points))
+  coast = coast_accel(speed)
+  spread = 0.2
 
+  if accel >= coast:
+    gas = accel_to_gas(accel, speed)
+    x = accel - coast  # start at 0 when at coast accel
+    if spread > x >= 0:  # ramp up gas output using s-shaped curve from coast accel to coast + spread
+      gas *= 1 / (1 + (x / (spread - x)) ** -3) if x != 0 else 0
```
ShaneSmiskol, line 36: Squaring a linear function also could work (for slowly smoothly increasing gas near 0) to reduce complexity, but doesn't smoothly lessen how quickly it increases gas near 100%. This whole thing is for some sharp decrease in gas output approaching the coasting acceleration from the right which can't be captured in the polynomial function

selfdrive/car/toyota/carcontroller.py
```diff
@@ -10,33 +10,33 @@
...
+    return gas if not braking else gas / 2.0  # while car is braking reduce gas output
```
ShaneSmiskol, line 37: With the new brakeLights message I found we know if the car is actually actuating brakes for TSS1 Corolla (not deriving from aEgo or requested accel which is nice) so we reduce output until the car releases brakes or the pedal applies enough output to move us. This reduces jerking since if you apply gas, all brakes are released immediately from the car

selfdrive/car/toyota/carcontroller.py
```diff
@@ -64,17 +64,17 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, hud_alert,
...
-        apply_gas = clip(compute_gb_pedal(apply_accel * CarControllerParams.ACCEL_SCALE, CS.out.vEgo), 0., 1.)
-      apply_accel = 0.06 - actuators.brake
+      apply_gas = compute_gb_pedal(apply_accel, CS.out.vEgo, CS.out.brakeLights)
+      if apply_accel > 0 and CS.out.vEgo <= CS.CP.minSpeedCan:  # increase accel to release brake quicker todo: might not be needed, test with and without in similar situations
+        apply_accel *= 2
 
-    apply_accel, self.accel_steady = accel_hysteresis(apply_accel, self.accel_steady, enabled)
-    apply_accel = clip(apply_accel * CarControllerParams.ACCEL_SCALE, CarControllerParams.ACCEL_MIN, CarControllerParams.ACCEL_MAX)
+    apply_accel = clip(apply_accel, CarControllerParams.ACCEL_MIN, CarControllerParams.ACCEL_MAX) if enabled else 0.
+    apply_gas = clip(apply_gas, 0., 1.)
```
ShaneSmiskol, line 77: Finally I removed hysteresis since it doesn't really make much since as far as I understand and added a final `if enabled` to just make sure it never sends non zero if not engaged

selfdrive/car/toyota/interface.py
```diff
@@ -91,7 +91,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.safetyParam = 88
       ret.wheelbase = 2.70
       ret.steerRatio = 18.27
-      ret.minSpeedCan = 0.1 * CV.KPH_TO_MS
+      ret.minSpeedCan = 0.4 * CV.KPH_TO_MS
```
ShaneSmiskol, line 94: Must have found an outlier before, this is more common

selfdrive/car/toyota/carstate.py
```diff
@@ -129,6 +129,7 @@ def get_can_parser(CP):
       ("STEER_RATE", "STEER_ANGLE_SENSOR", 0),
       ("CRUISE_ACTIVE", "PCM_CRUISE", 0),
       ("CRUISE_STATE", "PCM_CRUISE", 0),
+      ("STANDSTILL_ON", "PCM_CRUISE", 0),
```
ShaneSmiskol, line 132: Only 1 when the car is braking and engaged

ShaneSmiskol:
Almost ready, just need to solve the car not releasing brakes soon enough causing a spike of gas. In data, I've seen the desired speed at a constant low value but the car just doesn't release brakes (it's below the coast accel at that speed). Multiplying the des accel for the car at a standstill might be the only way to release brakes quicker.

ShaneSmiskol:
This relies on the removed `brakeLights` signal, I'll have to figure something else out to get the car to release brakes sooner (previously was multiplying accel when car was braking). Could just define the minimum desired acceleration that gets the car to release brakes partially or fully, and send it when we're stopped and want to resume.

ShaneSmiskol:
Closing, will be revisited in the future

PR #20159, opened 2021-02-25
Qt-UI fix ssh username

ghost:


selfdrive/ui/qt/widgets/input_field.cc
```diff
@@ -25,7 +25,7 @@ InputDialog::InputDialog(QString prompt_text, QWidget *parent): QDialog(parent)
     background-color: #444444;
   )");
   header_layout->addWidget(cancel_btn, 0, Qt::AlignRight);
-  QObject::connect(cancel_btn, SIGNAL(released()), this, SLOT(reject()));
+  connect(cancel_btn, &QPushButton::released, [=](){done(QDialog::Rejected); emit cancel();});
```
pd0wm, line 28: Don't change the original behavior. Make sure reject is still emitted.

pd0wm:
Please test in all other places the keyboard is used too.

PR #20168, opened 2021-02-26
Qt: Use QWebView for terms & conditions

ghost:
Add QWebView


grekiki:
It will need some apt updates. Look at the ubuntu_setup.sh diff.

grekiki:
No browsers yet I guess

PR #20171, opened 2021-02-26
qt fixups

adeebshihadeh:


selfdrive/ui/qt/api.cc
```diff
@@ -78,17 +81,26 @@ QString CommaApi::create_jwt() {
...
+  qDebug() << "Device supports OpenSSL: " << QSslSocket::supportsSsl();
 
   QTimer* timer = new QTimer(this);
   QObject::connect(timer, &QTimer::timeout, [=](){sendRequest(requestURL, payloads);});
   timer->start(period_seconds * 1000);
 
   networkTimer = new QTimer(this);
   networkTimer->setSingleShot(true);
-  networkTimer->setInterval(20000); // 20s before aborting
+  networkTimer->setInterval(5000); // 20s before aborting
```
grekiki, line 94: Probably want to keep the comment correct or remove it?

adeebshihadeh: just testing stuff here :)

PR #20176, opened 2021-02-27
camerad: move the duplicate function declaration to camera_common.h

deanlee:


adeebshihadeh:
Nice, let's start cleaning up camerad with small-ish PRs like we did with loggerd

deanlee:
ok

PR #20180, opened 2021-02-27
camera_qcom: move opencl stuff into class LapConv 

deanlee:


ZwX1616:
Looks clean. Can you change the class name to LapConv or something? (the kernel does rgb -> gray -> conv, sorry for the bad naming)

ZwX1616:
Can you rebase? Should be just deleting a few duplicate lines

PR #20212, opened 2021-03-02
oneplus cleanup (continue)

deanlee:


selfdrive/camerad/cameras/camera_qcom.cc
```diff
@@ -109,9 +93,9 @@ static void camera_init(VisionIpcServer *v, CameraState *s, int camera_id, int c
   s->line_length_pclk = line_length_pclk;
   s->max_gain = max_gain;
   s->fps = fps;
-
   s->self_recover = 0;
 
+  s->apply_exposure = (camera_id == CAMERA_ID_IMX298) ? imx298_apply_exposure : ov8865_apply_exposure;
```
ZwX1616, line 98: imx298_apply_exposure defined after this?

ZwX1616: sensor_write_regs as well. But other than that lgtm

deanlee: my bad@@ fixed

PR #20218, opened 2021-03-02
merge the china golf 7.5 missing values in the existing fingerprint

dingyf12593:


adeebshihadeh:
What's your dongle ID?

dingyf12593:
> What's your dongle ID?

Dongle ID: [3f84146622df5f39]

PR #20231, opened 2021-03-03
Manager cleanup

pd0wm:


selfdrive/manager/manager.py
```diff
@@ -0,0 +1,220 @@
...
+    params.put("Passive", str(int(os.getenv("PASSIVE"))))
+
+  if params.get("Passive") is None:
+    raise Exception("Passive must be set to continue")
+
+  if EON:
+    update_apks()
+
+  manager_init(spinner)
+  manager_prepare(spinner)
```
adeebshihadeh, line 178: merge these? unclear how these are different

adeebshihadeh: the top half of this main function is like a third init too

selfdrive/manager/process.py
```diff
@@ -0,0 +1,225 @@
...
+  # exec the process
+  os.chdir(cwd)
+  os.execvp(pargs[0], pargs)
+
+
+def join_process(process, timeout):
+  # Process().join(timeout) will hang due to a python 3 bug: https://bugs.python.org/issue28382
+  # We have to poll the exitcode instead
+  t = time.time()
+  while time.time() - t < timeout and process.exitcode is None:
```
adeebshihadeh, line 52:
```suggestion
  t = time.monotonic()
  while time.monotonic() - t < timeout and process.exitcode is None:
```

selfdrive/test/test_manager.py
```diff
@@ -23,37 +26,44 @@ def tearDown(self):
 
   def test_manager_prepare(self):
     os.environ['PREPAREONLY'] = '1'
-    manager.main()
+
+    with Spinner() as spinner:
```
adeebshihadeh, line 30: can't it just default to none? or maybe make it's own spinner?

adeebshihadeh:
Let's also move the manager test to selfdrive/manager/

selfdrive/manager/process.py
pd0wm: Maybe with python 3.8 this whole function can go. Will look into that after merging.

selfdrive/manager/manager.py
pd0wm: Init and prepare are different. prepare calls prepare on all the proceses, but doesn't do anything to the state in manager.

PR #20234, opened 2021-03-03
loggerd:slice message buffer

deanlee:


pd0wm:
What's the benefit?

deanlee:
No, but it is reasonable to send a buffer with the correct length to MessageReader.

adeebshihadeh:
I don't think we need this. Isn't the length in the message header?

Issue #20072, opened 2021-02-13
Centralize definitions of mathematical and physical constants


wolterhv:
Currently, I've seen π or some approximation to it be defined in 5 files. This can be shown with the output of `grep -r '3\.14' $OPENPILOT_TOPLEVEL`:

```
./selfdrive/car/ford/carcontroller.py:        curvature = self.vehicle_model.calc_curvature(actuators.steerAngle*3.1415/180., CS.out.vEgo)
./selfdrive/car/mock/interface.py:LPG = 2 * 3.1415 * YAW_FR * TS / (1 + 2 * 3.1415 * YAW_FR * TS)
./selfdrive/controls/lib/lateral_mpc/generator.cpp:#define PI 3.1415926536
./selfdrive/modeld/thneed/include/msm_kgsl.h: * are cherry picked into other trees out of order so it is frozen as of 3.14.
./selfdrive/locationd/ublox_msg.cc:      double gpsPi = 3.1415926535898;
./selfdrive/locationd/test/ephemeris.py:    gpsPi = 3.1415926535898
```

Something similar happens with the gravitational acceleration constant g (often defined as 9.81 or even as 10) and probably with other constants as well. Naturally, this occurs in python source code as much as in c++ code.

I propose that all mathematical and physical constants be defined in a central location. It may be defined in a language-agnostic format and have language-specific files generated at build time for inclusion in source code, e.g. with m4. My proposal is that this file be located in `common/constants.txt` and the make procedure can generate therefrom `constants.py` and `constants.h`.

What are other opinions in regard to this? It's not a big problem, probably not a problem at all -- I don't think it could be causing any bugs. It can be solved pretty effortlessly though, and I'm willing to implement it.

pd0wm:
Don't think reading from a configuration file will make things easier. But we can definitely clean up all the definitions of pi and G.

wolterhv:
I didn't mean reading from a configuration file. I'll explain what I had in mind of generating language-specific source files with m4.

Suppose you wanted to define three constants: pi, Earth's gravitational acceleration and the Earth's radius. Your codebase is in multiple languages and you don't want to maintain more than one file. So you use m4 and create the following file, `constants.m4`:

```m4
dnl Macros
define(`constdef',`ifdef(`cpp',`#define $1 $2',`$1 = $2')')dnl
define(`comment',`ifdef(`cpp',`// $1',`# $1')')dnl
dnl File headers
ifdef(`cpp',dnl
[...]

PR #20235, opened 2021-03-03
Replace hardcoded values for Pi

wolterhv:
**Description:** Replaces re-definitions of π for their equivalents in `math` (for python) or `math.h` (for C++). See discussion in https://github.com/commaai/openpilot/issues/20072.

**Verification:** All automatic checks passed.

### DIFF
--- a/selfdrive/car/ford/carcontroller.py
+++ b/selfdrive/car/ford/carcontroller.py
@@ -2,6 +2,7 @@ from cereal import car
 from selfdrive.car import make_can_msg
 from selfdrive.car.ford.fordcan import create_steer_command, create_lkas_ui, spam_cancel_button
 from opendbc.can.packer import CANPacker
+import math
 
 
 MAX_STEER_DELTA = 1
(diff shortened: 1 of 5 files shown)


selfdrive/car/ford/carcontroller.py
```diff
@@ -2,6 +2,7 @@
 from selfdrive.car import make_can_msg
 from selfdrive.car.ford.fordcan import create_steer_command, create_lkas_ui, spam_cancel_button
 from opendbc.can.packer import CANPacker
+import math
```
adeebshihadeh, line 5: standard library imports should be at the top

selfdrive/car/mock/interface.py
```diff
@@ -5,12 +5,13 @@
 import cereal.messaging as messaging
 from selfdrive.car import gen_empty_fingerprint
 from selfdrive.car.interfaces import CarInterfaceBase
+import math
```
adeebshihadeh, line 8: same

selfdrive/controls/lib/lateral_mpc/generator.cpp
```diff
@@ -1,8 +1,8 @@
 #include <acado_code_generation.hpp>
 #include "common/modeldata.h"
+#include <math.h> // M_PI
```
adeebshihadeh, line 3: remove the comment

wolterhv:
Thanks for your comments and for fixing the title!

PR #20237, opened 2021-03-04
refactor LogCameraInfo

deanlee:
refactor `LogCameraInfo` to simplify the initialization of encoders according to the device type

### DIFF
--- a/selfdrive/loggerd/loggerd.cc
+++ b/selfdrive/loggerd/loggerd.cc
@@ -347,11 +347,11 @@ int main(int argc, char** argv) {
     poller->registerSocket(sock);
     socks.push_back(sock);
 
-    for (auto &cam_info : cameras_logged) {
-      if (strcmp(it.name, cam_info.frame_packet_name) == 0 &&
-          (cam_info.type != DRIVER_CAM || Params().read_db_bool("RecordFront"))) {
-        s.rotate_states.push_back(std::make_unique<RotateState>(sock, cam_info));
-      }
+    auto cam_info = std::find_if(std::begin(cameras_logged), std::end(cameras_logged), [it](auto &ci) { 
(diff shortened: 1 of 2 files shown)


selfdrive/loggerd/loggerd.cc
```diff
@@ -44,13 +45,11 @@ namespace {
 constexpr int MAIN_FPS = 20;
 
 #ifndef QCOM2
-constexpr int MAIN_BITRATE = 5000000;
-constexpr int MAX_CAM_IDX = LOG_CAMERA_ID_DCAMERA;
-constexpr int DCAM_BITRATE = 2500000;
+#define MAIN_BITRATE 5000000
```
adeebshihadeh, line 48: why change these to defines?

selfdrive/loggerd/loggerd.cc
```diff
@@ -70,7 +69,8 @@ LogCameraInfo cameras_logged[LOG_CAMERA_ID_MAX] = {
     .downscale = false,
     .has_qcamera = true
   },
-  [LOG_CAMERA_ID_DCAMERA] = {
+#if defined(QCOM) || defined(QCOM2)
```
adeebshihadeh, line 72: we should remove these ifdefs, loggerd shouldn't really care about this stuff. should be handled automatically based on what kind of frame it gets

selfdrive/loggerd/loggerd.cc
deanlee: avoid unused variables error

selfdrive/loggerd/loggerd.cc
deanlee: if so, The code will be more complicated than it is now.

PR #20238, opened 2021-03-04
VW MQB: Get steering rate-change sign from the correct signal

jyoung8607:

Fix a very old bug processing the steeringRateDeg signal from the car. It was reading the absolute rate correctly, but not the sign (positive/negative, left/right). This would cause steeringRateDeg to be inverted quite frequently. We got away with this for a long time because PID lateral control doesn't consume steeringRateDeg.

**Verification**

**Route**

2018 Volkswagen Golf R: cae14e88932eb364|2021-03-04--16-08-22


adeebshihadeh:
Suspicious that there's no differences in process replay, but looked into it and seems the car isn't moving in the segment we use. I'll update the test route with one from our A3. Thanks for the fix!

PR #20242, opened 2021-03-04
VW MQB: Add support for blind spot monitoring radar

jyoung8607:

If present, read the signals for "blind spot occupied/mirror indicator on" and "blind spot warning/mirror indicator flashing" from the SWA_01 CAN message. SWA is short for Spurwechselassistent, a.k.a. German Engineered™ Lane Change Assist. The indicator data is fed into carState for use in blocking openpilot's automated lane changes.

For message and signal validation, the opendbc repo already has support for SWA_01 CRC and counter checks. I have NOT added it to the CANParser `checks[]` at this time, because I can't assume all supported cars will have blind spot radars. Other OP supported cars are also missing `checks[]` entries for their BSM messages. In talking with @adeebshihadeh we decided it would be nice to add fingerprint-time BSM support detection at some point.

**Verification**

**Route**

2018 Volkswagen Golf R: cae14e88932eb364|2021-03-04--16-38-06


selfdrive/car/volkswagen/carstate.py
```diff
@@ -60,6 +60,12 @@ def update(self, pt_cp):
     # We use the speed preference for OP.
     self.displayMetricUnits = not pt_cp.vl["Einheiten_01"]["KBI_MFA_v_Einheit_02"]
 
+    # Consume blind-spot radar info/warning LED states, if available
+    ret.leftBlindspot = any([bool(pt_cp.vl["SWA_01"]["SWA_Infostufe_SWA_li"]),
```
adeebshihadeh, line 64: can you add a comment explaining the info and warning states?

adeebshihadeh: since it's only two signals, I'd also just use an or instead of building a list and using any

jyoung8607: Sure, NP

PR #20247, opened 2021-03-05
VW MQB: Use raw steering angle signal

jyoung8607:

Read the steering angle from EPS_01 instead of LWI_01. I initially chose the signal from LWI_01 because it had higher precision (0.1 deg compared to 0.15 deg), looked "smoother", and for certain vehicles with variable ratio steering racks, it properly included the effect. But, the processing seems to cause undesirable effects with openpilot lane change, particularly when changing lanes across a shift in road-crown, possibly due to effects of straight-line stability/active centering (VW SSP 317). openpilot lane changes are far more consistent with the EPS_01 angle data.

**Verification**

Several months in the community port. Test driven as a standalone change to upstream master. Would be interested in any feedback from Comma's A3 or validation of our theories about the underlying data.

**Route**

2018 Volkswagen Golf R: cae14e88932eb364|2021-03-04--21-39-17


pd0wm:
@adeebshihadeh Can you try this on the A3?

The signal looks pretty different, but of course it's hard to tell which one is right.

image

adeebshihadeh:
Will test this on monday!

jyoung8607:
For cars with fixed ratio steering racks, they track very closely but not exactly. There's a bit of offset and it appears it can vary. All MQB steering angle sensors are at the bottom of the steering column, so there's no bushing flex or u-joint slop to be accounted for.

For cars with progressive variable ratio steering racks, LWI_01 accounts for the effect while EPS_01/LH_EPS_03 will show the actual steering column angle. Here's a graph of testing a 45°, 90°, 180°, full-lock turn of the wheel on my Golf R. Note They track each other up until they approach 90° where the ratio appears to change.

lwi-vs-eps

I have reason to believe LWI_01 is trying to show the vehicle's true direction of travel, which can differ from steering wheel angle, even when near center. I have information from an authoritative source which describes the signal thusly:

> "**Calculated steering wheel angle (BLW)**: Output of the calculated steering angle **_with straight-line correction_** from the position of the steering motor (absolute value) (only approved for specified use in ABS, ESP, PLA, HCA) "

PLA = park assist, HCA = Heading Control Assist / Lane Assist (LKAS). The EPS rack tries to accommodate for road crown or cross winds or slight problems with wheel alignment or tire inflation, and maintain straight-ahead driving along with properly intuitive return-to-center. Driver assist features like Park Assist and LKAS need to understand the true direction of travel.

It's hard to do repeatable tests and measurements of openpilot lane-change performance, but anecdotally I feel like it's a lot more consistent with the raw steering wheel signal than the calculated straight-ahead signal. When traversing a change in road crown, it's much more consistent with the raw angle than the calculated angle.

My line of thought is, when crossing a road crown split, both the EPS and openpilot itself have to re-learn their angle-offsets for straight-ahead driving, but the EPS re-learn is changing openpilot's view of reality during that process and making its job a lot harder. However, that's a hypothesis based on road feel and not on formally analyzing the data, which I'm not very well equipped for.

I feel this change helps stability in both normal driving and lane changes, and I feel like I understand why, and it's had good feedback in the community port. But, I acknowledge part of my case is guesswork and conjecture, and welcome feedback on whether Comma thinks it's a correct change to make. It sucks to lose the more-correct view of steering angle for progressive ratio racks, but 99.99% of openpilot driving happens in the center regime, and I can think of ways of identifying those racks and computing an angle/ratio correction should it be needed.

pd0wm:
If it's doing any processing it's probably better to use the raw angle. You're correct that those two learners might not interact too well.

The difference in angle due to variable ratio is probably very small in the range that openpilot is actually using (+- 45 degrees, probably even less in highway driving). The differences that will occur will be learnt as an angle offset as well. 

Do you still have that drive of the step changes? Like to create an x/y plot to quantify the error vs steering error we create by ignoring the changing ratio.

jyoung8607:
Thank you for taking a look!

Logs from that drive are long-gone, screenshot was more than a year old, so I repeated the test. Just for additional confirmation of what we expect, EPS_01/LH_EPS_03 is topping out at 346° and full-lock is indeed just slightly less than a full turn from center on this particular vehicle.

cae14e88932eb364|2021-03-08--07-03-55
Pauses at center, 45°, 90°, 180°, full right lock

lwi-vs-eps-pauses

cae14e88932eb364|2021-03-08--07-08-25
Gradual slow turn from center to full right lock

lwi-vs-eps-constant

pd0wm:
2021-03-08-151730_1659x1272_scrot

It's pretty hard to estimate the ratio at small angles (divide by zero and small offset). But at 45 deg the ratio seems to be 0.93, and at 100 degrees it's 0.90. It takes the full end-to end to go down to a 0.77 ratio. This confirms my belief that variable ratio racks only vary a lot all the way at the ends to make parking easier.

jyoung8607:
I don't think it's a showstopper right now, but can you think of a good way to describe those step functions in code? That way we'd get the best of both worlds, and we make sure to not confuse the params learner. It would be ideal if we could put something in the port like:

```python
raw_steer_angle = pt_cp.vl["EPS_01"]['Steering_Wheel_Angle'] * (1, -1)[int(pt_cp.vl["EPS_01"]['Steering_Wheel_Angle_VZ'])]
ret.steeringAngleDeg = get_true_steer_angle(raw_angle, characteristic_map)
```

Where get_true_steer_angle() is a function that multiplies `raw_angle` by the ratio at that point in the rack, and `characteristic_map` describes the ratios for that particular rack, which we could identify by its UDS fingerprint. I thought about doing a simple interp() between the breakpoints, but that will turn the breakpoint S-curves into diagonals. Unfortunately I don't have the mathematical background to clearly and simply express that curve in code.

Thoughts welcome. If you can think of some suitable way to calculate the mapping, I'm willing to do the legwork of checking cars as they come onboard and making sure the cars with progressive racks are identified and mapped out.

pd0wm:
I don't think the mapping is worth it. If we later decide we need it we'll make something generic for all cars, since VW is not the only with this problem.

adeebshihadeh:
Looks good on our A3. Good to merge after the tests pass.

PR #20248, opened 2021-03-05
refactor sensors_qcom

deanlee:
1. new struct `SensorState`
2. remove hardcoded value of sensor handles,  get them from `get_sensors_list`.
3. call `setDeay` after reactivate sensors
4. ~try to solve the following problem(sensor.which() will return 0(ACCELERATION) if the message is empty):~
https://github.com/commaai/openpilot/blob/59698344a2c1f6557b06485de1dcf296ec5eb542/selfdrive/ui/ui.cc#L216-L217
4. use callback function to build sensor message.

### DIFF
--- a/selfdrive/sensord/sensors_qcom.cc
+++ b/selfdrive/sensord/sensors_qcom.cc
@@ -1,23 +1,10 @@
-#include <stdio.h>
-#include <stdint.h>
-#include <stdlib.h>
-#include <string.h>
-#include <unistd.h>
-#include <assert.h>
-#include <sys/time.h>
-#include <sys/cdefs.h>
 #include <sys/types.h>
 #include <sys/resource.h>
-
 #include <map>
-#include <set>
-
-#include <cutils/log.h>
 #include <hardware/sensors.h>
 #include <utils/Timers.h>
 
 #include "messaging.hpp"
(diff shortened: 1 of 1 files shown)


pd0wm:
regarding 4. That happened while running sensors_qcom2. Also pretty sure it's on the receiving end, I can run multiple instances of ui at the same time, and only one will crash.

deanlee:
oic

adeebshihadeh:
I hesitate to merge these refactor PRs that don't solve any problems for daemons that are stable, small, and self-contained. The QCOM sensor daemon is ~230 lines and works well. What's the benefit of merging this? Does it use less CPU?

deanlee:
@pd0wm The reason that `sensorEvents` reads an empty list may be that the number of read connections exceeds the limit (NUM_READERS) :
https://github.com/commaai/cereal/blob/97bfe99980a3ef58e07eb8006ac5e59bf83ad986/messaging/msgq.hpp#L27-L29

deanlee:
> regarding 4. That happened while running sensors_qcom2. Also pretty sure it's on the receiving end, I can run multiple instances of ui at the same time, and only one will crash.

it's fixed in #21427

PR #20250, opened 2021-03-05
log to file and send through athena

gregjhogan:


selfdrive/hardware/base.py
```diff
@@ -105,3 +105,7 @@ def get_thermal_config(self):
   @abstractmethod
   def set_screen_brightness(self, percentage):
     pass
+
+  @abstractmethod
+  def get_cloudlog_enabled(self):
+    pass
```
adeebshihadeh, line 111: I would just put this in athenad, don't think it belongs in the HW abstraction layer

common/dict_helpers.py
```diff
@@ -1,6 +1,9 @@
 # remove all keys that end in DEPRECATED
 def strip_deprecated_keys(d):
-  for k in d.keys():
-    if type(k) == str and k.endswith('DEPRECATED'):
-      d.pop(k)
+  for k in list(d.keys()):
+    if type(k) is str:
```
adeebshihadeh, line 4:
```suggestion
    if isinstance(k, str):
```

common/logging_extra.py
```diff
@@ -62,8 +63,43 @@ def format_dict(self, record):
...
+      for ik, iv in v.items():
+        ik, iv = self.fix_kv(ik, iv)
+        nv[ik] = iv
+      v = nv
+    elif isinstance(v, list):
+      k += "$a"
+    return k, v
+
+  def format(self, record):
+    if type(record) == str:
```
pd0wm, line 91:
```suggestion
    if isinstance(record, str):
```

common/logging_extra.py
```diff
@@ -62,8 +63,43 @@ def format_dict(self, record):
     return record_dict
 
   def format(self, record):
+    if self.swaglogger is None:
+      raise Exception("must set swaglogger before calling format()")
     return json_robust_dumps(self.format_dict(record))
 
+class SwagLogFileFormatter(SwagFormatter):
+  def fix_kv(self, k, v):
```
pd0wm, line 71: docstring? What is the purpose of this function?

common/logging_extra.py
```diff
@@ -62,8 +63,43 @@ def format_dict(self, record):
...
+        ik, iv = self.fix_kv(ik, iv)
+        nv[ik] = iv
+      v = nv
+    elif isinstance(v, list):
+      k += "$a"
+    return k, v
+
+  def format(self, record):
+    if type(record) == str:
+      v = json.loads(record)
```
pd0wm, line 92: exception handling in case of non valid json?

selfdrive/swaglog.py
```diff
@@ -1,19 +1,69 @@
...
+  def get_existing_logfiles(self):
+    log_files = list()
+    base_dir = os.path.dirname(self.base_filename)
+    for fn in os.listdir(base_dir):
+      fp = os.path.join(base_dir, fn)
+      if fp.startswith(self.base_filename) and os.path.isfile(fp):
+        log_files.append(fp)
+    return sorted(log_files)
+
+  def shouldRollover(self, record):
```
pd0wm, line 50: snake case

selfdrive/swaglog.py
```diff
@@ -1,19 +1,69 @@
...
+    size_exceeded = self.max_bytes > 0 and self.stream.tell() >= self.max_bytes
+    time_exceeded = self.interval > 0 and self.last_rollover + self.interval <= time()
+    return size_exceeded or time_exceeded
+
+  def doRollover(self):
```
pd0wm, line 55: snake case

common/dict_helpers.py
```diff
@@ -0,0 +1,9 @@
+# remove all keys that end in DEPRECATED
```
pd0wm, line 1: add new files to `release/files_common`

selfdrive/swaglog.py
```diff
@@ -1,19 +1,69 @@
 import os
+from time import time
+from pathlib import Path
 import logging
+from logging.handlers import BaseRotatingHandler
 
-from logentries import LogentriesHandler
```
pd0wm, line 4: remove logentries from pipfile

selfdrive/swaglog.py
```diff
@@ -1,19 +1,69 @@
...
-  return LogentriesHandler(le_token, use_tls=False, verbose=False)
+def get_file_handler():
+  Path(SWAGLOG_DIR).mkdir(parents=True, exist_ok=True)
+  base_filename = os.path.join(SWAGLOG_DIR, "swaglog")
+  handler = SwaglogRotatingFileHandler(base_filename)
+  return handler
 
+class SwaglogRotatingFileHandler(BaseRotatingHandler):
+  def __init__(self, base_filename, interval=60, max_bytes=1024*256, backup_count=2500, encoding=None):
+    BaseRotatingHandler.__init__(self, base_filename, mode="a", encoding=encoding, delay=True)
```
pd0wm, line 25:
```suggestion
    super().__init__(self, base_filename, mode="a", encoding=encoding, delay=True)
```

common/logging_extra.py
gregjhogan: overlap cannot exist in properties once expanded (e.g. `log.info` creates `msg` vs `log.event()` creates `msg.deviceState.deviceState.ambientTempC`)

selfdrive/swaglog.py
gregjhogan: I don't think I can, it is `BaseRotatingHandler.shouldRollover`

selfdrive/swaglog.py
```diff
@@ -1,19 +1,69 @@
...
+    base_dir = os.path.dirname(self.base_filename)
+    for fn in os.listdir(base_dir):
+      fp = os.path.join(base_dir, fn)
+      if fp.startswith(self.base_filename) and os.path.isfile(fp):
+        log_files.append(fp)
+    return sorted(log_files)
+
+  def shouldRollover(self, record):
+    size_exceeded = self.max_bytes > 0 and self.stream.tell() >= self.max_bytes
+    time_exceeded = self.interval > 0 and self.last_rollover + self.interval <= time()
```
pd0wm, line 52: Use some monotonic time source. This will trigger a rotation when the time syncs.

selfdrive/athena/athenad.py
```diff
@@ -244,6 +254,82 @@ def b64jpeg(x):
...
+    log_path = os.path.join(SWAGLOG_DIR, log_entry)
+    # file without an extension is always the active log file
+    if log_entry == "swaglog":
+      continue
+    try:
+      time_sent = int.from_bytes(getxattr(log_path, LOG_ATTR_NAME), sys.byteorder)
+    except (ValueError, TypeError):
+      time_sent = 0
+    # assume send failed and we lost the response if sent more than one hour ago
+    if not time_sent or curr_time - time_sent > 3600:
```
pd0wm, line 270: This also needs to take into account the system time starting on midnight 2020/04/01 and switching to the current time after boardd is started.

selfdrive/swaglog.py
```diff
@@ -1,19 +1,69 @@
...
 
-class LogMessageHandler(logging.Handler):
+  def _open(self):
+    self.last_rollover = time()
+    next_filename = f"{self.base_filename}.{self.last_rollover}"
+    stream = open(next_filename, self.mode, encoding=self.encoding)
+    self.log_files.insert(0, next_filename)
+    return stream
+
+  def get_existing_logfiles(self):
```
pd0wm, line 41: Can we share some of this code with athena?

gregjhogan: I only see two lines that really overlap (for loop and joining dir with file name) so that seems ok, what were you thinking?

common/logging_extra.py
gregjhogan: added comments explaining

common/logging_extra.py
gregjhogan: python logging base formatter already catches exceptions and throws up the errors to stdout and keeps going (swallows exception after printing it) which I think is what we want to do in this unexpected situation

selfdrive/athena/athenad.py
gregjhogan: I think this is OK since there is only ever a single log file in flight at a time, and so worst case we send one file twice which shouldn't cause any problems, what do you think?

selfdrive/swaglog.py
gregjhogan: will switch to `sec_since_boot()` and then change file name to a monotonic counter

selfdrive/athena/athenad.py
pd0wm: I would just use sec_since_boot for this time, since it doesn't matter what it is as long as you can measure durations.

selfdrive/athena/athenad.py
```diff
@@ -244,6 +254,80 @@ def b64jpeg(x):
...
+        log_path = os.path.join(SWAGLOG_DIR, log_entry)
+        if log_entry and log_success:
+          try:
+            setxattr(log_path, LOG_ATTR_NAME, LOG_ATTR_VALUE_MAX_UNIX_TIME)
+          except OSError:
+            pass # file could be deleted by log rotation
+      except queue.Empty:
+        pass
+
+      curr_time = int(time.time())
```
pd0wm, line 296: use `sec_since_boot()`, see comment above.

selfdrive/athena/athenad.py
gregjhogan: it does matter, because if it reboots with a high monotonic time it will take forever to consider it eligible to re-send, right?

selfdrive/athena/athenad.py
gregjhogan: yeah, I missed this one, thanks!

PR #20253, opened 2021-03-06
VW MQB: Support for other transmission types

jyoung8607:

Add some necessary pieces to support the Volkswagen e-Golf, which has no transmission and needs to find gearshift position from a different signal. Also add support for manual transmission vehicles, which can be had with ACC and LKAS.

**Verification**

Test driven as a standalone change to upstream master. I myself can only test the auto trans branch of the code, but many others in the community have exercised this same code for months in the community port.

**Route**

2018 Volkswagen Golf R: cae14e88932eb364|2021-03-05--18-47-48

### DIFF
--- a/selfdrive/car/volkswagen/carstate.py
+++ b/selfdrive/car/volkswagen/carstate.py
@@ -165,7 +165,6 @@ class CarState(CarStateBase):
       ("ZV_HD_offen", "Gateway_72", 0),             # Trunk or hatch open
       ("BH_Blinker_li", "Gateway_72", 0),           # Left turn signal on
       ("BH_Blinker_re", "Gateway_72", 0),           # Right turn signal on
-      ("GE_Fahrstufe", "Getriebe_11", 0),           # Auto trans gear selector position
       ("AB_Gurtschloss_FA", "Airbag_02", 0),        # Seatbelt status, driver
       ("AB_Gurtschloss_BF", "Airbag_02", 0),        # Seatbelt status, passenger
       ("ESP_Fahrer_bremst", "ESP_05", 0),           # Brake pedal pressed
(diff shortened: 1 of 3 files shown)


selfdrive/car/volkswagen/carstate.py
```diff
@@ -42,8 +45,16 @@ def update(self, pt_cp):
...
     # Update gear and/or clutch position data.
-    can_gear_shifter = int(pt_cp.vl["Getriebe_11"]['GE_Fahrstufe'])
-    ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(can_gear_shifter, None))
+    if trans_type == TRANS.automatic:
+      ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(pt_cp.vl["Getriebe_11"]['GE_Fahrstufe'], None))
+    elif trans_type == TRANS.direct:
+      ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(pt_cp.vl["EV_Gearshift"]['GearPosition'], None))
+    elif trans_type == TRANS.manual:
+      ret.clutchPressed = not pt_cp.vl["Motor_14"]['MO_Kuppl_schalter']
+      if bool(pt_cp.vl["Gateway_72"]['BCM1_Rueckfahrlicht_Schalter']):
```
pd0wm, line 54: Is there a way to get reverse signal from the gearbox directly instead of using the light?

jyoung8607: In the case of a manual trans, there's no gearbox controller to get it from. There's a contact switch in the shifter for Reverse wired to the BCM, and a clutch pedal contact switch wired to the ECM. That's about it.

In this case, I believe the signal is truly referring to the switch and not the light on the back of the car. "Rueckfahrlicht schalter" means "reversing light switch". To put that in context, VW refers to the pair of redundant brake pedal contacts as the  "brake light switch" and the "brake test switch", and those are both fully distinct from the tail braking lights being activated, which we can see happen by watching BCM messaging when ACC is decelerating without driver input.

pd0wm: Ok, sounds good! The "switch" explanation makes sense to me.

selfdrive/car/volkswagen/values.py
```diff
@@ -24,6 +25,9 @@ class CANBUS:
   pt = 0
   cam = 2
 
+TRANS = car.CarParams.TransmissionType
+GEAR = car.CarState.GearShifter
```
adeebshihadeh, line 29:
```suggestion
TransmissionType = car.CarParams.TransmissionType
GearShifter = car.CarState.GearShifter
```

this is the convention used throughout the codebase

jyoung8607: Done!

PR #20258, opened 2021-03-06
CameraBuf: new function copy_cl_buffer

deanlee:


pd0wm:
Not sure if we should hide this in the function, especially since it's not just a copy but also applies a gain. The function is also only used in one place, so it doesn't save any code.

PR #20259, opened 2021-03-06
cameras_qcom: use macros for actuator moving dir

deanlee:


selfdrive/camerad/cameras/camera_qcom.cc
```diff
@@ -855,8 +855,7 @@ static void road_camera_start(CameraState *s) {
   s->cur_lens_pos = 0;
   s->cur_step_pos = inf_step;
 
-  actuator_move(s, s->cur_lens_pos);
-
+  actuator_move(s, 0);
```
pd0wm, line 858: Why change this?

deanlee: better not to change.reverted.

PR #20268, opened 2021-03-07
Honda Clarity Port

vanillagorillaa:
PR for Honda Clarity. Still a work in progress and requires a proxy panda (not sure if we want to upstream this, so put in community supported for now), and also needs a modified Nidec harness (Clarity sends on the additional FCAN, not the same FCAN that the rest of the Nidec Hondas send messages on).

TODOs:

- [x] test route added to test_car_models 
- [x] Route with openpilot
- [ ] Route with stock system
- [ ] Figure out braking (clarity has regen braking, figure out the hand off between that and friction brakes)


vanillagorillaa:
Due to requiring panda safety changes ("computer_brake" is located in different bytes of the x:1fa message and is not supported with out changing the checks in safety_honda) I will be closing this pull request.

gregjhogan:
For other Honda vehicles we use the safety mode param to differentiate the brake messages for bosch hardware, so maybe a new safety mode param value would work. I also have a PR for dynamic rx check list (controlled by safety mode param) if that is needed.

PR #20271, opened 2021-03-08
VW MQB: UDS fingerprinting support

jyoung8607:
**Prep for Volkswagen MQB FPv2**

This is the least obnoxious way I can make FPv2 work for us. Feedback welcome. It depends on commaai/panda#631 for the UDS library to accept query responses at different offsets. Will update the commit reference when it's available.

* Add support for different response offsets per-query-type
* Add two Volkswagen-specific queries (one at normal offset for engine and trans, another for the rest)
* ~~Sort all CAR references alphabetically (there's going to be a whole lot of them soon)~~
* ~~Refactor brands to widely-known abbreviations AU/SE/SK/VW (Audi, SEAT, Škoda, Volkswagen)~~
* ~~Add generation with widely-known Mk references (e.g. Golf Mk7, because Mk6 and Mk5 and hopefully Mk8 coming soon)~~
* Actually provide per-model mass and wheelbase for the first time ever YAY

**Verification**

Several minutes of testing on a single car in a private branch. Test driven as a standalone change to upstream master.

**Route**

2018 Volkswagen Golf R: cae14e88932eb364|2021-03-07--20-43-08


adeebshihadeh:
Merged the panda PR and bumped on master.

adeebshihadeh:
Let's keep this to just the fingerprinting. We can do a big car dump once this is in.

jyoung8607:
I can withdraw the Atlas and Tiguan if you like, but I was stopping here anyway, after adding enough to test and illustrate what it will look like. The Golf R and Tiguan are both in my own driveway. The Atlas used to be in my driveway (turned in off of lease) but I have archived diagnostic data sufficient to teach openpilot what it looked like. The BIG car dump comes later. 👍

adeebshihadeh:
I'd prefer a minimal PR to bring FP 2.0 to VW. The Golf and our A3 should be enough to see what this looks like.

selfdrive/car/volkswagen/values.py
```diff
@@ -46,19 +48,51 @@ class CANBUS:
 }
 
 class CAR:
-  GOLF = "VOLKSWAGEN GOLF"
-  AUDI_A3 = "AUDI A3"
+  AU_A3_MK3 = "AUDI A3 3RD GEN"
+  VW_GOLF_MK7 = "VOLKSWAGEN GOLF 7TH GEN"
```
adeebshihadeh, line 52: let's revert the changes here. we can come up with a convention in another PR

selfdrive/car/volkswagen/values.py
```diff
@@ -46,19 +48,51 @@ class CANBUS:
...
+
+MQB_CARS = {
+  CAR.AU_A3_MK3,             # Chassis 8V, 2013-2019, includes A3, A3 e-tron, A3 g-tron, S3, RS3
+  CAR.VW_GOLF_MK7            # Chassis AU, 2013-2020, includes Golf, Alltrack, Sportwagen, GTI, GTI TCR, GTE, GTD, Clubsport, Golf R, e-Golf
+}
 
 FINGERPRINTS = {
-  CAR.GOLF: [{
-    64: 8, 134: 8, 159: 8, 173: 8, 178: 8, 253: 8, 257: 8, 260: 8, 262: 8, 264: 8, 278: 8, 279: 8, 283: 8, 286: 8, 288: 8, 289: 8, 290: 8, 294: 8, 299: 8, 302: 8, 346: 8, 385: 8, 418: 8, 427: 8, 668: 8, 679: 8, 681: 8, 695: 8, 779: 8, 780: 8, 783: 8, 792: 8, 795: 8, 804: 8, 806: 8, 807: 8, 808: 8, 809: 8, 870: 8, 896: 8, 897: 8, 898: 8, 901: 8, 917: 8, 919: 8, 927: 8, 949: 8, 958: 8, 960: 4, 981: 8, 987: 8, 988: 8, 991: 8, 997: 8, 1000: 8, 1019: 8, 1120: 8, 1122: 8, 1123: 8, 1124: 8, 1153: 8, 1162: 8, 1175: 8, 1312: 8, 1385: 8, 1413: 8, 1440: 5, 1514: 8, 1515: 8, 1520: 8, 1529: 8, 1600: 8, 1601: 8, 1603: 8, 1605: 8, 1624: 8, 1626: 8, 1629: 8, 1631: 8, 1646: 8, 1648: 8, 1712: 6, 1714: 8, 1716: 8, 1717: 8, 1719: 8, 1720: 8, 1721: 8
+  CAR.AU_A3_MK3: [{
```
adeebshihadeh, line 60: revert these fingerprint additions too

selfdrive/car/volkswagen/values.py
```diff
@@ -46,19 +48,51 @@ class CANBUS:
 }
 
 class CAR:
-  GOLF = "VOLKSWAGEN GOLF"
-  AUDI_A3 = "AUDI A3"
+  AU_A3_MK3 = "AUDI A3 3RD GEN"
+  VW_GOLF_MK7 = "VOLKSWAGEN GOLF 7TH GEN"
+
+MQB_CARS = {
```
adeebshihadeh, line 54: same as above

jyoung8607:
Comments still welcome on the overall structure/approach, but don't merge yet. I'm still working on some fine-tuning of the exact queries we're running, and working out some inconsistencies in what we're getting back from the engine ECU in some cars.

I'm going to drop ABS/ESP from the fingerprint list for now. A subset of VW ABS/ESP controllers are initially replying with 0x78: 'request correctly received - response pending' and then sending what we need, and I'm not sure we can properly handle both variants from openpilot's parallel stateless UDS query code. We never send commands directly to the ABS/ESP controller, only to the drivetrain coordinator (TSK) in the engine ECU, and the ECU will already throw a fit if it doesn't like something about the ESP.

jyoung8607:
Tracked down a PEBCAK issue that was stopping multi-query from working for me, so I've added that to my implementation. We're now grabbing everything I desired for both strong vehicle/component identification and strong version validation:

- VEHICLE_MANUFACTURER_SPARE_PART_NUMBER returns the controller's "software part number", what feature/functionality it's been flashed with. Can be shared between vehicles. Sometimes (but not always) changes with software upgrades.
- VEHICLE_MANUFACTURER_ECU_SOFTWARE_VERSION_NUMBER returns something like a build number or subversion. Has absolutely nothing to do with what vehicle it's installed in. If VMSPN doesn't change after a software update, this one will.
- APPLICATION_DATA_IDENTIFICATION returns something that seems to identify the per-vehicle-model parameterization data block that gets uploaded to some types of modules, tweaking that control module to work for a certain model.

The SRS multi-ID ought to do a very good job of discriminating between vehicle models, especially when combined with the EPS multi-ID where I can now detect the per-vehicle parameterization for EPS power assist curves, which lets me tell apart a Golf vs e-Golf vs Golf R, etc for future tuning purposes.

Testing still underway but I'm pretty happy with this set of data.

gregjhogan:
> I'm going to drop ABS/ESP from the fingerprint list for now. A subset of VW ABS/ESP controllers are initially replying with 0x78: 'request correctly received - response pending' and then sending what we need, and I'm not sure we can properly handle both variants from openpilot's parallel stateless UDS query code. We never send commands directly to the ABS/ESP controller, only to the drivetrain coordinator (TSK) in the engine ECU, and the ECU will already throw a fit if it doesn't like something about the ESP.

Seems like ignoring messages with that negative response code is easy (no response is needed)?  It would be a good universal improvement, but perhaps should be a separate PR.
https://github.com/commaai/panda/blob/3bb0408ee62a983d54f9eab78c13cc1ee54703f4/python/uds.py#L545-L549

jyoung8607:
> Seems like ignoring messages with that negative response code is easy (no response is needed)? It would be a good universal improvement, but perhaps should be a separate PR.

Turns out that may be PEBCAK as well, since with multi-query I'm now seeing that from several modules and it's working fine. I will try to take another look at why ABS/ESP is giving us intermittent heartache, but I don't think it's that important to add back.

jyoung8607:
This FPv2 multi-query checks on a 2018 A3 e-tron Sportback: https://github.com/Ambroos/openpilot/commit/286b06bcb739c62fd12be59334a88a98a19774b5. Waiting for feedback from a couple other users, and I will try our Tiguan tonight or possibly tomorrow. Feedback welcome from Comma's A3.

jyoung8607:
Still having a heck of a time figuring out the right way to de-conflict the engine 0x7e0 and transmission 0x7e1, such that I consistently get the multi-query data want without possibly breaking other cars. Alternatives:
* Just remove engine and transmission from the VW FPv2 constellation (I see this for a few Hyundais)
* Try to find something super VW proprietary that Hyundai won't respond to, and put VW above them
* Only look for VW candidates if VIN matches ^((WVW)|(1V2)|(TMB)|(3VW)|(WAU)|(WVG)|(LFV)).{14}$
* ... ?

gregjhogan:
another option may be to send tester present (0x3E) to all addrs (no sub-addrs) first, then by process of elimination from the set of addrs that respond (even if you get error responses) I bet you will always land on a single make, then only re-query the addrs that responded for that specific make (or all makes if non matched)

jyoung8607:
I think I've figured out the last of our issues, partially an ID10T error when switching branches on my device 800 times this week, partially slow wake-up on the e-Golf fwdCamera.

After sleeping on it, I've realized if we later run into query order issues, all we have to do is run my snowflake 0x6A offset queries first, which probably won't evoke a response from any other cars, and will winnow down the candidate list nicely.

jyoung8607:
FYI, examples of what we're picking up so far: https://github.com/jyoung8607/openpilot/blob/0302a8c8b370114e2bd59e4433037d91662378b6/selfdrive/car/volkswagen/values.py#L73-L149

jyoung8607:
I've trimmed the fwdCamera data and its support code from this PR. fwdCamera is optional for VW, it's slow to respond on several cars we've tested, and we get the SW/API info we need from the steering rack.

I have two owners that are consistently having wakeup issues: they can FP fine if OP boots up after the car, but it misses if OP is running and tries to FP immediately at startup. I'm still working with the owners of those vehicles to see what's unique about their car or their hardware. We may end up needing a timer or retry tweak later, but I don't think it's an obstacle to upstreaming the rest of the work.

Aside from that, I'm confident in the dataset we're collecting here. We already have samples from 20 vehicles across 4 brands (VW, Audi, SEAT, Skoda) and 8 openpilot CARs representing 12 submodels. Once this passes review I can have the test users collect drive data for process_replay and start opening PRs.

https://github.com/jyoung8607/openpilot/blob/mqb-fpv2-temp/selfdrive/car/volkswagen/values.py

selfdrive/car/volkswagen/values.py
```diff
@@ -62,6 +68,43 @@ class CAR:
   }],
 }
 
+FW_VERSIONS = {
+  CAR.AUDI_A3: {
+    (Ecu.engine, 0x7e0, None): [
+      b'\xf1\x875G0906259L \xf1\x890002',  # 2017 Audi A3 Prestige (CNTC)
```
adeebshihadeh, line 74: remove all the comments for specific FW versions

selfdrive/car/volkswagen/values.py
```diff
@@ -53,6 +54,11 @@ class CAR:
   GOLF = "VOLKSWAGEN GOLF"
   AUDI_A3 = "AUDI A3"
 
+MQB_CARS = {
+  CAR.GOLF,                 # Chassis AU, 2013-2020, includes Golf, Alltrack, Sportwagen, GTI, GTI TCR, GTE, GTD, Clubsport, Golf R, e-Golf
```
adeebshihadeh, line 58: let's remove this list for the time being. I could easily see a not-obvious bug where we merge a car port that forgets to put the new car in the list and ends up with no car params set

selfdrive/car/isotp_parallel_query.py
```diff
@@ -7,9 +7,13 @@
 from selfdrive.boardd.boardd import can_list_to_can_capnp
 from panda.python.uds import CanClient, IsoTpMessage, FUNCTIONAL_ADDRS, get_rx_addr_for_tx_addr
 
+class RX_OFFSET:
```
adeebshihadeh, line 10: just make a variable called `DEFAULT_RX_OFFSET` in fw_versions.py and hardcode the VW offset, don't want to have make-specific stuff here

selfdrive/car/volkswagen/interface.py
```diff
@@ -25,7 +25,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None):
     # VW port is a community feature, since we don't own one to test
     ret.communityFeature = True
 
-    if candidate in [CAR.GOLF, CAR.AUDI_A3]:
+    if candidate in MQB_CARS:
```
adeebshihadeh, line 28: remove this if, same comment as above

selfdrive/car/volkswagen/values.py
jyoung8607: Done

selfdrive/car/isotp_parallel_query.py
jyoung8607: Done

selfdrive/car/volkswagen/values.py
jyoung8607: I request that we keep it.

The `candidate in MQB_CARS` block in interface.py is designed to collect all the MQB-specific work together, and allow the same for the PQ-specific work. It may be a long time (or never) before Comma accepts that work upstream. However, the task of maintaining a PQ development fork will involve a much larger diff with much more frequent merge conflicts if this part isn't included now.

TL;DR: Please don't make my job harder. :)

Since you requested process_replay drives for every make/model (and we intend to provide them) I think that would suffice for now. Beyond that, I think every other car port already lets get_params() run off the end with no data if there's no match. Robustly dealing with that is a laudable goal but this PR isn't really the vehicle to make it happen.

selfdrive/car/volkswagen/interface.py
jyoung8607: Same comment as above: I request that we keep it in.

selfdrive/car/volkswagen/values.py
adeebshihadeh: we need routes for test car models, but process replay wouldn't catch that anyway. process replay assumes that you start in a known, good state.

I understand wanting to minimize the diff. I propose we use simply use one list (PQ_CARS, MQB gets privileges for being first). See honda for precedent on this: https://github.com/commaai/openpilot/blob/master/selfdrive/car/honda/values.py#L1160.

adeebshihadeh: I think one list is preferably anyway. We'll run into the same thing where we have to make sure each car is in exactly one of the lists.

jyoung8607: I'm not sure that proposal advances either of our goals.

Your stated concern is "_a not-obvious bug where we merge a car port that forgets to put the new car in the list and ends up with no car params set_". The proposal makes the penalty for "forgetting" into a _wrong_ carParams situation rather than _no_ carParams, which is a large backward step in obviousness and safety.

My equivalent of the [Honda Bosch conditional section of interface.py](https://github.com/commaai/openpilot/blob/d72855ecb8749ca994af4f1b5cf1b575c068e88e/selfdrive/car/honda/interface.py#L127-L136) would be considerably longer (for trans type and network location detection) and would be a 100% diff just due to indentation, unless you let me hang it all behind an `if True:` or something until a later possible merge of PQ support.

Your point about making sure each CAR is on exactly one list is well-taken, but `test_car_interfaces.py` could be augmented to perform that check.

adeebshihadeh: I'm fine with an `if True:` for the time being. I'd just really like to avoid dead code in the hopes of one day upstreaming something which isn't in a very upstreamable state.

jyoung8607: Fair enough, I'll make the changes shortly. Thanks for the accommodation.

jyoung8607: Done. And yes, I'd like to retain the comments for what fits into each CAR/chassis type. Otherwise, people adding new FPv2 data will try to add new CARs for e.g., the dozen variants of VW Golf, which is a bad idea for (wall of text).

PR #20272, opened 2021-03-08
Run models in function run_model

deanlee:


pd0wm:
This shouldn't happen in practice though.Maybe better to not have this complexity?

deanlee:
If so, then the nested while loop is meaningless.

deanlee:
@pd0wm how about this one? run the model in  `run_mode()`,  keep the initialization and cleanup code in `main()`

deanlee:
@pd0wm how about this one? run the model in  `run_mode()`,  keep the initialization and cleanup code in `main()`

selfdrive/modeld/modeld.cc
```diff
@@ -101,90 +152,20 @@ int main(int argc, char **argv) {
...
-    break;
+  while (!do_exit && !vipc_client.connect(false)) {
+    util::sleep_for(100);
   }
 
-  // loop
-  while (!do_exit) {
-    VisionBuf *b = &vipc_client.buffers[0];
+  // run the models
+  if (vipc_client.connected) {
```
pd0wm, line 160: Can this ever be not true?

pd0wm:
Looks good!

selfdrive/modeld/modeld.cc
```diff
@@ -14,11 +13,11 @@
...
 ExitHandler do_exit;
 // globals
-bool run_model;
+bool live_calib_seen;
 mat3 cur_transform;
-pthread_mutex_t transform_lock;
+std::mutex transform_lock;
 
-void* live_thread(void *arg) {
+void live_thread() {
```
pd0wm, line 20: Guess we should rename this to `calibration_thread` since that's what it's doing.

selfdrive/modeld/modeld.cc
deanlee: vipc_client.connected is false only when do_exit is true

PR #20273, opened 2021-03-08
modeld: use sm.frame instead of run_count

deanlee:


adeebshihadeh:
sm.frame and run_count aren't the same thing. sm.frame increments whether the model runs or not.

deanlee:
yes,my bad.I did not notice the  run_model_this_iter

PR #20279, opened 2021-03-08
Spinner: wait for UI to start

sshane:
Fixes the jumping you get at the end where it goes from ~80% to a blank screen, now it makes it all the way to 100% with a delay to show the user it's done. Before it was ambiguous and you might have thought an error or crash occurred for a few seconds while the UI was loading.

### DIFF
--- a/common/spinner.py
+++ b/common/spinner.py
@@ -1,10 +1,14 @@
 import os
+import time
 import subprocess
 from common.basedir import BASEDIR
+from common.clock import sec_since_boot
 
 
 class Spinner():
   def __init__(self):
+    self.update_t = -1
+    self.sleep_t = 1 / 50.
     try:
       self.spinner_proc = subprocess.Popen(["./spinner"],
                                            stdin=subprocess.PIPE,
@@ -25,7 +29,12 @@ class Spinner():
         pass
 
   def update_progress(self, cur: int, total: int):
-    self.update(str(int(100 * cur / total)))
+    elapsed_t = sec_since_boot() - self.update_t
+    if elapsed_t < self.sleep_t:
+      time.sleep(self.sleep_t - elapsed_t)
+
+    self.update(str(round(100 * cur / total)))
+    self.update_t = sec_since_boot()
 
   def close(self):
     if self.spinner_proc is not None:
@@ -44,7 +53,6 @@ class Spinner():
 
 
 if __name__ == "__main__":
-  import time
   with Spinner() as s:
     s.update("Spinner text")
     time.sleep(5.0)
--- a/common/spinner.py
+++ b/common/spinner.py
@@ -44,6 +44,7 @@ class Spinner():
 
 
 if __name__ == "__main__":
+  import time
   with Spinner() as s:
     s.update("Spinner text")
     time.sleep(5.0)
--- a/common/spinner.py
+++ b/common/spinner.py
@@ -1,14 +1,10 @@
 import os
-import time
 import subprocess
 from common.basedir import BASEDIR
-from common.realtime import sec_since_boot
 
 
 class Spinner():
   def __init__(self):
-    self.update_t = -1
-    self.sleep_t = 1 / 50.
     try:
       self.spinner_proc = subprocess.Popen(["./spinner"],
                                            stdin=subprocess.PIPE,
@@ -29,12 +25,7 @@ class Spinner():
         pass
 
   def update_progress(self, cur: int, total: int):
-    elapsed_t = sec_since_boot() - self.update_t
-    if elapsed_t < self.sleep_t:
-      time.sleep(self.sleep_t - elapsed_t)
-
     self.update(str(round(100 * cur / total)))
-    self.update_t = sec_since_boot()
 
   def close(self):
     if self.spinner_proc is not None:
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
This is just a hack for the slow UI loading. We should fix the slow loading instead.

ShaneSmiskol:
The rate limiter is sort of needed, near the end some of the Python prepare items take a very small amount of time and it skips over those if the spinner is updated too fast.

pd0wm:
Are you saying the final 100% is ignored by the spinner?

ShaneSmiskol:
Yes, if too many things are sent at once it skips them. If I just put a 5 second sleep after the 100% update it stays at 90% or so

selfdrive/manager/manager.py
```diff
@@ -136,6 +134,9 @@ def manager_thread():
     pm_apply_packages('enable')
     start_offroad()
 
+  if spinner:
```
pd0wm, line 137: Also make sure you do one call to `ensure_running` before you close the spinner to give the ui a chance to start.

ShaneSmiskol: Thought that would add too much redundant code (sm update, then getting all the arguments correctly) so I put it in the manager loop. Unless it's fine to just do something like `False, False, []`

ShaneSmiskol:
By the way, it looks like if we continually try to get stdin with fgets (by replacing `stdin_input_available` with `true`), then it fully receives all progresses, but then it blocks all the time. The spinner code is just too slow to keep up above 50 Hz and by the time it checks if there's new stdin, the timeout is too large so it thinks there's nothing new, but the last progress it's received might be significantly older than the latest stdin input. I don't know too much C to fix this myself, so sleeping is probably the best hack. At most it adds 0.2 seconds near the end so it can be a smooth animation.

If you really want to, we can just sleep before sending the last 100% update

adeebshihadeh:
Things like spinner should never silently stop execution like this. I'm not too worried about fixing the old spinner since we'll be moving the comma two to Qt very soon, which has its own spinner implementation. We can just make this PR extend the time the spinner runs and fixing up the Makeifle.

common/spinner.py
```diff
@@ -25,7 +25,7 @@ def update(self, spinner_text: str):
         pass
 
   def update_progress(self, cur: int, total: int):
-    self.update(str(int(100 * cur / total)))
+    self.update(str(round(100 * cur / total)))
```
adeebshihadeh, line 28: why change this? strongly doubt the difference is noticeable

ShaneSmiskol: Just so it's more precise, int doesn't round. Something that's 99.9 won't make the spinner display 100%. It's a small change but more correct I think

ShaneSmiskol:
I couldn't get it compiling since the Qt commit, I just reverted to an old release of openpilot to mess around with it.

selfdrive/manager/manager.py
```diff
@@ -99,13 +100,13 @@ def manager_prepare(spinner=None):
...
   total = 100.0 - (0 if PREBUILT else MAX_BUILD_PROGRESS)
-
   for i, p in enumerate(managed_processes.values()):
-    perc = (100.0 - total) + total * (i + 1) / len(managed_processes)
-
+    p.prepare()
     if spinner:
+      perc = (100.0 - total) + total * (i + 1) / len(managed_processes)
+      if perc == 100:
+        time.sleep(1 / 50.)
```
ShaneSmiskol, line 108: Let me test this, I actually think it won't work since it will just be at the last progress+1

adeebshihadeh: the sleep should be reverted

ShaneSmiskol: Done. If it's okay, sending 100% once more right before launching the processes makes the bar finish and show for about the same amount of time I had it sleep for before.

adeebshihadeh: I don't want to merge in any hacks like this. feel free to fix it properly in the old spinner and open a PR, but we'll spend some time making this nice when we ship the Qt UI

ShaneSmiskol: Sounds good. I tried to fix it in the spinner C code but couldn't get anywhere. I'll open a PR if I do though

selfdrive/manager/manager.py
```diff
@@ -136,6 +134,10 @@ def manager_thread():
     pm_apply_packages('enable')
     start_offroad()
 
+  ensure_running(managed_processes.values(), False)
+  if spinner:  # close spinner when ui has started
```
adeebshihadeh, line 138: there's no spinner here

selfdrive/manager/manager.py
```diff
@@ -136,6 +134,10 @@ def manager_thread():
     pm_apply_packages('enable')
     start_offroad()
 
+  ensure_running(managed_processes.values(), False)
```
pd0wm, line 137: Pass in the ignore list, otherwise it might start processes here that aren't supposed to run.

selfdrive/manager/manager.py
```diff
@@ -118,7 +116,7 @@ def manager_cleanup():
   cloudlog.info("everything is dead")
 
 
-def manager_thread():
+def manager_thread(spinner):
```
pd0wm, line 119:
```suggestion
def manager_thread(spinner=None):
```

PR #20281, opened 2021-03-08
Docker compose for simulator

iejMac:


iejMac:
Closed because docker-compose introduces more complications

PR #20288, opened 2021-03-09
2021 Acura RDX firmware clean up + update readme

rodclem:
Rearranged recently added firmware entries for 2021 Acura RDX per suggestion from vanillagorillaa. Updated readme to state 2021 Acura RDX is now supported.


selfdrive/car/honda/interface.py
```diff
@@ -328,7 +328,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.centerToFront = ret.wheelbase * 0.41
       ret.steerRatio = 11.95  # as spec
       ret.lateralParams.torqueBP, ret.lateralParams.torqueV = [[0, 3840], [0, 3840]]
-      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.6], [0.18]]
+      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.5], [0.16]]
```
adeebshihadeh, line 331: this should be done in another PR. our policy on tuning changes is that at least a couple other users have to confirm that it's better

rodclem: My mistake - I did not intend to put the tuning changes in the PR. Sorry for the confusion.

PR #20298, opened 2021-03-10
VW MQB: Updated message and signal data, round 2

jyoung8607:

Match commaai/opendbc#358 with improved and mostly-canonical message naming, signal naming, and scaling properties for ACC_02. We're not using ACC_04 yet but I have some future plans.

**Verification**

The original ACC_02 was a hot mess; I had spitballed the ACC setpoint scaling factor and on rare occasions you could see see a 1mph rounding error between openpilot's UI and the car. The new scaling is authoritatively correct. CI/replay might flag the difference.

**Route**

2018 Volkswagen Golf R: cae14e88932eb364|2021-03-10--10-04-37


jyoung8607:
> The new scaling is authoritatively correct. CI/replay might flag the difference.

And so it did:

```
controlsd
		carControl.hudControl.setSpeed: 5997
		carState.cruiseState.speed: 5997
		controlsState.vCruise: 5997
```

The difference is expected and desired for this PR.

adeebshihadeh:
Re-ran process replay with a small tolerance and looks good.

PR #20297, opened 2021-03-10
VW MQB: Updated message and signal data, round 1

jyoung8607:
**Diff reduction with the VW Community Port**

Match commaai/opendbc#357 with improved and mostly-canonical message naming, signal naming, and scaling properties for LH_EPS_03 (formerly known as EPS_01) and LDW_02. The LDW_02 additions are required for an upcoming stock BSM bug fix.

**Verification**

Null effective change; renaming only except for one fixed two-byte signal that's now split into two fixed one-byte signals. Several months in the community port. Test driven as a standalone change to upstream master.

**Merge Review**

Requires both commaai/opendbc#357 and commaai/panda#632. Getting CI to pass prior to merge on all three will be tricky due to cross-dependencies in CI testing between opendbc, panda, and openpilot. Maybe merge opendbc, bump the opendbc submodule ref in this branch, pass CI here, merge this to master, then update panda's CI test refs to openpilot master to pass CI there, merge panda, then bump openpilot's master panda ref?

**Route**

2018 Volkswagen Golf R: cae14e88932eb364|2021-03-10--06-42-27

PR #20300, opened 2021-03-11
Continued: VW MQB: Updated message and signal data

adeebshihadeh:
@jyoung8607 is the replay difference to be expected here?


jyoung8607:
> @jyoung8607 is the replay difference to be expected here?

Difference is in sendcan. Two-bit signal "LDW_Unknown" has become known, and is now broken into two one-bit signals "LDW_SW_Warnung_links" and "LDW_SW_Warnung_rechts". process_replay is right to flag it because I'm now sending slightly different garbage placeholder data.

For the moment I'll just tweak the new garbage to match the old garbage. In an upcoming PR, those signals are going to change to proper/correct passthrough data from the stock camera, and at that time I'll want to bump the replay reference.

jyoung8607:
process_replay is now passing in #20297.

PR #20304, opened 2021-03-11
tuning simulator parameters

iejMac:

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -17,9 +17,11 @@ from selfdrive.test.helpers import set_params_enabled
 
 parser = argparse.ArgumentParser(description='Bridge between CARLA and openpilot.')
 parser.add_argument('--joystick', action='store_true')
-parser.add_argument('--town', type=str, default='Town04')
+parser.add_argument('--town', type=str, default='Town04_Opt')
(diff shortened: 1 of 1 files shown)


tools/sim/bridge.py
```diff
@@ -17,9 +17,11 @@
 
 parser = argparse.ArgumentParser(description='Bridge between CARLA and openpilot.')
 parser.add_argument('--joystick', action='store_true')
-parser.add_argument('--town', type=str, default='Town04')
+parser.add_argument('--town', type=str, default='Town04_Opt')
 parser.add_argument('--spawn_point', dest='num_selected_spawn_point',
         type=int, default=16)
+parser.add_argument('--quality', type=str, default='Regular')
```
adeebshihadeh, line 23: make this a bool and call it something like low_quality

tools/sim/bridge.py
```diff
@@ -135,7 +135,7 @@ def bridge(q):
   client.set_timeout(10.0)
   world = client.load_world(args.town)
 
-  if args.quality == "Low":
+  if args.low_quality is True:
```
adeebshihadeh, line 138:
```suggestion
  if args.low_quality:
```

PR #20310, opened 2021-03-11
camerad: refactor RGBToYUVState into class

deanlee:


selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -139,19 +136,17 @@ bool CameraBuf::acquire() {
...
+                               cur_rgb_buf->len, 0, 0, nullptr));
   }
 
-  clWaitForEvents(1, &debayer_event);
-  CL_CHECK(clReleaseEvent(debayer_event));
-
   cur_yuv_buf = vipc_server->get_buffer(yuv_type);
-  rgb_to_yuv_queue(&rgb_to_yuv_state, q, cur_rgb_buf->buf_cl, cur_yuv_buf->buf_cl);
+  rgb2yuv->queue(q, cur_rgb_buf->buf_cl, cur_yuv_buf->buf_cl);
+  CL_CHECK(clFinish(q));
```
adeebshihadeh, line 149: this isn't a null effect change and can easily have unintended side effects. since this queue isn't exclusive to rgb2yuv, this now waits for everything to finish instead of just `debayer_event`

PR #20318, opened 2021-03-11
set_exposure_target test

ZwX1616:
API is cleaned up a little for easier testability

<s>needs some more clean up and add gt</s>


.github/workflows/selfdrive_tests.yaml
```diff
@@ -207,7 +207,8 @@ jobs:
                         $UNIT_TEST selfdrive/locationd && \
                         $UNIT_TEST selfdrive/athena && \
                         $UNIT_TEST selfdrive/thermald && \
-                        $UNIT_TEST tools/lib/tests"
+                        $UNIT_TEST tools/lib/tests && \
+                        scons -j$(nproc) --test selfdrive/camerad && ./selfdrive/camerad/test/ae_gray_test"
```
adeebshihadeh, line 211: just add --test to the scons call on top

Issue #20232, opened 2021-03-03
Torque and steering angle rate values wrong for some cars


HaraldSchafer:
Ideally we would get from every car:
- Steering angle in degrees
- Steering angle rate in degrees/sec
- Human torque (torque human exerts on steering wheel) in NM
- EPS torque (the torque added by the EPS on the steering wheel) in NM

Human torque and steering angle are used in openpilot, so I assume those are correct. The other two seem to be wrong on many cars. I looked at some data to find the cars on which they are wrong, this isn't necessarily complete, but should give good picture. 

### Steering angle rate problems
**Sign flipped on these cars**
- [x] HONDA ODYSSEY 2018 EX-L 
- [x] HONDA FIT 2018 EX 
- [x] HONDA RIDGELINE 2017 BLACK EDITION
- [x] HONDA CR-V 2016 TOURING
- [x] HONDA PILOT 2017 TOURING
- [x] HONDA PILOT 2019 ELITE
- [x] HONDA HRV 2019 TOURING
- [x] ACURA ILX 2016 ACURAWATCH PLUS

**Strictly positive on these cars**
- [ ] HYUNDAI GENESIS 2015-2016
- [ ] GENESIS G70 2018
- [ ] HYUNDAI IONIQ HYBRID 2017-2019
- [ ] HYUNDAI SONATA 2019
- [ ] KIA NIRO EV 2020
- [ ] KIA STINGER GT2 2018
- [ ] HYUNDAI SANTA FE LIMITED 2019

**Not set on these cars**
- [ ] SUBARU ASCENT LIMITED 2019
- [ ] SUBARU FORESTER 2019
- [ ] SUBARU IMPREZA LIMITED 2019
- [x] CHEVROLET VOLT PREMIER 2017 
- [ ] NISSAN ROGUE 2019
- [ ] NISSAN LEAF 2018
- [ ] NISSAN X-TRAIL 2017
- [ ] SUBARU OUTBACK 2015 - 2017

### EPS Torque problems
**Strictly positive on these cars**
- [ ] HONDA ODYSSEY 2018 EX-L
- [ ] HONDA CR-V 2017 EX 
- [ ] HONDA FIT 2018 EX 
- [ ] HONDA ACCORD 2018 LX 1.5T
- [ ] HONDA RIDGELINE 2017 BLACK EDITION
- [ ] HONDA ACCORD 2018 SPORT 2T
- [ ] HONDA ACCORD 2018 HYBRID TOURING
[...]

PR #20319, opened 2021-03-11
Add GM steering rate and EPS torque to CarState

hewers:
**Description** 

Add steering rate, and EPS torque, to GM vehicles like Chevrolet Volt Premier 2017.

Sub-fix for https://github.com/commaai/openpilot/issues/20232 , adding torque and steering rates to multiple vehicles.

**Verification**

- ~~Deploy and drive route~~
- ~~Verify by plotting CarState fields~~

angle
torque

**Route**

4283e5cdbb717eff|2021-03-11--16-47-49


qadmus:
- Rebased
- Removed opendbc bump (steer rate scale can come later)
- Updated refcommit (not sure why, AUDI A3 was failing?)

qadmus:
Reverted refcommit. process replay still fails. I don't know what I'm doing.

jyoung8607:
Looked at this since I tweaked VW steering signals very recently... reading the log can be a little confusing. Audi A3 happens to be mentioned last because it's the last replay segment to be tested, but the segment that fails replay is actually for the GM Volt.

https://github.com/commaai/openpilot/blob/d023397ba94cd1b2a47acf0e8b900cd8d53eb360/selfdrive/test/process_replay/test_processes.py#L14-L24

```
***** results for segment ae3ed0eb20960a20|2021-01-15--15-04-06--8 *****
	controlsd
		carState.steeringRateDeg: 1044
		carState.steeringTorqueEps: 5909
		controlsState.lateralControlState.pidState.steeringRateDeg: 1044
	radard
	plannerd
	calibrationd
	dmonitoringd
	locationd
	paramsd
	ubloxd
```

Replay mismatch is actually expected and desired for what you're doing, so should just be a matter of bumping the replay ref once you're sure you're happy with it.

qadmus:
@jyoung8607 First time updating the ref, is there a guide (or I'll write one)?

- Anything special required about the route?
- I've preserved the example route, should that always be preserved or is it automatically preserved?
- Replace Volt route in test_processes.py : segments = []
- Run update_ref.py, commit ref_commit

jyoung8607:
I'm not sure you need to replace the current CI route for this purpose. I believe a Comma staffer has to do the actual replay ref bump for you, because to actually use the new replay data in CI, `update_refs` has to send the new baseline logs to an Azure bucket only they can write to.

adeebshihadeh:
Thanks for the PR!

PR #20334, opened 2021-03-12
adding unit testing for simulator

iejMac:

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -208,7 +208,7 @@ def bridge(q):
 
   vc = carla.VehicleControl(throttle=0, steer=0, brake=0, reverse=False)
 
-  is_openpilot_engaged = False
+  is_openpilot_engaged = True
   throttle_out = steer_out = brake_out = 0
   throttle_op = steer_op = brake_op = 0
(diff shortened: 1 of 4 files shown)


tools/sim/bridge.py
```diff
@@ -230,6 +231,12 @@ def destroy():
     throttle_manual = steer_manual = brake_manual = 0
 
     # --------------Step 1-------------------------------
+    if sm.updated["controlsState"] and not initial_speed_set:
```
adeebshihadeh, line 234: add a comment here

tools/sim/bridge.py
```diff
@@ -208,7 +208,7 @@ def destroy():
 
   vc = carla.VehicleControl(throttle=0, steer=0, brake=0, reverse=False)
 
-  is_openpilot_engaged = False
+  is_openpilot_engaged = True
```
adeebshihadeh, line 211: think openpilot should always be engaged in the sim, there's a bunch of hacks right now to make manual cotnrol work

adeebshihadeh: CARLA already has support for manual control anyway, don't think there's value in duplicating that

PR #20335, opened 2021-03-12
Improved FPv1 for SONATA_LF

ErichMoraga:
@jpwillett#7533 2019 Hyundai Sonata limited 2.0T  DongleID/route af26f4e5a30e0b53|2021-03-12--17-31-15
FPv2 is very problematic for this car...
https://discord.com/channels/469524606043160576/524611978208215070/819982100539113512

carParams generally shows something different every drive, in the logs.  Works great with extra msgs. in the CAN msg. based FPv1.


adeebshihadeh:
I'd prefer not to merge any new CAN fingerprints for HKG. There isn't an issue with the FW fingerprints - that's a bug with capnproto and dump.py.

ErichMoraga:
"There isn't an issue with the FW fingerprints - that's a bug with capnproto and dump.py."

Yeah, I totally get that, and mentioned it to the customer a few hours ago.  Not sure if/when the bug will be resolved, though improving PFv1 is pretty harmless, as it only required adding a few extra CAN msgs.

adeebshihadeh:
It's fixed: https://github.com/commaai/openpilot/blob/master/selfdrive/debug/dump.py#L8.

adeebshihadeh:
Closing for now, reopen or submit a new PR when we have FW fingerprints

ErichMoraga:
@adeebshihadeh I switched him over to `master-ci`, so it could dump `carParams` properly, but instead...
`carParams: Reader was evicted`

Only one person mentioned anything like that (in our private beta channel), and based on the feedback from that, I increased the max number of readers in `cereal/messaging/msgq.hpp` from 8 to 12, but that didn't help.

PR #20341, opened 2021-03-14
Honda CRV 2021 firmware versions

honkergoose:
Dongle Id: f56ab3f0035b2b7b

Honda CRV 2021 works out of the box with Comma Two but occasionally when the car is turned off and then turned back on later the message "Car Unrecognized: Dashcam Mode" appears. This is resolved with a restart but having to restart often is annoying.

I followed the instructions for Fingerprinting 2.0 to update the firmware versions and hopefully prevent this issue.


vanillagorillaa:
Would you mind putting the new versions in order with existing ones?

honkergoose:
@vanillagorillaa Updated! Thanks!

PR #20343, opened 2021-03-14
use class AlignedBuffer from cereal

deanlee:
use class `AlignedBuffer` from cereal/messaging.hpp : https://github.com/commaai/cereal/pull/125


pd0wm:
Merged cereal PR and bumped submodule

deanlee:
Done

PR #20344, opened 2021-03-14
dmoniitoring: new function crop_yuv

deanlee:


selfdrive/modeld/models/dmonitoring.cc
```diff
@@ -41,70 +41,67 @@ static inline auto get_yuv_buf(std::vector<uint8_t> &buf, const int width, int h
...
-  const int global_x_offset = full_width_tici / 2 - adapt_width_tici / 2;
-  const int global_y_offset = full_height_tici / 2 - cropped_height / 2;
-  const int crop_x_offset = adapt_width_tici - cropped_width + 32;
-  const int crop_y_offset = -196;
+  Rect crop_rect = {full_width_tici / 2 - adapt_width_tici / 2,
+                    full_height_tici / 2 - cropped_height / 2 - 196,
+                    cropped_height / 2,
+                    cropped_height};
+  if (!s->is_rhd) {
+    crop_rect.x += adapt_width_tici - crop_rect.w + 32;
```
ZwX1616, line 73: 32 only goes into non-rhd?

deanlee: yes,I think it's the same as before

ZwX1616: never mind. that's fine. my bad :)

PR #20351, opened 2021-03-15
fix update_lead crash

deanlee:


pd0wm:
Thanks for fixing this. Do  you have any idea of the underlying issue that's causing this? We can't start adding band-aids everywhere.

deanlee:
The problem in this case is caused by not received any `modelV2` message before `radarState`,which is not the same as the `sensorEvents` problem.

pd0wm:
Ah, ok. Any reason why you're checking the raw size then instead of the list length?

deanlee:
I made an assumption here, as long as there is data, the length of getZ() and getX() are both TRAJECTORY_SIZE, otherwise I have to check the two lists returned by getZ and getX

pd0wm:
In other places we usually check if the recv frame is non zero to see if the message was ever received.

selfdrive/ui/ui.cc
```diff
@@ -82,14 +82,17 @@ static int get_path_length_idx(const cereal::ModelDataV2::XYZTData::Reader &line
...
 
-static void update_lead(UIState *s, const cereal::RadarState::Reader &radar_state,
-                        const cereal::ModelDataV2::XYZTData::Reader &line, int idx) {
-  auto &lead_data = s->scene.lead_data[idx];
-  lead_data = (idx == 0) ? radar_state.getLeadOne() : radar_state.getLeadTwo();
-  if (lead_data.getStatus()) {
-    const int path_idx = get_path_length_idx(line, lead_data.getDRel());
-    // negative because radarState uses left positive convention
-    calib_frame_to_full_frame(s, lead_data.getDRel(), -lead_data.getYRel(), line.getZ()[path_idx] + 1.22, &s->scene.lead_vertices[idx]);
+static void update_leads(UIState *s, const cereal::RadarState::Reader &radar_state) {
```
pd0wm, line 85:
```suggestion
static void update_leads(UIState *s, const cereal::RadarState::Reader &radar_state, const cereal::ModelDataV2::XYZTData::Reader &line, bool model_rcvd) {
```

Looking at the function signature I don't think it's clear that the modelData is being used as input. Think it's best to keep passing in the model data and the `model_rcvd` flag seperately

pd0wm: Or use `std::optional<const cereal::ModelDataV2::XYZTData::Reader>` if you're feeling fancy

deanlee: coool

deanlee:
the reason that sensorEvents reads an empty list is possibly that the number of connections exceeds the limit (NUM_READERS) :

pd0wm:
Why do you think that could cause this? How many subscribers are there?

deanlee:
There are at least 4(or 5?) , it's likely to exceed the limit  if start more UI processes

pd0wm:
When I was testing I only ran sensord an ui instances.

If the number of readers is exceed warning messages are printed. Also a disconnected reader wouldn't be able to read a message anymore, it would just try to reconnect.

When I saw the bug the capnproto message seemed valid and had the correct union set for the specific sensor and also the valid flag was true, however the list with measurements was empty.

deanlee:
OIC.I understood it wrong.

PR #20380, opened 2021-03-17
2021 Pacifica Hybrid values

vanillagorillaa:
From 84db504dd06e33f6|2021-03-16--10-26-30

Seems as if there are some messages only seen when car is remotely started.


vanillagorillaa:
Test failing due to matching fingerprint within the 2019 Pacifica Hybrid, may need to combine into one definition. Will review tomorrow.

Issue #20263, opened 2021-03-07
Offroad UI enhancements


adeebshihadeh:
- [x] Show username for current SSH keys
- [x] Refresh developer page periodically
- [x] Cache home screen state (drive stats, prime status, etc.)
  - [x] show cached version instead of "No Internet connection" when offline
- [x] Click outside of confirmation dialog to dismiss
- [x] Close confirmation dialogs after settings timeout
- [x] Make driver view fast
- [x] Don't shift drive stats down when alert/update button is shown
- [x] Description text in settings
- [x] Close offroad alerts when all alerts go away
- [x] Scrolling in offroad alerts + release notes
- [x] Disable driver view button while onroad
- [x] Close android activities when exiting settings
- [x] Show calibration values

BrevanJ04:
- [ ] WPA2-EAP support
- [ ] Remove saved wifi networks
- [ ] UI change to denote the difference between connecting over wifi and ethernet

doktorsleepelss:
* [ ]  View mount placement degrees like on C2

jyoung8607:
- [ ] Optionally mask Wifi password during entry

speedking456:
- [ ] Reset SSH keys to default

michaelhonan:
- [ ] comma prime upgrade widget overlaps the home statistics view.

grekiki:
> * [ ]  comma prime upgrade widget overlaps the home statistics view.
I think that's only true on PC because it uses the narrower screen(https://github.com/commaai/openpilot/blob/fb5873bb6d00b4f625e9ee0cd1767fe8805b2476/selfdrive/ui/qt/qt_window.hpp#L16)

PR #20382, opened 2021-03-17
developer_pane(): use structured binding in loop

deanlee:


adeebshihadeh:
These small, opinionated PRs take a not insignificant time to review and there's nothing necessarily better or cleaner about this. If you want to work on the UI though, there's a bunch of things to work on in https://github.com/commaai/openpilot/issues/20263.

PR #20387, opened 2021-03-17
Hyundai Ioniq PHEV 2020

baldwalker:
added test route

cfad75360cd08054|2021-03-16--18-48-19

Fingerprint v1 and v2.0

can't get steering to work without setting gas to not disengage OP in selfdrive/boardd/boardd.cc and selfdrive/car/interfaces.py


pd0wm:
Sounds like you need to find a different gas pedal value for your car. Panda/controls probably thinks the gas is always pressed.

baldwalker:
> Sounds like you need to find a different gas pedal value for your car. Panda/controls probably thinks the gas is always pressed.

this is the second Ioniq Phev 2020 I've tried to get working for people and get same problem with both, how can I sort the gas pedal value out?

pd0wm:
Try using live cabana while parked and try to find another signal which correlates to pressing the gas pedal. Quickly looked in cabana for you, but find anything that stood out.

baldwalker:
> Try using live cabana while parked and try to find another signal which correlates to pressing the gas pedal. Quickly looked in cabana for you, but find anything that stood out.

I don't have access to the vehicle and looking at live cabana data is a bit over my head, might not be able to help

speedking456:
Do you have any errors on the dash?

Zach4422:
Im not very technically proficient in these matters but this is my car so would be willing to help in whatever way I can. 

No errors on my dash. The closer to stock fork simply didn't steer when OP was engaged. Tried it at lower and higher speeds.

baldwalker:
> Do you have any errors on the dash?

No errors on dash, will only steer when changing safety param to 1 and not disengaging op on gas

TK211X:
Oh yea, the signal you're looking for is on EV_PC / HEV_PC / PHEV_PC (Whichever variant you have, PHEV in this case) signal on C-CAN forwarded by CGW reported from P-CAN PCU. The pedal signal is reported directly to the PCU.

Pedal ---> PCU ---> PCAN ---> CGW ---> CCAN

It will cause a signal collision if you mark it on the public generic .dbc you have for all the HKMC vehicles. Separate power source vehicles have the PCU report the _PC formatted slightly differently. The proper trigger on CAN was found to differentiate the type vehicle so that this wouldn't have been an issue. Comma did not accept the proposed change because they wanted a universal signal across ALL HKMC vehicles. Yea, no.

If you like, try finding the specific PR that had the discussion and all involved details with the working code changes..... and try to get comma to upstream the fix for your vehicle.

************************
*Disclaimer* All related information seen here is prior to 01/11/2020 while I was still involved in the open source project. 
This information was derived / available at that time.

YawWashout:
Yeah this is my car but I can’t say I understand half of what was mentioned here to fix it.

happy to test anything for someone though!

PR #20392, opened 2021-03-18
terms fixups

iejMac:


selfdrive/ui/qt/offroad/onboarding.cc
```diff
@@ -75,7 +75,7 @@ QWidget* OnboardingWindow::terms_screen() {
   buttons->addWidget(new QPushButton("Decline"));
   buttons->addSpacing(50);
 
-  QPushButton *accept_btn = new QPushButton("Scroll to accept");
```
adeebshihadeh, line 78: only change this when the button is enabled

selfdrive/ui/qt/offroad/onboarding.cc
```diff
@@ -105,6 +105,8 @@ QWidget* OnboardingWindow::terms_screen() {
 #endif
   QObject::connect(sb, &QScrollBar::valueChanged, [sb, accept_btn]() {
     accept_btn->setEnabled(accept_btn->isEnabled() || (sb->value() == sb->maximum()));
+    if (sb->value() == sb->maximum())
```
adeebshihadeh, line 108: always use brackets and move the set enable here

PR #20393, opened 2021-03-18
QT: scrollable toggles panel, show description on click 

deanlee:


adeebshihadeh:
Nice, does this look like the current UI?

deanlee:
panel

adeebshihadeh:
Looks good :)

adeebshihadeh:
Tried it out on a device, and I like it! The scrolling doesn't feel quite right, but @iejMac is fixing that in https://github.com/commaai/openpilot/pull/20392. Waiting on fixing scrolling on QCOM2 to merge.

PR #20395, opened 2021-03-18
Qt: cache home screen state

deanlee:


selfdrive/ui/qt/widgets/drive_stats.cc
```diff
@@ -43,19 +43,23 @@ QLayout* build_stat(QString name, int stat) {
...
 }
 
-void DriveStats::parseError(QString response) {
-  clearLayouts(vlayout);
-  vlayout->addWidget(new QLabel("No Internet connection"), 0, Qt::AlignCenter);
-}
+void DriveStats::parseResponse(QString response, bool save) {
+  static QString prev_response;
+  response = response.trimmed();
+  if (prev_response == response) {
```
adeebshihadeh, line 49: hmm, we shouldn't need to do this. maybe build the layout in the constructor and just update the text values in this function?

adeebshihadeh: updating the text should be super cheap, Qt probably already does this if when updating values

pd0wm:
If you create a new param also make sure to add it to params.pyx (https://github.com/commaai/openpilot/blob/master/common/params_pyx.pyx#L16)

adeebshihadeh:
Can we do this more generically with the API class and RequestRepeater? Maybe they should take an arg in the constructor to first use the cached version if it exists? Then this would also cover the Prime widget too and anything else we end up using the API for.

common/params_pyx.pyx
```diff
@@ -16,6 +16,7 @@ cdef enum TxType:
 keys = {
   b"AccessToken": [TxType.CLEAR_ON_MANAGER_START],
   b"AthenadPid": [TxType.PERSISTENT],
+  b"CachedDriveStats": [TxType.PERSISTENT],
```
adeebshihadeh, line 19: can we make one param called something like api cache and use it for all of them?

deanlee: A json file containing all the results of different api calls?

PR #20396, opened 2021-03-18
Qt: update developer page when it's visible

deanlee:


adeebshihadeh:
Can we only update this when settings is opened? We're already using a bunch of timers, and I think the number of timers can easily get out of control since they're so easy to use.

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -217,37 +222,55 @@ QWidget * device_panel() {
...
+}
+
+void DeveloperPanel::updatePanel() {
+  auto dev_params = getParams();
+  for (auto &[name, value] : dev_params) {
+    labels.at(QString::fromStdString(name))->setContent(QString::fromStdString(value));
+  }
+}
+
+std::vector<std::pair<std::string, std::string>> DeveloperPanel::getParams() {
```
adeebshihadeh, line 260: why not put this in updatePanel?

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -217,37 +222,55 @@ QWidget * device_panel() {
...
     }
   )");
-  return widget;
+}
+
+void DeveloperPanel::showEvent(QShowEvent *event) {
+  updatePanel();
+}
+
+void DeveloperPanel::updatePanel() {
```
adeebshihadeh, line 253: rename to update? we'll need update functions for a lot of our widgets, so think it makes sense to call them all update

selfdrive/ui/qt/offroad/settings.cc
deanlee: Because I also create widgets from params in the constructor to simplify the initialization.

PR #1297, opened 2020-03-31
2020 Chrysler Pacifica

pd0wm:

PR #20399, opened 2021-03-18
include PACIFICA_2020 in late-model behavior

adhintz:
These models need to turn off gone_fast_yet bit at -3 to avoid LKAS faults.

This was likely an accidental omission from PR #1297 

To avoid future similar mistakes, we could define a list of these cars in values.py instead of having the list in multiple places.


adeebshihadeh:
Merging in this fix, but a list in values.py sounds good.

PR #20401, opened 2021-03-19
add color correction for eon

ihsakashi:


adeebshihadeh:
What is this color correction for?

ihsakashi:
It's just for people who have color blindness / weakness.

brianandyt:
This is super exciting wasn’t expecting accessibility features to be added till 0.9

He means color vision deficiency(color blindness isn’t a medical term). These are just filters for people with Anomalous trichromatism.

Because openpilot relays on so heavily on its colored boarders for relaying information to the user we need make sure the user actually see them. This become even more import when your team added traffic lights which aren’t standardize throughout the US.

adeebshihadeh:
We're not going to merge this for a couple reasons: it relies on android, and it's another setting in the UI. That being said, we're working on some new designs and will take this into consideration.

rav4kumar:
Interesting,  geo wants to sell overpriced hardware because of the software.... okay, yet you guys cannot even merge this to master because of one more things in the settings? lame. how are you going to sell the hardware to people who are visually impaired? where is the support for i18? why do people have to look "fork" but yet geo out here saying "we're upping the price because of the software" should be shame. you cannot even add simple little thing that could be a big thing. just because you are not visually impaired does not mean its a silly thing. classic comma.

adeebshihadeh:
This is consistent with our historical position on toggles and settings. We intend to address the issue in a design rework, and accessibility will be a goal in the new design.

ReFil:
An accessibility feature is always going to need a toggle, I don't understand why that is a gateway to improving the accessibility of openpilot

alexcomma:
We are working towards a design that accommodates for vision deficiency. If the design alone is not sufficient, then we will add a toggle. @ReFil an accessibility feature is not always going to need a toggle. Often times designs are too complex or too rigid that a toggle becomes the easiest solution (hence, its popularity). openpilot is optimized for stability/longevity and that includes mitigating maintenance costs; meaning as few dependencies as possible which is why we are going to pass on this PR, but not this feature, at this time.

alexcomma:
@rav4kumar GitHub is for development, not a public forum for feedback (maybe reddit?). Going forward, if your comments are not subject related, they may be deleted. It's 2021, everything is controversial, but our focus is working with those that believe in what we are doing and how we are doing it.

PR #20403, opened 2021-03-19
QT:show current calibration values in reset dialog

deanlee:
This is a temporary solution because there is no suitable place to show. the reset widget needs to be rewritten in the future.


adeebshihadeh:
Closing this since it should be fixed with #20417

PR #20404, opened 2021-03-19
Add mac requirements

ghost:
* Fix mac requirements
* Add `cryptography` package dependencies
* Rust probably needs to be setup properly

See [https://github.com/pyenv/pyenv/issues/1740](https://github.com/pyenv/pyenv/issues/1740) for more information regarding `zlib` and `bzip2` pyenv fix.

Tested on MacOS Big Sur. (Some environment variables are left out.)


adeebshihadeh:
Why is rust necessary? And I assume this was tested on an Intel Mac?

BraunPhilipp:
Building `cryptography` requires `rust`. More info on [https://github.com/pyca/cryptography/blob/main/docs/installation.rst](https://github.com/pyca/cryptography/blob/main/docs/installation.rst). 

Tested on Intel MacBook with MacOs Big Sur.

adeebshihadeh:
I see, guess it works in CI since rust is probably pre-installed on the runners. Can you add a comment to the rust line?

PR #20412, opened 2021-03-20
fix prius lame steering

cydia2020:
**Description** 
After upgrading to 0.8.2, OP was not able to handle curves as gracefully as before, this is due to actuator_delay being too small causing OP to send steer commands later than it should; this PR aims to fix this.

**Verification**
Was on A440 in Victoria, Australia last night, OP can't do any curves before applying this, it handled curves much better after applying this.

**Route**
Route: please test with your prius since my device is banned.


HaraldSchafer:
We determined all the lag values by doing data analysis on thousands of minutes of driving. So we feel pretty confident in them. 

That said, 0.8.2 did make controls more lazy overall and increasing the delay will make it less lazy. Especially the prius is sensitive to this stuff. But we want to deal with this is a more general way, we have an overhaul in mind that should improve the prius a lot.

PR #20413, opened 2021-03-20
more camerad jenkins test

ZwX1616:


selfdrive/camerad/test/check_exposure.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
```
adeebshihadeh, line 1: rename the file to `test_exposure.py` to match the convention used throughout the codebase

selfdrive/camerad/test/check_exposure.py
```diff
@@ -0,0 +1,56 @@
...
+    i_mean = np.mean(i) / 255.
+    print([i_median, i_mean])
+    return med_ex[0] < i_median < med_ex[1] and mean_ex[0] < i_mean < mean_ex[1]
+
+
+  @with_processes(['camerad'])
+  def test_camera_operation(self):
+    print("checking image outputs")
+
+    time.sleep(WAIT_TIME)
```
adeebshihadeh, line 45: instead of this fixed sleep, can we just check an image every second until the check passes?

selfdrive/camerad/test/test_exposure.py
```diff
@@ -0,0 +1,68 @@
...
+  def test_camera_operation(self):
+    print("checking image outputs")
+
+    start = time.time()
+    passed = 0
+    while(time.time() - start < TEST_TIME and passed < REPEAT):
+      rpic, dpic = get_snapshots(frame="roadCameraState", front_frame="driverCameraState")
+
+      res = self._is_exposure_okay(rpic)
+      res &= self._is_exposure_okay(dpic)
```
adeebshihadeh, line 52: use a logical and

adeebshihadeh: or just make `_is_exposure_okay` assert the validity instead of returning it

PR #20414, opened 2021-03-20
Add several missing f/w for SONATA_LF

ErichMoraga:
`@jpwillett#7533` 2019 Hyundai Sonata limited 2.0T DongleID/route `af26f4e5a30e0b53|2021-03-19--12-27-17`

FPv2 stabilized after switching to `master-ci`.  See also https://github.com/commaai/openpilot/pull/20335 regarding original FPv1 PR.


adeebshihadeh:
That route doesn't fingerprint as a SONATA LF with this patch. ESP and transmission FW seem to be wrong.

ErichMoraga:
`SONATA_LF = "HYUNDAI SONATA 2019"`

That route isn't patched with the f/w changes I made (fingerprinting via FPv1), because I couldn't get back in to add the patch.  If that's what I need to do, then I'll ask the user to let me back in.

ErichMoraga:
We played around with it a bit more, and it doesn't always dump the same firmware, and when there's a match, it doesn't really seem to match.  I give up for now...

PR #20415, opened 2021-03-20
QT: reduce offroad CPU usage from ~15% to ~4%

deanlee:


selfdrive/ui/qt/home.cc
```diff
@@ -284,7 +284,9 @@ void GLWindow::timerUpdate() {
   handle_display_state(&ui_state, false);
 
   ui_update(&ui_state);
-  repaint();
+  if (ui_state.vipc_client->connected) {
```
adeebshihadeh, line 287: hmm, this wouldn't update the sidebar at all while offroad

deanlee: Yes it is. This is indeed a problem

PR #20417, opened 2021-03-20
Qt: use the AbstractControl for multiple types of controls

deanlee:


adeebshihadeh:
Nice work. Can we do this in the following order: get this merged as a minimal refactor with no functional changes (visual changes are fine), then we can rebase #20393 to get scrolling and descriptions?

adeebshihadeh:
Hmm, maybe we can clean this up by using a Qt table or form layout?

deanlee:
still working on the margins & paddings

adeebshihadeh:
It doesn't really make sense for the line to be part of individual "control" widget. Ideally, a nice settings panel or layout class would handle that and the scroll area for us.

PR #20420, opened 2021-03-21
Add mac requirements

ghost:
* Fix mac requirements
* Add `cryptography` package dependencies

**For more information:**
* [https://github.com/pyenv/pyenv/issues/1740](https://github.com/pyenv/pyenv/issues/1740)
* [https://github.com/pyca/cryptography/blob/main/docs/installation.rst](https://github.com/pyca/cryptography/blob/main/docs/installation.rst)

Tested on MacOS Big Sur and MacOS Catalina. (Messed up rebase in other PR.)

### DIFF
--- a/tools/mac_setup.sh
+++ b/tools/mac_setup.sh
@@ -7,6 +7,11 @@ if [[ $(command -v brew) == "" ]]; then
 fi
 
 brew bundle --file=- <<-EOS
+brew "cmake"
+brew "zlib"
+brew "bzip2"
+brew "rust"
+brew "rustup-init"
 brew "capnp"
 brew "coreutils"
 brew "eigen"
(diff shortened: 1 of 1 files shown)


tools/mac_setup.sh
```diff
@@ -28,6 +33,24 @@ elif [[ $SHELL == "/bin/bash" ]]; then
   RC_FILE="$HOME/.bash_profile"
 fi
 
+# Build requirements for macOS
+# https://github.com/pyenv/pyenv/issues/1740
+# https://github.com/pyca/cryptography/blob/main/docs/installation.rst
+rustup-init -y
+
+if [ -n "$RC_FILE" ]; then
+  echo 'export LDFLAGS="$LDFLAGS -L/usr/local/opt/zlib/lib"' >> $RC_FILE
```
adeebshihadeh, line 42: it's not great to spam a user's bashrc. does brew not handle this for us?

BraunPhilipp: Unfortunately not. An alternative would be to only define the flags for `pyenv install`. `openssl` and `rust` would need to stay probably.

PR #20421, opened 2021-03-21
hw.h: trim os version string

deanlee:
the version string ends with a line terminator.cause the display of the `os version` in the devel panel  to be misplaced.


adeebshihadeh:
Is this still necessary?

deanlee:
I think it is still necessary, otherwise there will still be alignment problem in developer panel

adeebshihadeh:
fixed in #20434

PR #20422, opened 2021-03-21
no shift when update button pops up

iejMac:


adeebshihadeh:
Moved the button to the top header in the home screen instead.

PR #20423, opened 2021-03-21
Qt TrainingGuide: use QRect[] for bonding boxes

deanlee:


adeebshihadeh:
The training guide will be completely redone very soon, so doesn't make sense to merge any updates for it.

PR #20425, opened 2021-03-21
apply_steer rate limit comparison needs int(round( initial_steer ))

hewers:
**Description** CC.steer_rate_limited (thus CS.steeringRateLimited) was almost always true on GM, because the original float rarely equals its rounded value.

**Verification** Not tested yet, just inspection. I'll confirm on GM.

Alternative fix (I prefer it) is to return boolean from apply_std_steer_torque_limits() when limiting occurs.
`
apply_steer, limited = apply_std_steer_torque_limits(apply_steer, self.apply_steer_last, CS.out.steeringTorque, P)`

### DIFF
--- a/selfdrive/car/chrysler/carcontroller.py
+++ b/selfdrive/car/chrysler/carcontroller.py
@@ -24,7 +24,7 @@ class CarController():
 
     # *** compute control surfaces ***
     # steer torque
-    new_steer = actuators.steer * CarControllerParams.STEER_MAX
+    new_steer = int(round(actuators.steer * CarControllerParams.STEER_MAX))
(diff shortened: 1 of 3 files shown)


pd0wm:
Good catch! This probably prevented the alert to show up in cases where the steering didn't hit the limit yet.

qadmus:
Revert? @pd0wm this breaks Hyundai 2015 `LKAS11` `CR_Lkas_StrToqReq` which uses float Nm, and is int(round()) to zero :( 

Almost all cars uses full integers for steering commands, except `hyundai_2015_ccan.dbc`. Found this while trying to normalize steer commands.

pd0wm:
Good to know though. When merging I went through all cars to make sure they were integers, but was worried I missed one for a second.

PR #20427, opened 2021-03-22
2021 RAV4 Hybrid Value

jimmybnguyen:
**Fingerprint**

Added missing ecu.engine value for RAV4H_TSS2


adeebshihadeh:
What's your dongle ID?

jimmybnguyen:
> What's your dongle ID?
a4e50d70209cd491

PR #20436, opened 2021-03-22
Skoda Superb B8

netadvanced:
Car port for Skoda Superb III B8 (2015-2019)

**Checklist**
- [X] added to README
- [X] test route added to [test_car_models](../../selfdrive/test/test_car_models.py)
- [X] route with openpilot:

Note: I've also been testing the "universal" fingerprint for Volkswagen but it seems to disturb my radio and some accessories randomly, no more problems with the proper fingerprint.


README.md
```diff
@@ -169,6 +169,7 @@ Community Maintained Cars and Features
 | Nissan    | Leaf 2018-20                  | ProPILOT          | Stock            | 0mph               | 0mph         |
 | Nissan    | Rogue 2018-19                 | ProPILOT          | Stock            | 0mph               | 0mph         |
 | Nissan    | X-Trail 2017                  | ProPILOT          | Stock            | 0mph               | 0mph         |
+| Skoda     | Superb III B8 2015-2019       | Driver Assistance | Stock            | 0mph               | 0mph         |
```
adeebshihadeh, line 172: only put the years for which the port has been tested

selfdrive/test/process_replay/test_processes.py
```diff
@@ -21,7 +21,7 @@
   ("SUBARU", "4d70bc5e608678be|2021-01-15--17-02-04--5"),      # SUBARU.IMPREZA
   ("GM", "ae3ed0eb20960a20|2021-01-15--15-04-06--8"),          # GM.VOLT
   ("NISSAN", "e4d79cf6b8b19a0d|2021-01-17--14-48-08--7"),      # NISSAN.XTRAIL
-  ("VOLKSWAGEN", "e2a273d7e6eecec2|2021-03-03--16-05-26--4"),  # VW.AUDI_A3
+  ("VOLKSWAGEN", "c65dd238d105a170|2021-03-21--02-47-29--8"),  # VW.SKODA_SUPERB_B8
```
adeebshihadeh, line 24: revert this

jyoung8607:
I don't recommend adding any more legacy FPv1 models for the VW family. The reason we haven't upstreamed these in the past is because we ran into enormous amounts of fingerprint overlap almost immediately. It's not sustainable.

@netadvanced, mind popping into Discord #volkswagen and telling us more about the radio problem? Many VW radios and/or audio amps will indeed known to go into a special diagnostic mode and cut off normal audio when a UDS diagnostic session is opened, but absent us probing that specific address (which we don't in normal operation) and keeping a diagnostic session open (we which don't), there shouldn't be a problem.

adeebshihadeh:
We'll wait to merge this and try to get #20271 before 0.8.3 goes to release.

README.md
netadvanced: All done

jyoung8607:
@netadvanced, if you'll update to current commaai:master, we've got FPv2 support for the Superb set up and ready. You should be able to open a PR with your own carFw values out of any uploaded route.

https://github.com/commaai/openpilot/blob/ed3add5ae600e3eaf43234eb2ce669a7213afd0d/selfdrive/car/volkswagen/values.py#L325-L339

https://github.com/commaai/openpilot/wiki/Fingerprinting

netadvanced:
@jyoung8607 Thanks, I'll give it a try

adeebshihadeh:
Let's close this and open a new PR once you have the FW versions.

PR #20438, opened 2021-03-22
support for Tesla Model S with AutoPilot (AP1/AP2)

BogGyver:
Car port for Tesla Model S with AP1 (2014-2016) and AP2 (2016-2019). 

**Checklist**
- [x] added to README
- [ ] route with openpilot:

Also compatible (after fingerprint and configuration, but without code changes) with Tesla Model X with AP1 (2016) and AP2 (2016-2018)

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "panda"]
 	path = panda
-	url = ../../commaai/panda.git
+	url = ../../boggyver/panda.git
 [submodule "opendbc"]
 	path = opendbc
-	url = ../../commaai/opendbc.git
+	url = ../../boggyver/opendbc.git
 [submodule "laika_repo"]
 	path = laika_repo
 	url = ../../commaai/laika.git
--- a/README.md
+++ b/README.md
@@ -173,8 +173,8 @@ Community Maintained Cars and Features
 | Subaru    | Crosstrek 2018-19             | EyeSight          | Stock            | 0mph               | 0mph         |
 | Subaru    | Forester 2019-20              | EyeSight          | Stock            | 0mph               | 0mph         |
 | Subaru    | Impreza 2017-19               | EyeSight          | Stock            | 0mph               | 0mph         |
-| Tesla | Model S 2014-2016 | AutoPilot HW1 | Stock | 0mph | 0mph |
-| Tesla | Model S 2016-2019 | AutoPilot HW2 | Stock | 0mph | 0mph |
+| Tesla     | Model S 2014-2016             | AutoPilot HW1     | Stock            | 0mph               | 0mph         |
+| Tesla     | Model S 2016-2019             | AutoPilot HW2     | Stock            | 0mph               | 0mph         |
 | Volkswagen| Golf 2015-19                  | Driver Assistance | Stock            | 0mph               | 0mph         |
 
 <sup>1</sup>Requires an [OBD-II car harness](https://comma.ai/shop/products/comma-car-harness) and [community built ASCM harness](https://github.com/commaai/openpilot/wiki/GM#hardware). ***NOTE: disconnecting the ASCM disables Automatic Emergency Braking (AEB).*** <br />
--- a/README.md
+++ b/README.md
@@ -173,6 +173,8 @@ Community Maintained Cars and Features
 | Subaru    | Crosstrek 2018-19             | EyeSight          | Stock            | 0mph               | 0mph         |
 | Subaru    | Forester 2019-20              | EyeSight          | Stock            | 0mph               | 0mph         |
 | Subaru    | Impreza 2017-19               | EyeSight          | Stock            | 0mph               | 0mph         |
+| Tesla | Model S 2014-2016 | AutoPilot HW1 | Stock | 0mph | 0mph |
+| Tesla | Model S 2016-2019 | AutoPilot HW2 | Stock | 0mph | 0mph |
 | Volkswagen| Golf 2015-19                  | Driver Assistance | Stock            | 0mph               | 0mph         |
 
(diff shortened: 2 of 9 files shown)


geohot:
@robbederks Was messaging with @BogGyver. Now that you own this car, let's get it upstreamed!

BogGyver:
what works well:
 - LKA with stock ACC
 - ALC
 - disengage on brake, accel or steering override
 - panda code has already the logic for blocking messages based on safetyParam for future features:
    - Long Control
    - Instrument Cluster
    - Body Controls (keep the turn signal on through ALC for example)
 - coexistence of OpenPilot and AutoPilot (one pull for OP, two pulls for AP) 

what needs work:
 - long control for AP1 works on first engagement but ACC gets disabled after that by AP error DAS_W004... we can rewrite that warning message so that it does not disable ACC for the whole car, TBD
 - long control for AP2 needs to be tested (not sure if the message is sent on CH CAN or PT CAN at the moment) but in documentation it shows on CH CAN; logic is implemented but we need to test it.
 - smoothing of acceleration and jerking
 - once we fix the above, we need limits and safety for long control message
 
what needs to be done in the future:
 - Instrument Cluster integration: 
     - show lanes and objects from OP on IC when OP enabled
     - show status of OP enabled instead of AP (will also trigger native sounds)
 - Body Controls for ALC

robbederks:
Thanks for the PR, with it rewritten on my base the code is already a lot cleaner. Talked with @pd0wm about it, and we came to the following conclusions / remarks:
- We don't want to disengage on steer override, this is inconsistent with what the openpilot experience is on all other supported cars. Ideally, we want a nice user torque blending (which I'm looking into at the moment), and we don;t think it's merge-able without. We have an idea which might be a good first-order approximation for this, which we'll test out soon.
- The HUD integration seems a little unnecessarily complex (we don't have lane curvature in the OP stack anymore for example), and you mentioned that it doesn't really work yet. Let's get rid of this for now, and maybe revisit it in the future with a separate PR
- Long control should also be a separate PR
- We still need some improvements in the controls stack for keeping the blinker on (add as an actuator), as well as the desired speed / accell for long control (also an actuator)

I do think that this is a little too much to merge in at once though, so I propose:
- I'll fix up my PR with the minor improvements you made to the logic (thx for those!), as well as maybe a useable version of user blending, and get it merged in to master
- Once that's done, let's make new separate PRs for the following features:
1. The additions to actuator for long control and blinker hold
2. Radar parsing
3. Long control
4. Possibly HUD integration
- We'll also start working on getting the harnesses for sale

README.md
```diff
@@ -173,6 +173,8 @@ Community Maintained Cars and Features
 | Subaru    | Crosstrek 2018-19             | EyeSight          | Stock            | 0mph               | 0mph         |
 | Subaru    | Forester 2019-20              | EyeSight          | Stock            | 0mph               | 0mph         |
 | Subaru    | Impreza 2017-19               | EyeSight          | Stock            | 0mph               | 0mph         |
+| Tesla | Model S 2014-2016 | AutoPilot HW1 | Stock | 0mph | 0mph |
```
geohot, line 176: Spacing here

.gitmodules
```diff
@@ -1,9 +1,9 @@
 [submodule "panda"]
 	path = panda
-	url = ../../commaai/panda.git
```
geohot, line 3: Can't merge this obviously, are the needed PRs open on panda and opendbc?

README.md
BogGyver: @geohot @robbederks  So tesla has a VERY low threshold for CAN errors... it takes two mismatches for CAN counter to disable steering and ACC cababilities. 

If you look in https://github.com/boggyver/panda/tree/tesla_ap2_fwd you will see a little different approach to block and regenerate. 

I created a set of functions in https://github.com/BogGyver/panda/blob/tesla_ap2_fwd/board/safety/safety_forwards.h (ignore the location for now) that still allows us to generate the message on the Comma2 but it is checked, packed with the official counter and then sent via the fwd hook. All the messages that need to follow this protocol are declared and checked. There are safety checks for timers. If you want to send the message as is just set the counter to 0, otherwise set counter to 1. 

I can explain more if needed but with this approach the counter/frequency is preserved and the message is still generated in the Comma2 side (except of course counter and checkum). The same set of checks can be done in the TX side before the data is accepted. This preserves your logic and existing code (except computing the checksum and counter) on all existing functions to generate messages. 

This safety file implements that method as an example (and i have no more EPAS issues due to counters). https://github.com/BogGyver/panda/blob/tesla_ap2_fwd/board/safety/safety_tesla.h

.gitmodules
BogGyver: yes they are. but also see above for more complex needs

BogGyver:
@geohot @robbederks  So tesla has a VERY low threshold for CAN errors... it takes two mismatches for CAN counter to disable steering and ACC cababilities. 

If you look in https://github.com/boggyver/panda/tree/tesla_ap2_fwd you will see a little different approach to block and regenerate. 

I created a set of functions in https://github.com/BogGyver/panda/blob/tesla_ap2_fwd/board/safety/safety_forwards.h (ignore the location for now) that still allows us to generate the message on the Comma2 but it is checked, packed with the official counter and then sent via the fwd hook. All the messages that need to follow this protocol are declared and checked. There are safety checks for timers. If you want to send the message as is just set the counter to 0, otherwise set counter to 1. 

I can explain more if needed but with this approach the counter/frequency is preserved and the message is still generated in the Comma2 side (except of course counter and checkum). The same set of checks can be done in the TX side before the data is accepted. This preserves your logic and existing code (except computing the checksum and counter) on all existing functions to generate messages. 

This safety file implements that method as an example (and i have no more EPAS issues due to counters). https://github.com/BogGyver/panda/blob/tesla_ap2_fwd/board/safety/safety_tesla.h

SippieCup:
I think there might be a user safety issue with the car displaying two different paths, especially when the "passive" path is the one displayed directly in front of the user. 

The path provided on the IC for AP1 can differ greatly from OP's own predictions and can potentially mislead users.

I am also worried about the potential of frame drops causing desyncs if we are blocking the DAS message with steering and just replaying the last message. But it looks like @BogGyver's safety code may handle that.

BogGyver:
WHATEVER

PR #20446, opened 2021-03-23
APK purge

adeebshihadeh:


selfdrive/manager/manager.py
```diff
@@ -59,9 +57,6 @@ def manager_init(spinner=None):
   if params.get("Passive") is None:
     raise Exception("Passive must be set to continue")
 
-  if EON:
-    update_apks()
-
   os.umask(0)  # Make sure we can create files with 777 permissions
```
pd0wm, line 60: I _think_ we can also get rid of the umask (and all other 777 permissions) since everything runs under the same user now.

adeebshihadeh: will do another pass afterwards with the maybes

PR #20447, opened 2021-03-23
refresh OffroadAlert only when alert is changed

deanlee:


pd0wm:
What's the benefit of this? Redrawing the widget shouldn't be that expensive, right?

deanlee:
Personally, Ithink the added code is worth it.destroy and create widgets,recalc layouts&show, load release text from disk and set it in the Qlabel, all these operations are much more expensive compared to one comparison.
there is not much complexity added in this PR. If we reduce the refresh interval in the future, we can still benefit from it.

adeebshihadeh:
We don't really need to destroy the layout. Let's just write the OffroadAlerts widget properly, and I don't think we'll need to add this hack.

deanlee:
yes,that will be better

adeebshihadeh:
Cool, I think the re-write will be a nice red diff as opposed to this green diff :)

PR #20451, opened 2021-03-24
convience script for starting manager

gregjhogan:
you used to be able to just start `manager.py` and it would build and start it, now that they are separated into `build.py` and `manager.py` it is more convenient to have a single script to run when killing/starting manager inside the existing tmux session


gregjhogan:
`launch_openpilot.sh` is the convenient, single script you can run in tmux after killing manager

PR #20455, opened 2021-03-24
Qt: show current calibration values

deanlee:


adeebshihadeh:
Let's just do this now: https://github.com/commaai/openpilot/blob/master/selfdrive/locationd/calibrationd.py#L83. It should simplify the code in calibrationd and UI.

pd0wm:
I don't see the UI reading the param anywhere when we're offroad

deanlee:
Now read calibration from `CalibrationParams` at startup

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -93,8 +93,23 @@ DevicePanel::DevicePanel(QWidget* parent) : QWidget(parent) {
                                       GLWindow::ui_state.scene.driver_view = true; }
                                     ));
 
-  offroad_btns.append(new ButtonControl("Reset Calibration", "RESET",
-                                   "openpilot requires the device to be mounted within 4° left or right and within 5° up or down. openpilot is continuously calibrating, resetting is rarely required.", [=]() {
+  QString resetCalibDesc = "openpilot requires the device to be mounted within 4° left or right and within 5° up or down. openpilot is continuously calibrating, resetting is rarely required.";
+  if (std::string params = Params().get("CalibrationParams"); !params.empty()) {
```
adeebshihadeh, line 97: move reading the param outside of the if statement

adeebshihadeh:
Looks good, but it doesn't update now.

PR #20460, opened 2021-03-24
update_model: don't update lane_lines&road_edges if end_to_end is true

deanlee:
related to https://github.com/commaai/openpilot/commit/7ee79c99236f749d02a6779b006ee1ab187e0612


adeebshihadeh:
I don't think these kinds of micro-optimizations are necessary.

PR #20467, opened 2021-03-25
Qt input: pass QString by reference

deanlee:
1. ~use simple static QString array for keyboard layouts~
2. pass QString by reference


adeebshihadeh:
What's the point of this? I think the passing by const reference is good, but the other stuff just makes it less readable.

PR #20061, opened 2021-02-11
Merge pandad into boardd

ghost:
fixes #2102

PR #20470, opened 2021-03-25
Merge pandad and boardd

pd0wm:
Continuation of https://github.com/commaai/openpilot/pull/20061


pd0wm:
Going to do this in smaller PRs. Will write USB abstraction class first.

PR #20472, opened 2021-03-25
Params: new class FileLock

deanlee:


selfdrive/common/params.cc
```diff
@@ -78,6 +78,20 @@ static int ensure_dir_exists(std::string path) {
   return mkdir_p(path.c_str());
 }
 
+class FileLock {
+ public:
+  FileLock(const std::string& file_name) : fn_(file_name) {}
+  int try_lock(int operation) {
+    fd_ = open(fn_.c_str(), O_CREAT, 0775);
+    if (fd_ < 0) return -1;
```
pd0wm, line 86: set `fd_` to -1 in the error case so we don't try to close it?

selfdrive/common/params.cc
```diff
@@ -75,6 +75,20 @@ static int ensure_dir_exists(std::string path) {
...
+    return flock(fd_, operation);
+  }
+  ~FileLock() {
+    if (fd_ >= 0) close(fd_);
```
deanlee, line 87: Would this be better?

pd0wm:
This API is not consistent with std::mutex. mutex.lock() is blocking, while mutex.try_lock() is not. Maybe just implement https://en.cppreference.com/w/cpp/named_req/BasicLockable so we can also use `std::lock_guard<FileLock>`.

deanlee:
But if open fails, we can only throw an exception

pd0wm:
> But if open fails, we can only throw an exception

True. An exception might not be so bad actually. None of the users of the params class actually check if any of the operations actually suceeded, which is not great. Are there cases where the `open()` call can fail but we can still continue normal operation?

pd0wm:
@adeebshihadeh What do you think?

deanlee:
errno is 2 : No such file or directory
```
***** testing against commit 796df70c9c83a693f81e4a35d97f90dc5378ef06 *****
***** testing route segment 02c45f73a2e5c6e9|2021-01-01--19-08-22--1 *****

terminate called after throwing an instance of 'std::runtime_error'
  what():  Failed to open lock file, errno:2
Aborted (core dumped)
Error: Process completed with exit code 134.
```

deanlee:
how about this one? no exception, just log the error and return

pd0wm:
Now I'm interested to know why the lock fails in CI

deanlee:
failed to add common lib to params_pyx.so:

`dlopen failed: cannot locate symbol "PyExc_SystemError" referenced by "/data/openpilot/common/params_pyx.so"`

deanlee:
Failed to open lock file, lockfle:` /root/.comma/params/.lock`, errno: 2

pd0wm:
Looks like the params folder doesn't exist yet. Do we try to create the lock before ensuring it exists?

deanlee:
the params folder has not been created before call clear_all()
https://github.com/commaai/openpilot/blob/c23ec9f75389c9ad86400e9a9d23ee6bc5b648a2/selfdrive/test/process_replay/process_replay.py#L339

pd0wm:
Ok, let's merge those other two PRs and then get back to this one.

PR #20475, opened 2021-03-25
disable buttons when onroad

iejMac:


selfdrive/ui/qt/widgets/controls.hpp
```diff
@@ -67,16 +67,28 @@ class ButtonControl : public AbstractControl {
...
-      border-radius: 50px;
-      font-size: 35px;
-      font-weight: 500;
-      color: #E4E4E4;
-      background-color: #191919;
-    )");
+  void setEnabled(bool enabled) {
+    btn.setEnabled(enabled);
+    if (enabled == false){
+      btn.setStyleSheet(R"(
```
adeebshihadeh, line 73: these all go in the same style sheet. wrap the default in QPushButton and the disabled in QPushButton:disabled

PR #20484, opened 2021-03-26
VW MQB: Volkswagen Tiguan Mk2

jyoung8607:
**New MQB Model**

This CAR will cover all variants of the second generation Tiguan. 2018-current in North America, 2016-current RoW.

- [x] added to README
- [x] route with openpilot: 2cef8a0b898f331a|2021-03-25--20-13-57


selfdrive/car/volkswagen/interface.py
```diff
@@ -62,6 +62,11 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None):
       ret.mass = 1500 + STD_CARGO_KG
       ret.wheelbase = 2.64
 
+    if candidate in [CAR.TIGUAN_MK2]:
```
adeebshihadeh, line 65:
```suggestion
    if candidate == CAR.TIGUAN_MK2:
```

jyoung8607: Derp. Fixing.

adeebshihadeh:
Congrats on the 100th car port!

PR #20493, opened 2021-03-26
VW MQB: Volkswagen Passat Mk8

jyoung8607:
**Volkswagen Passat Mk8**

Covers all variants of the eighth generation Passat (2014-present), including Passat Alltrack and Passat GTE (PHEV). Notably, this does NOT include the Passat sold in North America (NMS/PQ), though I plan to support it in the future.

- [x] added to README
- [x] route with openpilot: 4d134e099430fba2|2021-03-26--00-26-06

Thanks to community Passat owners Hendrik212 (2016 GTE), and Shonggie (2017 Alltrack).


jyoung8607:
I expected CI to fail on this, because the route hadn't been uploaded to the CI bucket yet, but it succeeded. Can't think of how that would be, unless @adeebshihadeh happened to be watching and uploaded it right after I opened the PR?

pd0wm:
Route is definitely in the bucket.

PR #20499, opened 2021-03-26
Qt api: pass payloads as pointer

deanlee:


pd0wm:
What is the benefit besides skipping a copy?

selfdrive/ui/qt/api.hpp
```diff
@@ -33,16 +32,17 @@ class RequestRepeater : public QObject {
   Q_OBJECT
 
 public:
-  explicit RequestRepeater(QWidget* parent, QString requestURL, int period = 10, const QString &cache_key = "", QVector<QPair<QString, QJsonValue>> payloads = *(new QVector<QPair<QString, QJsonValue>>()), bool disableWithScreen = true);
```
deanlee, line 36: There may be a memory leak here

pd0wm:
The pointer argument becomes a huge foot-gun. If you let the original payloads vector go out of scope while the RequestRepeater is still around you will get unexpected behavior. 

I don't see the payloads arguments used anywhere. Should we just remove it?

PR #20501, opened 2021-03-26
Qt: refactor http request

deanlee:


deanlee:
performance looks good, grab and return context only take about 5~6ms

selfdrive/ui/qt/api.cc
```diff
@@ -72,76 +72,67 @@ QString CommaApi::create_jwt(QVector<QPair<QString, QJsonValue>> payloads, int e
...
+      request.setRawHeader(it.key().toUtf8(), it.value().toUtf8());
+    }
+  }
+#ifdef QCOM
+  QSslConfiguration ssl = QSslConfiguration::defaultConfiguration();
+  ssl.setCaCertificates(QSslCertificate::fromPath("/usr/etc/tls/cert.pem", QSsl::Pem, QRegExp::Wildcard));
+  request.setSslConfiguration(ssl);
+#endif
+
+  QEventLoop loop;
```
adeebshihadeh, line 88: is it safe to create another event loop outside the main one? doesn't this make the requests synchronous now?

deanlee: it's safe,the function is "block", but all events will be dispatched to their usual places during waiting.

PR #20505, opened 2021-03-27
KIA Ceed: increase max torque

ZanZaD13:
**Description** [Increases car handling in certain situations.

**Verification** [test driving tighter corners at lower speeds.


adeebshihadeh:
Do you have a stock route where you've seen values in this increased range or are there any other CEED users who can confirm the extra torque?

adeebshihadeh:
You're the only Ceed user over the past month, so we'll merge this.

PR #20506, opened 2021-03-27
Params: python-like interface

deanlee:


selfdrive/ui/qt/widgets/controls.hpp
```diff
@@ -109,12 +109,12 @@ class ParamControl : public ToggleControl {
...
   ParamControl(const QString &param, const QString &title, const QString &desc, const QString &icon, QWidget *parent = nullptr) : ToggleControl(title, desc, icon, parent) {
     // set initial state from param
-    if (Params().read_db_bool(param.toStdString().c_str())) {
+    if (Params().getBool(param.toStdString().c_str())) {
       toggle.togglePosition();
     }
     QObject::connect(this, &ToggleControl::toggleFlipped, [=](int state) {
       char value = state ? '1' : '0';
-      Params().write_db_value(param.toStdString().c_str(), &value, 1);
+      Params().put(param.toStdString().c_str(), &value, 1);
```
pd0wm, line 117: Maybe we should make a `putBool` helper as well.

selfdrive/common/params.h
```diff
@@ -12,35 +13,40 @@ class Params {
...
+
+  template <class T>
+  std::optional<T> get(const char *key, bool block = false) {
+    std::istringstream iss(get(key, block));
+    T value{};
+    iss >> value;
+    return iss.fail() ? std::nullopt : std::optional(value);
+  }
+
+  inline bool getBool(const char *key) {
```
pd0wm, line 42: Interesting that only `1` becomes true, not any other values. Did you find documentation for that? I couldn't find anything.

Seems that `1\n` also evaluates to true, but that's fine with me.

deanlee: fixed.

deanlee: fixed

pd0wm: Yeah, maybe good to keep old behavior

PR #20510, opened 2021-03-27
2021 Kia Seltos INDI/LQR Tuning

speedking456:
**Description** 
After KALE 🥬 was merged, PID tuning resulted in weird ping-pong/hugging issues no matter the parameters, so it was clear new tuning was required. I tried INDI, it was definitely better, but not as good as copying LQR from the RAV4.

**Verification** 
Multiple 50+ KM routes were driven with almost perfect turing and lane centering. KALE 🥬 definitely requires this new tuning; the previous PID tuning would not make turns correctly and would tend to hug one side or the other in different scenarios.

**Route**
Route: in progress, doing a bit more testing before I'm calling it good


speedking456:
So the best tuning I could get for the Seltos was by copying the RAV4 LQR tuning. Can this be upstreamed? Or is that too ghetto?

speedking456:
Closed because model could be the issue. Doing more testing, possibly wait for the next KL update.

PR #20511, opened 2021-03-28
hyundai - wind down torque on disengage

gregjhogan:
This is supposed to cause gentle wind-down of torque when disengaging

seems to work but need to review in cabana to confirm

### DIFF
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -40,6 +40,7 @@ class CarController():
     self.apply_steer_last = 0
     self.car_fingerprint = CP.carFingerprint
     self.steer_rate_limited = False
+    self.steer_wind_down = False
     self.last_resume_frame = 0
 
   def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert,
@@ -58,6 +59,10 @@ class CarController():
 
     if not lkas_active:
       apply_steer = 0
+      if self.apply_steer_last != 0:
+        self.steer_wind_down = True
+    if lkas_active or CS.out.steeringPressed:
+      self.steer_wind_down = False
 
     self.apply_steer_last = apply_steer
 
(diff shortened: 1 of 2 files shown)


AskAlice:
This is literally the reason I got banned from the openpilot discord. One day I got quite upset as I had almost gotten into an accident while going around a turn on a highway while adjusting to the new safety rules that required disengage on brake for vehicles without ACC/SCC. 

Is there panda code associated with this? I cant imagine this is the only thing necessary.

gregjhogan:
@AskAlice this isn't to make things safer and you may want to lower the max torque allowed for your vehicle.  The safety requirement  which would apply to the situation you described is [here](https://blog.comma.ai/how-to-write-a-car-port-for-openpilot/#background--safety-architecture):
> Steer loss: while in a turn, in the case of an immediate loss of steering torque, the driver shall have at least 1 second of reaction time before the car significantly deviates from the desired path (e.g. lane lines crossing).

The best way to ensure this is to limit the max torque allowed such that even a complete failure of the power steering motor still meets this requirement.  If the power steering motor failed obviously you cannot rely on this functionality.

I found this functionality in the power steering motor controller firmware of my vehicle and I think it makes things more chill.

No panda changes are necessary to make this work because the wind down is completely handled by the power steering motor controller.

AskAlice:
I will test this tomorrow.

AskAlice:
Tested it, this is much better than the existing behavior. MUCH safer. 100% should merge in my book.

gregjhogan:
if your car was unsafe before this change, it is still unsafe after this change because openpilot is QM and therefore you cannot depend on it to take action to keep you safe (as described [here](https://blog.comma.ai/how-to-write-a-car-port-for-openpilot/#background--safety-architecture))

AskAlice:
> if your car was unsafe before this change, it is still unsafe after this change because openpilot is QM and therefore you cannot depend on it to take action to keep you safe (as described [here](https://blog.comma.ai/how-to-write-a-car-port-for-openpilot/#background--safety-architecture))

Definitely, however, it is a great improvement.

TK211X:
This is one of the better changes for a safer OpenPilot in a while. Thank you @gregjhogan

PR #20512, opened 2021-03-28
Update actuator delay values for Subaru Ascent and Impreza

RCoaster:
Car bug fix

Description: Actuator delay too high for Subaru Ascent and Impreza, causing ping ponging due to new MPC in 0.8.2 and 0.8.3. Based on discussions in Discord, lowering the delay by .2 solved the ping ponging issue. 

Fix was tested on a 2019 Subaru Ascent. I did not have an Impreza to test, so the value was based on the original value and lowering it by .2 

Route: Two routes at different times: 

b75e98e3b7cf69dd|2021-03-27--20-28-16

b75e98e3b7cf69dd|2021-03-28--04-48-47

### DIFF
--- a/selfdrive/car/subaru/interface.py
+++ b/selfdrive/car/subaru/interface.py
@@ -36,7 +36,7 @@ class CarInterface(CarInterfaceBase):
       ret.wheelbase = 2.89
       ret.centerToFront = ret.wheelbase * 0.5
       ret.steerRatio = 13.5
-      ret.steerActuatorDelay = 0.3   # end-to-end angle controller
+      ret.steerActuatorDelay = 0.105  # end-to-end angle controller
       ret.lateralTuning.pid.kf = 0.00003
       ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0., 20.], [0., 20.]]
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.0025, 0.1], [0.00025, 0.01]]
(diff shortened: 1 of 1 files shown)


RCoaster:
Update: Got 2 routes uploaded and preserved. They are: 

b75e98e3b7cf69dd|2021-03-27--20-28-16

b75e98e3b7cf69dd|2021-03-28--04-48-47

HaraldSchafer:
subaru

We determined the values by doing batch tests on many cars and many segments. Don't really want to merge this since it contradicts the tests. We are working on have a better more general way of dealing with tuning.

RCoaster:
Were those tests done on 8.2 and above? There is a known ping pong issue for Subaru's. The actuator delay helped solve the ping pong.

HaraldSchafer:
These tests don't depend on anything that was changed in 0.8.3 or 0.8.2, they just check the response time from commanded steering angle to achieved yaw rate.

ClockeNessMnstr:
Trying to figure out if actuatordelay should be .2 for crosstrek/impreza in discord. Or rather trying to see if anyone can try it with poor results. 
for me .2 is very close to correct using just FF control turned up.

ClockeNessMnstr:
I'm getting the 2019 Subaru Forester delay of .3 in my logs between the LKAS request and the torque output. 

However I'm getting a ridiculously short delay for the crosstrek in the same messages. about .05 - .1
is there a good way to collect this data for a single car? logging the calculation time and imu data?

I think .2 is hardcoded somewhere right? I lost track of where that was if I saw it digging around.

HaraldSchafer:
I just logged the requested steering angle and measured steering angle and calculated the lag but taking the argmax of the convolution of those 2 signals.

ClockeNessMnstr:
https://cdn.discordapp.com/attachments/525718620517564446/852224627710754826/PXL_20210609_044301938.MP.jpg

ClockeNessMnstr:
PXL_20210609_044301938 MP

not the best shot as it's missing the timesteps / resolution to show the time, but I'm getting a .4 second delay on a 2018 crosstrek which is the same as the Impreza. 

This would be a ,2 second delay in the tuning as there is a hardcoded offset of .2 for whatever reason.

PR #20514, opened 2021-03-28
making no disengage on gas easier for Forks

ffhspa:
this PR should make it easier for Forks to implement no Disengage on Gas.


adeebshihadeh:
We added this functionality to panda so forks could run stock panda safety, but we won't merge code that is only intended for forks.

PR #20516, opened 2021-03-28
Qt: update ui in a separate thread

deanlee:
based on QT example: threadedqopenglwidget https://code.woboq.org/qt5/qtbase/examples/opengl/threadedqopenglwidget/
related document: https://doc.qt.io/qt-5/qopenglwidget.html#threading
QOpenGLWidget source code: https://code.woboq.org/qt5/qtbase/src/widgets/kernel/qopenglwidget.cpp.html
```

glwindow thread:   | swap |           | swap |             | swap |
UIUpdate thread:            | update |         |  update |  
```

### DIFF
--- a/selfdrive/thermald/thermald.py
+++ b/selfdrive/thermald/thermald.py
@@ -349,7 +349,7 @@ def thermald_thread():
     set_offroad_alert_if_changed("Offroad_TemperatureTooHigh", (not startup_conditions["device_temp_good"]))
 
     # Handle offroad/onroad transition
-    should_start = all(startup_conditions.values())
+    should_start = True#all(startup_conditions.values())
     if should_start != should_start_prev or (count == 0):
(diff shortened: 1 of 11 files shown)


deanlee:
performance looks good, "grab + return context + invoke update" only takes about 5~6ms

adeebshihadeh:
How's the CPU usage? Currently, it's around 15% onroad.

deanlee:
Reduce cpu usage by 3~4% and do not block the UI main thread. I think it's worth continuing to optimize this PR

adeebshihadeh:
Can we split out the class to wrap our current event loop? It should have similar behavior where we block on visionipc with a timeout and update sockets, then it should emit a signal with an update. Then this PR will be for scheduling the GL stuff once the other PR is merged.

deanlee:
@adeebshihadeh  how about this one?

PR #20518, opened 2021-03-28
Kia Niro EV 2021 FW versions

jonjander:
values for Kia Niro EV 2021


adeebshihadeh:
What's your dongle ID?

jonjander:
My ID is c06f375e2a2a220d

PR #20524, opened 2021-03-29
scrolling improvements

adeebshihadeh:
continuation of #20458


selfdrive/ui/qt/widgets/controls.cc
```diff
@@ -47,7 +46,6 @@ AbstractControl::AbstractControl(const QString &title, const QString &desc, cons
     description->setWordWrap(true);
     description->setVisible(false);
     vlayout->addWidget(description);
-    vlayout->addStretch();
```
deanlee: Without this stretch, the title will adjust the position and flash during folding

adeebshihadeh: I don't think this is the real solution, it breaks the centering

PR #20535, opened 2021-03-30
Add device_type to class Hardware

deanlee:
use this variable to simplify hardware-related code


selfdrive/hardware/base.h
```diff
@@ -4,11 +4,18 @@
 #include <fstream>
 
 
+enum HardwareDeviceType {
```
pd0wm, line 7: We already have a definition for device type: https://github.com/commaai/cereal/blob/master/log.capnp#L44

PR #20536, opened 2021-03-30
cloudlog on RTC read/write

pd0wm:

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -161,18 +161,18 @@ bool usb_connect() {
 #endif
 
   if (tmp_panda->has_rtc){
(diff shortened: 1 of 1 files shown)


selfdrive/boardd/boardd.cc
```diff
@@ -332,6 +334,9 @@ void panda_state_thread(bool spoofing_started) {
       struct tm sys_time = get_time();
       if (time_valid(sys_time)){
         panda->set_rtc(sys_time);
+        LOGW("Updating panda RTC: %d-%02d-%02d %02d:%02d:%02d",
```
gregjhogan, line 337: maybe only log if time is off by some threshold like 10 seconds?

PR #20541, opened 2021-03-30
ui closing dialogs on timeout

iejMac:


adeebshihadeh:
Let's make sure this works for both the confirmation dialog and the keyboard dialog as well as android activities.

selfdrive/ui/qt/window.cc
```diff
@@ -28,7 +28,11 @@ MainWindow::MainWindow(QWidget *parent) : QWidget(parent) {
       if(widget->metaObject()->className() != QString("MainWindow")) {
         widget->close();
       }
+
     }
+#ifdef QCOM
+	hardware.eon.close_activites();
```
adeebshihadeh:
```suggestion
	HardwareEon::close_activites();
```

PR #20542, opened 2021-03-30
add fpv2 values from 2015 audi a3 tdi

keeleysam:
this should remove the need for fpv1 on the a3


adeebshihadeh:
Can you add the A3 to the IGNORED_FINGERPRINTS list?

PR #20551, opened 2021-04-01
adds a note about serial steering cars working wOP

reddn:
Add a small note under Community Maintained cars saying OP will work with serial steering cars with additional hardware and minor code modifications


adeebshihadeh:
We'll add a page to the wiki for cars that are well-supported in forks and just link to that in the README. @VirtuallyChris will get the wiki page up sometime tomorrow, and you can add this info there.

PR #20553, opened 2021-04-01
VW: improve car's lane assist indication

Ambroos:
This makes a few small improvements to the car's lane assist status display for VW, Audi and other VW group brands.

The values display this, VW car display data from @jyoung8607:
- 0: nothing
- 1: grey lane line (used to show lane assist is active but not detecting lines)
- 2: green lane line, without arrow in VW (used to show lane assist is active and detecting lines, or active centering in Audi cars)
- 3: red lane line in Audi, green lane line with arrow in VW (used to show lane departure in Audi/VW, or active centering in VW)

To map openpilot as close as possible to the stock behaviour, after this diff the following is done:

- In any case, always show state 3 for lane departure
- With laneless: show state 2 when engaged, state 0 when disengaged
- With lanes (lane-specific states): show state 1 when engaged and no lane lines are visible, or forced when at a standstill (to avoid visual jitter), state 2 when engaged and lane lines detected, state 0 when disengaged

Test drives with my Audi A3:
892d124b54261a3a|2021-03-31--19-02-51 with lanes
892d124b54261a3a|2021-03-31--19-14-31 laneless

### DIFF
--- a/selfdrive/car/volkswagen/carcontroller.py
+++ b/selfdrive/car/volkswagen/carcontroller.py
@@ -1,4 +1,5 @@
 from cereal import car
+from common.params import Params
 from selfdrive.car import apply_std_steer_torque_limits
 from selfdrive.car.volkswagen import volkswagencan
 from selfdrive.car.volkswagen.values import DBC, CANBUS, MQB_LDW_MESSAGES, BUTTON_STATES, CarControllerParams
@@ -18,9 +19,11 @@ class CarController():
     self.graMsgStartFramePrev = 0
     self.graMsgBusCounterPrev = 0
 
+    self.use_lanelines = Params().get('EndToEndToggle') != b'1'
+
     self.steer_rate_limited = False
 
-  def update(self, enabled, CS, frame, actuators, visual_alert, left_lane_visible, right_lane_visible):
+  def update(self, enabled, CS, frame, actuators, visual_alert, left_lane_visible, right_lane_visible, left_lane_depart, right_lane_depart):
     """ Controls thread """
 
     P = CarControllerParams
@@ -110,18 +113,20 @@ class CarController():
     # filters LDW_02 from the factory camera and OP emits LDW_02 at 10Hz.
 
     if frame % P.LDW_STEP == 0:
-      hcaEnabled = True if enabled and not CS.out.standstill else False
-
       if visual_alert == car.CarControl.HUDControl.VisualAlert.steerRequired:
         hud_alert = MQB_LDW_MESSAGES["laneAssistTakeOverSilent"]
       else:
         hud_alert = MQB_LDW_MESSAGES["none"]
 
-      can_sends.append(volkswagencan.create_mqb_hud_control(self.packer_pt, CANBUS.pt, hcaEnabled,
+      left_lane_visible = (left_lane_visible and not CS.out.standstill) if self.use_lanelines else True
+      right_lane_visible = (right_lane_visible and not CS.out.standstill) if self.use_lanelines else True
+
+      can_sends.append(volkswagencan.create_mqb_hud_control(self.packer_pt, CANBUS.pt, enabled,
                                                             CS.out.steeringPressed, hud_alert, left_lane_visible,
                                                             right_lane_visible, CS.ldw_lane_warning_left,
                                                             CS.ldw_lane_warning_right, CS.ldw_side_dlc_tlc,
-                                                            CS.ldw_dlc, CS.ldw_tlc))
+                                                            CS.ldw_dlc, CS.ldw_tlc,
+                                                            left_lane_depart, right_lane_depart))
 
     #--------------------------------------------------------------------------
     #                                                                         #
--- a/selfdrive/car/volkswagen/carcontroller.py
+++ b/selfdrive/car/volkswagen/carcontroller.py
@@ -1,5 +1,4 @@
 from cereal import car
-from common.params import Params
 from selfdrive.car import apply_std_steer_torque_limits
 from selfdrive.car.volkswagen import volkswagencan
 from selfdrive.car.volkswagen.values import DBC, CANBUS, MQB_LDW_MESSAGES, BUTTON_STATES, CarControllerParams
@@ -18,8 +17,6 @@ class CarController():
     self.graMsgSentCount = 0
     self.graMsgStartFramePrev = 0
     self.graMsgBusCounterPrev = 0
-
-    self.use_lanelines = Params().get('EndToEndToggle') != b'1'
 
     self.steer_rate_limited = False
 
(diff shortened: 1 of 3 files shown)


jyoung8607:
I've been discussing this with @Ambroos and I'm good with this change in-concept, but I haven't tested it on either of my vehicles yet and won't be able to until the middle of next week.

There are some nuances in what the factory camera sends depending on whether it's coded for VW or Audi branding, and what we send today is visually correct for VW but is visually incorrect on the Audi virtual cockpit. This change makes it correct-enough for VW (and arguably better, given openpilot's different operation concept from the factory LKAS) and is now correct for Audi. We contemplated switching behaviors based on brand but there's no real reason to.

selfdrive/car/volkswagen/carcontroller.py
```diff
@@ -18,9 +19,11 @@ def __init__(self, dbc_name, CP, VM):
     self.graMsgStartFramePrev = 0
     self.graMsgBusCounterPrev = 0
 
+    self.use_lanelines = Params().get('EndToEndToggle') != b'1'
```
adeebshihadeh, line 22: this is a temporary change in openpilot UI's to make it clear what mode you're in. it doesn't make sense to carry this over to the car interface

Ambroos: @adeebshihadeh is there another way we can know if we should expect lanelines to be present? There's a difference in car HUD expectations.

If we shouldn't count on this param, I think we should drop the 'lane line visible' signal at all times, and just always show a green indicator when Openpilot is active, and grey when it isn't.

adeebshihadeh: let's just forget the laneless case and always go with the "With lanes (lane-specific states)" case

Ambroos:
OK, I've made some adjustments and I think this is the best available approach without depending on the laneless param.

- Engaged and lane visible: state 3 (red/centering)
- Engaged and lane not visible or laneless: state 2 (green)
- Disengaged (but available): state 1 (grey)

State 0 never happens in this scenario, which I think is fine. This model is closest to what the indicators do with VW/Audi's lane assist systems, where it's 1/grey for "ready but won't intervene", 2/green for "active and will intervene on lane departure/actively centering", 3/red for "lane departure/actively centering".

I'd love to test, but my car is in for repairs for a few days. I'll report back once I've given this a test drive.

adeebshihadeh:
Why state 1 for disengaged? Maybe I'm misunderstanding, but seems like it should be:

- Disengaged: state 0
- Lane departure warning: state 3
- Engaged and lane not visible: state 1
- Engaged and lane visible: state 3

Ambroos:
State 0 would be equivalent to having openpilot disabled/unplugged (vs disengaged). The UI shifts around to make room for lane lines when going from state 0 to any other state, which is visually distracting and should be avoided while driving. In my A3 the set speed moves over, for example, which just doesn't look right.

State 1 is a subtle idle state, while in state 1 no VW/Audi car will ever do any steering input. I think that's a good way to map it to openpilot states. ACC uses the same grey state when ACC is ready to be engaged.

State 2 is "lane assist understands what it's seeing and is actively steering" (in both VW and Audi with active centering enabled), which maps very well to openpilot's engaged state.

State 3 should probably only be used for LDW, and I'm not sure about the value of it at for anything else. Openpilot does great without lane lines. Seeing things change in the instrument cluster for something as trivial as lane line visibility seems unnecessary. It also looks wrong in Audi cars, where state 3 is red, and Audi only uses red icons for warnings and things that need attention. I'm only leaving it in because it's the current state, and it'll become less and less relevant/used automatically over time thanks to laneless (unless the indicators/UI are being brought back to laneless).

I'm pretty confident in only using 1/2 in normal operation, and 3 for LDW. It's the least distracting and most accurate mapping to the stock states.

jyoung8607:
> Why state 1 for disengaged? Maybe I'm misunderstanding, but seems like it should be:
> 
> * Disengaged: state 0
> * Lane departure warning: state 3
> * Engaged and lane not visible: state 1
> * Engaged and lane visible: state 3

I'm assuming you meant state 2 for engaged and lane visible. With that proviso, that seems reasonable.

I honestly don't remember why I didn't go to 0 when disengaged, back when I originally wrote this. I think openpilot native lane departure warning may have been new, and I wanted to show in the cluster whether we were tracking lanes, because openpilot's LDW still worked when disengaged. It was almost emulating the "late intervention" vs "early intervention" strategies for stock LKAS. For VW, I think going to 0 when disengaged is acceptable. I leave it to @Ambroos if that's too terribly obnoxious in the Audi virtual cockpit when engaging and disengaging. It isn't bothersome in the VW version of the virtual cockpit.

@Ambroos, I feel that leaving the lane-line indicators static while engaged, regardless of recognized lane lines, would violate [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment). If openpilot sees lane lines, there's no reason we shouldn't propagate those to the cluster, just like the car would, and just like the driver expects. Remember that openpilot "laneless" operation is kind of a misnomer, the model's path does what a human would do and humans look at lane-lines, "laneless" just means we don't try to forcibly snap/override the model's path to something between those lines as if we were bolting training-wheels onto a bike. That was only ever done because the model's path output used to be not-so-hot, and that time is coming to an end.

Ambroos:
@jyoung8607, I think my interpretation of POLA is a little different. For me state 2 means that LKAS is active, working well and steering, and state 3 means LKAS is correcting/detecting lane departure. VW group cars don't really do much besides simple lane departure warnings if they only detect one lane line, so a partial 2 state isn't very relevant to openpilot.

Whether openpilot can detect lane lines or not is just not relevant information in my opinion, because openpilot will do a good job steering anyway. The visual for lane lines in the openpilot UI is very subtle because it's not the most useful information.

0 is off, 1 is grey, 2 is green and 3 is red (the Audi setup). The ACC icon treats these as off / ready / active / distance warning. My mental model of the LKAS HUD would be the same. Black is completely off (no Comma 2 connected), gray is ready (disengaged), green is engaged, red is LDW. Using either gray or red for engaged states, mismatched from what those colours mean in the ACC status indicator feels really wrong.

If there was a better to propagate lane line status to the VW instrument cluster I'd be all for it, but with the limitations we have I think sticking closest to the stock LKAS behaviour (when it comes to when it's steering and when it isn't), which also matches the ACC behaviour, is by far the simplest and most POLA-compliant way to do it.

jyoung8607:
To be clear, as we discussed in Discord, I'm perfectly fine with your proposal for engaged to become state 2, and to reserve state 3 for lane departure warning. Combined with the behavior in the first version of this PR, perhaps I read more into this than you meant to say:

> ... and I'm not sure about the value of it to indicate lane line visibility. Openpilot does great without lane lines. Seeing things change in the instrument cluster for something as trivial as lane line visibility seems unnecessary.

POLA here means when the driver sees see lane-line recognition info in the cluster in stock mode, they should keep seeing it in openpilot mode. I was not fine with was the talk of discarding lane state entirely and/or locking it to a fixed value while engaged, just because openpilot is going toward "laneless". The two concepts aren't directly related.

I think we are in raging agreement here, perhaps there is some miscommunication between different versions of this code. The code in this PR currently still uses state 3 for engaged.

adeebshihadeh:
Closing as @jyoung8607 intends to open a new PR to succeed this one.

PR #20555, opened 2021-04-01
refactor webcam, use common run_camera function

deanlee:


selfdrive/camerad/cameras/camera_webcam.cc
```diff
@@ -64,24 +69,44 @@ void camera_init(VisionIpcServer * v, CameraState *s, int camera_id, unsigned in
   s->buf.init(device_id, ctx, s, v, FRAME_BUF_COUNT, rgb_type, yuv_type);
 }
 
-static void* road_camera_thread(void *arg) {
-  int err;
+void run_camera(CameraState *s, cv::VideoCapture &video_cap, float *ts) {
+  if (!video_cap.isOpened()) {
+    assert(0);
+  }
```
adeebshihadeh, line 75:
```suggestion
  assert(video_cap.isOpened());
```

selfdrive/camerad/cameras/camera_webcam.cc
```diff
@@ -19,13 +19,17 @@
 #pragma clang diagnostic pop
 
 
-extern ExitHandler do_exit;
+// id of the video capturing device
+#define ROAD_CAMERA_ID 1
```
adeebshihadeh, line 23: should we use environment variables here then fallback to 1 and 2 if they're not set?

ZwX1616:
lgtm.

PR #20560, opened 2021-04-01
Kia Seltos 2021 INDI Tuning + Increased Steer Limit

speedking456:
**Description**
KALE was not performing with PID tuning, I've improved its performance with INDI tuning.

**Verification**
Drives a lot smoother and doesn't fight the lane planner.

**Route**
c75a59efa0ecd502|2021-04-01--15-20-16


speedking456:
Ready to merge

adeebshihadeh:
Our policy for tuning PRs is that another user has to confirm that the new tuning is better. In this case, another user would have to confirm both changes.

speedking456:
> Our policy for tuning PRs is that another user has to confirm that the new tuning is better. In this case, another user would have to confirm both changes.

As far as I know, there is only one other Kia Seltos running OP. In such a case, would I have to get them to test it as well to have this merged?

adeebshihadeh:
Yes, usually we require 2-3 people to confirm, but seems there's only two Seltos users.

speedking456:
> Yes, usually we require 2-3 people to confirm, but seems there's only two Seltos users.

I did the original port, does that make a difference? I would have included more optimized tuning in my original pull request, but it was useable enough when I created it. I mostly wanted to get it added for those who have my vehicle as well.

adeebshihadeh:
Let's ask in the HKG channel in discord for feedback. If you get no response in a couple days, we'll just merge.

adeebshihadeh:
Any update on this? Did you ask in the HKG channel and get any feedback?

speedking456:
> Any update on this? Did you ask in the HKG channel and get any feedback?

I'm waiting for the other Seltos user to get back to me. They've said they'll try the fork but no feedback yet.

adeebshihadeh:
This good to merge now? If they haven't responded by now, I'm fine with merging.

speedking456:
> This good to merge now? If they haven't responded by now, I'm fine with merging.

Good to merge

selfdrive/car/hyundai/interface.py
```diff
@@ -144,13 +144,20 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
...
-      ret.wheelbase = 2.6
-      ret.steerRatio = 13.73  # Spec
-      tire_stiffness_factor = 0.5
-      ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0.], [0.]]
-      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.25], [0.05]]
+      ret.mass = 1337. + STD_CARGO_KG
+      ret.wheelbase = 2.63
+      ret.steerRatio = 14.56
+      tire_stiffness_factor = 1
+      ret.steerActuatorDelay = 0.096
```
adeebshihadeh, line 151: revert this

selfdrive/car/hyundai/interface.py
```diff
@@ -144,13 +144,20 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
...
-      ret.mass = 1310. + STD_CARGO_KG
-      ret.wheelbase = 2.6
-      ret.steerRatio = 13.73  # Spec
-      tire_stiffness_factor = 0.5
-      ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0.], [0.]]
-      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.25], [0.05]]
+      ret.mass = 1337. + STD_CARGO_KG
+      ret.wheelbase = 2.63
+      ret.steerRatio = 14.56
+      tire_stiffness_factor = 1
```
adeebshihadeh, line 150: delete this, it's already the default

selfdrive/car/hyundai/interface.py
```diff
@@ -144,13 +144,15 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0.], [0.]]
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.25], [0.05]]
     elif candidate == CAR.KIA_SELTOS:
-      ret.lateralTuning.pid.kf = 0.00005
-      ret.mass = 1310. + STD_CARGO_KG
```
adeebshihadeh, line 148: why are all these other lines deleted? only remove the ones for the PID tuning

speedking456: The PID lines were already removed, didn't get what you meant by
> delete this, it's already the default
I've reverted the changes, it should have already been correct I would think.

speedking456 (commented):
I changed these params as my original pull request was not entirely accurate. The updated params are more accurate.

adeebshihadeh:
Ah, this PR should have been split up in that case, but we can merge it in one this time. Can you update the params? Then this should be good to merge.

speedking456:
> Ah, this PR should have been split up in that case, but we can merge it in one this time. Can you update the params? Then this should be good to merge.

Un-reverted, sorry about the confusion, didn't realize it should be two PRs. I'll keep that in mind next time.

Ready to merge whenever you are.

adeebshihadeh:
One last thing: revert the change to `steerActuatorDelay`. It's an unnoticeable change and those values should be determined from lots of data.

speedking456:
> One last thing: revert the change to `steerActuatorDelay`. It's an unnoticeable change and those values should be determined from lots of data.

Reverted. 

I originally made that change thinking it improved performance on two car wide roads with only a center line, however I think it was the model improvements that fixed it.

PR #20561, opened 2021-04-01
Added disengage alert in modeld

mitchellgoffpc:
Added logic to check for spikes in disengagement probability and trigger a corresponding alert.


HaraldSchafer:
Looks much more readable without the ring buffer!

selfdrive/modeld/models/driving.cc
```diff
@@ -46,8 +46,29 @@ constexpr int OUTPUT_SIZE =  POSE_IDX + POSE_SIZE;
...
+uint64_t last_steering_frame = 0;
+uint64_t history_size = 0;
+
 // #define DUMP_YUV
 
+float smooth(float p, float a, float b) {
+  return p * a + (1 - p) * b;
+}
+
+float threshold_mse(float x, float threshold, bool gt) {
```
HaraldSchafer, line 67: What does gt mean?

mitchellgoffpc: greater than / less than, it's a flag to control whether the error should be taken above or below the threshold. I can split this up into two separate functions to make it more readable, or maybe just inline it.

HaraldSchafer: Yeah two functions sounds more readable

mitchellgoffpc:
Closing this PR since the new model outputs give a much easier way to detect collision risk

PR #20563, opened 2021-04-02
update openpilot simulation device option

nninept:
Description : There was an error for connect joystick to bridge.py because of ```--device``` option configuration.

*** Optional report
Although this one resolved device error, but I found another bug that ```manual_ctrl.py``` couldn't import ``` evdev ``` library. It could be resolved by manually install ``` evdev ``` with pip in container, but  when stop docker it flash.
I think ``` evdev ``` should be installed when build openpilot docker but I can't find how to do it.  Is there any idea?


adeebshihadeh:
Any reason this is a draft?

PR #20572, opened 2021-04-03
below lane change speed alert

### LINKED ISSUE #20565: Make clear when too slow for lane changes
If you command a lane change under 45mph, nothing will happen. It it should be clear to the user that no lane change can be done because the speed is too low.

Perhaps instead of the message "Steer Left to Start Lane Change" that appears above 45mph something like **"Speed too low for lane change"**

Winvoker:
related to #20565
Fixes #20565

Unit test failed because of cereal. I made PR https://github.com/commaai/cereal/pull/129

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -206,6 +206,9 @@ class Controls:
                                                  LaneChangeState.laneChangeFinishing]:
       self.events.add(EventName.laneChange)
 
+    if self.sm['lateralPlan'].laneChangeState == LaneChangeState.belowSpeed:
+      self.events.add(EventName.belowSteerSpeed)
+
     if self.can_rcv_error or (not CS.canValid and self.sm.frame > 5 / DT_CTRL):
       self.events.add(EventName.canError)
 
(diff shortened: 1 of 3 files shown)


Winvoker:
fixed attr error

selfdrive/controls/controlsd.py
```diff
@@ -206,6 +206,9 @@ def update_events(self, CS):
                                                  LaneChangeState.laneChangeFinishing]:
       self.events.add(EventName.laneChange)
 
+    if self.sm['lateralPlan'].laneChangeState == LaneChangeState.belowSpeed:
+      self.events.add(EventName.belowSteerSpeed)
```
cydia2020, line 210: isn't belowsteerspeed for vehicles that can't steer below a certain speed? See "No ALC Below" in the readme.

Winvoker: Oh yes , thanks :) In that case I need to create a new event.

adeebshihadeh:
#20565 was closed since we lowered the min lane change speed to 30mph

PR #20576, opened 2021-04-04
lane change ready icon + code

cydia2020:
#20565

an icon would display when assisted lane change conditions are met.

see https://github.com/commaai/openpilot/issues/20565#issuecomment-812994242 for demo

[icon credit](https://smashicons.com/)


deanlee:
Personally, I think these icons can be distracting and confusing. The simpler the interface, the better.

cydia2020:
> Personally, I think these icons can be distracting and confusing. The simpler the interface, the better.

I guess so, however an alert that popping up everytime the users uses the indicator isn't ideal either. A light is still better than a text from a human factor standpoint once the human knows what to look for.

selfdrive/loggerd/uploader.py
```diff
@@ -217,7 +217,7 @@ def uploader_fn(exit_event):
     on_wifi = network_type == NetworkType.wifi
     allow_raw_upload = params.get("IsUploadRawEnabled") != b"0"
 
-    d = None
+    d = uploader.next_file_to_upload(with_raw=allow_raw_upload and on_wifi and offroad)
```
cydia2020: This was pushed to the wrong branch and I do apologise

adeebshihadeh:
Thanks for the PR, but this icon looks out of place with the rest of openpilot. I also don't think adding more icons to the UI is the right solution here.

cydia2020:
> Thanks for the PR, but this icon looks out of place with the rest of openpilot. I also don't think adding more icons to the UI is the right solution here.

Np :) the best solution is probably to just add a page to the training guide, most users can figure stuff out pretty quickly anyway

adeebshihadeh:
It's part of the new training guide in 0.8.3 :)

PR #20577, opened 2021-04-04
updater: fix makefile, it works now

deanlee:


installer/updater/Makefile
```diff
@@ -46,9 +39,11 @@ updater: updater.o $(OBJS)
 	@echo "[ LINK ] $@"
 	$(CXX) $(CPPFLAGS) -fPIC -o 'updater' $^ \
     $(FRAMEBUFFER_LIBS) \
-    $(CURL_LIBS) \
-    $(BORINGSSL_LIBS) \
     -L/system/vendor/lib64 \
+    -L/usr/lib \
+    -lcurl \
```
adeebshihadeh, line 44: why didn't the ones in phonelibs work? now it's linking against different libs, which are dynamic

PR #20579, opened 2021-04-04
Qt NEOS updater

deanlee:


adeebshihadeh:
I think we should keep the NEOS updater as is for two reasons:

adeebshihadeh:
I think we want to keep the updater as-is for a couple of reasons. Users coming from old, pre-Qt releases should be able to update in the normal flow. Also, we know the current updater is reliable, so there's no reason for this big change given the limited number of NEOS releases.

PR #20584, opened 2021-04-05
loggerd: log frame socket in encoder_thread

deanlee:


pd0wm:
Closing because of https://github.com/commaai/openpilot/pull/20613

PR #20590, opened 2021-04-06
New AR sensor

robbederks:
Still very much a WIP


geohot:
I'd like visiond to support both sensors, does the 233 really have less height?

robbederks:
> I'd like visiond to support both sensors, does the 233 really have less height?

Of course, this was just to get it to a point of streaming something. Full res should be 2048x1280, these values were just the defaults in the registers.

PR #20600, opened 2021-04-06
Add more 2017 A3 FW (Agent Random)

godoppl:
Added fingerprints for:
2017 Audi A3 e-tron Sportback - Premium Plus (ACC & LKAS)

A similar make and model is already in the system.


adeebshihadeh:
What's your dongle ID?

VirtuallyChris:
Just got the dongle ID and verified it 👍

PR #20610, opened 2021-04-07
locationd and paramsd using cython version of ekfsym

jwooning:


SConstruct
```diff
@@ -356,6 +356,28 @@ else:
 
 Export('common', 'gpucommon', 'visionipc')
 
+# Build rednose library and ekf models
+
+rednose = {
```
pd0wm, line 361:
```suggestion
rednose_config = {
```

PR #20613, opened 2021-04-07
Move loggerd rotation logic to encoder thread

pd0wm:
After this PR it's no longer guaranteed that the `XXXCameraState` packets are in the same segment as the actual frame, however it makes the rotation logic more stable.

### DIFF
--- a/selfdrive/loggerd/loggerd.cc
+++ b/selfdrive/loggerd/loggerd.cc
@@ -151,6 +151,16 @@ struct LoggerdState {
 };
 LoggerdState s;
 
+static void trigger_rotate() {
+  int err = logger_next(&s.logger, LOG_ROOT.c_str(), s.segment_path, sizeof(s.segment_path), &s.rotate_segment);
+  assert(err == 0);
+  LOGW("rotated to %s", s.segment_path);
+
+  for (auto &r : s.rotate_state) r.rotate();
+
+  s.last_rotate_tms = millis_since_boot();
+}
+
 void encoder_thread(int cam_idx) {
   assert(cam_idx < LOG_CAMERA_ID_MAX-1);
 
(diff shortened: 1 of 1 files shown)


deanlee:
how about refactor rotation after this ?  https://github.com/commaai/openpilot/pull/20584

pd0wm:
If we change some logic in our backend we don't need the frame packets in the same segment at all, the encodeIdx should contain all we need (start of frame time)

selfdrive/loggerd/loggerd.cc
```diff
@@ -214,16 +204,28 @@ void encoder_thread(int cam_idx) {
...
+          break; // only look at fcamera frame id if not QCOM2
+#endif
+      }
+
+      if (new_segment) trigger_rotate();
+      pthread_mutex_unlock(&s.rotate_lock);
+
+      // check if this encoder needs to rotate
       {
         pthread_mutex_lock(&s.rotate_lock);
```
pd0wm, line 226: @deanlee Do your remember the purpose of this lock & unlock sequence?

deanlee: I don’t know why they are here too, seems unnecessary.

pd0wm: ok, will remove then!

pd0wm: Let me know if you have other comments on this PR!

deanlee:
If we don't need to consider the integrity of the data in the same segment, it should be fine to do rotation based on time in the main thread.did I get that right?

pd0wm:
We still want exactly 1200 frames in each segment, so that's why the encoders have to decide on rotation.

deanlee:
I am trying to do rotation in the main thread, which may reduce the complexity  https://github.com/commaai/openpilot/pull/20696

pd0wm:
The problem with rotating in the main thread is that loggerd is not a real-time process. It's possible that more than 1200 frames have been encoded while the main thread is not scheduled for say 100ms.

PR #20615, opened 2021-04-08
Toyota: use FW query for DSU detection

adeebshihadeh:


pd0wm:
Did you run something on a few thousand segments to see if the DSU always responds nicely? If it wouldn't respond it would still fingerprint, so you wouldn't directly notice this.

adeebshihadeh:
Checked all the Toyota routes from the last week and this matches the old behavior for 15824 / 16163. Most of the mismatches don't get any FW at all, not just the DSU. We can hold off on merging this cleanup until we look into firmware fingerprinting reliability.

VirtuallyChris:
Noticed some people don't plug in the comma power v2, or they do have it plugged in but have some sort of connection issue. They likely wouldn't know if their car works using fp1.0 anyway :/

pd0wm:
FW query is reliable enough for DSU detection. We will only emulate the DSU if
1) The fingerprint source is FW (e.g. ecu fw query works)
2) The DSU is not present in the fw list

This will drop support for cars with unplugged DSUs without comma power plugged in. Since CAN fingerprinting will be dropped soon anyway this shouldn't be a problem.

VirtuallyChris:
How will this work with S-DSU? Will it still work as it did before?

adeebshihadeh:
Doesn't affect SDSU.

pd0wm:
We don't have access to the fingerprint source in `get_params` so changed it to just check the length of the firmware array to check if the comma power was plugged in.

PR #20616, opened 2021-04-08
MPC retune for laneless fix

haraschax:


selfdrive/controls/lib/lateral_planner.py
```diff
@@ -47,9 +47,13 @@
 
 
 class LateralPlanner():
-  def __init__(self, CP):
+  def __init__(self, CP, use_lanelines=None):
```
pd0wm, line 50: Should we just pass in both toggle values and have later_planner read the params?

HaraldSchafer: Ah yes good idea

Issue #20591, opened 2021-04-06
Spinner during local compilation consumes ~17% CPU


crdnl:
**Describe the bug**

During compilation on the local device, the spinner "./qt/spinner" consumes 17% of the CPU on the device. Considering the already constrained processor, it may be worth investigating on how to decrease the utilization.

**How to reproduce or log data**
1. Open an SSH connection to a device.
2. `rm -rf /data/openpilot`
3. `git clone --recurse-submodules https://github.com/commaai/openpilot.git /data/openpilot`
4. Reboot the device
5. Let the device start up and display the spinner on screen while compiling (a progress bar should appear at the bottom)
6. Reconnect your SSH session to the device
7. `htop`

**Device/Version information (please complete the following information):**
 - Device: comma two
 - Dongle ID: 40ca690add6c053f
 - Version: d40891739fa72e0ed658c2beca15f303def08681

**Additional context**
Although not important for an end user, for a developer this could result in a decrease in time for developer to work on the project.

adeebshihadeh:
@deanlee interested in this one?

deanlee:
Perhaps no better way to do this, 30fps requires 8% cpu even without any drwaing. the quicker solution is to set it to 20fps

deanlee:
Perhaps no better way to do this, 30fps requires 8% cpu even without any drwaing. the quicker solution is to use OpenGL 2d painting ,and set fps to 20

adeebshihadeh:
Reduced a bit in #20618. Further improvements are welcome, but at its current CPU usage, this isn't really of concern as it doesn't affect the experience for release2 users.

PR #20618, opened 2021-04-08
Qt: Faster spinner, reduce cpu usage from ~17% to ~7%

deanlee:
related to https://github.com/commaai/openpilot/issues/20591, use OpenGL 2D painting and set fps to 20

30fps : ~11% cpu usage
25fps: ~9% cpu usage
20fps: ~%7 cpu usage


pd0wm:
Sure the asset path needs to change? No need to check in any binaries, we will update those after merging.

Issue #20091, opened 2021-02-16
Bootlogs are written before time is fetched from Panda/NTP


spektor56:
**Describe the bug**

I have two folders, "1970-01-01--00-00-53--0" and "1970-01-01--00-01-03--0" in my realdata folder, the changed time are "1969-12-31 7:00:53" and "1969-12-31 7:01:03" respectively.  The rest of the dates are correct.

**How to reproduce or log data**

Haven't tried to reproduce yet, just notice when browsing through my C2 via SFTP.  Date command returns "Tue Feb 16 16:32:42 GMT 2021" which is correct for me.
[dates.zip](https://github.com/commaai/openpilot/files/5990126/dates.zip)

**Expected behavior**

No files from 1969/1970

**Device/Version information (please complete the following information):**
 - Device: comma two
 - Dongle ID: f280907e439de88d
 - Version: 0.8.1
 - Car make/model: 2017 Honda CRV LX AWD (Canadian model)

pd0wm:
Those are bootlogs created by manager before the time was pulled from the panda RTC. We should probably wait with writing bootlogs before we have a time.

PR #20620, opened 2021-04-08
Fix: bootlogs are written before time is fetched 

deanlee:
related to: https://github.com/commaai/openpilot/issues/20091


pd0wm:
We should probably sync the time before starting manager than to wait with the bootlogs. This problem affects more than just bootlogs.

We already wrote a small script to talk to the panda directly (if it's online) and fetch the time, we just need to find a good place to put it.

PR #20628, opened 2021-04-08
updated readme for community supported cars

VirtuallyChris:
After looking at the wiki, I realized most makes already had this section. I cleaned them up for all makes to make them consistent, and added a rudimentary Tesla page until someone from the #tesla channel fixes it up.


adeebshihadeh:
Moved it down to prevent confusion between "community maintained" and "community supported".

PR #20629, opened 2021-04-08
Loosen ecamera calib checks

haraschax:

### DIFF
--- a/selfdrive/locationd/calibrationd.py
+++ b/selfdrive/locationd/calibrationd.py
@@ -64,8 +64,7 @@ class Calibrator():
     # Read saved calibration
     params = Params()
     calibration_params = params.get("CalibrationParams")
-    self.wide_camera = (params.get('EnableWideCamera') == b'1') if TICI else False
-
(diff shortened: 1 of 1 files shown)


selfdrive/locationd/calibrationd.py
```diff
@@ -63,6 +64,7 @@ def __init__(self, param_put=False):
     # Read saved calibration
     params = Params()
     calibration_params = params.get("CalibrationParams")
+    self.wide_camera = (params.get('EnableWideCamera') == b'1') if TICI else False
```
adeebshihadeh, line 67:
```suggestion
    self.wide_camera = TICI and params.get('EnableWideCamera') == b'1'
```

pd0wm: also use params.get_bool('EnableWideCamera')

pd0wm, line 67:
```suggestion
    self.wide_camera = TICI and params.get_bool('EnableWideCamera')
```

Issue #20593, opened 2021-04-06
Car fw versions in useradmin don't have proper formatting


VirtuallyChris:
Before 0.8.3, carfw was shown in the same format as it is in the values.py file.

Since many people still use useradmin to fingerprint their cars, we should make the logs match what they would see in values.py to prevent PRs [like this](https://github.com/commaai/openpilot/pull/20583)

![image](https://user-images.githubusercontent.com/37757984/113651090-366e2f80-9646-11eb-9a56-20cdd0320462.png)

gregjhogan:
This was a temporary fix until capnproto 0.9.0 is released with https://github.com/capnproto/capnproto/pull/1143

VirtuallyChris:
Ah I see, sucks that there hasn't been a release in over a year :/

adeebshihadeh:
There's unlikely to be a new capnp release soon. @gregjhogan can we just build capnp from source in useradmin for now?

gregjhogan:
Tried using the bleeding edge capnproto and it is actually worse (all non-ascii characters change to octal notation)

![image](https://user-images.githubusercontent.com/4112046/117245778-394c7380-adf0-11eb-8209-dc491b9ee138.png)

not exactly sure why it is doing this, it is somewhere inside capnproto

an alternative would be to switch from `str()` to `to_dict()` and then use `pprint` to format it, but this approach leads to re-ordering of the keys and turning off all the sorting doesn't fix it (`logMonoTime` goes to the bottom)

adeebshihadeh:
This isn't really an actionable openpilot issue and personally don't think the format is a huge deal. We have a test to check for duplicate FW. Closing and will open an issue in xx to track this when we get some more time to look into it.

PR #20642, opened 2021-04-10
Add carFw values to support 2020 Acura RDX Base

afoltz3:

**Description** 2020 Acura RDX Base model is showing Car Unsupported error. Obtained carFw values from useradmin and added them to this file. Also renamed the car class description to be appropriately broad.

Note that I did not add or change any v1 fingerprinting as (1) I understand this feature is deprecated, and (2) it appears that it is ignored for purposes of the ACURA_RDX_3G car.

**Verification** I have not verified this fix, as I'm still having issues SSHing into the Comma device after the 0.8.3 update.

**Route** No route.


selfdrive/car/honda/values.py
```diff
@@ -50,7 +50,7 @@ class CAR:
   ODYSSEY = "HONDA ODYSSEY 2018 EX-L"
   ODYSSEY_CHN = "HONDA ODYSSEY 2019 EXCLUSIVE CHN"
   ACURA_RDX = "ACURA RDX 2018 ACURAWATCH PLUS"
-  ACURA_RDX_3G = "ACURA RDX 2020 TECH"
+  ACURA_RDX_3G = "ACURA RDX GEN 3 (2019 - 2021)"
```
adeebshihadeh, line 53: revert this

adeebshihadeh:
We'll need your dongle ID before merging this.

selfdrive/car/honda/values.py
afoltz3: Done - thanks

afoltz3:
> We'll need your dongle ID before merging this.

Great - dongle ID is 3a24e49ad10d01e9

afoltz3:
I see unit tests failed due to duplicate FW versions. I did not think through that x00 would be seen the same as 000. I'll modify the file by removing the 7 duplicates that caused that test to fail.

VirtuallyChris:
Please replace the 000 with x00 like the others. This is a visual bug in useradmin. See https://github.com/commaai/openpilot/issues/20593

PR #20643, opened 2021-04-10
CameraBuf: avoid busy waiting on safe_queue

deanlee:


adeebshihadeh:
What does this fix? Seems like more code to do nearly the same thing.

adeebshihadeh:
Why raise the timeout? Are we hitting it now when we shouldn't?

deanlee:
I'm still testing, I think it is unreasonable to wait for 1ms in a loop, which may cause an unnecessary increase in cpu usage

deanlee:
by the way, I encountered some problems when compiling master on EON:
```

/data/data/com.termux/files/usr/bin/aarch64-linux-android-ld: rednose/helpers/common_ekf.os: Relocations in generic ELF (EM: 62)
/data/data/com.termux/files/usr/bin/aarch64-linux-android-ld: rednose/helpers/common_ekf.os: Relocations in generic ELF (EM: 62)

```

deanlee:
It can save 3~4% CPU usage according to my test,

deanlee:
Maybe 50ms will be better

adeebshihadeh:
I see, we're hitting this time out all the time when we should only hit this on exit. Tested 50ms and seems to be a nice spot where it's still responsive and we get all the CPU savings. Nice find :)

PR #20651, opened 2021-04-11
camerad: SubMaster never updated

deanlee:


adeebshihadeh:
Hmm, this is definitely a bug, but this isn't the right place to put the update. Putting things like the SubMaster update in functions like `driver_cam_auto_exposure` can have weird side effects that are hard to debug later if you want to add another service to that SubMaster.

deanlee:
Submaster is not thread safe and can only be used in one thread. yes you're right, we need to find a good place to put it. currently I think static it in  driver_cam_auto_exposure  looks reasonable.

selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -387,6 +387,7 @@ static void driver_cam_auto_exposure(CameraState *c, SubMaster &sm) {
 
 void common_process_driver_camera(SubMaster *sm, PubMaster *pm, CameraState *c, int cnt) {
   if (cnt % 3 == 0) {
+    sm->update(0);
```
adeebshihadeh, line 390: this doesn't really make sense either unless we restrict the scope of this SubMaster more to the driver cam

adeebshihadeh:
@deanlee ended up just putting the update in `common_process_driver_camera` for now to get this fixed, but we should really refactor camerad to do this kind of stuff cleanly.

PR #20652, opened 2021-04-11
camerad: refactor exposure related functions

deanlee:
1.  use struct `ExpRect` as parameter for `set_exposure_target`.
2.  remove  `SubMaster *sm` from `MultiCameraState`. use static SubMaster in `driver_cam_auto_exposure`
3. new function `road_cam_auto_exposure` in `camera_common`


adeebshihadeh:
While this might make camerad more readable locally within these functions, I think we should refactor towards make all of camerad more readable. What if I want to use `driverState` outside of `driver_cam_auto_exposure`?

deanlee:
merged to https://github.com/commaai/openpilot/pull/20653

PR #20653, opened 2021-04-11
camerad: refactor MultipleCameraState to class CameraServer

deanlee:


pd0wm:
Is there any way we can break this PR up in smaller diffs?

There is too much stuff being moved around at the same time, so it's really hard to make sure nothing is omitted or broken. Camerad is pretty brittle and doesn't have a lot of tests around things like auto exposure.

deanlee:
ok, I'll split it into small PRs

adeebshihadeh:
I'm going to close this. Most of the code is pretty good now, so we can't afford to spend a bunch of time reviewing and verifying these refactors anymore. Bugfixes are always worth it though. Thanks for helping us get to a pretty good point with the codebase :)
 
We'll come back to camerad soon for some improvements we want to make for the comma three.

PR #20655, opened 2021-04-11
util.cc: reduce two read_file functions into one

deanlee:
`read_file` is only used here:
https://github.com/commaai/openpilot/blob/91504176e403ca1bb6bd9491ce9e2075b7f4bae0/selfdrive/modeld/runners/snpemodel.cc#L28

### DIFF
--- a/selfdrive/common/util.cc
+++ b/selfdrive/common/util.cc
@@ -11,16 +11,6 @@
 #include <sched.h>
 #endif // __linux__
 
-int write_file(const char* path, const void* data, size_t size, int flags, mode_t mode) {
-  int fd = open(path, flags, mode);
-  if (fd == -1) {
-    return -1;
-  }
-  ssize_t n = write(fd, data, size);
-  close(fd);
-  return (n >= 0 && (size_t)n == size) ? 0 : -1;
-}
-
 void set_thread_name(const char* name) {
 #ifdef __linux__
   // pthread_setname_np is dumb (fails instead of truncates)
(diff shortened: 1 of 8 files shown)


adeebshihadeh:
Can we just use the other read_file instead of having two different implementations?

deanlee:
`util::read_file `is not suitable to read large file because it has redundant copies and many resize operations during read large file. it's almost 10x slower than read_file (tested on EON):
```
util::read_file("supercombo.dlc") : 2599 ms
read_file("supercombo.dlc", &out_size) : 260 ms
```

I suggest to keep `read_file` and optimize `util::read_file `later

pd0wm:
Can we just optimize it now? Is it fast if you reserve the correct size before reading?

deanlee:
updated. This implementation is almost as fast as the C version.

selfdrive/common/util.cc
```diff
@@ -84,3 +54,19 @@ int set_core_affinity(int core) {
...
 #endif
 }
+
+namespace util {
+
+std::string read_file(const std::string& fn) {
+  std::ifstream ifs(fn, std::ios::binary | std::ios::ate);
+  if (ifs) {
+    std::ifstream::pos_type pos = ifs.tellg();
+    std::string result(pos, '\0');
```
pd0wm, line 64: Why not use resize? Now you're writing the array twice.

selfdrive/common/util.cc
```diff
@@ -84,3 +54,20 @@ int set_core_affinity(int core) {
   return -1;
 #endif
 }
+
+namespace util {
```
adeebshihadeh, line 58: why is read_file in the namespace but not write_file?

selfdrive/common/util.cc
```diff
@@ -84,3 +54,20 @@ int set_core_affinity(int core) {
...
+  std::ifstream ifs(fn, std::ios::binary | std::ios::ate);
+  if (ifs) {
+    std::ifstream::pos_type pos = ifs.tellg();
+    std::string result;
+    result.resize(pos);
+    ifs.seekg(0, std::ios::beg);
+    ifs.read(result.data(), pos);
+    if (ifs) return result;
+  }
+  return {};
```
adeebshihadeh, line 70: why not? think this is more readable
```suggestion
  return "";
```

PR #20656, opened 2021-04-11
2019 RDX Tech fingerprint

afoerster:

**Description** 
2019 RDX Tech fingerprint

**Verification**
Test driven. It will disengage periodically on master branch, this didn't happen though when running on release-2. On both branches there were bad oscillations, I hope to add some steering improvements in a later PR.

**Route**
6a8fa21108f02593%7C2021-04-11--08-41-17--0

Dongle ID 6a8fa21108f02593

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ Supported Cars
 | ----------| ------------------------------| ------------------| -----------------| -------------------| ------------------|
 | Acura     | ILX 2016-19                   | AcuraWatch Plus   | openpilot        | 25mph<sup>1</sup>  | 25mph             |
 | Acura     | RDX 2016-18                   | AcuraWatch Plus   | openpilot        | 25mph<sup>1</sup>  | 12mph             |
-| Acura     | RDX 2020-21                   | All               | Stock            | 0mph               | 3mph              |
+| Acura     | RDX 2019-21                   | All               | Stock            | 0mph               | 3mph              |
 | Honda     | Accord 2018-20                | All               | Stock            | 0mph               | 3mph              |
 | Honda     | Accord Hybrid 2018-20         | All               | Stock            | 0mph               | 3mph              |
 | Honda     | Civic Hatchback 2017-21       | Honda Sensing     | Stock            | 0mph               | 12mph             |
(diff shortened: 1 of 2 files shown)


vanillagorillaa:
Is the 2019 RDX Bosch or Nidec hardware?

afoerster:
It's Bosch

afoerster:
> It will disengage periodically on master branch, this didn't happen though when running on release-2

It's possible this had to do with weather conditions, it was clear last night (release-2) and wet roads, very light sprinkle this morning (master).

vanillagorillaa:
Could you put the new firmware version in numerical order with the ones already in there?

adeebshihadeh:
I'd like to understand the random disengagements before merging this. Can you provide a segment where this happened? Does it silently disengage or does an alert show?

afoerster:
They're not random, it happens going around corners. It happens when I'd expect a `steerSaturated` warning but instead it's a `steerError` and a 'steering not available message' is shown.

Is `steerError` always the result of the steer_status in carstate.py? I'm trying to print out what the steer_status is when steerError is set but I don't see a way to log and my print statements don't get to the main tmux page.

I don't actually want to use the cloudlog, I'm guessing that goes into the qfiles and I don't know how to read those. This is what I'm trying to do:
```
     if ret.steerError:
      message = "steerError " + ret.steerError
      print(message)
      cloudlog.warning(message)
``` 

Here's a route you'll find some: https://my.comma.ai/useradmin/?onebox=6a8fa21108f02593%7C2021-04-11--15-31-20

adeebshihadeh:
Ah, this is a result of a [recent update](2fa1b2572dfb63e367cfe2868da576090b227ca4) to the steering unavailable alert, which is why this doesn't happen on release2. Instead of a simple alert on the temporary loss of steering, we now soft disengage. If this happens frequently for you, can you open a bug report and we'll look into it?

As a side note, if you want to read your logs, we have some nice tools in the tools directory. Check out [PlotJuggler](https://github.com/commaai/openpilot/tree/master/tools/plotjuggler) if you want to plot existing signals and [LogReader](https://github.com/commaai/openpilot/blob/master/tools/lib/logreader.py) if you want all the data. Here's a simple plot of steerWarning and steerError from plotjuggler.
image

PR #20657, opened 2021-04-11
hyundai: disable steering on all steer warnings

gregjhogan:
The steering warning now covers the 90 deg steering limit as well as other checks mentioned in https://github.com/commaai/openpilot/pull/19628

Need others to test and verify this is true.  It should actually handle this situation better now by slowing winding down when the wheel exceeds 90 deg.

This PR depends on https://github.com/commaai/openpilot/pull/20511


adeebshihadeh:
Though #20511 was closed, I think the change to carState.steerWarning is better than the hardcoded angle limit. Once the wind down is removed, this should be good to merge.

gregjhogan:
If I remove #20511 from this PR to the user the behavior is the same - the power steering will fault (but we didn't inject it) when the angle limit is hit (or any of the checks mentioned in #19628) and then wind down happens.  I think this it is better, but I guess I would like you to test this out and decide if you like it or not before I put any more effort into this PR.

adeebshihadeh:
As long as we only send the wind down bit while engaged, I think this is a nice use for it.

gregjhogan:
chery-pick the last commit and try it on the sonata to see if the experience is acceptable.  I believe it was determined that #20511 wasn't the desired openpilot experience (similar to no disengage on gas), so I want to see how people feel about it when they experience it when you disengage while it is winding down, and it would be nice to have someone test this on another car (for broader testing of this PR and because I no longer have an angle limit).

adeebshihadeh:
This will need accompanying panda safety changes to enforce the fault bit is only sent while engaged.

gregjhogan:
I can't think of any reason panda safety should need to be changed.  I thought we were going to remove the first commit (inducing a fault) if the experience where it might fight you on disengage after fault is acceptable.

PR #20674, opened 2021-04-13
PQ-VW

pd0wm:
Not sure if this will be upstreamed yet. For internal testing only.

Known bugs:
- [ ] No warning for the 6 minute steering lockout
- [x] No warning for no steering below minimum steer speed
- [ ] Disengage on gas only sometimes works

Our hardware currently intercepts the powertrain, requiring moving some signals to the `cam` parser. Once we start using the official J533 harness this will be reverted, and all signals will be send/received from bus 1 (instead of 0 and 2).

### DIFF
--- a/selfdrive/car/volkswagen/carstate.py
+++ b/selfdrive/car/volkswagen/carstate.py
@@ -257,7 +257,6 @@ class CarState(CarStateBase):
 
     # Pick up the GRA_ACC_01 CAN message counter so we can sync to it for
     # later cruise-control button spamming.
-    # FIXME: will need msg counter and checksum algo to spoof GRA_neu
     self.graMsgBusCounter = pt_cp.vl["GRA_Neu"]['GRA_Neu_Zaehler']
 
     # Check to make sure the electric power steering rack is configured to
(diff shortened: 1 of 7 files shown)


actuallylemoncurd:
" No warning for the 6 minute steering lockout"
I have a fix for this! I am swamped with work at the moment, but will hopefully dump all the stuff needed for it today
You get a warning at 5.5min to grab the wheel, if you ignore it for 15 seconds you get an alarm, if you ignore again then it will auto bypass as to not lose steering until you turn op control off then on

selfdrive/car/volkswagen/carstate.py
```diff
@@ -408,15 +408,13 @@ def get_pq_can_parser(CP):
...
       ("Bremse_1", 100),          # From J104 ABS/ESP controller
-      ("Gate_Komf_1", 10),        # From J533 CAN gateway
     ]
 
     if CP.transmissionType == TransmissionType.automatic:
       signals += [("Waehlhebelposition__Getriebe_1_", "Getriebe_1", 0)]  # Auto trans gear selector position
       checks += [("Getriebe_1", 100)]  # From J743 Auto transmission control module
     elif CP.transmissionType == TransmissionType.manual:
-      signals += [("Kupplungsschalter", "Motor_1", 0),  # Clutch switch
-                  ("GK1_Rueckfahr", "Gate_Komf_1", 0)]  # Reverse light from BCM
```
jyoung8607: GK1_Rueckfahr is definitely used in the manual trans case, if nothing else, to provide openpilot's flip to the "Reverse Gear" display overlay when backing up, to avoid driver confusion with any stock backup camera that may exist.

jyoung8607:
I forgot to include the openpilot-level checksum validation in the quickie merge-to-current: commaai/opendbc#259. That PR is a little stale, but that code or something like it should be included in the overall car port.

PR #20676, opened 2021-04-13
VW lane lines visual indicator changes

keeleysam:
**Description** more updates of #20553 to make the car's indicators more in line with stock, especially on audi.  this also keeps visual indicators online regardless of OP enablement state, as we already have the status LED for this function.

**Verification**

**Route**
Route: 07667b885add75fd|2021-04-13--13-53-11

### DIFF
--- a/selfdrive/car/volkswagen/volkswagencan.py
+++ b/selfdrive/car/volkswagen/volkswagencan.py
@@ -15,18 +15,22 @@ def create_mqb_steering_control(packer, bus, apply_steer, idx, lkas_enabled):
   }
   return packer.make_can_msg("HCA_01", bus, values, idx)
 
-def create_mqb_hud_control(packer, bus, hca_enabled, steering_pressed, hud_alert, left_lane_visible, right_lane_visible,
-                           ldw_lane_warning_left, ldw_lane_warning_right, ldw_side_dlc_tlc, ldw_dlc, ldw_tlc):
-  if hca_enabled:
-    left_lane_hud = 3 if left_lane_visible else 1
-    right_lane_hud = 3 if right_lane_visible else 1
-  else:
+def create_mqb_hud_control(packer, bus, enabled, steering_pressed, hud_alert, left_lane_visible, right_lane_visible,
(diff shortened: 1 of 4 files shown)


jyoung8607 (approved):

jyoung8607:
LGTM in concept, I'll try to road-test this on my 2018 Golf R tonight or tomorrow.

jyoung8607:
@keeleysam Don't worry about the process_replay failure for now. It's spotting the difference in sendcan output for LDW_02, by design. Once we're otherwise okay with the change, the commit-reference has to be updated and the matching logs have to be uploaded to a special storage bucket only comma employees have access to. There's no need to change out the test reference drive (and the process_replay test drive is actually separate from this list) so you can revert that change.

selfdrive/car/volkswagen/volkswagencan.py
```diff
@@ -15,20 +15,26 @@ def create_mqb_steering_control(packer, bus, apply_steer, idx, lkas_enabled):
...
+  # 0 (LKAS disabled) - off 
+  # 1 (LKAS enabled, no lane detected) - dark gray 
+  # 2 (LKAS enabled, lane detected) - light gray on VW, green or white on Audi depending on year or virtual cockpit.  On a color MFD on a 2015 A3 TDI it is white, virtual cockpit on a 2018 A3 e-Tron its green. 
+  # 3 (LKAS enabled, lane departure detected) - white on VW, red on Audi 
+
+  # not using standstill state right now, though we may want to do something with this if there's too much jitter at a stop to just freeze lane visibility.
+
+  # since the LDW_Status_LED already indicates whether OP is enabled or not, we can always show the lane visibility.
+  # this does differ from stock where it would always be 0 if LKAS is disabled, but this should be less distracting and users should use the status LED as a visual indicator of OP enabled ment state instead of the absence or presence of lane lines.
+  # as OP will notify users of LDW regardless of enablement state, this should provide the best user experience and users shouldn't go from 0/1 directly to 3.
```
adeebshihadeh, line 31: remove these lines

keeleysam: ok

PR #20678, opened 2021-04-14
update time from RTC on manager init

### LINKED ISSUE #20091: Bootlogs are written before time is fetched from Panda/NTP

adeebshihadeh:
fixes #20091


selfdrive/manager/manager.py
```diff
@@ -21,6 +22,10 @@
 
 
 def manager_init():
+
+  # update system time from panda
+  set_time(cloudlog)
```
pd0wm, line 27: Should we also put a call in the launch script just before starting the agnos/neos updater?

pd0wm (approved):

selfdrive/manager/manager.py
adeebshihadeh: think it’s best to keep as much in manager as possible, there’s no logging or error handling in the launch script

pd0wm: Would be nice to sync the time before trying to download an OS update though. AGNOS updater will retry until it succeeds, but think the NEOS downloader will just fail and prompt you to reboot if you start it too quickly.

PR #20690, opened 2021-04-16
Switch to new sentry_sdk

### LINKED ISSUE #20373: Migrate to new sentry_sdk
Migrate from `raven` to `sentry_sdk`

pd0wm:
~For some reason exceptions don't show up in the sentry dashboard. If running with debug=True, it looks like the events are captured correctly.~

The new sentry sdk doesn't automatically flush on multiprocess exit. Fixed with an explicit flush.

- More consistent metadata
- up to 16kb tombstones using new extra data field

### DIFF
--- a/selfdrive/crash.py
+++ b/selfdrive/crash.py
@@ -1,7 +1,6 @@
 """Install exception handler for process crash."""
 import os
 import sys
-import threading
 import capnp
 from selfdrive.version import version, dirty, origin, branch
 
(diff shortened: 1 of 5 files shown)


selfdrive/crash.py
```diff
@@ -18,60 +17,24 @@ def bind_user(**kwargs):
...
-        except Exception:
-          sys.excepthook(*sys.exc_info())
-
-      self.run = run_with_except_hook
+    for k, v in kwargs.items():
+      sentry_sdk.set_tag(k, v)
 
-    threading.Thread.__init__ = init
+  sentry_sdk.init("https://a8dc76b5bfb34908a601d67e2aa8bcf9@o33823.ingest.sentry.io/77924",
+                  default_integrations=False, integrations=[ThreadingIntegration(propagate_hub=True)],
```
pd0wm, line 39: the threading integration replaced the old `install()` function.

PR #20693, opened 2021-04-17
Refactor cancel button spam for Chrysler

xps-genesis:
Modify button spam logic to avoid acc faults when pcm_cancel_cmd = True

### DIFF
--- a/selfdrive/car/chrysler/carcontroller.py
+++ b/selfdrive/car/chrysler/carcontroller.py
@@ -13,6 +13,7 @@ class CarController():
     self.car_fingerprint = CP.carFingerprint
     self.gone_fast_yet = False
     self.steer_rate_limited = False
+    self.wheel_button_counter_prev = 0
 
     self.packer = CANPacker(dbc_name)
 
@@ -45,10 +46,14 @@ class CarController():
     can_sends = []
 
     #*** control msgs ***
+    
+    wheel_button_counter_change = CS.wheel_button_counter != self.wheel_button_counter_prev
+    if wheel_button_counter_change:
+      self.wheel_button_counter_prev = CS.wheel_button_counter
 
-    if pcm_cancel_cmd:
+    if pcm_cancel_cmd and (self.ccframe % 10 < 5) and wheel_button_counter_change:
       # TODO: would be better to start from frame_2b3
-      new_msg = create_wheel_buttons(self.packer, self.ccframe, cancel=True)
+      new_msg = create_wheel_buttons(self.packer, CS.wheel_button_counter + 1, cancel=True)
       can_sends.append(new_msg)
 
     # LKAS_HEARTBIT is forwarded by Panda so no need to send it here.
--- a/selfdrive/car/chrysler/carcontroller.py
+++ b/selfdrive/car/chrysler/carcontroller.py
@@ -51,7 +51,7 @@ class CarController():
     if wheel_button_counter_change:
       self.wheel_button_counter_prev = CS.wheel_button_counter
 
-    if pcm_cancel_cmd and (self.ccframe % 10 < 5) and wheel_button_counter_change:
+    if pcm_cancel_cmd and wheel_button_counter_change:
       # TODO: would be better to start from frame_2b3
       new_msg = create_wheel_buttons(self.packer, CS.wheel_button_counter + 1, cancel=True)
       can_sends.append(new_msg)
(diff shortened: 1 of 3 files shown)


selfdrive/car/chrysler/carcontroller.py
```diff
@@ -45,10 +46,14 @@ def update(self, enabled, CS, actuators, pcm_cancel_cmd, hud_alert):
     can_sends = []
 
     #*** control msgs ***
+    
+    wheel_button_counter_change = CS.wheel_button_counter != self.wheel_button_counter_prev
+    if wheel_button_counter_change:
+      self.wheel_button_counter_prev = CS.wheel_button_counter
 
-    if pcm_cancel_cmd:
+    if pcm_cancel_cmd and (self.ccframe % 10 < 5) and wheel_button_counter_change:
```
adeebshihadeh, line 54: what does `self.ccframe % 10 < 5` do?

selfdrive/car/chrysler/carstate.py
```diff
@@ -96,6 +98,7 @@ def get_can_parser(CP):
       ("COUNTER", "EPS_STATUS", -1),
       ("TRACTION_OFF", "TRACTION_BUTTON", 0),
       ("SEATBELT_DRIVER_UNLATCHED", "SEATBELT_STATUS", 0),
+      ("COUNTER", "WHEEL_BUTTONS", 0),
```
adeebshihadeh, line 101: add this to the checks

selfdrive/car/chrysler/carcontroller.py
xps-genesis: sends cancel for 50msec and gives a 50msec break before resending for 50msec

adeebshihadeh: how often is the first message not accepted?

xps-genesis: don't have any data, but have had issue with continuous spamming.

adeebshihadeh: can you try removing the that condition? I'd also like some information on how often the first cancel isn't taken.

xps-genesis: alright.  u probably can gather that info from all the chrysler users. my vehicle is a my21 pacifica and it had more issues with than older pacificas.

selfdrive/car/chrysler/chryslercan.py
```diff
@@ -48,10 +48,10 @@ def create_lkas_command(packer, apply_steer, moving_fast, frame):
   return packer.make_can_msg("LKAS_COMMAND", 0, values)
 
 
-def create_wheel_buttons(packer, frame, cancel=False):
+def create_wheel_buttons(packer, counter, cancel=False):
   # WHEEL_BUTTONS (571) Message sent to cancel ACC.
   values = {
     "ACC_CANCEL": cancel,
-    "COUNTER": frame % 10
+    "COUNTER": counter
```
j-vanetten, line 55: Since we are adding 1 to the `WHEEL_BUTTONS`'s `COUNTER`, we need a mod 16.  This is in case the `COUNTER` is 15 and when we add 1, it should be 0 and not 16 since the `COUNTER` is 0-15

```suggestion
    "COUNTER": counter % 16
```

selfdrive/car/chrysler/carcontroller.py
j-vanetten: It shouldn't be needed as the cancel button is accepted on the down press and not up.  We should continue to send the press until OP disengages.

j-vanetten:
kudos to @xps-genesis for making the PR.

selfdrive/car/chrysler/carcontroller.py
xps-genesis: thanks

selfdrive/car/chrysler/chryslercan.py
xps-genesis: since counter itself is a 4 bit signal, it is not necessary to do the % 16, the signal will wrap around when packed on to the CAN bus. 16 will be went as 0 and 17 as 1 and so on.. but either way it is correct.

selfdrive/car/chrysler/carcontroller.py
j-vanetten: I haven't tested that.  But I know you can physically hold down buttons on the wheel without faults.

xps-genesis: ya, but the steering wheel is already sending a msg continuously and OP is trying to double send the same message when OP wants to send cancel, so the receiving module may fault out if it see discrepancy in the counter at some point. because it is see a counter value from steering button and OP. we may not see any issue if the cancel from OP got accepted quickly, which is what we are mostly seeing on our vehicles.

PR #20694, opened 2021-04-17
Chrysler BSM info signal

xps-genesis:
Add BSM signals for chrysler


adeebshihadeh:
Nice, any idea what the signal values mean? Would be nice if we can add some comments or values to the DBC.

adeebshihadeh:
@xps-genesis any update on this?

xps-genesis:
> @xps-genesis any update on this?

I don't have access to cabana. but ==1 is blind spot active

selfdrive/car/chrysler/carstate.py
```diff
@@ -96,6 +99,8 @@ def get_can_parser(CP):
       ("COUNTER", "EPS_STATUS", -1),
       ("TRACTION_OFF", "TRACTION_BUTTON", 0),
       ("SEATBELT_DRIVER_UNLATCHED", "SEATBELT_STATUS", 0),
+      ("BLIND_SPOT_RIGHT", "BLIND_SPOT_WARNINGS", 0),
+      ("BLIND_SPOT_LEFT", "BLIND_SPOT_WARNINGS", 0),
```
adeebshihadeh, line 103: are you sure this is on all cars and trims?

adeebshihadeh:
This should be good to merge if other users can confirm that this works and the checks are behind the carParams.enableBsm flag. Just check for the address in interface.py (https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/interface.py#L240).

j-vanetten:
Works on my 2018 Jeep Grand Cherokee Trailhawk.

adeebshihadeh:
@j-vanetten thanks for confirming. A lot of the Chrysler's are missing test routes, including yours. Would you be interested in giving us a route? The only requirements for the route are openpilot needs to be engaged for at least a minute, and you're ok with making it public.

j-vanetten:
> @j-vanetten thanks for confirming. A lot of the Chrysler's are missing test routes, including yours. Would you be interested in giving us a route? The only requirements for the route are openpilot needs to be engaged for at least a minute, and you're ok with making it public.

Yup, I have that bookmarked, just haven't had time to get on to stock for a drive to share.

PR #20695, opened 2021-04-17
Add fingerprint for 2017 VW Tiguan

ghost:
Cannot yet confirm openpilot works on the car, but it is from the same generation as the 2020 Tiguan that is already supported, so I don't see any reason it shouldn't work.
Route: 
650b1e4383a9baf2|2021-04-16--09-16-00

### DIFF
--- a/selfdrive/car/volkswagen/values.py
+++ b/selfdrive/car/volkswagen/values.py
@@ -238,18 +238,23 @@ FW_VERSIONS = {
   },
   CAR.TIGUAN_MK2: {
     (Ecu.engine, 0x7e0, None): [
+      b'\xf1\x8704L906026HR\xf1\x895174',
       b'\xf1\x8783A907115B \xf1\x890005',
     ],
     (Ecu.transmission, 0x7e1, None): [
       b'\xf1\x8709G927158DT\xf1\x893698',
+      b'\xf1\x870DL300045R \xf1\x892001',
     ],
     (Ecu.srs, 0x715, None): [
+      b'\xf1\x875Q0959655BJ\xf1\x890336\xf1\x82\02331310031333334313132013730379333423100',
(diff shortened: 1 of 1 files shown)


jyoung8607:
It should, those SW part numbers match known MQB vehicles.

If this is merged, use caution if/when adding to the README, and don't add it to the comma compatibility checker. They are both US-market-centric, and 2016-2018 Tiguans are vehicles where market matters.

- Everywhere but USA got the openpilot-supported MQB Mk2 Tiguan starting with MY2016.
- USA got the MQB Mk2 Tiguan starting with MY2018 *except for the special carryover 2018-only Tiguan Limited*.
- The USA 2016-2017 Tiguan and 2018 Tiguan Limited are still PQ Mk1. Untested, but they probably work with commaai/openpilot#20674.

This is exactly as strange and confusing as it sounds.
https://en.wikipedia.org/wiki/Volkswagen_Tiguan

grekiki:
Looks like adding PQ simplifies things :) Yeah this is an EU Tiguan with 2.0 TDI DSG configuration.

pd0wm:
Let me know when you can confirm openpilot support so we can merge it!

adeebshihadeh:
If this won't be tested anytime soon, let's close the PR until it's been tested.

DoskJS:
I have a 2018 VW Tiguan SE 4motion, would my car be useful for testing this? @jyoung8607

jyoung8607:
2018+ Tiguan is already fully supported by stock openpilot, provided it has factory Lane Assist. Unfortunately, for Tiguan I think that's bundled with SEL and above. Your 2018 SE can still be made to work, but needs a community built harness (due to no factory camera) and a community-supported fork of openpilot.

PR #20696, opened 2021-04-18
Refactor loggerd rotations

deanlee:
continuation of #20613


deanlee:
if reply (https://github.com/commaai/openpilot/pull/21004) can search for eidx in different segments, there is no need to synchronize frames anymore

PR #20697, opened 2021-04-18
No docker openpilot sim for testing modifications

creikey:
Useful for locally testing UI stuff without needing to build a docker image

### DIFF
--- a/tools/sim/start_openpilot.sh
+++ b/tools/sim/start_openpilot.sh
@@ -9,4 +9,9 @@ export PASSIVE="0"
 export NOBOARD="1"
 export SIMULATION="1"
 
-./launch_openpilot.sh
+tmux new -d -s htop
+tmux send-keys "./launch_openpilot.sh" ENTER
+tmux neww
+tmux send-keys "PYTHONPATH=$(pwd) python tools/sim/bridge.py" ENTER
+tmux a
+
--- a/tools/sim/start_openpilot.sh
+++ b/tools/sim/start_openpilot.sh
@@ -1,0 +1,12 @@
+#!/bin/bash
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+cd $DIR
+
+cd ../../
+
+export PASSIVE="0"
+export NOBOARD="1"
+export SIMULATION="1"
+
+./launch_openpilot.sh


tools/sim/start_openpilot.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null && pwd)"
+cd $DIR
+
+cd ../../
+
+export PASSIVE="0"
```
grekiki, line 8: Is this necessary?

creikey: Without them I get communication errors

adeebshihadeh:
There's no need for a new script. Just add the necessary environment variables to `tmux_script.sh`.

creikey:
> There's no need for a new script. Just add the necessary environment variables to `tmux_script.sh`. `launch_openpilot.sh` already sets up the environment. `tmux_script.sh` should already do what you want.

```sh
./bridge.py 
creikey@pc:~/openpilot/tools/sim$ ./bridge.py 
  File "./bridge.py", line 11, in <module>
ModuleNotFoundError: No module named 'cereal'
```
```sh
./launch_openpilot.sh
creikey@pc:~/openpilot/tools/sim$ ./launch_openpilot.sh
  File "./manager.py", line 9, in <module>
ModuleNotFoundError: No module named 'cereal'
creikey@pc:~/openpilot/tools/sim$
```

Starting from the tmux script does not properly set the `PYTHONPATH` as is done in `start_openpilot_docker.sh:14`

```sh
  EXTRA_ARGS="-v $PWD/../..:/root/openpilot -e PYTHONPATH=/root/openpilot:$PYTHONPATH"
```

adeebshihadeh:
All of our tools kind of assume that PYTHONPATH is already set to your openpilot directory, but I'd merge a PR to set it in `tools/sim/launch_openpilot.sh`.

creikey:
> All of our tools kind of assume that PYTHONPATH is already set to your openpilot directory, but I'd merge a PR to set it in `tools/sim/launch_openpilot.sh`.

I'll just set it in my bashrc, and `bridge.py` is started with no script anyways. Maybe this should be added to the sim wiki? Something like

> You can run the simulator on locally edited files by first setting your pythonpath (from the `tools/sim` directory):

`export PYTHONPATH="$(cd ../../ && pwd)"`

> Then by running `./tmux_script.sh`

PR #20713, opened 2021-04-20
params put in locationd async

### LINKED ISSUE #20700: Locationd uses blocking params.put()
Locationd uses a blocking call to params.put() to write the last gps position. This could block the whole thread, potentially leading to a communication issue alert.

jwooning:
closes #20700

### DIFF
--- a/selfdrive/locationd/locationd.cc
+++ b/selfdrive/locationd/locationd.cc
@@ -366,7 +366,7 @@ int Localizer::locationd_thread() {
         std::string lastGPSPosJSON = util::string_format(
           "{\"latitude\": %.15f, \"longitude\": %.15f, \"altitude\": %.15f}", posGeo(0), posGeo(1), posGeo(2));
 
-        std::thread([&params] (const std::string& gpsjson) {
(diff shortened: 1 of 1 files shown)


selfdrive/locationd/locationd.cc
```diff
@@ -365,7 +365,10 @@ int Localizer::locationd_thread() {
         VectorXd posGeo = this->get_position_geodetic();
         std::string lastGPSPosJSON = util::string_format(
           "{\"latitude\": %.15f, \"longitude\": %.15f, \"altitude\": %.15f}", posGeo(0), posGeo(1), posGeo(2));
-        params.put("LastGPSPosition", lastGPSPosJSON);  // TODO write async
+
+        std::thread([&params] (const std::string& gpsjson) {
```
pd0wm, line 369: Does this still work if it takes long enough for `lastGPSPosJSON` to go out of scope and the memory to be reused?

jwooning: it seems that the std::thread call made a copy anyway, but I removed the reference to make it more clear

PR #20715, opened 2021-04-20
unittests for locationd uncommon events

jwooning:


selfdrive/locationd/test/test_locationd.py
```diff
@@ -0,0 +1,165 @@
...
+from common.params import Params
+
+from selfdrive.manager.process_config import managed_processes
+
+SENSOR_DECIMATION = 10
+VISION_DECIMATION = 2
+
+LIBLOCATIOND_PATH = os.path.abspath(os.path.join(os.path.dirname(__file__), '../liblocationd.so'))
+
+random.seed(123489234)
```
pd0wm, line 20: Put this in the `setUp()`? Now it runs at import time, so other things can use the RNG inbetween.

PR #20717, opened 2021-04-21
sim: get GPS data from CARLA

creikey:
Not sure if just the lat/long/alt, which is what carla provides in the gnss callback, is enough to send to OP as there are many more fields as seen here:

https://github.com/commaai/cereal/blob/master/log.capnp#L219

If these extra fields are necessary, I can modify this PR to copy data from carla's IMU into the gps packet (stuff like heading)

I tested this PR by fetching the gps data in `ui.cc`

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -29,7 +29,7 @@ REPEAT_COUNTER = 5
 PRINT_DECIMATION = 100
 STEER_RATIO = 15.
 
-pm = messaging.PubMaster(['roadCameraState', 'sensorEvents', 'can'])
+pm = messaging.PubMaster(['roadCameraState', 'sensorEvents', 'can', "gpsLocationExternal"])
 sm = messaging.SubMaster(['carControl','controlsState'])
 
 class VehicleState:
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
@iejMac can you test this?

pd0wm:
We definitely need to set the following fields
 - flags (set to 1 to indicate valid fix)
 - vNED (3d velocity in NED reference frame, see common/transformations for a converter to compute this)
 - verticalAccuracy (1 m or something reasonable)
 - speedAccuracy (0.1 m/s or something reasonable)
 - bearingDeg

tools/sim/bridge.py
```diff
@@ -92,13 +92,17 @@ def panda_state_function():
...
-    pm.send('gpsLocationExternal', dat)
-    time.sleep(0.01)
+def gps_callback(gps):
+  dat = messaging.new_message('gpsLocationExternal')
+
+  dat.gpsLocationExternal = {
+    "latitude": gps.latitude,
+    "longitude": gps.longitude,
+    "altitude": gps.altitude,
+    "source": 5, # external, copied from log.capnp
```
pd0wm, line 102: set to ublox using capnp enums

adeebshihadeh:
Any update on this? If this will stay a draft for a while, let's close and reopen when ready.

creikey:
> Any update on this? If this will stay a draft for a while, let's close and reopen when ready.

Oh uh I've been learning some slam stuff for the programming challenge and messing with gpus, I'll finish this today

creikey:
@adeebshihadeh @pd0wm should be good to go now, although I don't have a way to verify that my vNED coordinates are correct other than printing them out while carla is running along with the bearing and them matching (180 deg bearing x is negative).

tools/sim/bridge.py
```diff
@@ -65,7 +68,8 @@ def cam_callback(image):
   pm.send('roadCameraState', dat)
   frame_id += 1
 
-def imu_callback(imu):
+def imu_callback(imu, vehicle_state):
+  vehicle_state.bearing_deg = ((imu.compass) / (2.0 * math.pi)) * 360.0
```
adeebshihadeh, line 72:
```suggestion
  vehicle_state.bearing_deg = math.degrees(imu.compass)
```

PR #20721, opened 2021-04-21
Force deceleration for seatbelt unlatched or door open

sshane:
Changes the previous behavior of disabling after a few seconds to forcing deceleration and still steering.

I think it's a little nicer to not stop steering, for example: if a driver is in the middle of a sweeping turn on a highway and they open their door to make sure it's closed/not ajar, or any other reason and they aren't aware that openpilot disengages under these circumstances. The deceleration rate is pretty slow though, if you're going up a steep hill it will still accelerate hard, and it takes quite a while to come to a stop on flat road. For example, it will theoretically take almost 2 minutes to come to a stop at 45 mph.

With this code though, the driver can close the door or relatch the seatbelt and openpilot will continue driving, but I can make the alert and decel stick until a disengage occurs, even if they close the door, etc. if that's desired

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -489,8 +489,8 @@ class Controls:
       can_sends = self.CI.apply(CC)
       self.pm.send('sendcan', can_list_to_can_capnp(can_sends, msgtype='sendcan', valid=CS.canValid))
 
-    force_decel = (self.sm['driverMonitoringState'].awarenessStatus < 0.) or \
-                  (self.state == State.softDisabling)
+    force_decel = (self.sm['driverMonitoringState'].awarenessStatus < 0. or
+                   self.state == State.softDisabling or CS.seatbeltUnlatched or CS.doorOpen)
 
     # Curvature & Steering angle
     params = self.sm['liveParameters']
--- a/selfdrive/controls/lib/events.py
+++ b/selfdrive/controls/lib/events.py
@@ -601,12 +601,20 @@ EVENTS: Dict[int, Dict[str, Union[Alert, Callable[[Any, messaging.SubMaster, boo
   },
 
   EventName.doorOpen: {
-    ET.SOFT_DISABLE: SoftDisableAlert("Door Open"),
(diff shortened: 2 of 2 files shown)


pd0wm:
We immediately disengage to prevent drivers leaving the vehicle while when they don't realize openpilot is still engaged, for example stopped behind a parked car. Most stock ACC systems would engage the EPB in this situation, but we can't do that.

If we keep the car stopped the driver would be able to leave the vehicle, if then later ACC disengages for some reason or the lead car moves away the car could start driving without a driver present.

I understand the current behavior is not ideal, but I think the alternative is worse.

ShaneSmiskol:
Interesting, I hadn't thought of that. But that seems pretty rare to me, a driver accidentally or purposefully engaging at a stop behind a parked car and then leaving the vehicle? While it's still running, engaged, and in drive? I don't have an electronic parking brake so I cannot try to figure out how to engage it over CAN sadly.

Instead we could just revert to previous behavior of soft disabling if they engage at a stop and and take off the seatbelt/door, which is already an edge case in my opinion. I understand users can and will try to break your software no matter what but this seems really specific to me.

pd0wm:
Even though the situation is rare it's still very relevant. I doesn't have to happen intentionally, if you're stopped behind a car you can mistake it for being in park and step out of the vehicle. Any audible alerts could be mistaken for a regular seatbelt alert. It's one of the reasons many car manufacturers don't do stop & go ACC if yo have no EPB.

Changing the behavior at standstill is also not really an option, since the slow down function would eventually reach standstill. Do you want it to start moving again?

Don't open the door while driving on the highway. You should never rely on openpilot to take action, it's an L2 system.

ShaneSmiskol:
That makes sense! With my car it's obvious if it's engaged or not since there's a deafening ABS pump, plus if someone has an EV it would be completely silent if it was on and in drive. Well, would you consider raising the decel rate? If someone happens to fall asleep or has another emergency 0.2 m/s/s is quite a slow rate to come to a stop.

PR #20737, opened 2021-04-23
Kill loggerd when red thermal status

sshane:
Kills loggerd (uses a fair amount of CPU when onroad) when thermal status is red (engage is disallowed anyway) so the temperature comes down quicker.

Let me know if this would affect anything else, I haven't extensively tested it or anything, it just seems like a good idea when you can leave your C2 in the car on 100F+ days.


adeebshihadeh:
If we can't support all the onroad processes, none should be running. Besides, overheating shouldn't really happen while onroad, and we don't go onroad until the temperature is low enough.

PR #20743, opened 2021-04-25
ignore_avg_freq of longitudinalPlan, which is triggered by radarState

hewers:
**Description** GM radar updates at 15 Hz not standard 20 Hz. 'radarState' and subsequent longitudinal plan is published at 15 Hz. The expected frequency is set to 20 Hz in cereal/services.py. This causes commIssue on engage for GM.

Related:
https://github.com/commaai/openpilot/commit/64d2ec38f290f7146af4196d240fb70cc11653b0
https://github.com/commaai/openpilot/pull/20558

**Verification** Engaged openpilot successful, no commIssue.

**Route**
Route: Uploading


qadmus:
+ @pd0wm

PR #20747, opened 2021-04-25
Hyundai: add BSM detection

adeebshihadeh:


Dr-Diesel:
Hello,
OP takes 1419: 8 from fingerprinted CAR.IONIQ_EV_2020 and requires BSM can signal . Getting Can Error on european Ioniq EV 2021 Smart, that lacks BSM. When deleted 1419: 8 from fingerprint, OP calibrates and runs. (fails afterwards on Can Error again with Check LKAS, AntiCollision malfunction and other errors again, but I that's not BSM related I assume).

PR #20750, opened 2021-04-26
Parse DBCs dynamically

deanlee:
related to: https://github.com/commaai/opendbc/pull/376


release/files_common
```diff
@@ -531,17 +531,13 @@ opendbc/can/common.h
...
 opendbc/can/packer_pyx.pyx
 opendbc/can/parser.cc
 opendbc/can/parser.py
 opendbc/can/parser_pyx.pyx
-opendbc/can/process_dbc.py
-opendbc/can/dbc_out/.gitkeep
-opendbc/can/dbc_out/.gitignore
+opendbc/can/dbc_pyx.pyx
 
 opendbc/chrysler_pacifica_2017_hybrid.dbc
```
adeebshihadeh, line 542: we should remove the DBC whitelist and include all the ones in the root. we didn't do this before because they got built into libdbc which was huge

adeebshihadeh:
Converting to draft until the opendbc PR is ready for review

PR #20751, opened 2021-04-26
Mapbox widget

pd0wm:


creikey:
Why use `libmapboxgl` instead of the Location api built into Qt?

selfdrive/ui/ui
```diff
@@ -1,3 +1,4 @@
 #!/bin/sh
 export LD_LIBRARY_PATH="/system/lib64:$LD_LIBRARY_PATH"
+export QT_PLUGIN_PATH="../../phonelibs/qt-plugins/$(uname -m)"
```
adeebshihadeh, line 3: should we move this stuff to process_config.py?

selfdrive/ui/qt/widgets/map.cc
```diff
@@ -0,0 +1,500 @@
...
+const int PAN_TIMEOUT = 100;
+const bool DRAW_MODEL_PATH = false;
+const qreal REROUTE_DISTANCE = 25;
+const float MAX_ZOOM = 17;
+const float MIN_ZOOM = 14;
+const float MAX_PITCH = 50;
+const float MIN_PITCH = 0;
+const float MAP_SCALE = 2;
+
+static void clearLayout(QLayout* layout) {
```
adeebshihadeh, line 23: already defined in selfdrive/ui/qt/offroad/networking.cc

pd0wm: should we make a qt_common library? In this case I thought it was better to copy it than to link against some code from the networking widget.

selfdrive/ui/ui
pd0wm: Yes. But maybe something for a standalone PR.

selfdrive/ui/SConscript
```diff
@@ -4,9 +4,16 @@ Import('qt_env', 'arch', 'common', 'messaging', 'gpucommon', 'visionipc',
 
 base_libs = [gpucommon, common, messaging, cereal, visionipc, transformations, 'zmq',
              'capnp', 'kj', 'm', 'OpenCL', 'ssl', 'crypto', 'pthread'] + qt_env["LIBS"]
+
+maps = arch in ['larch64', 'x86_64']
+
 if arch == 'aarch64':
   base_libs += ['log', 'utils', 'gui', 'ui', 'CB', 'gsl', 'adreno_utils', 'cutils', 'uuid']
 
+if arch in ['x86_64']:
```
pd0wm, line 13:
```suggestion
if maps and arch in ['x86_64']:
```

selfdrive/ui/qt/widgets/map.cc
adeebshihadeh: https://github.com/commaai/openpilot/blob/master/selfdrive/ui/qt/util.h

selfdrive/ui/ui
adeebshihadeh: @deanlee interested in fixing this up?

PR #20752, opened 2021-04-26
Replay: print timestamp every 5 seconds

iejMac:


selfdrive/ui/replay/Unlogger.hpp
```diff
@@ -31,6 +31,7 @@ Q_OBJECT
     QMap<std::string, PubSocket*> socks;
     Context *ctx;
     uint64_t tc = 0;
+    float printed_at = 0;
```
adeebshihadeh, line 34: call this last_print

PR #20774, opened 2021-04-29
Civic 2019 EU new ECU values added

GhostOfSatoshi:

### DIFF
--- a/selfdrive/car/honda/values.py
+++ b/selfdrive/car/honda/values.py
@@ -494,6 +494,7 @@ FW_VERSIONS = {
       b'37805-5AW-G720\x00\x00',
       b'37805-5AZ-E850\x00\x00',
       b'37805-5AZ-G740\x00\x00',
+      b'37805-5AZ-G840\x00\x00',
       b'37805-5BB-A530\x00\x00',
       b'37805-5BB-A540\x00\x00',
       b'37805-5BB-A630\x00\x00',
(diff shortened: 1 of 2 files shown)


pd0wm:
What is your dongle id?

selfdrive/car/honda/values.py
```diff
@@ -613,6 +615,9 @@ class CAR:
       b'38897-TBA-A110\x00\x00',
       b'38897-TBA-A020\x00\x00',
     ],
+    (Ecu.electricBrakeBooster, 0x18da2bf1, None): [
```
VirtuallyChris, line 618: Adding another ECU will break all other cars being able to fingerprint. Maybe we don't include this one in this PR?

GhostOfSatoshi:
> What is your dongle id?


205552e5b47d25a5

selfdrive/car/honda/values.py
GhostOfSatoshi: I see. But that also means that future releases won't recognize my car which has this ECU (I mean without adding this manually)?

pd0wm: I guess we could remove the brakeBooster from the list of essential ECUs: https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py#L207

pd0wm: Can you add that change to your PR?

GhostOfSatoshi: > I guess we could remove the brakeBooster from the list of essential ECUs: https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py#L207

This would be great (as far as I understand). Wouldn't brake backward compatibility but would let EU Civics work.

PR #20785, opened 2021-05-01
HttpRequest: cache SSL certificate

deanlee:


adeebshihadeh:
Can we set this globally for the QApplication somehow? `SSL_CERT_DIR` doesn't seem to work.

adeebshihadeh:
Done in 23faedd25eb6895c86bb890fdbf4f48889ecf4c3

PR #20793, opened 2021-05-02
Add route for 17 Corolla

sshane:


selfdrive/test/test_routes.py
```diff
@@ -251,20 +251,11 @@
...
-  "b0f5a01cf604185c|2018-01-26--10-54-38": {
-    'carFingerprint': TOYOTA.COROLLA,
-    'enableCamera': True,
-    'enableDsu': False,
-  },
-  "b0f5a01cf604185c|2018-01-26--10-59-31": {
-    'carFingerprint': TOYOTA.COROLLA,
-    'enableCamera': False,
-    'enableDsu': False,
+    'enableGasInterceptor': True,
```
adeebshihadeh, line 258: no need for this, it's not used anymore

selfdrive/test/test_routes.py
```diff
@@ -251,20 +251,11 @@
     'enableCamera': True,
     'enableDsu': False,
   },
-  "b0f5a01cf604185c|2018-01-26--00-54-32": {
+  "4e45c89c38e8ec4d|2021-05-02--02-49-28--15": {
```
adeebshihadeh, line 254: remove the segment

selfdrive/test/test_routes.py
ShaneSmiskol: Should enableDsu be false since I've unplugged mine?

adeebshihadeh: you can just leave out everything besides carFingerprint

adeebshihadeh:
Can you add an additional check to this line so routes with a missing enableCamera are also included? https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_models.py#L22

ShaneSmiskol:
Assuming `enableCamera` is True if it's missing? Got it. It might be nicer to switch to a base class for the routes and then define the defaults in the init function.

adeebshihadeh:
Yes. No need for a class, I'll clean this up soon.

ShaneSmiskol:
Oh whoops I already went ahead and started to refactor test_models. I can revert if you want to do it yourself, but I'll push to let you see.

selfdrive/test/test_routes.py
```diff
@@ -35,473 +35,130 @@
...
+  def __init__(self, route, car_fingerprint, *, enable_camera=True, enable_dsu=False, enable_gas_interceptor=False):
+    self.route = route
+    self.car_fingerprint = car_fingerprint
+    self.enable_camera = enable_camera
+    self.enable_dsu = enable_dsu
+    self.enable_gas_interceptor = enable_gas_interceptor
+
+
+routes = [
+  TestRoute("420a8e183f1aed48|2020-03-05--07-15-29", CHRYSLER.PACIFICA_2017_HYBRID),
```
adeebshihadeh, line 49: the other info was only for some old test we don't run anymore. we only need the route and car now, so let's revert these changes

ShaneSmiskol: No problem

ShaneSmiskol: Not sure why one of the routes is failing now...

adeebshihadeh: that's unrelated, it's a recent process replay regression

ShaneSmiskol: Thought so. Let me know if you want me to make a PR to clean up the routes dict

PR #20794, opened 2021-05-02
Torque unification testing

sshane:
Just a draft for now to record changes and receive feedback.

With the current ~lateral accel model for torque (from latcontrol_pid), it applies too little at low speeds and too much at high speeds. This normalizes the response, assuming `speed ^ 2` is correct for required torque at any speed. We add a new function similar to `compute_gb` that converts the torque before clipping.

Todo:
- [ ] Fit an (*accurate*) ff function for many cars and implement
- [x] Fit a starting point ff function for '17 Corolla and implement (fitted on only one car at 0.12 sec delay)
- [x] Implement in latcontrol_pid (uses pid's `convert` functionality)
- [ ] ~See what is needed to fit a function to output multiplier without calculating both standard ff and car-specific ff~
  - Edit: Calculating both standard ff and car-specific ff each frame usually is quicker than a fitted function, and more accurate. Will look into this more
- [ ] Implement convert function in other latcontrollers (can't do in carcontroller or controlsd because it should be applied before range clipping. ie. we'd never be able to reach max torque when multiplier is under 1)
- [ ] Quantify results. Maybe compare stock, unification on whole torque output, and just modifying feedforward (not prop and integral)

Results so far is much better control at a higher range of speeds, it should increase the chances that a single proportional and integral gain works over a larger speed range.

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -41,6 +41,10 @@ class CarInterfaceBase():
     return 1.
 
   @staticmethod
+  def compute_torque(torque, speed):
+    return float(torque)
+
+  @staticmethod
   def compute_gb(accel, speed):
     raise NotImplementedError
 
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -11,6 +11,19 @@ EventName = car.CarEvent.EventName
 
 class CarInterface(CarInterfaceBase):
   @staticmethod
+  def compute_torque(torque, speed):
+    def std_feedforward(_speed):
+      return _speed ** 2
+
+    def acc_feedforward(_speed):
+      return 0.35189607550172824 * _speed ** 2 + 7.506201251644202 * _speed + 69.226826411091
+
+    # todo: find nicer way to avoid multiplier going to inf
+    speed = max(speed, 20 * CV.MPH_TO_MS)
+    torque_mult = acc_feedforward(speed) / std_feedforward(speed)
+    return float(torque) * torque_mult
+
+  @staticmethod
   def compute_gb(accel, speed):
     return float(accel) / CarControllerParams.ACCEL_SCALE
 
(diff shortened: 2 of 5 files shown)


pd0wm:
How did you measure improvements in control? I'd like the quantify the effect of these changes on desired angle setpoint tracking.

ShaneSmiskol:
No testing yet, but if you remember from that PR a while ago to make feedforward more accurate (speed ^ 2 didn't apply enough torque at low speed, too much at high speeds, etc), this at least multiplies feedforward to match that (since it applies it to torque as a whole).

Should I use [measure_steering_accuracy.py](https://github.com/commaai/openpilot/blob/master/selfdrive/debug/internal/measure_steering_accuracy.py) live in the car on the same route, multiple times with each? Or can you export data with PlotJuggler to analyze with Python now?

pd0wm:
I imagine it better matches the torque required, but I wonder how that actually influences the final control loop performance. You should be able to change `measure_steering_accuracy` to work on Logs (see how I did it for https://github.com/commaai/openpilot/blob/master/selfdrive/debug/filter_log_message.py)

ShaneSmiskol:
I'm pretty sure it's going to be more accurate in general due to the lower kp and ki values, since the stock ones are much too high for my car. Causes jerking in turns. I'll do both runs with about the mean between stock and what I found to work great, or maybe just stock.

gregjhogan:
I think measure_steering_accuracy.py is broken now because there is no longer a desired steering angle that is logged?  Now you have to run things through the vehicle model or something to get the desired steering angle.

When I wrote that script I was trying to improve steering accuracy on my CR-V and I noticed that to get good steering accuracy I could set kp really high at low speed and much lower at high speed to improve things.  I didn't try playing with kf, will be interested to see if this is a better approach.

pd0wm:
`controlsState.steeringAngleDesiredDeg` has the correct desired angle

Issue #20799, opened 2021-05-02
C++ cleanup


adeebshihadeh:
- [x] standardize file extensions to .cc and .h
- [ ] cleanup include paths, include from the root
- [x] replace ifdefs with the HW abstraction layer where possible

adeebshihadeh:
@deanlee interested in this one? probably best to do each in a separate PR.

adeebshihadeh:
@deanlee I think we can close this if we get rid of the opendbc/cereal paths in the include path

ahowe444:
I notice there are still some .hpp in phonelibs/ directory. Are these meant to be standardized as well?

PR #20800, opened 2021-05-03
c++ cleanup: standardize file extensions to .cc and .h

deanlee:
https://github.com/commaai/openpilot/issues/20799


pd0wm:
Sorry, I created a lot of these inconsistent header file names :(

Let's do the submodules next, I know cereal has a few too.

PR #20806, opened 2021-05-03
Define frame&screen size in hardware abstraction layers

deanlee:


pd0wm:
Shouldn't we define these in the hardware abstraction layer?

pd0wm:
Merged the ifdef PR. I was thinking we add methods to the hardware abstraction layers to fetch screen size, camera sizes, and intrinsic matrices?

PR #20810, opened 2021-05-04
add CLEAR_ON_IGNITION param type

adeebshihadeh:


adeebshihadeh:
@deanlee want to move the cython specific params stuff into the cpp version? We're just missing the list of all the params and the TxType.

deanlee:
https://github.com/commaai/openpilot/pull/20814

PR #20816, opened 2021-05-04
Locationd 100 Hz

pd0wm:
- Change frequency to 100Hz
- Fix bug where stds were wrong for cameraOdometry packets
- Normalize quaternions in filter more aggressively
- drop car odometry packets to avoid unstable jacobians
- reset when non-finite values detected
- retune initial covariance of filter

Batch tests on 2k min of driving show this PR reduces yaw rate error by ~40%. Forward velocity error is increased by ~20% and lateral velocity error is reduced by ~20%. Cases where the  filter is extremely wrong or unstable is reduced by ~65%.


selfdrive/locationd/locationd.cc
```diff
@@ -239,36 +245,33 @@ void Localizer::handle_gps(double current_time, const cereal::GpsLocationData::R
...
-  this->speed_counter++;
-
-  if (this->speed_counter % SENSOR_DECIMATION == 0) {
-    this->kf->predict_and_observe(current_time, OBSERVATION_ODOMETRIC_SPEED, { (VectorXd(1) << log.getVEgoRaw()).finished() });
-    this->car_speed = std::abs(log.getVEgo());
-    if (this->car_speed < 1e-3) {
-      this->kf->predict_and_observe(current_time, OBSERVATION_NO_ROT, { Vector3d(0.0, 0.0, 0.0) });
-    }
+  //this->kf->predict_and_observe(current_time, OBSERVATION_ODOMETRIC_SPEED, { (VectorXd(1) << log.getVEgoRaw()).finished() });
+  this->car_speed = std::abs(log.getVEgo());
```
pd0wm, line 249: Why not vEgoRaw?

HaraldSchafer: It's not put into the filter so I thought smoothed is better, but can change if you want.

pd0wm: The filtered version is a little weird around sudden stops where it can overshoot and become a little bit negative. In this case I would recommend the raw speed.

If you only care about standstill I would use the dedicated standstill flag. For some cars this is a different signal and is a more accurate at very low speeds (where the wheel speeds would be zero even though you’re moving)

Issue #19739, opened 2021-01-12
Toyota: fix EPS_STATUS message length


adeebshihadeh:
See #19692. @rav4kumar pointed out that EPS_STATUS is currently 5 bytes in the Avalon DBC, but the actual length is 8 bytes. I checked our Rav 4, and it has the same discrepancy. I suspect this is wrong for many or all Toyotas.

adeebshihadeh:
We should also have strict can parser checks that would catch things like this.

adeebshihadeh:
It appears this only wrong for the 2019+ Avalon. Addressed in #21058

PR #20824, opened 2021-05-04
Add ESP f/w for AVALON

ErichMoraga:
`@JRyan#4759` 2020 Toyota Avalon Limited  DongleID/route 619c87373e917a0c|2021-05-04--17-08-14


ErichMoraga:
Note: This is affected by...
https://github.com/commaai/openpilot/issues/19739

adeebshihadeh:
We're actually missing a test route for the Avalon - would be great if we could get one in with this PR. Only requirements are it has to be engaged for at least a segment and user is okay with making it public.

ErichMoraga:
> We're actually missing a test route for the Avalon - would be great if we could get one in with this PR. Only requirements are it has to be engaged for at least a segment and user is okay with making it public.

Sure, no problem.  I'll get on top of that shortly.

nelsonjchen:
The 2019+ XX50 vehicles are different from the XX40 vehicles. It might be a good idea to make these fingerprint differently.

nelsonjchen:
Helped @JRyan#4759 on Discord today. Followed #20864

nelsonjchen:
Helped `JRyan#4759` on Discord today. We installed a [hacked fork of master-ci](https://github.com/nelsonjchen/openpilot/commit/04052020ceaf2cc7427e6d321e49f0914ab08e6c) that used a different DBC with an 8 byte EPS value. This allowed `JRyan#4759` to activate openpilot and see openpilot working whereas he wasn't able to before.

619c87373e917a0c|2021-05-09--16-05-44

---

On that note, I want to reemphasize my support for making an additional `CAR.AVALON_2019` and `CAR.AVALONH_2019` model in Toyota instead of overloading `CAR.AVALON` with the 2019+ Avalons. 

fpv1+fpv2 for a future additional model split: https://gist.github.com/nelsonjchen/e42999408d40bfe40d21b7910cde0c83

PR #20837, opened 2021-05-06
Lexus IS 200T FW versions

johnlu6518:
Lexus IS 200T 2017 in Taiwan..
Engine fw for lexus IS

Device: comma two
Dongle ID: ec30903bb15db3ad


johnlu6518:
Thank you!

PR #20847, opened 2021-05-07
match nvg sidebar

adeebshihadeh:


selfdrive/ui/qt/sidebar.cc
```diff
@@ -3,185 +3,110 @@
...
-    Sidebar {
-      background-color: #393939;
-    }
-    * {
-      color: white;
-    }
-  )");
-  setLayout(layout);
-}
+  temp_val = (int)s.scene.deviceState.getAmbientTempC();
```
adeebshihadeh, line 75: @deanlee can we fix up the C++ submaster so we always have a valid reader like the python version? then we can remove all these extra readers in UI.

deanlee: good idea

deanlee: https://github.com/commaai/cereal/pull/139

PR #20849, opened 2021-05-07
Cleanup util.h

deanlee:
1. Reduce header file size and dependencies
2. fix the issue that `std:atomic<bool>` in ExitHandler conflict with `android_system_core/include/cutils/atomic.h` on EON if we include util.h after ISurfaceComposer
``
In file included from phonelibs/android_frameworks_native/include/gui/ISurfaceComposer.h:23:
In file included from phonelibs/android_system_core/include/utils/RefBase.h:20:
In file included from phonelibs/android_system_core/include/cutils/atomic.h:22:
In file included from /system/comma/usr/lib/clang/8.0.0/include/stdatomic.h:31:
/data/data/com.termux/files/usr/include/stdatomic.h:238:42: error: reference to 'memory_order' is ambiguous
static __inline void atomic_thread_fence(memory_order __order __attribute__((unused))) {
```

### DIFF
--- a/selfdrive/common/util.cc
+++ b/selfdrive/common/util.cc
@@ -152,7 +152,7 @@ std::string dir_name(std::string const &path) {
 
 // class ExitHandler
 
-struct ExitHandleHelper {
+struct ExitHandlerHelper {
   static void set_do_exit(int sig) {
 #ifndef __APPLE__
     power_failure = (sig == SIGPWR);
@@ -164,22 +164,22 @@ struct ExitHandleHelper {
 };
 
 ExitHandler::ExitHandler() {
-  std::signal(SIGINT, (sighandler_t)ExitHandleHelper::set_do_exit);
-  std::signal(SIGTERM, (sighandler_t)ExitHandleHelper::set_do_exit);
+  std::signal(SIGINT, (sighandler_t)ExitHandlerHelper::set_do_exit);
+  std::signal(SIGTERM, (sighandler_t)ExitHandlerHelper::set_do_exit);
 #ifndef __APPLE__
-  std::signal(SIGPWR, (sighandler_t)ExitHandleHelper::set_do_exit);
+  std::signal(SIGPWR, (sighandler_t)ExitHandlerHelper::set_do_exit);
 #endif
 }
 
 ExitHandler::operator bool() { 
-  return ExitHandleHelper::do_exit; 
+  return ExitHandlerHelper::do_exit; 
 }
 
 ExitHandler& ExitHandler::operator=(bool v) {
-  ExitHandleHelper::do_exit = v;
+  ExitHandlerHelper::do_exit = v;
   return *this;
 }
 
 bool ExitHandler::powerFailure() {
-  return ExitHandleHelper::power_failure;
+  return ExitHandlerHelper::power_failure;
 }
--- a/selfdrive/common/util.cc
+++ b/selfdrive/common/util.cc
@@ -2,6 +2,11 @@
 
 #include <errno.h>
 
+#include <atomic>
+#include <cassert>
+#include <csignal>
+#include <cstring>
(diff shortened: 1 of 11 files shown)

### CODE REVIEW

File: selfdrive/modeld/runners/onnxmodel.cc [hunk 1]
```diff
@@ -14,6 +14,12 @@
 #include "selfdrive/common/swaglog.h"
 #include "selfdrive/common/util.h"
 
+static std::string dir_name(std::string const &path) {
```
adeebshihadeh, line 17: why move all these helpers outside util?

deanlee: moved back

selfdrive/common/util.h
```diff
@@ -31,12 +24,14 @@ namespace util {
 // ***** math helpers *****
 
 // map x from [a1, a2] to [b1, b2]
-template<typename T>
+template <typename T>
 T map_val(T x, T a1, T a2, T b1, T b2) {
-  x = std::clamp(x, a1, a2);
+  if (x < a1) x = a1;
+  if (x > a2) x = a2;
```
adeebshihadeh, line 30: why change this? std::clamp is more readable

selfdrive/common/util.cc
```diff
@@ -97,5 +102,84 @@ int write_file(const char* path, const void* data, size_t size, int flags, mode_
...
+  if (pos == std::string::npos) return "";
+  return path.substr(0, pos);
+}
 
 }  // namespace util
+
+
+// class ExitHandler
+
+struct ExitHandlerHelper {
```
adeebshihadeh, line 155: why make this helper struct?

selfdrive/common/util.cc
```diff
@@ -97,5 +102,85 @@ int write_file(const char* path, const void* data, size_t size, int flags, mode_
...
+  if (pos == std::string::npos) return "";
+  return path.substr(0, pos);
+}
 
 }  // namespace util
+
+
+// class ExitHandler
+
+class ExitHandlerHelper {
```
adeebshihadeh, line 155: why not just put this in the ExitHandler class?

deanlee: Putting it in the header file will conflict with android_system_core/include/cutils/atomic.h

adeebshihadeh: hmm, can we just wrap the include in a namespace?

deanlee: It may not be simple

pd0wm: Do we really want to go through all this effort to "hide" the std::atomic<bool>? I'd rather force the include order, since the problem will just come back if we just std::atomic<bool> somewhere else.

deanlee: yes that would be better

selfdrive/locationd/ubloxd.cc
```diff
@@ -1,3 +1,5 @@
+#include <assert.h>
```
pd0wm, line 1:
```suggestion
#include <cassert>
```

selfdrive/locationd/ublox_msg.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <assert.h>
```
pd0wm, line 3:
```suggestion
#include <cassert>
```

pd0wm: Or is there a specific reason to use the c header?

selfdrive/loggerd/logger.h
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <assert.h>
```
pd0wm, line 3:
```suggestion
#include <cassert>
```

PR #20852, opened 2021-05-07
Refactor hardware abstraction layers 

deanlee:
1. Convert static classes to singleton.
2. Add screen size, camera sizes, and intrinsic matrices to hardware abstraction layers.

### DIFF
--- a/selfdrive/hardware/base.h
+++ b/selfdrive/hardware/base.h
@@ -37,7 +37,7 @@ public:
   inline mat3 driver_cam_intrinsic_matrix() {
     return get_matrix(driver_cam_focal_len, driver_cam_size);
   }
-  inline mat3 wide_cam_intrinsic_matrix() {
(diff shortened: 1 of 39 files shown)


selfdrive/hardware/base.h
```diff
@@ -1,25 +1,50 @@
...
 
 // no-op base hw class
-class HardwareNone {
+class HardwareBase {
 public:
-  static constexpr float MAX_VOLUME = 0;
-  static constexpr float MIN_VOLUME = 0;
+  float MAX_VOLUME = 0;
+  float MIN_VOLUME = 0;
+  int road_cam_focal_len = 910;
```
adeebshihadeh, line 13: think it’s better for this to live in camerad

pd0wm: I think for now it's fine to be here. Ideally it lives in camerad, but it is then communicated over visionipc when you get the frames, in that case the subscriber doesn't have to know which hardware it's running on.

PR #20855, opened 2021-05-08
UI: remove duplicate capnp readers

deanlee:
readers in `SubMaster` is always valid after https://github.com/commaai/cereal/pull/139


deanlee:
@adeebshihadeh : Is there any special reason to update sm in a separate function?
https://github.com/commaai/openpilot/blob/9c60d8a7cb95adf81b3abcd839ae7427a7d2ac4f/selfdrive/ui/ui.cc#L123-L126

deanlee:
it's better not to be static

selfdrive/ui/ui.h
```diff
@@ -107,50 +102,52 @@ typedef struct UIScene {
   uint64_t started_frame;
 } UIScene;
 
-typedef struct UIState {
-  VisionIpcClient * vipc_client;
-  VisionIpcClient * vipc_client_front;
-  VisionIpcClient * vipc_client_rear;
-  VisionBuf * last_frame;
+class UIState {
```
adeebshihadeh, line 105: what's the benefit of making this a class? let's just do the minimal refactor to remove the other readers

deanlee: the benefit is that we can use the syntax of ` s->sm["radarState"] `to get the reader, otherwise we can only use `(*(s->sm))["radarState"] `. (or use A temporary variable, such as `Submaster &sm = *(s->sm)`)

adeebshihadeh:
> @adeebshihadeh : Is there any special reason to update sm in a separate function?
> https://github.com/commaai/openpilot/blob/9c60d8a7cb95adf81b3abcd839ae7427a7d2ac4f/selfdrive/ui/ui.cc#L123-L126

yes, it's to support replay without sockets

selfdrive/ui/paint.cc
```diff
@@ -213,15 +214,15 @@ static void ui_draw_vision_maxspeed(UIState *s) {
...
+  const float speed = std::max(0.0, (*s->sm)["carState"].getCarState().getVEgo() * (s->scene.is_metric ? 3.6 : 2.2369363));
   const std::string speed_str = std::to_string((int)std::nearbyint(speed));
   nvgTextAlign(s->vg, NVG_ALIGN_CENTER | NVG_ALIGN_BASELINE);
   ui_draw_text(s, s->viz_rect.centerX(), 240, speed_str.c_str(), 96 * 2.5, COLOR_WHITE, "sans-bold");
   ui_draw_text(s, s->viz_rect.centerX(), 320, s->scene.is_metric ? "km/h" : "mph", 36 * 2.5, COLOR_WHITE_ALPHA(200), "sans-regular");
 }
 
 static void ui_draw_vision_event(UIState *s) {
-  if (s->scene.controls_state.getEngageable()) {
+  if ((*s->sm)["controlsState"].getControlsState().getEngageable()) {
```
adeebshihadeh, line 225: would be nice if this could become `sm["controlsState"].getEngageable()` like the python version, should be the last remaining big difference between the two

PR #20863, opened 2021-05-09
CI: Valgrind assert leaks

ghost:
Now valgrind test will fail if it detects a leak. Also added improvements from https://github.com/commaai/openpilot/pull/20828

### DIFF
--- a/selfdrive/test/test_valgrind_replay.py
+++ b/selfdrive/test/test_valgrind_replay.py
@@ -83,7 +83,9 @@ class TestValgrind(unittest.TestCase):
     thread.daemon = True
     thread.start()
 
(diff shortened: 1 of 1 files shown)


pd0wm:
If we don't wait for a response we still need to wait till the message is received, otherwise it's not deterministic. We should probably find a good way to re-use the infrastructure from process replay instead of writing it again.

adeebshihadeh:
Instead of keeping this in a draft, let's just revert the other changes and keep the assert.

grekiki:
Yeah sounds good, doing more work on my car port right now. I'd still want to replace the sleep with the check though, should speed up the test by a percent or so.

PR #20866, opened 2021-05-10
Subaru false positive brake pressed when ACC is braking

ClockeNessMnstr:
Single messages for user braking are appearing on Subaru vehicles causing unsafe ACC cancelation.
The vehicle cancels ACC and "coast" into/towards a vehicle that should be being braked for.
It seems to happen with moderate braking when a new lead is acquired.
It is also appearing more frequently with gas pedal enabled.

My testing is only on subaru_global but change preserves the previous logic to ensure compatibility with all vehicles.
When vehicle is in motion this checks for 2 consecutive brake messages before exiting controls.
This prevents ACC from being canceled by itself if OP or stock ACC is actively braking. I do not know if this would be an issue with long vehicles but I imagine it would be similar behavior.

Panda PR https://github.com/commaai/panda/pull/645

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -124,11 +124,14 @@ class CarInterfaceBase():
       else:
         events.add(EventName.steerTempUnavailable)
 
-    # Disable on rising edge of gas or brake. Also disable on brake when speed > 0.
+    # Disable on rising edge of gas.
+    # Disable on rising edge of brake if vehicle is speed = 0
+    # Debounce rising edge of brake AND Disable on brake while speed > 0.
     # Optionally allow to press gas at zero speed to resume.
     # e.g. Chrysler does not spam the resume button yet, so resuming with gas is handy. FIXME!
     if (cs_out.gasPressed and (not self.CS.out.gasPressed) and cs_out.vEgo > gas_resume_speed) or \
-       (cs_out.brakePressed and (not self.CS.out.brakePressed or not cs_out.standstill)):
+       (cs_out.brakePressed and not self.CS.out.brakePressed and cs_out.standstill) or \
+       (cs_out.brakePressed and self.CS.out.brakePressed and not cs_out.standstill):
       events.add(EventName.pedalPressed)
 
     # we engage when pcm is active (rising edge)


pd0wm:
Do you have an example route where this happens?

ClockeNessMnstr:
> Do you have an example route where this happens?

I still need to run stock around traffic and get a good capture of it happening again with lead car switching where it's the most dangerous feeling.
My first logs of it expired a while ago.

Currently I'm testing with "unsafe_gas" as I can get the same brake "blips" frequently just following a car and using the accelerator a few times to let ACC do some braking.

Depends what is wanted in the route but I'm assuming we'd like to see the issue on a clean install?

pd0wm:
Preferable a clean install, but a route with unsafe_gas will do for now.

ClockeNessMnstr:
This is from a test build with panda working and OP's "pedalPressed" disabled in interfaces.py. (already reverted interfaces.py)

https://my.comma.ai/cabana/?route=a3586311666ae480%7C2021-05-13--16-39-33&exp=1652475871&sig=%2FLlJ9t7Jcl%2Fz2n4mPO7CEOLk043dDtdJgnEyZ0Ghm2k%3D&max=12&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2Fa3586311666ae480%2F68d179979fa00a4a0b3209e7eb1f184f_2021-05-13--16-39-33

616 - 618 seconds. Hard braking resulted in consecutive frames of brake pedal = 16
(16 is the minimum value on my car based on the dbc)

362 - 371 seconds. about 17 messages are debounced in Panda. 

The update loops are checking values from the same CAN message for the break multiple times. 
This is why the debounce isn't working in the current commit. 

In Panda for every CAN message, the generic_rx_checks are checking brake / gas status. needed to add a boolean input for gas and brake to only update on a new message. This is working already but has little to no effect as OP is still checking the single messages twice and defeating the debounce.

ClockeNessMnstr:
If I'm not mistaken, controlsd.py is calling "CS = self.CI.update(self.CC, can_strs)" at 100 hz meaning that a 50hz brake message update is actually read 2, possibly 3, times depending on sync. 

This means that a single message cannot be debounced without making a flag for whether or not the brake or gas was updated. I don't believe this exists and it looks like a bit of a PITA. Another solution would be to get the "brake_pressed" and "brake_pressed_prev" read from Panda but doing the CAN reads twice (100hz in OP, onchange in panda) and verifying controls match seems to be an integral part of the design intentional or not. 

The issue being seen on Subaru's sometimes has a consecutive erroneous message on the bus as well as not being limited to a LSB only error. As such, neither debouncing, thresholding, or the combination of the two guarantees safe operation. 

As noted before, MartinL has been testing using a different brake status signal. That appears to be the correct fix for the vehicle specific issue. 

If other vehicles see a similar issue then perhaps debouncing is still a good idea. but it looks like it requires a bit of a refactor.

pd0wm:
Why did you close this PR? The canParser has timestamps of when the message was update. See how we have some debounce logic in honda: https://github.com/commaai/openpilot/blob/master/selfdrive/car/honda/carstate.py#L327

ClockeNessMnstr:
Ah, excellent! I'll take another look at this.

ClockeNessMnstr:
As a car bug fix, a different signal will be used on subaru global. MartinL was already working to make that change. 

I was looking originally to see if debouncing the brake signal would be useful as it is an issue that could be hiding on more vehicles.

For Honda, I don't see the corresponding debounce in panda, but I'm guessing it's there somehow to prevent controls mismatch?

pd0wm:
Yeah, I know it's not in panda. Somewhere on my list is [figuring out the origins](https://wiki.lesswrong.com/wiki/Chesterton%27s_Fence) of the debounce logic and probably removing it.

ClockeNessMnstr:
This can be closed still I think. 

Brake signal update is coming from MartinL's branch for Subaru which will fix this. there is a "Brake_Status.Brake" signal that has a filtered 0/1 value for the user brake press and does not have the noise seen in the brake pedal message. 

It also looks like it needs to be implemented per vehicle line's "carstate.py" anyway in order to use the correct ().ts property of the brake message for each vehicle line.

PR #20877, opened 2021-05-11
Update README.md to include 21 Subaru Forester as supported car

tgun:


VirtuallyChris:
Can you give us your device dongle ID so we can confirm it works with your car?

tgun:
Sure! a30eb761d1b1a9cb

PR #20883, opened 2021-05-11
Add new firmware versions for LEXUS_ESH_TSS2

rpaulhanna:


pd0wm:
2021 model I assume based on VIN?

rpaulhanna:
> 2021 model I assume based on VIN?

Yes. 2021 on paper, but made last Nov I think.

PR #20892, opened 2021-05-13
VW MQB: Add FW values for 2018 Volkswagen Golf  

dingyf12593:
Just new FW values.

Dongle ID: [3f84146622df5f39]


jyoung8607:
If you would be so kind, can you add them in alphanumeric sorted order rather than tacking them on the end of the lists? Makes it a lot easier to spot already-existing entries when adding new cars in the future. Thanks!

PR #20916, opened 2021-05-15
2021 Hyundai Elantra port

CruiseBrantley:
Fingerprint for 2021 Hyundai Elantra
Test Route: 82e9cdd3f43bf83e|2021-05-15--02-42-51

Minor issues left to resolve: 
- ~~Check Lane Follow Assist Warning~~ -Fixed
- Sometimes on brake message appears: "Smart Cruise Control Conditions Not Met"


selfdrive/car/hyundai/values.py
```diff
@@ -577,10 +581,18 @@ class Buttons:
     (Ecu.fwdCamera, 0x7c4, None): [b'\xf1\x00JFA LKAS AT USA LHD 1.00 1.02 95895-D5000 h31'],
     (Ecu.transmission, 0x7e1, None): [b'\xf1\x816U2V8051\x00\x00\xf1\x006U2V0_C2\x00\x006U2V8051\x00\x00DJF0T16NL0\t\xd2GW'],
   },
+  CAR.ELANTRA_21: {
```
adeebshihadeh, line 584: add to the IGNORED_FINGERPRINTS list

adeebshihadeh:
Are you sure this requires a new car? Why can't the existing ELANTRA be used?

selfdrive/car/hyundai/values.py
CruiseBrantley: Added.

CruiseBrantley:
> Are you sure this requires a new car? Why can't the existing ELANTRA be used?

Model year is "All new" which also includes a new harness (A2 I believe) I used Sonata as a reference as another A harness vehicle, I assumed the features being different would be problematic.

selfdrive/car/hyundai/values.py
```diff
@@ -21,6 +21,7 @@ def __init__(self, CP):
 class CAR:
   # Hyundai
   ELANTRA = "HYUNDAI ELANTRA 2017"
+  ELANTRA_21 = "HYUNDAI ELANTRA 2021"
```
pd0wm, line 24:
```suggestion
  ELANTRA_2021 = "HYUNDAI ELANTRA 2021"
```

Let's stay consistent with the Ionic naming.

CruiseBrantley: Got it, fixed.

adeebshihadeh:
Thanks for the PR!

PR #20919, opened 2021-05-15
UI: add decline page for the Terms and Conditions

maxime-desroches:
Add a decline page as per this TODO:

https://github.com/commaai/openpilot/blob/533bc30c2e4ecac5e022f2ba4f3919842dfa0a1a/selfdrive/ui/qt/offroad/onboarding.cc#L80

It looks like this:

decline

Let me know if something is wrong/bad with this


pd0wm:
Looks good!

Discussed internally, and we'd like the call to action to be "Decline, uninstall openpilot" instead of power off. Just set the param `DoUninstall` and everything should happen automatically. See: https://github.com/commaai/openpilot/blob/master/selfdrive/ui/qt/offroad/settings.cc#L159

PR #879, opened 2019-11-14
Fix grinding with Honda pedal + Civic specific pedal tuning

VirtuallyChris:
This pull request solves 3 issues relating to the Honda Pedal.

### 1. Pedal fix discovered by @runchman
This change improves brake pump logic for Honda vehicles allowing for reduced stress on the brake pump. It also fixes the grinding noise found in many cars such as the Odyssey and Civic while using a Honda pedal during low speed braking. (under 15mph)

[Here's an example of the sound that is now fixed in my Civic.](https://www.youtube.com/watch?v=CrcZB59LmAc)

### 2. Gas breakpoints
The Toyota interface.py has code that greatly improves the pedal feel and greatly reduces the ability for the pedal to jerk the car unexpectedly at low speeds. This wasn't present in the Honda file, I've added this.

### 3. Civic specific tuning
The Civic needed pedal tuning more than any car due to the very high kpV compared to all other Honda cars. The stock values would make the car constantly floor the gas and slam the brakes until around ~20mph. The new values reduced that issue greatly.

I still do believe pedal tuning will still need some work in the future. It never seems to be as smooth as lets say a Toyota with full stop and go controlled by openpilot. I believe it may need the decoupling of gas and braking into two PI loops. Trying to tune the pedal to accel slower from 0 causes the brakes to be affected and for the car to not stop in time.
Another potential solution would be to allow tuning over the accel_max at low speeds. It seems no matter how low I make the PI loop, the pedal still finds a way to quickly rise to gasMaxV.

With that being said, this pull request turns the Civic pedal experience from being awful and dangerous to slightly jerky at low speeds but still a decent experience. I think for now this is the best we're going to get.

-VC 😀

PR #20922, opened 2021-05-16
Honda - Don't send cancel cmd when using comma pedal

sshane:
@VirtuallyChris Care to test this? Hopefully it works as expected and is upstreamable!

I originally thought to change the `enableCruise` field in controlsd, and it wouldn't change behavior for cars other than Honda (since GM never has a pedal), but it was more confusing.

This shouldn't change the code where it spams the cancel button if we don't have longitudinal control and we want to cancel, since the user wouldn't have a pedal, and `enableCruise` would be True, and `pcm_cancel_cmd` would always be equal to itself. Let me know if I'm wrong in this assumption though.

### DIFF
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -101,6 +101,10 @@ class CarController():
       # send pcm acc cancel cmd if drive is disabled but pcm is still on, or if the system can't be activated
       pcm_cancel_cmd = True
 
+    # Never send cancel command if we never enter cruise state (pedal)
+    # Cancel cmd causes brakes to release at a standstill
+    pcm_cancel_cmd &= CS.CP.enableCruise
+
     # *** rate limit after the enable check ***
     self.brake_last = rate_limit(brake, self.brake_last, -2., DT_CTRL)
 
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -103,7 +103,7 @@ class CarController():
 
     # Never send cancel command if we never enter cruise state (no cruise if pedal)
     # Cancel cmd causes brakes to release at a standstill causing grinding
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
What's the goal of this change?

ShaneSmiskol:
To eliminate the brake grinding at low speeds, Chris said that if openpilot controls long (has pedal, so `enableCruise` is false) it constantly sends a cancel request to the car, causing it to release brakes. https://github.com/commaai/openpilot/pull/879

VirtuallyChris:
Will test tonight!

ShaneSmiskol:
If it's possible, can you also unplug the pedal and use stock Honda ACC to verify it doesn't change behavior for non-openpilot long controlled Hondas? Specifically that it cancels stock ACC when openpilot disengages

selfdrive/car/honda/carcontroller.py
```diff
@@ -101,6 +101,10 @@ def update(self, enabled, CS, frame, actuators,
       # send pcm acc cancel cmd if drive is disabled but pcm is still on, or if the system can't be activated
       pcm_cancel_cmd = True
 
+    # Never send cancel command if we never enter cruise state (no cruise if pedal)
+    # Cancel cmd causes brakes to release at a standstill causing grinding
+    pcm_cancel_cmd &= CS.CP.enableCruise
```
pd0wm, line 106:
```suggestion
    pcm_cancel_cmd = pcm_cancel_cmd and CS.CP.enableCruise
```

Probably best to not use bitwise and to prevent weird issues later.

VirtuallyChris:
Did some test drives tonight - behavior is exactly the same as before for both non pedal ACC and pedal ACC, but no brake grinding anymore! This will make all the pedal users _very_ happy! For a reminder, here's what it sounds like before this PR https://youtu.be/CrcZB59LmAc

PR #20929, opened 2021-05-17
Sidebar: repaint only when properties updated

deanlee:


adeebshihadeh:
Does this actually save any CPU?  When the sidebar isn't visible, it should only update the state without redrawing, which should be cheap.

deanlee:
Just want to make it consistent with the update logic of other widgets (update only when visible), free is always better than cheap.

deanlee:
If you think the added complexity is not worth it, I will remove the `isVisible` part

deanlee:
If you think the added complexity is not worth it, I will remove the isVisible part

adeebshihadeh:
I don't think this PR is worth merging as is, but I like the idea. There's probably a nice way to setup the properties in Qt, so it'll automatically repaint when certain properties are changed. Let's also remove the visible check.

PR #20933, opened 2021-05-17
New model outputs

mitchellgoffpc:
Updated driving.cc to work with new model output sizes
- [x] Log new meta outputs in cereal
- [x] Get rid of x-indexed plans
- [x] Interpolate plan_t_arr from t-indexed plans
- [x] Add new FCW alert based on the new meta outputs


selfdrive/controls/lib/events.py
```diff
@@ -479,6 +479,14 @@ def startup_fuzzy_fingerprint_alert(CP: car.CarParams, sm: messaging.SubMaster,
       Priority.LOW, VisualAlert.steerRequired, AudibleAlert.chimePrompt, 1., 1., 1.),
   },
 
+  EventName.modelFcw: {
+    ET.WARNING: Alert(
+      "BRAKE!",
+      "Model FCW: Risk of Collision",
```
adeebshihadeh, line 485:
```suggestion
      "Risk of Collision",
```

adeebshihadeh: actually, I would just use the existing fcw event

selfdrive/controls/lib/events.py
```diff
@@ -479,6 +479,14 @@ def startup_fuzzy_fingerprint_alert(CP: car.CarParams, sm: messaging.SubMaster,
       Priority.LOW, VisualAlert.steerRequired, AudibleAlert.chimePrompt, 1., 1., 1.),
   },
 
+  EventName.modelFcw: {
```
adeebshihadeh, line 482: why not use the existing fcw event?

PR #20934, opened 2021-05-17
Subaru: show LDW on dash for correct lane

trailtacos:
Wire up the left and right lane departures to the correct
Subaru messages


pd0wm:
Why do you need to create new alerts for this? Toyota already does this using the values from hudcontrol:

See:
 - https://github.com/commaai/openpilot/blob/master/selfdrive/controls/controlsd.py#L483-L484
 - https://github.com/commaai/openpilot/blob/master/selfdrive/car/toyota/interface.py#L386-L390
 - https://github.com/commaai/openpilot/blob/master/selfdrive/car/toyota/carcontroller.py#L143

CrispinFlowerday:
On the subaru, we can only show one alert on the dash to the user at a time (keep hands on wheel, left lane departure, right lane departure), It seemed to me that the Events model inside OP was the best place to determine what should be shown as opposed to trying to guess in the car implementation.

E.g. what should the car show if a lane is being departed, but OP has some other higher priority alert to display (e.g. FCW).  By using the VisualAlert value we leave OP fully in control of what is shown to the user

adeebshihadeh:
There's no guessing needed in the car controller. If you get a `VisualAlert.ldw` in the car controller, use CC.hudControl to determine the lane.

CrispinFlowerday:
Ahh right, that will work, thanks - note that I will still need to alter EventName.ldw to emit the VisualAlert.ldw and adjust all car implementations to understand VisualAlert.ldw (to do the same as VisualAlert.steerRequired) - will that work for you?

adeebshihadeh:
That should be fine.

adeebshihadeh:
Any update on this? Otherwise, will close due to inactiviy.

CrispinFlowerday:
Yeah, sorry - it took a while to fully test the changes on the road.  I've updated the patch now.

selfdrive/car/chrysler/chryslercan.py
```diff
@@ -8,7 +8,7 @@
 def create_lkas_hud(packer, gear, lkas_active, hud_alert, hud_count, lkas_car_model):
   # LKAS_HUD 0x2a6 (678) Controls what lane-keeping icon is displayed.
 
-  if hud_alert == VisualAlert.steerRequired:
+  if hud_alert in [ VisualAlert.steerRequired, VisualAlert.ldw ]:
```
adeebshihadeh, line 11: remove the extra whitespace

selfdrive/car/subaru/subarucan.py
```diff
@@ -27,12 +27,19 @@ def create_es_distance(packer, es_distance_msg, pcm_cancel_cmd):
...
   return packer.make_can_msg("ES_Distance", 0, values)
 
-def create_es_lkas(packer, es_lkas_msg, visual_alert, left_line, right_line):
+def create_es_lkas(packer, es_lkas_msg, visual_alert, left_line, right_line, left_lane_depart, right_lane_depart):
 
   values = copy.copy(es_lkas_msg)
   if visual_alert == VisualAlert.steerRequired:
     values["Keep_Hands_On_Wheel"] = 1
 
+  if visual_alert == VisualAlert.ldw and values["LKAS_Alert"] == 0:
```
adeebshihadeh, line 36: why checks LKAS_ALERT?

CrispinFlowerday: the `LKAS_Alert` field is used by stock to display all dash warnings including collision warnings.  I feel it is best not to over-write with the left/right lane departure if the stock collision warning is wanting to warn the driver about an impending collision.

selfdrive/car/chrysler/chryslercan.py
CrispinFlowerday: Oh damn, sorry, will do.

PR #20940, opened 2021-05-18
Filter only initialized if up for 1sec

haraschax:


pd0wm:
Should we add another state? Something like INITIALIZING? That makes it easier to check downstream if we want.

HaraldSchafer:
I think it's fine like this. Can just look at the timeSinceReset field if you want to know exactly what is happening.

Will probably need to rethink the state when we add magnetometer. We could make it only fully trusted when moving or have good magnetometer reading, and then subscribers can decide if they want to use the values when those conditions are not met.

PR #20942, opened 2021-05-18
Fix Terms and Training guide skip when the screen turns off

### LINKED ISSUE #20926: Terms and training guide close when screen turns off
When the screen turns off the terms or training guide are closed and you are brought to the home screen. Thermald does correctly prevents startup (but no indication why).

maxime-desroches:
Fixes this issue #20926

**Description:**
After ~30s of doing nothing, the Terms  and Training guide were closed/skipped and replaced by the home screen.

The closeSettings function in window.cc was setting the current widget to the home screen without checking for anything.  
Also fix the same problem when you want to review the training guide in the settings.

Let me know if something is wrong/bad with this


pd0wm:
LGTM!

PR #20949, opened 2021-05-18
Also shutdown after one hour if no ignition seen

pd0wm:
Previously disabling charging would turn off the C2, this is no longer the case. So a device that hasn't been onroad was able to drain a battery.


ChadBailey:
I'm just a random stranger creeping on pull requests, but I do wonder if some users may be depending on this behavior? I could see an argument to be made that they want to use the internal hotspot and keep it active for a dashcam in parking mode.

I may be way off base here, just thinking aloud - perhaps what I'm suggesting isn't even possible much less relied upon.

pd0wm:
All the other conditions for amount of power used still need to be met, so it won't just shut down after an hour.

PR #20951, opened 2021-05-18
Toyota: fix brakeLights while engaged

sshane:
And it should include a few other TSS1 cars as well. BRAKE_LIGHTS_ACC is never 1 while engaged and car is braking, but `BRAKE_PRESSED` in `BRAKE_MODULE2` works. It doesn't always line up perfectly, sometimes being a split second delayed, but works.

### DIFF
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -28,7 +28,10 @@ class CarState(CarStateBase):
     ret.seatbeltUnlatched = cp.vl["SEATS_DOORS"]['SEATBELT_DRIVER_UNLATCHED'] != 0
 
     ret.brakePressed = cp.vl["BRAKE_MODULE"]['BRAKE_PRESSED'] != 0  # only user
-    ret.brakeLights = bool(cp.vl["ESP_CONTROL"]['BRAKE_LIGHTS_ACC'] or cp.vl["BRAKE_MODULE2"]['BRAKE_PRESSED'] != 0 or ret.brakePressed)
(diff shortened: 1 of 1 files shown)


selfdrive/car/toyota/carstate.py
```diff
@@ -27,8 +27,11 @@ def update(self, cp, cp_cam):
                         cp.vl["SEATS_DOORS"]['DOOR_OPEN_RL'], cp.vl["SEATS_DOORS"]['DOOR_OPEN_RR']])
     ret.seatbeltUnlatched = cp.vl["SEATS_DOORS"]['SEATBELT_DRIVER_UNLATCHED'] != 0
 
-    ret.brakePressed = cp.vl["BRAKE_MODULE"]['BRAKE_PRESSED'] != 0
+    ret.brakePressed = cp.vl["BRAKE_MODULE"]['BRAKE_PRESSED'] != 0  # only user
     ret.brakeLights = bool(cp.vl["ESP_CONTROL"]['BRAKE_LIGHTS_ACC'] or ret.brakePressed)
+    if self.CP.carFingerprint not in TSS2_CAR:
+      ret.brakeLights = bool(ret.brakeLights or cp.vl["BRAKE_MODULE2"]['BRAKE_PRESSED'] != 0)
```
adeebshihadeh, line 33: let's look for a message that's universal across toyotas

adeebshihadeh: distinction seems to be whether it's a hybrid, not TSS2

PR #20953, opened 2021-05-18
Add missing several missing Camry firmwares

ErichMoraga:
`@oxigen#0649` 2019 Toyota Camry ICE  DongleID/route 04b2f59611cf9e7f|2021-05-18--11-11-17


pd0wm:
Sounds like this list needs to be updated too: https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py#L212

PR #20959, opened 2021-05-19
Remove non-existing include in setup.cc

maxime-desroches:
Remove a non-existing + not used include


adeebshihadeh:
Fixed in efca6eaa702dedb08a6bfa7c0f562e434ac6e8c8

PR #20960, opened 2021-05-19
Move MCLK from 24 -> 19.2 mhz + Add accuracy in sidebar

geohot:
All in title.

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -169,6 +169,9 @@ static void update_state(UIState *s) {
       scene.satelliteCount = data.getMeasurementReport().getNumMeas();
     }
   }
+  if (sm.updated("gpsLocationExternal")) {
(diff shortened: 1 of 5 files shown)


selfdrive/ui/ui.cc
```diff
@@ -169,6 +169,9 @@ static void update_state(UIState *s) {
       scene.satelliteCount = data.getMeasurementReport().getNumMeas();
     }
   }
+  if (sm.updated("gpsLocationExternal")) {
+    scene.gpsAccuracy = sm["gpsLocationExternal"].getGpsLocationExternal().getAccuracy();
```
adeebshihadeh, line 173: not necessary anymore, you can just directly use submaster now

geohot: I don't understand this comment, what can I remove?

geohot: Oh, you mean from the other thread? Still, I like this abstraction

PR #20980, opened 2021-05-20
better scons cache handling

adeebshihadeh:
- [x] remove cache on scons --clean
- [x] minimize cache size

### DIFF
--- a/selfdrive/manager/build.py
+++ b/selfdrive/manager/build.py
@@ -26,7 +26,7 @@ def build(spinner, dirty=False):
   j_flag = "" if nproc is None else f"-j{nproc - 1}"
 
   for retry in [True, False]:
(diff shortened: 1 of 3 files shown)


selfdrive/manager/build.py
```diff
@@ -80,6 +80,9 @@ def build(spinner, dirty=False):
           t.wait_for_exit()
         exit(1)
     else:
+      shutil.rmtree("/tmp/scons_cache", ignore_errors=True)
+      shutil.rmtree("/data/scons_cache", ignore_errors=True)
+      os.system(f"cd {BASEDIR} && scons {j_flag} --cache-populate")
```
pd0wm, line 85: This will add quite a bit of time to the build process. Sure we want to do this?

adeebshihadeh: changed it to enforce a max size instead of just keeping the last build products. new implementation is pretty fast

PR #20985, opened 2021-05-21
stricter memory usage check

adeebshihadeh:
Testing closet devices use 48-50%.


pd0wm:
Do we have any leaks right now? Would this cause any issues on long drives?

adeebshihadeh:
There's leaks, but it wouldn't cause issues on normal drives. The 48-50% numbers came from devices after 22 days of uptime.

PR #20996, opened 2021-05-22
C3: affine camera IRQs to camera core

adeebshihadeh:


selfdrive/camerad/test/setup_cam_core.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# camerad gets core 6
```
pd0wm, line 3: Should we make camerad fail to start with an error message if the core is not online? Probably gets into a very weird state if we don’t.

adeebshihadeh: if the core isn't online, the state shouldn't be too weird, though think we should add the assert anyway. the affinity stuff will fail, but camerad will still get its realtime priority.

PR #20478, opened 2021-03-26
VW MQB: Integration support, bugfixes, minor cleanup

jyoung8607:
**Bug fixes and integration support from the VW Community Port**

- Feature: auto-detect whether we're hooked up at the LKAS camera or CAN gateway, move messages and signals to match
- Bugfix: pass-through lane-departure warning info from the LKAS camera to the blind-spot warning radars
- Bugfix: garbage-collect some signals from ACC_06 that we haven't used for a long time
- Style: camelCase -> PEP8

**Verification**

This is a refactored version of what we've been running in community since the very beginning. The refactor has been tested successfully against several community cars before opening this PR.

**Merge Review**

The LDW bugfix will change LDW_02 messages as seen by process_replay. The change is correct and desired.

**Route**

cae14e88932eb364|2021-03-25--16-23-20 (still uploading)

PR #20997, opened 2021-05-23
VW MQB: Cleanup, prep for J533 gateway integration

jyoung8607:

Preparatory refactor for gateway integration: rearrange the ACC and BSM radar signal and message lists, such that we can add them to either (or neither) bus as circumstances dictate.

Remove unused signal `TSK_Fahrzeugmasse_02` and its associated check `Motor_16`.

Touch up some other signal and message comments.

### DIFF
--- a/selfdrive/car/volkswagen/carstate.py
+++ b/selfdrive/car/volkswagen/carstate.py
@@ -229,12 +229,12 @@ class CarState(CarStateBase):
       checks += [("Motor_14", 10)]  # From J623 Engine control module
 
     # TODO: Detect ACC radar bus location
-    signals += MqbExtraSignals.fwd_radar[0]
-    checks += MqbExtraSignals.fwd_radar[1]
+    signals += MqbExtraSignals.fwd_radar_signals
+    checks += MqbExtraSignals.fwd_radar_checks
     # TODO: Detect BSM radar bus location
     if CP.enableBsm:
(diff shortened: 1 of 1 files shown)


jyoung8607:
The goal here is to implement network-location detection from #20478. I'm breaking it into two pieces for ease of review. This first one is nil-effect. The next PR will bring in the `NetworkLocation` switch and move the radars to the `ext_cp` alias.

selfdrive/car/volkswagen/carstate.py
```diff
@@ -267,3 +256,23 @@ def get_cam_can_parser(CP):
     ]
 
     return CANParser(DBC[CP.carFingerprint]['pt'], signals, checks, CANBUS.cam)
+
+class MqbExtraSignals:
+  # Additional signal and message lists for optional or bus-portable controllers
+  fwd_radar = ([
+    ("ACC_Wunschgeschw", "ACC_02", 0),              # ACC set speed
```
adeebshihadeh, line 263: the tuple is a little confusing. let's call them `xxx_signals` and `xxx_checks` or just put them back in the function broken out from the main lists

jyoung8607: I took them out of the regular function because they can appear on either of two (and in the future, three) CAN buses, and I don't want to carry and maintain multiple copies of the relevant signal and check lists. I'm onboard with clarifying how they're referenced though.

PR #20999, opened 2021-05-23
VW MQB: Add manumatic to allowed openpilot gears

jyoung8607:
VW stock ACC allows use of Tiptronic (manumatic shifting) gears. It doesn't make a ton of sense to do so, but it's allowed. I get a fair number of complaints that openpilot doesn't allow it.

It's not because people want to drive that way, but because it's easy to accidentally bump a steering-wheel paddle-shifter. Ordinarily that would just bump you up-or-down a gear for a few seconds, and then automatically revert to Drive/Sport/Eco. However, if you do so while openpilot is engaged, it throws a spectacular audiovisual tantrum to warn you of impending doom.

Go ahead and permit manumatic as an openpilot gear for VW MQB. This might get revisited (perhaps a softer alert?) when I implement openpilot long control on MQB.

### DIFF
--- a/selfdrive/car/volkswagen/interface.py
+++ b/selfdrive/car/volkswagen/interface.py
@@ -152,7 +152,7 @@ class CarInterface(CarInterfaceBase):
         be.pressed = self.CS.buttonStates[button]
         buttonEvents.append(be)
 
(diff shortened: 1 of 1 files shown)


pd0wm:
Makes sense while we're still using stock longitudinal!

PR #21000, opened 2021-05-23
LogReader: fix leaks

deanlee:
1. fix leaks in LogReader
2.  do not merge events, seek in segments.(faster and thread-safe)
3. pub messages directly from buffer, without  deep copy from MessageReader to MessageBuilder.


deanlee:
merged to https://github.com/commaai/openpilot/pull/21004

PR #21001, opened 2021-05-24
VW MQB: Network location detection, J533 gateway integration

jyoung8607:

At startup, look for certain messages on CAN 1. If yes, we're wired-in at the gateway. If no, we're wired-in at the camera. Knowing this location determines:

* Which CAN bus the forward and blind spot radar signals are collected from
* Which CAN bus we target with resume button spamming

Holding this in draft, because I need commaai/cereal#156 or something like it to finish the job. Most people integrating at the gateway are doing it because they don't have a stock camera, and I need to make all the inbound LDW_02 stuff conditional.


jyoung8607:
Closing; will resubmit in a new PR.

PR #21005, opened 2021-05-24
INDI use limited torque of CC.apply_last_steer

hewers:
**Description** 

Start with `selfdrive/controls/lib/latcontrol_indi.py` :
INDI for Toyota uses locally-applied torque and torque rate limits, by importing Toyota specific code. This is useful for any car. Adding CarInterface to all the lateral controller's update() signatures gives access to CarController's apply_last_steer and params.STEER_MAX, used to scale back to [-1,1] output. So, the exact torque-limited steer output is available to the INDI controller regardless of control frequency or unique torque-limit methods.

This was useful to me because GM steer control is 50 Hz, and it's tricky to apply the same rate limits at 100 Hz after rounding and all.

To use CS.out and pass unit tests, we add CS.out to Mock CarInterface.update() and CarState, and clean up Chrysler and GM to match style of all other makes.

**Verification**
* Process replay for Toyota Prius (INDI) proves no change to actuation - right?
* In use for Volt INDI tuning.

### DIFF
--- a/selfdrive/car/chrysler/interface.py
+++ b/selfdrive/car/chrysler/interface.py
@@ -50,7 +50,7 @@ class CarInterface(CarInterfaceBase):
     ret.tireStiffnessFront, ret.tireStiffnessRear = scale_tire_stiffness(ret.mass, ret.wheelbase, ret.centerToFront)
 
     ret.enableCamera = True
-    
+
     ret.enableBsm = 720 in fingerprint[0]
 
     return ret
@@ -77,9 +77,7 @@ class CarInterface(CarInterfaceBase):
 
     ret.events = events.to_msg()
 
-    # copy back carState packet to CS
     self.CS.out = ret.as_reader()
-
     return self.CS.out
 
   # pass in a car.CarControl
--- a/selfdrive/car/gm/interface.py
+++ b/selfdrive/car/gm/interface.py
@@ -171,9 +171,7 @@ class CarInterface(CarInterfaceBase):
 
     ret.events = events.to_msg()
 
-    # copy back carState packet to CS
     self.CS.out = ret.as_reader()
-
     return self.CS.out
 
   def apply(self, c):
--- a/selfdrive/car/mock/carstate.py
+++ b/selfdrive/car/mock/carstate.py
@@ -1,0 +1,7 @@
+from selfdrive.car.interfaces import CarStateBase
+
+class CarState(CarStateBase):
+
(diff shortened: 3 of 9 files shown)


selfdrive/car/chrysler/interface.py
```diff
@@ -77,9 +77,7 @@ def update(self, c, can_strings):
 
     ret.events = events.to_msg()
 
-    # copy back carState packet to CS
     self.CS.out = ret.as_reader()
-
     return self.CS.out
```
qadmus, line 81: Match style of all other makes.

selfdrive/car/gm/interface.py
```diff
@@ -171,9 +171,7 @@ def update(self, c, can_strings):
 
     ret.events = events.to_msg()
 
-    # copy back carState packet to CS
     self.CS.out = ret.as_reader()
-
     return self.CS.out
```
qadmus, line 175: Match style of all other makes.

selfdrive/car/mock/interface.py
```diff
@@ -84,7 +84,8 @@ def update(self, c, can_strings):
     curvature = self.yaw_rate / max(self.speed, 1.)
     ret.steeringAngleDeg = curvature * self.CP.steerRatio * self.CP.wheelbase * CV.RAD_TO_DEG
 
-    return ret.as_reader()
+    self.CS.out = ret.as_reader()
+    return self.CS.out
```
qadmus, line 88: Mock CS.out was not implemented, added here.

selfdrive/car/mock/carstate.py
```diff
@@ -0,0 +1,7 @@
+from selfdrive.car.interfaces import CarStateBase
+
+class CarState(CarStateBase):
+
+  @staticmethod
+  def get_can_parser(CP):
+    return None
```
qadmus: Minimal mock?

qadmus:
No pressure, redirect me and/or add work to do. Just thought this was worth upstream practice..

pd0wm:
Not sure if this is the cleanest solution. The output from a single iteration can still exceed the rate limit like this.

How about adding a function to car interface that deals with the rate limit and the clip? (just make it return the input on cars not using this). Then we can call that from both latcontrol_indi and use if from the carcontroller itself.

qadmus:
@pd0wm sounds very good! I'm going to close and reopen when ready.

PR #21012, opened 2021-05-24
Subaru: update brake pressed signal

martinl:
**Description** Forester 2019-21 and some other newer models (2020+ Crosstrek, Outback, Legacy) produce intermittent Brake_Pedal Brake_Pedal > 1 signals on can0 while stock ACC is engaged and braking, causing openpilot to disengage before stopping behind lead car (https://github.com/martinl/openpilot/issues/27 and https://github.com/commaai/openpilot/pull/20866). We found that using alternative Brake_Status Brake signal for brakePressed fixes the issue since it is only triggered when brake pedal is pressed.

**Verification** We have verified the new signal with Crosstrek 2018, Forester 2019, Crosstrek 2020 and upcoming Outback 2020 - openpilot disengages on brake press and there have been no reported issues with openpilot disengaging while ACC is braking behind lead car after the signal change.

Prerequisite: https://github.com/commaai/panda/pull/650


adeebshihadeh:
Can you rebase?

adeebshihadeh:
Need to verify on the following cars:
- [ ] ASCENT
- [ ] IMPREZA
- [ ] FORESTER

adeebshihadeh:
Can we use Brake_Pedal Brake_Pedal to populate the carState.brake field for the global cars? We don't set this for any of the Subarus right now.

adeebshihadeh:
Can we use Brake_Pedal Brake_Pedal to populate the carState.brake field for the global cars? We don't set this for any of the Subarus right now.

PR #21013, opened 2021-05-24
VW MQB: Škoda Octavia Mk3

jyoung8607:
**Škoda Octavia Mk3**

Covers all variants of the third generation Octavia (2013-2019).

- [x] added to README
- [x] test route added to [test_routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_models.py)
- [x] route with openpilot: 66e5edc3a16459c5|2021-05-25--19-00-29

Thanks to community Octavia owners cben (2015) and some friend of Saber422's (2019).

### DIFF
--- a/selfdrive/test/test_routes.py
+++ b/selfdrive/test/test_routes.py
@@ -442,6 +442,10 @@ routes: dict = {
     'carFingerprint': VOLKSWAGEN.SKODA_KODIAQ_MK1,
     'enableCamera': True,
   },
+  "66e5edc3a16459c5|2021-05-25--19-00-29": {
+    'carFingerprint': VOLKSWAGEN.SKODA_OCTAVIA_MK3,
+    'enableCamera': True,
+  },
   "026b6d18fba6417f|2021-03-26--09-17-04": {
     'carFingerprint': VOLKSWAGEN.SKODA_SCALA_MK1,
     'enableCamera': True,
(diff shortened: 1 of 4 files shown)


selfdrive/test/test_routes.py
```diff
@@ -442,6 +442,10 @@
     'carFingerprint': VOLKSWAGEN.SKODA_KODIAQ_MK1,
     'enableCamera': True,
   },
+  "66e5edc3a16459c5|2021-05-23--23-44-46": {
```
adeebshihadeh, line 445: need to look into why CI passed, but this route isn't uploaded

adeebshihadeh: make sure the user has the upload raw toggle on

jyoung8607: I noticed that. Not only that, but it passed on an earlier commit where I forgot to merge the route in entirely, which should have failed immediately.

jyoung8607:
Kicking back to draft till I can confirm we have a viable route.

jyoung8607:
I withdrew the test route. At the time of the drive, the device was running a fork that doesn't disengage on brake. Hopefully someone else with an Octavia and modern hardware will make themselves known. There are a bunch of Octavia folks, unfortunately they mostly have older hardware and/or are located in Germany.

@adeebshihadeh: CI should be failing because Octavia isn't in `non_tested_cars`, and it's still passing. I haven't tried to track down why just yet. Might look later tonight or tomorrow if you aren't already aware of the cause.

jyoung8607:
Ready for review per our DMs.

PR #21016, opened 2021-05-24
Add LEXUS_ESH_TSS2 to improved longitudinal tune

rpaulhanna:
2021 Lexus ES 300h.  Tested improved longitudinal tune parameters on master-ci for better following, stopping and resuming behind lead vehicles.

Route:  3a5c5101bd71ad5d|2021-05-24--15-05-04

Examples around timestamps 15:06:45 and 15:14:00


adeebshihadeh:
Our poilicy on tuning PRs is that a couple of users have to confirm that the change is better.

rpaulhanna:
Understood.  Thanks.

adeebshihadeh:
I think one other user is fine in this case since it's an existing tune.

adeebshihadeh:
Merging, as I only see a handful of active Lexus ES Hybrids over the last month.

rpaulhanna:
Thanks, Adeeb.

rpaulhanna:
Looks like the change dropped out of the latest release 0.8.5 when the "improved longitudinal" section was re-organized.  Switched back to my custom fork to re-implement the improved long.  If it's possible to merge this again, I would prefer to return to stock.  Thanks.

PR #21017, opened 2021-05-24
Refactor joystick script

### LINKED ISSUE #20984: Fix up joystick and tools/carcontrols
The joystick code has been broken for a while. Let's rewrite it, so it's robust to refactors.

I'm thinking we keep joystickd, which a joystick packet. If controlsd sees the joystick on startup, it goes into a debug mode and takes the command from the joystick packet. A mid-size permanent alert should also be shown with the title "DEBUG MODE" and perhaps the current command on the second line.

sshane:
Planned changes:
- [x] Enabling keyboard support for gas and brakes and steering torque.
- [x] Remove debug_controls.py and replace with auto detection in controlsd to go into a debug/joystick mode if `testJoystick` is detected to be alive on start up, with an alert.
- [x] onroad QT UI now updates alerts when alertText2 changes

Todos:
- [x] Clean up Joystick abstracting class (specifically the keyboard part)
- [x] Add joystick support without pygame
  - [x] Axis support
  - [x] Button support (convert joystick buttons to key buttons to combine logic)
    - [x] Add reset button for joystick
- [x] Add buttons back
  - [x] CC cancel
  - [x] steerRequired hud alert
  - [x] Toggle engaged
- [ ] Fix param not loading properly
- [x] Deadzone never lets us get to 100%. Just remove?
- [ ] Add `inputs` so Actions can pass
  - [x] Alternatively for now, force keyboard if we can't import package

joystick now operates similarly to unlogger, since keyboard receives are most easily done blockingly, we spin up a thread at 10hz (customizable) that the main loop sends the axes and buttons to.


selfdrive/common/params.cc
```diff
@@ -216,6 +216,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"Offroad_UnofficialHardware", CLEAR_ON_MANAGER_START},
     {"Offroad_NvmeMissing", CLEAR_ON_MANAGER_START},
     {"ForcePowerDown", CLEAR_ON_MANAGER_START},
+    {"JoystickDebugMode", CLEAR_ON_IGNITION},
```
adeebshihadeh, line 219: add CLEAR_ON_MANAGER_START too

selfdrive/car/car_helpers.py
```diff
@@ -13,7 +13,7 @@
 EventName = car.CarEvent.EventName
 
 
-def get_startup_event(car_recognized, controller_available, fuzzy_fingerprint):
+def get_startup_event(car_recognized, controller_available, fuzzy_fingerprint, debug_mode):
```
adeebshihadeh, line 16: don't do this in the startup event

PR #21024, opened 2021-05-25
Compute on-road screen brightness based on camera exposure

### LINKED ISSUE #20705: Use exposure for screen brightness
Sometimes, a bad calibrated light sensor can get in the field. It would be much better if we use the camera exposure while onroad instead of the light sensor. Likely would fix things like https://github.com/commaai/openpilot/issues/20419 since the light sensor values vary from device to device.

pd0wm:

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -177,8 +177,9 @@ static void update_state(UIState *s) {
   }
   if (sm.updated("sensorEvents")) {
     for (auto sensor : sm["sensorEvents"].getSensorEvents()) {
-      if (!Hardware::TICI() && sensor.which() == cereal::SensorEventData::LIGHT) {
-        scene.light_sensor = sensor.getLight();
+      // Use light sensor when offroad, camera exposure when onroad
+      if (!scene.started && Hardware::EON() && sensor.which() == cereal::SensorEventData::LIGHT) {
+        scene.light_sensor = std::clamp<float>(sensor.getLight(), 0.0, 1023.0);
       }
       if (!scene.started && sensor.which() == cereal::SensorEventData::ACCELERATION) {
         auto accel = sensor.getAcceleration().getV();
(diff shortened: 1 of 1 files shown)


selfdrive/ui/ui.cc
```diff
@@ -177,8 +177,9 @@ static void update_state(UIState *s) {
   }
   if (sm.updated("sensorEvents")) {
     for (auto sensor : sm["sensorEvents"].getSensorEvents()) {
-      if (!Hardware::TICI() && sensor.which() == cereal::SensorEventData::LIGHT) {
-        scene.light_sensor = sensor.getLight();
+      // Use light sensor when offroad, camera exposure when onroad
+      if (!scene.started && Hardware::EON() && sensor.which() == cereal::SensorEventData::LIGHT) {
+        scene.light_sensor = std::clamp<float>(10.0 * sensor.getLight(), 0.0, 1023.0);
```
pd0wm, line 182: On older releases the screen brightness was max at a sensor value of `light_sensor = 255 / 2.6 = 98`. Currently we reach max brightness at `light_sensor = 100 / 0.1 = 1000`. By multiplying by 10, the old behavior is preserved.

selfdrive/ui/ui.cc
```diff
@@ -319,7 +322,7 @@ void QUIState::update() {
 
 Device::Device(QObject *parent) : brightness_filter(BACKLIGHT_OFFROAD, BACKLIGHT_TS, BACKLIGHT_DT), QObject(parent) {
   brightness_b = Params(true).get<float>("BRIGHTNESS_B").value_or(10.0);
-  brightness_m = Params(true).get<float>("BRIGHTNESS_M").value_or(0.1);
+  brightness_m = Params(true).get<float>("BRIGHTNESS_M").value_or(2.6) / 26.0;
```
pd0wm, line 325: Old default was 2.6, and it was written to disk when the param was not present. We're stuk with this value.

selfdrive/ui/ui.cc
```diff
@@ -193,10 +194,15 @@ static void update_state(UIState *s) {
       }
     }
   }
-  if (Hardware::TICI() && sm.updated("roadCameraState")) {
+  if (sm.updated("roadCameraState")) {
     auto camera_state = sm["roadCameraState"].getRoadCameraState();
-    float gain = camera_state.getGainFrac() * (camera_state.getGlobalGain() > 100 ? 2.5 : 1.0) / 10.0;
-    scene.light_sensor = std::clamp<float>((1023.0 / 1757.0) * (1757.0 - camera_state.getIntegLines()) * (1.0 - gain), 0.0, 1023.0);
+    if (Hardware::EON()) {
```
adeebshihadeh, line 199: should we check the inverse of the condition for the light sensor? you'll get weird behavior here with driver view for example

adeebshihadeh: or better yet use different variables and handle this logic in the brightness code

selfdrive/ui/ui.cc
```diff
@@ -193,10 +194,15 @@ static void update_state(UIState *s) {
...
+      float gain = camera_state.getGainFrac();
+      scene.light_sensor = std::clamp<float>((1023.0 / 5408.0) * (5408.0 - camera_state.getIntegLines()) * (1.0 - gain), 0.0, 1023.0);
```
ZwX1616, line 201: This formula (max_t - t) * (max_gain - gain) feels a bit weird?

ZwX1616: shouldn't it be
(max_t * max_gain) - (t * gain)
?

ZwX1616: I think of it as the proportion of EV

VirtuallyChris:
Is there any reason to keep using the light sensor while offroad? I've seen the consistency of the reading vary greatly depending on the device and lighting conditions. Can we just lock brightness to max while offroad?

selfdrive/ui/ui.cc
pd0wm: But if the gain goes up, wouldn't that influence the meaning of the max time as well? The current output seems to be pretty linear when I go through the whole range of exposure.

ffhspa:
On my C2 at Night Screen its very bright, that disturb my Driving. Screen brightness should be lovered at Night

VirtuallyChris (approved):

PR #21028, opened 2021-05-25
deleter.py: delete boot and crash folders last

### LINKED ISSUE #21022: Deleter behavior not consistent for boot and crashlogs
The deleter looks at creation date of the folder. This doesn't work for the bootlogs and crashlogs, since the folder is created once but grows over time. When the deleter decides to wipe it, it removes the whole folder. These two folders need to be excluded from the folder logic, and needs be handled on a file-by-file basis by deleter.

pd0wm:
Fixes #21022 

Boot and crash should never be deleted. If they get really large we will delete them as a last resort.

### DIFF
--- a/selfdrive/loggerd/deleter.py
+++ b/selfdrive/loggerd/deleter.py
@@ -9,6 +9,8 @@ from selfdrive.loggerd.uploader import listdir_by_creation
 MIN_BYTES = 5 * 1024 * 1024 * 1024
 MIN_PERCENT = 10
 
+DELETE_LAST = ['boot', 'crash']
+
 
 def deleter_thread(exit_event):
   while not exit_event.is_set():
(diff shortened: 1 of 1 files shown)


gregjhogan:
maybe we should exclude these folder from the normal delete logic and add something new to keep a max of 100 files in each folder (seems like now they could grow until there is no space left for drives)

gregjhogan:
also seems like tests were never updated when boot logs were moved and crash logs were added (tests still use `bootlog.bz2` and no refs to anything named crash)

pd0wm:
Bootlogs are tested on the loggerd side, but not on the uploader side: https://github.com/commaai/openpilot/blob/master/selfdrive/loggerd/tests/test_loggerd.py#L133

My reasoning was that under normal behavior the bootlog folder will never become large enough to be relevant, the deleting is more as a last resort in case of another bug. But since both you and Adeeb like something on a file-by-file basis I'll add some helpers for that. We can reuse that to maintain a maximum scons cache size.

adeebshihadeh:
It would be nice if we could have something like a list of paths to watch with a particular policy (file or directory size).

adeebshihadeh:
Merging this, we could add the per-file stuff later.

PR #21031, opened 2021-05-25
Improve focus when using snapshot

### LINKED ISSUE #1669: Poor focus in snapshots from athena
**Example snapshots**

**Still from driving video**
Screenshot 2020-06-08 23 35 33

**Describe the bug**
Focus appears to be off significantly in rear camera snapshots

Capture snapshot via athena.

**Expected behavior**
Snapshot with good focus.

 - Device: comma two
 - Dongle ID: 95e637a323d6e9fc
 - Version: master-ci branch based off of 9d8ff2887c280388eb0abc589d5f562a6dfac0ad
 - Car make/model: Civic Hatchback 2017

**Additional context**
Focus during normal driving footage is fine and looks great. Only the snapshots appear to be affected. I query snapshots constantly throughout the day and it always exhibits the same behavior. I thought maybe it was compression at play, but I'm starting to thing that it's not the issue.

sshane:
Now waits for at least 10 of 12 (~0.83%) regions of interest to be under blurry focus threshold. `camera_qcom.cc`'s `setup_self_recover` function recovers if under 92% focus.

C2 reaches 10 of 12 regions almost instantly indoors, unless in weird situations like if pointed at a close subject, then the ROI code gives camerad a little longer to focus, up to 10 seconds.

Old (left) vs. new (right):

Screenshot from 2021-05-26 14-45-35

Keep in mind that when given more time, camerad's resulting frames are much more underexposed compared to previously, so the new (right) image was exposure compensated so you can see similar amounts of detail.

This was the actual image:
back_new 6


take_snapshot.py
```diff
@@ -0,0 +1,27 @@
+# DO NOT PUSH!
+import io
```
ShaneSmiskol, line 2: this is temporary

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -40,8 +50,19 @@ def get_snapshots(frame="roadCameraState", front_frame="driverCameraState"):
     sockets.append(front_frame)
 
   sm = messaging.SubMaster(sockets)
-  while min(sm.logMonoTime.values()) == 0:
+  t = sec_since_boot()
+  # wait_for_exposure =
+  i = 0
+  while sec_since_boot() - t < 10:
     sm.update()
+    i += 1
```
adeebshihadeh, line 58: use sm.frame

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -103,6 +126,7 @@ def snapshot():
   if pic is not None:
     print(pic.shape)
     jpeg_write("/tmp/back.jpg", pic)
-    jpeg_write("/tmp/front.jpg", fpic)
+    if fpic is not None:  # can be none if not front_camera_allowed
```
adeebshihadeh, line 129: remove the comment

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -80,11 +101,13 @@ def snapshot():
 
   proc = subprocess.Popen(os.path.join(BASEDIR, "selfdrive/camerad/camerad"),
                           cwd=os.path.join(BASEDIR, "selfdrive/camerad"), env=env)
-  time.sleep(3.0)
+  time.sleep(4.0)
```
adeebshihadeh, line 104: remove the sleep

ShaneSmiskol: It's overexposed for the first 2-4 seconds, and the focus code reports high focus during this time.

adeebshihadeh: handle this in get_snapshots

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -40,8 +48,13 @@ def get_snapshots(frame="roadCameraState", front_frame="driverCameraState"):
     sockets.append(front_frame)
 
   sm = messaging.SubMaster(sockets)
-  while min(sm.logMonoTime.values()) == 0:
+  time.sleep(4.0)  # wait for startup and AF
+  t = time.time()
```
adeebshihadeh, line 52:
```suggestion
  t = time.monotonic()
```

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -63,31 +76,25 @@ def snapshot():
...
   if front_camera_allowed:
-    env["SEND_DRIVER"] = "1"
-
-  proc = subprocess.Popen(os.path.join(BASEDIR, "selfdrive/camerad/camerad"),
-                          cwd=os.path.join(BASEDIR, "selfdrive/camerad"), env=env)
-  time.sleep(3.0)
+    os.environ["SEND_DRIVER"] = "1"
 
+  managed_processes['camerad'].start()
   frame = "wideRoadCameraState" if TICI else "roadCameraState"
```
adeebshihadeh, line 93: do all the tici conditions in one place

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -40,8 +48,13 @@ def get_snapshots(frame="roadCameraState", front_frame="driverCameraState"):
     sockets.append(front_frame)
 
   sm = messaging.SubMaster(sockets)
-  while min(sm.logMonoTime.values()) == 0:
+  time.sleep(4.0)  # wait for startup and AF
+  t = time.monotonic()
+  while time.monotonic() - t < 10:
     sm.update()
+    if min(sm.logMonoTime.values()):
+      if rois_in_focus(sm[frame].sharpnessScore) >= focus_perc_threshold:
```
adeebshihadeh, line 56: combine these

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -40,8 +48,13 @@ def get_snapshots(frame="roadCameraState", front_frame="driverCameraState"):
     sockets.append(front_frame)
 
   sm = messaging.SubMaster(sockets)
-  while min(sm.logMonoTime.values()) == 0:
+  time.sleep(4.0)  # wait for startup and AF
+  t = time.monotonic()
+  while time.monotonic() - t < 10:
     sm.update()
+    if min(sm.logMonoTime.values()):
```
adeebshihadeh, line 55:
```suggestion
    if min(sm.rcv_frames.values()) > 1:
```

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -40,8 +48,13 @@ def get_snapshots(frame="roadCameraState", front_frame="driverCameraState"):
     sockets.append(front_frame)
 
   sm = messaging.SubMaster(sockets)
-  while min(sm.logMonoTime.values()) == 0:
+  time.sleep(4.0)  # wait for startup and AF
```
adeebshihadeh, line 51: put a threshold on the frame id

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -1,18 +1,20 @@
...
 import cereal.messaging as messaging
-from common.basedir import BASEDIR
 from common.params import Params
+from typing import List
 from common.transformations.camera import eon_f_frame_size, eon_d_frame_size, leon_d_frame_size, tici_f_frame_size
 from selfdrive.hardware import TICI
 from selfdrive.controls.lib.alertmanager import set_offroad_alert
+from selfdrive.manager.process_config import managed_processes
+
+LM_THRESH = 120
```
adeebshihadeh, line 17: add a comment with the path of the header this came from

selfdrive/camerad/snapshot/snapshot.py
```diff
@@ -40,8 +48,13 @@ def get_snapshots(frame="roadCameraState", front_frame="driverCameraState"):
     sockets.append(front_frame)
 
   sm = messaging.SubMaster(sockets)
-  while min(sm.logMonoTime.values()) == 0:
+
+  start_t = time.monotonic()
+  while time.monotonic() - start_t < 10:
     sm.update()
+    # wait 4 sec for startup and AF
+    if sm.rcv_frame[frame] >= 80 and rois_in_focus(sm[frame].sharpnessScore) >= focus_perc_threshold:
```
adeebshihadeh, line 56:
```suggestion
    if sm.rcv_frame[frame] * DT_MDL > 4. and rois_in_focus(sm[frame].sharpnessScore) >= focus_perc_threshold:
```

ShaneSmiskol:
@adeebshihadeh I think the camera exposure part is failing, the resulting images are much darker compared to just waiting 3 seconds, the previous behavior

ShaneSmiskol:
Should be working now! I can combine the loops, I just thought that since the first one only runs the first time you start camerad that it should be separated with a clear comment.

PR #21035, opened 2021-05-26
adding pqtimebomb bypass

dkiiv:
So I couldnt figure out how to update the dang cereal submodule
Just add this:
`	
    pqTimebombWarn @99;
    pqTimebombBypassing @100;
    pqTimebombBypassed @101;
    pqTimebombTERMINAL @102;`
into the "enum EventName @0xbaa8c5d505f727de {" list

NOTE: This code does NOT differenciate between a PQ and an MQB vehicle!! So you will get the grab wheel message on both PQ and MQB as this code sits!

What this code does:
It starts a counter once you engage OP
Once this counter reaches its time limit it prompts for the driver to grab the wheel, once the driver grabs the wheel it shuts **off** HCA for 1.05 seconds. If the driver does NOT grab the wheel in time it will sound an alarm "TERMINAL" and do the bypass automagically. This is to prevent OP from hitting the timebomb rendering OP steering helpless and useless until you manually disengage and re-engage

It is worth noting, if you do hit the timebomb you must wait 2-3 seconds before sending HCA commands again. However not sending HCA commands for 1.05 seconds any time BEFORE the timebomb resets the counter.

The value for the timebomb on the steering rack is 360, 6 minutes.

Here is the car.capnp file to go along with this commit, as again I could not noodle out the submodule
[car.capnp.zip](https://github.com/commaai/openpilot/files/6543411/car.capnp.zip)

### DIFF
--- a/selfdrive/car/volkswagen/carcontroller.py
+++ b/selfdrive/car/volkswagen/carcontroller.py
@@ -56,6 +56,10 @@ class CarController():
       # commanding HCA if there's a fault, so the steering rack recovers.
       if enabled and not (CS.out.standstill or CS.steeringFault):
 
+        #STUFF FOR PQTIMEBOMB BYPASS
+        if CS.out.stopSteering:
+          apply_steer = 0
+
         # FAULT AVOIDANCE: Requested HCA torque must not exceed 3.0 Nm. This
         # is inherently handled by scaling to STEER_MAX. The rack doesn't seem
         # to care about up/down rate, but we have some evidence it may do its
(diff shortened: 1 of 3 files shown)


actuallylemoncurd:
Wow the code portion did not post correctly
image
The highlighted bits are the stuff you need to add to car.capnp in cereal

actuallylemoncurd:
Oops, I forgot the steering pressed Boolean, will have to add that tomorrow

pd0wm:
We don't have any clear plans yet to upstream any of the `pq` stuff yet. Once we do, we will integrate something like this timeout. I'll keep a reference to the PR, but will close it for now.

actuallylemoncurd:
Perfectly understood! I just wanted to get what i have done out there so we are all on the same page. I have a couple other things I want to try first for bypassing the timebomb so we will see how those go

Do you have any plans on bypassing the timebomb? Maybe we can send our idea's back and forth @pd0wm

Issue #20983, opened 2021-05-21
UI: expose updater in settings


adeebshihadeh:
I'm thinking:
* label for last update check (LastUpdateTime param)
* button to trigger update check (send SIGHUP to updated)
* release notes for current release

PR #21036, opened 2021-05-26
UI: add update in settings

maxime-desroches:
Addressing #20983

This currently adds a new Panel in the settings but could be changed if you want.

Because of how updated.py is currently made, sending a SIGHUP will trigger a full fetch and not just a quick update check, hence why it is a Download and not just check for updates.

Currently the timer to check the status of the update (in progress, done, error, unavailable) is set at every 5 secondes but could be changed.

Since updated is not enabled on PC, you must edit process_config.py in order to test/use the Download feature on PC.

This is a design tentative. Let me know if something is wrong/bad with any of this.

main
notes
updates
no_updates
checking

### DIFF
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -204,6 +204,81 @@ DevicePanel::DevicePanel(QWidget* parent) : QWidget(parent) {
   )");
 }
 
+ReleaseNotes::ReleaseNotes(QWidget* parent) : QFrame(parent) {
+  QVBoxLayout *layout = new QVBoxLayout;
+  layout->setMargin(50);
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Thanks for the PR, but this is something we want to do ourselves as it requires new design work. I made a new "internal" tag for issues that we want to do ourselves to avoid confusion in the future. Feel free to work on any of the issues without that tag!

adeebshihadeh:
It's looking like we won't have new designs in time for the next release, so we thought of a nice way to implement most of this without new design if you're interested:
* add last update time to the developer page
* add the release notes as a description to the version label
* send SIGHUP when last update time is clicked

maxime-desroches:
Will do!

PR #21041, opened 2021-05-26
C++ cleanup: standardize phonelibs

whowe444:
Continuing C++ cleanup, standardizing hpp/cpp files to h/cc in phonelibs directory per issue #20799 , only did one file so far


pd0wm:
No need to standardize these since they are third party libraries.

PR #21004, opened 2021-05-24
Refactor c++ replay

deanlee:
- [x]  fix leaks
- [x] lthread-safe
- [x] search encodeIdx in adjacent segments.
- [x]  pub messages directly from buffer, without  deep copy
- [x] send frames for all cameras
- [x] falls back to qlogs + qcameras
- [x] replay from local/remote route
- [x] deal with missing segments or files
- [x] memory usage is about 1GB
- [x] fix segment fault

It has been split into small PRs for easier review:

- Refactor framereader  #21141
- Refactor logreader #21152
- CameraServer #21155

Usage/Test:

1. run `selfdrive/ui/replay/replay dongle_id\|YYYY-MM-DD--HH-MM-SS` to start the replay
Replay will first try to load segments from LOG_ROOT, if it fails, it will be loaded from the server.
2. run `selfdrive/ui/_ui `to watch the replay

test

PR #21042, opened 2021-05-26
Refactor framereader

deanlee:
this is part of https://github.com/commaai/openpilot/pull/21004 

- [x] inherit from QThread
- [x] add error checks for decoding
- [x] fix leaks

### DIFF
--- a/selfdrive/ui/replay/framereader.h
+++ b/selfdrive/ui/replay/framereader.h
@@ -44,7 +44,7 @@ private:
     AVPacket pkt;
     AVFrame *picture;
   };
-  std::vector<Frame*> frames;
+  std::vector<Frame> frames;
 
   AVFormatContext *pFormatCtx = NULL;
   AVCodecContext *pCodecCtx = NULL;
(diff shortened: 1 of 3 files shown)


selfdrive/ui/replay/framereader.h
```diff
@@ -5,49 +5,61 @@
...
 }
 
+class FrameReader : public QThread {
+  Q_OBJECT
 
-class FrameReader {
-public:
-  FrameReader(const std::string &fn);
+ public:
+  FrameReader(const std::string &url, VisionStreamType stream_type);
```
adeebshihadeh, line 26: what's the point of the vision stream type arg? it's unused

deanlee: it will be used to create vipc server buffers in https://github.com/commaai/openpilot/pull/21004

adeebshihadeh: this shouldn’t be done in framereader, see python unclogged and framereader for an example

PR #21043, opened 2021-05-26
Update CI tests for CAN fingerprinting deprecation

jyoung8607:
`test_car_interfaces.py` and `test_models.py` weren't running at all for cars without a legacy CAN fingerprint. Should fix CI passing incorrectly in cases like #21013.


adeebshihadeh:
Thanks for tracking this down!

PR #21045, opened 2021-05-26
athena: version endpoint

gregjhogan:


selfdrive/athena/athenad.py
```diff
@@ -132,6 +132,16 @@ def getMessage(service=None, timeout=1000):
   return ret.to_dict()
 
 
+@dispatcher.add_method
+def getVersion():
+  return {
+    "version": version,
```
pd0wm, line 138: add get_version method for consistency?

gregjhogan: good idea - done

selfdrive/athena/tests/test_athenad.py
```diff
@@ -174,6 +174,13 @@ def test_getSshAuthorizedKeys(self):
     keys = dispatcher["getSshAuthorizedKeys"]()
     self.assertEqual(keys, MockParams().params["GithubSshKeys"].decode('utf-8'))
 
+  def test_getVersion(self):
+    resp = dispatcher["getVersion"]()
+    keys = ["version", "remote", "branch", "commit"]
+    self.assertEqual(list(resp.keys()), keys)
+    for k in keys:
+      self.assertTrue(resp[k], f"{k} has no value")
```
adeebshihadeh, line 182: check type instead?

gregjhogan: well I want an error if it returns None or an empty string (the default if an error happens), so this seems better than just checking if it is a string?

adeebshihadeh: can we check those explicitly? it's a bit unclear from the way it's written
```suggestion
      self.assertTrue(isinstance(resp[k], str) and len(k), f"{k} has no value")
```

gregjhogan: yeah, sounds good!

PR #21049, opened 2021-05-27
Hello openpilot

RicardLee:


adeebshihadeh:
Thanks for the PR, but this isn't something we'd add to the UI.

PR #21052, opened 2021-05-27
Avalon fixup

sshane:
Todo: check if there's any Avalon route that has STEERING_LKA at all. In Cabana, there's very similar number of messages on bus 0 and 2 compared to a Corolla, yet parser.cc never receives any for the Avalon.

Edit: Issue is not in parser.cc. 740 (steering_lka) isn't even in the log reader in test_models. Need to figure out why Cabana has the messages, but not the parsed log. Issue with the specific segment it's grabbing?

### DIFF
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -140,7 +140,7 @@ class CarState(CarStateBase):
       ("STEERING_LEVERS", 0.15),
       ("SEATS_DOORS", 3),
       ("ESP_CONTROL", 3),
-      ("EPS_STATUS", 0),
+      ("EPS_STATUS", 25),
       ("BRAKE_MODULE", 40),
       ("GAS_PEDAL", 33),
       ("WHEEL_SPEEDS", 80),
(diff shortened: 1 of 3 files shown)


selfdrive/car/toyota/carstate.py
```diff
@@ -188,8 +188,8 @@ def get_cam_can_parser(CP):
 
     # use steering message to check if panda is connected to frc
     checks = [
-      ("STEERING_LKA", 42),
+      # ("STEERING_LKA", 42),
```
ShaneSmiskol, line 191: Is this even needed anymore?

selfdrive/car/toyota/carstate.py
```diff
@@ -140,7 +140,7 @@ def get_can_parser(CP):
       ("STEERING_LEVERS", 0.15),
       ("SEATS_DOORS", 3),
       ("ESP_CONTROL", 3),
-      ("EPS_STATUS", 25),
+      ("EPS_STATUS", 0),
```
ShaneSmiskol, line 143: Parser is saying it never got this signal from logs, but it exists on Cabana under bus 0

PR #21058, opened 2021-05-27
Split Avalon 2016-18 and Avalon 2019+

### LINKED ISSUE #19739: Toyota: fix EPS_STATUS message length

sshane:
CAN fingerprint from a 2020 Avalon (https://gist.github.com/nelsonjchen/e42999408d40bfe40d21b7910cde0c83 ignore it saying 2016)

Reverted added fw from the PRs listed in the commits, this is assuming that the fw previous to these commits was strictly 2016-2018. Also need to make sure that having fw in the 2019 Avalon that exists in the older Avalon candidate won't cause any issues.

- [x] Pick known 2016 vs. 2019+ Avalons and test fingerprint them, making sure the Avalons with a longer EPS_STATUS msg fp as AVALON_2019


selfdrive/car/toyota/values.py
```diff
@@ -204,6 +205,9 @@ class CAR:
   CAR.AVALON: [{
     36: 8, 37: 8, 170: 8, 180: 8, 426: 6, 452: 8, 464: 8, 466: 8, 467: 8, 547: 8, 550: 8, 552: 4, 562: 6, 608: 8, 610: 5, 643: 7, 705: 8, 740: 5, 767: 4, 800: 8, 835: 8, 836: 8, 849: 4, 869: 7, 870: 7, 871: 2, 896: 8, 897: 8, 905: 8, 911: 1, 916: 2, 921: 8, 933: 6, 944: 8, 945: 8, 951: 8, 955: 8, 956: 8, 979: 2, 1005: 2, 1014: 8, 1017: 8, 1041: 8, 1042: 8, 1044: 8, 1056: 8, 1059: 1, 1114: 8, 1161: 8, 1162: 8, 1163: 8, 1176: 8, 1177: 8, 1178: 8, 1179: 8, 1180: 8, 1181: 8, 1190: 8, 1191: 8, 1192: 8, 1196: 8, 1200: 8, 1201: 8, 1202: 8, 1203: 8, 1206: 8, 1227: 8, 1235: 8, 1237: 8, 1264: 8, 1279: 8, 1552: 8, 1553: 8, 1554: 8, 1555: 8, 1556: 8, 1557: 8, 1558: 8, 1561: 8, 1562: 8, 1568: 8, 1569: 8, 1570: 8, 1571: 8, 1572: 8, 1584: 8, 1589: 8, 1592: 8, 1593: 8, 1595: 8, 1596: 8, 1597: 8, 1664: 8, 1728: 8, 1779: 8, 1904: 8, 1912: 8, 1990: 8, 1998: 8
   }],
+  CAR.AVALON_2019: [{
```
adeebshihadeh, line 208: remove this, can fingerprints are deprecated

selfdrive/car/toyota/values.py
```diff
@@ -263,6 +267,31 @@ class CAR:
 
 FW_VERSIONS = {
   CAR.AVALON: {
+    (Ecu.esp, 0x7b0, None): [
```
adeebshihadeh, line 270: this is now a subset of the avalon 2019 fw

adeebshihadeh:
Merged the opendbc PR, bump it to opendbc master

selfdrive/car/fw_versions.py
```diff
@@ -209,7 +209,7 @@ def match_fw_to_car_exact(fw_versions_dict):
         continue
 
       # On some Toyota models, the engine can show on two different addresses
-      if ecu_type == Ecu.engine and candidate in [TOYOTA.CAMRY, TOYOTA.COROLLA_TSS2, TOYOTA.CHR, TOYOTA.LEXUS_IS, TOYOTA.AVALON] and found_version is None:
+      if ecu_type == Ecu.engine and candidate in [TOYOTA.CAMRY, TOYOTA.COROLLA_TSS2, TOYOTA.CHR, TOYOTA.LEXUS_IS, TOYOTA.AVALON_2019] and found_version is None:
```
ShaneSmiskol, line 212: @adeebshihadeh Can we remove this? So far I've seen the engine on 7e0 for AVALON and 700 for AVALON_2019, but not both for either. It was probably added when the 19+ years were added

toyboxZ:
and Avalon2019h use different dbc file toyota_nodsu_hybrid_pt_generated

ShaneSmiskol:
@toyboxZ are you sure? The 2021 Avalon hybrid is still TSSP, like the 2020 Avalon. I'm thinking it's the one signal that's different

ShaneSmiskol:
Getting a user with a 2019 Avalon to test this!

toyboxZ:
> @toyboxZ are you sure? The 2021 Avalon hybrid is still TSSP, like the 2020 Avalon. I'm thinking it's the one signal (`EPS_STATUS`) that's different. Are you saying the hybrid has more unmatched signals?

yes，Avalon hybrid can use smartdsu s&g. dbc file toyota_nodsu_hybrid_pt_generated.
also can join the group NO_STOP_TIMER_CAR

ShaneSmiskol:
The only difference between the new Avalon dbc (from the 17 Avalon dbc) is the addition of the BSM signal (which doesn't even seem to work for the hybrid Avalon), the `STEERING_LTA` message (which also isn't present in the Hybrid since it's not TSS2), the brake module (scales are different), and the addition of a few sport/econ messages. So I don't think it's critical to use a different DBC for the hybrid.

PR #21064, opened 2021-05-28
Chrysler Lkas request flag sent even when OP is not enabled

xps-genesis:
Issue seen with Chrysler. Lkas request flag is set to True at all times if the vehicle is above min steer speed. Lkas request flag should be only set when OP is enabled.

Issue noticed in vehicles with speed spoofing for steer down to zero. LKAS fault occurs with min steer speed = 0. since LKAS request is True from vehicle startup.

### DIFF
--- a/selfdrive/car/chrysler/carcontroller.py
+++ b/selfdrive/car/chrysler/carcontroller.py
@@ -61,7 +61,7 @@ class CarController():
         can_sends.append(new_msg)
         self.hud_count += 1
 
-    new_msg = create_lkas_command(self.packer, int(apply_steer), self.gone_fast_yet, frame)
+    new_msg = create_lkas_command(self.packer, int(apply_steer), (self.gone_fast_yet and enalbed), frame)
     can_sends.append(new_msg)
 
     self.ccframe += 1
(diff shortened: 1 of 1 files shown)


selfdrive/car/chrysler/carcontroller.py
```diff
@@ -61,7 +61,7 @@ def update(self, enabled, CS, actuators, pcm_cancel_cmd, hud_alert):
         can_sends.append(new_msg)
         self.hud_count += 1
 
-    new_msg = create_lkas_command(self.packer, int(apply_steer), self.gone_fast_yet, frame)
+    new_msg = create_lkas_command(self.packer, int(apply_steer), (self.gone_fast_yet and enalbed), frame)
```
vanillagorillaa, line 64: enabled is spelled wrong here

adeebshihadeh:
Is this actually LKAS request? The signal is called `LKAS_HIGH_TORQUE`. @adhintz thoughts on this PR?

xps-genesis:
yes it is the lkas request. not sure who and why they named it high torque

j-vanetten:
This change causes an LKAS fault on my 2018 Jeep Grand Cherokee Trailhawk.  

* Without engaging, as soon as I reach 9mph I get an LKAS fault
* If I engage below 9mph everything is fine until I disengage.  At that point I get an LKAS fault

Seems for my vehicle, this flag must be set even when not engaged.

adeebshihadeh:
I'll close for now then since it seems this isn't well understood.

PR #21069, opened 2021-05-28
Update Elantra 2021 FPv2 values

CruiseBrantley:
Previously I was getting only fuzzy matches on Elantra 2021. 

So I took out the x00 transformations I manually made before (which seem to be done automatically anyway) and added additional detected carFw values verbatim.

I have not had a fuzzy detection in my last few tests with these new values!

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -562,16 +562,21 @@ FW_VERSIONS = {
     (Ecu.transmission, 0x7e1, None): [b'\xf1\x816U2V8051\x00\x00\xf1\x006U2V0_C2\x00\x006U2V8051\x00\x00DJF0T16NL0\t\xd2GW'],
   },
   CAR.ELANTRA_2021: {
-    (Ecu.fwdRadar, 0x7d0, None): [b'\xf1\x00CN7_ SCC FHCUP      1.00 1.01 99110-AA000         '],
+    (Ecu.fwdRadar, 0x7d0, None): [b'\xf1\000CN7_ SCC FHCUP      1.00 1.01 99110-AA000         '],
     (Ecu.eps, 0x7d4, None): [
       b'\xf1\x87\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf1\x00CN7 MDPS C 1.00 1.06 \x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00 4CNDC106\xf1\xa01.06',
-      b'\xf1\x8756310AA050\x00\xf1\x00CN7 MDPS C 1.00 1.06 56310AA050\x00 4CNDC106\xf1\xa01.06',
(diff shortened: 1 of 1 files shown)


pd0wm:
No need to replace '\x00` by `\000` in the existing fingerprints. It's the same thing, and we prefer `\x00`

CruiseBrantley:
> No need to replace `\x00` by `\000` in the existing fingerprints. It's the same thing, and we prefer `\x00`

Fixed with just adding new firmware instead.

adeebshihadeh:
What's your dongle ID?

CruiseBrantley:
> What's your dongle ID?

8f0b6ca149481580

selfdrive/car/hyundai/values.py
```diff
@@ -568,9 +568,14 @@ class Buttons:
       b'\xf1\x8756310AA050\x00\xf1\x00CN7 MDPS C 1.00 1.06 56310AA050\x00 4CNDC106\xf1\xa01.06',
     ],
     (Ecu.fwdCamera, 0x7c4, None): [b'\xf1\x00CN7 MFC  AT USA LHD 1.00 1.00 99210-AB000 200819'],
-    (Ecu.esp, 0x7d1, None): [b'\xf1\x8758910-AB800\xf1\x00CN ESC \t 101 \x10\x03 58910-AB800\xf1\xa01.01'],
+    (Ecu.esp, 0x7d1, None): [
+      b'\xf1\x8758910-AB800\xf1\x00CN ESC \t 101 \x10\x03 58910-AB800\xf1\xa01.01',
+      b'\xf1\x8758910-AB800\xf1\x00CN ESC \t 101 \x20\x03 58910-AB800\xf1\xa01.01',
```
pd0wm, line 573: Where did you get this one from? Don't see it in any of your routes.

CruiseBrantley: I took them all from routes on useradmin (some from more than 3 days ago) or the boardd method, I'm not sure how to determine which is actively being used.
I kept finding and adding new values when it would get a fuzzy match, which I haven't seen in a while now.

pd0wm: I wrote a script to go through all your routes as well, but only saw the other two lines you added, no `b'\xf1\x8758910-AB800\xf1\x00CN ESC \t 101 \x20\x03 58910-AB800\xf1\xa01.01'` Can you tell me the exact route in useradmin where you found it?

pd0wm, line 573:
```suggestion

```

selfdrive/car/hyundai/values.py
```diff
@@ -568,9 +568,14 @@ class Buttons:
...
+      b'\xf1\x00CN ESC \t 101 \020\003 58910-AB800',
```
pd0wm, line 574:
```suggestion
      b'\xf1\x00CN ESC \t 101 \x10\x03 58910-AB800',
```

selfdrive/car/hyundai/values.py
CruiseBrantley: Ok, that makes sense. I don't know the exact route but yeah hex x10 = 20 makes a lot of sense on the confusion.

PR #21073, opened 2021-05-29
Update Kia Seltos FPv2 Values

speedking456:
User EAE75 on discord has a newer Seltos with different firmwares, adding the values here.


adeebshihadeh:
What's the dongle ID?

speedking456:
> What's the dongle ID?

dongle ID is 143f8e46ecc22dab

PR #21074, opened 2021-05-29
Add missing Chrysler routes

sshane:


adeebshihadeh:
We need to do something similar here for gasPressed as we do for brakePressed on honda bosch. Let's remove [this list](https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_models.py#L32) and replace it with something like [this](https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_models.py#L181).

selfdrive/test/test_models.py
```diff
@@ -173,8 +173,9 @@ def test_panda_safety_carstate(self):
     failed_checks = {k: v for k, v in checks.items() if v > 25}
 
     # TODO: the panda and openpilot interceptor thresholds should match
-    if "gasPressed" in failed_checks and self.CP.enableGasInterceptor:
-      if failed_checks['gasPressed'] < 150:
+    skip_gas_check = self.car_model in [CHRYSLER.JEEP_CHEROKEE_2019, CHRYSLER.PACIFICA_2018, CHRYSLER.PACIFICA_2018_HYBRID]
```
adeebshihadeh, line 176: just check if it's a chrysler

selfdrive/test/test_models.py
```diff
@@ -173,7 +173,7 @@ def test_panda_safety_carstate(self):
     failed_checks = {k: v for k, v in checks.items() if v > 25}
 
     # TODO: the panda and openpilot interceptor thresholds should match
-    skip_gas_check = self.car_model in [CHRYSLER.JEEP_CHEROKEE_2019, CHRYSLER.PACIFICA_2018, CHRYSLER.PACIFICA_2018_HYBRID]
+    skip_gas_check = self.car_model in self.car_model.startswith(('CHRYSLER', 'JEEP'))
```
adeebshihadeh, line 176: use CP.carName

PR #21080, opened 2021-05-30
Add type hints, small cleanups

cmyui:
Just some low hanging fruit I optimized/improved while looking around the codebase, just simple surface-level changes.

I love what you guys are doing! I'm planning on reading more through the codebase and hopefully submitting more useful targeted PRs in the future, probably starting with improving type hinting coverage.

### DIFF
--- a/selfdrive/hardware/tici/agnos.py
+++ b/selfdrive/hardware/tici/agnos.py
@@ -39,18 +39,27 @@ class StreamingDecompressor:
     self.sha256.update(result)
     return result
 
+SPARSE_CHUNK_FMT = struct.Struct('H2xI4x')
 def unsparsify(f: StreamingDecompressor) -> Generator[bytes, None, None]:
+  # https://source.android.com/devices/bootloader/images#sparse-format
   magic = struct.unpack("I", f.read(4))[0]
   assert(magic == 0xed26ff3a)
 
   # Version
-  major, minor = struct.unpack("HH4x", f.read(8)) # (ignore header sizes)
+  major = struct.unpack("H", f.read(2))[0]
+  minor = struct.unpack("H", f.read(2))[0]
   assert(major == 1 and minor == 0)
 
-  block_sz, num_chunks = struct.unpack("I4xI4x", f.read(16))
+  f.read(2)  # file header size
+  f.read(2)  # chunk header size
+
+  block_sz = struct.unpack("I", f.read(4))[0]
+  f.read(4)  # total blocks
+  num_chunks = struct.unpack("I", f.read(4))[0]
+  f.read(4)  # crc checksum
 
   for _ in range(num_chunks):
-    chunk_type, out_blocks = struct.unpack("H2xI4x", f.read(12))
+    chunk_type, out_blocks = SPARSE_CHUNK_FMT.unpack(f.read(12))
 
     if chunk_type == 0xcac1:  # Raw
       # TODO: yield in smaller chunks. Yielding only block_sz is too slow. Largest observed data chunk is 252 MB.
--- a/selfdrive/hardware/tici/agnos.py
+++ b/selfdrive/hardware/tici/agnos.py
@@ -21,7 +21,7 @@ class StreamingDecompressor:
     self.eof = False
     self.sha256 = hashlib.sha256()
 
-  def read(self, length: int) -> None:
+  def read(self, length: int) -> bytes:
     while len(self.buf) < length:
       self.req.raise_for_status()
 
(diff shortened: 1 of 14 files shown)


selfdrive/hardware/tici/agnos.py
```diff
@@ -38,30 +39,18 @@ def read(self, length):
...
     return result
 
-
-def unsparsify(f):
+def unsparsify(f: StreamingDecompressor) -> Generator[bytes, None, None]:
   magic = struct.unpack("I", f.read(4))[0]
   assert(magic == 0xed26ff3a)
 
   # Version
-  major = struct.unpack("H", f.read(2))[0]
```
pd0wm, line 47: Don't think this makes it more clear. Currently all lines match the spec: https://source.android.com/devices/bootloader/images#sparse-format. Feel free to add comment linking to the spec.

selfdrive/hardware/tici/agnos.py
```diff
@@ -6,12 +6,13 @@
...
 import subprocess
 import os
+from typing import Generator
 
 from common.spinner import Spinner
 
 
 class StreamingDecompressor:
-  def __init__(self, url):
+  def __init__(self, url: str) -> None:
```
adeebshihadeh, line 15: do we really want to annotate __init__?

tools/sim/lib/keyboard_ctrl.py
```diff
@@ -14,55 +14,56 @@
...
+    elif c == 'd':
+      q.put("steer_%f" % -0.15)
+    elif c == 'q':
       exit(0)
 
-def test(q):
-  while 1:
-    print("hello")
-    time.sleep(1.0)
+def test(q: 'Queue') -> NoReturn:
```
adeebshihadeh, line 59: why is Queue a string?

selfdrive/hardware/tici/agnos.py
cmyui: I'm not sure how standard the return type is (I've seen it quite a bit), but yea I would definitely type hint the params, helps like any other function (this example seems pretty trivial since it's just a url but it'd be more important in other situations with many possibly confusing params).
image

tools/sim/lib/keyboard_ctrl.py
cmyui: since it's only imported `if __name__ == '__main__'` i used a [forward reference](https://www.python.org/dev/peps/pep-0484/#forward-references) so we don't have to always import it, the type checkers can figure out that it's a multiprocessing.Queue this way 
image.

you can also use this with TYPE_CHECKING to reduce imports at runtime while still supporting type checkers if it's the only reason you're importing stuff https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING, which i use pretty frequently in my own code.

selfdrive/hardware/tici/agnos.py
cmyui: Ah yea this puts it into perspective more, it was literally 1:1 with the spec lol.

The stuff that runs once doesn't really matter then, but depending on how many times this loop runs though, it might be worth keeping the changes in there and even using [struct.Struct](https://docs.python.org/3/library/struct.html#struct.Struct) since calling `struct.pack` repeatedly recompiles this object in the background each time.. I'll get to work :P

selfdrive/hardware/tici/agnos.py
cmyui: Turns out this return value type hinting is actually a bit of a rabbit hole with [this](https://github.com/python/mypy/issues/604#issuecomment-348525995) being the implemented conclusion.

pd0wm: So if I read that issue correctly, we don't need to annotate the return type since we annotate url and the `-> None` is implied?

pd0wm:
CI still fails. Can you fix the tests?

pd0wm:
@cmyui If you make the tests pass we can merge this.

selfdrive/hardware/tici/agnos.py
cmyui: Yeah, correct. Personally I find this pretty unobvious and being that the second line of the [Zen of Python](https://www.python.org/dev/peps/pep-0020/) reads "Explicit is better than implicit.", this seems pretty unpythonic lol.

I type hint everything in my codebase personally (so I would keep the ` -> None` in this case), but please tell me which you prefer and I'll keep that standard while type hinting openpilot :) cheers

cmyui:
My apologies I've been away from my PC for a bit, time to get to work B)

selfdrive/hardware/tici/agnos.py
pd0wm: Leaving `None` is fine with me.

PR #21089, opened 2021-05-31
Update ui fixes

pd0wm:
- Correctly parse `LastUpdateTime` as UTC to prevent any timezone issues.
- Change text to "checking..." after initiating an update to indicate something is happening
- Set up a file watch on `LastUpdateTime` to update the ui when the check is done
- Handle case where LastUpdateTime is not set. Show nothing instead of "now".

### DIFF
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -211,9 +211,11 @@ DeveloperPanel::DeveloperPanel(QWidget* parent) : QFrame(parent) {
 
 
   QFileSystemWatcher *fs_watch = new QFileSystemWatcher(this);
-  fs_watch->addPath(QString::fromStdString(Params().get_params_path()) + "/d/LastUpdateTime");
+  QString update_param_path = QString::fromStdString(Params().get_params_path()) + "/d/LastUpdateTime";
+  fs_watch->addPath(update_param_path);
 
   QObject::connect(fs_watch, &QFileSystemWatcher::fileChanged, [=](const QString path) {
+    fs_watch->addPath(update_param_path); // Add again after the param was renamed
     updateLabels();
   });
 }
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -209,30 +208,49 @@ DeveloperPanel::DeveloperPanel(QWidget* parent) : QFrame(parent) {
...
+  }
 
   if (labels.size() < dev_params.size()) {
-    versionLbl = new LabelControl("Version", version, QString::fromStdString(params.get("ReleaseNotes", false)).trimmed());
+    versionLbl = new LabelControl("Version", version, QString::fromStdString(params.get("ReleaseNotes")).trimmed());
     layout()->addWidget(versionLbl);
     layout()->addWidget(horizontal_line());
 
-    lastUpdateTimeLbl = new LabelControl("Last Update Check", lastUpdateTime, "The last time openpilot checked for an update.");
+    lastUpdateTimeLbl = new LabelControl("Last Update Check", lastUpdateTime, "The last time openpilot checked for an update. Updates are only checked while off-road.");
```
adeebshihadeh, line 246: don't think we use the offroad/onroad language in any user-facing stuff. maybe "while car is off"?

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -209,30 +208,49 @@ DeveloperPanel::DeveloperPanel(QWidget* parent) : QFrame(parent) {
...
     connect(lastUpdateTimeLbl, &LabelControl::showDescription, [=]() {
-      std::system("pkill -1 -f selfdrive.updated");
+      Params params = Params();
+      if (params.getBool("IsOffroad")) {
+        fs_watch->addPath(QString::fromStdString(params.getParamsPath()) + "/d/LastUpdateTime");
```
adeebshihadeh, line 250: what happens when the update fails?

pd0wm: Will keep saying checking. Will add support for detecting failures in a new PR.

PR #21095, opened 2021-06-01
VW MQB: Add FW for 2015 Audi A3

jyoung8607:
New FW values from a 2015 Audi A3.

**Dongle ID:** 8d486e223e697835

Thanks to community A3 owner iamhtoo.


pd0wm:
Do you also have a route with the FW versions from this user (can be in dashcam mode)? All routes from this dongle ID lack carFw.

jyoung8607:
I don't. The firmware versions he provided were from `fw_versions.py` CLI, and well-formed/reasonable looking for the car in question. That said, I've inquired with the user to find out why his uploaded routes don't have carFw, and I'll revert to draft until we can check.

pd0wm:
Yeah, it's nice to have if we want to check later why something was added, and so we can cross reference with the VIN.

jyoung8607:
Dashcam-mode route 8d486e223e697835|2021-06-01--18-01-29 looks to have carFw. The user reports he's now locally edited values.py and the car is recognized.

PR #21101, opened 2021-06-01
Toyota: Only use gas interceptor under 19 mph

sshane:
Uses pedal tuning from 0 to 19 mph, then transitions to stock tuning from 19 to 22 mph. There's also some one-way hysteresis for switching between the two controllers; 3 mph above the switch speed.

Some notes for future PR to convert acceleration to gas:

I think the minimum required component to make the solution as universal as possible is a function that takes in desired acceleration and current speed and output a gas percentage. The function would have an offset that changes based on speed (think how the required amount of gas to maintain your speed increases as you speed up), and a linear component that ramps up the gas based on desired acceleration (it seems to be pretty linear from the data I've gathered). The problem to decide when to switch between giving gas and not (if we want to coast or brake) *should* mostly be solved by an accurate gas function (hopefully without needing a coasting acceleration function).

Then we can remove the pedal specific tuning and return braking performance to stock at all speeds, and we can parameterize the minimum speed we can send acc requests to the car, by car. (The Prius only uses the pedal for auto resume, so setting it to something like 1 mph will let us use the more reliable ACC commands sooner for those vehicles)


selfdrive/car/toyota/interface.py
```diff
@@ -345,17 +345,15 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
...
     # intercepting the DSU is a community feature since it requires unofficial hardware
     ret.communityFeature = ret.enableGasInterceptor or ret.enableDsu or smartDsu
 
     if ret.enableGasInterceptor:
-      ret.gasMaxBP = [0., 9., 35]
-      ret.gasMaxV = [0.2, 0.5, 0.7]
-      ret.longitudinalTuning.kpV = [1.2, 0.8, 0.5]
-      ret.longitudinalTuning.kiV = [0.18, 0.12]
+      ret.gasMaxBP = [0., MIN_ACC_SPEED, MIN_ACC_SPEED]
+      ret.gasMaxV = [0.2, 0.48, 0.5]
```
ShaneSmiskol, line 356: Difference could be small enough, need to test. Also, gas response will be different if we want the same braking performance as stock. Unless we want to keep the same not as strong braking for now

selfdrive/car/toyota/interface.py
```diff
@@ -352,10 +352,13 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
     ret.communityFeature = ret.enableGasInterceptor or ret.enableDsu or smartDsu
 
     if ret.enableGasInterceptor:
-      ret.gasMaxBP = [0., 9., 35]
-      ret.gasMaxV = [0.2, 0.5, 0.7]
-      ret.longitudinalTuning.kpV = [1.2, 0.8, 0.5]
-      ret.longitudinalTuning.kiV = [0.18, 0.12]
+      # Keeps same pedal tuning below MIN_ACC_SPEED, with stock tuning above
+      ret.gasMaxBP = [0., ret.minEnableSpeed]
```
adeebshihadeh, line 356: check if minEnableSpeed is > 0

selfdrive/car/toyota/carcontroller.py
```diff
@@ -6,6 +6,7 @@
                                            create_fcw_command, create_lta_steer_command
 from selfdrive.car.toyota.values import Ecu, CAR, STATIC_MSGS, NO_STOP_TIMER_CAR, TSS2_CAR, CarControllerParams
 from opendbc.can.packer import CANPacker
+from selfdrive.config import Conversions as CV
```
adeebshihadeh, line 9: imports are sorted alphabetically

selfdrive/car/toyota/carcontroller.py
```diff
@@ -32,9 +33,12 @@ def __init__(self, dbc_name, CP, VM):
     self.alert_active = False
     self.last_standstill = False
     self.standstill_req = False
-
     self.steer_rate_limited = False
 
+    self.can_use_pedal = True
```
adeebshihadeh, line 38:
```suggestion
    self.use_interceptor = True
```

selfdrive/car/toyota/carcontroller.py
```diff
@@ -32,9 +33,11 @@ def __init__(self, dbc_name, CP, VM):
     self.alert_active = False
     self.last_standstill = False
     self.standstill_req = False
-
     self.steer_rate_limited = False
 
+    self.use_interceptor = True
```
adeebshihadeh, line 38:
```suggestion
    self.use_interceptor = CP.enableGasInterceptort
```

selfdrive/car/toyota/carcontroller.py
```diff
@@ -32,9 +33,11 @@ def __init__(self, dbc_name, CP, VM):
     self.alert_active = False
     self.last_standstill = False
     self.standstill_req = False
-
     self.steer_rate_limited = False
 
+    self.use_interceptor = CP.enableGasInterceptor
+    self.pedal_hyst_gap = 3. * CV.MPH_TO_MS
```
pd0wm, line 39: Does this need to be an attribute. A global constant sounds more appropriate (e.g. `PEDAL_HYST_GAP`).

selfdrive/car/toyota/carcontroller.py
```diff
@@ -49,15 +52,21 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, hud_alert,
...
-      apply_accel = 0.06 - actuators.brake
-    else:
-      apply_accel = actuators.gas - actuators.brake
+      # handle hysteresis when around the minimum acc speed
+      if CS.out.vEgo < MIN_ACC_SPEED:
+        self.use_interceptor = True
+      elif CS.out.vEgo > MIN_ACC_SPEED + self.pedal_hyst_gap:
+        self.use_interceptor = False
+
+      if self.use_interceptor and enabled:
```
pd0wm, line 65: `use_interceptor` can be a local variable

selfdrive/car/toyota/carcontroller.py
```diff
@@ -32,9 +33,11 @@ def __init__(self, dbc_name, CP, VM):
     self.alert_active = False
     self.last_standstill = False
     self.standstill_req = False
-
     self.steer_rate_limited = False
 
+    self.use_interceptor = CP.enableGasInterceptor
```
pd0wm, line 38: Why not init to False?

selfdrive/car/toyota/interface.py
```diff
@@ -345,17 +345,20 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
...
 
     if ret.enableGasInterceptor:
-      ret.gasMaxBP = [0., 9., 35]
-      ret.gasMaxV = [0.2, 0.5, 0.7]
-      ret.longitudinalTuning.kpV = [1.2, 0.8, 0.5]
-      ret.longitudinalTuning.kiV = [0.18, 0.12]
+      # Keeps same pedal tuning below MIN_ACC_SPEED with stock tuning above
+      ret.gasMaxBP = [0., MIN_ACC_SPEED]
+      ret.gasMaxV = [0.2, 0.5]
+      ret.longitudinalTuning.kpBP = [0., 5., MIN_ACC_SPEED, MIN_ACC_SPEED, 35.]
```
pd0wm, line 358: Use the same hysteresis here for a more smooth transition?

pd0wm: e.g. [0., 5., MIN_ACC_SPEED, MIN_ACC_SPEED + PEDAL_HYST_GAP, 35.]

selfdrive/car/toyota/carcontroller.py
ShaneSmiskol: That's fair, I forget most cars openpilot can start while driving. I was assuming we always starts from 0

selfdrive/car/toyota/carcontroller.py
ShaneSmiskol: That sounds appropriate. I was thinking about moving them into CarControllerParams, but I think we want to keep that class more unified between makes

selfdrive/car/toyota/interface.py
```diff
@@ -345,17 +345,20 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
...
 
     if ret.enableGasInterceptor:
-      ret.gasMaxBP = [0., 9., 35]
-      ret.gasMaxV = [0.2, 0.5, 0.7]
-      ret.longitudinalTuning.kpV = [1.2, 0.8, 0.5]
-      ret.longitudinalTuning.kiV = [0.18, 0.12]
+      # Keeps same pedal tuning below MIN_ACC_SPEED with stock tuning above
+      ret.gasMaxBP = [0., MIN_ACC_SPEED]
+      ret.gasMaxV = [0.2, 0.5]
+      ret.longitudinalTuning.kpBP = [0., 5., MIN_ACC_SPEED, MIN_ACC_SPEED + PEDAL_HYST_GAP, 35.]
```
adeebshihadeh, line 358: this doesn't preserve the tuning set above

RELEASES.md
```diff
@@ -6,6 +6,7 @@ Version 0.8.5 (2021-XX-XX)
  * Lexus UX Hybrid 2019 support thanks to brianhaugen2!
  * SEAT Leon 2017 & 2020 support thanks to jyoung8607!
  * Škoda Octavia 2015 & 2019 support thanks to jyoung8607!
+ * Toyota: improved longitudinal above 25 mph when using a comma pedal
```
adeebshihadeh, line 9: car ports are generally on the bottom of the release notes

selfdrive/car/toyota/carcontroller.py
```diff
@@ -49,15 +50,21 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, hud_alert,
     # *** compute control surfaces ***
 
     # gas and brake
-
-    apply_gas = clip(actuators.gas, 0., 1.)
+    apply_gas = 0.
+    apply_accel = actuators.gas - actuators.brake
```
adeebshihadeh, line 54: think this is more clear:
```suggestion
    interceptor_gas_cmd = 0.
    pcm_accel_cmd = actuators.gas - actuators.brake
```

RELEASES.md
```diff
@@ -2,6 +2,7 @@ Version 0.8.5 (2021-XX-XX)
 ========================
  * NEOS update: improved reliability and stability
  * Smart model-based Forward Collision Warning
+ * Toyota: improved longitudinal above 25 mph when using a comma pedal
```
adeebshihadeh, line 5:
```suggestion
 * Improved longitudinal control on Toyotas with a comma pedal
```

ShaneSmiskol:
Tune transition, bottom lines are pedal, top is stock:

Screenshot from 2021-06-03 11-30-18

PR #21109, opened 2021-06-02
Cleanup carstate quoting styles

jyoung8607:
Per discussion with @pd0wm: cleanup after many years of organic growth, copypasta, and cross-vehicle pollination: use a consistent quoting style in each car port's `carstate.py`.


pd0wm:
Looks good to me. Let's do all of them.

PR #21120, opened 2021-06-03
steerActualDelay actual value

zorrobyte:
**Description**
Controls tunables shouldn't be offset outside of interface.py, otherwise, the values will not work as expected. if the new MPC needs higher delays, that's fine; but let's make the values in interface accurate.

**Verification**
Should be obvious

### DIFF
--- a/selfdrive/car/chrysler/interface.py
+++ b/selfdrive/car/chrysler/interface.py
@@ -26,7 +26,7 @@ class CarInterface(CarInterfaceBase):
     ret.lateralTuning.pid.kpBP, ret.lateralTuning.pid.kiBP = [[9., 20.], [9., 20.]]
     ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.15, 0.30], [0.03, 0.05]]
     ret.lateralTuning.pid.kf = 0.00006   # full torque for 10 deg at 80mph means 0.00007818594
-    ret.steerActuatorDelay = 0.5
+    ret.steerActuatorDelay = 0.3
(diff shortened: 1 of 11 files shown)


adeebshihadeh:
The values in the car interfaces are only for the steering actuator. As the comment says, the other 0.2s is an estimate of the delay in other parts of the stack.

zorrobyte:
steerActuatorDelay is calculated by looking at the steer cmd and the torque output. If this delay if left in, regardless of where the lag is in controls, the calculated value will be skewed +0.2s

@pd0wm what are your thoughts here? steerActuatorDelay is a kludge that's used anyway to account for full stack delay and isn't represented by the actual plant model, anyway.

pd0wm:
The 0.2 is about an upper limit of the time it takes to go from light hitting the sensor to reach plannerd. There are many steps from exposure, reading out the sensor, copying the picture a couple times and running modeld. I agree with @adeebshihadeh this value shouldn't be included in steerActuatorDelay since they come from other sources. Something like indi shouldn't use this extra 0.2 and only the actual steer actuator delay.

zorrobyte:
@pd0wm @adeebshihadeh 
Maybe you guys can revert the .2s hack now that the stack lag is coming down, see @geohot 's enhancement here:
https://github.com/commaai/openpilot/pull/23795

```
from debayering, c3 cams are like 2 ms lag vs on C2 what used to be 8 and now is 3.5 
C3 has a completely new camera stack vs C2
```

geohot:
Doubt it. .2s is 200 ms. I don't think 4 ms will make a difference.

PR #21123, opened 2021-06-03
Add support for steering angle sensor precision

jyoung8607:
Potential improvement for PIF tuning/noise. Discussion (or rejection?) welcome.  Depends on commaai/cereal#164.

* Permit setting `carParams.steerAnglePrecision` within a car port, per-CAR if desired
* Default it to zero (no behavior/tuning change for any car port that doesn't use the new setting yet)
* Use that precision to set the deadzone for the lateral PIF controller
* Set the steering angle precision to 0.15° for VW MQB
  * Mainly as a test to show that process_replay rightfully breaks for VW but not for other ports
  * Significant follow-ups coming to improve the VW tune which is currently "meh"
  * I will very likely ask for revert of #20247 to regain 0.1° precision, looks like better tuning with Kale is the right path

Credit to @zorrobyte for the idea.

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -59,6 +59,7 @@ class CarInterfaceBase():
 
     # standard ALC params
     ret.steerControlType = car.CarParams.SteerControlType.torque
+    ret.steerAnglePrecision = 0.
     ret.steerMaxBP = [0.]
     ret.steerMaxV = [1.]
     ret.minSteerSpeed = 0.
--- a/selfdrive/car/volkswagen/interface.py
+++ b/selfdrive/car/volkswagen/interface.py
@@ -46,6 +46,7 @@ class CarInterface(CarInterfaceBase):
     # Global tuning defaults, can be overridden per-vehicle
 
     ret.steerRateCost = 1.0
+    ret.steerAnglePrecision = 0.15
     ret.steerLimitTimer = 0.4
     ret.steerRatio = 15.6  # Let the params learner figure this out
     tire_stiffness_factor = 1.0  # Let the params learner figure this out
--- a/selfdrive/controls/lib/latcontrol_pid.py
+++ b/selfdrive/controls/lib/latcontrol_pid.py
@@ -36,7 +36,7 @@ class LatControlPID():
       steer_feedforward = angle_steers_des_no_offset  # offset does not contribute to resistive torque
       steer_feedforward *= CS.vEgo**2  # proportional to realigning tire momentum (~ lateral accel)
 
-      deadzone = 0.0
+      deadzone = CP.steerAnglePrecision
 
       check_saturation = (CS.vEgo > 10) and not CS.steeringRateLimited and not CS.steeringPressed
       output_steer = self.pid.update(angle_steers_des, CS.steeringAngleDeg, check_saturation=check_saturation, override=CS.steeringPressed,


zorrobyte:
This value would need to be set in all interfaces to build properly, eg, add `ret.steerAnglePrecision = 0` to each brand's `interface.py`.

For reference, TSS2 "good" sensor is 0.05~ degrees precise. TSS-P is 0.1~. ZSS is 0.000001~ degrees precise. I had this idea a long time ago as there's no reason to wind up PI control if error can never be zero between steps due to desiredangle being like 0.00001 while most cars are like 0.1~0.05.

Of course, this PR could be moot if comma is eventually eliminating the PI controller, such as moving to INDI with online tuning. 0.9 will tell.

jyoung8607:
I defaulted it to `0.` up in `interfaces.py`, among other reasons, so that we wouldn't have to touch other ports or vehicles until desired. A lot of templated/default stuff was collected up there, several releases ago.

The build isn't breaking because of the other ports, it's breaking because comma has to review and approve commaai/cereal#164 before I can bump cereal in this PR. That's the only reason the PR is in Draft right now.

ShaneSmiskol:
@jyoung8607 You can always change your submodule branch to your own so the tests in this PR can pass, then switch it back to master once it's merged. Just so it's not a blocker on testing process replay, unit tests, etc. Not super important, but it's an option!

jyoung8607:
With a proper Audi A3 route, now process_replay is highlighting the differences we want to see, on just the one port.

Holding in Draft because it'll need a final cereal bump before merging, but this is otherwise ready for consideration.

`  ("VOLKSWAGEN", "ef895f46af5fd73f|2021-05-22--14-06-35--6"),  # VW.AUDI_A3_MK3`

```
***** results for segment ef895f46af5fd73f|2021-05-22--14-06-35--6 *****
	controlsd
		['sendcan', 0, 'dat']: 909
		carControl.actuators.steer: 1884
		carParams.steerAnglePrecision: 2
		carState.steeringRateLimited: 450
		controlsState.lateralControlState.pidState.i: 1983
		controlsState.lateralControlState.pidState.output: 1884
		controlsState.lateralControlState.pidState.p: 2019
	radard
	plannerd
	calibrationd
	dmonitoringd
	locationd
	paramsd
	ubloxd
```

pd0wm:
Do you have some logs of the same route with and without this change? Wonder if it has a measurable effect on driving performance.

zorrobyte:
as requested @pd0wm 

0.5 degree deadzone (feels like too much):
d5b9bf105ad64d5d|2021-06-07--11-09-09

0.2 degree deadzone:
d5b9bf105ad64d5d|2021-06-07--11-26-17

0.1 degree deadzone:
d5b9bf105ad64d5d|2021-06-07--11-38-55

0 deadzone (stock):
d5b9bf105ad64d5d|2021-06-07--11-50-32

I'm likely going to set mine to 0.05 to account for TSS2 sensor accuracy of ~0.05. @jyoung8607 has a sensor ~0.1 degrees precise. I'd recommend a tuning value of ~0.1 for his vehicle. Please test and report back identical routes of:
0.0,
0.05,
0.1

jyoung8607:
I've just driven a pair of identical test routes. On further thought, I'm not sure what I drove was the best test-case for this change, it's some roads with annoyingly variable camber so there is a lot of error to deal with, but it was quick and easy. I can redo the drives on a more level stretch if really needed.

`7aae592fc08e895f|2021-06-16--14-12-45` on today's master
`7aae592fc08e895f|2021-06-16--14-26-57` on sas-deadzone (this PR)

I haven't yet tried to plotjuggle anything, I will try to do so later tonight. I myself can't make a strong assertion this change is useful, only that it seems logical. We've got the inventor of the ZSS over here that thinks it's worth a try, so I'm trying. @zorrobyte, you have access to this device if you feel inclined to look. 

I believe a deadzone change has been submitted a couple times over the years, I think from Ross and Gernby, but as part of  larger patchsets that were harder to review. My contribution here is to take *just this one change*, test it out in isolation, and see if it's a gain, or a loss, or nothing at all. If we look at the data and it doesn't seem useful, this is fine to close.

pd0wm:
Didn't have time yet to look into this. @jyoung8607 preserved your routes. Noticed that @zorrobyte already did that.

pd0wm:
Made some quick plots of @jyoung8607's drive. Left: master, right: this PR

2021-06-22-165314_2552x1410_scrot

Moving average of error (500 samples):
2021-06-22-165502_2266x327_scrot

Error of the deadzone seems slightly higher on average, but the torque request is less noisy. Which in hindsight, we would could have probably expected.

Under what results would you call this PR an improvement? Is a noisy torque signal a problem if the jitter is so small it's imperceptible in wheel motion (I assume 0.15 degrees is).

jyoung8607:
> Under what results would you call this PR an improvement? Is a noisy torque signal a problem if the jitter is so small it's imperceptible in wheel motion (I assume 0.15 degrees is)?

@zorrobyte, what say you? Was this an issue for say, non-ZSS older Toyota?

adeebshihadeh:
Still want to keep this open? Seems like there's no noticeable affect.

zorrobyte:
@jyoung8607 I’d say, and the reason why I never opened a PR l, is that while it’s an improvement, the effects aren’t very noticeable even if there is a positive effect at the micro level.

I do appreciate your enthusiasm and that this avenue has been throughly explored! If we didn’t have INDI for Prius and was stuck with only the shitty steering sensor, I believe it would be more of a useful enhancement.

jyoung8607:
If it isn't helpful, then it isn't helpful, question answered!

PR #21128, opened 2021-06-03
disable uploading raw by default

adeebshihadeh:


ErichMoraga:
Why is this disabled by default now?  To save a little on data uploads?  I see a lot of questions from people trying to review logs in Cabana, who weren't aware of this change.

Issue #19672, opened 2021-01-06
hyundai: palisade has large variation in learned steer ratio


gregjhogan:
current learned steer ratio for palisade is definitely too low, but idk what to update it to when it swings this much in a single drive
```
17.8 -> 1ad763dd22ef1a0e|2020-12-29--15-15-43--0
20.7 -> 1ad763dd22ef1a0e|2020-12-29--15-15-43--40
22.3 -> 1ad763dd22ef1a0e|2020-12-29--15-15-43--80
20.6 -> 1ad763dd22ef1a0e|2020-12-29--15-15-43--180
16.5 -> 1ad763dd22ef1a0e|2020-12-29--15-15-43--290
```

gregjhogan:
apparently the spec is 15.6, so it is probably variable but haven't found anything that states this is the case
https://www.hyundainews.com/en-us/models/hyundai-palisade-2020-palisade/specifications

zorrobyte:
2019 Rav4
![image](https://user-images.githubusercontent.com/7257172/120910732-3d69ec00-c64f-11eb-972d-ec9f4c334030.png)
https://www.toyota.com/content/ebrochure/2021/rav4_ebrochure.pdf

Learned ratio:
![image](https://user-images.githubusercontent.com/7257172/120910741-4ce93500-c64f-11eb-924c-90191e674364.png)

I don't think the steer ratio has ever been very spot on, eh?

pd0wm:
See my comment in your other PR. I expect a little higher than the factory spec. 

In this case some variability in the learned value between 16.5 and 17 shouldn't be an issue. At a given speed there are many combinations of steer ratio & stiffness that leads to the same handling of the car. The learner needs to look at data at different speeds to find the correct value. Unfortunately the learner has no access to previous data, just the current estimates and current driving data and tries it's best to match that to a steer ratio & stiffness. If you plot stiffness and speed as well you'll probably see a correlation.

I wouldn't be too worried about these small changes. I'm sure at each point in time the stiffness & steer ratio & angle offset estimates matched how the car was driving at that speed. Greg's steer ratio is changing between 16 and 22, which is a much larger variation and is more of a concern.

adeebshihadeh:
Is this still an issue after all the locationd/paramsd fixes? @nuwandavek

nuwandavek:
We can close this for now. This is not a big deal, but we will look at this in the new paramsd rewrite. There are quite a few things we can change that would circumvent any such filter state issue.

[...]

PR #21139, opened 2021-06-04
Rav4 TSS2 ICE & Hybrid, better tune

zorrobyte:
The stock tune is very lazy and will go out of the lanelines on my 2019 Rav4 TSS2 due to low PI F gains. 

My vehicle is identified as `CAR.RAV4_TSS2` without EPS FW `b"8965B42170\x00\x00\x00\x00\x00\x00`

This has also been tested on @C0mpl3t3N00b#0074 's `CAR.RAV4H_TSS2` without EPS FW `b"8965B42170\x00\x00\x00\x00\x00\x00`

The learner also ends up around 16 steerRatio, so I went ahead and updated it.

EPS motor is on rack for TSS2 Rav vs old TSS-P Rav which is on column.

I also interp'd gains at speed as controls are either too reactive, or not reactive enough at various speeds. This should have good lowspeed and highspeed perf.


zorrobyte:
@yufeng66 @mattbsea @dpanych @RICKYMAMA1 @Taik 

You guys want to test this out? I put 100 or so miles into a usable Rav4 tune and if so, we can unify tuning instead of splitting by FW versions.

Taik:
@zorrobyte I don't have much time to tweak but I'll be happy to try out and give my feedback.

yufeng66:
I am happy to try it out this weekend

mattbsea:
To test,  do I just install zorrobyte:rav4tss2_tune ?

zorrobyte:
> To test, do I just install zorrobyte:rav4tss2_tune ?

You may install based on the PR and be sure to compile (git submodule init && git submodule update && scons -j4). If you don't want to do this, I have compiled a branch based on ci here: https://github.com/zorrobyte/openpilot/tree/rav4tune_ci

You can also use shane's installer url here:
https://smiskol.com/fork/zorrobyte/rav4tune_ci

zorrobyte:
I got a route back from @C0mpl3tn00b
image

Metrics look good!
image

zorrobyte:
Went on two interstate drives! Haven't been on the highway for a year and a half... it was fun!

d5b9bf105ad64d5d|2021-06-05--15-50-17
and
d5b9bf105ad64d5d|2021-06-05--15-16-44

Tune is solid to me! Just waiting on @Taik @yufeng66 @mattbsea

C0mpl3t3N00b:
Great work and many thanks @zorrobyte! Please get this man's work up streamed to stock OpenPilot!

zorrobyte:
@pd0wm before we merge this, is it expected for the steerRatio learner to jack up the steerRatio to 16, even though Toyota documents it as 14.3/14.4? Some of this variation in the learner can be seen on https://github.com/commaai/openpilot/issues/19672

if this is expected behavior, than no worries; but how useful is a ratio learner if:
1. The spec is published and well known
2. The learner changes the ratio during a drive, changing controls behavior
3. 
I suppose I've just never understood the ratio learner. Is it more of an adjustment that accounts for other vehicle dynamics? Ie, how much to turn at curvature?

yufeng66:
Tried the turning. Here is my impression.
1.	On highway it’s perfect
2.	Local speed straight road, also perfect
3.	30-40mph winding road, the tuning is slightly more aggressive than I like. I felt it is overcompensating from time to time.  The steering wheel will go back and forth a little bit hunting for the perfect angle. This is the opposite of the turning in release2 now, which is too relaxed with the current laneless model.

ErichMoraga (commented):
Coding it to 14.3 won't do any good if the learner eventually drives it higher (same on other vehicle models).  It would be better to dig into why the learner does that, like... an incorrect algorithm, etc.

zorrobyte:
> Coding it to 14.3 won't do any good if the learner eventually drives it higher (same on other vehicle models). It would be better to dig into why the learner does that, like... an incorrect algorithm, etc.

Oh, I know. I just couldn't decide between setting it to the actual ratio or what the learner sets it at. I decided to go with the factory value.

> 30-40mph winding road, the tuning is slightly more aggressive than I like. I felt it is overcompensating from time to time. The steering wheel will go back and forth a little bit hunting for the perfect angle. This is the opposite of the turning in release2 now, which is too relaxed with the current laneless model.

I think the PI F values are good, actuator response in PJ looks good. This is a symptom of too high of a steerRatio and I bet if I hardcoded it to actual sR, it'd decrease the undesired behavior.

Anyway, I'd be happy to just get better values upstreamed and I'm sure we'll see continued enhancement of the vehicle model, liveparams

jyoung8607:
> 3. 30-40mph winding road, the tuning is slightly more aggressive than I like. I felt it is overcompensating from time to time.  The steering wheel will go back and forth a little bit hunting for the perfect angle. This is the opposite of the turning in release2 now, which is too relaxed with the current laneless model.

For better or worse, at this stage of development, laneless does like to wander a little bit in these situations. Retry with lanes enabled (laneless toggle off) before assuming you have a tune problem.

adnanlanewla:
I tired this tune with lane model. My impressions is that it performs worst than stock op.
The wheel kept wobbling for me. It was more noticeable at lower speeds but even had higher speed the wheel wobbles a lot.

ErichMoraga:
> I tired this tune with lane model.

Sorry to hear it tired you out.  I suggest Zorro disable dynamic/learned SR the way I discussed in #tuning, so you're comparing apples to apples.

zorrobyte:
> I tired this tune with lane model. My impressions is that it performs worst than stock op.
> The wheel kept wobbling for me. It was more noticeable at lower speeds but even had higher speed the wheel wobbles a lot.

Could you please share a route with me? Here or DM on Discord is okay. I'd like to see what is so different

zorrobyte:
I have more work to do and will submit a new PR once I have something more solid.

pd0wm:
Regarding steer ratio, the spec listed by the manufacturer is usually end-to-end. Most racks have variable ratio's, and have a slightly higher ratio in the center. Therefore, it makes sense that the learner finds a slightly higher value. 14.3 vs ~16 is about what I would expect here. The learner doesn't always perfectly converge, but if it's stable around 16 I would put that in instead of 14.3.

ErichMoraga:
Wouldn't it be more interesting to learn/map that vehicles SR curve (could be defined w/ multiple breakpoints, instead of an equation), instead of restricting it to an ever changing value?

zorrobyte:
Ideally, in MPC theory, a controls engineer would breakpoint between various linear MPC configs; but that’s assuming one has a single vehicle model, only one car to optimize.

OP does well considering it’s almost vehicle agnostic. For some ideas you can see how the Apollo group handles MPC and LQR controls for their OSS project.

I wish I was a controls engineer, but I suck at math and things like physics, calculus, and linear algebra is often required when writing good controls engineering code. It’s besides the point, but I hope that in some alternative reality, I am one 😉

mattbsea:
@zorrobyte Sorry I didn't have time to install and test the branch, but sounds like it's closed for now.  Please let me know if there's anything I can help with in the future.  I drive a 2020 RAV4 Hybrid.

PR #21144, opened 2021-06-04
CARLA: save calibration params as capnp

siggie0815:
When starting simulation on latest master, OP tells us **Calibration in Progress**. However, nothing happens as we cannot drive over 15mph.
The problem is that the calibration data changed from JSON format to a message byte stream in OP. This change was not reflected in bridge.py in the simulator.


siggie0815:
OK, we can drive manually... just figured out :rofl: 

Well I still think this is usefull...

PR #21152, opened 2021-06-05
Refactor C++ LogReader

deanlee:
- fix leaks in `FileReader` & `LogReader` (this PR still has not released the `LogReader` in reply.cc, it will be resolved with a new merging mechanism in the next pr) 
- merge events in the main thread.
- publish messages from raw buffer, without deep copy


selfdrive/ui/replay/filereader.h
```diff
@@ -1,70 +1,90 @@
...
 #include <QElapsedTimer>
 #include <QMultiMap>
 #include <QNetworkAccessManager>
-#include <QReadWriteLock>
 #include <QString>
-#include <QVector>
-#include <QWidget>
+#include <QThread>
+#include <unordered_map>
+#include <vector>
```
adeebshihadeh, line 11: fix the include order

selfdrive/ui/replay/filereader.h
```diff
@@ -1,70 +1,90 @@
...
-  QNetworkAccessManager *qnam;
-  QElapsedTimer timer;
-  QString file;
+  void startHttpRequest();
+  QNetworkReply *reply_ = nullptr;
+  QUrl url_;
 };
 
-typedef QMultiMap<uint64_t, cereal::Event::Reader> Events;
+enum CameraType {
```
adeebshihadeh, line 33: fine here for now, but we should really clean this up in camerad and have a nice header for all this

selfdrive/ui/replay/filereader.cc
```diff
@@ -1,139 +1,132 @@
...
-void FileReader::readyRead() {
-  QByteArray dat = reply->readAll();
-  printf("got http ready read: %d\n", dat.size());
+void FileReader::startHttpRequest() {
+  QNetworkAccessManager *qnam = new QNetworkAccessManager(this);
+  QNetworkRequest request(url_);
+  request.setAttribute(QNetworkRequest::FollowRedirectsAttribute, true);
+  reply_ = qnam->get(request);
+  connect(reply_, &QNetworkReply::finished, [=]() {
+    emit !reply_->error() ? finished(reply_->readAll()) : failed(reply_->errorString());
```
adeebshihadeh, line 53: use an if statement here

selfdrive/ui/replay/filereader.cc
```diff
@@ -1,139 +1,132 @@
...
-
-      // increment
-      event_offset = (char*)cmsg.getEnd() - raw.data();
-    } catch (const kj::Exception& e) {
+      words = kj::arrayPtr(evt->reader.getEnd(), words.end());
+      events.insert(evt->mono_time, evt.release());
+    } catch (const kj::Exception &e) {
       // partial messages trigger this
-      //qDebug() << e.getDescription().cStr();
+      // qDebug() << e.getDescription().cStr();
```
adeebshihadeh, line 124: let's uncomment this debug print

selfdrive/ui/replay/filereader.cc
```diff
@@ -1,139 +1,132 @@
...
-      // TODO: rewrite with callback
-      if (event.which() == cereal::Event::ROAD_ENCODE_IDX) {
-        auto ee = event.getRoadEncodeIdx();
-        eidx_local.insert(ee.getFrameId(), qMakePair(ee.getSegmentNum(), ee.getSegmentId()));
+  valid_ = true;
+  kj::ArrayPtr<const capnp::word> words((const capnp::word *)raw_.data(), raw_.size() / sizeof(capnp::word));
+  while (!exit_ && words.size() > 0) {
+    try {
+      std::unique_ptr<Event> evt = std::make_unique<Event>(words);
+      switch (evt->which) {
```
adeebshihadeh, line 107: why not clean this up with capnp's dynamic api?

adeebshihadeh: here's an example: https://github.com/commaai/openpilot/blob/master/selfdrive/camerad/cameras/camera_frame_stream.cc#L47

deanlee: Because I think it's faster and straightforward

selfdrive/ui/replay/replay.cc
```diff
@@ -104,6 +101,14 @@ void Replay::removeSegment(int n) {
   }
 }
 
+void Replay::mergeEvents() {
```
adeebshihadeh, line 104: I like this move :)

selfdrive/ui/replay/filereader.h
```diff
@@ -1,70 +1,91 @@
...
+
+  uint64_t mono_time;
+  cereal::Event::Which which;
+  cereal::Event::Reader event;
+  capnp::FlatArrayMessageReader reader;
+  kj::ArrayPtr<const capnp::word> words;
+};
+
+typedef QMultiMap<uint64_t, Event*> Events;
+typedef std::unordered_map<uint32_t, EncodeIdx> EncodeIdxMap;
```
adeebshihadeh, line 65: not a fan of these typedefs, I think it's pretty confusing when reading "Events" and "EncodeIdxMap" later on

selfdrive/ui/replay/filereader.cc
```diff
@@ -1,139 +1,134 @@
 #include "selfdrive/ui/replay/filereader.h"
 
+#include <bzlib.h>
 #include <QtNetwork>
 
-FileReader::FileReader(const QString& file_) : file(file_) {
-}
+#include "selfdrive/common/timing.h"
```
adeebshihadeh, line 6: seems unused?

deanlee: Yes, it is for testing

selfdrive/ui/replay/filereader.h
```diff
@@ -1,70 +1,87 @@
...
+const int MAX_CAMERAS = std::size(ALL_CAMERAS);
 
-class LogReader : public FileReader {
-Q_OBJECT
+struct EncodeIdx {
+  int segmentNum;
+  uint32_t frameEncodeId;
+};
+
+class Event {
```
adeebshihadeh, line 47: think it would be cleaner without this class. we should just use the event reader directly

deanlee: In addition, `which` and `mono_time` will be used  for sorting in  #21004, to avoid the problem that event may be lost or resent if sorted only by mono_time

adeebshihadeh:
Seems to work, but I'm getting this:
```
QObject: Cannot create children for a parent that is in a different thread.
(Parent is Replay(0x5aa300), parent's thread is QThread(0x544480), current thread is QThread(0x5f29e0)
```

deanlee:
fixed. AddSegment is called in segmentQueueThread,we can't set parent to replay in it

PR #21159, opened 2021-06-05
Add chime to lane changed blocked alert

speedking456:
**Description:** Changed the lane changed blocked alert to be more prominent, it's now orange and makes an alert sound.


adeebshihadeh:
I like this change, what do you think @pd0wm?

pd0wm:
Yes. I like it!

adeebshihadeh:
Merging this one, let's do that in a separate PR

PR #21161, opened 2021-06-06
Updated FPv2 values for Santa Fe 2020

CruiseBrantley:
Fix Fuzzy match for 2020 Santa Fe
dongleid: c66fe4bb6baaa968


selfdrive/car/hyundai/values.py
```diff
@@ -291,6 +291,7 @@ class Buttons:
       b'\xf1\x00TM ESC \x03 103\x18\x11\x07 58910-S2600\xf1\xa01.03',
       b'\xf1\x00TM ESC \x0c 103\x18\x11\x08 58910-S2650',
       b'\xf1\x00TM ESC \x0c 103\x18\x11\x08 58910-S2650\xf1\xa01.03',
+      b'\xf1\x00TM ESC \x02 104\031\a\a 58910-S2600\xf1\xa01.04',
```
CruiseBrantley, line 294: Tried `x1f` to replace `031`, however it didn't fix the match until I added it unaltered. 🤷‍♂️

pd0wm: `031` is octal. Try this:
```suggestion
      b'\xf1\x00TM ESC \x02 104\x19\x07\x07 58910-S2600\xf1\xa01.04,
```

pd0wm: Ah, that one is already present!

CruiseBrantley: Oh alright, strange that it didn't detect when I added the value there with `x1f` the last time I booted it seemed to detect even from release2 with no changes though. HKG fingerprints seems a little confusing but I think the transmission value should still be added at least. I'll remove the extra esp value.

PR #21175, opened 2021-06-07
Rav4 TSS2 - much better long tuning and SnG

zorrobyte:
Stock long tuning for Rav4 TSS2 is uncomfortable and will press brakes hard with leads, cut-ins and in stop and go. My insurance company's tracking also picks it up as an unsafe braking event in their telematics, increasing my insurance rate; which is hilarious.
4B75FCCC-7C69-44D0-9467-3598A4A137C0

This long tune feels very comfortable, even for 40MPH to stopped lead and in stop and go. Acceleration from stock feels natural and keeps up with traffic without overshoot.

Routes:
d5b9bf105ad64d5d|2021-06-07--11-09-09
d5b9bf105ad64d5d|2021-06-07--11-26-17
d5b9bf105ad64d5d|2021-06-07--11-38-55
d5b9bf105ad64d5d|2021-06-07--11-50-32

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -175,6 +175,16 @@ class CarInterface(CarInterfaceBase):
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.15], [0.05]]
       ret.mass = 3370. * CV.LB_TO_KG + STD_CARGO_KG
       ret.lateralTuning.pid.kf = 0.00004
+      #known good values
+      ret.longitudinalTuning.deadzoneBP = [0., 8.05]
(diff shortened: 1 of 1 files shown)


pd0wm:
Any good stock tuning route I can use from your device?

zorrobyte:
> Any good stock tuning route I can use from your device?

Here is a route on `master`
d5b9bf105ad64d5d|2021-06-01--19-48-18

zorrobyte:
This was a case of confirmation bias, more care and thought will need to go into future PRs.

The tuning values, in exception to stopping/starting/accel rate, was already being used. There may be some SnG enhancement to be had, but such would require better validation and testing.

Closing. Thank you for being kind Willem.

PR #21177, opened 2021-06-07
 startup alert for no fw returned

adeebshihadeh:


selfdrive/controls/lib/events.py
```diff
@@ -268,6 +268,14 @@ def startup_fuzzy_fingerprint_alert(CP: car.CarParams, sm: messaging.SubMaster,
     ET.PERMANENT: startup_fuzzy_fingerprint_alert,
   },
 
+  EventName.startupNoFw: {
+    ET.PERMANENT: Alert(
+      "Car Unrecognized",
+      "Check All Connections",
```
VirtuallyChris, line 274: Maybe 'Check OBD-II Connection'?

PR #21180, opened 2021-06-08
Use correct gas signal for Optima Hybrid

sshane:
Fixes the Kia Optima Hybrid never disengaging on gas, it was using the ICE signals which aren't present. Probably should get a route with this.

This car uses the current EV signal, not the new hybrid signal. It exists, but from one example the current EV signal captured a gas press event that the new HEV signal did not


adeebshihadeh:
This car would have never worked due to the CAN parser checks.

adeebshihadeh:
Let's revisit this when we have a user and test route.

PR #21198, opened 2021-06-08
Fingerprint for Lexus NX Hybrid 2019

neneboe:
Added fingerprint for Lexus NX Hybrid 2019. All it needed was the engine fwVersion.

My dongle id is ed54b42db6f29263.


selfdrive/car/toyota/values.py
```diff
@@ -53,7 +53,7 @@ class CAR:
   LEXUS_IS = "LEXUS IS 2018"
   LEXUS_CTH = "LEXUS CT HYBRID 2018"
   RAV4H_TSS2 = "TOYOTA RAV4 HYBRID 2019"
-  LEXUS_NXH = "LEXUS NX HYBRID 2018"
+  LEXUS_NXH = "LEXUS NX HYBRID 2018 & 2019"
```
VirtuallyChris, line 56: Revert this change

VirtuallyChris (changes requested):

selfdrive/car/toyota/values.py
VirtuallyChris: Just in the readme, the platform is the 2018 and up but it identifies as 2018. That's why all cars only have 1 year listed

neneboe: Understood. Thanks, Chris.

PR #21070, opened 2021-05-28
Remove radard

haraschax:
- Add model that predicts lead trajectory, not just lead position and velocity. This allows longitudinal MPC to anticipate the future more and be more smooth
- Remove radar
- Replace handcoded cruise control logic with MPC
- Replace handcoded FCW logic with a simple threshold check
- Simplify overall architecture
- Make velocity + acceleration API with controls explicit
- Increase thresholds for cruise accel and deccel
- Remove controlsd from longitudinal tests
- Change packets to output control trajectory
- Let controls do anticipation

**Still todo**
- Reduce cpu usage of planner and controls
- Tune accel limits and cruise mpc
- Tune lead mpc costs

PR #21199, opened 2021-06-08
fcw.py: small cleanup

deanlee:


adeebshihadeh:
This file is being deleted in #21070

PR #21207, opened 2021-06-09
new class QParams, used to gain access to the params in Qt

deanlee:


deanlee:
qParams will be initialized before default_path&persistent_path.put  default_path&persistent_path in ctor to make sure paths get initialized before it is used

adeebshihadeh:
What's the point of this class?

deanlee:
no need to convert between qstring and std::string every time, and vice versa.

adeebshihadeh:
I don't think this is cleaner than the std::string and QString conversions. Willem opened #21229 to address the possibility of failing to create one of the Params instances.

PR #21213, opened 2021-06-10
ui: networking enhancements

sshane:
Todo:
- [ ] Able to forget a connected network (do we want to press the ssid name, or have a separate button?)
- [x] Detect if a network is active (connected at any point) and don't ask for password.
- [x] See how many times we can reuse list_connections_ssid, or completely replace the old function (some functions use it differently, need to look into that)
- [ ] How should we handle Forget button for hotspot network?
- [ ] Make refreshing more responsive


ShaneSmiskol:
still a draft, will finish up tomorrow

ShaneSmiskol:
This PR is getting split up.

PR to remember networks: https://github.com/commaai/openpilot/pull/21233

PR #21218, opened 2021-06-10
Qt: remove chain of signals, emit through a singleton class signalMap

deanlee:
- [x] remove chain of signals, emit system-level signals through a singleton class signalMap
- [x] emit signals for toggles
- [x] remove  the periodic reading, connect to the toggle signals,


pd0wm:
Both Adeeb and I weren't a fan of this. I don't like the spaghetti of signals either, but this seems to be a bit of an anti-pattern. We will think about something to clean this up.

PR #21224, opened 2021-06-11
[Feature] Enable Wifi tethering when on road

mogorman:
Enable wifi tethering when car is started,  disable it when car is stopped.

I realize where I put the activation code, `ui.cc` might not be the best place. Happy to move it to a better one if suggested.

Code is tested and working on my C2 in my pacifica.


adeebshihadeh:
Thanks for the PR, but this isn't a feature we're interested in merging. The settings are intentionally very minimal, but this sounds like a nice feature for a fork!

PR #21226, opened 2021-06-11
Move all MPC to acados

haraschax:
Initial PR to switch to acados. acados is signficantly cleaner than ACADO, ACADO is also no longer maintained.
Before we can decide if we can move to acados we need to:
- [ ] add acados to openpilot
- [ ] make sure it runs on device
- [ ] make sure it is feasible to run  without too much cpu
- [ ] do we need to use C interface?

If we decide to move to acados we still need to

- [ ] clean up
- [ ] confirm results are the same
- [ ] port longitudinal


pd0wm:
Am I correct in assuming the python package generates C code and builds it on the fly? How much hacking do we have to do split the code generation out and put that in our regular build process?

HaraldSchafer:
Yes that is correct. Not that much hacking I think. From what I understand it was designed to be split up. So that you can generate on pc and then run on embedded system in C. Adeeb was going to look into it

selfdrive/controls/lib/lateral_mpc/SConscript
```diff
@@ -1,48 +1,48 @@
...
+lenv["ENV"]["LD_LIBRARY_PATH"] = os.getenv("LD_LIBRARY_PATH", "") + libpath
+lenv.Command(generated_files,
+             ["lat_mpc.py"],
+             f"cd {Dir('.').abspath} && python lat_mpc.py")
 
-mpc_files = ["lateral_mpc.c"] + generated_c
-env.SharedLibrary('mpc', mpc_files, LIBS=['m', 'qpoases'], LIBPATH=['lib_qp'], CPPPATH=cpp_path)
+lenv.Command(File(f"{gen}/libacados_ocp_solver_lat.so"),
+             generated_files,
+             f"cd {Dir(gen).abspath} && make ocp_shared_lib")
```
pd0wm, line 48: We should make scons compile this so we control optimization levels and compile with `--mcpu=cortex-a57`

PR #21231, opened 2021-06-11
util::read_file: better fallback

deanlee:


adeebshihadeh:
Can you add some tests for this? The current implementation was thoroughly tested, and I'm hesitant to review this without any tests.

deanlee:
tested locally

adeebshihadeh:
I'd like unit tests for this if we're going to merge it.

PR #21232, opened 2021-06-11
UI widgets: remove unnecessary setLayout

deanlee:


adeebshihadeh:
We do this in all the widgets right now, although unnecessary because we set the layout parent. Any docs on the best practice here? If we decide to go with this, let's fix all instances.

deanlee:
I think construct it with parent,and unified name to main_layout will be clearer.

adeebshihadeh:
I'm fine with that. Let's make it consistent across all widgets.

selfdrive/ui/qt/maps/map.cc
```diff
@@ -441,18 +441,19 @@ void MapWindow::offroadTransition(bool offroad) {
...
     icon_01 = new QLabel;
     layout->addWidget(icon_01);
     layout->addStretch();
-    layout_outer->addLayout(layout);
+    main_layout->addLayout(layout);
   }
 
   {
-    QVBoxLayout *layout = new QVBoxLayout;
+    QWidget * w = new QWidget;
```
adeebshihadeh, line 455:
```suggestion
    QWidget *w = new QWidget;
```

PR #21233, opened 2021-06-11
ui: remember previous connections

sshane:


selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -392,6 +403,27 @@ void WifiManager::disconnect() {
   }
 }
 
+QString WifiManager::ssid_from_path(const QDBusObjectPath &path) {
```
ShaneSmiskol, line 406: will make other functions that call list_connections use this function in another PR

ShaneSmiskol:
Should be ready

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -279,6 +283,31 @@ QVector<QDBusObjectPath> WifiManager::get_active_connections() {
   return conns;
 }
 
+bool WifiManager::isKnownNetwork(const QString &ssid) {
+  for(QDBusObjectPath path : list_connections()) {
```
adeebshihadeh, line 287: const ref?

adeebshihadeh, line 287: this loop already exists twice, abstract into a function

ShaneSmiskol: I did the refractor in the other PR but I can do it here

adeebshihadeh: yeah, just abstract into a function

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -107,7 +107,7 @@ void Networking::refresh() {
 }
 
 void Networking::connectToNetwork(const Network &n) {
-  if (n.security_type == SecurityType::OPEN) {
+  if ((n.security_type == SecurityType::OPEN) || wifi->isKnownNetwork(n.ssid)) {
```
adeebshihadeh, line 110: think it makes more sense to call activate_wifi_connection here

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -412,56 +429,27 @@ QVector<QDBusObjectPath> WifiManager::list_connections() {
...
-        if (key == "ssid") {
-          if (val == tethering_ssid.toUtf8()) {
-            QDBusInterface nm3(nm_service, nm_path, nm_iface, bus);
-            nm3.setTimeout(dbus_timeout);
-            nm3.call("ActivateConnection", QVariant::fromValue(path), QVariant::fromValue(QDBusObjectPath(devicePath)), QVariant::fromValue(QDBusObjectPath("/")));
-            return true;
-          }
-        }
-      }
+  for (const QDBusObjectPath path : list_connections()) {
```
adeebshihadeh, line 447: seems like this loop should be moved into the function too, all the instances do the same thing.

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -212,8 +212,6 @@ void WifiManager::connect(const Network &n, const QString &password) {
 
 void WifiManager::connect(const Network &n, const QString &username, const QString &password) {
   connecting_to_network = n.ssid;
-  // disconnect();
-  clear_connections(n.ssid); //Clear all connections that may already exist to the network we are connecting
```
ShaneSmiskol, line 216: Only here if new connection

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -392,6 +381,37 @@ void WifiManager::disconnect() {
...
+}
+
+QDBusObjectPath WifiManager::path_from_ssid(const QString &ssid) {  // ensure ssid is a known connection before calling this function
+  for (const QDBusObjectPath path : list_connections()) {
+    if (ssid_from_path(path) == ssid) {
+      return path;
+    }
+  }
+  QDBusObjectPath path;
+  return path;
```
ShaneSmiskol, line 412: Not sure what to do here, you should only call this function if it's a known connection, and every referencing location does check that, so is returning an empty QDBusObjectPath variable okay, or something like `NULL`? Not sure what best practice is here.

selfdrive/ui/qt/offroad/wifiManager.cc
adeebshihadeh, line 216: why remove this?

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -255,7 +253,7 @@ void WifiManager::deactivate_connections(const QString &ssid) {
       if (Ssid == ssid) {
         QDBusInterface nm2(nm_service, nm_path, nm_iface, bus);
         nm2.setTimeout(dbus_timeout);
-        nm2.call("DeactivateConnection", QVariant::fromValue(active_connection_raw));// TODO change to disconnect
```
adeebshihadeh, line 258: are you sure that's what the comment meant?

selfdrive/ui/qt/offroad/wifiManager.cc
ShaneSmiskol: connect is now only reached if it's a new connection, `clear_connections` actually deleted the connection, not disconnected from it

selfdrive/ui/qt/offroad/wifiManager.cc
ShaneSmiskol: I'm pretty sure, `DeactivateConnection` disconnects from the network

selfdrive/ui/qt/offroad/wifiManager.cc
ShaneSmiskol: We can uncomment the disconnect line though, that will only deactivate the connection if we're currently connected to a network

selfdrive/ui/qt/offroad/wifiManager.cc
adeebshihadeh: let's just check if the path is empty

pd0wm:
There is no indication that it's starting to connect to the new network. Can we show "connecting" like when connecting to a new network? At a minimum the buttons needs to change color while they are held down so you know it registered the click.

ShaneSmiskol:
Yes, I'm working on refactoring the experience now. The issue is that it takes 100 ms to refresh the networks at the office, and the refresh rate is every 5 seconds. It also does it in the main UI thread so it's blocking and lags UI. I'm working on making it poll the networks in a background thread, and then have it update the UI as the state is changed, and information is updated

PR #21241, opened 2021-06-12
camerad: added replay camera

deanlee:


selfdrive/camerad/SConscript
```diff
@@ -9,6 +9,8 @@ elif arch == "larch64":
   libs += ['atomic']
   cameras = ['cameras/camera_qcom2.cc']
 else:
+  libs += ['avutil', 'avcodec', 'avformat', 'swscale']
```
adeebshihadeh, line 12: moves this down into the else

selfdrive/camerad/cameras/camera_replay.cc
```diff
@@ -0,0 +1,116 @@
...
+
+#include "selfdrive/common/clutil.h"
+#include "selfdrive/common/util.h"
+
+extern ExitHandler do_exit;
+
+const char *BASE_URL = "https://commadataci.blob.core.windows.net/openpilotci/";
+
+const std::string road_camera_routes[] = {
+    "0c94aa1e1296d7c6|2021-05-05--19-48-37",
```
adeebshihadeh, line 14: let's just use on route. fine to loop over the same segment too

adeebshihadeh: this is mostly for quick testing

adeebshihadeh:
I just realized CARLA still uses frame stream. This will need to be moved to visionipc before we can merge this - https://github.com/commaai/openpilot/blob/master/tools/sim/bridge.py#L65.

selfdrive/camerad/cameras/camera_replay.cc
```diff
@@ -0,0 +1,116 @@
...
+}
+
+void process_driver_camera(MultiCameraState *s, CameraState *c, int cnt) {
+}
+
+void cameras_run(MultiCameraState *s) {
+  std::vector<std::thread> threads;
+  threads.push_back(start_process_thread(s, &s->road_cam, process_road_camera));
+  threads.push_back(start_process_thread(s, &s->driver_cam, process_driver_camera));
+  std::thread road_cam_thread = std::thread(road_camera_thread, &s->road_cam);
```
adeebshihadeh, line 109: why not put all of these in threads?

adeebshihadeh:
Can we just add this as a new camera for now? I don't have time to finish #21324, but I'd like to get this merged. This should be the default camera, and frame stream gets built with a new flag.

deanlee:
How about merging #21379 first?  framereader is depends on qt , I had to do some ugly hacks in this PR to make it work.

deanlee:
driver cam temporarily disabled. I don’t have the video file address for it

deanlee:
saved a lot of code after #21379

release/files_common
```diff
@@ -190,6 +190,8 @@ selfdrive/common/version.h
 
 selfdrive/common/framebuffer.h
 selfdrive/common/framebuffer.cc
+selfdrive/common/framereader.h
+selfdrive/common/framereader.cc
```
adeebshihadeh, line 194: why add these to the release files?

release/files_common
```diff
@@ -376,6 +378,8 @@ selfdrive/camerad/cameras/camera_common.h
 selfdrive/camerad/cameras/camera_common.cc
 selfdrive/camerad/cameras/camera_frame_stream.cc
 selfdrive/camerad/cameras/camera_frame_stream.h
+selfdrive/camerad/cameras/camera_replay.cc
+selfdrive/camerad/cameras/camera_replay.h
```
adeebshihadeh, line 382: same as above

deanlee:
`selfdrive/build release` will fail after remove them from release files

adeebshihadeh:
Hmm, guess we have to add them to the release files, since we'll delete frame stream anyway.

adeebshihadeh:
Let's leave framereader in the replay folder for now. Other than that, this should be good to merge.

PR #21244, opened 2021-06-12
UI: refactor SoftwarePanel

deanlee:
`SoftwarePanel`  is unnecessarily complicated.
by the way, it seems that it is not safe to create update button and `lastUpdateTimeLbl` in `showEvent`,  they needs to be created before the `fileChanged` signal is emitted.

### DIFF
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -207,11 +207,11 @@ SoftwarePanel::SoftwarePanel(QWidget* parent) : QWidget(parent) {
   gitBranchLbl = new LabelControl("Git Branch");
   gitCommitLbl = new LabelControl("Git Commit");
   osVersionLbl = new LabelControl("OS Version");
-  versionLbl = new LabelControl("Version", "", QString::fromStdString(Params().get("ReleaseNotes")).trimmed());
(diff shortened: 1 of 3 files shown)


selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -204,17 +204,40 @@ DevicePanel::DevicePanel(QWidget* parent) : QWidget(parent) {
...
+    std::system("pkill -1 -f selfdrive.updated");
+  }, "", this);
+
   QVBoxLayout *main_layout = new QVBoxLayout(this);
-  setLayout(main_layout);
+  QWidget *widgets[] = {versionLbl, lastUpdateLbl, updateBtn, gitBranchLbl, gitCommitLbl, osVersionLbl};
+  for (int i = 0; i < std::size(widgets); ++i) {
+    main_layout->addWidget(widgets[i]);
+    if (i < std::size(widgets) - 1) {
+      main_layout->addWidget(horizontal_line());
```
adeebshihadeh, line 228: we should really fix the horizontal line mess. we may want to build it into a layout, but for now maybe a function that inserts the lines into a populated vlyaout?

deanlee: added a new class ListWidget to do this https://github.com/commaai/openpilot/pull/21273

PR #21245, opened 2021-06-12
Update lane line indicators on car HUD to be linked to OP enabled state

alamo3:
The lane line indicators on the car HUD flicker a lot at times
since there are times when the lane probability fluctuates a lot
in places like parking lots. The purpose of the lane line indicators
on most of the car HUDs is to indicate whether the system detects
lane lines and if both are there, the system is available (mostly).
Since OP can be enabled pretty much anywhere, I think this change
makes sense.

Can be verified by enabling OP, lane indicators on HUD should turn on.
Otherwise, they are off.

Video of said issue can be found here: https://cdn.discordapp.com/attachments/524328425415245827/851598155585945639/20210607_190620.mp4

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -521,8 +521,10 @@ class Controls:
 
     right_lane_visible = self.sm['lateralPlan'].rProb > 0.5
     left_lane_visible = self.sm['lateralPlan'].lProb > 0.5
-    CC.hudControl.rightLaneVisible = bool(right_lane_visible)
-    CC.hudControl.leftLaneVisible = bool(left_lane_visible)
+
+    #Do this since it causes A LOT of flickering on the HUD at times.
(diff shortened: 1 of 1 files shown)


VirtuallyChris:
This is already how it is on Nidec Hondas, so either we accept this change to make all cars consistent, or I can open an issue for the Nidecs to make it match the behavior of other cars

mob9221:
The main reason for this change is to deal with the flickering issue that happens from time to time. Not sure how many people actually look at the lane line indicators but it can get pretty distracting at times if it keeps going on and off like in the video (car was parked in a parking lot). We can also make the change only apply to specific cars if necessary but not sure if thats a good route to go.

pd0wm:
We should probably remove the lane line indicators completely as we're moving to laneless. Won't be merging this, since there is also a bunch of car specific code that can be removed.

PR #21248, opened 2021-06-13
default carParams.stoppingControl to True

gregjhogan:


pd0wm:
Should we just change the default? Think this is a toyota quirk.

gregjhogan:
seems like a good idea to me, updated

PR #21251, opened 2021-06-13
common/util.cc: add unit test, fix bug in util::read_file

deanlee:

### DIFF
--- a/selfdrive/common/tests/test_util.cc
+++ b/selfdrive/common/tests/test_util.cc
@@ -19,7 +19,6 @@ TEST_CASE("util::read_file") {
 
   auto read_directory = [=](const std::string &dir, const std::string &path_suffix = "") {
     DIR *d = opendir(dir.c_str());
-    assert(d);
     struct dirent *de = nullptr;
     while ((de = readdir(d))) {
       if (!isalnum(de->d_name[0])) continue;
--- a/selfdrive/common/tests/test_util.cc
+++ b/selfdrive/common/tests/test_util.cc
@@ -9,6 +9,20 @@
 
 #include "selfdrive/common/util.h"
 
+std::string random_string(size_t length) {
+  auto randchar = []() -> char {
+    const char charset[] =
+        "0123456789"
+        "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
+        "abcdefghijklmnopqrstuvwxyz";
+    const size_t max_index = (sizeof(charset) - 1);
+    return charset[rand() % max_index];
+  };
+  std::string str(length, 0);
+  std::generate_n(str.begin(), length, randchar);
+  return str;
+}
+
 TEST_CASE("util::read_file") {
   auto read_file = [](const std::string &fn) -> std::string {
     std::ifstream f(fn);
@@ -17,13 +31,11 @@ TEST_CASE("util::read_file") {
     return buffer.str();
   };
 
-  auto read_directory = [=](const std::string &dir, const std::string &path_suffix = "") {
-    DIR *d = opendir(dir.c_str());
+  auto test_read_proc = [=]() {
+    DIR *d = opendir("/proc");
     struct dirent *de = nullptr;
     while ((de = readdir(d))) {
-      if (!isalnum(de->d_name[0])) continue;
-
-      const std::string path = util::string_format("%s/%s%s", dir.c_str(), de->d_name, path_suffix.c_str());
+      const std::string path = util::string_format("/proc/%s/cmdline", de->d_name);
       std::string ret1 = util::read_file(path);
       std::string ret2 = read_file(path);
       REQUIRE(ret1 == ret2);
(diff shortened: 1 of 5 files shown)


deanlee:
got am error in test case : `basic_string::_M_replace_aux`.

tellg returns -1 causing this problem.

selfdrive/common/tests/test_util.cc
```diff
@@ -0,0 +1,98 @@
...
+    for (auto &data : test_data) {
+      write(fd, data.data(), data.size() * sizeof(uint8_t));
+      file_content.insert(file_content.end(), data.begin(), data.end());
+      std::string ret = util::read_file(filename);
+      REQUIRE(ret.size() == file_content.size());
+      REQUIRE(memcmp(ret.data(), file_content.data(), file_content.size() * sizeof(uint8_t)) == 0);
+    }
+    close(fd);
+  }
+  SECTION("read non-exits file") {
```
pd0wm, line 92:
```suggestion
  SECTION("read non-existant file") {
```

selfdrive/common/tests/test_util.cc
```diff
@@ -0,0 +1,98 @@
...
+    for (int i = 0; i < 5; ++i) {
+      std::string ret = util::read_file(random_string(5));
```
pd0wm, line 94:
```suggestion
      std::string ret = util::read_file("/does_not_exist");
```

selfdrive/common/tests/test_util.cc
```diff
@@ -0,0 +1,98 @@
...
+      REQUIRE(ret == "");
```
pd0wm, line 95:
```suggestion
      REQUIRE(ret.empty());
```

selfdrive/common/tests/test_util.cc
```diff
@@ -0,0 +1,98 @@
...
+    while ((de = readdir(d))) {
+      const std::string path = util::string_format("/proc/%s/cmdline", de->d_name);
+      std::string ret1 = util::read_file(path);
+      std::string ret2 = read_file(path);
+      REQUIRE(ret1 == ret2);
+    }
+    closedir(d);
+  }
+  SECTION("read ascii file") {
+    std::vector<std::string> test_data;
```
pd0wm, line 54: Why not one long string?

selfdrive/common/tests/test_util.cc
```diff
@@ -0,0 +1,98 @@
...
+    for (auto &data : test_data) {
+      write(fd, data.c_str(), data.size());
+      file_content += data;
+      std::string ret = util::read_file(filename);
+      REQUIRE(ret == file_content);
+    }
+    close(fd);
+  }
+  SECTION("read binary file") {
+    std::vector<std::vector<uint8_t>> test_data;
```
pd0wm, line 73: Why a vector? An std::string can hold raw bytes, right?

selfdrive/common/tests/test_util.cc
```diff
@@ -0,0 +1,98 @@
...
+#include <algorithm>
+#include <climits>
+#include <random>
+#include <string>
+
+#define CATCH_CONFIG_MAIN
+#include "catch2/catch.hpp"
+#include "selfdrive/common/util.h"
+
+std::string random_string(size_t length) {
```
pd0wm, line 14: Do we really need random data? Why not a function that takes in a length, and an upper and lower bound on the byte values? `random_bytes` can be between 0 and 255, `random_string` is between 32 and 126.

But why this distinction between ascii and non ascii? Nothing in the code that we test does anything different for ascii.

selfdrive/common/tests/test_util.cc
deanlee: I used a single temp file to continue writing to it.

selfdrive/common/tests/test_util.cc
```diff
@@ -0,0 +1,64 @@
...
+    std::ifstream f(fn);
+    std::stringstream buffer;
+    buffer << f.rdbuf();
+    return buffer.str();
+  };
+
+  SECTION("read /proc") {
+    DIR *d = opendir("/proc");
+    struct dirent *de = nullptr;
+    while ((de = readdir(d))) {
```
pd0wm, line 33: Why not use `/proc/self/cmdline` so simplify the test? You can also simplify the test by just checking if it contains `test_util` instead of writing a 2nd implementation of `read_file`

selfdrive/common/tests/test_util.cc
pd0wm: What's wrong with an std::string of 64k?

selfdrive/common/tests/test_util.cc
deanlee: It is impossible to test that the open is successful, but tellg return -1 in this way

deanlee: It may be a file permission issue

pd0wm: What situation are you trying to catch with this test? A file you don't have permission to? I don't like the randomness of loping over `/proc`. If you need something without read access try `/proc/kmsg`, or `/etc/shadow` if you also want to test with a regular file.

deanlee: There are indeed many problems, I need to learn how to write a good test case:D

pd0wm: This looks so much cleaner :)

pd0wm:
Test looks pretty clean now. However, I could net get the test to fail when reverting the changes to util.cc, so think we're still missing a critical test case.

deanlee:
read a directory will cause that failure

deanlee:
read a directory will cause that failure

PR #21254, opened 2021-06-14
VW MQB: Add FW for 2016 Golf R

jyoung8607:
New FW values from a 2016 VW Golf R.

Most of the values are typical, but his tuner saw fit to graffiti over the ECU SW part number. 😠

**Dongle ID:** b5a5d867d5cd2bc1

Thanks to community Golf R owner edwyck.


adeebshihadeh:
Let's add a comment at the top of FW_VERSIONS for known modifications to the versions.

selfdrive/car/volkswagen/values.py
```diff
@@ -123,7 +129,7 @@ class CAR:
       b'\xf1\x878V0906264F \xf1\x890003',
       b'\xf1\x878V0906264L \xf1\x890002',
       b'\xf1\x878V0906264M \xf1\x890001',
-      b'\xf1\x878V09C0BB01 \xf1\x890001',  # Malformed ECU SW part number due to COBB/EQT tune
+      b'\xf1\x878V09C0BB01 \xf1\x890001',  # Corrupted by COBB/EQT tune
```
adeebshihadeh, line 132: remove the comment and just make a list of expected modifications in the comment

jyoung8607: Disagree:

1) I don't know that I can enumerate all expected modifications
2) If I have knowledge a particular FW entry is unreliable for vehicle ID and/or compatibility checks, I want to retain that

Frankly it pisses me off that I have to have that ECU entry at all. It's stupid and it's been several years since I've seen a reputable tuner do so. Having checked some other available scan data from Ross-Tech's forum, it looks like even COBB quit doing it on more recent tunes. I'm inquiring with the user to see if he can have COBB fix it so I don't have to.

jyoung8607:
How's that? I can't really write a regex that defines vandalism, but I can kinda write one that is less-likely-to-be vandalism.

I'm still having the end user work with COBB to repair this particular instance of stupid, but IDK if or when that will succeed.

adeebshihadeh:
I'll merge this, but would be great if we could remove the COBB fw version.

Issue #20675, opened 2021-04-13
Take Control immediately, communication issue between processes


defyent:
Twice today, at what seemed to be random times (first was within 10 minutes of the drive, and probably 5 minutes of initiating OP, second was about 30 minutes into a different drive, and within 10 minutes of initiating OP) I got the alarm "Take Control immediately, communication issue between processes". I took over immediately and disengaged OP, but was able to immediately reengage OP both times. 

**How to reproduce or log data**

Seemingly random times, nothing eventful happened to trigger the alarm that I noticed.

**Expected behavior**

Not to alarm

**Device/Version information (please complete the following information):**
 - Device: comma two
 - Dongle ID: cc5727706f6dd3c3
 - Route: [e.g. cc5727706f6dd3c3|2021-04-13--01-00-16
 - Timestamp: 20:40:42 NZST
 - Version: 0.8.3
 - Car make/model: Toyota Corolla Hatchback Hybrid 2018

**Additional context**

I've found the other bug reports that are closed from a few days ago, one mentioned commit 0dca8cb, I looked at the change and the code on my device and it seems my device already has this commit applied, "ubloxRaw" does not exist anywhere in the controlsd.py file.

speedking456:
I recently ran into this issue as well, as did @KDKimura. Apparently this is a long standing issue—pd0wm mentions it in my issue #20601—however only recently has this actually become an issue for me. Just trying to find similarity, but are you using a sim card in your device?

defyent:
Hrm okay, nope no sim card, wifi only.

speedking456:
> Hrm okay, nope no sim card, wifi only.

Alright well, pd0wm's solution of removing the sim card won't work for you. Are you able to go into developer of your comma two and check what commit you are running?

defyent:
9a0006fa0a

pd0wm:
Looks like `liveParameters` and `liveLocationKalman` lagged for about a second triggering the alert. My current best guess it this is caused by the garbage collector. This is turned off for for most processes like controlsd, but looks like it's still enabled for locationd and paramsd. 

We're currently rewriting locationd and paramsd in C++, which should of course solve this problem. I'll see if I can disable the garbage collector in the meantime.

![image](https://user-images.githubusercontent.com/1314752/114726593-d2cabd00-9d3d-11eb-9618-d1b858af3cfe.png)

pd0wm:
Think this specific case should be fixed now. Feel free to open a new issue when you run into the problem again.

PR #21257, opened 2021-06-14
Release GIL when calling C++ params functions

pd0wm:
Run updater continuously:
```python
import psutil

from selfdrive.updated import init_overlay, finalize_update

# Set low io priority (doesn't seem to affect the results much, but match updated)
proc = psutil.Process()
proc.ionice(psutil.IOPRIO_CLASS_BE, value=7)

init_overlay()
while True:
  finalize_update()
```

Try writing to params in the background. Observe this loop runs slower than 10 Hz indicating the GIL is not released.
```python
import time

from common.params import put_nonblocking
from common.realtime import set_realtime_priority

# set_realtime_priority(1)

t = time.time()
while True:
  dt = time.time() - t
  if dt > 0.2:
    print(dt)
  t = time.time()

  put_nonblocking("GitCommit", "abcdjkldsjflksfd")
  time.sleep(0.1)

```

Probably also the actual issue behind https://github.com/commaai/openpilot/issues/20675, although disabling the GC for paramsd was a good change as well.

Without these changes lags between 0.5 and 5 seconds are observed often. With the changes from this PR the situation is much better, but can still produce small lags (0.2-0.5) from time to time, not exactly sure yet where those come from. Pinning both scripts to a different core doesn't help.

By setting real time priority (even a very low one) this is fixed as well, but not sure if we want to do that for paramsd/locationd. Setting real time priority without the changes from this PR has no effect.

### DIFF
--- a/common/params_pxd.pxd
+++ b/common/params_pxd.pxd
@@ -17,12 +17,12 @@ cdef extern from "selfdrive/common/params.h":
     ALL
 
   cdef cppclass Params:
-    Params(bool)
-    Params(string)
+    Params(bool) nogil
+    Params(string) nogil
     string get(string, bool) nogil
-    bool getBool(string)
-    int remove(string)
-    int put(string, string)
-    int putBool(string, bool)
-    bool checkKey(string)
+    bool getBool(string) nogil
+    int remove(string) nogil
+    int put(string, string) nogil
+    int putBool(string, bool) nogil
+    bool checkKey(string) nogil
     void clearAll(ParamKeyType)
--- a/common/params_pyx.pyx
+++ b/common/params_pyx.pyx
@@ -57,13 +57,12 @@ cdef class Params:
 
     return key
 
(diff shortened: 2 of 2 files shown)


jyoung8607:
> ```python
> ...
> # Set low io priority (doesn't seem to affect the results much, but match updated)
> proc = psutil.Process()
> proc.ionice(psutil.IOPRIO_CLASS_BE, value=7)

The [NEOS kernel has the IO schedulers removed](https://github.com/commaai/android_kernel_comma_msm8996/blob/31f1524d0bac766ee71dde42886edceb6b2e7f2b/arch/arm64/configs/comma_defconfig#L35-L37) which results in FIFO and ignores ionice. If I had to guess, running at realtime CPU priority probably plasters over the problem by making the kernel prioritize its IO to avoid priority inversion. It would be good to have the CFQ scheduler.

gregjhogan:
You can use a cgroup to throttle I/O

PR #21259, opened 2021-06-14
sidebar: display full signal when network strength is GREAT

deanlee:


pd0wm:
Why the gap? Isn't 2, 3, 4, 5 cleaner? In that case you can also get rid of the map and do something along the lines of:
```c++
  setProperty("netStrength", deviceState.getNetworkStrength() > 0 ? deviceState.getNetworkStrength()  + 1 : 0);
```

PR #21261, opened 2021-06-14
Don't check RAM/disk in simulation

### LINKED ISSUE #21252: [simulation] "Low Memory / Reboot your Device" in carla simulator
**Describe the bug**

OP complains about low memory when reading from CARLA.

I followed the instructions [here](https://github.com/commaai/openpilot/wiki/CARLA) to download and run CARLA and OpenPilot. The emulator screen pops up, briefly flashes some orange warning, then says "Low Memory / Reboot your Device".

Additionally, if I hit the 1 key in the terminal, the message changes to "openpilot Unavailable / Low Memory: Reboot Your Device" for maybe 5 seconds.

I dutifully click into the menu and click "Reboot" > "Ok", but this only clears the "are you sure you want to reboot" message, without rebooting.

**Expected behavior**

I should be able to hit the 1 key and see the simulator start driving.

**Additional context**

I tried adding `--memory 8G` and `--kernel-memory 1G` and `--memory-swap 8G` and `--shm-size 5G` to `start_openpilot_docker.sh`. I also tried starting with `./start_openpilot_docker.sh --low_quality`, although CARLA seems to run fine, both here and when downloaded directly from carla.org

Host RAM usage during this never goes above 21.4G (out of 31.3 in htop).

Machine has 32G of installed RAM, GTX 1080 GPU with driver 460.80, and i7-4770k CPU.

Operating system: Ubuntu 20.04

pd0wm:
Fixes #21252 

These percentages don't work on PC since ram/disk can have very different sizes.

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -23,7 +23,7 @@ from selfdrive.controls.lib.alertmanager import AlertManager
 from selfdrive.controls.lib.vehicle_model import VehicleModel
 from selfdrive.controls.lib.longitudinal_planner import LON_MPC_STEP
 from selfdrive.locationd.calibrationd import Calibration
-from selfdrive.hardware import HARDWARE, TICI
+from selfdrive.hardware import HARDWARE, TICI, PC
 
 LDW_MIN_SPEED = 31 * CV.MPH_TO_MS
 LANE_DEPARTURE_THRESHOLD = 0.1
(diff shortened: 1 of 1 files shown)


pd0wm:
@adeebshihadeh Looks like one of our test routes in not engaging due to low disk/memory. Do you want to:
 - Change the route
 - Change the reference
 - Change this PR to only check for `SIMULATION` instead of `PC`

pd0wm:
Will just check simulation for now, but not ideal.

 - We need to fix the test route
 - We need to put absolute values for ram/disk in the deviceState packets and move the thresholds to the HW abstraction layer.

PR #21264, opened 2021-06-14
make locationd and paramsd RT

adeebshihadeh:


pd0wm:
@adeebshihadeh Do you want to verify this in the testing closet before merging?

PR #21267, opened 2021-06-15
new planning api

haraschax:


selfdrive/controls/lib/longcontrol.py
```diff
@@ -12,6 +14,7 @@
 BRAKE_STOPPING_TARGET = 0.5  # apply at least this amount of brake to maintain the vehicle stationary
 
 RATE = 100.0
+DEFAULT_LONG_LAG = 0.5
```
ShaneSmiskol, line 17: I'm pretty sure this changes based on speed, for my Corolla 0.5s was too much at low speed and it jerked much worse than it normally does. 0.2 at 10 mph to 0.5 at 70 mph is what I settled on, but it most likely varies by car.

HaraldSchafer: Hmm ok thanks, I just want to use a default value for now. I will do 0.3 then. And later we can implement car specific lags like we do for lateral control.

PR #21271, opened 2021-06-15
threaded CameraView

deanlee:
perform updating and rendering on worker threads.


deanlee:
There is a more efficient way to use QWindow instead of QOpenglWindow, but I haven't found how to display a overlay QWidget on navtive window.

PR #21273, opened 2021-06-15
qt: new class ListWidget

deanlee:
remove the horizontal line mess.


deanlee:
added a QSpaceItem to AbstractControl to make sure the title bar will hold it's position while expanding

selfdrive/ui/qt/widgets/controls.h
```diff
@@ -127,3 +126,28 @@ class ParamControl : public ToggleControl {
 private:
   Params params;
 };
+
+class ListWidget : public QWidget {
```
adeebshihadeh, line 130: why not make this a layout?

deanlee: I think do painting  is better, performance is better too.It can save a lot of styled horizontal_line widgets. The layouting is faster, and the spacing between controls can be controlled more precisely.

selfdrive/ui/qt/widgets/controls.h
```diff
@@ -127,3 +126,28 @@ class ParamControl : public ToggleControl {
...
+  Q_OBJECT
+ public:
+  explicit ListWidget(QWidget *p = nullptr) : QWidget(p), layout_(this) {
```
adeebshihadeh, line 133:
```suggestion
  explicit ListWidget(QWidget *parent = 0) : QWidget(parent), layout_(this) {
```
let's keep it consistent

selfdrive/ui/qt/widgets/controls.h
adeebshihadeh: we should get the better perf if make a horizontal line widget drawn with paintEvent. the reason I'd like this to be a layout is because this looks like pseudo-Qt code with a weird widget class that has the API of a layout, but it really seems like there's no good reason for that. I think we can get what we want, and it can still feel like familiar Qt code

deanlee: it's more looks like a QListWidget if we remove the function addLayout. what do you think? addLayout is only used in one place

deanlee: An extra line widget is difficult to control the spacing between controls. a bool variable also needs to be added to let the caller tell us which one is the last.we can think this widget as a  styled qlistwidget with bottom line between items

adeebshihadeh:
Any reason this is still a draft? Can you also rebase?

adeebshihadeh:
The title bar is still moving and flashing when expanding and collapsing the description.

deanlee:
@adeebshihadeh  It should be ok now

PR #21277, opened 2021-06-15
Qt: new class ElidedLabel

deanlee:

### DIFF
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -250,8 +250,8 @@ void SoftwarePanel::updateLabels() {
   updateBtn->setText("CHECK");
   updateBtn->setEnabled(true);
   gitBranchLbl->setText(QString::fromStdString(params.get("GitBranch")));
-  gitCommitLbl->setText(QString::fromStdString(params.get("GitCommit")).left(10));
-  osVersionLbl->setText(QString::fromStdString(Hardware::get_os_version()).trimmed());
+  gitCommitLbl->setText(QString::fromStdString(params.get("GitCommit")));
+  osVersionLbl->setText(QString::fromStdString(Hardware::get_os_version()));
 }
 
 QWidget * network_panel(QWidget * parent) {
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -250,8 +250,8 @@ void SoftwarePanel::updateLabels() {
   updateBtn->setText("CHECK");
   updateBtn->setEnabled(true);
   gitBranchLbl->setText(QString::fromStdString(params.get("GitBranch")));
-  gitCommitLbl->setText(QString::fromStdString(params.get("GitCommit")));
-  osVersionLbl->setText(QString::fromStdString(Hardware::get_os_version()));
+  gitCommitLbl->setText(QString::fromStdString(params.get("GitCommit")).left(10));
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/util.h
```diff
@@ -13,6 +13,14 @@ inline QString getBrandVersion() {
   return getBrand() + " v" + QString::fromStdString(Params().get("Version")).left(14).trimmed();
 }
 
+inline QString truncate(const QString& str, size_t width, bool show_ellipsis = true) {
+  if (str.length() > width) {
+    return show_ellipsis ? str.left(width) + "..." : str.left(width);
```
pd0wm, line 18:
```suggestion
    return show_ellipsis ? str.left(width) + "…" : str.left(width);
```

pd0wm:
How did you choose 18? Isn't there space for more?

deanlee:
maybe we should do truncate in AbstractControl later

deanlee:
> How did you choose 18? Isn't there space for more?
can adjust to 22

adeebshihadeh:
Maybe we should make a nice helper function to fit some text in a max width within a range of font sizes?

deanlee:
maybe a helper function like this.but it's hard to use.will increase a lot of code

```
inline QString elideString(QWidget *w, const QString &str, int width = 0, Qt::TextElideMode mode = Qt::ElideRight) {
  QFontMetrics fm(w->font());
  return fm.elidedText(str, mode, width <= 0 ? w->width() : width);
}
```

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -264,8 +264,8 @@ void SoftwarePanel::updateLabels() {
   updateBtn->setText("CHECK");
   updateBtn->setEnabled(true);
   gitBranchLbl->setText(QString::fromStdString(params.get("GitBranch")));
-  gitCommitLbl->setText(QString::fromStdString(params.get("GitCommit")).left(10));
-  osVersionLbl->setText(QString::fromStdString(Hardware::get_os_version()).trimmed());
+  gitCommitLbl->setText(QString::fromStdString(params.get("GitCommit")));
+  osVersionLbl->setText(QString::fromStdString(Hardware::get_os_version()));
```
adeebshihadeh, line 268: short hash is better than cutoff long one
```suggestion
  gitCommitLbl->setText(QString::fromStdString(params.get("GitCommit")).left(10));
  osVersionLbl->setText(QString::fromStdString(Hardware::get_os_version()).trimmed());
```

selfdrive/ui/qt/widgets/controls.cc
```diff
@@ -85,3 +87,14 @@ ButtonControl::ButtonControl(const QString &title, const QString &text, const QS
   QObject::connect(&btn, &QPushButton::released, this, &ButtonControl::released);
   hlayout->addWidget(&btn);
 }
+
+void ElidedLabel::paintEvent(QPaintEvent *event) {
+  QFrame::paintEvent(event);
+
+  QPainter painter(this);
```
adeebshihadeh, line 94: can't we just make a custom QLabel and override setText?

deanlee: We may not know the width of qlabel when calling settext

adeebshihadeh: we can use the showEvent

deanlee:
override resizeEvent and paintEvent, so that it can be displayed correctly in all situation

PR #21278, opened 2021-06-15
Networking refactor

sshane:
It takes 100ms to poll for all networks, getting AP strength, ssid, etc. This moves that into a thread so it doesn't block the main UI thread.

Todo:
- [x] Abstract wifi Networking code into WifiManager
- [x] Scanning takes multiple seconds and is non-blocking, now we wait for a DBus signal to update the UI!
- [ ] Lock connect buttons when pressed so you can't crash the UI by spamming it (this was also possible before)
  - [ ] If you're waiting for a wrong password connect to time out and click another connect button, it will prompt for a password for the secondary known network. This should fix that.
- [ ] Update UI as updates are made available from the wifi manager thread, no waiting for 5 seconds to update


pd0wm:
Do you want to poll this? Dbus + Qt also has ways to get signals on certain changes.

ShaneSmiskol:
I saw that with the `change` function, I'm going to update UI based on the signals it receives, with some minimum update rate imposed in the main loop if there's no updates/changes.

PR #21279, opened 2021-06-15
Added 2018 Accord 2.0T Touring transmission FW Vers

tanmayparikh:
**Description**
The 2018 Honda Accord 2.0T Touring with a 10speed auto transmission (Canadian Model) showed the following warning:
`"WARNING: No Exact Match on Car Model"`

It was failing the fingerprint match on boot in the `startup_fuzzy_fingerprint_alert()` function

Compared the fingerprint from useradmin from my latest drive and the `Ecu.transmission` for my car in FW_VERSION was missing.
I have added it to the fingerprint check

**Verification**
* Applied the fix
* Rebooted the car
* Warning gone

EDIT: dongle ID 1e3df82cadc6591b


adeebshihadeh:
What's your dongle ID?

tanmayparikh:
> What's your dongle ID?

Updated the description

PR #21282, opened 2021-06-15
Qt: cleanup includes

deanlee:
- remove unused includes
- only keep the minimum necessary includes in header file
- remove include qtwidgets,it's too big


selfdrive/ui/qt/onroad.h
```diff
@@ -2,13 +2,16 @@
 
 #include <map>
 
+#include <QOpenGLFunctions>
+#include <QOpenGLWidget>
 #include <QSoundEffect>
-#include <QtWidgets>
```
adeebshihadeh, line 6: does this actually make a difference? anything unused should get optimized out

deanlee: From what I understand, compiler will read all includes and  "insert" their content  to the source file,then do the compile and optimize steps.it's our duty to cleanup the includes

PR #21288, opened 2021-06-16
UI: skip SetupWidget if the dongle id is not valid

maxime-desroches:
Since #21275 openpilot does not show the SetupWidget if the dongle id is not valid, therefore we can skip most of the constructor + PrimeUserWidget, PrimeAdWidget and PairingQRWidget


adeebshihadeh:
I don't want to start adding this everywhere for these minor speedups. The other checks are purposeful since they prevent unnecessary network requests.

PR #21297, opened 2021-06-16
VW MQB: Simplify adding new cars (1 of 2)

jyoung8607:
We don't need different DBCs for every car, so lose the extra steps and extra code needed to onboard new models. If/when PQ comes onboard, those will be using entirely different CAN parser and packer functions anyway.

### DIFF
--- a/selfdrive/car/volkswagen/carcontroller.py
+++ b/selfdrive/car/volkswagen/carcontroller.py
@@ -10,7 +10,7 @@ class CarController():
   def __init__(self, dbc_name, CP, VM):
     self.apply_steer_last = 0
 
-    self.packer_pt = CANPacker(dbc_dict('vw_mqb_2010', None)["pt"])
+    self.packer_pt = CANPacker("vw_mqb_2010")
 
     self.hcaSameTorqueCount = 0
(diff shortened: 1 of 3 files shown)


selfdrive/car/volkswagen/carcontroller.py
```diff
@@ -10,7 +10,7 @@ class CarController():
   def __init__(self, dbc_name, CP, VM):
     self.apply_steer_last = 0
 
-    self.packer_pt = CANPacker(DBC[CP.carFingerprint]['pt'])
+    self.packer_pt = CANPacker("vw_mqb_2010")
```
pd0wm, line 13: Put string in values.py and import?

jyoung8607: Changed to something friendly to adding PQ later.

selfdrive/car/volkswagen/values.py
```diff
@@ -26,6 +25,9 @@ class CANBUS:
   pt = 0
   cam = 2
 
+class DBC:
+  mqb = "vw_mqb_2010"
```
adeebshihadeh, line 29: add a comment that this covers all VW MQB vehicles

PR #21299, opened 2021-06-16
VW MQB: Simplify adding new cars (2 of 2)

jyoung8607:
The `get_params()` if-elif chain is starting to get a little out of control, and it's only going to get worse. Refactor the ugly key-value lookup pattern into an actual key-value lookup pattern.

It's not clear if we'll end up needing per-car tunes at all, so I haven't added a tune dict at this time. If we do end up needing per-car tunes, I'll have to write something to look at a couple bytes of the 0xF182 parameterization identifier, in order to know what sub-model we have (e.g., base Golf vs Golf R, I'm not making CARs for all those) and do a separate tune-table dict lookup from there.

Additional cleanup and consolidation to the top of `get_params()` for better alignment with a future PQ port. Fixed the scope of the new enableBsm check to be MQB-only, PQ will use a different message.

### DIFF
--- a/selfdrive/car/volkswagen/interface.py
+++ b/selfdrive/car/volkswagen/interface.py
@@ -49,11 +49,11 @@ class CarInterface(CarInterfaceBase):
     # Optional per-CAR attributes, with defaults
     ret.steerActuatorDelay = ATTRIBUTES[candidate].setdefault("steer_actuator_delay", 0.05)  # Seems good for most MQB
     ret.steerRatio = ATTRIBUTES[candidate].setdefault("steer_ratio", 15.6)  # Updated by params learner
+    tire_stiffness_factor = ATTRIBUTES[candidate].setdefault("tire_stiffness", 1.0)  # Updated by params learner
 
     # Tuning values, currently using the same tune for all MQB
     # If we need to tune individual models, we'll need a dict lookup by EPS parameterization, not just CAR
     ret.steerRateCost = 1.0
-    tire_stiffness_factor = 1.0  # Updated by params learner
     [ ret.lateralTuning.pid.kpBP, ret.lateralTuning.pid.kiBP,
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV,
       ret.lateralTuning.pid.kf ] = ([0.], [0.], [0.6], [0.2], 0.00006)
--- a/selfdrive/car/volkswagen/interface.py
+++ b/selfdrive/car/volkswagen/interface.py
@@ -1,6 +1,6 @@
 from cereal import car
 from selfdrive.swaglog import cloudlog
-from selfdrive.car.volkswagen.values import CAR, BUTTON_STATES, TransmissionType, GearShifter
+from selfdrive.car.volkswagen.values import ATTRIBUTES, BUTTON_STATES, TransmissionType, GearShifter
 from selfdrive.car import STD_CARGO_KG, scale_rot_inertia, scale_tire_stiffness, gen_empty_fingerprint
 from selfdrive.car.interfaces import CarInterfaceBase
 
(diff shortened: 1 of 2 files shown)


selfdrive/car/volkswagen/interface.py
```diff
@@ -22,115 +22,41 @@ def compute_gb(accel, speed):
...
+    ret.steerActuatorDelay = ATTRIBUTES[candidate].setdefault("steer_actuator_delay", 0.05)  # Seems good for most MQB
+    ret.steerRatio = ATTRIBUTES[candidate].setdefault("steer_ratio", 15.6)  # Updated by params learner
+    tire_stiffness_factor = ATTRIBUTES[candidate].setdefault("tire_stiffness", 1.0)  # Updated by params learner
 
-    ret.enableCamera = True
-    ret.enableBsm = 0x30F in fingerprint[0]
+    # Tuning values, currently using the same tune for all MQB
+    # If we need to tune individual models, we'll need a dict lookup by EPS parameterization, not just CAR
+    ret.steerRateCost = 1.0
+    [ ret.lateralTuning.pid.kpBP, ret.lateralTuning.pid.kiBP,
```
adeebshihadeh, line 57: revert this

selfdrive/car/volkswagen/values.py
```diff
@@ -71,6 +71,26 @@ class CAR:
   SKODA_SUPERB_MK3 = "SKODA SUPERB 3RD GEN"   # Chassis 3V/NP, Mk3 Skoda Superb and variants
   SKODA_OCTAVIA_MK3 = "SKODA OCTAVIA 3RD GEN" # Chassis NE, Mk3 Skoda Octavia and variants
 
+# Required attributes: mass, wheelbase
+# Optional attributes: steer_actuator_delay, steer_ratio, tire_stiffness
+
+ATTRIBUTES = {
```
adeebshihadeh, line 77: not sure how I feel about this, since it breaks the pattern we use for all the other cars

gregjhogan: Another alternative would be to make these classes and use inheritance to share common values, then load the appropriate class by name and call a method on the class (implemented in a base class everything inherits from) which returns a `CarParams` instance.  The class constructor can take params to support as much flexibility as we like (e.g. `eps_modified` for different tuning) but in general I think the value of doing something like this is eliminating future bugs by eliminating branch logic as much as possible.

selfdrive/car/volkswagen/interface.py
jyoung8607: Ok, I can undo the lateralTuning.pid line rollup. Was a remnant from when I did have a tune-table but decided against.

selfdrive/car/volkswagen/values.py
jyoung8607: Yes, it's different. However, a [13-going-on-32 item](https://github.com/commaai/openpilot/discussions/20346) if-elif chain (before PQ, before MQBevo, before MEB) seems like an undesirable anti-pattern.

Nothing says other ports couldn't be adapted to use a similar form. In fact, the ATTRIBUTE reading code I proposed could be centralized into `selfdrive/car/interfaces.py` such that it gets used to set up any car-ports that use ATTRIBUTE and leaves alone the ones that haven't been converted.

I am not married to this particular implementation, and am also willing to discuss something like @gregjhogan's proposal. I'm just not a fan of what's there today.

jyoung8607: Forgot to mention: the if-elif chain style has no protection against falling off the end of get_params() with unset parameters if someone forgets or mistypes a car model name. My implementation requires no touches to interface.py at all to add a new car, and will fail earlier and more clearly if someone forgets to add the right ATTRIBUTEs.

adeebshihadeh:
This feels like a very intermediate refactor that breaks the pattern used everywhere else. Discussed this with Willem, and we agree that the current abstraction isn't great, but this isn't merge-able as is. We'll do a nice refactor of this ourselves at some point, but feel free open to open a new PR to propose a new refactor too.,

PR #21306, opened 2021-06-16
Split Civic Sedan and Coupe in readme

VirtuallyChris:
Looking at the VWs as an example, we include each 'model' in its own line of the readme even if they all identify as the same. This would fix the only car model where we show two in one line


pd0wm:
Why not order them like coupe, coupe, sedan, sedan?

VirtuallyChris:
Oops you're right, that makes way more sense

PR #21310, opened 2021-06-17
Scan for networks on UI init

sshane:
On start up ***only*** we request a scan from NM which when complete a signal is sent to `propertyChange` to update the networks and UI. Without waiting, we only see the current connection and have to wait 5 seconds to see the rest, which takes longer.

### DIFF
--- a/selfdrive/ui/qt/offroad/networking.cc
+++ b/selfdrive/ui/qt/offroad/networking.cc
@@ -23,27 +23,19 @@ void NetworkStrengthWidget::paintEvent(QPaintEvent* event) {
 
 // Networking functions
 
-Networking::Networking(QWidget* parent, bool is_setup) : QWidget(parent), is_setup(is_setup) {
+Networking::Networking(QWidget* parent, bool show_advanced) : QWidget(parent), show_advanced(show_advanced) {
   main_layout = new QStackedLayout(this);
 
-  QLabel* scanning = new QLabel("Scanning for networks");
-  scanning->setAlignment(Qt::AlignCenter);
-  scanning->setStyleSheet(R"(font-size: 65px;)");
-  main_layout->addWidget(scanning);
+  QLabel* warning = new QLabel("Network manager is inactive!");
+  warning->setAlignment(Qt::AlignCenter);
+  warning->setStyleSheet(R"(font-size: 65px;)");
 
-  if (is_setup) {
-    while (!ui_setup_complete) {
-      attemptInitialization();
-      qDebug() << "Attempted initialization";
-    }
-    refresh(true);
-  } else {
-    attemptInitialization();
-    wifi->requestScan();
-  }
+  main_layout->addWidget(warning);
 
+  attemptInitialization();
+  wifi->requestScan();
   QTimer* timer = new QTimer(this);
-  QObject::connect(timer, &QTimer::timeout, this, [=] { refresh(false); });
+  QObject::connect(timer, &QTimer::timeout, this, &Networking::refresh);
   timer->start(5000);
 }
 
@@ -60,7 +52,7 @@ void Networking::attemptInitialization() {
 
   QWidget* wifiScreen = new QWidget(this);
   QVBoxLayout* vlayout = new QVBoxLayout(wifiScreen);
-  if (!is_setup) {
+  if (show_advanced) {
     QPushButton* advancedSettings = new QPushButton("Advanced");
     advancedSettings->setStyleSheet("margin-right: 30px;");
     advancedSettings->setFixedSize(350, 100);
@@ -99,10 +91,8 @@ void Networking::attemptInitialization() {
   ui_setup_complete = true;
 }
 
-void Networking::refresh(bool force) {
-  qDebug() << "Refresh, force:" << force;
-  if (!this->isVisible() && !force) {
-    qDebug() << "Not refreshing";
+void Networking::refresh() {
+  if (!this->isVisible()) {
     return;
   }
   if (!ui_setup_complete) {
--- a/selfdrive/ui/qt/offroad/networking.cc
+++ b/selfdrive/ui/qt/offroad/networking.cc
@@ -34,9 +34,9 @@ Networking::Networking(QWidget* parent, bool show_advanced) : QWidget(parent), s
 
   attemptInitialization();
   wifi->requestScan();
-  timer = new QTimer(this);
-  QObject::connect(timer, &QTimer::timeout, this, &Networking::refresh);
-  timer->setInterval(5000);
+  refreshTimer = new QTimer(this);
+  QObject::connect(refreshTimer, &QTimer::timeout, this, &Networking::refresh);
+  refreshTimer->setInterval(5000);
 }
 
 void Networking::attemptInitialization() {
@@ -124,11 +124,11 @@ void Networking::wrongPassword(const QString &ssid) {
 }
 
 void Networking::hideEvent(QHideEvent* event) {
-  timer->stop();
+  refreshTimer->stop();
 }
 
 void Networking::showEvent(QShowEvent* event) {
-  timer->start();
+  refreshTimer->start();
 }
 
 // AdvancedNetworking functions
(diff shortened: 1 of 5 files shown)


selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -124,6 +123,14 @@ void Networking::wrongPassword(const QString &ssid) {
   }
 }
 
+void Networking::hideEvent(QHideEvent* event) {
+  timer->stop();
```
adeebshihadeh, line 127: rename to refreshTimer

selfdrive/ui/qt/offroad/wifiManager.h
```diff
@@ -75,10 +75,11 @@ class WifiManager : public QWidget {
...
-  void change(unsigned int new_state, unsigned int previous_state, unsigned int change_reason);
 signals:
   void wrongPassword(const QString &ssid);
-  void successfulConnection(const QString &ssid);
-  void refresh();
+  void refreshed();
+
+private slots:
+  void state_change(unsigned int new_state, unsigned int previous_state, unsigned int change_reason);
+  void property_change(const QString &interface, const QVariantMap &props, const QStringList &invalidated_props);
```
adeebshihadeh, line 84: camel case

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -75,13 +75,13 @@ WifiManager::WifiManager(QWidget* parent) : QWidget(parent) {
   }
 
   QDBusInterface nm(nm_service, adapter, device_iface, bus);
-  bus.connect(nm_service, adapter, device_iface, "StateChanged", this, SLOT(change(unsigned int, unsigned int, unsigned int)));
+  bus.connect(nm_service, adapter, device_iface, "StateChanged", this, SLOT(state_change(unsigned int, unsigned int, unsigned int)));
+  bus.connect(nm_service, adapter, props_iface, "PropertiesChanged", this, SLOT(property_change(QString, QVariantMap, QStringList)));
```
adeebshihadeh, line 79: use function pointers

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -291,10 +291,10 @@ void WifiManager::forgetNetwork(const QString &ssid) {
   }
 }
 
-void WifiManager::request_scan() {
+void WifiManager::requestScan() {
   QDBusInterface nm(nm_service, adapter, wireless_device_iface, bus);
   nm.setTimeout(dbus_timeout);
-  nm.call("RequestScan",  QVariantMap());
+  nm.call("RequestScan",  QVariantMap());  // a signal is sent to WifiManager::property_change once the scan is complete
```
adeebshihadeh, line 297: should be clear without the comment that this is what happens

selfdrive/ui/qt/offroad/wifiManager.cc
sshane: `QDBusConnection` only supports `const char *slot` for the slot.

sshane:
I'll split the network manager inactive code to another PR in a bit!

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -479,8 +488,6 @@ bool WifiManager::tetheringEnabled() {
 
 void WifiManager::changeTetheringPassword(const QString &newPassword) {
   tetheringPassword = newPassword;
-  if (isKnownNetwork(tethering_ssid.toUtf8())) {
-    forgetNetwork(tethering_ssid.toUtf8());
-  }
+  forgetNetwork(tethering_ssid.toUtf8());
```
adeebshihadeh, line 491: why did we check this before?

sshane: I added it forgetting that I also added the check in `forgetNetwork` for stuff like this. The previous clear_connections command just iterated through all the connections and performed no action if it didn't exist.

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -89,8 +91,8 @@ void Networking::attemptInitialization() {
   ui_setup_complete = true;
 }
 
-void Networking::refresh() {
-  if (!this->isVisible()) {
+void Networking::refresh(bool force) {
+  if (!this->isVisible() && !force) {
```
adeebshihadeh, line 95: just check ui_setup_complete instead

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -29,24 +29,27 @@ Networking::Networking(QWidget* parent, bool show_advanced) : QWidget(parent), s
...
   warning->setStyleSheet(R"(font-size: 65px;)");
-
   main_layout->addWidget(warning);
 
-  QTimer* timer = new QTimer(this);
-  QObject::connect(timer, &QTimer::timeout, this, &Networking::refresh);
-  timer->start(5000);
-  attemptInitialization();
+  if (attemptInitialization()) {
+    wifi->requestScan();
```
adeebshihadeh, line 35: move this to attemptInitialization

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -26,23 +26,22 @@ void NetworkStrengthWidget::paintEvent(QPaintEvent* event) {
 Networking::Networking(QWidget* parent, bool show_advanced) : QWidget(parent), show_advanced(show_advanced) {
   main_layout = new QStackedLayout(this);
 
-  QLabel* warning = new QLabel("Network manager is inactive!");
+  QLabel* warning = new QLabel("Scanning for networks");
```
adeebshihadeh, line 29:
```suggestion
  QLabel* warning = new QLabel("Network manager is inactive!");
```

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -87,10 +88,11 @@ void Networking::attemptInitialization() {
   )");
   main_layout->setCurrentWidget(wifiScreen);
   ui_setup_complete = true;
+  wifi->requestScan();
 }
 
 void Networking::refresh() {
-  if (!this->isVisible()) {
+  if (!this->isVisible() && !wifi->firstRefresh) {
```
adeebshihadeh, line 95: just use ui_setup_complete

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -87,10 +88,11 @@ void Networking::attemptInitialization() {
   )");
   main_layout->setCurrentWidget(wifiScreen);
   ui_setup_complete = true;
+  wifi->requestScan();
```
adeebshihadeh, line 91: switch the order here so you can use this in refresh()

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -261,6 +263,8 @@ void WifiUI::refresh() {
 
 void WifiUI::handleButton(QAbstractButton* button) {
   QPushButton* btn = static_cast<QPushButton*>(button);
+  btn->setDisabled(true);
+  btn->setText("Connecting");
   Network n = wifi->seen_networks[connectButtons->id(btn)];
```
adeebshihadeh, line 268: remove until we do this properly

selfdrive/ui/qt/offroad/networking.cc
sshane: So we want to request the scan immediately on start up and update the networks in the background once, so that means two calls to Networking::refresh if we use `ui_setup_complete`, and the second call won't allow the update in the background since `ui_setup_complete` will be true the second time. I'll think of a cleaner way though now that we're waiting for the scan all the time

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -99,6 +101,10 @@ void Networking::refresh() {
       return;
     }
   }
+  wifi->requestScan();
+}
+
+void Networking::refresh() {
   wifiWidget->refresh();
   an->refresh();
 }
```
sshane, line 110: We can leave out `firstRefresh` if we believe that NM will never send us a scan complete signal (`LastScan`) without us requesting it, since we don't check that Networking is visible here (so we can update once in the background). It can block UI for 70 ms+ if NM ever does though

selfdrive/ui/qt/offroad/networking.cc
sshane: It was non-blocking so the order wouldn't have mattered

Issue #21266, opened 2021-06-14
Update alert incorrectly supersedes overtemp alert


jyoung8607:
**Describe the bug**

When attempting to go onroad with a hot device that has an update available, the update alert displays instead of the overtemp alert. It keeps displaying until the device cools down, at which point it goes onroad.

**How to reproduce or log data**

1) Have an update available
2) Be hot
3) Start car
4) Observe device correctly refuse to go onroad until it cools down, but fail to properly tell the user why

**Expected behavior**

In this situation, the overtemp alert should supersede the update available alert.

**Device/Version information (please complete the following information):**
 - Dongle ID: 7aae592fc08e895f
 - Route: 7aae592fc08e895f|2021-06-14--16-20-37
 - Timestamp: beginning of the route, and prior
 - Version: day-old master + specialsauce (I was still running the branch from #21253)
 - Car make/model: 2018 Volkswagen Golf R

PR #21312, opened 2021-06-17
OffroadHome: merge stylesheets

deanlee:


selfdrive/ui/qt/home.cc
```diff
@@ -139,11 +137,23 @@ OffroadHome::OffroadHome(QWidget* parent) : QFrame(parent) {
...
   setStyleSheet(R"(
+    * {
+     color: white;
+    }
     OffroadHome {
       background-color: black;
     }
-    * {
-     color: white;
+    OffroadHome>QPushButton {
```
adeebshihadeh, line 146: this is really only for the update button, the selector should be more specific

selfdrive/ui/qt/home.cc
```diff
@@ -138,11 +137,23 @@ OffroadHome::OffroadHome(QWidget* parent) : QFrame(parent) {
...
   setStyleSheet(R"(
+    * {
+     color: white;
+    }
     OffroadHome {
       background-color: black;
     }
-    * {
-     color: white;
+    QPushButton#alert_notification {
```
adeebshihadeh, line 146:
```suggestion
    #alert_notification {
```

adeebshihadeh:
Want to fix #21266 too?

deanlee:
Ok, let me check

adeebshihadeh:
Hmm, think it's also fine to separate the update and alert buttons.

deanlee:
I'll fix it in another PR

PR #21318, opened 2021-06-17
input.cc: close dialog when parent is hidden

deanlee:


selfdrive/ui/qt/widgets/input.cc
```diff
@@ -5,7 +5,20 @@
...
-InputDialog::InputDialog(const QString &prompt_text, QWidget *parent) : QDialog(parent) {
+QDialogBase::QDialogBase(QWidget *parent) : QDialog(parent) {
+  Q_ASSERT(parent != nullptr);
+  parent->installEventFilter(this);
+}
+
+bool QDialogBase::eventFilter(QObject *o, QEvent *e) {
+  if (o == parent() && e->type() == QEvent::Hide) {
+    close();
+    return false;
```
adeebshihadeh, line 16: do we need to return here?

selfdrive/ui/qt/widgets/input.cc
```diff
@@ -5,7 +5,20 @@
...
 
-InputDialog::InputDialog(const QString &prompt_text, QWidget *parent) : QDialog(parent) {
+QDialogBase::QDialogBase(QWidget *parent) : QDialog(parent) {
+  Q_ASSERT(parent != nullptr);
+  parent->installEventFilter(this);
+}
+
+bool QDialogBase::eventFilter(QObject *o, QEvent *e) {
+  if (o == parent() && e->type() == QEvent::Hide) {
+    close();
```
adeebshihadeh, line 15: should we reject instead?

selfdrive/ui/qt/widgets/ssh_keys.cc
```diff
@@ -26,7 +26,7 @@ SshControl::SshControl() : AbstractControl("SSH Keys", "Warning: This grants SSH
 
   QObject::connect(&btn, &QPushButton::released, [=]() {
     if (btn.text() == "ADD") {
-      QString username = InputDialog::getText("Enter your GitHub username");
+      QString username = InputDialog::getText("Enter your GitHub username", 39, this);
```
adeebshihadeh, line 29: min length added by mistake?

adeebshihadeh (changes requested):
nice cleanup

selfdrive/ui/qt/widgets/ssh_keys.cc
deanlee: github limits the username to 39 characters.

pd0wm: This is a minimum length (e.g. for wifi passwords).

deanlee: ....my bad

adeebshihadeh:
Nice, I really like this.

PR #21322, opened 2021-06-18
HomeWindow: small cleanup

deanlee:
@adeebshihadeh OffroadHome will be refreshed on showEvent. It seems that this connection is unnecessary,am I right? In addition, I don’t understand why we need to refresh alerts when open settings


adeebshihadeh:
Yes, the refresh on openSettings is unnecessary.

PR #21324, opened 2021-06-18
CARLA: send frames over visionipc

adeebshihadeh:

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -10,6 +10,7 @@ from cereal import log
 from typing import Any
 
 import cereal.messaging as messaging
+from cereal.visionipc.visionipc_pyx import VisionIpcServer, VisionStreamType  # pylint: disable=no-name-in-module, import-error
 from common.params import Params
 from common.realtime import Ratekeeper, DT_DMON
 from lib.can import can_function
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
@deanlee this is mostly being held up because the YUV -> RGB conversion is slow. want to write a GL shader so we can ditch the RGB buffers?

deanlee:
Do we still need to use EGLImageTexture?

jwolffe:
I have started work on this, as part of cleaning up the Carla simulation. Got the UI video working via vipc located in the bridge. I discovered "the hard way" that bridge.py needs to produce both RGB and YUV video, to satisfy both the user interface and modeld. No problem, lots of learning along the way. I am wondering, however, why RGB <-> YUV conversions aren't centralized in the vipc server. Server could listen to either, but produce both formats. I will continue to focus on adding the conversion to YUV into the bridge. Do you think openCL (as in existing camera code) is the right way to convert, or perhaps something else? I'm doing my own investigation of conversion methods too.

adeebshihadeh:
I believe @nuwandavek is writing a small hack to get this merged in during the hackathon, but it seems like you're interested in doing this properly. Ultimately, we want to remove the RGB stream. The only thing left to do here is make the UI work with the YUV stream, which should be relatively straightforward. Everything else that uses visionipc already uses the YUV streams.

jwolffe:
Thanks for the guidance. Moving toward a single protocol makes a lot of sense. I'll focus on producing YUV from the bridge.

jwolffe:
Looking into ways to do the conversion, I came across pyopencl. Seems like this would allow me to port the existing rgb_to_yuv conversion routine (written in C++/opencl ) over to the python environment of bridge.py without too much fuss. It looks like it fits well with numpy too. Does this approach make sense to you?

jwolffe:
I have modified bridge.py to do the conversion to YUV, using pyopencl to call the existing rgb_to_yuv.cl kernel. Seems to run,  invoking the kernel and moving buffers around, but it needs testing. I tried the draft ui mods for using YUV, but got a green screen, and as I recall that is in line with some troubles that were reported for that mod. Is a working version of the YUV UI screen expected soon? I could try fixing it myself, but that would probably be a duplicated effort, with a whole new learning curve.

deanlee:
#22905 can display YUV correctly.  you can run replay --demo --yuv to see the result .have you checked your conversion in other ways?

jwolffe:
Thanks, knowing the display is good helps a lot. I'll have to look elsewhere for the problems.

PR #21325, opened 2021-06-18
util::read_file: fix reading from sysfs and add test case

adeebshihadeh:


deanlee:
OMG.....fail again?

deanlee:
telg(/sys/power/wakeup_count) return 4kb length, but actually only a few bytes can be read.so the f.read return false.but in fact the data in result is correct

PR #21326, opened 2021-06-18
OffroadHome: separate the update and alert buttons. 

### LINKED ISSUE #21266: Update alert incorrectly supersedes overtemp alert

deanlee:
fix issue: #21266

### DIFF
--- a/selfdrive/ui/qt/widgets/offroad_alerts.cc
+++ b/selfdrive/ui/qt/widgets/offroad_alerts.cc
@@ -41,6 +41,7 @@ OffroadAlert::OffroadAlert(QWidget* parent) : QFrame(parent) {
   QObject::connect(dismiss_btn, &QPushButton::released, this, &OffroadAlert::closeAlerts);
 
   rebootBtn.setText("Reboot and Update");
+  rebootBtn.setObjectName("rebootBtn");
   rebootBtn.setFixedSize(600, 125);
   rebootBtn.setVisible(false);
   footer_layout->addWidget(&rebootBtn, 0, Qt::AlignBottom | Qt::AlignRight);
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/widgets/offroad_alerts.cc
```diff
@@ -61,52 +61,61 @@ OffroadAlert::OffroadAlert(QWidget* parent) : QFrame(parent) {
       border-radius: 30px;
       background-color: white;
     }
+    #rebootBtn {
+      background-color: #E22C2C;
+    }
```
adeebshihadeh, line 66:
```suggestion
```

deanlee:
think it's better to split the update from OffroadAlert into a separate widget

adeebshihadeh:
I'm fine with that

adeebshihadeh:
Noticed some regressions:
* missing stretch at the bottom of the alerts layout
* update window doesn't go away when UpdateAvailable is deleted (button goes away though)

Used `selfdrive/ui/tests/cycle_offroad_alerts.py` to test.

deanlee:
fixed

adeebshihadeh:
There's a flash on startup now while the layout resizes. I think it's only if alerts exist on startup.

PR #21331, opened 2021-06-18
SoftwarePanel: relase file watcher in time to reduce resource consumption

deanlee:


adeebshihadeh:
Is this really a concern? The file system watcher should be pretty light.

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -240,13 +241,23 @@ SoftwarePanel::SoftwarePanel(QWidget* parent) : QWidget(parent) {
...
   });
+  std::system("pkill -1 -f selfdrive.updated");
 }
 
 void SoftwarePanel::showEvent(QShowEvent *event) {
   updateLabels();
 }
 
+void SoftwarePanel::hideEvent(QHideEvent* event) {
+  if (fs_watch) {
```
adeebshihadeh, line 255: why not just remove and add the paths?

deanlee:
If I understand correctly, although we only added two files, fswatch will hook the entire params/d directory,it's not cheap. we should release it after use Instead of always hooking up

selfdrive/ui/qt/offroad/settings.cc
deanlee: I think delete is more clearer, or I have to save paths in the class variable, or write it here  again

pd0wm:
Where in the docs do you see it hooks the whole folder? In my understanding it would hook the param file. When `updated` writes to the param, the old file gets overwritten. This removes it from the watch list:

https://doc.qt.io/qt-5/qfilesystemwatcher.html#details
> Note that QFileSystemWatcher stops monitoring files once they have been renamed or removed from disk, and directories once they have been removed from disk.

deanlee:
Sorry for wasting your time.I read a wrong and outdated document.

PR #21332, opened 2021-06-18
Fix triangle size

pd0wm:


deanlee:
It's nice to see that finally getting rid of s->zoom

PR #21336, opened 2021-06-18
PID use limited torque of CC.apply_last_steer

ClockeNessMnstr:
https://github.com/commaai/openpilot/pull/21005

I'm trying to implement the same as Qadmus but for PID. 

@pd0wm 
@qadmus 

Testing this in a few hours to see if it works correctly. I was previously passing around the value quite "jankily" hoping this is more along the lines of what was mentioned in the other PR

*TODO*:
- [x] Pass the last requested data from the PID to the call from CI
- [x] implement the function in each vehicles car controller? just needs apply_last /  STEERMAX from controller

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -87,7 +87,7 @@ class CarInterfaceBase():
 
   def calc_last_outputs(self):
     lat_out = None
-    if self.CC.get_last_output is not None:
+    if hasattr(self.CC, 'get_last_output'):
       lat_out = self.CC.get_last_output()
     return lat_out
     
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -38,7 +38,7 @@ class CarInterfaceBase():
     self.CC = None
     if CarController is not None:
       self.CC = CarController(self.cp.dbc_name, CP, self.VM)
-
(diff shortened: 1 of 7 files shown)


pd0wm:
See my comment in the previous PR.

> How about adding a function to car interface that deals with the rate limit and the clip? (just make it return the input on cars not using this). Then we can call that from both latcontrol_indi and use if from the carcontroller itself.

This method should contain all the car specific rate limiting and clipping.

ClockeNessMnstr:
> This method should contain all the car specific rate limiting and clipping.

Only issue with this is that the rate limiting and clipping is in a vehicle specific scale for CarController but (-1, 1) for latControl_XXX

It might not make sense to have the scaling as part of the output so it seems that this function would need to:
take in a (-1, 1)
apply vehicle specific scaling -> limiting ->
then re-scale (-1, 1)

Then car controller would need to scale again? 
seems a bit weird. 

I updated my branch to be as close as possible to this but just grabbing the last value and descaling. 
call is made in latcontrol, it passes through the old value

ClockeNessMnstr:
is the Car Parameter for Steer Max available for all cars? maybe that's the answer

ClockeNessMnstr:
Could we re-open to get the updated commits?

PR #21337, opened 2021-06-19
Simplify longitudinal tests

haraschax:


selfdrive/test/longitudinal_maneuvers/test_longitudinal.py
```diff
@@ -1,17 +1,13 @@
...
-
-from selfdrive.config import Conversions as CV
-from selfdrive.car.honda.values import CruiseButtons as CB
+from selfdrive.test.process_replay.test_processes import get_segment
 from selfdrive.test.longitudinal_maneuvers.maneuver import Maneuver
 from selfdrive.manager.process_config import managed_processes
 from common.file_helpers import mkdirs_exists_ok
 from common.params import Params
+from tools.lib.logreader import LogReader
+import unittest
```
adeebshihadeh, line 9: import std libs on top

selfdrive/test/longitudinal_maneuvers/test_longitudinal.py
```diff
@@ -25,65 +21,45 @@ def check_fcw(log):
 def check_engaged(log):
   return log['controls_state_msgs'][-1][-1].active
 
+def put_default_car_params():
+  cp = [m for m in LogReader(get_segment('0982d79ebb0de295|2021-01-08--10-13-10--0')) if m.which() == 'carParams']
```
adeebshihadeh, line 25: no need for a log here. use `CarInterface.get_params`

selfdrive/test/longitudinal_maneuvers/plant.py
```diff
@@ -1,467 +1,158 @@
...
-      Plant.sendcan = messaging.sub_sock('sendcan')
-      Plant.model = messaging.pub_sock('modelV2')
-      Plant.live_params = messaging.pub_sock('liveParameters')
-      Plant.live_location_kalman = messaging.pub_sock('liveLocationKalman')
-      Plant.pandaState = messaging.pub_sock('pandaState')
-      Plant.deviceState = messaging.pub_sock('deviceState')
-      Plant.driverMonitoringState = messaging.pub_sock('driverMonitoringState')
-      Plant.cal = messaging.pub_sock('liveCalibration')
-      Plant.controls_state = messaging.sub_sock('controlsState')
+      Plant.radar = messaging.pub_sock('radarState')
```
adeebshihadeh, line 29: ideally this shouldn't use any real sockets. should be able to reuse process replay's fake pub and sub master

selfdrive/test/longitudinal_maneuvers/plant.py
```diff
@@ -1,467 +1,158 @@
...
-
-def to_3s_byte(x):
-  s = struct.pack("!h", int(x))
-  return binascii.hexlify(s)[1:]
 
 class Plant():
   messaging_initialized = False
 
-  def __init__(self, lead_relevancy=False, rate=100, speed=0.0, distance_lead=2.0):
+  def __init__(self, lead_relevancy=False, rate=20, speed=0.0, distance_lead=2.0):
```
adeebshihadeh, line 25: import DT_MDL from common.realtime

selfdrive/test/longitudinal_maneuvers/plant.py
adeebshihadeh: we can do this later

selfdrive/test/longitudinal_maneuvers/maneuver.py
```diff
@@ -30,56 +25,20 @@ def evaluate(self):
...
-        for k, v in log.items():
-          logs[k].append(v)
-
-    valid = True
-    for check in self.checks:
-      c = check(logs)
-      if not c:
-        print(check.__name__ + " not valid!")
-      valid = valid and c
+    if d_rel < 1.0:
```
adeebshihadeh, line 39: was this intentional? only check at the end? if so, let's also remove the unused checks in `test_longitudinal.py`

HaraldSchafer: No it was not. it should check every iteration

adeebshihadeh: if we're removing the fcw check, should we remove the fcw test cases? https://github.com/commaai/openpilot/blob/simplify_longitudinal_tests/selfdrive/test/longitudinal_maneuvers/test_longitudinal.py#L31

HaraldSchafer: You can remove them, but put a todo to make new ones. I want to make new fcw and test that accordingly with this test.

PR #21338, opened 2021-06-19
Add 2021 Avalon Hybrid Fingerprint

nelsonjchen:
**Checklist**
- [x] added to README
- [ ] test route added to [test_routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_models.py)
- [ ] route with openpilot:


nelsonjchen:
Follow up of #19928 after all the Avalon confusion was resolved.

Jamil's device ID is 1d09d9186e2e8a48. No drives have been uploaded yet. Will note here when they are.

adeebshihadeh:
Full logs aren't necessary for verifying fingerprints.

nelsonjchen:
@adeebshihadeh Thanks for looking into this!

PR #21341, opened 2021-06-19
Toyota C-HR Hybrid FW versions from Japan model DAA-ZYX10-AHXNB

belm0:
This model was produced from 2016-2018 in Japan.

http://www.hoyusys.co.jp/catalog/TOYOTA/CHR/201612/

test route: 2c4292a5cd10536c|2021-06-20--01-04-14


adeebshihadeh:
Converting to draft until this is tested

belm0:
updated description with test route

PR #21342, opened 2021-06-19
Qt: smooth scrolling

deanlee:


adeebshihadeh:
I don't think we can really fix this by tuning the scroller. The issue seems to be lack of CPU, so we either need to move to the GPU or dramatically reduce the amount of CPU scrolling uses.

deanlee:
trying to reimplement QScrollArea first

deanlee:
Or we can try to scroll a snapshot of the viewport instead of move it.

PR #21343, opened 2021-06-19
Qt: move settings to GPU

deanlee:


adeebshihadeh:
We have to make sure this doesn't use much GPU at all because of the way the GPU scheduling works with the adreno driver. Once we ship big model, we won't have much GPU headroom while we're onroad. I'd recommend checking with the [Snapdragon profiler](https://developer.qualcomm.com/software/snapdragon-profiler).

deanlee:
Even drawing with OpenGL, the CPU usage is still high when scrolling.

adeebshihadeh:
Sounds like we're doing something else wrong. Does a simple page of only QLabels scroll nicely or does it use a bunch of CPU too?

deanlee:
I don't think we have done anything wrong so far.no matter what we scroll(e.g: simple qlabels in QScrollArea or some text items in QListView),  the cpu usage is always high if scroll the content by pixels.

dorkmo:
How often is scrolling needed while onroad?

deanlee:
> How often is scrolling needed while onroad?
It's better to make scrolling smooth,  I'm trying to  figured out an approach to solve the problem.

dorkmo:
> 
> 
> > How often is scrolling needed while onroad?
> 
> It's better to make scrolling smooth, I'm trying to figured out an approach to solve the problem.

i was just thinking that maybe the performance penalty to cpu or gpu doesnt hurt driving model because scrolling usually only takes places while you are parked and not engaged. Could the scrolling gpu performance be disabled while onroad?

PR #20550, opened 2021-04-01
Add 2021 Accord 2.0 Touring & Hybrid Touring

csouers:
- Create the HONDA_BOSCH_EXT set for HONDA_BOSCH cars that utilize the new extended IDs for LKAS_HUD
- Create a new CAR for each 2021 variant. Let's us differentiate between which LKAS_HUD frames to send to the radar.
 
These cars are basically the same as prior models with the aforementioned changes to LKAS_HUD on the private CAN.

**Pre reqs**
- opendbc > [pr](https://github.com/commaai/opendbc/pull/372)
- panda > [pr](https://github.com/commaai/panda/pull/615)
- add `085ac1d942c35910|2021-03-25--20-11-15` to the test bucket

**Checklist**
- [x] added to README
- [x] test routes added to [test_car_models](../../selfdrive/test/test_car_models.py)
- [x] routes with openpilot: [2.0 Touring](https://my.comma.ai/cabana/?route=085ac1d942c35910%7C2021-03-31--19-35-33&exp=1648783384&sig=aqbsNvoXEpgy4IjKdQ8g%2FZESG%2FQJP9uFQNQDGq%2Fx02I%3D&max=9&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2F085ac1d942c35910%2Ff24803fb096e557ecae92aa832461cd4_2021-03-31--19-35-33) [Hybrid Touring](https://my.comma.ai/cabana/?route=f29e2b57a55e7ad5%7C2021-04-01--13-04-26&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2Ff29e2b57a55e7ad5%2F3d99aebd0e6345d0a239b7a55c950dca_2021-04-01--13-04-26&seekTime=2)
[...]

PR #21347, opened 2021-06-20
Added Firmware 2021 Accord Hybrid

feyded1020:
***** Template: Fingerprinting for 2021 Honda Accord Hybrid *****

**Description** No support for 2021 Honda Accord Hybrid, added firmwares to values.py

**Verification** Followed fingerprinting wiki

**Route**
Route: None yet

-->

### DIFF
--- a/selfdrive/car/honda/values.py
+++ b/selfdrive/car/honda/values.py
@@ -268,6 +268,7 @@ FW_VERSIONS = {
       b'57114-TWA-A040\x00\x00',
       b'57114-TWA-A050\x00\x00',
       b'57114-TWA-B520\x00\x00',
+      b'57114-TWA-A530\x00\x00',
     ],
     (Ecu.srs, 0x18da53f1, None): [
       b'77959-TWA-A440\x00\x00',
@@ -281,6 +282,7 @@ FW_VERSIONS = {
       b'78109-TWA-A120\x00\x00',
       b'78109-TWA-A210\x00\x00',
       b'78109-TWA-A220\x00\x00',
+      b'78109-TWA-A230\x00\x00',
       b'78109-TWA-L010\x00\x00',
     ],
     (Ecu.shiftByWire, 0x18da0bf1, None): [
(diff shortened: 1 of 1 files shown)


selfdrive/car/honda/values.py
```diff
@@ -1301,4 +1303,4 @@ class CAR:
 }
 
 HONDA_BOSCH = set([CAR.ACCORD, CAR.ACCORDH, CAR.CIVIC_BOSCH, CAR.CIVIC_BOSCH_DIESEL, CAR.CRV_5G, CAR.CRV_HYBRID, CAR.INSIGHT, CAR.ACURA_RDX_3G])
-HONDA_BOSCH_ALT_BRAKE_SIGNAL = set([CAR.ACCORD, CAR.CRV_5G, CAR.ACURA_RDX_3G])
+HONDA_BOSCH_ALT_BRAKE_SIGNAL = set([CAR.ACCORD, CAR.CRV_5G, CAR.ACURA_RDX_3G])
```
belm0, line 1306: suggest restoring the file's terminating newline-- there shouldn't be a change here

adeebshihadeh:
What's your dongle ID?

selfdrive/car/honda/values.py
adeebshihadeh: yeah, let's revert this

feyded1020:
> What's your dongle ID?

It is:
1949e3587c4885e7

selfdrive/car/honda/values.py
feyded1020: Weird, not sure how that happened. I'm still relatively new to this, but all I did was input 2 new lines in Textmate and save it. 

Should I retry and reupload to see if it works correctly?

adeebshihadeh:
Does it work after adding the fingerprint? There's a car port PR open for the 2021 Accords. #20550

feyded1020:
> Does it work after adding the fingerprint? There's a car port PR open for the 2021 Accords. #20550

Yes, car no longer says 'No Exact Match Found', and haven't had any issues with the driving capabilities.

csouers:
@feyded1020 Can you please provide the VIN for the vehicle and also a cabana link that is from the "Copy Share Link" button? You can message me on the comma Discord with the username `@csouers` with any questions.  I am doing the the port for the 2021+ Accord and Odyssey and this is very curious. I would expect to see different radar and camera firmware IDs for this model year.

PR #21358, opened 2021-06-21
DevicePanel: set specific stylesheet for reboot&poweroff buttons

deanlee:
The current setstylesheet  has affected all pushbuttons in this page, including the title button in AbstractControl:


pd0wm:
I don't see a difference in background before/after this PR. For me the device panel looks fine.

deanlee:
because there is a hack used to fix this problem in AbstractControl:
https://github.com/commaai/openpilot/blob/22b632f4078a979d34b833a4a265ac79014465c3/selfdrive/ui/qt/widgets/controls.cc#L58

remove this line,you can see the difference.

pd0wm:
Can you do both in the same PR so we can remove that hack immediately?

PR #21362, opened 2021-06-21
fix FileLock: fd could be closed twice

deanlee:
since we no longer throw exception in the `lock()`, we should only close fd_ in `unlock()`


adeebshihadeh:
Can we start adding tests for these small helper classes if you're touching them anyway?

PR #21366, opened 2021-06-21
General networking improvements

sshane:
- Fix race condition; individually connect buttons to `connectToNetwork`, passing the network
- Instead of calling `ListConnections` each time we call `pathFromSsid` (checking if connection is known, activating a connection, forgetting a network, etc) we update class variable `known_connections` once per update, and when a connection is added or removed by user.
- Shorter syntax to parse data from nm calls

Previous avg time to update UI: 208, 187, 210, 189, 207, 207, 212, 200, 188. AVG: 200.89 ms
New avg time to update UI: 84, 70, 81, 83, 84, 89, 81, 81, 85. AVG: 82 ms

### DIFF
--- a/selfdrive/ui/qt/offroad/wifiManager.cc
+++ b/selfdrive/ui/qt/offroad/wifiManager.cc
@@ -68,6 +68,7 @@ bool compare_by_strength(const Network &a, const Network &b) {
 WifiManager::WifiManager(QWidget* parent) : QWidget(parent) {
   qDBusRegisterMetaType<Connection>();
   qDBusRegisterMetaType<IpConfig>();
+  qDBusRegisterMetaType<QDBusObjectPath>();
   connecting_to_network = "";
   adapter = get_adapter();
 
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -380,46 +386,38 @@ void WifiManager::disconnect() {
   }
 }
 
-QDBusObjectPath WifiManager::pathFromSsid(const QString &ssid) {
-  QDBusObjectPath path;  // returns uninitialized path if network is not known
-  for (auto const& [conn_ssid, conn_path] : listConnections()) {
+QDBusObjectPath WifiManager::pathFromSsid(const QString &ssid, int &index) {
+//  QDBusObjectPath path;  // returns uninitialized path if network is not known
+  index = 0;
```
adeebshihadeh, line 391: what's the point of the index? why is it written even if the path isn't found?

sshane: This was a hack to avoid iterating through `known_connections` twice in `forgetConnection`, and I didn't want to ignore using it in that one function for some unknown reason. I'll find a cleaner way to do this

sshane:
Waiting for PR #21310 to be merged, then I'll rebase this

sshane:
In the next PR I'll tackle `// TODO: only refresh on firstScan, then use AccessPointAdded and Removed signals`

sshane:
The `firstScan` parameter will be used in the future for getting the initial list of access points, as well as the initial list of known connections, whenever the signal for that one can be figured out (AP signals work strangely)

PR #21374, opened 2021-06-22
Added lots of new type hinting/minor fixes

ChadBailey:
* Added some comments
* Fixed up minor issues exposed by type checking
  (i.e. stderr possibly being None)
* Floored spinner progress to prevent erroneous status message

### DIFF
--- a/selfdrive/hardware/eon/hardware.py
+++ b/selfdrive/hardware/eon/hardware.py
@@ -2,6 +2,7 @@ import binascii
 import itertools
 import os
 import re
+import serial
 import struct
 import subprocess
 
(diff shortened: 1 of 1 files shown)


ChadBailey:
Seems this is broken... unfortunately unable to get all tests running locally, so I was unaware. I'll get it fixed up and commit

ChadBailey:
Seems my first issue was related to using `Any` in place of `Union`, oops. I believe mypy should catch that, so I filed an upstream bug report https://github.com/python/mypy/issues/10695

Pushing fixed version... this passes all pre-commit hook items, so hoping this will pass!
image

PR #21377, opened 2021-06-23
TEST NEEDED: Feature/improved installers

ChadBailey:
This pull request is a combination of https://github.com/commaai/openpilot/blob/master/tools/mac_setup.sh and https://github.com/commaai/openpilot/blob/master/tools/ubuntu_setup.sh along with several other checks and niceties added.

The goal is to improve the user experience by reducing steps & giving more friendly status messages.

- [ ] Merge changes from #21380 
- [ ] Test on Linux
- [ ] Test Linux - non-home directory
- [ ] Mac
- [ ] Mac - non-home directory
- [ ] CI Pipeline

Additionally, the verbiage should probably be approved by whoever does such a thing (not my strong suit, so please feel free to post verbiage improvements - i'll get them added)


ChadBailey:
Based on line 52 of the Mac setup she'll script, it checks if "$CI" which was the only evidence I found this may be used in any CICD pipeline. Can anyone confirm if it is used for a pipeline?

I have t had the chance to look closely at the GitHub actions yet but even if I had I didn't want to overlook it if say it was hidden as a variable name etc.

ChadBailey:
This now conflicts with #21380 which I didn't notice while building. I'll get the merge conflict resolved by implementing those improvements later today

tools/install_tools.sh
```diff
@@ -0,0 +1,282 @@
...
+echo
+echo "----   FINISH OPENPILOT SETUP   ----"
+echo
+echo "Congratulations, your dev environment has been set up successfully!"
+echo 
+echo "Useful Commands:"
+echo "  Here are some useful commands you may want to"
+echo "  consider adding as aliases to your $RC_FILE:"
+echo
+echo '    alias op-pch="pre-commit run --all"'
```
adeebshihadeh, line 269: don't think these should be added as aliases. I was thinking something more like this https://github.com/ApolloAuto/apollo/blob/master/apollo.sh

tools/install_tools.sh
```diff
@@ -0,0 +1,282 @@
...
+    echo "[PASS] OP_HOME: $OP_HOME"
+  fi
+}
+
+check_shell
+check_op_home
+
+install_deps_apt() {
+  APT_VERSION=$(apt-get --version | awk -F "." '/apt/ {print $1}')
+  if [[ $APT_VERSION != "apt 2" ]]; then
```
adeebshihadeh, line 101: this seems pretty excessive. has anyone actually run into this?

tools/install_tools.sh
```diff
@@ -0,0 +1,282 @@
...
+pyenv global 3.8.5
+pyenv rehash
+eval "$(pyenv init -)"
+pip install --upgrade pip==20.2.4
+pip install pipenv==2020.8.13
+pipenv install --system --deploy
+git -C $OP_HOME submodule update --init
+pre-commit install
+
+if [[ `git branch --show-current` == "release2" ]]; then
```
adeebshihadeh, line 253: think this check is also excessive. the tools/ directory is stripped out of release builds anyway (see release/files_common)

tools/install_tools.sh
ChadBailey: thanks for the review... the aliases were just the ones i happened to create with little thought, then figured i'd share as a helpful tip lol.

at first i was going to suggest just using pipenv scripts but i really like the idea of creating a make-style script like the one you linked. i also wonder if scons has make-style build objectives we could leverage. i'll take a stab at it tomorrow

tools/install_tools.sh
ChadBailey: this was the more explicit form of checking for apt package manager as a whole. my intent was to make it fail on distros which don't use apt, while not penalizing non-ubuntu users so long as their system meets requirements.

tools/install_tools.sh
adeebshihadeh: this should really be a separate PR though. let's remove from this one for now

ChadBailey:
I'll be doing a somewhat significant update to this tomorrow or over the weekend, please wait for the update before testing/etc.

ChadBailey:
I didn't realize I had some other stuff planned over the weekend so I wasn't able to update. I plan to do so sometime throughout the week. I'll close this for now then reopen once ready

tools/install_tools.sh
ChadBailey: sounds good

PR #21378, opened 2021-06-23
Added lots of new type hinting/minor fixes

ChadBailey:
* Added some comments
* Fixed up minor issues exposed by type checking
  (i.e. stderr possibly being None)
* Floored spinner progress to prevent erroneous status message

Edit: If anyone is confused I originally opened this under https://github.com/commaai/openpilot/pull/21374 but then accidentally wiped out the commit & pull request by removing the commit from my fork, oops!

The original PR failed to build and initiated an upstream bug report in mypy https://github.com/python/mypy/issues/10695

### DIFF
--- a/common/spinner.py
+++ b/common/spinner.py
@@ -25,7 +25,7 @@ class Spinner():
         pass
 
   def update_progress(self, cur: int, total: int):
-    self.update(str(round(100 * cur / total)))
+    self.update(str(100 * cur // total))
 
   def close(self):
     if self.spinner_proc is not None:
--- a/selfdrive/manager/manager.py
+++ b/selfdrive/manager/manager.py
@@ -5,6 +5,7 @@ import signal
 import subprocess
 import sys
 import traceback
+from typing import Union, NoReturn
 
 import cereal.messaging as messaging
 import selfdrive.crash as crash
@@ -102,14 +103,25 @@ def manager_prepare():
     p.prepare()
 
 
-def manager_cleanup():
+def manager_cleanup() -> None:
+  """Kills all processes in `managed_processes`"""
   for p in managed_processes.values():
     p.stop()
 
   cloudlog.info("everything is dead")
 
 
-def manager_thread():
+def manager_thread() -> Union[NoReturn, None]:
+  """
+  Initializes the openpilot manager thread which launches and manages all
+  processes found in `manager.process_config.managed_processes()`.
+
+  NoReturn:
+    Does not return (typical operation)
+
+  None:
+    Returns None When `Params().get_bool("DoUninstall")` is True
+  """
   cloudlog.info("manager start")
   cloudlog.info({"environ": os.environ})
 
@@ -123,8 +135,10 @@ def manager_thread():
     ignore += ["manage_athenad", "uploader"]
   if os.getenv("NOBOARD") is not None:
     ignore.append("pandad")
-  if os.getenv("BLOCK") is not None:
-    ignore += os.getenv("BLOCK").split(",")
+  try:
+    ignore += os.environ["BLOCK"].split(",")
+  except (KeyError,AttributeError):
+    pass
 
   ensure_running(managed_processes.values(), started=False, not_run=ignore)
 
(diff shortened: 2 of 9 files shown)


selfdrive/manager/manager.py
```diff
@@ -193,6 +215,7 @@ def main():
   if Params().get_bool("DoUninstall"):
     cloudlog.warning("uninstalling")
     HARDWARE.uninstall()
+  return None
```
adeebshihadeh, line 218: why was this added?

selfdrive/manager/manager.py
```diff
@@ -177,7 +199,7 @@ def main():
   manager_prepare()
 
   if prepare_only:
-    return
+    return None
```
adeebshihadeh, line 202: and this one?

selfdrive/manager/manager.py
```diff
@@ -124,7 +136,7 @@ def manager_thread():
   if os.getenv("NOBOARD") is not None:
     ignore.append("pandad")
   if os.getenv("BLOCK") is not None:
-    ignore += os.getenv("BLOCK").split(",")
+    ignore += os.getenv("BLOCK","").split(",")
```
adeebshihadeh, line 139: there's already a check for this. either revert this or remove the if statement

common/spinner.py
```diff
@@ -25,7 +25,10 @@ def update(self, spinner_text: str):
         pass
 
   def update_progress(self, cur: int, total: int):
-    self.update(str(round(100 * cur / total)))
+    # Always floor progress bars because they indicate fully completed chunks
```
adeebshihadeh, line 28: no need for the comment here

selfdrive/manager/process.py
```diff
@@ -58,11 +59,11 @@ def join_process(process, timeout):
...
-  daemon = False
-  sigkill = False
-  proc = None
-  enabled = True
+  unkillable: bool = False
+  daemon: bool = False
+  sigkill: bool = False
+  proc: Optional[Process] = None
+  enabled: bool = True
   name = ""
```
adeebshihadeh, line 67: type hint?

selfdrive/timezoned.py
```diff
@@ -12,7 +13,12 @@
 from selfdrive.swaglog import cloudlog
 
 
-def set_timezone(valid_timezones, timezone):
+def set_timezone(valid_timezones: list, timezone: str):
```
adeebshihadeh, line 16: annotate return type?

selfdrive/manager/manager.py
ChadBailey: mypy doesn't like implied returns, so it will error with either "Missing return statement" or "Return value expected" when adding any return type hints.

selfdrive/manager/manager.py
ChadBailey: see prior comments

selfdrive/manager/manager.py
ChadBailey: this is another type related oddity... since `os.getenv`'s return value is optional, it would potentially raise an exception in the event of it being None... i couldn't just remove the if statement though because it would add an empty string to the list which is unintended behavior. i'm going to submit a rewritten version of this section to solve all concerns

common/spinner.py
ChadBailey: sorry, i've had to have this conversation a surprising amount of times so i suppose i was trying to head it off at the pass

ChadBailey:
i made all the changes based on the feedback... i also renamed watchdog_max_dt to watchdog_max_seconds since dt implies "datetime". from what i was able to find, it looks like it's using nanoseconds downsampled to seconds via / 1e9. let me know if i've gotten that wrong

selfdrive/manager/manager.py
```diff
@@ -123,8 +135,10 @@ def manager_thread():
     ignore += ["manage_athenad", "uploader"]
   if os.getenv("NOBOARD") is not None:
     ignore.append("pandad")
-  if os.getenv("BLOCK") is not None:
-    ignore += os.getenv("BLOCK").split(",")
+  try:
```
adeebshihadeh, line 138: revert to the if

adeebshihadeh:
It's dt as in delta time, not datetime.

adeebshihadeh:
Converting to draft until it's ready

adeebshihadeh:
Any update on this? If it's going to sit as a draft for a while, let's close until it's ready.

adeebshihadeh:
Closing due to inactivity. Feel free to reopen after addressing comments.

PR #21380, opened 2021-06-23
fix: fix pathing issues in ubuntu_setup.sh and openpilot_env.sh

iamorphen:
**Intro**

Hi all, this is my first MR here, so I took some low-hanging fruit to get used to the process.

**Description**

The [openpilot tools setup instructions](https://github.com/commaai/openpilot/blob/master/tools/README.md#setup) does not work as-is.

When run from the home directory, git operations fail before `cd`'ing into the git tree. We could solve this by telling people to `cd` into the git tree, but it requires another step for the end-user, so I wanted to handle it programmatically.

Also, `pyenv` is not set up correctly, causing executables managed by it to not be on the `PATH`. This includes things like `scons`, without which we cannot build.

Additionally, `PYTHONPATH` [is `PATH`-like](https://docs.python.org/3.8/using/cmdline.html#envvar-PYTHONPATH), so we now extend it instead of overwriting it.

**Other Things**
This also writes a newline into .bashrc before extending it, to visually separate openpilot additions from existing content.

**Verification**

In short, I followed the setup instructions again with my changes in place.

In detail, I:
1. Removed my `pyenv` installation: `rm -rf ~/.pyenv`
2. Changed to my home directory
3. Ran `openpilot/tools/ubuntu_setup.sh`
4. [EDIT] Sourced my .bashrc
4. Confirmed `pyenv`, `scons`, etc. are on my path
5. Ran a successful build with `scons -j$(nproc)`

### DIFF
--- a/tools/ubuntu_setup.sh
+++ b/tools/ubuntu_setup.sh
@@ -71,7 +71,6 @@ fi
 # in the openpilot repo
 cd $HOME/openpilot
 
-# maybe extend bashrc
 source ~/.bashrc
 if [ -z "$OPENPILOT_ENV" ]; then
   OP_DIR=$(git rev-parse --show-toplevel)
--- a/tools/ubuntu_setup.sh
+++ b/tools/ubuntu_setup.sh
@@ -68,17 +68,17 @@ if ! command -v "pyenv" > /dev/null 2>&1; then
   curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
 fi
 
-# install bashrc
+# in the openpilot repo
+cd $HOME/openpilot
+
+# maybe extend bashrc
 source ~/.bashrc
 if [ -z "$OPENPILOT_ENV" ]; then
   OP_DIR=$(git rev-parse --show-toplevel)
-  echo "source $OP_DIR/tools/openpilot_env.sh" >> ~/.bashrc
+  printf "\nsource %s/tools/openpilot_env.sh" "$OP_DIR" >> ~/.bashrc
   source ~/.bashrc
   echo "added openpilot_env to bashrc"
 fi
-
-# in the openpilot repo
-cd $HOME/openpilot
 
 # do the rest of the git checkout
 git lfs pull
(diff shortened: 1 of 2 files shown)


tools/ubuntu_setup.sh
```diff
@@ -68,18 +68,18 @@ if ! command -v "pyenv" > /dev/null 2>&1; then
   curl -L https://github.com/pyenv/pyenv-installer/raw/master/bin/pyenv-installer | bash
 fi
 
-# install bashrc
+# in the openpilot repo
+cd $HOME/openpilot
+
+# maybe extend bashrc
```
adeebshihadeh, line 74:
```suggestion
# maybe extend bashrc
```

ChadBailey:
@iamorphen do you happen to have access to a mac? I'd like help validating these changes on mac once I merge them in #21377 if you would be willing/able

iamorphen:
> do you happen to have access to a mac? I'd like help validating these changes on mac once I merge them in #21377 if you would be willing/able

@ChadBailey , thanks for asking. My changes only affect installation on Ubuntu with the exception of the adjustment of `PYTHONPATH`. Is it the `PYTHONPATH` change in particular you're looking to validate?

ChadBailey:
my changes were changing some of the same files, but i also combined the linux installer with the mac installer. that's why i asked about the mac, i just thought maybe i would luck out and you have a mac since i was in need of a mac tester and you are familiar with the setup process lol.

i do have some more updating to do before it's ready to be tested, but if you're interested i'll let you know when i've finished with the updates.

iamorphen:
> i also combined the linux installer with the mac installer

I see! Sorry if I threw a wrench in there anywhere. As for the installers--as you probably already know--what I learned in this work is that manual installation for Ubuntu involves different steps (and different potential adjustments) than for macOS. You could give the ["easy" installer](https://github.com/pyenv/pyenv-installer) a shot, but according to its wiki, if it fails for whatever reason, some potential fixes are different across platforms.

As for me, I will only be working with openpilot on Ubuntu at this time, but I salute your work on the installer!

PR #21386, opened 2021-06-23
Stay in ELM327 mode during controls init

pd0wm:
Set a safety param to make sure bus 1 is no longer connected to the OBD-II port


adeebshihadeh:
We should add some testing around boardd's safety setting.

PR #21387, opened 2021-06-23
Honda: emulate convenience blinker from stalk

### LINKED ISSUE #21130: Honda Nidec using wrong blinker message
**Describe the bug**
When doing lane changes on Honda Nidec, the stalk state is used. This causes the lane change alert to show for a split second and doesn't allow the convenience blinker to be used for lane change assist like on other makes. 

Get in a Nidec, try to change lanes.

**Device/Version information:**
_Route Coming Soon_
 - Dongle ID:
 - Route:
 - Timestamp:
 - Commit hash:
 - Car make/model: Honda Civic 2017

pd0wm:
Closes #21130

- [x] Check convenience blinker duration (currently guessed at 3 seconds)
- [x] Check what happens after two short pressed of the stalk. Does the second press add to the convenience blinker time?
- [x] Test in car


selfdrive/car/interfaces.py
```diff
@@ -186,11 +186,38 @@ def update_speed_kf(self, v_ego_raw):
     v_ego_x = self.v_ego_kf.update(v_ego_raw)
     return float(v_ego_x[0]), float(v_ego_x[1])
 
-  def update_blinker(self, blinker_time: int, left_blinker_lamp: bool, right_blinker_lamp: bool):
+  def update_blinker_from_light(self, blinker_time: int, left_blinker_lamp: bool, right_blinker_lamp: bool):
```
adeebshihadeh, line 189: both light and lamp are used here, let's pick one and stick with it

ChadBailey:
Finished testing in 2018 Accord 1.5T, all went well - route 053abdc1032b7ecf|2021-06-23--18-15-03--8

Admittedly, I intended to test immediately swapping between left-right blinker positions but forgot.

PR #21394, opened 2021-06-24
Refactor Hyundai gear parsing to use DBC definitions

sshane:
*when available. opendbc PR: https://github.com/commaai/opendbc/pull/405

The 2017 Elantra could use TCU gear parsing so I moved it out of cluster gears. Couldn't get any routes for the two remaining cars, but we might be able to get rid of cluster gears if we can verify they work in one of the other methods.


selfdrive/car/hyundai/values.py
```diff
@@ -650,8 +650,8 @@ class Buttons:
 
 FEATURES = {
   # which message has the gear
-  "use_cluster_gears": set([CAR.ELANTRA, CAR.ELANTRA_GT_I30, CAR.KONA]),
-  "use_tcu_gears": set([CAR.KIA_OPTIMA, CAR.SONATA_LF, CAR.VELOSTER]),
+  "use_cluster_gears": set([CAR.ELANTRA_GT_I30, CAR.KONA]),  # TODO: see if we can remove cluster gears for these cars
```
adeebshihadeh, line 653: let's also get rid of these sets and detect this automatically

selfdrive/car/hyundai/carstate.py
```diff
@@ -1,13 +1,26 @@
...
-from selfdrive.car.hyundai.values import DBC, STEER_THRESHOLD, FEATURES, EV_CAR, HYBRID_CAR
+from selfdrive.car.hyundai.values import DBC, STEER_THRESHOLD, USE_FCA, EV_CAR, HYBRID_CAR
 from selfdrive.car.interfaces import CarStateBase
 from opendbc.can.parser import CANParser
 from selfdrive.config import Conversions as CV
 
 GearShifter = car.CarState.GearShifter
 
 
+def parse_cluster_gear(cluster_gear):
```
adeebshihadeh, line 11: set these values in the DBC

sshane:
Some cars support cluster gears which weren't using it before, and some don't

PR #21395, opened 2021-06-24
keyboard: backspace repeat

sshane:
- [x] key repeat
- [ ] fix sticky buttons


pd0wm:
I sometimes hit a key, the touch down registers so the button changes color. However, it seems there is no corresponding touch up so the character never shows up and the button stays pressed.

selfdrive/ui/qt/widgets/keyboard.cc
```diff
@@ -66,7 +66,7 @@ Keyboard::Keyboard(QWidget *parent) : QFrame(parent) {
     {"q","w","e","r","t","y","u","i","o","p"},
     {"a","s","d","f","g","h","j","k","l"},
     {"⇧","z","x","c","v","b","n","m","⌫"},
-    {"123","  ","⏎"},
+    {"123",",","  ",".","⏎"},
```
adeebshihadeh, line 69: revert these changes, let's stick to the ios layout

PR #21399, opened 2021-06-24
tici: flash bootloader partitions from manifest

pd0wm:

### DIFF
--- a/launch_chffrplus.sh
+++ b/launch_chffrplus.sh
@@ -104,62 +104,11 @@ function tici_init {
 
   # Check if AGNOS update is required
   if [ $(< /VERSION) != "$AGNOS_VERSION" ]; then
-    # Get number of slot to switch to
-    CUR_SLOT=$(abctl --boot_slot)
-    if [[ "$CUR_SLOT" == "_a" ]]; then
-      OTHER_SLOT="_b"
-      OTHER_SLOT_NUMBER="1"
-    else
-      OTHER_SLOT="_a"
-      OTHER_SLOT_NUMBER="0"
-    fi
-    echo "Cur slot $CUR_SLOT, target $OTHER_SLOT"
-
-    # Get expected hashes from manifest
     MANIFEST="$DIR/selfdrive/hardware/tici/agnos.json"
-    SYSTEM_HASH_EXPECTED=$(jq -r ".[] | select(.name == \"system\") | .hash_raw" $MANIFEST)
-    SYSTEM_SIZE=$(jq -r ".[] | select(.name == \"system\") | .size" $MANIFEST)
-    BOOT_HASH_EXPECTED=$(jq -r ".[] | select(.name == \"boot\") | .hash_raw" $MANIFEST)
-    BOOT_SIZE=$(jq -r ".[] | select(.name == \"boot\") | .size" $MANIFEST)
-    echo "Expected hashes:"
-    echo "System: $SYSTEM_HASH_EXPECTED"
-    echo "Boot: $BOOT_HASH_EXPECTED"
+    $DIR/selfdrive/hardware/tici/agnos.py --swap $MANIFEST
 
-    # Read hashes from alternate partitions, should already be flashed by updated
-    SYSTEM_HASH=$(dd if="/dev/disk/by-partlabel/system$OTHER_SLOT" bs=1 skip="$SYSTEM_SIZE" count=64 2>/dev/null)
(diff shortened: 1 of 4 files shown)


launch_chffrplus.sh
```diff
@@ -104,62 +104,11 @@ function tici_init {
...
-
-      abctl --set_active "$OTHER_SLOT_NUMBER"
-
-      sleep 1
-      sudo reboot
-    else
-      echo "Hash mismatch, downloading agnos"
-      if $DIR/selfdrive/hardware/tici/agnos.py $MANIFEST; then
-        echo "Download done, swapping active slot to $OTHER_SLOT_NUMBER"
+    $DIR/selfdrive/hardware/tici/agnos.py --swap $MANIFEST
```
adeebshihadeh, line 108: think it would be nice to also move the version check to python for testing purposes

selfdrive/hardware/tici/agnos.py
```diff
@@ -74,25 +76,75 @@ def unsparsify(f: StreamingDecompressor) -> Generator[bytes, None, None]:
...
+  cloudlog.info(f"Downloading and writing {partition['name']}")
+
+  if verify_partition(target_slot_number, partition):
+    cloudlog.info(f"Already flashed {partition['name']}")
+    return
+
+  downloader = StreamingDecompressor(partition['url'])
+
+  # Clear hash before flashing in case we get interrupted
+  full_check = partition.get('full_check', False)
```
adeebshihadeh, line 140: safer to default to true
```suggestion
  full_check = partition.get('full_check', True)
```

selfdrive/hardware/tici/agnos.py
```diff
@@ -120,17 +172,23 @@ def flash_partition(cloudlog, spinner, target_slot, partition):
...
 
 
-def flash_agnos_update(manifest_path, cloudlog, spinner=None):
+def swap(manifest_path, target_slot_number):
   update = json.load(open(manifest_path))
+  for partition in update:
+    if not partition.get('full_check', False):
+      clear_partition_hash(target_slot_number, partition)
+
+  os.system(f"abctl --set_active {target_slot_number}")
```
adeebshihadeh, line 185: return code not checked. should we make a nice abctl wrapper or reimplement in python?
```suggestion
  subprocess.check_output(f"abctl --set_active {target_slot_number}", shell=True)
```

selfdrive/hardware/tici/agnos.py
```diff
@@ -154,21 +212,39 @@ def flash_agnos_update(manifest_path, cloudlog, spinner=None):
...
-    exit(1)
+  parser.add_argument("--swap", action="store_true", help="Verify and perform swap, downloads if necessary")
+  parser.add_argument("manifest", help="Manifest json")
+  args = parser.parse_args()
 
   spinner = Spinner()
   spinner.update("Updating AGNOS")
   time.sleep(5)
 
   logging.basicConfig(level=logging.INFO)
```
adeebshihadeh, line 239: should we try to import cloudlog?

selfdrive/hardware/tici/agnos.py
```diff
@@ -74,25 +76,75 @@ def unsparsify(f: StreamingDecompressor) -> Generator[bytes, None, None]:
...
+  path = f"/dev/disk/by-partlabel/{partition['name']}"
+
+  if partition.get('has_ab', True):
+    path += slot_number_to_suffix(target_slot_number)
+
+  return path
+
+
+def verify_partition(target_slot_number, partition):
+  full_check = partition.get('full_check', False)
```
adeebshihadeh, line 99: same as above
```suggestion
  full_check = partition.get('full_check', True)
```

launch_chffrplus.sh
pd0wm: Like to keep that logic the same between agnos/neos. Once we move the NEOS updater to python as well we can chage it.

selfdrive/hardware/tici/agnos.py
pd0wm: Was for backwards compatibility with the old manifest. But that's updated now.

selfdrive/hardware/tici/agnos.py
pd0wm: We should reimplement it, it's very slow right now.

selfdrive/hardware/tici/agnos.py
pd0wm: Think that relies on quite a few built stuff. I'm worried it might import successfully, but then fail in subtle ways.

PR #21407, opened 2021-06-25
NEOS 18

adeebshihadeh:


selfdrive/manager/process_config.py
```diff
@@ -28,7 +28,6 @@
   PythonProcess("paramsd", "selfdrive.locationd.paramsd"),
   PythonProcess("plannerd", "selfdrive.controls.plannerd"),
   PythonProcess("radard", "selfdrive.controls.radard"),
-  PythonProcess("rtshield", "selfdrive.rtshield", enabled=EON),
```
pd0wm, line 31: Also delete rtshield?

adeebshihadeh: will keep around in selfdrive/debug. it's pretty handy for testing

PR #21408, opened 2021-06-25
Quicker refreshing and use connection signals

sshane:


selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -106,27 +106,34 @@ void Networking::requestScan() {
...
 }
 
-void Networking::refreshSlot() {
-  wifiWidget->refresh();
-  an->refresh();
+void Networking::refresh() {
+  if (this->isVisible() || wifi->firstScan) {
+    wifiWidget->refresh();
+    an->refresh();
+    wifi->firstScan = false;
```
adeebshihadeh, line 113: why is this in wifi? only used in networking

sshane: We need to allow a refresh once in the background. If we get rid of `isVisible` NM can sometimes send a scan signal and the UI will update in the background at any time which isn't desired. And we can't use `ui_setup_complete` because it will be true already when we're here (since we're only here if the scan is complete)

sshane: In the future when UI updating is quick, we can remove this and only update when the view is tapped

selfdrive/ui/qt/offroad/wifiManager.h
```diff
@@ -32,11 +32,13 @@ class WifiManager : public QWidget {
 
   void requestScan();
   QVector<Network> seen_networks;
+  QMap<QDBusObjectPath, QString> known_connections;
```
adeebshihadeh, line 35: match case

PR #21411, opened 2021-06-26
Use first order filter

sshane:
Abstracts some some of the more in depth calculations into the first order filter which was added for stuff like this

Also renamed the variables to match https://en.wikipedia.org/wiki/Low-pass_filter and INDI

### DIFF
--- a/common/filter_simple.py
+++ b/common/filter_simple.py
@@ -1,12 +1,12 @@
 class FirstOrderFilter:
   # first order filter
-  def __init__(self, x0, RC, dt):
+  def __init__(self, x0, rc, dt):
     self.x = x0
     self.dt = dt
-    self.update_alpha(RC)
+    self.update_alpha(rc)
 
-  def update_alpha(self, RC):
-    self.alpha = (self.dt / RC) / (1. + self.dt / RC)
+  def update_alpha(self, rc):
+    self.alpha = (self.dt / rc) / (1. + self.dt / rc)
 
   def update(self, x):
     self.x = (1. - self.alpha) * self.x + self.alpha * x
--- a/common/filter_simple.py
+++ b/common/filter_simple.py
@@ -6,7 +6,7 @@ class FirstOrderFilter:
     self.update_alpha(rc)
 
   def update_alpha(self, rc):
-    self.alpha = (self.dt / rc) / (1. + self.dt / rc)
+    self.alpha = self.dt / (rc + self.dt)
 
   def update(self, x):
     self.x = (1. - self.alpha) * self.x + self.alpha * x
--- a/common/filter_simple.py
+++ b/common/filter_simple.py
@@ -1,8 +1,8 @@
 class FirstOrderFilter:
   # first order filter
   def __init__(self, x0, RC, dt):
-    self.dt = dt
     self.x = x0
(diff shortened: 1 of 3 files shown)


sshane:
```python
old = []
new = []
lane_width_certainty = 1
lane_width_certainty_filter = FirstOrderFilter(1.0, 0.95, DT_MDL)
to = 4
for i in range(60):
  lane_width_certainty += 0.05 * (to - lane_width_certainty)
  lane_width_certainty_filter.update(to)
  new.append(lane_width_certainty_filter.x)
  old.append(lane_width_certainty)
print(np.array_equal(old, new))
print(np.allclose(old, new))
```
```python
False
True
```

```
np.array(new) - np.array(old)
Out[178]: 
array([ 0.00000000e+00, -2.22044605e-16, -4.44089210e-16, -6.66133815e-16,
       -6.66133815e-16, -6.66133815e-16, -8.88178420e-16, -8.88178420e-16,
       -8.88178420e-16, -8.88178420e-16, -8.88178420e-16, -8.88178420e-16,
       -4.44089210e-16, -4.44089210e-16, -4.44089210e-16, -4.44089210e-16,
       -4.44089210e-16, -4.44089210e-16,  0.00000000e+00,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16,
        4.44089210e-16,  4.44089210e-16,  4.44089210e-16,  4.44089210e-16])
```

ClockeNessMnstr:
Could multiply (RC / RC) in update_alpha as well.

self.alpha = (self.dt / RC) / (1. + self.dt / RC)
 = (self.dt) / (RC + self.dt)

Makes the division safe and then RC = 0 "turns off" the filter.

common/filter_simple.py
```diff
@@ -1,9 +1,13 @@
-class FirstOrderFilter():
+class FirstOrderFilter:
   # first order filter
-  def __init__(self, x0, ts, dt):
-    self.k = (dt / ts) / (1. + dt / ts)
+  def __init__(self, x0, RC, dt):
```
pd0wm, line 3:
```suggestion
  def __init__(self, x0, rc, dt):
```

pd0wm:
The Prius segment is not engaging, so we can't verify that there are no changes in INDI. Before merging this we should make sure that segment engages again or find a new Prius segment.

PR #21413, opened 2021-06-26
fix: sim: gracefully shutdown the bridge

iamorphen:
# Description

Currently pressing `q` in the simulator to shut the bridge down "just works" but is not a true graceful exit. This MR adjusts the control flow and management of threads and processes to achieve graceful shutdown.

## In Detail

We need multiple changes to address the control flow issues.

Currently `bridge` was attempting to clean up CARLA resources in a function registered with `atexit`. However, `bridge` runs as a subprocess, so functions registered in it with `atexit` are not invoked. (We can confirm this visually--`destroy()`'s print statements were not executing.) We fix this by cleaning up the resources ourselves at the end of `bridge`.

Currently `bridge` is terminated immediately when the keyboard-polling function reads `q` from its message queue. Instead of terminating immediately, they keyboard-polling function now emits a `quit` message when it encounters `q`, and then the keyboard-polling function itself returns. In `main`, we can now join the `bridge` process and wait for it to exit gracefully.

`bridge` also spawns some threads, and `join`ing `bridge` hangs until those threads are shut down. So, we now create a queue for each thread and use the queue to issue control flow messages to the threads so that we may stop their loops and join the threads at shutdown time.

### Design Decisions for Threads

There are many choices for issuing the shutdown signal to the threads. One is global variables, but I'd like to avoid those at all costs. The others are mostly all synchronization primitives from `threading` or `multiprocessing`, for example `Event`. I chose queues because the semantics are most obvious with them. Synchronization primitives suggest that threads suspend execution for some time while waiting for a signal, which is not what we're doing here.

## Limitations

This fix doesn't cover using joystick input. I don't have access to a joystick-like device right now and so can't test any modifications to such functions.

# Verification

1. Start the simulator [as directed by the docs](https://github.com/commaai/openpilot/blob/master/tools/sim/README.md), except start openpilot with `MOUNT_OPENPILOT=1 ./start_openpilot_docker.sh` for development's sake.
2. Engage openpilot and let it run for a couple frames.
3. Press `q`.
4. The sim will show a red CAN error warning, and the terminal will show nothing concerning.

```text
root# ./bridge.py 
frame:  engaged: False ; throttle:  0.0 ; steer(c/deg):  -0.0 0.0 ; brake:  0.0
frame:  engaged: False ; throttle:  0.0 ; steer(c/deg):  -0.0 0.0 ; brake:  0.0
frame:  engaged: False ; throttle:  0.0 ; steer(c/deg):  -0.0 0.0 ; brake:  0.0
frame:  engaged: True ; throttle:  0.125 ; steer(c/deg):  -0.003 3.368 ; brake:  -0.0
frame:  engaged: True ; throttle:  0.146 ; steer(c/deg):  0.004 -3.803 ; brake:  -0.0
frame:  engaged: True ; throttle:  0.54 ; steer(c/deg):  0.004 -4.425 ; brake:  -0.0
frame:  engaged: True ; throttle:  0.548 ; steer(c/deg):  0.002 -2.045 ; brake:  -0.0
root#
```

# Future Work

`bridge.py` needs some more love. In my next passes, I'll be doing things like cleaning up the control flow in large functions more and getting rid of some cruft.

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -1,12 +1,12 @@
 #!/usr/bin/env python3
 import argparse
-import atexit
 import carla # pylint: disable=import-error
 import math
 import numpy as np
 import time
 import threading
 from cereal import log
+from multiprocessing import Process, Queue
 from typing import Any
 
 import cereal.messaging as messaging
@@ -82,9 +82,9 @@ def imu_callback(imu, vehicle_state):
   dat.sensorEvents[1].gyroUncalibrated.v = [imu.gyroscope.x, imu.gyroscope.y, imu.gyroscope.z]
   pm.send('sensorEvents', dat)
 
-def panda_state_function():
+def panda_state_function(exit_event: threading.Event):
   pm = messaging.PubMaster(['pandaState'])
-  while 1:
+  while not exit_event.is_set():
     dat = messaging.new_message('pandaState')
     dat.valid = True
     dat.pandaState = {
@@ -125,10 +125,9 @@ def gps_callback(gps, vehicle_state):
 
   pm.send('gpsLocationExternal', dat)
 
-def fake_driver_monitoring():
+def fake_driver_monitoring(exit_event: threading.Event):
   pm = messaging.PubMaster(['driverState','driverMonitoringState'])
-  while 1:
-
+  while not exit_event.is_set():
     # dmonitoringmodeld output
     dat = messaging.new_message('driverState')
     dat.driverState.faceProb = 1.0
@@ -145,9 +144,9 @@ def fake_driver_monitoring():
 
     time.sleep(DT_DMON)
 
-def can_function_runner(vs):
+def can_function_runner(vs: VehicleState, exit_event: threading.Event):
   i = 0
-  while 1:
+  while not exit_event.is_set():
     can_function(pm, vs.speed, vs.angle, i, vs.cruise_button, vs.is_engaged)
     time.sleep(0.01)
     i+=1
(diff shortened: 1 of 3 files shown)


tools/sim/bridge.py
```diff
@@ -82,9 +83,13 @@ def imu_callback(imu, vehicle_state):
   dat.sensorEvents[1].gyroUncalibrated.v = [imu.gyroscope.x, imu.gyroscope.y, imu.gyroscope.z]
   pm.send('sensorEvents', dat)
 
-def panda_state_function():
+def panda_state_function(q: Queue):
   pm = messaging.PubMaster(['pandaState'])
   while 1:
+    if not q.empty():
+      if q.get() == "quit":
+        break
```
adeebshihadeh, line 91: I think a threading.Event is much cleaner, for example this becomes:
```suggestion
  while not exit_event.is_set():
```

adeebshihadeh:
As for the joystick part, that's fine. I'd actually like to remove the joystick code (feel free to do so in a separate PR). CARLA already provides a UI with manual input, and I see no value in duplicating that.

tools/sim/bridge.py
iamorphen: Thanks; done! I've renamed some things as a result of this change, too.

iamorphen:
> As for the joystick part, that's fine. I'd actually like to remove the joystick code (feel free to do so in a separate PR). CARLA already provides a UI with manual input, and I see no value in duplicating that.

Got it; thanks for the note on this!

tools/sim/bridge.py
```diff
@@ -211,18 +211,25 @@ def bridge(q):
...
-  atexit.register(destroy)
-
   # launch fake car threads
-  threading.Thread(target=panda_state_function).start()
-  threading.Thread(target=fake_driver_monitoring).start()
-  threading.Thread(target=can_function_runner, args=(vehicle_state,)).start()
+  StoppableThread = namedtuple("StoppableThread", ["thread", "exit_event"])
+  stoppable_threads = []
+
+  event = threading.Event()
```
adeebshihadeh, line 218: we only need one event

tools/sim/bridge.py
```diff
@@ -211,18 +211,25 @@ def bridge(q):
...
-    camera.destroy()
-    vehicle.destroy()
-    print("done")
-  atexit.register(destroy)
-
   # launch fake car threads
-  threading.Thread(target=panda_state_function).start()
-  threading.Thread(target=fake_driver_monitoring).start()
-  threading.Thread(target=can_function_runner, args=(vehicle_state,)).start()
+  StoppableThread = namedtuple("StoppableThread", ["thread", "exit_event"])
```
adeebshihadeh, line 215: no need for this. just use a single event and set it

iamorphen: I agree that will work. I chose distinct events to preclude the possibility of a future bug of one thread influencing the flow control of another since the shared event is mutable. If you don't think we need to worry about that much, please confirm, and I'll make the change.

adeebshihadeh: yeah, that's really not a concern

iamorphen: Thanks for confirming. I'll make the change when I'm back in front of my dev machine.

iamorphen: Done.

PR #21415, opened 2021-06-27
HttpRequest:  use a single QNetworkAccessManager instance

deanlee:
https://doc.qt.io/qt-5/qnetworkaccessmanager.html#details

> . It contains the proxy and cache configuration, as well as the signals related to such issues, and reply signals that can be used to monitor the progress of a network operation. One QNetworkAccessManager instance should be enough for the whole Qt application.

### DIFF
--- a/selfdrive/ui/qt/api.cc
+++ b/selfdrive/ui/qt/api.cc
@@ -9,7 +9,10 @@
 #include <QDebug>
 #include <QFile>
 #include <QJsonDocument>
+#include <QNetworkAccessManager>
+#include <QNetworkReply>
 #include <QNetworkRequest>
+#include <QTimer>
 
 #include "selfdrive/common/params.h"
 #include "selfdrive/common/util.h"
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Is this really necessary?

deanlee:
I have a cursory look at the source code https://code.woboq.org/qt5/qtbase/src/network/access/qnetworkaccessmanager.cpp.html,  I think it makes sense for the documentation to suggest using only one QNetworkAccessManager for whole application.The initialization of qnam has to do a lot of work, If we only use it in a single thread, there is no need to create multiple instance

pd0wm:
Does this (measurably) affect ui startup times? I can imaging we shouldn't be creating them for every request, but we only create them in the constructors.

deanlee:
it does not affect. it will be instantiated on first use, not the startup.

adeebshihadeh:
If this doesn't affect startup times, I think the way we do it is fine.

deanlee:
yes,it's an inconsequential change.

PR #21418, opened 2021-06-27
VW MQB: Add FW for 2016 Volkswagen Tiguan

jyoung8607:
Add firmware for the EU/RoW 2016 Volkswagen Tiguan.

**Dongle ID:** 21aed36c451a54e9

Thanks to community Tiguan owner mikkelamby.

I'm not really sure what to do with README on this. It's a new model-year for us, but it falls into the confusing area I described in #20695. I could asterisk it somewhat like the Passat, add a footnote to the 2020 mentioning EU/RoW 2016 is supported too?


RELEASES.md
```diff
@@ -5,6 +5,7 @@ Version 0.8.6 (2021-XX-XX)
  * Škoda Octavia RS 2016 support thanks to jyoung8607!
  * Toyota Alphard 2020 support thanks to belm0!
  * Volkswagen Golf SportWagen 2015 support thanks to jona96!
+ * Volkswagen Tiguan 2016 support thanks to jyoung8607!
```
adeebshihadeh, line 8: we normally don't add new MYs to the release notes

adeebshihadeh:
```suggestion
```

jyoung8607:
Could swear we'd done MYs in the release notes before, but I guess not. Whichever way you want is fine.

Any thoughts on the README? I know it's supposed to be US-centric, but it seems odd to leave out entirely.

adeebshihadeh:
The README is a bit odd since it's US-centric for model years plus models that don't exist in the U.S. Let's merge this to get the FW in, and we'll think about how to handle this case going forward.

PR #21420, opened 2021-06-27
Extend FirstOrderFilter (LPF) 

ClockeNessMnstr:
https://github.com/commaai/openpilot/pull/21411
looked like a good time to jump in with some updates / extensions. 

Extend filter with reset function.
Add HZ mode for setting time constant alpha with a cutoff frequency.
add a reset function with new initial value
make division safe with 0hz or RC=0 (possibly important with dynamic updates)
RC=0 is passthrough
0hz will output x0

### DIFF
--- a/common/filter_simple.py
+++ b/common/filter_simple.py
@@ -1,8 +1,8 @@
 class FirstOrderFilter:
   # first order filter
   def __init__(self, x0, RC, dt):
-    self.dt = dt
     self.x = x0
+    self.dt = dt
     self.update_alpha(RC)
 
   def update_alpha(self, RC):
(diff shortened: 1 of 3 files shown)


pd0wm:
Any immediate use for this?

ClockeNessMnstr:
tagging along with #21411

I think these changes are useful if not needed to use the filter in a controller. 

I'm using a filter in my PID controller, looks like Shane is adding it in INDI with this branch and merging to master. 

a reset function is needed for using it in a controller if the function can be skipped while disengaged to avoid a discontinuity. 
Hz mode I find much more intuitive when looking at a graph and deciding what frequency is noise vs inputs to select a cutoff frequency instead of a time constant.
Also using it with an update-able "alpha" needs to be safe division.

pd0wm:
Converting this to draft until we decide on the other PR. Won't merge this without an immediate use case.

PR #21422, opened 2021-06-27
PID backward-Euler implementation. add Derivative, track actual actuator output, back-calculate integrator. 

ClockeNessMnstr:
I have been running this controller for a few weeks. 

Using a derivative term and back-calculating the integrator based on the actual commanded output from saturating the torque ramp up allows for much better tunes without hitting instabilities from actuator / safety saturation. 

There is also a input filter to reduce the noise in the P-term. most of the noise seems to be in the Desired angle so it should not be vehicle dependent.

Includes/depends/related:
https://github.com/commaai/openpilot/pull/21336
https://github.com/commaai/openpilot/pull/21005

Cereal updates for kdBP, kdV, pid.d

PI portion of the controller has the same behavior with current tunes. Only difference would be the 5hz LPF on the error currently.

### DIFF
--- a/selfdrive/car/chrysler/carcontroller.py
+++ b/selfdrive/car/chrysler/carcontroller.py
@@ -16,6 +16,9 @@ class CarController():
 
     self.packer = CANPacker(dbc_name)
 
+  def get_last_output(self):
+    return self.apply_steer_last / CarControllerParams.STEER_MAX
+
   def update(self, enabled, CS, actuators, pcm_cancel_cmd, hud_alert):
     # this seems needed to avoid steering faults and to force the sync with the EPS counter
     frame = CS.lkas_counter
--- a/selfdrive/car/ford/carcontroller.py
+++ b/selfdrive/car/ford/carcontroller.py
@@ -11,6 +11,7 @@ TOGGLE_DEBUG = False
 
 class CarController():
   def __init__(self, dbc_name, CP, VM):
+    self.apply_steer_last = 0.0
     self.packer = CANPacker(dbc_name)
     self.enabled_last = False
     self.main_on_last = False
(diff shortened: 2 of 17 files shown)


ClockeNessMnstr:
Screenshot from 2021-06-27 18-34-28

recoveries from steering overrides short/sharp & windup/long tests.

ClockeNessMnstr:
Screenshot from 2021-06-27 18-31-50

highway stretch, long bends / straights

ClockeNessMnstr:
a3586311666ae480|2021-06-27--18-15-51

ClockeNessMnstr:
Screenshot from 2021-06-26 20-09-52

35mph max torque turns

ClockeNessMnstr:
a3586311666ae480|2021-06-27--18-15-51

pd0wm:
1) Why a LPF on the error? If there is noise on the desired angle this should be fixed in the planner, not masked with a low pass filter. Those changes might provide small benefits on the short term, but make the whole stack hard to reason about and debug.
2) Why not a Kalman filter to compute the rate of change? Many cars also provide a steering rate signal we can use to provide a better estimate.
3) Do you have any comparison drives before/after these changes that show that the (average) tracking error is reduced? You need to measure this error before it enters the LPF.

ClockeNessMnstr:
I'm just going to pull out the error filter. 

I'm seeing a small improvement down to 1.6hz (as far as I had time for) using an average of  (1+err)^2 but it's between 1.2-1.6 and my results were depending pretty heavy on traffic.

Controller was still working fine without it. 

I'll end up trying the lowest filter that doesn't cause problems and seeing how far I can push a tune on it. If it provides a better tune than without a filter I'll make a new PR for the filter.

adeebshihadeh:
any update on this?

ClockeNessMnstr:
> any update on this?

Extra filters are out. 
I can try to capture some comparisons like the tests in the screenshots. 
Haven't had much time the past few weeks.
What do we want to see?
There's two parts essentially. 
PID implementation and integrator back calculation

adeebshihadeh:
These are all very different changes. I'm not sure we're interested in adding a derivative term, but I think uses the final requested torque in the controller makes sense. Can you make a clean PR that only does that?

PR #21427, opened 2021-06-28
sensors_qcom2:  fix segmentation fault, use orphans to build sensor events

deanlee:
If get_event fails, it will return an empty `SensorEventData` with default type `SensorEventData::ACCELERATION(0)`, that will cause segmentation fault when get value from `getAcceleration().getV()`

### DIFF
--- a/selfdrive/sensord/sensors/bmx055_accel.cc
+++ b/selfdrive/sensord/sensors/bmx055_accel.cc
@@ -43,7 +43,7 @@ fail:
   return ret;
 }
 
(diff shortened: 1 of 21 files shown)


pd0wm:
Sensord will be rewritten soon to be interrupt based. We will probably split out every sensor in a separate socket, to make processing on the receiving end simpler. Like to not touch sensord in the meantime to not introduce regressions, it has been quite stable (as long as you properly check the messages on the receiving end).

PR #21432, opened 2021-06-28
Wait for adapter to setup WifiManager

sshane:
Should start a scan immediately when the correct wifi adapter is detected, instead of checking every 5 seconds previously

- Wait for a `DeviceAdded` signal to start scanning immediately (on setup)
- When adapter is available (should be always), get tethering status as normal on init. In setup it is assumed false.

### DIFF
--- a/selfdrive/ui/qt/offroad/networking.cc
+++ b/selfdrive/ui/qt/offroad/networking.cc
@@ -26,26 +26,7 @@ void NetworkStrengthWidget::paintEvent(QPaintEvent* event) {
 Networking::Networking(QWidget* parent, bool show_advanced) : QWidget(parent), show_advanced(show_advanced) {
   main_layout = new QStackedLayout(this);
 
-  QLabel* warning = new QLabel("Network manager is inactive!");
-  warning->setAlignment(Qt::AlignCenter);
-  warning->setStyleSheet(R"(font-size: 65px;)");
-
-  main_layout->addWidget(warning);
-
-  QTimer* timer = new QTimer(this);
-  QObject::connect(timer, &QTimer::timeout, this, &Networking::requestScan);
-  timer->start(5000);
-  attemptInitialization();
-}
-
-void Networking::attemptInitialization() {
-  // Checks if network manager is active
-  try {
-    wifi = new WifiManager(this);
-  } catch (std::exception &e) {
-    return;
-  }
-
+  wifi = new WifiManager(this);
   connect(wifi, &WifiManager::wrongPassword, this, &Networking::wrongPassword);
   connect(wifi, &WifiManager::refreshSignal, this, &Networking::refresh);
 
@@ -66,7 +47,6 @@ void Networking::attemptInitialization() {
   wifiWidget->setObjectName("wifiWidget");
   connect(wifiWidget, &WifiUI::connectToNetwork, this, &Networking::connectToNetwork);
   vlayout->addWidget(new ScrollView(wifiWidget, this), 1);
-
   main_layout->addWidget(wifiScreen);
 
   an = new AdvancedNetworking(this, wifi);
@@ -89,29 +69,12 @@ void Networking::attemptInitialization() {
     }
   )");
   main_layout->setCurrentWidget(wifiScreen);
-  ui_setup_complete = true;
-  wifi->requestScan();
-}
-
-void Networking::requestScan() {
-  if (!this->isVisible()) {
-    return;
-  }
-  if (!ui_setup_complete) {
-    attemptInitialization();
-    if (!ui_setup_complete) {
-      return;
-    }
-  }
-  wifi->requestScan();
 }
 
 void Networking::refresh() {
-  if (this->isVisible() || firstRefresh) {
-    wifiWidget->refresh();
-    an->refresh();
-    firstRefresh = false;
-  }
+  qDebug() << "REFRESHED UI!";
+  wifiWidget->refresh();
+  an->refresh();
 }
 
 void Networking::connectToNetwork(const Network &n) {
--- a/selfdrive/ui/qt/offroad/networking.cc
+++ b/selfdrive/ui/qt/offroad/networking.cc
@@ -72,7 +72,6 @@ Networking::Networking(QWidget* parent, bool show_advanced) : QWidget(parent), s
 }
 
 void Networking::refresh() {
-  qDebug() << "REFRESHED UI!";
   wifiWidget->refresh();
   an->refresh();
 }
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -89,29 +68,13 @@ void Networking::attemptInitialization() {
...
-    if (!ui_setup_complete) {
-      return;
-    }
-  }
-  wifi->requestScan();
 }
 
 void Networking::refresh() {
-  if (this->isVisible() || firstRefresh) {
+  if (wifi->seen_networks.size() != 0) {
```
adeebshihadeh, line 74: does this make a difference? refresh should be performant enough to not need this

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -69,13 +70,16 @@ WifiManager::WifiManager(QWidget* parent) : QWidget(parent) {
   qDBusRegisterMetaType<Connection>();
   qDBusRegisterMetaType<IpConfig>();
   connecting_to_network = "";
-  adapter = get_adapter();
 
-  bool has_adapter = adapter != "";
-  if (!has_adapter) {
-    throw std::runtime_error("Error connecting to NetworkManager");
+  bus.connect(nm_service, nm_path, nm_iface, "DeviceAdded", this, SLOT(deviceAdded(QDBusObjectPath)));
```
adeebshihadeh, line 74: do we need this if we already have the adapter?

selfdrive/ui/qt/offroad/networking.cc
sshane: This was so we don't show a blank screen when there's no networks around, but we can instead add the "Scanning for networks" whenever there's no networks

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -387,10 +380,28 @@ void WifiManager::propertyChange(const QString &interface, const QVariantMap &pr
...
+}
+
+void WifiManager::updateUI() {
+  if (this->isVisible()) {
     emit refreshSignal();
   }
 }
 
+void WifiManager::deviceAdded(const QDBusObjectPath &path) {
+  if (isWirelessAdapter(path)) {
```
adeebshihadeh, line 399: can this be run multiple times?

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -39,9 +39,10 @@ const QString ipv4config_iface       = "org.freedesktop.NetworkManager.IP4Config
 
 const QString nm_service             = "org.freedesktop.NetworkManager";
 
-const int state_connected = 100;
 const int state_need_auth = 60;
+const int state_connected = 100;
 const int reason_wrong_password = 8;
+const int device_type_wifi = 2;
```
adeebshihadeh, line 45: is there a header with all these constants we can include?

selfdrive/ui/qt/offroad/wifiManager.cc
sshane: Yes, there's another adapter on PC that's called `NM_DEVICE_TYPE_WIFI_P2P` that's added at the same time

adeebshihadeh: shouldn't this only be called once?

selfdrive/ui/qt/offroad/wifiManager.cc
sshane: Yeah, this header would be nice to have: https://github.com/NetworkManager/NetworkManager/blob/main/src/libnm-core-public/nm-dbus-interface.h

sshane: I'll do this in another PR

selfdrive/ui/qt/offroad/wifiManager.cc
sshane: `deviceAdded`? It's called twice when the WiFi adapter is plugged in, I'll check the device

sshane: Made it so that it will only get the adapter once, which if we're certain it never is removed is okay (it's a different path each time it's added)

pd0wm:
> Tethering is assumed to be false on Networking init (it never persisted through reboots, and you aren't going to kill and restart UI)

I don’t think you can make this assumption

selfdrive/ui/qt/offroad/wifiManager.cc
sshane: I just added a commit to make it resilient to losing the adapter (equivalent to unplugging and replugging in the USB on PC) by reconnecting the signals using the old adapter path. If we don't care about this, I can revert the commit.

sshane:
@pd0wm True, it would just require a new signal to Networking as we can't always know on start up since it's non-blocking. I'll add it back in another PR when I introduce more signals as I didn't want to make this PR too big, and we can't keep the same functionality without adding more code

adeebshihadeh:
We shouldn't remove functionality like that and bring it back in another PR. Let's handle it in this PR.

selfdrive/ui/qt/offroad/wifiManager.h
```diff
@@ -33,7 +33,7 @@ class WifiManager : public QWidget {
   void requestScan();
   QVector<Network> seen_networks;
   QMap<QDBusObjectPath, QString> knownConnections;
-  QString ipv4_address;
+  QString ipv4_address = "";
```
adeebshihadeh, line 36: should already be empty

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -69,13 +70,36 @@ WifiManager::WifiManager(QWidget* parent) : QWidget(parent) {
...
+    if (!adapter.isEmpty() && this->isVisible()) {
+      requestScan();
+    }
+  });
+  timer->start(5000);
+
+  // Create dbus interface for tethering button. This populates the introspection cache,
+  // making sure all future creations are non-blocking
+  // https://bugreports.qt.io/browse/QTBUG-14485
+  QDBusInterface(nm_service, nm_settings_path, nm_settings_iface, bus);
```
adeebshihadeh, line 99: does this delay startup?

sshane: Yes, but it takes 1.1-1.45 ms. Future calls take 0.08 to 0.09 ms so this could probably be removed

PR #21439, opened 2021-06-29
modeld: fill_model: make sure tidx+1 doesn't exceed TRAJECTORY_SIZE-1

deanlee:


HaraldSchafer:
@adeebshihadeh I want to run jenkins on this before merge, any quick way to do that?

PR #21444, opened 2021-06-30
Car port: AP1 Tesla Model S

kennedyshead:


kennedyshead:
Kudos to @BogGyver for all help, and making the port possible.

adeebshihadeh:
Does this branch actually work? You're setting the safety param, but the upstream Tesla safety doesn't use it.

kennedyshead:
Yes it works, as showen in video. Bit there are some minor bugs I`m working on regarding LKAS.

adeebshihadeh:
Where did you post this info?

kennedyshead:
I referred to the cabana video in the testroute, but I can post a "kinda-working" video if you want! For dashCam mode (wich this is) it works real nice though.

kennedyshead:
So... Did a testdrive with long activated and it works.
The only two bugs that I can find is that when I manually turn the wheel there is a message "LKAS fault: restart to engage..."(something like that like that), and when I disengage by turning the wheel the same message and a red flashing. I guess it should not disengage hen you turn the wheel, that should be considered a steering correction?

selfdrive/car/tesla/interface.py
```diff
@@ -35,6 +35,11 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None):
       ret.wheelbase = 2.959
       ret.centerToFront = ret.wheelbase * 0.5
       ret.steerRatio = 13.5
+    elif candidate == CAR.AP1_MODELS:
```
robbederks, line 38: let's merge the two since they are the same: `if candidate in [CAR.AP2_MODELS, CAR.AP1_MODELS]:`

kennedyshead: :+1:

robbederks:
So just adding the fingerprint is enough to get lateral working on AP1 without errors? Or is the manual steering override LKAS fault etc still present?

kennedyshead:
> So just adding the fingerprint is enough to get lateral working on AP1 without errors? Or is the manual steering override LKAS fault etc still present?

After https://github.com/commaai/openpilot/pull/21469 the lateral (and long actually) works great, driven 600km with no major issues.

There is a more extensive fingerprint that I'm trying out but part from that this port is good to go `dashcamOnly = False` imo

robbederks:
Great, if you have the updated fingerprint I think this is probably good to merge.
We still have `dashcamOnly` because of the lack of user steer torque blending, which is a pretty big deviation from the openpilot experience on all other supported cars.

kennedyshead:
Fingerprint was already here :D tested and works!

silvercurrent:
> So... Did a testdrive with long activated and it works, without any problems at all! Its so much better with VOACC then stock that I almost cried a little.
> The only two bugs that I can find is that when I manually turn the wheel there is a message "LKAS fault: restart to engage..."(something like that like that), and when I disengage by turning the wheel the same message and a red flashing. I guess it should not disengage hen you turn the wheel, that should be considered a steering correction?

@kennedyshead , how do you enable VOACC?

kennedyshead:
> > So... Did a testdrive with long activated and it works, without any problems at all! Its so much better with VOACC then stock that I almost cried a little.
> > The only two bugs that I can find is that when I manually turn the wheel there is a message "LKAS fault: restart to engage..."(something like that like that), and when I disengage by turning the wheel the same message and a red flashing. I guess it should not disengage hen you turn the wheel, that should be considered a steering correction?
> 
> @kennedyshead , how do you enable VOACC?

Its all enabled in the "stable" fork in my repo. If you try it you should be aware that it is not really stable yet despite the name, so be very careful.

PR #21447, opened 2021-07-01
Use AP signals

sshane:
Replaces: https://github.com/commaai/openpilot/pull/21431


sshane:
Realized if we use signals, we can't stop receiving signals in the background if we want to keep the network list up to date, and it takes 5 ms per AP to add and get all data we need (checking strength, security, ssid, etc), so the best solution for now is to only update when the view is visible and call the DBUS command instead of using the signals in the background.

PR #21455, opened 2021-07-01
Networking: populate tethering password

sshane:
On clicking the edit or toggle, a tethering connection is created if it doesn't already exist. Then on editing the password we get and modify the connection settings.

### DIFF
--- a/selfdrive/ui/qt/offroad/wifiManager.cc
+++ b/selfdrive/ui/qt/offroad/wifiManager.cc
@@ -275,9 +275,9 @@ bool WifiManager::isWirelessAdapter(const QDBusObjectPath &path) {
 }
 
 void WifiManager::requestScan() {
-  QDBusInterface nm(NM_DBUS_SERVICE, adapter, NM_DBUS_INTERFACE_DEVICE_WIRELESS, bus);
-  nm.setTimeout(DBUS_TIMEOUT);
-  nm.call("RequestScan",  QVariantMap());
+  QDBusInterface nm(nm_service, adapter, wireless_device_iface, bus);
+  nm.setTimeout(dbus_timeout);
+  nm.call("RequestScan", QVariantMap());
 }
 
 uint WifiManager::get_wifi_device_state() {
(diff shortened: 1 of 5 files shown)


selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -498,6 +500,18 @@ bool WifiManager::tetheringEnabled() {
...
+QString WifiManager::getTetheringPassword() {
+  const QDBusObjectPath &path = getConnectionPath(tethering_ssid.toUtf8());
+  if (!path.path().isEmpty()) {
+    QDBusInterface nm(nm_service, path.path(), nm_settings_conn_iface, bus);
+    nm.setTimeout(dbus_timeout);
+
+    const QDBusReply<QMap<QString, QMap<QString, QVariant>>> response = nm.call("GetSecrets", "802-11-wireless-security");
+    return response.value().value("802-11-wireless-security").value("psk").toString();
+  }
+  return tetheringPassword;
```
adeebshihadeh, line 512: probably better to return an empty string here

sshane:
Takes approx 10 ms to create the tethering connection on init, but it's expected to only run once, ever. Is this okay?

If not, it can be moved to when the scan is complete.

adeebshihadeh:
No, let's do it once the toggle is flipped for the first time (shouldn't block the UI though) or perhaps add it to agnos.

sshane:
We then need to add it to the edit password as well, so two duplicate checks, is the only reason I didn't do that

sshane:
Waiting on https://github.com/commaai/openpilot/pull/21453

selfdrive/ui/qt/widgets/input.h
```diff
@@ -14,10 +14,9 @@ class InputDialog : public QDialog {
 
 public:
   explicit InputDialog(const QString &prompt_text, QWidget* parent = 0);
-  static QString getText(const QString &prompt, int minLength = -1);
+  static QString getText(const QString &prompt, int minLength = -1, const QString &defaultText = "");
   QString text();
   void setMessage(const QString &message, bool clearInputField = true);
-  void setMinLength(int length);
```
adeebshihadeh, line 20: why delete this? believe this was to match a similar class in Qt

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -413,6 +415,9 @@ void WifiManager::activateWifiConnection(const QString &ssid) {
 
 // Functions for tethering
 void WifiManager::addTetheringConnection() {
+  if (isKnownConnection(tethering_ssid)) {
```
adeebshihadeh, line 418: I would check this before calling addTetheringConnection

PR #21457, opened 2021-07-02
Tethering: Edit password while enabled

sshane:


selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -152,12 +152,7 @@ void AdvancedNetworking::refresh() {
 }
 
 void AdvancedNetworking::toggleTethering(bool enable) {
-  if (enable) {
-    wifi->enableTethering();
-  } else {
-    wifi->disableTethering();
-  }
-  editPasswordButton->setEnabled(!enable);
+  wifi->toggleTethering(enable);
```
adeebshihadeh, line 155: toggleTethering implies you're just flipping the state. setTetheringEnabled would be a better name

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -117,13 +118,13 @@ AdvancedNetworking::AdvancedNetworking(QWidget* parent, WifiManager* wifi): QWid
   main_layout->addWidget(back, 0, Qt::AlignLeft);
 
   // Enable tethering layout
-  ToggleControl *tetheringToggle = new ToggleControl("Enable Tethering", "", "", wifi->tetheringEnabled());
+  ToggleControl *tetheringToggle = new ToggleControl("Enable Tethering", "", "", wifi->isTetheringEnabled());
   main_layout->addWidget(tetheringToggle);
-  QObject::connect(tetheringToggle, &ToggleControl::toggleFlipped, this, &AdvancedNetworking::toggleTethering);
+  QObject::connect(tetheringToggle, &ToggleControl::toggleFlipped, [=](bool enabled) { emit setTetheringEnabled(enabled); });
```
adeebshihadeh, line 123: we should probably add a new signal to the toggle that also has the state

PR #21469, opened 2021-07-02
FIX: Tesla spamming "LKAS fault"

kennedyshead:
EAC_INHIBITED is not an actual error, when this is removed all "LKAS fault" errors are fixed.

**Description**
Tesla is throwing "LKAS fault" errors.

**Verification**
Testdrive and tried to get the errors to show up, but they are gone.

**Route**
Route: 6c14ee12b74823ce|2021-07-02--15-27-53


robbederks:
How did you reproduce this before this change? I think I put this in there when we didn't disengage on steer override, but since that isn't a thing anymore it should be fine to remove.

I would like to know exactly what actually causes this "INHIBIT" to happen though

robbederks:
I'm running this code daily, and haven't run into this afaik

kennedyshead:
Turning the wheel fast to max produces the error, could be AP1 only if you dont see it? I also run it daily now but with AP1.

robbederks:
I see, not engaged I assume? Haven't noticed that. I'm fine with merging the change though!

PR #21473, opened 2021-07-03
lag compensation by steering actuator delay

ClockeNessMnstr:
The current lag compensation doesn't seem to be functioning correctly. 

actuator delay can be compensated by simply using the curvatures + curvature_rates future values. 
The biggest improvement was with KL which was previously unusable. So far KL is now as usable or better than the regular model. 
Previously standard model would be entering turns correctly and then KL would be very early. 

I'm very interested to understand how psi is supposed to work but it should not be unstable at delay values approaching 0. 
it also seems that psi includes vEgo so could be able to compensate for long acceleration but I could not tell if accelerating in a turn affected the logged psi values.

### DIFF
--- a/selfdrive/controls/lib/drive_helpers.py
+++ b/selfdrive/controls/lib/drive_helpers.py
@@ -71,14 +71,15 @@ def get_lag_adjusted_curvature(CP, v_ego, psis, curvatures, curvature_rates):
   # TODO this needs more thought, use .2s extra for now to estimate other delays
   delay = CP.steerActuatorDelay + .2
   current_curvature = curvatures[0]
-  psi = interp(delay, T_IDXS[:CONTROL_N], psis)
-  desired_curvature_rate = curvature_rates[0]
+  desired_curvature = interp(delay, T_IDXS[:CONTROL_N], curvatures)
+  #psi = interp(delay, T_IDXS[:CONTROL_N], psis)
+  desired_curvature_rate = interp(delay, T_IDXS[:CONTROL_N], curvature_rates)
 
   # MPC can plan to turn the wheel and turn back before t_delay. This means
   # in high delay cases some corrections never even get commanded. So just use
   # psi to calculate a simple linearization of desired curvature
-  curvature_diff_from_psi = psi / (max(v_ego, 1e-1) * delay) - current_curvature
-  desired_curvature = current_curvature + 2 * curvature_diff_from_psi
+  #curvature_diff_from_psi = psi / (max(v_ego, 1e-1) * delay) - current_curvature
+  #desired_curvature = current_curvature + 2 * curvature_diff_from_psi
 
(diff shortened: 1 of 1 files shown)


HaraldSchafer:
1. There is no such thing as 0 delay, so the stability at 0 does not concern me, at the least the delay will always be larger then the timestep/2. In which case the divide by delay causes no issues.
2. The lateral planning is done under the assumption of constant speed. So long acceleration is not considered. Perhaps one day it should be, I don't think it matters much.
3. You cannot compensate for the actuator delay by simply taking future curvature values. This is explained in the comment. If the delay is very high, like 0.6s as in subarus, the plan would be to turn the wheel and turn it back in less than 0.6s. So if you take the curvature at T_delay it would already be 0 again. And the car would never turn.

Why do you say current lag compensation is not working well? What issues are you having?

ClockeNessMnstr:
The .2 second hardcoded increase in delay seems to be there for the purpose of avoiding using delays in the range of 0-.2
using a delay of 0 should still be controllable (but not optimal) even though we would be ignoring physical actuator delay. 

Actuator delay for PID control of steering angle using torque should be the time between the PID loop and physical torque output on the EPS.
Using future curvature I found that actual value around 0.1 without the +.2 seems to give good timing entering turns on the Forester which is much closer to the EPS lag which is not delayed near 0.3 seconds.

Reading future values does not necessarily mean that the command will never be actuated unless there are significant discontinuities, which does not seem to be the case as the values appear to be smoothly shifted/scaled from curvature[0:T]. even with a delay of .6 seconds for example, the period between t+0 and t+0.6 will have been commanded between t minus 0.6 and t minus 0. 

I also don't know how psi should work in the current compensation. is it angular offset from the path?

HaraldSchafer:
No, a delay of 0 the way it is written here does not make sense. 0.2 was picked by doing analysis on all the data.

Using future curvature only works for low delay values, as explained above. The delay values currently set were gotten from 100k segments of driving data, your car may be different than the subaru's we tested.

Psi is the integral of curvature.

ClockeNessMnstr:
I must have broke something while testing tired that made it unstable. 

It makes sense now how it's calculated and should be stable for all values other than 0 barring any accuracy issues from interpolating psi closer to zero which I don't think would matter too much.

Switched it back for a sanity check but kept the same steering delay and it was fine/stable.

HaraldSchafer:
sounds good.

rgilman33:
> No, a delay of 0 the way it is written here does not make sense. 0.2 was picked by doing analysis on all the data.
> 
> Using future curvature only works for low delay values, as explained above. The delay values currently set were gotten from 100k segments of driving data, your car may be different than the subaru's we tested.
> 
> Psi is the integral of curvature.

@HaraldSchafer what sort of things is that .2s hardcoded delay composed of? Is this various lags within OP itself?

HaraldSchafer:
Camera image processing, model running on the image, planner running on model outputs, etc...

rgilman33:
> Camera image processing, model running on the image, planner running on model outputs, etc...

Perfect, thank you

PR #21474, opened 2021-07-03
Pj improvement

ghost:
Some plot juggler additions I am using locally but don't want to rewrite all the time

-- parse longer parts of a route than a single segment
-- ~~cache the logs so if you parse the same route again, loading becomes much faster~~

### DIFF
--- a/tools/plotjuggler/juggle.py
+++ b/tools/plotjuggler/juggle.py
@@ -50,7 +50,7 @@ def juggle_route(route_name, segment_number, segment_count, qlog, can, layout):
     logs = r.qlog_paths() if qlog else r.log_paths()
 
   if segment_number is not None:
-    logs = logs[segment_number:segment_number + segment_count]
+    logs = logs[segment_number:segment_number+segment_count]
 
   if None in logs:
     fallback_answer = input("At least one of the rlogs in this segment does not exist, would you like to use the qlogs? (y/n) : ")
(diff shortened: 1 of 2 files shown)


tools/plotjuggler/juggle.py
```diff
@@ -13,7 +13,11 @@
 
 juggle_dir = os.path.dirname(os.path.realpath(__file__))
 
-def load_segment(segment_name):
+def load_segment(segment_name, cache):
+  if cache:
+    env = os.environ
+    env["FILEREADER_CACHE"] = "1"
```
adeebshihadeh, line 19: just set the environment variable instead of adding the --cache flag, like this:
`FILEREADER_CACHE=1 ./juggle.py <route>`

tools/plotjuggler/juggle.py
```diff
@@ -42,33 +42,28 @@ def start_juggler(fn=None, dbc=None, layout=None):
...
       return
 
-  all_data = []
   pool = multiprocessing.Pool(24)
-  for d in pool.map(load_segment, logs):
-    all_data += d
-
-  if not can:
-    all_data = [d for d in all_data if d.which() not in ['can', 'sendcan']]
+  all_data = [d for d_list in pool.map(load_segment, logs) for d in d_list if d.which() not in ['can', 'sendcan'] or can]
```
adeebshihadeh, line 66: revert this, it's less readable

grekiki: Another idea is using 
```
all_data = [d for d_list in pool.map(load_segment, logs) for d in d_list]
if not can:
  all_data = [d for d in all_data if d.which() not in ['can', 'sendcan']]
```

PR #21476, opened 2021-07-04
Honda Civic Bosch longitudinal support

cfranyota:
This is needed for long control.

### DIFF
--- a/selfdrive/car/honda/hondacan.py
+++ b/selfdrive/car/honda/hondacan.py
@@ -182,7 +182,7 @@ def create_ui_commands(packer, pcm_speed, hud, car_fingerprint, is_metric, idx, 
     }
     commands.append(packer.make_can_msg('RADAR_HUD', bus_pt, radar_hud_values, idx))
     
-    if car_fingerprint in CAR.CIVIC_BOSCH:
+    if car_fingerprint == CAR.CIVIC_BOSCH:
       commands.append(packer.make_can_msg("LEGACY_BRAKE_COMMAND", bus_pt, {}, idx))
 
   return commands
--- a/selfdrive/car/honda/hondacan.py
+++ b/selfdrive/car/honda/hondacan.py
@@ -1,7 +1,7 @@
 from selfdrive.car.isotp_parallel_query import IsoTpParallelQuery
 from selfdrive.swaglog import cloudlog
 from selfdrive.config import Conversions as CV
-from selfdrive.car.honda.values import HONDA_BOSCH
+from selfdrive.car.honda.values import HONDA_BOSCH, CAR
 
 # CAN bus layout with relay
 # 0 = ACC-CAN - radar side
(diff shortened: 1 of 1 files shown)


selfdrive/car/honda/hondacan.py
```diff
@@ -181,6 +181,9 @@ def create_ui_commands(packer, pcm_speed, hud, car_fingerprint, is_metric, idx,
       'SET_TO_1' : 0x01,
     }
     commands.append(packer.make_can_msg('RADAR_HUD', bus_pt, radar_hud_values, idx))
+    
+    if car_fingerprint in CAR.CIVIC_BOSCH:
```
pd0wm, line 185: I doubt this will ever be true since `CAR.CIVIC_BOSCH` is string, not a list. Are you sure this change does anything?

cfranhonda: Without this, the car won't accelerate/brake. The civic_bosch guys can vouch for this one.

pd0wm: Ah, my bad. It would be true, but so would a bunch of other subsets. You want either:

`if car_fingerprint in (CAR.CIVIC_BOSCH, )` or `if car_fingerprint == CAR.CIVIC_BOSCH`

adeebshihadeh:
What exactly does this do and why is it needed? I believe @csouers also knows about this.

cfranhonda:
> What exactly does this do and why is it needed? I believe @csouers also knows about this

https://github.com/commaai/opendbc/blob/80e7b94ae9f7451235bd802fec04d2f5e601083b/honda_civic_hatchback_ex_2017_can_generated.dbc#L451

It's for gas/brake control for vision radar. Without this on vision radar the car just sits there when you engage,  no gas or brakes.

csouers:
Civic Bosch and the Diesel variant used a blank 0x1FA frame (except the Beep bit) during the interim years when they were switching the line to Bosch only. IIRC, the 2019+ Civics don't send it at all. Probably won't hurt to send it on those cars but it would be nice to test that and check other cars (Hatch vs Sedan between 2018 and 2019) for it.

cfranhonda:
> Civic Bosch and the Diesel variant used a blank 0x1FA frame (except the Beep bit) during the interim years when they were switching the line to Bosch only. IIRC, the 2019+ Civics don't send it at all. Probably won't hurt to send it on those cars but it would be nice to test that and check other cars (Hatch vs Sedan between 2018 and 2019) for it.
> 
> @cfranhonda have you found a fault signal for when this isn't being sent, but should be? It doesn't trigger the usual brake_error bits that are already checked.

I'm not getting any faults without this. Everything looks happy just doesn't move. Also, a 19' sedan tried it, it works as well.

cfranhonda:
don't forget this is needed for Civics to join the vision party

Issue #19747, opened 2021-01-13
Beep when using assistance lane change when it shouldn't.


JustStig:
**Describe the bug**
Prior to 0.8.1, when using the assisted lane change, there would be no beeps or alerts, but now in 0.8.1 when using assisted lane change, the comma would beep as if the driver wasn't paying attention during assisted lane changes.

This behavior (beeps) did not happen in 0.8.0 or the prior version.

**How to reproduce or log data**

Using 0.8.1 using assisted lane change at highway speed, the car would beep during lane change.

**Expected behavior**
No beeps or sound made during lane change

**Device/Version information (please complete the following information):**
 - Device: Comma 
 - Dongle ID: 05f2b20ed4718f04
 - Route: Any Route
 - Timestamp: Any time
 - Version: 0.8.1
 - Car make/model: 2020 Hyundai Santa Fe Ultimate

**Additional context**
This did not happen in .8.0 or prior version.

adeebshihadeh:
Is the sound from the car or the comma two?

JustStig:
The sound from the car and the dash display "keep hands on steering wheel"

adeebshihadeh:
Most likely a consequence of #2490. The dash alert is intentional, but the car beeping isn't. @cdwillie76 do you see the same behavior?

cdwillie76:
The beep yes. When comma isn’t connected, my Santa Fe beeps also when the keep hands on wheel alert comes up with at stock so I think that is part of the message. Not sure if there is a way to do the hands on wheel without a beep on Santa Fe. Prior to 0.8.1, did the lane change cause the hands on wheel?

JustStig:
Prior to 8.1 ALC did not trigger the hands on wheel

adeebshihadeh:
I understand that. There was a bug fix that shipped in 0.8.1 for the Santa Fe which is why you now see the hands on wheel dash alert. Ideally we want to use the car's dash alerts without the car playing sounds, but it seems the message we currently have for the Santa Fe plays a sound. We'd certainly merge a PR that uses a dash alert that doesn't play a sound if you can find one, but this works as intended.

PR #2490, opened 2020-11-06
2019 Hyundai Santa Fe fixes

cdwillie76:

PR #21481, opened 2021-07-05
HKG: 2019 Kia Niro PHEV

### LINKED ISSUE #21479: 2019 Kia Niro PHEV - No steering under 32 MPH
**Describe the bug**: When system is active (ACC turned on) if the driving speed (not the set speed) goes below 32 MPH, the steering control will disengage with out warning. When driving under 32 MPH then moving the set speed from 29 to 32 MHP, the system can violently (quickly turning right or left) as the system assumes control as the car transitions from 31 to 32 MPH. This can be quite alarming and renders the system unsafe under those operations. Rolling the set speed from above 32 MPH to below 32 results in a benign disengagement of the steering (i.e. just stops controlling the steering but doesn’t do so in a erratic behavior). This behavior makes the system not usable under 32 MPH and dangerous in freeway stop-n-go traffic. 

I’ve tried both Circuit-pro fork “hkg_085” and Donfyffe, also with v0.8.5, and both have the same 32 MPH issue described above.

**How to reproduce or log data**: As stated above. 

**Expected behavior**: I would like the system to control the steering wheel once ACC is set in the car and allow the car’s steering wheel to be controlled from 0 MPH, there by allowing the software to function in stop-n-go traffic. At the very least there should be a warning to NOT use the system in stop-n-go traffic especially if the car speeds bounces above and below 32 MPH. It would also be nice to have something on the C2 display showing when C2 is actively controlling the steering wheel. 

 - Device: comma two
 - Dongle ID: cd5179e1
 - Route: ?? Not using comma services so my.comma.ai/useradmin is blank
 - Panda FW: 46fda13fee69be84
 - Timestamp: Anytime the conditions are ACC engaged and speed is set below 32 MPH
 - Version: v0.8.5 from “Circuit-Pro” branch “hkg_085”
 - Car make/model: Kia Niro Plug-in Hybrid 2019

**Additional context**

jyoung8607:
**Checklist**
- [x] added to README
- [x] route with openpilot: 173219cf50acdd7b|2021-07-05--10-27-41

Although this was never a bug in upstream openpilot, this PR effectively resolves #21479.

**Important note:** This removes conflicting UDS fingerprint entries for OPTIMA_H, because they appear to be erroneous:
* A single OPTIMA_H FPv2 entry was added in #1682, and never touched again
* They are an exact 100% collision with this Kia Niro PHEV
* If this car doesn't match a UDS fingerprint, the fallback CAN match is to OPTIMA_H (the source of the above FW batch)
* Owner and VIN lookup verify this is a Kia Niro PHEV
* Engine is different enough (2.0L vs 1.6L) to make it improbable Optima shares this exact engine firmware
* Mass and wheelbase are different enough to make it improbable Optima shares this exact steering firmware


README.md
```diff
@@ -169,6 +169,7 @@ Community Maintained Cars and Features
 | Jeep      | Grand Cherokee 2016-18        | Adaptive Cruise   | Stock            | 0mph               | 9mph         |
 | Jeep      | Grand Cherokee 2019-20        | Adaptive Cruise   | Stock            | 0mph               | 39mph        |
 | Kia       | Forte 2018-2021               | SCC + LKAS        | Stock            | 0mph               | 0mph         |
+| Kia       | Niro PHEV 2019                | SCC + LKAS        | Stock            | 0mph               | 32mph        |
```
adeebshihadeh, line 172: move this under the EV

selfdrive/car/hyundai/interface.py
```diff
@@ -229,7 +231,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
 
     # these cars require a special panda safety mode due to missing counters and checksums in the messages
     if candidate in [CAR.HYUNDAI_GENESIS, CAR.IONIQ_EV_2020, CAR.IONIQ_EV_LTD, CAR.IONIQ_PHEV, CAR.IONIQ, CAR.KONA_EV, CAR.KIA_SORENTO,
-                     CAR.SONATA_LF, CAR.KIA_NIRO_EV, CAR.KIA_OPTIMA, CAR.VELOSTER, CAR.KIA_STINGER, CAR.KIA_SELTOS,
+                     CAR.SONATA_LF, CAR.KIA_NIRO_EV, CAR.KIA_NIRO_HEV, CAR.KIA_OPTIMA, CAR.VELOSTER, CAR.KIA_STINGER, CAR.KIA_SELTOS,
```
adeebshihadeh, line 234: did you check if this is actually required? the car model unit tests check this, so if you remove it and CI passes, it's not necessary

jyoung8607: No, I just aligned to existing Niro EV. I will check.

selfdrive/car/hyundai/values.py
```diff
@@ -569,6 +563,24 @@ class Buttons:
       b'\xf1\x00OSE LKAS AT EUR LHD 1.00 1.00 95740-K4100 W40',
     ],
   },
+  CAR.KIA_NIRO_HEV: {
+    (Ecu.engine, 0x7e0, None): [
+      b'\xf1\x816H6F4051\000\000\000\000\000\000\000\000',
+    ],
+    (Ecu.transmission, 0x7e1, None): [
+      b"\xf1\x816U3J2051\000\000\xf1\0006U3H0_C2\000\0006U3J2051\000\000PDE0G16NS2\xf4\'\\\x91",
+      b'\xf1\x816U3J2051\000\000\xf1\0006U3H0_C2\000\0006U3J2051\000\000PDE0G16NS2\000\000\000\000',
```
jyoung8607, line 572: We're getting both of these variants back intermittently, verified in his uploaded logs. I'm not sure what causes there to be trailing garbage vs trailing nulls, and I don't have the cycles to find out. Carrying both variants in the FW dict is the only thing I can really do in response, unless comma have some better idea what's happening here.

adeebshihadeh: known issue. we've been including both for now

jyoung8607:
Car is running and driving, and the test route is ready for upload to CI bucket if you would be so kind, but holding in draft till we see if we can work around/fix the dash beeping along with the instrument cluster "take control of steering" alert.

selfdrive/car/hyundai/interface.py
jyoung8607: Looks like we don't need the legacy safety thing.

jyoung8607:
As far as I can tell, the LFA message and associated LKA signal tweaks aren't a match for this 2019 vehicle. [2019 Kia Niro](https://www.kiamedia.com/us/en/models/niro/2019/features) had LKA only, verified no LFA message in Cabana except one sent by OP in an earlier version of this PR. The [2020 Kia Niro](https://www.kiamedia.com/us/en/models/niro/2020/features) gained LFA. With LKA only, it appears there isn't a visual-only alert even from the stock system, only visual alert with audible beep, and the beep is obnoxious for our purposes.

@adeebshihadeh if you agree with this analysis, it has implications for #2490, #19747, etc.

Other than that, the car seems G2G except I don't like what his car does when coasting to a stop. It oscillates between the yellow minSteerSpeed state and the red steering fault state, I'm guessing because of this: https://github.com/commaai/openpilot/blob/615ebb9f82b44864d4b09ba1430db5258357c6ac/selfdrive/car/hyundai/carcontroller.py#L53.

I don't see any reason why the HKG port should be applying nonzero steering torque below minSteerSpeed, for any car, but I'm hesitant to tamper with HKG carcontroller global behaviors without a consult. Any thoughts on this?

https://youtu.be/LatWw_0NhOM

jyoung8607:
I tested no torque under minSteerSpeed with the user, and he's now happy with the state of the port.

I didn't really set out to file a car port PR that modifies the behavior of other existing cars, but we seem to have gotten there. If you'd prefer I break this PR into two separate PRs, I can do that.

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -42,19 +42,15 @@ def __init__(self, dbc_name, CP, VM):
...
     self.steer_rate_limited = new_steer != apply_steer
 
-    # disable when temp fault is active
-    lkas_active = enabled and not CS.out.steerWarning
-
-    # fix for Genesis hard fault at low speed
-    if CS.out.vEgo < 16.7 and self.car_fingerprint == CAR.HYUNDAI_GENESIS:
-      lkas_active = False
+    # disable when temp fault is active, or below LKA minimum speed
+    lkas_active = enabled and not CS.out.steerWarning and CS.out.vEgo >= min_steer_speed
```
adeebshihadeh, line 53:
```suggestion
    lkas_active = enabled and not CS.out.steerWarning and CS.out.vEgo >= CS.CP.minSteerSpeed
```

adeebshihadeh:
I'd prefer to find a flag that's thrown when below the minSteerSpeed, but this is fine for now.

PR #21495, opened 2021-07-06
qt/spinner: reduce CPU usage from 17% to 12%

deanlee:


selfdrive/ui/qt/spinner.cc
```diff
@@ -115,7 +106,9 @@ void Spinner::update(int n) {
 
 int main(int argc, char *argv[]) {
   setQtSurfaceFormat();
-
+  if (Hardware::EON()) {
```
adeebshihadeh, line 109: let's make a function like initApp in util for this stuff. it should do all the common stuff like:
* turn on display
* set brightness
* set format
* set gl attributes

selfdrive/ui/qt/util.cc
```diff
@@ -67,6 +69,15 @@ void setQtSurfaceFormat() {
   QSurfaceFormat::setDefaultFormat(fmt);
 }
 
+void initApp(int brightness) {
+  Hardware::set_display_power(true);
+  Hardware::set_brightness(brightness);
```
adeebshihadeh, line 74:
```suggestion
void initApp() {
  Hardware::set_display_power(true);
  Hardware::set_brightness(65);
```

selfdrive/ui/qt/util.h
```diff
@@ -14,6 +14,7 @@ void clearLayout(QLayout* layout);
 void setQtSurfaceFormat();
 QString timeAgo(const QDateTime &date);
 void swagLogMessageHandler(QtMsgType type, const QMessageLogContext &context, const QString &msg);
+void initApp(int brightness = 65);
```
adeebshihadeh, line 17:
```suggestion
void initApp();
```

PR #21511, opened 2021-07-07
Fix active AP

sshane:
initConnections was also blocking, so move to first scan


adeebshihadeh:
Good to merge?

sshane:
Yep, tested and working

PR #21512, opened 2021-07-07
Toyota: handle models with permanent low speed lockout

sshane:
If `SET_ME_X01` isn't 1, low speed lockout should be guaranteed to be 2. Going to verify with a few routes from always faulting cars and never faulting cars

Cars experiencing `lowSpeedLockout`s:
- Toyota Camry Nightshade 2021 (US) - `SET_ME_X01` is `2` on startup (during stock route)
- Toyota / Camry LE 2021 (US) - `SET_ME_X01` is `2` on startup
- Toyota Corolla non-hybrid 2021 (Brazil made - Argentina market) - `SET_ME_X01` is `2` on startup
- Corolla Hybrid 2021 (Brazil made - Brazilian market) - `SET_ME_X01` is `2` on startup
- Toyota / Corolla Hybrid 2020 (Brazil made - Latinamerican market) - `SET_ME_X01` is `2` on startup
- Toyota Corolla Cross Hybrid (Thailand made and Asian market eg Thailand, Vietnam and Taiwan) - `SET_ME_X01` is `2` on startup

Cars not experiencing `lowSpeedLockout`s:
- 2021 Camry Hybrid TSS 2.5 - `SET_ME_X01` is `1` on startup
- TSS2 Corolla from test_routes.py - `SET_ME_X01` is `1` on startup

Not in this PR, but perhaps if you block 343 while the device is coming up and send your own with `SET_ME_X01` set to 1 you may be able to get stop and go acceleration

### DIFF
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -210,6 +210,6 @@ class CarState(CarStateBase):
 
     if CP.carFingerprint in TSS2_CAR:
       signals.append(("ACC_TYPE", "ACC_CONTROL", 1))
-      checks.append(("ACC_CONTROL", 33))
+      checks.append(("ACC_CONTROL", 0))
 
     return CANParser(DBC[CP.carFingerprint]["pt"], signals, checks, 2)
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -164,7 +164,7 @@ class CarState(CarStateBase):
     ]
 
     if CP.carFingerprint in TSS2_CAR:
-      signals.append(("SET_ME_X01", "ACC_CONTROL", 0))
+      signals.append(("SET_ME_X01", "ACC_CONTROL", 1))
       checks.append(("ACC_CONTROL", 33))
 
     if CP.carFingerprint == CAR.LEXUS_IS:
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -4,7 +4,7 @@ from opendbc.can.can_define import CANDefine
 from selfdrive.car.interfaces import CarStateBase
 from opendbc.can.parser import CANParser
 from selfdrive.config import Conversions as CV
-from selfdrive.car.toyota.values import CAR, DBC, STEER_THRESHOLD, NO_STOP_TIMER_CAR
+from selfdrive.car.toyota.values import CAR, DBC, STEER_THRESHOLD, NO_STOP_TIMER_CAR, TSS2_CAR
 from selfdrive.swaglog import cloudlog
 
 
@@ -76,11 +76,9 @@ class CarState(CarStateBase):
     if self.CP.carFingerprint == CAR.LEXUS_IS:
       ret.cruiseState.available = cp.vl["DSU_CRUISE"]["MAIN_ON"] != 0
       ret.cruiseState.speed = cp.vl["DSU_CRUISE"]["SET_SPEED"] * CV.KPH_TO_MS
-      self.low_speed_lockout = False
     else:
       ret.cruiseState.available = cp.vl["PCM_CRUISE_2"]["MAIN_ON"] != 0
       ret.cruiseState.speed = cp.vl["PCM_CRUISE_2"]["SET_SPEED"] * CV.KPH_TO_MS
-      self.low_speed_lockout = cp.vl["PCM_CRUISE_2"]["LOW_SPEED_LOCKOUT"] == 2
 
       val = cp_cam.vl["ACC_CONTROL"]["SET_ME_X01"]
       if val not in self.logged_setme_vals and val != 1:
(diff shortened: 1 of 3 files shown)


sshane:
Does the CANParser return the last seen signal value? ACC_CONTROL on bus 0 is only available for a split second on start up

adeebshihadeh:
Yes, you only see it until the bus is broken by setting the safety mode.

sshane:
Oh I think we can use bus 2 (cam) since it's always available there, and then we won't run into the rate issues (since it's not sending anything on bus 0 anymore)

sshane:
When checking 50 TSS2 routes, I found a US TSS2 Corolla (meaning normally doesn't lockout) that had a genuine low speed lockout with SET_ME_X01 set to 1, so with this PR it would correctly catch that case with the same behavior as previous (show cruise fault alert).

Checked 50 TSS2 routes (including some of the faulting ones in the beginning) and I haven't seen a route where SET_ME_X01 isn't 1 without a low speed lockout...except in rare cases where a usually permanently locked out Camry allows stop and go acceleration (in @Nyck310's `2702185606b65689|2021-07-08--06-06-35`).

selfdrive/car/toyota/carstate.py
```diff
@@ -76,18 +74,19 @@ def update(self, cp, cp_cam):
...
-      val = cp_cam.vl["ACC_CONTROL"]["SET_ME_X01"]
-      if val not in self.logged_setme_vals and val != 1:
-        self.logged_setme_vals.add(val)
-        cloudlog.event("setme_vals", vals=self.logged_setme_vals, car=self.CP.carFingerprint)
 
     self.pcm_acc_status = cp.vl["PCM_CRUISE"]["CRUISE_STATE"]
+
+    # some TSS2 cars have low speed lockout permanently set. ignore those cars, while still checking all others.
+    # when a car has a permanent low speed lockout, ACC_TYPE is 2
+    self.low_speed_lockout = False
```
adeebshihadeh, line 85: just set in the constructor

selfdrive/car/toyota/carstate.py
```diff
@@ -76,18 +76,19 @@ def update(self, cp, cp_cam):
...
-
-      val = cp_cam.vl["ACC_CONTROL"]["SET_ME_X01"]
-      if val not in self.logged_setme_vals and val != 1:
-        self.logged_setme_vals.add(val)
-        cloudlog.event("setme_vals", vals=self.logged_setme_vals, car=self.CP.carFingerprint)
 
     self.pcm_acc_status = cp.vl["PCM_CRUISE"]["CRUISE_STATE"]
+
+    # some TSS2 cars have low speed lockout permanently set. ignore those cars, while still checking all others.
+    # when a car has a permanent low speed lockout, ACC_TYPE is 2
```
adeebshihadeh, line 86:
```suggestion
    # some TSS2 cars have low speed lockout permanently set, so ignore on those cars
    # these cars are identified by an ACC_TYPE value of 2
```

selfdrive/car/toyota/carstate.py
```diff
@@ -76,18 +76,19 @@ def update(self, cp, cp_cam):
...
+    self.acc_type = cp_cam.vl["ACC_CONTROL"]["ACC_TYPE"]
+    if self.CP.carFingerprint != CAR.LEXUS_IS and (self.CP.carFingerprint not in TSS2_CAR or
+                             (self.CP.carFingerprint in TSS2_CAR and self.acc_type == 1)):
```
adeebshihadeh, line 89:
```suggestion
    if (self.CP.carFingerprint not in TSS2_CAR and self.CP.carFingerprint != CAR.LEXUS_IS) or \
       (self.CP.carFingerprint in TSS2_CAR and self.acc_type == 1):
```

selfdrive/car/toyota/carstate.py
```diff
@@ -192,7 +193,7 @@ def get_cam_can_parser(CP):
     signals = [
       ("FORCE", "PRE_COLLISION", 0),
       ("PRECOLLISION_ACTIVE", "PRE_COLLISION", 0),
-      ("SET_ME_X01", "ACC_CONTROL", 1)
+      ("ACC_TYPE", "ACC_CONTROL", 1),
```
adeebshihadeh, line 196:
```suggestion
      ("ACC_TYPE", "ACC_CONTROL", 0),
```

adeebshihadeh: also enable the checks

sshane:
@adeebshihadeh the cars affected by this seem to have a hard disengage at 12-15 mph (uncertain for the Camry, Corolla definitely is 12 mph), so we don't have to worry about forcing a disengage with `speedTooLow`. But should we show the `belowEngageSpeed` alert if they try to engage at a low speed? Right now minEnableSpeed is -1 for these cars, but we first need to know the speed, 12 or 15, or different speeds per locked out car.

tomascw:
Cars experiencing lowSpeedLockouts:

Toyota Rav4 hybrid 2019 (Japan made - Argentina market)

silvercurrent:
@adeebshihadeh ,@sshane, and @Nyck310, I'd also be interested in getting Stop and Go working on Toyotas with lowspeed lockout. Besides this fix https://github.com/commaai/openpilot/blob/master/selfdrive/car/toyota/carstate.py#L89, could we also revert to the behavior in ca. OP 0.8.2/0.8.3 after the car has reached 15mph and engaged OP for the first time on the ignition cycle? If I remember correctly, Stop and Go did work on my 2021 Camry LE without full-speed DRCC, if it was possible to get OP to engage without the CAN fault.

Nyck310:
> @adeebshihadeh ,@sshane, and @Nyck310, I'd also be interested in getting Stop and Go working on Toyotas with lowspeed lockout. Besides this fix https://github.com/commaai/openpilot/blob/master/selfdrive/car/toyota/carstate.py#L89, could we also revert to the behavior in ca. OP 0.8.2/0.8.3 after the car has reached 15mph and engaged OP for the first time on the ignition cycle? If I remember correctly, Stop and Go did work on my 2021 Camry LE without full-speed DRCC, if OP engaged without the CAN fault.
> 
> Edit: The Camry has a hard disengage at 15 mph.

I'm  unsure of the change that happened that made the restart workaround obsolete so 21 camry could have stop n go ( something about the timing of a message being sent in versions post  0.8.3 ) but if there were a way to...reverse just that, give it more time, and not modify the lowspeedlockout, I think that would solve the issue.
What that entails I have no idea, but I'm  up for being a test rat with a bounty for anyone who wants to take a crack at it.

silvercurrent:
Alternatively, would VOACC accomplish the same thing? I believe it's been around since at least 0.8.3, but has to be manually enabled.

PR #21518, opened 2021-07-08
Installer: progress bar

sshane:
Todo:
- [x] Fix `Two environments with different actions were specified for the same target: qt/setup/moc_installer.o`
- [x] Clean up imports
- [ ] Doesn't draw UI without hacks

Mean runs for release: `[7079.25, 143.75125, 562.3125 ]`


selfdrive/ui/qt/setup/installer.cc
```diff
@@ -23,19 +23,19 @@
 #define BRANCH "master"
 #endif
 
+#ifdef MASTER
```
adeebshihadeh, line 26: let's not worry about this right now. let's just make sure it works well on release2

adeebshihadeh:
No need for the spinner or submodule handling.

adeebshihadeh:
ported progress to #21588

PR #21522, opened 2021-07-08
camera_qcom: power down the actuator when is_road_cam is true

deanlee:


adeebshihadeh:
What's the goal of this? camera_qcom is stable and somewhat frozen, so we really don't want to make changes like this without purpose.

PR #21529, opened 2021-07-09
Mazda:  Latest FP2 FW roundup

Jafaral:


adeebshihadeh:
We'll need dongle IDs for these new FW versions

Jafaral:
@adeebshihadeh  see individual commit messages plz.

PR #21535, opened 2021-07-09
Tesla: more fingerprint for AP1

kennedyshead:
**Description** In rare occasions my AP1 was not recognised, this fixes the problem.

**Verification** tested in all different states I could imagine the car being in


selfdrive/car/tesla/values.py
```diff
@@ -21,6 +21,9 @@ class CAR:
...
     },
+    {
```
adeebshihadeh, line 24: this is just a subset of the original fingerprint, but 660 and 857 have different lengths

kennedyshead: Nope, I saw that and I have tried to figure out why that is for a week now. It might be a EU version?!

adeebshihadeh: are these values taken from another car besides your own?

kennedyshead: No these new ones are from mine, the ones before where borrowed from tinkla

kennedyshead: should I just replace?

adeebshihadeh: yeah, let's just replace with yours

PR #21552, opened 2021-07-11
Qt/spinner: reduce cpu usage from 12% to 9%

deanlee:
continue #21495.  

1. do not fill background, use `WA_OpaquePaintEvent` to paint all in `paintEvent` (cpu recuded from 12% to 9%)
2. ~set fps to 20, But the angle of rotation at each step is the same as 30fps.  (cpu recuded from 9% to 6%)~

### DIFF
--- a/selfdrive/ui/qt/spinner.cc
+++ b/selfdrive/ui/qt/spinner.cc
@@ -15,20 +15,24 @@
 #include "selfdrive/ui/qt/util.h"
 
 TrackWidget::TrackWidget(QWidget *parent) : QWidget(parent) {
+  setAttribute(Qt::WA_OpaquePaintEvent);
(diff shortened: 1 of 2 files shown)


selfdrive/ui/qt/spinner.cc
```diff
@@ -15,23 +15,27 @@
 #include "selfdrive/ui/qt/util.h"
 
 TrackWidget::TrackWidget(QWidget *parent) : QWidget(parent) {
+  setAttribute(Qt::WA_OpaquePaintEvent);
```
adeebshihadeh, line 18: I think 9% is enough :). let's revert the other changes

deanlee:
done

PR #21556, opened 2021-07-11
sidebar: indicator for GPS fix

adeebshihadeh:


geohot:
I like that it relies on liveLocationKalman too, abstracted away from the type of GPS.

PR #21560, opened 2021-07-12
Qt: significantly lower CPU usage while scrolling on PC

deanlee:
the contents of parent widgets are propagated by default. all parents and the entire scroll area will be re-paint while scrolling,that cause high CPU usage.


deanlee:
@adeebshihadeh :  It's works as expected on PC, only the newly exposed parts will be repaint, the maximum cpu usage is below 7% while scrolling.
but it is different on eon. although the parent will no longer receive the paint message after this PR, the entire scroll area will still be repainted. I suspect it is caused by some problem with the qt library on the eon.

PR #21570, opened 2021-07-12
Less toyota max accel

haraschax:


HaraldSchafer:
Some corollas tend to overshoot the desired acceleration, this would prevent that. Not sure if we want to do it this way though. This isn't really using the right abstraction layer and really the PID should not overshoot in the first place.

sshane:
Why not `gasMax` if this is to be merged? Should do the same thing

selfdrive/car/toyota/carcontroller.py
```diff
@@ -61,7 +63,10 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, hud_alert,
         pcm_accel_cmd = 0.06 - actuators.brake
 
     pcm_accel_cmd, self.accel_steady = accel_hysteresis(pcm_accel_cmd, self.accel_steady, enabled)
-    pcm_accel_cmd = clip(pcm_accel_cmd * CarControllerParams.ACCEL_SCALE, CarControllerParams.ACCEL_MIN, CarControllerParams.ACCEL_MAX)
+    planner_max_accel = get_max_accel(CS.out.vEgo)
```
sshane, line 66: Why not `gasMax`? Clipping after PID might introduce a lot of deadzone where integral can build up but with no change to accel command, causing it to be slow to come back down.

PR #21575, opened 2021-07-13
networking: update on show

sshane:
`LastScan` signal seems to only be unreliable if requested multiple times in a short time period, when requested right after adapter comes online it returns quick. The times it doesn't return quick enough, refreshing on show view will keep the UI feeling responsive enough.

### DIFF
--- a/selfdrive/ui/qt/offroad/networking.cc
+++ b/selfdrive/ui/qt/offroad/networking.cc
@@ -168,16 +168,22 @@ void AdvancedNetworking::toggleTethering(bool enabled) {
 
 WifiUI::WifiUI(QWidget *parent, WifiManager* wifi) : QWidget(parent), wifi(wifi) {
   main_layout = new QVBoxLayout(this);
+  scanningWidget();
+  main_layout->setSpacing(25);
+}
 
-  // Scan on startup
+void WifiUI::scanningWidget() {
   QLabel *scanning = new QLabel("Scanning for networks");
   scanning->setStyleSheet(R"(font-size: 65px;)");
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -101,6 +101,16 @@ void Networking::wrongPassword(const QString &ssid) {
   }
 }
 
+void Networking::showEvent(QShowEvent* event) {
+  // Wait to refresh to avoid lag when cycling through settings widgets
```
adeebshihadeh, line 105:
```suggestion
  // Wait to refresh to avoid queuing up too many scans if shown and hidden quickly
```

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -169,6 +179,12 @@ WifiUI::WifiUI(QWidget *parent, WifiManager* wifi) : QWidget(parent), wifi(wifi)
 
 void WifiUI::refresh() {
   clearLayout(main_layout);
+  if (wifi->seen_networks.size() == 0) {
+    QLabel *scanning = new QLabel("No networks");
```
adeebshihadeh, line 183:
```suggestion
    QLabel *scanning = new QLabel("No networks found. Scanning...");
```

selfdrive/ui/qt/offroad/networking.cc
sshane: Scans aren't really a problem, it's the blocking calls done when refreshing the networks, it causes significant lag if you click and click out of the Network pane

PR #21582, opened 2021-07-14
Add 2021 Camry to DRCC exception

sshane:


sshane:
No 2021 Camry L, but looks good aside from that

PR #21592, opened 2021-07-14
small keyboard fixes

sshane:
Keys had white border around them in setup from `setup.cc`'s `border: 7px solid white;`


adeebshihadeh:
This should be fixed in setup, not the keyboard.

PR #21594, opened 2021-07-14
Qt: draw border in OnroadWindow

deanlee:


pd0wm:
Can we make sure the border is non transparent? Currently it still relies on black bars around the camera view to render correctly. However, with the map open it looks wrong.

pd0wm:
Can we extend the camera view all the way to the edge? That way there are no black borders in split view.

2021-07-14-123051_1395x1083_scrot

deanlee:
Is this OK? I am traveling and there is no device to test now.

pd0wm:
Now the video is completely gone. The scissor is probably necessary.

pd0wm:
Also saw the video was now gone on master, so that must be a problem with my machine. So disregard.

pd0wm:
Can we also continue the lane lines behind the border? With the split the lane lines are cut off too soon.

deanlee:
how about now?

pd0wm:
Yes, looks good! Feel free to remove draft status.

@adeebshihadeh does this look good to you?

adeebshihadeh:
I don't think we want this as part of the onroad alerts. The border should be part of the layout (a margin in the base layout everything else is drawn on), otherwise you have to think about what's hidden by the border.

deanlee:
> I don't think we want this as part of the onroad alerts. The border should be part of the layout (a margin in the base layout everything else is drawn on), otherwise you have to think about what's hidden by the border.

moved to OnroadWindow,looks better now.

selfdrive/ui/qt/onroad.cc
```diff
@@ -7,42 +7,49 @@
...
+  nvg->updateState(s);
+  QColor bgColor = alerts->updateState(s);
+  if (bg != bgColor) {
+    bg = bgColor;
+    update();
+  }
+}
 
 void OnroadWindow::offroadTransition(bool offroad) {
+  alerts->offroadTransition(offroad);
```
adeebshihadeh, line 52: this should use signals/slots

selfdrive/ui/qt/onroad.cc
```diff
@@ -54,17 +61,24 @@ void OnroadWindow::offroadTransition(bool offroad) {
...
-      split->addWidget(m);
-
-      map = m;
+      map = new MapWindow(settings);
+      split->addWidget(map);
     }
-
+  }
+  if (map) {
+    ((MapWindow*)map)->offroadTransition(offroad);
```
adeebshihadeh, line 69: same as above

selfdrive/ui/qt/onroad.h
```diff
@@ -17,19 +17,19 @@ class OnroadAlerts : public QWidget {
   Q_OBJECT
 
 public:
-  OnroadAlerts(QWidget *parent = 0) {};
+  OnroadAlerts(QWidget *parent = 0) : QWidget(parent) {};
+  QColor updateState(const UIState &s);
```
adeebshihadeh, line 21: updateState should be a standard slot all UI widgets have, don't like that this one returns a QColor. we can add a getter for the color or just abstract out the logic to get the background color from the alert status

deanlee: Yes,that's better

deanlee:
It feels a bit strange to use getter

deanlee:
@adeebshihadeh  what do you think?

adeebshihadeh:
The border now frequently gets out of sync with the alerts.

deanlee:
sorry! my bad. forgot to call update after this change

adeebshihadeh:
Looks good now. Can we keep the background until we get a frame though? I think the black frame on startup looks worse now.

PR #21595, opened 2021-07-14
make sure tidx+1 doesn't exceed TRAJECTORY_SIZE-1

pd0wm:
Continuation of #21439 to run in jenkins


deanlee:
:heart:

deanlee:
@HaraldSchafer does get_best_data has the same problem?

https://github.com/commaai/openpilot/blob/48020e686eacba1dea76d14ff9e807594f7b68e4/selfdrive/modeld/models/driving.cc#L124-L131

PR #21600, opened 2021-07-14
driving.cc/get_best_data: for discussion only

deanlee:
@HaraldSchafer  :does get_best_data has the same issue with  https://github.com/commaai/openpilot/pull/21595 ?


HaraldSchafer:
No it is correct as is, hopefully this is more readable

PR #21334, opened 2021-06-18
Refactor lateral lag compensation

haraschax:

PR #21611, opened 2021-07-15
Add angleError to pid_log

ghost:
After https://github.com/commaai/openpilot/pull/21334 it became harder to check how well the tune is working in PJ because it's not possible to plot desired steering angle any more, however cereal already has angle error defined for LateralPIDState but not used. 
As for the sign of the error should match the p term


selfdrive/controls/lib/latcontrol_pid.py
```diff
@@ -42,6 +42,7 @@ def update(self, active, CS, CP, VM, params, desired_curvature, desired_curvatur
       output_steer = self.pid.update(angle_steers_des, CS.steeringAngleDeg, check_saturation=check_saturation, override=CS.steeringPressed,
                                      feedforward=steer_feedforward, speed=CS.vEgo, deadzone=deadzone)
       pid_log.active = True
+      pid_log.angleError = angle_steers_des - CS.steeringAngleDeg
```
pd0wm, line 45: Think we can always set the value, even when the controller is not active. Just move this line to above the if statement.

grekiki:
Think someone with azure token might need to update the process replay

Issue #21317, opened 2021-06-17
athenad: send websocket data using multiple frames


gregjhogan:
The websocket protocol has this concept of chunking data into frames so that you can easily send/receive large amounts of data.  You can send huge amounts of data in a single frame, but the advantage of chunking data into smaller frames is that you can set your timeouts much lower.  Since the image snapshot prime feature and logs sent from the device through the websocket can be large relative to the speed of a mobile data connection (2G can be as low as 5 KB/s) and we currently always send data in a single frame, we have to use a high timeout value.  The current websocket library we use doesn't have anything built-in to chunk sending data into frames for you, but it is probably easy to put this logic in athenad and switch from calling [send()](https://websocket-client.readthedocs.io/en/latest/core.html?highlight=frame#websocket._core.WebSocket.send) to [send_frame()](https://websocket-client.readthedocs.io/en/latest/core.html?highlight=frame#websocket._core.WebSocket.send_frame) in a loop.

related to https://github.com/commaai/openpilot/pull/21316

PR #21615, opened 2021-07-15
Chunking websocket sends into frames

shopped:
PR for https://github.com/commaai/openpilot/issues/21317

Splitting web socket data into 5KB frames, this might be too small
Unsure of how to test this, but the logic works on a local web socket server
Open to any feedback, this is my first attempt at contribution


gregjhogan:
thanks for the PR, but since this was a lot more complicated than necessary I created https://github.com/commaai/openpilot/pull/21918

PR #21632, opened 2021-07-17
ui: multitouch

sshane:
Start on multi touch support, overriding Qt's default mouse event synthesizing (setting `AA_SynthesizeMouseForUnhandledTouchEvents` to false) which only uses the first touch point.

Todo:
- [ ] Fix losing scrolling in new widget if you leave old widget while scrolling


sshane:
@deanlee would you want to take a crack at doing this properly? So far I haven't found a way to set a global attribute that's applied to new QWidgets without recursively going through and setting each one to accept touch events.

deanlee:
how about overriding QApplication::notify ?(pseudocode, not tested yet)

```
void Application::notify(QObject *receiver, QEvent *e) {
  if (e->type() == QEvent::TouchBegin) {
    QWidget *widget = static_cast<QWidget *>(receiver);
    bool acceptTouchEvents = widget->testAttribute(Qt::WA_AcceptTouchEvents);
    if (!acceptTouchEvents) {
      widget->setAttribute(Qt::WA_AcceptTouchEvents);
    }
  }
  return QAplication(receiver, e);
}
```

deanlee:
@sshane Is that okay? #21757

Issue #21557, opened 2021-07-11
GM: steering faults


Verylukyguy:
While driving, a "Steering Temporarily Unavailable" error will display and the Comma Two will no longer engage.
It also show a "Cruise Fault" error.

Comma Two
Dongle ID:  e38a02cae1ef1e0c
Route:  e38a02cae1ef1e0c|2021-07-11--18-54-39 at 19:03:43
Route:  e38a02cae1ef1e0c|2021-07-11--19-05-29 at 19:12:59
Timestamp:  The error came at the end of the first route at 19:03:43 and pulled into a parking lot and reset the vehicle and there was another error at 19:12:59 at the end of the second route. 
They are from one trip, where I had to reset the vehicle and Comma after the first error.
Version:  Comma.ai Master .0.8.6
GMC Acadia 2018

If you pull over and turn off the vehicle and allow the vehicle and Comma device to reset, approximately three minutes, you can restart the vehicle and reengage OpenPilot.

My understanding is that this is a LKA timing issue and has been on-going on multiple GM vehicles. 

----------------------------------------------------------------------------------------------------------------------------------------------
This is information that I received from someone else about this issue.

The issue only occurs when a lane keep assist message is sent less than 20 milliseconds after the one before it
My theory:  The code currently has a buffer that contains only the last Lane Keep Assist message it does not have a queue. 
 Another part of the software pulls this buffer close to exactly every 20 milliseconds.  Should two messages arrive within a 20 millisecond period, the earlier message is probably replaced by the newer one before the steering column had a chance to read it
So there's kind of two ways to fix this. Tighten up the timing - which is what the code currently does by moving the message generation onto the panda and using a timer on the panda.
The other option would be to simply skip three messages until you were back at the correct counter value.
Three messages is only 60 milliseconds
Nobody's going to notice a darn thing the car or you.

adeebshihadeh:
Please add more relevant information to the issue. How often does this happen? Is there a reparable procedure to reproduce it? Does it affect all GMs?

Verylukyguy:
I was not watching the CAN data in real time, but, in my example route, I was accelerating from a stop when the road shifts slightly to the right.  Many people with GM vehicle's (i.e. Volts and Bolts) have experienced the same issue.  There is a partial community fix on JShuler's fork, but it involves a timer on the Panda and a Buffer to regulate the LKA message timing.

jyoung8607:
[...]

PR #21641, opened 2021-07-18
GM: Correctly handle EPS warning vs error states

jyoung8607:
**Description**

The GM port never grew support for the separate steerError vs steerWarning states; they're newer than the GM port. As a result, it doesn't separately handle recoverable/temporary vs nonrecoverable/restart-required errors, and doesn't tell the driver when there's a permanent restart-required fault.

**Verification**

EPS LKA state 2 vs 3 was already documented in GM's carstate, and in the process of troubleshooting with an affected user, it was confirmed that 2 is recoverable and 3 requires a restart.

It's possible that users experiencing #21557 (which involves EPS LKA state 3) might think the problem has gotten worse after this fix. The underlying problem hasn't gotten worse, we're just doing a better job of reporting it, including while disengaged.

**Route**

Route: `e38a02cae1ef1e0c|2021-07-16--00-36-57`

### DIFF
--- a/selfdrive/car/gm/carcontroller.py
+++ b/selfdrive/car/gm/carcontroller.py
@@ -33,7 +33,7 @@ class CarController():
 
     # STEER
     if (frame % P.STEER_STEP) == 0:
-      lkas_enabled = enabled and not CS.out.steerWarning and CS.out.vEgo > P.MIN_STEER_SPEED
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Thanks for the bug fix!

PR #21642, opened 2021-07-18
Fingerprint: Correct value for fingerprint Leaf_IC

Mr-SourBalls:
**Description** Fixed the fingerprint for the nissan LEAF_IC model
**Verification** Confirmed working by some discord members with the same Model


adeebshihadeh:
Why is this titled correct and why are values removed? Is there only one user or is this for supporting a new user?

robin-reckmann:
This breaks at least my Leaf IC. How about adding a new fingerprint instead of modifying the current one?

Mr-SourBalls:
That might be a better idea. :)

PR #21643, opened 2021-07-18
add check for submodules in release builds

adeebshihadeh:


release/build_release2.sh
```diff
@@ -55,6 +55,13 @@ scons -j3
 python selfdrive/manager/test/test_manager.py
 selfdrive/car/tests/test_car_interfaces.py
 
+# Ensure no submodules in release
+if test "$(git submodule--helper list | wc -l)" -gt "0"; then
+  echo "submodules found:"
+  git submodule--helper list
```
adeebshihadeh, line 61:
```suggestion
  git submodule--helper list
  exit 1
```

Issue #21637, opened 2021-07-17
UI: fix http request caching


adeebshihadeh:

adeebshihadeh:
@deanlee want to look into this one? would be nice to also write tests

deanlee:
ok

adeebshihadeh:
this is for things like the API requests for drive stats that we cache to a param

deanlee:
I see... but I think it's still better to use QNetworkConfigurationManager to test internet connection in qt/setup. it's more "realtime" and if we clear nam'scache in HttpRequest, the time of clear/re-init may be more than 300ms on EON

deanlee:
> this is for things like the API requests for drive stats that we cache to a param

could it be expressed in a more specific way?

adeebshihadeh:
We cache the last successful request in RequestRepeater so that things like the drive stats are populated even without internet. https://github.com/commaai/openpilot/blob/master/selfdrive/ui/qt/request_repeater.cc#L15

deanlee:
I still don't quite understand how to fix the request caching. can it  be solved if we don't emit receivedResponse from cache when there is no network?

adeebshihadeh:
Fixed the caching with #21693, but the prime widget doesn't seem to take advantage of the caching properly.

PR #21664, opened 2021-07-20
qt/setup: fix connection cache problem

deanlee:
related issue #21637
use QNetworkConfigurationManager to test internet connection.


adeebshihadeh:
QNetworkConfigurationManager::isOnline() doesn't necessarily return whether a device has an internet connection. Also from the docs: `This class is obsolete.`.

PR #21668, opened 2021-07-20
move DriverStatus params to class

ZwX1616:


selfdrive/monitoring/driver_monitor.py
```diff
@@ -99,17 +102,22 @@ def __init__(self):
     self.cfactor = 1.
 
 class DriverStatus():
-  def __init__(self, rhd=False):
+  # pylint: disable=no-member
+  def __init__(self, rhd=False, **kwargs):
+    # init settings
+    self.__dict__.update(DRIVER_MONITOR_SETTINGS)
+    self.__dict__.update((k, v) for k, v in kwargs.items() if k in DRIVER_MONITOR_SETTINGS.keys())
```
adeebshihadeh, line 109: seems like this is for tests? leave a comment so it's clear what this is for, because nothing in openpilot seems to use it

selfdrive/monitoring/driver_monitor.py
```diff
@@ -99,17 +102,22 @@ def __init__(self):
     self.cfactor = 1.
 
 class DriverStatus():
-  def __init__(self, rhd=False):
+  # pylint: disable=no-member
+  def __init__(self, rhd=False, **kwargs):
+    # init settings
+    self.__dict__.update(DRIVER_MONITOR_SETTINGS)
```
adeebshihadeh, line 108: this is pretty dirty. some cleaner options:
* move these constants into the DriverStatus class
* make the dict a class and acces with `MonitoringSettings.` instead of `self.`
* make the dict a class and DriverStatus a subclass

selfdrive/monitoring/dmonitoringd.py
```diff
@@ -50,7 +50,9 @@ def dmonitoringd_thread(sm=None, pm=None):
     driver_status.get_pose(sm['driverState'], sm['liveCalibration'].rpyCalib, sm['carState'].vEgo, sm['controlsState'].enabled)
 
     # Block engaging after max number of distrations
-    if driver_status.terminal_alert_cnt >= MAX_TERMINAL_ALERTS or driver_status.terminal_time >= MAX_TERMINAL_DURATION:
+    # pylint: disable=no-member
```
adeebshihadeh, line 53: this is only needed because of the hacky dict update

selfdrive/monitoring/driver_monitor.py
adeebshihadeh: also, if this is for tests, think it makes more sense to pass in the class of constants as an argument instead of this update thing

selfdrive/monitoring/test_monitoring.py
```diff
@@ -5,19 +5,15 @@
...
 
 EventName = car.CarEvent.EventName
 
-_TEST_TIMESPAN = 120  # seconds
-_DISTRACTED_SECONDS_TO_ORANGE = _DISTRACTED_TIME - _DISTRACTED_PROMPT_TIME_TILL_TERMINAL + 1
-_DISTRACTED_SECONDS_TO_RED = _DISTRACTED_TIME + 1
-_INVISIBLE_SECONDS_TO_ORANGE = _AWARENESS_TIME - _AWARENESS_PROMPT_TIME_TILL_TERMINAL + 1
-_INVISIBLE_SECONDS_TO_RED = _AWARENESS_TIME + 1
+TEST_TIMESPAN = 120  # seconds
+DISTRACTED_SECONDS_TO_ORANGE = dm_settings["_DISTRACTED_TIME"] - dm_settings["_DISTRACTED_PROMPT_TIME_TILL_TERMINAL"] + 1
```
adeebshihadeh, line 13: these super long names make it hard to read the tests. maybe something like `WARN_DT`?

selfdrive/monitoring/driver_monitor.py
ZwX1616: kk let's use the setting class. Will fix after model work

selfdrive/monitoring/test_monitoring.py
ZwX1616: will do a names cleanup in another pr

PR #21671, opened 2021-07-21
VW MQB: Auto network location detection

jyoung8607:
**Description**

Autodetect whether openpilot is wired in at the Lane Assist camera or the CAN gateway, based on whether we see powertrain messages on bus 1. We're on Extended CAN either way, but when the relay opens to split the bus, we need to know whether the ACC and BSM radars end up on the bus 0 or bus 2 side.

In a follow-up PR, existence of the Lane Assist camera will be made optional.

**Verification**

Works on my 2018 Volkswagen Golf R with a prototype J533 gateway harness.

**Route**

`7aae592fc08e895f|2021-07-21--11-16-53` (still uploading at this moment)

### DIFF
--- a/selfdrive/car/volkswagen/interface.py
+++ b/selfdrive/car/volkswagen/interface.py
@@ -1,5 +1,5 @@
 from cereal import car
-from selfdrive.car.volkswagen.values import CAR, BUTTON_STATES, TransmissionType, GearShifter
+from selfdrive.car.volkswagen.values import CAR, BUTTON_STATES, CANBUS, NetworkLocation, TransmissionType, GearShifter
 from selfdrive.car import STD_CARGO_KG, scale_rot_inertia, scale_tire_stiffness, gen_empty_fingerprint
 from selfdrive.car.interfaces import CarInterfaceBase
 
@@ -12,6 +12,13 @@ class CarInterface(CarInterfaceBase):
 
     self.displayMetricUnitsPrev = None
     self.buttonStatesPrev = BUTTON_STATES.copy()
+
+    if CP.networkLocation == NetworkLocation.fwdCamera:
+      CANBUS.ext = CANBUS.pt
+      self.cp_ext = self.cp
+    else:
+      CANBUS.ext = CANBUS.cam
+      self.cp_ext = self.cp_cam
 
   @staticmethod
   def compute_gb(accel, speed):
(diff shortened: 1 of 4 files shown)


selfdrive/car/volkswagen/interface.py
```diff
@@ -36,6 +43,11 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None):
       else:  # No trans message at all, must be a true stick-shift manual
         ret.transmissionType = TransmissionType.manual
 
+      if 0x86 in fingerprint[1]:  # LWI_01 seen on bus 1, we're wired to the CAN gateway
```
adeebshihadeh, line 46: is this reliable with the current minimum fingerprinting time?

selfdrive/car/volkswagen/interface.py
```diff
@@ -13,6 +13,13 @@ def __init__(self, CP, CarController, CarState):
     self.displayMetricUnitsPrev = None
     self.buttonStatesPrev = BUTTON_STATES.copy()
 
+    if CP.networkLocation == NetworkLocation.fwdCamera:
+      CANBUS.ext = CANBUS.pt
```
adeebshihadeh, line 17: not a fan of this. think this would be cleaner
```suggestion
      self.ext_bus = CANBUS.pt
```

selfdrive/car/volkswagen/interface.py
adeebshihadeh: let's hardcode the network location to the camera until the follow up PR

jyoung8607: > is this reliable with the current minimum fingerprinting time?

Yes, it's a 100Hz message that seems to be available immediately upon openpilot startup in CAN logs. This is one of the major reasons we had powertrain added to bus 1 on gateway harnesses. Back in the day, I first tried to sense bus position by watching for object fusion messages on the *camera* harness, and those didn't show up quite fast enough, but this was way before CAN message collection window ran to 100ms for UDS fingerprinted cars.

jyoung8607: > let's hardcode the network location to the camera until the follow up PR

I don't think that's the right thing to do:

* This PR, as it stands alone, gateway integrated but with a factory LKAS camera, is a perfectly valid config to run and drive with. It's how I typically run my Golf R and Tiguan, and it's how many other folks are set up as well. It would be technically valid to accept this PR and reject the camera-optional PR.

* For the follow-up PR to make the camera optional, I actually do have some things to prove around fingerprint time. I think it'll be okay, and if it isn't I know how to solve it easily, but I intend to script up something to look at logs from several other MQB user devices before I move forward. Feel free to make me show my work on that one.

adeebshihadeh: fair enough, should be good to merge after addressing the other comment

selfdrive/car/volkswagen/interface.py
jyoung8607: Done.

adeebshihadeh:
Next PR, let's get a process replay route too.

PR #21676, opened 2021-07-22
networking: reuse drawn layouts

sshane:
Iterates through `seenNetworks` and checks if there's already an `hlayout` at that index to update with the new network's info. If not, add it. If networks are removed from `seenNetworks`, the loop after the first removes extra layouts. This is all so that we don't lose focus while scrolling, since every widget is removed and re-added.

- [x] Horizontal lines
- [x] Layout design is the same
- [x] Can scroll while updating
- [x] Fix Scanning widget
- [ ] ~fix stuttering when toggling off tethering~ (we're refreshing too many times in an instant, will be fixed in another PR

### DIFF
--- a/selfdrive/ui/qt/offroad/networking.cc
+++ b/selfdrive/ui/qt/offroad/networking.cc
@@ -279,6 +279,7 @@ QHBoxLayout* WifiUI::buildNetworkWidget(const Network &network, bool isTethering
 
   // Status icon
   QLabel *statusIcon = new QLabel();
+  statusIcon->setFixedWidth(lock.width());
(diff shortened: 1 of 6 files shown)


sshane:
Found a much simpler solution for the losing focus problem! Will open a new PR tomorrow

sshane:
When we're scrolling, none of the widgets have focus actually, obj: `QWidgetWindow(0x221b440, name = "MainWindowClassWindow")` does. Is this a bug in Qt?

adeebshihadeh:
Can you rebase? Seems this was based on another PR that got merged in.

PR #21683, opened 2021-07-22
HomeWindow: cleanup mousePressEvent

deanlee:
sidebar has its own mousePressEvent handler, this event will not be propagated to HomeWindow.so when  HomeWindow::mousePressEvent is called, e->x() must be in onroad.


adeebshihadeh:
any update on this?

PR #21686, opened 2021-07-22
elantra 2021: steering ratio correction

tecandrew:
**Description**: small adjustment to use actual steering ratio measurement


tecandrew:
specs found on Hyundai [press release site](https://www.hyundainews.com/en-us/models/hyundai-elantra-2021-elantra/specifications)
image

grekiki:
I'd be a bit worried because this is a suspiciosly low steering ratio. I suspect it might be a variable ratio steering rack because of the low ratio and because they say it's an "overall ratio", however for openpilot the ratio around low steering angles would be much more important.
Looked online for a bit but couldn't confirm whether or not the Elantra has a variable ratio steering, but would probably be a good idea to exclude that option.

PR #21687, opened 2021-07-22
FrameIterator that uses 1+ GB less RAM

gregjhogan:


gregjhogan:
previous code also left defunct ffmpeg processes behind

PR #21694, opened 2021-07-23
Add Fingerprint 2.0 for 2020 Toyota Corolla Cross (Taiwanese Version)

protonchang:
* Add fingerprint for engine, eps, esp, fwdCamera.
* This car does not support Stop & Go, also no Full-Speed DRCC
* No ALC and ACC below 18mph (30kph)

* Dongle ID: 94545a9472a6c942


selfdrive/car/toyota/values.py
```diff
@@ -649,12 +650,14 @@ class CAR:
       b'\x018965B12530\x00\x00\x00\x00\x00\x00',
       b'\x018965B1255000\x00\x00\x00\x00',
       b'8965B12361\x00\x00\x00\x00\x00\x00',
+      b'8965B16011\x00\x00\x00\x00\x00\x00'
```
VirtuallyChris, line 653: need a , after this line

protonchang:
Also, I'm not sure if I should update README.md or not, as Corolla Cross in the US market has Full-Speed DRCC and SnG feature, but it's not public available yet.
I suspect it will have a  different fingerprint compare to the Taiwanese version.

pd0wm:
Does openpilot stop steering under 18mph?

protonchang:
Yes, complete disengadge below 18mph

PR #21696, opened 2021-07-23
networking: cleanup style sheets

sshane:
Should be merged before: #21676

### DIFF
--- a/selfdrive/ui/qt/offroad/networking.cc
+++ b/selfdrive/ui/qt/offroad/networking.cc
@@ -217,6 +217,9 @@ WifiUI::WifiUI(QWidget *parent, WifiManager* wifi) : QWidget(parent), wifi(wifi)
       padding-top: 50px;
       padding-bottom: 50px;
     }
+    #ssidLabel[currentNetwork=true] {
+      font-weight: 500;
+    }
   )");
 }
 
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -221,34 +239,19 @@ void WifiUI::refresh() {
...
-      padding-top: 50px;
-      padding-bottom: 50px;
-    )").arg(weight));
-    QObject::connect(ssid_label, &QPushButton::clicked, this, [=]() { emit connectToNetwork(network); });
-    hlayout->addWidget(ssid_label, network.connected == ConnectedType::CONNECTING ? 0 : 1);
+    QPushButton *ssidLabel = new QPushButton(network.ssid);
+    ssidLabel->setObjectName("ssidLabel");
+    ssidLabel->setEnabled(network.connected == ConnectedType::DISCONNECTED &&
+                          network.security_type != SecurityType::UNSUPPORTED);
+    if (network.connected != ConnectedType::DISCONNECTED) {
```
adeebshihadeh, line 246: you can use a property for this

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -217,6 +217,9 @@ WifiUI::WifiUI(QWidget *parent, WifiManager* wifi) : QWidget(parent), wifi(wifi)
       padding-top: 50px;
       padding-bottom: 50px;
     }
+    #ssidLabel[currentNetwork=true] {
```
adeebshihadeh, line 220: call the property connected

adeebshihadeh: or disconnected and check if false

PR #21709, opened 2021-07-25
VW MQB: Volkswagen T-Cross Mk1

jyoung8607:
**Volkswagen T-Cross Mk1**

Covers all variants of the first generation Volkswagen T-Cross subcompact SUV (2019-present), including SWB and LWB variants.

- [x] added to README
- [x] added to release notes
- [x] test route added to [test_routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_models.py)
- [x] route with openpilot: 2744c89a8dda9a51|2021-07-24--21-28-06

Thanks to community T-Cross tester Saber422 (2021).


jyoung8607:
Note: Saber422 drove and tested this as a rental and it was force-fingerprinted as a Kodiaq on the fork he drove with, not sure if that'll create a problem for route replay testing.

PR #21429, opened 2021-06-28
Unify Rav4 tuning - lat is very unresponsive on non-FW version Rav4s

zorrobyte:
The stock tune for non-FW Ravs is much too lazy and OP is borderline unusable at lower speeds. OP will drift in the lane and sometimes will exit the lane as steering sensitivity is much too low. I've been running stock `master` on my device and better stock values need to be used.

~I don't think it's so much that differing FW versions are the culprit. Toyota wouldn't revise EPS firmware so darastically and the part numbers for the EPS motor are the same. It seems that more like the majority of Rav 4 users are on that FW version. I have a 2019 without that FW version and the stock tuning values result in sloppy lat control.~ I was proven wrong with research, TSS2 Rav4 has two rather different steering racks.

Issue #21707, opened 2021-07-24
0.8.6 update on RAV4 2020 hybrid 


amsalp:
Aggressive steering and oversteer over speeds and slow recovery.

The steering feels like it lets go and recovers halfway around a corner. Like a hold then release type scenario. 

**Device/Version information:**
 - Device: [comma two]
 - Dongle ID: [bde2c8ra474e7a8a]
 - Route: [2021-07-23--08-35-40]
 - Timestamp: [from start to finish]
 - Version: [release 2 (0.8.6)]
 - Car make/model: [Toyota Rav4 2020 hybrid ]
0.8.5 was working really well

HaraldSchafer:
I'll look into this tomorrow!

amsalp:
Ok no problem I've added more drives after this and tried laneless. I seems a little better. But I does cut corners and recovers very aggressively. Some parts it makes me a little seasick when it does it's overcorrection. Wobbles the car. I all so  hear the break relay clicking when it comes to a stop behind another car.

yufeng66:
My eps firm version is 8965B42171. I think neither of the two PID tuning works very well for me with the laneless model at this point.  The more relaxed one does slow and subtle ping pong on highway.  It is tolerable and car never go over the lane line. The more aggressive one got more noticeable ping pong at local speed on big curves. It is less tolerable. The steering wheel would turn too fast and make the experience less chill and almost unsafe. I think a new tuning might be needed. I try to use INDI a bit and liked result so far. The following is what I did. basically the Prius tuning with higher gain. 

      ret.lateralTuning.init('indi')
      ret.lateralTuning.indi.innerLoopGainBP = [0.]
      ret.lateralTuning.indi.innerLoopGainV = [6.0]
      ret.lateralTuning.indi.outerLoopGainBP = [0.]
      ret.lateralTuning.indi.outerLoopGainV = [4.0]
      ret.lateralTuning.indi.timeConstantBP = [0.]
      ret.lateralTuning.indi.timeConstantV = [1.0]
      ret.lateralTuning.indi.actuatorEffectivenessBP = [0.]
      ret.lateralTuning.indi.actuatorEffectivenessV = [1.0]
      ret.steerActuatorDelay = 0.2

HaraldSchafer:
[...]

PR #21715, opened 2021-07-25
Revert lateral tuning change for some TSS2 Rav4

adeebshihadeh:
A rav4 user with one of the unconfirmed EPS fw versions from #21429 reported a regression in lateral control.

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -180,7 +180,7 @@ class CarInterface(CarInterfaceBase):
       # 2019+ Rav4 TSS2 uses two different steering racks and specific tuning seems to be necessary.
       # See https://github.com/commaai/openpilot/pull/21429#issuecomment-873652891
       for fw in car_fw:
-        if fw.ecu == "eps" and fw.fwVersion.startswith(b'\x02'):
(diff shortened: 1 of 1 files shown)


yufeng66:
can we add b'8965B42171\x00\x00\x00\x00\x00\x00' to the list as well? my car have similar issue with that firmware

adeebshihadeh:
@yufeng66 that's the same fw as the author (@zorrobyte) of the PR. can you open an issue?

zorrobyte:
@adeebshihadeh could you share where such regression was reported?

I'm a little concerned as it took quite a bit of effort for that PR to be merged and that it can be undone based on one, unverified complaint. Tuning is a highly subjective matter and the values I proposed were considered "stock" OP values for mergability. Sure, better values can be found, but 0.15/0.05 is unusable for my vehicle.

adeebshihadeh:
The regression was reported in #21707. The tuning for your EPS's fw version wasn't reverted, only this one which was never confirmed.

zorrobyte:
> The regression was reported in #21707. The tuning for your EPS's fw version wasn't reverted, only this one which was never confirmed.

Gotcha! Thank you for the clarification. Rav is a rare breed indeed.

PR #21716, opened 2021-07-25
UI: decrease api request frequency

### LINKED ISSUE #21692: Reduce update frequency of rate limited API requests

adeebshihadeh:
@gregjhogan does this get us outside the rate limit? It's not great that the prime widget won't be as responsive, but maybe we should athena write that.


pd0wm:
We do have the billing server call an athena endpoint now, so we could use that to trigger a widget update somehow.

PR #21722, opened 2021-07-26
Tici sound improvements

robbederks:


selfdrive/ui/soundd.cc
```diff
@@ -14,18 +14,25 @@
 // TODO: detect when we can't play sounds
 // TODO: detect when we can't display the UI
 
+// TODO: merge again and add EQ in the amp config
+#ifdef QCOM2
```
pd0wm, line 18: Use `Hardware::TICI()` whenever possible.

PR #21731, opened 2021-07-27
implement display power

sshane:
no permissions. Can also set display brightness, and default to 65 when `on` but it may flicker if the brightness filter isn't 65 at that time

### DIFF
--- a/selfdrive/hardware/tici/hardware.h
+++ b/selfdrive/hardware/tici/hardware.h
@@ -28,7 +28,7 @@ public:
   static void set_display_power(bool on) {
     std::ofstream bl_power_control("/sys/class/backlight/panel0-backlight/bl_power");
     if (bl_power_control.is_open()) {
-      bl_power_control << (on ? "0" : "1") << "\n";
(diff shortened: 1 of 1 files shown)


selfdrive/hardware/tici/hardware.h
```diff
@@ -25,7 +25,13 @@ class HardwareTici : public HardwareNone {
       brightness_control.close();
     }
   };
-  static void set_display_power(bool on) {};
+  static void set_display_power(bool on) {
+    std::ofstream bl_power_control("/sys/class/backlight/panel0-backlight/bl_power");
+    if (bl_power_control.is_open()) {
+      bl_power_control << (on ? "0" : "1") << "\n";
```
adeebshihadeh, line 31: FB_BLANK_POWERDOWN = 4
```suggestion
      bl_power_control << (on ? "0" : "4") << "\n";
```

sshane:
This doesn't quite work yet, there's some permission errors accessing the file. only a command like `echo 0 | sudo tee /sys/class/backlight/panel0-backlight/bl_power` works

PR #21739, opened 2021-07-27
ui: elided button

sshane:
Converted the `ElidedLabel` from deanlee's #21277

Fixes ssid being cut off if too long (and connecting)

Code for Qt's ElidedLabel: https://doc.qt.io/qt-5/qtwidgets-widgets-elidedlabel-example.html

QObject classes don't support templates


deanlee:
how about inherit from ElidedLabel and make it clickable?  #21773

PR #21745, opened 2021-07-27
athenad: retry failed and timed out uploads

pd0wm:


selfdrive/athena/athenad.py
```diff
@@ -39,14 +39,17 @@
 LOG_ATTR_VALUE_MAX_UNIX_TIME = int.to_bytes(2147483647, 4, sys.byteorder)
 RECONNECT_TIMEOUT_S = 70
 
+RETRY_DELAY = 1.0  # seconds
+MAX_RETRY_COUNT = 300  # Try for at most 5 minutes if upload fails immediately
```
gregjhogan, line 43: I would raise the delay to 10 seconds and lower the retries to 30 because I think if it isn't working it is unlikely to improve in 1 second time, and waiting up to 10 seconds to get going once the connection is good seems fine

PR #21750, opened 2021-07-27
roaming toggle

sshane:
stores the `lteConnectionPath` on connections init instead of adding it to `knownConnections` with its id

`home-only` is still set to `true` in `launch_openpilot.sh` in case UI fails to start or we fail to modify the connection, but once UI starts it uses `UpdateUnsaved` to temporarily modify the setting if the value doesn't match the `GsmRoaming` param


selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -81,6 +81,12 @@ TogglesPanel::TogglesPanel(QWidget *parent) : QWidget(parent) {
                                   this));
 #endif
 
+  toggles.append(new ParamControl("GsmRoaming",
```
adeebshihadeh, line 84: this should go in AdvancedNetworking

launch_chffrplus.sh
```diff
@@ -101,7 +101,6 @@ function tici_init {
   sudo su -c 'echo "performance" > /sys/class/devfreq/soc:qcom,memlat-cpu0/governor'
   sudo su -c 'echo "performance" > /sys/class/devfreq/soc:qcom,memlat-cpu4/governor'
   nmcli connection modify --temporary lte gsm.auto-config yes
```
sshane, line 103: This is already set in the OS as well, remove?

pd0wm: Doesn't seem to work from the OS. Try removing it from launch, and read back the property.

pd0wm:
If possible I don't want to modify the LTE connection itself permanently. I think if you make non temporary changes it will copy the connection file to `/data/etc/NetworkManager/system-connections`, and then we can't make changes to it anymore when we do an OS update.

If you want the roaming state to persist across reboots store it in a param.

launch_chffrplus.sh
sshane: I restarted without launching launch_openpilot and auto-config was true for me. I'll make sure with a clean reflash though

sshane: It reverted to `no` after I deleted the `nmconnection` file in `/data`, I'll leave it in

launch_chffrplus.sh
```diff
@@ -101,7 +101,6 @@ function tici_init {
...
-  nmcli connection modify --temporary lte gsm.home-only yes
```
sshane, line 104: Should we keep this around, in case `ui` ever fails to start and we still don't want roaming?

pd0wm: Probably nice. Outside of the US roaming can cost quite a bit of money, so probably better to err on the side of caution.

PR #21751, opened 2021-07-27
keyboard: smaller password character

sshane:
This is the next size down from the current (U+25CF) but it seems pretty small.


adeebshihadeh:
fixed in d0a650a8ba829cb923f761e02232199eeddaf74b

PR #21753, opened 2021-07-28
ui: grey out unsupported networks

sshane:


selfdrive/assets/img_circled_slash.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="104" height="104" viewBox="0 0 104 104" fill="none" xmlns="http://www.w3.org/2000/svg">
```
adeebshihadeh, line 1: need to add this to assets.qrc

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -245,10 +249,11 @@ void WifiUI::refresh() {
     // Clickable SSID label
     QPushButton *ssidLabel = new QPushButton(network.ssid);
     ssidLabel->setObjectName("ssidLabel");
-    ssidLabel->setEnabled(network.connected == ConnectedType::DISCONNECTED &&
-                          network.security_type != SecurityType::UNSUPPORTED);
+    ssidLabel->setEnabled(network.security_type != SecurityType::UNSUPPORTED);
```
adeebshihadeh, line 252: why enable if we're connected or connecting?

sshane: enabled is tied to the greyed out style sheet. We only connect the button if we're disconnected below, but I can keep this the same and just set a property for the style sheet to use

PR #21754, opened 2021-07-28
H7 test

briskspirit:
Added static list() method to Panda class and also rewrote init to support connection by panda usb serial number. 
Still WIP to make it handle multiple pandas.

P.S.: not everything is going to be merged from this PR, have some helpers and hacks included here for my local tests!

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -437,7 +437,7 @@ void panda_state_thread(bool spoofing_started) {
       ps.setVoltage(pandaState.voltage);
       ps.setCurrent(pandaState.current);
     }
-    
+
     ps.setIgnitionLine(pandaState.ignition_line);
     ps.setIgnitionCan(pandaState.ignition_can);
     ps.setControlsAllowed(pandaState.controls_allowed);
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -37,8 +37,8 @@
 
 Panda *panda = nullptr;
 Panda *panda_aux = nullptr;
-uint8_t main_shift = 0;
-uint8_t aux_shift = 5;
+uint8_t panda_first_bus = 0;
+uint8_t panda_aux_first_bus = 5;
 std::vector<std::string> pandas_detected;
 std::vector<std::string> pandas_connected;
 std::atomic<bool> safety_setter_thread_running(false);
@@ -218,8 +218,8 @@ void can_recv(PubMaster &pm) {
   uint32_t data[(RECV_SIZE/4)*2]; // Double the size, for both pandas
   size_t num_msg = 0;
 
-  num_msg = panda->can_receive_raw(&data[num_msg*4], main_shift);
-  if (panda_aux != nullptr) num_msg += panda_aux->can_receive_raw(&data[num_msg*4], aux_shift);
+  num_msg = panda->can_receive_raw(&data[num_msg*4], panda_first_bus);
+  if (panda_aux != nullptr) num_msg += panda_aux->can_receive_raw(&data[num_msg*4], panda_aux_first_bus);
 
   MessageBuilder msg;
   auto evt = msg.initEvent();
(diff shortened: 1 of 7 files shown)


pd0wm:
Let me know when it's ready to review. Didn't look at the code much, but this PR has similar functionality: https://github.com/commaai/openpilot/pull/20061/files#diff-1895de1dfd645ff5591acfe43977a38fe28d542f6e124be200e4264a43744f01R37. Maybe there something useful in there for you as well.

briskspirit:
Sure, thank you! I will look into that PR and maybe later will take something from it, but for now it is slightly different task.

selfdrive/boardd/boardd.cc
```diff
@@ -165,27 +176,116 @@ bool usb_connect() {
   }
 
   connected_once = true;
-  panda = tmp_panda.release();
+  if (tmp_panda->hw_type != cereal::PandaState::PandaType::UNO && tmp_panda->hw_type != cereal::PandaState::PandaType::DOS) {
```
pd0wm, line 179: This would break EON gold or PC with black panda. Can we rewrite the check such that those keep working?

selfdrive/boardd/boardd.cc
```diff
@@ -165,27 +176,116 @@ bool usb_connect() {
...
   return true;
 }
 
 // must be called before threads or with mutex
 static bool usb_retry_connect() {
-  LOGW("attempting to connect");
-  while (!do_exit && !usb_connect()) { util::sleep_for(100); }
+  
+  LOGW("attempting to find boards");
+  size_t detect_count = 0;
```
pd0wm, line 191: No need for temp variable. Just use `pandas_detected.size()`

pd0wm:
Why not create a subclass of Panda for the red panda that override `can_send` and `can_recv`. Then that logic doesn't need to be moved to boardd.cc

selfdrive/boardd/boardd.cc
```diff
@@ -272,17 +372,38 @@ void panda_state_thread(bool spoofing_started) {
   // run at 2hz
   while (!do_exit && panda->connected) {
     health_t pandaState = panda->get_state();
+    health_t pandaState_aux;
+    if (panda_aux != nullptr) pandaState_aux = panda_aux->get_state();
+
+    uint8_t safety_model = pandaState.safety_model;
```
pd0wm, line 378: Why copy all these fields out?

selfdrive/boardd/boardd.cc
```diff
@@ -165,27 +176,116 @@ bool usb_connect() {
...
+    int len = data[i*4+1]&0xF;
+    canData[i].setDat(kj::arrayPtr((uint8_t*)&data[i*4+2], len));
+    canData[i].setSrc((data[i*4+1] >> 4) & 0xff);
+  }
+  auto can_data = capnp::messageToFlatArray(msg);
   auto bytes = can_data.asBytes();
   pm.send("can", bytes.begin(), bytes.size());
 }
 
+void split_messages(capnp::List<cereal::CanData>::Reader can_data_list) {
```
pd0wm, line 242: Why do we need to split the messages? Maybe it's better to move this into the panda class as well. When initializing a panda object you can give it a list of busses to use, and an optional shift. Then you can pass the full list of messages into the `can_send` function of both pandas.

pd0wm:
Should we make the shift 5? It's still on the list to to move the OBD bus on a regular panda to bus 4 (instead of sharing bus 1). If we commit to 4 now making that change later will be more annoying.

pd0wm:
Alternatively to the shifting the bus it might be cleaner to add a `pandaIndex` to the `CanData` struct. What do you think about that?

selfdrive/boardd/boardd.cc
briskspirit: I assume that secondary panda will be supported only by Comma 3, so I can check for DOS as primary and only on C3.

selfdrive/boardd/boardd.cc
briskspirit: I need it as I am removing from list of detected pandas, so line 203 will fail if I will use pandas_detected.size()

selfdrive/boardd/boardd.cc
briskspirit: They are used later in pandaState message if aux panda was chosen as a main "drive" panda. (But we still care about some state variables from DOS as it is driving IR, fan, camera clock etc.)

selfdrive/boardd/boardd.cc
briskspirit: I like that idea, thank you!

briskspirit:
> Should we make the shift 5? Or maybe even 10? It's still on the list to to move the OBD bus on a regular panda to bus 4 (instead of sharing bus 1). If we commit to 4 now making that change later will be more annoying.

In that case I would use shift 16, more logical to implement same way as we do for sent back messages from USB. And OBD definitely needs to be moved to its own bus number, because now it is very confusing (and having gmlan on bus 3 is also a bit confusing but seems like it is difficult to change that now)

> Alternatively to the shifting the bus it might be cleaner to add a pandaIndex to the CanData struct. What do you think about that? All the different bus numbers are already confusing enough as is.

Add pandaIndex and also pass this index as an optional argument to Panda instance? And then filter messages in can_send based on that index? I like it, definitely looks a lot less confusing, think we should use that approach.

selfdrive/boardd/boardd.cc
pd0wm: Why not two structs and copy over the necessary values? This adds a bunch of lines that are not needed.

PR #21757, opened 2021-07-28
Qt/input: multi-touch

deanlee:


adeebshihadeh:
Merging this as it's a big improvement to the keyboard, but can we make the whole UI multitouch?

PR #21764, opened 2021-07-28
Qt/SetingWindow: faster initialization and painting

deanlee:
1. reduce init time from ~520 ms to ~200ms (EON)
2. merge  #21560  faster painting(EON&PC), fast scrolling(PC only)


deanlee:
```
 panel_widget->setStyleSheet(R"(
    border-radius: 30px;
    background-color: #292929;
  )");
```
remove this can save 80 ms on EON.it apply this style to all children.

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -41,6 +41,7 @@ Networking::Networking(QWidget* parent, bool show_advanced) : QFrame(parent) {
   connect(wifiWidget, &WifiUI::connectToNetwork, this, &Networking::connectToNetwork);
 
   ScrollView *wifiScroller = new ScrollView(wifiWidget, this);
+  wifiScroller->setStyleSheet("background-color: #292929;");
```
adeebshihadeh, line 44: do we really need to set this everywhere? any docs on how this impacts perf or did you use something like GammaRay?

deanlee: There is no clear documentation. I made these changes through my own tests and by reading Qt source code.

pd0wm:
2021-11-17-122144_2153x1077_scrot

Something is off with the toggle background description.

deanlee:
It may be a merge problem, I will fix it later

deanlee:
Yes, I haven't fixed it yet. the code has changed too much since it was first submitted. I will open it for review after fixed the background.

PR #21779, opened 2021-07-29
Kia Niro HEV 2021 port.

tetious:
This is a basic port, based on the 2019 HEV and 2020 EV models. The steering ratio is from the spec sheet and seems to work well.

steering-specs

Steering also seems to work all the way to 0mph, unlike the 2019, though the manual claims otherwise. Given these differences, I decided to break this year into a new model. I suspect it will work with the 2020 too. (after fingerprinting) 

I do still get the stock hands on wheel warning, though not the audible one. (yet?) The HDA automatic speed limit updating might not be working with the stock SCC. Possibly a CAN message is not getting forwarded.

For other's reference, this car uses the Hyundai D harness.

Thanks for all your amazing work on Openpilot!

**Checklist**
- [x] added to README
- [x] route with openpilot: 34a875f29f69841a|2021-07-29--13-02-09


speedking456:
I did the port for the 2021 Kia Seltos and have the same issue with HDA speed limits not working. Maybe that's something we could look into.

tetious:
> I did the port for the 2021 Kia Seltos and have the same issue with HDA speed limits not working. Maybe that's something we could look into.

If you want to dig into the CAN data to try to find the message (if it is even being logged?) there is a speed change at around 2888 in Cabana for the test route: 34a875f29f69841a|2021-07-29--13-02-09

I poked around a bit and nothing jumped out at me.

It would be fun to try to figure this out. :) Feel free to contact me in Discord. (tetious#1124)

adeebshihadeh:
Can you rebase on master? There's a few merge conflicts after merging some other HKG car ports.

selfdrive/car/hyundai/interface.py
```diff
@@ -153,6 +153,14 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.25], [0.05]]
       if candidate == CAR.KIA_NIRO_HEV:
         ret.minSteerSpeed = 32 * CV.MPH_TO_MS
+    elif candidate == CAR.KIA_NIRO_HEV_2021:
```
adeebshihadeh, line 156: same parameters as CAR.KIA_NIRO_HEV, just add it to the list above

tetious: Done!

PR #21780, opened 2021-07-29
UI: lazy onboarding widgets init

deanlee:
`OnboardingWindow::showEvent` is always called on startup  which will cause the TermsPage to be initialized every time( in `TermsPage::showEvent`).~and the screen may flicker.~


adeebshihadeh:
Does this actually happen? I've never seen it.

deanlee:
yes.it's always visible  then hide immediately on startup,you can add an assert or printf here to check that:
https://github.com/commaai/openpilot/blob/1f341a28102cf9e8a1bdab465f47e85d301da8d7/selfdrive/ui/qt/offroad/onboarding.cc#L121-L125

adeebshihadeh:
I understand that, but does it ever flicker?

deanlee:
I'm not very sure, but when I was debugging on my laptop, I still can see a very fast flicker in the background after #21695. then I debugged to here.It seems that there is no such problem now

PR #21788, opened 2021-07-30
keyboard multitouch improvements

sshane:
supports multiple keys down, and each is 'clicked' when touch is released, like a real keyboard


selfdrive/ui/qt/widgets/keyboard.cc
```diff
@@ -20,13 +20,14 @@ const float key_spacing_horizontal = 15;
 
 KeyButton::KeyButton(const QString &text, QWidget *parent) : QPushButton(text, parent) {
   setAttribute(Qt::WA_AcceptTouchEvents);
+  setFocusPolicy(Qt::NoFocus);
```
adeebshihadeh, line 23: does this do anything?

PR #21800, opened 2021-07-30
proclogd: reduce the size of the procLog message by nearly half

deanlee:
- [x] no orphans
- [x] test case
- [x] cleanup

found the following bugs during unit testing that needed to be fixed:
- [x]  can't parse process stat if process name contains a `)` (e.g. `6560 ((sd-pam)) S 6556 6556 6556 0 -1 1077936448 52 0 0 0 0 0 0 0 20 0 1 0 8649 175263744 969 18446744073709551615 1 1 0 0 0 0 0 4096 0 0 0 0 17 2 0 0 0 0 0 0 0 0 0 0 0 0 0`)
- [x] set a wrong value(`irqtime`) to Softirq,it should be `sirqtime`


pd0wm:
Where do these orphans come from? Can we just prevent them being created in the first place?

deanlee:
https://capnproto.org/cxx.html
orphan's space is allocated in the same MessageBuilder. even though an orphan is adopted, it still resides inside the memory allocated. the orphan’s content will be part of the serialized message.the only way to remove this "memory hole" is to copy the entire message over to a new MessageBuilder before serializing, since only the reachable objects will be copied.
orphan is mostly used for internal states. we can use std::vector instead of orphan.

pd0wm:
Thanks for writing a test! Can you also add a more end-to-end test where it verifies a properly formatted message is sent over the socket?

deanlee:
ok

deanlee:
found some bugs in unit testing that needed to be fixed:

- [ ] can't parse process stat if process name contains '`)`' (e.g. `6560 ((sd-pam)) S 6556 6556 6556 0 -1 1077936448 52 0 0 0 0 0 0 0 20 0 1 0 8649 175263744 969 18446744073709551615 1 1 0 0 0 0 0 4096 0 0 0 0 17 2 0 0 0 0 0 0 0 0 0 0 0 0 0`)

PR #21806, opened 2021-07-30
fix cd in process scripts

gregjhogan:


selfdrive/ui/ui
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
+cd "$(dirname "$0")"
```
adeebshihadeh, line 2: if we do this, let's add to all scripts

PR #21807, opened 2021-07-30
2021 Elantra Fw versions

sternm33:


VirtuallyChris:
The PR needs to be done against master, and please provide your dongle ID.

PR #21809, opened 2021-07-30
Params: add function readMultiple

deanlee:


pd0wm:
Do you see any measurable performance improvement in refreshing the offroad alerts? Not sure if holding the lock for one longer period is better than multiple shorter locks.

adeebshihadeh:
I'm also skeptical that this is worth it. Maybe we could refactor the offroad alerts into one param if this actually matters, but it'd have to be a significant cleanup or performance improvement to be worth it.

deanlee:
closed

PR #21810, opened 2021-07-30
qt/util: new function getDongleId

deanlee:

### DIFF
--- a/selfdrive/common/util.cc
+++ b/selfdrive/common/util.cc
@@ -155,10 +155,6 @@ std::string dir_name(std::string const &path) {
   return path.substr(0, pos);
 }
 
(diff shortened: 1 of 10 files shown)


pd0wm:
Think it makes more sense to return an `std::optional<QString>`. Now you still have to check if it's empty, so there is not much benefit to calling the `is_valid_dongle_id` function.

pd0wm:
Looks much better. Think this is an improvement over the old code since it forces us to think about the case where there is no dongle id.

pd0wm:
I'm removing the static caching. Will cause some unexpected behavior if you're debugging/testing stuff around dongleIds.

deanlee:
OK. cache is really not necessary

PR #21811, opened 2021-07-30
logger.cc/logger_build_init_data(): reuse params_map

deanlee:


selfdrive/loggerd/logger.cc
```diff
@@ -94,23 +94,22 @@ kj::Array<capnp::word> logger_build_init_data() {
...
-      lentry.setValue(capnp::Data::Reader((const kj::byte*)kv.second.data(), kv.second.size()));
-      i++;
-    }
+  std::map<std::string, std::string> params_map;
+  Params().readAll(&params_map);
+
+  init.setGitCommit(params_map["GitCommit"]);
+  init.setGitBranch(params_map["GitBranch"]);
+  init.setGitRemote(params_map["GitRemote"]);
+  init.setPassive(params_map["Passive"] == "1");
```
pd0wm, line 103:
```suggestion
  init.setPassive(params.getBool("Passive"));
```

Let's keep using the helper function here in case we ever change the definition of what values classify as true.

PR #21814, opened 2021-07-31
api.h: add global variable for api base url

deanlee:


pd0wm:
Can you also update the usage from python (`common/api/__init__.py`)?

PR #21815, opened 2021-07-31
util/read_files_in_dir: return map by value

deanlee:
It's efficient to return std::map as value with NRVO  in C++17.


pd0wm:
Great. The original code was probably a premature optimization anyway, the maps are pretty small and the function is only used on process init.

PR #21819, opened 2021-08-01
Add Car Hyundai KONA Hybrid 2020

haram-KONA:
<!--- ***** Add hyundai KONA Hybrid 2020 *****

**Description** [](Add Car KONA Hybrid 2020.)

**Verification** [](results: 49f3c13141b6bc87|2021-07-28--08-05-13--0)

**Route**
Route: [49f3c13141b6bc87|2021-07-28--08-05-13]

DonglID : 49f3c13141b6bc87


haram-KONA:
I don't know how to solve this problem....
Why is the model verification of the car failing?
What are the common causes of these things?

It's difficult .. for a beginner to do ...

pd0wm:
Since this is a new car port we need an example route we can use for our regression tests. Do you have a route we can be made public and used in our test suite?

haram-KONA:
> Since this is a new car port we need an example route we can use for our regression tests. Do you have a route that can be made public and used in our test suite?

(results: 49f3c13141b6bc87|2021-07-28--08-05-13--0)

**Route**
Route: [49f3c13141b6bc87|2021-07-28--08-05-13]

haram-KONA:
Do I have to do something else to apply?
It is very cumbersome to drive by changing the settings every update.

adeebshihadeh:
Merged in #21912

PR #21826, opened 2021-08-02
hw.h: move Hardware-related paths to namespace Path

deanlee:


selfdrive/hardware/hw.h
```diff
@@ -16,3 +17,24 @@ class HardwarePC : public HardwareNone {
...
+inline std::string params() {
+  return Hardware::PC() ? HOME + "/.comma/params" : "/data/params";
+}
+inline std::string persistent_params() {
+  return Hardware::PC() ? HOME + "/.comma/params" : "/persist/comma/params";
+}
+inline std::string rsa_file() {
+  return Hardware::PC() ? HOME + "/.comma/persist/comma/id_rsa" : "/persist/comma/id_rsa";
+}
+inline std::string json_token() {
```
pd0wm, line 37: Don't think this is right. This path should not be device dependent. Think it's only used on PC though.

selfdrive/hardware/hw.h
```diff
@@ -16,3 +17,24 @@ class HardwarePC : public HardwareNone {
 };
 #define Hardware HardwarePC
 #endif
+
+namespace Path {
+inline static std::string HOME = util::getenv("HOME");
+inline std::string log_root() {
+  if (const char *env = getenv("LOG_ROOT"))
```
pd0wm, line 24: Use brackets or put statement on same line as if.

PR #21832, opened 2021-08-03
loggerd: implement DONT_LOG flag

geohot:
Some params have sensitive information. Don't put them in the logs.

### DIFF
--- a/selfdrive/loggerd/logger.cc
+++ b/selfdrive/loggerd/logger.cc
@@ -108,7 +108,9 @@ kj::Array<capnp::word> logger_build_init_data() {
   for (auto& [key, value] : params_map) {
     auto lentry = lparams[i];
     lentry.setKey(key);
-    lentry.setValue(capnp::Data::Reader((const kj::byte*)value.data(), value.size()));
+    if ( !(params.getKeyType(key) & DONT_LOG) ) {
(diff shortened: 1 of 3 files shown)


selfdrive/loggerd/logger.cc
```diff
@@ -108,7 +108,9 @@ kj::Array<capnp::word> logger_build_init_data() {
   for (auto& [key, value] : params_map) {
     auto lentry = lparams[i];
     lentry.setKey(key);
-    lentry.setValue(capnp::Data::Reader((const kj::byte*)value.data(), value.size()));
+    if ( !(params.getKeyType(key) & DONT_LOG) ) {
```
adeebshihadeh, line 111: hmm, are you sure the precedence is right?

adeebshihadeh: might be better to add a testKey helper

geohot: Not sure what you mean. Is that wrong somehow?

`batman@x840:/tmp$ cat a.c && ./a.out
#define DONT_LOG 0x40
int main() {
  int a = 0x40;
  int b = 0;
  if ( !(a & DONT_LOG) ) printf("not set on a\n");
  if ( !(b & DONT_LOG) ) printf("not set on b\n");
}
not set on b
`

adeebshihadeh: missed the parantheses

PR #21845, opened 2021-08-03
Cleanup honda long control

haraschax:
This should deal with the undershooting, shorten the follow distance and cleanup the api so it is similar to other cars.
Now Honda Nidec's rely on some hacky specific code that is only for honda Nidecs

### DIFF
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -156,6 +156,7 @@ class CarController():
         can_sends.append(hondacan.spam_buttons_command(self.packer, CruiseButtons.CANCEL, idx, CS.CP.carFingerprint))
       elif CS.out.cruiseState.standstill:
         can_sends.append(hondacan.spam_buttons_command(self.packer, CruiseButtons.RES_ACCEL, idx, CS.CP.carFingerprint))
+      pcm_accel = 0
 
     else:
       # Send gas and brake commands.
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -89,7 +89,7 @@ class CarController():
     self.params = CarControllerParams(CP)
 
   def update(self, enabled, CS, frame, actuators,
-             pcm_speed, pcm_override, pcm_cancel_cmd, pcm_accel,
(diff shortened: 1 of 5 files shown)


selfdrive/controls/controlsd.py
```diff
@@ -506,20 +506,10 @@ def publish_logs(self, CS, start_time, actuators, lac_log):
...
     if self.joystick_mode and self.sm.rcv_frame['testJoystick'] > 0 and self.sm['testJoystick'].buttons[0]:
       CC.cruiseControl.cancel = True
 
-    # TODO remove car specific stuff in controls
-    # Some override values for Honda
-    # brake discount removes a sharp nonlinearity
-    brake_discount = (1.0 - clip(actuators.brake * 3., 0.0, 1.0))
-    speed_override = max(0.0, (self.LoC.v_pid + CS.cruiseState.speedOffset) * brake_discount)
-    CC.cruiseControl.speedOverride = float(speed_override if self.CP.pcmCruise else 0.0)
-    CC.cruiseControl.accelOverride = float(self.CI.calc_accel_override(CS.aEgo, self.a_target,
```
adeebshihadeh, line 520: let's make sure to deprecate these fields

selfdrive/car/honda/carcontroller.py
```diff
@@ -148,11 +146,22 @@ def update(self, enabled, CS, frame, actuators,
...
+
+    # Prevent rolling backwards
+    accel = -1.0 if stopping else accel
+    apply_accel = interp(accel, P.BOSCH_ACCEL_LOOKUP_BP, P.BOSCH_ACCEL_LOOKUP_V)
+
+    gas_mult = interp(CS.out.vEgo, [0.0, 1.0], [2.0, 1.0])
+    pcm_speed = max(0.0, CS.out.vEgo + gas_mult * apply_accel)
+
+    # This is needed otherwise accel to decel is not smooth
+    pcm_accel = int(clip(apply_accel, 0.0, 1.0) * 0xc6)
```
gregjhogan, line 164: nidec is using bosch accel breakpoints/values?

HaraldSchafer: Yeah lemme rename

PR #21850, opened 2021-08-04
Sonata Hybrid 2021

Matt-Wash-Burn:
***** Hyundai Sonata Hybrid: Car port *****

**Checklist**
- [x] added to README
- [x] route with openpilot:


selfdrive/car/hyundai/carcontroller.py
```diff
@@ -78,7 +78,7 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert,
 
     # 20 Hz LFA MFA message
     if frame % 5 == 0 and self.car_fingerprint in [CAR.SONATA, CAR.PALISADE, CAR.IONIQ, CAR.KIA_NIRO_EV, CAR.KONA_EV,
-                                                   CAR.IONIQ_EV_2020, CAR.IONIQ_PHEV, CAR.KIA_CEED, CAR.KIA_SELTOS, CAR.ELANTRA_2021, CAR.ELANTRA_HEV_2021]:
+                                                   CAR.IONIQ_EV_2020, CAR.IONIQ_PHEV, CAR.KIA_CEED, CAR.KIA_SELTOS, CAR.ELANTRA_2021, CAR.ELANTRA_HEV_2021,CAR.SONATA_HYBRID]:
```
is343, line 81: Instead of `HYBRID`, should we stick with the `HEV` convention that's used elsewhere?

is343 (commented):
Hey, thanks for doing this. I was planning on adding my 2021 Sonata Hybrid Limited once my comma 3 came in, but it looks like you beat me to it. I'm glad this is getting added. 

I just had a comment regarding the current naming convention for hybrids. What do you think?

selfdrive/car/hyundai/values.py
```diff
@@ -36,6 +36,7 @@ class CAR:
   SONATA_LF = "HYUNDAI SONATA 2019"
   PALISADE = "HYUNDAI PALISADE 2020"
   VELOSTER = "HYUNDAI VELOSTER 2019"
+  SONATA_HYBRID= "HYUNDAI SONATA HYBRID 2021"
```
adeebshihadeh, line 39:
```suggestion
  SONATA_HYBRID = "HYUNDAI SONATA HYBRID 2021"
```

selfdrive/car/hyundai/carcontroller.py
adeebshihadeh: I want to update them all to hybrid anyway to match the other makes

is343: Ah alright, thanks. Figured we should keep it consistent. If we want to change the others instead, that makes sense too.

PR #21855, opened 2021-08-04
LoggerState: test cases

deanlee:
multiple threaded logging problem during testing that needs to be fixed:
- [x] SentinelType may not be the last msg in the log


pd0wm:
What's the benefit over the python test?

deanlee:
it can do functional testing

adeebshihadeh:
I'd like this to supplement the python tests. Let's keep the existing python test cases.

deanlee:
closed. move to c++ logger #21877

selfdrive/loggerd/logger.h
```diff
@@ -53,8 +54,12 @@ class BZFile {
   BZFILE* bz_file = nullptr;
 };
 
+typedef cereal::Sentinel::SentinelType SentinelType;
+
 typedef struct LoggerHandle {
   pthread_mutex_t lock;
+  SentinelType end_sentinel_type;
+  bool exit_signal;
```
pd0wm, line 62: exit signal is not a bool. We want to distinguish SIGINT from SIGPWR.

deanlee: fixed

PR #21857, opened 2021-08-04
Cleaner lane change alerts

haraschax:


selfdrive/controls/lib/events.py
```diff
@@ -482,33 +482,33 @@ def joystick_alert(CP: car.CarParams, sm: messaging.SubMaster, metric: bool) ->
...
-      AlertStatus.userPrompt, AlertSize.mid,
-      Priority.LOW, VisualAlert.steerRequired, AudibleAlert.chimePrompt, .1, .1, .1),
+      "",
+      AlertStatus.userPrompt, AlertSize.small,
+      Priority.LOW, VisualAlert.none, AudibleAlert.chimePrompt, .1, .1, .1),
   },
 
   EventName.laneChange: {
     ET.WARNING: Alert(
       "Changing Lane",
```
VirtuallyChris, line 508: Maybe we change this one line to "Monitor Other Vehicles" so at least it's somewhere? I think the user will know it's lane changing.

adeebshihadeh: I think it would be nicer to integrate "confirm" into the initial alert text

PR #21550, opened 2021-07-11
refactor loggerd: trigger rotate in encoder thread

deanlee:
related issue: https://github.com/commaai/openpilot/issues/21465

PR #21860, opened 2021-08-05
refactor loggerd: trigger rotate in encoder thread

pd0wm:
Continuation of https://github.com/commaai/openpilot/pull/21550

### DIFF
--- a/selfdrive/loggerd/loggerd.cc
+++ b/selfdrive/loggerd/loggerd.cc
@@ -150,7 +150,9 @@ void encoder_thread(int cam_idx) {
       VisionBuf* buf = vipc_client.recv(&extra);
       if (buf == nullptr) continue;
 
-      s.last_camera_seen_tms = millis_since_boot();
+      if (cam_info.trigger_rotate) {
+        s.last_camera_seen_tms = millis_since_boot();
+      }
 
       if (cam_info.trigger_rotate && (cnt >= SEGMENT_LENGTH * MAIN_FPS)) {
         // trigger rotate and wait logger rotated to new segment
--- a/selfdrive/loggerd/loggerd.cc
+++ b/selfdrive/loggerd/loggerd.cc
@@ -244,6 +244,23 @@ void logger_rotate() {
   LOGW((s.logger.part == 0) ? "logging to %s" : "rotated to %s", s.segment_path);
 }
 
+void rotator_thread() {
+  while (!do_exit) {
+    if (s.waiting_rotate == s.max_waiting) {
+      logger_rotate();
+    }
+
+    double tms = millis_since_boot();
+    if ((tms - s.last_rotate_tms) > SEGMENT_LENGTH * 1000 &&
+        (tms - s.last_camera_seen_tms) > NO_CAMERA_PATIENCE) {
+      LOGW("no camera packet seen. auto rotating");
+      logger_rotate();
+    }
+
+    util::sleep_for(1);
+  }
+}
+
 } // namespace
 
 int main(int argc, char** argv) {
@@ -299,6 +316,8 @@ int main(int argc, char** argv) {
     }
   }
 
+  auto rotator = std::thread(rotator_thread);
+
   uint64_t msg_count = 0, bytes_count = 0;
   double start_ts = millis_since_boot();
   while (!do_exit) {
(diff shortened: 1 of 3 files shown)


selfdrive/loggerd/loggerd.cc
```diff
@@ -198,62 +147,32 @@ void encoder_thread(int cam_idx) {
...
-          pthread_mutex_unlock(&s.rotate_lock);
+      s.last_camera_seen_tms = millis_since_boot();
 
-          // wait for all to finish rotating
-          for(auto &r : s.rotate_state) {
-             while(r.enabled && r.cur_seg != s.rotate_segment && !do_exit) util::sleep_for(5);
-          }
-          rotate_state.rotating = false;
-          rotate_state.finish_rotate();
+      if (cam_info.trigger_rotate && (cnt > 0 && (cnt % (SEGMENT_LENGTH * MAIN_FPS)) == 0)) {
```
pd0wm, line 154:
```suggestion
      if (cam_info.trigger_rotate && (cnt >= SEGMENT_LENGTH * MAIN_FPS)) {
```

pd0wm:
Also the `encoder_thread` is still not actually triggering the rotate, it just stops encoding frames until the main thread rotates.

selfdrive/loggerd/loggerd.cc
```diff
@@ -151,7 +151,7 @@ void encoder_thread(int cam_idx) {
 
       s.last_camera_seen_tms = millis_since_boot();
 
-      if (cam_info.trigger_rotate && (cnt > 0 && (cnt % (SEGMENT_LENGTH * MAIN_FPS)) == 0)) {
+      if (cam_info.trigger_rotate && (cnt >= SEGMENT_LENGTH * MAIN_FPS)) {
```
deanlee, line 154: if we use >=, we should set cnt = 0 after rotation

deanlee:
sorry, we can't just set cnd to zero, the encoderid will be incorrect:
https://github.com/commaai/openpilot/blob/1e00db152106219ab828ca5508a102d6a841cb31/selfdrive/loggerd/loggerd.cc#L196

How about using a new variable frame_cnt to handle rotation?

selfdrive/loggerd/loggerd.cc
```diff
@@ -335,138 +286,66 @@ int main(int argc, char** argv) {
...
-        logger_log(&s.logger, (uint8_t*)msg->getData(), msg->getSize(), qs.counter == 0 && qs.freq != -1);
-        if (qs.freq != -1) {
-          qs.counter = (qs.counter + 1) % qs.freq;
-        }
-
+      QlogState &qs = qlog_states[sock];
+      Message *msg = nullptr;
+      while (!do_exit && (msg = sock->receive(true))) {
+        const bool in_qlog = qs.freq != -1 && (qs.counter++ % qs.freq == 0);
+        logger_log(&s.logger, (uint8_t *)msg->getData(), msg->getSize(), in_qlog);
```
deanlee, line 333: logger_log will block logger_next wile writing. If we don’t use lh_log here, an additional rotation thread will not help much.

pd0wm: But that's only while a single message is logged. That should be fine, right? I don't want to wait till a while socket is drained.

deanlee: how about check  `if (s.waiting_rotate == s.max_waiting)`  in while loop after logger_log ? It may be better than checking it every 1ms in another thread

pd0wm: Guess I can put that all in a helper function and call that.

selfdrive/loggerd/loggerd.cc
```diff
@@ -198,62 +148,34 @@ void encoder_thread(int cam_idx) {
...
-        // wait if camera pkt id is older than stream
-        rotate_state.waitLogThread();
-
-        if (do_exit) break;
+      if (buf == nullptr) continue;
 
-        // rotate the encoder if the logger is on a newer segment
-        if (rotate_state.should_rotate) {
-          LOGW("camera %d rotate encoder to %s", cam_idx, s.segment_path);
+      s.last_camera_seen_tms = millis_since_boot();
```
adeebshihadeh, line 153: gate this on cam_info.trigger_rotate?

pd0wm: Sure. Don't think it matters much in practice.

PR #21877, opened 2021-08-08
c++ Logger

deanlee:


pd0wm:
I liked the test being in a separate PR that we can merge separately/first.

We already made quite a bit of changes to loggerd already, so want to make sure it's stable first before merging big changes like this.

deanlee:
move testing to #21855

PR #21880, opened 2021-08-08
fix: util::file_exists will return false if failed to open existing file

deanlee:
e.g. write protected file


pd0wm:
Can you add a test to `test_util.cc`?

PR #21883, opened 2021-08-09
fix: util::read_files_in_dir does not always return a correct result

deanlee:
fix two problems:
1.  return dir's name in map with empty value.
2.  can't return files which name starts with an non alphanumeric character.(e.g.  `.file`, `'file 1'`.


pd0wm:
How do we want to deal with symlinks?

PR #21890, opened 2021-08-10
Update long_mpc api for use with sim

mitchellgoffpc:


selfdrive/controls/lib/long_mpc.py
```diff
@@ -45,12 +45,14 @@ def set_cur_state(self, v, a):
     self.cur_state[0].v_ego = v_safe
     self.cur_state[0].a_ego = a_safe
 
-  def update(self, carstate, model, v_cruise):
+  def update(self, carstate, radarstate, v_cruise):
```
adeebshihadeh, line 48: where is this used?

HaraldSchafer: It is not, it is to keep api the same with the other mpcs, this will go away once they are merged.

PR #21891, opened 2021-08-10
Japanese Prius PHV 2017 fingerprint

ikawaoka:
#2017 JAPAN Prius PHV solar

Added the fingerprint of the Japanese Prius PHV 2017 solar charging system equipped car.

The fingerprint is obtained from my car and is recognized normally.


pd0wm:
CAN ID based fingerprints are obsolete. Can you submit your ECU VW versions?

What is your dongle id?

PR #21902, opened 2021-08-11
fix InitData: params/entries contains empty values

deanlee:


pd0wm:
What's the benefit of this? Would still be good to know if the param was present on disk or not.

PR #21903, opened 2021-08-11
athena: return current upload status in listUploadQueue

### LINKED ISSUE #21471: improve athena uploading of raw logs
Currently when you request that raw logs get uploaded for a segment the athena it is not very reliable.  This is because it stops on any error and never retries.  Instead of having a separate uploader for athena, I think we should update the normal uploader to upload files with an xattr (instead of files without it) and then have the athena upload endpoint set this xattr instead of managing it's own upload queue (because the normal uploader is reliable).  The user should be able to request uploading segments/routes from explorer (see https://github.com/commaai/explorer/issues/117) and these requests should be uploaded first.

Once we implement this, we can potentially eliminate the raw upload toggle because I think we can upload anything requested on both wifi and cellular.  Often when putting your device on wifi to upload raw logs currently you are looking for something recent but have to wait forever while it uploads all the old data you don't really care about first.

pd0wm:

### DIFF
--- a/selfdrive/athena/athenad.py
+++ b/selfdrive/athena/athenad.py
@@ -52,7 +52,7 @@ log_recv_queue: Any = queue.Queue()
 cancelled_uploads: Any = set()
 UploadItem = namedtuple('UploadItem', ['path', 'url', 'headers', 'created_at', 'id', 'retry_count', 'current', 'progress'], defaults=(0, False, 0))
 
-cur_upload_item = None
+cur_upload_items = {}
 
 
 def handle_long_poll(ws):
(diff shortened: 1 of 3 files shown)


selfdrive/athena/athenad.py
```diff
@@ -100,35 +103,53 @@ def jsonrpc_handler(end_event):
...
+      if cur_upload_items[tid].id in cancelled_uploads:
+        cancelled_uploads.remove(cur_upload_items[tid].id)
         continue
 
       try:
-        _do_upload(item)
-      except (requests.exceptions.Timeout, requests.exceptions.ConnectionError, requests.exceptions.SSLError) as e:
-        cloudlog.warning(f"athena.upload_handler.retry {e} {item}")
+        def cb(sz, cur):
+          cur_upload_items[tid] = cur_upload_items[tid]._replace(progress=cur / sz if sz else 1)
```
gregjhogan, line 119: maybe it is time to make `UploadItem` a class, put this callback method on it, and change all the `_replace()` calls to be setting props?

pd0wm: I kinda like the immutability of the UploadItem named tuple actually. I'm still not a fan of modifying items while in the queue, that feels like an anti-pattern.

gregjhogan: yeah immutable objects are good for multi-threading, it just seems hard to follow how some of the pieces tie together, maybe the problem is mixing the thread state and the pending queue items

PR #21904, opened 2021-08-11
swaglog: add test case

deanlee:


deanlee:
It seems that we should handle `EAGAIN` on `zmq_send` carefully to avoid losing messages

PR #21905, opened 2021-08-11
Add argument to Panda to connect by serial number

briskspirit:
Add static list() to C++ Panda library and also change connection method for Panda.

### DIFF
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -91,20 +91,20 @@ void Panda::cleanup() {
 std::vector<std::string> Panda::list() {
   // init libusb
   std::vector<std::string> serials;
-  libusb_context *ctx = NULL;
-  libusb_device **dev_list = NULL;
+  libusb_context *ctx_tmp = NULL;
+  libusb_device **dev_list_tmp = NULL;
   size_t serials_idx = 0;
-  int err = libusb_init(&ctx);
+  int err = libusb_init(&ctx_tmp);
   if (err != 0) { return serials; }
 
 #if LIBUSB_API_VERSION >= 0x01000106
-  libusb_set_option(ctx, LIBUSB_OPTION_LOG_LEVEL, LIBUSB_LOG_LEVEL_INFO);
+  libusb_set_option(ctx_tmp, LIBUSB_OPTION_LOG_LEVEL, LIBUSB_LOG_LEVEL_INFO);
 #else
-  libusb_set_debug(ctx, 3);
+  libusb_set_debug(ctx_tmp, 3);
 #endif
 
-  for (size_t i = 0; i < libusb_get_device_list(ctx, &dev_list); ++i) {
-    libusb_device *device = dev_list[i];
+  for (size_t i = 0; i < libusb_get_device_list(ctx_tmp, &dev_list_tmp); ++i) {
+    libusb_device *device = dev_list_tmp[i];
     libusb_device_descriptor desc = { 0 };
 
     libusb_get_device_descriptor(device, &desc);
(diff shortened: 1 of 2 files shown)


selfdrive/boardd/panda.h
```diff
@@ -44,19 +44,24 @@ class Panda {
...
   std::mutex usb_lock;
   void handle_usb_issue(int err, const char func[]);
   void cleanup();
 
  public:
-  Panda();
+  Panda(std::string serial);
   ~Panda();
 
+  std::string usb_serial;
```
adeebshihadeh, line 56: make this a getSerial function and read like the python lib does https://github.com/commaai/panda/blob/master/python/__init__.py#L431

briskspirit: we already have get_serial function. And this is usb_serial, it is different from one returned by get_serial() in C++ or Python function.

selfdrive/boardd/panda.cc
```diff
@@ -22,7 +22,28 @@ Panda::Panda() {
   libusb_set_debug(ctx, 3);
 #endif
 
-  dev_handle = libusb_open_device_with_vid_pid(ctx, 0xbbaa, 0xddcc);
+  for (size_t i = 0; i < libusb_get_device_list(ctx, &dev_list); ++i) {
```
adeebshihadeh, line 25: don't duplicate this loop and code

selfdrive/boardd/panda.cc
```diff
@@ -67,6 +88,42 @@ void Panda::cleanup() {
...
+
+    libusb_get_device_descriptor(device, &desc);
+    if (desc.idVendor == 0xbbaa && desc.idProduct == 0xddcc) {
+      libusb_device_handle *handle = NULL;
+      libusb_open(device, &handle);
+      unsigned char serial[25];
+      int ret = libusb_get_string_descriptor_ascii(handle, desc.iSerialNumber, serial, sizeof(serial));
+      libusb_close(handle);
+
+      if (ret < 0) { return serials; }
```
adeebshihadeh, line 118: why not just break?

selfdrive/boardd/panda.cc
```diff
@@ -67,6 +88,42 @@ void Panda::cleanup() {
...
+      serials.resize(serials_idx+1);
+      serials[serials_idx] = (char *)serial;
+      ++serials_idx;
```
adeebshihadeh, line 121: just use push_back

selfdrive/boardd/panda.cc
```diff
@@ -67,6 +88,42 @@ void Panda::cleanup() {
...
 }
 
+std::vector<std::string> Panda::list() {
+  // init libusb
+  std::vector<std::string> serials;
+  libusb_context *ctx_tmp = NULL;
+  libusb_device **dev_list_tmp = NULL;
+  size_t serials_idx = 0;
+  int err = libusb_init(&ctx_tmp);
+  if (err != 0) { return serials; }
```
adeebshihadeh, line 98: this should probably throw some error or indicate something's wrong

selfdrive/boardd/panda.cc
```diff
@@ -97,9 +152,7 @@ int Panda::usb_read(uint8_t bRequest, uint16_t wValue, uint16_t wIndex, unsigned
   int err;
   const uint8_t bmRequestType = LIBUSB_ENDPOINT_IN | LIBUSB_REQUEST_TYPE_VENDOR | LIBUSB_RECIPIENT_DEVICE;
 
-  if (!connected) {
-    return LIBUSB_ERROR_NO_DEVICE;
-  }
+  if (!connected) return LIBUSB_ERROR_NO_DEVICE;
```
adeebshihadeh, line 155: this should be done in another PR

selfdrive/boardd/panda.cc
```diff
@@ -302,14 +349,9 @@ int Panda::can_receive(kj::Array<capnp::word>& out_buf) {
...
-    if (data[i*4] & 4) {
-      // extended
-      canData[i].setAddress(data[i*4] >> 3);
-      //printf("got extended: %x\n", data[i*4] >> 3);
-    } else {
-      // normal
-      canData[i].setAddress(data[i*4] >> 21);
-    }
+    if (data[i*4] & 4) canData[i].setAddress(data[i*4] >> 3); // extended
+    else canData[i].setAddress(data[i*4] >> 21); // normal
```
adeebshihadeh, line 353: this is harder to read and isn't within the scope of this PR

PR #21906, opened 2021-08-11
athena: add prefix param to listDataDirectory

### LINKED ISSUE #21471: improve athena uploading of raw logs

gregjhogan:


pd0wm:
Would using something like https://docs.python.org/3/library/glob.html be more efficient since it doesn't require walking the tree in python?

gregjhogan:
> Would using something like https://docs.python.org/3/library/glob.html be more efficient since it doesn't require walking the tree in python?

maybe I am missing something, but it looks to me like `glob` will walk the entire tree if you search recursively (we would need to use `**` to support partial directory name matches)

since this only supports a prefix I should be able to build something simple that uses `scandir()` more intelligently

pd0wm:
I assumed `glob` would talk directly to the os/filesystem, but's basically what you implemented but then with a regex. So disregard!

gregjhogan:
should be good to merge once tests pass

selfdrive/athena/athenad.py
```diff
@@ -171,11 +171,29 @@ def setNavDestination(latitude=0, longitude=0):
   return {"success": 1}
 
 
-@dispatcher.add_method
-def listDataDirectory():
-  files = [os.path.relpath(os.path.join(dp, f), ROOT) for dp, dn, fn in os.walk(ROOT) for f in fn]
+def scan_dir(path, prefix):
```
pd0wm, line 174:
```suggestion
def scan_dir(path, prefix):
  """Returns a list of files where either any of the parent folders or the filename itself matches the prefix"""
```

selfdrive/athena/tests/test_athenad.py
```diff
@@ -34,6 +35,13 @@ def tearDown(self):
     athenad.upload_queue = queue.Queue()
     athenad.cur_upload_items.clear()
 
+    for i in os.listdir(athenad.ROOT):
+      p = os.path.join(athenad.ROOT, i)
+      if os.path.isdir(p):
+        shutil.rmtree(p)
+      else:
+        os.unlink(p)
+
```
gregjhogan, line 44: I put this stuff in `setUp` because then if a test fails I can still inspect the state of things, is there some advantage to putting it in `tearDown`?  I just always felt better putting this kind of stuff in setup to ensure tests always start clean, too (would rather have exceptions in this code happen in `setUp`).

pd0wm: Oh, didn't notice I used tearDown. It's fine to put it back!

PR #21911, opened 2021-08-12
Change car controller interface from gas/brake to acceleration

haraschax:
This isn't a strict refactor, I will try to get the behavior mostly the same. Most cars do not have an interface to gas and/or brake. So it does not make sense for controls to output gas and brake to car interface. Convert from accel in car layers for clarity.


selfdrive/car/honda/interface.py
```diff
@@ -236,10 +188,6 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.lateralParams.torqueBP, ret.lateralParams.torqueV = [[0, 4096], [0, 4096]]  # TODO: determine if there is a dead zone at the top end
       tire_stiffness_factor = 0.677
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.6], [0.18]]
-      ret.longitudinalTuning.kpBP = [0., 5., 35.]
```
adeebshihadeh, line 239: this should've gone in another PR

selfdrive/car/honda/carcontroller.py
```diff
@@ -148,34 +177,30 @@ def update(self, enabled, CS, frame, actuators,
...
-    accel = -1.0 if stopping else accel
-    if CS.CP.carFingerprint in HONDA_BOSCH:
-      apply_accel = interp(accel, P.BOSCH_ACCEL_LOOKUP_BP, P.BOSCH_ACCEL_LOOKUP_V)
-    else:
-      apply_accel = interp(accel, P.NIDEC_ACCEL_LOOKUP_BP, P.NIDEC_ACCEL_LOOKUP_V)
+    accel = -4.0 if stopping else accel
 
     # wind brake from air resistance decel at high speed
-    wind_brake = interp(CS.out.vEgo, [0.0, 2.3, 35.0], [0.0, 0.0, 0.15])
+    wind_brake = interp(CS.out.vEgo, [0.0, 2.3, 35.0], [0.001, 0.002, 0.15])
```
pd0wm, line 189: Why did this change?

selfdrive/car/interfaces.py
```diff
@@ -72,15 +68,11 @@ def get_std_params(candidate, fingerprint):
...
-    ret.gasMaxV = [.5]       # half max brake
-    ret.brakeMaxBP = [0.]
-    ret.brakeMaxV = [1.]
     ret.openpilotLongitudinalControl = False
     ret.startAccel = 0.0
     ret.minSpeedCan = 0.3
-    ret.stoppingBrakeRate = 0.2 # brake_travel/s while trying to stop
-    ret.startingBrakeRate = 0.8 # brake_travel/s while releasing on restart
+    ret.stoppingDecelRate = 0.8 # brake_travel/s while trying to stop
+    ret.startingAccelRate = 3.6 # brake_travel/s while releasing on restart
```
pd0wm, line 75: Why a factor of 4.5 here?

selfdrive/car/toyota/interface.py
```diff
@@ -331,8 +324,8 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.longitudinalTuning.kpV = [1.3, 1.0, 0.7]
       ret.longitudinalTuning.kiBP = [0., 5., 12., 20., 27.]
       ret.longitudinalTuning.kiV = [.35, .23, .20, .17, .1]
-      ret.stoppingBrakeRate = 0.1  # reach stopping target smoothly
-      ret.startingBrakeRate = 2.0  # release brakes fast
+      ret.stoppingDecelRate = 0.4  # reach stopping target smoothly
+      ret.startingAccelRate = 8.0  # release brakes fast
```
pd0wm, line 328: Why a factor of 4 here? Shouldn't it be 3 for toyota since that's the `ACCEL_SCALE`?

selfdrive/car/toyota/interface.py
```diff
@@ -363,7 +356,7 @@ def update(self, c, can_strings):
       events.add(EventName.lowSpeedLockout)
     if ret.vEgo < self.CP.minEnableSpeed and self.CP.openpilotLongitudinalControl:
       events.add(EventName.belowEngageSpeed)
-      if c.actuators.gas > 0.1:
+      if c.actuators.accel > 0.4:
```
pd0wm, line 359: See above about factor of 3.

selfdrive/car/toyota/interface.py
HaraldSchafer: Yes, good catch

selfdrive/car/toyota/interface.py
HaraldSchafer: Yes

selfdrive/car/interfaces.py
HaraldSchafer: Yes, typo

selfdrive/car/honda/carcontroller.py
HaraldSchafer: Hmmmm, this was messing up the interpolation a few lines down at 0 speed. I thought this was causing cruise faults, but in retrospect that was likely something else. I'll make a different PR for this.

Issue #19613, opened 2020-12-29
Security hardening


capstarall:
**Local Hardening**
Sure, “it’s just a devkit” for now, understood. But it would be wise to get a headstart on hardening the system so that when it’s driving 10k+ cars, it’s got to have a reliable and tested security model.

While by default the remote / network attack surface is quite limited (kudos for that), here’s a few suggestions to make things better yet still functional, primarily for the local environment, in order to limit the blast radius and make practical attacks non-trivial for anyone who manages to get some sort of internet or wifi rw/x access on the device (via comma’s services or otherwise).

**/dev/shm permissions**
777 is a little too open. Would 700 or 770 still let the necessary services read and write without disruption (some testing was done here, but more needed), while maintaining some sense of defense-in-depth?

Eg. if some random service did get compromised on the device, poking at stuff in /dev/shm seems like one of the things we’d only want comma components / root to do.

**Processes should run as non-root**
So that if there are ever bugs in these processes, they don’t automatically lead to a root shell. Seems like adding/reusing a local sandboxed user and doing a little work to the local services to make sure everybody is still communicating ok would be the easiest approach.

**SSH key should not be public**
Picking the SSH key off of the openpilot github is great for easy mode, but it probably should be randomly generated either 1) per device or 2) per openpilot install so you’re less scared of connecting this thing to the wifis.

**Installation warning on URLs other than openpilot official**
Measures should be taken to warn users in case they typo or are otherwise led to use a URL that looks or sounds like the official openpilot one (we all know the comma two is meant to use openpilot.comma.ai, it can still be non-obvious just with an addition confirmation from the other if they want to really install something else).

One typo or bad URL shouldn’t be a straight shot to running a backdoored image.

Anyways, hope this is helpful! Would be great to get to know if some of these were already on the list of things to do, or if they would even be items that have a shot on the current roadmap.

Really love your product. Keep up the chill work!

abdulrahman004:
> Installation warning on URLs other than openpilot official

We already get a warning of Warning: Untested branch for any non-official branch.

pd0wm:
We mainly focus our efforts on preventing code execution on the device. If you get code execution all bets are probably off, unless you're going to spend absurd amount of time sandboxing everything. That time is better spent on the former.

My thoughts on your questions:

[...]

PR #21922, opened 2021-08-13
Less open file permissions

pd0wm:
Last part of https://github.com/commaai/openpilot/issues/19613

- [x] Test clean install
- [x] Test C2
- [x] Test C3


deanlee:
It seems that umask(0) is not needed anymore, do I understand it right?

https://github.com/commaai/openpilot/blob/0ea90b41b51b63bbea9e85f360c05aa7269cf3ad/selfdrive/loggerd/logger.cc#L148

PR #21923, opened 2021-08-13
Smarter thermal controls for tici

### LINKED ISSUE #21917: Improved C3 fan policy

robbederks:


selfdrive/thermald/thermald.py
```diff
@@ -216,9 +240,12 @@ def thermald_thread():
       if handle_fan is None and pandaState.pandaState.pandaType != log.PandaState.PandaType.unknown:
         is_uno = pandaState.pandaState.pandaType == log.PandaState.PandaType.uno
 
-        if (not EON) or is_uno:
+        if (not EON and not TICI) or is_uno:
           cloudlog.info("Setting up UNO fan handler")
           handle_fan = handle_fan_uno
+        elif TICI:
+          cloudlog.info("Setting up TICI fan handler")
+          handle_fan = handle_fan_tici
```
adeebshihadeh, line 248:
```suggestion
        if TICI:
          cloudlog.info("Setting up TICI fan handler")
          handle_fan = handle_fan_tici
        elif is_uno:
          cloudlog.info("Setting up UNO fan handler")
          handle_fan = handle_fan_uno
```

selfdrive/thermald/thermald.py
```diff
@@ -121,6 +121,30 @@ def handle_fan_uno(max_cpu_temp, bat_temp, fan_speed, ignition):
...
+def handle_fan_tici(max_cpu_temp, bat_temp, fan_speed, ignition):
+  global prev_time, error_integral, prev_cpu_temp_error
+
+  CPU_TEMP_SETPOINT = 80 if ignition else 60
+  FAN_PWR_RANGE = (0, 80 if ignition else 30)
+  KI = 1e-5
+
+  if prev_time is None:
+    prev_time = time.monotonic()
+  dt = max(time.monotonic() - prev_time, 0.01)
```
adeebshihadeh, line 135: does this work? looks like prev_time is only set once

robbederks: lol, that's why it gets unstable after long drives, good catch!

selfdrive/thermald/thermald.py
```diff
@@ -104,16 +103,12 @@ def handle_fan_eon(max_cpu_temp, bat_temp, fan_speed, ignition):
     # update speed if using the low thresholds results in fan speed decrement
     fan_speed = new_speed_l
 
-  if bat_temp < _BAT_TEMP_THRESHOLD:
```
adeebshihadeh, line 107: are you sure we can remove this? we still support the EON gold

selfdrive/controls/lib/pid.py
```diff
@@ -1,3 +1,4 @@
+from numbers import Number
 import numpy as np
 from common.numpy_fast import clip, interp
```
adeebshihadeh, line 3:
```suggestion
import numpy as np
from numbers import Number

from common.numpy_fast import clip, interp
```

selfdrive/controls/lib/pid.py
```diff
@@ -29,10 +31,14 @@ def __init__(self, k_p, k_i, k_f=1., pos_limit=None, neg_limit=None, rate=100, s
 
   @property
   def k_p(self):
+    if isinstance(self._k_p, Number):
+      return self._k_p
```
adeebshihadeh, line 35: we shouldn't handle two types here. why not just pass in a list?

robbederks: Just a list doesn't work, needs to be a list of lists containing x,y pairs for interpolation based on speed (which doesn't make much sense outside of a vehicle controls context). Had it this way first, but looks pretty ugly

selfdrive/thermald/thermald.py
robbederks: this is fine, worst case the EON fans might spin a little faster

PR #21927, opened 2021-08-14
sidebar: simplify updateState with QPair

deanlee:
`updateState` is always called before `paintevent`, so there is no need to initializing status in header

### DIFF
--- a/selfdrive/ui/qt/sidebar.h
+++ b/selfdrive/ui/qt/sidebar.h
@@ -6,14 +6,13 @@
 #include "selfdrive/common/params.h"
 #include "selfdrive/ui/ui.h"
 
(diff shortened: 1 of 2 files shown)


selfdrive/ui/qt/sidebar.h
```diff
@@ -6,14 +6,13 @@
 #include "selfdrive/common/params.h"
 #include "selfdrive/ui/ui.h"
 
+typedef QPair<QString, QColor> ItemStatus;
+Q_DECLARE_METATYPE(ItemStatus);
```
adeebshihadeh, line 10:
```suggestion
Q_DECLARE_METATYPE(ItemStatus);

```

PR #21929, opened 2021-08-14
swaglog: c++ LogState

deanlee:
1. move related C funtions into class LogState
2. remove variable `inited` and  function `cloudlog_init`. init `LogState` in static function `logInstance`
3. protect log_j.dump() with mutex. (ctx_j_ may be changed by `bind()`）


adeebshihadeh:
Can we add these tests without the refactor?

PR #21932, opened 2021-08-14
Tesla radar interface

robbederks:
Implemented a radar interface for Tesla Model S

### DIFF
--- a/selfdrive/car/tesla/radar_interface.py
+++ b/selfdrive/car/tesla/radar_interface.py
@@ -67,7 +67,7 @@ class RadarInterface(RadarInterfaceBase):
     self.updated_messages.update(values)
 
     # Trigger update only on new first track message
-    if RADAR_MSGS_A[0] not in self.updated_messages:
+    if RADAR_MSGS_B[-1] not in self.updated_messages:
       return None
 
     ret = car.RadarData.new_message()
@@ -110,4 +110,5 @@ class RadarInterface(RadarInterfaceBase):
       self.pts[i].measured = bool(msg_a['Meas'])
 
     ret.points = list(self.pts.values())
+    self.updated_messages.clear()
     return ret
--- a/selfdrive/car/tesla/radar_interface.py
+++ b/selfdrive/car/tesla/radar_interface.py
@@ -1,5 +1,113 @@
 #!/usr/bin/env python3
+from cereal import car
+from opendbc.can.parser import CANParser
+from selfdrive.car.tesla.values import DBC, CANBUS
 from selfdrive.car.interfaces import RadarInterfaceBase
 
+RADAR_MSGS_A = list(range(0x371, 0x37E, 3))
+RADAR_MSGS_B = list(range(0x372, 0x37F, 3))
+NUM_POINTS = len(RADAR_MSGS_A)
+SGU_INDEX_VALUES = [f"RADC_ACCTargObj{i+1}_sguIndex" for i in range(NUM_POINTS)]
+SGU_INDEX_NONE = 63
+
+def get_radar_can_parser(CP):
+  # Status messages
+  signals = [
+    ('RADC_HWFail', 'TeslaRadarSguInfo', 0),
+    ('RADC_SGUFail', 'TeslaRadarSguInfo', 0),
+    ('RADC_SensorDirty', 'TeslaRadarSguInfo', 0),
+  ]
+
+  checks = [
+    ('TeslaRadarSguInfo', 10),
+    ('TeslaRadarTguInfo', 10),
+  ]
+
+  # Radar tracks. There are also raw point clouds available,
+  # we don't use those.
+  for i in range(NUM_POINTS):
+    msg_id_a = RADAR_MSGS_A[i]
+    msg_id_b = RADAR_MSGS_B[i]
+
+    # There is a bunch more info in the messages,
+    # but these are the only things actually used in openpilot
+    signals.extend([
+      ('LongDist', msg_id_a, 255),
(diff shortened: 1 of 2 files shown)


robbederks:
Jenkins failure doesn't seem to be related to this PR

selfdrive/car/tesla/radar_interface.py
```diff
@@ -1,5 +1,113 @@
...
+      errors.append('fault')
+    ret.errors = errors
+
+    # Radar tracks
+    for i in range(NUM_POINTS):
+      msg_a = self.rcp.vl[RADAR_MSGS_A[i]]
+      msg_b = self.rcp.vl[RADAR_MSGS_B[i]]
+
+      # Make sure msg A and B are together
+      if msg_a['Index'] != msg_b['Index2']:
```
pd0wm, line 90: When does this happen?

selfdrive/car/tesla/radar_interface.py
```diff
@@ -1,5 +1,113 @@
...
+
+  def update(self, can_strings):
+    if self.rcp is None:
+      return super().update(None)
+
+    values = self.rcp.update_strings(can_strings)
+    self.updated_messages.update(values)
+
+    # Trigger update only on new first track message
+    if RADAR_MSGS_A[0] not in self.updated_messages:
```
pd0wm, line 70: Why not on the last message? Maybe use the same structure as other radar interfaces with `self.trigger_message`

selfdrive/car/tesla/radar_interface.py
robbederks: this happens when one of the messages has come in already while the previous one hasn't yet. Checking this makes sure the A and B messages are in sync

selfdrive/car/tesla/radar_interface.py
robbederks: didn't really see the benefit of using `self.trigger_message`, since it's just another way to specify the last message (most often hard-coded in the other car ports, doesn't seem very clean).
I'll change it to `RADAR_MSGS_B[-1]`

robbederks: not sure if they always get sent out in sequence, but it does make more intuitive sense to use the last one instead of the first one I guess

robbederks: just noticed that this trigger doesn't actually do anything (or at least not what I imagined it did), since `self.updated_messages` is not cleared, so this is always true

robbederks:
> Jenkins failure doesn't seem to be related to this PR

It might have been after all, since I wasn't waiting on a new trigger message, hence the increased CPU usage

selfdrive/car/tesla/radar_interface.py
pd0wm: Isn't the trigger message field used by process replay?

robbederks:
Switched from using the radar tracks the radar computes to outputting the raw data points (up to 32). Should be good to merge now!

robbederks:
The radar also outputs the variance for all its measurements. We don't use this now, but cool to add in the future maybe? Data per point:
```
BO_ 877 M_36Dhex: 8 Radar
 SG_ LongDist : 0|12@1+ (0.0625,0) [0|255.9] "meters"  FrontCamera
 SG_ LongSpeed : 12|12@1+ (0.0625,-128) [-128|128] "meters/sec"  FrontCamera
 SG_ LatDist : 24|11@1+ (0.125,-128) [-128|128] "meters"  FrontCamera
 SG_ ProbExist : 35|5@1+ (3.125,0) [0|96.875] "%"  FrontCamera
 SG_ LongAccel : 40|10@1+ (0.03125,-16) [-16|16] "meters/sec/sec"  FrontCamera
 SG_ ProbObstacle : 50|5@1+ (3.125,0) [0|96.875] "%"  FrontCamera
 SG_ Valid : 55|1@1+ (1,0) [0|1] ""  FrontCamera
 SG_ ProbNonObstacle : 56|5@1+ (3.125,0) [0|96.875] "%"  FrontCamera
 SG_ Meas : 61|1@1+ (1,0) [0|1] ""  FrontCamera
 SG_ Tracked : 62|1@1+ (1,0) [0|1] ""  FrontCamera
 SG_ Index : 63|1@1+ (1,0) [0|1] ""  FrontCamera

BO_ 878 M_36Dhex2: 8 Radar
 SG_ LatSpeed : 0|10@1+ (0.125,-64) [-64|64] "meters/sec"  FrontCamera
 SG_ Length : 10|6@1+ (0.125,0) [0|7.875] "m"  FrontCamera
 SG_ dZ : 16|6@1+ (0.25,-5) [-5|10.75] "m"  FrontCamera
 SG_ MovingState : 22|2@1+ (1,0) [0|3] ""  FrontCamera
 SG_ dxSigma : 24|6@1+ (1,0) [0|63] ""  FrontCamera
 SG_ vxSigma : 30|6@1+ (1,0) [0|63] ""  FrontCamera
 SG_ axSigma : 36|6@1+ (1,0) [0|63] ""  FrontCamera
 SG_ dySigma : 42|6@1+ (1,0) [0|63] ""  FrontCamera
 SG_ ProbClass : 48|5@1+ (3.125,0) [0|96.875] "%"  FrontCamera
 SG_ Class : 53|3@1+ (1,0) [0|7] ""  FrontCamera
 SG_ dxRearEndLoss : 56|6@1+ (1,0) [0|63] ""  FrontCamera
 SG_ NotUsed : 62|1@1+ (1,0) [0|1] ""  FrontCamera
 SG_ Index2 : 63|1@1+ (1,0) [0|1] ""  FrontCamera
```

PR #21933, opened 2021-08-15
Update MISRA link in SAFETY.md

davetapley:
Link is now a 404.
Was only a blog post anyway:
https://web.archive.org/web/20201028092101/https://www.misra.org.uk/MISRAHome/MISRAC2012/tabid/196/Default.aspx


doktorsleepelss:
Might as well just link https://www.misra.org.uk/misra-c/ instead of the old deleted page.  Or the wikipedia article.

PR #21936, opened 2021-08-15
camerad: use the YUV buffer to create the jpeg thumbnail

deanlee:
1.5~2x faster than current rgb->jpeg.

### DIFF
--- a/selfdrive/camerad/cameras/camera_common.cc
+++ b/selfdrive/camerad/cameras/camera_common.cc
@@ -250,11 +250,11 @@ static std::pair<uint8_t *, size_t> yuv420_to_jpeg(const CameraBuf *b, int thumb
   // cinfo.dct_method = JDCT_IFAST;
   cinfo.raw_data_in = TRUE;
   // configure sampling factors for yuv420.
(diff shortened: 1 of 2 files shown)


jmayes41 (approved):

jmayes41 (approved):

selfdrive/camerad/cameras/camera_common.cc
```diff
@@ -216,66 +216,83 @@ kj::Array<uint8_t> get_frame_image(const CameraBuf *b) {
   return kj::mv(frame_image);
 }
 
-static void publish_thumbnail(PubMaster *pm, const CameraBuf *b) {
-  uint8_t* thumbnail_buffer = NULL;
-  unsigned long thumbnail_len = 0;
-
-  unsigned char *row = (unsigned char *)malloc(b->rgb_width/4*3);
+static std::pair<uint8_t *, size_t> yuv420_to_jpeg(const CameraBuf *b, int thumbnail_width, int thumbnail_height) {
```
adeebshihadeh, line 219: why not use an std::vector so we can avoid the std::pair?

PR #21937, opened 2021-08-15
update hyundai palisade steer ratio

gregjhogan:
short-term improvement until https://github.com/commaai/openpilot/issues/19672 is resolved


pd0wm:
Where did you get 15.6? 13.75 seems very low.

gregjhogan:
> Where did you get 15.6? 13.75 seems very low.

There is a link in https://github.com/commaai/openpilot/issues/19672

PR #21939, opened 2021-08-16
UI: indicate why buttons are disabled while onroad

nelsonjchen:
The abbreviated text and non-upper casing is because the button sizes
seem to be hardcoded.

Not sure if the code quality is up to snuff, but this was what I could
hack in and get working.

I just got really tired answering these kinds of Discord questions:

https://discord.com/channels/469524606043160576/728680686101463062/876213799198294046

https://discord.com/channels/469524606043160576/538741329799413760/847980306766233671

https://discord.com/channels/469524606043160576/524327905937850394/841496197025628220

https://discord.com/channels/469524606043160576/524592892627517450/832061398569189427

https://discord.com/channels/469524606043160576/524592892627517450/835642789890359318

https://discord.com/channels/469524606043160576/524592892627517450/842170990280835173

Offroad is still the same:

PXL_20210815_233007605

Onroad is not, and the button is still disabled:

PXL_20210815_232842425 MP


nelsonjchen:
I can do a similar one for the "Device" buttons if this passes muster.

adeebshihadeh:
I agree this is confusing, but I can already see the "what is veh.?" questions. We'll think about a better way to handle this as this also applies to most of the toggles.

nelsonjchen:
> I agree this is confusing, but I can already see the "what is veh.?" questions. We'll think about a better way to handle this as this also applies to most of the toggles.

I'm not too attached to the wording. "Turn off car" still fits fine and was an earlier string (it's even in the branch name). It's just a cosmetic issue for non-cars.  What do you think about "Turn off car" for the interim? Even if it's not final, maybe some usage data and anecdotes could be gleaned from this non-perfect solution.

jyoung8607:
Fully agreed about the user confusion, I've had three or four people ask me why the Uninstall button doesn't work, and it turns out their car is running. That said...

> I'm not too attached to the wording. "Turn off car" still fits fine and was an earlier string (it's even in the branch name). It's just a cosmetic issue for non-cars. What do you think about "Turn off car" for the interim? Even if it's not final, maybe some usage data and anecdotes could be gleaned from this non-perfect solution for a future more proper revision.

It would probably be best to pop up a dialog with more room for a clear explanation, the same dialog code that pops up for reboot or recalibration. You could do that in addition to graying out the button exactly as it's done today. Remember that we need room to explain the situation clearly not just in English, but in every language the localized UI will support in the future, and we may not be able to find sufficiently clear and compact phrases for all those languages.

nelsonjchen:
It would be very weird and poor discoverability to have a dialog behind a "disabled" button. 

Yeah, and you're right, a cursory look at some translations of "Turn off car" shows not all of them would fit into the button. 🤔

adeebshihadeh:
I'm going to close this as it'll just sit around for a bit. I talked with @alexcomma, and we don't want to merge this as is. However, we'll keep this in mind as new design is done.

nelsonjchen:
Sounds good! Looking forward to the new design.

PR #21947, opened 2021-08-17
enclose macro arguments in parentheses

deanlee:


adeebshihadeh:
Can we also do this for the other macros like CL_CHECK?

PR #21948, opened 2021-08-17
Handle EINTR for syscalls that can return it

deanlee:


pd0wm:
Is it possible to set the `SA_RESTART` flag when registering the `SIGUSR2` handlers so we don't need to deal with all of this?

deanlee:
I’m not sure, I would recommend  not to make any assumptions and always writing code to handle EINTR.

PR #21953, opened 2021-08-18
add TSS2 replay route

adeebshihadeh:


selfdrive/test/process_replay/process_replay.py
```diff
@@ -10,6 +10,8 @@
 import capnp
 from tqdm import tqdm
 
+os.environ["SIMULATION"] = "1"
```
pd0wm, line 13: I don't think we want to set this on import time. Would probably cause all kind of unexpected stuff to happen.

PR #21959, opened 2021-08-18
Thermald thermal state cleanup

robbederks:


selfdrive/thermald/thermald.py
```diff
@@ -294,35 +307,25 @@ def thermald_thread():
 
     current_filter.update(msg.deviceState.batteryCurrent / 1e6)
 
-    # TODO: add car battery voltage check
-    max_cpu_temp = cpu_temp_filter.update(max(msg.deviceState.cpuTempC))
-    max_comp_temp = max(max_cpu_temp, msg.deviceState.memoryTempC, max(msg.deviceState.gpuTempC))
+    max_comp_temp = temp_filter.update(
```
pd0wm, line 310: Make sure the change from max_cpu_temp to using max_comp_temp everywhere is safe

PR #21961, opened 2021-08-18
boardd:   wait for safety_setter_thread to finish while quitting panda_state_thread

deanlee:
use std::future to wait for `safety_setter_thread` to finish before quitting.

`safety_setter_thread` should not run after `panda_state_thread`,  or the `panda ` could be set to nullptr

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -36,6 +36,7 @@
 #define CUTOFF_IL 200
 #define SATURATE_IL 1600
 #define NIBBLE_TO_HEX(n) ((n) < 10 ? (n) + '0' : ((n) - 10) + 'a')
+using namespace std::chrono_literals;
 
 std::atomic<bool> ignition(false);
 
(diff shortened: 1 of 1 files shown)


pd0wm:
Any reason the safety setter thread needs to be global like this? Can we make the `panda_state_thread` "own" the thread and join it on disconnect?

deanlee:
> Any reason the safety setter thread needs to be global like this? Can we make the `panda_state_thread` "own" the thread and join it on disconnect?

`std::future<bool> safety_future` is used for this purpose.it's owned by panda_state_thread and will "join" the `safety_setter_thread` in destructor before the `panda_state_thread` return.

pd0wm:
Thanks for rebasing!

pd0wm:
While testing I noticed the thread never finishes if you go offroad before the fingerprinting finishes. The panda state threat then puts you back in noOutput and you're stuck. The safety setter thread should also exit when ignition goes low.

pd0wm:
https://github.com/commaai/openpilot/pull/22426

deanlee:
Just realized this is my 500th PR:)

PR #21963, opened 2021-08-18
boardd: atomic ignition status

deanlee:


adeebshihadeh:
I don't think this is cleaner. Checking `ignition` is more clear than `ignition_status >= IgnitionStatus::starting`.

PR #21967, opened 2021-08-19
base LatControl class

### LINKED ISSUE #22419: Create base class for low level lateral control classes (PID, INDI, LQR)

sshane:
Removes duplicate code for saturation checking, moves sat check out of pid.py since only lat is using it. Can be expanded in the future for other common code between controllers

We can also do the sat check for the angle inside itself, but we need lat_plan passed as well


selfdrive/controls/lib/latcontrol_pid.py
```diff
@@ -2,18 +2,20 @@
...
 from selfdrive.controls.lib.pid import PIController
 from selfdrive.controls.lib.drive_helpers import get_steer_max
+from selfdrive.controls.lib.latcontrol import LatControl, MIN_STEER_SPEED
 from cereal import log
 
 
-class LatControlPID():
+class LatControlPID(LatControl):
   def __init__(self, CP, CI):
+    super().__init__(CP)
```
sshane, line 11: We can also take in `CI` for the other two controllers in controlsd, so you don't have the mismatching arguments

sshane:
How does that look? That way we don't need to duplicate the code that manages the sat_count. The only difference is that now with angle sat checking, it checks if steering rate is not limited, and only above 10 m/s. Is that okay?

selfdrive/controls/lib/latcontrol_angle.py
```diff
@@ -20,11 +20,6 @@ def update(self, active, CS, CP, VM, params, desired_curvature, desired_curvatur
       angle_steers_des += params.angleOffsetDeg
 
     angle_control_saturated = abs(angle_steers_des - CS.steeringAngleDeg) > STEER_ANGLE_SATURATION_THRESHOLD
-    if angle_control_saturated and not CS.steeringPressed and active:
-      self.sat_count += 1
-    else:
-      self.sat_count = 0
-
-    angle_log.saturated = self.sat_count > STEER_ANGLE_SATURATION_TIMEOUT
```
pd0wm: Let's also remove the constant on line 6

pd0wm:
Looks good. Let me write some unit tests around these controllers and we should be good to merge.

pd0wm:
```
***** results for segment fakedata|2022-01-20--17-49-04--0 *****
	controlsd
***** results for segment fakedata|2022-01-20--17-50-51--0 *****
	controlsd
***** results for segment fakedata|2022-01-20--17-52-36--0 *****
	controlsd
***** results for segment fakedata|2022-01-20--17-54-50--0 *****
	controlsd
***** results for segment fakedata|2022-01-20--17-56-40--0 *****
	controlsd
***** results for segment fakedata|2022-01-20--17-58-25--0 *****
	controlsd
***** results for segment fakedata|2022-01-20--18-00-11--0 *****
	controlsd
***** results for segment fakedata|2022-01-20--18-01-57--0 *****
	controlsd
***** results for segment fakedata|2022-01-20--18-03-41--0 *****
	controlsd
		controlsState.lateralControlState.pidState.saturated: 1
***** results for segment fakedata|2022-01-20--18-05-29--0 *****
	controlsd
		carParams.steerLimitTimer: 2
***** results for segment fakedata|2022-01-20--18-07-15--0 *****
	controlsd
***** results for segment fakedata|2022-01-20--18-09-32--0 *****
	controlsd
```

PR #21821, opened 2021-08-01
Qt/onroad: use CameraViewWidget

deanlee:

PR #21969, opened 2021-08-19
ui: add Lead struct to encapsulate related data

deanlee:


adeebshihadeh:
This isn't particularity more readable than the existing code. If you want to do some UI cleanup though, let's the rest of the NVG to Qt. #21821 is a good start on this.

deanlee:
opened #21821  for review

PR #21974, opened 2021-08-19
Params::put:  fixed the wrong call to Params::remove instead of global ::remove

deanlee:


pd0wm:
I assume this was only relevant if the rename failed?

deanlee:
yes, but even if the rename is successful, this bug will cause extra filelock and fsync calls which are expansive.

selfdrive/common/params.cc
```diff
@@ -285,7 +285,7 @@ int Params::put(const char* key, const char* value, size_t value_size) {
   } while (false);
 
   close(tmp_fd);
-  remove(tmp_path.c_str());
+  ::remove(tmp_path.c_str());
```
adeebshihadeh, line 288: can't we just unlink?

PR #21975, opened 2021-08-19
params: remove persistent params

deanlee:


pd0wm:
@adeebshihadeh How do you feel about removing the persistent params code since it's unused? Or do you see a need for it in the future?

adeebshihadeh:
Yeah, let's remove persistent params. I don't think we'll need it in the future.

deanlee:
removed

PR #21989, opened 2021-08-20
add list() to panda.cc

briskspirit:
add list() to panda.cc, returns vector of usb serial numbers of connected pandas

### DIFF
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -28,7 +28,7 @@ Panda::Panda(std::string serial) {
       libusb_open(dev_list[i], &dev_handle);
       if (dev_handle == NULL) { goto fail; }
 
-      unsigned char desc_serial[25];
+      unsigned char desc_serial[26] = { 0 };
       int ret = libusb_get_string_descriptor_ascii(dev_handle, desc.iSerialNumber, desc_serial, std::size(desc_serial));
       if (ret < 0) { goto fail; }
 
@@ -125,7 +125,7 @@ std::vector<std::string> Panda::list() {
     if (desc.idVendor == 0xbbaa && desc.idProduct == 0xddcc) {
       libusb_device_handle *handle = NULL;
       libusb_open(device, &handle);
-      unsigned char desc_serial[25];
+      unsigned char desc_serial[26] = { 0 };
       int ret = libusb_get_string_descriptor_ascii(handle, desc.iSerialNumber, desc_serial, std::size(desc_serial));
       libusb_release_interface(handle, 0);
       libusb_close(handle);
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -32,7 +32,7 @@ Panda::Panda(std::string serial) {
       int ret = libusb_get_string_descriptor_ascii(dev_handle, desc.iSerialNumber, desc_serial, std::size(desc_serial));
       if (ret < 0) { goto fail; }
 
-      usb_serial = std::string(desc_serial, desc_serial+std::size(desc_serial)).c_str();
+      usb_serial = std::string((char *)desc_serial, ret).c_str();
       if (serial.empty() || serial == usb_serial) {
         break;
       }
@@ -131,7 +131,7 @@ std::vector<std::string> Panda::list() {
       libusb_close(handle);
 
       if (ret < 0) { goto finish; }
-      serials.push_back(std::string(desc_serial, desc_serial+std::size(desc_serial)).c_str());
+      serials.push_back(std::string((char *)desc_serial, ret).c_str());
     }
   }
 
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -28,7 +28,7 @@ Panda::Panda(std::string serial) {
       libusb_open(dev_list[i], &dev_handle);
       if (dev_handle == NULL) { goto fail; }
 
-      unsigned char desc_serial[25];
+      unsigned char desc_serial[26];
       int ret = libusb_get_string_descriptor_ascii(dev_handle, desc.iSerialNumber, desc_serial, std::size(desc_serial));
       if (ret < 0) { goto fail; }
 
@@ -124,7 +124,7 @@ std::vector<std::string> Panda::list() {
     if (desc.idVendor == 0xbbaa && desc.idProduct == 0xddcc) {
       libusb_device_handle *handle = NULL;
       libusb_open(device, &handle);
-      unsigned char serial[25];
+      unsigned char serial[26];
       int ret = libusb_get_string_descriptor_ascii(handle, desc.iSerialNumber, serial, std::size(serial));
       libusb_release_interface(handle, 0);
       libusb_close(handle);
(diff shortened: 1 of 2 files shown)


selfdrive/boardd/panda.cc
```diff
@@ -92,6 +92,57 @@ void Panda::cleanup() {
...
 }
 
+std::vector<std::string> Panda::list() {
+  // init libusb
+  ssize_t num_devices;
+  libusb_context *context = NULL;
+  libusb_device **dev_list = NULL;
+  std::vector<std::string> serials;
+  
+  int err = libusb_init(&context);
```
pd0wm, line 102: Refactor USB init functions into a helper functions instead of copy-pasting.

selfdrive/boardd/panda.cc
```diff
@@ -92,6 +92,57 @@ void Panda::cleanup() {
...
+    libusb_get_device_descriptor(device, &desc);
+    if (desc.idVendor == 0xbbaa && desc.idProduct == 0xddcc) {
+      libusb_device_handle *handle = NULL;
+      libusb_open(device, &handle);
+      unsigned char serial[25];
+      int ret = libusb_get_string_descriptor_ascii(handle, desc.iSerialNumber, serial, sizeof(serial));
+      libusb_release_interface(handle, 0);
+      libusb_close(handle);
+
+      if (ret < 0) { goto finish; }
```
pd0wm, line 131: No need for brackets if on the same line

selfdrive/boardd/panda.cc
```diff
@@ -92,6 +92,57 @@ void Panda::cleanup() {
...
+  }
+  for (size_t i = 0; i < num_devices; ++i) {
+    libusb_device *device = dev_list[i];
+    libusb_device_descriptor desc;
+    libusb_get_device_descriptor(device, &desc);
+    if (desc.idVendor == 0xbbaa && desc.idProduct == 0xddcc) {
+      libusb_device_handle *handle = NULL;
+      libusb_open(device, &handle);
+      unsigned char serial[25];
+      int ret = libusb_get_string_descriptor_ascii(handle, desc.iSerialNumber, serial, sizeof(serial));
```
pd0wm, line 127:
```suggestion
      int ret = libusb_get_string_descriptor_ascii(handle, desc.iSerialNumber, serial, std::size(serial));
```

selfdrive/boardd/panda.cc
```diff
@@ -92,6 +92,57 @@ void Panda::cleanup() {
...
+      libusb_release_interface(handle, 0);
+      libusb_close(handle);
+
+      if (ret < 0) { goto finish; }
+      serials.push_back((char *)serial);
```
pd0wm, line 132: Is the serial buffer guaranteed to be null terminated (if the descriptor is 25 characters)? Maybe safer to use `std::string(serial, ret)`

selfdrive/boardd/panda.cc
briskspirit: It will stop only on error, when return value of libusb_get_string_descriptor_ascii is less than 0.
Also why would it encounter descriptor with 25+ characters? We are checking only for our devices and this descriptor is filled out by us in the panda code (usb.h)

selfdrive/boardd/panda.cc
briskspirit: std::string(serial, ret) gives an error on compile

selfdrive/boardd/panda.cc
briskspirit: I can just skip this "faulty" device and continue to check other devices in the list

briskspirit: Also about brackets - I know, but boardd.cc is full of those brackets and I was told that they should be removed in another PR, so I keep style consistent for now.

selfdrive/boardd/panda.cc
briskspirit: Don't get how to use it, returns empty string.

briskspirit: Never mind, got it after reading docs few times more )) Check please.

selfdrive/boardd/panda.cc
```diff
@@ -124,7 +124,7 @@ std::vector<std::string> Panda::list() {
     if (desc.idVendor == 0xbbaa && desc.idProduct == 0xddcc) {
       libusb_device_handle *handle = NULL;
       libusb_open(device, &handle);
-      unsigned char serial[25];
+      unsigned char serial[26];
```
pd0wm: You're not initializing the array to zero, so still no null byte at the end.

pd0wm: But since you're using the length when converting to std::string it shouldn't be necessary to have a null byte.

selfdrive/boardd/panda.cc
```diff
@@ -92,6 +87,64 @@ void Panda::cleanup() {
...
+    if (desc.idVendor == 0xbbaa && desc.idProduct == 0xddcc) {
+      libusb_device_handle *handle = NULL;
+      libusb_open(device, &handle);
+      unsigned char desc_serial[25];
+      int ret = libusb_get_string_descriptor_ascii(handle, desc.iSerialNumber, desc_serial, std::size(desc_serial));
+      libusb_release_interface(handle, 0);
+      libusb_close(handle);
+
+      if (ret < 0) { goto finish; }
+      serials.push_back(std::string(desc_serial, desc_serial+std::size(desc_serial)).c_str());
```
pd0wm, line 134:
```suggestion
      serials.push_back(std::string(desc_serial, ret));
```

selfdrive/boardd/panda.cc
```diff
@@ -35,10 +29,11 @@ Panda::Panda(std::string serial) {
       if (dev_handle == NULL) { goto fail; }
 
       unsigned char desc_serial[25];
-      int ret = libusb_get_string_descriptor_ascii(dev_handle, desc.iSerialNumber, desc_serial, sizeof(desc_serial));
+      int ret = libusb_get_string_descriptor_ascii(dev_handle, desc.iSerialNumber, desc_serial, std::size(desc_serial));
       if (ret < 0) { goto fail; }
 
-      if (serial.empty() || serial.compare(reinterpret_cast<const char*>(desc_serial)) == 0) {
+      usb_serial = std::string(desc_serial, desc_serial+std::size(desc_serial)).c_str();
```
pd0wm, line 35:
```suggestion
      usb_serial = std::string(desc_serial, ret);
```

selfdrive/boardd/panda.cc
briskspirit: This is not gonna work.
error: no matching constructor for initialization of 'std::string' (aka 'basic_string<char>')
      serials.push_back(std::string(desc_serial, ret));

selfdrive/boardd/panda.cc
briskspirit: Same as above, that's what I tried first.

selfdrive/boardd/panda.cc
briskspirit: I suppose because I am using unsigned char. Method I am using now is recommended in this case.

pd0wm: Now you still rely on the null byte instead of the length returned by the libusb function. Just cast it to (char*) if needed.

PR #21995, opened 2021-08-20
Subaru: universal signal for ACC speed units

martinl:
**Description**

This PR updates Subaru Global and Preglobal ACC speed units signals in carstate. Both previously used signals were unreliable and signal values were different across markets/models.

Prerequsite: https://github.com/commaai/opendbc/pull/414

**Verification** 

Both signals have been in use in subaru-community branch since beginning of July and ACC speed has been correct for all tested models


adeebshihadeh:
This looks like a nice cleanup, but there may also be a cruise speed signal that doesn't change units. Let's get this merged (left a comment on the DBC PR), but after that, we may be able to find a new speed signal that would make this obsolete.

adeebshihadeh (changes requested):
This looks like a nice cleanup, but there may also be a cruise speed signal that doesn't change units. Let's get this merged (left a comment on the DBC PR), but after that, we may be able to find a new speed signal that would make this obsolete.

martinl:
I updated preglobal to use Dash_State2 UNITS signal to match opendbc PR

PR #21999, opened 2021-08-20
Use an LPF in longitudinal_planner

ghost:
Replace custom math with the FirstOrderFilter class


grekiki:
Can add more decimals if process replay needs to pass

adeebshihadeh:
I don't think using the `FirstOrderFilter` class make this cleaner or more readable unless you add some new helper functions, so that `x` doesn't need to be set outside the class.

PR #22003, opened 2021-08-21
car port: Honda Freed

belm0:
Popular 3-row compact minivan in Japan based on Fit platform.

**Checklist**
- [x] added to README
- [x] route with openpilot:  2c4292a5cd10536c|2021-08-19--21-32-15
- [x] https://github.com/commaai/opendbc/pull/419 dependency needs to be merged
- [x] follow braking seems insufficient with the recent Honda nidec long tuning.  Confirm that recent fixes for ODYSSEY help here.


.gitmodules
```diff
@@ -3,7 +3,7 @@
 	url = ../../commaai/panda.git
 [submodule "opendbc"]
 	path = opendbc
-	url = ../../commaai/opendbc.git
+	url = git@github.com:belm0/opendbc.git
```
belm0, line 6: temporarily point to pending PR https://github.com/commaai/opendbc/pull/413

belm0: I can't explain this test failure.    `honda_hrv_touring_2019_can_generated` is the same after my opendbc change, only with a few items shifted around.
```
...................E........................................................CANParser: could not find message 0x0 in DBC honda_hrv_touring_2019_can_generated
python: opendbc/can/parser.cc:140: CANParser::CANParser(int, const std::string &, const std::vector<MessageParseOptions> &, const std::vector<SignalParseOptions> &): Assertion `false' failed.
```

adeebshihadeh: it needs to be added to release/files_common

belm0: thank you, I've added the freed dbc to `files_common`

but I don't think it explains the error regarding the existing honda_hrv_touring_2019_can_generated file

```
CANParser: could not find message 0x0 in DBC honda_hrv_touring_2019_can_generated
```

belm0: error in HRV test was due to a bug in my carstate refactoring-- fixed

selfdrive/car/honda/carstate.py
```diff
@@ -238,6 +232,9 @@ def update(self, cp, cp_cam, cp_body):
       ret.doorOpen = bool(cp.vl["SCM_BUTTONS"]["DRIVERS_DOOR_OPEN"])
     elif self.CP.carFingerprint == CAR.HRV:
       ret.doorOpen = bool(cp.vl["SCM_BUTTONS"]["DRIVERS_DOOR_OPEN"])
+    elif self.CP.carFingerprint == CAR.FREED:
+      ret.standstill = not cp.vl["STANDSTILL"]["WHEELS_MOVING"]
+      ret.doorOpen = bool(cp.vl["SCM_BUTTONS"]["DRIVERS_DOOR_OPEN"])
```
belm0, line 237: Why doesn't HRV set `ret.standstill`?  I think it's a bug.

belm0: I'll make HRV and FREED the same case here unless there is objection.

selfdrive/car/honda/carstate.py
```diff
@@ -155,19 +155,13 @@ def get_can_signals(CP, gearbox_msg="GEARBOX"):
...
-                ("MAIN_ON", "SCM_BUTTONS", 0),
-                ("BRAKE_HOLD_ACTIVE", "VSA_STATUS", 0)]
-    checks += [
-      ("GAS_PEDAL_2", 100),
-    ]
-  elif CP.carFingerprint == CAR.HRV:
-    signals += [("CAR_GAS", "GAS_PEDAL", 0),
+  elif CP.carFingerprint in (CAR.FIT, CAR.FREED, CAR.HRV):
+    gas_pedal = "GAS_PEDAL" if CP.carFingerprint == CAR.HRV else "GAS_PEDAL_2"
+    signals += [("CAR_GAS", gas_pedal, 0),
```
belm0, line 160: By the way, on HRV, this `CAR_GAS` signal is unused by carstate, which seems like a bug.  I think it should be used in the following code (rather than grouping HRV with CRV):

```python
HRV has vl["GAS_PEDAL"]["CAR_GAS"] available, so seems like it should be used.

self.pedal_gas = cp.vl["POWERTRAIN_DATA"]["PEDAL_GAS"]
# crv doesn't include cruise control
if self.CP.carFingerprint in (CAR.CRV, CAR.CRV_EU, CAR.HRV, CAR.ODYSSEY, CAR.ACURA_RDX, CAR.RIDGELINE, CAR.PILOT_2019, CAR.ODYSSEY_CHN):
  ret.gas = self.pedal_gas / 256.
else:
  ret.gas = cp.vl["GAS_PEDAL_2"]["CAR_GAS"] / 256.
```

belm0:
I've taken up apparent HR-V bugs in a separate PR #22009

belm0:
I'll set this PR to draft pending the following:
  * https://github.com/commaai/opendbc/pull/419 dependency needs to be merged
  * follow braking seems insufficient with the recent Honda nidec long tuning.  Resolve the problem, or opt-in to legacy long control.

selfdrive/car/honda/interface.py
```diff
@@ -198,6 +198,17 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       tire_stiffness_factor = 0.75
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.2], [0.05]]
 
+    elif candidate == CAR.FREED:
```
adeebshihadeh, line 201: merge this with the fit parameters, i.e. `elif candidate in [CAR.FIT, CAR.FREED]`

selfdrive/car/honda/values.py
```diff
@@ -828,6 +829,25 @@ class CAR:
       b'77959-T5R-A230\x00\x00',
     ],
   },
+  CAR.FREED: {
+    # NOTE: omitting gateway ECU since it doesn't appear on all models
```
adeebshihadeh, line 833: no need to omit, as the gateway isn't an essential ecu

belm0: I see-- I misunderstood the docs on `match_fw_to_car_exact()`

selfdrive/car/honda/interface.py
belm0: Oh, actually I thought I had the proper mass and wheelbase here.  I'll correct those, and this entry will be different than FIT.  (Freed is a minivan, Fit is a compact.)

adeebshihadeh:
Is this ready for review or are we still blocked on something?

belm0:
> Is this ready for review or are we still blocked on something?

Thanks for the ping-- by chance I'll be trying the car again over the next day to confirm that the long planning is reasonable.

selfdrive/car/honda/values.py
```diff
@@ -838,6 +839,29 @@ class CAR:
...
+    ],
+    (Ecu.eps, 0x18da30f1, None): [
+      b'39990-TDK-J050\x00\x00',
+      b'39990-TDK-N020\x00\x00',
+    ],
+    # TODO: vsa is "essential" for fpv2 but doesn't appear on some models
+    (Ecu.vsa, 0x18da28f1, None): [
+      b'57114-TDK-J120\x00\x00',
+      b'57114-TDK-J330\x00\x00',
+    ],
```
belm0, line 854: are there other options for this vsa fingerprint issue than 1) add another CAR entry, 2) rely on fuzzy match?

belm0 (commented):
Took a highway drive with all the long changes on master-- it's passable.  No more speeding into slowing lead car.

selfdrive/car/honda/values.py
adeebshihadeh: we should look into why that is, but there's already precedent for exceptions like this https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py#L224

belm0: (is it correct that "vsa" is in the essential set?)

belm0: so perhaps this?

```python
      # vsa is missing on some Honda Freed sub-models.  (Does this ECU need to be essential?)
      if ecu_type == Ecu.vsa and candidate == HONDA.FREED and found_version is None:
        continue
```

adeebshihadeh: yeah, something like that, but I'd like to understand why this happens first

belm0: `VSA_STATUS` is active in the route, so this implies it's not coming from the vsa ECU

Screen Shot 2021-09-25 at 9 29 17 AM

2c4292a5cd10536c|2021-09-24--22-21-53

belm0: Oh never mind-- in this route the vsa actually appeared in carFw.

My first two engine starts didn't have vsa, then I removed the entry from the fingerprint, and it appeared in canFw on subsequent engine starts 🤷

SunDuke1:
Hello! Thanks to you, I bought a Honda Freed+ Crosstar 2020 (gasoline, 5 seats), as the only Japanese right-hand drive car for the domestic market, which I knew about at that time, compatible with Comma Openpilot))). I have been driving it for 9 months on right-hand traffic roads, I am very happy with the car, but not very happy with the driver assistance system, many restrictions. I plan to buy a Comma 3X at the next sale to expand the capabilities of autonomous driving. 
Since you are the only one who has successfully adapted Openpilot for Honda Freed. 
I have questions: 
1) Have you compiled a FAQ, README or any instructions for setting up and operating Comma Openpilot on Honda Freed? 
2) Is there a video instruction and a video of trips, in how the autopilot behaves in real use. I read your message that you drove a lot and most likely at different times of the day and at different speeds and weather; 
 3) You  are have tested Openpilot on a 7-seater Freed, possibly a hybrid or 4wd. How and where to change the weight in Openpilot, perhaps the clearance of the car? 
4) Are there any special settings for a right-hand drive car for right-hand traffic? 
5) There is no stop-and-go on the Honda Freed, but I want Openpilot to work at the entire speed range from 0 km/h to the maximum permitted speed. For this, Honda owners with Nidec are recommended to use the Comma Pedal device. There are installation videos on Honda Fit, Pilot, etc. Have you installed this add-on? Can you tell me how to set up Openpilot in this case? 
I have not found any other sources of knowledge, so thanks in advance for your answers

belm0:
@SunDuke1  About right-hand drive, it's no worry.  It affects driver monitoring only.  Comma 2 has a manual setting, and Comma 3 detects the driver seat automatically.  I drove the Freed with OpenPilot in Japan.

Weight and other detailed vehicle properties aren't too important.  The existing approximate values will be fine.

Installing Comma on Freed won't be different than other vehicles.  You have to be a little brave removing the plastic cover at the mirror.

I don't have experience with Comma Pedal.  It's true that stop-and-go is nice, but OpenPilot still has a good value on the Freed without it.

About Freed+ Crosstar 2020 specifically, you may have to accept some uncertainty.  Maybe the ECU fingerprint will have to be added, maybe some CAN register or usage is a little different.  You can only know by trying, and ask for help on Comma.ai Discord.   If it doesn't work, you can return the Comma 3, though you may have to pay international shipping.

PR #22008, opened 2021-08-22
Honda cleanup

adeebshihadeh:


selfdrive/car/honda/hondacan.py
```diff
@@ -181,7 +181,7 @@ def create_ui_commands(packer, pcm_speed, hud, car_fingerprint, is_metric, idx,
       'SET_TO_1' : 0x01,
```
Vaggysag, line 181: There is also an extra space in this line

PR #22009, opened 2021-08-22
Honda HR-V: set standstill and gas signals

belm0:
* CarState.standstill was not set
  * ~CarState.gas was not derived from GAS_PEDAL/CAR_GAS~

These bugs were inferred by reading the code, and may be benign.  I'm not aware of problem symptoms being reported.

I don't have an HR-V for testing.


Chinech:
Nicx

selfdrive/car/honda/carstate.py
```diff
@@ -237,6 +237,7 @@ def update(self, cp, cp_cam, cp_body):
       ret.standstill = cp.vl["ENGINE_DATA"]["XMISSION_SPEED"] < 0.1
       ret.doorOpen = bool(cp.vl["SCM_BUTTONS"]["DRIVERS_DOOR_OPEN"])
     elif self.CP.carFingerprint == CAR.HRV:
+      ret.standstill = not cp.vl["STANDSTILL"]["WHEELS_MOVING"]
```
belm0, line 240: It's probably best to double-check that HRV has STANDSTILL / WHEELS_MOVING signal.  Would the registered HRV test route be suitable?

adeebshihadeh: yes, you can use this to pull the rlogs from the test route https://github.com/commaai/openpilot/blob/master/selfdrive/test/openpilotci.py#L10

selfdrive/car/honda/carstate.py
```diff
@@ -295,8 +296,10 @@ def update(self, cp, cp_cam, cp_body):
 
     self.pedal_gas = cp.vl["POWERTRAIN_DATA"]["PEDAL_GAS"]
     # crv doesn't include cruise control
-    if self.CP.carFingerprint in (CAR.CRV, CAR.CRV_EU, CAR.HRV, CAR.ODYSSEY, CAR.ACURA_RDX, CAR.RIDGELINE, CAR.PILOT_2019, CAR.ODYSSEY_CHN):
+    if self.CP.carFingerprint in (CAR.CRV, CAR.CRV_EU, CAR.ODYSSEY, CAR.ACURA_RDX, CAR.RIDGELINE, CAR.PILOT_2019, CAR.ODYSSEY_CHN):
       ret.gas = self.pedal_gas / 256.
+    elif self.CP.carFingerprint == CAR.HRV:
+      ret.gas = cp.vl["GAS_PEDAL"]["CAR_GAS"] / 256.
```
adeebshihadeh, line 302: what's the difference between these two signals? better to use the common one if possible

adeebshihadeh: carState.gas is meant as a user gas signal, so the existing signal is correct

belm0: > carState.gas is meant as a user gas signal, so the existing signal is correct

@adeebshihadeh The following models are using `GAS_PEDAL_2/CAR_GAS` for carState.gas, which is _not_ the user gas signal: CIVIC, ACURA_ILX, FIT, PILOT

Is it because `POWERTRAIN_DATA/PEDAL_GAS` is not available on these models?  It seems unlikely.  Should we try to confirm that and move everything to the actual user gas signal?

```python
# crv doesn't include cruise control
if self.CP.carFingerprint in (CAR.CRV, CAR.CRV_EU, CAR.HRV, CAR.ODYSSEY, CAR.ACURA_RDX, CAR.RIDGELINE, CAR.PILOT_2019, CAR.ODYSSEY_CHN):
  ret.gas = self.pedal_gas / 256.
else:
  ret.gas = cp.vl["GAS_PEDAL_2"]["CAR_GAS"] / 256.
```
(I don't understand the comment about CRV in this code.)

adeebshihadeh: I'm not sure why the comment says that either. Moving the other cars over sounds good if the signal is there. You should be able to use the test routes to verify it like you did with the HRV.

belm0: All the models have `POWERTRAIN_DATA/PEDAL_GAS`, we can infer that because the code is using it for `gasPressed`:

```python
self.pedal_gas = cp.vl["POWERTRAIN_DATA"]["PEDAL_GAS"]
...
ret.gasPressed = self.pedal_gas > 1e-5
```

> The following models are using GAS_PEDAL_2/CAR_GAS for carState.gas, which is not the user gas signal: CIVIC, ACURA_ILX, FIT, PILOT

correction:  all of the Bosch cars are using `GAS_PEDAL_2/CAR_GAS` too...

> carState.gas is meant as a user gas signal

How certain are you about it?

adeebshihadeh: it may not always have been intended that way, but @pd0wm and I agreed that makes the most sense

adeebshihadeh: feel free to make a PR to cereal to change the name to userGas is that makes it more clear

belm0: looking at `car.capnp`, it explicitly states that `gas` is user + computer

```
  # gas pedal, 0.0-1.0
  gas @3 :Float32;        # this is user + computer
  gasPressed @4 :Bool;    # this is user pedal only
```

The user of the data could always gate `gas` with `gasPressed` to derive user gas.

Can you say a little about why userGas makes sense for this API?

https://github.com/commaai/cereal/blob/0c929f74ce2ca96ff95bd9c1b50b5840ccc4503d/car.capnp#L150-L152

adeebshihadeh: currently, carState.gas isn't used anywhere, and I can't imagine a case where we'd want to start using it as a user + computer signal

PR #22010, opened 2021-08-22
Honda carstate: use common gas pedal signal

belm0:
CarState.gas is now derived exclusively from user pedal signal.  Previously it was derived
from user pedal or user pedal + computer gas, depending on model.

Subsequently, this allowed several cases to be collapsed in the the signal config regarding SCM messages.

NOTE: Cereal currently defines CarState.gas as user + computer, but it's agreed that user pedal is sufficient going forward.


selfdrive/car/honda/carstate.py
```diff
@@ -207,6 +180,13 @@ def __init__(self, CP):
     self.gearbox_msg = "GEARBOX"
     if CP.carFingerprint == CAR.ACCORD and CP.transmissionType == TransmissionType.cvt:
       self.gearbox_msg = "GEARBOX_15T"
+    # message containing CAR_GAS, if any
+    self.gas_pedal_msg = "GAS_PEDAL_2"
+    if self.CP.carFingerprint in (CAR.CRV, CAR.CRV_EU, CAR.ODYSSEY, CAR.ACURA_RDX, CAR.RIDGELINE, CAR.PILOT, CAR.PILOT_2019, CAR.ODYSSEY_CHN):
+      # crv et al don't include cruise control
+      self.gas_pedal_msg = None
```
belm0, line 187: note: I added CAR.PILOT to this case

The original code was adding `("CAR_GAS", "GAS_PEDAL_2", 0)` to signals, but then had a comment on the `GAS_PEDAL_2` check that the message wasn't available on some Pilots.  If that did happen, it would mean CarState.gas would always be 0, which sounds problematic.  So it seems better to put PILOT entirely in the "GAS_PEDAL not available" group, so that `gas` will be derived from `PEDAL_GAS`.

belm0:
setting status to draft, may want to resolve https://github.com/commaai/openpilot/pull/22009#discussion_r701549992 first

adeebshihadeh:
Let's use the common user gas signal as discussed in #22009.

selfdrive/car/honda/carstate.py
belm0: now irrelevant since all models are using PEDAL_GAS

selfdrive/car/honda/carstate.py
```diff
@@ -10,7 +10,7 @@
 TransmissionType = car.CarParams.TransmissionType
 
 
-def get_can_signals(CP, gearbox_msg="GEARBOX"):
+def get_can_signals(CP, *, gearbox_msg):
```
adeebshihadeh, line 13: why change this?

adeebshihadeh:
I updated the comment in cereal (https://github.com/commaai/cereal/commit/6f842ba52468537248fa6361146279063352fb47). Interestingly, carState.brake is user only. I suspect carState.gas was defined as user + computer because that's the signal we knew of on the first openpilot cars, the Civic and Acura ILX.

selfdrive/car/honda/carstate.py
belm0: This is a remnant from abstracting the pedal message, but I suggest keeping it since there are further opportunities for abstracting messages (e.g. `SCM_FEEDBACK` vs. `SCM_BUTTONS` for `MAIN_ON` signal).

  * this function only has one caller, and the arg default isn't used-- avoid maintaining the `GEARBOX` string literal in two places
  * when additional string args are added in the future, there's potential to mix them up and cause silent bugs rather than runtime exceptions-- safer to burden the single caller with keyword args

selfdrive/car/honda/carstate.py
```diff
@@ -277,20 +254,15 @@ def update(self, cp, cp_cam, cp_body):
...
 
     # this is a hack for the interceptor. This is now only used in the simulation
     # TODO: Replace tests by toyota so this can go away
     if self.CP.enableGasInterceptor:
       user_gas = (cp.vl["GAS_SENSOR"]["INTERCEPTOR_GAS"] + cp.vl["GAS_SENSOR"]["INTERCEPTOR_GAS2"]) / 2.
-      ret.gasPressed = user_gas > 1e-5  # this works because interceptor read < 0 when pedal position is 0. Once calibrated, this will change
+      ret.gasPressed = user_gas > 1e-5  # this works because interceptor reads < 0 when pedal position is 0. Once calibrated, this will change
     else:
-      ret.gasPressed = pedal_gas > 1e-5
+      ret.gasPressed = ret.gas > 1e-5
```
belm0, line 265: This refactor isn't equivalent, since `ret.gas` is scaled by 1/256.  Is that OK?

adeebshihadeh: yeah, that should be fine

selfdrive/car/honda/carstate.py
adeebshihadeh: hmm, made it a non kw arg

PR #22013, opened 2021-08-22
accel/decel button long press

gregjhogan:
currently speed always changes 5 mph at a time when openpilot handles the accel/decel button press events

update to do 1 mph for short press and 5 mph for long press

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -48,6 +48,7 @@ Desire = log.LateralPlan.Desire
 LaneChangeState = log.LateralPlan.LaneChangeState
 LaneChangeDirection = log.LateralPlan.LaneChangeDirection
 EventName = car.CarEvent.EventName
+ButtonEvent = car.CarState.ButtonEvent
 
 
 class Controls:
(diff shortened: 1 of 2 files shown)


martinl:
I would vote for 5mph for short and 1mph for long press, since imo 1mph fine tuning is less frequently needed than switching between different speed limits, which would be faster by using short button presses with 5mph step

jyoung8607:
> I would vote for 5mph for short and 1mph for long press, since imo 1mph fine tuning is less frequently needed than switching between different speed limits, which would be faster by using short button presses with 5mph step

I can't speak for every manufacturer, but for VW family, short 5/long would be an even bigger [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) violation than the current openpilot longitudinal behavior: it's the exact opposite of how stock ACC works.

martinl:
There are both L1/S5 and S5/L1 (and probably more diverse) implementations for stock ACC controls and some users will have to adapt a bit when picking one or the other. Either way, this is useful additional capability for openpilot longitudinal control.

PR #22014, opened 2021-08-23
Add Kodiaq 2019 fingerprint

bluetulippon:
Add Kodiaq 2019 fingerprint


selfdrive/car/volkswagen/values.py
```diff
@@ -463,16 +463,19 @@ class CAR:
   },
   CAR.SKODA_KODIAQ_MK1: {
     (Ecu.engine, 0x7e0, None): [
+      b'\xf1\x875NA907115E \xf1\x890003',
```
grekiki, line 466: Don't add this, it's already included

jyoung8607:
Thanks for the FP data! Since this is the first 2019 Kodiaq we're adding data for, please expand the supported model year range to "2018-19" in the README as well.

selfdrive/car/volkswagen/values.py
adeebshihadeh:
```suggestion
```

adeebshihadeh:
What's your dongle ID?

bluetulippon:
Hi, adeebshihadeh, 

My dongle_id is f8a8f3e52154e16a


Hi, jyoung8607,

I try to add the model to README.md, but it's a conflicting file https://github.com/commaai/openpilot/commit/8c44cb24349734204eef0df90c990bb0a7acb1d9


Hi, grekiki,

Thanks for your review, adeebshihadeh has already changed the file.

jyoung8607:
> Hi, jyoung8607,
> 
> I try to add the model to README.md, but it's a conflicting file [8c44cb2](https://github.com/commaai/openpilot/commit/8c44cb24349734204eef0df90c990bb0a7acb1d9)

Škoda Octavia, right underneath it, was touched to expand its model-year support range after you initially checked out your branch, so the diff doesn't quite match anymore. I think github will give you a UI you can use to fix the conflict, or you can rebase this branch on commaai/master on your side, or apply your changes to a fresh branch based on commaai/master.

If you have trouble with it and need help, I can resubmit this on your behalf, but for those that are willing and able to make their contributions directly, I certainly encourage that.

bluetulippon:
I created a new pull request https://github.com/commaai/openpilot/pull/22039.

So close this, thank you.

grekiki:
Well that's an interesting solution I suppose :)

PR #22040, opened 2021-08-24
Params: do blocking read in params_pyx.pyx

deanlee:


pd0wm:
What do we gain by moving this functionality out of the python code? Are there situations where our signal handling isn't working correctly?

deanlee:
The implementation of Params::get is not complete, the prev handle will miss the exit signals. It relies on the cython caller to throw exceptions to work properly. I think it might be more reasonable to put the complete blocking read logic in cython.

pd0wm:
What do you mean by "prev handle will miss the exit signals". Which signals besides SIGINT and SIGTERM are you talking about?

deanlee:
we replaced the prev signal handle with a new one,  and did not pass the signal to the prev handle after SIGINT and SIGTERM occurred. for example, If it is called in a c++ module, the ExitHandler will not catch the exit signal.

pd0wm:
You're right, that is concerning, we should make sure the original handler is called. Let's fix up the C++ side of this, instead of moving implementation back into python and losing functionality on the C++ side that we might need later.

PR #22041, opened 2021-08-24
UI: show thermal status instead of ambient temp

adeebshihadeh:


grekiki:
That's going to make it harder to determine the "temperature direction" though...

PR #22050, opened 2021-08-27
C3: Fix keyboard buttons on the right getting stuck in pressed state

### LINKED ISSUE #21907: C3: Fix keyboard buttons on the right getting stuck in pressed state
@sshane made some progress on this in #21774

deanlee:
fixed #21907
This is not the final solution.there is no problem with KeyButton.  KeyButton::PaintEvent and QAbstractButton::setDown(https://code.woboq.org/qt5/qtbase/src/widgets/widgets/qabstractbutton.cpp.html#_ZN15QAbstractButton7setDownEb) are always be called with correct state. 
but the relevant area of ​​the screen is not updated. I don't know the reason yet. the quick solution is to call parentWidget()->update() in KeyButton::event  or set focus on line edit.

### DIFF
--- a/selfdrive/ui/qt/widgets/input.cc
+++ b/selfdrive/ui/qt/widgets/input.cc
@@ -13,7 +13,6 @@ QDialogBase::QDialogBase(QWidget *parent) : QDialog(parent) {
   Q_ASSERT(parent != nullptr);
   parent->installEventFilter(this);
 
-  setWindowFlags(Qt::Popup);
   setStyleSheet(R"(
     * {
       outline: none;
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
The keyboard feels less responsive after this for some reason.

sshane:
@adeebshihadeh keyboard multitouch was broken in https://github.com/commaai/openpilot/commit/a96f35896681cbbf9752b2ba4176b4d7b43d6d1b somehow

deanlee:
@adeebshihadeh it's ok now

adeebshihadeh:
Thanks for tracking that down @sshane. 

Nice work @deanlee! There still seems to be something weird going on with the right side of the keyboard. If you hold down a button on the left side and tap on buttons on the right, the buttons on the right don't show the pressed style. Will merge as this is a big improvement, but would be nice to track that down.

deanlee:
@adeebshihadeh  : fixed this issue in #22127

Issue #20375, opened 2021-03-16
KIA ceed 2019 dashboard warning


ZanZaD13:
FCA (Forward collision assistence) Dashboard warning upon engage.

When engaging OP FCA warning comes up on dashboard and stays until disengage.
Warning can be delayd 5-10 seconds when both lanelines are visible.

Issue has ocurred since either 0.7.4 or 0.7.5 since this was not happening on 0.7.3

**Device/Version information (please complete the following information):**
 - Device: Comma two
 - Dongle ID: e0e98335f3ebc58f
 - Route: e0e98335f3ebc58f|2021-03-14--16-54-08
 - Timestamp: Upon every engage. 17:59
 - Version: master 0.8.3
 - Car make/model: Kia ceed 2019

VirtuallyChris:
Try adding your car to the 20hz LFA MFA message area. A similar PR is open here https://github.com/commaai/openpilot/pull/20320

If that fixes it, submit a PR and we'll merge it.

ZanZaD13:
Car is allready there, does not change the problem.

adeebshihadeh:
Is this still an issue?

ZanZaD13:
Yes this is still happening in 0.8.5.

sshane:
Are you still experiencing this issue?

ZanZaD13:
> Are you still experiencing this issue?

Yes this is still an issue on 0.8.13.1 Release2

Why isn't it auto updating to 0.8.14?

sshane:
@ZanZaD13 please try this branch if you're still using a comma two: https://github.com/commaai/openpilot/commits/fix-lfa-warn-c2

ZanZaD13:
I will try this as soon as i can, I'm still on C2.

ZanZaD13:
> @ZanZaD13 if you're still using a comma two, please try this branch and let me know how it goes: https://installer.comma.ai/commaai/fix-lfa-warn-c2
> 
> If you're using a comma three, try this branch: https://installer.comma.ai/commaai/fix-lfa-warn

I've only had time for a 5 minute drive, weird result, seemed like it was working on first engage, (op drove about 20 seconds without FCA warning) I had to make a turn, then on next engage i got FCA warning again..

sshane:
Hi @ZanZaD13, please try that branch again, either by using SSH and `git pull`, or re-install that branch entirely. Let me know with a route if it works!

ZanZaD13:
[...]

Issue #20113, opened 2021-02-19
Hyundai Kona EV 2019, Collision Sensor error message.


mstoteler:
When engaging open pilot, the car gives an 
error message to check the forward collision sensor. And gives a few orange error symbols on the car dashboard. (see attached image)

Device: Comma 2
Dongle ID: 9e02651652ef6c6f
Software version: Master-Ci (0.8.2)

Car make/model: HYUNDAI KONA EV 2019
Route: 9e02651652ef6c6f|2021-02-19--07-48-20

![PXL_20210219_075631910 MP2](https://user-images.githubusercontent.com/79313978/108501396-8325c300-72b1-11eb-950f-0e247f42994f.jpg)

adeebshihadeh:
Has this always happened on your car or is it a recent regression?

mstoteler:
I only know that it is the case now. The FCA warning always shows op when engaging OP.

adeebshihadeh:
Can you record a route with the stock system and post it here? All you need to do is disable openpilot in the settings before starting your car. No need to drive, just leave your car on for a minute or so.

mstoteler:
I recorded a drive, disabled openpilot and on stock OP on release2.

9e02651652ef6c6f|2021-03-11--12-02-15

adeebshihadeh:
Can you test this branch? https://github.com/commaai/openpilot/pull/20320

The discord community (https://discord.comma.ai) can help you with checking out the branch.

mstoteler:
I've tested it this morning. But still the same message..

9e02651652ef6c6f|2021-03-12--08-56-28

mstoteler:
Is this problem not fixable? My car doesn't emergency break with OpenPilot engaged?

adeebshihadeh:
This problem is certainly fixable, but your car is a community maintained car, meaning support largely comes from the community. Yes, emergency braking is likely affected given the fault you're seeing.

sshane:
Is this still an issue on stock openpilot, but not a fork?

sshane:
@xx979xx Do you happen to remember if any of your changes fix this user's issue?

DerGue69:
Hi. I have Kia Niro MY 19 - EU Austria 1st Edition. I am facing the same problem. Yesterday  I installed stock OpenPilot with fingerprint modifications. 
When I engage Openpilot I instantly get the FCA Warning message.

Can you give me a hint what to test, which fork I should try?

Günther

[...]

PR #22051, opened 2021-08-27
HKG: Add FW for 2018 Hyundai Kona EV

jyoung8607:
Add firmware for the 2018 Hyundai Kona EV. Expand supported model-year range.

**Dongle ID:** 3f4095e3f5a6d997

Thanks to community Kona EV owner Jazzah.


jyoung8607:
The user here is affected by the the same FCA-error-while-engaged malady as the users in #20113 and #20375. At least some of the popular HKG community forks have done something to prevent that from happening, but on cursory inspection I can't see what it is. It doesn't appear to be one of the usual HKG feature-group membership things.

Taking this out of review and proposing a merge because the data is valid, and it's at least as functional as the other listed cars. I'll see if I can track down the issue but it'll be outside of this PR.

pd0wm:
That data in the VIN request is very strange. Especially because that's the first request we send before other UDS interactions. Do you have the rlogs for one of those interactions, would love to see what was going on on the CAN bus.

selfdrive/car/hyundai/values.py
```diff
@@ -32,7 +32,7 @@ class CAR:
   IONIQ_EV_2020 = "HYUNDAI IONIQ ELECTRIC 2020"
   IONIQ_PHEV = "HYUNDAI IONIQ PHEV 2020"
   KONA = "HYUNDAI KONA 2020"
-  KONA_EV = "HYUNDAI KONA ELECTRIC 2019"
+  KONA_EV = "HYUNDAI KONA ELECTRIC 2018-2019"
```
pd0wm, line 35:
```suggestion
  KONA_EV = "HYUNDAI KONA ELECTRIC 2019"
```
We try to not change car strings after the initial port.

jyoung8607:
`3f4095e3f5a6d997|2021-08-27--23-54-57` (preserved, with rlogs)
`3f4095e3f5a6d997|2021-08-28--12-06-31` (preserved, with partial rlogs)

That's pretty much what I have to choose from. The associated C3 is off in retropilot-associated forkworld at this particular moment, but I think he would probably come back if there was something that needed testing or validation.

pd0wm:
Checked the can logs, and it's actually sending that as a response. Probably related to the other weird responses where the ECU isn't fully initialized yet.

```
129 0x7df b'0209020000000000' b'\x02\t\x02\x00\x00\x00\x00\x00'
129 0x18db33f1 b'0209020000000000' b'\x02\t\x02\x00\x00\x00\x00\x00'
0 0x7df b'0209020000000000' b'\x02\t\x02\x00\x00\x00\x00\x00'
2 0x7df b'0209020000000000' b'\x02\t\x02\x00\x00\x00\x00\x00'
1 0x7ec b'10144902014f5345' b'\x10\x14I\x02\x01OSE'
129 0x7e4 b'3000000000000000' b'0\x00\x00\x00\x00\x00\x00\x00'
1 0x7ec b'21565f5f20424d53' b'!V__ BMS'
1 0x7ec b'2220000000000000' b'" \x00\x00\x00\x00\x00\x00'
```

PR #22054, opened 2021-08-28
retry ioctls while errno == EINTR

deanlee:


deanlee:
frames may be dropped if don't retry on EINTR
https://github.com/commaai/openpilot/blob/3b0b20ba4bab2a32d28145000a4f260ad38ea402/selfdrive/camerad/cameras/camera_qcom2.cc#L1099-L1116

PR #22067, opened 2021-08-29
camera_qcom2: refactor open_v4l_by_name_and_index

deanlee:


selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -540,41 +540,23 @@ static void camera_init(MultiCameraState *multi_cam_state, VisionIpcServer * v,
...
-        snprintf(nbuf, sizeof(nbuf), "/dev/v4l-subdev%d", v4l_index);
-        LOGD("open %s for %s index %d", nbuf, name, index);
-        return open(nbuf, flags);
-      }
-      cnt_index--;
+int open_v4l_by_name_and_index(const char name[], int index, int flags = O_RDWR | O_NONBLOCK) {
+  for (int v4l_index = 0; /**/; ++v4l_index) {
+    std::string v4l_name = util::read_file(util::string_format("/sys/class/video4linux/v4l-subdev%d/name", v4l_index));
+    if (v4l_name.empty()) return -1;
+    if (v4l_name.compare(0, v4l_name.length() - 1, name) == 0 && index-- == 0) {
```
pd0wm, line 547: Maybe `v4l_name.find(name) == 0` is clearer? Also keep the comment about there being a `\n`.

Think relying on the short circuiting of `&&` makes the code a little bit harder to read than necessary. Let's keep the nested if from the original PR.

deanlee: much nicer :+1:

PR #22070, opened 2021-08-30
Take camera snapshot using VisionIPC

adeebshihadeh:


selfdrive/camerad/cameras/camera_common.h
```diff
@@ -34,6 +34,7 @@
 #define LOG_CAMERA_ID_QCAMERA 3
 #define LOG_CAMERA_ID_MAX 4
 
+// TODO: remove these once all the internal tools are moved to vipc
```
pd0wm, line 37: Are they used for focussing lenses in production?

adeebshihadeh: yes, lens focusing and sharpness tests

pd0wm:
Probably no point in fixing up the RGB padding if we're going to merge https://github.com/commaai/openpilot/pull/22905 soon.

PR #22077, opened 2021-08-30
camera_qcom2: new functions device_acquire & device_config

deanlee:


selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -76,12 +76,29 @@ int cam_control(int fd, int op_code, void *handle, int size) {
...
+  return cam_control(fd, CAM_CONFIG_DEV, &cmd, sizeof(cmd));
+}
+
 int device_control(int fd, int op_code, int session_handle, int dev_handle) {
   // start stop and release are all the same
-  static struct cam_release_dev_cmd release_dev_cmd;
-  release_dev_cmd.session_handle = session_handle;
-  release_dev_cmd.dev_handle = dev_handle;
-  return cam_control(fd, op_code, &release_dev_cmd, sizeof(release_dev_cmd));
+  struct cam_start_stop_dev_cmd cmd { .session_handle = session_handle, .dev_handle = dev_handle };
```
pd0wm, line 100: Why did the struct type change?

selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -76,12 +76,29 @@ int cam_control(int fd, int op_code, void *handle, int size) {
   return ret;
 }
 
+int32_t device_acquire(int fd, int32_t session_handle, void *data) {
+  struct cam_acquire_dev_cmd cmd = {
+      .session_handle = session_handle,
+      .handle_type = CAM_HANDLE_USER_POINTER,
+      .num_resources = (uint32_t)(data ? 1 : 0),
+      .resource_hdl = (uint64_t)data};
```
pd0wm, line 84:
```suggestion
      .resource_hdl = (uint64_t)data,
  };
```

selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -76,12 +76,29 @@ int cam_control(int fd, int op_code, void *handle, int size) {
   return ret;
 }
 
+int32_t device_acquire(int fd, int32_t session_handle, void *data) {
```
pd0wm, line 79: return `std::optional<int32_t>` and let the calling site assert on error condition.

selfdrive/camerad/cameras/camera_qcom2.cc
deanlee: I was confused by the name when I was reading the code

selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -650,23 +648,16 @@ static void camera_open(CameraState *s) {
     .comp_grp_id = 0x0, .split_point = 0x0, .secure_mode = 0x0,
   };
 
-  ret = cam_control(s->multi_cam_state->isp_fd, CAM_ACQUIRE_DEV, &acquire_dev_cmd, sizeof(acquire_dev_cmd));
-  LOGD("acquire isp dev: %d", ret);
+  s->isp_dev_handle = device_acquire(s->multi_cam_state->isp_fd, s->session_handle, &isp_resource).value_or(-1);
```
pd0wm, line 651: This `value_or` with a magic -1 seems kind of a hack. Think we should make a temporary variable and assert that. It's one more line but should be more clear.
```suggestion
  auto isp_dev_handle = device_acquire(s->multi_cam_state->isp_fd, s->session_handle, &isp_resource);
  assert(isp_dev_handle);
  s->isp_dev_handle = *isp_dev_handle;
```

selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -76,12 +76,36 @@ int cam_control(int fd, int op_code, void *handle, int size) {
...
+  return cam_control(fd, CAM_CONFIG_DEV, &cmd, sizeof(cmd));
+}
+
 int device_control(int fd, int op_code, int session_handle, int dev_handle) {
-  // start stop and release are all the same
-  static struct cam_release_dev_cmd release_dev_cmd;
-  release_dev_cmd.session_handle = session_handle;
-  release_dev_cmd.dev_handle = dev_handle;
-  return cam_control(fd, op_code, &release_dev_cmd, sizeof(release_dev_cmd));
+  if (op_code == CAM_START_DEV || op_code == CAM_STOP_DEV) {
```
pd0wm, line 100: Think the old code was fine since all the structs are the same, I was just wondering why it was changed with respect to the original code.

deanlee: so how about using cam_start_stop_dev_cmd ? release is a bit confusing.

pd0wm: ah, I guess. If you think that's clearer it's fine

pd0wm:
Looks good now!

PR #22084, opened 2021-08-31
ONNXModel: use the actual model file name

deanlee:


adeebshihadeh:
Using SNPE on PC should still be supported.

selfdrive/modeld/models/driving.cc
```diff
@@ -63,10 +63,12 @@ void model_init(ModelState* s, cl_device_id device_id, cl_context context) {
   constexpr int output_size = OUTPUT_SIZE + TEMPORAL_SIZE;
   s->output.resize(output_size);
 
-#if (defined(QCOM) || defined(QCOM2)) && defined(USE_THNEED)
+#if defined(USE_THNEED)
```
adeebshihadeh, line 66:
```suggestion
#ifdef USE_THNEED
```

selfdrive/modeld/models/driving.cc
```diff
@@ -63,10 +63,12 @@ void model_init(ModelState* s, cl_device_id device_id, cl_context context) {
...
   s->m = std::make_unique<ThneedModel>("../../models/supercombo.thneed", &s->output[0], output_size, USE_GPU_RUNTIME);
+#elif defined(USE_ONNX_MODEL)
```
adeebshihadeh, line 68:
```suggestion
#elif USE_ONNX_MODEL
```

PR #22089, opened 2021-08-31
Honda carstate: fix forcing is_metric to False

belm0:
A bug in !CIVIC case would override the user's IsMetric option with False, and subsequently that forced value is written back to the car HUD (via IMPERIAL_UNIT).

### DIFF
--- a/selfdrive/car/honda/carstate.py
+++ b/selfdrive/car/honda/carstate.py
@@ -348,7 +348,8 @@ class CarState(CarStateBase):
         ret.brakePressed = True
 
     # TODO: discover the CAN msg that has the imperial unit bit for all other cars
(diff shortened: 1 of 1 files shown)


belm0:
I misunderstood was that the is_metric in controlsd is different than is_metric in carstate.  They are independent:
  * controlsd - Controls.is_metric is only used for call to create_alerts()
  * carstate - CarState.is_metric is written out to ACC_HUD / IMPERIAL_UNIT

But it means that:
  *  IMPERIAL_UNIT is forced True for all Honda's other than Civic.  I don't know if that causes problems in practice.
  * the user's IsMetric toggle is not propagated to the HUD

PR #22092, opened 2021-08-31
Correct Hyundai Steering Torque sensor

oricalworks:
Fixed by using the sensor between steering wheel and steering column!

### DIFF
--- a/selfdrive/car/hyundai/carstate.py
+++ b/selfdrive/car/hyundai/carstate.py
@@ -180,6 +180,7 @@ class CarState(CarStateBase):
     checks = [
       # address, frequency
       ("MDPS12", 50),
+      ("MDPS11", 100),
       ("TCS13", 50),
       ("TCS15", 10),
       ("CLU11", 50),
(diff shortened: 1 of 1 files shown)


pd0wm:
Is this value signed correctly? Or is it always positive?

oricalworks:
> Is this value signed correctly? Or is it always positive?

https://api.retropilot.org/cabana/index.html?retropilotIdentifier=de647b2e|662b29cc0cd29d6959b61565247f3a361f680c2fd3ed3aa0faf281841085b98e|2021-07-15--19-01-57|e26245a402dd5e4044ea054896cdeca1f602abd278293e39584791b988ba12e5&retropilotHost=https%3A%2F%2Fapi.retropilot.org%2F&demo=1

look at MDPS11 CR_Mdps_DrvTq

oricalworks:
fixed missing check

oricalworks:
The failed check is because of HONDA ACCORD 2018 which is weird bc I didn't touch master and only made changes to  Hyundai.

pd0wm:
Oh my bad, it's the angle rate that's not correctly signed on HKG.

What's the benefit of this signal over the existing one?

oricalworks:
> Oh my bad, it's the angle rate that's not correctly signed on HKG.
> 
> What's the benefit of this signal over the existing one? Is this driver torque instead of total torque? 
YES exactly its driver applied torque and not total.

pd0wm:
Seems like there was some discussion about this already:
 - https://github.com/commaai/openpilot/pull/970#discussion_r368311619
 - https://github.com/commaai/panda/pull/427

From those discussions it seems like not all cars have this signal and it's better to use the current one.

oricalworks:
> Seems like there was some discussion about this already:
> 
> * [Hyundai-Kia-Genesis (HKG) #970 (comment)](https://github.com/commaai/openpilot/pull/970#discussion_r368311619)
> * [Hyundai-Kia-Genesis (HKG) panda#427](https://github.com/commaai/panda/pull/427)
> 
> From those discussions it seems like not all cars have this signal and it's better to use the current one.

Thank you

PR #22096, opened 2021-09-01
Ensure that first segment starts with all cameras on the same frameId

pd0wm:
If logging starts at the wrong time the first frame of the video file can be mismatched between cameras. This then persists for the whole drive.


selfdrive/loggerd/loggerd.cc
```diff
@@ -124,9 +131,11 @@ void encoder_thread(const LogCameraInfo &cam_info) {
   std::vector<Encoder *> encoders;
   VisionIpcClient vipc_client = VisionIpcClient("camerad", cam_info.stream_type, false);
 
+  bool ready = false;
+
   while (!do_exit) {
     if (!vipc_client.connect(false)) {
-      util::sleep_for(100);
+      util::sleep_for(1);
```
adeebshihadeh, line 138: how much CPU does this use if camerad isn't up or dies?

PR #22100, opened 2021-09-01
Add script to rerun process on route and generate new logs

pd0wm:


selfdrive/debug/.gitignore
```diff
@@ -0,0 +1 @@
+*.bz2
```
adeebshihadeh, line 1: add to root gitignore?

PR #22108, opened 2021-09-02
Honda e Port

vanillagorillaa:
**Checklist**
- [x] added to README
- [x] route with openpilot: 2d5808fae0b38ac6|2021-09-01--17-14-11
- [x] route with stock system:  2d5808fae0b38ac6|2021-08-27--18-55-09


selfdrive/car/honda/interface.py
```diff
@@ -278,6 +278,16 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       tire_stiffness_factor = 0.82
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.6], [0.18]]
 
+    elif candidate == CAR.HONDA_E:
+      stop_and_go = True
+      ret.mass = 3338.8 * CV.LB_TO_KG + STD_CARGO_KG
+      ret.wheelbase = 2.5
+      ret.centerToFront = ret.wheelbase * 0.39
+      ret.steerRatio = 15.0  # Cant find a number for this in spec sheet
```
grekiki, line 286: Had a bit of fun with that. Using the specs from https://hondanews.eu/eu/en/cars/media/pressreleases/199768/2020-honda-e-33 I got the wheelbase of 2.53m and the turning radius at the wheels of 4.3 meters. I also get that the steering wheel needs 3.1 turns lock to lock so 1.55 turns in one direction or 558 degrees is the full steering wheel turn. Using https://calculator.academy/turning-radius-calculator/#f1p0|f2p0 I got the actual front wheel angle of 30.45 degress combining the wheelbase and turning radius from before. This gives a steering ratio of 18.3, but given the variable ratio steering it should probably be 15-20% higher.

vanillagorillaa: Logs show 16.710999 as the learned steer ratio, will update and see how it changes.

grekiki: center_to_front should also be *0.5 instead of *0.39
https://www.google.com/amp/s/www.motor1.com/news/354582/honda-e-tech-details-released/amp/

vanillagorillaa: Updated

PR #22118, opened 2021-09-03
panda.cc: new function enumerate_usb_devices

deanlee:


selfdrive/boardd/panda.cc
```diff
@@ -106,44 +111,16 @@ void Panda::cleanup() {
...
-  for (size_t i = 0; i < num_devices; ++i) {
-    libusb_device *device = dev_list[i];
-    libusb_device_descriptor desc;
-    libusb_get_device_descriptor(device, &desc);
-    if (desc.idVendor == 0xbbaa && desc.idProduct == 0xddcc) {
-      libusb_device_handle *handle = NULL;
-      libusb_open(device, &handle);
-      unsigned char desc_serial[26] = { 0 };
-      int ret = libusb_get_string_descriptor_ascii(handle, desc.iSerialNumber, desc_serial, std::size(desc_serial));
-      libusb_release_interface(handle, 0);
```
deanlee, line 132: we don't need to call libusb_release_interface here. it's used to release interface previously claimed with libusb_claim_interface()

adeebshihadeh:
In this case, I think using niche C++ features increases complexity and reduces readability.

PR #22121, opened 2021-09-03
Hyundai longitudinal prerequisites

pd0wm:
TODO

- [x] ~Can we always use `TCS13`?~ 30ms difference, but not worth it since we still need to check available from SCC.
- [x] ~Can we use `TCS13` in the panda instead of button presses.~ No, otherwise we won't be allow to send the acc request.
- [x] Fix radar knockout only working on first boot
- [x] Combine radar knockout code with Honda Bosch
- [x] Fix tuning


selfdrive/car/hyundai/carcontroller.py
```diff
@@ -62,24 +65,48 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert,
...
+          # send 25 messages at a time to increases the likelihood of resume being accepted
+          can_sends.extend([create_clu11(self.packer, frame, CS.clu11, Buttons.RES_ACCEL)] * 25)
+          self.last_resume_frame = frame
+
+    if frame % 2 == 0 and CS.CP.openpilotLongitudinalControl:
+      lead_visible = True
+      accel = actuators.accel if enabled else 0
+      accel = clip(accel, CarControllerParams.ACCEL_MIN, CarControllerParams.ACCEL_MAX)
+
+      stopping = actuators.longControlState == LongCtrlState.stopping
```
gregjhogan, line 94: I added requiring `CS.out.vEgo <= 0.1` when setting this for my vehicle because otherwise I would sometimes get a really hard, uncomfortable jerk coming to a stop

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -62,24 +65,48 @@ def update(self, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert,
                         left_lane, right_lane, left_lane_depart, right_lane_depart)
 
     can_sends = []
+
+    # tester present - w/ no response (keeps radar disabled)
+    if CS.CP.openpilotLongitudinalControl:
+      if (frame % 10) == 0:
+        can_sends.append([0x7D0, 0, b"\x02\x3E\x80\x00\x00\x00\x00\x00", 0])
```
gregjhogan, line 72: slow this down to 2 Hz or even 1 Hz? only needs to be once every 5 seconds, and diagnostic tools generally send it at 1 Hz

PR #22126, opened 2021-09-04
Test case for enable BSM on CH-R - TSS-P with BSM

cferra:


pd0wm:
Feel free to re open with a description once you're done testing.

PR #22127, opened 2021-09-04
c3 keyboard: fix button status not updated

deanlee:


adeebshihadeh:
Keyboard feels great now :)

PR #22129, opened 2021-09-05
Fixes for C3 LTE connection being deleted

### LINKED ISSUE #22115: No LTE after fresh install on new C3
**Describe the bug**

After installing release3 openpilot using the normal methods on a factory-fresh C3, LTE connectivity wasn't working. This was discovered and noted as part of a separate issue #22105.

Unknown.

 - Device: CCC3
 - Dongle ID: 3cfdec54aa035f3f
 - Route: N/A / all
 - Timestamp: N/A / all prior to late this afternoon
 - Version: 0.8.8 release3
 - Car make/model: 2018 Volkswagen Golf R

**Additional context**

Upon further digging, it was found the default LTE system connection was either never created, or not completely created, or the connection profile didn't make it to stable storage _[edit: it was found to be deleted programmatically]_. In the output below, the UUID-nmmeta symlink to `/dev/null` is of the same UUID as the system default LTE connection.

I don't really know how I got to this state, or if anyone else might conceivably be affected. If so, this does seem something easy to detect and mitigate in startup scripts. I fixed the issue manually by removing the nmmeta link and copying the default LTE connection from `/usr/lib/NetworkManager` and restarting NetworkManager. _[edit: there's no need to manually copy the system default LTE connection, just delete the nmmeta shadowing link.]_ In retrospect, I think it likely a factory reset would also have fixed it.

**Example NetworkManager info from a good C3 with operating LTE:**
```
comma@tici:/data/openpilot$ nmcli conn show
NAME                              UUID                                  TYPE  DEVICE   
openpilot connection young        fa834e4b-2ad9-4ca8-b6bb-013f50914db7  wifi  wlan0    
lte                               6586fb1f-4a66-46d4-9164-0100b73750ee  gsm   cdc-wdm0 
Hotspot                           cbdf53cc-83dc-4539-9161-a7e6aeb0c4be  wifi  --       
openpilot connection jypixel      8939f9c8-ae58-46ed-b12c-6f138ad84980  wifi  --       
openpilot connection STRESS_TEST  27629e64-81f1-4b93-afe8-9956022a34ef  wifi  --  
```

```
comma@tici:/data/etc/NetworkManager/system-connections$ ll
total 24
drwxr-xr-x 2 root root 4096 Aug 11 19:24  ./
drwxr-xr-x 3 root root 4096 May 27 11:16  ../
-rw------- 1 root root  441 Aug 11 19:24  Hotspot.nmconnection
-rw------- 1 root root  363 Aug  2 23:35 'openpilot connection jypixel.nmconnection'
-rw------- 1 root root  308 Jul 30 20:27 'openpilot connection STRESS_TEST.nmconnection'
-rw------- 1 root root  376 Aug  3 00:34 'openpilot connection young.nmconnection'
```

**Example NetworkManager info from this C3 in a bad state:**
```
comma@tici:/data/openpilot$ nmcli conn show
NAME                        UUID                                  TYPE  DEVICE 
openpilot connection young  67419148-9117-4993-bd7f-a5b65a957ebd  wifi  wlan0
```

```
comma@tici:/data/etc/NetworkManager/system-connections$ ll
total 12
drwxr-xr-x 2 root root 4096 Aug 31 17:50  ./
drwxr-xr-x 3 root root 4096 Jul 21 15:00  ../
lrwxrwxrwx 1 root root    9 Aug 31 17:50  6586fb1f-4a66-46d4-9164-0100b73750ee.nmmeta -> /dev/null
-rw------- 1 root root  376 Aug 31 11:16 'openpilot connection young.nmconnection'
```

jyoung8607:
**Description**

Partially resolve #22115 by fixing two `wifiManager.cc` code paths that could delete the LTE connection. 

The first fix will make it so we don't wrongly try to forget connections outside of `ui`'s new connection process. It was possible to trigger this if wpa_supplicant tried to re-authenticate to an existing Wifi network while driving out of range. In that case, it didn't even try to delete the current Wifi connection; instead it was trying to delete an empty SSID connection.

The second fix will make sure the LTE connection can't possibly make it onto the `knownConnections` SSID list. We prevent that at init time, but there was a possible path if NetworkManager ever sent a DBus NewConnection event for the LTE. Unsure if this would do it, but it's possible that `ui` could come up before the LTE modem finishes booting and enumerating. If that ever happened, the first bug that was trying to forget empty SSID connections could actually match the empty SSID LTE connection. 

**Verification**

None. The associated bug is difficult to reproduce on-demand; I've made no attempt. Please feel 150% free to check my logic.

**Additional context**

As already noted in the bug, we should also make an AGNOS-side fix to always un-delete the LTE connection at boot-up. If nothing else, we need to recover LTE on any devices that got stuck in this state before receiving the fix.

Related to the current issue, I think the current `ui` assumption that `lteConnection` is the only non-802-11-wireless connection is asking for trouble when we can hook up USB Ethernet devices. I don't have a USB Ethernet interface handy to write code for handling that case tonight.

### DIFF
--- a/selfdrive/ui/qt/offroad/wifiManager.cc
+++ b/selfdrive/ui/qt/offroad/wifiManager.cc
@@ -341,9 +341,14 @@ void WifiManager::connectionRemoved(const QDBusObjectPath &path) {
 }
 
 void WifiManager::newConnection(const QDBusObjectPath &path) {
-  knownConnections[path] = getConnectionSettings(path).value("802-11-wireless").value("ssid").toString();
-  if (knownConnections[path] != tethering_ssid) {
-    activateWifiConnection(knownConnections[path]);
+  const Connection &settings = getConnectionSettings(path);
+  if (settings.value("connection").value("type") == "802-11-wireless") {
+    knownConnections[path] = settings.value("802-11-wireless").value("ssid").toString();
+    if (knownConnections[path] != tethering_ssid) {
+      activateWifiConnection(knownConnections[path]);
+    }
+  } else if (path.path() != "/") {
+    lteConnectionPath = path.path();
   }
 }
 
--- a/selfdrive/ui/qt/offroad/wifiManager.cc
+++ b/selfdrive/ui/qt/offroad/wifiManager.cc
@@ -303,7 +303,7 @@ QString WifiManager::getAdapter() {
 
 void WifiManager::stateChange(unsigned int new_state, unsigned int previous_state, unsigned int change_reason) {
   raw_adapter_state = new_state;
-  if (new_state == NM_DEVICE_STATE_NEED_AUTH && change_reason == NM_DEVICE_STATE_REASON_SUPPLICANT_DISCONNECT) {
+  if (new_state == NM_DEVICE_STATE_NEED_AUTH && change_reason == NM_DEVICE_STATE_REASON_SUPPLICANT_DISCONNECT && !connecting_to_network.isEmpty()) {
     forgetConnection(connecting_to_network);
     emit wrongPassword(connecting_to_network);
   } else if (new_state == NM_DEVICE_STATE_ACTIVATED) {
--- a/selfdrive/ui/qt/offroad/wifiManager.cc
+++ b/selfdrive/ui/qt/offroad/wifiManager.cc
@@ -347,8 +347,6 @@ void WifiManager::newConnection(const QDBusObjectPath &path) {
     if (knownConnections[path] != tethering_ssid) {
       activateWifiConnection(knownConnections[path]);
     }
-  } else if (path.path() != "/") {
-    lteConnectionPath = path.path();
   }
 }
 


jyoung8607:
For S&Gs I instrumented `wifiManager.cc` to show which connections it was adding at init time, and show any new connections that came in via DBus events. Then I forced a situation where the LTE modem started well after regular power-on, after UI.

LTE is still picked up as a defined connection at init time even when the hardware isn't present, and powering on LTE well after boot/UI startup doesn't kick off the NewConnection event I thought possible.

This PR is still valid and applicable, it still fixes bugs, and I think the first fix will almost certainly mitigate the original problem. I just wish I could see the complete chain of events.

```diff
diff --git a/selfdrive/ui/qt/offroad/wifiManager.cc b/selfdrive/ui/qt/offroad/wifiManager.cc
index 81da7bd..72854ee 100644
--- a/selfdrive/ui/qt/offroad/wifiManager.cc
+++ b/selfdrive/ui/qt/offroad/wifiManager.cc
@@ -3,6 +3,7 @@
 #include <algorithm>
 #include <set>
 #include <cstdlib>
+#include <cstdio>
 
 #include "selfdrive/common/params.h"
 #include "selfdrive/common/swaglog.h"
@@ -341,6 +342,7 @@ void WifiManager::connectionRemoved(const QDBusObjectPath &path) {
 }
 
 void WifiManager::newConnection(const QDBusObjectPath &path) {
+  printf("UI debug: newConnection called for connection %s\n", qPrintable(path.path()));
   knownConnections[path] = getConnectionSettings(path).value("802-11-wireless").value("ssid").toString();
   if (knownConnections[path] != tethering_ssid) {
     activateWifiConnection(knownConnections[path]);
@@ -376,8 +378,10 @@ void WifiManager::initConnections() {
   for (const QDBusObjectPath &path : response.value()) {
     const Connection &settings = getConnectionSettings(path);
     if (settings.value("connection").value("type") == "802-11-wireless") {
+      printf("UI debug: init adding Wifi connection %s\n", qPrintable(path.path()));
       knownConnections[path] = settings.value("802-11-wireless").value("ssid").toString();
     } else if (path.path() != "/") {
+      printf("UI debug: init adding LTE connection %s\n", qPrintable(path.path()));
       lteConnectionPath = path.path();
     }
   }
```

**Started UI:**

```
comma@tici:/data/openpilot/selfdrive/ui$ ./ui 
GBM_INFO::msmgbm_mapper(125)::gbm mapper instantiated

gbm_create_device(156): Info: backend name is: msm_drm
UI debug: init adding Wifi connection /org/freedesktop/NetworkManager/Settings/1
UI debug: init adding LTE connection /org/freedesktop/NetworkManager/Settings/2
: "wl-shell" is a deprecated shell extension, prefer using "xdg-shell-v6" or "xdg-shell" if supported by the compositor by setting the environment variable QT_WAYLAND_SHELL_INTEGRATION
: Wayland does not support QWindow::requestActivate()
```

**Started LTE:**
```
comma@tici:/data/openpilot$ mmcli -m 0                                                                                                                                                
error: couldn't find modem                                                                                                                                                            
comma@tici:/data/openpilot$ nmcli conn                                                                                                                                                
NAME                        UUID                                  TYPE  DEVICE                                                                                                        
openpilot connection young  67419148-9117-4993-bd7f-a5b65a957ebd  wifi  wlan0                                                                                                         
lte                         6586fb1f-4a66-46d4-9164-0100b73750ee  gsm   --                                                                                                            
comma@tici:/data/openpilot$ time /usr/comma/lte/lte.sh start                                                                                                                          
Starting LTE...                                                                                                                                                                       
 Resetting...                                                                                                                                                                         
 Pulsing power button...                                                                                                                                                              
 Waiting until the modem comes online...                                                                                                                                              
  Waiting...                                                                                                                                                                          
  Waiting...                                                                                                                                                                          
  Waiting...                                                                                                                                                                          
  Waiting...                                                                                                                                                                          
  Waiting...                                                                                                                                                                          
  Waiting...
  Waiting...
  Waiting...
  Waiting...
  Waiting...
  Waiting...
  Waiting...
  Waiting...
  Modem is online!

real    0m15.080s
user    0m0.020s
sys     0m0.000s
comma@tici:/data/openpilot$ nmcli conn
NAME                        UUID                                  TYPE  DEVICE   
openpilot connection young  67419148-9117-4993-bd7f-a5b65a957ebd  wifi  wlan0    
lte                         6586fb1f-4a66-46d4-9164-0100b73750ee  gsm   cdc-wdm0
```

No further output from UI.

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -341,9 +341,14 @@ void WifiManager::connectionRemoved(const QDBusObjectPath &path) {
...
-  knownConnections[path] = getConnectionSettings(path).value("802-11-wireless").value("ssid").toString();
-  if (knownConnections[path] != tethering_ssid) {
-    activateWifiConnection(knownConnections[path]);
+  const Connection &settings = getConnectionSettings(path);
+  if (settings.value("connection").value("type") == "802-11-wireless") {
+    knownConnections[path] = settings.value("802-11-wireless").value("ssid").toString();
+    if (knownConnections[path] != tethering_ssid) {
+      activateWifiConnection(knownConnections[path]);
+    }
+  } else if (path.path() != "/") {
```
pd0wm, line 350: Should we just delete this 2nd branch then since it's already picked up?

jyoung8607: Even though I couldn't trigger the second bug/fix with a late LTE startup, technically nothing stops a user from SSHing in and deleting and recreating the LTE connection, and *that* would surely do it. Won't be common, but shouldn't make UI explode either.

For various reasons I went with the simplest possible approach: make sure new-connection import works just like init-time/existing-connection import. That way it doesn't matter when a connection shows up, it's always handled the same way. Perhaps that code should move to a shared function.

pd0wm: I meant that we keep the check, but don't update `lteConnectionPath`.

jyoung8607: I can take it out if you want, but it IS possible (admittedly vanishingly rare) for the LTE connection to be destroyed and/or created at runtime. If you want to keep things simple, I'd rather push the init-time connection handling out into a separate function:

https://github.com/commaai/openpilot/blob/ec71aa0458d7f655be1204e1fa7f3060c5d264d3/selfdrive/ui/qt/offroad/wifiManager.cc#L377-L382

And call that from both `initConnections()` and `newConnection()`, which is probably what I should have done in the first place. Then we can come back and make sure USB Ethernet devices don't break things no matter when they're plugged in, etc. Is that amenable to you?

pd0wm: Think it's fine to not deal with LTE being recreated at runtime. That should make things easier to reason about.  It's only used for the roaming toggle anyway.

PR #22131, opened 2021-09-05
loggerd: fix concurrency issues in sync encoders

deanlee:
The frame_id of the last camera may not be sent to update_max_atomic for comparison

### DIFF
--- a/selfdrive/loggerd/logging.cc
+++ b/selfdrive/loggerd/logging.cc
@@ -1,0 +1,20 @@
+#include "selfdrive/loggerd/logging.h"
+// Wait for all encoders to reach the same frame id
+bool sync_encoders(LoggerdState *s, CameraType cam_type, uint32_t frame_id) {
+  if (s->max_waiting > 1 && s->encoders_ready != s->max_waiting) {
+    if (std::exchange(s->camera_ready[cam_type], true) == false) {
+      ++s->encoders_ready;
+      LOGE("camera %d encoder ready", cam_type);
+    }
+    if (s->latest_frame_id < frame_id) {
+      s->latest_frame_id = frame_id;
+    }
+    return false;
+  } else {
+    // Small margin in case one of the encoders already dropped the next frame
+    uint32_t start_frame_id = s->latest_frame_id + 2;
+    bool synced = frame_id >= start_frame_id;
+    if (!synced) LOGE("camera %d waiting for frame %d, cur %d", cam_type, start_frame_id, frame_id);
+    return synced;
+  }
+}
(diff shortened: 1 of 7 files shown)


deanlee:
merged to  #22130

pd0wm:
I don't understand completely. You can describe in which situation the last camera does not update the field? And is the 2 frame margin not covering this already?

deanlee:
```
if (cam_info.trigger_rotate && (s.max_waiting > 1)) {
  if (!ready) {
    LOGE("%s encoder ready", cam_info.filename);
    ++s.encoders_ready;
    ready = true;
  }
  // this is not atomic, if ++s.encoders_ready == max_waiting, 
  // the main thread may set s.encoders_synced to true at this time, which will cause the following if check to fail.
  //  we can easily reproduce this problem in unit testing
  if (!s.encoders_synced) {
    update_max_atomic(s.latest_frame_id, extra.frame_id);
```

pd0wm:
Does changing the order of setting ready and updating the frame id to solve the problem?

deanlee:
yes, it's can be fixed like this https://github.com/commaai/openpilot/pull/22131/commits/5e66cd85071cf9ac08965e84cdab837f4683f00e 

But I think only std::mutex can really guarantee consistency. and there is no need to check whether it is synchronized in the main thread, so I put it in a function with std::mutex.

pd0wm:
Just like to prevent locks as much as possible, so if we can do it without locks I'm much more happy with that. That commit is what I had in mind.

deanlee:
I will try, but I feel there will be other synchronization issues without mutex

deanlee:
can't pass the test,  must be protected by a mutex to ensure consistency

selfdrive/loggerd/logging.cc
```diff
@@ -0,0 +1,21 @@
+#include "selfdrive/loggerd/logging.h"
+
+#include "selfdrive/common/util.h"
+
+// Wait for all encoders to reach the same frame id
+bool sync_encoders(LoggerdState *s, CameraType cam_type, uint32_t frame_id) {
```
adeebshihadeh, line 6: I like how the sync logic is nicely contained in this function, but can we do this without adding new files? having all of `logger`, `logging`, and `loggerd` for filenames is confusing.

deanlee: need to move functions out of loggerd.cc for unit testing.
how about rename loggerd.cc-> main.cc, logging.cc->loggerd.cc, logging.h->loggerd.h?

deanlee: sync_encoders needs to be moved out from loggerd.cc for unit test.
how about rename loggerd.cc->main.cc, logging.cc->loggerd.cc, logging.h->loggerd.h ? just like proclogd.

PR #22142, opened 2021-09-06
params: support sub keys

deanlee:
It can be used to group other parameters later, and can also be used to track the logging directories without the .lock files.


adeebshihadeh:
This subkey thing sounds nice to have, but there's a bunch of code to support it and look much cleaner. I think params is in a pretty good place as is.

PR #22149, opened 2021-09-06
C3: clear nmmeta files on boot

adeebshihadeh:


launch_chffrplus.sh
```diff
@@ -111,10 +111,15 @@ function tici_init {
...
 
+  # setup governors
   sudo su -c 'echo "performance" > /sys/class/devfreq/soc:qcom,memlat-cpu0/governor'
   sudo su -c 'echo "performance" > /sys/class/devfreq/soc:qcom,memlat-cpu4/governor'
+
+  # TODO: move this to agnos
+  # network manager config
   nmcli connection modify --temporary lte gsm.auto-config yes
   nmcli connection modify --temporary lte gsm.home-only yes
+  rm -rf /data/etc/NetworkManager/system-connections/*.nmmeta
```
pd0wm, line 122: needs sudo

pd0wm: Let's also only use `-f` no need to delete folders.

pd0wm:
```suggestion
  sudo rm -f /data/etc/NetworkManager/system-connections/*.nmmeta
```

PR #22158, opened 2021-09-07
boardd: new function get_time_str

deanlee:


selfdrive/boardd/boardd.cc
```diff
@@ -40,6 +40,12 @@ std::atomic<bool> ignition(false);
 
 ExitHandler do_exit;
 
+std::string get_time_str(const struct tm &time) {
+  std::string s(30, '\0');
+  std::strftime(&s[0], s.size(), "%Y-%m-%d %H:%M:%S", &time);
```
pd0wm, line 45: Resize the string using the return value of strftime in case we want to use the result somewhere that doesn't call `c_str()`.

PR #22160, opened 2021-09-07
util: add safe write functions

deanlee:


selfdrive/common/util.cc
```diff
@@ -104,6 +104,32 @@ int write_file(const char* path, const void* data, size_t size, int flags, mode_
   return (n >= 0 && (size_t)n == size) ? 0 : -1;
 }
 
+FILE* safe_fopen(const char* filename, const char* mode) {
```
adeebshihadeh, line 107: isn't this the same thing as HANDLE_EINTR(fopen)?

deanlee: not the same, HANDLE_EINTR can only handle functions that return -1

PR #22165, opened 2021-09-07
loggerd: do qlog decimation for EncodeIdx

deanlee:


pd0wm:
We never want to decimate the encodeIdx in the qlog anyway since we need it on the backend to process the video properly. I'll remove the comment.

PR #22166, opened 2021-09-07
loggerd: add variable initEncodeIdx to struct LogCameraInfo

deanlee:


pd0wm:
Don't think adding function pointers makes this more clear.

PR #22170, opened 2021-09-08
boardd: don't set_safety_mode if params is empty

deanlee:


selfdrive/boardd/boardd.cc
```diff
@@ -81,19 +81,22 @@ void safety_setter_thread(Panda *panda) {
     }
     util::sleep_for(100);
   }
-  LOGW("got %d bytes CarParams", params.size());
 
-  AlignedBuffer aligned_buf;
-  capnp::FlatArrayMessageReader cmsg(aligned_buf.align(params.data(), params.size()));
-  cereal::CarParams::Reader car_params = cmsg.getRoot<cereal::CarParams>();
-  cereal::CarParams::SafetyModel safety_model = car_params.getSafetyModel();
+  if (params.size() > 0) {
```
pd0wm, line 85: When does this happen? Won't it stay in the above loop if the size is 0?

deanlee: sorry I was wrong

PR #22173, opened 2021-09-09
Qt: onroad HUD

deanlee:
- [x] qt onroad HUD
- [x] remove nvg related functions after confirming that the ui is the same as before.


adeebshihadeh:
Would be nice to finish off the NVG -> Qt move

PR #22179, opened 2021-09-09
Remove Kia Seltos from Legacy Safety Model

speedking456:
Added this during the initial port thinking it was the issue as to why steering inputs were not being accepted. Turned out to be the checksum and I just never undid this commit.

It has secretly been causing issues and headaches for me, and I assume the other Seltos owners out there.

### DIFF
--- a/selfdrive/car/hyundai/interface.py
+++ b/selfdrive/car/hyundai/interface.py
@@ -218,7 +218,7 @@ class CarInterface(CarInterfaceBase):
 
     # these cars require a special panda safety mode due to missing counters and checksums in the messages
     if candidate in [CAR.HYUNDAI_GENESIS, CAR.IONIQ_EV_2020, CAR.IONIQ_EV_LTD, CAR.IONIQ_PHEV, CAR.IONIQ, CAR.KONA_EV, CAR.KIA_SORENTO,
-                     CAR.SONATA_LF, CAR.KIA_NIRO_EV, CAR.KIA_OPTIMA, CAR.VELOSTER, CAR.KIA_STINGER, CAR.KIA_SELTOS,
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
> It has secretly been causing issues and headaches for me, and I assume the other Seltos owners out there.
What kind of issues? The legacy safety mode is exactly the same, except more relaxed since some HKGs are missing checksums and counters in some of the messages.

speedking456:
> > It has secretly been causing issues and headaches for me, and I assume the other Seltos owners out there.
> 
> What kind of issues? The legacy safety mode is exactly the same, except more relaxed since some HKGs are missing checksums and counters in some of the messages.

It appears to have magically solved an issue I’ve been having where the resume button will engage openpilot, but the car does not accept steering commands. 

Even if it didn’t it’s probably better that it uses a more strict safety model if it can no?

adeebshihadeh:
> Even if it didn’t it’s probably better that it uses a more strict safety model if it can no?
Yes, the stricter safety mode is better. I'd just like to ensure that the behavior between the two safety modes isn't different on your car.

speedking456:
> Yes, the stricter safety mode is better. I'd just like to ensure that the behavior between the two safety modes isn't different on your car.

I haven’t noticed any difference, the resume issue I thought was fixed was not and still occurs intermittently.

Unrelated to this PR. Related to the resume issue. Looked at plot juggler and noticed a big output spike right at the point of engagement. Could that be triggering the car to not accept steering commands for a period of time? How long depends, sometimes it’s like 5 seconds, others it’s probably closer to 20-30s. It happens both on the legacy and current safety models if you were wondering.

I’ll open up a new issue for it soon as it’s been  boggling my mind for a while now.

PR #22185, opened 2021-09-10
replay: cache route files

deanlee:


selfdrive/ui/replay/replay.cc
```diff
@@ -1,13 +1,66 @@
...
+    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, write_data);
+    curl_easy_setopt(curl, CURLOPT_WRITEDATA, fp);
+    res = curl_easy_perform(curl);
+    curl_easy_cleanup(curl);
+    fclose(fp);
+  }
+  return CURLE_OK == res;
+}
+
+std::string sha256_string(const std::string &string) {
```
pd0wm, line 37: https://github.com/commaai/openpilot/blob/master/selfdrive/ui/qt/api.cc#L58

Ggg900800 (approved):

PR #22190, opened 2021-09-10
VW MQB: Škoda Karoq Mk1 round 2

jyoung8607:
Resubmission of #22133 now that openpilot is okay without a factory Lane Assist camera.

@adeebshihadeh, per some of our earlier conversations, this would be a good one for our second `process_replay` route. I'm not sure how you do the fake-data reprocessing on those.


adeebshihadeh:
I'll add to process replay after merging. The regenerated data is used when we add new outputs to the model or other processes (i.e. not necessary in this case).

PR #22192, opened 2021-09-11
test_replay.cc: temporarily comment out the rand seek test

deanlee:


adeebshihadeh:
Can you revert the disable commit to the CI config?

adeebshihadeh:
Let's also make sure to re-run CI a few times to ensure it doesn't randomly fail before merging.

PR #22193, opened 2021-09-11
Mazda: fix cruise disabled

Jafaral:
On some disengagement events (brake/gas/etc) OP may occasionally
end up disabling the main cruise control where the driver has
to press the main cruise button before they can engage ACC again.
Try to detect this situation and automatically turn on cruise
when that happens.

depends on: https://github.com/commaai/opendbc/pull/423

### DIFF
--- a/selfdrive/car/mazda/carcontroller.py
+++ b/selfdrive/car/mazda/carcontroller.py
@@ -8,6 +8,7 @@ class CarController():
     self.apply_steer_last = 0
     self.packer = CANPacker(dbc_name)
     self.steer_rate_limited = False
+    self.brake_counter = 0
 
   def update(self, enabled, CS, frame, actuators):
     """ Controls thread """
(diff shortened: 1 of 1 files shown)


Jafaral:
I tested this with a few drives and it worked as expected. OP now detects any accidental cruise disable and successfully restore to enable state. The driver can still manually disable the main CC and OP doesn't interfere with the driver in that case.

adeebshihadeh:
Why does this happen? Is it always caused by openpilot cancelling cruise control? This seems like a bug with how we're interfacing with the car, so I'd prefer to fix that instead of hacking around it like this.

Jafaral:
It happens because the stock system disengages ACC on a brake press and that works just fine most of the time. In some rare cases, OP might see the brake press before  the updated ACC signal is propagated up to OP (both brake signal cruise signal runs at 50hz, so there is a window of at most 20ms to get out of sync when it comes to OP state), OP then sends a signal to disengage stock ACC. In Mazda, the signal that cancels ACC is the same as the one that turns main CC off. If OP sends the signal after stock ACC was disengaged you end up with Main CC disabled which the situation that this PR  tried to detect and handle.

This is not a new issue, it has been there since day one of the Mazda port. It is rare tough. I tried to fix it in the early days and I had two ideas: 
 
1- Don't send disengage signal on brake and leave it to the stock system to disengage 
2- Wait a cycle (20 ms) before sending disengage signal to give OP time to be in sync with the stock system

Neither of the of these two options sounded appealing when I discussed them with you @adeebshihadeh at the time and considered them against OP safety requirement. My third alternative solution was to do exactly what this PR provides. You considered that an acceptable solution. I didn't have the Main CC signal figured out at the time and the issue hasn't been common enough to have me go after it. the other day I decided to fix it, and I did :-)

adeebshihadeh:
Ideally, I think we should be able to find a new cancellation signal to use that we can always send without side effects. For example, we send a seatbelt unlatched message for the Nissans. For now though, I think the 20ms delay is fine, though we should try to only to do it when necessary (just brake press?). I probably misunderstood when we discussed this before.

Jafaral:
@adeebshihadeh , will add the a 25ms (20+5 grace period) delay after brake press before sending the cancel signal. I will keep the new added logic to track the cruise state and turn it back on as a catch all case, or do you want that to be  dropped ?

adeebshihadeh:
Yes, let's drop the new logic to re-enable cruise.

PR #22195, opened 2021-09-11
c++ replay: send YUV/RGB for all cameras

### LINKED ISSUE #20991: Move to C++ unlogger
We have parallel implementations of unlogger right now, the original python version and the new C++ version. The new unlogger doesn't lag and will support replaying from the UI (#2547). The old unlogger lives in `tools/replay/` and the new one lives in `selfdrive/ui/replay`. Once these TODOs are complete, we can completely remove the old unlogger.

- [x] fallback to qcams, ideally we don't require video either
- [x] add pause/play using spacebar
- [x] stop replay while waiting for video to load (the time is still progressing)
- [x] occasional OpenCL error on VisionIPC init
- [x] work with only 1 segment routes
- [x] show error when route is not found
- [x] sometimes replay starts in the second segment
- [x] add optional argument for start time
- [x] seeking by timestamp is not reliable, sometimes it seeks to a point and them immediately goes back to where you were
- [x] segfaults randomly
- [x] send YUV/RGB for all cameras

### LINKED ISSUE #22221: replay: issues/improvements
- [ ] fallback to qcams, ideally we don't require video either
- [ ] add pause/play using spacebar
- [x] stop replay while waiting for video to load (the time is still progressing)
- [ ] occasional OpenCL error on VisionIPC init
- [ ] work with only 1 segment routes
- [ ] show error when route is not found
- [x] sometimes replay starts in the second segment
- [x] add optional argument for start time
- [x] seeking by timestamp is not reliable, sometimes it seeks to a point and them immediately goes back to where you were
- [x] segfaults randomly
- [ ] send YUV/RGB for all cameras

deanlee:
send YUV/RGB for all cameras

resolve #20991 :
- [x]  fallback to qcams
~- [ ] occasional OpenCL error on VisionIPC init~
- [x] send YUV/RGB for all cameras ( FrameReader need to be refactored)

### DIFF
--- a/selfdrive/ui/replay/camera.cc
+++ b/selfdrive/ui/replay/camera.cc
@@ -3,62 +3,86 @@
 #include <cassert>
 #include <iostream>
 
-static const VisionStreamType stream_types[] = {
-    [RoadCam] = VISION_STREAM_RGB_BACK,
-    [DriverCam] = VISION_STREAM_RGB_FRONT,
-    [WideRoadCam] = VISION_STREAM_RGB_WIDE,
-};
+const int YUV_BUF_COUNT = 50;
 
 CameraServer::CameraServer() {
   device_id_ = cl_get_device_id(CL_DEVICE_TYPE_DEFAULT);
   context_ = CL_CHECK_ERR(clCreateContext(NULL, 1, &device_id_, NULL, NULL, &err));
-
-  start();
-  camera_thread_ = std::thread(&CameraServer::thread, this);
 }
 
 CameraServer::~CameraServer() {
-  // stop camera thread
-  queue_.push({});
-  camera_thread_.join();
-
-  delete vipc_server_;
+  stop();
   CL_CHECK(clReleaseContext(context_));
 }
 
 void CameraServer::start() {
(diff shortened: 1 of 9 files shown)


selfdrive/ui/replay/camera.cc
```diff
@@ -0,0 +1,73 @@
...
+
+CameraServer::~CameraServer() {
+  queue_.push({});
+  camera_thread_.join();
+  delete vipc_server_;
+  CL_CHECK(clReleaseContext(context_));
+}
+
+void CameraServer::thread() {
+  for (;;) {
```
adeebshihadeh, line 29: let's stick to a while true in this case

adeebshihadeh:
```suggestion
  while (true) {
```

deanlee:
this PR should be refactored and merged after #22203

deanlee:
clGetPlatformIDs returns (-1001 CL_PLATFORM_NOT_FOUND_KHR) sometimes.

deanlee:
re-open for reivew

PR #22204, opened 2021-09-12
PHV solar fingerprint V1

ikawaoka:
device: 5af2e3aacd2eb919

fingerprint V2

    carFw = [],

....?


pd0wm:
Do you have your comma power plugged in to the OBD port? That is required to get FP 2.0 FW versions.

PR #22217, opened 2021-09-13
c++ replay: refactor FrameReader

deanlee:
the decode time of each frame is reduced from ~20ms to less than 1ms after refactor.there is no need to use a separate decode thread in FrameReader now.
If the performance is still not enough, we can do additional caching in CameraServer #22195


deanlee:
https://ffmpeg.org/doxygen/3.4/pthread__frame_8c_source.html#l00477

>   506     if (fctx->delaying) {
>   507         *got_picture_ptr=0;
>   508         if (avpkt->size) {
>   509             err = avpkt->size;
>   510             goto finish;
>   511         }
>   512     }

got_picture_ptr will be set to 0 if delaying ,but ret value is > 0. It may be necessary to read a few frames in ctor to "warm up" the decoder thread

PR #22218, opened 2021-09-13
c++ replay: new function eventSocketName

deanlee:
get event socket name from cache. it's  faster than get socket name from DynamicStruct every time.


adeebshihadeh:
Let's focus on reliability and matching the python unlogger functionality before optimizations. The python unlogger is super slow and laggy, so we are already doing much better with replay.

adeebshihadeh:
I'm going to close this for now. It's important to get replay into a state where we can replace unlogger, and replay performance is already way better than unlogger. I think this also hurts the readability.

deanlee:
OK,I got it.

PR #22222, opened 2021-09-14
add gnureadline to Pipfile

sshane:
I'm not sure of the practice for adding packages, how easy it is, but this package adds Python readline functionality to the interactive Python environment, so you can actually use your arrow keys to go to previous lines or left and right. Really speeds up quick dev work where you want to jot something down on your C3!


adeebshihadeh:
I'll look into fixing this in the next AGNOS, but don't think we want to add a package for this. Try out `ipython` on your comma three instead :)

PR #22226, opened 2021-09-14
Honda carstate: refactor MAIN_ON signal

belm0:
MAIN_ON handling was scattered throughout the code since
the enclosing message differed by model.  Store the message
name in the instance.

Several cases in get_can_signals() and update() were
subsequently collapsed.

### DIFF
--- a/selfdrive/car/honda/carstate.py
+++ b/selfdrive/car/honda/carstate.py
@@ -63,7 +63,6 @@ def get_can_signals(CP, gearbox_msg, main_on_sig_msg):
       ("SCM_FEEDBACK", 10),
       ("SCM_BUTTONS", 25),
     ]
-  assert any(msg == main_on_sig_msg for msg, _ in checks)
 
   if CP.carFingerprint in (CAR.CRV_HYBRID, CAR.CIVIC_BOSCH_DIESEL, CAR.ACURA_RDX_3G, CAR.HONDA_E):
     checks += [
--- a/selfdrive/car/honda/carstate.py
+++ b/selfdrive/car/honda/carstate.py
@@ -10,7 +10,7 @@ from selfdrive.car.honda.values import CAR, DBC, STEER_THRESHOLD, SPEED_FACTOR, 
 TransmissionType = car.CarParams.TransmissionType
 
 
-def get_can_signals(CP, gearbox_msg):
+def get_can_signals(CP, gearbox_msg, main_on_sig_msg):
   # this function generates lists for signal, messages and initial values
   signals = [
     ("XMISSION_SPEED", "ENGINE_DATA", 0),
@@ -38,6 +38,7 @@ def get_can_signals(CP, gearbox_msg):
     ("PEDAL_GAS", "POWERTRAIN_DATA", 0),
     ("CRUISE_SETTING", "SCM_BUTTONS", 0),
     ("ACC_STATUS", "POWERTRAIN_DATA", 0),
+    ("MAIN_ON", main_on_sig_msg, 0),
   ]
 
   checks = [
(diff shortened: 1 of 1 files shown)


selfdrive/car/honda/carstate.py
```diff
@@ -10,7 +10,7 @@
 TransmissionType = car.CarParams.TransmissionType
 
 
-def get_can_signals(CP, gearbox_msg):
+def get_can_signals(CP, gearbox_msg, main_on_sig_msg):
```
adeebshihadeh, line 13: just call it `main_on_signal`, msg is confusing because usually that refers to the address the signal is on

belm0: This variable is the message name, and the naming intends to convey that it's the name of the message containing the MAIN_ON signal.

Just as the existing `gearbox_msg` is e.g. `GEARBOX` (a message name), `main_on_sig_msg` is e.g. `SCM_FEEDBACK` (also a message name).

If it were called `main_on_signal`, then I would expect it to be a signal name such as `MAIN_ON`.

adeebshihadeh: my bad, then let's just call it `main_on_msg`

belm0: (The situation with `SCM_FEEDBACK` and `SCM_BUTTON` messages is a little unusual, because *both* are actively used within each model, with signals like `MAIN_ON` and `DRIVERS_DOOR_OPEN` coming from one or the other based on model.)

belm0: > my bad, then let's just call it main_on_msg

We agree about the `_msg` suffix, since it's a CAN message name.

I think it's useful to know whether we're referring to a message by the message name proper vs. the name of a signal contained within the message.   This little extra hint may save some confusion by a future person maintaining this code, or may reduce the need for comments.

More `get_can_signals()` args will likely be added-- please consider it.

adeebshihadeh: sounds good to me

selfdrive/car/honda/carstate.py
```diff
@@ -62,6 +63,7 @@ def get_can_signals(CP, gearbox_msg):
       ("SCM_FEEDBACK", 10),
       ("SCM_BUTTONS", 25),
     ]
+  assert any(msg == main_on_sig_msg for msg, _ in checks)
```
adeebshihadeh, line 66: can parser already does this

adeebshihadeh:
```suggestion
```

selfdrive/car/honda/carstate.py
```diff
@@ -168,6 +157,10 @@ def __init__(self, CP):
     self.gearbox_msg = "GEARBOX"
     if CP.carFingerprint == CAR.ACCORD and CP.transmissionType == TransmissionType.cvt:
       self.gearbox_msg = "GEARBOX_15T"
+    self.main_on_sig_msg = "SCM_FEEDBACK"
```
adeebshihadeh, line 160:
```suggestion

    self.main_on_sig_msg = "SCM_FEEDBACK"
```

selfdrive/car/honda/carstate.py
belm0: The intention of this assert is to confirm that the passed-in `main_on_sig_msg` is represented in the list of checks.  The CAN parser can't possibly know about our local args.

In other words, we don't want someone to pass `main_on_sig_msg="FOO"` later on, and have no check that "FOO" messages are occurring.

We can't naively append `checks += [main_on_sig_msg, 10]`, because both `SCM_FEEDBACK` and `SCM_BUTTONS` need to have checks, and the rates depend on model-- see above.

So I think this assert is relevant and concisely provides some safeguard.

adeebshihadeh: the can parser doesn't need to know about the args, it uses the list of signals. i.e. if you pass in FOO, it'll complain that FOO isn't in checks. see https://github.com/commaai/opendbc/blob/master/can/parser_pyx.pyx#L81

belm0: oh, that's great-- thank you for explaining!

adeebshihadeh:
Nice, another car state cleanup :)

PR #22228, opened 2021-09-14
c++ replay: thread-safe seek

deanlee:
1. thread-safe seek
2. move the keyboard thread out of Replay.


adeebshihadeh:
Why move the keyboard thread?

deanlee:
to make class Replay more generic.   we may use replay in other places or call seekTo in other ways

PR #22232, opened 2021-09-14
Add 2022 Corolla Hatchback Engine FP

d412k5t412:


adeebshihadeh:
Which route did this FW come from?

d412k5t412:
Used fw_versions.py to get the FP. Couldn't get it from logs. Successful route edf70cea06b79900|2021-09-14--17-30-07

PR #22236, opened 2021-09-15
Use thneed directly on the loaded YUV data

geohot:
Avoids making 2 needless copies of the image, one in `ModelFrame::prepare` and one (of both images) in `Thneed::copy_inputs` 

This also removes the clFinish from `ModelFrame::prepare`, should work if thneed is using the same queue.

This is completely untested. I don't have a set up device on my desk. I suspect the first pass is wrong.

Two more ways to improve:
1. Put the preparation steps into thneed itself to avoid more enqueues.
2. Remove the clEnqueueCopyBuffer by switching the first layer weights to accept the images in either order.

### DIFF
--- a/selfdrive/modeld/models/commonmodel.cc
+++ b/selfdrive/modeld/models/commonmodel.cc
@@ -22,16 +22,23 @@ ModelFrame::ModelFrame(cl_device_id device_id, cl_context context) {
   loadyuv_init(&loadyuv, context, device_id, MODEL_WIDTH, MODEL_HEIGHT);
 }
 
-float* ModelFrame::prepare(cl_mem yuv_cl, int frame_width, int frame_height, const mat3 &transform) {
+float* ModelFrame::prepare(cl_mem yuv_cl, int frame_width, int frame_height, const mat3 &transform, cl_mem *output) {
   transform_queue(&this->transform, q,
                   yuv_cl, frame_width, frame_height,
                   y_cl, u_cl, v_cl, MODEL_WIDTH, MODEL_HEIGHT, transform);
-  loadyuv_queue(&loadyuv, q, y_cl, u_cl, v_cl, net_input_cl);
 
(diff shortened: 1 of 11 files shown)


geohot:
Hmm, it's wrong because ModelFrame is using a different command queue from thneed.

geohot:
Seems correct now, though it's laggy on my desk (with or without this change). Let's see what CI says.

And with openpilot running, it seems like it saves over a millisecond!

selfdrive/modeld/models/commonmodel.cc
```diff
@@ -22,16 +22,23 @@ ModelFrame::ModelFrame(cl_device_id device_id, cl_context context) {
...
+  if (output == NULL) {
+    loadyuv_queue(&loadyuv, q, y_cl, u_cl, v_cl, net_input_cl, 0);
+
+    std::memmove(&input_frames[0], &input_frames[MODEL_FRAME_SIZE], sizeof(float) * MODEL_FRAME_SIZE);
+    CL_CHECK(clEnqueueReadBuffer(q, net_input_cl, CL_TRUE, 0, MODEL_FRAME_SIZE * sizeof(float), &input_frames[MODEL_FRAME_SIZE], 0, nullptr, nullptr));
+    clFinish(q);
+    return &input_frames[0];
+  } else {
+    loadyuv_queue(&loadyuv, q, y_cl, u_cl, v_cl, *output, MODEL_FRAME_SIZE);
+    clFinish(q);
```
geohot, line 39: You still have to clFinish because thneed is on a different command queue.

I tried piping the ModelFrame command queue into thneed, but it doesn't work. Once you send direct ioctls to the driver, the nonce is out of sync. While this is fixable, it's probably better to just put the ModelFrame::prepare inside thneed.

geohot:
If that CI passes, this should be good to merge

geohot:
@adeebshihadeh any more tests to do, or you think this is good to merge? Maybe run for a per hours in testing closet?

adeebshihadeh:
The CI tests should be sufficient here. Looking into the lags on the UI side.

PR #22203, opened 2021-09-12
refactor c++ replay

deanlee:
- [x]  cache route files
- [x]  skip invalid/missing segments.
- [x]  read frames from all cameras.(fall back to qcamera)
- [x]  thread safe merging segments in correct order
- [x] cancel downloading & remove segments in time
- [x] merged #22239 

resolve #22221 :
1. fallback to qcams, ideally we don't require video either
2. stop replay while waiting for video to load (the time is still progressing)
3.  work with only 1 segment routes
4. show error when route is not found
5. sometimes replay starts in the second segment
6. add optional argument for start time
7. segfaults randomly

the remain issues related with camera will be resolved in #22195 
seeking related issues will be fixed in #22319

PR #22239, opened 2021-09-15
c++ replay: fast and reliable merge events

deanlee:
fast and reliable merge events with std::vector. sort events by mono_time and which, the time for merge and sort 5 segments is less than 4ms.

QMultiMap is not suitable for events. it can't guarantee that the events are sent in the correct order without duplicates （mono_time is not UUID), it also takes up more memory and  merging  time. QMultipMap::lower_bound  will also cause the last event to be sent again.

### DIFF
--- a/selfdrive/ui/replay/replay.cc
+++ b/selfdrive/ui/replay/replay.cc
@@ -166,7 +166,7 @@ void Replay::stream() {
     if (!next_evt) {
       lk.unlock();
       qDebug() << "waiting for events";
-      QThread::msleep(10);
+      QThread::msleep(100);
       continue;
     }
     seek_ts = -1;
--- a/selfdrive/ui/replay/replay.cc
+++ b/selfdrive/ui/replay/replay.cc
@@ -154,7 +154,7 @@ std::optional<std::vector<Event*>::iterator> Replay::nextEvent(cereal::Event::Wh
 }
 
 void Replay::stream() {
-  route_start_ts = 0;
+  float last_print = 0;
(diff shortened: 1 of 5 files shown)


adeebshihadeh:
Nice, once we get these bugs fixed we can delete unlogger. Another pretty bad bug is that replay doesn't work for routes with less segments that replay's window (i.e. only first segment uploaded).

deanlee:
that bug will be fixed in #22203

selfdrive/ui/replay/replay.cc
```diff
@@ -63,13 +63,18 @@ void Replay::mergeEvents() {
   const int end_idx = std::min(current_segment + FORWARD_SEGS, log_paths.size());
 
   // merge logs
-  QMultiMap<uint64_t, Event *> *new_events = new QMultiMap<uint64_t, Event *>();
+  std::vector<Event *> *new_events = new std::vector<Event *>();
   std::unordered_map<uint32_t, EncodeIdx> *new_eidx = new std::unordered_map<uint32_t, EncodeIdx>[MAX_CAMERAS];
   for (int i = start_idx; i <= end_idx; ++i) {
     if (auto it = lrs.find(i); it != lrs.end()) {
```
deanlee, line 69: here is a historic bug. LogerrReader may be downloading and updating events in another thread at this point. that will cause invalid data to be merged, and in the worst case, the program may crash.

deanlee: fixed in ##22203

PR #22241, opened 2021-09-15
Hyundai radar parser

### LINKED ISSUE #1294: Add support for Hyundai radar -- $500 bounty

pd0wm:
See `selfdrive/debug/hyundai_enable_radar_points.py` to enable radar points on some cars.

> Some Hyundai radars can be reconfigured to output (debug) radar points on bus 1.
> Reconfiguration is done over UDS by reading/writing to 0x0142 using the Read/Write Data By Identifier
> endpoints (0x22 & 0x2E). This script checks your radar firmware version against a list of known 
> firmware versions. If you want to try on a new radar make sure to note the default config value
> in case it's different from the other radars and you need to revert the changes.
> 
> After changing the config the car should not show any faults when openpilot is not running. 
> These config changes are persistent accross car reboots. You need to run this script again
> to go back to the default values.
> 
> USE AT YOUR OWN RISK! Safety features, like AEB and FCW, might be affected by these changes.

Looks like some cars already have this enabled by default. A quick check on our test routes showed the following cars:
- ~KIA_SORENTO_GT_LINE_2018~ Wrong format
- HYUNDAI_IONIQ_ELECTRIC_LIMITED_2019
- KIA_NIRO_HYBRID_2019
- GENESIS_G80_2017
- ~HYUNDAI_SONATA_2019~ Wrong format

### DIFF
--- a/selfdrive/car/hyundai/radar_interface.py
+++ b/selfdrive/car/hyundai/radar_interface.py
@@ -6,12 +6,15 @@ from opendbc.can.parser import CANParser
 from selfdrive.car.interfaces import RadarInterfaceBase
 from selfdrive.car.hyundai.values import DBC
 
+RADAR_START_ADDR = 0x500
+RADAR_MSG_COUNT = 32
+
 
 def get_radar_can_parser(CP):
   signals = []
   checks = []
 
-  for addr in range(0x500, 0x500 + 32):
+  for addr in range(RADAR_START_ADDR, RADAR_START_ADDR + RADAR_MSG_COUNT):
     msg = f"R_{hex(addr)}"
     signals += [
       ("NEW_SIGNAL_2", msg, 0),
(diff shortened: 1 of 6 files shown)


geohot:
Haha you and Greg are going to win the #1294 bounty

selfdrive/car/hyundai/radar_interface.py
```diff
@@ -46,25 +57,34 @@ def update(self, can_strings):
 
   def _update(self, updated_messages):
     ret = car.RadarData.new_message()
-    cpt = self.rcp.vl
+    if self.rcp is None:
+      return ret
```
adeebshihadeh, line 61: this needed? already checked in update

selfdrive/car/hyundai/radar_interface.py
```diff
@@ -46,25 +57,34 @@ def update(self, can_strings):
...
-          self.pts[ii] = car.RadarData.RadarPoint.new_message()
-          self.pts[ii].trackId = self.track_id
-          self.track_id += 1
-        self.pts[ii].dRel = cpt["SCC11"]['ACC_ObjDist']  # from front of car
-        self.pts[ii].yRel = -cpt["SCC11"]['ACC_ObjLatPos']  # in car frame's y axis, left is negative
-        self.pts[ii].vRel = cpt["SCC11"]['ACC_ObjRelSpd']
-        self.pts[ii].aRel = float('nan')
-        self.pts[ii].yvRel = float('nan')
-        self.pts[ii].measured = True
+    for addr in range(0x500, 0x500 + 32):
```
adeebshihadeh, line 69:
```suggestion
    for addr in range(RADAR_START_ADDR, RADAR_START_ADDR + RADAR_MSG_COUNT):
```

selfdrive/car/hyundai/radar_interface.py
pd0wm: Needed for the car interface test

PR #22251, opened 2021-09-16
Honda carstate: refactor door open signals

belm0:
There were varying number of doors, signal names, and message
names.  Make it data-driven.


selfdrive/car/honda/carstate.py
```diff
@@ -105,23 +105,15 @@ def get_can_signals(CP, gearbox_msg, main_on_sig_msg):
...
-    ]
+  signals.extend((signal, door_sig_msg, 1) for signal in door_signals)
+
+  if CP.carFingerprint not in HONDA_NO_WHEELS_MOVING_SIGNAL:
+    signals += [("WHEELS_MOVING", "STANDSTILL", 1)]
+    if CP.carFingerprint != CAR.HRV:
+      checks += [
+        ("DOORS_STATUS", 3),
+        ("STANDSTILL", 50),
+      ]
```
belm0, line 116: I *think* there should not be an HRV exception for the STANDSTILL check.

And we should have separate sets for cars without WHEELS_MOVING and without DOORS_STATUS.

But anyway I left the original logic intact for now.

belm0: certainly HRV hits this code path, so it should not have the STANDSTILL check exception
```python
ret.standstill = not cp.vl["STANDSTILL"]["WHEELS_MOVING"]
```

HRV only happens to work because STANDSTILL check is added again later:
```python
  if CP.openpilotLongitudinalControl:
    ...
    checks += [("STANDSTILL", 50)]
```

selfdrive/car/honda/values.py
```diff
@@ -1302,4 +1302,5 @@ class CAR:
 
 HONDA_NIDEC_ALT_PCM_ACCEL = set([CAR.ODYSSEY])
 HONDA_BOSCH = set([CAR.ACCORD, CAR.ACCORDH, CAR.CIVIC_BOSCH, CAR.CIVIC_BOSCH_DIESEL, CAR.CRV_5G, CAR.CRV_HYBRID, CAR.INSIGHT, CAR.ACURA_RDX_3G, CAR.HONDA_E])
-HONDA_BOSCH_ALT_BRAKE_SIGNAL = set([CAR.ACCORD, CAR.CRV_5G, CAR.ACURA_RDX_3G])
\ No newline at end of file
+HONDA_BOSCH_ALT_BRAKE_SIGNAL = set([CAR.ACCORD, CAR.CRV_5G, CAR.ACURA_RDX_3G])
+HONDA_NO_WHEELS_MOVING_SIGNAL = {CAR.ACCORD, CAR.ACCORDH, CAR.CIVIC_BOSCH, CAR.CIVIC_BOSCH_DIESEL, CAR.CRV_HYBRID, CAR.INSIGHT, CAR.ACURA_RDX_3G, CAR.HONDA_E, CAR.ODYSSEY_CHN}
```
belm0: Instead of these lists, somehow I'd like to represent the setting in car params, so that the configuration of each model can be centralized in `interface.py`.

something like:
```python
  class HondaDetails(NamedTuple):
    nidec_alt_pcm_accel = False
    is_bosch = False
    ...

  def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):
    ret = CarInterfaceBase.get_std_params(candidate, fingerprint)
    ret.carName = "honda"
    ret.details = HondaDetails()

    ...
    elif candidate == CAR.ODYSSEY:
      ret.details.nidec_alt_pcm_accel = True
      ...
      

```

adeebshihadeh:
Can you rebase?

selfdrive/car/honda/values.py
belm0: the new `flags` field in CarParams would allow this, but I'll try it in a separate PR

belm0:
@adeebshihadeh any concerns with this PR?

I've had to rebase it about 4 times...

selfdrive/car/honda/values.py
```diff
@@ -1444,3 +1444,5 @@ class CAR:
 HONDA_BOSCH = set([CAR.ACCORD, CAR.ACCORD_2021, CAR.ACCORDH, CAR.ACCORDH_2021, CAR.CIVIC_BOSCH, CAR.CIVIC_BOSCH_DIESEL, CAR.CRV_5G,
                    CAR.CRV_HYBRID, CAR.INSIGHT, CAR.ACURA_RDX_3G, CAR.HONDA_E])
 HONDA_BOSCH_ALT_BRAKE_SIGNAL = set([CAR.ACCORD, CAR.ACCORD_2021, CAR.CRV_5G, CAR.ACURA_RDX_3G])
+HONDA_NO_WHEELS_MOVING_SIGNAL = {CAR.ACCORD, CAR.ACCORD_2021, CAR.ACCORDH, CAR.ACCORDH_2021, CAR.CIVIC_BOSCH, CAR.CIVIC_BOSCH_DIESEL,
+                                 CAR.CRV_HYBRID, CAR.INSIGHT, CAR.ACURA_RDX_3G, CAR.HONDA_E, CAR.ODYSSEY_CHN}
```
adeebshihadeh, line 1448:
```suggestion
HONDA_NO_WHEELS_MOVING_SIGNAL = set([CAR.ACCORD, CAR.ACCORD_2021, CAR.ACCORDH, CAR.ACCORDH_2021, CAR.CIVIC_BOSCH, CAR.CIVIC_BOSCH_DIESEL,
                                 CAR.CRV_HYBRID, CAR.INSIGHT, CAR.ACURA_RDX_3G, CAR.HONDA_E, CAR.ODYSSEY_CHN])
```

adeebshihadeh:
@sshane let's try to find a common signal

adeebshihadeh:
I don't think this minor cleanup is worth resolving the conflicts for. @sshane will clean this up as part of the Honda DBC cleanup project.

PR #22263, opened 2021-09-17
Qt/ui: update vipc in thread

### LINKED ISSUE #22257: C3: UI state updates at 10Hz
Camera view seems fine, but model and other state only updates at 10Hz

deanlee:
fix UI state updates at 10Hz while onroad. resolved  #22257

### DIFF
--- a/selfdrive/ui/ui.h
+++ b/selfdrive/ui/ui.h
@@ -117,7 +117,9 @@ typedef struct UIScene {
   uint64_t started_frame;
 } UIScene;
 
+class QUIState;
 typedef struct UIState {
+  QUIState *qUIState = nullptr;
   int fb_w = 0, fb_h = 0;
   NVGcontext *vg;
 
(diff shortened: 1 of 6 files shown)


selfdrive/ui/ui.h
```diff
@@ -117,7 +117,9 @@ typedef struct UIScene {
   uint64_t started_frame;
 } UIScene;
 
+class QUIState;
```
pd0wm, line 120: Can we just remove the `typedef struct`? That's a leftover from the original c days.

adeebshihadeh:
Can we move the camera view updating to another thread instead?

deanlee:
moved vipc to thread.it doesn't block the main thread now.

deanlee:
need to stop "try to connect or recv again" when offroad. otherwise it will  loop infinite

deanlee:
there is a resource leak in VisionIpcClient, fixed in   https://github.com/commaai/cereal/pull/201

deanlee:
moved to #22354

PR #22280, opened 2021-09-19
c++ replay: graceful shutdown

deanlee:


pd0wm:
I'm still seeing some camerad server shutdown messages after the replay destructor. Is that intended?

deanlee:
it's ok now.release cameraserver before print "shoutdown: done"

PR #22285, opened 2021-09-20
agnos 1.6

adeebshihadeh:


ErichMoraga:
Any idea when this will be merged?  I've been testing it for a couple days, and it's been quite stable.

adeebshihadeh:
This won't get merged. It was only to test an agnos-builder refactor.

PR #22287, opened 2021-09-20
fix yuv420_to_jpeg: thumbnail_width & thumbnail_height must be aliged with 16 pixel.

deanlee:
from document: https://github.com/libjpeg-turbo/libjpeg-turbo/blob/main/libjpeg.txt
`The MCU height is max_v_samp_factor = 2 DCT rows so you must pass at least 16 scanlines on each call to jpeg_write_raw_data()`
 It will cause illegal memory violation  if the height is not aligned with 16pixel.

sorry for this silly mistake.


pd0wm:
This would change the output size of the image, right? I don't think we want to make that change.

In that document they talk about padding the input data, is that something we can do?

deanlee:
we can make the input buffer big enough with zero filled.this can also fix the problem

pd0wm:
If that generates the same jpegs as before the switch to yuv that sounds good to me.

deanlee:
it's the same as before

PR #22293, opened 2021-09-21
c++ replay: add some test cases

deanlee:


MSH2005P:
How test app with this library ?? 
Are the app any for python ?!

deanlee:
> How test app with this library ??

scons --test
run replay/test_replay

PR #22308, opened 2021-09-22
c++ replay: chunking and concurrent downloads

deanlee:


deanlee:
added CURLlGlobalInitializer
> If you did not already call curl_global_init(3), curl_easy_init(3) does it automatically. This may be lethal in multi-threaded cases, since curl_global_init(3) is not thread-safe, and it may result in resource problems because there is no corresponding cleanup.

pd0wm:
Video looks strange when running from qcameras:

2021-09-23-224108_1905x1076_scrot

deanlee:
how about  #22195? I think it can handle switch between qcamera and normal video

pd0wm:
Sorry, this comment was not supposed to be in this PR. It's happening on master and also on 22195. Let me move it into the appropriate issue.

PR #22319, opened 2021-09-23
c++ replay: improve seeking&updating events

deanlee:
1. improve robustness of seeking&updating events.
2. fix seeking stuck while segments is loading. replay can continually seek in the entire route now.
3. cancel downloads and release segments immediately if they are not in the current segment window.


deanlee:
lgtm. the seeking and upating is much better. 
since the FrameReader needs to decode the previous 15 frames of the current frame 。 the video will have a delay of 300-500ms after seeking. this will be resolved in #22217.

adeebshihadeh:
Nice, I think we're almost ready to delete unlogger :)

Any other known bugs in replay? Feel free to list them in the "Move to C++ unlogger" issue.

deanlee:
I haven't found other bugs yet. need to write more test cases to ensure that.

adeebshihadeh:
I got into a state where the replay kept going, but it wouldn't seek anywhere else despite the seek input working (was at 1141 seconds into the demo route if that helps).

deanlee:
I found the problem. I should not set `events_updated_` to true in `seekto()` when segment changed.

deanlee:
> I got into a state where the replay kept going, but it wouldn't seek anywhere else despite the seek input working (was at 1141 seconds into the demo route if that helps).
> 
> Edit: this is happening consistently. Repro: `./replay --demo`, let it start, seek to 1000, let it seek, seek to 0, won't go to 0.

fixed this issue

deanlee:
fixed this issue

PR #22323, opened 2021-09-23
Temp Good

Achilles308:
This fixes an annoying variation between the temp, connect, and panda status fields in the UI of openpilot.

GOOD
TEMP

is now

TEMP
GOOD

which is more inline with the better structure of

CONNECT (item of interest)
ONLINE (condition)


cydia2020:
can you change "NO PANDA" to "VEHICLE OFFLINE"?

Achilles308:
> can you change "NO PANDA" to "VEHICLE OFFLINE"?

I wracked my brain to come up with a better wording for these... I'm afraid I couldn't come up with a better one

Maybe:

PANDA
OFF

PANDA
DISCONNECTED

PANDA
UNREPSONSIVE

I think it'll run into limited characters to fit in the bounded area.

VEHICLE
OFFLINE

Is that the correct description of what it is though?

PR #22324, opened 2021-09-23
VW MQB: Min steer speed support, other cleanup

jyoung8607:
**Description**

Sadly, VW Transporter (#22283) turns out to have a minimum steer speed. This was a real surprise. No other MQB racks have one, and most contemporaneous MQBs have Traffic Jam Assist and/or Travel Assist, which bring factory Lane Assist down to zero.

Added the relevant checks and events. While here, added some missing error handling, and cleaned up redundant comments.
 
**Verification**

Test driven by the affected owner.

**Route**

`a459f4556782eba1|2021-09-24--08-47-54`

### DIFF
--- a/selfdrive/car/volkswagen/carcontroller.py
+++ b/selfdrive/car/volkswagen/carcontroller.py
@@ -39,7 +39,7 @@ class CarController():
       # torque value. Do that anytime we happen to have 0 torque, or failing that,
       # when exceeding ~1/3 the 360 second timer.
 
-      if enabled and not (CS.out.standstill or CS.out.steerError or CS.out.steerWarning):
+      if enabled and CS.out.vEgo > CS.CP.minSteerSpeed and not (CS.out.standstill or CS.out.steerError or CS.out.steerWarning):
         new_steer = int(round(actuators.steer * P.STEER_MAX))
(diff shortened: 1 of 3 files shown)


selfdrive/car/volkswagen/carcontroller.py
```diff
@@ -39,7 +39,7 @@ def update(self, enabled, CS, frame, ext_bus, actuators, visual_alert, left_lane
       # torque value. Do that anytime we happen to have 0 torque, or failing that,
       # when exceeding ~1/3 the 360 second timer.
 
-      if enabled and not (CS.out.standstill or CS.out.steerError or CS.out.steerWarning):
+      if enabled and CS.out.vEgo > CS.CP.minSteerSpeed and not (CS.out.standstill or CS.out.steerError or CS.out.steerWarning):
```
jyoung8607, line 42: `CS.out.vEgo > CS.CP.minSteerSpeed` where minSteerSpeed == default of 0 is almost the same thing as `CS.out.standstill`, but I'd like to keep both conditions.

In the early days of the port, I used `vEgo < 0.01` as my standstill condition and had problems with EPS faulting coming to a standstill until I raised it to `vEgo < 0.1` (since replaced with the EPS hydraulic hold flag).

jyoung8607:
Test completed, ready for review.

PR #22344, opened 2021-09-24
Improve Hyundai FW query stability

pd0wm:


selfdrive/car/fw_versions.py
```diff
@@ -42,15 +42,12 @@ def p16(val):
   p16(uds.DATA_IDENTIFIER_TYPE.APPLICATION_SOFTWARE_IDENTIFICATION)
 
 
-HYUNDAI_VERSION_REQUEST_SHORT = bytes([uds.SERVICE_TYPE.READ_DATA_BY_IDENTIFIER]) + \
```
sshane, line 45: Looks like this request might have been used to get the Optima's eps FW. We get `0x11 - service not supported` with both our requests now @pd0wm

PR #22345, opened 2021-09-24
Mazda: Cruise Speed

Jafaral:
Finally found the cruise speed signal!
Depends on:  https://github.com/commaai/opendbc/pull/430


selfdrive/car/mazda/carstate.py
```diff
@@ -64,15 +64,9 @@ def update(self, cp, cp_cam):
     elif speed_kph < LKAS_LIMITS.DISABLE_SPEED:
       self.lkas_allowed = False
 
-    # if any of the cruize buttons is pressed force state update
```
adeebshihadeh, line 67: Nice, glad to remove this :)

PR #22348, opened 2021-09-25
c++ replay:  more accurate sleep

deanlee:
QThread::usleep is quite inaccurate. use `nanosleep` and spin lock for more accurate sleeping


deanlee:
the CPU usage increased by 2%~5%, but it’s worth it

adeebshihadeh:
That's fine, but let's keep in mind that we want to be able to run this on C2 and C3 at some point too.

deanlee:
> That's fine, but let's keep in mind that we want to be able to run this on C2 and C3 at some point too.
Ok i will pay attention to this. 

precise_nano_sleep should be ok on c2 &c3

PR #22349, opened 2021-09-25
use TSS2 long tuning on all Toyotas

pd0wm:


ntegan1:
all toyotas without a pedal :(

selfdrive/car/toyota/interface.py
```diff
@@ -320,8 +320,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.longitudinalTuning.kpV = [1.2, 0.8, 0.765, 2.255, 1.5]
```
cydia2020, line 320: Need to change here as well.

cydia2020:
Just tested this, doesn't work well on my Prius 2018, constantly overshoots and undershoots

pd0wm:
Do you have a drive I can look at? With or without pedal? Still need to update the tuning for the pedal.

cydia2020:
> Do you have a drive I can look at? With or without pedal? Still need to update the tuning for the pedal.

Pedal, but min acc speed is 5m/s for me on my fork so it shouldn't matter, i only left the TSS2.0 long tuning for the test, so I wasn't using the pedal tuning for the test, so for the oscillation issue I think we can ignore the pedal for now.

I'll put a drive here hopefully by tomorrow.

PR #22358, opened 2021-09-27
VW MQB: Fix min steer speed alerts

jyoung8607:
Fix logic error introduced in #22324: don't show low-speed alerts when stopping for cars that steer down to zero.


adeebshihadeh:
Do we also need to fix this in the car controller?

jyoung8607:
I believe the logic there is correct. The EPS doesn't like to be steered *at* zero, but anywhere above zero seems fine. During the drive I discovered this problem, I didn't have any actual EPS faults or anything, just superfluous openpilot low-steer-speed alerts. I commented on the intentional check for both `standstill` and `> minSteerSpeed` in #22324.

Issue #22095, opened 2021-08-31
Logging/loggerd requirements for training


HaraldSchafer:
- [ ] Encoder/Loggerd should keep up with camera streams such that no frames are dropped there, if there are, this should raise a flag
- [ ] All segments should have synchronized camera streams on C3 (i.e. all frames are from the same time)
- [ ] All segments should have same number of frames in 3 camera streams for C3
- [ ] All segments should have 1200 frames
- [ ] If the camera drops a frame, a blank frame should be inserted and a flag raised

Training Data-Quality Checks
- Logging time lags : abs(`roadEncodeIdx/timestampSof` - `controlsState/t`) < 1 sec
- Camera FPS check : Frame diffs of `<camera>EncodeIdx/timestampEof` are between 0.045 and 0.055 secs
- Inter Camera Lag : Lag between same frame of different cameras < 0.005

deanlee:
>     * [ ]  If the camera drops a frame, a blank frame should be inserted and a flag raised
It is very difficult to do this perfectly in loggerd. and will greatly increase the complexity and performance overhead.
It may be more appropriate to do this in other places, such as framereader

adeebshihadeh:
@nuwandavek can you expand the list above with all the checks we do offline?

gregjhogan:
> > ```
> > * [ ]  If the camera drops a frame, a blank frame should be inserted and a flag raised
> > ```
> 
> It is very difficult to do this perfectly in loggerd. and will greatly increase the complexity and performance overhead. I think it may be more appropriate to do this in other places, such as framereader or replay

I really want blank frames to easily keep video playback in connect and cabana always in sync with the log data, but now that I think about it more it seems kind of impractical.  For example, an easy case to detect and try to compensate for is when the encoder fails to write out a frame.  But how do we insert a blank frame in?  I assume we have to shove another frame through the encoder which may fail again.  We also don't put anything into the qlog/rlog when a frame is dropped, but again I suppose that can fail, too.

jpc:
Hi, I work at Collabora and we invest a lot of work into improving GStreamer – the real-time multimedia framework. Recently we were researching into how people are doing data sampling when uploading videos from embedded devices.

One thing we are curious about is whether it would be useful to have Open Source GStreamer components that could automate the process of: getting the camera data, accurately time-stamping it, running it through a (ONNX?) DL model, compressing it and uploading/streaming it over the internet based on the output from the model.

[...]

PR #22361, opened 2021-09-28
replay: send frames based on encodeIdx packet

pd0wm:
This should simplify things a bit. @deanlee what do you think?

### DIFF
--- a/selfdrive/ui/replay/replay.cc
+++ b/selfdrive/ui/replay/replay.cc
@@ -232,40 +232,43 @@ void Replay::stream() {
         }
 
         // publish frame
-        // TODO: publish all frames
-        if (evt->which == cereal::Event::ROAD_ENCODE_IDX) {
-          auto idx = evt->event.getRoadEncodeIdx();
-          auto &seg = segments_[idx.getSegmentNum()];
+        if (evt->frame) {
+          // TODO: publish all frames
+          if (evt->which == cereal::Event::ROAD_ENCODE_IDX) {
+            auto idx = evt->event.getRoadEncodeIdx();
+            auto &seg = segments_[idx.getSegmentNum()];
 
-          if (seg && seg->isLoaded() && idx.getType() == cereal::EncodeIndex::Type::FULL_H_E_V_C) {
-            auto &frm = seg->frames[RoadCam];
+            if (seg && seg->isLoaded() && idx.getType() == cereal::EncodeIndex::Type::FULL_H_E_V_C) {
+              auto &frm = seg->frames[RoadCam];
 
-            if (vipc_server == nullptr) {
-              cl_device_id device_id = cl_get_device_id(CL_DEVICE_TYPE_DEFAULT);
-              cl_context context = CL_CHECK_ERR(clCreateContext(NULL, 1, &device_id, NULL, NULL, &err));
+              if (vipc_server == nullptr) {
+                cl_device_id device_id = cl_get_device_id(CL_DEVICE_TYPE_DEFAULT);
+                cl_context context = CL_CHECK_ERR(clCreateContext(NULL, 1, &device_id, NULL, NULL, &err));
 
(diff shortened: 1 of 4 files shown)


deanlee:
lgtm. the only problem is that the missing frames cannot be traced after this.are we going to do some special processing for lost frames in the future? for example, output a blank frame

pd0wm:
What do you mean by missing frames? A frame that has an encodeIdx packet but is not in the hevc?

deanlee:
> What do you mean by missing frames? A frame that has an encodeIdx packet but is not in the hevc?

yes

pd0wm:
Where do we still call `getRoadCameraState`? I think I made everything driven by encodeIdx now.

deanlee:
Sorry for my bad expression, I mean, do we need to consider the situation where the frame does not has a corresponding encodeid? and do some special processing in the future: #22095

selfdrive/ui/replay/replay.cc
```diff
@@ -201,6 +194,13 @@ void Replay::stream() {
       cur_which = evt->which;
       cur_mono_time_ = evt->mono_time;
 
+      if (evt->which == cereal::Event::ROAD_ENCODE_IDX) {
+        uint64_t sof = evt->event.getRoadEncodeIdx().getTimestampSof();
+        if (sof > 0) {
+          cur_mono_time_ = sof;
```
deanlee, line 200: if we send frame based on encode_idx's mono_time, the frame will be sent later than actual time.

deanlee: and I think it’s not safe to modify cur_mono_time_ here.  it will make cur_mono_time_ old, the events sent before may be re-sent again.

deanlee: I personally recommend still using ROAD_CAMERA_STATE to send frames

pd0wm:
Think it's fine for replay if we only show frames that are actually in the video file for now. We will deal with dropped frames later.

selfdrive/ui/replay/replay.cc
pd0wm: Looks like C2 has EOF set, so I'll use that as will if set.

pd0wm: Would it be fine if I user evt_start_ts? Since that only seems to affect the sleeps?

pd0wm: moved the timestamp handling into the Event clas

pd0wm:
After this change the demo route doesn't perform very well since it doesn't have eof/sof timestaps set correctly. This is no longer a problem on recent routes. Shouldn't be a problem since we will replace the demo route anyway by something that has `modelV2` packets.

deanlee:
I suggest to set the ENCODE_IDX's mono_time to SOF to ensure the orders

pd0wm:
Do you want to handle this outside of the event class? We still need to check for both eof and sof. C2 only has eof set, and in earlier routes none of them are set.

deanlee:
How about put aside this simplification for now, back later and continue to do it?

pd0wm:
This change is necessary for replay to work on qlogs, since they don't contain full frequency cameraState packets.

deanlee:
hmmm...OIC

pd0wm:
Let's merge this for now, but the real solution is probably to insert a special event type that only handles sending the frame data over visionIPC. That should allow us to send the encodeIdx packet at the correct logMonoTime, while also maintaining timing accuracy of the video.

deanlee:
@pd0wm I have submitted a new PR #22364 after read your code. insert new custom encode events  in `logreader`.  the new event's memory will be handled by logreader as before.

deanlee:
@pd0wm It seems that there are duplicate eidx in the dmeo route, each frame has two identical eidx

pd0wm:
It shouldn't since it checks for the type to be HEVC before publishing. We should really replace the demo route with something more recent though.

PR #22362, opened 2021-09-28
replay: fall back to qlog

pd0wm:
requires: https://github.com/commaai/openpilot/pull/22361

@adeebshihadeh do you want to do this silently or require confirmation/command line flag like plotjuggler?


adeebshihadeh:
I like the juggle script's confirmation. juggle.py doesn't do this, but let's make the command line flag use rlogs when possible and fill in the gaps with qlogs.

PR #22364, opened 2021-09-28
c++ replay: send frames based on custom encodeIdx packet

deanlee:


deanlee:
there are no ROAD_ENCODE_IDX in the qlogs of demo_route and (a8e8bf6a3864361b|2021-01-04--03-01-18). I don't know why.

pd0wm:
They are present on more recent routes. It's fine if replay on qlog doesn't work with old routes.

selfdrive/ui/replay/logreader.cc
```diff
@@ -38,33 +38,68 @@ bool LogReader::load(const std::string &file) {
...
+  auto insertEidx = [&](CameraType type, const cereal::EncodeIndex::Reader &reader, const Event* evt) {
+    eidx_map[type][reader.getFrameId()] = {evt, reader};
   };
 
   kj::ArrayPtr<const capnp::word> words((const capnp::word *)raw_.data(), raw_.size() / sizeof(capnp::word));
   while (words.size() > 0) {
     try {
       std::unique_ptr<Event> evt = std::make_unique<Event>(words);
       switch (evt->which) {
+        case cereal::Event::ROAD_CAMERA_STATE:
```
pd0wm, line 52: I don't think we need all this extra tracking of the camera state timestamp. In the rare cases that neither EOF or SOF are set we can just use the logMonoTime of the encode packet.

pd0wm:
Also saw some weird skipping around, so decided to steal your idea of a isCustomEvent flag on the event

PR #22366, opened 2021-09-28
car sort

crwusiz:
# hyundai / kia / genesis

fw2.0 sort
# fwdRadar, fwdCamera, eps, esp, engine, transmission

ceed 0x7d4 esp -> eps fix


adeebshihadeh:
We're not gonna merge something like this. It's hard to verify for a small benefit. The fingerprints will move to a data file somewhat soon. However, the code sorting is worth doing (i.e. the DBC section). Feel free to submit a new PR with just code sorted.

PR #22367, opened 2021-09-28
c++ replay: get socket name from capnp::Schema

deanlee:


selfdrive/ui/replay/replay.cc
```diff
@@ -201,12 +204,7 @@ void Replay::stream() {
       cur_which = evt->which;
       cur_mono_time_ = evt->mono_time;
 
-      std::string type;
-      KJ_IF_MAYBE(e_, static_cast<capnp::DynamicStruct::Reader>(evt->event).which()) {
-        type = e_->getProto().getName();
-      }
-
-      if (socks.find(type) != socks.end()) {
+      if (sockets_[cur_which] != nullptr) {
```
adeebshihadeh, line 207: this will break if we remove a field from the event struct. can you handle the case where the socket isn't in sockets_?

PR #22370, opened 2021-09-29
tools: update demo route

adeebshihadeh:


adeebshihadeh:
@mitchellgoffpc will record a good route tomorrow

PR #22371, opened 2021-09-29
agnos updater: support non-sparse images

adeebshihadeh:


pd0wm:
LGTM. The only change is that it now prints the progress, right?

adeebshihadeh:
And checks the raw hash. Basically, everything goes through the old sparse code path now.

pd0wm (approved):

pd0wm:
Guess those used to be checked on swap, but won't anymore for large partitions.

PR #22374, opened 2021-09-29
windows setup instructions

adeebshihadeh:


adeebshihadeh:
@sshane @jyoung8607 @grekiki I believe all of you guys use Windows primarily. Anything else worth mentioning here?

jyoung8607:
For pure CLI stuff it's solid, works out of the box like real Ubuntu. Build, run, CI-style tests are all good AFAIK.

GUI stuff is more complex. By some black magic I've never looked into, `webbrowser.open` in `tools/lib/auth.py` fires off the _Windows_-side browser and then somehow DTRT with the token. That part just works. But you have to roll your own X server setup to have plotjuggler or UI, or use the *extremely* new native GUI capabilities in the Preview builds. I've used the native stuff and IMO it's not there yet, lots of visual artifacts with plotjuggler.

I'm not sure there's anything to add to openpilot's documentation except perhaps a pointer to a decent quality X server setup guide. I need to do this on another laptop soon and will open a PR if I find a guide I like.

grekiki:
image
Works fine on win11 here :) No X server setup needed, just run plotjuggler with the ubuntu instructions in WSL2.

PR #22375, opened 2021-09-29
c++ replay: more robust seek

deanlee:
- [x] more robust seek .
- [x] skip invalid segments
- [x] faster merge events
- [x] test cases
- [x] reduce cpu usage ~20%


pd0wm:
Does this also handle seeking past the end of the route?

deanlee:
It should be handled, I need to write some test cases int this PR to ensure the stability of seek

deanlee:
should use std::ceil(seconds/60.0) to get the  segment num

pd0wm:
How much faster is the merge by doing things in parallel. I'm worried that depending on `tbb` is going to give us issues when we try to run on device.

deanlee:
not too much, merge 5 segments, from ~9ms-> ~0.5 ms.
Do I need to remove it now?

pd0wm:
Yeah, let's remove it. It's not installed on AGNOS.

PR #22378, opened 2021-09-29
c++ replay: publish all frames in CameraServer

deanlee:


pd0wm:
Any reason YUV frames are not enabled? If we have support for that, and routes with only 1 segment we can delete unlogger.py since we reached feature parity.

deanlee:
There is no particular reason. I wanted to support it after #22217. but #22195 can also work. we can use it before #22217

pd0wm:
How big is the performance hit by always broadcasting YUV? I.e. is it worth it to gate it behind a flag like we did for unlogger.py?

deanlee:
the yuv data is already cached by framereader in the decoder thread.  the only time consumed while publish is memcpy to vipc buffer. it's extreme fast

pd0wm:
Last PR needed to delete unlogger.py!

pd0wm:
Confirmed YUV works by running modeld live

PR #22382, opened 2021-09-29
notebook precommit install

haraschax:

### DIFF
--- a/update_requirements.sh
+++ b/update_requirements.sh
@@ -40,3 +40,4 @@ pyenv rehash
 echo "precommit install ..."
 pre-commit install
 [ -d "./xx" ] && cd xx && pre-commit install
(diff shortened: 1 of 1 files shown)


update_requirements.sh
```diff
@@ -39,4 +39,5 @@ pyenv rehash
 
 echo "precommit install ..."
 pre-commit install
-[ -d "./xx" ] && cd xx && pre-commit install
\ No newline at end of file
+[ -d "./xx" ] && cd xx && pre-commit install
+[ -d "./notebooks" ] && cd xx && pre-commit install
```
adeebshihadeh, line 43: does this work? current directory should already be xx, and think you want to be in notebooks, right?

HaraldSchafer: yeah you're right

PR #22383, opened 2021-09-30
Fix second lead car indicator for cut-in detection

sunnyhaibin:
**Description**: The second lead car indicator has been missing since 0.8.10 `master`. Second lead car indicator does not show up and only the first lead car indicator moves behind the cut-in car when being cut-in.

**Verification**: Second lead car indicator is showing up along with the first lead car indicator, and it moves behind the cut-in car independently when the car is being cut-in.

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -144,7 +144,7 @@ static void ui_draw_world(UIState *s) {
     if (lead_one.getProb() > .5) {
       draw_lead(s, lead_one, s->scene.lead_vertices[0]);
     }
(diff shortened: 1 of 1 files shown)


pd0wm:
I'm confused. This PR is only a whitespace change.

cydia2020:
> I'm confused. This PR is only a whitespace change.

I think the missing white space is the issue, I've been seeing the same thing for a while.

pd0wm:
For C/C++ whitespace is not relevant. So it's impossible this change fixes anything.

adeebshihadeh:
This whitespace change certainly doesn't affect any behavior, but I'll merge anyway to fix the missing space.

PR #22394, opened 2021-10-01
No more auth

ghost:
Auth is not needed anymore


pd0wm:
It still is if you want to run on non public routes

PR #22395, opened 2021-10-01
(potentially) better C3 amplifier config

robbederks:
This increases the base volume and rolls off the higher frequencies quite significantly.
image (6)


speedking456:
tested this, it sounds less distorted and much closer to the C2 sound profile.

sshane:
All the sounds sound great now, however `warning_repeat` is much quieter than before, I don't know if that's an improvement.

robbederks:
> All the sounds sound great now, however `warning_repeat` is much quieter than before, I don't know if that's an improvement. I could definitely see not hearing it in a Corolla at 70 mph with loud music

Yeah, it's definitely still a WIP. It's supposed to go with a new set of sounds too

PR #22398, opened 2021-10-01
Support for external driving panda

pd0wm:
After this PR boardd is "pinned" to a specific panda FW serial number. It's no longer needed to enforce the FW version in thermald since there is no risk of hot swapping to a different panda.


robbederks:
Continuing this in #22402

PR #22399, opened 2021-10-01
accel/decel button metric increments

speedking456:
Freedom units have weird multiples compared to metric. Modified so that it increments by 10 km/h on long press and 1 km/h on short press, rather than 8 and 1.6 km/h. This mimics stock hyundai behaviour on metric vehicles.

### DIFF
--- a/selfdrive/controls/lib/drive_helpers.py
+++ b/selfdrive/controls/lib/drive_helpers.py
@@ -1,16 +1,25 @@
 import math
 from cereal import car
 from common.numpy_fast import clip, interp
+from common.params import Params
 from common.realtime import DT_MDL
 from selfdrive.config import Conversions as CV
 from selfdrive.modeld.constants import T_IDXS
 
+IS_METRIC = Params().get_bool("IsMetric")
 
 # kph
 V_CRUISE_MAX = 135
-V_CRUISE_MIN = 8
-V_CRUISE_DELTA = 1.6
-V_CRUISE_ENABLE_MIN = 40
+if IS_METRIC:
+  V_CRUISE_DELTA = 1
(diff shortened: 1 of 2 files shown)


selfdrive/controls/lib/drive_helpers.py
```diff
@@ -1,16 +1,25 @@
 import math
 from cereal import car
 from common.numpy_fast import clip, interp
+from common.params import Params
 from common.realtime import DT_MDL
 from selfdrive.config import Conversions as CV
 from selfdrive.modeld.constants import T_IDXS
 
+IS_METRIC = Params().get_bool("IsMetric")
```
adeebshihadeh, line 9: this needs to update as the toggle is flipped

speedking456: Gonna need some help. I'm not sure how I would go about doing that

pd0wm:
You're sure hyundai uses 10 km/h increments? Think Honda uses 5 km/h.

KyleGreenlaw:
Can confirm Hyundai stock system uses +/- 10km/h on long press.

KyleGreenlaw:
Also min set speed on Hyundai stock is 30km/h, not 40.

pd0wm:
ISO 15622:2018 suggests a min set speed of 4.4 m/s.

spektor56:
> You're sure hyundai uses 10 km/h increments? Think Honda uses 5 km/h.

Yea Honda increments by 5 and for Canadians makes more sense.  Most people drive above speed limits and you can go 15 km/h over without getting demerit points.  65 on a 50, 75 on a 60 etc.

pd0wm:
What change broke things? The short/long press, or the metric one?

PR #22401, opened 2021-10-01
Merge pandad into boardd 

deanlee:
continue #20061, split to smaller PRs:
- [x] #22405 
  - [ ] panda_dfu
  - [ ] panda_flash


adeebshihadeh:
Feel free to squash all the previous commits to clean up the history :)

pd0wm:
https://github.com/commaai/openpilot/pull/22402 will change behavior to exit boardd after disconnect. This will make sure pandad can flash new pandas as they show up, which is most of what we wanted by merging boardd and pandad. We decided it was not worth it anymore to maintain and test two separate flashers so will keep pandad in python.

Sorry for the wasted time! But think it's not worth the complexity and the time we need to spend on this in the future.

PR #22402, opened 2021-10-01
Multipanda support

robbederks:
~~Hacky implementation for RX on both pandas now. Very much WIP~~


robbederks:
Little less hacky now. Have yet to test this with actual HW, but at least it compiles and has most of the functionality (filled with bugs still probably) that I would want.

One thing still missing is that we'd probably want PandaState to be a capnp list (same order as the `pandas` list), not separate messages like now, since there's not really a way to tell them apart

selfdrive/boardd/boardd.cc
```diff
@@ -176,20 +181,33 @@ void can_send_thread(Panda *panda, bool fake_send) {
...
+        auto can_data = event.getSendcan();
+
+        for (uint32_t i = 0; i < pandas.size(); i++) {
+          std::list<cereal::CanData::Reader> filtered_can_data;
+
+          for (const auto& msg : can_data) {
+            if ((msg.getSrc() / 4) == i) {
+              filtered_can_data.push_back(msg);
+            }
+          }
```
robbederks, line 193: This can probably be cleaned up, not sure what the best way to write this is

selfdrive/boardd/boardd.cc
```diff
@@ -561,34 +588,55 @@ int main(int argc, char* argv[]) {
...
+
+        // Sort by hardware type
+        if (a->hw_type != b->hw_type) {
+          return a->hw_type > b->hw_type;
+        }
+
+        // Last resort, sort by serial
+        return a->usb_serial.compare(b->usb_serial);
+      }
+    );
```
robbederks, line 622: This needs testing. Can never remember what returning true/false exactly does to the order of the sorted list

selfdrive/boardd/boardd.cc
```diff
@@ -561,34 +588,55 @@ int main(int argc, char* argv[]) {
...
+    threads.emplace_back(can_send_thread, pandas, getenv("FAKESEND") != nullptr);
+    threads.emplace_back(can_recv_thread, pandas);
 
     for (auto &t : threads) t.join();
 
 fail:
-    if (peripheral_panda_serial != panda_serial) {
-      delete peripheral_panda;
+    for (const auto& panda : pandas){
+      delete panda;
```
robbederks, line 639: does this work?

selfdrive/boardd/panda.cc
```diff
@@ -341,14 +354,16 @@ void Panda::send_heartbeat() {
...
+void Panda::can_send(std::list<cereal::CanData::Reader> can_data_list) {
   static std::vector<uint32_t> send;
   const int msg_count = can_data_list.size();
 
   send.resize(msg_count*0x10);
 
   for (int i = 0; i < msg_count; i++) {
-    auto cmsg = can_data_list[i];
+    auto cmsg = can_data_list.front();
+    can_data_list.pop_front();
```
robbederks, line 365: refactor to write this cleaner? linked list does not have indexing, so we either need to do this, or use an iterator and keep track of the msg num in a counter

Issue #22254, opened 2021-09-16
Locationd not using carState packets as expected, possible timing issue


HaraldSchafer:
In this segment
` 09c32a592a49b2da|2021-08-04--13-41-56/18 `

The carState.standstill flag is always true. This should cause serious issues with the locationd that expects this value to be correct and enforces standstill when it is true. Somehow locationd did not react to this at runtime at all (it does in offline reruns). I suspect there are timing issues between the messages causing messages to be dropped, but unconfirmed.

nuwandavek:
This is not a bug. The `valid` attribute of this segment's `carState` message is always `False`. So, it fails this check https://github.com/commaai/openpilot/blob/1cfabc3388443a5480310930e12331d508c92bdd/selfdrive/locationd/locationd.cc#L453

So, the `carState` message is never used to update the filter.

HaraldSchafer:
Ok this is just a car bug then adding to this
https://github.com/commaai/openpilot/issues/22253

PR #22403, opened 2021-10-01
paramsd: Sort messages in each update iteration before processing

nuwandavek:
Currently, `liveLocationKalman` messages are processed first, and the `carState` messages are processed later, even through the `carState` message was earlier. This leads the filter to ignore some `carState` messages as it is too old. The fix is to sort the messages on logMonoTime first before processing them.

Img below shows the difference in process replay, which is confirmed to be correct by liveParamsD.

image


adeebshihadeh:
Looks like there's still a diff from process replay. Do you have uncommitted local changes?

HaraldSchafer:
Yes it is, we were just talking about how to add logging for this stuff

HaraldSchafer:
https://github.com/commaai/openpilot/issues/22254

pd0wm:
@adeebshihadeh Should we add something to SubMaster instead? SubMaster.update() can return a list of socket names that are sorted by message arrival time for example.

adeebshihadeh:
I don't think it makes sense for SubMaster to do the sorting. The sorting is really only important for the filter, so think it makes sense to enforce as close to the filter as possible.

pd0wm:
I don't say we always enforce it, but by allowing you to loop over the result of `sm.update()` it makes it as easy as possible when you want it and hides some of the implementation details.

nuwandavek:
Do we want to put this in `sm.update()`?

adeebshihadeh:
I think this is fine to merge as-is, but if we do want to add it to sm.update, it should be an optional argument that defaults to False.

pd0wm:
@adeebshihadeh `sm.update()` currently doesn't return anything, so there is not backwards compatibility needed.

@nuwandavek feel free to merge this now, we can always come back later and refactor

PR #22404, opened 2021-10-02
GM: EPS fault workaround

jyoung8607:
**Description**

Proposal to fix the infamous #21557. Thanks to @gregjhogan for the loopback idea!

- ~~Reduce GM EPS steer commands from 50Hz to 25Hz~~
- ~~Scale openpilot and Panda safety ramp-rates to match the prior effective ramp-rate~~
- Sync the EPS steer command counter based on Panda loopback of openpilot's last successful send

Still needs discussion and feedback before it's finalized.

**Verification**

Fully clean 148 mile drive and a collection of other shorter drives.

**Route**

`e38a02cae1ef1e0c|2021-10-01--19-20-12`

**Additional context**

@JasonJShuler, how do you feel about this? It wasn't intended to deal with any potential controlsd lags; if we still see devices on current openpilot that are incrementing cumLagMS, we need to identify them and fix the lags. It seems like it's effective for minor Panda policing drops. Do you think this is okay to run on all other GM vehicles, or do we need to special-case the sensitive ones? I have no idea there. Do you think we need to add any init-time counter sync? We seem to be doing okay without it.

@adeebshihadeh, how do you feel about this? Unconventional, but doesn't break service abstractions/contracts.


qadmus:
Is it enough just to reduce the frequency to 25 Hz?

jyoung8607:
> Does the issue resolve by just reducing ASCMLKASteeringCmd from 50 Hz to 25 Hz?

It's reduced somewhat, but not fixed. We do need something like the loopback sync for controls allowed transitions, because it can't seem to tolerate any counter gaps at all.

adeebshihadeh:
Yeah, this looks like a pretty robust solution.

As a side note, I don't really like how the steering limits are configured. It's easy to bypass the limits, etiher purposefully or accidentally, by sending at a higher frequency. Though, this is a refactor for another PR.

JasonJShuler:
@jyoung8607 this is fantastic! Thanks for pouring your resources into this - we get this upstreamed and I predict a flood of new car ports!! So glad I was wrong, and so grateful for your help!

To your question re: applicability to other GM vehicles: The disengage update sounds non-impactful.
The LKAS rate change I cannot offer anything evidence-based, apologies. I have only dealt with this obnoxious variety of EPS. The stock cameras send frames at 10hz while inactive and 50hz while active. Any deviation from this opens the possibilty of compatibility issues, and I don't know what impact it would have on steering performance.


That said, I'd avoid the condition if possible - till FP2.0 is solid for GM it would be a bit of a pain
~~~

I know I claimed the pause-and-catch-up couldn't be detected from the OP side, but I'm no longer so sure. If we can see the timing in cabana doesn't that mean it passed through OP? How long is the actual wall-clock delay from OP sending a frame to OP receiving it? Could we track it this way?


Thanks again!

JasonJShuler:
Oh, regarding the Bolt port. There's a lot left to do and unfortunately a lot of it was previously Panda code. Could we arrange some time to discuss on Discord?

JasonJShuler:
> Yeah, this looks like a pretty robust solution.
> 
> As a side note, I don't really like how the steering limits are configured. It's easy to bypass the limits, etiher purposefully or accidentally, by sending at a higher frequency. Though, this is a refactor for another PR.

On the bright side, this EPS not accepting messages any faster than 50hz makes it a bit harder to work around 😆

jyoung8607:
I prefer to make it an across-the-board change, if we can. If that doesn't work out, it should be easy enough to identify the cars involved and pick rates, and pass a parameter down to the GM Panda safety do the same. But I'd like to avoid it.

There is some risk here we need to figure out how to retire. If we don't yet know if it's safe to use 25Hz globally, we need to develop that understanding. It would help to get folks with a few different GM models, including the not-so-picky kind, to make sure their cars work. @qadmus, I've seen a lot of work from you on GM tuning, it would be nice to have your evaluation, to measure whether this change materially affects tunes.

If all that looks good, I'd ask comma to consider this (with a companion Panda PR) for _just after_ the 0.8.10 cutoff, beginning of 0.8.11. That starts getting it in front of openpilot-nightly/master-ci users as well as the more actively maintained forks. Then, we could ask for comma help to survey GM vehicle qlogs a week or two later, to look for new concentrations of steerWarning or steerError well in advance of the next monthly release.

jyoung8607:
Re: the Bolt port and other questions, not sure I followed what you were saying or understand the problems involved, but happy to chat on Discord. A weekday next week would be best, I'm otherwise committed this weekend.

qadmus:
Standby today for detecting any change in tune behavior for 50 Hz -> 25 Hz on Volt. 

I expect no change in behavior, and prefer stock 25 Hz.

Using 50 Hz works but it's probably wishful thinking that it has any improvement in responsiveness. The torque rate limits of the actuator are the same at 25 Hz and 50 Hz, so it's not message-to-message limits like Toyota (?). On one weak attempt, 100 Hz caused a bad actuator status in Volt, that I did not dig into.

jyoung8607:
IIRC (please correct me if I'm wrong) the GM harness topology doesn't use the relay, and going to dashcam mode doesn't let us easily observe stock LKAS's active send rate, only the 1Hz idle rate. So, I can't test your assertion about the stock send rate.

But if you're certain the stock send rate is 25Hz, this changes from an unknown-risk mitigation into a no-brainer bugfix. We should absolutely be matching the stock send rate unless there's an exceedingly good reason to do otherwise. Do you have evidence for this?

JasonJShuler:
> IIRC (please correct me if I'm wrong) the GM harness topology doesn't use the relay, and going to dashcam mode doesn't let us easily observe stock LKAS's active send rate, only the 1Hz idle rate. So, I can't test your assertion about the stock send rate.
> 
> But if you're certain the stock send rate is 25Hz, this changes from an unknown-risk mitigation into a no-brainer bugfix. We should absolutely be matching the stock send rate barring some really good reason not to, and no such reason is in evidence. Do you have evidence to support 25Hz being the stock LKAS camera active send rate?

**Every GM vehicle I've encountered uses a 50hz stock LKAS rate while active, not 25.** I would recommend double-checking with Andrew C if he is still around - he was very familiar with the technical details of ASCM-bearing vehicles. I recall old conversations where the 100ms gap vs 20ms gap thing was discussed - back when my Bolt was the first to run OP. I can provide evidence of this on my Bolt anyway if you like.


(More of an FYI) The current GM harness topology is probably out-of-date. Many newer cars have a filtered OBD-II port, so the OBD-II harness doesn't really make sense - sort of a step backwards. (Think Giraffe) For vehicles without an ASCM (ie without radar) the front camera functions the same as your Toyota or Honda or Kia or whatever, and the harness *could* work if GMLAN support were included and we can capture rolling counter prior to relay break. (In fact, there seems to be sufficient power at the camera connector as to make the Comma Power optional).

jyoung8607:
I had a follow-up Discord conversation with @qadmus and after checking, he agrees with you, and corrected his earlier comment. So we're back to being almost-but-not-quite certain this is a good change for everyone.

Next step is some more feedback from @qadmus (or anyone else with input) to confirm this doesn't materially change the tuning situation, which we are hoping to have today.

If modern GM vehicles would allow changing the harness/topology toward the more supportable relay intercept setup, that would be a positive thing, but that's a complex discussion for another day.

selfdrive/car/gm/carstate.py
```diff
@@ -131,3 +133,15 @@ def get_can_parser(CP):
...
     return CANParser(DBC[CP.carFingerprint]["pt"], signals, checks, CanBus.POWERTRAIN)
+
+  @staticmethod
+  def get_loopback_can_parser(CP):
+    signals = [
+      ("RollingCounter", "ASCMLKASteeringCmd", 0),
+    ]
+
+    checks = [
+      ("ASCMLKASteeringCmd", 25),  # reduced openpilot send rate for sensitive EPS firmware
```
qadmus, line 144:
```suggestion
      ("ASCMLKASteeringCmd", 50),  # reduced openpilot send rate for sensitive EPS firmware
```
Check at 50 Hz

JasonJShuler:
> Non-update update: This PR is still alive, but due to other commitments I haven't been able to spend time on openpilot related stuff for the last couple weeks, and don't expect I'll be able to until next week at the earliest.
> 
> General thoughts:
> 
> * The original code in this PR, 25Hz send rate / counter sync / scaled rates, had zero EPS faults but it did hurt the tune
> * The current code, 50Hz send rate / counter sync only, does occasionally fault EPS but much less than stock
> * Next logical step is to try 33Hz / counter sync only, don't scale the rates, verify zero EPS faults, fix the tune if needed
> * Could also try 33Hz with rate scaling, but the ramp rates for GM are relatively high and could stand to come down

I'm not a big fan of having OP send LKAS frames at a different rate than what the system was designed for. If OP is able to detect when it happens the same way I did on the panda, couldn't we do something similar? That is, either stop the sending till the RC wraps around, or unbind the RC from the loop counter so we only increment when we actually send? If we did that, there would only be 1 dropped frame.

Apologies if this is already happening - I haven't dug into this code yet

jyoung8607:
And to add some clarity: technically the current version of code in this PR implements something like solution 4 (transmit drops if the counter doesn't increment in time) and it's not reliable.

PR #22405, opened 2021-10-02
boardd: new class PandaComm

deanlee:


deanlee:
moved to https://github.com/commaai/openpilot/pull/23015

PR #22406, opened 2021-10-02
add intro to openpilot docs

barbinbrad:
Added intro to openpilot docs


adeebshihadeh:
Nice work, but I think these kinds of docs are best kept in the wiki. @pd0wm what do you think?

pd0wm:
@barbinbrad very nice work! I agree on the wiki page. Should lower the barrier to help contribute and expand the missing sections.

I went ahead and put the page here: https://github.com/commaai/openpilot/wiki/Introduction-to-openpilot

PR #22411, opened 2021-10-02
Lateral PID: move steer feedforward to CarInterface

hewers:
**Description** This is useful for cars with strange steer commands that aren't proportional to angle * speed ** 2.

I want to implement this in the best way, and I'm still figuring out Python's tricks (this is cargo cult coding), so suggest any improvements and we'll iterate.

On top of this change, I'll add GM steer feedforward and lateral tuning.

**Verification** Process replay is unchanged.

### DIFF
--- a/selfdrive/controls/lib/latcontrol_pid.py
+++ b/selfdrive/controls/lib/latcontrol_pid.py
@@ -6,11 +6,12 @@ from cereal import log
 
 
 class LatControlPID():
-  def __init__(self, CP):
+  def __init__(self, CP, CI):
     self.pid = PIController((CP.lateralTuning.pid.kpBP, CP.lateralTuning.pid.kpV),
                             (CP.lateralTuning.pid.kiBP, CP.lateralTuning.pid.kiV),
                             k_f=CP.lateralTuning.pid.kf, pos_limit=1.0, neg_limit=-1.0,
                             sat_limit=CP.steerLimitTimer)
+    self.feedforward = CI.get_steer_feedforward
 
   def reset(self):
     self.pid.reset()
@@ -23,7 +24,7 @@ class LatControlPID():
     angle_steers_des_no_offset = math.degrees(VM.get_steer_from_curvature(-desired_curvature, CS.vEgo))
     angle_steers_des = angle_steers_des_no_offset + params.angleOffsetDeg
 
-    pid_log.angleError = angle_steers_des - CS.steeringAngleDeg 
+    pid_log.angleError = angle_steers_des - CS.steeringAngleDeg
     if CS.vEgo < 0.3 or not active:
       output_steer = 0.0
       pid_log.active = False
@@ -33,9 +34,8 @@ class LatControlPID():
       self.pid.pos_limit = steers_max
       self.pid.neg_limit = -steers_max
 
-      # TODO: feedforward something based on lat_plan.rateSteers
-      steer_feedforward = angle_steers_des_no_offset  # offset does not contribute to resistive torque
-      steer_feedforward *= CS.vEgo**2  # proportional to realigning tire momentum (~ lateral accel)
+      # offset does not contribute to resistive torque
+      steer_feedforward = self.feedforward(angle_steers_des_no_offset, CS.vEgo)
 
       deadzone = 0.0
 
--- a/selfdrive/controls/lib/latcontrol_pid.py
+++ b/selfdrive/controls/lib/latcontrol_pid.py
@@ -11,7 +11,7 @@ class LatControlPID():
                             (CP.lateralTuning.pid.kiBP, CP.lateralTuning.pid.kiV),
                             k_f=CP.lateralTuning.pid.kf, pos_limit=1.0, neg_limit=-1.0,
                             sat_limit=CP.steerLimitTimer)
-    self.feedforward = CI.get_steer_feedforward
+    self.get_steer_feedforward = CI.get_steer_feedforward
 
   def reset(self):
     self.pid.reset()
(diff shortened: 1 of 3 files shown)


selfdrive/controls/lib/latcontrol_pid.py
```diff
@@ -6,11 +6,12 @@
 
 
 class LatControlPID():
-  def __init__(self, CP):
+  def __init__(self, CP, CI):
     self.pid = PIController((CP.lateralTuning.pid.kpBP, CP.lateralTuning.pid.kpV),
                             (CP.lateralTuning.pid.kiBP, CP.lateralTuning.pid.kiV),
                             k_f=CP.lateralTuning.pid.kf, pos_limit=1.0, neg_limit=-1.0,
                             sat_limit=CP.steerLimitTimer)
+    self.feedforward = CI.get_steer_feedforward
```
qadmus, line 14: Is this more efficient? Or should we pass CI to update() and call get_steer_feedforward() there?

selfdrive/controls/lib/latcontrol_pid.py
```diff
@@ -23,7 +24,7 @@ def update(self, active, CS, CP, VM, params, desired_curvature, desired_curvatur
     angle_steers_des_no_offset = math.degrees(VM.get_steer_from_curvature(-desired_curvature, CS.vEgo))
     angle_steers_des = angle_steers_des_no_offset + params.angleOffsetDeg
 
-    pid_log.angleError = angle_steers_des - CS.steeringAngleDeg 
+    pid_log.angleError = angle_steers_des - CS.steeringAngleDeg
```
qadmus, line 27: end of line whitespace

grekiki:
I've done some testing by making 
```
steer_feedforward = angle_steers_des_no_offset + params.averageOffsetDegrees
```
and setting P and I to zero. Then I tuned the kf to ensure that the wheel had no centering anymore(reducing kf if it was unstable and increasing if the wheel was centering).

I think this is a nice way to tune the kf term, but it also confirmed in my case that the values need to be quite a bit different depending on speed(which makes sense if we consider openpilot torque output as the simulatet driver torque, which the EPS is boosting by different amounts depending on the speed).

Haven't checked for nonlinearity in angle a lot, but if it's just the speed, wouldn't breakpoints like for P and I be a nicer solution?

qadmus:
@grekiki That's a smart idea to attempt tuning kf, very clean if we had the right function of (speed, angle) -> (steer) to begin with.

I also think that feedforward is best with (desired_curvature + params.angleOffsetDegrees - params.angleOffsetAverageDegrees), as the constant offset doesn't need counter torque (angle sensor not zeroed) but the instant vehicle model error does. Working on proving that out.

Non-linearity: for Volt steer command it's not just speed but a sigmoid of angle, so speed breakpoints won't work.
```
  # Volt determined by iteratively plotting and minimizing error for f(angle, speed) = steer.
  def get_steer_feedforward(self, desired_angle, v_ego):
    if self.CS.car_fingerprint in [CAR.VOLT]:
      desired_angle *= 0.02904609
      sigmoid = desired_angle / (1 + fabs(desired_angle))
      return 0.10006696 * sigmoid * (v_ego + 3.12485927) # 6.99 ~= 7 mph minimum steer speed
```
solution

grekiki:
@qadmus For the first formula, looking at this simulation for example (https://www.youtube.com/watch?v=cJyIYjwakKo) I'd say it's quite reasonable to assume that the front tyres are being "pushed" down by the weight of the engine thereby generating lateral loads to keep the car straight, which should generate centering force. (@pd0wm any opinion on that?) (thinking a bit more, I am not confident this is true. Slip seems complicated...)

For the sigmoid, seems interesting. From just physics, I'd say that the centering torque should be about proportional to the $sin(steering_angle)*v^2$ where steering_angle=steering_wheel_angle/steering_ratio(combine a=v^2/r with http://www.davdata.nl/math/turning_radius.html), though it doesn't seem like that would approximate your curve too well(with about 15 steering ratio you don't get to the "interesting part" of the sin quite yet).

Are there plots something you can make from any rlogs?

qadmus:
@grekiki the plots are almost universal from any rlogs, but there are a few hardcoded values like GM's max steer command. See my branch `tuning-scripts` for work in progress. @sshane created all the principal work, I just riffed on it.

That sigmoid will hopefully come after this refactor, I tried fitting all the sigmoids with all the degrees of freedom for variables, arctan is close but 1/(1+|x|) is best.

selfdrive/controls/lib/latcontrol_pid.py
pd0wm:
```suggestion
    self.get_steer_feedforward = CI.get_steer_feedforward
```

Think this is a better name to indicate it's a function not a variable.

pd0wm: Think this is fine, no need to pass to update.

qadmus:
Renamed LatControlPID.get_steer_feedforward()

@pd0wm , an alternative implementation is to do the best with the original LateralPID output, then try to re-map the steer command in CarController. But, that would interfere with clipping / saturation behavior.

 I like how each car could get their own feedforward estimate here.

pd0wm:
Yeah like that too!

If you're looking for another small project you could make all lateral control classes inherit from the same base class. After this change the PID controller has a different API for the constructor than the others, which could be confusing.

grekiki:
That sounds quite simmilar to https://github.com/commaai/openpilot/pull/21967

sshane:
I can rebase it if you want the PR!

PR #22412, opened 2021-10-02
Ubuntu setup script now works if openpilot not in $HOME

johndavidmartinez:
**Description** 
Ubuntu setup script assumed openpilot repo's root was in $HOME/openpilot. Now the script assumes openpilot repo's root is the directory above openpilot/tools (which it is).

**Verification** 
Re-ran ./ubuntu_setup.sh after making fix.


johndavidmartinez:
It looks like a lot of tools just assume openpilot is in $HOME/openpilot. This is fine.

PR #22420, opened 2021-10-04
convert .format() to f-sting: PEP 498.

bbennett80:


pd0wm:
Did you use a tool to do this or change it manually?

bbennett80:
@pd0wm i used a tool. It is called flynt. Might this be something to add to a pre-commit workflow?

pd0wm:
Pre-commit isn't necessary. Don't think I've seen a new `format()` been committed after we moved to python3.

bbennett80:
Understood. Is the f-string conversion something you would consider given the speed benchmark improvements?

pd0wm:
The diff is quite large to review, but if I can regenerate it myself using the tool it might be worth it. What exact flags did you use to run `flynt`? Is the PR ready for review?

bbennett80:
I used "flynt openpilot" without other flags. 

I believe both pyextra and phonelibs were included. Happy to exclude those and resubmit. 

PR currently not ready for review. There are a few outstanding string concatenations for urls.

bbennett80:
Final flynt command for string concats: flynt --transform-concats openpilot --exclude pyextra phonelibs

bbennett80:
We just had the power go out. Let me close this and I will open another. Is that OK with you?

PR #22426, opened 2021-10-04
safety_setter_thread: exit on ignition low

pd0wm:
The thread never finishes if you go offroad before the fingerprinting finishes. The panda state threat then puts you back in noOutput and you're stuck. The safety setter thread should also exit when ignition goes low.


deanlee:
how about move global ignition into class Panda?

pd0wm:
Once we start supporting multiple pandas I think it makes sense to have one notion of a global ignition signal (which is an `or` of all pandas ignition signal). I already changed `send_panda_state` to return the ignition value for that specific panda.

PR #22430, opened 2021-10-04
replace unlogger.py with c++ replay

pd0wm:
Thanks to @deanlee!

rqplot is replaced by plotjuggler


doktorsleepelss:
It would be nice to keep unlogger around for use with WSL because the new replay doesn't work with it.

pd0wm:
Is this because OpenCL is not supported?

grekiki:
Seems like it. UI works fine here on WSL, but replay gives
```
seeking to 0
downloading segment 0 ...
downloading segment 1 ...
downloading segment 2 ...
replay: selfdrive/common/clutil.cc:63: cl_device_id cl_get_device_id(cl_device_type): Assertion `CL_SUCCESS == (clGetPlatformIDs(0, __null, &num_platforms))' failed.
```

pd0wm:
@grekiki Does https://github.com/commaai/openpilot/pull/22431 work?

grekiki:
@pd0wm  Works nicely :)
image

PR #22433, opened 2021-10-04
replay/ui.py: get image from vipc

deanlee:
ui


tools/replay/ui.py
```diff
@@ -109,19 +111,16 @@ def ui_thread(addr, frame_address):
     top_down = top_down_surface, lid_overlay
 
     # ***** frame *****
-    fpkt = messaging.recv_one(frame)
-    rgb_img_raw = fpkt.roadCameraState.image
+    rgb_img_raw = vipc_client.recv()
```
pd0wm, line 114: Can this be `None`?

pd0wm:
Can you add reconnect for when you restart `replay`. Now it's still connected to the wrong vipc server.

deanlee:
how about this? added is_connected  to VisionIpcClient

PR #22442, opened 2021-10-05
Set CarControl.active

pd0wm:


adeebshihadeh:
I think we should just deprecate these duplicate fields. Any reason to have it in controlsState and carControl?

pd0wm:
Think the car controller code should be able to work with just the information from the CarControl struct.

PR #22448, opened 2021-10-06
Fix boardd FD leak

### LINKED ISSUE #22423: Boardd: libusb initialization error after too many connect retries
If you leave boardd running with no pandas connected it will start printing errors after a few minutes and no longer connects to a panda.

```
...
libusb: error [usbi_pipe] failed to create pipe (24)
selfdrive/boardd/panda.cc: libusb initialization error
libusb: error [usbi_pipe] failed to create pipe (24)
selfdrive/boardd/panda.cc: libusb initialization error
libusb: error [usbi_pipe] failed to create pipe (24)
selfdrive/boardd/panda.cc: libusb initialization error
...
```

robbederks:
Fixes #22423

### DIFF
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -11,20 +11,27 @@
 #include "selfdrive/common/swaglog.h"
 #include "selfdrive/common/util.h"
 
-static int init_usb_ctx(libusb_context *context) {
-  int err = libusb_init(&context);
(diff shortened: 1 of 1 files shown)


pd0wm:
You want to use https://en.cppreference.com/w/cpp/thread/call_once to make sure it's thread safe. Is it also possible to fix the fd leak itself?

PR #22451, opened 2021-10-06
boardd: refactor function list()

deanlee:


pd0wm:
It looks like there might be locks involved around libusb contexts, so I don't think we want to share the context between `Panda` objects. https://github.com/commaai/openpilot/pull/22472 was the source of the leak.

PR #22454, opened 2021-10-06
Make pandaState and safetyMode a list

robbederks:
In preparation for multipanda support


pd0wm:
As discussed, we need to add this to the process replay regeneration script.

robbederks:
> As discussed, we need to add this to the process replay regeneration script.

Tried this, but getting an error trying to extract a bz2:
```
diff --git a/selfdrive/test/process_replay/regen.py b/selfdrive/test/process_replay/regen.py
index 0e8a0eaf1..b4871d06a 100755
--- a/selfdrive/test/process_replay/regen.py
+++ b/selfdrive/test/process_replay/regen.py
@@ -7,6 +7,7 @@ import argparse
 # run DM procs
 os.environ["USE_WEBCAM"] = "1"
 
+from cereal import log
 from cereal.visionipc.visionipc_pyx import VisionIpcServer, VisionStreamType  # pylint: disable=no-name-in-module, import-error
@@ -29,11 +30,16 @@ FAKEDATA = os.path.join(process_replay_dir, "fakedata/")
 def replay_service(s, msgs):
   pm = messaging.PubMaster([s, ])
   rk = Ratekeeper(service_list[s].frequency, print_delay_threshold=None)
-  smsgs = [m for m in msgs if m.which() == s]
+
+  smsgs = [m for m in msgs if m.which() in [s, "pandaState"]]
   while True:
     for m in smsgs:
-      # TODO: use logMonoTime
-      pm.send(s, m.as_builder())
+      if s == "pandaStates" and m.which() == "pandaState":
+        # Migrate pandaState to a list
+        pm.send(s, [m.as_builder()])
+      else:
+        # TODO: use logMonoTime
+        pm.send(s, m.as_builder())
       rk.keep_time()
```
@adeebshihadeh mind taking a look how to fix this?

robbederks:
> Regen should work and generate valid routes in this case. However, I don't understand how that patch works since the migration tries to send a python list? The migration should make a new pandaStates message and append the old pandaState into that list.

This is exactly what it does, a part of the diff of `regen,py` was already committed to this PR. Will run and upload then

PR #22472, opened 2021-10-07
boardd: fix context leak in init_usb_ctx

### LINKED ISSUE #22423: Boardd: libusb initialization error after too many connect retries

pd0wm:
This was the actual cause of the leak. `libusb_init` got a `libusb_context **` that was pointing to the argument of the `init_usb_ctx` not the context that the caller passed in.

Introduced in https://github.com/commaai/openpilot/commit/603ad435bef2a1db28b294b36768fda5f53fc814

Fixes https://github.com/commaai/openpilot/issues/22423

### DIFF
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -11,17 +11,19 @@
 #include "selfdrive/common/swaglog.h"
 #include "selfdrive/common/util.h"
 
-static int init_usb_ctx(libusb_context *context) {
-  int err = libusb_init(&context);
+static int init_usb_ctx(libusb_context **context) {
+  assert(context != nullptr);
+
+  int err = libusb_init(context);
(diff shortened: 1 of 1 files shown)


deanlee:
This bug is really hard to find if you don’t look at it carefully :+1::+1::+1: I have read the code many times but never noticed this problem

PR #21823, opened 2021-08-01
2018 Sonata 2.0T Limited Fingerprint

Felixowns:
Makes the 2018 Sonata identify properly and not require multiple ecu power cycles for comma to detect vehicle. Properly identifies as the sonata LF (2018-2019 are the same car) instead of previously identifying as "KIA OPTIMA SX 2019 & 2016" which worked but took multiple power cycles before openpilot would enable. Did a quick road test but will do a longer one to see if any other vehicle specific parameters need to be adjusted.

Dongle Id:0481820f6e50c2b7

PR #22481, opened 2021-10-08
2018 Sonata Fingerprint update

Felixowns:
Fixed a typo from previous fingerprint submission. Works fine now.


pd0wm:
Do you have a new route for these fingerprints? I doubt those older FW were wrong since I probably verified them against the actual route (see https://github.com/commaai/openpilot/pull/21823).

Felixowns:
Here is my route from this morning. I still got a "no exact match" this morning so maybe there is still something off. 
DongleID: 0481820f6e50c2b7
Route : 0481820f6e50c2b7|2021-10-14--09-17-42

adeebshihadeh:
All the FW from the route you posted are already on master. Your car should fingerprint correctly in the next release.

PR #22494, opened 2021-10-08
boardd: reduce unnecessary allocations

adeebshihadeh:


robbederks:
@adeebshihadeh both before and after this PR the send vector allocated is way too large (and before this commit 3/4s of the data sent to the panda was garbage / null), since `.size()` and `.resize()` is in elements (4 bytes here) instead of bytes.

I refactored it a bit more in the multipanda PR #22402

PR #22496, opened 2021-10-09
draw gps icon add

crwusiz:
Gps status is OK gps icon display


crwusiz:
<a href="https://imgur.com/3RjQCP7"></a>

adeebshihadeh:
Thanks for the PR, but we don't merge external design changes.

PR #22497, opened 2021-10-09
sim: update CARLA to 0.9.12

xia0long:
We can now install the latest version of Carla with pip.


adeebshihadeh:
Glad we can finally get rid of that massive CARLA download just to get the python API. `start_carla.sh` isn't working on my machine anymore though.

xia0long:
do we need to add the instructions for installing Carla to the `README.md` if we remove `install_carla.sh`？

xia0long:
> Glad we can finally get rid of that massive CARLA download just to get the python API. `start_carla.sh` isn't working on my machine anymore though.

Carla 0.9.12 supports only Vulkan, which may be the reason why it does not work. I rarely run Carla on docker and am looking at solutions to this problem.

reference link: https://carla.readthedocs.io/en/0.9.12/build_docker/

PR #22498, opened 2021-10-09
replay: write CarParams on start

deanlee:
carParams is logged every 50 s. it can be read from any segment.


deanlee:
merged to #22473

Issue #22429, opened 2021-10-04
Replay feature requests


pd0wm:
- [ ] Automatically add services to the blocklist. E.g. after hitting a multiple publishers error, don't try to broadcast anymore

deanlee:
> Ask confirmation before falling back to qlogs/qcameras, --qlog to skip confirmation

What to do after user chooses No? skip to the next non-qcamera/qlog segment or just exit?

pd0wm:
Just exiting is fine

deanlee:
@gregjhogan :  #22503

adeebshihadeh:
@deanlee seems qcams are broken, at least for C3 routes
![image](https://user-images.githubusercontent.com/8762862/137558630-b001841a-aee4-45cf-a126-93402486d4b6.png)

deanlee:
It may be a transform problem (not very sure yet). I checked the image output by frameread,  it doesn't seem to have any problems.

PR #22504, opened 2021-10-09
replay: improve segment downloading

deanlee:
- [x] better error handling
- [x] retry on failure
- [x] log download information when it's blocking the replay. resolve https://github.com/commaai/openpilot/issues/22429:
`downloading segments at: 736.25 KB/S`


adeebshihadeh:
We don't want to spam with too much output. Can we only show the progress when it's blocking the replay?

PR #22505, opened 2021-10-10
LogReader: use parallel sorting on PC

deanlee:


adeebshihadeh:
Unless @gregjhogan think this will give us significant speedups in the pipeline, I don't think this is worth it. We should be going for optimizations that'll work on both PC and the devices. I'm hoping that replay will also work on C2.

PR #22507, opened 2021-10-10
tools/replay/camera.py: get image from vipc

deanlee:


adeebshihadeh:
I don't think this is used anymore. Deleted in 0d10c000260262c11c58452d41be845b96134d40.

PR #22508, opened 2021-10-10
remove tools/replay/unlog_ci_segment.py

deanlee:


adeebshihadeh:
This is still needed until the CI segments can be accessed with the Route class.

PR #22531, opened 2021-10-12
replay: refactor Route and Segment

deanlee:
- [x] refactor to simplify code and improve readability.
- [x] handle segment download/load failure
- [x] fix issues with loading from local route
- [x] fix can't seek to the last segment.
- [x] merge one segment backward. this fix the following 2 issue:
   1.  get stuck for a while when playing to the next segment
   2. the event time between segments is not continuous, if the previous segment is not merged, there may be duplicate messages sent when playing to the next segment
- [x] fix other known issues, improve robustness.


deanlee:
Haven't noticed yet. There shouldn't be much difference if segments already cached

PR #22539, opened 2021-10-12
add acados and replay on Darwin

maimaixiong:


adeebshihadeh:
Thanks for the PR, but we generally like to build generated things that get checked in ourselves out of caution. #23048 is a more complete version of this.

PR #22540, opened 2021-10-12
modeld: more readable fill_model

deanlee:


gregjhogan:
Would it make sense to use a struct with named fields that are arrays of fixed size instead of these offsets?

Issue #22465, opened 2021-10-07
Carla camera data not showing in openpilot UI. Openpilot not working with Carla


pgchui:
**Describe the bug**
openpilot did not work with carla. openpilot UI did not show camera data. Client in carla did not move at all.

**How to reproduce or log data**
Console output:
```
Killing old publisher: roadCameraState
Traceback (most recent call last):
  File "./bridge.py", line 69, in cam_callback
    pm.send('roadCameraState', dat)
  File "/openpilot/cereal/messaging/__init__.py", line 248, in send
    self.sock[s].send(dat)
  File "cereal/messaging/messaging_pyx.pyx", line 149, in cereal.messaging.messaging_pyx.PubSocket.send
cereal.messaging.messaging_pyx.MultiplePublishersError
```
1. follow steps in [setup](https://github.com/commaai/openpilot/wiki/Requirements)
2. `./start_carla.sh`
3. `./start_openpilot_docker.sh`

**Additional context**

Operating system: Ubuntu 20.04
NVIDIA-SMI: 470.63.01
CUDA Version: 11.4

pgchui:
Adding `USE_FRAME_STREAM=1` fixed the camera feed issue but new issue arose. After press 1, in console output of bridge.py, engaged changed from false to true. Right before the engaged value change, console printed:
carControl: Reader was evicted, reconnecting
controlsState: Reader was evicted, reconnecting
And openpilot UI showed "openpilot unavailable. Communication issue between processes."

PR #22541, opened 2021-10-12
sim: send peripheralState message

jwolffe:
This fixes the "Openpilot not working with Carla" portion of Issue #22465. In recent Discord discussion on openpilot-simulation thread, @adeeb suggested someone create a PR. This is my first pull request ever, apologies if anything is wrong.

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -216,8 +216,6 @@ class Controls:
           self.events.add(EventName.fanMalfunction)
       else:
         self.last_functional_fan_frame = self.sm.frame
-    #print("peripheralState: ", self.sm['peripheralState'].pandaType, self.sm['peripheralState'].voltage, 
-    #  self.sm['peripheralState'].current,self.sm['peripheralState'].fanSpeedRpm, self.sm['peripheralState'].usbPowerMode)
 
     # Handle calibration status
     cal_status = self.sm['liveCalibration'].calStatus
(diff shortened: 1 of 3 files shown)


jwolffe:
> It would be great to start removing this SIMULATION flag. Now that there's a flag in managerState to indicate when a process should be running, it should be easy to remove it from the list. Not sure what we want to do about driverCameraState, though. That depends on whether we want to always expect that camera is there, which doesn't make much sense for this use case.



> It would be great to start removing this SIMULATION flag. Now that there's a flag in managerState to indicate when a process should be running, it should be easy to remove it from the list. Not sure what we want to do about driverCameraState, though. That depends on whether we want to always expect that camera is there, which doesn't make much sense for this use case.

I'm new around here, but..... it seems like the test for checking running status of processes should be moved to manager.py, where they are controlled. Does this actually make sense?

jwolffe:
I added the send of a fake peripheralState in manager.py as suggested, and supplied some commented-out test code in controlsd.py that verifies the correct transmission of the fake data. I had quite a time getting through the git learning curve (total of 5 commits!). The fake data is just some values I pulled out of the air- maybe something more realistic is desired?

jwolffe:
> Nice, you got the idea, but some of the units were wrong. Can you clean this up and test again after my changes? Should be good to merge after that.

Ok, will do. It may take a while... I got into a real mess with git, now starting over with a fresh clone into my local repo.

PR #22544, opened 2021-10-13
ModelDataRaw struct with fixed size arrays

gregjhogan:
All of these hand calculated numbers seem prone to error and like they will be a PITA the next time we change something
https://github.com/commaai/openpilot/blob/master/models/README.md

Perhaps we can use a struct with doc strings (or something similar) to generate this instead?

I converted pose as an example to get started, looks like some of the other outputs will require some more serious refactoring to do the same, but still seems possible.

In the end we should be able to cast the `ModelState` output `&s->output` as the struct instead of copying a bunch of pointers?
https://github.com/commaai/openpilot/blob/cbac96f50cbd959841e8fb24e33d95bd11a89b91/selfdrive/modeld/models/driving.cc#L113-L122

This should also make it easier for others to play with the model outside openpilot (we could even add a simple pyx wrapper if we wanted to).

### DIFF
--- a/selfdrive/modeld/models/driving.cc
+++ b/selfdrive/modeld/models/driving.cc
@@ -363,10 +363,10 @@ void posenet_publish(PubMaster &pm, uint32_t vipc_frame_id, uint32_t vipc_droppe
 
   for (int i =0; i < 3; i++) {
     trans_arr[i] = net_outputs.pose->trans_arr[i];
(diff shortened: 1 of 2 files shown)


HaraldSchafer:
Yeah I agree this would be a nice improvement.

HaraldSchafer:
Yeah sounds good, I'll merge this.

PR #22551, opened 2021-10-14
update to python 3.8.10

adeebshihadeh:


EwoutH:
Why 3.8.10 and not 3.8.12, if I may ask?

PR #22554, opened 2021-10-14
ModelDataRaw struct part 2

gregjhogan:
follow up to https://github.com/commaai/openpilot/pull/22544

### DIFF
--- a/selfdrive/common/modeldata.h
+++ b/selfdrive/common/modeldata.h
@@ -1,11 +1,14 @@
 #pragma once
+
+#include <array>
+
 const int  TRAJECTORY_SIZE = 33;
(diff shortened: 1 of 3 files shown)


gregjhogan:
ready for review, process replay failure is due to us previously logging position std as value directly from the model, but apparently it should have been exp(std) like we do for pose

gregjhogan:
> ready for review, process replay failure is due to us previously logging position std as value directly from the model, but apparently it should have been exp(std) like we do for pose

will be fixed by https://github.com/commaai/openpilot/pull/22593 at which point outputs will match exactly and tests will pass

gregjhogan:
@HaraldSchafer I think this is good to merge now

Issue #22555, opened 2021-10-14
replay: malloc(): corrupted top size, corrupted size vs. prev_size, double free or corruption (!prev)


osekom:
### Describe the bug

hi, I am trying to reproduce the command "./replay '0c7f0c7f0c7f0c7f|2021-10-13--13-00-00' --dcam --ecam" for the CTF comma_hack but I get the following errors:

./replay '0c7f0c7f0c7f0c7f|2021-10-13--13-00-00' --dcam --ecam
services  std::vector(sensorEvents, gpsNMEA, deviceState, can, controlsState, pandaStates, peripheralState, radarState, roadEncodeIdx, liveTracks, sendcan, logMessage, liveCalibration, androidLog, carState, carControl, longitudinalPlan, procLog, gpsLocationExternal, ubloxGnss, clocks, ubloxRaw, liveLocationKalman, liveParameters, cameraOdometry, lateralPlan, thumbnail, carEvents, carParams, roadCameraState, driverCameraState, driverEncodeIdx, driverState, driverMonitoringState, wideRoadEncodeIdx, wideRoadCameraState, modelV2, managerState, uploaderState, testJoystick)
load route "0c7f0c7f0c7f0c7f|2021-10-13--13-00-00" with 4 valid segments
seeking to 0
downloading segment 0 ...
downloading segment 1 ...
downloading segment 2 ...
merge segments std::vector(0, 1, 2)
camera[2] frame size 1928x1208
start vipc server
Starting listener for: camerad
camera[1] frame size 1928x1208
restart vipc server
Stopping listener for: camerad
Starting listener for: camerad
camera[0] frame size 1928x1208
restart vipc server
Stopping listener for: camerad
Starting listener for: camerad
malloc(): corrupted top size
Aborted (core dumped)

or:

./replay '0c7f0c7f0c7f0c7f|2021-10-13--13-00-00' --dcam --ecam
[...]

PR #22572, opened 2021-10-16
Replay/framereader: fix c3 qcamera padding problem

deanlee:
use libyuv to convert yuv to rgb , and fixed the padding problem.it's also 2x faster than  `avpicture_fill`&`sws_scale`

before:
wrong2

after:
gd2


pd0wm:
Great. Looks like there were other issues as well with the previous implementation:

> I ran into the same thing on Debian 11, I think the RGB buffer isn't padded out to the right alignment when it's allocated, and sws_scale may write off the end, depending on which vectorized implementation is used. I haven't looked into it further yet, but my workaround was to edit line 167 and double the size of the allocation.
https://discord.com/channels/469524606043160576/524594418628558878/898977054551339038

deanlee:
> Great. Looks like there were other issues as well with the previous implementation:
> 
> > I ran into the same thing on Debian 11, I think the RGB buffer isn't padded out to the right alignment when it's allocated, and sws_scale may write off the end, depending on which vectorized implementation is used. I haven't looked into it further yet, but my workaround was to edit line 167 and double the size of the allocation.
> > https://discord.com/channels/469524606043160576/524594418628558878/898977054551339038

yes, that might lead to random crash like  #22555

PR #22579, opened 2021-10-16
WIP: Support for Intel macOS Monterey 12.0

jererobles:
Solves weird quirks I had trying to compile openpilot on MBP 16"
* Homebrew path was different
* openssl path was different
* Acados was pointing to non existing dir from repo in Sconstruct
* Linker had issues finding OpenGL/CL
* rpath usage had incorrect format for macOS
* Missing array import


bravech:
I believe the missing array import is due to using libc++ instead of libstdc++

adeebshihadeh:
Thanks for the PR, but I'm going to close this in favor of the more complete #23365.

PR #22580, opened 2021-10-16
hyundai: match stock when aeb is disabled

gregjhogan:
openpilot longitudinal for hyundai doesn't have AEB so we should match stock when you manually disable AEB


gregjhogan:
confirmed this matches manually disabled AEB (show orange indicator on dash) should be good to merge

pd0wm:
Do you have a picture of what the alert looks like?

pd0wm:
That looks ok to me. I'm sure we're going to get questions about it though ("Why is my car faulted?"), so maybe put a comment here: https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/interface.py#L28

PR #22584, opened 2021-10-17
Mazda: silence hands off warning in cars with no lockout

Jafaral:
block eps lkas state from getting to the camera

Depends on: https://github.com/commaai/panda/pull/741

### DIFF
--- a/selfdrive/car/mazda/carcontroller.py
+++ b/selfdrive/car/mazda/carcontroller.py
@@ -46,6 +46,8 @@ class CarController():
 
     self.apply_steer_last = apply_steer
 
+    can_sends.append(mazdacan.create_steer_rate(self.packer, CS.CP.carFingerprint, CS.steer_rate_msg))
+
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
We shouldn't really fake things like this. Why don't we just take over sending the alert message? We can start by copying the message from the camera and only filtering out the alert.

adeebshihadeh:
Regardless of this PR, we want to show HUD alerts ourselves anyway, and it looks like this is missing from the Mazda port. [These](https://github.com/commaai/cereal/blob/a0850ebef69c5a3c43c875fb730528b049bad2e1/car.capnp#L335) are the alerts that should be implemented, but just implementing steerRequired and fcw gets you most of the way there.

Jafaral:
Not sure I follow @adeebshihadeh , do you mean showing alerts on OP screen?  plus we already have things like  gear info and seatbelt. Can you please clarify ?

Jafaral:
Also, let me know if this PR is OK or if you want things done differently.

adeebshihadeh:
Fixed in #22730

PR #22589, opened 2021-10-17
tools refactor: unified setup script

dotpredicate:
**Description**
This PR provides a unified setup script for all platforms.
Platform-specific and platform-agnostic features have been separated. It is easy to debug if something goes wrong and some features can be skipped (to avoid wasting time on reinstalling for example).
Support for a platform can be added by creating a function and registering it in the OS list switch-case, see `install_ubuntu_lts_requirements` as an example.
Hard dependencies on openpilot folder being in the home directory and requirement of pyenv have been removed from the `openpilot_env.sh` script.

**Verification**
The setup script was tested brutally on Ubuntu 20.10 and should also work flawlessly with Ubuntu 20.04 LTS.
The install dependencies were copied from old files (with the only difference - removing deprecated libswscale) so nothing changes.
I am not sure about `./setup.sh --os macos` working, there may be bugs.


pd0wm:
Few quick comments since the PR is still in draft:

1) Can you detect the os automatically?
2) Can you prevent the duplication in the package list between Ubuntu LTS and Ubuntu latest?

@adeebshihadeh How do you feel about using the setup script as a base for our docker files? That way the packages don't get out of sync and the setup script is tested. Would be worse for caching, but guess that's acceptable.

lipniak:
>  1. Can you detect the os automatically?

Added a `detect_os` function which is executed if `--os` was not provided. Detects all supported systems.

> 2. Can you prevent the duplication in the package list between Ubuntu LTS and Ubuntu latest?

Added a `install_ubuntu_common_requirements` function - executed in ubuntu-lts and ubuntu-latest setup.

PR #22612, opened 2021-10-19
Mazda: Fix OP steer warning in cars with lockout

Jafaral:
It turned out cars with lockout don't have LKAS_BLOCK at the same time when HANDS_OFF signal is set. Once LKAS_BLOCK is set the other is unset causing OP warning to not show. Restore the old logic, but limit it to cars with lockout since it is only applicable to those, and it causes disengage in cars without lockout.
  
### DIFF
--- a/selfdrive/car/mazda/carstate.py
+++ b/selfdrive/car/mazda/carstate.py
@@ -78,8 +78,9 @@ class CarState(CarStateBase):
         self.low_speed_lockout = False
         self.low_speed_alert = False
 
-    # Check if LKAS is disabled due to lack of driver torque
-    ret.steerWarning = (cp.vl["STEER_RATE"]["HANDS_OFF_5_SECONDS"] == 1) and (cp.vl["STEER_RATE"]["LKAS_BLOCK"] == 1)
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Is `HANDS_OFF_5_SECONDS` ever set on cars without a lockout? If not, I'd prefer to use that instead of using the hardcoded list.

Jafaral:
> Is `HANDS_OFF_5_SECONDS` ever set on cars without a lockout? If not, I'd prefer to use that instead of using the hardcoded list.

It is. The behavior is a bit different though. Cars with lockout, the signal timeouts after a few seconds, `LKAS_BLOCK` gets set, `HANDS_OFF_5_SECONDS` gets unset. Whereas in cars with NO lockout, the signal continues for as long as there is no torque and `LKAS_BLOCK` doesn't get set. 

Thinking about this more, I don't think we care about hands off really, what actually care about is  wether LKAS is disabled (lockout) or not. I think I have a better solution for that:

```
ret.steerWarning = self.lkas_allowed  and (cp.vl["STEER_RATE"]["LKAS_BLOCK"] == 1))

```
T

PR #22631, opened 2021-10-20
Toyota longitudinal: Illuminate PCS Warning Light If DSU is Unplugged

cydia2020:
**Description**
Turn on amber PCS off for better indication, sorry didn't realise this earlier.

https://github.com/commaai/opendbc/pull/454


cydia2020:
> I changed the signal to be only 1 bit to reflect the comments and the usage in openpilot. Can you ensure everything still works after my opendbc changes?
> 
> @adeebshihadeh can you quickly test on a TSS2 car once the PR is rebased?

The signal is 2 bit as the pcs light can also flash.

cydia2020:
0 = light off
1 = PCS light constant ON
2 = PCS light blinking
3 = it does nothing

adeebshihadeh:
> @adeebshihadeh can you quickly test on a TSS2 car once the PR is rebased?

This message is only sent on TSS1 cars, since it's gated behind CarParams.enableDsu.

PR #22643, opened 2021-10-20
Car HUD Enhancement: Prevent Lane Lines From Flickering at Low Speed

cydia2020:
**Description**
Keep `left_lane_visible` and `right_lane_visible` off at low speeds (below `LDW_MIN_SPEED`).

This prevents lane line indicators from turning on/off constantly on vehicle's HUD when a front car is blocking the view of openpilot.

Most stock LDW systems do this as well. 

Toyota vehicles also put out a warning like this to remind drivers.
image


pd0wm:
Have you tried using hysteresis? That might be a nicer solution since it improves things at all speeds.

PR #22653, opened 2021-10-21
VW MQB: Volkswagen Taos Mk1

jyoung8607:
**Volkswagen Taos Mk1**

Covers all worldwide versions of the Taos sub-compact SUV (called Tharu in China). The Russia and China market variants have a smidge shorter wheelbase due to local parts-bin cost-saving measures, but it shouldn't be enough to matter.

- [x] added to CARS
- [x] added to RELEASES
- [x] test route added to [test_models.py](https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_models.py)
- [x] route with openpilot: `7d82b2f3a9115f1f|2021-10-21--15-39-42`

Thanks to community Taos owner adamrut.


jyoung8607:
I'm not sure what that Athena unit test fail is about. CI failed the same way on both commits some hours apart, but I can't see what I could possibly have done to cause it.

Issue #21333, opened 2021-06-18
hyundai: inconsistent fingerprinting


gregjhogan:
sonata and palisade (perhaps others?) will sometimes fuzzy fingerprint even though no ecus have had a firmware update

example route: 24804955a125718c|2021-06-13--15-31-24

I see what happened is we query a few data ids and it does not get a response for data id \xf1\xa0 (normally I guess it does).  It looks like there are many cases where we have the same ecu fw versions for a vehicle with and without \xf1\xa0  at the end, and other ecus never return a value for this data id.  I also see the data returned by this id is pretty much always included in another id that is being queried.

Since you cannot run a script and capture all combinations easily we could
* drop \xf1\xa0 id from the query list
* retry fingerprinting a few times

adeebshihadeh:
#19887 is similar

Jon8RFC:
Since 0.8.5, my 2018 Hyundai Sonata has often displayed the "no exact match" warning.  Is it due to fuzzy and/or because of the [camera fw removal from sonata_lf](https://github.com/commaai/openpilot/pull/20850/commits/f28f966c3447f1105b20c376812711f3631632e3#diff-211ba5315f2af5d2eb34a89b5ac67ade2db803c4f61a92150a0541aa41e19f51)?
This firmware is one of those inconsistent ones which rarely shows up, but here it is:  6363a0e48f515f27|2021-06-04--17-09-42--0

sunnyhaibin:
This issue also affects my 2021 Hyundai Sonata N Line. Here’s the route ID for one of the drives: `36e10531feea61a4|2021-09-22--17-51-10`

sunnyhaibin:
2021 Kia K5 affected by the same issue: `007d5e4ad9f86d13|2021-09-25--17-38-12`

pd0wm:
Just merged some improvements (https://github.com/commaai/openpilot/pull/22344/files). Let's see if that improves things.

sunnyhaibin:
> Just merged some improvements (https://github.com/commaai/openpilot/pull/22344/files). Let's see if that improves things.

Tried it already, the route IDs provided earlier were using those improvements, the issue still comes in frequently. Sometimes it would take about 5 times to have the car recognized.

sunnyhaibin:
The latest PR of the improvements seen to have made the matter worse... I had to turn on/off for almost 15 times in 1 instance for my car to get recognized.

Car: 2021 Hyundai Sonata N Line
[...]

PR #22655, opened 2021-10-22
Fingerprinting v2 support for Nissan vehicles

robin-reckmann:
Added fpv2 support for Nissan vehicles.

Tested with Nissan Leaf 2018 (IC),Nissan X-Trail 2017 and Nissan Altima.


pd0wm:
@adeebshihadeh Can you quickly test this PR on a Toyota and the Audi and post the routes? There is an overlap in engine address but the query (this PR requests `VEHICLE_MANUFACTURER_ECU_SOFTWARE_NUMBER` instead of `APPLICATION_SOFTWARE_IDENTIFICATION`) is different.

@robin-reckmann Is it really necessary to go through the whole sequence of `TESTER_PRESENT_REQUEST, DEFAULT_DIAGNOSTIC_REQUEST, EXTENDED_DIAGNOSTIC_REQUEST, NISSAN_VERSION_REQUEST_STANDARD`? What happens if you leave out the first 1, 2, or 3 steps.

robin-reckmann:
@avolmensky Could you provide a route for the xtrail?

robin-reckmann:
@pd0wm : Using only NISSAN_VERSION_REQUEST_STANDARD works! Never tested that :-) Thanks for the hint. I'll remove the first 3 commands from the sequence.

pd0wm:
> @pd0wm : Using only NISSAN_VERSION_REQUEST_STANDARD works! Never tested that :-) Thanks for the hint. I'll remove the first 3 commands from the sequence.

Ah, great! That also reduces the chance of a different car responding to the query.

robin-reckmann:
> > @pd0wm : Using only NISSAN_VERSION_REQUEST_STANDARD works! Never tested that :-) Thanks for the hint. I'll remove the first 3 commands from the sequence.
> 
> 
> 
> Ah, great! That also reduces the chance of a different car responding to the query.

I removed the first command of the nissan specific queries as wel and it still works for 13/15 Ecus using --Scan (including all the ones used for fingerprinting the leaf). Should we also remove these if the other Nissans also work without it?

pd0wm:
Yeah, the shorter the better since that will improve fingerprinting speed for all cars.

robin-reckmann:
The route using the latest commit (with reduced standard query): fd3b0ef05a984a10|2021-11-01--22-03-13

robin-reckmann:
I stumbled upon the same issues (missing frames in multi-frame packers) as discussed in #21333. For the Leaf is only happens for 0x79d. After confirming, that this happens also for other vehicles, we might remove this particular ECU to ensure reliable fingerprinting.

avolmensky:
Route from the X-Trail: `2c976ff547f21516|2021-11-03--03-26-28`

adeebshihadeh:
Here's a route from our Rav4: `4cf7a6ad03080c90|2021-11-03--15-14-40`. Our A3 isn't at the office and may not be available until Friday. Perhaps @jyoung8607 can get us a quick route.

jyoung8607:
> Our A3 isn't at the office and may not be available until Friday. Perhaps @jyoung8607 can get us a quick route.

A test startup with this branch on my Tiguan: `3cfdec54aa035f3f|2021-11-03--23-24-53`

At a quick glance everything seems normal.

pd0wm:
Looks good! Toyota engines don't respond to uds 0x22 at all (service not supported), while the VW gives an out of range error for `VEHICLE_MANUFACTURER_ECU_SOFTWARE_NUMBER`

robin-reckmann:
I'm not entirely sure that 0x742 is really the EPS since the FW looks a bit different. But I added it nevertheless.

robin-reckmann:
Altima EPS needs the diagnostic request part. Leaving it in then...

Issue #21921, opened 2021-08-13
athena: persist upload queue to disk


pd0wm:
Make sure the upload queue survives device restarts

DevinLeamy:
Hi,

I'm a new contributor to openpilot, and this looks like it would be a good first issue for me.
Can I be assigned to this issue?

pd0wm:
- Caching can be done in a persistent param. (`selfdrive/common/params.cc`), e.g. `AthenaUploadQueue`. I would just json to store a list of upload items.
 - They would indeed need to be stored real time, we can loose power at any moment. Probably upon receipt, and after every successful or failed upload in the upload handler. The queue can be loaded back in on startup.

PR #22659, opened 2021-10-22
athena: persist upload queue to disk

DevinLeamy:
**Description**
- Addresses https://github.com/commaai/openpilot/issues/21921

**Verification**
- Added relevant test to test suite.

**Also**
- Made `athenad.cancelled_uploads` reset between tests.
- Made `MockParams` persist data (like `common.params`) during a test, and reset between tests.


pd0wm:
LGTM. Need to update some stuff in the backend to make sure URLs are valid for 30 days, and can only be used once. Will merge once that is done.

pd0wm (approved):

pd0wm:
The backend now generates URLs with longer validity, so that change is merged. However, we currently have no way in useradmin/connect to cancel pending uploads, which is probably something you want as well if the queue persists. So we'll get that in as well before merging this.

adeebshihadeh:
@pd0wm are we ready to merge this now?

pd0wm:
Yes! Will merge it once 0.8.12 is in `devel-staging`.

PR #22670, opened 2021-10-23
GM: 2016 Escalade ESV

Gibby:
Added basic info the 2016 Escalade ESV and fingerprint data.

Note: None of the options from https://github.com/commaai/openpilot/wiki/Fingerprinting worked in getting a fingerprint that OP would recoginze. Ended up having to get the the fingerprint from controlsd.


adeebshihadeh (changes requested):
Remaining TODOs:
- [x] fix merge conflicts
- [x] revert extra CARS.md changes
- [x] add test route in selfdrive/test/test_routes.py

PR #22678, opened 2021-10-25
Toyota Long Enhancement: Allow 5 km/h (5 mph) speed increments on Toyota

cydia2020:
**Description**
Allow 5 km/h (or 5 mph if car uses imperial units) speed increments on Toyota, fixes https://github.com/commaai/openpilot/issues/22636

Mimics stock behaviour:
short press = 1 km/h (1 mph) increment
long press = 5 km/h (5 mph) increment

Needs to be used with https://github.com/commaai/opendbc/pull/457


AlexandreSato (approved):
its working!

PR #22695, opened 2021-10-26
boardd/safety_setter_thread: use lambda to block read parameters

deanlee:


pd0wm:
Not sure this lamdba with std::function as argument makes things easier to understand. There is some repetition in the old code, but it's clear what's going on.

PR #22744, opened 2021-10-29
params: code cleanup

deanlee:


selfdrive/common/params.h
```diff
@@ -17,67 +15,24 @@ enum ParamKeyType {
...
-
-  inline int putBool(const char *key, bool val) {
-    return put(key, val ? "1" : "0", 1);
-  }
-
-  inline int putBool(const std::string &key, bool val) {
-    return putBool(key.c_str(), val);
-  }
+  int put(const char *key, const char *val, size_t value_size);
+  inline int put(const std::string &key, const std::string &val) { return put(key.c_str(), val.data(), val.size()); }
```
adeebshihadeh, line 34: let's leave these expanded. no need to save lines like this

deanlee: ok :smile:

adeebshihadeh:
can we abstract away the `d/` too?

PR #22754, opened 2021-10-31
ui: render markdown in release notes

### LINKED ISSUE #22650: ui: render markdown in release notes

mayfieldiv:
Resolves #22650

`updated` now sets the `ReleaseNotes` param as HTML generated by a barebones mardown parsing function

QLabel already detects HTML correctly, so `offroad_alerts.cc` didn't need any updates

release-notes

Blocked by:
- [x] #22793


pd0wm:
Thanks! Looks so much better.

We need to decide what to do with this extra dependency. We can do either:
1) Ship `markdown` with the next AGNOS/NEOS, keeping this PR open in the meantime
2) Put `markdown` in pyextra, and removing it after the next AGNOS/NEOS
3) Write our own small subset of `markdown` only supporting headers and (nested) lists, to not add another dependency.

I'll get back to you about this!

selfdrive/updated.py
```diff
@@ -112,10 +113,10 @@ def set_params(new_version: bool, failed_count: int, exception: Optional[str]) -
 
   if new_version:
     try:
-      with open(os.path.join(FINALIZED, "RELEASES.md"), "rb") as f:
+      with open(os.path.join(FINALIZED, "RELEASES.md"), "r") as f:
         r = f.read()
-      r = r[:r.find(b'\n\n')]  # Slice latest release notes
-      params.put("ReleaseNotes", r + b"\n")
+      r = r[:r.find('\n\n')]  # Slice latest release notes
+      params.put("ReleaseNotes", markdown.markdown(r, tab_length=2))
```
pd0wm, line 119: Can you add another layer of exception handling here, and put in the raw release notes in case `markdown.markdown()` throws an exception?

deanlee:
>     1. Ship `markdown` with the next AGNOS/NEOS, keeping this PR open in the meantime
> 
>     2. Put `markdown` in pyextra, and removing it after the next AGNOS/NEOS
> 
>     3. Write our own small subset of `markdown` only supporting headers and (nested) lists, to not add another dependency.

QTextDocument support markdown in Qt 5.14 https://doc.qt.io/qt-5/qtextdocument.html#setMarkdown . I think option 3 may be beter.

pd0wm:
The Qt markdown support seems to rely on https://github.com/mity/md4c under the hood. As a fourth option we could throw that library in `third_party`.

mayfieldiv:
md4c was the other approach I was considering if adding python dependencies wasn't straightforward

deanlee:
It may not be worth importing an entire library for such a small feature

selfdrive/test/test_updated.py
```diff
@@ -53,7 +54,7 @@ def setUp(self):
       f"cd {self.basedir} && scons -j{os.cpu_count()} cereal/ common/"
     ])
 
-    self.params = Params(db=os.path.join(self.basedir, "persist/params"))
```
mayfieldiv, line 56: I'm pretty sure this was a bug in the test, since the parameter seems to be named `d`

selfdrive/test/test_updated.py
```diff
@@ -176,6 +177,13 @@ def _check_update_state(self, update_available):
     self.assertEqual(update == "1", update_available, f"UpdateAvailable: {repr(update)}")
     self.assertEqual(self._read_param("UpdateFailedCount"), "0")
 
+    if update_available:
+      # validate our simple markdown parser produces the same output as `markdown` from pip
+      release_notes = self._read_param("ReleaseNotes")
+      with open(os.path.join(self.basedir, "RELEASES.md")) as f:
+        r = f.read().split("\n\n", 1)[0]
+      self.assertEqual(markdown.markdown(r, tab_length=2), release_notes)
+
```
mayfieldiv, line 186: I wasn't able to fully run these tests locally, but I'm pretty sure this check will work correctly. Are these tests only run on comma devices? I assume the `markdown` library won't be a problem here, since I think this file is the unittest @pd0wm was referring to in https://github.com/commaai/openpilot/pull/22768#issuecomment-956503933

pd0wm: I'd put the test in a separate unit test that we only run in github actions. Make sure you put the `markdown` dependency under dev-packages.

adeebshihadeh:
What's the difference between `markdown-it-py` and `markdown`?

mayfieldiv:
no meaningful difference, I just noticed it was already in the Pipfile.lock so using it would produce a bit smaller diff for the PR

mayfieldiv:
I feel like the build failure is a transient error with an existing unrelated test

```
FAIL: selfdrive.athena.tests.test_athenad_old.test_athena
----------------------------------------------------------------------
  File "/tmp/openpilot/selfdrive/athena/tests/test_athenad_old.py", line 36, in expect_athena_starts
    athenad_pid = subprocess.check_output(["pgrep", "-P", manage_athenad_pid], encoding="utf-8").strip()
  File "/root/.pyenv/versions/3.8.10/lib/python3.8/subprocess.py", line 415, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/root/.pyenv/versions/3.8.10/lib/python3.8/subprocess.py", line 516, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['pgrep', '-P', b'1805']' returned non-zero exit status 1.

During handling of the above exception, another exception occurred:

  File "/root/.pyenv/versions/3.8.10/lib/python3.8/site-packages/nose/case.py", line 198, in runTest
    self.test(*self.arg)
  File "/tmp/openpilot/selfdrive/athena/tests/test_athenad_old.py", line 82, in test_athena
    athenad_pid = expect_athena_starts()
  File "/tmp/openpilot/selfdrive/athena/tests/test_athenad_old.py", line 40, in expect_athena_starts
    assert False, f"Athena did not start within {timeout} seconds"
AssertionError: Athena did not start within 30 seconds

----------------------------------------------------------------------
Ran 1 test in 30.025s

FAILED (failures=1)
```

pd0wm:
let me merge https://github.com/commaai/openpilot/pull/22793 first

PR #22756, opened 2021-10-31
enable wshadow

### LINKED ISSUE #22714: Enable Wshadow flag
Need to set the flag and fix all the violations. Submodules should be done first

mayfieldiv:
Resolves #22714
Not sure how strict of naming conventions y'all want when there's a collision between a parameter and a class member (e.g. `ack_` vs `ack`, `init_rgb_type` vs `rgb_type`, `state` vs `s`)

Blocked by:
- [x] https://github.com/commaai/cereal/pull/214
- [x] https://github.com/commaai/rednose/pull/21

Related:
- [x] https://github.com/commaai/opendbc/pull/461

### DIFF
--- a/selfdrive/ui/qt/maps/map.cc
+++ b/selfdrive/ui/qt/maps/map.cc
@@ -576,9 +576,9 @@ void MapInstructions::updateDistance(float d) {
   distance->setText(distance_str);
 }
 
-void MapInstructions::showError(QString error) {
+void MapInstructions::showError(QString error_text) {
   primary->setText("");
-  distance->setText(error);
+  distance->setText(error_text);
   distance->setAlignment(Qt::AlignCenter);
 
   secondary->setVisible(false);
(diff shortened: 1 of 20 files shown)


pd0wm:
Can you add the flag to the cereal sconstruct so we catch any errors immediately? Same for rednose and opendbc.

selfdrive/ui/qt/maps/map.cc
```diff
@@ -576,9 +576,9 @@ void MapInstructions::updateDistance(float d) {
   distance->setText(distance_str);
 }
 
-void MapInstructions::showError(QString error) {
+void MapInstructions::showError(QString error_text) {
```
pd0wm, line 579: This was an actual bug!

pd0wm: Ran into it in https://github.com/commaai/openpilot/pull/22767

mayfieldiv: haha good to hear!

PR #22758, opened 2021-10-31
loggerd: single lock file per segment

deanlee:
The qlog/rlog is always closed last, one .lock file in logger.cc is enough


adeebshihadeh:
I support this, but let's formalize it a bit more. Maybe just make a single `.lock` file in the directory and remove the rlog and qlog lock files?

adeebshihadeh (changes requested):
I support this, but let's formalize it a bit more. Maybe just make a single `.lock` file in the directory and remove the rlog and qlog lock files?

adeebshihadeh:
Maybe loggerd itself should make the lock? We want to make sure the cases where either the rlog or cameras don't rotate are handled.

deanlee:
Think it's safer to create/remove lock in loggerstate

adeebshihadeh:
I didn't realize that these locks are used in a bunch of other places besides openpilot's uploader, such as the connect web app. This should stay as it is.

PR #22763, opened 2021-11-01
loggerd: async clear locks

deanlee:


adeebshihadeh:
Why do we want this? Is this actually slow, especially in a real world case with only one segment lock?

deanlee:
The logs are kept locally until the disk space reaches the limit, there may be thousands of directories that need to be traversed. clear_locks can be very slow on the device.

pd0wm:
Think we should just move this to the bootlog logger and only clear the locks once on boot. I like the simplicity of the old remove locks function. By running this async we have to deal with all the complexity of excluding the current route.

pd0wm:
https://github.com/commaai/openpilot/pull/23060

PR #22767, opened 2021-11-01
Move navigation into separate daemon

pd0wm:


selfdrive/ui/qt/maps/map.cc
```diff
@@ -177,61 +167,36 @@ void MapWindow::timerUpdate() {
...
-          if (d > REROUTE_DISTANCE) {
-            clearRoute();
-          }
-        }
+      if (localizer_valid) {
+        m_map->setPitch(MAX_PITCH); // TODO: smooth pitching based on maneuver distance
+        emit distanceChanged(i.getManeuverDistance()); // TODO: combine with instructionsChanged
+        emit instructionsChanged(i);
       }
+    } else {
```
pd0wm, line 180: UI seems laggy when there is no gps. Is `clearRoute` an expensive call?

pd0wm: Can't reproduce?

PR #22776, opened 2021-11-02
GM: units of steer command for driver and eps torque

hewers:
Not Newton-meters. Much easier to use the same unit in steering for now: -300 to 300, steer command units.

**Note: Behavior change: steer pressed threshold dropped from 100 to 50, to match the STEER_DRIVER_ALLOWANCE value where openpilot limits how much it opposes the driver. Want this as another PR?**

opendbc PR: https://github.com/commaai/opendbc/pull/462


qadmus:
opendbc pull request: https://github.com/commaai/opendbc/pull/462

qadmus:
The Nissan Leaf change appears because opendbc/master is ahead of openpilot/opendbc

PR #22777, opened 2021-11-02
replay: fix qcam download error

deanlee:


pd0wm:
Can we also add a test for loading qcams so we don't accidentally break it again?

selfdrive/ui/replay/route.h
```diff
@@ -28,7 +28,7 @@ class Route {
   inline const QString &name() const { return route_.str; }
   inline const RouteIdentifier &identifier() const { return route_; }
   inline const std::map<int, SegmentFile> &segments() const { return segments_; }
-  inline const SegmentFile &at(int n) { return segments_.at(n); }
+  inline SegmentFile &at(int n) { return segments_.at(n); }
```
deanlee, line 31: Temporarily modify the return type for test case. I'll restore it back after the class Segment can support switch between qlog&log

Issue #22574, opened 2021-10-16
Put camera/lens type enum in VisionIpcBufExtra


pd0wm:
This can be used to choose intrinsics instead of relying on the combination of frame size and stream type (or worse, the HW type for which things were compiled). Not sure yet about the best way to handle this in replays. Do we also put this enum in EncodeIdx, and write some heuristics for older routes?

deanlee:
It is better to find out a way without adding new fields in EncodeIdx

Issue #22230, opened 2021-09-14
ui: choose fcam_intrinsic matrix, zoom and y_offset based on video size


pd0wm:

pd0wm:
@deanlee Are you interested in this as part of the replay project? Now that we have the camera view widget this should be a lot more doable.

PR #22787, opened 2021-11-04
fix model prediction issue in PC mode

eFiniLan:
when running OP in PC mode, the path prediction is messed up, so use Hardware::TICI() to check instead of Hardware::EON()


adeebshihadeh:
This was done to fix the UI for C3 replays. The real fix is either #22230 or #22574.

PR #22788, opened 2021-11-04
replay: new flags

deanlee:
added new flags:
`--qcam` load qcamera file 
`--yuv` send yuv frame

### DIFF
--- a/selfdrive/ui/replay/camera.h
+++ b/selfdrive/ui/replay/camera.h
@@ -8,7 +8,7 @@
 
 class CameraServer {
 public:
-  CameraServer(std::pair<int, int> camera_size[MAX_CAMERAS] = nullptr);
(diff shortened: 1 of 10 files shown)


selfdrive/ui/replay/util.h
```diff
@@ -4,6 +4,16 @@
 #include <ostream>
 #include <string>
 
+enum REPLAY_FLAGS {
```
adeebshihadeh, line 7: don't think these should be defined in util.h

selfdrive/ui/replay/camera.h
```diff
@@ -38,4 +38,5 @@ class CameraServer {
   };
   std::atomic<int> publishing_ = 0;
   std::unique_ptr<VisionIpcServer> vipc_server_;
+  bool send_yuv_;
```
adeebshihadeh, line 41:
```suggestion
  bool send_yuv;
```

PR #22789, opened 2021-11-04
ui: singleton uistate

deanlee:
- [x] combine class `QUIState`&`UIState` into one.
- [x] singleton `UIState` for simple access.
- [x] simplify signals and slots connections


selfdrive/ui/qt/home.cc
```diff
@@ -19,7 +19,7 @@ HomeWindow::HomeWindow(QWidget* parent) : QWidget(parent) {
 
   sidebar = new Sidebar(this);
   main_layout->addWidget(sidebar);
-  QObject::connect(this, &HomeWindow::update, sidebar, &Sidebar::updateState);
+  QObject::connect(uiState(), &UIState::uiUpdate, sidebar, &Sidebar::updateState);
```
adeebshihadeh, line 22: if we're using uiState(), doesn't it make more sense to do these connections in Sidebar?

selfdrive/ui/qt/window.cc
```diff
@@ -37,8 +33,8 @@ MainWindow::MainWindow(QWidget *parent) : QWidget(parent) {
     main_layout->setCurrentWidget(onboardingWindow);
   }
 
-  QObject::connect(&qs, &QUIState::uiUpdate, &device, &Device::update);
-  QObject::connect(&qs, &QUIState::offroadTransition, [=](bool offroad) {
+  QObject::connect(uiState(), &UIState::uiUpdate, &device, &Device::update);
```
adeebshihadeh, line 36: move to device?

adeebshihadeh:
Not sure how I feel about the singleton, but this is a nice cleanup!

PR #22799, opened 2021-11-05
driving/get_best_prediction: use std::max_element

deanlee:


pd0wm:
Think "Explicit is better than implicit" applies here. It doesn't make the code faster or easier to read.

PR #22807, opened 2021-11-05
Mazda: fix occasional main cruise disable

Jafaral:

### DIFF
--- a/selfdrive/car/mazda/carcontroller.py
+++ b/selfdrive/car/mazda/carcontroller.py
@@ -33,13 +33,14 @@ class CarController():
         can_sends.append(mazdacan.create_button_cmd(self.packer, CS.CP.carFingerprint, CS.crz_btns_counter, Buttons.RESUME))
 
     if c.cruiseControl.cancel or (CS.out.cruiseState.enabled and not c.enabled):
-      # if brake is pressed, let us wait >20ms before trying to disable crz to avoid
+      # If brake is pressed, let us wait >70ms before trying to disable crz to avoid
       # a race condition with the stock system, where the second cancel from openpilot
-      # will disable the crz 'main on'
+      # will disable the crz 'main on'. crz ctrl msg runs at 50hz. 70ms allows us to
+      # read 3 messages and most likely sync state before we attempt cancel.
       self.brake_counter = self.brake_counter + 1
-      if frame % 20 == 0 and not (CS.out.brakePressed and self.brake_counter < 3):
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
I'd like to understand this better before we delay sending the cancel 250ms. Are we sending multiple cancels? Is there another message we can use for cancellation?

Jafaral:
@adeebshihadeh, let me run with this a couple more days to be sure. If the issue stopped happening then it wouldn't be because we are sending two cancels. The fact that I only see this happens with brake and not gas makes me think that it is more of a timing issues between the stock system reading the state and responding with cancel and OP also trying to do the same. The cruise state signal runs at 10Hz, slow enough for state to get out of sync. Hence I ended up going with 250ms delay in that case.

Jafaral:
Not a single cruise disable since I applied this patch.  BTW, 250ms is nothing more than an upper bound. In most situation we are going to get the update as soon as we we pick up cruise state (10hz)  from the stock system. I'd say on average that would be ~150ms.

adeebshihadeh:
It's an upper bound to cancel, assuming the PCM handles this. The 250ms is now the minimum time until openpilot tries to cancel. I'd like a more targeted solution than this. If we think it's the delay for a cruise state update, then we should explicitly check that and trigger on cruise state updates.

Jafaral:
* We send cancel cruise at 5hz, the 250ms was to ensure we miss two cycles. We can take that down to 210ms and probably still make the see guarantee.
* I don't know how long the stock system takes to react to barke by canceling cruise. Let us say 50ms for the sake of argument. In worst case scenario it will take us 50ms+100ms(10hz cruise update) = 150ms to get the cancel cruise and disengage. In practice it is probably half of that time. Over the last 5 days, I didn't see a single out of sync cruise state between stock and OP. I also didn't see any delay whatsoever in disengage events when pressing brake. It is instant, at least from a human perception point of view.

Jafaral:
@adeebshihadeh , not  sure  where I came up with 10hz for the crz signal, I just checked , we  have it at 50hz. That means we shouldn't have more than 20ms delay to sync with the stock crz state.

Jafaral:
@adeebshihadeh , here is a new attempt to fix this issue. instead of waiting 250ms, we now only wait 70ms, this wait is only for brake press. That is enough to read 3 `CRZ_CTRL` msgs which runs at 50hz  and sync cruise state. On the flip side, I doubled the rate we send cancel signal up to 10hz up from 5hz. That means OP is more aggressive now  in canceling cruise for any disengage events.

PR #22808, opened 2021-11-05
Mazda: disable hud alerts until we find a way to silence audible alerts 

Jafaral:


selfdrive/car/mazda/mazdacan.py
```diff
@@ -65,6 +65,8 @@ def create_steering_control(packer, car_fingerprint, frame, apply_steer, lkas):
 
 def create_alert_command(packer, cam_msg: dict, ldw: bool, steer_required: bool):
   values = copy.copy(cam_msg)
+  # TODO: disable alerts for now until we find a way to silence audible warnings
+  steer_required = False
```
adeebshihadeh, line 69: this should go in the carcontroller

Jafaral: sounds good

adeebshihadeh:
I’ll merge this for now since I agree that it’s super annoying, but I expect this will be addressed soon. I don’t care so much about the actual HUD alert, the most important thing is the dash alert.

Jafaral:
> I’ll merge this for now since I agree that it’s super annoying, but I expect this will be addressed soon. I don’t care so much about the actual HUD alert, the most important thing is the dash alert.

I don't disagree @adeebshihadeh. BTW, I just pushed a commit to enable hud alerts for steer lockout

adeebshihadeh:
I think this makes more sense.

PR #22810, opened 2021-11-06
Handle Toyota brake hold active

### LINKED ISSUE #22791: Integral can build up with brake hold
### Describe the bug

If you engage brake hold and then engage openpilot (since user isn't pressing the brake), the long PI controller is allowed to build up integral while we're held at a standstill without a lead, and upon pressing the resume button the car is being sent 1.44 m/s/s acceleration instead of 1.2 m/s/s causing overshoot to 2 m/s/s.

Screenshot from 2021-11-04 07-19-06

comma three, 2022 Toyota Camry Hybrid

f15e3c37c118e841|2021-11-04--06-25-15--1

### openpilot version

0.8.10

### Additional info

_No response_

sshane:
Needs: https://github.com/commaai/opendbc/pull/465

Fixes: #22791


selfdrive/car/toyota/interface.py
```diff
@@ -355,16 +355,19 @@ def update(self, c, can_strings):
...
-      events.add(EventName.belowEngageSpeed)
-      if c.actuators.accel > 0.3:
-        # some margin on the actuator to not false trigger cancellation while stopping
-        events.add(EventName.speedTooLow)
-      if ret.vEgo < 0.001:
-        # while in standstill, send a user alert
-        events.add(EventName.manualRestart)
+    if self.CS.CP.openpilotLongitudinalControl:
+      if self.CS.brake_hold:
+        events.add(EventName.brakeHold)
```
sshane, line 360: Move this to common events?

adeebshihadeh: yes, added to carState for this https://github.com/commaai/cereal/commit/aa349f6c314b1fcc7f5a42e6f4bfe0d466335634

adeebshihadeh:
LGTM - any reason it's still a draft?

BrettLynch123:
Hi I think this change introduced unwanted bahaviour. In my 2020 Toyota corolla hybrid if I have break hold on at the traffic lights and i hit the resume button in the past it would engage openpilot and sit behind the car infront until it moved. This change makes it so that if im behind the car and have break hold on it disengages the break hold. openpilot beeps and shows error message and then the car starts rolling forward uncontrolled.
I would have either expected the break hold not disengage or not show the warning and openpilot take over. 
Scared me first time it happened. didn't know what had happened. Should I create a new issue? I'm not sure how to fix. or even if it would be a priority to do so.

PR #22818, opened 2021-11-08
Fix pygame and knack dependencies

jyoung8607:
Update `Pipfile` and `Pipfile.lock` to fix two dependencies needed for a clean Ubuntu 20.04 install:

* pygame 2.0.0 dev8 is [no longer indexed on PyPI](https://pypi.org/project/pygame/#history); updated to 2.0.0 release. This was the most reasonable minimum change. Don't know if there's a good reason not to let this version float to current.
* Revert knack to 0.8.2 because 0.8.3 was withdrawn from PyPI due to a breaking conflict with the Azure CLI tools (microsoft/knack#252) which are also used by this project.
* Includes a variety of other minor updates by nature of regenerating Pipfile.lock the proper way.


pd0wm:
The opencv update seems to generate a version mismatch with PyQt. We don't use any of the UI features of cv2, so switching to the headless version makes sense I think.

PR #22819, opened 2021-11-08
Fix momentary false blind spot alerts when changing lanes

jyoung8607:
**Description**

Fix a bug that, under very specific circumstances, causes momentary false BSM alerts when changing lanes.

**Verification**

Haven't road-tested this fix yet. Will try to do so this week. Feedback welcome.

**Route**

Route: `3cfdec54aa035f3f|2021-11-07--15-35-07` segment 22 demonstrates the problem.

**Additional context**

Have been hitting this occasionally for months. This weekend it ticked me off enough to investigate. Scenario:

PXL_20211108_043307624

- Driving with openpilot engaged on road with >= 3 lanes of travel
- Ego vehicle at position 1 performs a normal ALC to the right **using comfort/one-touch turn signal** 
- An arbitrary amount of time passes with openpilot engaged and driving normally
- Ego vehicle arrives at position 2 with vehicle B in **right blind spot**
- Ego vehicle initiates ALC **to the left**
- openpilot momentarily shows laneChangeBlocked audio/text (incorrect)
- openpilot then completes the lane change normally to position 3 with no other intervention

All `carState` data for left/right turn signals and left/right blind spot monitors are working correctly.

After the first ALC, the state machine makes it through Finishing to Off without resetting `lateralPlan.LaneChangeDirection`. It remains set to right. If the signal stalk was left on all the way through the finish and THEN turned off, the state machine would have moved to `preLaneChange` which resets it properly.

At the moment of the second ALC, `lateralPlan.laneChangeState` transitions to `preLaneChange`, but doesn't execute the code for that state or otherwise set the direction until the next run-through. Controls then has a lateral plan to change lanes right instead of left, and incorrectly warns the driver. The next and subsequent lateralPlan messages are correct and the ALC completes.

The bug can be triggered with an arbitrarily long delay between the first and second lane changes, provided openpilot remains engaged the entire time.


VirtuallyChris:
I can confirm I’ve seen this on the Audi a few times recently.

jyoung8607:
Test behavior looks good to me: `3cfdec54aa035f3f|2021-11-09--10-21-17`

**Before: lane change direction not reset after one-touch lane changes**

lateral-plan-before

**After: lane change direction reset to `none` after lane changes**

lateral-plan-after

jyoung8607:
> I can confirm I’ve seen this on the Audi a few times recently.

I've seen it on both my VWs, and for the record, this should be repeatable on other car ports as well.

PR #22832, opened 2021-11-09
Longcontrol: do not clip vEgo to minSpeedCan

sshane:
Just a draft PR to open discussion. I found that the reason some hybrids and other TSS2 vehicles are slow to start moving in stop and go traffic is that the PID controller believes the car is moving at 0.3 m/s while stopped, so proportional and integral keep the car stopped until the desired plan speed (and/or desired acceleration) passes that.

Also, does it still make sense to have `minCanSpeed`, when it's used to compare against v_target, as well as the ego's speed, that's already gone through a kalman filter? The actual minimum can speed on the wheel speed sensors can also go down to 0.04 m/s/s or less before dropping to 0.

Comparison of stock (on left) to openpilot before change: 
Screenshot from 2021-11-05 02-28-10


pd0wm:
Are there even jumps in vEgo on start/stop since it's kalman filtered?

sshane:
It seems from krkeegan's test, just sending acceleration earlier doesn't fix the slow starts on TSS2, I think we also might need to reduce the standstill acceleration. Toyota usually sends -0.4 m/s/s when on flat ground, but I'm not sure about sharp hills.

I'll get some routes with this change, as well as this + changing `stopAccel` to what Toyota requests when stopped.

pd0wm:
Even if it doesn't help it's a nice cleanup anyway.

krkeegan:
I will try out the stopAccel change today.

If that doesn't work, am I safe to roll back to 0.8.9, I know there have been concerns about downgrading Agnos?  Perhaps some data from that would help us determine why things are different all of a sudden?

HaraldSchafer:
LGTM

HaraldSchafer:
I think a big problem is just there is extreme lag in acceleration requests when starting from stop, which we don't account for anywhere. Perhaps some lag compensation that takes this into account could help.

sshane:
I think the car responds quicker when sending less deceleration when at a standstill, as well as setting startAccel to 0. Plus with it overshooting large accelerations, the system might actually be sensitive to the derivative of the request (jerk).

I have some comparison data, just need to go home and upload it.

krkeegan:
I changed `stopAccel` to -.4 and it doesn't look any better.  I am leaving town on Saturday, but if I have a chance before then I will try to put 0.8.9 back on and take a look at what that looked like.

image

sshane:
Hmm your car takes about 1.2 seconds to start moving from vPid being non-zero, while mine only 0.855 seconds. This is with this PR's change as well as setting `stopAccel` to -0.4 and `startAccel` to 0, copying the TSS2 ACC. However, the hysteresis in carcontroller seems to be messing with the desired requests we want to send as it actually sent near -1 m/s/s while stopped, I hadn't accounted for that.

Screenshot from 2021-11-10 23-23-41

(`./juggle.py "f15e3c37c118e841|2021-11-10--17-02-26" 58 2 --can`)

sshane:
Here's how the two systems handle coming to a stop and re-accelerating: https://imgur.com/a/3Y0Kk6m

Stock moves to its standstill acceleration no matter what the requested acceleration is when it stopped, and when re-accelerating, the request instantly jumps up to 0 m/s/s.

Do we want to mimic this in openpilot? Maybe move this to another PR?

pd0wm:
Yes, let's mimic that in a new PR. This large deceleration was probably needed when we didn't enter the standstill state.

PR #22833, opened 2021-11-09
loggerd: new class FFmpegEncoder

deanlee:
do ffmpeg encoding and remuxing in class FFmpegEncoder

### DIFF
--- a/release/files_common
+++ b/release/files_common
@@ -299,6 +299,7 @@ selfdrive/proclogd/proclog.cc
 selfdrive/proclogd/proclog.h
 
 selfdrive/loggerd/SConscript
+selfdrive/loggerd/encoder.cc
 selfdrive/loggerd/encoder.h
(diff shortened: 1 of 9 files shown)


adeebshihadeh:
Is there any immediate benefit to this? Seems like a nice cleanup, but we don't have the time to spend reviewing and validating something like this needs. In addition to that, we need as much comma three data as we can get, so we can't ship a subtly broken logger for a release.

Currently, we have offline checks that run on incoming data to track logger quality, but those metrics lag the actual changes we make since so few people run master and many don't run the latest commits. We may be able to start considering stuff like this once all those checks are also run in openpilot CI. We'd merge PRs that only add tests (#22095) and possibly minor refactors to support those tests.

deanlee:
The main purpose of this pr is  to write test cases for ffmpeg encoding and remuxing. I have found several problems through this refactor. I will submit the fix separately.

PR #22841, opened 2021-11-10
controlsd: alert on cruise mismatch for >1s

### LINKED ISSUE #22780: Alert if cruiseState.enabled does not match openpilot for >1 second

adeebshihadeh:
fixes #22780


selfdrive/test/test_onroad.py
```diff
@@ -112,7 +112,7 @@ def check_cpu_usage(first_proc, last_proc):
       cpu_usage = cpu_time / dt * 100.
       if cpu_usage > max(normal_cpu_usage * 1.15, normal_cpu_usage + 5.0):
         # cpu usage is high while playing sounds
-        if proc_name == "./_soundd" and cpu_usage < 25.:
+        if proc_name == "./_soundd" and cpu_usage < 65.:
```
adeebshihadeh, line 115: want to look into this weird CPU usage? maybe it has to do with the audio backend since this doesn't seem to happen on c3? c2 uses the opensl es backend

cydia2020:
This creates an annoying issue on Toyotas whereby if the cruise is cancelled by the brakes while the car is in standstill, it throws an alert.

https://youtube.com/shorts/A9rEyBFc_-g

PR #22842, opened 2021-11-10
ui: auto scaling on resolution below 1920 x 1080

kernyan:
note special case when on PC and screen resolution is exactly 1920 x 1080, we intentionally avoid fixed sized window to account for taskbar/menubar area

Revised from #22831

### DIFF
--- a/selfdrive/hardware/hw.h
+++ b/selfdrive/hardware/hw.h
@@ -13,7 +13,7 @@
 class HardwarePC : public HardwareNone {
 public:
   static std::string get_os_version() { return "openpilot for PC"; }
-  static constexpr bool PC() { return true; }
+  static bool PC() { return true; }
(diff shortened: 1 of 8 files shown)


pd0wm:
@kernyan Can you set a maximum size to (2160x1080)?

@adeebshihadeh Our ui looks terrible on less than 1080p, and doesn't handle dynamic resizing well. Do we want to merge this PR and fix that incrementally?

adeebshihadeh:
> @adeebshihadeh Our ui looks terrible on less than 1080p, and doesn't handle dynamic resizing well. Do we want to merge this PR and fix that incrementally?

I just thought this would be nice for cases like my MBP where the UI window is bigger than the screen. I'm not too interested in actually making the UI look good.

PR #22843, opened 2021-11-10
ui: async uploading texture with PBO

deanlee:
1. fixed the "slow frame time". the FPS is stable at 20hz
2. CPU usage goes from ~22% to ~9%.

### DIFF
--- a/selfdrive/common/visionimg.cc
+++ b/selfdrive/common/visionimg.cc
@@ -55,9 +55,17 @@ EGLImageTexture::EGLImageTexture(const VisionBuf *buf) {
   glBindTexture(GL_TEXTURE_2D, frame_tex);
   glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB, buf->width, buf->height, 0, GL_RGB, GL_UNSIGNED_BYTE, nullptr);
   glGenerateMipmap(GL_TEXTURE_2D);
+
+  glGenBuffers(1, &frame_buf);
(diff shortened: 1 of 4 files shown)


deanlee:
the cpu usage of UI goes from 29% to 8% on my laptop.

adeebshihadeh:
Nice, how do the frame timings look?

deanlee:
looks much better

adeebshihadeh:
Did you try to remove the copy instead?

deanlee:
Not yet.  my plan is to display yuv in opengl first, and then see how to remove the copy. what do you think?

adeebshihadeh:
Big fan of removing the RGB vipc buffer! This PR is irrelevant though if we're able to remove the copy, right?

deanlee:
Yes, but it may take a lot of time. I'll try to display yuv frame in UI first.

pd0wm:
Looks like CPU usage on C3 went up from 20% to 30%. Anything we can do about that?

deanlee:
how about remove the glflush in vipc_thread, will it improve the cpu usage?

pd0wm:
Removing the glFlush seems to bring cpu usage back down to 20%. What is the effect of removing that?

deanlee:
glmap will block wait until OpenGL is done. I use glflush to make glmap return faster.

deanlee:
I'll try to improve the cpu usage on c3

PR #22846, opened 2021-11-10
controlsd: optimize alert management

adeebshihadeh:
Goes from 75% of a core to 40% with this patch applied. Building the capnp list for all those events still eats through a ton of CPU though.  
``` python
comma@tici:/data/openpilot/selfdrive/controls$ git diff .
diff --git a/selfdrive/controls/controlsd.py b/selfdrive/controls/controlsd.py
index 92785b07d..53171af96 100755
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -185,6 +185,9 @@ class Controls:
     self.events.add_from_msg(CS.events)
     self.events.add_from_msg(self.sm['driverMonitoringState'].events)

+    for _ in range(50):
+      self.events.add(EventName.localizerMalfunction)
+
     # Handle startup event
     if self.startup_event is not None:
       self.events.add(self.startup_event)
```

This has a slight behavior change that I think makes sense. Instead of using different durations for the car HUD, alert text, and sound, the max of the three durations is used for all three. I imagine this was done to control the amount of beeps when the car's chimes were used, but `soundd` handles this properly now. The only effect of this change is the car HUD and alert text durations will now be the same.

I think it still makes sense to specify the text and sound durations separately, but we can probably get rid of `duration_hud` altogether as it seems like this should match `duration_text`.


adeebshihadeh:
On second thought, I think it makes more sense to only use a single duration for each alert that would be used for the sound, alert text, and HUD. What do you think @pd0wm?

pd0wm:
A single duration makes sense. Think the sound duration was used as a hack anyway.

selfdrive/controls/lib/alertmanager.py
```diff
@@ -44,42 +51,35 @@ def clear_current_alert(self) -> None:
...
-        self.alert_rate = current_alert.alert_rate
+    current_alert = AlertEntry()
+    for k, v in self.activealerts.items():
+      if v.alert is None:
+        continue
+
+      if v.alert.event_type == clear_event_type:
+        self.activealerts[k].end_frame = -1
+
+      # sort by priority first and then by end_frame
```
pd0wm, line 67: why sort by end frame? Think it makes more sense to do based on start_frame (which we can store when we create the alert).

pd0wm (approved):

PR #22848, opened 2021-11-10
loggerd: trigger rotate on frame id instead of frame count

adeebshihadeh:


selfdrive/loggerd/loggerd.cc
```diff
@@ -138,18 +139,18 @@ bool sync_encoders(LoggerdState *state, CameraType cam_type, uint32_t frame_id)
     return false;
   } else {
     // Small margin in case one of the encoders already dropped the next frame
-    uint32_t start_frame_id = state->latest_frame_id + 2;
-    bool synced = frame_id >= start_frame_id;
+    state->start_frame_id = state->latest_frame_id + 2;
```
pd0wm, line 142: It seems like this branch can only be hit when all the encoders are ready and `latest_frame_id` should no longer change. But, maybe use `update_max_atomic` atomic here as well to be sure.

gregjhogan: I don't think we need this new `start_frame_id`.  When frame `state->latest_frame_id + 2` is dropped by a camera we currently start out of sync, but now we will recover the next segment.  So, why not get rid of the `+ 2` and rename `latest_frame_id` to be `start_frame_id` to make things simpler?  If there is some other reason to offset the start frame id to something in the future, rename the variable and add the offset when calling `update_max_atomic` above.

selfdrive/loggerd/loggerd.cc
```diff
@@ -187,20 +188,20 @@ void encoder_thread(const LogCameraInfo &cam_info) {
...
         }
-      }
 
-      if (cam_info.trigger_rotate && (cnt >= SEGMENT_LENGTH * MAIN_FPS)) {
-        // trigger rotate and wait logger rotated to new segment
-        ++s.waiting_rotate;
-        std::unique_lock lk(s.rotate_lock);
-        s.rotate_cv.wait(lk, [&] { return s.rotate_segment > cur_seg || do_exit; });
+        const int frames_per_seg = SEGMENT_LENGTH * MAIN_FPS;
+        if (extra.frame_id >= (cur_seg * frames_per_seg) + s.start_frame_id) {
```
gregjhogan, line 193: This seems wrong, it looks like it checks if `extra.frame_id` is higher than the first frame id of the current segment?

PR #22853, opened 2021-11-10
Params: move keys to seperate file

deanlee:


adeebshihadeh:
Why not put them in params.h?

deanlee:
> Why not put them in params.h?

to limit the scope. otherwise a global variable parameters will be exposed everywhere. but it's only used in params.cc

PR #22857, opened 2021-11-10
locationd: Set ECEF_POS valid flag to false if in no-gps mode

nuwandavek:
The `last_gps_fix` is initialized to 0 in the Localizer, and updated with the `current_time` when proceessing GPS messages. So, we can check if that's greater than 0. This flag can now be used in navigation to verify if the filter has ever received a GPS message.


selfdrive/locationd/locationd.cc
```diff
@@ -131,7 +131,7 @@ void Localizer::build_live_location(cereal::LiveLocationKalman::Builder& fix) {
 
   // write measurements to msg
   init_measurement(fix.initPositionGeodetic(), fix_pos_geo_vec, nans, true);
```
pd0wm, line 133: Also set flag on positionGeodetic since it's also wrong

nuwandavek: Done

PR #22859, opened 2021-11-10
Cleanup toyota tuning

haraschax:

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -112,7 +112,7 @@ class CarInterface(CarInterfaceBase):
       ret.steerRatio = 16.0
       tire_stiffness_factor = 0.8
       ret.mass = 4700. * CV.LB_TO_KG + STD_CARGO_KG  # 4260 + 4-5 people
-      set_lat_tune(ret.lateralTuning, LatTunes.PID_F)
+      set_lat_tune(ret.lateralTuning, LatTunes.PID_G)
 
     elif candidate in [CAR.HIGHLANDER, CAR.HIGHLANDERH]:
       stop_and_go = True
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -1,11 +1,15 @@
 #!/usr/bin/env python3
 from cereal import car
 from selfdrive.config import Conversions as CV
-from selfdrive.car.toyota.values import Ecu, CAR, TSS2_CAR, NO_DSU_CAR, MIN_ACC_SPEED, PEDAL_HYST_GAP, CarControllerParams
+from selfdrive.car.toyota.tunes import long_tunes
+from selfdrive.car.toyota.values import Ecu, CAR, TSS2_CAR, NO_DSU_CAR, MIN_ACC_SPEED, CarControllerParams
 from selfdrive.car import STD_CARGO_KG, scale_rot_inertia, scale_tire_stiffness, gen_empty_fingerprint, get_safety_config
 from selfdrive.car.interfaces import CarInterfaceBase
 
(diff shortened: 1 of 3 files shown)


selfdrive/car/toyota/tunes.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+from selfdrive.car.toyota.values import MIN_ACC_SPEED, PEDAL_HYST_GAP
+
+
+
+###### LONG ######
+def set_long_tune(tune, name):
+  if name == 'PEDAL':
```
adeebshihadeh, line 8: use an enum for the tunes https://docs.python.org/3/library/enum.html

selfdrive/car/toyota/tunes.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+from selfdrive.car.toyota.values import MIN_ACC_SPEED, PEDAL_HYST_GAP
+
+
+
+###### LONG ######
+def set_long_tune(tune, name):
```
adeebshihadeh, line 7: return a car.CarParams.LongitudinalPIDTuning
``` python
from cereal import car
tune = car.CarParams.LongitudinalPIDTuning.new_message()
```

selfdrive/car/toyota/interface.py
```diff
@@ -1,13 +1,17 @@
...
-from selfdrive.car.toyota.values import Ecu, CAR, TSS2_CAR, NO_DSU_CAR, MIN_ACC_SPEED, PEDAL_HYST_GAP, CarControllerParams
+from selfdrive.car.toyota.tunes import set_long_tune, set_lat_tune
+from selfdrive.car.toyota.values import Ecu, CAR, TSS2_CAR, NO_DSU_CAR, MIN_ACC_SPEED, CarControllerParams
 from selfdrive.car import STD_CARGO_KG, scale_rot_inertia, scale_tire_stiffness, gen_empty_fingerprint, get_safety_config
 from selfdrive.car.interfaces import CarInterfaceBase
 
 EventName = car.CarEvent.EventName
 
 
+
```
adeebshihadeh, line 12: remove these extra lines

adeebshihadeh:
I like this, it should add more friction to create new tunes. After merging this, we should do the same for at least Honda and Hyundai.

selfdrive/car/toyota/interface.py
```diff
@@ -23,267 +24,212 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
...
+    if candidate in [CAR.LEXUS_CTH,]:
+      ret.safetyConfigs[0].safetyParam = 100
+    elif candidate in [CAR.LEXUS_IS,]:
+      ret.safetyConfigs[0].safetyParam = 77
+    elif candidate in [CAR.COROLLA,]:
+      ret.safetyConfigs[0].safetyParam = 88
+    elif candidate in [CAR.PRIUS,]:
+      ret.safetyConfigs[0].safetyParam = 66  # see conversion factor for STEER_TORQUE_EPS in dbc file
+    else:
+      ret.safetyConfigs[0].safetyParam = 73
```
adeebshihadeh, line 232: doesn't really make sense to do this for the safetyParam. you could also just set a default of 73 on top and override for specific cars

HaraldSchafer: Done

selfdrive/car/toyota/interface.py
HaraldSchafer: Done

selfdrive/car/toyota/tunes.py
HaraldSchafer: Done

PR #22860, opened 2021-11-10
ui/alert: add blinking back

deanlee:
fill rect in NvgWindow is much more faster than in OnroadAlerts(0.05ms, full screen). the CPU usage almost will not increase while blinking

https://user-images.githubusercontent.com/27770/141195181-1ada9b05-c47d-49e4-8d57-cda5aec9602e.mp4

### DIFF
--- a/selfdrive/ui/qt/onroad.cc
+++ b/selfdrive/ui/qt/onroad.cc
@@ -10,6 +10,13 @@
 #include "selfdrive/ui/qt/maps/map.h"
 #endif
 
+int getAlertHeight(cereal::ControlsState::AlertSize size, int full_height) {
+  if (size == cereal::ControlsState::AlertSize::SMALL) return 271;
+  if (size == cereal::ControlsState::AlertSize::MID) return 420;
+  if (size == cereal::ControlsState::AlertSize::FULL) return full_height;
+  return 0;
+}
+
 OnroadWindow::OnroadWindow(QWidget *parent) : QWidget(parent) {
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
The way alerts is now split between NvgWindow and OnroadAlerts is pretty dirty. Why is it faster to draw the background in NvgWindow? Can we just optimize OnroadAlerts?

deanlee:
do blinking in onroadAlerts is very slow which is basically impossible.
It is difficult to optimize OnroadAlerts, there is no way to stack two openglwidgets together. OnroadAlerts is suitable for drawing some content that does not change much

adeebshihadeh:
I'd rather not do this, but maybe we should merge OnroadAlerts into NvgWindow. Though I'm not sure blinking is worth it.

deanlee:
OnroadAlerts is still useful, we can draw other information on it, for example, maxspeed, speed... :   https://github.com/commaai/openpilot/pull/22173 . I renamed it to OnroadHud in that pr

adeebshihadeh:
I was thinking that the HUD overlays would be their own widget. Does stacking have a big perf hit for normal QWidgets?

deanlee:
stacking multiple small qwidgets is expensive. QGraphicsScene is a better choice

deanlee:
I'll try to do blinking in OnroadAlerts after https://github.com/commaai/openpilot/pull/22843

adeebshihadeh:
We've decided to just cut alert blinking.

PR #22861, opened 2021-11-10
sidebar/updateState: use uistate.has_prime instead of read from params

deanlee:


selfdrive/ui/qt/sidebar.cc
```diff
@@ -53,7 +53,7 @@ void Sidebar::updateState(const UIState &s) {
   ItemStatus connectStatus;
   auto last_ping = deviceState.getLastAthenaPingTime();
   if (last_ping == 0) {
-    connectStatus = params.getBool("PrimeRedirected") ? ItemStatus{"NO\nPRIME", danger_color} : ItemStatus{"CONNECT\nOFFLINE", warning_color};
```
adeebshihadeh, line 56: this isn't the same. `PrimeRedirected` means that the cell plan is inactive, where as the prime status is from our backend

deanlee:
One possible problem is that if last_ping==0, params.getBool will be called every time.

PR #22862, opened 2021-11-10
Release2

bluesforte:
This is a working port for the Hyundai Tucson 2021.  It has been tested by one of my clients on multiple long drives (3+hours each), and seems to be working well.  I didn't do any lateral PID tune due to time constraints, but the default one handles pretty well. 

Some notes:
- SteerRatio was determined empirically - 16 "felt" right after a few test drives. 
- Car data (wheelbase, weight etc) was determined from the Hyundai website
- The Tucson cruise control appears not to turn on unless above ~20MPH, though does correctly stay on and allows for following cars even in stop-and-go traffic.  May be able to be overridden somewhere in openpilot, but minimizing changes and don't want to have any impact on the safety model.
- Wish I had saved more test data while I had access to the car, but wasn't thinking ahead to the pull request :(.  If I do get access again, I will grab the route data.

Enjoy!


adeebshihadeh:
We only merge PRs against the master branch. The car port will make its way to release2 in the release after your PR is merged.

Issue #22837, opened 2021-11-09
Loggerd: 100% CPU- should not use bz2 compression on camera video


jwolffe:
### Describe the bug

When simulating with Carla, I observed 100% usage of the CPU core that loggerd runs on. The roadCameraState is being handled through the logger_log routine that uses bz2 compression. The video is not compressible by that lossless method, and the high data rate (~60MB/s) causes loggerd to be overloaded.

### OS Version

Ubuntu20.04LTS

### openpilot version or commit

53e095343de56ff838f36ed9feda9f31ff6358b6

### Additional info

I tried fixing it by changing line 334 of loggerd.cc to:
     if (!it.should_log || strstr(it.name,"CameraState")) continue;

This prevents any of roadCameraState, driverCameraState, or wideRoadCameraState from being logged through the bz2 path, and seems to solve the immediate problem. The camera video is compressed to .hvec by the encoder threads. There is a question of how this issue should be fixed, whether simply applying my one-liner to loggerd.cc, or perhaps some sort of system-level solution that coordinates the should_log boolean from services.py together with the .hvec logging done by the encoder threads.

adeebshihadeh:
We'd like to keep our issue tracker focused on comma two/three, so I'm gonna close this. loggerd for PC mostly exists for allowing the tests to run on PC. We'd certainly merge PRs to improve the PC experience though!

As for the frames being the in cameraState packets, we should fix this. CARLA uses a fake camera in camerad called "frame stream" as it was easier to stuff the frames in the capnp messages than send over visionipc from python as camerad normally does. However, we now have a nice python visionipc server implementation, so we can do this properly now. I started that in #21324 but didn't have time to finish it. Feel free to take that PR and finish it up.

jwolffe:
First, my apologies, I didn't intend to misuse the "Issues". I'm still learning how things are done in this community. 

I will look into that PR #21324 and see what I can figure out. It is a little difficult for me to see how things work- I only have the PC simulation environment, and I don't have much understanding of the "big picture" yet. I've just been finding and fixing bugs in the simulation.

So, the cereal implementation supports the ...CameraState messages (and loggerd erroneously applies bz2 compression), but that was just to provide a temporary "hack" for the PC simulation. Am I correct that they should be removed from the cereal code when the PR gets finished, or is something else intended?

Thank you for the guidance.

adeebshihadeh:
No worries :)

[...]

Issue #22736, opened 2021-10-29
100% CPU on onnxruntime threads caused by spin-waits


jwolffe:
### Describe the bug

I observed onnx_runner.py threads consuming nearly 100% of 4 CPU cores on my 8 core PC when running Openpilot simulations with Carla. Basically, one thread seems to consume less than a full core, while the others are at a solid 100%. I then tried different values for "options.intra_op_num_threads" in onnxrunner.py. The following results are from the System Monitor utility, where 100% means all 8 cores are busy. 

11% with CPUExecutionProvider  (1 thread)
21% with CPUExecutionProvider  (2 threads)
35% with CPUExecutionProvider  (3 threads)
44% with CPUExecutionProvider  (4 threads)
63% with CPUExecutionProvider  (6 threads)
 3% with CUDAExecutionProvider (4 threads)

### OS Version

Ubuntu 20.04LTS

### openpilot version or commit

e178bed3016957ae09228a91f793c8d5c00d5573

### Additional info

I was not able to get a "pip install onnxruntime-gpu" configuration working with my RTX2070 CUDA/CUDNN setup. I then built onnxruntime from source to see if it would help- it gave the same results for CPU operation, but now I got the CUDA/GPU working, with dramatic reduction of CPU time (after minor mods to onnx_runner.py to select CUDA execution). After a good deal of investigation it seems that onnxruntime CPU operation defaults to spin-waits on the input queues, which fully accounts for the measured results. Great for benchmarking, but not all that practical in this case. Given the low execution rate of the supercombo model, spin waits are major overkill; the operating system's scheduling should be entirely adequate.

Unfortunately, the latest onnxruntime APIs only support changing the allow_spinning option from C++; it isn't supported in the Python API. While the CUDAExecutionProvider solves my immediate simulation problems, it should be possible to rewrite the Python code for C++ disabling of the spinning. 

There are probably broader issues than just my Carla simulation on a PC- the Comma2 and Comma3 (and other) devices may suffer from the spinning, resulting in higher power dissipation than necessary, along with loss of available processing capability. At the very least, it would be a good idea to allow tuning the "options.intra_op_num_threads" value from an environment variable, to better match hardware capability with processing load.

pd0wm:
[...]

Issue #22705, opened 2021-10-27
sim: current issues preventing PC simulation with Carla, and some fixes


jwolffe:
### Describe the bug

Over the last month or so, the "out of the box" docker-based Carla simulations have not functioned correctly. I have created some temporary native (non-docker) fixes for several issues that allow the current master code to "drive" on my Ubuntu PC as a non-root user. Hopefully this information will prove useful to other developers.

### OS Version

Ubuntu 20.04LTS

### openpilot version or commit

e178bed3016957ae09228a91f793c8d5c00d5573

### Additional info

Issues and Fixes:
**1) Undo PR #22546 to allow correct camera operation (EON vs TICI switching of values)- temporary only.** The PR is assumed to fit together with a planned future implementation.

**2) CS.canValid problem- fix in controlsd.py:**
    ....
    self.button_timers = {ButtonEvent.Type.decelCruise: 0, ButtonEvent.Type.accelCruise: 0}
    self.canValid = False  #add to intialization
    ...
    self.canValid = CS.canValid or SIMULATION #add after reading message
    all_valid = self.canValid and self.sm.all_alive_and_valid()
    ...
    [replace all other instances of CS.canValid with self.canValid]

**3) Fix .comma path to prevent generating path errors**
sudo mkdir /.comma
sudo ln -s $HOME/.comma/* /.comma/

**4) Add exports to ~/.bashrc:**
export USE_FRAME_STREAM=1
export SIMULATION="1"
export FINGERPRINT="HONDA CIVIC 2016" (related to PR #22679)

**5) Modify util.cc  --- allow non-root simulation, return 0 for priority and affinity setters**
int set_realtime_priority(int level) {
#ifdef __linux__
  if ((getenv("SIMULATION") != nullptr) && (std::string(getenv("SIMULATION")) == "1")) {
    return 0;
  }
  else {
  .... [ linux priority code ]
  }
#else  

  int set_core_affinity(std::vector<int> cores) {
#ifdef __linux__
  if ((getenv("SIMULATION") != nullptr) && (std::string(getenv("SIMULATION")) == "1")) {
    return 0;
  }
  else {
  .... [ linux affinity code ]
  )
#else  
**6) Add the PythonAPI egg for Carla into the PYTHONPATH, allowing bridge.py to run**

jwolffe:
[...]

PR #22875, opened 2021-11-11
PC/Carla simulation fixes

jwolffe:
This PR addresses multiple issues affecting simulation on a PC
1) prevent excessive CPU usage by loggerd (#22837)
2) fix PC intrinsics selection (#22705,#22546)
3) add missing "HUD_SETTING" message in can.py (#22705)
4) don't alert on cruise_mismatch when simulating (#22780)
5) add onnx CUDAExecutionProvider (#22736)
6) fix PC path setup (#22705)

An additional change is needed to get good simulation. Adding the missing can message revealed a problem in the simulated car's database, which lacks COUNTER and CHECKSUM values for "HUD_SETTING" messages in the opendbc/generator/honda/honda_civic_touring_2016_can.dbc file. The following appears to work:

BO_ 493 HUD_SETTING: 8 XXX
 SG_ IMPERIAL_UNIT : 5|1@0+ (1,0) [0|1] "" EON
 SG_ COUNTER : 61|2@0+ (1,0) [0|3] "" EON
 SG_ CHECKSUM : 59|4@0+ (1,0) [0|3] "" EON

### DIFF
--- a/selfdrive/common/modeldata.h
+++ b/selfdrive/common/modeldata.h
@@ -41,7 +41,7 @@ const auto X_IDXS_FLOAT = convert_array_to_type<double, float, TRAJECTORY_SIZE>(
 #include "selfdrive/common/mat.h"
 #include "selfdrive/hardware/hw.h"
 const mat3 fcam_intrinsic_matrix =
-    Hardware::EON() ? (mat3){{910., 0., 1164.0 / 2,
+    (Hardware::EON() || Hardware::PC()) ? (mat3){{910., 0., 1164.0 / 2,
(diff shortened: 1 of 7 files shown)


jwolffe:
Something "smells wrong" with the malformed HUD_SETTING message. How has this not been a problem with actual operation in cars? I'm wondering if the correct solution might be to remove it rather than fix it with the added counter and checksum.

pd0wm:
Looks like we never defined the counter/checksum in the HUD_SETTING message. Fixed this on the bridge side for now.

PR #22877, opened 2021-11-12
nav: add arrow images for RHD instructions

eFiniLan:
Navigation instruction icon on RHD is not handled:
DSC_0711

This will show the right icon instead:
DSC_0712


selfdrive/ui/qt/maps/map.cc
```diff
@@ -319,6 +320,7 @@ void MapWindow::offroadTransition(bool offroad) {
 }
 
 MapInstructions::MapInstructions(QWidget * parent) : QWidget(parent) {
+  is_rhd = Params().getBool("IsRHD");
```
adeebshihadeh, line 323: can we get this from mapbox instead?

eFiniLan: Hi @adeebshihadeh,
thanks! according to the [direction api](https://docs.mapbox.com/api/navigation/directions/):

`driving_side` column is `The legal driving side at the location for this step. Either left or right.`

however, it is under `routes[0] > legs[0] > steps[0] > driving_side`
 
image

any idea where I will have access to that? looks like `cur_maneuver.extendedAttributes()` does not have it.

pd0wm: The parser from the qtlocation plugin is not very complete. It's on my own list to write my own parser so we can also get speed limits properly.

pd0wm:
Did you make the icons yourself? I got the other ones from https://github.com/mapbox/directions-icons/tree/master/src/png/light and didn't include rhd versions.

selfdrive/ui/qt/maps/map.cc
pd0wm: So we can use the RHD toggle/param for now and switch to this output once we have our own parser.

eFiniLan:
> Did you make the icons yourself? I got the other ones from https://github.com/mapbox/directions-icons/tree/master/src/png/light and didn't include rhd versions.

yeah... I just flip them : )

selfdrive/ui/qt/maps/map.cc
```diff
@@ -470,8 +478,15 @@ void MapInstructions::updateInstructions(cereal::NavInstruction::Reader instruct
     } else if (straight) {
       fn += "turn_straight";
     }
+    fn +=  + ".png";
```
pd0wm, line 481: I assume the lane arrows don't need a RHD version?

eFiniLan: this section do not append modifier, I've removed them :+1:

pd0wm:
What do you think? No need for string replacements this way.

pd0wm:
2021-11-12-141437_1122x760_scrot

What program did you use to flip the images? They are larger in size, but lower in quality.

eFiniLan:
> 2021-11-12-141437_1122x760_scrot
> 
> What program did you use to flip the images? They are larger in size, but lower in quality.
> 
> Maybe it's better to let Qt deal with the flipping.

This was done with gimp, I just tried macos "Preview" and photoshop, photoshop version looks better and smaller.

flipping could work too, but then we will need to specify *roundabout*/*rotary* and *uturn* types so I guess using images is cleaner.

pd0wm:
Still seeing some more aliasing on the flipped ones.

We don't need to detect roundabout/rotary, right? We can just swap left/right for every instruction and then flip the image. For a lot instruction this won't do anything, but that doesn't really matter.

eFiniLan:
> Still seeing some more aliasing on the flipped ones.
> 
> We don't need to detect roundabout/rotary, right? We can just swap left/right for every instruction and then flip the image. For a lot instruction this won't do anything, but that doesn't really matter.

ahh that will do, I'll give it a shot :)

pd0wm:
Much more elegant this way!

PR #22881, opened 2021-11-12
Fingerprint for Japanese 2017 Nissan X-Trail

avolmensky:
Adding an additional fingerprint for a 2017 Nissan X-Trail (Rogue)


pd0wm:
What's the dongle id/route?

PR #22882, opened 2021-11-12
Move shutdown/reboot handling to manager

### LINKED ISSUE #22821: Graceful reboot/shutdown while onroad

pd0wm:
Closes #22821

We should probably add something to boardd to put the panda back into elm/no output mode when it's shut down cleanly. Think that should go in a separate PR.

2021-11-12-151102_2140x1072_scrot
2021-11-12-151047_2148x1072_scrot

### DIFF
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -168,7 +168,7 @@ DevicePanel::DevicePanel(QWidget* parent) : ListWidget(parent) {
   power_layout->addWidget(reboot_btn);
   QObject::connect(reboot_btn, &QPushButton::clicked, [=]() {
     if (ConfirmationDialog::confirm("Are you sure you want to reboot?", this)) {
-      Hardware::reboot();
+      Params().putBool("DoReboot", true);
     }
   });
 
@@ -177,7 +177,7 @@ DevicePanel::DevicePanel(QWidget* parent) : ListWidget(parent) {
   power_layout->addWidget(poweroff_btn);
   QObject::connect(poweroff_btn, &QPushButton::clicked, [=]() {
     if (ConfirmationDialog::confirm("Are you sure you want to power off?", this)) {
-      Hardware::poweroff();
+      Params().putBool("DoShutdown", true);
     }
   });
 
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
I think we want to keep it short. What about "Disengage to Reboot"?

A side note: we should have a timeout on the settings when onroad. All the dialogs already handle this, so it should be as easy as adding a QTimer like we do for turning off the display.

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -167,17 +167,31 @@ DevicePanel::DevicePanel(QWidget* parent) : ListWidget(parent) {
...
-      Hardware::reboot();
+    if (QUIState::ui_state.status == UIStatus::STATUS_DISENGAGED) {
+      if (ConfirmationDialog::confirm("Are you sure you want to reboot?", this)) {
+        // Check engaged again in case it changed while the dialog was open
+        if (QUIState::ui_state.status == UIStatus::STATUS_DISENGAGED) {
+          Params().putBool("DoReboot", true);
+        }
+      }
+    } else {
+      ConfirmationDialog::alert("Reboot not Possible While Engaged", this);
```
adeebshihadeh, line 178:
```suggestion
      ConfirmationDialog::alert("Disengage to Reboot", this);
```

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -167,17 +167,31 @@ DevicePanel::DevicePanel(QWidget* parent) : ListWidget(parent) {
...
-      Hardware::poweroff();
+    if (QUIState::ui_state.status == UIStatus::STATUS_DISENGAGED) {
+      if (ConfirmationDialog::confirm("Are you sure you want to power off?", this)) {
+        // Check engaged again in case it changed while the dialog was open
+        if (QUIState::ui_state.status == UIStatus::STATUS_DISENGAGED) {
+          Params().putBool("DoShutdown", true);
+        }
+      }
+    } else {
+      ConfirmationDialog::alert("Power Off not Possible While Engaged", this);
```
adeebshihadeh, line 194:
```suggestion
      ConfirmationDialog::alert("Disengage to Power Off", this);
```

PR #22889, opened 2021-11-12
Toyota long retunes

haraschax:

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -10,21 +10,6 @@ from opendbc.can.packer import CANPacker
 VisualAlert = car.CarControl.HUDControl.VisualAlert
 
 
-def accel_hysteresis(accel, accel_steady, enabled):
-
-  # for small accel oscillations within ACCEL_HYST_GAP, don't change the accel command
-  if not enabled:
-    # send 0 when disabled, otherwise acc faults
-    accel_steady = 0.
-  elif accel > accel_steady + CarControllerParams.ACCEL_HYST_GAP:
-    accel_steady = accel - CarControllerParams.ACCEL_HYST_GAP
-  elif accel < accel_steady - CarControllerParams.ACCEL_HYST_GAP:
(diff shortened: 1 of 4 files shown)


selfdrive/car/toyota/carcontroller.py
```diff
@@ -5,63 +5,38 @@
...
-    pcm_accel_cmd = clip(pcm_accel_cmd, CarControllerParams.ACCEL_MIN, CarControllerParams.ACCEL_MAX)
+    if CS.CP.enableGasInterceptor and enabled:
+      MAX_INTERCEPTOR_GAS = 0.5
+      # RAV4 has very sensitive has pedal
+      if CS.CP.carFingerprint in [CAR.RAV4, CAR.RAV4H]:
+        PEDAL_SCALE = interp(CS.out.vEgo, [0.0, MIN_ACC_SPEED, MIN_ACC_SPEED + PEDAL_TRANSITION], [0.1, 0.3, 0.0])
+      else:
+        PEDAL_SCALE = interp(CS.out.vEgo, [0.0, MIN_ACC_SPEED, MIN_ACC_SPEED + PEDAL_TRANSITION], [0.4, 0.5, 0.0])
+      #pitch_accel = -sin(pitch) * 9.81
+      interceptor_gas_cmd = clip(PEDAL_SCALE * actuators.accel, 0., MAX_INTERCEPTOR_GAS)
```
sshane, line 36: Some offsetting might be a good idea too. Now it might not cause any jerking since it's now feedforward, but 0 acceleration means we want to apply *some* gas, so there still will be some highway oscillation.

HaraldSchafer: I was hoping to not use the pedal above the minimum ACC speed

sshane: This issue would also apply at low speeds as well I'm thinking, just maybe not as much. If you try stay at a speed it will always be undershooting without some gas offset based on speed.

Also this seems to be applying some gas, albeit small, above the min ACC speed, so some strange things might happen. Like if going downhill without gravity compensation, gas pedal will keep the car from braking if coasting accel is positive.

PR #22892, opened 2021-11-12
watchdog_kick: non-allocating

deanlee:


pd0wm:
Should we just write the 4 bytes of the uint64 timestamp in the watchdog file and `struct.unpack` it on the python side?

deanlee:
~25% of the QUIState::update()'s time is spent on  watchdog_kick when onroad. think it's better not to  waste any cpu on unnecessary things even just a tiny bit.

pd0wm:
We also don't need to kick the watchdog on every `update()` call. Maybe that got changed in one of the refactors? The timeout is 5 seconds, so it's fine to only call `kick` every second.

pd0wm:
Good to merge?

deanlee:
yes

PR #22901, opened 2021-11-14
ui: draw all radard leads

sshane:
This draws any lead that the long MPC would stop for, specifically leads detected by the radar at low speed where the model doesn't agree.

Like in this example, radarState and modelV2 disagree if there's a lead (happens to be a speed bump the radar is picking up in low speed override mode). It could be confusing why the car is stopping or not accelerating when there's no indication of a lead on the screen. Just a temp fix. Real fix: remove radard

(`f15e3c37c118e841|2021-11-13--17-47-08--34`)
Screenshot from 2021-11-13 19-29-48


HaraldSchafer:
I was hoping to remove radard or at least radar override, but both seemed to be useful according to latest tests, so those are staying for a while. Let's just always draw the radard leads again?

PR #22905, opened 2021-11-14
ui: use YUV vipc stream

deanlee:


deanlee:
There is a problem with the output, may be caused by padding problem
yuv

deanlee:
watch3 looks good

adeebshihadeh:
Is this ready for review? If not, what's missing?

deanlee:
The only problem is c2, do we need to use rgb on c2 and yuv on other devices?

pd0wm:
What do you mean by C2 is optimized for RGB?

We want all devices to run the same code.

deanlee:
how about EGLImageTexture? it's used for avoid rgb memory copy on C2, should we remove it?

pd0wm:
You can remove it in this PR. We have to verify using the snapdragon profiler that the end result is the same or better.

deanlee:
let's keep it until find a general way to avoid copy

deanlee:
Now there is an additional VISIONSTREAM RGB MAP, Can it be changed to yuv format？

pd0wm:
Yes, you can convert it to yuv using something like libYUV. It's only 256x256, and prefer to do it on the CPU.

pd0wm:
Not sure if it makes sense to convert that to YUV. Is it a problem to keep it as RGB for now?

deanlee:
it can't be displayed after this Pr.

pd0wm:
That's ok. I'll think of a way to display it again after merging this.

deanlee:
The cpu usage on c2 is the same as before after use YUV  stream (15%).  I think it's ready for review now.

selfdrive/ui/qt/offroad/driverview.cc
```diff
@@ -12,7 +12,7 @@ DriverViewWindow::DriverViewWindow(QWidget* parent) : QWidget(parent) {
   layout = new QStackedLayout(this);
   layout->setStackingMode(QStackedLayout::StackAll);
 
-  cameraView = new CameraViewWidget("camerad", VISION_STREAM_RGB_FRONT, true, this);
+  cameraView = new CameraViewWidget("camerad", VISION_STREAM_YUV_FRONT, true, this);
```
adeebshihadeh, line 15: wanna make a PR to rename the YUV streams? something like VISION_STREAM_YUV_FRONT -> VISION_STREAM_ROAD and leave RGB as is

jwooning:
I did some profiling on this PR and it seems there is no real performance or memory benefit yet. 

Can you also look into the openCL kernels in camerad to use the yuv buffers, this should then allow to remove some of the used rgb buffers.

pd0wm:
To save on memory bandwidth we need to skip the intermediate step into an RGB buffer, and make the debayer step go straight to a YUV output buffer. Feel free to only do the C2 debayer kernel first, so we can profile and see it there are any actual benefits.

PR #22906, opened 2021-11-14
framereader: support hardware decoding

deanlee:
currently supports CUDA, ~MediaCodec and VideoToolbox~


pd0wm:
Can we detect which ones are available and automatically enable them?

deanlee:
It may take at least 150 lines of code to accurately detect the hardware type （https://ffmpeg.org/doxygen/trunk/hwdevice_8c_source.html）.it might be better to implement it in another Pr later if necessary. what do you think?

pd0wm:
Can you add the flag to override the choice (and force no hw decoding)?

pd0wm:
Is it worth it to support videtoolbox and mediacodec? From my quick testing they don't seem to support hevc. Or is that a problem with my setup?

pd0wm:
Noticed ui.py doesn't work when using `cuda` as a hw decoder (can't match the buffer size to a camera).

deanlee:
Then we can only support CUDA for the time being.
auto detection is complicated.  I suggest still using flag for hw-decode, what do you think?

deanlee:
> Noticed ui.py doesn't work when using `cuda` as a hw decoder (can't match the buffer size to a camera).

Hardware decoder requires 16-pixels aligned height. I'll fix this issue

deanlee:
> Noticed ui.py doesn't work when using `cuda` as a hw decoder (can't match the buffer size to a camera).

fixed

pd0wm:
Just supporting cuda is fine. Think it can be default on, but disabled using `--no-cuda` or something like that.

PR #21992, opened 2021-08-20
fingerprint for toyota highlander hybrid 2020

evdoks:
**Description:** Fingerprint for Toyota Highlander hybrid 2020

**Verification:** Checked in Toyota Highlander hybrid 2020 (EU model)

PR #22911, opened 2021-11-15
Fixed someone else's missing 000 to x00 conversion

ErichMoraga:


ErichMoraga:
https://github.com/commaai/openpilot/pull/21992 was where it originated.

PR #22923, opened 2021-11-16
Add cost to long plan change for smoother lag comp

haraschax:


pd0wm:
Jonathan suggested constraining the plan up to the lag compensation amount with what we computed previous iteration. Have you experimented with that already?

HaraldSchafer:
That's what this PR does?

pd0wm:
This does a cost for the whole horizon right? I meant only up to `carParams.longitudinalActuatorDelayUpperBound` and no cost after.

HaraldSchafer:
I'm trying up to 1second now. Don't really like the idea of it being car dependent. Then all the tests and cost tuning is then car dependent.

sshane:
Just tested a few times since there's not a lot of traffic, but coming to a stop seems to be much smoother. Used to over brake and then slowly creep up. Could be placebo but seems better so far. Excited to try it in traffic

HaraldSchafer:
@sshane Especially for large lags it should make a big difference :)

PR #22924, opened 2021-11-16
Fix missing pip command issue.

eFiniLan:
Screen Shot 2021-11-16 at 1 58 01 PM

### DIFF
--- a/tools/mac_setup.sh
+++ b/tools/mac_setup.sh
@@ -61,6 +61,7 @@ fi
 
 # install python
 PYENV_PYTHON_VERSION=$(cat $OP_ROOT/.python-version)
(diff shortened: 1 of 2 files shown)


pd0wm:
Should this also be to openpilot_env.sh or zshrc.sh?

eFiniLan:
> Should this also be to openpilot_env.sh or zshrc.sh?

Thanks @pd0wm, I've adjusted openpilot_env.sh for that, I've got that far with my macbook m1, still trying to figure out why it failed when install backports.lzma==0.0.14.

pd0wm:
Let me see if that package can be removed. Seems like some leftover from the python 2 era.

PR #22927, opened 2021-11-16
ui: update params on the show event

deanlee:

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -186,6 +186,10 @@ static void update_state(UIState *s) {
   scene.started = sm["deviceState"].getDeviceState().getStarted() && scene.ignition;
 }
 
+void ui_update_params(UIState *s) {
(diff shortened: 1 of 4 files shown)


selfdrive/ui/ui.cc
```diff
@@ -211,10 +203,12 @@ static void update_status(UIState *s) {
   static bool started_prev = false;
   if (s->scene.started != started_prev) {
     if (s->scene.started) {
+      Params params;
       s->status = STATUS_DISENGAGED;
       s->scene.started_frame = s->sm->frame;
-      s->scene.end_to_end = Params().getBool("EndToEndToggle");
-      s->wide_camera = Hardware::TICI() ? Params().getBool("EnableWideCamera") : false;
+      s->scene.is_metric = params.getBool("IsMetric");
```
adeebshihadeh, line 209: this is supposed to update live. maybe update it on the show event?

PR #22928, opened 2021-11-16
ui/alert: return std::optional<alert>

deanlee:
update bgColor before `alerts->updateAlert` to fix the wrong bg of controlsUnresponsive:

Screenshot from 2021-11-16 22-03-07

@adeebshihadeh  I think it may makes more sense to return an std::optional now.


adeebshihadeh:
Fixed the background in 0e35db7c05172b154c70216d2b1b9ed8c9c7b163

PR #22933, opened 2021-11-16
panda/can_send: remove pos++

deanlee:
I don’t understand why pos++ is here,  it might be a mistake, right?


briskspirit:
We add counter(1 byte) to send vector. Pos represents byte count(length) of send vector that we pass to usb_bulk_write later.

deanlee:
https://github.com/commaai/openpilot/blob/70c1f19304cf6c81baddee2f67d64e48b7891d3f/selfdrive/boardd/panda.cc#L398-L406
@briskspirit should we call send.clear() after usb_bulk_write? otherwise in the next cycle, the old data will be sent again. (The counter will also be re-inserted into the old data, which should also be a problem).  Is my understanding correct ?

PR #22935, opened 2021-11-16
Add Linux Mint 20.2 support

JMPZ11:
**Description** [] Added Linux Mint 20.2 to list of supported versions in ubuntu_setup.sh. Linux Mint 20.2 is based on Ubuntu 20.04 LTS. Prior to the addition of the case statement, the script worked without error

**Verification** [] Ran script in clean Linux Mint 20.2 install, launched pipenv shell, ran pre-commit and scons; built and flashed panda / pedal. OP UI ran prior to the addition of these conditions.


adeebshihadeh:
Added a generic override in 0674726cf4e0dfffc30d4596e2c650821b792588

PR #22940, opened 2021-11-17
Mazda: limit hud warnings until we find a way to silence them

Jafaral:
The hud warning is  bit annoying with a chime. Let us give ourselves some time to sort this out before we enable the hud warning for several  events. This PR limits the warning to steer lockout situations.


adeebshihadeh:
I'd like to keep this as is. Perhaps you're getting excessive alerts for some other reason? Are there particular alerts you're seeing often?

Jafaral:
@adeebshihadeh  we still have an unsolved situation where we throw warnings at lkas speed allowed transition , I don't know and that certainly needs fixing.  There are also steer limit warnings. 

I'm OK leaving it for now. We can revisit later.

adeebshihadeh:
Let's leave this for now. Hopefully we can figure out an alert without sound soon. Not sure if I mentioned this already, but we only really care about the dash alert, showing up on the HUD is just an extra.

PR #22941, opened 2021-11-17
Add Lexus RC support

ErichMoraga:
`@ph33rl355#529`  2020 Lexus RC F Track Edition  DongleID/route 32696cea52831b02|2021-11-16--23-12-02

Doesn't seem to need a full port, per user feedback...
https://discord.com/channels/469524606043160576/524327905937850394/910400788319326229


pd0wm:
Any specific reason to move this under the IS? Is it based off the same car?

ErichMoraga:
> Any specific reason to move this under the IS? Is it based off the same car?
> 
> The FW versions seems to have a distinct prefix of `24` vs `53` for the IS

The RC was moved under the IS mostly out of simplicity, as the cars are extremely similar... part of the Lexus sports car line-up.  Furthermore, the RC seems to drive fine, with only some slight ping pong... typical for a pre-LSS2 car.  Lastly, this will possibly be the only RC running OP.

pd0wm:
Let's split it out in a new car, but you reuse the IS dbc.

ErichMoraga:
> Let's split it out in a new car, but you can reuse the IS dbc.

DONE.

ErichMoraga (commented):
Minimum speed might actually be closer to 25, but the user doesn't seem sure...
https://discord.com/channels/469524606043160576/524327905937850394/910963149329563778

pd0wm:
Do you have a test route?

ErichMoraga:
> 

32696cea52831b02|2021-11-19--09-07-37 is his latest drive (he's been driving a few days with the f/w hacked into `values.py`).  No idea about the alternate cancel message.

ErichMoraga:
Thanks for your help @pd0wm (I have a lot of distractions).  I've manually loaded the user's C3 w/ all the changes via wget (DL'd raw files onto his device), and manually compiled.  Rebooted his C3, so next time he drives, that route will have changes identical to what's in this PR ;-)

ErichMoraga:
> Do you have a test route?

Added test route... `32696cea52831b02|2021-11-19--18-13-30`

adeebshihadeh:
The route doesn't have rlogs uploaded. I just requested them, but the user should keep their device on wifi until it's fully uploaded.

ErichMoraga:
> The route doesn't have rlogs uploaded. I just requested them, but the user should keep their device on wifi until it's fully uploaded.

All 37 rlogs for that route just finished uploading.

adeebshihadeh:
This looks very similar to the quirky IS. Also looks like the IS cancel works in the test route.

PR #22946, opened 2021-11-17
panda: refactor can_send

deanlee:


gregjhogan:
The counter is supposed to be every 64 bytes starting with byte zero (not before each message).  This makes the first byte of every USB packet a counter.

https://github.com/commaai/panda/blob/7393783bc856fb6c700e3d6cbde04a1ac0195783/USB_packet_structure.png

It looks like we could use some unit tests for the send and receive since nothing failed!

deanlee:
my bad. I'll add unit tests for it.

briskspirit:
@deanlee I will add more comments to code today so it will be easier to follow. In the mean time it would be awesome if you can add unit tests first!

deanlee:
@briskspirit I'm writing tests cases in https://github.com/commaai/openpilot/pull/22955. please feel free to let me know if I am doing anything wrong

PR #22948, opened 2021-11-17
sidebar: dont update states when hidden

deanlee:


pd0wm:
Would this not show the old state briefly when opening the sidebar?

deanlee:
i think it's ok. it's updated at 20hz when visible.

PR #22949, opened 2021-11-17
cameraview: remove glflush

deanlee:


pd0wm:
You can revert https://github.com/commaai/openpilot/commit/7a3754d50b2325863259cb31b8da6c5e31990246  if the cpu usage is back to 20%

pd0wm:
Do you still want to look into the blocking of glMap, or is this the best we can do?

deanlee:
glmap usually blocking for a few milliseconds, and occasionally more than 10ms, it's ok because it's run in vipc thread, the main  thread will not be blocked. 
I can't think of other ways for the time being to improve the performance of async pbo.

pd0wm:
Oh, assumed it was much more than a few ms. Then this is fine.

pd0wm:
Weird, I really though CPU usage was lower when I tested this change before. Guess I didn't rebuild properly or something. There must be something else we can do to lower CPU.

PR #22950, opened 2021-11-17
swaglog.py: respect LOGPRINT env variable

pd0wm:
match swaglog.cc


pd0wm:
@adeebshihadeh if you agree with this change I'll fix up the uploader tests to not depend on our logging.

adeebshihadeh:
I like that it matches swaglog.cc, but those info to warning changes don’t seem correct?

pd0wm:
I'll revert those changes. Most of them don't really need to be printed anyway.

PR #22959, opened 2021-11-17
locationd: Fix Nav localization reliability

nuwandavek:


HaraldSchafer:
This PR contains a lot of duplicate code. 

Stick to using one initial P matrix and then just set dealing with it in the reset function.

Make the new reset function the only reset function and overload it to use less arguments like is already done, or use optional parameters.

selfdrive/locationd/locationd.cc
```diff
@@ -243,9 +243,24 @@ void Localizer::handle_sensors(double current_time, const capnp::List<cereal::Se
...
+  // This is done to make sure that the error estimate of the position does not blow up
+  // when the filter is in no-gps mode or in a tunnel
+  VectorXd current_x = this->kf->get_x();
+  
+  VectorXd ecef_pos = current_x.segment<3>(0);
+  VectorXd ecef_vel = current_x.segment<3>(7);
+  MatrixXdr ecef_pos_R = this->kf->get_fake_gps_pos_cov();
+  MatrixXdr ecef_vel_R = this->kf->get_fake_gps_vel_cov();
+  
+  this->kf->predict_and_observe(current_time, OBSERVATION_ECEF_POS, { ecef_pos }, { ecef_pos_R });
```
HaraldSchafer, line 256: This isn't ideal, you want to predict, then get the values then observe those. This pressures the position and velocity to stay the same, not move as predicted by the other sensors.

nuwandavek: If the observation noise is high enough, this is not true right? The more certain observations (from other sensors) have a higher say in the estimate than the low certainty sensor (as seen in the experiments).

HaraldSchafer: The effect may be minimal, but it's still more correct to get the predicted state.

You can just expose this function to do that.
https://github.com/commaai/rednose/blob/5b526a8e00bdc1c3922be470af1602cf9dc72dde/rednose/helpers/ekf_sym.cc#L197

nuwandavek: Yo, added this.

PR #22967, opened 2021-11-18
VW MQB: Volkswagen T-Roc Mk1

jyoung8607:
**Volkswagen T-Roc Mk1**

Covers all worldwide versions of the T-Roc sub-compact crossover SUV, including the T-Roc R and T-Roc Cabriolet. Averaged some mild wheelbase and mass variance between the versions for simplicity.

- [x] added to CARS
- [x] added to RELEASES
- [x] test route added to [test_models.py](https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_models.py)
- [x] route with openpilot: `0cd0b7f7e31a3853|2021-11-18--00-38-32`

Thanks AGAIN to community member Saber422!


jyoung8607:
The test route is 12 segments long, but the segment 1 rlogs are MIA, athena can't see them at all. If we can make use of this test route, great, if not, we can go back to the user.

Saber422:
All logs are completely uploaded. (0cd0b7f7e31a3853|2021-11-18--00-38-32)
Thank you!

PR #22976, opened 2021-11-19
Hyundai: remove longitudinal whitelist

adeebshihadeh:


pd0wm:
After we added the check https://github.com/commaai/openpilot/pull/22976/files#diff-601a5bb0df30ad46a5907d06a35bb7bb694fc195a49935692c97ff341d12f257R253 here it's indeed a bit unnecessary.

PR #22977, opened 2021-11-19
panda: refactor can_recv

deanlee:
unpack all data to the cache(`recv_buf)` at once


selfdrive/boardd/panda.cc
```diff
@@ -425,64 +425,41 @@ void Panda::can_send(capnp::List<cereal::CanData>::Reader can_data_list) {
...
-
-  if (recv == RECV_SIZE) {
-    LOGW("Receive buffer full");
-  }
-
   if (!comms_healthy) {
     return false;
   }
-  return unpack_can_buffer(data, recv, out_vec);
+  if (recv == RECV_SIZE) { LOGW("Panda receive buffer full"); }
```
adeebshihadeh, line 431: use multiple lines here

selfdrive/boardd/panda.cc
```diff
@@ -425,64 +425,41 @@ void Panda::can_send(capnp::List<cereal::CanData>::Reader can_data_list) {
...
+    canData.busTime = 0;
+    canData.address = header.addr;
+    canData.src = header.bus + bus_offset;
+    if (header.rejected) { canData.src += CANPACKET_REJECTED; }
+    if (header.returned) { canData.src += CANPACKET_RETURNED; }
+
+    const uint8_t data_len = dlc_to_len[header.data_len_code];
+    canData.dat.assign((char *)&recv_buf[pos + CANPACKET_HEAD_SIZE], data_len);
+
+    pos += CANPACKET_HEAD_SIZE + data_len;
```
adeebshihadeh, line 462: move this into the for?

briskspirit: ?

briskspirit: No difference but don't need to declare pos on extra line. Think that's why he chose it?

PR #22980, opened 2021-11-19
 UI: add timeout to close settings window while onroad

### LINKED ISSUE #22978: UI: add timeout to close settings window while onroad
We can probably reuse the display timeout timer used while offroad for this.

deanlee:
resolved https://github.com/commaai/openpilot/issues/22978

1. close settings window and set display off if user does not tap the screen within 30s while off-road.
2. close settings window and return to onroad window if user does not tap the screen within 10s while on-road


pd0wm:
The sidebar should not close after the timeout, only if the settings was open.

PR #22981, opened 2021-11-19
panda: refactor can_send

deanlee:

### DIFF
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -363,8 +363,9 @@ uint8_t Panda::len_to_dlc(uint8_t len) {
   }
 }
 
-void Panda::can_send(capnp::List<cereal::CanData>::Reader can_data_list) {
-  PacketWriter packet_writer(this);
+void Panda::pack_can_buffer(const capnp::List<cereal::CanData>::Reader &can_data_list,
+                         std::function<void(uint8_t *, size_t)> write_func) {
+  PacketWriter packet_writer(write_func);
   for (auto cmsg : can_data_list) {
     // check if the message is intended for this panda
     uint8_t bus = cmsg.getSrc();
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -363,56 +363,25 @@ uint8_t Panda::len_to_dlc(uint8_t len) {
   }
 }
 
-void Panda::pack_can_buffer(const capnp::List<cereal::CanData>::Reader &can_data_list,
-                         std::function<void(uint8_t *, size_t)> write_func) {
-  if (send.size() < (can_data_list.size() * CANPACKET_MAX_SIZE)) {
-    send.resize(can_data_list.size() * CANPACKET_MAX_SIZE);
-  }
-
-  int msg_count = 0;
-  while (msg_count < can_data_list.size()) {
-    uint32_t pos = 0;
-    while (pos < USB_TX_SOFT_LIMIT) {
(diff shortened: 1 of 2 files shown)


briskspirit:
Not sure that this refactoring is needed or if it makes code more readable and clear. Now it is even more hard to understand pack_buffer function imo.

gregjhogan:
I think there is an opportunity to decrease the complexity of `Panda::pack_can_buffer` while increasing readability (even if this isn't the best approach)

briskspirit:
I am not sure that readability was increased.

deanlee:
There should be a better way, I will try to figure it out.

deanlee:
how about this one?

selfdrive/boardd/panda.cc
```diff
@@ -363,57 +363,47 @@ uint8_t Panda::len_to_dlc(uint8_t len) {
...
 }
 
-void Panda::pack_can_buffer(const capnp::List<cereal::CanData>::Reader &can_data_list,
-                         std::function<void(uint8_t *, size_t)> write_func) {
-  if (send.size() < (can_data_list.size() * CANPACKET_MAX_SIZE)) {
-    send.resize(can_data_list.size() * CANPACKET_MAX_SIZE);
+static void write_packet(uint8_t *dest, int *write_pos, const uint8_t *src, size_t size) {
+  for (int i = 0, &pos = *write_pos; i < size; ++i, ++pos) {
+    if (pos % USBPACKET_MAX_SIZE == 0) {
+      dest[pos++] = pos / USBPACKET_MAX_SIZE;
```
briskspirit, line 369:
```suggestion
      dest[pos] = pos / USBPACKET_MAX_SIZE;
      pos++;
```
This is MISRA violation. We are not using MISRA for OP, but I want code related to panda be as close as possible to MISRA rules.

selfdrive/boardd/panda.cc
```diff
@@ -363,57 +363,48 @@ uint8_t Panda::len_to_dlc(uint8_t len) {
...
+
+    write_packet(send_buf, &pos, (uint8_t *)&header, sizeof(can_header));
+    write_packet(send_buf, &pos, (uint8_t *)can_data.begin(), can_data.size());
+    if (pos >= USB_TX_SOFT_LIMIT) {
+      write_func(send_buf, pos);
+      pos = 0;
     }
   }
+
+  // flush the remain packets
```
briskspirit, line 406:
```suggestion
  // send remaining packets
```

selfdrive/boardd/panda.cc
```diff
@@ -363,57 +363,48 @@ uint8_t Panda::len_to_dlc(uint8_t len) {
   }
 }
 
-void Panda::pack_can_buffer(const capnp::List<cereal::CanData>::Reader &can_data_list,
-                         std::function<void(uint8_t *, size_t)> write_func) {
-  if (send.size() < (can_data_list.size() * CANPACKET_MAX_SIZE)) {
-    send.resize(can_data_list.size() * CANPACKET_MAX_SIZE);
+static void write_packet(uint8_t *dest, int *write_pos, const uint8_t *src, size_t size) {
+  for (int i = 0, &pos = *write_pos; i < size; ++i, ++pos) {
+    if (pos % USBPACKET_MAX_SIZE == 0) {
```
briskspirit, line 368:
```suggestion
    // Insert counter every 64 bytes (first byte of 64 bytes USB packet)
    if (pos % USBPACKET_MAX_SIZE == 0) {
```

PR #22986, opened 2021-11-20
longitudinal: only apply overshoot prevention when braking

sshane:
The previous code activated any time you're under 1.5 m/s and the v_target_future is less than 0.7 m/s, so essentially no acceleration would be commanded until after the future v_target got above a certain threshold, which instead we only want to apply this when braking. Adding an and so it only applies when requesting brake means we start accelerating ~0.8 seconds earlier in this example:

Screenshot from 2021-11-20 00-01-06

I have openpilot longitudinal in sng behaving nearly exactly like stock on TSS2, but I'll keep the changes to a minimum per PR and try to split them up under distinct fixes.

### DIFF
--- a/selfdrive/controls/lib/longcontrol.py
+++ b/selfdrive/controls/lib/longcontrol.py
@@ -107,7 +107,7 @@ class LongControl():
 
       # Toyota starts braking more when it thinks you want to stop
       # Freeze the integrator so we don't accelerate to compensate, and don't allow positive acceleration
-      prevent_overshoot = not CP.stoppingControl and CS.vEgo < 1.5 and v_target > v_target_future < 0.7
+      prevent_overshoot = not CP.stoppingControl and CS.vEgo < 1.5 and v_target_future < 0.7 and v_target_future < v_target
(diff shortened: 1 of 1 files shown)


pd0wm:
I'm not sure this will work properly when stopping. If you dip under the desired set speed the planner might command positive acceleration which won't be actuated by the PCM.

Isn't it better to tweak the conditions from v_future to correctly distinguish stopping from starting?

also, do we know if this hack is needed for TSS2 cars? I'm not sure if we ever checked that.

sshane:
I think you're right, my car only comes to a stop at 0.2 m/s or so, above that we seem to have full control.

sshane:
Also usually the planned acceleration is always negative when this is happening, but I'll switch to use v_target

sshane:
Re TSS2: I think after we get under 0.15-0.2 m/s the car comes to a stop if you're requesting 0 acceleration, however the lack of stopping control is not present no where near 1.5 m/s as described for the conditions of the overshoot prevention. And Once we're under 0.5 m/s we're usually in a stopping state, so this might not even be used for TSS2. Should we set stoppingControl to True for TSS2 cars as well in this PR?

PR #22991, opened 2021-11-20
Remove minSpeedCan

sshane:
Doesn't seem to make sense to keep `minSpeedCan` since it's not being compared to raw CAN data, and is now more used as a stopping speed threshold. With this PR we compare whether `vEgo` and `vTargetFuture` is less than `vEgoStopping`, so that we only stop if we're in the correct speed range, and the future desired speed is telling us to stop soon.

Changes in behavior:
- When in starting state, we can enter stopping state again up to 0.5 m/s instead of 0.3 m/s. Shouldn't matter that much since the car has a large delay to acceleration, and we exit starting state when output_accel is high enough, which is pretty quick.
- When coming to a stop, [vTarget is ideally essentially vEgo](https://i.imgur.com/9HudeRn.png), so previously the stopping state condition was: `vEgo < 0.5 and (vEgo < 0.31 and vTargetFuture < 0.31)`, now it's `vEgo < 0.5 and vTargetFuture < 0.5`

As expected, we now enter a stopping state when vEgo is less than 0.5, and vTargetFuture is already less than 0.3 m/s, so that didn't seem to be the limiting factor: https://i.imgur.com/jEg21FQ.png

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -82,6 +82,7 @@ class CarInterfaceBase():
     ret.minEnableSpeed = -1. # enable is done by stock ACC, so ignore this
     ret.steerRatioRear = 0.  # no rear steering, at least on the listed cars aboveA
     ret.openpilotLongitudinalControl = False
+    ret.minSpeedCan = 0.3
     ret.startAccel = -0.8
     ret.stopAccel = -2.0
     ret.startingAccelRate = 3.2 # brake_travel/s while releasing on restart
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Make sure fields are deprecated once they're unused.

adeebshihadeh:
Also make sure the tests pass. Process replay is now failing on master.

HaraldSchafer:
Yeah my bad, I'm running refs

PR #22992, opened 2021-11-20
framereader: replace swscale with libyuv,  reduce cpu usage by half 

deanlee:
libyuv has better performace than libswscale


pd0wm:
What was the reason to switch to swscale in this PR then? https://github.com/commaai/openpilot/pull/22710/files

deanlee:
switched back to libswscale in https://github.com/commaai/openpilot/pull/22710/files is because padding issue is solved, and swscale can support scaling and multiple format conversions more flexibly(since we only suport cuda and hevc,  this is not a problem anymore. )
I did not consider performance issues at that time.

PR #22995, opened 2021-11-21
panda: use static all_connected to track connectivity changes

deanlee:


adeebshihadeh:
This reduces lines, but seems worse conceptually. Why would a Panda care or have a concept of "all conencted"? I don't see the problem with how it's done now.

deanlee:
yes u r right, it's not very clear. closed.

PR #22996, opened 2021-11-21
CI: test build on latest Ubuntu

adeebshihadeh:


adeebshihadeh:
Turns out GH Action's `ubuntu-latest` is just 20.04. We'll probably want to do this on our Jenkins instance if we want it to be fast.

Issue #22062, opened 2021-08-28
Openpilot 0.8.8 triggers immediate LKAS, ACC, and Lane Departure Warning on Civic Hatchback


keijikage:
Installing Openpilot 0.8.8 on a C3 hooked up to a 2021 Civic Hatchback causes an immediate LKAS, ACC, and LDW warning.  Re-seated harness with no resolution.

Downgrading to  0.8.7. version resolves issue.

**Device/Version information (please complete the following information):**
 - Device:Comma 3
 - Dongle ID: 546b3e758686e75a
 - Route: 546b3e758686e75a|2021-08-28--16-26-33
 - Timestamp: Immediately on startup
 - Version: 0.8.8 Release 3
 - Car make/model: 2021 Civic Hatchback Sport

May impact more than one honda model, multiple users in discord indicating similar symptoms on C2 hardware as well

keijikage:
New route -  546b3e758686e75a|2021-08-29--01-42-44--0
Reset C3 prior to uploading prior logs.

riley206:
I'm having the same issue on my 2020 civic hatch. How do you go back to a previous update?

adeebshihadeh:
The issue appears to be caused by https://github.com/commaai/panda/pull/615. The new HUD msgs `0x33DA` and `0x33DB` found in the 2021 Accord also appear to be used on the Civic Hatch. #22063 should fix the issue.

riley206:
I'm honestly not to sure what all that means yet but I appreciate the response.

adeebshihadeh:
The fix is in master-ci and will go to the release branches shortly. If you'd like to install master-ci, you can use `openpilot-nightly.comma.ai`.

csouers:
@riley206 and @keijikage; When you engage openpilot on 0.8.7 or older and have NOT pressed the LKAS button on the steering wheel, do the LKAS lines show up solid white on the car's display?

keijikage:
> @riley206 and @keijikage; When you engage openpilot on 0.8.7 or older and have NOT pressed the LKAS button on the steering wheel, do the LKAS lines show up solid white on the car's display?

The last stock version I ran on any regular basis was 0.8.4, but it did not have this behavior - e.g. the observed behavior is that the lines were off when OP was off (LKAS not engaged).

The only exception to this is when the Comma is booting AND LKAS is engaged prior to fully booting - the car display can become inverted from the expected.

csouers:
[...]

PR #23000, opened 2021-11-21
Honda: use new Bosch LKAS_HUD messages

### LINKED ISSUE #22073: Honda: use new HUD msgs on newer Bosch models
I'm using stock OP 0.8.8. When I hit the LKAS button after OP is engaged using the set button, I get the lkas solid lines on the dash. After a few moments, I get "steering required" on the dashboard, if I don't grab the steering wheel within a few seconds, those solid lines become broken lines. The steering required message stays on until I grab the wheel. My C2 can see my face during this whole ordeal. My understanding is that the LKAS button is not suppose to do anything when OP is engaged.

adeebshihadeh:
resolves #22073

- [x] <s>determine all cars this applies to</s> auto detect instead


adeebshihadeh:
Hmm, appears `CIVIC_BOSCH` can be either the old or new style msg. The car from #22062 has the new msgs, but @csouers's has the old `0x33D`.

csouers:
Would it be sufficient to simply check for the new msg in the can fingerprint to determine what kind of hud frames to send? It would overall be less new lines in OP in exchange for a few new lines in cereal (a new struct). 

So far I recall seeing the new messaging on 2021+ Civic Bosch, Accord, CRV (not sure of which variant), Honda E, and Odyssey. Haven't seen a new Insight yet and I'm not 100% sure on the later RDXs. 

I've refactored my port in this way and have some users testing it now. I'll link it here in the morning.

csouers:
For example, [interface.py](https://github.com/csouers/openpilot/blob/1af371a3ca5afff0ad719ceb8d1da50cede1dc3f/selfdrive/car/honda/interface.py#L38) and [hondacan.py](https://github.com/csouers/openpilot/blob/1af371a3ca5afff0ad719ceb8d1da50cede1dc3f/selfdrive/car/honda/hondacan.py#L149)

[The cereal changes that I mentioned.](https://github.com/csouers/cereal/commit/a1eb2a93dc40b44894906b42796d00280df6119d)
[Here's the branch comparison of what I have against comma master](https://github.com/commaai/openpilot/compare/master...csouers:bosch2021_test?expand=1)

Several 2021+ Accord firmware IDs are already in master, so this seems like a simple way to handle all of those without having to split them back out into new CARs.

selfdrive/car/honda/hondacan.py
```diff
@@ -142,16 +142,24 @@ def create_ui_commands(packer, pcm_speed, hud, car_fingerprint, is_metric, idx,
     'SOLID_LANES': hud.lanes,
     'BEEP': 0,
   }
-  commands.append(packer.make_can_msg('LKAS_HUD', bus_lkas, lkas_hud_values, idx))
 
-  if radar_disabled and car_fingerprint in HONDA_BOSCH:
+  if CP.flags & Flag.BOSCH_EXT_HUD:
+    lkas_hud_values['SET_ME_X48'] = 0x48
+
+  if CP.flags & Flag.BOSCH_EXT_HUD and not CP.openpilotLongitudinalControl:
```
adeebshihadeh, line 149: @csouers why is `openpilotLongitudinalControl` checked here?

adeebshihadeh:
Auto detection definitely seems to be the way to go here. @csouers are you saying that we can just merge the new 2021 Accord cars into the old ones after this PR?

selfdrive/car/honda/hondacan.py
csouers: The radar passes along the new LKAS_HUD messages to the powertrain bus on 0x33d. If it's disabled, we just send LKAS_HUD like usual.

csouers:
> Auto detection definitely seems to be the way to go here. @csouers are you saying that we can just merge the new 2021 Accord cars into the old ones after this PR?

Should be fine. I've been using the same tuning as the old cars and haven't received any complaints about poor behavior. At least a few 2021 Civic and CR-V users have also been doing the same as well.

selfdrive/car/honda/values.py
```diff
@@ -37,6 +39,11 @@ def __init__(self, CP):
     self.STEER_LOOKUP_V = [v * -1 for v in CP.lateralParams.torqueV][1:][::-1] + list(CP.lateralParams.torqueV)
 
 
+class Flag(IntFlag):
+  # Bosch models with alternate set of LKAS_HUD messages
+  BOSCH_EXT_HUD = 1
+
+
```
belm0, line 46: suggestions:
  * use a name like `CarParamFlags`.  `Flag` is going to cause confusion in any module that happens to import `enum.Flag`.
  * derive from `Flag` rather than `IntFlag`.  This won't affect your existing code like `x |= BOSCH_EXT_HUD.value`.  `IntFlag` is intended more for use with legacy API's, letting you pass an `IntFlag` instance where an `int` is expected-- the downside is less type safety.

belm0: After experimenting with this CP.flags approach, I found it limiting because Honda-specific details include non-flag values like STEER_THRESHOLD and SPEED_FACTOR.

Consider adopting #23053 instead, as it works generally across bool, number, string, list, etc.

adeebshihadeh: renamed to HondaFlags. I'll respond to your alternative approach on that PR

adeebshihadeh:
@csouers The tests say this is good as far as catching the regression from last time we merged the 2021 Accord, but can you confirm everything's working as expected on master-ci with the users of the relevant cars?

selfdrive/car/honda/hondacan.py
```diff
@@ -142,16 +142,24 @@ def create_ui_commands(packer, pcm_speed, hud, car_fingerprint, is_metric, idx,
     'SOLID_LANES': hud.lanes,
     'BEEP': 0,
   }
-  commands.append(packer.make_can_msg('LKAS_HUD', bus_lkas, lkas_hud_values, idx))
 
-  if radar_disabled and car_fingerprint in HONDA_BOSCH:
+  if CP.flags & HondaFlags.BOSCH_EXT_HUD:
```
adeebshihadeh, line 146:
```suggestion
  if not (CP.flags & HondaFlags.BOSCH_EXT_HUD):
```

adeebshihadeh:
@csouers any update?

csouers:
> @csouers any update?

Haven't heard back. I'll ask a few more people that I have in mind.

adeebshihadeh:
Thanks, would be great to get some confirmation before 0.8.12 goes to release.

csouers:
Civic 2021+ is confirmed working. Waiting on Accord and CR-V.

csouers:
Accord 2021 2.0 is also confirmed.

adeebshihadeh:
Thanks @csouers!

PR #23004, opened 2021-11-22
VW MQB: Longitudinal prereqs

jyoung8607:
Getting some of the diff noise / prerequisite pieces out of the way for VW MQB longitudinal:

- [x] Detect and alert on TSK faults as ACC faults (works and is useful for both stock and OP long)
- [x] Remember TSK (TreibStrang-Koordinator, drivetrain coordinator) state
- [x] Remember ABS/ESP hydraulic brake hold state
- [x] Apply fix from #22176 directly to button-spam code instead of standstill
- [x] Gate button-spam code behind `CP.pcmCruise`
- [x] Gate stock ACC setpoint handling behind `CP.pcmCruise`

### DIFF
--- a/selfdrive/car/volkswagen/interface.py
+++ b/selfdrive/car/volkswagen/interface.py
@@ -43,7 +43,7 @@ class CarInterface(CarInterfaceBase):
       else:
         ret.networkLocation = NetworkLocation.fwdCamera
 
-    # Global tuning defaults, can be overridden per-vehicle
+    # Global lateral tuning defaults, can be overridden per-vehicle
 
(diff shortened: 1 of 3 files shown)


selfdrive/car/volkswagen/interface.py
```diff
@@ -190,6 +190,8 @@ def update(self, c, can_strings):
     # Vehicle health and operation safety checks
     if self.CS.parkingBrakeSet:
       events.add(EventName.parkBrake)
```
adeebshihadeh, line 192: side ntoe: we should add this to carState and move this to the common events

PR #23012, opened 2021-11-23
CARLA: send YUV and RGB over visionipc from bridge.py

jwolffe:
This PR incorporates visionIPC outputs for RGB frames as in #21324 together with YUVframes. Conversion is via pyopencl, using the GPU,. It uses the existing OpenCL kernel file rgb_to_yuv.cl in camerad. The camerad process is replaced by bridge.py during simulation. This also solves the problem previously addressed by #22837. The pyopencl package is added to the Pipfile. The YUV output has been tested together with the changes in #22905.


tools/sim/bridge.py
```diff
@@ -54,17 +57,65 @@ def steer_rate_limit(old, new):
...
 frame_id = 0
+vipc_server = None
+ctx = None
+queue = None
+krnl = None
+Wdiv4 = 0 
+Hdiv4 = 0 
 def cam_callback(image):
   global frame_id
+  global vipc_server
```
adeebshihadeh, line 68: this is way too many globals. let's put all this stuff in a camerad class

adeebshihadeh:
Nice, how fast is this? Does it run at 20Hz comfortably?

tools/sim/bridge.py
jwolffe: My use of the globals was to prevent redundant recalculations and allocation/deallocation on the repeated callbacks, particularly on the GPU that may not handle it that well. I'm not very familiar with Python. Can you offer an example of how to avoid the globals and achieve the same result? Or are you simply saying that you want to "bundle them up" in a single global class?

adeebshihadeh: yeah, just wrap all this in a class

jwolffe: I haven't studied performance very carefully, but it seems to consume pretty neglible gpu and cpu on my machine.  No problem keeping up at 20 Hz. I will try to characterize it a bit better.

adeebshihadeh: cool, main concerns is is it doing 20Hz comfortably on modest hardware?

jwolffe: I have a pretty powerful machine, 8 core Alienware area51 with rtx2070.  I'll try setting up some "torture testing", perhaps by repeating the whole process N times and see where it breaks down.

jwolffe: Camerad class is incorporated now. I took some performance measurements, and they look pretty good. I iterated the yuv conversion routine N times and observed the total bridge.py CPU usage, where 12.5 % is one fully occupied CPU core. I did not repeat any of the I/O. When I got up to 14 iterations some rare "slow frame" events happened. Hardware able to run Carla should easily work.

       N   %CPU
  	1	5
  	3	6
 	5	7
 	7	8
 	8	9
 	9	9
 	10	10
 	11	10
 	12	11
 	13	12
 	14	12 (very rare slow frame)
 	15	12 (rare slow frame)
 	16    12 (some slow frames)

pd0wm:
Made few more changes and everything seems to work out of the box now.
2021-11-26-160030_2235x1162_scrot

Will look at the onnx-runtime in a separate PR.

jwolffe:
> 
That's nice! If you like, I will put together a PR for the onnxruntime issue. I have already done some experimentation.

pd0wm:
Yeah, sounds good! I quickly installed `onnxrutime-openmp` but didn't see any changes.

jwolffe:
> Yeah, sounds good! I quickly installed `onnxrutime-openmp` but didn't see any changes.
export OMP_WAIT_POLICY=PASSIVE should make it work.

pd0wm:
Is it supposed to show up as it's own execution provider? I'm only seeing `CPUExecutionProvider`

jwolffe:
> Is it supposed to show up as it's own execution provider? I'm only seeing `CPUExecutionProvider`

I didn't check that, but I did see the CPU usage drop by more than half. I'm in the middle of a fresh install right now, will let you know what provider is reported when I get up and running again.

jwolffe:
It looks like we have some performance problems, possibly due to the use of higher resolution video. My fresh install of the latest code runs terribly, with lots of "slow frame time" reports. Interestingly, the CPU usage is low. I think Carla is having problems keeping up with the increased video resolution. Running bridge.py with --low_quality does much better, with very few slow frame times reported, but the video still seems to "stutter" badly.

jwolffe:
A bit more on the performance issue. The image.frame counter from Carla used to run at three times the frame_id rate (when using the 1164X874 frames).  Apparently Carla was trying to generate frames at 3X the 20Hz basic rate, and output only every third one. Now the image.frame and frame_id rates are the same. We now only complete 8-10 frames per second. I suspect that too much time may be lost in transfer time, such that the callback routine doesn't return in time for the next "tick" in Carla.

Using the low_quality option causes many more frames to be completed per second, looks to be at or very close to the nominal 20. But the frame.id count sometimes steps by 6, not the previously expected steps of 3.

I suspect the problem may be related to trying to move too much data, leading to latency issues. Perhaps the transition to the higher resolution video should wait until the UI starts using YUV, when the RGB output from the bridge can be eliminated.

pd0wm:
Have you tried running carla in fixed time step mode (-fps=20), and maybe also in synchronous mode? That might prevent it from rendering useless frames.

pd0wm:
Made a PR with my changes: https://github.com/commaai/openpilot/pull/23046. Let's continue our discussing there!

PR #23022, opened 2021-11-24
cameraview: fix frameskip

deanlee:
use repaint to fix the frameskip when there are too many events in the queue (e.g.  watch3)

### DIFF
--- a/selfdrive/ui/qt/widgets/cameraview.cc
+++ b/selfdrive/ui/qt/widgets/cameraview.cc
@@ -247,7 +247,7 @@ void CameraViewWidget::vipcConnected(VisionIpcClient *vipc_client) {
 
 void CameraViewWidget::vipcFrameReceived(VisionBuf *buf) {
   latest_frame = buf;
-  update();
+  repaint();
 }
 
 void CameraViewWidget::vipcThread() {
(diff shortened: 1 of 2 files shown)


pd0wm:
Doesn't it make more sense to skip frames so we can keep up realtime?

adeebshihadeh:
What events are dominating the queue? Perhaps we're spamming the queue with nonsense timer events or something else?

deanlee:
We can easily implement an OffroadTimer like this if  use singleton uistate  https://github.com/commaai/openpilot/pull/22789

```
OffroadTimer(QObject *parent) : QTimer(parent) {
  QObject::connect(uiState(), &UIState::offroadTransition, [=](bool offroad) {
    offroad ? start() : stop();
  });
}
```

deanlee:
> What events are dominating the queue? Perhaps we're spamming the queue with nonsense timer events or something else?

At present, this drawing problem is only in wath3,  it is mainly caused by a large number of  update events from the three cameras.

deanlee:
Think we don't need this hack if use YUV vipc stream.

pd0wm:
Can we ensure the texture is not being modified while Qt is actually drawing the ui? Even with YUV that might still be an issue, right?

deanlee:
Fixed, the uploaded and render have been synchronized.

deanlee:
> Can we ensure the texture is not being modified while Qt is actually drawing the ui? Even with YUV that might still be an issue, right?

YUV https://github.com/commaai/openpilot/pull/22905 does not has this issue,  the performance looks good enough , so I did not do asynchronous upload for it

PR #23031, opened 2021-11-25
Clean up planner files

sshane:


selfdrive/controls/lib/lateral_planner.py
```diff
@@ -67,12 +67,8 @@ def __init__(self, CP, use_lanelines=True, wide_camera=False):
   def reset_mpc(self, x0=np.zeros(6)):
     self.x0 = x0
     self.lat_mpc.reset(x0=self.x0)
-    self.desired_curvature = 0.0
-    self.safe_desired_curvature = 0.0
-    self.desired_curvature_rate = 0.0
-    self.safe_desired_curvature_rate = 0.0
```
pd0wm, line 73: @HaraldSchafer Was the rate limit removed in the MPC refactor?

sshane: Think it was moved here https://github.com/commaai/openpilot/blob/e376a621baa125063ffbf1fbd91560e25d45e8fd/selfdrive/controls/lib/drive_helpers.py#L98

pd0wm: Ah, I see :)

PR #23032, opened 2021-11-25
UI minor changes to make it translation friendly

eFiniLan:
* SshControl - add a button state instead of text() to check it's status, so we can safely translate "ADD", "REMOVE", "LOADING"
* having "ALERT" and "ALERTS" as 2 string instead of appending "S".

### DIFF
--- a/selfdrive/ui/qt/widgets/ssh_keys.cc
+++ b/selfdrive/ui/qt/widgets/ssh_keys.cc
@@ -46,13 +46,13 @@ void SshControl::getUserKeys(const QString &username) {
       params.put("GithubUsername", username.toStdString());
       params.put("GithubSshKeys", resp.toStdString());
     } else {
-      ConfirmationDialog::alert("Username '" + username + "' has no keys on GitHub", this);
+      ConfirmationDialog::alert(QString("Username '%1' has no keys on GitHub").arg(username), this);
     }
     refresh();
     request->deleteLater();
   });
   QObject::connect(request, &HttpRequest::failedResponse, [=] {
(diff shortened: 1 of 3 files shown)


selfdrive/ui/qt/widgets/ssh_keys.cc
```diff
@@ -4,16 +4,20 @@
 #include "selfdrive/ui/qt/api.h"
 #include "selfdrive/ui/qt/widgets/input.h"
 
+const int BTN_STATE_REMOVE = 0;
```
adeebshihadeh, line 7: make this an enum

eFiniLan: updated! thanks!

selfdrive/ui/qt/widgets/ssh_keys.cc
```diff
@@ -10,10 +10,10 @@ SshControl::SshControl() : ButtonControl("SSH Keys", "", "Warning: This grants S
   hlayout->insertWidget(1, &username_label);
 
   QObject::connect(this, &ButtonControl::clicked, [=]() {
-    if (text() == "ADD") {
```
deanlee, line 13: why not use `if (text() == tr("ADD"))` ?  doesn't it work?

eFiniLan: Hi @deanlee,
ahh.. sorry my fault, I tested it back in 0.8.7/0.8.8 and it wasn't working, I must have forgot to translate one of the other strings, I will reverse the changes to ssh_keys :)

PR #23033, opened 2021-11-25
boardd: new function sync_time

deanlee:
two-way sync time in function sync_time

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -78,7 +78,7 @@ bool check_all_connected(const std::vector<Panda *> &pandas) {
   return true;
 }
 
-enum SYNC_TIME_DIR { SYNC_TO_PANDA, SYNC_FROM_PANDA };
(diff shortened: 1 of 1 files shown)


pd0wm:
I like to prevent changes to the system time as much as possible, so like to do this only once on connect. Can you add `direction` argument to the sync time function to specify which direction we want to sync?

selfdrive/boardd/boardd.cc
```diff
@@ -78,6 +78,35 @@ bool check_all_connected(const std::vector<Panda *> &pandas) {
   return true;
 }
 
+enum SYNC_TIME_DIR { SYNC_TO_PANDA, SYNC_FROM_PANDA };
```
pd0wm, line 81:
```suggestion
enum SyncTimeDir { SYNC_TO_PANDA, SYNC_FROM_PANDA };
```

pd0wm:
The actual values should still be all caps.

pd0wm:
why the lambda? Just make the function static.

PR #23037, opened 2021-11-25
Python minor style improvement

ghost:


adeebshihadeh:
This is actually faster, not just a style improvement, since you're creating a generator instead of a list comprehension.

PR #23039, opened 2021-11-26
refactor httprequest: emit single signals

deanlee:


pd0wm:
Not sure one is better than the other. Any specific reason for making this change?

deanlee:
HttpRequest has four signals after request finished. I think this is obviously too much. we only need to know whether it succeeded or failed.  it seems that one signal is enough

PR #23040, opened 2021-11-26
UI: move onroad hud to Qt

deanlee:
split from https://github.com/commaai/openpilot/pull/22173
hud2

### DIFF
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -142,7 +142,7 @@ static void ui_draw_vision(UIState *s) {
   if (scene->world_objects_visible) {
     ui_draw_world(s);
   }
-  // TODO: do we need this?
+  // TODO: move this to Qt
   ui_draw_vision_header(s);
 }
 
--- a/selfdrive/ui/paint.cc
+++ b/selfdrive/ui/paint.cc
@@ -16,16 +16,7 @@
 #include <nanovg_gl.h>
 #include <nanovg_gl_utils.h>
 
-#include "selfdrive/common/util.h"
 #include "selfdrive/hardware/hw.h"
-#include "selfdrive/ui/ui.h"
-
-static void ui_draw_text(const UIState *s, float x, float y, const char *string, float size, NVGcolor color, const char *font_name) {
-  nvgFontFace(s->vg, font_name);
-  nvgFontSize(s->vg, size);
-  nvgFillColor(s->vg, color);
-  nvgText(s->vg, x, y, string, NULL);
-}
 
 static void draw_chevron(UIState *s, float x, float y, float sz, NVGcolor fillColor, NVGcolor glowColor) {
   // glow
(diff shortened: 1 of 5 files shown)


selfdrive/ui/qt/onroad.cc
```diff
@@ -162,6 +172,108 @@ void OnroadAlerts::paintEvent(QPaintEvent *event) {
   }
 }
 
+// OnroadHud
+OnroadHud::OnroadHud(QWidget *parent) : QWidget(parent) {
+  engage_img = QPixmap("../assets/img_chffr_wheel.png").scaled(img_size, img_size, Qt::KeepAspectRatio, Qt::SmoothTransformation);
```
adeebshihadeh, line 177: let's make a helper for this in qt utils. we do this in quite a few places

adeebshihadeh: think it's best if the helper and using it is its own PR

deanlee: i was thinking about this :D

adeebshihadeh:
Can you post some image diffs for verification with these UI refactors?

deanlee:
the "red color" on the background is painted by nanovg at the same time with COLOR_RED_ALPHA:

head3

they are almost the same, but it’s hard to be exactly the same because nanovg scales the font on painting.

deanlee:
improved the boundrect:
h1

old hud(nanovg) paint at the same time with Qt hud, no diff now. 

good

selfdrive/ui/paint.cc
```diff
@@ -213,11 +142,8 @@ static void ui_draw_vision(UIState *s) {
   if (scene->world_objects_visible) {
     ui_draw_world(s);
   }
-  // Set Speed, Current Speed, Status/Events
+  // TODO: do we need this?
```
adeebshihadeh, line 145: yes, we still need this. let's move it into it's own widget

selfdrive/ui/qt/onroad.cc
```diff
@@ -27,11 +27,16 @@ OnroadWindow::OnroadWindow(QWidget *parent) : QWidget(parent) {
 
   stacked_layout->addWidget(split_wrapper);
 
+  hud = new OnroadHud(this);
+  hud->setAttribute(Qt::WA_TransparentForMouseEvents, true);
+  stacked_layout->addWidget(hud);
```
adeebshihadeh, line 32: this should be part of a new stacked layout together with the nvg widget, then you can get rid of setMapWidth

selfdrive/ui/qt/onroad.cc
```diff
@@ -150,6 +159,99 @@ void OnroadAlerts::paintEvent(QPaintEvent *event) {
...
+// OnroadHud
+OnroadHud::OnroadHud(QWidget *parent) : QWidget(parent) {
+  engage_img = QPixmap("../assets/img_chffr_wheel.png").scaled(img_size, img_size, Qt::KeepAspectRatio, Qt::SmoothTransformation);
+  dm_img = QPixmap("../assets/img_driver_face.png").scaled(img_size, img_size, Qt::KeepAspectRatio, Qt::SmoothTransformation);
+
+  connect(this, &OnroadHud::valueChanged, [=] { update(); });
+}
+
+void OnroadHud::updateState(const UIState &s) {
+  auto getMaxSpeed = [&s](float maxspeed) {
```
adeebshihadeh, line 171: no need for the lambda here

selfdrive/ui/qt/onroad.cc
```diff
@@ -17,13 +17,18 @@ OnroadWindow::OnroadWindow(QWidget *parent) : QWidget(parent) {
   stacked_layout->setStackingMode(QStackedLayout::StackAll);
   main_layout->addLayout(stacked_layout);
 
+  QStackedLayout *hud_stacked_layout = new QStackedLayout;
```
adeebshihadeh, line 20: maybe call this road_view_layout?

PR #23044, opened 2021-11-26
Update hyundai elantra hb values.py

maxy0619:
Added hyundai elantra hybrid 2021 fwd camera, eps firmware versions.


pd0wm:
What's the dongle id?

PR #23046, opened 2021-11-26
sim: improve carla performance

pd0wm:
- Change simulator to `High`. Change weather settings to something reasonable to compensate
- Run in synchronous mode at 20 fps to prevent rendering useless frames

Still seeing some weird stutter in the UI, but not `tools/replay/ui.py` so think that's a separate problem (maybe made worse by the high CPU usage from onnx runner).

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -221,13 +221,20 @@ def bridge(q):
   client.set_timeout(10.0)
   world = client.load_world(args.town)
 
+  settings = world.get_settings()
+  settings.synchronous_mode = True # Enables synchronous mode
+  settings.fixed_delta_seconds = 0.05
+  world.apply_settings(settings)
+
+  world.set_weather(carla.WeatherParameters.ClearSunset)
+
   if args.low_quality:
     world.unload_map_layer(carla.MapLayer.Foliage)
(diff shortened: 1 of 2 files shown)


jwolffe:
I tried your fixes, and things seem much better. No slow frames, and CPU usage seems reasonable. The bridge.py shows up as about 7% (equiv to 56% of a single core). I do see the video corruption you mentioned. Looks like some sort of "tearing" where portions of old and new buffers are getting mixed. Reading while writing, perhaps? I tried increasing the RGB buffer size, but that just makes the UI crash- obviously I don't understand the buffer scheme. I will try to look into the video corruption and see if I can come up with a fix. I don't think its a CPU loading issue- I'm only at about 45% for my whole machine.

pd0wm:
No need to dive to deep into the ui issues. Think it was introduced when we started to copy the video into the texture asynchronously. Works fine on the comma three but not on PC for some reason.

PR with more discussion: https://github.com/commaai/openpilot/pull/23022

Would be nice if you can figure out a nice solution to the onnx cpu usage!

jwolffe:
Synchronous mode certainly seems the right approach.  I'll put aside the video issue for now and focus on onnxruntime.

PR #23048, opened 2021-11-27
builds on M1 mac

geohot:
Should be okay to merge


geohot:
The paren thing seems to be an issue

geohot:
The web downloads from replay are coming up all 0s

geohot:
Working to replay in UI

pd0wm:
https://github.com/commaai/openpilot/blob/master/tools/mac_setup.sh works correctly on M1?

deanlee:
is there any special reason for removed the support downloading to a file?

geohot:
The stringstream doesn't work on mac for some reason, the returned string was all 0s. I looked for a bit and couldn't find the bug, so I switched it to this and it worked.

geohot:
> https://github.com/commaai/openpilot/blob/master/tools/mac_setup.sh works correctly on M1?

I ran that in pieces. The brew stuff works correctly. For pipenv I installed into my system Python of 3.9.9

deanlee:
> The stringstream doesn't work on mac for some reason, the returned string was all 0s. I looked for a bit and couldn't find the bug, so I switched it to this and it worked.

fixed this issue in https://github.com/commaai/openpilot/pull/23052

khoi:
I ran `./tools/mac_setup.sh` in a fresh M1 machine. Got an error when installing `scipy `

```
  Error: 'pybind11' must be installed before running the build.
```

adeebshihadeh:
M1 phonelibs are under `mac_arm64`, but we should probably move those to `Darwin` and move the x86 stuff to `mac_x86`.

geohot:
Rebased, the new changes are a lot simpler.

adeebshihadeh:
LGTM. I’ll clean up the phonelibs convention for Intel and M1 after this is merged.

PR #23049, opened 2021-11-27
replay: use std::unique_ptr to manage memory, reuse events buffer.

deanlee:


adeebshihadeh:
This is somewhat less readable. How much of a difference does this make?

deanlee:
which part is less readable and how I can improve it?
use std::unique_ptr and swap to manage memory is better than  manually.  it can also reduce the overhead of frequent memory re-allocation during merge.

PR #23050, opened 2021-11-27
logreader: support reading from corrupt log

deanlee:
1. logreader: support reading from corrupt log
2. logreader&framereader: support reading from buffer.


pd0wm:
Does this fix the replay getting stuck if the last log is corrupt?

deanlee:
no, I think it can’t. where is it stuck ?  Is it in decompressbz2 or FlatArrayMessageReader?

pd0wm:
Will do some more investigation the next time it happens.

deanlee:
@pd0wm  It's stuck in decompressbz2 if log is corrupt

PR #23051, opened 2021-11-27
PC: fix onnxruntime excessive CPU

### LINKED ISSUE #22736: 100% CPU on onnxruntime threads caused by spin-waits

jwolffe:
Excessive CPU was observed in simulations with Carla #22736. This PR  fixes it by supporting additional Execution Providers for onnxruntime. Apparently the basic onnxruntime package from PyPi has spin-waits that cannot be disabled. Several alternatives were explored, using pip installs of onnxruntime-gpu, onnruntime-openmp, and onnxruntime built from source to support GPU operation with CUDA on the RTX2070 test platform. Attempts to get the "stock" PyPi onnxruntime-gpu package to work with CUDA were unsuccessful, with the onnxruntime defaulting to CPU operation. Based on the measurements of CPU load and execution time of the model, I have tentatively chosen to set onnxruntime-openmp to be installed automatically from the pipfile, using PASSIVE wait policy. This PR supports users changing to other execution providers without modification of code, by installing alternative packages with pip. The following measurements were made with the changes in PR #23046 incorporated.

Performance measurements (100% CPU = 8 cores fully busy):
onnxruntime-openmp  PASSIVE policy	.019 sec	11% CPU
onnxruntime-openmp  ACTIVE policy 	.018 sec 	23% CPU
onnxruntime-gpu (cuda,  compiled source)  	.012 sec	 3% CPU
onnxruntime-gpu (no cuda, CPU 2 threads)	.038 sec 	20% CPU
onnxruntime (CPU spin-wait, 4 threads)	.018 sec 	41% CPU


pd0wm:
Looks like the openmp version is no longer published since version 1.8.0 (https://github.com/microsoft/onnxruntime/releases/tag/v1.8.0). Any reason this was deprecated?

jwolffe:
My understanding is that openvino is intended to provide GPU acceleration with Intel graphics hardware. Onnxruntime-gpu gives acceleration for Nvidia/CUDA hardware. I assume (perhaps incorrectly) that more users trying to run Carla/openpilot would have Nvidia.  My understanding is that most Intel GPU hardware is too "lightweight" to be of much help running Carla. CPU-only, with the help of threading on multicore processors, seems entirely reasonable for running the model in openpilot, though.

A version of CPU threading is now incorporated in the basic onnxruntime package, and as far as I can tell, they dropped ongoing support  for openmp threading because it became functionally redundant.  The only reason I recommended the stale onnxruntime-openmp package over basic onnxruntime is because it doesn't have the unavoidable spin-waits. When/if onnxruntime allows disabling spin waits I would recommend using it over the onnxruntime-openmp.

pd0wm:
OpenVINO also has a CPU only mode, which we use extensively in our data processing pipeline (uses Intel MKL).

jwolffe:
> OpenVINO also has a CPU only mode, which we use extensively in our data processing pipeline (uses Intel MKL).

I never tried openvino.  If you are comfortable with the threading support it offers, I would go along with making it the default choice in the Pipfile.

pd0wm:
Let me know if it works well for you. Not sure if we can add it the Pipfile by default, might give issues with installing on MacOS.

jwolffe:
I have been trying to get openvino working with onnxruntime,  no success so far. Have you made it work?

pd0wm:
Apparently we already moved away from it again internally. @gregjhogan just mentioned that the onnxrutime-gpu also includes the cpu runtime. If that's the case installing that package by default might make the most sense.

jwolffe:
> Apparently we already moved away from it again internally. @gregjhogan just mentioned that the onnxrutime-gpu also includes the cpu runtime. If that's the case installing that package by default might make the most sense.

Unfortunately, the onnxruntime-gpu also has the unwanted spin waits when defaulting to CPU. Perhaps a "good enough" solution might be to just reduce the intra-op threads. I suggest 2 might be a reasonable choice, and just keep the plain onnxruntime in the Pipfile as the default.  The mods to onnx_runner.py would then allow further tailoring by the user in a wide variety of ways.

pd0wm:
Wouldn't most people be able to use the cuda version though? We can change the intra-op threads to 2 as a fallback. 

What would be the reason to put the plain onxxruntime in the pipfile over the gpu one?

jwolffe:
> Wouldn't most people be able to use the cuda version though? We can change the intra-op threads to 2 as a fallback.
> 
> What would be the reason to put the plain onxxruntime in the pipfile over the gpu one?

There is no particular advantage to the plain one. My experience was that I couldn't get the pypi onnxruntime-gpu to work with my Nvidia CUDA setup, I had to build it from source.  Others may find the magic combination to make the stock onnxruntime-gpu version work. So, perhaps we can converge on a solution here- onnxruntime-gpu in the Pipfile,  and intra-op threads set to 2 for CPU provider. Does that sound good to you? At least there will be only 2 threads running at ~100% in the absence of a working CUDA config.

jwolffe:
I did some investigation into why my CUDA setup didn't work with the stock Pypi version of onnxruntime-gpu. It appears that there is a cuda bug in the stock Pypi package. When I ran "ldd /home/wolffja/.local/lib/python3.8/site-packages/onnxruntime/capi/libonnxruntime_providers_cuda.so" I got the following:
Inconsistency detected by ld.so: dl-version.c: 205: _dl_check_map_versions: Assertion `needed != NULL' failed!

I then ran "LD_DEBUG=all ldd /home/wolffja/.local/lib/python3.8/site-packages/onnxruntime/capi/libonnxruntime_providers_cuda.so" and the rather lengthy output included the following lines:
...
      5839:	
      5839:	checking for version `GCC_3.0' in file /lib/x86_64-linux-gnu/libgcc_s.so.1 [0] required by file /home/wolffja/.local/lib/python3.8/site-packages/onnxruntime/capi/libonnxruntime_providers_cuda.so [0]
      5839:	checking for version `GLIBC_2.3' in file /lib64/ld-linux-x86-64.so.2 [0] required by file /home/wolffja/.local/lib/python3.8/site-packages/onnxruntime/capi/libonnxruntime_providers_cuda.so [0]
Inconsistency detected by ld.so: dl-version.c: 205: _dl_check_map_versions: Assertion `needed != NULL' failed!
...

A solution to this is not obvious to me. Hopefully, a future update from 1.9.0 may fix this. Until then, it appears that openpilot users will have to build their own onnxruntime from source to get CUDA support.

PR #23053, opened 2021-11-28
honda refactor: associate a "details" structure with CarParams

belm0:
Previously we maintained ad-hoc fingegprint lists and
car-to-value mappings (e.g. HONDA_BOSCH, STEER_THRESHOLD, etc.).

Now there is a HondaDetails structure that can be accessed anywhere
given the CarParams object.  HondaDetails fields are typed and
have a default.

advantages:
  * honda-specific details can be set together with CarParams
    fields in the candidate cases of interface.py, making car
    configuration more centralized
  * references to CAR and fingerprints are reduced (e.g. hondacan.py
    no longer has such references)
  * treatment of flag and non-flag details are unified

This approach has specific advantages over introducing a CarParams.flags
field (#23000):
  * not limited to flags-- we can have any type of value (number, string),
    and specify a default.  (See e.g. speed_factor and steer_threshold.)
  * testing of flags is more readable, e.g.
    "details.is_bosch_ext_hud" vs. "flags & HondaFlags.BOSCH_EXT_HUD"


adeebshihadeh:
Thanks for the PR, and thinking about how to improve our abstractions. However, I don't think this is the way we want to solve this. I think this car/make specific pattern will let us avoid figuring out how to fit these things in carParams, when in fact most things are generic across cars and should live in carParams. For example, we just moved Honda's specific `SPEED_FACTOR` into the new `carParams.wheelSpeedFactor` field.

belm0:
> I think this car/make specific pattern will let us avoid figuring out how to fit these things in carParams, when in fact most things are generic across cars and should live in carParams. For example, we just moved Honda's specific `SPEED_FACTOR` into the new `carParams.wheelSpeedFactor` field.

Absolutely, parameters that can be general across OEM's should be, and `wheelSpeedFactor` is a good addition.

On the other hand, there will always be Bosch vs. not, and cars within an OEM that deviate in their use of certain CAN messages and signals.  It's at the whim of the OEM how fractured their use of CAN is.  But it's fair if your premise is that OEM quirks can likely be limited to booleans (which covers car sets). 

**Would you agree with a PR that replaces `HONDA_BOSCH` and the other car sets with use of `CP.flags`?**  The current situation is a little unwieldy, because these lists are separate from the config of each CAR in `interface.py`.

adeebshihadeh:
> **Would you agree with a PR that replaces `HONDA_BOSCH` and the other car sets with use of `CP.flags`?** The current situation is a little unwieldy, because these lists are separate from the config of each CAR in `interface.py`.

I don't think so. CP.flags still isn't great, but it was necessary since the information to determine whether a car uses the extended HUD msgs isn't entirely in CarParams, where as it is enough for HONDA_BOSCH and the other lists. I'm open to replace the flags field with something better, though. `CP.carFingerprint in HONDA_BOSCH` isn't too bad, so any proposed improved would have to be significantly better to warrant the move.

belm0:
> `CP.carFingerprint in HONDA_BOSCH` isn't too bad, so any proposed improved would have to be significantly better to warrant the move.

Set tests like `CP.carFingerprint in HONDA_BOSCH` are concise and efficient-- that isn't the issue.

The issue is from the point of view of people adding and maintaining car ports.  Rather than have a local 20 lines of code in interfaces.py where everything is configured for a car, they need to know about the various car sets that may need amending.  If we make those part of HondaFlags, then the surface area of each car config is reduced.

adeebshihadeh:
> The issue is from the point of view of people adding and maintaining car ports. Rather than have a local 20 lines of code in interfaces.py where everything is configured for a car, they need to know about the various car sets that may need amending. If we make those part of HondaFlags, then the surface area of each car config is reduced.

I agree, but I think there's more low hanging fruit in car port complexity than this. The Honda DBCs are a huge mess, for example. You're welcome to try to tackle this, it'll just need to be enough of an improvement for us to want to move over.

belm0:
> The Honda DBCs are a huge mess, for example. You're welcome to try to tackle this, it'll just need to be enough of an improvement for us to want to move over.

A lot within the DBC files can probably be consolidated-- I did a little with the Freed port.

PR to move to HondaFlags wouldn't be hard at this point, I may try and see how it looks.

PR #23069, opened 2021-11-30
loggerd: write out logs uncompressed

adeebshihadeh:


deanlee:
Does this save CPU usage? I used to think that most of the cpu is consumed by the encoders

PR #23076, opened 2021-11-30
Sound: use setLoopCount(0) to stop repeating sound

deanlee:
don't stop at the middle of the playing.


adeebshihadeh:
Does this work? Any reason it's still a draft?

deanlee:
lgtm

PR #23083, opened 2021-11-30
add Elantra '22 model to CARS.md

tecandrew:
'22 model (hybrid/non-hybrid) are carryover models of the '21 (no changes between the two). some new users on Discord have asked this a few times

https://www.hyundainews.com/en-us/models/hyundai-elantra-2022-elantra


adeebshihadeh:
We only bump the model years once openpilot has actually driven on one. Has this been confirmed yet?

tecandrew:
Not confirmed, just basing this bump off Hyundai's press release notes. Unsure if there are new/current users that have the '22 model, but it's been asked before if OP supports it.

adeebshihadeh:
Until openpilot has driven a 2022, the answer is “it’s likely supported but unconfirmed”.

PR #23101, opened 2021-12-02
Decay PI integrator term over time

haraschax:


neokii:
An error occurs when assigning to actuators.accel due to type mismatch.

self.i_decay_factor = **float(**np.exp(-1.0/(rate*i_decay_tau + 1e-6))**)**

pd0wm:
We decided to replace this by a clip based on current kI

PR #23103, opened 2021-12-02
qt/util: new function scaledPixmap

deanlee:


selfdrive/ui/qt/util.h
```diff
@@ -22,6 +23,9 @@ void swagLogMessageHandler(QtMsgType type, const QMessageLogContext &context, co
 void initApp();
 QWidget* topWidget (QWidget* widget);
 
+inline QPixmap scaledPixmap(const QString &fileName, const QSize &size, Qt::AspectRatioMode aspectRatioMode = Qt::KeepAspectRatio) {
```
adeebshihadeh, line 26: can we default the size to the image size?

selfdrive/ui/qt/spinner.cc
```diff
@@ -19,8 +19,8 @@ TrackWidget::TrackWidget(QWidget *parent) : QWidget(parent) {
   setFixedSize(spinner_size);
 
   // pre-compute all the track imgs. make this a gif instead?
-  QPixmap comma_img = QPixmap("../assets/img_spinner_comma.png").scaled(spinner_size, Qt::KeepAspectRatio, Qt::SmoothTransformation);
-  QPixmap track_img = QPixmap("../assets/img_spinner_track.png").scaled(spinner_size, Qt::KeepAspectRatio, Qt::SmoothTransformation);
+  QPixmap comma_img = scaledPixmap("../assets/img_spinner_comma.png", spinner_size);
```
adeebshihadeh, line 22: this would be for its own PR, but what do you think of putting these utils in their own namespace? it's unclear where this scaledPixmap comes from right now.

adeebshihadeh:
LGTM, any reason it's still a draft?

PR #23104, opened 2021-12-02
keyboard: replace backtick with apostrophe

MyNameIsCosmo:
Without `'`, I cannot type my WiFi password... though it might be time
to rotate it. This replaces a repeated period (`.`) with an apostrophe.

**Description** 
Apostrophe is missing from the layers of the qt keyboard. This was noticed when trying to type in a WiFi password.

**Verification**
No testing has been done on this patch.
Can test once the sim docker image is built, and I can build the source locally.

### DIFF
--- a/selfdrive/ui/qt/widgets/keyboard.cc
+++ b/selfdrive/ui/qt/widgets/keyboard.cc
@@ -126,7 +126,7 @@ Keyboard::Keyboard(QWidget *parent) : QFrame(parent) {
   std::vector<QVector<QString>> specials = {
     {"[","]","{","}","#","%","^","*","+","="},
     {"_","\\","|","~","<",">","€","£","¥","•"},
-    {"123",".",",","?","!","`",BACKSPACE_KEY},
+    {"123","'",",","?","!","`",BACKSPACE_KEY},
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/widgets/keyboard.cc
```diff
@@ -126,7 +126,7 @@ Keyboard::Keyboard(QWidget *parent) : QFrame(parent) {
   std::vector<QVector<QString>> specials = {
     {"[","]","{","}","#","%","^","*","+","="},
     {"_","\\","|","~","<",">","€","£","¥","•"},
-    {"123",".",",","?","!","`",BACKSPACE_KEY},
+    {"123","'",",","?","!","`",BACKSPACE_KEY},
```
adeebshihadeh, line 129: this is supposed to be the same as the iphone keyboard, but it seems the apostrophe was mistaken for a backtick
```suggestion
    {"123",".",",","?","!","'",BACKSPACE_KEY},
```

MyNameIsCosmo: Thanks! Applied.
Apologies for the extra spam, I've grown used to web pages that don't dynamically reload.

PR #23106, opened 2021-12-02
2019 Lexus NX FW versions

flightguy2:
Adding fingerprint only for 2019 Lexus NX.  Currently, only 2018 & 2020 NX exists.


selfdrive/car/toyota/values.py
```diff
@@ -55,6 +55,7 @@ class CAR:
   LEXUS_ESH_TSS2 = "LEXUS ES HYBRID 2019"
   LEXUS_IS = "LEXUS IS 2018"
   LEXUS_NX = "LEXUS NX 2018"
+  LEXUS_NX = "LEXUS NX 2019"
```
vanillagorillaa, line 58: This isn't needed and should be removed. The year doesn't matter as long as the platform is the same

vanillagorillaa:
Please provide the dongleID for the device this was taken from

selfdrive/car/toyota/values.py
adeebshihadeh:
```suggestion
```

flightguy2:
Dongle ID:  f61143b2bcdedf19

Regards,

Steve

flightguy2:
> 

Original route used to fingerprint was from several weeks ago, but I believe one of the more recent route segments should demonstrate the same fingerprint values.  I hope this works...

Segment:  f61143b2bcdedf19|2021-12-02--13-41-54

PR #23108, opened 2021-12-02
Move over all tss2 long tunes

haraschax:

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -244,11 +244,7 @@ class CarInterface(CarInterfaceBase):
 
     if ret.enableGasInterceptor:
       set_long_tune(ret.longitudinalTuning, LongTunes.PEDAL)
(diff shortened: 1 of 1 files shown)


selfdrive/car/toyota/interface.py
```diff
@@ -269,7 +269,9 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
     if ret.enableGasInterceptor:
       set_long_tune(ret.longitudinalTuning, LongTunes.PEDAL)
     elif candidate in [CAR.COROLLA_TSS2, CAR.COROLLAH_TSS2, CAR.RAV4_TSS2, CAR.RAV4H_TSS2, CAR.LEXUS_NX_TSS2,
```
pd0wm, line 271:
```suggestion
    elif candidate in TSS2_CAR:,
```

ErichMoraga: If doing this, then consider...
https://discord.com/channels/469524606043160576/524327905937850394/919675894568734750

PR #23109, opened 2021-12-02
soundd: scale volume using ambient noise

### LINKED ISSUE #23084: soundd: scale volume with ambient noise instead of speed

adeebshihadeh:
resolves #23084


adeebshihadeh:
Seems this will require an AGNOS update.

PR #23113, opened 2021-12-03
comma two: remove lens sag compensation

pd0wm:
This seems strictly better on the device I tested on. Need to do some more analysis on fleet data before shipping this, so holding off until 0.8.13


pd0wm:
image

PR #23116, opened 2021-12-03
Python: Replace more lists with generators

ghost:
I also replaced one np.sum with sum which seems to be faster for the usecase according to https://stackoverflow.com/questions/10922231/pythons-sum-vs-numpys-numpy-sum

### DIFF
--- a/scripts/code_stats.py
+++ b/scripts/code_stats.py
@@ -4,7 +4,7 @@ import ast
 import stat
 import subprocess
 
-fouts = set([x.decode('utf-8') for x in subprocess.check_output(['git', 'ls-files']).strip().split()])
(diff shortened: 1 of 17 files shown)


scripts/code_stats.py
```diff
@@ -4,7 +4,7 @@
 import stat
 import subprocess
 
-fouts = set([x.decode('utf-8') for x in subprocess.check_output(['git', 'ls-files']).strip().split()])
+fouts = set(x.decode('utf-8') for x in subprocess.check_output(['git', 'ls-files']).strip().split())
```
adeebshihadeh, line 7:
```suggestion
fouts = {x.decode('utf-8') for x in subprocess.check_output(['git', 'ls-files']).strip().split()}
```

grekiki: Will fix others

PR #23122, opened 2021-12-04
Toyota: only use pedal to get out of standstill on supported cars

cydia2020:
Only use pedal to get the car out of standstill on cars with full speed DRCC (namely Prius, ~~RAV4 Hybrid, Kluger~~, CTH, ~~ESH~~, NX, and RX).

This eliminates the jerk introduced by the pedal at low speeds, it also helps simplify tuning. Tested on my Prius.


selfdrive/car/toyota/interface.py
```diff
@@ -268,7 +268,12 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
     ret.communityFeature = ret.enableGasInterceptor or ret.enableDsu or smartDsu
 
     if ret.enableGasInterceptor:
-      set_long_tune(ret.longitudinalTuning, LongTunes.PEDAL)
+      # these cars have full speed DRCC, use stock tune
+      if candidate in [CAR.PRIUS, CAR.RAV4H, CAR.HIGHLANDER, CAR.HIGHLANDERH, CAR.LEXUS_CTH, CAR.LEXUS_ESH,
+                       CAR.LEXUS_NX, CAR.LEXUS_NXH, CAR.LEXUS_RX, CAR.LEXUS_RXH]:
```
adeebshihadeh, line 273: put this list in values.py

selfdrive/car/toyota/interface.py
```diff
@@ -268,7 +268,12 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
     ret.communityFeature = ret.enableGasInterceptor or ret.enableDsu or smartDsu
 
     if ret.enableGasInterceptor:
```
adeebshihadeh, line 270: move the new condition in the body up here

selfdrive/car/toyota/carcontroller.py
```diff
@@ -38,7 +38,16 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
       # offset for creep and windbrake
       pedal_offset = interp(CS.out.vEgo, [0.0, 2.3, MIN_ACC_SPEED + PEDAL_TRANSITION], [-.4, 0.0, 0.2])
       pedal_command = PEDAL_SCALE * (actuators.accel + pedal_offset)
-      interceptor_gas_cmd = clip(pedal_command, 0., MAX_INTERCEPTOR_GAS)
+      # Prius, Kluger and most LSS Lexus have full speed ACC, but do not automatically resume
```
adeebshihadeh, line 41: let's use the US names for the models, Kluger -> Highlander

selfdrive/car/toyota/carcontroller.py
```diff
@@ -38,7 +38,16 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
...
+      # Send pedal briefly if car is in standstill and openpilot wants to accelerate to resume ACC
+      # Pedal command is only sent very briefly, so no need to differentiate between Kluger and other cars
+      if CS.CP.carFingerprint in [CAR.PRIUS, CAR.RAV4H, CAR.HIGHLANDER, CAR.HIGHLANDERH, CAR.LEXUS_CTH, CAR.LEXUS_ESH, CAR.LEXUS_NX, CAR.LEXUS_NXH, CAR.LEXUS_RX, CAR.LEXUS_RXH]:
```
adeebshihadeh, line 44: use your new list in values.py for this

selfdrive/car/toyota/carcontroller.py
```diff
@@ -38,7 +38,16 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
...
+        if (actuators.accel > 0.1) and CS.pcm_acc_status == 7:
```
adeebshihadeh, line 45: this should use cruiseState.standstill. we should probably remove the condition in carState and always set it

adeebshihadeh (changes requested):
@HaraldSchafer do you think this is worth the extra complexity?

selfdrive/car/toyota/interface.py
cydia2020: Is the RAV4 hybrid a no stop timer car? Does the driver need to press the resume button for it to resume?

cydia2020: * asking because the RAV4H is mentioned in NO_STOP_TIMER_CAR, but it's also a pedal car in the new pedal code https://github.com/commaai/openpilot/blob/9ceb1a15eb87ba9ea5d7c8dbcc176f2466c82338/selfdrive/car/toyota/carcontroller.py#L32

HaraldSchafer:
, CAR.RAV4H, CAR.HIGHLANDER, CAR.HIGHLANDERH], those cars don't need pedals?

@adeebshihadeh Don't have any objections to this. I think it's good to use the pedal as little as possible.

cydia2020:
> , CAR.RAV4H, CAR.HIGHLANDER, CAR.HIGHLANDERH], those cars don't need pedals?
> 
> @adeebshihadeh Don't have any objections to this. I think it's good to use the pedal as little as possible.

That's one of the things that I am not sure of, standstill is not requested on those cars. So they should just "go" when op sends accel command

selfdrive/car/toyota/interface.py
```diff
@@ -267,7 +267,9 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
     # intercepting the DSU is a community feature since it requires unofficial hardware
     ret.communityFeature = ret.enableGasInterceptor or ret.enableDsu or smartDsu
 
-    if ret.enableGasInterceptor:
+    if ret.enableGasInterceptor and candidate in NO_AUTO_RESUME_CAR:
+      set_long_tune(ret.longitudinalTuning, LongTunes.TSS)
```
adeebshihadeh, line 271: why was this added?

cydia2020: I didn't quite get your "move the new condition in the body up here", so I changed it this way

adeebshihadeh: just delete this new branch

VirtuallyChris:
At least for the Rav4H and the HighlanderH they both do stop and go without hitting resume so exclude them from this list. Not 100% sure about the regular Highlander.

cydia2020:
> At least for the Rav4H and the HighlanderH they both do stop and go without hitting resume so exclude them from this list. Not 100% sure about the regular Highlander.

Highlander is also marked as a no stop timer car, so I won't add it to the no auto resume car list

selfdrive/car/toyota/interface.py
cydia2020: done

selfdrive/car/toyota/carcontroller.py
```diff
@@ -38,7 +38,15 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
       # offset for creep and windbrake
       pedal_offset = interp(CS.out.vEgo, [0.0, 2.3, MIN_ACC_SPEED + PEDAL_TRANSITION], [-.4, 0.0, 0.2])
       pedal_command = PEDAL_SCALE * (actuators.accel + pedal_offset)
-      interceptor_gas_cmd = clip(pedal_command, 0., MAX_INTERCEPTOR_GAS)
+      # Prius, and most LSS Lexus have full speed DRCC, but do not automatically resume
+      # Send pedal briefly if car is in standstill and openpilot wants to accelerate to resume ACC
```
adeebshihadeh, line 42:
```suggestion
      # Prius, and most Lexus, have full speed cruise, but require a button or pedal press to resume from standstill
      # For these cars, only use the pedal to resume from the standstill cruise state
```

selfdrive/car/toyota/values.py
```diff
@@ -1632,5 +1632,8 @@ class CAR:
 
 NO_DSU_CAR = TSS2_CAR | set([CAR.CHR, CAR.CHRH, CAR.CAMRY, CAR.CAMRYH])
 
+# full speed radar cruise
```
adeebshihadeh, line 1635:
```suggestion
# these cars have full speed ACC, but require a resume button or gas pedal press to resume from a standstill
```

selfdrive/car/toyota/values.py
```diff
@@ -1632,5 +1632,8 @@ class CAR:
 
 NO_DSU_CAR = TSS2_CAR | set([CAR.CHR, CAR.CHRH, CAR.CAMRY, CAR.CAMRYH])
 
+# these cars have full speed ACC, but require a resume button or gas pedal press to resume from a standstill
+NO_AUTO_RESUME_CAR = set([CAR.PRIUS, CAR.LEXUS_CTH, CAR.LEXUS_NX, CAR.LEXUS_NXH, CAR.LEXUS_RX, CAR.LEXUS_RXH])
```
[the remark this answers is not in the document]
cydia2020: I'm only 100% confident about the TSS-P Prius since I've been running this code for a while now. The other cars in this list are extrapolated from `stop_and_go = True` in `interface.py` and not `NO_STOP_TIMER_CAR` in `values.py`.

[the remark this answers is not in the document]
adeebshihadeh: let's just stick to the prius for now then

adeebshihadeh, line 1636:
```suggestion
NO_AUTO_RESUME_CAR = set([CAR.PRIUS])
```

selfdrive/car/toyota/carcontroller.py
```diff
@@ -29,7 +29,7 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
     if CS.CP.enableGasInterceptor and enabled:
       MAX_INTERCEPTOR_GAS = 0.5
       # RAV4 has very sensitive gas pedal
-      if CS.CP.carFingerprint in [CAR.RAV4, CAR.RAV4H, CAR.HIGHLANDER, CAR.HIGHLANDERH]:
+      if CS.CP.carFingerprint in [CAR.RAV4]:
```
adeebshihadeh, line 32: why was this changed?

cydia2020: All these cars except the RAV4 are marked as `NO_STOP_TIMER_CAR`, so standstill isn't requested when the car stops, thus a pedal isn't needed to get them moving again if my understanding is correct.

cydia2020:
```
# no resume button press required
NO_STOP_TIMER_CAR = TSS2_CAR | set([CAR.RAV4H, CAR.HIGHLANDERH, CAR.HIGHLANDER, CAR.SIENNA, CAR.LEXUS_ESH])
```

selfdrive/car/toyota/carstate.py
```diff
@@ -100,12 +100,7 @@ def update(self, cp, cp_cam):
       self.low_speed_lockout = cp.vl["PCM_CRUISE_2"]["LOW_SPEED_LOCKOUT"] == 2
 
     self.pcm_acc_status = cp.vl["PCM_CRUISE"]["CRUISE_STATE"]
-    if self.CP.carFingerprint in NO_STOP_TIMER_CAR or self.CP.enableGasInterceptor:
-      # ignore standstill in hybrid vehicles, since pcm allows to restart without
-      # receiving any special command. Also if interceptor is detected
-      ret.cruiseState.standstill = False
-    else:
-      ret.cruiseState.standstill = self.pcm_acc_status == 7
+    ret.cruiseState.standstill = self.pcm_acc_status == 7
```
cydia2020, line 103: This seems to cause problems, I'll look into it a bit more.

cydia2020: adding this block back.

selfdrive/car/toyota/carcontroller.py
cydia2020: reverted this per #23991

selfdrive/car/toyota/values.py
```diff
@@ -1843,6 +1843,9 @@ class ToyotaCarInfo(CarInfo):
 
 NO_DSU_CAR = TSS2_CAR | {CAR.CHR, CAR.CHRH, CAR.CAMRY, CAR.CAMRYH}
 
+# these cars have full speed ACC, but require a resume button or gas pedal press to resume from a standstill
+NO_AUTO_RESUME_CAR = set([CAR.PRIUS])
```
sshane, line 1847:
```suggestion
NO_AUTO_RESUME_CAR = {CAR.PRIUS}
```

selfdrive/car/toyota/interface.py
```diff
@@ -235,7 +235,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[], disabl
     # to a negative value, so it won't matter.
     ret.minEnableSpeed = -1. if (stop_and_go or ret.enableGasInterceptor) else MIN_ACC_SPEED
 
-    if ret.enableGasInterceptor:
+    if ret.enableGasInterceptor and candidate not in NO_AUTO_RESUME_CAR:
```
sshane, line 238: Probably still want this car param to describe if there's a pedal or not. How can you know it has an interceptor in carcontroller?

cydia2020:
```python
    if self.CP.enableGasInterceptor and CC.longActive:
      MAX_INTERCEPTOR_GAS = 0.5
      # RAV4 has very sensitive gas pedal
      if self.CP.carFingerprint in (CAR.RAV4, CAR.RAV4H, CAR.HIGHLANDER, CAR.HIGHLANDERH):
        PEDAL_SCALE = interp(CS.out.vEgo, [0.0, MIN_ACC_SPEED, MIN_ACC_SPEED + PEDAL_TRANSITION], [0.15, 0.3, 0.0])
      elif self.CP.carFingerprint in (CAR.COROLLA,):
        PEDAL_SCALE = interp(CS.out.vEgo, [0.0, MIN_ACC_SPEED, MIN_ACC_SPEED + PEDAL_TRANSITION], [0.3, 0.4, 0.0])
      else:
        PEDAL_SCALE = interp(CS.out.vEgo, [0.0, MIN_ACC_SPEED, MIN_ACC_SPEED + PEDAL_TRANSITION], [0.4, 0.5, 0.0])
      # offset for creep and windbrake
      pedal_offset = interp(CS.out.vEgo, [0.0, 2.3, MIN_ACC_SPEED + PEDAL_TRANSITION], [-.4, 0.0, 0.2])
      pedal_command = PEDAL_SCALE * (actuators.accel + pedal_offset)
      # Some cars, like the Prius, have full speed cruise, but require a button or pedal press to resume from standstill
      # For these cars, only use the pedal to resume from the standstill cruise state
      if CS.CP.carFingerprint in NO_AUTO_RESUME_CAR:
        if (actuators.accel > 0.1) and CS.cruiseState.standstill:
          interceptor_gas_cmd = 0.15
        else:
          interceptor_gas_cmd = 0
      else:
        interceptor_gas_cmd = clip(pedal_command, 0., MAX_INTERCEPTOR_GAS)
    else:
      interceptor_gas_cmd = 0.
    pcm_accel_cmd = clip(actuators.accel, CarControllerParams.ACCEL_MIN, CarControllerParams.ACCEL_MAX)
```

It's wrapped in the interceptor condition.

sshane: Whoops, got this confused with setting the car params

cydia2020:
Think this is ready for merge, I've been running this for almost a year now and haven't experienced any hiccups.

sshane:
Can you provide a route with the changes? Want to quickly see if you overshoot at all, and how much

cferra:
This also occurs on my CH-R - have to press to resume, can this car be added to the list for this? Or can it be addressed an another way?

cydia2020:
> This also occurs on my CH-R - have to press to resume, can this car be added to the list for this? Or can it be addressed an another way?

If it's tss-p, then openpilot doesn't control longitudinal, so that'd be a no, there are forks out there that does, but it requires another piece of hardware.

sshane:
@cydia2020 I wonder, what if you just adjust MIN_ACC_SPEED to be 1 mph for your car?

adeebshihadeh:
Thanks for the PR, but this is too much code and discussion than we can afford for niche things like the pedal.

PR #23123, opened 2021-12-04
Add fingerprint for 2021 Toyoto Corolla Cross Non-Hybrid (Thailand version)

khoi:
- Add missing fingerprints
- This car does not support Stop & Go
- No ALC and ACC below 18 miles/h (30km/h)

Dongle ID: ec07911a91f863f3


adeebshihadeh:
Which route did these FW versions come from? I checked a couple routes from that dongle ID, but there were no FW versions in them.

PR #23133, opened 2021-12-05
UI: move rest of NVG to Qt

deanlee:


selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -6,11 +6,7 @@
 namespace {
 
 const char frame_vertex_shader[] =
-#ifdef NANOVG_GL3_IMPLEMENTATION
```
adeebshihadeh, line 9: I think we still need this for mac, let's just give it a better name and add a comment to confirm that it's actually needed

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -21,11 +17,7 @@ const char frame_vertex_shader[] =
   "}\n";
 
 const char frame_fragment_shader[] =
-#ifdef NANOVG_GL3_IMPLEMENTATION
```
adeebshihadeh, line 24: same here

SConstruct
```diff
@@ -90,7 +90,6 @@ if arch == "aarch64" or arch == "larch64":
     "/usr/lib",
     "/system/vendor/lib64",
     "/system/comma/usr/lib",
-    "#third_party/nanovg",
```
adeebshihadeh, line 93: let's keep this stuff for now to reduce the diff for this PR. you can make a follow up to cleanup all the references and delete nanovg from phonelibs

selfdrive/ui/qt/onroad.cc
```diff
@@ -256,14 +255,110 @@ void NvgWindow::initializeGL() {
   qInfo() << "OpenGL renderer:" << QString((const char*)glGetString(GL_RENDERER));
   qInfo() << "OpenGL language version:" << QString((const char*)glGetString(GL_SHADING_LANGUAGE_VERSION));
 
-  ui_nvg_init(&QUIState::ui_state);
   prev_draw_t = millis_since_boot();
   setBackgroundColor(bg_colors[STATUS_DISENGAGED]);
 }
 
+void NvgWindow::updateFrameMat(int w, int h) {
```
adeebshihadeh, line 262: I was thinking we get rid of the NvgWindow and make the path/lane lines their own widget. Then the stack can be: CameraViewWidget -> lane lines / path -> HUD -> alerts.

deanlee: I have tried,  `drawLaneLines` takes 2-4ms to draw in it's own widget, but only takes 0.2-0.4ms  in NvgWindow(10x faster). I suggest to still put it in NvgWindow for now. it will be easy to move out to it's own widget after the performance is not an issue.

adeebshihadeh: why is it slower?

deanlee: It should be caused by the performance of anti-aliasing. there is no  MSAA in normal widget.

deanlee: Nanovg use MSAA or a shader (if flag NVG_ANTIALIAS is set) to do anti-aliasing.

adeebshihadeh: ok, let's merge this for now, but I'd love to separate these out at some point

PR #23150, opened 2021-12-07
Toyota: Display Blue Barriers When Engaged, Matches OEM Design

krkeegan:
**Description** The Toyota OEM system displays the blue barriers whenever it is contributing to steering.  So when the LKAS system is engaged it looks like this:

PXL_20211021_163532657-3

Whereas currently when OP is engaged the blue barriers are not present.

This PR changes the `barriers` value based on the engaged state of OP in the LKAS_HUD can message. So that OP matches the OEM system.

*Note* 
The original code linked the blue barrier with lane departure warnings which doesn't match the OEM system.  In the OEM system a single blue barrier is only used with lane departure + steering assist, in other words when the car is contributing to steering.  If you depart the lane at an angle steeper than what steering assist will work on, then no blue barrier is displayed only the orange lane marker.

Since the OP lane departure warnings do not include steering assist, a single blue barrier does not need to be displayed ever.  The Orange lane markers are still displayed as part of the lane departure warnings.

**Verification** I used the OEM system with OP disabled and monitored how the HUD display changed when the LKAS system was engaged and not.  I then ran this commit on my device with OP engaged and verified that it displayed the same HUD contents when OP was engaged.

**Route**
Route:  d50727925674b294|2021-12-06--16-38-13

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -120,7 +120,7 @@ class CarController():
       send_ui = True
 
     if (frame % 100 == 0 or send_ui):
-      can_sends.append(create_ui_command(self.packer, steer_alert, pcm_cancel_cmd, left_line, right_line, left_lane_depart, right_lane_depart))
+      can_sends.append(create_ui_command(self.packer, steer_alert, pcm_cancel_cmd, left_line, right_line, left_lane_depart, right_lane_depart, enabled))
 
     if frame % 100 == 0 and CS.CP.enableDsu:
       can_sends.append(create_fcw_command(self.packer, fcw_alert))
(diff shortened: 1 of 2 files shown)


pd0wm:
Is this a TSS2 thing when lane trace is active? What does the TSS1 Prius show when sending 1 to the barriers field? Does it even have support for blue barriers?

krkeegan:
Both barriers are only displayed on the OEM system when Lane Tracing Assist is active, which is only available on TSS2.  On TSS1 vehicles the blue barriers are present but only one is displayed at a time, and only when the Steering Assist is active for Lane Departure.

I do not have a TSS1 vehicle, but I believe from all of the documentation and notes in discord that 1 will display both barriers on TSS1 vehicles as well.

illumiN8i:
Here's a picture of what this change looks like on TSS-P Prius Prime.
blue barriers

PR #23156, opened 2021-12-07
locationd_thread: fix possible dangling reference to params

deanlee:


selfdrive/locationd/locationd.cc
```diff
@@ -523,8 +521,8 @@ int Localizer::locationd_thread() {
         std::string lastGPSPosJSON = util::string_format(
           "{\"latitude\": %.15f, \"longitude\": %.15f, \"altitude\": %.15f}", posGeo(0), posGeo(1), posGeo(2));
 
-        std::thread([&params] (const std::string gpsjson) {
-          params.put("LastGPSPosition", gpsjson);
```
deanlee, line 527: 'params' might be invalidated here

PR #23158, opened 2021-12-07
Navigation speed limit

mayfieldiv:
Cereal: https://github.com/commaai/cereal/pull/230

### DIFF
--- a/selfdrive/ui/navd/route_parser.cc
+++ b/selfdrive/ui/navd/route_parser.cc
@@ -1,396 +1,390 @@
 #include "selfdrive/ui/navd/route_parser.h"
 
-#include <math.h>
 #include <QDebug>
 #include <QGeoPath>
+#include <QJsonArray>
 #include <QJsonDocument>
 #include <QJsonObject>
-#include <QJsonArray>
 #include <QLocale>
 #include <QUrlQuery>
+#include <math.h>
 
 #include "selfdrive/common/params.h"
 #include "selfdrive/ui/qt/maps/map_helpers.h"
 
-static QList<QGeoCoordinate> decodePolyline(const QString &polylineString)
-{
-    QList<QGeoCoordinate> path;
-    if (polylineString.isEmpty())
-        return path;
-
-    QByteArray data = polylineString.toLatin1();
-
-    bool parsingLatitude = true;
-
-    int shift = 0;
-    int value = 0;
-
-    QGeoCoordinate coord(0, 0);
-
-    for (int i = 0; i < data.length(); ++i) {
-        unsigned char c = data.at(i) - 63;
-
-        value |= (c & 0x1f) << shift;
-        shift += 5;
-
-        // another chunk
-        if (c & 0x20)
-            continue;
-
-        int diff = (value & 1) ? ~(value >> 1) : (value >> 1);
-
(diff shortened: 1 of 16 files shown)


selfdrive/ui/navd/route_parser.cc
```diff
@@ -0,0 +1,1228 @@
...
+        if (wayName.isEmpty())
+            return laneInstruction + MapboxRouteParser::tr(" and make a U-turn");
+        else
+            return laneInstruction + MapboxRouteParser::tr(" and make a U-turn onto %1").arg(wayName);
+    default:
+        return laneInstruction;
+    }
+}
+
+static QString instructionText(const QJsonObject &step, const QJsonObject &maneuver, QGeoManeuver::InstructionDirection direction)
```
pd0wm, line 722: You're aware we don't use the instructionText anywhere? We only use the banner objects from the response to decide what to show to the user.

mayfieldiv: gotcha, thanks

selfdrive/ui/navd/route_parser.cc
```diff
@@ -0,0 +1,300 @@
...
+#include <QJsonDocument>
+#include <QJsonObject>
+#include <QLocale>
+#include <QUrlQuery>
+#include <math.h>
+
+#include "selfdrive/common/params.h"
+#include "selfdrive/ui/qt/maps/map_helpers.h"
+
+// TODO: this seems to create duplicate coordinates in the path when decoding the polyline, so the corresponding annotations don't match up by index
```
pd0wm, line 15: You can also use `geometries=geojson` in the route request to just get a list of coordinates. Then you don't have to write hacks to match the annotations by distance instead of index.

selfdrive/ui/navd/route_parser.cc
```diff
@@ -0,0 +1,300 @@
...
+
+      auto distance = routeObject.value("distance").toDouble();
+      auto travel_time = routeObject.value("duration").toDouble();
+      auto error = false;
+      QList<RouteSegment> segments;
+      QList<RouteAnnotation> annotations;
+
+      auto legs = routeObject.value("legs").toArray();
+      Route route;
+      for (int leg_index = 0; leg_index < legs.size(); ++leg_index) {
```
pd0wm, line 184:
```suggestion
      for (auto const &l: routeObject.value("legs").toArray()) {
```

selfdrive/ui/navd/route_parser.cc
mayfieldiv: good idea

mayfieldiv:
Turns out the route path coordinate duplication was caused by the way the route path was constructed - concatenating the paths of the segments. Since the segment paths share a border, these border coordinates were duplicated. I changed the parser to parse the geometry of the whole route, since it's available in the response

zorrobyte:
Not sure where mapbox gets their speed data, maybe OSM?

An easy way to update their database would be to use the Toyota stock camera sign reader to log speeds for contribution as they are often wildly incorrect. Of course, doing so via video/ML model would also work. Just throwing it out there.

adeebshihadeh:
We decided to re-implement this in python.

PR #23164, opened 2021-12-08
onroad test fixes

adeebshihadeh:


selfdrive/test/test_onroad.py
```diff
@@ -26,7 +26,7 @@
   "./loggerd": 45.0,
   "./locationd": 9.1,
   "selfdrive.controls.plannerd": 22.6,
-  "./_ui": 15.0,
```
adeebshihadeh, line 29: @deanlee any idea why this went up after the recent refactors?

deanlee: imo, the new OnroadHud window Increased the workload of window compositing. 
we may need to consider using plugin(https://doc.qt.io/qt-5/embedded-linux.html) or QGraphicsScene(with opengl viewport) to reduce the cpu usage if necessary.

deanlee:
@adeebshihadeh qt constantly re-upload the raster memory buffers (onroadHud & OnroadAlerts painted on raster memory) into OpenGL textures for compositing, this can have a significant impact on performance.

deanlee:
A possible quick solution is to keep the existing structures and call hud and alerts's  paint functions in NvgWindow::paintGL

PR #23168, opened 2021-12-08
Use std::unique_ptr to manage memory

deanlee:

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -208,14 +208,14 @@ void can_send_thread(std::vector<Panda *> pandas, bool fake_send) {
   set_thread_name("boardd_can_send");
 
   AlignedBuffer aligned_buf;
-  Context * context = Context::create();
-  SubSocket * subscriber = SubSocket::create(context, "sendcan");
+  std::unique_ptr<Context> context(Context::create());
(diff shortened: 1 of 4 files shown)


selfdrive/boardd/boardd.cc
```diff
@@ -208,38 +208,31 @@ void can_send_thread(std::vector<Panda *> pandas, bool fake_send) {
...
+      for (const auto& panda : pandas) {
+        panda->can_send(event.getSendcan());
       }
     }
-
-    delete msg;
   }
-
-  delete subscriber;
-  delete context;
```
pd0wm, line 242: For ZMQ it's important to delete all subscribers before terminating the context, or that might hang forever. It seems like unique pointers are destructed in reverse order of their creation, so that should work out here. Is there any way we can make this relation more explicit?

deanlee: we can use braces to make it "more explicit",  But I think it might be redundant since objects are always destructed in the reverse order of construction (LIFO). 

```
unique_ptr ctx
{
  unique_ptr msg
}
```

pd0wm: Guess it's fine since it's impossible to create the subscriber before the context.

PR #23171, opened 2021-12-08
add fingerprint for European 2018 Toyota C-HR Hybrid

Metalbird1997:
**Description** 
FP2.0 was missing for the (European) 2018 Toyota C-HR Hybrid. The current release3 showed Dashcam Mode, Unsupported Car message. The only thing different was the engine FP.

**Verification** 
Implemented FP and went for a test drive. Now working as expected.

**Route**
Route: Please contact me on Discord `Metalbird#0001` if you actually need to see a route for this. I'll gladly send you a link.


adeebshihadeh:
We'll need a route for this. Can you post it here?

PR #23178, opened 2021-12-09
cameras: remove function get_frame_image

deanlee:
Looks like it's not needed anymore


adeebshihadeh:
Unfortunately, it's still used for some internal tools.

PR #23186, opened 2021-12-10
alert text simplification

adeebshihadeh:


selfdrive/controls/lib/events.py
```diff
@@ -579,7 +586,7 @@ def joystick_alert(CP: car.CarParams, sm: messaging.SubMaster, metric: bool) ->
   },
 
   EventName.wrongGear: {
-    ET.SOFT_DISABLE: SoftDisableAlert("Gear not D"),
+    ET.SOFT_DISABLE: UserSoftDisableAlert("Gear not D"),
     ET.NO_ENTRY: NoEntryAlert("Gear not D"),
```
qadmus, line 590: I like how warnings give actions, like "touch wheel" and "take control".

Maybe NO_ENTRY should too, because action is required to engage: "Shift to Drive", "Close doors", "Fasten seatbelts"

qadmus: For soft disable, it's talking about past-tense "what happened", so those sound right.

Also I don't know the nitty gritty here so call out if I'm missing deets.

adeebshihadeh: I think it's fine as is since it's more a response to the engagement attempt. @alexcomma what do you think?

selfdrive/controls/lib/events.py
```diff
@@ -153,6 +153,13 @@ def __init__(self, alert_text_2):
                      AudibleAlert.warningSoft, 2.),
 
 
+# less harsh version of SoftDisable, where the condition is user-triggered
+class UserSoftDisableAlert(SoftDisableAlert):
+  def __init__(self, alert_text_2):
+    super().__init__(alert_text_2),
+    self.alert_text_1 = "openpilot disabling soon"
```
adeebshihadeh, line 160:
```suggestion
    self.alert_text_1 = "openpilot will disengage"
```

PR #23189, opened 2021-12-10
Remove calibrationd reset logic

haraschax:


HaraldSchafer:
May not be relevant with this fix
https://github.com/commaai/openpilot/pull/23211

PR #23200, opened 2021-12-12
LapConv: remove hardcoded  full_stride_x

deanlee:
the correct value of FULL_STRIDE_X should be rgb_width(1164), am I right ?


pd0wm:
I assume there is some alignment going on in the RGB buffer? `rgb_stride` seems to be set to 3840, which is the same as the current `FULL_STRIDE_X * 3`. Maybe it's better to use that than hard-coded values. @ZwX1616 is that correct?

deanlee:
We use frame_width*3 as rgb_stride by default in many other places, e.g. framereader. I'm not sure if this will cause potential problems.

pd0wm:
This is only relevant for comma two, but for correctness you should always check the VisionBuf for the stride.

PR #23201, opened 2021-12-12
webcam/replay_cam: setImage if env_send_road is set

deanlee:


adeebshihadeh:
No need for this, I'd like to entirely remove this soon. I just need to find some time to fix up some of our internal tools first.

PR #22751, opened 2021-10-30
Add C/C++ Doxygen docs

tecandrew:
adds C/C++ docs to sphinx-generated docs.comma.ai site.

rel: #22732

Issue #22732, opened 2021-10-29
Add C++ code to sphinx API docs


pd0wm:
https://devblogs.microsoft.com/cppblog/clear-functional-c-documentation-with-sphinx-breathe-doxygen-cmake/

tecandrew:
done doxygen docs in the past before, could I take a crack at this? not sure how to integrate it into a sphinx site though

![image](https://user-images.githubusercontent.com/21319730/139530533-23461cfa-fc2b-4d5e-963d-e1fefe5b9b6a.png)

pd0wm:
Feel free to open a wip PR once you have some kind of build process going!

PR #23204, opened 2021-12-13
c docs to docs.comma.ai

tecandrew:
# Overview 
redo of PR #22751 
rel: #22732

## Features 

* add c docs to https://docs.comma.ai 
  * doxygen-based documentation using `sphinx` and `breathe`
* adds some SEO stuff
* make it look prettier

image

### Misc
* uses `cmake` instead of `Makefile`
  * initially used to auto-generate _all_ C code using traditional `doxygen/cmake` template, but `sphinx-breathe` doesn't like some code (selfdrive/ui/qt, selfdrive/ui/widgets)
    * now only documents explicitly listed files in `docs/sphinx/c_docs.rst.in` 
* regenerates `Pipfile.lock`, adding `breathe` module


pd0wm:
Nice!

Same comments as on the previous PR:
1. Can you convert the cmake build file into scons instead of introducing another build system? If you can't figure it out I can take a look as well.
2. Can we only check in the parts of the `Doxyfile.in` that are not the default settings?

tecandrew:
I'll take a stab at scons. Might take a while to convert since i've never dev'd with it before

> 2. Can we only check in the parts of the `Doxyfile.in` that are not the default settings?

Not sure what you meant by this. 😅

pd0wm:
You added a 2600 line that only contains the default settings. Does it also work if only add the options that we actually change?

tecandrew:
I should take that out: sphinx is doing the XML creation, not doxygen Doxyfile explicitly: it's generating docs based off found .h and .c pairs. docs/sphinx/conf.py.in

adeebshihadeh:
Can you split out all the listed features into their own PR?

tecandrew:
> Can you split out all the listed features into their own PR?

yeah, i could do that, but it'll be based off `cmake` in it's current state.

PR #23214, opened 2021-12-14
remove community features toggle

adeebshihadeh:


pd0wm:
Let's replace this by a startup alert if AEB is disabled.

PR #23215, opened 2021-12-14
C2: display ip address in network panel

deanlee:


pd0wm:
I agree it's quite a hassle to find the IP address on a comma two and showing it in the settings would help. I don't want to start managing networks settings from the ui. On the comma three we really didn't have any other choice, but it was way more work than anticipated.

For this PR only include showing the IP address.

PR #23217, opened 2021-12-14
put logcat in bootlog

deanlee:
Thousands of logs will be pushed every time logcatd is started. It may be better to put the system logs at startup in the bootlog

1. put system logs in bootlog
2. only log the latest messages to androidLog

### DIFF
--- a/selfdrive/loggerd/bootlog.cc
+++ b/selfdrive/loggerd/bootlog.cc
@@ -11,6 +11,8 @@ static kj::Array<capnp::word> build_boot_log() {
   if (Hardware::TICI()) {
     bootlog_commands.push_back("journalctl");
     bootlog_commands.push_back("sudo nvme smart-log --output-format=json /dev/nvme0");
+  } else if(Hardware::EON()) {
+    bootlog_commands.push_back("logcat");
   }
 
   MessageBuilder msg;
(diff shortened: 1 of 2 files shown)


selfdrive/loggerd/bootlog.cc
```diff
@@ -11,6 +11,8 @@ static kj::Array<capnp::word> build_boot_log() {
   if (Hardware::TICI()) {
     bootlog_commands.push_back("journalctl");
     bootlog_commands.push_back("sudo nvme smart-log --output-format=json /dev/nvme0");
+  } else if(Hardware::EON()) {
+    bootlog_commands.push_back("logcat");
```
adeebshihadeh, line 15: does this work? I would think you would need `logcat -d`

selfdrive/loggerd/bootlog.cc
```diff
@@ -11,6 +11,8 @@ static kj::Array<capnp::word> build_boot_log() {
   if (Hardware::TICI()) {
     bootlog_commands.push_back("journalctl");
     bootlog_commands.push_back("sudo nvme smart-log --output-format=json /dev/nvme0");
+  } else if(Hardware::EON()) {
```
adeebshihadeh, line 14:
```suggestion
  } else if (Hardware::EON()) {
```

adeebshihadeh:
Can you revert c671f8ca7160691791bade508c4a8eb5cdb73fb8 in the spirit of our new PR policy? This PR has a nicely stated goal, but that commit doesn't contribute to it. If you want to get that change in, make it a separate PR, however those are the kinds of things we're not really interested in anymore.

PR #23219, opened 2021-12-14
PI control integrator backcalculation

ClockeNessMnstr:
partial from https://github.com/commaai/openpilot/pull/21422
@adeebshihadeh 

Backfeeding the actual clipped / limited outputs into the controllers should help reduce complexity of managing "integral" behavior.

I've implemented the PI+FF portion. This makes the override and freeze functions redundant. I only have lat on my Subarus so  I'm interested to see if it's a noticeable improvement or change on long vehicles. 
"override" is/should be very smooth with these changes. 

For PI control this is essentially recalculating the integrator towards any clipping or limits applied to the output. 
this is an antiwindup for steermax, max_up/max_down, driver_allowance, or any other limit applied by the vehicle values. 

This should also be useful for the autotune-INDI since I believe that is using the output vs the response. so an accurate value for the actual request is needed.


adeebshihadeh:
@pd0wm what do you think of CarInterface.apply also returning another CarControl with the actual values used?

pd0wm:
Yes I like that. I indeed wanted that for auto tuning. Shall we return and log another version of the CarControl struct, but then with the values that were actually sent to the car?

jyoung8607 (commented):
It would be nice (whether before, after, or during this PR) to grow a `CarControllerBase` and harmonize the five different names different car ports use to import `CarControllerParams`, initialize things like `apply_steer_last`, maybe handle `CANPacker` setup, etc. Then we wouldn't need several near-copies of `get_last_output()`.

pd0wm:
Little bit annoying that GitHub forced this PR closed. Let me try to revive it.

PR #23220, opened 2021-12-14
util::check_output: bigger line buffer

deanlee:


pd0wm:
Any noticeable performance improvements?

adeebshihadeh:
For these kinds of PR, can you include a goal and post metrics to prove it's better?

deanlee:
closed. It is not obvious. it may save 5-10ms when executing logcatd.although it can save many calls to fgets and string alloc, most of the time is spent on executing commands.

PR #23225, opened 2021-12-15
cloudlog: add daemon name to context for python daemons

adeebshihadeh:


selfdrive/manager/process.py
```diff
@@ -33,6 +33,8 @@ def launcher(proc):
     # create new context since we forked
     messaging.context = messaging.Context()
 
+    cloudlog.bind(daemon=name)
```
adeebshihadeh, line 36:
```suggestion
    # add daemon name to cloudlog ctx
    cloudlog.bind(daemon=name)
```

PR #23226, opened 2021-12-15
swaglog: add test case. fixed the ctx of the first message is empty

deanlee:


deanlee:
fixed a bug
```
selfdrive/common/tests/test_swaglog.cc:51: FAILED:
  REQUIRE( ctx["dongle_id"].string_value() == dongle_id )
with expansion:
  "" == "test_dongle_id"
```

cloudlog_init should be called before log,  otherwise the ctx_j will be empty

deanlee:
Do we really need to insert the same ctx in every message? the same content has been already written in the log by `log_init_data`.

pd0wm:
> Do we really need to insert the same ctx in every message? the parameters in ctx have been already written in log by `log_init_data`.

Those are processed by two completely different code paths in the backend. The swaglogs are uploaded by athena and processed separately (and also logged while offroad).

PR #23227, opened 2021-12-15
RequestRepeater: stop timer while onroad

deanlee:


pd0wm:
Don't think this is worth debugging issues with quickly going onroad/offroad messing with the current timer state.

PR #23231, opened 2021-12-15
fixed the background issue of the dcamera preview on C2

deanlee:
the background need to be cleared before display dcam frame on C2.

**before:**

**after:**

### DIFF
--- a/selfdrive/ui/qt/widgets/cameraview.cc
+++ b/selfdrive/ui/qt/widgets/cameraview.cc
@@ -199,7 +199,7 @@ void CameraViewWidget::updateFrameMat(int w, int h) {
 }
 
 void CameraViewWidget::paintGL() {
-  if (latest_texture_id == -1 || stream_type == VISION_STREAM_RGB_FRONT) {
+  if (latest_texture_id == -1 || (stream_type == VISION_STREAM_RGB_FRONT && Hardware::EON())) {
     glClearColor(bg.redF(), bg.greenF(), bg.blueF(), bg.alphaF());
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -199,12 +199,13 @@ void CameraViewWidget::updateFrameMat(int w, int h) {
 }
 
 void CameraViewWidget::paintGL() {
-  if (latest_texture_id == -1) {
+  if (latest_texture_id == -1 || stream_type == VISION_STREAM_RGB_FRONT) {
```
adeebshihadeh, line 202: this isn't a real way to fix the issue. how expensive is it to always do this?

deanlee: It's very cheap and doesn't even make a difference

deanlee: We can only clear the background every frame, otherwise there will be problems on C2

deanlee: It should be caused by  by the difference between c2's GraphicBuffer/visonbuf and other devices

adeebshihadeh: I don't understand. is it not cheap to always glClearColor and glClear on all platforms for all cameras?

deanlee: it's same on all platforms. think we should always call glclear before drawing to avoid weird artifacts

PR #23235, opened 2021-12-15
modeld: update live calibration in main thread

### LINKED ISSUE #23234: model replay random failures
Should be easy to catch running a loop. @deanlee want to take a look at and see why this isn't fully deterministic? There's a chance it's a test bug too.

https://jenkins.comma.life/blue/organizations/jenkins/openpilot/detail/master/1233/pipeline

Logs attached from that failed jenkins run.
[log.txt](https://github.com/commaai/openpilot/files/7722250/log.txt)

deanlee:
possible resolve https://github.com/commaai/openpilot/issues/23234

update_calibration is fast enough, there should be no need to do it in a separate thread which will increase the complexity and   unnecessary overhead.

### DIFF
--- a/selfdrive/modeld/modeld.cc
+++ b/selfdrive/modeld/modeld.cc
@@ -22,13 +22,15 @@ mat3 update_calibration(cereal::LiveCalibrationData::Reader live_calib, bool wid
      medmodel_frame_from_ground = medmodel_frame_from_road_frame[:, (0, 1, 3)]
      ground_from_medmodel_frame = np.linalg.inv(medmodel_frame_from_ground)
   */
-  Eigen::Matrix<float, 3, 3> ground_from_medmodel_frame;
-  ground_from_medmodel_frame <<
-    0.00000000e+00, 0.00000000e+00, 1.00000000e+00,
-    -1.09890110e-03, 0.00000000e+00, 2.81318681e-01,
-    -1.84808520e-20, 9.00738606e-04,-4.28751576e-02;
+  static const Eigen::Matrix<float, 3, 3> ground_from_medmodel_frame = [] {
+    Eigen::Matrix<float, 3, 3> tmp;
+    tmp << 0.00000000e+00, 0.00000000e+00, 1.00000000e+00,
+          -1.09890110e-03, 0.00000000e+00, 2.81318681e-01,
+          -1.84808520e-20, 9.00738606e-04, -4.28751576e-02;
+    return tmp;
+  }();
 
-  Eigen::Matrix<float, 3, 3> cam_intrinsics = Eigen::Matrix<float, 3, 3, Eigen::RowMajor>(wide_camera ? ecam_intrinsic_matrix.v : fcam_intrinsic_matrix.v);
+  auto cam_intrinsics = Eigen::Matrix<float, 3, 3, Eigen::RowMajor>(wide_camera ? ecam_intrinsic_matrix.v : fcam_intrinsic_matrix.v);
   const mat3 yuv_transform = get_model_yuv_transform();
 
   auto extrinsic_matrix = live_calib.getExtrinsicMatrix();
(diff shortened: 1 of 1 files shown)


selfdrive/modeld/modeld.cc
```diff
@@ -80,20 +65,22 @@ void run_model(ModelState &model, VisionIpcClient &vipc_client) {
...
-    bool valid = live_calib_seen;
-    transform_lock.unlock();
-
     // TODO: path planner timeout?
     sm.update(0);
     int desire = ((int)sm["lateralPlan"].getLateralPlan().getDesire());
     frame_id = sm["roadCameraState"].getRoadCameraState().getFrameId();
+    if (sm.updated("liveCalibration")) {
+      model_transform = update_calibration(sm["liveCalibration"].getLiveCalibration(), wide_camera);
+      live_calib_seen = true;
```
adeebshihadeh, line 82: use sm.rcv_frame instead?

adeebshihadeh:
Can you test this scenario: send liveCalibration at something >20Hz and run modeld for several minutes? Does this affect the timings at all? Can you post stats for how long updating the calibration matrix takes?

deanlee:
The benchmarking outputs after running test_modeld:


```
benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
update_calibration                             100           306     2.1726 ms 
                                        65.2796 ns     65.215 ns    65.5357 ns 
                                       0.601714 ns  0.0752819 ns    1.42598 ns 
```

selfdrive/modeld/modeld.cc
deanlee: We still need a variable to save the result of sm.rcv_frame and pass it to 2 functions.the current implementation may be more efficient which can reduce a map lookup in sm.rcv_frame

selfdrive/modeld/models/driving.cc
```diff
@@ -343,3 +343,41 @@ void posenet_publish(PubMaster &pm, uint32_t vipc_frame_id, uint32_t vipc_droppe
...
 }
+
+mat3 update_calibration(cereal::LiveCalibrationData::Reader live_calib, bool wide_camera) {
+  /*
+     import numpy as np
+     from common.transformations.model import medmodel_frame_from_road_frame
+     medmodel_frame_from_ground = medmodel_frame_from_road_frame[:, (0, 1, 3)]
+     ground_from_medmodel_frame = np.linalg.inv(medmodel_frame_from_ground)
+  */
+  static const Eigen::Matrix<float, 3, 3> ground_from_medmodel_frame = [] {
```
adeebshihadeh, line 354: I think this was cleaner and more readable the way it was originally

selfdrive/modeld/tests/test_modeld.cc
```diff
@@ -0,0 +1,27 @@
+#include <random>
```
adeebshihadeh, line 1: no need to include the test. I just wanted to verify this initially

deanlee:
removed test

selfdrive/modeld/modeld.cc
```diff
@@ -14,64 +14,49 @@
...
      from common.transformations.model import medmodel_frame_from_road_frame
      medmodel_frame_from_ground = medmodel_frame_from_road_frame[:, (0, 1, 3)]
      ground_from_medmodel_frame = np.linalg.inv(medmodel_frame_from_ground)
   */
-  Eigen::Matrix<float, 3, 3> ground_from_medmodel_frame;
-  ground_from_medmodel_frame <<
-    0.00000000e+00, 0.00000000e+00, 1.00000000e+00,
-    -1.09890110e-03, 0.00000000e+00, 2.81318681e-01,
-    -1.84808520e-20, 9.00738606e-04,-4.28751576e-02;
+  static const Eigen::Matrix<float, 3, 3> ground_from_medmodel_frame = [] {
```
adeebshihadeh, line 25: I also meant to remove the lambda

deanlee: remove static and init it every time?

PR #23241, opened 2021-12-16
Convert format strings strings to f-strings

EwoutH:
Reformats all the text from the old "%-formatted" and .format(...) format to the newer f-string format, as defined in [PEP 498](https://www.python.org/dev/peps/pep-0498/). This requires Python 3.6+.

[Flynt](https://github.com/ikamensh/flynt) 0.69 was used to reformat the strings. 120 f-strings were created in 51 files.

F-strings are in general more readable, concise and performant. See also: [PEP 498#Rationale](https://www.python.org/dev/peps/pep-0498/#rationale)

```
Running flynt v.0.69

Flynt run has finished. Stats:

Execution time:                            3.342s
Files modified:                            51
Character count reduction:                 778 (0.04%)

Per expression type:
Old style (`%`) expressions attempted:     84/142 (59.2%)
`.format(...)` calls attempted:            53/53 (100.0%)
No concatenations attempted.
F-string expressions created:              120
```


deanlee:
f-string may not be multi-language friendly, which may be a problem.

PR #23244, opened 2021-12-16
PID: Use last actuators to limit integrator

pd0wm:
Follow up https://github.com/commaai/openpilot/pull/23219. Code from @ClockeNessMnstr

### DIFF
--- a/selfdrive/controls/lib/pid.py
+++ b/selfdrive/controls/lib/pid.py
@@ -1,4 +1,3 @@
-import numpy as np
 from numbers import Number
 
 from common.numpy_fast import clip, interp
@@ -26,7 +25,7 @@ class PIController():
     self.neg_limit = neg_limit
 
     self.sat_count_rate = 1.0 / rate
-    self.i_unwind_rate = 0.3 / rate
+    self.i_unwind_rate = 1.0 / rate
     self.i_rate = 1.0 / rate
     self.sat_limit = sat_limit
 
@@ -60,27 +59,20 @@ class PIController():
     self.saturated = False
     self.control = 0
 
-  def update(self, setpoint, measurement, speed=0.0, check_saturation=True, override=False, feedforward=0., deadzone=0., freeze_integrator=False):
+  def update(self, setpoint, measurement, last_output, speed=0.0, check_saturation=True, feedforward=0., deadzone=0.):
     self.speed = speed
+
+    i_bf = self.i_unwind_rate * (self.p + self.i + self.f - last_output)
 
     error = float(apply_deadzone(setpoint - measurement, deadzone))
     self.p = error * self.k_p
+    self.i = self.i + error * self.k_i * self.i_rate - i_bf
     self.f = feedforward * self.k_f
 
-    if override:
-      self.i -= self.i_unwind_rate * float(np.sign(self.i))
-    else:
-      i = self.i + error * self.k_i * self.i_rate
-      control = self.p + self.f + i
+    # ensure PI controller action is not clipped when ff is large
+    self.f = clip(self.f, self.neg_limit, self.pos_limit)
 
-      # Update when changing i will move the control away from the limits
-      # or when i will move towards the sign of the error
-      if ((error >= 0 and (control <= self.pos_limit or i < 0.0)) or
-          (error <= 0 and (control >= self.neg_limit or i > 0.0))) and \
-         not freeze_integrator:
-        self.i = i
-
-    control = self.p + self.f + self.i
+    control = self.p + self.i + self.f
     self.saturated = self._check_saturation(control, check_saturation, error)
 
     self.control = clip(control, self.neg_limit, self.pos_limit)
--- a/selfdrive/controls/lib/pid.py
+++ b/selfdrive/controls/lib/pid.py
@@ -1,3 +1,4 @@
+import numpy as np
 from numbers import Number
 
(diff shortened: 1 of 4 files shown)


pd0wm:
@ClockeNessMnstr I tried to keep the diff as small as possible, and applied the change to lateral only for now. Is there a specific reason you removed the checks on moving integrator away from the limits?

ClockeNessMnstr:
>     * Is there a specific reason you removed the checks on moving integrator away from the limits?
back-calculation is the anti-windup method. it replaces this clamping and driver override. 
Easy to see with "instant-clip", if the limits are hit then the Integrator is set so that P+I+FF = limit. 

>     * Any reason why you used an unwind based on last actuator instead of an instant clip?
this is just setting the unwind rate to (100 / rate) (cancelling out the rate) . I don't think we'll like that. it's not very stable if it's even still stable.

pd0wm:
Yeah, I see how it's equivalent to the other limit. Since we do this for lateral only, we have to keep it around for longitudinal, but can remove it later.

I don't see how this would make things unstable. Can you explain that more?

ClockeNessMnstr:
Also, anecdotally: when hitting the max step limits any noise in P or FF could cause the controller to reverse direction for a frame before the integrator action can keep pushing it higher. This Can cause delays in applying the proper torque. Similar to a dead time delay this would be destabilizing

ClockeNessMnstr:
Yes, I think that's the same effect. Something like below is what I was imagining.

Output is clipped  to .75 with p = 0.5, i = 0.5 (assume ff = 0)

Last output sets i = .25
Error noise happens (drops to 0), p = 0, i = 0.25 , output is clipped to .5 (.25 max step)

Next iteration i will be set to .5 (last output - p = .5).
The iteration after the noise is gone, p = 0.5, i = .5
Clipped to .75

Frames are: .75, .5 (noise), .75

Frames should be .75, 1.0, 1.0 (max steer)

HaraldSchafer:
Sorry this took a while to review, I really struggled to understand what exactly this change implies.

I think I get it now and I do agree it is conceptually cleaner. From what I understand it just deals better with getting the most of the actuators given the limits and some noise, as well as doing antiwindup?

I do have comments about the code itself which is inline.

selfdrive/controls/lib/pid.py
```diff
@@ -60,27 +61,33 @@ def reset(self):
...
     error = float(apply_deadzone(setpoint - measurement, deadzone))
     self.p = error * self.k_p
     self.f = feedforward * self.k_f
 
-    if override:
-      self.i -= self.i_unwind_rate * float(np.sign(self.i))
+    # Clip integrator based on the last output value
+    if last_output is not None:
+      if not freeze_integrator:
+        i_bf = self.p + self.i + self.f - last_output
```
HaraldSchafer, line 74: bf? can we give this a name that makes it's function clear?

HaraldSchafer: Can we make it clear self.i refers to the previous iteration and self.p and self.f do not, by renaming the appropriately? Now the code looks similar to this below `control = self.p + self.i + self.f`,  but has a completely different meaning.

ClockeNessMnstr: i_bf was "integrator BackFeed" so probably not even an accurate description anymore. 

Maybe "actuator_difference"

selfdrive/controls/lib/pid.py
```diff
@@ -60,27 +61,33 @@ def reset(self):
...
+        self.i = self.i +  self.k_i * (error * self.i_rate - i_bf * self.i_bf_rate)
     else:
```
HaraldSchafer, line 76: There are now entirely different codepaths to do very similar things. Can we put it all in the same codepath? Or if this just a temporary thing, add TODOs making this clear what should still be changed.

selfdrive/controls/lib/pid.py
ClockeNessMnstr: > Can we make it clear self.i refers to the previous iteration and self.p and self.f do not, by renaming the appropriately? Now the code looks similar to this below `control = self.p + self.i + self.f`, but has a completely different meaning.

Could save self.control_last = control before the return and use that instead of each component.

selfdrive/controls/lib/pid.py
```diff
@@ -27,6 +27,7 @@ def __init__(self, k_p, k_i, k_f=1., pos_limit=None, neg_limit=None, rate=100, s
 
     self.sat_count_rate = 1.0 / rate
     self.i_unwind_rate = 0.3 / rate
+    self.i_bf_rate = 2.0 / rate
```
ClockeNessMnstr, line 30: Unwind rate would work better here for clarity but it's used in the old logic still

HaraldSchafer:
@ClockeNessMnstr I realized you don't have access to make the improvements to this branch. I can do it when back in office. The last thing we probably need is some unit tests that test the cases like noisy plan, override, etc...

Then I'll test in the car when I am back in office.

ClockeNessMnstr:
@pd0wm, I've got a cleanup commit ready for @HaraldSchafer's review if i can push it to this PR branch. 

Looking at setting up/starting unit tests too. Are there already some steering tests to go off of?

pd0wm:
Feel free to open a new PR. I'll close this one if you open it.

PR #23245, opened 2021-12-16
modeldata: use a constexpr function to build the IDXS

deanlee:


pd0wm:
@HaraldSchafer any benefit to computing these at compile time? Or are you copying them from somewhere else anyway?

deanlee:
There is a python version index_function does the same thing.

https://github.com/commaai/openpilot/blob/55390d273f36f49da3896e687ac5530f40c1d150/selfdrive/modeld/constants.py#L1-L7

HaraldSchafer:
Yeah I think it's cleaner not to have to manually generate them when they change

PR #23247, opened 2021-12-16
add BackCalculation to " PID: Use last actuators to limit integrator #23244 "

ClockeNessMnstr:
rebased again. 

kept old controller for long but still replaced override and clamping behavior when last_output is provided to the controller.


ClockeNessMnstr:
This is moving https://github.com/commaai/openpilot/pull/23219 to https://github.com/commaai/openpilot/pull/23244

PR #23248, opened 2021-12-16
replay: Add next engagement / disengagement jump capabilities

### LINKED ISSUE #23242: Add next engagement / disengagement jump capabilities to replay
It would be nice if replay had shortcuts for jumping to a few seconds (2?) before the next engagement or disengagement edge, maybe by typing `e` or `d` respectively.

deanlee:
use qlogs to do a quick search for the next engagement/disengagement  in route. resolve https://github.com/commaai/openpilot/issues/23242
`e `- seek to 2 seconds before the next engagement
`d` - seek to 2 secconds before the disengagement

### DIFF
--- a/selfdrive/ui/replay/replay.cc
+++ b/selfdrive/ui/replay/replay.cc
@@ -116,19 +116,19 @@ void Replay::doSeek(int seconds, bool relative) {
 void Replay::doSeekToFlag(FindFlag flag) {
   if (flag == FindFlag::nextEngagement) {
     qInfo() << "seeking to the next engagement...";  
-  } else {
+  } else if (flag == FindFlag::nextEngagement) {
     qInfo() << "seeking to the disengagement...";  
   }
 
   updateEvents([&]() {
-    auto next = find(flag);
-    if (next) {
+    if (auto next = find(flag)) {
       cur_mono_time_ = *next;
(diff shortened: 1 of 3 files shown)


deanlee:
My network is now very slow and almost impossible to do testing (current download speed is 15kb/s).  From the code point of view, there should be no major problems.so I opened this PR for review.

selfdrive/ui/replay/replay.cc
```diff
@@ -111,6 +113,58 @@ void Replay::doSeek(int seconds, bool relative) {
...
+    qInfo() << "seeking to the next engagement...";  
+  } else if (flag == FindFlag::nextEngagement) {
+    qInfo() << "seeking to the disengagement...";  
+  }
+
+  updateEvents([&]() {
+    if (auto next = find(flag)) {
+      double tm = *next - 2 * 1e9; // seek to 2 seconds before next
+      if (tm <= cur_mono_time_ && flag == FindFlag::nextEngagement) {
+        qInfo() << "already in engagement";
```
pd0wm, line 127: This just seeks back by two seconds now. Maybe just not do anything?

adeebshihadeh:
I think it would also be cool to include a colored timeline like connect.comma.ai using controlsState.alertStatus.

PR #23251, opened 2021-12-17
Road Roll Compensation Rebased

haraschax:
Squashed version of

https://github.com/commaai/openpilot/pull/22651


adeebshihadeh:
No need for a new PR/branch:
```
git fetch
git checkout roll-vm
git pull origin master
<resolve conflicts...>
git reset origin/master
git add .
git commit -m "squashed version of roll-vm"
git push -f
```

selfdrive/controls/lib/vehicle_model.py
```diff
@@ -19,6 +19,9 @@
 
 from cereal import car
 
+ACCELERATION_DUE_TO_GRAVITY = 9.8
```
geohot, line 22: Is this accurate enough? Isn't it 9.81, and also altitude dependent?

selfdrive/locationd/paramsd.py
```diff
@@ -41,16 +44,34 @@ def handle_log(self, t, which, msg):
       yaw_rate = msg.angularVelocityCalibrated.value[2]
       yaw_rate_std = msg.angularVelocityCalibrated.std[2]
 
+      localizer_roll = msg.orientationNED.value[0]
+      roll_valid = msg.orientationNED.valid and ROLL_MIN < localizer_roll < ROLL_MAX
+      if roll_valid:
```
geohot, line 49: Why isn't the std from the localizer used?

Why do you have to put in roll observation if invalid but not yaw?

selfdrive/controls/lib/vehicle_model.py
GregorKikelj: image
It varies by more than the error, and the relative error between 9.81 and 9.8 is 0.1%.
Going from 6400km from the center of the earth at sea level to 6403km (+10k feet) also drops the acceleration by less than 0.1%

PR #22726, opened 2021-10-28
Hyundai: Add Car Port for 2022 Kia Niro EV

sunnyhaibin:
**Checklist**
- [x] added to RELEASES, CARS
- [x] test route added to [test_routes.py](https://github.com/sunnyhaibin/openpilot-1/blob/94ddfc7163f2f0cb5dbc9eea7df69606e4db0ab8/selfdrive/test/test_routes.py#L92)
- [x] route with openpilot: `c011e480f2430f67|2021-10-28--11-09-58`

Thanks to community Niro EV owner Kolibabas.

PR #23253, opened 2021-12-17
HKG: Add and Fix FW for 2022 Kia Niro EV

sunnyhaibin:
Add firmware for the 2022 Kia Niro EV. Expand supported model-year range.

Remove `esp` from `KIA_NIRO_EV` as it was incorrectly added to `KIA_NIRO_EV` previously without functionality check from a Niro EV user. The incorrect `esp` firmware was confirmed by @adeebshihadeh that it was not from a Kia Niro EV.

Before removing `esp` firmware query from `KIA_NIRO_EV`:
- Car could not be recognized, even with the correct firmware versions added

After removing `esp` firmware query from `KIA_NIRO_EV`:
- Car could be recognized

Route ID:
* `51e260f2c91cb05f|2021-12-16--22-26-36`
* `0807a60245c2867e|2021-12-28--19-06-33`

Harness Type: [Hyundai H](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)

Thanks to community Niro EV owner Kolibabas and itsTreble.

This is a revised Pull Request of #22726.


pd0wm:
Can you fix the tests?

sunnyhaibin:
> Can you fix the tests? Anything we can do about the overlap between the Niro and the Kona?

@pd0wm The tests are fixed now. It seems like the FW versions below were randomly added under `KIA_NIRO_EV` from the FW dump from #20702 previously. Looking back into the Git history, seems like these FW versions were never used for `KIA_NIRO_EV`.
```
  CAR.KIA_NIRO_EV: {
    (Ecu.fwdRadar, 0x7D0, None): [
      b'\xf1\x00OSev SCC F-CUP      1.00 1.01 99110-K4000         ',
    ],
    (Ecu.eps, 0x7D4, None): [
      b'\xf1\x00OS  MDPS C 1.00 1.04 56310K4050\x00 4OEDC104',
    ],
    (Ecu.fwdCamera, 0x7C4, None): [
      b'\xf1\x00OSE LKAS AT EUR LHD 1.00 1.00 95740-K4100 W40',
    ],
  },
```

After removing the above duplicated FW versions, the tests ran successfully.

PR #23254, opened 2021-12-17
Panda class usbdevice

deanlee:


deanlee:
fixed 2 bugs:

1. control_transfer missed parameters data & length
2. no need to check `transferred` after bulk read. (the transferred is 0 after bulk read)

PR #23256, opened 2021-12-17
C2: add button to launch wifi settings in text window

deanlee:
If there is a startup error and the Wifi is not automatically connected, the only way to ssh to the C2 is re-flash the device, and run setup again

### DIFF
--- a/selfdrive/ui/qt/text.cc
+++ b/selfdrive/ui/qt/text.cc
@@ -2,6 +2,7 @@
 #include <QLabel>
 #include <QPushButton>
 #include <QScrollBar>
+#include <QSet>
 #include <QVBoxLayout>
 #include <QWidget>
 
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
It would be cool to add the wifi settings as a button in the power menu in NEOS, though I'm not sure if we're doing another NEOS yet.

pd0wm:
I assume this needs the event filtering, otherwise touches will also be handled by the Qt app and you can accidentally click Reboot/Exit. https://github.com/commaai/openpilot/blob/master/selfdrive/ui/qt/window.cc#L88

deanlee:
tested on c2,lgtm

adeebshihadeh:
We'll likely ship a new NEOS for 0.8.13, so I'll look into adding this into the power menu.

adeebshihadeh:
NEOS 19 will have a settings button in the power menu.

PR #23258, opened 2021-12-17
Adding Corolla FingerPrint

rogerioaguas:
**Description** Adding a Corolla Cross Finger Print

**Verification** Test on my car and run **python -m compileall value.py to verifcation crash**


rogerioaguas:
I will close pull request because i dont create unit tests

ErichMoraga:
"Verification Test on my car and run python -m compileall value.py to verifcation crash"

There were a couple improper indentations in your commit.

PR #23262, opened 2021-12-17
docs: setup docs for C code

tecandrew:
add c docs

* `breathe` does the generation, not explicit `Doxygen`
* documents only nearby (.h, .cc) pairs
* only explicit listed files is documented: `sphinx` can't display some `breathe`-generated xml files

rel: #23204 (comment)


tecandrew:
figured out how to do it without `cmake`.

tecandrew:
😄
this was in reference to #22732. think it could be closed

PR #23263, opened 2021-12-18
plotjuggler: support segment names

gregjhogan:
I got annoyed with not being able to copy/paste segment names from connect/useradmin
- support passing in segment name (e.g. `4cf7a6ad03080c90|2021-09-29--13-46-36--1`)
- support passing in route name and segment count


tools/plotjuggler/juggle.py
```diff
@@ -4,20 +4,29 @@
...
 from tools.lib.robust_logreader import RobustLogReader
-from tools.lib.route import Route
+from tools.lib.route import Route, RouteSegmentName, SEGMENT_NAME_RE, ROUTE_NAME_RE
 from urllib.parse import urlparse, parse_qs
 
 juggle_dir = os.path.dirname(os.path.realpath(__file__))
 
 DEMO_ROUTE = "4cf7a6ad03080c90|2021-09-29--13-46-36"
 
+def get_route_and_segment_num(name):
```
adeebshihadeh, line 22: seems like this should be handled in RouteSegmentName, then pass in a RouteSegmentName instead of a string to juggle_route?

tools/plotjuggler/juggle.py
```diff
@@ -47,28 +56,31 @@ def start_juggler(fn=None, dbc=None, layout=None):
...
-    query = parse_qs(urlparse(route_name).query)
+def juggle_route(route_or_segment_name, segment_count, qlog, can, layout):
+  segment_start = 0
+  if 'cabana' in route_or_segment_name:
+    query = parse_qs(urlparse(route_or_segment_name).query)
     api = CommaApi(get_token())
     logs = api.get(f'v1/route/{query["route"][0]}/log_urls?sig={query["sig"][0]}&exp={query["exp"][0]}')
-  elif route_name.startswith("http://") or route_name.startswith("https://") or os.path.isfile(route_name):
-    logs = [route_name]
+  elif route_or_segment_name.startswith("http://") or route_or_segment_name.startswith("https://") or os.path.isfile(route_or_segment_name):
```
adeebshihadeh, line 65: would be nice to have RouteSegmentName handle cabana URLs too

tools/lib/route.py
```diff
@@ -7,7 +7,8 @@
 from tools.lib.auth_config import get_token
 from tools.lib.api import CommaApi
 
-SEGMENT_NAME_RE = r'[a-z0-9]{16}[|_][0-9]{4}-[0-9]{2}-[0-9]{2}--[0-9]{2}-[0-9]{2}-[0-9]{2}--[0-9]+'
+ROUTE_NAME_RE = r'[a-z0-9]{16}[|_][0-9]{4}-[0-9]{2}-[0-9]{2}--[0-9]{2}-[0-9]{2}-[0-9]{2}'
+SEGMENT_NAME_RE = r'{}--[0-9]+'.format(ROUTE_NAME_RE)
```
adeebshihadeh, line 11: support slash format too?

tools/plotjuggler/juggle.py
gregjhogan: yeah I can add a parser that parses route and segment names

tools/plotjuggler/juggle.py
gregjhogan: this seems like a good candidate for a separate PR to enhance the `Route` class (I assume you need to maintain the cabana URL scheme for files in case you don't have API access to the device)

tools/lib/route.py
gregjhogan: added

adeebshihadeh (commented):
Looks good, but I still think `parse_route_or_segment_name` is cleaner as a class. `RouteIdentifier`, like replay uses, seems like an adequate name.

gregjhogan:
> Looks good, but I still think `parse_route_or_segment_name` is cleaner as a class. `RouteIdentifier`, like replay uses, seems like an adequate name.

I added `RouteName` and `SegmentName` classes to handle parsing different input formats and generating the canonical name (regardless of the input format) with `SegmentName` having an optional flag so it can parse either a route or segment name

PR #23268, opened 2021-12-19
Add More Types for mypy

geohot:
Was bored on the plane. Fixed a few subtle potential bugs too.

Why is "driverview" discussed in "process.py"? Can this be refactored better?

### DIFF
--- a/selfdrive/manager/manager.py
+++ b/selfdrive/manager/manager.py
@@ -100,12 +100,12 @@ def manager_init() -> None:
                    device=HARDWARE.get_device_type())
 
 
-def manager_prepare():
+def manager_prepare() -> None:
   for p in managed_processes.values():
     p.prepare()
 
 
-def manager_cleanup():
+def manager_cleanup() -> None:
   # send signals to kill all procs
   for p in managed_processes.values():
     p.stop(block=False)
--- a/selfdrive/manager/manager.py
+++ b/selfdrive/manager/manager.py
@@ -5,6 +5,7 @@ import signal
 import subprocess
 import sys
 import traceback
+from typing import List, Tuple, Union
 
 import cereal.messaging as messaging
 import selfdrive.crash as crash
@@ -25,7 +26,7 @@ from selfdrive.version import is_dirty, get_commit, get_version, get_origin, get
 sys.path.append(os.path.join(BASEDIR, "pyextra"))
 
 
-def manager_init():
+def manager_init() -> None:
   # update system time from panda
   set_time(cloudlog)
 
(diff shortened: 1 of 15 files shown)


selfdrive/thermald/thermald.py
```diff
@@ -166,8 +166,8 @@ def thermald_thread():
   onroad_conditions = {
     "ignition": False,
   }
-  startup_conditions = {}
-  startup_conditions_prev = {}
+  startup_conditions: Dict[str, Any]= {}
+  startup_conditions_prev: Dict[str, Any] = {}
```
adeebshihadeh, line 170:
```suggestion
  startup_conditions: Dict[str, bool]= {}
  startup_conditions_prev: Dict[str, bool] = {}
```

selfdrive/thermald/thermald.py
```diff
@@ -166,8 +166,8 @@ def thermald_thread():
   onroad_conditions = {
```
adeebshihadeh, line 166:
```suggestion
  onroad_conditions: Dict[str, bool] = {
```

selfdrive/thermald/thermald.py
```diff
@@ -3,7 +3,7 @@
 import os
 import time
 from pathlib import Path
-from typing import Dict, Optional, Tuple
+from typing import Dict, NoReturn, Optional, Tuple, Any
```
adeebshihadeh, line 6:
```suggestion
from typing import Dict, NoReturn, Optional, Tuple
```

selfdrive/manager/process.py
```diff
@@ -148,14 +151,18 @@ def stop(self, retry=True, block=True):
...
-  def signal(self, sig):
+  def signal(self, sig: int) -> None:
     if self.proc is None:
       return
 
     # Don't signal if already exited
     if self.proc.exitcode is not None and self.proc.pid is not None:
       return
 
+    # Can't signal if we don't have a pid
```
pd0wm, line 162: This shouldn't happen in practice since the child classes immediately start the process after creating the object. But doesn't hurt to check in case we change the implementation later.

selfdrive/manager/process.py
```diff
@@ -37,7 +38,7 @@ def launcher(proc, name):
     cloudlog.bind(daemon=name)
 
     # exec the process
-    mod.main()
+    getattr(mod, 'main')()
```
pd0wm, line 41: Why change this? Both cases will result in `AttributeError` if main doesn't exist

selfdrive/manager/manager.py
```diff
@@ -56,7 +57,7 @@ def manager_init():
 
   # is this dashcam?
   if os.getenv("PASSIVE") is not None:
-    params.put_bool("Passive", bool(int(os.getenv("PASSIVE"))))
+    params.put_bool("Passive", bool(int(os.getenv("PASSIVE", "0"))))
```
pd0wm, line 60: Does mypy need this? We already check that `PASSIVE` is set in the line above, and now it looks like this`"0"` is special but it will never actually be used.

geohot: It does, otherwise it doesn't know the type of the os.getenv

selfdrive/manager/process.py
geohot: mypy fails without this change. I agree they are functionally the same

PR #23272, opened 2021-12-20
add missing toyota fw engine

AlexandreSato:
test_route: 541c7e27e413735c|2021-12-17--14-21-53

Adding fingerprint of a corolla gr here from Brazil


pd0wm:
I don't see any entries in `carParams.fw`. Can you provide a route with firmware versions present?

AlexandreSato:
> I don't see any entries in `carParams.fw`. Can you provide a route with firmware versions present?

541c7e27e413735c|2022-01-10--08-10-15

PR #23275, opened 2021-12-20
Pyupgrade 3.6: Update syntax with Python 3.6+ features

EwoutH:
Updated Python code with Python 3.6+ features:
- utf-8 encoding is now the default ([PEP 3120](https://www.python.org/dev/peps/pep-3120/), Python 3.0+)
- Replace list comprehensions by Generator Expressions ([PEP 289](https://www.python.org/dev/peps/pep-0289/), Python 2.4+)
- Replace `yield` loop by `yield from` ([PEP 380](https://www.python.org/dev/peps/pep-0380/), Python 3.3+)
- Replace the IOError alias by OSError ([PEP 3151](https://www.python.org/dev/peps/pep-3151/), Python 3.3+)
- Remove the default subclass `(object)` when defining a class
- Define sets with curly braces `{}` instead of `set()`
- Remove `"r"` parameter from open function, which is default

Used [pyupgrade](https://github.com/asottile/pyupgrade) 2.29.1 with `--py36-plus` as starting point for this PR.

### DIFF
--- a/docs/CARS.md
+++ b/docs/CARS.md
@@ -150,7 +150,7 @@
 | Subaru    | Crosstrek 2018-20             | EyeSight          | Stock            | 0mph               | 0mph         |
 | Subaru    | Forester 2019-21              | EyeSight          | Stock            | 0mph               | 0mph         |
 | Subaru    | Impreza 2017-19               | EyeSight          | Stock            | 0mph               | 0mph         |
-| Volkswagen| Arteon 2021<sup>4</sup>       | Driver Assistance | Stock            | 0mph               | 0mph         |
+| Volkswagen| Arteon 2018, 2021<sup>4</sup> | Driver Assistance | Stock            | 0mph               | 0mph         |
 | Volkswagen| Atlas 2018-19                 | Driver Assistance | Stock            | 0mph               | 0mph         |
 | Volkswagen| California 2021<sup>4</sup>   | Driver Assistance | Stock            | 0mph               | 32mph        |
 | Volkswagen| e-Golf 2014, 2019-20          | Driver Assistance | Stock            | 0mph               | 0mph         |
--- a/selfdrive/car/volkswagen/values.py
+++ b/selfdrive/car/volkswagen/values.py
@@ -101,19 +101,24 @@ class CAR:
 FW_VERSIONS = {
   CAR.ARTEON_MK1: {
     (Ecu.engine, 0x7e0, None): [
+      b'\xf1\x873G0906259F \xf1\x890004',
       b'\xf1\x873G0906259P \xf1\x890001',
     ],
     (Ecu.transmission, 0x7e1, None): [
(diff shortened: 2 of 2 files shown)


selfdrive/controls/radard.py
```diff
@@ -135,7 +135,7 @@ def update(self, sm, rr, enable_lead):
       self.tracks[ids].update(rpt[0], rpt[1], rpt[2], v_lead, rpt[3])
 
     idens = list(sorted(self.tracks.keys()))
-    track_pts = list([self.tracks[iden].get_key_for_cluster() for iden in idens])
+    track_pts = list(self.tracks[iden].get_key_for_cluster() for iden in idens)
```
GregorKikelj, line 139: Could list be replaced with [] here?

selfdrive/controls/radard.py
```diff
@@ -135,7 +135,7 @@ def update(self, sm, rr, enable_lead):
...
```
GregorKikelj, line 138:
```suggestion
    track_pts = [self.tracks[iden].get_key_for_cluster() for iden in idens]
```

EwoutH: Good catch, implemented it.

EwoutH:
Thanks for the suggestions, I implemented them, squashed the commits, rebased and signed it.

selfdrive/test/process_replay/compare_logs.py
```diff
@@ -58,7 +58,7 @@ def compare_logs(log1, log2, ignore_fields=None, ignore_msgs=None, tolerance=Non
   if ignore_msgs is None:
     ignore_msgs = []
 
-  log1, log2 = [list(filter(lambda m: m.which() not in ignore_msgs, log)) for log in (log1, log2)]
+  log1, log2 = (list(filter(lambda m: m.which() not in ignore_msgs, log)) for log in (log1, log2))
```
GregorKikelj, line 61: I think this should also work.
```suggestion
  log1, log2 = ([filter(lambda m: m.which() not in ignore_msgs, log)] for log in (log1, log2))
```

GregorKikelj: Btw squashing is done automatically on merge, don't think it's needed in a PR

adeebshihadeh:
Can you fix the failing tests?

selfdrive/test/process_replay/compare_logs.py
GregorKikelj: Sorry, though that [] and list() were interchangeable, apparently not in this case

EwoutH:
I think replacing the `list()` by `[]` as @GregorKikelj suggested didn't play well with the lambda, so I reverted it. I did this a bit to quick, and accidentally reset the whole branch. Luckily `git reflow` is a thing and I recovered it. I couldn't get this PR reopened though, so it's now #23305.

(note to self: be really careful touching `git reset --hard`)

PR #23279, opened 2021-12-20
Ford FPv2 support

incognitojam:
I believe the FW version request should be the same as Nissan and Mazda (as per phoenixpilot fork), so am creating this PR provisionally until I can test.

I've included the ECU FW for the Focus 2018 which I managed to retrieve already. Will keep as a draft for now.

**Checklist**
- [ ] Get raw responses that are returned from the ECUs and see if these requests are actually required, or if the responses are just in a different format
- [ ] Compare responses from REQUEST and REQUEST_A


incognitojam:
It seems like I might have to wait for DBC updates before adding the Focus... see https://github.com/commaai/opendbc/pull/476

incognitojam:
In fact, I could drop the Focus commit for now and simply merge the FW request alone if you are happy with that?

incognitojam:
I realise that it creates a lot of problems to by adding a new car model, so should this just be included with the car port instead?

incognitojam:
I have some code for the carstate, interface etc. to add the Focus. Would it be better to open a new PR with that first and then look at FPv2 afterwards?

incognitojam:
Currently get no FW results when sending the request on bus 1, however phoenixpilot sends the request on both bus 0 and bus 1, so will give that a go today

incognitojam:
Here's the carFw with these changes:
Screenshot from 2021-12-23 14-52-37

Note that I've removed the car model from this PR.

The implementation is a little janky. If it doesn't get a result for the VIN it tries on bus 0 instead, and if there are less than 3 results for carFw on bus 1, it tries bus 0 and joins the results. Possibly need to remove duplicates from the list though.

adeebshihadeh:
I'm not sure it makes sense to merge this before we have a full car port. Perhaps we should leave this as a draft until that's merged in. How close is a port for an LCA vehicle? Does it have the lockout like the other Fords?

incognitojam:
I've driven with it for several minutes without lockout, but it is not super responsive and I had to help steer at times. I don't know if this is poor tuning or something else. It would be good to get feedback with a PR!

incognitojam:
I'm closing this for now, can look at it again after some progress is made with the other Ford PRs

PR #23280, opened 2021-12-21
GM: Switch to AcceleratorPedal2

JMPZ11:
**Description** [] Update carstate to use AcceleratorPedal2 for gas and remove the requirement for AcceleratorPedal. The Pedal value in the two messages is identical, but AcceleratorPedal is not present on the 2020 Silverado.

**Verification** [] Has been running without issue on several vehicles. I also checked all the fingerprints to confirm that if 417 is present, 452 is also -- meaning it is unlikely to cause any problems. passes testing.

Pairs with Panda PR https://github.com/commaai/panda/pull/813

### DIFF
--- a/selfdrive/car/gm/carstate.py
+++ b/selfdrive/car/gm/carstate.py
@@ -36,7 +36,7 @@ class CarState(CarStateBase):
     if ret.brake < 10/0xd0:
       ret.brake = 0.
 
-    ret.gas = pt_cp.vl["AcceleratorPedal"]["AcceleratorPedal"] / 254.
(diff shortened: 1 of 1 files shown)


JasonJShuler:
FYI - the failures are Mazda related, doesn't look like there were any GM issues

JasonJShuler:
@adeebshihadeh per your request in the Panda PR, how should I go about providing the necessary validation?

adeebshihadeh:
I couldn't push to your branch for some reason so moved this to #23376.

PR #23273, opened 2021-12-20
Mazda: alert when LKAS is disabled

Jafaral:
Fixes: #22746
Depends: https://github.com/commaai/cereal/pull/235

Signed-off-by: Jafar Al-Gharaibeh <to.jafar@gmail.com>

PR #23288, opened 2021-12-22
Mazda:  alert when LKAS is disabled, a second attempt

Jafaral:


Jafaral:
re #23273

Jafaral:
@adeebshihadeh , I fixed all failed test, but it is not clear what is wrong with process_replay, any idea?

Jafaral:
Ah, I think I got it

Jafaral:
scratch that, some Mazda tests are seeing the new event lkasDisabled and failed to match expected output. Not sure where to fix that quickly

PR #23289, opened 2021-12-22
replay: add macros for logging

deanlee:


adeebshihadeh:
I'd like to propose something to replace this entirely, a terminal UI that looks something like this:

```bash
Replaying: bd6a637565e91581|2021-10-29--22-18-31

[===============>                                                   ]   00:10 (0/12)

-----------------------
|  Seek instructions  |
-----------------------
|  s  +10s |  S -10s  |
|  m  +60s |  M -60s  |
-----------------------

```

We could also do some progress bars when blocked on downloading some of the log files. What do you think?

PR #23291, opened 2021-12-22
painter opacity should be reset in OnroadHud::drawIcon to prevent wro…

alfhern:
painter opacity should be reset in `OnroadHud::drawIcon` to prevent wrong opacity in future painter operations


deanlee:
why it should be reset? which paint operation is affected by it? I assume that the drawIcon is the last drawing operation in the paintevent.

alfhern:
Fair enough, as long as that is the case it is fine as it is.
I noticed the issue while working on our fork, in which we have paint actions after the calls to `drawIcon`. 
Thought of it as making sense to just prevent issues like this in the future. Anyhow, I made the change on my code, I guess this was a suggestion more than a bug report, feel free to ignore.

adeebshihadeh:
This doesn't even preserve the old opacity from before the call. I agree with @deanlee and think this it's fine as is.

PR #23293, opened 2021-12-22
CameraView:  sync pbo with fence

deanlee:
save some cpu usage and reduce many "slow frame time" messages

### DIFF
--- a/selfdrive/ui/qt/widgets/cameraview.cc
+++ b/selfdrive/ui/qt/widgets/cameraview.cc
@@ -106,6 +106,11 @@ CameraViewWidget::~CameraViewWidget() {
     glDeleteVertexArrays(1, &frame_vao);
     glDeleteBuffers(1, &frame_vbo);
     glDeleteBuffers(1, &frame_ibo);
+    for (auto sync : gl_fences) {
+      if (glIsSync(sync)) {
+        glDeleteSync(sync);
+      }
+    }
   }
   doneCurrent();
 }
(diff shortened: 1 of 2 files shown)


selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -289,20 +295,23 @@ void CameraViewWidget::vipcThread() {
...
+          memcpy(texture_buffer, buf->addr, buf->len);
+          gl_buffer->unmap();
+
+          // copy pixels from PBO to texture object
+          glBindTexture(GL_TEXTURE_2D, texture[buf->idx]->frame_tex);
+          glTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, buf->width, buf->height, GL_RGB, GL_UNSIGNED_BYTE, 0);
+          glBindTexture(GL_TEXTURE_2D, 0);
+          assert(glGetError() == GL_NO_ERROR);
+
+          wait_fence.reset(new WaitFence());
```
pd0wm, line 311: Shouldn't we have a fence per texture?

deanlee: I think one is enough here.  In fact, we don’t actually need multiple textures on c3, one texture is enough. This can be optimized later

adeebshihadeh:
The CPU usage test is failing now on master with UI using up to 75% on the comma two.

pd0wm:
Think the CI failure was a fluke. This PR shouldn't have any effect on the C2. Made a new PR that also passes the cpu usage test: https://github.com/commaai/openpilot/pull/23409

PR #23295, opened 2021-12-22
LongitudinalMpc: Remove class accel_limit_arr

sshane:
Removes unused class variable except in the e2e xva update function

### DIFF
--- a/selfdrive/controls/lib/longitudinal_mpc_lib/long_mpc.py
+++ b/selfdrive/controls/lib/longitudinal_mpc_lib/long_mpc.py
@@ -194,6 +194,7 @@ class LongitudinalMpc:
   def __init__(self, e2e=False):
     self.e2e = e2e
     self.reset()
+    self.e2e_accel_limit_arr = np.full((N+1, 2), (-10., 10.))
     self.source = SOURCES[2]
 
   def reset(self):
@@ -345,9 +346,8 @@ class LongitudinalMpc:
     for i in range(N):
       self.solver.cost_set(i, "yref", self.yref[i])
     self.solver.cost_set(N, "yref", self.yref[N][:COST_E_DIM])
-    accel_limit_arr = np.full((N+1, 2), (-10., 10.))
     x_obstacle = 1e5*np.ones((N+1))
-    self.params = np.concatenate([accel_limit_arr,
+    self.params = np.concatenate([self.e2e_accel_limit_arr,
                                   x_obstacle[:,None],
                                   self.prev_a[:,None]], axis=1)
     self.run()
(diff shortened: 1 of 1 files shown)


pd0wm:
@mitchellgoffpc  Can these be removed or are they used in training?

selfdrive/controls/lib/longitudinal_mpc_lib/long_mpc.py
```diff
@@ -347,15 +344,13 @@ def update_with_xva(self, x, v, a):
     for i in range(N):
       self.solver.cost_set(i, "yref", self.yref[i])
     self.solver.cost_set(N, "yref", self.yref[N][:COST_E_DIM])
-    self.accel_limit_arr[:,0] = -10.
-    self.accel_limit_arr[:,1] = 10.
+    accel_limit_arr = np.full((N+1, 2), (-10., 10.))
```
pd0wm, line 347: Let's only build it once since it's static.

sshane: Since we only change the prev_a live parameter I just set all the other parameters once in the `set_weights` function

selfdrive/controls/lib/longitudinal_mpc_lib/long_mpc.py
```diff
@@ -224,6 +221,9 @@ def reset(self):
   def set_weights(self):
     if self.e2e:
       self.set_weights_for_xva_policy()
+      self.params[:,0] = -10.
```
pd0wm, line 224: I assume this should happen regardless of `self.e2e`?

pd0wm: Weird that the tests still pass

selfdrive/controls/lib/longitudinal_mpc_lib/long_mpc.py
```diff
@@ -224,6 +221,9 @@ def reset(self):
...
+      self.params[:,1] = 10.
+      self.params[:,2] = 1e5*np.ones((N+1))
```
pd0wm, line 226: Do we still need the `np.ones`?

pd0wm:
```suggestion
      self.params[:,2] = 1e5
```

selfdrive/controls/lib/longitudinal_mpc_lib/long_mpc.py
```diff
@@ -348,15 +348,9 @@ def update_with_xva(self, x, v, a):
     for i in range(N):
       self.solver.cost_set(i, "yref", self.yref[i])
     self.solver.cost_set(N, "yref", self.yref[N][:COST_E_DIM])
-    self.accel_limit_arr[:,0] = -10.
-    self.accel_limit_arr[:,1] = 10.
-    x_obstacle = 1e5*np.ones(N+1)
-    self.params = np.concatenate([self.accel_limit_arr,
-                                  x_obstacle[:, None],
-                                  self.prev_a[:,None]], axis=1)
+    self.params[:,3] = self.prev_a
```
pd0wm, line 351:
```suggestion
    self.params[:,3] = np.copy(self.prev_a)
```

Just to be sure I'll match `update` here.

PR #23300, opened 2021-12-23
Qt/ui: Add replay capabilities

### LINKED ISSUE #2547: Add replay (unlogger) to QT UI
So at first it sounds crazy. But then you remember how bad unlogger is and you think about it more.

A dashcam has the ability to replay the "recording" on the device. openpilot should too, and it can replace unlogger. Replay with the same messages on the bus. Grab the files from the API, or fast fast if they are already on the device. We do a really bad job surfacing the ability to replay routes in the detail we record them with explorer and the app.

I imagine a nice movie timeline scroll overlayed on the bottom of UI while replaying. This completes the device as a dashcam, but...

The real win would be for PCs. I run ui on my PC, pair the PC to my account, and can access all the routes from my device. I replay one, I can inspect the messages, replace services, etc...all the things we do now with unlogger.

The docker is great for carla (minus the current bug). I was pleasantly surprised when the openpilot UI came up. Imagine I could, with just the one docker command and a quick pair to my account from my phone, start replaying one of my drives (or of course some demo drives without pairing).

Either way, we need to surface playing back drives better to (prospective) developers.

deanlee:
resolve https://github.com/commaai/openpilot/issues/2547
- [ ] Add widgets needed for replay in ui
- [ ] Add a small program replay_ui to replay route in a simple UI
- [ ] Integrate into qt/ui

simple replay ui:

### DIFF
--- a/selfdrive/ui/.gitignore
+++ b/selfdrive/ui/.gitignore
@@ -5,6 +5,7 @@ _mui
 watch3
 installer/installers/*
 replay/replay
+replay/replay_ui
 replay/tests/test_replay
 qt/text
 qt/spinner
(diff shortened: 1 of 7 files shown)


adeebshihadeh:
I was actually thinking we would use the normal onroad UI for this. Would that simplify things?

deanlee:
> I was actually thinking we would use the normal onroad UI for this. Would that simplify things?

 I need time to think about this. I am worried that reusing onroad ui will introduce new complexity.

pd0wm:
We also have some design already for an on-device replay. We will get back to you in the new year about our plans with this!

PR #23304, opened 2021-12-24
replay: add download progress bar

deanlee:
<pre>
Downloading [====>                                         ]  6% 71.53 MB
</pre>

### DIFF
--- a/selfdrive/ui/replay/util.cc
+++ b/selfdrive/ui/replay/util.cc
@@ -76,8 +76,7 @@ struct DownloadProgressBar {
 
   void removeDownload(const std::string &url) {
     std::lock_guard lk(lock);
-    auto it = items.find(url);
-    if (it != items.end()) {
+    if (auto it = items.find(url); it != items.end()) {
       items.erase(it);
     }
   }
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Can we implement the rest of the terminal UI I mentioned [here](https://github.com/commaai/openpilot/pull/23289#issuecomment-1000585908)? Perhaps it would be easier if we also introduced a new library like ncurses, no reason to do the terminal UI stuff from scratch.

deanlee:
Do we need to use replay in an environment without qt?  If not, I think it may be better to implement it with qt instead of importing a new library, installing it on all devices, and writing bunch of code to implement a console ui.
we can reuse qt controls (such as timeline, progressbar...)  between replay and qt/ui  to save a lot of code. @adeebshihadeh what do you think?

adeebshihadeh:
Hmm, I like the idea of the terminal UI right now. Also, I’m thinking the replay UI is only for controls, since we use the normal UI for the video. @pd0wm what do you think, terminal UI or GUI?

adeebshihadeh:
I thought about this a bit more, and I'm really open to both. I still slightly prefer a terminal UI, mostly to avoid a second window, but a GUI would be super easy to do with Qt.

pd0wm:
I like a terminal UI since you can use it on a device over SSH, but don't think it's worth it over a Qt ui if it adds more than ~100 lines.

 @adeebshihadeh do we have ncurses on NEOS/AGNOS?

adeebshihadeh:
> I like a terminal UI since you can use it on a device over SSH, but don't think it's worth it over a Qt ui if it adds more than ~100 lines.

I'm willing to spend a few more lines (500?) if everything's nicely contained in its own file and doesn't really need to touch replay.

> @adeebshihadeh do we have ncurses on NEOS/AGNOS?

Yes, both have ncurses.

PR #23307, opened 2021-12-25
Car Port: 2022 Honda Civic

vanillagorillaa:

**Checklist**
- [x] added to README
- [x] route with openpilot: f44aa96ace22f34a|2021-12-22--06-22-31
- [x] route with stock system: f44aa96ace22f34a|2021-12-25--21-08-31
- [x] confirm spam resume works
- [x] ~~figure out possible new LKAS_HUD message~~

Requires: 
- [x] https://github.com/commaai/cereal/pull/234 - will make safety param
- [ ] https://github.com/commaai/opendbc/pull/618
- [ ] https://github.com/commaai/panda/pull/815

Lateral only for now, long should be pretty easy. There is no radar on this car so openpilot can send ACC commands directly to the car.


ZgLijian:
请问您是什么地区的，我是中国的，中国这边的2022款思域可以吗

ZgLijian:
What region are you from? I'm from China. Is the 2022 civic here ok

selfdrive/car/honda/hondacan.py
```diff
@@ -140,7 +145,10 @@ def create_ui_commands(packer, CP, pcm_speed, hud, is_metric, idx, stock_hud):
     'SET_ME_X48': 0x48,
     'STEERING_REQUIRED': hud.steer_required,
     'SOLID_LANES': hud.lanes,
+    'LANE_LINES': 3,
+    'DASHED_LANES': hud.lanes,
     'BEEP': 0,
+    'LKAS_PROBLEM': 0,
```
sshane, line 151:
```suggestion
    'LKAS_PROBLEM': 0 if frame > 200 else 1,
```

sshane:
Moved to

PR #23308, opened 2021-12-25
ui:  set the bg color of controllerUnresponsivePermanent to blue

deanlee:


adeebshihadeh:
Nice catch, but the permanent once should actually be blue.

PR #23311, opened 2021-12-26
firmware fingerprinting: order brand requests

gregjhogan:
Currently every new vehicle manufacturer we add to firmware fingerprinting slows down fingerprinting for all vehicles.  We can instead check what ECUs respond to UDS messages quickly, then only query firmware versions for the vehicle manufacturers that match (which I expect will normally be a single vehicle manufacturer).  If we find there is never more than two matching vehicle manufacturers, this probably even allows us to run fingerprinting for each vehicle manufacturer serial (eliminating any ordering problems of overlapping addresses when querying vehicle manufacturers in parallel).

Needs testing on all vehicles.  I currently have it fall back to current behavior if no candidates are found, but we probably need to ship this in shadow mode for a release and monitor if any vehicles would have failed to fingerprint correctly.


pd0wm:
If we build something like this it would be best if we can identify the brand exactly before firing off the actual queries.

Sometimes we have two queries for the same address, one brand responds to both and the other to only one. We then use the ordering of the queries to get the desired outcome, which is not ideal. If we manage to separate the brands completely, it will also make it easier to add cars or make changes to a specific brand without having to test all of them.

gregjhogan:
> If we build something like this it would be best if we can identify the brand exactly before firing off the actual queries.
> 
> Sometimes we have two queries for the same address, one brand responds to both and the other to only one. We then use the ordering of the queries to get the desired outcome, which is not ideal. If we manage to separate the brands completely, it will also make it easier to add cars or make changes to a specific brand without having to test all of them.

I tried to address this above, does my above idea of running any matching brands in serial not accomplish the same thing?  Allowing multiple is just a safety net and my guess is currently it will always identify a single brand.  It should also be valid to require a complete set of ECUs respond for a brand and we could validate ahead of time this will always identify a unique brand.

This currently sends 512 CAN messages which probably takes at least 0.1 seconds (the length of time I wait for responses) to send on a 500 kbps bus, so we should only send messages for CAN addresses.  I probably didn't notice any issues on my vehicle because it doesn't use 29 bit addrs and they are sent last.

I probably won't have time to finish this anytime soon but thought it would be a good idea to explain this idea with some code since I have suggested it a few times in the past.

adeebshihadeh:
@sshane let's try to get this merged soon after 0.8.14 goes to release. we'll need to test it on each make

sshane:
@gregjhogan is this good to start testing?

gregjhogan:
I think the idea is still solid, but I mentioned some things that may need to be addressed or improved above, so feel free to take it over.

sshane:
@gregjhogan I made it only query addresses in the FW versions, as well as take the bus into account (as now it's not always 1). This reduces the number of messages needed to be sent to only 41, let me know if I did that right.

Do we also need to account for different response offsets? Some makes aren't the same and we currently overwrite with the latest ones

selfdrive/car/ecu_addrs.py
```diff
@@ -0,0 +1,75 @@
...
+    if msg.dat[1] == (SERVICE_TYPE.TESTER_PRESENT + 0x40):
+      return True
+    # error response
+    if msg.dat[1] == 0x7F and msg.dat[2] == SERVICE_TYPE.TESTER_PRESENT:
+      return True
+  return False
+
+
+def get_all_ecu_addrs(logcan: messaging.SubSocket, sendcan: messaging.PubSocket, bus: int, timeout: float = 1, debug: bool = True) -> Set[int]:
+  addr_list = [0x700 + i for i in range(256)] + [0x18da00f1 + (i << 8) for i in range(256)]
```
sshane, line 28: Should we add an offset here since addresses at the top of the range won't be in the list?

sshane:
Results from a Hyundai Sonata with the engine both on and off: https://pastebin.com/cGCjvvL1

sshane:
Unfortunately this doesn't always work only including when there's >= 4 ECUs. This is on a RAV4 turned on into accessory mode (no engine start). It ended up failing to fingerprint:

```python
{'event': 'candidate brand intersection', 'brand': 'body', 'intersection': 'set()'}
{'event': 'candidate brand intersection', 'brand': 'honda', 'intersection': 'set()'}
{'event': 'candidate brand intersection', 'brand': 'hyundai', 'intersection': '{2024, 1996, 1976, 2008}'}
{'event': 'candidate brand intersection', 'brand': 'mazda', 'intersection': '{2024}'}
{'event': 'candidate brand intersection', 'brand': 'nissan', 'intersection': 'set()'}
{'event': 'candidate brand intersection', 'brand': 'subaru', 'intersection': '{1976, 2024}'}
{'event': 'candidate brand intersection', 'brand': 'toyota', 'intersection': '{2024, 1961, 1976}'}
{'event': 'candidate brand intersection', 'brand': 'volkswagen', 'intersection': '{2024}'}
```

And when the engine is turned on (ran normal queries):
```python
{'event': 'candidate brand intersection', 'brand': 'body', 'intersection': 'set()'}
{'event': 'candidate brand intersection', 'brand': 'honda', 'intersection': 'set()'}
{'event': 'candidate brand intersection', 'brand': 'hyundai', 'intersection': '{1976}'}
{'event': 'candidate brand intersection', 'brand': 'mazda', 'intersection': 'set()'}
{'event': 'candidate brand intersection', 'brand': 'nissan', 'intersection': 'set()'}
{'event': 'candidate brand intersection', 'brand': 'subaru', 'intersection': '{1976}'}
{'event': 'candidate brand intersection', 'brand': 'toyota', 'intersection': '{1976, 1961}'}
{'event': 'candidate brand intersection', 'brand': 'volkswagen', 'intersection': 'set()'}
```

gregjhogan:
4 was arbitrary, but yeah something is wrong if all the ECU addrs used for fingerprinting are not responding.
also if a single addr is used for two essential ecus that use sub addrs on that single addr, it probably needs to count as two matches

gregjhogan:
It seems like requiring 4 definitely won't work because we have some cars that only have firmware versions for 3 ECUs.  It should be fine to wait until you get a complete set of response addrs for a single platform in a brand before querying that brand for firmware versions (although this may require querying essential and non-essential ECU addrs).

Also, realize that if ECUs are not giving you any response to tester present (even an error response is acceptable) there is no way it is going to give you a firmware version.

selfdrive/car/ecu_addrs.py
```diff
@@ -0,0 +1,76 @@
...
+import traceback
+from typing import Dict, Set
+
+import cereal.messaging as messaging
+from panda.python.uds import SERVICE_TYPE
+from selfdrive.car import make_can_msg
+from selfdrive.boardd.boardd import can_list_to_can_capnp
+from selfdrive.swaglog import cloudlog
+
+TESTER_PRESENT_DAT = bytes([0x02, SERVICE_TYPE.TESTER_PRESENT, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0])
```
sshane, line 13: The first byte should be the subaddr here, correct?

gregjhogan: yes, but using them might complicate things quite a bit, and if it is only for toyota camera and radar maybe try counting that response addr as two matches for toyota first.

sshane:
Results from logging 4 days in master. Seems like the VWs and Subaru (only one route available) might have lots of ECUs in common with Hyundai.

```python
VOLKSWAGEN GOLF 7TH GEN: More than one brand candidate (correct is volkswagen)!
tx addr matches: {'subaru': {1968, 2017, 2016}, 'hyundai': {2016, 2017, 1968, 2001, 2004}, 'toyota': {1968, 2001, 2016}, 'volkswagen': {2016, 2017, 1810, 1813, 1879}, 'mazda': {2016, 2017}, 'nissan': {2016}}
max(): hyundai
intersection: ['hyundai', 'subaru', 'toyota']

VOLKSWAGEN GOLF 7TH GEN: Wrong brand candidate! Method returned: hyundai, actual: volkswagen
tx addr matches: {'subaru': {1968, 2017, 2016}, 'hyundai': {2016, 2017, 1968, 2001, 2004}, 'toyota': {1968, 2001, 2016}, 'volkswagen': {2016, 2017, 1810, 1813, 1879}, 'mazda': {2016, 2017}, 'nissan': {2016}}
max(): hyundai
intersection>=3: ['hyundai', 'subaru', 'toyota']

mock: Wrong brand candidate! Method returned: hyundai, actual: mock
tx addr matches: {'hyundai': {2000, 2001, 1988, 2004}, 'toyota': {1953, 2001}, 'mazda': {1840, 1798}}
max(): hyundai
intersection>=3: ['hyundai']

SUBARU IMPREZA LIMITED 2019: More than one brand candidate (correct is subaru)!
tx addr matches: {'subaru': {2016, 2017, 1862, 1927, 1968}, 'hyundai': {2016, 2017, 1988, 1968, 2000}, 'toyota': {1968, 2016}, 'volkswagen': {2016, 2017}, 'mazda': {2016, 2017}, 'nissan': {2016, 1858, 1859}}
max(): subaru
intersection: ['hyundai', 'subaru']

VOLKSWAGEN GOLF 7TH GEN: More than one brand candidate (correct is volkswagen)!
tx addr matches: {'subaru': {1968, 2017, 2016}, 'hyundai': {2016, 2017, 1968, 2001, 2004}, 'toyota': {1968, 2001, 2016}, 'volkswagen': {2016, 2017, 1810, 1813, 1879}, 'mazda': {2016, 2017}, 'nissan': {2016}}
max(): hyundai
intersection: ['hyundai', 'subaru', 'toyota']

VOLKSWAGEN GOLF 7TH GEN: Wrong brand candidate! Method returned: hyundai, actual: volkswagen
tx addr matches: {'subaru': {1968, 2017, 2016}, 'hyundai': {2016, 2017, 1968, 2001, 2004}, 'toyota': {1968, 2001, 2016}, 'volkswagen': {2016, 2017, 1810, 1813, 1879}, 'mazda': {2016, 2017}, 'nissan': {2016}}
max(): hyundai
intersection>=3: ['hyundai', 'subaru', 'toyota']

VOLKSWAGEN GOLF 7TH GEN: More than one brand candidate (correct is volkswagen)!
tx addr matches: {'subaru': {1968, 2017, 2016}, 'hyundai': {2016, 2017, 1968, 2001, 2004}, 'toyota': {1968, 2001, 2016}, 'volkswagen': {2016, 2017, 1810, 1813, 1879}, 'mazda': {2016, 2017}, 'nissan': {2016}}
max(): hyundai
intersection: ['hyundai', 'subaru', 'toyota']

VOLKSWAGEN GOLF 7TH GEN: Wrong brand candidate! Method returned: hyundai, actual: volkswagen
tx addr matches: {'subaru': {1968, 2017, 2016}, 'hyundai': {2016, 2017, 1968, 2001, 2004}, 'toyota': {1968, 2001, 2016}, 'volkswagen': {2016, 2017, 1810, 1813, 1879}, 'mazda': {2016, 2017}, 'nissan': {2016}}
max(): hyundai
intersection>=3: ['hyundai', 'subaru', 'toyota']


Checked 112 unique fingerprinted routes
Checked 29 cars
Checked 35 unique dongles
```

jyoung8607:
This wasn't the favored approach a year ago, but ... bear in mind you only *have* to solve it this way this for JDM vehicles, and they don't seem to be stepping on each other too badly, and openpilot has already onboarded all the Japanese makes I can think of.

For all other openpilot supported vehicles, and most planned or desired future vehicles, VIN WMIs should do nicely. Run VW queries only for `^((WVW)|(1V2)|(TMB)|(3VW)|(WAU)|(WVG)|(LFV)).{14}$` (haven't verified this list recently). Never run VW queries for anything else. Same down the line for all other brands. Problem solved fully and forever.

Can do it for the Japanese makes outside Japan too, if you get back a 17-character VIN and you feel so inclined. I think the Toyota WMI list is absurdly long due to their manufacturing org structure, but you can do that big data thing you do.

gregjhogan:
Yeah matching 3 or more isn't going to work, but even if it only narrows it down to two candidates at some point running those two in serial is going to be a lot faster than attempting all brands in parallel as we add more brands.  I also don't think this is fragile if we accept it is fine to not guarantee a single brand.

gregjhogan:
another idea would be to pick an ecu per brand (perhaps the power steering motor ecu) and query all brands for a firmware version of that address list first (perhaps using tester present first to filter the list of addrs down to just one or two addrs) then fuzzy match on just that single ECU to determine the brand

this approach should still work if we stop fingerprinting through the OBD port at some point (may no longer be able to get the VIN), and it seems reasonable to go into dashcam mode if we don't get a single brand from this brand filtering process

it isn't ideal, but may be the best compromise to get brand growth under control

sshane:
From all the fingerprint events (not cached) on master since logging was added, no car ever returned the wrong brand taking the max. For each platform below is a count of how many times it brand filtered successfully:

```python
{'HONDA CIVIC (BOSCH) 2019': 9, 'HONDA INSIGHT 2019': 12, 'HYUNDAI SONATA 2020': 13, 'TOYOTA COROLLA TSS2 2019': 48, 'LEXUS RX HYBRID 2017': 2, 'TOYOTA RAV4 2017': 8, 'HONDA CIVIC 2016': 11, 'KIA EV6 2022': 17, 'TOYOTA HIGHLANDER HYBRID 2020': 10, 'KIA OPTIMA SX 2019 & 2016': 7, 'HYUNDAI ELANTRA 2021': 8, 'TOYOTA CAMRY HYBRID 2021': 6, 'HONDA ACCORD 2018': 10, 'TOYOTA HIGHLANDER HYBRID 2018': 4, 'LEXUS RC 2020': 7, 'HYUNDAI KONA HYBRID 2020': 8, 'TOYOTA COROLLA HYBRID TSS2 2019': 9, 'TOYOTA PRIUS 2017': 22, 'HYUNDAI ELANTRA HYBRID 2021': 6, 'TOYOTA AVALON 2022': 3, 'HONDA PILOT 2017': 7, 'TOYOTA PRIUS TSS2 2021': 9, 'HYUNDAI PALISADE 2020': 13, 'TOYOTA RAV4 HYBRID 2022': 11, 'HYUNDAI SANTA FE 2019': 5, 'TOYOTA HIGHLANDER 2017': 6, 'VOLKSWAGEN ATLAS 1ST GEN': 7, 'TOYOTA RAV4 HYBRID 2017': 4, 'VOLKSWAGEN GOLF 7TH GEN': 8, 'TOYOTA RAV4 HYBRID 2019': 8, 'COMMA BODY': 1, 'HONDA CR-V 2017': 2, 'NISSAN LEAF 2018 Instrument Cluster': 3, 'mock': 2, 'SUBARU IMPREZA LIMITED 2019': 1, 'MAZDA CX-5 2022': 1}
```

Will do some testing to figure out how robust this is to missing some ecus.

sshane:
Out of 689 fingerprint events so far, I tried random removing an ECU response to see how that would affect the final filtering, it seems pretty fragile.

Just removing one ECU, there was never an instance where we successfully returned the correct brand, and the maximum number of bad brand filtering events was 18 instances out of the 689. I haven't tried adding an ECU yet, I'll do that next.

sshane:
Greg and I talked and we think the safest way to do this is to fingerprint after each brand's request, with the brand requests ordered by number of matching ecus. It might be different to how we do this now, but should be the safest way in case a car doesn't respond to tester present properly.

sshane:
So far from 1,659 segments (only first segs) on master, across 57 platforms and 125 unique dongles, there's no instances where if we get any responding ECUs, the most likely brand is incorrect. And now since this PR orders the brands instead of creating a sub-set to query, in the rare case a car is an outlier and brand filtering would leave out the correct brand, we just instead query later (potentially the very next brand in the ordered list).

Final PR Notes:
- FW version responses are now logged with brands, so two responses should not overwrite themselves (no ordering problem)
- This PR does not remove brands to query, only order and break early if we get an exact match.
- If no exact match, then we try to fuzzy fingerprint as normal (after querying all brands)

Some examples of this in action: https://pastebin.com/W4VJBYWe

All the cars I looked at: https://pastebin.com/HK0qRbQN

PR #23320, opened 2021-12-27
PlotJuggler: Parse routes from connect

sshane:
This makes it a bit easier to start inspecting data when copy/pasting a route name from comma connect. Instead of dealing with an inconspicuous error message and then manually fixing the route name for juggle.py, this PR parses the segment number correctly from the pasted string.

### DIFF
--- a/tools/plotjuggler/juggle.py
+++ b/tools/plotjuggler/juggle.py
@@ -121,8 +121,6 @@ if __name__ == "__main__":
   else:
     route = DEMO_ROUTE if args.demo else args.route_name.strip()
     if route.count("--") > 1:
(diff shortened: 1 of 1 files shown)


pd0wm:
@adeebshihadeh What do you think of bringing over (part of) `CanonicalRouteName` and CanonicalRouteSegmentName` from `xx`? We're implementing this parsing in a bunch of places.

pd0wm:
Ah, that's already in progress in this PR https://github.com/commaai/openpilot/pull/23263 :). I'm closing this one.

sshane:
Ah didn't see that PR, nice!

PR #23321, opened 2021-12-27
replay/logreader: handle abort in load

deanlee:


pd0wm:
When is this needed? Does decompressing ever take that long?

deanlee:
we need to abort loading asap  when seeking to a new segment or exiting replay to reduce unnecessary CPU consumption and waiting time (framereader has handled the abort, but logreader has not). decompressing is the most time-consuming operation in logreader::load, especially on slower devices, it may take about 500ms to finish.and the loop read events may only take tens of milliseconds

PR #23322, opened 2021-12-27
PC: dynamically set PYTHONPATH from real openpilot path

sshane:
As a person who installs openpilot at `~/openpilot/openpilot` I find myself changing the openpilot_env.sh file often to make sure my PYTHONPATH is correct for the openpilot modules. This should let you install openpilot wherever you want!

### DIFF
--- a/tools/openpilot_env.sh
+++ b/tools/openpilot_env.sh
@@ -1,5 +1,5 @@
 if [ -z "$OPENPILOT_ENV" ]; then
-  OP_ROOT="$(dirname "$(dirname "$(realpath -s "$0")")")"
+  OP_ROOT="$(dirname "$(dirname "${BASH_SOURCE[0]}")")"
   export PYTHONPATH="$OP_ROOT:$PYTHONPATH"
   export PATH="$HOME/.pyenv/bin:$PATH"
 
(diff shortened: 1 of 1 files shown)


tools/openpilot_env.sh
```diff
@@ -1,5 +1,6 @@
 if [ -z "$OPENPILOT_ENV" ]; then
-  export PYTHONPATH="$HOME/openpilot:$PYTHONPATH"
+  OP_ROOT="$(dirname "$(dirname "${BASH_SOURCE[0]}")")"
```
pd0wm, line 2:
```suggestion
  OP_ROOT="$( cd $(dirname "$(dirname "${BASH_SOURCE[0]}")") && pwd )"
```

Think you need this to get the absolute path.

sshane: Both work for me on Ubuntu 20.04 opening a new terminal window, is this suggestion safer?

pd0wm: Not sure how it works for you. For me the `OP_ROOT` is jus equal to `.`

sshane: Oh if you print `OP_ROOT` when sourcing from your .bashrc it gets the absolute path, but when you source or run the script outside `.bashrc` it's relative.

PR #23324, opened 2021-12-27
replay: add flag REPLAY_FLAG_FULL_SPEED to play at full speed

deanlee:
press `x` to play at full speed
press `x` again to play at normal speed.

https://user-images.githubusercontent.com/27770/147486315-ef37a566-f276-412d-a21b-9fbf3d2ad43d.mp4

### DIFF
--- a/selfdrive/ui/replay/main.cc
+++ b/selfdrive/ui/replay/main.cc
@@ -69,6 +69,14 @@ void keyboardThread(Replay *replay_) {
       replay_->seekTo(-10, true);
     } else if (c == 'G') {
       replay_->seekTo(0, true);
+    } else if (c == 'x') {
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
Cool, but what's the use case for this?

deanlee:
I think we can add this capability to ui replay. let use choose a fast speed

pd0wm:
Does it makes sense to play at something like 5x the speed (or slower if your computer can't handle it)? This would make playback a bit more stable.

deanlee:
Currently, the speed is automatically adjusted according to the decoding speed of the video. I think this is more reasonable than specifying a speed.

deanlee:
We may not be able to reach 5x speed at all, especially on c2/c3

deanlee:
We may not be able to reach 5x speed at all, especially on c2/c3

Issue #23285, opened 2021-12-21
thermald: refactor into multiple services


adeebshihadeh:

BBBmau:
When it comes to refactoring, are we wanting to treat each service inside thermald_thread as its own function that we call so that it's much cleaner inside thermald_thread? I'm seeing that we can accomplish this after seeing how much is in the function already but would like to know if this is the right approach.

PR #23325, opened 2021-12-27
add primeRedirected to deviceState

deanlee:
`params.getBool("PrimeRedirected") `will be called at 20hz `if last_ping == 0`.  this will affect the performance of  UI.


adeebshihadeh:
We've been discussing breaking up thermald recently, but we're not exactly sure how it'll look yet. Maybe for this, we make athena its own packet instead of using params? Let's discuss this in #23285.

PR #23326, opened 2021-12-27
UI: run set_brightness in QtConcurrent

deanlee:
~use std::future to control the life time of the thread and avoid accidentally starting multiple threads to set brightness.~
move to thread

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -3,6 +3,8 @@
 #include <cassert>
 #include <cmath>
 
+#include <QtConcurrent>
+
 #include "common/transformations/orientation.hpp"
 #include "selfdrive/common/params.h"
 #include "selfdrive/common/swaglog.h"
@@ -290,9 +292,11 @@ void Device::updateBrightness(const UIState &s) {
   }
 
   if (brightness != last_brightness) {
-    std::thread{Hardware::set_brightness, brightness}.detach();
+    if (!brightness_future.isRunning()) {
+      brightness_future = QtConcurrent::run(Hardware::set_brightness, brightness);
+      last_brightness = brightness;
+    }
   }
-  last_brightness = brightness;
 }
 
 bool Device::motionTriggered(const UIState &s) {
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Generally, there's a simpler way to write a PR that looks like "XYZ: use more C++". In this case, why don't we just spin up another thread that sets the brightness at like 10Hz like we used to pre-Qt? A simple thread that sets the brightness and sleeps in a loop is way simpler than: 
``` c++
    if (!brightness_future.valid() || brightness_future.wait_for(0ms) == std::future_status::ready) {
      brightness_future = std::async(std::launch::async, Hardware::set_brightness, brightness);
      last_brightness = brightness;
    }
```

deanlee:
use QThread to facilitate moving updateWakefulness to thread if needed.

adeebshihadeh:
This is a massive diff to cleanup the original stated issue of "multiple brightness threads can be created". Has the goal changed? If not, why move the whole device update to a thread? If this is merged as is, I can easily see a future PR to refactor this again and "simplify" everything into a single thread, a la [Chesterton's Fence](https://wiki.lesswrong.com/wiki/Chesterton%27s_Fence).

deanlee:
There may be no simple way to spin up another thread to update brightness because SubMaster and UIState are not thread safe.
It may be better to use QFuture and QtCurrent at present, this can avoid starting multiple threads,and eliminate the overhead of creating thread

selfdrive/ui/ui.cc
```diff
@@ -255,7 +257,9 @@ Device::Device(QObject *parent) : brightness_filter(BACKLIGHT_OFFROAD, BACKLIGHT
 }
 
 void Device::update(const UIState &s) {
-  updateBrightness(s);
+  if (s.sm->frame % 2 == 0) {
```
pd0wm, line 260: Why was the frequency changed? This messes with the `brightness_filter` time constant.

pd0wm:
If you revert the panda changes I'll merge.

deanlee:
rebased

PR #23327, opened 2021-12-28
Add comment about vEgoStarting to interfaces

sshane:
Do we want an assert after `CarInterface.get_params()` in car_helpers.py too or is this good?

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -89,7 +89,7 @@ class CarInterfaceBase(ABC):
     ret.startingAccelRate = 3.2 # brake_travel/s while releasing on restart
     ret.stoppingDecelRate = 0.8 # brake_travel/s while trying to stop
     ret.vEgoStopping = 0.5
-    ret.vEgoStarting = 0.5
+    ret.vEgoStarting = 0.5  # needs to be >= vEgoStopping to avoid state transition oscillation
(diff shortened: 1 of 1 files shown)


jyoung8607:
> Do we want an assert after `CarInterface.get_params()` in car_helpers.py too or is this good?

For so long as those conditions interact the way they do, I think an assert would be a good idea. I tripped over the oscillation condition while working on #22963. It shouldn't be permitted to happen and it's not obvious what's happening when it does.

sshane:
Yeah it can be confusing, but when `vEgoStarting` is less than stopping, it's possible for both the starting condition and stopping condition to be true in longcontrol.py, so it oscillates between them both.

For example if `v_target_future = 0.75`, then `v_target_future < (CP.vEgoStopping := 1.0)` for stopping and `v_target_future > (CP.vEgoStarting := 0.5)` for starting are both true for this future target speed (logic in `long_control_state_trans` function`).

pd0wm:
Can we fix the state machine so this oscillation doesn't happen?

sshane:
I believe it would still oscillate, just between stopping and PID, rather than starting. I'll try testing a fix in #23333

pd0wm:
Merging this now. Once we fix the oscillation we should remove the comment again.

sshane:
Sounds good!

PR #23328, opened 2021-12-28
Longcontrol: remove starting state

sshane:
I don't see a point in having a starting state which "releases brake," when it's an acceleration request. PID can do the same, and we don't have to wait for the starting code to move the `output_accel` to an arbitrary number only for it to jump anyways when PID control takes over. In Toyota stock ACC, the accel request jumps to whatever their controller wants immediately (there's some weirdness if you continually press the resume button while stopped as I did in one of the test cases below though). This means we should also start a bit earlier as well (not *that* noticeable).

Here's an album of some PJ graphs, two accelerations from this PR, stock openpilot, and stock Toyota TSS2 ACC: https://imgur.com/a/UXu8pfU.

Maybe there was a reason to have the state in the past (or for other makes?), but why not hand off control to PID as soon as possible if we don't care about smoothly ramping output accel which doesn't make that much of a difference if any?

Route for this PR: `f15e3c37c118e841|2021-12-27--19-22-18--6`
Route for stock openpilot: `f15e3c37c118e841|2021-12-27--19-40-52--1`
Route for stock ACC: `f15e3c37c118e841|2021-12-27--19-30-29`


HaraldSchafer:
Yeah I like this. Will double check a couple things and then get back to this.

HaraldSchafer:
Can you update the refs? Then I'll merge it.

adeebshihadeh:
> Can you update the refs? Then I'll merge it.

External contributors can't update the refs, since they need write access to the azure bucket.

adeebshihadeh:
Let's also make sure any cereal depredations are done too.

sshane:
Opened a PR here: https://github.com/commaai/cereal/pull/238

HaraldSchafer:
Cool, I'll update, use cereal and merge

selfdrive/car/honda/hondacan.py
```diff
@@ -53,7 +53,7 @@ def create_acc_commands(packer, enabled, active, accel, gas, idx, stopping, star
   accel_command = accel if active else 0
   braking = 1 if active and accel < min_gas_accel else 0
   standstill = 1 if active and stopping else 0
-  standstill_release = 1 if active and starting else 0
+  standstill_release = 1 if active else 0
```
sshane, line 56: Maybe change to pid? It looks like both standstill and standstill_release will be 1 when stopped. But no clue how Honda works here

HaraldSchafer: Hmm, yeah changed to not stopping, not sure how it behaves. Will have to test before we merge.

PR #23329, opened 2021-12-28
LanePlanner: correct path offset for TICI

sshane:
The model's path predictions are in the perspective of the EON/C2's left camera of -0.06 m, however with TICI it's treated as it's centered, but it should be offsetting the path extra 0.06 m to the right, this might be causing some minor left hugging.

Even if offsetting doesn't fully offset the path due to the model behaving unexpectedly with path offsets, since they were added at all I think getting it right is a good idea.

Or with the smaller offset are we sacrificing hugging for less path oscillation if the model behaves unexpectedly with path offsets?


HaraldSchafer:
I should have added comments to make this clear, but it is the way it is intended now. There are pretty complex effects at work. We look at the data from the model to see the average prediction difference between the model path from the TICI and C2 and that's how we get the current model path offsets.

Eventually we will input the camera offset into the model and it will all be a lot cleaner :).

Issue #23302, opened 2021-12-23
openpilot stops too far back from lead on incline


gregjhogan:
openpilot seems to consistently stop further back on an incline compared to flat ground
example: `5decac85ab437ed1\|2021-12-22--12-29-05--16`
first engagement it stops too far back, then next 3 I try to get it to creep forward
![image](https://user-images.githubusercontent.com/4112046/147296852-3bb4cc2d-3518-4671-a820-ccd504641bb3.png)
once stopped too far back, I see it oscillates between the stopping (longControlState==2) and starting (longControlState==3) state @ 50 Hz
![image](https://user-images.githubusercontent.com/4112046/147297915-930f72c3-3dd2-4b28-ae58-78ba6b349417.png)

sshane:
I believe the stopping/starting oscillation is due to `vEgoStopping` being greater than `vEgoStarting`, which will always lead to a state transition oscillation due to the logic in longcontrol.py (assuming this car is a Hyundai, the route isn't public to check out).

pd0wm:
Think I added the high vEgoStopping because the acceleration started oscillating a lot around the set point at low speeds. But maybe we should check if that's still the case, since there probably have been a bunch of changes since then.

PR #23333, opened 2021-12-28
Longcontrol: fix control state oscillation

sshane:
Just a draft/test PR that should fix one of the issues in #23302. The previous conditions never actually checked if the plan was to accelerate or decelerate, just if the future target speed was within range, so you could run into state oscillations if `vEgoStarting` was lower than `vEgoStopping`, making both starting and stopping conditions True at once.

This should fix both cases where oscillation can occur: braking to 0 and accelerating from 0. Wrote in 5 minutes so it needs some testing which I'll get data for soon!


sshane:
Route with test `f15e3c37c118e841|2021-12-28--14-36-17` segments 10-11. Seeing no oscillation with `vEgoStarting` at 0.2 and `vEgoStopping` higher at 0.5.

pd0wm (approved):

PR #23336, opened 2021-12-30
ui/get_path_length_idx:  line_x[i] shoud be less than or equal to path_height

deanlee:
the condition should be less than or equal to `path_height`,  otherwise the last valid point may be dropped incorrectly.

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -36,7 +36,7 @@ static bool calib_frame_to_full_frame(const UIState *s, float in_x, float in_y, 
 static int get_path_length_idx(const cereal::ModelDataV2::XYZTData::Reader &line, const float path_height) {
   const auto line_x = line.getX();
   int max_idx = 0;
-  for (int i = 1; i < TRAJECTORY_SIZE && line_x[i] < path_height; ++i) {
+  for (int i = 1; i < TRAJECTORY_SIZE && std::islessequal(line_x[i], path_height); ++i) {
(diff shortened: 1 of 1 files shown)


deanlee:
the condition should be less than or equal to `path_height`,  otherwise the last valid point will be dropped incorrectly.

selfdrive/ui/ui.cc
```diff
@@ -36,7 +36,7 @@ static bool calib_frame_to_full_frame(const UIState *s, float in_x, float in_y,
 static int get_path_length_idx(const cereal::ModelDataV2::XYZTData::Reader &line, const float path_height) {
   const auto line_x = line.getX();
   int max_idx = 0;
-  for (int i = 0; i < TRAJECTORY_SIZE && line_x[i] < path_height; ++i) {
+  for (int i = 1; i < TRAJECTORY_SIZE && std::islessequal(line_x[i], path_height); ++i) {
```
pd0wm, line 39: Why `std::islessequal` instead of `<=`?

deanlee: it's not very safe to do floating comparison with ==. the right way is `std::abs(a, b) < std::numeric_limits<float>::epsilon`. or use std::islessequal in this case.

deanlee: I was worried that  the built-in operator<= may raise FE_INVALID if line_x[i] is NaN. but it seems unnecessary for now.

PR #23338, opened 2021-12-31
Longitudinal planner: set accel to 0 when not engaged

sshane:
When engaging (or using gas pedal if on a fork) after accelerating or decelerating quickly, the planner commands close to `aEgo` for a short while, delaying following the true plan. An example is if you accelerate into a lead and engage (or again use gas pedal), the car continues to accelerate for a short while even though we're closer than `T_FOLLOW` and should be slowing down.

Engaging after accelerating:
- Stock openpilot: `f15e3c37c118e841|2021-12-30--20-17-23--1` (took 1.62 seconds from engaging to reach negative acceleration)
- This PR: `f15e3c37c118e841|2021-12-30--20-12-59--1` (took 0.55 seconds to reach negative acceleration, though initial accel was much lower. Started requesting brake immediately)

Left side is stock, right side is this PR. With stock you can see after I take my foot off the gas and engage, openpilot requests positive acceleration even though the plan clearly shows it wants to decelerate (from the `longitudinalPlan.speeds` signal) and so the car further accelerates into the lead.
 
Screenshot from 2021-12-31 00-11-04


sshane:
So while this lets us follow the plan much quicker when at high speeds after engaging, at low speeds since we always start at 0 m/s/s feedforward, [the stock ACC system always wants to brake](https://i.imgur.com/wDsK1M3.png). With stock openpilot, this wasn't an issue if `aEgo` was positive, however it was still an issue if you engage shortly after braking, as `aEgo` would be negative and we'd be requesting it.

I'll make this a draft for now, however I experimented with smoothing in `aEgo` to `a_desired` when not active, and it seems to fix both issues of braking when engaging at a stop, no matter what our acceleration is/was, as well as following the plan much earlier at high speeds. [Here's a graph](https://i.imgur.com/YmY92Mx.png) and a route: `f15e3c37c118e841|2021-12-31--03-06-31--2`. You can see how we don't brake at all engaging at zero, and the acceleration is smooth and consistent!

sshane:
Both methods here potentially add different issues, so I'll close it for now and probably open an issue.

- Setting `a_desired` to 0.0 meant we follow the plan a bit earlier in the specific acceleration case detailed above, however with Toyota ACC it meant that we always start at 0 requested acceleration at low speeds, so the car always braked for under a second before accelerating.
- Smoothing in `a_ego` to `a_desired` means that if the set speed is far away from the actual speed and we set the new speed, we have the same issue as before this PR, except the plan's residual acceleration is causing us to accelerate or decelerate instead of `a_ego`.

sshane:
@pd0wm Going to revert to setting to 0 m/s/s. The issue is that if we engage shortly after starting from a stop, aEgo is positive but we'll start requesting 0 so Toyota follows the request (correctly) by braking until feedforward raises

PR #23339, opened 2021-12-31
ui: set ignition to false if no panda

deanlee:


pd0wm:
This is on purpose. If you loose connection to the panda, you don't want to go offroad so you can still alert the user.

PR #23341, opened 2021-12-31
Longitudinal planner: make v_desired a FirstOrderFilter

sshane:
With `v_desired` initialized at 0 and updating the value to any number, there's a 1.22% difference between current vs. with FOF. If it's not worth it, feel free to close. I just thought it might be nice to use.


HaraldSchafer:
I like this change, but it is pretty ugly to have v_desire to be a filter object and a_desired be a float. Can you keep v_desired a float and make a new v_desired_filter object?

PR #23346, opened 2022-01-01
LogReader: add arg to sort by time

gregjhogan:
It is often useful to iterate through the data in chronological order.

### DIFF
--- a/tools/lib/logreader.py
+++ b/tools/lib/logreader.py
@@ -13,7 +13,7 @@ from cereal import log as capnp_log
 
 # this is an iterator itself, and uses private variables from LogReader
 class MultiLogIterator:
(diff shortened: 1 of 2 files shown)


tools/lib/logreader.py
```diff
@@ -13,7 +13,7 @@
 
 # this is an iterator itself, and uses private variables from LogReader
 class MultiLogIterator:
-  def __init__(self, log_paths, wraparound=False):
+  def __init__(self, log_paths, wraparound=False, sort_by_time=False):
```
adeebshihadeh, line 16: should we get rid of wraparound? think it was only used for unlogger

pd0wm: Sounds good to me!

PR #23348, opened 2022-01-01
plotjuggler: fix layout param

gregjhogan:

### DIFF
--- a/tools/plotjuggler/juggle.py
+++ b/tools/plotjuggler/juggle.py
@@ -66,12 +66,12 @@ def start_juggler(fn=None, dbc=None, layout=None):
 
   extra_args = ""
   if fn is not None:
-    extra_args += f"-d {fn}"
+    extra_args += f" -d {fn}"
   if layout is not None:
-    extra_args += f"-l {layout}"
+    extra_args += f" -l {layout}"
 
-  subprocess.call(f'plotjuggler --plugin_folders {INSTALL_DIR} {extra_args}',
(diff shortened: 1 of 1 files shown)


tools/plotjuggler/juggle.py
```diff
@@ -66,12 +66,12 @@ def start_juggler(fn=None, dbc=None, layout=None):
...
   if fn is not None:
-    extra_args += f"-d {fn}"
+    extra_args += f" -d {fn}"
   if layout is not None:
-    extra_args += f"-l {layout}"
+    extra_args += f" -l {layout}"
 
-  subprocess.call(f'plotjuggler --plugin_folders {INSTALL_DIR} {extra_args}',
-                  shell=True, env=env, cwd=juggle_dir)
+  cmd = f'plotjuggler --plugin_folders {INSTALL_DIR}{extra_args}'
```
adeebshihadeh, line 73: why remove the space between INSTALL_DIR and extra_args?

adeebshihadeh:
Let's add a test case for the layouts in test_plotjugger.py

tools/plotjuggler/juggle.py
gregjhogan: because then we have a trailing space with no extra params and two spaces before the first param

gregjhogan:
feel free to take this one over if any additional changes are desired before merging

adeebshihadeh:
Hmm, I guess the way the tests are written now wouldn't catch the bug. I'll merge for now.

PR #23354, opened 2022-01-02
ui: remove fb_w fb_h, clip to +/-32767 pixels

deanlee:
The previous nanovg's clipping problem was due to the pointer may exceeds +-32767 pixels,  qt has the same restriction too. 
it's safe to use +/-std::numeric_limits<int16_t>::max() as the clip_region. and I think it's better than using window_rect +/- a hardcoded margin.


pd0wm:
Closing this since I don't see a clear benefit. Getting rid of `fb_w` and `fb_h` is not that big of a deal.

PR #23355, opened 2022-01-02
ui: update lead's z position if modelV2 has been updated after startup.

deanlee:


pd0wm:
Did you observe this being a problem? As far as I can tell radard doesn't output leads before the model is ready.

selfdrive/ui/ui.cc
```diff
@@ -112,7 +112,7 @@ static void update_state(UIState *s) {
   }
   if (sm.updated("radarState")) {
     std::optional<cereal::ModelDataV2::XYZTData::Reader> line;
-    if (sm.rcv_frame("modelV2") > 0) {
```
deanlee, line 115: we added this if statement before because we encountered  segfalut without it. so I think the new if statement is more logically correct and safe.

adeebshihadeh: why is line optional for update_leads? should we just skip the call if there is no modelV2 yet?

PR #23356, opened 2022-01-02
ui/update_model: do not recalculate the max_idx If there is no lead

deanlee:


pd0wm:
This was the source of all the extra CPU usage. This whole PR doesn't make sense now that I'm looking at it again.

adeebshihadeh:
@deanlee it would be great to come up with some tests to ensure UI CPU usage is pretty constant. we already have nice abstractions in UI, so it should be pretty easy to write some fuzzing tests

PR #23357, opened 2022-01-02
Toyota: gate pedal behind long

cydia2020:


pd0wm:
Any specific reason for this change?

cydia2020:
From what I can tell, if someone has their DSU plugged in with a pedal (this scenario is very rare but there is **nothing** to stop users from doing that), openpilot will still send pedal commands without a clear way to decelerate if the stock ACC isn't detecting the lead car. This feels like a potential safety hazard.

pd0wm:
Maybe it makes more sense to check the condition at the point where we actually command the gas. The gas interceptor *is* present, so it makes sense that the value in carParams is set.

cydia2020:
> Maybe it makes more sense to check the condition at the point where we actually command the gas. The gas interceptor _is_ present, so it makes sense that the value in carParams is set.

So should this be done from car controller instead?

pd0wm:
Yes, I think that makes the most sense.

selfdrive/car/toyota/carcontroller.py
```diff
@@ -26,7 +26,7 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
              left_line, right_line, lead, left_lane_depart, right_lane_depart):
 
     # gas and brake
-    if CS.CP.enableGasInterceptor and enabled:
+    if CS.CP.enableGasInterceptor and CS.CP.openpilotLongitudinalControl and enabled:
```
pd0wm, line 29: We would still send the CAN packets to the interceptor on line 103. Probably best to combine these two blocks.

pd0wm, line 29: It's fine to make the `interceptor_gas_cmd`. You want to gate the sending on line 103

PR #23361, opened 2022-01-03
tools: configure PYTHONPATH in pipenv

gregjhogan:
https://pipenv-fork.readthedocs.io/en/latest/advanced.html#automatic-loading-of-env

I wish I didn't have to generate the .env file, but I didn't find a way to create a .env file which automatically resolves the root directory of the venv (this won't handle directory renaming or moving)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 venv/
+.env
 .clang-format
 .DS_Store
 .tags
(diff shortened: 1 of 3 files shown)


gregjhogan:
just noticed this [commit](https://github.com/commaai/openpilot/commit/fed7cfeef064b12f76b413298ffdf71fec1f537e#diff-d3def02d69328ccf42e97ae059f342221f494c36b5e08c19458c210064ecd1dc), but this seems superior because the scope of changing the PYTHONPATH is narrower and it is shell agnostic?

adeebshihadeh:
I agree this seems preferable. Would be nice if we could automatically activate the pipenv too.

gregjhogan:
> I agree this seems preferable. Would be nice if we could automatically activate the pipenv too.

probably the best solution: https://github.com/Intersec/pyvenv-activate#manual-installation

PR #23364, opened 2022-01-03
tools: add python interface to c++ Logreader

### LINKED ISSUE #22500: Unify framereader/logreader implementations
- [ ] python interface to cpp logreader/framereader
- [ ] logreader support for processing a single segment from url without writing out any files
- [ ] framereader support for processing a single segment from url without writing out any files
- [ ] framereader support for [index video stream](https://github.com/commaai/openpilot/blob/537dff3ae0bac7ca3d40f1441297cc146dc1a9bb/tools/lib/framereader.py#L132-L144) from url without writing out any files

support from command line is not needed, this is for spinning through thousands of random segments as fast as possible without disk I/O becoming a bottleneck (or wearing out SSD drives)

### LINKED ISSUE #23358: Make RobustLogReader fast
https://github.com/commaai/openpilot/blob/master/tools/lib/robust_logreader.py can deal with corrupted logs, but is very slow and needs an external call to `bzip2recover`. Our C++ replay handles corrupted logs in a much more elegant way.

deanlee:
resolve: https://github.com/commaai/openpilot/issues/23358,  resolve: https://github.com/commaai/openpilot/issues/22500
it's fully backward compatibility with the old python interface. and added a new function `findAll` for faster filtering events


selfdrive/test/process_replay/process_replay.py
```diff
@@ -156,7 +156,7 @@ def fingerprint(msgs, fsm, can_sock, fingerprint):
   fsm.wait_on_getitem = True
 
   # populate fake socket with data for fingerprinting
-  canmsgs = [msg for msg in msgs if msg.which() == "can"]
+  canmsgs = msgs.findAll(["can"])
```
adeebshihadeh, line 159: this is pretty nice, but can you split out new functionality into a new PR?

deanlee:
There are some other optimizations that need to be done. for example, `msgs = list(LogReader(path))` should be changed to `msgs = LogReader(path)`. This can reduce unnecessary copies, which can greatly improve performance.  I will commit these changes after this Pr is merged.

Issue #22562, opened 2021-10-14
Fix up macOS support


adeebshihadeh:
Now that WSL2 is known to work well for Windows users, we just need to fix up mac support.

- [x] fix x86 build
- [x] fix M1 build
- [x] fix setup script
- [x] get CI up again
- [x] test replay
- [x] build + test plotjuggler (https://github.com/commaai/PlotJuggler/issues/13)

bravech:
I have a build on M1 working in https://github.com/bravech/openpilot/commit/c037ac9a35108b161d48f97e8421dfd415ccaeca , but there are several issues with running the tools after building, such as: `ui` slows down after running for a few minutes and eventually crashes, `watch3` crashes with Qt-related errors, . One possible culprit may be that `clang` uses `libc++` instead of `libstdc++`, but I'm not sure if it would cause these issues.

vinperdom:
I'm having trouble building with scons with @bravech's branch. Below is the error message I receive. I have rosetta 2 installed, btw.

```
ld: warning: ignoring file /opt/homebrew/lib/libzmq.dylib, building for macOS-x86_64 but attempting to link with file built for macOS-arm64
ld: symbol(s) not found for architecture x86_64
```

jererobles:
@vinperdom are you on M1? Try setting your arch manually via flags. Checkout the troubleshooting section from tools/README.md in the PR I mentioned above.

jererobles:
@vinperdom if you have M1 then make sure to use `ARCHFLAGS="-arch arm64" ./tools/mac_setup.sh`

lukasloetkolben:
You can have try on my repo. 
I have a readme, when following the steps you should be able to build OP.
After cloning the Repository you have to make a 'git checkout 5ae276fb2ab4468736e73dd5a829493b028320df'

vinperdom:
I think I got better results with @lukasloetkolben's branch. Now I'm getting `OSError: Cannot load native module 'Crypto.Hash._MD5': Not found '_MD5.cpython-38-darwin.so'`. I'm trying to solve it but if someone already has the answer, let me know.

lukasloetkolben:
Casadi must be built from source right now.

`git clone https://github.com/casadi/casadi.git casadi
cd casadi
git checkout 3.5.5
mkdir build
cd build
cmake -DWITH_PYTHON=ON -DWITH_PYTHON3=ON ..
make -j 8
sudo make install`

smurfd:
[...]

Issue #20910, opened 2021-05-14
scons -j4 error when trying to launch openpilot outside a docker for PC simulation


AIasd:
Hi,

I am trying to run Openpilot in Carla outside of a docker by running `./launch_openpilot.sh` as well as `./bridge.py` in addition to the CARLA server.

I encountered the following error when I run  `./launch_openpilot.sh`:
```
{"event": "dirty comma branch", "version": "0.8.4", "dirty": true, "origin": "https://github.com/commaai/openpilot.git", "branch": "origin/master", "dirty_files": ":100755 100755 1ac31bcc3a54cac01cdb41b67ffb52f641f40c62 0000000000000000000000000000000000000000 M\ttools/ubuntu_setup.sh", "commit": "c4d00134ee18296ef4b498f2c1c623ffa3fda55f", "origin_commit": "c4d00134ee18296ef4b498f2c1c623ffa3fda55f"}
System time valid
starting process ui
preimporting selfdrive.locationd.calibrationd
preimporting selfdrive.controls.controlsd
./ui: 3: exec: ./_ui: not found
Manager failed to start
Traceback (most recent call last):
  File "./manager.py", line 200, in <module>
    main()
  File "./manager.py", line 175, in main
    manager_prepare()
  File "./manager.py", line 100, in manager_prepare
    p.prepare()
  File "/home/zhongzzy9/openpilot/selfdrive/manager/process.py", line 213, in prepare
    importlib.import_module(self.module)
  File "/home/zhongzzy9/.pyenv/versions/3.8.5/lib/python3.8/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1014, in _gcd_import
  File "<frozen importlib._bootstrap>", line 991, in _find_and_load
  File "<frozen importlib._bootstrap>", line 975, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 671, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 783, in exec_module
  File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
  File "/home/zhongzzy9/openpilot/selfdrive/controls/controlsd.py", line 12, in <module>
    from selfdrive.boardd.boardd import can_list_to_can_capnp
  File "/home/zhongzzy9/openpilot/selfdrive/boardd/boardd.py", line 4, in <module>
[...]

PR #23365, opened 2022-01-03
mac: fix setup and re-enable CI

tecandrew:
* update `tools/mac_setup.sh`
* manual build & install of `casadi` to openpilot virtualenv
* replace `acados` Darwin libs with universal2 (`x86_64`/`arm64`) libs
* skips `onnxruntime` install for Darwin until the [universal2 pip package](https://github.com/microsoft/onnxruntime/releases/tag/v1.10.0) is available
* remove rust stuff

🎯 #22562


adeebshihadeh:
Nice, can you re-enable the mac build in `.gtihub/workflows/selfdrive_tests.yaml`? Fixing this should be sufficient to get that workflow running again.

tools/mac_setup.sh
```diff
@@ -14,8 +16,8 @@ brew "cmake"
 brew "git-lfs"
 brew "zlib"
 brew "bzip2"
-brew "rust"
-brew "rustup-init"
+brew "unzip"
+brew "wget"
```
adeebshihadeh, line 20: what are these for?

tools/mac_setup.sh
```diff
@@ -29,6 +31,8 @@ brew "openssl"
 brew "pyenv"
 brew "qt@5"
 brew "zeromq"
+brew "protobuf"
+brew "protobuf-c"
```
adeebshihadeh, line 35: and these?

tools/mac_setup.sh
```diff
@@ -55,15 +53,30 @@ export PATH="$PATH:/usr/local/bin"
...
   source "$ROOT/tools/openpilot_env.sh"
   echo "Added openpilot_env to RC file: $RC_FILE"
 fi
 
 # install python dependencies
-$ROOT/update_requirements.sh
+$ROOT/update_requirements.sh || true
+
+# install casadi
+echo "-- casadi manual install"
```
adeebshihadeh, line 65: do we need to do this for the x86 macs?

adeebshihadeh: actually, I'm fine with keeping as is so it has CI coverage

tools/mac_setup.sh
tecandrew: getting casadi source files. i could probably do away w wget. does osx have curl by default?

tools/mac_setup.sh
tecandrew: building casadi

tools/mac_setup.sh
tecandrew: i dont think so, but the pipfile doesnt have a way to distinguish b/t architecture except for the host OS (markers in the pipfile) afaik

tools/mac_setup.sh
adeebshihadeh: yeah, or feel free to use git

tools/mac_setup.sh
adeebshihadeh: can you leave a comment so we can get rid of this once we can pip install casadi?

.github/workflows/selfdrive_tests.yaml
```diff
@@ -69,52 +69,52 @@ jobs:
...
-  #      cd /usr/local/Cellar
-  #      new_cellar=$(ls -1)
-  #      comm -12 <(echo "$EXISTING_CELLAR") <(echo "$new_cellar") | while read pkg; do
-  #        if [[ $pkg != "zstd" ]]; then # caching step needs zstd
-  #          rm -rf "$pkg"
-  #        fi
-  #      done
-  #      comm -13 <(echo "$EXISTING_CELLAR") <(echo "$new_cellar") | tee ~/github_brew_cache_entries.txt
+  build_mac:
+   name: build macos
```
adeebshihadeh, line 73: looks like a space got removed on this indentation level

.github/workflows/selfdrive_tests.yaml
```diff
@@ -69,52 +69,52 @@ jobs:
...
+   runs-on: macos-11
```
adeebshihadeh, line 74:
```suggestion
   runs-on: macos-latest
```

tools/openpilot_env.sh
```diff
@@ -1,8 +1,20 @@
...
+  ARCH="$(uname -m)"
+  HOST="$(uname -s)"
+
   export PYTHONPATH="$OP_ROOT:$PYTHONPATH"
   export PATH="$HOME/.pyenv/bin:$PATH"
 
+  # for qt resource compiler "rcc"
+  if [[ "$HOST" == "Darwin" ]]; then
+    if [[ "$ARCH" == "arm64" ]]; then
+      export PATH="/opt/homebrew/opt/qt@5/bin:$PATH"
```
adeebshihadeh, line 12: let's  handle this in scons, since we already know the qt path there

.github/workflows/selfdrive_tests.yaml
tecandrew: gh actions looks like its working? 🤔

tecandrew:
at this point it seems good to go. i have both my intel and m1 mac able to get a working build, but ci env isn't behaving correctly? I can't seem to replicate the issue on my local machine.

did some googling around and found that `clang` doesn't like the `-rpath=<>` syntax and prefers/accepts the `-rpath,<>` instead. some googling around and I found that this bug was replicated inadvertently at least once before

* https://github.com/commaai/openpilot/issues/20910

might be a `scons` bug?

relevant: 
* https://stackoverflow.com/questions/69754783/replacing-w1-rpath-origin-with-w1-rpath-origin-in-scons
* https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-rpath

adeebshihadeh:
Can you pull in master to fix the conflicts?

tecandrew:
`pipfile.lock` needs to be regenerated to accomodate for `casadi` and `onnxruntime-gpu` skipping on Darwin. 

also, this weird `./selfdrive/ui/ui` bug is only shown on apple m1.  it also sometimes errors when closing the app (same w/ plotjuggler). apple intel has no issue.

image

.github/workflows/selfdrive_tests.yaml
adeebshihadeh: just a formatting issue, one space instead of two

tecandrew:
```
if arch == "Darwin":
  env['RPATHPREFIX'] = "-rpath "
```
works (note the whitespace). build still works on my local m1 and intel machines. strange why ci macos needs this.

https://github.com/tecandrew/openpilot/runs/4730989175?check_suite_focus=true

adeebshihadeh:
> ```
> if arch == "Darwin":
>   env['RPATHPREFIX'] = "-rpath "
> ```
> 
> works (note the whitespace). build still works on my local m1 and intel machines. strange why ci macos needs this.
> 
> https://github.com/tecandrew/openpilot/runs/4730989175?check_suite_focus=true

Nice. I'm optimizing the cache usage and dependency install time now, then I'll add that.

Issue #2106, opened 2020-08-30
OP Low Gear/High RPM issue on 2020 Lexus RX 350


rpaulhanna:
**Describe the bug**
Most notably at freeway speeds, engaging OP results in two downshifts shortly thereafter and an unnecessarily high RPM until disengaged or upon encountering a vehicle within following distance intervention.  Manual or stock cruise control driving (OP disabled) over the same stretch of road at the same speed does not result in this low gear/high RPM behavior.  As indicated by several Discord posts, this same behavior is also evident on other RX’s of earlier year models as well.

**How to reproduce or log data**
Engage OP at freeway speeds and observe downshifts/high RPMs (2500-4000) shortly thereafter.  Alternatively, accelerate to freeway speeds with OP engaged and observe high RPMs.  Disengage OP and observe the car promptly upshift and RPMs return to a normal freeway cruising range 1500-1800.

**Expected behavior**
Engaging OP or using it at freeway speeds should not result in downshifts and maintaining a highly inefficient DOUBLING of engine RPM just because OP is in use.  OP long should result in mimicking driving the car manually at speed or as the stock cruise control does on the freeway—overdrive gear and 1500-1800 RPM.

**Device/Version information**
- Dongle ID:  3a5c5101bd71ad5d
- Route:  3a5c5101bd71ad5d|2020-08-30--13-12-43
- Timestamp:  Reference multiple engagements/disengagements starting 2 minutes into the clip.
- Version:  release2 v0.7.8
- Car make/model:  2020 Lexus RX 350

**Additional context**
There appears to be a discrepancy between UI_SET_SPEED and SPEED (actual) that could be causing the problem.  @nelsonjchen developed a “set speed scaling” solution that does resolve the problem.  Reference his comment to @cecnic1989’s PR about the same issue:  https://github.com/commaai/opendbc/pull/279#issuecomment-679826539

When implementing @nelsonjchen’s solution and adjusting the speed scale factor to 0.974 (which represents the 2.6% difference between UI_SET_SPEED and SPEED on my vehicle at approximately 65 mph), the low gear/high RPM issue is completely resolved.  Reference route 3a5c5101bd71ad5d|2020-08-30--14-30-28 where over the same stretch of freeway the correct gear is maintained and the RPM stays within 1500-1800 except for a brief period of road surface incline and then it promptly upshifts and returns to the normal RPM cruising range.

[...]

PR #23367, opened 2022-01-03
Toyota: extend wheel speed factor to 2017 Lexus RX

cecnic1989:
**Description** applying the wheelSpeedFactor fix discussed in #2106 to LEXUS_RX

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -64,6 +64,7 @@ class CarInterface(CarInterfaceBase):
       tire_stiffness_factor = 0.5533
       ret.mass = 4387. * CV.LB_TO_KG + STD_CARGO_KG
       set_lat_tune(ret.lateralTuning, LatTunes.PID_B)
+      ret.wheelSpeedFactor = 1.035
 
     elif candidate == CAR.LEXUS_RXH:
       stop_and_go = True


pd0wm:
Thanks! Wasn't sure if the older model was affected too. Do you have a drive/dongle ID so I can quickly check if the factor is correct for your car too?

Do you know if the hybrid version is affected as well?

cecnic1989:
Sure, my dongle Id is 515dd251c598ea16. I just checked comma connect, and I only have three drives there. I don't think any contain highway drives to show the rev issue. Let me know if you need me to do some highway driving for this PR.

cecnic1989:
I'm not sure if the hybrid version is affected. All the community members that I've been talking to are non-hybrid RX owners.

adeebshihadeh:
The parameters for all supported RXs are very similar. We should probably just merge them.

adeebshihadeh:
Merged all the RX params in #23374.

cecnic1989:
Cheers thanks Adeeb! Yeah I’m not sure why all the other params are slightly different. It’s should be fine to merge them.

adeebshihadeh:
@cecnic1989 the tuning for your car changed slightly after merging that PR. Let me know how it feels now!

PR #23369, opened 2022-01-03
Remove unused typing import

BrianWisniewski:
Removed unused list import as recommended here:  https://lgtm.com/projects/g/commaai/openpilot/alerts/?mode=list

### DIFF
--- a/selfdrive/pandad.py
+++ b/selfdrive/pandad.py
@@ -4,7 +4,7 @@ import os
 import usb1
 import time
 import subprocess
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Hmm, why didn't any of the linters catch this?

BrianWisniewski:
Could the call to "List()" on line 111 have confused the linter?

adeebshihadeh:
Fixed in #23375. Thanks for pointing this out .

PR #23370, opened 2022-01-03
lateral_planner.py/update: follow capnp best practices

deanlee:


adeebshihadeh:
Hmm, I don't know about this one. `sm['carState']` is just a pure python dict lookup, which should be cheap. This is less readable and probably has no performance impact.

PR #23373, opened 2022-01-03
2020 Toyota Highlander Hybrid Engine fingerprint

mattbsea:
Engine Fingerprint from 2020 Toyota Highlander Hybrid Platinum


vanillagorillaa:
Please include the dongleID from the device that this firmware was taken from

mattbsea:
DongleID: b1927e255a1630fe

PR #23377, opened 2022-01-03
Joystick: learn axis min/max

incognitojam:
Some joysticks/gamepads have negative values in the axes, [-x, x], not just in the range [0, x]. These changes allow for negative values, and for both axes values to be learnt when we see a new higher value. This means the values don't have to be modified in the code when plugging in a new controller, just move the controller to the maximum positions of each axis at least once before engaging.


incognitojam:
I'd like if you didn't have to configure the steer/accel axes names too, but it probably varies a lot between controllers.

PR #23384, opened 2022-01-04
setup/spinner/text: add display timeout 

### LINKED ISSUE #23138: setup/spinner/text: add display timeout
We should abstract away the display timeout in UI to make it easy to use in the smaller standalone UIs.

whowe444:
resolves #23138 
- Moved timeout functionality into interface called Wakeable
- Removed "awake" boolean from UIState and replaced with slot/signal, appears to only be needed in request_repeater.cc, as recommended by a TODO comment
- Add interface to spinner, setup, text

Don't yet have a C2 or C3 so haven't tested on device


selfdrive/ui/ui.cc
```diff
@@ -241,22 +241,26 @@ void UIState::update() {
...
+  static float accel_prev = 0;
+  static float gyro_prev = 0;
 
-void Device::update(const UIState &s) {
-  updateBrightness(s);
-  updateWakefulness(s);
+  bool accel_trigger = abs(s.scene.accel_sensor - accel_prev) > 0.2;
+  bool gyro_trigger = abs(s.scene.gyro_sensor - gyro_prev) > 0.15;
 
-  // TODO: remove from UIState and use signals
```
whowe444, line 255: Also addressed this TODO

adeebshihadeh:
Thanks, but we've decided to instead implement this in AGNOS using the window manager.

PR #23388, opened 2022-01-04
controlsd: fix vCruise jumping on pcmCruise cars

sshane:
Previous behavior:
- When engaging on Toyota with `pcmCruise`, [`vCruise` goes to `vEgo` with some clipping for one frame](https://i.imgur.com/HUv47zS.png), then back to the true set speed.

This PR:
- `vCruise` doesn't jump when engaging (only for `pcmCruise` cars).


selfdrive/controls/controlsd.py
```diff
@@ -463,7 +463,8 @@ def state_transition(self, CS):
           else:
             self.state = State.enabled
           self.current_alert_types.append(ET.ENABLE)
-          self.v_cruise_kph = initialize_v_cruise(CS.vEgo, CS.buttonEvents, self.v_cruise_kph_last)
+          if not self.CP.pcmCruise:
```
sshane, line 466: Prevents vCruise skipping when engaging for `pcmCruise` cars only.

pd0wm:
Looks good. Indeed should have been behind the same check as `update_v_cruise`

PR #23389, opened 2022-01-04
Subaru FPv2 values for currently supported models

martinl:
**Description** Here are Subaru FW values from subaru-PR-test branch for each model currently supported in upstream

**Verification** FW values have been submitted and verified working by discord users mentioned in comments

How should we handle adding the FW values - do you need dongle id / route for each car?

### DIFF
--- a/selfdrive/car/subaru/values.py
+++ b/selfdrive/car/subaru/values.py
@@ -83,12 +83,6 @@ FINGERPRINTS = {
 
 FW_VERSIONS = {
   CAR.ASCENT: {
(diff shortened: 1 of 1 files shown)


pd0wm:
Can you remove the comments?

@adeebshihadeh I assume we're ok with accepting a big batch at first?

adeebshihadeh:
I’m cool with it. Let’s just remove the comments on the values.

martinl:
I removed the comments

adeebshihadeh:
Thanks @martinl!

PR #23390, opened 2022-01-04
SConstruct: add -DKJ_NDEBUG to CCFLAGS

deanlee:
This is one of the capnp's Tip & best practices : https://capnproto.org/cxx.html

> The Cap’n Proto library has lots of debug-only asserts that are removed if you #define NDEBUG, including in headers. If you care at all about performance, you should compile your production binaries with the -DNDEBUG compiler flag. 

https://github.com/capnproto/capnproto/blob/09b91a303073a5c6d7ea8409f3fddf6686b03d76/c%2B%2B/src/kj/debug.h#L103-L105

we can use -DKJ_NDEBUG to disable these asserts

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -173,6 +173,10 @@ if arch != "Darwin":
 cflags += ["-DSWAGLOG"]
 cxxflags += ["-DSWAGLOG"]
 
+# Disable capnp's asserts
+cflags += ["-DKJ_NDEBUG"]
+cxxflags += ["-DKJ_NDEBUG"]
+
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
We build capnp and kj with `O2`, so these should already be disabled right? https://github.com/commaai/agnos-builder/blob/7b51a1a69772c76a43c9fd6c9fa99e1f46aac4ba/userspace/compile-capnp.sh#L9

deanlee:
I'll confirm again

PR #23400, opened 2022-01-04
calibrationd: follow capnp best practices

deanlee:


adeebshihadeh:
No need to do this where it will have basically no performance impact.

deanlee:
just think it's better to follow the same rules

adeebshihadeh:
I agree in general, but it's not exactly natural to write like this, so I think it's an exception.

PR #23404, opened 2022-01-05
Toyota: Cancel cruise without chime

Smartype:

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -3,7 +3,8 @@ from common.numpy_fast import clip, interp
 from selfdrive.car import apply_toyota_steer_torque_limits, create_gas_interceptor_command, make_can_msg
 from selfdrive.car.toyota.toyotacan import create_steer_command, create_ui_command, \
                                            create_accel_command, create_acc_cancel_command, \
-                                           create_fcw_command, create_lta_steer_command
+                                           create_fcw_command, create_lta_steer_command, \
+                                           create_acc_cancel_command_2
 from selfdrive.car.toyota.values import CAR, STATIC_DSU_MSGS, NO_STOP_TIMER_CAR, TSS2_CAR, \
                                         MIN_ACC_SPEED, PEDAL_TRANSITION, CarControllerParams
 from opendbc.can.packer import CANPacker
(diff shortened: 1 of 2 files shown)


alancyau:
I can confirm this fix works on my TSS2 Prius. Great work @Smartype!

selfdrive/car/toyota/carcontroller.py
```diff
@@ -95,10 +96,13 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
       if pcm_cancel_cmd and CS.CP.carFingerprint in [CAR.LEXUS_IS, CAR.LEXUS_RC]:
```
sshane, line 96: If `create_acc_cancel_command_2` works on all cars (the Lexi), we can replace `create_acc_cancel_command` and just do it here

pd0wm:
Can you explain some more how this works?

On first glance it looks like this simulates the first stage of an AEB to force a cancel. That doesn't sound like a good solution, especially on cars where the DSU is not unplugged (or TSS2).

 - On TSS2 cars or cars with DSU plugged in this might be interacting with the `PRE_COLLISION` message sent by the stock system, and cause messages from stock to be dropped due to a counter mismatch.
 - On cars with DSU unplugged this might interact with the currently static message we are sending, and also cause reliability issues due to counter issues. However, that is fixable.

Smartype:
I don't know why this worked. Only toyota knows why. As there is not public doc about how it works. 


For the counter: if the car has a strict counter check, why this works every time with any counter value(the frame var)? Further more, if it does accept my counter value, why not accept the stock one? (There could be a chance that the injected msg causes an error state and as a result, acc got disengaged) 

Some tests can be down:
1. Reuse the stock counter, or counter + 1, if this really matter
2. Try trigger aeb alarm after this msg, see if it break/isolate the aeb module(dsu or radar)

Right now, we'd check if this hack work on other Toyotas, if not, we do not have to spent much time perfecting it.

Smartype:
In my opinion, counter here indicates the dsu is alive and updating. There is no strict check for "next expected value". 

And about stock aeb, i think it can be done in panda, when aeb is active, panda can switch to passthrough mode.

nelsonjchen:
Just want to leave this anecdoet I posted in the issue about this branch here:

071ba9916a1da2fa|2022-01-04--20-12-54--5 with @Smartype's changes on a 2020 Corolla SE Hatchback. While does deactivate without a chime, mashing the brake pedal down (e.g. enough to activate brake hold) at a standstill causes a cruise control malfunction. Needs to restart car to re-enable cruise control.

sshane:
@nelsonjchen could that be because while pcm_cancel_cmd is True, CC isn't sending 0x343 and the car faults? While stopped, sometimes the stock system takes a while to disengage under certain circumstances and pcm_cancel_cmd may be true for longer than it should

selfdrive/car/toyota/carcontroller.py
```diff
@@ -95,10 +96,13 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
...
         can_sends.append(create_acc_cancel_command(self.packer))
       elif CS.CP.openpilotLongitudinalControl:
-        can_sends.append(create_accel_command(self.packer, pcm_accel_cmd, pcm_cancel_cmd, self.standstill_req, lead, CS.acc_type))
-        self.accel = pcm_accel_cmd
+        if pcm_cancel_cmd:
+          can_sends.append(create_acc_cancel_command_2(self.packer, frame))
+        else:
+          can_sends.append(create_accel_command(self.packer, pcm_accel_cmd, pcm_cancel_cmd, self.standstill_req, lead, CS.acc_type))
+          self.accel = pcm_accel_cmd
```
cydia2020, line 103:
```
      if pcm_cancel_cmd:
        can_sends.append(create_acc_cancel_command(self.packer, frame))

      if CS.CP.openpilotLongitudinalControl:
        can_sends.append(create_accel_command(self.packer, pcm_accel_cmd, self.standstill_req, lead, CS.acc_type, fcw_alert))
        self.accel = pcm_accel_cmd
```

I feel like this is a better way of handling this.

cydia2020:
Nice work 🎉, I was able to test this on my Prius 2018 (TSS-P) and I can confirm this works fine overall.

However, this doesn't work when the CAN speed is zero, this is regardless of the car's cruise state whether it'd be standstill, creeping, or resumed. (This is also why during my testing I ruled out this signal as all my testing were conducted with the vehicle stopped).

There are a few options to tackle this though.

- One is to send the `CANCEL_REQ` (a.k.a. create an error) command that we're currently using when the vehicle is stopped. This will create the error chime.
- Another one is to make the car think there's no lead by setting `MINI_CAR` to `0` when we want to cancel at low speed. This will create 4 consecutive beeps, and will make the dash display "NO CLOSE LEAD CAR".

adeebshihadeh:
Unfortunately, it seems this solution isn't going to work. It uses the AEB message in a way that isn't well understood and doesn't work in all conditions.

PR #23407, opened 2022-01-05
Remove Ascent FPv1

martinl:

**Description** Remove Ascent FPv1 which is known to cause identification issues for Impreza/Crosstrek since FPv1 are almost identical

**Verification** Tested on subaru-PR-test branch which does not have any Subaru FPv1 fingerprints (and has all known Ascent FPv2 FW versions)

I'll add Ascent route once it is available


adeebshihadeh:
Can you make a PR to remove all the pre-global can fingerprints? I’m fine with removing them all now since they’re dashcam only.

PR #23408, opened 2022-01-05
setup: perform all pyenv setup in a single place

gregjhogan:
The `pyenv` env setup isn't part of openpilot so I don't think it should be in `openpilot_env.sh`

also, the whole `pyenv init --path` thing seems to have only been an issue running `pyenv update` the first time after they transitioned to recommending ~/.profile (I am not seeing any issue on a fresh install or updating from a recent version)


update_requirements.sh
```diff
@@ -3,11 +3,21 @@
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 cd $DIR
 
+RC_FILE="${HOME}/.$(basename ${SHELL})rc"
```
adeebshihadeh, line 6: this isn’t always the case. mac defaults to ~/.bash_profile

gregjhogan: .bash_profile normally calls .bashrc so I thought this is a superset, didn't realize mac doesn't do this :(

adeebshihadeh:
Still want this?

gregjhogan:
yeah I will finish it up, it seems weird to me to have pyenv environment config in `openpilot_env.sh` since it applies much wider than just openpilot

gregjhogan:
This should be good now once CI passes (assuming CI sufficiently tests the setup)

pd0wm:
The openpilot env still sets some variables on Mac. Why do you want to remove it?

PR #23409, opened 2022-01-05
CameraView: sync pbo with glFence instead of calling glFinish

pd0wm:


adeebshihadeh:
@deanlee are you able to reproduce the random high CPU usage we're seeing on the comma two now? It started happening after merging this, but reverting hasn't fixed it.

deanlee:
it's wired, the recent cameraview-related updates will not affect c2.
I haven't encountered this problem, it may be that my test time is not enough. I'll check this issue.

deanlee:
Tested for a few hours and it's working fine so far, I'll continue to test.

PR #23413, opened 2022-01-05
Statsd

robbederks:


selfdrive/athena/athenad.py
```diff
@@ -48,7 +49,7 @@
 recv_queue: Any = queue.Queue()
 send_queue: Any = queue.Queue()
 upload_queue: Any = queue.Queue()
-log_send_queue: Any = queue.Queue()
+secondary_send_queue: Any = queue.Queue()
```
pd0wm, line 52: `low_priority_send_queue`?

selfdrive/athena/athenad.py
```diff
@@ -472,6 +474,33 @@ def log_handler(end_event):
       cloudlog.exception("athena.log_handler.exception")
 
 
+def stat_handler(end_event):
+  while not end_event.is_set():
+    last_scan = 0
+    try:
+      curr_scan = sec_since_boot()
+      if curr_scan - last_scan > 10:
```
pd0wm, line 482: This needs some kind of sleep to not use 100% CPU. Don't sleep too long or it will delay shutdown.

selfdrive/athena/athenad.py
```diff
@@ -472,6 +474,33 @@ def log_handler(end_event):
...
+        for stat_entry in os.listdir(STATS_DIR):
+          stat_path = os.path.join(STATS_DIR, stat_entry)
+          with open(stat_path) as f:
+            stats = json.loads(f.read())
```
pd0wm, line 486:
```suggestion
          with open(stat_path) as f:
            stats = json.load(f)
```

Rest of the code can be outside of the context manager.

selfdrive/statsd.py
```diff
@@ -0,0 +1,105 @@
...
+from collections import deque
+from datetime import datetime
+from selfdrive.hardware import PC
+from cereal.messaging import SubMaster
+from common.file_helpers import atomic_write_on_fs_tmp
+
+
+AGGREGATION_TIME_S = 60
+STATS_SOCKET = "ipc:///tmp/stats"
+if PC:
```
pd0wm, line 16: Maybe put this logic in `selfdrive.loggerd.config`?

pd0wm:
After some discussion we decided to try to follow the statds input and output formats as closely as reasonably possible. This should make it easier if we need to interface with existing tooling.

This is the json output sent to a normal statsd backend for some timers:
```js
{
  counters: {
    'statsd.bad_lines_seen': 0,
    'statsd.packets_received': 5,
    'statsd.metrics_received': 5
  },
  timers: { baz: [ 20, 20, 20, 20, 50 ] },
  gauges: {},
  timer_data: {
    baz: {
      count_90: 5,
      mean_90: 26,
      upper_90: 50,
      sum_90: 130,
      sum_squares_90: 4100,
      count_99: 5,
      mean_99: 26,
      upper_99: 50,
      sum_99: 130,
      sum_squares_99: 4100,
      std: 12,
      upper: 50,
      lower: 20,
      count: 5,
      count_ps: 0.5,
      sum: 130,
      sum_squares: 4100,
      mean: 26,
      median: 20
    }
  },
  counter_rates: {
    'statsd.bad_lines_seen': 0,
    'statsd.packets_received': 0.5,
    'statsd.metrics_received': 0.5
  },
  sets: {},
  pctThreshold: [ 90, 99 ]
}
```

And some gauges:
```js
{
  counters: {
    'statsd.bad_lines_seen': 0,
    'statsd.packets_received': 0,
    'statsd.metrics_received': 0
  },
  timers: {},
  gauges: { foo: 30, bar: 20, 'statsd.timestamp_lag': 0 },
  timer_data: {},
  counter_rates: {
    'statsd.bad_lines_seen': 0,
    'statsd.packets_received': 0,
    'statsd.metrics_received': 0
  },
  sets: {},
  pctThreshold: [ 90 ]
}
```

selfdrive/athena/athenad.py
```diff
@@ -472,6 +474,32 @@ def log_handler(end_event):
...
+            "method": "storeStats",
+            "params": {
+              **stats
+            },
+            "jsonrpc": "2.0",
+            "id": stat_entry
+          }
+          low_priority_send_queue.put_nowait(json.dumps(jsonrpc))
+          os.remove(stat_path)
+        last_scan = curr_scan
```
pd0wm, line 497: Should you set this in case of an exception? Otherwise it might get into a tight loop if it keeps failing.

selfdrive/athena/athenad.py
```diff
@@ -472,6 +474,32 @@ def log_handler(end_event):
       cloudlog.exception("athena.log_handler.exception")
 
 
+def stat_handler(end_event):
+  while not end_event.is_set():
+    last_scan = 0
+    try:
+      curr_scan = sec_since_boot()
+      if curr_scan - last_scan > 10:
+        for stat_entry in os.listdir(STATS_DIR):
```
pd0wm, line 483: maybe better to only process one file every ten seconds. That should still allow to work a backlog, but not allow athena to use too much CPU.

selfdrive/statsd.py
```diff
@@ -0,0 +1,96 @@
...
+  last_flush_time = time.monotonic()
+  gauges = {}
+  started_prev = False
+  while True:
+    try:
+      metric = sock.recv_string(zmq.NOBLOCK)
+      try:
+        metric_type = metric.split('|')[1]
+        metric_name = metric.split(':')[0]
+        metric_value = metric.split('|')[0].split(':')[1]
```
pd0wm, line 62: You can probably skip these splits by storing the unused results of the previous splits.

selfdrive/statsd.py
```diff
@@ -0,0 +1,96 @@
...
+        'started': started,
+        'gauges': {**gauges},
+      }
+
+      # clear intermediate data
+      gauges = {}
+
+      stats_path = os.path.join(STATS_DIR, str(flush_result['time_utc']))
+      with atomic_write_on_fs_tmp(stats_path) as f:
+        f.write(json.dumps(flush_result))
```
pd0wm, line 87:
```suggestion
        json.dump(flush_result, f)
```

selfdrive/athena/athenad.py
```diff
@@ -472,6 +475,32 @@ def log_handler(end_event):
       cloudlog.exception("athena.log_handler.exception")
 
 
+def stat_handler(end_event):
+  while not end_event.is_set():
+    last_scan = 0
+    try:
+      curr_scan = sec_since_boot()
+      if curr_scan - last_scan > 10:
+        for stat_entry in filter(lambda name: not name.startswith(tempfile.gettempprefix()), os.listdir(STATS_DIR)):
```
pd0wm, line 484:
```suggestion
        for stat_entry in (for fn in os.listdir(STATS_DIR) if not fn.startswith(tempfile.gettempprefix())):
```

The result is also not a `stat_entry` but a filename.

selfdrive/statsd.py
```diff
@@ -0,0 +1,104 @@
...
+        'started': started,
+        'gauges': {**gauges},
+      }
+
+      # clear intermediate data
+      gauges = {}
+
+      stats_path = os.path.join(STATS_DIR, str(flush_result['time_utc']))
+      with atomic_write_in_dir(stats_path) as f:
+        f.write(json.dumps(flush_result))
```
gregjhogan, line 95: serialize to this format: https://docs.influxdata.com/influxdb/v2.1/reference/syntax/line-protocol/

selfdrive/statsd.py
```diff
@@ -0,0 +1,104 @@
...
+      flush_result = {
+        'time_utc': int(datetime.utcnow().timestamp()),
+        'started': started,
+        'gauges': {**gauges},
+      }
+
+      # clear intermediate data
+      gauges = {}
+
+      stats_path = os.path.join(STATS_DIR, str(flush_result['time_utc']))
```
gregjhogan, line 93: We should limit the total number of files similar to cloud logs

selfdrive/statsd.py
robbederks: I'm doing this serialization on the backend now using the influx library. Shouldn't be too expensive, but up for discussion

pd0wm: Even if we do this client side we probably need to parse it again server side to do sanity checks.

selfdrive/statsd.py
```diff
@@ -0,0 +1,104 @@
...
+    if os.getpid() != self.pid:
+      self.connect()
+
+    try:
+      self.sock.send_string(metric, zmq.NOBLOCK)
+    except zmq.error.Again:
+      # drop :/
+      pass
+
+  def gauge(self, name: str, value: float):
```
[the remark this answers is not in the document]
robbederks: not sure if this would be that useful to have per metric, any use cases you can think of? would be easy to add once necessary I imagine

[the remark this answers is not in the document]
pd0wm: tags is also not part of the default statsd protocol. I assumed we wanted to follow that as much a possible.

selfdrive/athena/athenad.py
robbederks: personally like lambdas better, but up to you.

selfdrive/statsd.py
robbederks: would be great if there was something like f-strings for splitting... maybe regex?

selfdrive/statsd.py
```diff
@@ -0,0 +1,109 @@
...
+      metric = sock.recv_string(zmq.NOBLOCK)
+      try:
+        metric_type = metric.split('|')[1]
+        metric_name = metric.split(':')[0]
+        metric_value = metric.split('|')[0].split(':')[1]
+
+        if metric_type == METRIC_TYPE.GAUGE:
+          gauges[metric_name] = metric_value
+        else:
+          cloudlog.error(f"unknown metric type: {metric_type}")
```
adeebshihadeh, line 76: cloudlog.event makes these easier to parse
```suggestion
          cloudlog.error("unknown metric type", metric_type=metric_type)
```

selfdrive/statsd.py
```diff
@@ -0,0 +1,109 @@
...
+      except Exception:
+        cloudlog.error(f"malformed metric: {metric}")
+    except zmq.error.Again:
+      time.sleep(1e-3)
+
+    started = sm['deviceState'].started
```
adeebshihadeh, line 82: the SubMaster is never updated

adeebshihadeh: also no need to store started_prev
```suggestion
    started_prev = sm['deviceState'].started
    sm.update(0)
```

selfdrive/statsd.py
adeebshihadeh: yeah, let's keep all this stuff in one place

selfdrive/statsd.py
```diff
@@ -0,0 +1,120 @@
...
+  # open statistics socket
+  ctx = zmq.Context().instance()
+  sock = ctx.socket(zmq.PULL)
+  sock.bind(STATS_SOCKET)
+
+  # initialize stats directory
+  Path(STATS_DIR).mkdir(parents=True, exist_ok=True)
+
+  # initialize tags
+  tags = {
```
adeebshihadeh, line 60: can we add an engaged tag too?

selfdrive/statsd.py
```diff
@@ -0,0 +1,120 @@
...
+    'dongleId': Params().get("DongleId", encoding='utf-8'),
+    'started': False,
+    'version': get_short_version(),
```
adeebshihadeh, line 63: why short version? I often use the full version for filteeing

adeebshihadeh: it’s also confusing because we don’t use the short version anywhere

pd0wm: If we fix master-ci to include the hash again it will blow up the influxdb cardinality.

selfdrive/statsd.py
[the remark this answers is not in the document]
gregjhogan: tags will definitely be extremely useful for things like the version, but I guess maybe they will all be static

the alternative is always using metric names namespaced by the tags in the name (e.g. metric `cpu_temp` vs `cpu_temp.0_8_12`) but this can be a lot more difficult to work with

selfdrive/statsd.py
```diff
@@ -0,0 +1,120 @@
...
+    'branch': get_short_branch(),
+    'dirty': is_dirty(),
+    'origin': get_origin(),
```
gregjhogan, line 66: it would be nice if the origin was normalized (probably everywhere) for git vs http as well as having or not having .git at the end

gregjhogan: another option would be adding `get_normalized_origin()` I guess

robbederks: Normalized the origin here, we can switch over other places if wanted in another PR!

selfdrive/statsd.py
robbederks: don't think that's useful for all metrics (and would practically double cardinality again). I'm contemplating adding optional tags, where you could include it if wanted

selfdrive/statsd.py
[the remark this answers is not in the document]
robbederks: oh, things like version etc will all be in the tags by default (see where it generates the lines). I do maybe want to add optional tags to metrics as well where wanted (see adeeb's comment about engaged status for example)

selfdrive/statsd.py
robbederks: that's indeed why. If you want release, you could filter on version + branch + origin for example

PR #23414, opened 2022-01-05
 PID: Use last actuators to limit integrator

ClockeNessMnstr:
cleanup for harald's review of https://github.com/commaai/openpilot/pull/23244

bf -> backfeed
TODO: long feedback
save control_last

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -491,7 +491,7 @@ class Controls:
     if not self.joystick_mode:
       # accel PID loop
       pid_accel_limits = self.CI.get_pid_accel_limits(self.CP, CS.vEgo, self.v_cruise_kph * CV.KPH_TO_MS)
-      actuators.accel = self.LoC.update(self.active, CS, self.CP, long_plan, pid_accel_limits)
+      actuators.accel = self.LoC.update(self.active, CS, self.CP, self.last_actuators, long_plan, pid_accel_limits)
 
       # Steering PID loop and lateral MPC
       lat_active = self.active and not CS.steerWarning and not CS.steerError and CS.vEgo > self.CP.minSteerSpeed
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -491,7 +491,7 @@ class Controls:
     if not self.joystick_mode:
       # accel PID loop
       pid_accel_limits = self.CI.get_pid_accel_limits(self.CP, CS.vEgo, self.v_cruise_kph * CV.KPH_TO_MS)
-      actuators.accel = self.LoC.update(self.active, CS, self.CP, self.last_actuators, long_plan, pid_accel_limits)
+      actuators.accel = self.LoC.update(self.active, CS, self.CP, long_plan, pid_accel_limits)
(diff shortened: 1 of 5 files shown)


pd0wm:
Just create a new PR into master.

ClockeNessMnstr:
moved it to master

HaraldSchafer:
So like we mentioned, a couple unit tests testing the override and the noise when clipping integrator would be good before merging. 

I've been driving on it and didn't notice anything strange.

adeebshihadeh:
Converting to draft until tests are written.

ClockeNessMnstr:
I've been throwing a couple things at a noise unit test. mostly just checking a step response with noise added to check that the control stays moving towards target. 
does anyone have an idea what we want to catch there? 
once the error is low, the noise dominates so any directional checks trigger. Not sure what meaningful thresholds to set. 

override shouldn't be too hard, 
I'm thinking some short pulses of override and longer duration override to just check states/response when released.

HaraldSchafer:
Yep looks good!

Yeah just a test that there's no windup against the override should be enough. And that should be enough tests.

ClockeNessMnstr:
Think I got all the tests.

goal of the noise test is to detect excess noise entering the I-term and reversing it's action significantly in relation to the ramp up limits. 
i_term is allowed to reverse as long as the control is not moving away from the target. 
The control is allowed to move away from target due to the p term.
assertion is testing a mix of these conditions making it a bit arbitrary but should catch any destabilized I-term as is. 

Need to sanity check the limits of the tests with some graphs and update the limits on the checks to be as effective/meaningful as possible.

ClockeNessMnstr:
Got good coverage of different behavior. Going to go through and re-define the tests later and remove or gate off all the extras.

ClockeNessMnstr:
Alright, Unit tests are done.
Last commit is a plotter/debugger so the tests can be plotted.
Thinking of splitting off the plots to a PID tool / demo. Interactive readme of sorts for PID tuning.

ClockeNessMnstr:
Rebased for removal of saturation check from PID class.

HaraldSchafer:
Cool we'll try get this merged for 0.8.14

ClockeNessMnstr:
Last commit with the plot functions can be dropped to keep it cleaner if needed. 
Moved everything plot-related to the last commit for that reason. It's just easier to see what the test is doing when running manually.

Does "coverage run --append -m unittest discover" find the new unit test? or do I need to list the file somewhere like .github/workflows/selfdrive_tests.yaml

ClockeNessMnstr:
adjusted the noise test to use +/- 10% noise. 
Previously was adjusting the noise based on the Kp values being checked to ensure that the rate limits were hit aggressively. 
This was producing somewhat meaningless results. 

now it checks that the average error is bounded to less than 1/10 of the magnitude of the noise. This just checks that the output is stable.

pd0wm:
@HaraldSchafer Any updates on this?

HaraldSchafer:
I've decided not to merge this. I'm still not really sure what behavior we want. This backfeed approach doesn't seem ideal, I've decided to just freeze the integrator when it's clipping,

PR #23415, opened 2022-01-05
Add meta string to events

pd0wm:
@adeebshihadeh is knowing this in FindInteresting worth the complexity?

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import os
 import math
+import json
 from numbers import Number
 
 from cereal import car, log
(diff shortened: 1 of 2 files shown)


selfdrive/controls/lib/events.py
```diff
@@ -73,31 +74,37 @@ def create_alerts(self, event_types: List[str], callback_args=None):
...
+          if DT_CTRL * (self.events_t[e] + 1) >= alert.creation_delay:
             alert.alert_type = f"{EVENT_NAME[e]}/{et}"
             alert.event_type = et
             ret.append(alert)
     return ret
 
   def add_from_msg(self, events):
     for e in events:
-      self.events.append(e.name.raw)
+      meta = e.meta if len(e.meta) else None
```
pd0wm, line 93:
```suggestion
      meta = e.meta if e._has('meta') else None
```
Not sure if using this hidden method is cleaner

selfdrive/controls/controlsd.py
```diff
@@ -274,11 +275,15 @@ def update_events(self, CS):
...
       self.events.add(EventName.usbError)
     elif not self.sm.all_alive_and_valid() or self.can_rcv_error:
-      self.events.add(EventName.commIssue)
+      meta = {
+        'invalid': [s for s, valid in self.sm.valid.items() if not valid],
+        'not_alive': [s for s, alive in self.sm.alive.items() if not alive],
+        'can_rcv_error': self.can_rcv_error,
+      }
+
+      self.events.add(EventName.commIssue, meta=json.dumps(meta))
```
pd0wm, line 284: move json.dumps  to `Events.add()` if it's not a string already?

adeebshihadeh:
I don't think this is worth it. These events should infrequent, and you'll want to look into the logs anyway.

pd0wm:
ok! Let's see if we can get some critical logs into the qlogs then?

PR #23418, opened 2022-01-05
lateral_planner: follow capnp best practices

deanlee:


adeebshihadeh:
This is another case where this basically won't have a performance impact, and the code now looks a bit unnatural. I don't think it's worth it in these cases.

PR #23420, opened 2022-01-05
Add Niro EV to 384 STEER_MAX

VirtuallyChris:
Confirmed working great without faults by Kate from Transport Evolved.

`e809296d6ef28372|2021-12-09--16-52-57--2`


adeebshihadeh:
It's not just that it doesn't fault. We want to ensure that the value isn't just clipped. An easy way to check is find the max value the stock system requests.

VirtuallyChris:
I'll ask for a stock route. Kate confirmed that there's more torque by making turns that would hit torque limit before the change.

adeebshihadeh:
That should be fine.

PR #23425, opened 2022-01-06
log ERROR messages in qlogs too

pd0wm:

### DIFF
--- a/selfdrive/logmessaged.py
+++ b/selfdrive/logmessaged.py
@@ -30,11 +30,12 @@ def main() -> NoReturn:
     # then we publish them
     msg = messaging.new_message()
     msg.logMessage = record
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
The volume of error messages is way too high, but I like this idea. Perhaps we use the critical level instead, and call the service name something a bit more generic in case we want to change the level in the future.

pd0wm:
I'm fine with critical. I chose error since exceptions are logged at that level.

adeebshihadeh:
I think this should ultimately be error, but are we ok with massive qlogs until the error volume is more reasonable? I’m fine with it either way.

pd0wm:
I'm fine with going over everything that logs as an error and check if that's really necessary

PR #23426, opened 2022-01-06
TSS2 Toyota: improve stop and go performance

sshane:
Lowering `vEgoX` to 0.2 m/s means we let MPC/PID control acceleration for longer (without as harsh of a dip to 0 from 0.5 m/s), and start moving earlier.

I have PJ graphs showing TSS2 behavior, but I'll probably need to get new routes since they're old. I'll update this PR once I gather the data. Been running this on my fork for all TSS2 users for almost a month.

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -233,7 +233,9 @@ class CarInterface(CarInterfaceBase):
       set_long_tune(ret.longitudinalTuning, LongTunes.PEDAL)
     elif candidate in TSS2_CAR:
       set_long_tune(ret.longitudinalTuning, LongTunes.TSS2)
-      ret.stoppingDecelRate = 0.3  # reach stopping target smoothly
+      ret.vEgoStopping = 0.2  # car is near 0.1 to 0.2 when car starts requesting stopping accel
+      ret.vEgoStarting = 0.2
+      ret.stoppingDecelRate = 0.8  # reach stopping target smoothly - seems to take 0.5 seconds to go from 0 to -0.4
     else:
       set_long_tune(ret.longitudinalTuning, LongTunes.TSS)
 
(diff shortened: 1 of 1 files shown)


selfdrive/car/toyota/interface.py
```diff
@@ -246,7 +246,9 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       set_long_tune(ret.longitudinalTuning, LongTunes.PEDAL)
     elif candidate in TSS2_CAR:
       set_long_tune(ret.longitudinalTuning, LongTunes.TSS2)
-      ret.stoppingDecelRate = 0.3  # reach stopping target smoothly
+      ret.vEgoStopping = 0.2
+      ret.vEgoStarting = 0.2
+      ret.stoppingDecelRate = 0.8  # reach stopping target smoothly
```
sshane, line 251: We can either reach -2.0 m/s/s in the same time stock ACC gets to its standstill accel of -0.4 m/s/s, or use the same rate (take longer to get to -2.0 than stock to -0.4). Not sure if it makes too much of a difference, but this is using the same decel rate stock ACC uses.

sshane:
From a route from https://github.com/commaai/openpilot/issues/23778, shows an example where stock Toyota ACC wants to accelerate when our vTargetFuture is around 0.2 m/s, which matches this PR's changes. https://user-images.githubusercontent.com/25857203/154400565-f0581a42-71f8-4299-b9cd-bba931ba1db5.png

We can't just lower `vEgoStarting` until https://github.com/commaai/openpilot/pull/23333 is merged

zorrobyte:
@sshane 
Tested on latest master
```4aa5d874f367adbc|2022-02-17--21-21-11```

sshane:
This change seems to make us start requesting accel almost exactly when stock ACC does: https://imgur.com/a/xRnMAuw

BrettLynch123:
This greatly reduces the number of times i wished i had no disengage on accellerator (which is now in master thanks). but would like to test its effect on stopping on hills behind a car to make sure it doesn't cause worse rollback. I can sometimes get rollback in my 2020 corolla hybrid sedan. But even if it does I'd say its still a big improvement and i will just have to hit the break to avoid rollback when combing to a stop on an up hill.

sshane:
will re-open with testing

PR #23431, opened 2022-01-06
locationd: disable falling device detector

adeebshihadeh:


deanlee:
looks like that test_locationd.py/test_device_fell(self) should be disabled too

PR #23435, opened 2022-01-07
controlsd: cache car events

deanlee:
Don't rebuild car events every time. it's not cheap.

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -686,7 +686,8 @@ class Controls:
     self.pm.send('controlsState', dat)
 
     # carState
-    if events_changed := (self.events.names != self.events_prev):
+    events_changed = self.events.names != self.events_prev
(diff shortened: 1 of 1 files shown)


selfdrive/controls/controlsd.py
```diff
@@ -685,19 +686,21 @@ def publish_logs(self, CS, start_time, actuators, lac_log):
     self.pm.send('controlsState', dat)
 
     # carState
-    car_events = self.events.to_msg()
+    if events_changed := (self.events.names != self.events_prev):
```
pd0wm, line 689: The walrus operator has it's valid uses, but don't think this makes things more clear. Just write this as two lines.

```suggestion
    events_changed = self.events.names != self.events_prev
    if events_changed:
```

adeebshihadeh:
We should just make this cheap. Last time I profiled it, I found some low hanging fruit here.

PR #23436, opened 2022-01-07
constrolsd: read speeds in if statement

deanlee:


pd0wm:
Not sure this is much of an improvement. Having less code behind the if statement makes it easier to catch potential bugs.

PR #23440, opened 2022-01-07
Do not disable loggerd when below 5% space

pd0wm:
We already have an alert in controlsd that alerts when space is below 7%. Let's check if the outOfSpace event ever triggers before merging this.


adeebshihadeh:
Looks `outOfSpace` isn't in FindInteresting. I checked all the segments from the last week in mkv (52k) and didn't find any instances, so think this is fine to merge.

PR #23441, opened 2022-01-07
Longitudinal Tune equal to stock

AlexandreSato:
I tested this adjustment and it was desirablely equal to the STOCK ACC, in heavy traffic the cars do not cut and nor honk complaining that it does not start to move as soon as the front car begins to move. When you're on the move following a car in front, you don't look "loose" and you don't get jerk acceleration. The distance when the car to the rear of the other is equal to the stock or it is possible to see the rear tires of the car in front.

### DIFF
--- a/selfdrive/controls/lib/longitudinal_mpc_lib/long_mpc.py
+++ b/selfdrive/controls/lib/longitudinal_mpc_lib/long_mpc.py
@@ -34,7 +34,7 @@ X_EGO_COST = 0.
 V_EGO_COST = 0.
 A_EGO_COST = 0.
 J_EGO_COST = 5.0
-A_CHANGE_COST = .5
+A_CHANGE_COST = 0.
 DANGER_ZONE_COST = 100.
 CRASH_DISTANCE = .5
 LIMIT_COST = 1e6
@@ -49,9 +49,9 @@ T_IDXS_LST = [index_function(idx, max_val=MAX_T, max_idx=N+1) for idx in range(N
 T_IDXS = np.array(T_IDXS_LST)
 T_DIFFS = np.diff(T_IDXS, prepend=[0.])
 MIN_ACCEL = -3.5
-T_FOLLOW = 1.45
+T_FOLLOW = 1.25
 COMFORT_BRAKE = 2.5
-STOP_DISTANCE = 6.0
+STOP_DISTANCE = 5.0
 
 def get_stopped_equivalence_factor(v_lead):
   return (v_lead**2) / (2 * COMFORT_BRAKE)


HaraldSchafer:
There was a decent amount of thought put into the current policy, and probably won't change it anytime soon. Many cars start moving too late, though that's usually caused more by the car-specific controls rather than the policy. You can debug this by using plotjuggler.

krkeegan:
Hi @HaraldSchafer.  Thank you for all of your work on the new MPC, I have enjoyed watching the development of it and realize that you put a lot of thought into it.

I agree with you that some brands *cough Toyota* have a lag in the initial acceleration.  But, in my research, that is not the main cause of the lag we are seeing.  Many of us noticed the lag was introduced in 8.10 with the new MPC.  In 8.9 and prior, I had no complaints about the starting of my vehicle from a stop.  Immediately from 8.10 I noticed issues, with the same identical tuning.  Fast forward to 8.12, and there has been some minor tuning tweaks and some improvements to the MPC, but I am still feeling a significant and problematic lag on starting from a stop.

Below is a graph showing a simulated manuever behind a vehicle starting from a stop accelerating at 1.2m/s2 I tested the same manuever on v8.9 and v8.12 of Openpilot.  To generate this, I used the longitudinal tests and manuever.  I am happy to share the code if anyone wants to verify.

There is a lot to take in, so here are my comments on what I think is relevant:

- First the blue lines are vEgo, the dashed is 8.9 and solid is 8.12.  We can see that 8.12 is much slower to accelerate, appearing to sit still for nearly a full second.
- Second the red lines are the distance to lead again dashed is 8.9 and solid is 8.12.  Here is where I think you can best see the issue.  Not only does the distance grow more quickly in 8.12, but at its peak it is ~12m farther away from the lead.
- Finally, I included the 8.12 desired distance calculation, as a further example of how far away we get from ideal.  In 8.9 we drifted a maximum of ~13m from our desired distance, while in 8.12 we drift a maximum of ~25m.

image

Obviously, this is a simulation, but it really simulates the best case scenario.

I have been driving on a setup that dynamically decreases the `A_CHANGE_COST` and `J_EGO_COST` linearly depending on speed.  At 0m/s the jerk cost is .5 tapering back stock at 10m/s.  Similarly the a_change cost is .1 at 0m/s and also tapers back to stock at 10m/s.  This significantly improves my acceleration from a stop without any jerkiness, hard braking, creeping, or any other side effects that I can detect.

Would you be open to considering a solution like that?  I am happy to make a clean PR so you can see it better if you are interested.

HaraldSchafer:
@krkeegan I suspect the biggest reason the 0.8.9 and 0.8.12 plots look different is because the stationary follow distance was changed from 4m to 6m. This was done to allow for cars that do not follow brake profiles correctly to still brake on time. An unfortunate trade-off I don't see a way to avoid atm.

I'm in principle not very opposed to lowering the jerk and a_change costs at low speed. But merging these type of "policy tunes" takes a lot of verification work from us, even when the PRs are perfect. And I don't want to spend any time on this now. All the time we spend on this we aren't spending on e2e long, which is supposed to start replacing the take-off from stop behavior ASAP, it is intended to be the first e2e long feature.

Making sure toyotas actually follow the plan in a clean mergable will benefit e2e too, this won't, which is why I keep mentioning it. 

I appreciate the posts with plots, they and the other toyota tests you've done, are helpful

krkeegan:
Fair enough, e2e is important.  I did try changing the stopping distance to 4, it improves things, but isn't the main cause of the sluggishness.  The low jerk rate from a dead stop is the biggest contributor.

I appreciate the thoughtful response.  Have a nice weekend.

PR #23446, opened 2022-01-07
read carParams from namedtuple instead of pycapnp

deanlee:
Use namedtuple to read CarParams for faster and safer state(the namedtuple is immutable). it is nearly identical to native performance.

### DIFF
--- a/common/namedtuple.py
+++ b/common/namedtuple.py
@@ -1,0 +1,18 @@
+import collections
+
+
+def create_namedtuple_from_dict(obj):
+  if isinstance(obj, dict):
+    fields = sorted(obj.keys())
+    namedtuple_type = collections.namedtuple(typename='OPNamedType', field_names=fields, rename=True, defaults=(None,))
+    field_value_pairs = collections.OrderedDict((str(field), create_namedtuple_from_dict(obj[field])) for field in fields)
+    return namedtuple_type(**field_value_pairs)
+  elif isinstance(obj, (list, set, tuple, frozenset)):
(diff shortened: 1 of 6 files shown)


adeebshihadeh:
I'd like to do this optimization more generically. My two immediate ideas are to either move to compiling the schema to Cython or implement something like this in SubMaster. I think we should try the Cython approach first. What do you think?

deanlee:
Here are some issues I've found so far after doing a quick view on the cython approach: 
1. can't use keyword from,to ,the struct Segment needs to be modified to
>   struct Segment @0xa5b39b4fc4d7da3f {
    fromXXX @0 :LatLng;
    toXXX @1 :LatLng;
2.  doesn't support anyporint in capnp, struct Map needs to be modified to

> struct Map {
>   entries @0 :List(Entry);
>   struct Entry {
>     key @0 :Text;
>     value @1 :Text;
>   }
> }
> 

The above two issues are easy to solve, but the most serious problem is that the cython wrapper does not support custom namespace: `cereal,` and even if we added the namespace cereal to the generated cython files through hack, we still need to change all read function in python from sm['abc'].abc to sm['abc'].getXXXX. this is very unreasonable. In addition, the cython wrapper still return the pointer data member through dynamicstruct, I'm wondering how much the performance will improve.

the parsing part of pycapcp should be almost as fast as c++. the biggest problem is the way it read data. currently I prefer to see how much performance we can improve by optimizing the submaster rather than hacking the cython wrapper.

deanlee:
closed. moved to cereal with a better cache class https://github.com/commaai/openpilot/pull/23454

PR #23449, opened 2022-01-08
use roll std from locationd

nuwandavek:


nuwandavek:
image

selfdrive/locationd/paramsd.py
```diff
@@ -45,10 +45,15 @@ def handle_log(self, t, which, msg):
...
+      localizer_roll_std = msg.orientationNED.std[0]
       roll_valid = msg.orientationNED.valid and ROLL_MIN < localizer_roll < ROLL_MAX
       if roll_valid:
         roll = localizer_roll
-        roll_std = np.radians(1.0)
+        # A multiplier is needed as the std at localizer-filter-reset point is not high enough.
+        # A high multiplier (~10) has a better effect in filtering out jumps, but introduces a lag
+        # in the learnt roll value.
+        # A multiplier of 2, has steady state std ~1deg, and sufficiently filters out jumps
+        roll_std = localizer_roll_std * 2
```
HaraldSchafer, line 56: 2 * localizer_roll_std

selfdrive/locationd/paramsd.py
```diff
@@ -45,10 +45,15 @@ def handle_log(self, t, which, msg):
...
       localizer_roll = msg.orientationNED.value[0]
+      localizer_roll_std = msg.orientationNED.std[0]
       roll_valid = msg.orientationNED.valid and ROLL_MIN < localizer_roll < ROLL_MAX
       if roll_valid:
         roll = localizer_roll
-        roll_std = np.radians(1.0)
+        # A multiplier is needed as the std at localizer-filter-reset point is not high enough.
+        # A high multiplier (~10) has a better effect in filtering out jumps, but introduces a lag
+        # in the learnt roll value.
+        # A multiplier of 2, has steady state std ~1deg, and sufficiently filters out jumps
```
HaraldSchafer, line 55: Change whole comment block

` Experimentally found multiplier of 2 to be best trade-off between stability and accuracy` or similar?

HaraldSchafer:
@adeebshihadeh Looks like proc replay is passing here when it should fail? Comparison with nan always passes?

adeebshihadeh:
> @adeebshihadeh Looks like proc replay is passing here when it should fail? Comparison with nan always passes?

Fixed in 9941c99f7c1d807acd440f29b4de3493b6e11e50

PR #23453, opened 2022-01-08
Toyota: use common DBC for TSSP Camry Hybrid

sshane:
Needs opendbc PR merged first: https://github.com/commaai/opendbc/pull/487


adeebshihadeh:
Nice! I'm very interested in cleaning up the DBCs as much as possible.

PR #23454, opened 2022-01-08
submaster: optimize pycapnp reader

deanlee:
https://github.com/commaai/cereal/pull/240

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -95,7 +95,8 @@ class Controls:
     print("Waiting for CAN messages...")
     get_one_can(self.can_sock)
 
-    self.CI, self.CP = get_car(self.can_sock, self.pm.sock['sendcan'])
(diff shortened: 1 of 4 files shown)


pd0wm:
Are there any measurable performance increases from this?

Besides carParams, are we really reading the same value often enough this becomes worth it?

deanlee:
the SubMaster will be read multiple times in many place (event not updated). 
this PR can also speed up the access speed of child elements even if read only after update:

> sm["abc"].a.x //  first read, pycapnp performance
> sm["abc"].a.b // native performance
> sm["abc"].a.c // native performance
> 
> sm["abc"].a.b[0] // first read. pycapnp performance
> sm["abc"].a.b[1] // native performance

pd0wm:
How many percent of CPU does this save in practice?

deanlee:
it depends.  it may save 1~2% cpu  of controlsd on my laptop (I am not very sure at the moment, I did not systematically test it) 
I have not tested on devices yet

PR #23455, opened 2022-01-08
networking: cleanup&optimize WifiManager

### LINKED ISSUE #23445: UI: fix network related lags
Lags are obvious in these cases, but it's probably all fundamentally the same cause.

* wifi network scrolling isn't smooth
* wifi network scrolling freezes while updating the list
* keyboard isn't responsive in networking page
* open settings and rapidly switch between the different panels

deanlee:
1. inherit from QObject instead of QWidget
2. Added function WifiManager::call to simplify the d-bus calls
3. async scan/refresh

~Next PR: I plan to use QDBusAbstractInterface::asyncCall in WifiManager::call to solve the blocking problem~ i'll do async scan/refresh in this pr

resolve: https://github.com/commaai/openpilot/issues/23445


selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -243,63 +164,32 @@ QVector<QDBusObjectPath> WifiManager::get_active_connections() {
 }
 
 bool WifiManager::isKnownConnection(const QString &ssid) {
-  return !getConnectionPath(ssid).path().isEmpty();
+  return (bool)getConnectionPath(ssid);
```
adeebshihadeh, line 167: the explicit check is more clear here

adeebshihadeh, line 167: the explicit check is more clear here

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -243,63 +164,32 @@ QVector<QDBusObjectPath> WifiManager::get_active_connections() {
...
 }
 
 void WifiManager::forgetConnection(const QString &ssid) {
-  const QDBusObjectPath &path = getConnectionPath(ssid);
-  if (!path.path().isEmpty()) {
-    QDBusInterface nm2(NM_DBUS_SERVICE, path.path(), NM_DBUS_INTERFACE_SETTINGS_CONNECTION, bus);
-    nm2.call("Delete");
+  if (auto path = getConnectionPath(ssid)) {
```
adeebshihadeh, line 171: don't do assignment in conditions

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -243,63 +164,32 @@ QVector<QDBusObjectPath> WifiManager::get_active_connections() {
...
+  }
 }
 
-unsigned int WifiManager::get_ap_strength(const QString &network_path) {
-  QDBusInterface device_props(NM_DBUS_SERVICE, network_path, NM_DBUS_INTERFACE_PROPERTIES, bus);
-  device_props.setTimeout(DBUS_TIMEOUT);
-
-  QDBusMessage response = device_props.call("Get", NM_DBUS_INTERFACE_ACCESS_POINT, "Strength");
-  return get_response<unsigned int>(response);
+QByteArray WifiManager::get_property(const QString &network_path, const QString &property) {
```
adeebshihadeh, line 187: can you fix all the case mismatches? most of the file seems to be camel case

selfdrive/ui/qt/offroad/wifiManager.h
```diff
@@ -31,76 +36,93 @@ struct Network {
   ConnectedType connected;
   SecurityType security_type;
 };
-bool compare_by_strength(const Network &a, const Network &b);
+inline bool compare_by_strength(const Network &a, const Network &b) {
+  return a.connected > b.connected || a.strength > b.strength;
```
adeebshihadeh, line 40: connected networks don't seem to show on top anymore

adeebshihadeh: connecting should also show on top

adeebshihadeh (commented):
Nice cleanup. I did a quick review and test. Definitely seemed smoother in my quick test!

selfdrive/ui/qt/offroad/wifiManager.cc
deanlee: think it's more concise and clear to check std::optional variable in this way. How about keeping it as is? Otherwise we have to change many places.

adeebshihadeh:
My IP address is showing as `0.0.0.0` on wifi. Seems to work fine while tethering.

selfdrive/ui/qt/offroad/wifiManager.cc
adeebshihadeh: I guess this is alright for now, but I think std::optional is super confusing. Just reading `auto path = getConnectionPath(ssid)`, I have no idea what type `path` is.

deanlee:
reverted to the previous code to get the ip address.it should be ok know

adeebshihadeh:
Now it's not showing any networks besides the connected one.

adeebshihadeh:
> The scan will start after the net panel is displayed for 5 seconds. it still no refreshed after waiting 5-6 seconds?

Weird, I tried again (without 1a0b26c49de0ed1d29ce0dd763700b3695abeee9) and it works, but the scrolling is very laggy.

deanlee:
There are two possibilities to cause the laggy
1.isTetheringEnabled (it make call to d-bus) is called in a loop by UI to rebuild the list.
2.clear/rebuild widigets in scrollview may also cause laggy

adeebshihadeh:
Can we fix those cases too or do you want to do that in another PR?

deanlee:
I try to fix them together in this Pr

deanlee:
doun't know why failed to build docs

deanlee:
@adeebshihadeh  Why does the "build docs" fail?  Have I done something wrong?

deanlee:
Looks like this is the issue  https://github.com/michaeljones/breathe/issues/616

adeebshihadeh:
Let's make sure to not make the PR too big, so it's hard to review. Feel free to split this up if something's already done and self-contained.

deanlee:
closed.it's all split up

PR #23456, opened 2022-01-08
thermald: add function for blocking cloudlog instead of sleep

whowe444:
Use blocking cloudlog info function instead of sleeping


selfdrive/thermald/thermald.py
```diff
@@ -381,9 +380,7 @@ def thermald_thread() -> NoReturn:
 
     # Check if we need to shut down
     if power_monitor.should_shutdown(peripheralState, onroad_conditions["ignition"], in_car, off_ts, started_seen):
-      cloudlog.info(f"shutting device down, offroad since {off_ts}")
-      # TODO: add function for blocking cloudlog instead of sleep
-      time.sleep(10)
+      cloudlog.blocking_info(f"shutting device down, offroad since {off_ts}")
```
sshane, line 383: Maybe you should add a timeout flag as well in case it gets stuck. And maybe overriding the info function instead with a blocking argument if possible.

whowe444: Added the timeout. Not sure about overriding since that would mean every call to cloudlog.info() would have to do an additional boolean check which doesn't really make sense when there is only one call to the blocking version

sshane: Either way works, just a suggestion! I'd also add an argument with default for the timeout, so you can see that explicitly in thermald

whowe444: agreed

common/logging_extra.py
```diff
@@ -163,6 +163,14 @@ def event(self, event_name, *args, **kwargs):
     else:
       self.info(evt)
 
+  def blocking_info(self, msg, timeout=10):
```
pd0wm, line 166: Can you explain why this is blocking?

whowe444: My understanding is that this spawns a new thread, starts it, and the join function blocks the calling thread until the target function of thread t finishes.

adeebshihadeh:
This change doesn't actually do anything. In order to resolve the TODO, we want a function that blocks until the log hits disk, so that it can be uploaded on the subsequent boot. Check out `selfdrive/logmessaged.py` to see how this works more in more detail.

PR #23460, opened 2022-01-09
controlsd: wait for initialized to add events

sshane:
When starting the device while driving, and if you engage before the relay trips, various alerts can show, like `doorOpen`, `seatbeltNotLatched`, `espDisabled`, `commIssue`, etc. Wait until we're initialized for any alerts besides `controlsInitializing` and the startup alerts.

`initialized` at most should take 3.5 seconds, are there any alerts we need to show before then? I'm thinking not, but I haven't worked with the events that much before.

Screenshot from 2022-01-08 23-20-57

Segment with a bunch of events being added for a frame: `f15e3c37c118e841|2022-01-08--21-20-46--0`

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -190,9 +190,6 @@ class Controls:
     """Compute carEvents from carState"""
 
     self.events.clear()
-    self.events.add_from_msg(CS.events)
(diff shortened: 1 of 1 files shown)


sshane:
Duh, this was due to something on my fork, closing!

PR #23464, opened 2022-01-09
paramsd: remove unnecessary if statement

deanlee:
the event is already updated.


adeebshihadeh:
sm.updated is a dict with all the sockets, so this check isn’t redundant.

deanlee:
hmmm.my bad

PR #23465, opened 2022-01-09
paramsd: cache the params with the capnp struct

deanlee:


adeebshihadeh:
Can you write a test for this to ensure we read the cache correctly? There's a similar one for calibrationd you can use as an example.

selfdrive/locationd/test/test_paramsd.py
```diff
@@ -6,7 +6,6 @@
 import cereal.messaging as messaging
 from cereal import car
 from common.params import Params
-from selfdrive.locationd.paramsd import load_params
```
deanlee: import here will cause `self.localizer = self.lib.localizer_init()` failed in `selfdrive/locationd/test/test_locationd.py` with the following assertion:

> python3.8: rednose/helpers/ekf_sym.cc:14: EKFS::EKFSym::EKFSym(std::string, Map<EKFS::MatrixXdr>, Map<Eigen::VectorXd>, Map<EKFS::MatrixXdr>, int, int, int, int, int, std::vector<int>, std::vector<int>, std::vector<std::string>, double): Assertion `this->ekf' failed.

pd0wm:
Not a fan of this bytes in JSON. If the message isn't completely statically typed I don't see much of an advantage.

deanlee:
I'm not sure where to call CLEAR_ON_CAR_CHANGE would be more appropriate.

selfdrive/common/params.cc
```diff
@@ -137,7 +137,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"LastSystemShutdown", CLEAR_ON_MANAGER_START},
     {"LastUpdateException", PERSISTENT},
     {"LastUpdateTime", PERSISTENT},
-    {"LiveParameters", PERSISTENT},
+    {"LiveParameters", PERSISTENT | CLEAR_ON_CAR_CHANGED},
```
pd0wm, line 140: You can update `CalibrationParams` as well.

pd0wm: never mind, let's leave calibrationd for what it is. It currently also doesn't reset on car change.

pd0wm:
That should probably happen in controlsd when the fingerprint happens, before it writes the carParams.

deanlee:
how about use CarParamsCache and set it to PERSISTENT?

PR #23476, opened 2022-01-10
LateralPlanner: Move lane change logic to another class

sshane:
**Description** Moved the lane change state transition logic and calculations to a new class called DesireHelper. This removes a lot of bloat from the LateralPlanner which made it hard to glance at the different functions of the code (ie. model parsing, lane change logic, calculation of driving path, setting of MPC costs, and running the MPC).

**Verification** process replay!

### DIFF
--- a/selfdrive/controls/lib/desire_helper.py
+++ b/selfdrive/controls/lib/desire_helper.py
@@ -1,0 +1,113 @@
+from common.realtime import DT_MDL
+from selfdrive.config import Conversions as CV
+from cereal import log
+
+LaneChangeState = log.LateralPlan.LaneChangeState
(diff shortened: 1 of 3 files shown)


pd0wm:
Let's hold of on merging this until findInteresting has engaged segments again. Will be working on that next week.

selfdrive/controls/lib/desire_helper.py
```diff
@@ -0,0 +1,113 @@
+from common.realtime import DT_MDL
+from selfdrive.config import Conversions as CV
+from cereal import log
```
adeebshihadeh, line 3: sort alphabetically
```suggestion
from cereal import log
from common.realtime import DT_MDL
from selfdrive.config import Conversions as CV
```

PR #23481, opened 2022-01-10
LongControl: remove redundant check on active

deanlee:
I know it's useless, but there is a chinese proverb: "勿以善小而不为" (Don't miss doing any good thing no matter how insignificant it looks.).  so I still submitted this PR


pd0wm:
As the Dutch say: "Wie het kleine niet eert, is het grote niet weerd" (Who does not honour small things, is not worthy of great things)

PR #23485, opened 2022-01-10
use ndarray.tolist  to convert numpy array to float

deanlee:


pd0wm:
Seems to be about 10x faster :)

```
In [9]: timeit.timeit('y = [float(a) for a in x]', setup='import numpy as np; x = np.zeros(20)')
Out[9]: 3.55299404499965

In [10]: timeit.timeit('y = x.tolist()', setup='import numpy as np; x = np.zeros(20)')
Out[10]: 0.37758147599924996
```

PR #23489, opened 2022-01-11
thermald: add function for blocking cloudlog instead of sleep

### LINKED ISSUE #23484: Joystick 🕹️
### Describe the bug

Joystickd.py running on PC does not connect to comma device. Instructions should be improved since everyone I asked was unable to get it to work.

### OS Version

Ubuntu 20.04

_No response_

### Additional info

_No response_

whowe444:
perform fsync after cloudlog.info to make sure logs are written to disk before shutdown


common/logging_extra.py
```diff
@@ -163,6 +163,16 @@ def event(self, event_name, *args, **kwargs):
     else:
       self.info(evt)
 
+  def blocking_info(self, msg, fsync_dir):
+    # if logmessaged is running
+    self.info(msg)
+    # else get_file_handler().emit(msg)
+    #   pass
+    fd = os.open(fsync_dir, os.O_RDONLY)
```
pd0wm, line 171: You're probably looking for something like 
```
    for h in self.handlers:
      h.flush()
```

However, that also doesn't guarantee the message has been picked up by logmessaged already and has hit disk.

adeebshihadeh:
Making a blocking cloudlog is quite a bit more complicated than this. I would recommend something simpler for a first contribution.

whowe444:
> Making a blocking cloudlog is quite a bit more complicated than this. I would recommend something simpler for a first contribution.

Does it have to use logmessaged? In one of the comments it says that you can use the get_file_handler in the case that logmessaged isn't running. This uses SwaglogRotatingFileHandler which can call emit directly, which then calls flush on the file stream. So might just have to make sure to fsync

common/logging_extra.py
whowe444: So now I'm adding the file handler which bypasses logmessaged since it uses the filestream, calling handle which calls handle on all the handlers, which emits in all the handlers, but emit for the file_handler is just writing to a filestream and calling flush... then I make sure to fsync the whole directory

pd0wm:
Think the cleanest solution is to set the `DoShutdown` param. Then manager will handle the rest of the shutdown. That shuts down `logmessaged` cleanly before shutting down.

pd0wm:
https://github.com/commaai/openpilot/pull/23528

PR #23491, opened 2022-01-11
show prime widget on PC

sshane:
Makes for easier developing


pd0wm:
Want to keep differences between actual devices and PC to a mimimum. If you're developing on this you can just uncomment.

PR #23498, opened 2022-01-12
boardd: remove ublox timestamp logging

pd0wm:
This check was too unreliable to do anything with client side.


deanlee:
#include <unordered_map> & #include "selfdrive/locationd/ublox_msg.h" can be removed too.

PR #23499, opened 2022-01-12
optimize Events

deanlee:
The `Events.clear()`  rebuilds `events_prev` from all EVENTS  every time and deep copy `static_events`, which can hurt performance.

it should be 20-30x faster when there are events, and almost zero overhead when there are no events


adeebshihadeh:
I've got another optimized implementation that's halfway done in a branch somewhere. I'll come back to this once I have some time.

adeebshihadeh:
> I've got another optimized implementation that's halfway done in a branch somewhere. I'll come back to this once I have some time.

I didn't end up getting around to this since controlsd CPU usage isn't a problem post-comma two. Closing for now.

PR #23503, opened 2022-01-13
cleanup .gitattributes

gregjhogan:
should be fine to remove all of these?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,8 +1,2 @@
-*.keras filter=lfs diff=lfs merge=lfs -text
 *.dlc filter=lfs diff=lfs merge=lfs -text
 *.onnx filter=lfs diff=lfs merge=lfs -text
-*.pb filter=lfs diff=lfs merge=lfs -text
-*.bin filter=lfs diff=lfs merge=lfs -text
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
LGTM, maybe we should block certain file types with pre commit hooks? It is somewhat easy to check a big file into master and get it stuck in the history if it's not caught quickly.

adeebshihadeh:
Added a precommit for large files: #23505

gregjhogan:
check out partial clone, super useful as the history grows (even without accidental large files slipping in)
https://github.blog/2020-12-21-get-up-to-speed-with-partial-clone-and-shallow-clone/

`git clone --filter=blob:none git@github.com:commaai/openpilot.git` gives you
```
$ du -sh ~/tmp/openpilot/.git/objects/
93M	/home/greg/tmp/openpilot/.git/objects/
```
vs a non-partial clone
```
$ du -sh ~/openpilot/.git/objects/
1.7G	/home/greg/openpilot/.git/objects/
```
with partial clone you still have access to the full history, it just pulls objects  on-demand

adeebshihadeh:
Shall we put this in the instructions in tools/?

PR #23506, opened 2022-01-13
ui/networking: build wifi list inplace

deanlee:


selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -21,12 +21,13 @@ T get_response(const QDBusMessage &response) {
...
 }
 
-bool compare_by_strength(const Network &a, const Network &b) {
-  if (a.connected == ConnectedType::CONNECTED) return true;
-  if (b.connected == ConnectedType::CONNECTED) return false;
-  if (a.connected == ConnectedType::CONNECTING) return true;
-  if (b.connected == ConnectedType::CONNECTING) return false;
-  return a.strength > b.strength;
+bool compare_network(const Network &a, const Network &b) {
+  if (a.connected > b.connected) return true;
```
adeebshihadeh, line 25: I think the old implementation was a lot more readable.

selfdrive/ui/qt/offroad/networking.h
```diff
@@ -36,6 +54,9 @@ class AdvancedNetworking : public QWidget {
   explicit AdvancedNetworking(QWidget* parent = 0, WifiManager* wifi = 0);
 
 private:
+  void showEvent(QShowEvent* event) override {
```
adeebshihadeh, line 57: can you put this in the cc file?

adeebshihadeh:
There seems to be a couple bugs:
* font weight is inverted for connected/disconnected
* forget button isn't showing for the connected network
* the "connecting..." label should be aligned right

deanlee:
sorry, my bad.

adeebshihadeh:
Looks good, except now it seems the CONNECTED and CONNECTING states aren't getting set properly. If I load the UI on my PC while already connected to a network, I only see the network I'm connected to and it's properly bolded. However, once the other networks show up, it's no longer bold.

adeebshihadeh:
Closing until it's ready for review again

PR #23510, opened 2022-01-13
boardd: cleanup safety_setter_thread

deanlee:


adeebshihadeh:
Can you write some tests for this?

deanlee:
I feel like there is no way to initialize an valid or dummy vector<Panda*> for testing. any suggestion?

deanlee:
submitted a prerequisite for test cases https://github.com/commaai/openpilot/pull/23564

PR #23512, opened 2022-01-13
Subaru: fix missing commas in fingerprints

adhintz:
comma.ai was missing some commas.

*****Bug fix*****

**Description** 
Without the comma, python does string concatenation across lines.

**Verification**
Nothing beyond the automated tests. If you'd like someone with a Subaru to test this, please let me know.


adhintz:
values.py has plenty of types of encoding. \x00 \000 \a and ascii characters. Perhaps standardizing on all \x00 encoding would be nice.

I should remember how to run the tests locally instead of pushing to this PR to trigger the github workflows.

adeebshihadeh:
Nice catch. The linter should catch this after #23513.

PR #23517, opened 2022-01-14
Hyundai: Car Port for 2022 Kona Electric Camera SCC

sunnyhaibin:
**Checklist**
- [x] added to RELEASES, entry to CarInfo in selfdrive/car/*/values.py and ran selfdrive/car/docs.py to generate new docs
- [x] test route added to test_routes.py
- [x] route with openpilot:
  - `1d6ecb67ec23a1b8|2022-06-12--17-53-09`
  - `ff973b941a69366f|2022-07-28--22-01-19`
- [x] Harness Type:
  - [Hyundai N](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)
  - [Hyundai O](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)
- Dongle IDs of the FW versions:
  - `1d6ecb67ec23a1b8`
  - `248133d966e0182c`
  - `bcce6204298475be`
  - `ff973b941a69366f`

**Requires**

- https://github.com/commaai/panda/pull/965

Thanks to community Kona Electric 2022 owner Shfqrkhn, awesomebytes, and Zack1010OP.


RELEASES.md
```diff
@@ -4,6 +4,7 @@ Version 0.8.13 (2022-XX-XX)
  * Roll compensation
  * Improved camera focus on the comma two
  * Subaru ECU firmware fingerprinting thanks to martinl!
+ * Hyundai Kona Electric (EV) 2022 support thanks to sunnyhaibin!
```
adeebshihadeh, line 7:
```suggestion
 * Hyundai Kona Electric 2022 support thanks to sunnyhaibin!
```

docs/CARS.md
```diff
@@ -112,6 +112,7 @@
 | Hyundai   | Ioniq PHEV 2020-21              | SCC + LKAS        | Stock            | 0mph               | 0mph         |
 | Hyundai   | Kona 2020                       | SCC + LKAS        | Stock            | 0mph               | 0mph         |
 | Hyundai   | Kona EV 2018-19                 | SCC + LKAS        | Stock            | 0mph               | 0mph         |
+| Hyundai   | Kona EV 2022                    | All               | Stock            | 0mph               | 0mph         |
```
adeebshihadeh, line 115: for stuff like this, we should stick to the official name
```suggestion
| Hyundai   | Kona Electric 2018-19          | SCC + LKAS        | Stock            | 0mph               | 0mph         |
| Hyundai   | Kona Electric 2022               | All               | Stock            | 0mph               | 0mph         |
```

sshane:
This car seems to be a radar-less car, with the camera sending the SCC messages, so panda safety may be required to enable openpilot longitudinal control

sshane:
After talk with @sunnyhaibin and looking at routes, some 2021 Kona Electrics have a radar, however the definitions and addresses are different, and the camera sends the SCC messages. The longitudinal safety should be the same as the 2022 Kanas, however we need to detect these different 2021 model years to use a different radar parser and not try to disable the radar. Hate to make a new platform for the same model year, what do you think @pd0wm?

sshane:
We will merge the car initially thought to be 2021 into this PR as we believe it's 2022 now and all these cars *do* have a radar just for FCA, then sunny will finish up the openpilot port and I'll help out with safety. If it turns out there's actually a difference (only thing would be the 2022's don't broadcast radar anywhere), then we can split out later

sunnyhaibin:
@sshane User confirmed that button send from openpilot to resume from standstill works! 🎉

`ff973b941a69366f|2022-07-28--22-01-19`

This port should be ready to merge.

sshane:
Can you have him test gas to disengage?

sunnyhaibin:
> Can you have him test gas to disengage?

Will do! User mentioned he will have it tested in the AM his time.

sunnyhaibin:
@sshane  `ff973b941a69366f|2022-07-29--23-20-18` is the route that has disengage on gas per the user.

sshane:
@sunnyhaibin That route has no disengage on gas on

sshane:
Looks like sending the cancel button to disengage is working properly now!
Screenshot from 2022-08-02 14-01-29

sshane:
Didn't have too much data, but this car also seems like it has a lot of torque, similar to the other Konas. We'll need to limit it the same:
Screenshot from 2022-08-02 15-59-44

PR #23523, opened 2022-01-14
ui/WifiManager: inherit from QObject instead of QWidget

deanlee:

### DIFF
--- a/selfdrive/ui/qt/offroad/wifiManager.cc
+++ b/selfdrive/ui/qt/offroad/wifiManager.cc
@@ -29,7 +29,7 @@ bool compare_by_strength(const Network &a, const Network &b) {
   return a.strength > b.strength;
 }
 
-WifiManager::WifiManager(QWidget* parent) : QWidget(parent) {
+WifiManager::WifiManager(QObject *parent) : QObject(parent) {
   qDBusRegisterMetaType<Connection>();
   qDBusRegisterMetaType<IpConfig>();
   connecting_to_network = "";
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -315,21 +318,14 @@ void WifiManager::stateChange(unsigned int new_state, unsigned int previous_stat
...
 
 // https://developer.gnome.org/NetworkManager/stable/gdbus-org.freedesktop.NetworkManager.Device.Wireless.html
 void WifiManager::propertyChange(const QString &interface, const QVariantMap &props, const QStringList &invalidated_props) {
   if (interface == NM_DBUS_INTERFACE_DEVICE_WIRELESS && props.contains("LastScan")) {
-    if (this->isVisible() || firstScan) {
-      refreshNetworks();
-      emit refreshSignal();
-      firstScan = false;
-    }
+    refreshNetworks();
```
deanlee, line 328: always refresh after scan

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -74,10 +68,18 @@ void WifiManager::setup() {
   requestScan();
 }
 
+void WifiManager::start() {
+  timer.start(5000);
+  refreshNetworks();
+}
```
sshane, line 74: @deanlee Active scan takes a second, are we leaving enough time until we get the APs? This could be why it doesn't show a lot of APs right after opening the networking view

sshane: oh I see we have a signal set up to refresh again later. I suppose it's best to leave this here in case that breaks some how

PR #23525, opened 2022-01-14
thermald: move dbus and other slow calls into thread

### LINKED ISSUE #23516: Fix dbus related deviceState lags

pd0wm:
TODO:

- [x] Verify exceptions in threads still go to sentry

### DIFF
--- a/selfdrive/thermald/thermald.py
+++ b/selfdrive/thermald/thermald.py
@@ -250,8 +250,8 @@ def thermald_thread(end_event, hw_queue):
 
     msg = read_thermal(thermal_config)
 
-    if sm.updated['pandaStates'] and len(pandaStates.pandaStates) > 0:
-      pandaState = pandaStates.pandaStates[0]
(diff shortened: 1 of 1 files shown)


deanlee:
how about subscribing to the network event pushed by ui/wifimanager?   we can modify wifimanager to use d-bus's signals  to update the network state when it changes instead of current polling.

pd0wm:
We would still need to request some things using AT commands over dbus, so it wouldn't remove the need for the thread. Interesting idea though.

deanlee:
the following functions are also slow. it it possible to consider putting some of them into the thread ?:

get_available_percent
psutil.virtual_memory().percent
read_thermal

pd0wm:
think everything besides read_thermal will eventually move to this thread. This PR is to fix cases were the dbus calls are extremely slow, sometimes they seem to take more than 5 seconds even though all timeouts are set shorter.

PR #23530, opened 2022-01-14
Subaru Crosstrek Torque increase

ClockeNessMnstr:
Some EPS for Early Impreza/Crosstreks to not respond to the same torque requests as stock. 
The vehicle limits this message at 3071 on these EPS and this value works much better with the stock tune. 

I've done some joystick testing in the past with this and it's quite slow, not jerky/unsafe. 
Subaru EPS actually enforces a slew rate of about half what OP sets. 0-max torque in ~2 sec. 

I think my route with max torque in a curve just expired (3 days) but lateral acceleration can be verified as well. 

**Checklist**
- [x ] added to README
- [ ] test route added to [test_routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/test/test_routes.py)
- [ ] route with openpilot:

The 3 EPS were tested by myself, prlifestyle, budney,  michaelhonan, Moodkiller.

### DIFF
--- a/selfdrive/car/subaru/interface.py
+++ b/selfdrive/car/subaru/interface.py
@@ -41,9 +41,9 @@ class CarInterface(CarInterfaceBase):
       ret.centerToFront = ret.wheelbase * 0.5
       ret.steerRatio = 15
       ret.steerActuatorDelay = 0.4   # end-to-end angle controller
-      ret.lateralTuning.pid.kf = 0.00005
+      ret.lateralTuning.pid.kf = 0.00003333
       ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0., 20.], [0., 20.]]
-      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.2, 0.3], [0.02, 0.03]]
+      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.133, 0.2], [0.0133, 0.02]]
 
     if candidate == CAR.IMPREZA_2020:
       ret.mass = 1480. + STD_CARGO_KG
--- a/selfdrive/car/subaru/values.py
+++ b/selfdrive/car/subaru/values.py
@@ -6,6 +6,8 @@ class CarControllerParams:
   def __init__(self, CP):
     if CP.carFingerprint == CAR.IMPREZA_2020:
       self.STEER_MAX = 1439
+    elif CP.carFingerprint == CAR.IMPREZA:
+      self.STEER_MAX = 3071
     else:
       self.STEER_MAX = 2047
     self.STEER_STEP = 2                # how often we update the steer cmd


ClockeNessMnstr:
checks are broken, still moving from a PR to subaru community.

pd0wm:
This increases the max torque sent to these cars? I assume this also requires panda changes?

Is there less torque than other Subarus if you send 2047? Or is it just a question of changing the tuning.

We usually rent a car to verify this kind of changes, but sounds like this requires a very specific make/model.

ClockeNessMnstr:
Actually, we're looking over this in Subaru, it looks like the torque isn't different between Crosstreks.
Strange part there is the stock tune was still "weak" on mine specifically. 

I've hit max torque exceeded on a highway bend at 2047.
I also have a Subaru Forester. The foresters throw a lane keep assist error past 2047 but has significantly more torque/lateral acceleration at that point)
3071 puts the Crosstrek/Impreza just below the forester's turning capability. 

We can do the needed testing since it's an increase in torque. We could also do logging comparisons between the Forester and Crosstrek.  

Subaru eps for both models enforces a pretty low slew rate (~2 sec to max) and the new Max torque doesn't increase turning performance past the Forester so I'm confident we can get what we need tested.

ClockeNessMnstr:
There is less torque than other Subarus. 

The 2020 Crosstrek's limited torque at 1439 actually appears to be the same as the rest of the Crosstreks at 3071. The EPS message is 2x as responsive to the 2020's requests.

I believe The torque from the eps on the Forester is the same as the 2020 Crosstrek's or the same as the 17/18/19/20Hybrid using 3071.


What needs to be done exactly for testing? Would logs be enough to compare. 
If we need to do injection testing, what exactly needs to be recorded?

ClockeNessMnstr:
I cleaned up the change. It seems this applies to all impreza/crosstreks other than the 2020+ ICE models so it's much simpler

ClockeNessMnstr:
panda @ https://github.com/commaai/panda/pull/871

ClockeNessMnstr:
@adeebshihadeh, 

with the recent focus on lateral accel in controllers is there a data point I could get to validate this? 
I think the internal data on lateral accel also showed the low values of torque available with the low limit on the message. 
Currently steering torque limit can be hit on the highway @~80mph.

adeebshihadeh:
In the past, we've rented each car to establish and verify these limits, but I think we can verify this without having one, given the right data. Let's leave this open until we have a procedure to share.

ClockeNessMnstr:
Sounds good! Thanks

jprous:
> There is less torque than other Subarus.
> 
> The 2020 Crosstrek's limited torque at 1439 actually appears to be the same as the rest of the Crosstreks at 3071. The EPS message is 2x as responsive to the 2020's requests.
> 
> I believe The torque from the eps on the Forester is the same as the 2020 Crosstrek's or the same as the 17/18/19/20Hybrid using 3071.
> 
> What needs to be done exactly for testing? Would logs be enough to compare. If we need to do injection testing, what exactly needs to be recorded?

@ClockeNessMnstr I am stuck with my Forester (2019 2.0-L ES - with ACC, without LKAS - SA model). The EON and panda pairs perfectly with the car - no errors. No errors on the OBD system either. Car does not steer though, but it shows the LKAS icon on the dash in green when activated. At speed, on highway, steering feels stiff with EON activated, loose when deactivated. Could there be some steering response, but with too low torque? Steering rack P/N is 34110SJ010, same as some Aus models with LKAS. What is your thoughts? Steering ECU ROM problem? Or steering torque setting problem
[qlog (2).zip](https://github.com/commaai/openpilot/files/9045264/qlog.2.zip)
/

ClockeNessMnstr:
> > There is less torque than other Subarus.
> > The 2020 Crosstrek's limited torque at 1439 actually appears to be the same as the rest of the Crosstreks at 3071. The EPS message is 2x as responsive to the 2020's requests.
> > I believe The torque from the eps on the Forester is the same as the 2020 Crosstrek's or the same as the 17/18/19/20Hybrid using 3071.
> > What needs to be done exactly for testing? Would logs be enough to compare. If we need to do injection testing, what exactly needs to be recorded?
> 
> @ClockeNessMnstr I am stuck with my Forester (2019 2.0-L ES - with ACC, without LKAS - SA model). The EON and panda pairs perfectly with the car - no errors. No errors on the OBD system either. Car does not steer though, but it shows the LKAS icon on the dash in green when activated. At speed, on highway, steering feels stiff with EON activated, loose when deactivated. Could there be some steering response, but with too low torque? Steering rack P/N is 34110SJ010, same as some Aus models with LKAS. What is your thoughts? Steering ECU ROM problem? Or steering torque setting problem [qlog (2).zip](https://github.com/commaai/openpilot/files/9045264/qlog.2.zip) /

Find me on discord (same name) and send me the route id. I'll try to see what's showing up.

ClockeNessMnstr:
> In the past, we've rented each car to establish and verify these limits, but I think we can verify this without having one, given the right data. Let's leave this open until we have a procedure to share.

Just a bump. Hoping there's a procedure we can use to validate soon. 
Thanks

adeebshihadeh:
We don't have a procedure for you yet, but a good place to start is with some stock lane centering routes using EyeSight. Also, what's the max lateral acceleration you're seeing on this branch?

ClockeNessMnstr:
About 2.4-2.8 m/ss. 

I'll try to get out and get a stock route recorded

michaelhonan:
> stock lane centering routes

Impreza 2017-2020 doesn't have lane centering btw, it's only got the nudge. I've got my max lat accel set to `3.0` on my 18 Impreza, and it's been spot on. I'm pretty terrible at plotjuggler though, so take it as a grain of salt.

dbzx6r:
@ClockeNessMnstr is there anything I can do to help push this along? Gathering routes, etc. I've self tested with the controller and taking over and I still had ample time to take over

dbzx6r:
Bump, hoping to eventually have this implemented, as my impreza seems to have a very low torque with stock OP.

Been using this for over a year on first ClockeNessMnstr's fork, and then one of Sunny's old forks.

Willing to do the controller test with a 360 camera, share logs, and I would love to get the requested Stock Eyesight routes recorded but I'm not sure how to get my C3 to record routes without OP enabled, as OP disables eyesight control. Please advise and I can get the requested data

michaelhonan:
I can supply routes too if needed - been running this for well over a year same as dbzx6r.
Stock OP really is unusable once you've tried the increased torque.

dbzx6r:
I have got some routes. They are all the same section of road, except for 2 which were just random drives that were decent in length. All of these routes are using 0.9.1, on MLP's subaru-community. Please note my impreza does not have lane centering, only lane keep assists which nudges me back into the lane if I start to cross the line. I tried to let it ping pong as much as possible to see if I could get the max torque it would do as stock.

- f284f22298d498cc|2023-05-24--20-25-16--0 **| Long drive with 2047**

- f284f22298d498cc|2023-05-25--21-47-32--0 **| Long drive with 3071**

- f284f22298d498cc|2023-05-25--23-37-31--0 **| Route 1: 3071**

- f284f22298d498cc|2023-05-25--23-45-24--0 **| Route 2: Stock Eyesight**

- f284f22298d498cc|2023-05-26--00-20-30--0 **| Route 3: 2047**

jnewb1:
closing this and moving it to an issue

PR #23532, opened 2022-01-14
manager: log exit reason

adeebshihadeh:


pd0wm:
LGTM. Not sure how often we will need to reference this in practice.

PR #23534, opened 2022-01-14
camerad: Out of bounds memory write

geohot:
This may break things, but it's correct.

### DIFF
--- a/selfdrive/camerad/cameras/camera_qcom2.cc
+++ b/selfdrive/camerad/cameras/camera_qcom2.cc
@@ -227,7 +227,7 @@ void sensors_init(int video0_fd, int sensor_fd, int camera_num) {
   buf_desc[0].size = buf_desc[0].length = sizeof(struct cam_cmd_i2c_info) + sizeof(struct cam_cmd_probe);
   buf_desc[0].type = CAM_CMD_BUF_LEGACY;
   struct cam_cmd_i2c_info *i2c_info = (struct cam_cmd_i2c_info *)alloc_w_mmu_hdl(video0_fd, buf_desc[0].size, (uint32_t*)&buf_desc[0].mem_handle);
-  struct cam_cmd_probe *probe = (struct cam_cmd_probe *)((uint8_t *)i2c_info + sizeof(struct cam_cmd_i2c_info));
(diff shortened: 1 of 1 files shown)


geohot:
@adeebshihadeh Are we running things in valgrind?

adeebshihadeh:
> @adeebshihadeh Are we running things in valgrind? (maybe it wouldn't find this anyway since it's a weird allocation)

Only ubloxd right now. It's on the list for the upcoming testing sprint.

selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -227,7 +227,7 @@ void sensors_init(int video0_fd, int sensor_fd, int camera_num) {
   buf_desc[0].size = buf_desc[0].length = sizeof(struct cam_cmd_i2c_info) + sizeof(struct cam_cmd_probe);
   buf_desc[0].type = CAM_CMD_BUF_LEGACY;
   struct cam_cmd_i2c_info *i2c_info = (struct cam_cmd_i2c_info *)alloc_w_mmu_hdl(video0_fd, buf_desc[0].size, (uint32_t*)&buf_desc[0].mem_handle);
-  struct cam_cmd_probe *probe = (struct cam_cmd_probe *)((uint8_t *)i2c_info) + sizeof(struct cam_cmd_i2c_info);
+  struct cam_cmd_probe *probe = (struct cam_cmd_probe *)((uint8_t *)i2c_info + sizeof(struct cam_cmd_i2c_info));
```
deanlee, line 230: Would this be clearer and less error-prone:
`auto probe = (struct cam_cmd_probe *)(i2c_info + 1);`

geohot:
I'm bringing up the IMX390 and it was causing issues with that sensor. It seems okay with the AR0231, but the version check doesn't work. (it's reading 0)

selfdrive/camerad/cameras/camera_qcom2.cc
geohot: Good point

geohot:
Should be good we merge if we trust that CI checks cameras

PR #23535, opened 2022-01-14
ui/networking: async dbus calls

deanlee:

### DIFF
--- a/selfdrive/ui/qt/offroad/networking.cc
+++ b/selfdrive/ui/qt/offroad/networking.cc
@@ -139,7 +139,7 @@ AdvancedNetworking::AdvancedNetworking(QWidget* parent, WifiManager* wifi): QWid
   list->addItem(editPasswordButton);
 
   // IP address
-  ipLabel = new LabelControl("IP Address");
+  ipLabel = new LabelControl("IP Address", wifi->ipv4_address);
   list->addItem(ipLabel);
 
   // SSH keys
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -178,8 +178,11 @@ AdvancedNetworking::AdvancedNetworking(QWidget* parent, WifiManager* wifi): QWid
   main_layout->addStretch(1);
 }
 
+void AdvancedNetworking::showEvent(QShowEvent *event) {
+  ipLabel->setText(wifi->getIp4Address());
```
adeebshihadeh, line 182: why remove this out of the refresh?

deanlee: This is a blocking call and I want to minimize the time it takes to refresh. Since ip is only displayed in the AdvanceNetworking, I feels it's reasonable to put it here.

adeebshihadeh:
Still seeing `0.0.0.0` if UI is launched while already connected to wifi.

selfdrive/ui/qt/offroad/networking.cc
deanlee: The downside is that if the ip changed, the user can only reopen an to see the new ip

deanlee:
> Still seeing `0.0.0.0` if UI is launched while already connected to wifi.

that's wired. Well, I'll reverse this change

PR #23537, opened 2022-01-15
Disengage on gas

KyleGreenlaw:
Allow the ability to press the gas pedal and keep openpilot engaged. Gated behind a toggle


geohot:
This is just a param.

PR #23538, opened 2022-01-15
Add Disengage on Gas Toggle

sunnyhaibin:
Adding the community-requested toggle. This toggle allows users to turn on or turn off disengage openpilot on gas press.


geohot:
Won't work. You have to set the unsafe mode in panda.

spektor56 (changes requested):
panda->set_unsafe_mode(p.getBool("DisengageOnGas")

This is not scalable if other toggles are added

geohot:
CI is failing. You have bugs.

krkeegan:
May I suggest a better icon 

https://github.com/krkeegan/openpilot/blob/a4c621103a0f3fd17f5f0fe4f4f27be098557b4e/selfdrive/assets/offroad/icon_gas.png

selfdrive/car/interfaces.py
```diff
@@ -32,6 +33,7 @@ def __init__(self, CP, CarController, CarState):
     self.steering_unpressed = 0
     self.low_speed_alert = False
     self.silent_steer_warning = True
+    self.disengage_on_gas = Params().get("DisengageOnGas", encoding='utf8')
```
pd0wm, line 36:
```suggestion
    self.disengage_on_gas = Params().get_bool("DisengageOnGas")
```

selfdrive/boardd/boardd.cc
```diff
@@ -174,7 +174,7 @@ bool safety_setter_thread(std::vector<Panda *> pandas) {
 
     LOGW("panda %d: setting safety model: %d with param %d", i, (int)safety_model, safety_param);
 
-    panda->set_unsafe_mode(0);  // see safety_declarations.h for allowed values
+    panda->set_unsafe_mode(p.getBool("DisengageOnGas") ? 0 : 1);  // see safety_declarations.h for allowed values
```
pd0wm, line 177:
```suggestion
    panda->set_unsafe_mode(p.getBool("DisengageOnGas") ? 0 : 1);  // see safety_declarations.h for allowed values
```

Think we should add the unsafe mode to CarParams and the panda health packet. Then we can read it back like we do for safety model and car params.

pd0wm:
Process replay is failing since the param is not set. You want to set the param here as well: https://github.com/commaai/openpilot/blob/master/selfdrive/test/process_replay/process_replay.py#L349

spektor56:
Also, gas is not the ISO term and is a bit confusing if you are driving a Tesla or some other EV. The proper term is accelerator

cydia2020:
> Also, gas is not the ISO term and is a bit confusing if you are driving a Tesla or some other EV. The proper term is accelerator

I agree, gas is a very ambiguous term in this scenario.

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -71,6 +71,12 @@ TogglesPanel::TogglesPanel(SettingsWindow *parent) : ListWidget(parent) {
       "In this mode openpilot will ignore lanelines and just drive how it thinks a human would.",
       "../assets/offroad/icon_road.png",
     },
+    {
+      "DisengageOnGas",
+      "Disengage openpilot On Gas Press",
```
cydia2020, line 76: Disengage openpilot on Accelerator Press would be better imho, but idk if it'd fit on the c2 display.

GregorKikelj: Driving a diesel here, not really confused by the gas pedal personally

cydia2020: I personally don't see the term gas pedal being used anywhere in the automotive industry *formally*, it's usually called the accelerator pedal. [Example](https://toyotamanuals.com.au/document/landing_page/prius-50-series-owners-manual-nov-17-dec-18), [Example](https://www.tacomaworld.com/threads/accelerator-and-brake-pedals-depressed-simultaneously-error-light.557335/)

Verylukyguy: Disable OpenPilot Disengagement on Acceleration

spektor56: > Driving a diesel here, not really confused by the gas pedal personally

Disengage when you put gas in the car if you saw the suggested icon lol

krkeegan: Fair enough, I deleted the suggestion.  The speed limit icon is better.

spektor56: > Fair enough, I deleted the suggestion. The speed limit icon is better.

I agree it needs a better icon, accelerator pedal should do

nelsonjchen: Would the shorter string `Disengage on Accel. Pedal` work?

nelsonjchen:
Just wanted to mention the issue counterpart: #23588

selfdrive/ui/qt/offroad/settings.cc
sshane: How does that look?

twilsonco: Here's an icon I made for disengage on gas move-fast toggle a while ago.
icon_car_pedal

sshane:
Can't push any more (due to LFS?), re-opening here: https://github.com/commaai/openpilot/pull/23977

PR #23539, opened 2022-01-15
dmonitoringd: remove duplicate call on update when driver engaged

deanlee:


pd0wm:
@ZwX1616 any reason that call was there?

ZwX1616:
> @ZwX1616 any reason that call was there?

Just saw this. Yeah the full loop used to run despite driverState updated or not, so this was to make resetting with wheel a bit more responsive. We should remove it now since everything in dm is synced to dcam frequency

adeebshihadeh:
@ZwX1616 feel free to merge if this looks good to you

PR #23542, opened 2022-01-15
ui: skip drawing invisible lanelines and road edges

deanlee:


pd0wm:
Don't think this is an improvement. We should make sure ui always uses the same amount of CPU, regardless of what the model is outputting. Creating all these different cases makes profiling much harder.

PR #23544, opened 2022-01-15
tools/replay: replace deprecated usleep(0) with std::this_thread::yield

deanlee:


deanlee:
got random test failure in test_swaglog:

>  selfdrive/common/tests/test_swaglog.cc:67: FAILED:
>   REQUIRE( thread_msgs[i] == thread_msg_cnt )
> with expansion:
>   0 == 100

fix it in https://github.com/commaai/openpilot/pull/23546

PR #23549, opened 2022-01-16
onroad ui speed 'max' -> 'set'

hewers:
**Description** Change onroad cruise set point string from "MAX" to "SET"

**Verification** Ran replay --demo and ui:

image


adeebshihadeh:
This may seem like a minor change, but design changes like this need to go through a formal process. We have a designer starting soon, and hope to do a comprehensive redesign somewhat soon.

Verylukyguy:
I like this idea.  Set speed seems like a slightly more accurate description than Max speed.

PR #23554, opened 2022-01-17
dmonitoringd: make real-time process

pd0wm:
driverState is frequency checked by controlsd. We have to either make it real-time or relax the timing.

### DIFF
--- a/selfdrive/monitoring/dmonitoringd.py
+++ b/selfdrive/monitoring/dmonitoringd.py
@@ -1,13 +1,19 @@
 #!/usr/bin/env python3
+import gc
+
+import cereal.messaging as messaging
 from cereal import car
 from common.params import Params
-import cereal.messaging as messaging
+from common.realtime import set_realtime_priority
 from selfdrive.controls.lib.events import Events
-from selfdrive.monitoring.driver_monitor import DriverStatus
 from selfdrive.locationd.calibrationd import Calibration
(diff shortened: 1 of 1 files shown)


selfdrive/monitoring/dmonitoringd.py
```diff
@@ -1,13 +1,19 @@
...
+from common.realtime import set_realtime_priority
 from selfdrive.controls.lib.events import Events
-from selfdrive.monitoring.driver_monitor import DriverStatus
 from selfdrive.locationd.calibrationd import Calibration
+from selfdrive.monitoring.driver_monitor import DriverStatus
 
 
 def dmonitoringd_thread(sm=None, pm=None):
+  gc.disable()
+  set_realtime_priority(5)
```
adeebshihadeh, line 15: why so high?

pd0wm: Copied from pramsd, since that was our other soft-realtime process. Controls is around 50. Feel free to change to something more appropriate.

adeebshihadeh, line 15:
```suggestion
  set_realtime_priority(2)
```

adeebshihadeh: made it 2, so it's one higher than rtshield. I'll paramsd to match too

PR #23557, opened 2022-01-17
Add sample metric type to statsd

robbederks:


adeebshihadeh:
This should be split up into two PRs: adding the new metric type and the new logging.

selfdrive/statsd.py
```diff
@@ -84,10 +92,14 @@ def get_influxdb_line(measurement: str, value: float, timestamp: datetime, tags:
         try:
           metric_type = metric.split('|')[1]
           metric_name = metric.split(':')[0]
-          metric_value = metric.split('|')[0].split(':')[1]
+          metric_value = float(metric.split('|')[0].split(':')[1])
 
           if metric_type == METRIC_TYPE.GAUGE:
             gauges[metric_name] = metric_value
+          if metric_type == METRIC_TYPE.SAMPLE:
```
adeebshihadeh, line 99:
```suggestion
          elif metric_type == METRIC_TYPE.SAMPLE:
```

selfdrive/statsd.py
```diff
@@ -73,6 +80,7 @@ def get_influxdb_line(measurement: str, value: float, timestamp: datetime, tags:
 
   last_flush_time = time.monotonic()
   gauges = {}
+  samples = {}
```
adeebshihadeh, line 83:
```suggestion
  samples = defaultdict(list)
```

selfdrive/statsd.py
```diff
@@ -84,10 +92,14 @@ def get_influxdb_line(measurement: str, value: float, timestamp: datetime, tags:
...
           metric_type = metric.split('|')[1]
           metric_name = metric.split(':')[0]
-          metric_value = metric.split('|')[0].split(':')[1]
+          metric_value = float(metric.split('|')[0].split(':')[1])
 
           if metric_type == METRIC_TYPE.GAUGE:
             gauges[metric_name] = metric_value
+          if metric_type == METRIC_TYPE.SAMPLE:
+            if metric_name not in samples.keys():
+              samples[metric_name] = []
```
adeebshihadeh, line 101: this can be removed with a defaultdict

selfdrive/statsd.py
```diff
@@ -104,8 +116,22 @@ def get_influxdb_line(measurement: str, value: float, timestamp: datetime, tags:
...
+        result += get_influxdb_line(f"sample.{sample_key}.min", samples[sample_key][0], current_time, tags)
+        result += get_influxdb_line(f"sample.{sample_key}.max", samples[sample_key][-1], current_time, tags)
+        result += get_influxdb_line(f"sample.{sample_key}.mean", sample_sum / sample_count, current_time, tags)
+        for percentile in [0.05, 0.5, 0.95]:
+          value = samples[sample_key][int(round(percentile * (sample_count - 1)))]
+          result += get_influxdb_line(f"sample.{sample_key}.p{int(percentile * 100)}", value, current_time, tags)
+
       # clear intermediate data
       gauges = {}
+      samples = {}
```
adeebshihadeh, line 134:
```suggestion
      gauges.clear()
      samples.clear()
```

selfdrive/statsd.py
```diff
@@ -104,8 +116,22 @@ def get_influxdb_line(measurement: str, value: float, timestamp: datetime, tags:
...
+        sample_count = len(samples[sample_key])
+        sample_sum = sum(samples[sample_key])
+
+        result += get_influxdb_line(f"sample.{sample_key}.count", sample_count, current_time, tags)
+        result += get_influxdb_line(f"sample.{sample_key}.min", samples[sample_key][0], current_time, tags)
+        result += get_influxdb_line(f"sample.{sample_key}.max", samples[sample_key][-1], current_time, tags)
+        result += get_influxdb_line(f"sample.{sample_key}.mean", sample_sum / sample_count, current_time, tags)
+        for percentile in [0.05, 0.5, 0.95]:
+          value = samples[sample_key][int(round(percentile * (sample_count - 1)))]
+          result += get_influxdb_line(f"sample.{sample_key}.p{int(percentile * 100)}", value, current_time, tags)
```
adeebshihadeh, line 130:
```suggestion
      for key, values in samples.items():
        values.sort()
        sample_count = len(values)
        sample_sum = sum(values)

        result += get_influxdb_line(f"sample.{key}.count", sample_count, current_time, tags)
        result += get_influxdb_line(f"sample.{key}.min", values[0], current_time, tags)
        result += get_influxdb_line(f"sample.{key}.max", values[-1], current_time, tags)
        result += get_influxdb_line(f"sample.{key}.mean", sample_sum / sample_count, current_time, tags)
        for percentile in (0.05, 0.5, 0.95):
          value = values[int(round(percentile * (sample_count - 1)))]
          result += get_influxdb_line(f"sample.{sample_key}.p{int(percentile * 100)}", value, current_time, tags)
```

gregjhogan: I feel like these should be fields on a single metric instead of separate metrics?

robbederks: that does indeed sound better, did it this way since statsd seems to also make multiple entries (not exactly clear why).

gregjhogan: tags/fields are not features of vanilla statsd

selfdrive/statsd.py
```diff
@@ -104,8 +116,22 @@ def get_influxdb_line(measurement: str, value: float, timestamp: datetime, tags:
       for gauge_key in gauges:
         result += get_influxdb_line(f"gauge.{gauge_key}", gauges[gauge_key], current_time, tags)
```
gregjhogan, line 117: Should we eliminate the metric type prefix?  Using datadog for example I can't think of a time where I wished I had this.  The fields on the metric make it obvious I guess.

robbederks: yeah, I'm open to that. Would have to think about all the metrics we might want to implement in the future (counters, ...), but in any case it's usually indeed pretty obvious or easily looked up in the code if needed

gregjhogan:
We should probably move dongle id from a tag to a field.  Then it doesn't contribute to series cardinality and I expect using it as a secondary filter will still be fast.  The trade-off is poor performance grouping by it at the top level, which I am inclined to say is not useful.

robbederks:
> We should probably move dongle id from a tag to a field. Then it doesn't contribute to series cardinality and I expect using it as a secondary filter will still be fast. The trade-off is poor performance grouping by it at the top level, which I am inclined to say is not useful.

Sounds reasonable, will make a PR

PR #23561, opened 2022-01-18
boardd: add new class USBDeviceList

deanlee:
split from https://github.com/commaai/openpilot/pull/23254. tested on device


selfdrive/boardd/boardd.cc
```diff
@@ -426,9 +427,8 @@ void panda_state_thread(PubMaster *pm, std::vector<Panda *> pandas, bool spoofin
     send_peripheral_state(pm, peripheral_panda);
     ignition = send_panda_states(pm, pandas, spoofing_started);
 
-    // TODO: make this check fast, currently takes 16ms
     // check if we have new pandas and are offroad
-    if (!ignition && (pandas.size() != Panda::list().size())) {
+    if (!ignition && (pandas.size() != USBDeviceList(usb_ctx).size())) {
```
deanlee, line 431: USBDeviceList(usb_ctx).size() takes less than 0.5ms

deanlee, line 431: USBDeviceList(usb_ctx).size() takes less than 0.5ms

PR #23564, opened 2022-01-18
boardd: move main() to main.cc for test cases

deanlee:


pd0wm:
Why are all the includes in boardd.h?

PR #23579, opened 2022-01-19
thermald: check startup_conditions after ignition is on

deanlee:


adeebshihadeh:
We were thinking of moving the startup conditions to manager instead.

PR #23581, opened 2022-01-19
swaglog: retry zmq_send on EAGAIN

deanlee:
From zmq documentation zmq_send(3)  http://api.zeromq.org/3-2:zmq-send:

```
EAGAIN
    Non-blocking mode was requested and the message cannot be sent at the moment.
```


pd0wm:
Think this is on purpose. We'd rather drop a log message than block the calllsite.

deanlee:
Yes, after reading the docs carefully and doing some testing, I also don't think this PR makes sense.

PR #23582, opened 2022-01-19
Toyota: move gas pedal definitions into common DBC

adeebshihadeh:


cydia2020:
I don't have the non-hybrid accelerator address on my Prius, so if anyone can confirm the same for their cars, then I don't think we need to use a car list, we can just use an `or`

adeebshihadeh:
I checked an ICE and a Hybrid, and unfortunately the other signals are missing on those cars. Looks like we're stuck with this unless we can find a new signal to replace them.

PR #23585, opened 2022-01-20
Best practice for f/w values... \0 -> \x

ErichMoraga:
I did this among other changes in a PR I submitted a while back for @martinl, but since that got closed out, and he got Subaru FPv2 merged into Comma's code (even though it had issues last time I checked), I'll chisel away here in piecemeal fashion.


selfdrive/car/subaru/values.py
```diff
@@ -280,7 +280,7 @@ class CAR:
       b'\x00\x00c\x94\x00\x00\x00\x00',
       b'\x00\x00c\x94\x1f@\x10\b',
       b'\x00\x00c\xb7\x1f@\x10\x16',
-      b'\000\000c\x90\037@\020\016',
+      b'\x00\x00c\x90\x37@\x20\x16',
```
pd0wm, line 283: Be careful, the `\0xxx` values are in octal, so `\020` is `\x10`. Can you also replace all the regular ascii characters by their hex equivalent?

spektor56: curious why you guys decided to have ascii character escapes and such and display FW as a string instead of just using the raw byte array values like 1C-C7-71-1C-C7-71-BC-3F.

pd0wm: On some cars they do encode to ascii nicely (e.g. Toyota/Honda). For subaru they clearly don't so let's convert them all to hex.

pd0wm: That's due to some auto format detection that goes on inside the capnp C library when you print the struct. About a month ago I changed useradmin.comma.ai to show the bytearray correctly by converting to a dict, and then using `pprint` to format it.

gregjhogan:
Doing this by hand seems unwise (it could be a script) and if we care about this it should be checked in CI so we aren't fixing this perpetually?

adeebshihadeh:
I agree with @gregjhogan. This is impossible to review, and I don't think it matters all that much.

PR #23587, opened 2022-01-20
ui/camerad: use only yuv buffers

jwooning:
continuation of #22905


geohot:
Can the debayer change be done separately from the UI change to YUV? This pull request is quite long, is it possible to break into sane <300 line chunks?

PR #334, opened 2018-08-27
Adds 2019 Honda Pilot

energee:

PR #23590, opened 2022-01-21
Reverse Car Port: Merge Pilot_2019 into Pilot

vanillagorillaa:
The 2019 Pilot was made it's own car port due to having a different radar keep alive command (back when camera was disconnected and we had to keep the radar happy). https://github.com/commaai/openpilot/pull/334 as well as using a different message and signal for car_gas. The car_gas code was refactored in https://github.com/commaai/openpilot/pull/22010 so now all Honda's use the same message/signal. 

Because of these two differences no longer being an issue, I have gone ahead and merged these into one. Since the Honda Pilot had some firmware versions that the Pilot_2019 did not, and to avoid breaking support for these, I have removed these values and left only the common values between each vehicle. 

Probably want to get someone with a 2019 Pilot to test this


pd0wm:
Any reason why we shouldn't just merge the FW versions?

vanillagorillaa:
For ECU's that both vehicle definitions had this is what I did. I removed a few ECU's that were only defined in one vehicle because I think this would cause some vehicles to fail fingerprinting unless they had that version. Unless I'm wrong on this, in which case I can certainly add those back in.

VirtuallyChris:
Yeah you’re right about that

pd0wm:
If the ECUs are not in the list of essential ECUs (https://github.com/commaai/openpilot/blob/master/selfdrive/car/fw_versions.py#L255) `[Ecu.engine, Ecu.eps, Ecu.esp, Ecu.fwdRadar, Ecu.fwdCamera, Ecu.vsa]`, being absent shouldn't be an issue.

vanillagorillaa:
The ones originally removed are non essential, so added back in, and now we just have a simple FW merge

pd0wm:
LGTM

PR #23591, opened 2022-01-21
controlsd: cleaner check for softDisabling

deanlee:


adeebshihadeh:
Simplified in e50efd7671e8c6a69f50eab794d33fa0c27b5196

PR #23592, opened 2022-01-21
thermald: cleanup check for thermal status

deanlee:


robbederks:
Not sure if I like this change, this relies on the fact that the capnp `ThermalStatus` enum is sorted in the same way `THERMAL_BANDS` is, which might not be the case in the future (say if we want to add another ThermalStatus which is between green and yellow for example).

The way it is written now, the order of the bands is explicitly set by the `OrderedDict`.

PR #23595, opened 2022-01-21
Add more type hinting.

RyanDale:
**Description** Add type hinting to increase mypy coverage in the selfdrive module.


selfdrive/statsd.py
```diff
@@ -4,6 +4,7 @@
 import time
 from pathlib import Path
 from datetime import datetime, timezone
+from typing import NoReturn
```
adeebshihadeh, line 7:
```suggestion
from typing import NoReturn

```

PR #23600, opened 2022-01-21
wifimanager: use tuple for comparision

deanlee:
use tuple for comparision https://en.cppreference.com/w/cpp/utility/tuple/operator_cmp


adeebshihadeh:
The existing code is more readable.

PR #23608, opened 2022-01-23
replay: ncurses GUI 

deanlee:
~it's still very drafty.~

update:

replay

### DIFF
--- a/selfdrive/ui/replay/consoleui.cc
+++ b/selfdrive/ui/replay/consoleui.cc
@@ -14,12 +14,12 @@ enum Color {
   White,
   Debug,
   Yellow,
+  Green,
   Red,
   bgWhite,
   BrightWhite,
   Engaged,
   Disengaged,
-  Green,
 };
 
 template <typename T>
@@ -27,12 +27,9 @@ void add_str(WINDOW *w, T str, Color color = Color::None, bool bold = false) {
   if (color != Color::None) wattron(w, COLOR_PAIR(color));
   if (bold) wattron(w, A_BOLD);
 
-  if constexpr (std::is_same<T, const char *>::value)
-    waddstr(w, str);
-  else if constexpr (std::is_same<T, std::string>::value)
-    waddstr(w, str.c_str());
-  else
-    waddch(w, str);
+  if constexpr (std::is_same<T, const char *>::value) waddstr(w, str);
+  else if constexpr (std::is_same<T, std::string>::value) waddstr(w, str.c_str());
+  else waddch(w, str);
 
   if (bold) wattroff(w, A_BOLD);
   if (color != Color::None) wattroff(w, COLOR_PAIR(color));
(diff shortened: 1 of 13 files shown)


adeebshihadeh:
Wow, looks great so far!

selfdrive/ui/replay/replay.cc
```diff
@@ -269,10 +293,11 @@ void Replay::startStream(const Segment *cur_segment) {
   // write CarParams
   it = std::find_if(events.begin(), events.end(), [](auto e) { return e->which == cereal::Event::Which::CAR_PARAMS; });
   if (it != events.end()) {
+    car_name_ = (*it)->event.getCarParams().getCarName();
```
adeebshihadeh, line 296: use carParams.carFingerprint instead

deanlee:
Show different types of alerts
alert

adeebshihadeh:
Is there a nice library we can use on top or instead of ncurses to make this more readable?

deanlee:
I think the c library is more clear and straightforward than other libraries in this case. there are less than 200 lines of code in consoleui responsible for displaying the interface, the usage pattern of other libraries is very similar to C (set attr, move to position, print, close attr).  they basically don't make the code smaller or easier to read.
I'll cleanup the current code to make it easier to read.

selfdrive/ui/replay/consoleui.cc
```diff
@@ -0,0 +1,340 @@
...
+  sm.update(0);
+  if (sm.updated("carState")) {
+    write_item(0, 25, "SPEED:   ", util::string_format("%.2f", sm["carState"].getCarState().getVEgo()), " m/s");
+  }
+  if (sm.updated("liveParameters")) {
+    auto p = sm["liveParameters"].getLiveParameters();
+    write_item(1, 0, "STIFFNESS: ", util::string_format("%.2f", p.getStiffnessFactor() * 100), " deg");
+    write_item(1, 25, "STEER RATIO: ", util::string_format("%.2f", p.getSteerRatio()), "");
+    auto angle_offsets = util::string_format("%.2f|%.2f", p.getAngleOffsetAverageDeg(), p.getAngleOffsetDeg());
+    write_item(2, 0, "ANGLE OFFSET(AVG|INSTANCE): ", angle_offsets, " deg");
```
robbederks, line 162:
```suggestion
    write_item(2, 0, "ANGLE OFFSET(AVG|INSTANT): ", angle_offsets, " deg");
```

selfdrive/ui/replay/consoleui.cc
```diff
@@ -0,0 +1,356 @@
...
+
+  sm.update(0);
+
+  if (status != Status::Paused) {
+    status = (sm.updated("carState") || sm.updated("liveParameters")) ? Status::Playing : Status::Waiting;
+  }
+  auto [status_str, status_color] = status_text[status];
+  write_item(0, 0, "STATUS:    ", status_str, "      ", false, status_color);
+  write_item(0, 25, "TIME:  ", format_seconds(replay->currentSeconds()),
+             (" / " + format_seconds(replay->totalSeconds())).c_str(), true);
```
adeebshihadeh, line 158: can we add current segment too? like 1/11

PR #23609, opened 2022-01-23
setup: pipenv auto-activate

gregjhogan:
RE: https://github.com/commaai/openpilot/pull/23361#issuecomment-1003829972

### DIFF
--- a/tools/ubuntu_setup.sh
+++ b/tools/ubuntu_setup.sh
@@ -116,11 +116,21 @@ fi
 # install python dependencies
 $ROOT/update_requirements.sh
 
-source ~/.bashrc
+RC_FILE="${HOME}/.$(basename ${SHELL})rc"
+source "$RC_FILE"
 if [ -z "$OPENPILOT_ENV" ]; then
   printf "\nsource %s/tools/openpilot_env.sh" "$ROOT" >> ~/.bashrc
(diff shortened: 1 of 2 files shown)


tools/ubuntu_setup.sh
```diff
@@ -116,13 +116,23 @@ fi
...
 $ROOT/update_requirements.sh
 
-source ~/.bashrc
+RC_FILE="${HOME}/.$(basename ${SHELL})rc"
+source "$RC_FILE"
 if [ -z "$OPENPILOT_ENV" ]; then
   printf "\nsource %s/tools/openpilot_env.sh" "$ROOT" >> ~/.bashrc
   source ~/.bashrc
   echo "added openpilot_env to bashrc"
 fi
```
gregjhogan, line 125: I didn't add using `$RC_FILE` because https://github.com/commaai/openpilot/pull/23361 deletes these lines

gregjhogan:
I didn't put this in `update_requirements.sh` because this certainly isn't a requirement and it should only run once.  I guess we should move the pyenv installation into the OS specific setup scripts, too?

adeebshihadeh:
It sounded nice at the time, but thinking about it more, this just spams more stuff on the user's system. How about we just link to this project in the setup readme?

PR #23610, opened 2022-01-23
C3: Use Wide Road Camera as Light Sensor for Screen Brightness

krkeegan:
**Description**
Using the narrow camera as a light sensor meant that tail lights and head lights of opposing vehicles and even the reflection of our own headlights dominated the sensor value at night even though the broader overall scene is very dark.  Personally, I have felt that the screen brightness at night can be too bright, and I have also seen a number of comments in discord about how bright the C3 is at night.  

This has very little, I dare say no, effect during daylight hours.  Even when driving at dusk, the resulting brightness doesn't change much at all. The change affects nighttime driving the most, resulting in a dimmer screen when it is dark, but still turning brighter when in a bright intersection. 

I think this makes more sense as the wide angle camera's field of view is much closer to the human eye's field of view than the narrow road camera.

**Verification**
I drove around during daylight, dusk, and nighttime to view the screen brightness.  I also plotted a comparison of the calculated screen brightness for both the wide and narrow camera for all drives.

**Routes**
I defined two Custom Series in Plot Juggler to plot the resulting brightness values calculated from the narrow and wide cameras, the only thing I wasn't able to do, was to incorporate the integration over time, so the values are more spiky than they would actually be.  I can share the plot juggler template used to plot this if it helps:

- Nighttime - d50727925674b294|2022-01-18--18-00-44
image
As you can see, the resulting brightness is much lower.  The segment where the wide camera is at 10% and the road camera averages 30% looks like below, this seems pretty dark to me:
vlcsnap-2022-01-23-10h00m15s586
The narrow camera sees the scene as this:
vlcsnap-2022-01-23-10h11m55s184

- Dusk - d50727925674b294|2022-01-21--17-12-28
image
There is a slight difference here, but really only in the darker scenes.

- Daytime - d50727925674b294|2022-01-20--14-35-59
image
There is essentially no difference during daylight hours.

### DIFF
--- a/selfdrive/ui/ui.cc
+++ b/selfdrive/ui/ui.cc
@@ -166,16 +166,22 @@ static void update_state(UIState *s) {
       }
     }
   }
-  if (sm.updated("roadCameraState")) {
+  if (!Hardware::TICI() && sm.updated("roadCameraState")) {
     auto camera_state = sm["roadCameraState"].getRoadCameraState();
 
     float max_lines = Hardware::EON() ? 5408 : 1904;
     float max_gain = Hardware::EON() ? 1.0: 10.0;
     float max_ev = max_lines * max_gain;
 
-    if (Hardware::TICI()) {
-      max_ev /= 6;
-    }
+    float ev = camera_state.getGain() * float(camera_state.getIntegLines());
+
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
This looks good to me. @pd0wm what do you think?

PR #23613, opened 2022-01-24
Refactor condition where sound_stats are effected depending on s->isPlaying()

BBBmau:


adeebshihadeh:
I think these lines are fine as is.

PR #23617, opened 2022-01-25
GM: Change from brake value threshold to Brake_Pressed signal

JMPZ11:
**Description** [] The Suburban (and Silverado) were encountering random disengages while operating with Stock ACC.
random disengage

Logs indicated a Pedal pressed event. It was discovered that the brake position value when unpressed fluctuates normally between 6 and 9 (after scaling) but occasionally hits 10. Openpilot's threshold is >10 - meaning anything over 9 was seen as a press. The 10 was erroneously interpreted as a brake press. Since we know the value likes to reach 10, I bumped the threshold to 12 (So a max value of 11 means zero).

In this screenshot from Cabana, you can see the LKAS disabling immediately after a brake position value of 10:

suburban_high_brake_float

Whereas here is the brake tap to disengage, followed by actual application of the brake while taking an exit:
BrakePosition-tap

**Verification** [] After applying this (and the corresponding panda change) vehicle was driven 3 hours without any spurious pedal press events.
none_of_those_disengages
no_unexpected_disengages

NOTE: This fix also requires the same adjustment be made panda side - here is the associated PR: https://github.com/commaai/panda/pull/837

### DIFF
--- a/selfdrive/car/gm/carstate.py
+++ b/selfdrive/car/gm/carstate.py
@@ -31,10 +31,11 @@ class CarState(CarStateBase):
     ret.standstill = ret.vEgoRaw < 0.01
 
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(pt_cp.vl["ECMPRDNL"]["PRNDL"], None))
-    ret.brake = pt_cp.vl["EBCMBrakePedalPosition"]["BrakePedalPosition"] / 0xd0
-    # Brake pedal's potentiometer returns near-zero reading even when pedal is not pressed.
-    if ret.brake < 10/0xd0:
-      ret.brake = 0.
+    ret.brakePressed = bool(pt_cp.vl["ECMEngineStatus"]["Brake_Pressed"])
+    ret.brake = 0.
+    
+    if ret.brakePressed:
+      ret.brake = pt_cp.vl["EBCMBrakePedalPosition"]["BrakePedalPosition"] / 0xd0
 
     ret.gas = pt_cp.vl["AcceleratorPedal2"]["AcceleratorPedal2"] / 254.
     ret.gasPressed = ret.gas > 1e-5
@@ -67,7 +68,6 @@ class CarState(CarStateBase):
     ret.espDisabled = pt_cp.vl["ESPStatus"]["TractionControlOn"] != 1
     self.pcm_acc_status = pt_cp.vl["AcceleratorPedal2"]["CruiseState"]
 
-    ret.brakePressed = ret.brake > 1e-5
     # Regen braking is braking
     if self.car_fingerprint == CAR.VOLT:
       ret.brakePressed = ret.brakePressed or bool(pt_cp.vl["EBCMRegenPaddle"]["RegenPaddle"])
@@ -106,6 +106,7 @@ class CarState(CarStateBase):
       ("TractionControlOn", "ESPStatus", 0),
       ("EPBClosed", "EPBStatus", 0),
       ("CruiseMainOn", "ECMEngineStatus", 0),
+      ("Brake_Pressed", "ECMEngineStatus", 0),
     ]
 
     checks = [


adeebshihadeh:
Is there a "brake pressed" bit we can use here instead? Generally, we find single bits like that are more generic across models, which will help as more GMs are supported.

JasonJShuler:
> Is there a "brake pressed" bit we can use here instead? Generally, we find single bits like that are more generic across models, which will help as more GMs are supported.

Turns out there is! (There are a couple) ECMEngineStatus (which is already required) has Brake_Pressed which appears to be good.

good_brake_pressed

This will require more testing. Also I am not familiar with the process for confirming that this behaves the same as the old method on existing cars.

Shall I close these while I investigate / possibly switch?

jyoung8607:
> This will require more testing. Also I am not familiar with the process for confirming that this behaves the same as the old method on existing cars.
> 
> I will investigate if the new signal is reliable across vehicles.

DM me and I'll get you through it.

adeebshihadeh:
For the brake pressed signal, we want to be very certain of the new signal. I assume @jyoung8607 is setting you up to check all the CI routes, which is a good first step. Once it looks good there, I'll verify the signal matches on a lot more data.

JasonJShuler:
> For the brake pressed signal, we want to be very certain of the new signal. I assume @jyoung8607 is setting you up to check all the CI routes, which is a good first step. Once it looks good there, I'll verify the signal matches on a lot more data. Converting to draft until the new signal is verified on all the test routes.

Is there anything I should do different when submitting interdependent PRs? Several times you mentioned you have encountered an unexpected issue while merging - am I doing something wrong?

Presuming it passes this initial review, should I do anything different (beyond the implementation and test updates)?

jyoung8607:
> Is there anything I should do different when submitting interdependent PRs? Several times you mentioned you have encountered an unexpected issue while merging - am I doing something wrong?
> 
> Presuming it passes this initial review, should I do anything different (beyond the implementation and test updates)?

You're not doing anything wrong. On a change like this with safety relevance (disengage on brake is about as important as it gets) for already-upstream supported cars, we need to do some diligence to make 100.000% sure it's a good change.

The first pass is to check what we can against the data available to us, which we can quickly eyeball with PlotJuggler against the CI routes. You have the means for that in #23625. An actual scripted check of the entirety of the CI logs would be ideal. The second pass is to crunch all the GM vehicle routes available to comma, same thing we're doing but at much larger scale. That has to be done by comma because we (for obvious reasons) can't look at non-public routes from devices that don't belong to us.

@adeebshihadeh: do you already have something we could use? I could script something to mechanically parse the CAN logs, to check that ECMEngineStatus.Brake_Pressed is always true when EBCMBrakePedalPosition.BrakePedalPosition exceeds some threshold, call it 12 or 15 or something, but if you have something that's already friendly to your pipeline that we could adapt, I'd rather not reinvent the wheel.

adeebshihadeh:
> @adeebshihadeh: do you already have something we could use? I could script something to mechanically parse the CAN logs, to check that ECMEngineStatus.Brake_Pressed is always true when EBCMBrakePedalPosition.BrakePedalPosition exceeds some threshold, call it 12 or 15 or something, but if you have something that's already friendly to your pipeline that we could adapt, I'd rather not reinvent the wheel.

Nope, I'll probably just write a quick notebook once you guys hand this off. There's so few supported GMs that it's feasible to manually check with PlotJuggler. I'd suggest writing a custom function with the current brakePressed threshold to compare against the new bit.

jyoung8607:
@adeebshihadeh: I think this is a good change. The old signal is clearly a brake pressure, and the new proposed signal is a Boolean, most likely an electrical brake-light switch. Plots below, `Brake_Pressed` scale set to 10x to visualize the current GM detection threshold against the current brake pressure value.

There may be some tiny differences in behavior. In the Acadia screenshot, I zoom in to show a couple places where the precise time of rising and falling edges may shift a tiny bit. In the Regal screenshot, you can see two instances (~655 and ~905) where we would have had spurious disengagement just like Suburban and Silverado. I think changing is correct in both cases.

@JasonJShuler: When you're ready to proceed, instead of changing this threshold from 10 to 12, instead let's use this new Boolean signal for brake_pressed. We'll need it changed in both Panda and openpilot at the same time.

Side note: I really have no idea what's going on with the brake pressure [divisions by 0xd0](https://github.com/commaai/openpilot/blob/818d4a0a96cce12d9952c0a111f755fe7ac9fdc1/selfdrive/car/gm/carstate.py#L34-L37). If that's a legitimate scaling factor, it should be done in the DBC, not in CarState.

GM.ACADIA
gm-brake-signal-acadia

GM.BUICK_REGAL
gm-brake-signal-regal

GM.ESCALADE_ESV
gm-brake-signal-escalade

GM.VOLT
gm-brake-signal-volt

JasonJShuler:
I do believe this is all set and ready for further review. Both PRs are updated, tested onroad.

adeebshihadeh:
@sshane will verify this next week

sshane:
Unfortunately I'm seeing quite a few instances of this new bit not catching what the previous threshold check would. This is on a Volt (looks like the RegenPaddle signal also doesn't work): https://user-images.githubusercontent.com/25857203/152100124-8a75c0dc-4f04-402d-848c-d9a2c4dcf76f.png

JasonJShuler:
> Unfortunately I'm seeing quite a few instances of this new bit not catching what the previous threshold check would. This is on a Volt (looks like the RegenPaddle signal also doesn't work): https://user-images.githubusercontent.com/25857203/152100124-8a75c0dc-4f04-402d-848c-d9a2c4dcf76f.png
> 
> Did a check on all the bits in that segment, and here's the bits that match the closest to the current brake pressed check. You can see your signal here. There is another bit inside `EBCMBrakePedalPosition` that's slightly more accurate and activates earlier/deactivates later, so I'll check that out. There also may be special cases where it might not match that we don't care about, like in reverse or park which I'll also check.
> 
> ```
> hex(msg)='0xbe', bit_mismatches=245 of 4838 (5.06%), byt_idx=1, bit_idx=3
> hex(msg)='0xf1', bit_mismatches=295 of 6181 (4.77%), byt_idx=1, bit_idx=3
> hex(msg)='0xf1', bit_mismatches=276 of 6181 (4.47%), byt_idx=1, bit_idx=2
> hex(msg)='0xc9', bit_mismatches=59 of 4837 (1.22%), byt_idx=5, bit_idx=7
> hex(msg)='0xf1', bit_mismatches=66 of 6181 (1.07%), byt_idx=0, bit_idx=6
> ```
> 
> Bit on `0xf1`: https://imgur.com/a/NEejIWD

I'm not sure I'm seeing this right - the brake position value was in the 30s and the brake pressed bit wasn't set? How confident are we of the data? Are we certain that these instances of differing behavior aren't improvements?

Speaking purely anecdotally, I haven't encountered any significant change in functionality... other than no false disengages

sshane:
Those mismatches actually may be due to the internal threshold being somewhere around 16, that's why it doesn't match up with our check of 10 currently. So it still may be usable, I'll check it out more when I have time

JasonJShuler:
> Those mismatches actually may be due to the internal threshold being somewhere around 16, that's why it doesn't match up with our check of 10 currently. So it still may be usable, I'll check it out more when I have time
> 
> The mismatches in the text are frame counts

Unfortunately this value has no units, so it is difficult to know the difference from say 9 to 11 to 17. I would lean toward letting the vehicle decide and use the flag, but I can implement either way and I accept whatever you guys decide

selfdrive/car/gm/carstate.py
```diff
@@ -106,6 +106,7 @@ def get_can_parser(CP):
       ("TractionControlOn", "ESPStatus", 0),
       ("EPBClosed", "EPBStatus", 0),
       ("CruiseMainOn", "ECMEngineStatus", 0),
+      ("Brake_Pressed", "ECMEngineStatus", 0),
```
sshane, line 109:
```suggestion
      ("Brake_Pressed", "ECMEngineStatus"),
```

selfdrive/car/gm/carstate.py
```diff
@@ -31,10 +31,11 @@ def update(self, pt_cp, loopback_cp):
     ret.standstill = ret.vEgoRaw < 0.01
 
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(pt_cp.vl["ECMPRDNL"]["PRNDL"], None))
-    ret.brake = pt_cp.vl["EBCMBrakePedalPosition"]["BrakePedalPosition"] / 0xd0
-    # Brake pedal's potentiometer returns near-zero reading even when pedal is not pressed.
-    if ret.brake < 10/0xd0:
-      ret.brake = 0.
+    ret.brakePressed = bool(pt_cp.vl["ECMEngineStatus"]["Brake_Pressed"])
```
sshane, line 34:
```suggestion
    ret.brakePressed = pt_cp.vl["ECMEngineStatus"]["Brake_Pressed"] != 0
```

selfdrive/car/gm/carstate.py
sshane: @JasonJShuler Can you make some of these changes?

sshane:
PR moved to https://github.com/commaai/openpilot/pull/23712

JasonJShuler:
Apologies, I was taking a couple days off - looks like you have it under control?

PR #23621, opened 2022-01-25
Prepare for using the ISP

geohot:
First parsing out the structs we know more about.

Also, bpp was 0xc and should be 0xa (for now)

Will push more to this before merge, want to make sure I don't break CI.


geohot:
There's some issue with the i2c of the road cam and wide road cam being mixed up. Wonder what issues this is causing?

geohot:
Should be a no-op, if anything a very minor bug fix.

PR #23625, opened 2022-01-25
PlotJuggler: Support CI test routes

jyoung8607:
Fast and dirty, but helpful for situations like commaai/panda#813 and commaai/panda#837.


jyoung8607:
As written, this loses a behavior wherein juggling a specific segment (e.g. `3cfdec54aa035f3f|2022-01-24--14-24-23--3`) implicitly juggled only that segment, with an additional segment count of zero. Didn't realize it at the time, was trying to get something together quickly for @JasonJShuler. I'll re-refactor to fix it before taking out of Draft.

jyoung8607:
Withdrawing this for now. I still want the capability to easily look at CI routes for other cars, but I think the right way to go about this is to write a downloader script to fetch them, and then all the existing tools and methods for local routes will work.

PR #23634, opened 2022-01-27
Toyota: use common gear packet signal

sshane:
- [x] Needs https://github.com/commaai/opendbc/pull/464


adeebshihadeh:
Make sure to bump opendbc to the merged commit on master, or else CI will fail after this is merged because of an additional check that only runs on master.

PR #23635, opened 2022-01-27
Toyota: Move EPS torque factor to openpilot

sshane:
Previously the factors were defined in the DBC as well as here as the safety param. Now we read non-standard factors from a dict and use the default of `73` for all others. Also need to scale EPS torque in interface now since it's always 1 from the parser.

Needs:
- [x] https://github.com/commaai/opendbc/pull/522

### DIFF
--- a/selfdrive/car/toyota/values.py
+++ b/selfdrive/car/toyota/values.py
@@ -1684,6 +1684,8 @@ DBC = {
   CAR.ALPHARD_TSS2: dbc_dict('toyota_nodsu_pt_generated', 'toyota_tss2_adas'),
 }
 
+# These cars have non-standard EPS torque scale factors. All others are 73
+EPS_SCALE = {CAR.PRIUS: 66, CAR.COROLLA: 88, CAR.LEXUS_IS: 77, CAR.LEXUS_RC: 77, CAR.LEXUS_CTH: 100}
 
 # Toyota/Lexus Safety Sense 2.0 and 2.5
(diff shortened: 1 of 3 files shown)


selfdrive/car/toyota/values.py
```diff
@@ -1684,6 +1685,8 @@ class CAR:
   CAR.ALPHARD_TSS2: dbc_dict('toyota_nodsu_pt_generated', 'toyota_tss2_adas'),
 }
 
+# These cars have non-standard EPS torque scale factors. All others are 73
+EPS_SCALE = defaultdict(lambda: 73, {CAR.PRIUS: 66, CAR.COROLLA: 88, CAR.LEXUS_IS: 77, CAR.LEXUS_RC: 77, CAR.LEXUS_CTH: 100})
```
adeebshihadeh, line 1689: just use a normal dict, then do `EPS_SCALE.get(car, 100)`

sshane: Sure, just means we define it in two places if that's okay.

adeebshihadeh (approved):
LGTM. Feel free to merge after using a normal dict and bumping opendbc

selfdrive/car/toyota/values.py
adeebshihadeh: actually, think you're right. the defaultdict is cleaner

sshane:
Let's wait for tests and I'll merge

PR #23639, opened 2022-01-27
longitudinal planner: disable change cost when stopped. not engaged or gas pressed

pd0wm:


krkeegan:
This might help remove a small amount of lag on starting from a standstill.

HaraldSchafer:
LGTM

PR #23642, opened 2022-01-27
CANParser: remove safe default values

sshane:
Since we now wait for can to be valid, we will never need safe default signal values (like setting doors to open, seatbelt unbuckled, etc).

For the parsers in the radar interfaces, some interfaces set the default value of the x distance to 0, some set it to 0 but say they set it to 255, and some set it to 255. So it doesn't seem like it matters.
 - [x] ~~Make sure radar packet valid includes can valid, and that all radar interfaces have checks~~ All radar interfaces add a canError if the parser isn't valid: https://github.com/commaai/openpilot/blob/892e1162c0bfa0970cbe262452a70745eccb8ba6/selfdrive/car/chrysler/radar_interface.py#L64


sshane:
Think the failing process replay is fine. There's less `carEvents` in the new replayed log, meaning that before there were events being added for `doorOpen` and such as they had non-zero defaults (all while not initialized, so never shown to the user). Now these signals are defaulted to 0, so it makes sense that silent car events may change.

adeebshihadeh (approved):
LGTM, just make sure that the only process replay diffs are in carState and at the very beginning of the segment

sshane:
- `fakedata|2022-01-20--17-58-25--0`:
  - Reason why alert text is different: `STEER_STATUS` used to have a different default signal, meaning we sent a `steerUnavailable` alert every time before (now `startupMaster`): https://github.com/commaai/openpilot/blob/master/selfdrive/car/honda/carstate.py#L36
- `fakedata|2022-01-20--17-56-40--0`
  - Reason why alert text is different: used to set brake error to 1 as default value with event `brakeUnavailable` (now `startupMaster`)

PR #23661, opened 2022-01-29
Add COROLLA_TSS2 carFw

VAHAMAISEN:
Added Toyota COROLLA_TSS2 fingerprint.


vanillagorillaa:
Please include the dongleID for the device that these firmware versions were gathered from

VAHAMAISEN:
thank you for your reply.

The device dongle ID is f1c4e2a016e4de93.

2022年1月29日(土) 18:57 vanillagorillaa ***@***.***>:

PR #23665, opened 2022-01-30
selfdrive: Increase typing in the selfdrive module.

RyanDale:
**Description** Increase typing in the selfdrive module.


selfdrive/car/car_helpers.py
```diff
@@ -8,12 +12,13 @@
...
 import cereal.messaging as messaging
 from selfdrive.car import gen_empty_fingerprint
+from selfdrive.controls.lib.events import Alert
 
 from cereal import car
 EventName = car.CarEvent.EventName
 
 
-def get_startup_event(car_recognized, controller_available, fw_seen):
+def get_startup_event(car_recognized: bool, controller_available: bool, fw_seen: bool) -> Dict[str, Alert]:
```
pd0wm, line 21: Shouldn't this return a `car.CarEvent.EventName`? Why is mypy not catching this?

RyanDale: Good catch, I've got the wrong return type. mypy won't allow EventName as a type, but all paths result in an int being returned so I set it to that.

I'm unsure as to why mypy missed this. I'll look into that a bit deeper.

RyanDale:
@pd0wm, are there any additional changes you would like for this PR? Thanks!

RyanDale:
> Did you figure out why mypy accepted those wrong types? If I can't trust mypy it makes reviewing this kind of PRs quite time consuming.

Yes, I believe I've tracked that down. `EventName` is defined in capnp and in my testing, I cannot get mypy to recognize any types from a capnp file. For any entity in a capnp file, mypy will allow you to treat it as any type you wish (which is what happened with my initial commit). I've not been able to find a place in openpilot or in any other open source repositories where capnp files are type checked. I'm not too versed in capnp so if this is incorrect, I would be thrilled to know how capnp types can be recognized by mypy.

adeebshihadeh:
> Yes, I believe I've tracked that down. `EventName` is defined in capnp and in my testing, I cannot get mypy to recognize any types from a capnp file. For any entity in a capnp file, mypy will allow you to treat it as any type you wish (which is what happened with my initial commit). I've not been able to find a place in openpilot or in any other open source repositories where capnp files are type checked. I'm not too versed in capnp so if this is incorrect, I would be thrilled to know how capnp types can be recognized by mypy.

I don't think this is a capnp issue. This sounds more like mypy not being strict about things it doesn't understand.

pd0wm:
Also in your first commit the function returned a `Dict[str, Alert]` instead of an `EventName`. If it replaces capnp stuff by `any` this still should have been caught, since surely it knows what a dict is.

oskarkocol:
>Also in your first commit the function returned a Dict[str, Alert] instead of an EventName. If it replaces capnp stuff by any this still should have been caught, since surely it knows what a dict is.

I think this behavior can be enabled with the [--warn-return-any](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-return-any) flag in the [.pre-commit-config.yaml file](https://github.com/commaai/openpilot/blob/e5edbe98da16aeef7c7b9814c918bc7d646427db/.pre-commit-config.yaml#L21-L24). For ambiguous situations like this it'd throw:
```
selfdrive/car/car_helpers.py:31: error: Returning Any from function declared to return "int"
```

I had a quick look and there are only 22 errors like this in 14 files.

RyanDale:
[--warn-return-any](https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-warn-return-any) looks promising, thank you. I'm working through the 22 errors now and will update the PR once they're resolved.

PR #23683, opened 2022-02-01
Honda: fix brake pressed signal

### LINKED ISSUE #23124: Honda: disengagement via braking sometimes triggers unnecessary red alert
### Describe the bug

Since updating to 0.8.11, a frequent occurrence has been that disengaging via braking causes an immediate red alert.

comma three

930ad2d3aa9a7438|2021-12-04--10-08-21

### openpilot version

0.8.11

### Additional info

I do not know if this is model-specific or not.

### LINKED ISSUE #23643: Honda: cleanup brake press detection
openpilot considers two extra conditions for brake pressed that panda doesn't. 

- [ ] determine if brake_switch is necessary, and if it is, do we need two samples?
Going back to the one history, this was done to prevent a mismatch between openpilot and the Honda PCM. Do we need this at all? Is the noise real? What signal does the PCM use?

- [ ] remove the [exception for the Ridgeline, Pilot, and Passport](https://github.com/commaai/openpilot/blob/3ca8e3653b4dd7552c2cca6d3eb6ed76c98240be/selfdrive/car/honda/carstate.py#L291)
 #475 has the history for this one.

The PCM seems to use something different than us for brake presses. Perhaps it's using a different brake pressed signal or it simply has a threshold on the brake pedal position.

sshane:
This fixes the cruise disabled alert (#23124, #23643) when sometimes pressing the brake on most Hondas. This is due to cruise PCM looking at brake switch (and disabling after a few frames), however openpilot looks at brake pressed which is usually significantly delayed. panda also looks at brake pressed, so this never caused a controls mismatch.

Checked 42 total dongles, 199 total routes, and 1349 total segments. So far the switch is never 0 when brake pressed is 1. The only exception is when the car is in park, where brake switch is always 0. Though `or`ing against brake pressed should get us what we want.

Todo:
- [x] ~Check that there's no timestep noise~ - there is https://pastebin.com/jecb2ZRa
- [x] Do same check in panda - https://github.com/commaai/panda/pull/844/
- [x] Needs opendbc PR merged here: https://github.com/commaai/opendbc/pull/548
- [x] ~Verify this fixes the brake grinding issue for:~ - another PR
  - Pilot
  - Passport
  - Ridgeline
- [x] Check more routes for mismatches: https://pastebin.com/4x1f8k1P (these aren't in park but always near the end of drive)


adeebshihadeh:
Let's make sure to check on significantly more segments and across all affected platforms. Also feel free to break out the "pedal grinding" removal into another PR.

selfdrive/car/honda/carstate.py
```diff
@@ -260,18 +259,17 @@ def update(self, cp, cp_cam, cp_body):
...
     else:
       # brake switch has shown some single time step noise, so only considered when
       # switch is on for at least 2 consecutive CAN samples
-      # panda safety only checks BRAKE_PRESSED signal
-      ret.brakePressed = bool(cp.vl["POWERTRAIN_DATA"]["BRAKE_PRESSED"] or
-                              (self.brake_switch and self.brake_switch_prev and cp.ts["POWERTRAIN_DATA"]["BRAKE_SWITCH"] != self.brake_switch_prev_ts))
+      # brake switch rises earlier than brake pressed but is never 1 when in park
+      if cp.updated["POWERTRAIN_DATA"]:
+        brake_switch = cp.vl["POWERTRAIN_DATA"]["BRAKE_SWITCH"] != 0
+        self.brake_switch_on = self.brake_switch_on + 1 if brake_switch else 0
```
sshane, line 270: Occasionally we get two `POWERTRAIN_DATA` messages on one 100Hz cycle, and then the next cycle we don't get any causing update status to be False and brakePressed to fall in openpilot, while panda keeps it True until it gets an update.

Need to keep a counter, as panda only updates its previous value when updated and has no way of knowing if something isn't updated or not as it's fed messages as they come in.

sshane:
Double `POWERTRAIN`/`BRAKE_SWITCH` message brake checks:

```
CS: brake_switch_vals: [], brake_switch_on: 0
test_models/panda vals: []
CS: False, panda: False

CS: brake_switch_vals: [0.0, 0.0], brake_switch_on: 0
test_models/panda vals: [{'CHECKSUM': 0.0, 'COUNTER': 0.0, 'BRAKE_SWITCH': 0.0}, {'CHECKSUM': 3.0, 'COUNTER': 1.0, 'BRAKE_SWITCH': 0.0}]
CS: False, panda: False

CS: brake_switch_vals: [], brake_switch_on: 0
test_models/panda vals: []
CS: False, panda: False

CS: brake_switch_vals: [1.0, 1.0], brake_switch_on: 2.0
test_models/panda vals: [{'CHECKSUM': 9.0, 'COUNTER': 2.0, 'BRAKE_SWITCH': 1.0}, {'CHECKSUM': 0.0, 'COUNTER': 3.0, 'BRAKE_SWITCH': 1.0}]
CS: True, panda: True

CS: brake_switch_vals: [1.0], brake_switch_on: 3.0
test_models/panda vals: [{'CHECKSUM': 3.0, 'COUNTER': 0.0, 'BRAKE_SWITCH': 1.0}]
CS: True, panda: True
```

selfdrive/car/honda/carstate.py
```diff
@@ -260,18 +259,20 @@ def update(self, cp, cp_cam, cp_body):
...
-                              (self.brake_switch and self.brake_switch_prev and cp.ts["POWERTRAIN_DATA"]["BRAKE_SWITCH"] != self.brake_switch_prev_ts))
-
-      self.brake_switch_prev = self.brake_switch
-      self.brake_switch_prev_ts = cp.ts["POWERTRAIN_DATA"]["BRAKE_SWITCH"]
+      # brake switch rises earlier than brake pressed but is never 1 when in park
+      brake_switch_vals = cp.updated["POWERTRAIN_DATA"]["BRAKE_SWITCH"]
+      if len(brake_switch_vals):
+        if cp.vl["POWERTRAIN_DATA"]["BRAKE_SWITCH"] != 0:
+          # use all updated values this cycle, sometimes we receive two messages at once
+          self.brake_switch_on += sum(brake_switch_vals)
```
sshane, line 272: We increment the `self.brake_switch_on` variable if the message is updated this frame, and the last value isn't 0. So if we get two 1's, we increment twice and set `brakePressed` to true. If we get one or more 1's and one 0, then we reset it to 0.

selfdrive/test/test_models.py
```diff
@@ -197,11 +197,12 @@ def test_panda_safety_carstate(self):
 
     checks = defaultdict(lambda: 0)
     for can in self.can_msgs:
-      for msg in can_capnp_to_can_list(can.can, src_filter=range(64)):
+      msgs = can_capnp_to_can_list(can.can, src_filter=range(64))
```
sshane, line 200: Verified through both the CANParser and a full end to end test (sending CAN messages through jungle to panda to boardd to can socket) that the order you send messages in one cycle is the order you get the messages.

selfdrive/test/test_models.py
```diff
@@ -197,11 +197,12 @@ def test_panda_safety_carstate(self):
...
+      CS = self.CI.update(CC, (can_list_to_can_capnp(msgs),))
```
sshane, line 201: Verified through both the CANParser and a full end to end test (sending CAN messages through jungle to panda to boardd to can socket) that the order you send messages in one cycle is the order you get the messages.

adeebshihadeh, line 201:
```suggestion
      msgs = can_capnp_to_can_list(can.can, src_filter=range(64))
      CS = self.CI.update(CC, (msg.as_builder().to_bytes(),))
```

selfdrive/car/honda/carstate.py
```diff
@@ -260,18 +259,20 @@ def update(self, cp, cp_cam, cp_body):
...
+        else:
+          self.brake_switch_on = 0
+      ret.brakePressed = cp.vl["POWERTRAIN_DATA"]["BRAKE_PRESSED"] != 0 or self.brake_switch_on > 1
```
adeebshihadeh, line 275:
```suggestion
      brake_switch = cp.vl["POWERTRAIN_DATA"]["BRAKE_SWITCH"]
      brake_switch_vals = cp.updated["POWERTRAIN_DATA"]["BRAKE_SWITCH"]
      if len(brake_switch_vals) > 1:
        self.brake_switch_prev = brake_switch_vals[-2]
      ret.brakePressed = (cp.vl["POWERTRAIN_DATA"]["BRAKE_PRESSED"] != 0) or (brake_switch and self.brake_switch_prev)
```

selfdrive/test/test_models.py
adeebshihadeh, line 200: this should go in a separate PR

selfdrive/car/honda/carstate.py
```diff
@@ -260,18 +260,20 @@ def update(self, cp, cp_cam, cp_body):
...
-
-      self.brake_switch_prev = self.brake_switch
-      self.brake_switch_prev_ts = cp.ts["POWERTRAIN_DATA"]["BRAKE_SWITCH"]
+      # brake switch rises earlier than brake pressed but is never 1 when in park
+      brake_switch_vals = cp.updated["POWERTRAIN_DATA"]["BRAKE_SWITCH"]
+      brake_switch = cp.vl["POWERTRAIN_DATA"]["BRAKE_SWITCH"] != 0
+      if len(brake_switch_vals):
+        if len(brake_switch_vals) > 1:
+          self.brake_switch_prev = brake_switch_vals[-2] != 0
+        self.brake_switch_active = brake_switch and self.brake_switch_prev
```
sshane, line 274: If we reset active to False each time, then cases where we don't receive an update (so we don't know if brake switch is 0 or 1), but check was previously True, we would fall to 0. Panda maintains its previous `brakePressed` value if it doesn't receive a message.

sshane:
Tests that pass and their CPU usages:

- `fix-honda-brake-pressed/32/pipeline`:
  - C2: 
    ```
    selfdrive.controls.controlsd         54.27%
    selfdrive.controls.radard            6.11%
    ```
  - C3:
    ```
    selfdrive.controls.controlsd         31.10%
    selfdrive.controls.radard            4.24%
    ```
- `fix-honda-brake-pressed/33/pipeline`:
  - C2:
  - C3:

PR #23686, opened 2022-02-01
handle usb error in reading panda state

pd0wm:
Reading the panda health packet can fail. This currently silently returns a zero health packet, which as the ignition flag false, but also sets connected to false. This causes openpilot to go offroad, and makes boardd think it's ok to start reconnecting while onroad.

### DIFF
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -311,10 +311,10 @@ void Panda::set_ir_pwr(uint16_t ir_pwr) {
   usb_write(0xb0, ir_pwr, 0);
 }
 
-health_t Panda::get_state() {
+std::optional<health_t> Panda::get_state() {
   health_t health {0};
-  usb_read(0xd2, 0, 0, (unsigned char*)&health, sizeof(health));
-  return health;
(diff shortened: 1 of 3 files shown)


selfdrive/boardd/panda.cc
```diff
@@ -311,10 +311,10 @@ void Panda::set_ir_pwr(uint16_t ir_pwr) {
   usb_write(0xb0, ir_pwr, 0);
 }
 
-health_t Panda::get_state() {
+std::optional<health_t> Panda::get_state() {
   health_t health {0};
-  usb_read(0xd2, 0, 0, (unsigned char*)&health, sizeof(health));
-  return health;
+  int err = usb_read(0xd2, 0, 0, (unsigned char*)&health, sizeof(health));
+  return err >= 0 ? std::make_optional(health) : std::nullopt;
```
robbederks, line 317: Doesn't usb_read return the amount of read bytes? Maybe we can also return null if the size is different than expected.

pd0wm: usb_read is a control transfer, so happens in a single packet.

PR #19638, opened 2021-01-03
Fix Lexus ES brake message

TheInventorMan:
**Description** Currently, Openpilot reads the BRAKE_PRESSED signal in the BRAKE_MODULE message in order to detect a brake press and disengage. It appears that there is a redundant brake module broadcasting the BRAKE_MODULE2 message, also containing a BRAKE_PRESSED signal. This bugfix logically ORs both signals for brake detection. 

There was an instance where pressing the brake pedal did *not* disengage OP. It was later found that the BRAKE_MODULE2 message was showing (correctly) that the brake was pressed, while the original BRAKE_MODULE message was not. This can be seen at around t=486s [here](https://my.comma.ai/cabana/?route=dc7629e7a6909176%7C2020-01-02--19-39-21&exp=1609555181&sig=ykGfLaAjc3OaQpGrsCsyXQe3ZsfbrXU1N4fAq12cfFs%3D&max=13&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2Fdc7629e7a6909176%2F765ec378546d40a87b75faf6ce5882be_2020-01-02--19-39-21)

**Verification** This is difficult to verify, since a mismatch between BRAKE_MODULE and BRAKE_MODULE2 is an extremely rare event. However, this fix has been tested for thousands of miles in the past year and no issues with disengagement ever happened again.

**Route**
Route: https://my.comma.ai/37041c500fd30100/1609710437282/1609712371455

PR #23694, opened 2022-02-02
Toyota: use proper DBC for the Lexus ESH

adeebshihadeh:
Proper fix for #19638. Now that we have sensible Toyota DBCs, this makes sense. The 2018 Lexus ES is based on the Toyota GA-K platform, but we're using the New MC platform DBC for the ES. The diff between the GA-K and New MC DBCs is just the brake message.

@sshane can you verify this? Let's make sure we can find the behavior mentioned in #19638 and confirm it's fixed.


sshane:
If you check this route for this car: 37041c500fd30100|2020-12-26--15-56-23 the `BRAKE_MODULE` message on 0x226 doesn't exist on bus 0. the `BRAKE_MODULE` message currently used however does exist works.

PR #23696, opened 2022-02-03
Safety Doc: Allow gas non-disengagement

zorrobyte:
Per discussions in discord, PR for gas disengagement toggle


spektor56:
No mention of "MAIN" button? Cancel only turns off ACC in any car I have driven whilst the main button disables both.

adeebshihadeh:
Not sure these are all the changes we want to make, and no use keeping this open. We'll do this ourselves when it's time.

PR #23700, opened 2022-02-03
PC: build with --test by default

sshane:
This builds with --test by default on PC. Useful for if you don't know about this flag and changed panda safety and am running something like `test_models.py`

Time to build all on 32 cores:

before: 83.7986 seconds
after:    86.5663 seconds

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -6,11 +6,14 @@ import sysconfig
 import platform
 import numpy as np
 
+from selfdrive.hardware import PC
+
 TICI = os.path.isfile('/TICI')
 Decider('MD5-timestamp')
 
 AddOption('--test',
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
What does `scons -h` show now? How do I disable tests?

sshane:
The help shows the same args, I figured we always want to build if on PC and only optionally on-device. But I believe you can do `--test false` or 0

adeebshihadeh:
Yeah, always building on PC sounds fine.

adeebshihadeh:
On second thought, I'm not sure how much sense this makes. There shouldn't really be a difference between building for PC and a comma three/two.

PR #23702, opened 2022-02-03
scons: fix duplicate environments with test flag

sshane:
Pre-req for https://github.com/commaai/openpilot/pull/23700

### DIFF
--- a/selfdrive/loggerd/SConscript
+++ b/selfdrive/loggerd/SConscript
@@ -1,13 +1,12 @@
 Import('env', 'arch', 'cereal', 'messaging', 'common', 'visionipc', 'gpucommon')
 
 
-logger_lib = env.Library('logger', ["logger.cc"])
-libs = [logger_lib, common, cereal, messaging, visionipc,
+libs = [common, cereal, messaging, visionipc,
         'zmq', 'capnp', 'kj', 'z',
         'avformat', 'avcodec', 'swscale', 'avutil',
         'yuv', 'bz2', 'OpenCL']
 
-src = [env.Object('loggerd.cc')]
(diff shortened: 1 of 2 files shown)


selfdrive/ui/SConscript
```diff
@@ -47,7 +47,7 @@ asset_obj = qt_env.Object("assets", assets)
 qt_env.Program("soundd/_soundd", ["soundd/main.cc", "soundd/sound.cc"], LIBS=qt_libs)
 if GetOption('test'):
   qt_env.Program("tests/playsound", "tests/playsound.cc", LIBS=base_libs)
-  qt_env.Program('tests/test_sound', ['tests/test_runner.cc', 'soundd/sound.cc', 'tests/test_sound.cc'], LIBS=[base_libs])
+  qt_env.Program('tests/test_sound', ['tests/test_runner.cc', 'soundd/sound.cc', 'tests/test_sound.cc'], LIBS=qt_libs)
```
sshane, line 50: Here, does including `soundd/sound.cc` a second time mean we re-use what we built above? If so we can keep `base_libs` and just wrap sound.cc in a `env.Object()`

selfdrive/loggerd/SConscript
```diff
@@ -7,7 +7,7 @@ libs = [logger_lib, common, cereal, messaging, visionipc,
         'avformat', 'avcodec', 'swscale', 'avutil',
         'yuv', 'bz2', 'OpenCL']
 
-src = ['loggerd.cc']
+src = [env.Object('loggerd.cc')]
```
adeebshihadeh, line 10: build this into a library instead

PR #23708, opened 2022-02-04
Remove dashcam GM cars from cars.md

VirtuallyChris:
We have some dashcam mode cars on the supported cars list.

This PR removes them until someone tests them again


Verylukyguy:
Possibly the Malibu also?

PR #23713, opened 2022-02-05
logcatd_android.cc: Fix for repeating the same message.

deanlee:


adeebshihadeh:
Can we make the test stricter to catch this bug? Maybe we can clear the logcat buffer and whitelist the PID of our test before starting logcat, then we can check all the messages.

https://github.com/commaai/openpilot/blob/master/selfdrive/logcatd/tests/test_logcatd_android.py

deanlee:
the better(probably best) solution is to use blocking mode like `logcat` : 
https://android.googlesource.com/platform/system/core/+/refs/heads/android10-dev/logcat/logcat.cpp#1592
and catch  signal to exit
https://android.googlesource.com/platform/system/core/+/refs/heads/android10-dev/logcat/logcat_main.cpp#25

PR #23714, opened 2022-02-05
Toyota: clean up CarInterface apply function

sshane:
The access of `actuators`, `hudControl`, other values, etc is 2x+ as slow as if we just access it once and made a variable for it, but that adds many lines. I'll see if we can add some kind of caching on the capnp DynamicStructReader

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -22,11 +22,12 @@ class CarController():
     self.gas = 0
     self.accel = 0
 
-  def update(self, c, CS, frame, actuators, pcm_cancel_cmd, hud_alert,
-             left_line, right_line, lead, left_lane_depart, right_lane_depart):
+  def update(self, CC, CS, frame):
+    actuators = CC.actuators
+    hud_control = CC.hudControl
 
     # gas and brake
-    if CS.CP.enableGasInterceptor and c.longActive:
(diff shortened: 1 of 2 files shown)


selfdrive/car/toyota/carcontroller.py
```diff
@@ -121,8 +122,10 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
       # forcing the pcm to disengage causes a bad fault sound so play a good sound instead
       send_ui = True
 
-    if (frame % 100 == 0 or send_ui):
-      can_sends.append(create_ui_command(self.packer, steer_alert, pcm_cancel_cmd, left_line, right_line, left_lane_depart, right_lane_depart, enabled))
+    if frame % 100 == 0 or send_ui:
+      can_sends.append(create_ui_command(self.packer, steer_alert, pcm_cancel_cmd,
+                                         CC.hudControl.leftLaneVisible, CC.hudControl.rightLaneVisible,
+                                         CC.hudControl.leftLaneDepart, CC.hudControl.RightLaneDepart, CC.enabled))
```
sshane, line 128:
```suggestion
                                         CC.hudControl.leftLaneDepart, CC.hudControl.rightLaneDepart, CC.enabled))
```

adeebshihadeh:
Wow, this is noticeably slower?

PR #23717, opened 2022-02-06
remove unused int_rnd function

GregorKikelj:
I've removed some noop int casts from rounded numbers, unofrtunatly not 100% sure why they were there, but the tests seem to pass.


adeebshihadeh:
I reverted some of the changes, since I think it's better to be explicit in this case. For example, the capnp int fields will throw an exception instead of casting for you if you give it a float.

PR #23720, opened 2022-02-07
Corrected EPS value of Prius V

CT921:
* Correct EPS based on tuning information on Discord.
Adjust EPS value to 118 to reduce offset. The difference as linked picture.
[https://drive.google.com/file/d/1NJvYk40zJ3iDzdXfZGuvd_LGhCbOwuHJ/view?usp=sharing](https://drive.google.com/file/d/1NJvYk40zJ3iDzdXfZGuvd_LGhCbOwuHJ/view?usp=sharing)

* Modify the control parameters
Based on the correct EPS value, readjust the control parameters of PRIUS V to improve stability and solve the ping-pong problem.

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -41,10 +41,10 @@ class CarInterface(CarInterfaceBase):
     elif candidate == CAR.PRIUS_V:
       stop_and_go = True
       ret.wheelbase = 2.78
-      ret.steerRatio = 17.4
+      ret.steerRatio = 18.1
       tire_stiffness_factor = 0.5533
       ret.mass = 4387. * CV.LB_TO_KG + STD_CARGO_KG
-      set_lat_tune(ret.lateralTuning, LatTunes.LQR_RAV4)
+      set_lat_tune(ret.lateralTuning, LatTunes.LQR_PV)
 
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
It doesn't really make sense to hand tune the LQR parameters. I suggest trying PID instead.

sshane:
@adeebshihadeh I can check some Prius V routes if this new scaling is correct (and how the value changes based on any factors in the process so we understand it better), or we can keep closed if the non-linearity makes it not matter too much.

CT921:
@sshane Thanks for your support. You can check this route as below, there should be enough distance to confirm that the new EPS is closer to the correct value.
Test route: "14623aae37e549f3|2021-10-24--01-20-49"

@adeebshihadeh 
The STEER RATIO & EPS in the PR that I submitted is closer to the correct value, is it recommended to resubmit PR with PID?
The LQR I modified is based on the article as below, and is provided to about 20 Prius V owners. They all agree that the stability and performance have been greatly improved.
[LQR tuning article](https://hsin-chen5716.medium.com/openpilot%E7%9A%84lqr%E5%8F%83%E6%95%B8%E8%AA%BF%E6%95%B4-3691ce33cc9e)

PR #23721, opened 2022-02-07
thermald: accept ignition from any panda

briskspirit:
Now we check only internal panda for ignition. Let's check for ignition from any panda connected.

### DIFF
--- a/selfdrive/thermald/thermald.py
+++ b/selfdrive/thermald/thermald.py
@@ -253,10 +253,7 @@ def thermald_thread(end_event, hw_queue):
     if sm.updated['pandaStates'] and len(pandaStates) > 0:
 
       # Set ignition based on any panda connected
-      onroad_conditions["ignition"] = False
-      for pandaState in pandaStates:
-        if pandaState.pandaType != log.PandaState.PandaType.unknown:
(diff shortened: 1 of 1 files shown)


selfdrive/thermald/thermald.py
```diff
@@ -251,10 +251,14 @@ def thermald_thread(end_event, hw_queue):
...
     if sm.updated['pandaStates'] and len(pandaStates) > 0:
-      pandaState = pandaStates[0]
 
-      if pandaState.pandaType != log.PandaState.PandaType.unknown:
-        onroad_conditions["ignition"] = pandaState.ignitionLine or pandaState.ignitionCan
+      # Set ignition based on any panda connected
+      onroad_conditions["ignition"] = False
+      for pandaState in pandaStates:
+        if pandaState.pandaType != log.PandaState.PandaType.unknown:
+          onroad_conditions["ignition"] |= pandaState.ignitionLine or pandaState.ignitionCan
```
adeebshihadeh, line 259:
```suggestion
      onroad_conditions["ignition"] = any(ps.ignitionLine or ps.ignitionCan for ps in pandaStates if ps != pandaType != log.PandaState.PandaType.unknown)
```

PR #23726, opened 2022-02-08
paramsd: fix VehicleModelInvalid errors

nuwandavek:
Problem: Nearly ~90% of `vehicleModelInvalid` errors are caused by `steerRatio`/`stiffnessFactor` going out of acceptable bounds. This typically happens when there is extended periods of straight driving when there's no effective information about steer_ratio or stiffness. Additionally, the filter standard deviations are currently unbounded. So the values keep changing (uncertainty keeps increasing) and "correct" rapidly with new information (turns). 

Solution: Bound the  `steerRatio`/`stiffnessFactor` uncertainty by inputting current values at each time step as observations with a high observation noise. Also, increase the steering angle observation noise (Prius least count is 0.1 degrees!)

image

There was no drastic change in offset/yaw rate due to this (check new report).


HaraldSchafer:
Can you just add a comment of what the purpose is of inputting the current stiffness and steer ratio?

zorrobyte:
This is pretty awesome, artificially high steer ratio has been an issue since 0.6!

PR #23732, opened 2022-02-09
Toyota: improve stop and go performance by setting PERMIT_BRAKING to 0 under some conditions

cydia2020:
Improving sng performance by manipulating the `PERMIT_BRAKING` bit on `ACC_CONTROL`

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -108,7 +108,7 @@ class CarController():
       # handle permit braking
       if (CS.out.gasPressed or 1. / DT_CTRL > (frame - self.last_gas_press_frame)) \
          or ((actuators.accel > - 1.95) and (2. / DT_CTRL > (frame - self.last_standstill_frame))) \
-         or ((actuators.accel > - 1.95) and (2. / DT_CTRL > (frame - self.last_zero_speed_frame))):
+         or (2. / DT_CTRL > (frame - self.last_zero_speed_frame)):
         self.permit_braking = False
       else:
         self.permit_braking = True
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -98,7 +98,7 @@ class CarController():
         can_sends.append(create_accel_command(self.packer, pcm_accel_cmd, pcm_cancel_cmd, self.standstill_req, lead, CS.acc_type, CS.out.gasPressed))
         self.accel = pcm_accel_cmd
       else:
-        can_sends.append(create_accel_command(self.packer, 0, pcm_cancel_cmd, False, lead, CS.acc_type))
+        can_sends.append(create_accel_command(self.packer, 0, pcm_cancel_cmd, False, lead, CS.acc_type, 1))
 
     if frame % 2 == 0 and CS.CP.enableGasInterceptor and CS.CP.openpilotLongitudinalControl:
       # send exactly zero if gas cmd is zero. Interceptor will send the max between read value and gas cmd.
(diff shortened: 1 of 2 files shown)


cydia2020:
Would it be preferable if we introduce a 0.75 second delay before setting `PERMIT_BRAKING` to 1 after releasing the accelerator?

Thinking of doing something like this.
```
      # handle permit braking logic
      # record accelerator depression frame
      if CS.out.gasPressed:
        self.last_gas_press_frame = frame
      # set permit braking to false briefly after accelerator release
      if CS.out.gasPressed or 0.75 / DT_CTRL > (frame - self.last_gas_press_frame) and actuators.accel > - 0.07:
        self.permit_braking = False
      else:
        self.permit_braking = True
```

jyoung8607:
Seems like it should be handled higher up in controlsd?

* if it's appropriate, it should be done for all car ports, not just Toyota
* disabling the actuator at the car port level could result in integral windup for the brake input
* what about AEB? (if controlsd really really wants to brake, in the future)

cydia2020:
> Seems like it should be handled higher up in controlsd?
> 
> * if it's appropriate, it should be done for all car ports, not just Toyota (and not by making eleven copies)
> * disabling the actuator at the car port level could result in integral windup for the brake input
> * what about AEB? (if controlsd really really wants to brake, in the future)

This is mainly to resolve how Toyota's PCM handles the brake when the user is overriding the accelerator. If this is set to 1 (which it is in the current version of openpilot), the car's PCM will try to fight the brakes with the user.

KyleGreenlaw:
I think this is what’s happening on Hyundais, specifically my Sonata Hybrid. With gas override, the brakes will start and prevent acceleration. 

I agree it should be higher up not just in Toyota.

sshane:
Usually the car's PCM will prevent any braking when the user has their foot on the gas. Do you have a route with the car actually braking while you're accelerating @cydia2020? And windup is handled in longcontrol.py and longitudinal_planner.py, we treat it as disengaged.

cydia2020:
I have added some other stuff to this, I have been testing this setup for a few days now, and it works great, the car responds better to acceleration commands during SnG.

Some notes:
On Toyotas, openpilot requests -2.0m/s^2 acceleration when the vehicle is stopped, I used -1.95 as a condition for when `permit_braking` should be `1`, this is so that the car won't release the brakes and creep forward when say the user accidently bumps the resume button. It also ensures that openpilot can still request braking force if the lead decelerates suddenly.

After looking into the log data a bit, I have realised that the "fighting for brakes" condition was created by me pressing the accelerator only very lightly, so the accelerator depression did not register with the vehicle's PCM, creating the "fighting for brakes" condition.

selfdrive/car/toyota/carcontroller.py
```diff
@@ -91,14 +96,31 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd, hud_aler
...
+      # record standstill exit frame
+      if CS.pcm_acc_status == 7:
+        self.last_standstill_frame = frame
+      # record last frame when vego is 0
+      if CS.out.vEgo == 0:
+        self.last_zero_speed_frame = frame
+      # handle permit braking
+      if (CS.out.gasPressed or 1. / DT_CTRL > (frame - self.last_gas_press_frame)) \
+         or ((actuators.accel > - 1.95) and (2. / DT_CTRL > (frame - self.last_standstill_frame))) \
+         or (2. / DT_CTRL > (frame - self.last_zero_speed_frame)):
```
sshane, line 111: So we can not brake for up to a second after pressing the gas? I'm not sure this is what we want exactly, and how it affects stop and go performance while openpilot is engaged.

sshane:
Have you tried master after https://github.com/commaai/openpilot/pull/23639 was merged? I'm not seeing that much braking at low speeds at all anymore.

selfdrive/car/toyota/carcontroller.py
cydia2020: Yes I do, however I'm using the pedal only to get the car out of standstill, the pcm takes care of the acceleration afterwards, as seen in #23122

sshane: Perhaps we could do something like if we're undershooting our speed and we don't want to brake, set the bit

sshane:
Is there any update to this?

cydia2020:
Don't think this is needed anymore now, SnG has been smooth. Closing

PR #23734, opened 2022-02-09
C3: detect metered Android hotspot

### LINKED ISSUE #23731: Improve metered connection settings for prime lite
Better detection of phone hotspots:
- [x] C3 Android (https://github.com/commaai/openpilot/pull/23734)
- [ ] C3 iOS
- [ ] C3, allow forcing metered property from wifi settings menu on a connection basis.

Manual override
- [ ] Mark wifi as metered
- [x] Mark cellular connection as unmetered

pd0wm:
Resolves #23731

### DIFF
--- a/selfdrive/hardware/base.py
+++ b/selfdrive/hardware/base.py
@@ -2,7 +2,10 @@ from abc import abstractmethod, ABC
 from collections import namedtuple
 from typing import Dict
 
+from cereal import log
+
 ThermalConfig = namedtuple('ThermalConfig', ['cpu', 'gpu', 'mem', 'bat', 'ambient', 'pmic'])
+NetworkType = log.DeviceState.NetworkType
 
 class HardwareBase(ABC):
   @staticmethod
@@ -66,6 +69,9 @@ class HardwareBase(ABC):
   @abstractmethod
   def get_network_strength(self, network_type):
     pass
+
+  def get_network_metered(self, network_type) -> bool:
+    return network_type not in [NetworkType.none, NetworkType.wifi, NetworkType.ethernet]
 
   @staticmethod
   def set_bandwidth_limit(upload_speed_kbps: int, download_speed_kbps: int) -> None:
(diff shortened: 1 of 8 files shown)


adeebshihadeh:
I was thinking we make this is a separate property. For example, I'd like to treat my hotspot as normal wifi since it's unlimited. I imagine prime lite users with unlimited sims feel similarly.

pd0wm:
I did it like this since a lot things currently depend on just the networkType (uploader, connect, internal scripts). We can also write a metered property with the wifi networkmanager config manually, where you can force mark it as unmetered.

selfdrive/hardware/tici/hardware.py
```diff
@@ -91,11 +99,10 @@ def get_network_type(self):
       primary_connection = self.nm.Get(NM, 'PrimaryConnection', dbus_interface=DBUS_PROPS, timeout=TIMEOUT)
       primary_connection = self.bus.get_object(NM, primary_connection)
       primary_type = primary_connection.Get(NM_CON_ACT, 'Type', dbus_interface=DBUS_PROPS, timeout=TIMEOUT)
-      primary_id = primary_connection.Get(NM_CON_ACT, 'Id', dbus_interface=DBUS_PROPS, timeout=TIMEOUT)
 
       if primary_type == '802-3-ethernet':
         return NetworkType.ethernet
-      elif primary_type == '802-11-wireless' and primary_id != 'Hotspot':
```
pd0wm, line 98: primary connection switches over properly when hotspot enabled, no need for this check.

pd0wm:
Should be using the properties from networkmanager, which can be set on a connection by connection basis for both wifi and lte. We just need to build out the ui side to actually set them.

selfdrive/hardware/base.py
```diff
@@ -67,6 +71,12 @@ def get_sim_info(self):
   def get_network_strength(self, network_type):
     pass
 
+  def get_network_metered(self, network_type) -> bool:
+    if network_type in [NetworkType.none, NetworkType.wifi, NetworkType.ethernet]:
```
adeebshihadeh, line 75:
```suggestion
    return network_type not in (NetworkType.none, NetworkType.wifi, NetworkType.ethernet)
```

pd0wm: Ah, of course. There used to be other code here, but that's now gone.

PR #23735, opened 2022-02-09
VW MQB: Volkswagen Caravelle T6.1

jyoung8607:
**Volkswagen Caravelle T6.1**

The Caravelle is a passenger variant of the Transporter T6.1 platform. See #22283 for more details on the platform.

- [x] owner dongle ID: `76ff0d290ba96cbd`
- [x] added to CARS
- [x] added to RELEASES

Thanks to community Caravelle owner Hormold.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -9,6 +9,7 @@ Version 0.8.13 (2022-XX-XX)
  * Subaru Impreza 2020 support thanks to martinl!
  * Toyota Avalon 2022 support thanks to sshane!
  * Toyota Prius v 2017 support thanks to CT921!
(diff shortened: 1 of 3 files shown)


jyoung8607:
Further note to self: the SWB California and LWB Caravelle actually had identical steering racks with identical characteristic maps loaded. Except for possibly airbags, none of the other fingerprinted components will differ. If this becomes an issue, we're going to have to find two like-model, like-year, like-regulatory-domain T6.1s and see if any of the airbag config bits change. Failing that, we may have to do a special followup query to something else TBD in get_params().

PR #23738, opened 2022-02-10
[PID] sanitize gains - cleanup

ClockeNessMnstr:
ran into division by zero in numpy_fast.py in selfdrive:process-replay
it was empty lists from the capnp list builder being passed into the pid controller for the gains. 

for me it happened because I had a kd term and the toyota interface calls a tune list that invokes the lateraltuning.init() function clearing the values from cars/interfaces.py. 
Although I'm not sure what changed since I last updated that caused this since it looks like the toyota stuff was there on by branch before I updated. 

if you were to exclude BP's or V's for Ki / Kp from the interface for a vehicle it would crash here as well. 

I'm inclined to say "who cares" but I see some sanitation was getting added here for passing values in the wrong format already so I moved it to a dedicated function and added sanitation for missing BP's or excluded gains.

### DIFF
--- a/selfdrive/controls/lib/pid.py
+++ b/selfdrive/controls/lib/pid.py
@@ -33,27 +33,28 @@ class PIController():
     return interp(self.speed, self._k_i[0], self._k_i[1])
 
   def sanitize_gains(self, k_p, k_i, k_f):
+    if isinstance(k_p, Number):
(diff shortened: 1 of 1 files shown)


ClockeNessMnstr:
This wasn't supposed to be difficult. I'll try again some other time lol

PR #23742, opened 2022-02-10
Mazda: add 2017 CX-9 firmware

suryad:


adeebshihadeh:
Which route did you pull these from?

suryad:
I just followed the instructions from https://github.com/commaai/openpilot/wiki/Fingerprinting

adeebshihadeh:
What's your dongle ID?

PR #23743, opened 2022-02-11
Honda: use common DBC for CRV H and Accord Diesel

sshane:
Similar situation to https://github.com/commaai/openpilot/pull/23659. Used the ILX gear definitions and we maintain all previous gears + we get neutral (previously was unknown to openpilot as it wasn't defined correctly for these cars).


sshane:
Ready to merge if the jenkins test fail isn't related

PR #23744, opened 2022-02-11
Hyundai: fix cancel command occasionally not working

### LINKED ISSUE #23674: Hyundai Elantra Hybrid: disengage failure on gas
### Describe the bug

pressing gas did not disable cruise control / openpilot. 

image

comma three

eb5a50f9e98f50b9|2022-01-28--10-48-41--5

### openpilot version

0.8.12

### Additional info

has happened a few times before, but wasn't certain that this was the behavior I experienced.

with the recent development to _not_ disengage on gas, this might be a moot bug report, but submitting it for record's sake.

sshane:
Resolves #23674

Old silent (but *UNSAFE*) branch moved here: https://github.com/sshane/openpilot/tree/hyundai-cancel-cmd-silent

This sends the driver door open bit at a frequency that can't perfectly align with the car's message, as the car reads at 10Hz and we can sometimes always send at the wrong time to not get our message accepted by the SCC module (but CLU reads all messages as the dash flickers?)

When only sending 1 message at 10Hz vs. 25Hz and 10 messages: https://user-images.githubusercontent.com/25857203/157397440-8b0eb4c6-8ad1-4587-83ff-1c354a7073d3.png

### DIFF
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -3,7 +3,8 @@ from common.realtime import DT_CTRL
 from common.numpy_fast import clip, interp
 from selfdrive.config import Conversions as CV
 from selfdrive.car import apply_std_steer_torque_limits
-from selfdrive.car.hyundai.hyundaican import create_lkas11, create_clu11, create_lfahda_mfc, create_acc_commands, create_acc_opt, create_frt_radar_opt
+from selfdrive.car.hyundai.hyundaican import create_lkas11, create_clu11, create_lfahda_mfc, create_acc_commands, \
+                                             create_acc_opt, create_frt_radar_opt, create_cancel_command
 from selfdrive.car.hyundai.values import Buttons, CarControllerParams, CAR
 from opendbc.can.packer import CANPacker
 
@@ -45,6 +46,8 @@ class CarController():
     self.steer_rate_limited = False
     self.last_resume_frame = 0
     self.accel = 0
+    self.cancel_frames = 0
+    self.prev_pcm_cancel = False
 
   def update(self, c, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert, hud_speed,
              left_lane, right_lane, left_lane_depart, right_lane_depart):
@@ -78,14 +81,23 @@ class CarController():
                                    left_lane_warning, right_lane_warning))
 
     if not CS.CP.openpilotLongitudinalControl:
+      if pcm_cancel_cmd and not self.prev_pcm_cancel:
+        self.cancel_frames = 0
       if pcm_cancel_cmd:
-        can_sends.append(create_clu11(self.packer, frame, CS.clu11, Buttons.CANCEL))
+        self.cancel_frames += 1
+        if frame % 10 == 0:
+          print('Sending cancel')
+          can_sends.extend([create_cancel_command(self.packer, CS.cgw1)] * 1)
       elif CS.out.cruiseState.standstill:
         # send resume at a max freq of 10Hz
         if (frame - self.last_resume_frame) * DT_CTRL > 0.1:
           # send 25 messages at a time to increases the likelihood of resume being accepted
           can_sends.extend([create_clu11(self.packer, frame, CS.clu11, Buttons.RES_ACCEL)] * 25)
           self.last_resume_frame = frame
+
+      if not pcm_cancel_cmd and self.prev_pcm_cancel:
+        print('Took {} send frames to cancel'.format(self.cancel_frames))
+      self.prev_pcm_cancel = pcm_cancel_cmd
 
     if frame % 2 == 0 and CS.CP.openpilotLongitudinalControl:
       lead_visible = False
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -46,8 +46,6 @@ class CarController():
     self.steer_rate_limited = False
     self.last_resume_frame = 0
     self.accel = 0
-    self.cancel_frames = 0
-    self.prev_pcm_cancel = False
 
   def update(self, c, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert, hud_speed,
              left_lane, right_lane, left_lane_depart, right_lane_depart):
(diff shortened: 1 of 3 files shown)


gregjhogan:
I believe we have found in the past that the counter doesn't matter specifically for cancel, probably because the implications of not cancelling when the user requests it is bad (but cancelling at any time is acceptable for level 2 systems).  Also, while adding 1 to the counter makes it more likely to be correct, it still isn't guaranteed (the cluster could sneak in a message with the same counter).

It seems a proper fix for this might involve sending 1 (or perhaps more - see how we send resume) cancel messages every cycle until the disengagement happens?

gregjhogan:
Looks like we already should be sending the cancel button as long as we tihnk ACC is engaged but openpilot is disabled?
https://github.com/commaai/openpilot/blob/47211400f55881c4063579826764a42374ae66be/selfdrive/controls/controlsd.py#L566

Was it not spamming the cancel button?  If we are the counter will eventually be correct, so it seems unlikely trying to make the counter match closer will help.

sshane:
Yeah, right now we send a message every cycle with `Buttons.CANCEL` (4) until the car's cruise control turns off (that checks `SCC12`, `ACCMode`. However in the route provided in that issue, the car's cruise control actually disabled for a couple of frames, and then re-enabled. So openpilot thought the car disengage and then the user re-engaged in just a couple of frames: https://user-images.githubusercontent.com/25857203/153541425-709baae6-8528-45a8-a7cb-650b71684d94.png

gregjhogan:
Seems unlikely that cruise control disengaged then spontaneously re-engaged (enabled state comes from the radar, and the radar manages the enabled state), I think we must be missing something.

sshane:
Let me see if there's any other messages/bits set while trying to cancel.

sshane:
I think you're right, `CF_Clu_CruiseSwState` from the radar briefly goes to 4, indicating a user enable (or some kind of enable).

gregjhogan:
> I think you're right, `CF_Clu_CruiseSwState` briefly goes to 4, indicating a user enable (or some kind of enable).

I believe 4 is the cancel button, so not an indication of enable.

sshane:
At least in this route, it always goes to 4 when enabling.

gregjhogan:
Interesting, maybe buttons have different values for this vehicle?  Need to figure out what the value is when cancel is pressed, seems like it can't be 4?

sshane:
Hmm, well the bit is surrounded by already-defined values, and they remain zero when engaging/disengaging.

sunnyhaibin:
@sshane @gregjhogan In recent discovery, we have found out that the Asian (Vietnamese) 2022 Hyundai Santa Fe has some different quirks in the signals. Their `CANCEL` button is named `Paused/Resume`, meaning that the same button acts as a resume.

Also, when you press `Cruise MAIN`, it would also engage SCC, in turn engages openpilot altogether.

Noticed some of the values are different in some signals as well.

Vietnamese 2022 Santa Fe:
image

`Cruise MAIN` press when `SCC11 | MainMode_ACC` was `0`:
image

Stock drive route ID: `ec07911a91f863f3|2022-02-23--21-01-26`

adeebshihadeh:
Need to merge https://github.com/commaai/panda/pull/860 along with this.

CruiseBrantley:
> @sshane @gregjhogan In recent discovery, we have found out that the Asian (Vietnamese) 2022 Hyundai Santa Fe has some quirks in the signals. When you press `Cruise MAIN`, it would also engage SCC, in turn engages openpilot altogether.
> 
> Their `CANCEL` button is named `Paused/Resume`, meaning that the same button acts as a cancel and resume.
> 
> Noticed some of the values are different in some other signals as well. I've attached the stock drive route ID for review.
> 
> Stock drive route ID: `ec07911a91f863f3|2022-02-23--21-01-26`
> 
> `Cruise MAIN` press when `SCC11 | MainMode_ACC` was `0`: image
> 
> Vietnamese 2022 Santa Fe cruise control buttons on the steering wheel: image

Elantra 2021 looks the same and we've seen an error message (Something along the lines of "SCC Requirements not met") on cancelling SCC occasionally that I think may be related. It's just a small alert but always cancels in practice fine. If any routes or anything could help let me know.

selfdrive/car/hyundai/hyundaican.py
```diff
@@ -69,6 +69,14 @@ def create_clu11(packer, frame, clu11, button):
   return packer.make_can_msg("CLU11", 0, values)
 
 
+def create_cancel_command(packer):
+  tcs13_values = {
+    "ACC_EQUIP": 1,
+    "PBRAKE_ACT": 1,
```
sshane, line 75: When setting all the bits in byte 6, it brought the cancellation time down to ~10 frames and I narrowed it down to these two. If you set either one, it becomes more inconsistent and can take up to 70 frames to cancel

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -78,15 +81,24 @@ def update(self, c, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert,
...
 
     if not CS.CP.openpilotLongitudinalControl:
-      if pcm_cancel_cmd:
-        can_sends.append(create_clu11(self.packer, frame, CS.clu11, Buttons.CANCEL))
+      if pcm_cancel_cmd and not self.prev_pcm_cancel:
+        self.cancel_frames = 0
+      if pcm_cancel_cmd and frame % 10 == 0:
+        # cancel by causing a temporary fault in the SCC module, causes a chime
+        # TODO: find a way to make this silent
+        can_sends.extend([create_cancel_command(self.packer)] * 15)
```
sshane, line 89: Then spamming the message reduces our maximum time to cancel. We *usually* cancel in 10 frames without this, but on occasion it jumps up to 30-40 frames. With this, I haven't seen it above 15 on a short test drive

sshane:
Some notes: sending CGW1->CF_Gway_DrvDrSw for the driver door bit only required 1 to 2 messages to cancel the car on a 2020 Sonata, however we were only sending it at 10 Hz, so the fact that on occasion it required 2 messages meant the door open bit could have taken up to 19 openpilot frames/190 ms to cancel. With faulting the SCC module using TCS13, we cancel on average in 10 frames/100 ms which feels as instant as the door open bit.

sshane:
Closing as https://github.com/commaai/panda/commit/5a7af82f06bf5f8039df8f58f9b1114f7d3436ee was merged into master for testing

PR #23745, opened 2022-02-11
Hyundai longitudinal: driver override state when gas is pressed

### LINKED ISSUE #22439: hyundai longitudinal: send correct ACC state in pre-enable state
Need to send the gas override state instead of regular acc mode

sshane:
The radar seems to send `ACCMode` as 2 while the gas is pressed, and for a few frames after: https://user-images.githubusercontent.com/25857203/153542865-7465e43c-33c9-4f56-9cca-f6711ff90e67.png

Another plot from stock ACC: https://user-images.githubusercontent.com/25857203/153551097-7a2a5980-9c63-4382-b205-32dd4cc80b7d.png

To resolve https://github.com/commaai/openpilot/issues/22439


selfdrive/car/hyundai/hyundaican.py
```diff
@@ -95,10 +95,10 @@ def create_acc_commands(packer, enabled, accel, jerk, idx, lead_visible, set_spe
   commands.append(packer.make_can_msg("SCC11", 0, scc11_values))
 
   scc12_values = {
-    "ACCMode": 1 if enabled else 0,
+    "ACCMode": 2 if enabled and gas_pressed else 1 if enabled else 0,
```
sshane, line 98: If this does correspond to gas pressed, should we include the buffer after gas is released?

gregjhogan:
You also need to do this for SCC14 ACCMode to match stock.  It probably tells the ESC to suspend commanding acceleration without disabling.

sshane:
Sounds good, I'll add that as well. However, that comment (`stock will always be 4 instead of 0 after first disengage`) doesn't seem to be true. Here's a route where it switches between 4 and 0 for disabled: https://user-images.githubusercontent.com/25857203/153555378-4f490a4b-0d5f-4a75-83ec-fc3a603b42bb.png Is this a concern or is 4 a superset of 0 behavior? As in, the car would fault if we always sent 0 for `SCC14`, but not 4?

gregjhogan:
SCC12 ACCMode values:
0 - off
1 - enabled
2 - driver override
3 - off (maybe main off or disabled via some options menu?)

SCC14 ACCMode values:
0 - off
1 - enabled
2 - driver override
3 - off (maybe main off or disabled via some options menu?)
4 - cancelled

sshane:
Tested and works well on our Sonata! No changes in behavior, however a user reports it fixed braking while pressing the gas. Good to merge @gregjhogan

gregjhogan:
LGTM

PR #23747, opened 2022-02-11
controlsd: Fix Toyota steer angle sensor race condition

nuwandavek:
Currently, if the STEER_TORQUE_SENSOR message is read, and the STEER_ANGLE_SENSOR message is not read, it initializes the offset difference to a wrong value, and it takes a really long time to correct (due to the small decay in the first order filter).
image


selfdrive/car/toyota/carstate.py
```diff
@@ -64,7 +64,7 @@ def update(self, cp, cp_cam):
 
     if self.accurate_steer_angle_seen:
       # Offset seems to be invalid for large steering angles
-      if abs(ret.steeringAngleDeg) < 90:
+      if abs(ret.steeringAngleDeg) < 90 and cp.can_valid:
```
sshane, line 67: Maybe we can move this to the check on line 62? `if abs(torque_sensor_angle_deg) > 1e-3:` The next check relies on the update call

pd0wm: This is fine, right? The next lines already check for initialized.

sshane: Yeah, this is more clear now. It would just be a very small optimization

PR #23748, opened 2022-02-11
Hardcode mass in KG

vanillagorillaa:
Removed conversions for some vehicle mass that were in pounds and hardcoded to kg, plus added the "STD_CARGO_KG" to a few vehicles that didn't have it.


sshane:
Rebasing and updating refs again

adeebshihadeh:
@sshane make sure that test_processes.py passes with a tolerance like 1e-3

selfdrive/car/hyundai/interface.py
```diff
@@ -63,7 +63,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.25], [0.05]]
     elif candidate == CAR.SONATA_LF:
       ret.lateralTuning.pid.kf = 0.00005
-      ret.mass = 4497. * CV.LB_TO_KG
+      ret.mass = 2040. + STD_CARGO_KG
```
sshane, line 66: Some of these might've been wrong initially actually. Google says the 2019 Sonata has a curb weight of 3,247 to 3,527 lbs. But is the LF a special varianrt?

adeebshihadeh:
Closing as this is too hard to review, and it's unclear what the goal is. We'll add the missing `STD_CARGO_KG` in a separate PR.

PR #23752, opened 2022-02-11
Longitudinal: add Kf to carParams

ClockeNessMnstr:
"1" isn't really a sane value here.
Edit:  Updated long to be tuneable as well.


pd0wm:
What is the purpose of this PR? Having no feedforward should be the default case.

ClockeNessMnstr:
> What is the purpose of this PR? Having no feedforward should be the default case.

Yes, thought I'd just throw it up. Lateral feed forward is much lower than one. It also appears that longitudinal tuning has had a feed forward of one with no value to tune. (I don't have long though)

pd0wm:
See: https://github.com/commaai/openpilot/pull/23740

ClockeNessMnstr:
I agree about having the input sanition somewhere else actually. This is different though. 

Also, not sure if people tuning long realize that there's a feedforward?

pd0wm:
Sorry, you're right! Let me reopen.

PR #23755, opened 2022-02-12
pandad: better handling of internal panda failures

adeebshihadeh:


selfdrive/pandad.py
```diff
@@ -34,6 +37,24 @@ def flash_panda(panda_serial: str) -> Panda:
...
+      panda.reset(enter_bootstub=True)
+      if not panda.bootstub:
+        cloudlog.error(f"Panda unable to enter bootstub. Attempting to recover.")
+
+        gpio_set(GPIO.STM_RST_N, 1)
+        gpio_set(GPIO.STM_BOOT0, 1)
+        time.sleep(2)
+        gpio_set(GPIO.STM_RST_N, 0)
+
+        panda.recover(skip_enter_dfu=True)
```
robbederks, line 51: Should be fine to release `BOOT0` a second or so after releasing reset, that way you don't have to reset again after the recovery.

Also, `STM_RST_N` would suggest that it would reset when set to 0, which isn't the case. Would rename that to `STM_RST` to avoid confusion :)

selfdrive/boardd/pandad.py
```diff
@@ -27,6 +28,7 @@ def flash_panda(panda_serial: str) -> Panda:
   panda = Panda(panda_serial)
 
   fw_signature = get_expected_signature(panda)
+  internal_panda = panda.is_internal() and not panda.bootstub
```
pd0wm, line 31: how do you recover an internal panda using gpio if it's already stuck in bootstub on startup?

adeebshihadeh: not handled in this PR since the bootstub doesn't implement the endpoint to query hw type. I'm thinking we can handle this nicely by adding internal panda types to the HW abstraction layer and wait until all internal pandas are up, but want to keep this PR simple

PR #23756, opened 2022-02-14
HKG: Add FW for 2019 Veloster

sunnyhaibin:
Add firmware for the 2019 Hyundai Veloster.

Route ID: `6112eef7212ce785|2022-02-13--21-40-10`

Thanks to community Veloster owner Xenorites.

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -651,6 +651,7 @@ FW_VERSIONS = {
     (Ecu.engine, 0x7e0, None): [
       b'\x01TJS-JNU06F200H0A',
       b'\x01TJS-JDK06F200H0A',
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Looks like we're also missing this
```
  (0x7d1, None): [
    b'\xf1\x816V8RAC00121.ELF\xf1\x00\x00\x00\x00\x00\x00\x00',
  ]
```

sunnyhaibin:
> Looks like we're also missing this value
> 
> ```
>   (0x7d1, None): [
>     b'\xf1\x816V8RAC00121.ELF\xf1\x00\x00\x00\x00\x00\x00\x00',
>   ]
> ```

Good catch! Added 🙂

PR #23760, opened 2022-02-14
New car docs

adeebshihadeh:
- [ ] Add dashcam exception

### DIFF
--- a/docs/CARS.md
+++ b/docs/CARS.md
@@ -1,185 +1,174 @@
 # Supported Cars
+Cars are sorted into three tiers:
 
-## comma.ai supported cars
+  Gold - a full openpilot experience
 
+  Silver - a pretty good, albeit limited experience
+
+  Bronze - signifincatly limited
+
+## Gold Cars
 | Make      | Model (US Market Reference)   | Supported Package | ACC              | No ACC accel below | No ALC below      |
 | ----------| ------------------------------| ------------------| -----------------| -------------------| ------------------|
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
> It might be interesting to utilize the "silver" designation as a "grey zone" where cars that would be bronze in stock form, could be promoted via modifications (eg. SDSU, Pedal, ZSS, EPS f/w hack, etc.)... or perhaps have a separate page entirely like `docs/CARS_modded.md` that incorporates a more flexible interpretation.

The plan right now is to categorize each car based on the "stock" install and perhaps add a column indicating that it can achieve a gold-level experience with modifications. I kind of like your second README idea though.

sshane:
- [ ] Add dashcam exception

vibhusapra (commented):
The No ALC below column seems to be have been copied wrong. For example, looking at the first two cars, the ILX should have no ALC below 25mph and the RDX below 12mph.

sshane:
@vibhusapra That's because these values aren't in the CarInterface. Feel free to submit PRs for missing metadata like `CP.minSteerSpeed` and `CP.minEnableSpeed`

geohot:
I propose doing this sorting manually. I posted a "score sheet" on discord.

All cars start with 3 points. You lose one for:
* low torque (honda accord, nidec civic)
* no radar, likely inaccessible without reflash (vw, honda bosch)
* non mainline software support (chrysler)

3 = gold, 2 = silver, 0-1 = bronze

Toyota Corolla = Gold
Honda Civic (nidec) = Silver
Honda Accord = Bronze

PR #23761, opened 2022-02-14
Add unicode brand name after docs change

GregorKikelj:


adeebshihadeh:
We can't change these strings just yet since everything relies on the strings to identify the platforms. We're going to make the car platforms proper enums. Once that's done, you can reopen this to fix the brand names.

PR #23763, opened 2022-02-15
add thneed optimizer to bigmodel

geohot:
Not ready yet. Requires hyperthneed (private repo) to be checked out to run to get patched CL kernels, will probably need to just ship the kernel diffs, or could rewrite them. This also adds a little (very cheap) benchmark to the thneed compile script.

On the comma three, it goes from 442 kernels -> 253 kernels and 21.88 ms -> 19.25 ms. (13% speedup)

Will test on comma two. Also, from my testing last time, the comma three ".thneed" file ran faster on the comma two since it uses a different first kernel, which is actually faster on both devices. Will test this too.

### DIFF
--- a/release/files_common
+++ b/release/files_common
@@ -427,7 +427,9 @@ selfdrive/modeld/transforms/transform.cl
 selfdrive/modeld/thneed/thneed.*
 selfdrive/modeld/thneed/serialize.cc
 selfdrive/modeld/thneed/compile.cc
+selfdrive/modeld/thneed/optimizer.cc
 selfdrive/modeld/thneed/include/*
+selfdrive/modeld/thneed/kernels/*.cl
 
 selfdrive/modeld/runners/snpemodel.cc
 selfdrive/modeld/runners/snpemodel.h
--- a/selfdrive/modeld/SConscript
+++ b/selfdrive/modeld/SConscript
@@ -1,3 +1,5 @@
+import os
+
 Import('env', 'arch', 'cereal', 'messaging', 'common', 'gpucommon', 'visionipc')
 lenv = env.Clone()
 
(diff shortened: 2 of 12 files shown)


geohot:
On comma two it's 33.37 ms -> 30.89 ms. I'll try the comma three ".thneed" file also.

geohot:
Ahh, it does work with the comma three ".thneed" file if you don't compile with "--binary"!

```
    modelExecutionTime = 0.029528541,
    modelExecutionTime = 0.029566405,
    modelExecutionTime = 0.029415729,
    modelExecutionTime = 0.029673178,
    modelExecutionTime = 0.029825989,
    modelExecutionTime = 0.029519323,
    modelExecutionTime = 0.029588178,
    modelExecutionTime = 0.029472085,
```

Now we are talking! This is building the model on the 845 and running on the 820...a bit of a scam if you ask me. That's 5ms of savings!

geohot:
Never mind, it's not that one kernel. It's the local group sizes.

I think I can write something pretty generic for this, might even improve the C3.

geohot:
Even with that dead simple optimizer:
```
    modelExecutionTime = 0.028995939,
    modelExecutionTime = 0.029106146,
    modelExecutionTime = 0.029217761,
    modelExecutionTime = 0.029344948,
    modelExecutionTime = 0.028962813,
    modelExecutionTime = 0.029171614,
    modelExecutionTime = 0.029064478,
```

geohot:
Final take
```
    modelExecutionTime = 0.02885651,
    modelExecutionTime = 0.028595781,
    modelExecutionTime = 0.028761301,
    modelExecutionTime = 0.028757552,
    modelExecutionTime = 0.028669061,
    modelExecutionTime = 0.028583072,
    modelExecutionTime = 0.028702917,
```

vs (without optimizer)
```
    modelExecutionTime = 0.034263074,
    modelExecutionTime = 0.034599949,
    modelExecutionTime = 0.034288749,
    modelExecutionTime = 0.034150574,
    modelExecutionTime = 0.034248073,
    modelExecutionTime = 0.034374688,
    modelExecutionTime = 0.034223542,
```
for a savings of 5.5ms. I'm going to put the kernels in this repo, I've pretty much rewritten them at this point, so I'm not sure the patching style makes sense. Aside from what is needed to be compatible with SNPE, there's nothing creative in them, just a barebones implementation of a publicly well-known algorithm.

geohot:
On C3 we save about 3ms.

Before:
```
    modelExecutionTime = 0.02254045,
    modelExecutionTime = 0.022594668,
    modelExecutionTime = 0.023056382,
    modelExecutionTime = 0.022510814,
```

After:
```
    modelExecutionTime = 0.019777769,
    modelExecutionTime = 0.021569887,
    modelExecutionTime = 0.02000235,
    modelExecutionTime = 0.020021932,
```

geohot:
Overall: this PR only affects things at compile time. Sadly the local workgroup optimizer is non deterministic since it uses benchmarks, so if we are trying for reproducible builds, we have to rethink this. But otherwise it's ready to go.

adeebshihadeh:
Here's an example of explicitly setting a dependency: https://github.com/commaai/opendbc/blob/46a942d6790531cf5b94b14266140e43afcfda3e/can/SConscript#L24

geohot:
Hmm, CI is failing. Looking into it.

PR #23766, opened 2022-02-15
Hyundai: move Optima Hybrid to dashcam only

sshane:
No routes in the last year, and it's not in CARS.md

### DIFF
--- a/selfdrive/car/hyundai/interface.py
+++ b/selfdrive/car/hyundai/interface.py
@@ -29,6 +29,7 @@ class CarInterface(CarInterfaceBase):
     ret.openpilotLongitudinalControl = Params().get_bool("DisableRadar") and (candidate not in LEGACY_SAFETY_MODE_CAR)
 
     ret.pcmCruise = not ret.openpilotLongitudinalControl
+    ret.dashcamOnly = candidate == CAR.KIA_OPTIMA_H
 
     ret.steerActuatorDelay = 0.1  # Default delay
     ret.steerRateCost = 0.5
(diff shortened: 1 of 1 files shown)


selfdrive/car/hyundai/interface.py
```diff
@@ -175,7 +176,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.lateralTuning.indi.timeConstantV = [1.4]
       ret.lateralTuning.indi.actuatorEffectivenessBP = [0.]
       ret.lateralTuning.indi.actuatorEffectivenessV = [1.8]
-    elif candidate in (CAR.KIA_OPTIMA, CAR.KIA_OPTIMA_H):
+    elif candidate == CAR.KIA_OPTIMA:
```
adeebshihadeh, line 179: I would still keep it here

selfdrive/car/hyundai/interface.py
```diff
@@ -29,6 +29,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
     ret.openpilotLongitudinalControl = Params().get_bool("DisableRadar") and (candidate not in LEGACY_SAFETY_MODE_CAR)
 
     ret.pcmCruise = not ret.openpilotLongitudinalControl
+    ret.dashcamOnly = candidate == CAR.KIA_OPTIMA_H
```
adeebshihadeh, line 32: add same comment as GM

sshane, line 32:
```suggestion
    # These cars have been put into dashcam only due to both a lack of users and test coverage.
    # These cars likely still work fine. Once a user confirms each car works and a test route is
    # added to selfdrive/test/test_routes, we can remove it from this list.
    ret.dashcamOnly = candidate in [CAR.KIA_OPTIMA_H, CAR.ELANTRA_GT_I30]
```

selfdrive/car/hyundai/interface.py
sshane, line 179:
```suggestion
    elif candidate in (CAR.KIA_OPTIMA, CAR.KIA_OPTIMA_H):
```

sshane:
Woops, the branch is named dashcam so I can't push to it anymore.

PR #23767, opened 2022-02-15
Hyundai: move unused cars to dashcam only

sshane:
Both of these cars are missing test routes and had no routes in the last year

### DIFF
--- a/selfdrive/car/hyundai/interface.py
+++ b/selfdrive/car/hyundai/interface.py
@@ -29,6 +29,7 @@ class CarInterface(CarInterfaceBase):
     ret.openpilotLongitudinalControl = Params().get_bool("DisableRadar") and (candidate not in LEGACY_SAFETY_MODE_CAR)
 
     ret.pcmCruise = not ret.openpilotLongitudinalControl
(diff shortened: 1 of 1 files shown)


selfdrive/car/hyundai/interface.py
```diff
@@ -30,6 +30,11 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
 
     ret.pcmCruise = not ret.openpilotLongitudinalControl
 
+    # These cars have been put into dashcam only due to both a lack of users and test coverage.
+    # These cars likely still work fine. Once a user confirms each car works and a test route is
+    # added to selfdrive/test/test_routes, we can remove it from this list.
+    ret.dashcamOnly = candidate in [CAR.KIA_OPTIMA_H, CAR.ELANTRA_GT_I30]
```
adeebshihadeh, line 36:
```suggestion
    ret.dashcamOnly = candidate in {CAR.KIA_OPTIMA_H, CAR.ELANTRA_GT_I30}
```

PR #23771, opened 2022-02-15
omx encoder: move writing into thread

pd0wm:

### DIFF
--- a/selfdrive/loggerd/omx_encoder.cc
+++ b/selfdrive/loggerd/omx_encoder.cc
@@ -155,7 +155,6 @@ static const char* omx_color_fomat_name(uint32_t format) {
 
 
 // ***** encoder functions *****
-
 OmxEncoder::OmxEncoder(const char* filename, int width, int height, int fps, int bitrate, bool h265, bool downscale, bool write) {
   this->filename = filename;
   this->write = write;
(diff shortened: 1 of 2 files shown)


geohot:
Something like this is absolutely needed, and hopefully this fixes the broken segments.

My thought was to put a msgq here, but if you are confident about the thread safety of this, it's certainly simpler.

CI is failing because the "try_pop" with the continue is spinning. Hopefully there's a blocking API, or worst case add a sleep before continue.

I'm also curious about the max size of these queues. I never want the encoder thread to block.

geohot:
Yea, looks like the "try_pop" can just be a "pop"

pd0wm:
Weird that spinning on the SafeQueue::try_pop at 1000 Hz used so much CPU. I assumed that would be basically free.

Let me write something to move the `OMX_BUFFERHEADERTYPE` into a new struct that has the data at the end and we can safely copy around between processes.

pd0wm:
```
        The FillBufferDone method is used to return filled buffers from an
        output port back to the application for emptying and then reuse.  
        This is a blocking call so the application should not attempt to 
        empty the buffers during this call, but should queue the buffers 
        and empty them in another thread.  There is no error return, so 
        the application shall handle any errors generated internally.  The 
        application shall also update the buffer header to indicate the
        number of bytes placed into the buffer.  

        The application should return from this call within 5 msec.
```

Added two threads now, one to handle the copy and one for the actual writing. I can also put the copy inside of the callback. However, since OMX seems to be so sensitive I decided to follow the documentation for now and not empty the buffers in the callback.

geohot:
I'm not sure why you need a callback handler thread. I would just leave that in the main loop. My concern was over the write in the main loop, not the copy, and I worry more about a thread.

pd0wm:
Thought about it a bit more, and putting in the main loop would always introduce a one frame delay in freeing up an output buffer. Since we only have 6 that's quite significant.

Will merge like this if the testing closet results look good.

PR #23772, opened 2022-02-15
add thneed optimizer

geohot:
See #23763, I rebased this on master for merge. I suspect model replay may have been failing because it wasn't updated for bigmodel.


geohot:
Ahh, replay still fails, but I think it's close enough now. These new values should be more accurate (deterministically so) since there's no float16 rounding between conv and activation.

geohot:
Do we have a check to compare the model outputs to the computer? If not, we should.

I updated the ref by hand looking at the comparison and seeing the numbers very close.

adeebshihadeh:
I believe research runs a test like that manually before releases. For cases like this, I like to increase the tolerance a bit rather than checking by hand. https://github.com/commaai/openpilot/blob/7176f5c4013e09142bfac577952b4a524db77cfb/selfdrive/test/process_replay/model_replay.py#L145

DiqahLer (commented):
Reboot

PR #23762, opened 2022-02-14
Docs: auto-generate supported cars documentation

sshane:
To automatically generate docs like [CARS.md](https://github.com/commaai/openpilot/blob/master/docs/CARS.md)

- [x] Add car info for all models
- [x] Add exceptions
- [x] Finish rest of todos in cars.py

New naming conventions:
- Rav4 -> RAV4
- PHEV -> Plug-In Hybrid
- Plug-in -> Plug-In
- EV -> Electric

Notes for categorizing:
- All HKG that aren't in the [legacy safety mode list](https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/values.py#L1064) and have a radar DBC have op long
- Only Nidec Hondas have op long, as before (checked radarOffCan and added exception for Hyundai)

PR #23773, opened 2022-02-15
values: make CAR classes enums

sshane:
Takes out a chunk of the diff from #23762, and we should probably support this first


sshane:
Closing to re-think this

PR #23779, opened 2022-02-16
Snpemodel: Clean up after #23772

redacid95:
After #23772 it appears the time testing to get a baseline of speed for the thneed optimizer was left behind when merged into master branch. I am not sure if this was done with intent or not. Please feel free to close this merge request if this is meant to stay like this. I did try asking in discord first, but no one responded.


geohot:
Actually, if you want to remove the loop but leave the benchmark I'd merge that.

PR #23787, opened 2022-02-17
Longitudinal control: interpolate longitudinal plan

sshane:
So I tried just interpolating v_target, as well as a_target also, and v_target is the sole reason we get this oscillation. However, when also interpolating a_target, we also get rid of a bit of the steppiness in the output accel: https://imgur.com/a/woHWglC

More plots: https://imgur.com/a/W1JTORm


sshane:
Wait there's a bug, making sure we correctly interp

HaraldSchafer:
rcv_frames is a confusing name, how about iterations_since_plan or something like that?

HaraldSchafer:
Besides that LGTM , can you also post the plot here of the improvement?

selfdrive/controls/lib/longcontrol.py
```diff
@@ -53,21 +53,23 @@ def reset(self, v_pid):
...
+      v_target = interp(interp_t, T_IDXS[:CONTROL_N], speeds)
+      a_target = interp(interp_t, T_IDXS[:CONTROL_N], long_plan.accels)
 
-      v_target_upper = interp(CP.longitudinalActuatorDelayUpperBound, T_IDXS[:CONTROL_N], speeds)
-      a_target_upper = 2 * (v_target_upper - speeds[0])/CP.longitudinalActuatorDelayUpperBound - long_plan.accels[0]
+      v_target_lower = interp(CP.longitudinalActuatorDelayLowerBound + interp_t, T_IDXS[:CONTROL_N], speeds)
+      a_target_lower = 2 * (v_target_lower - v_target) / CP.longitudinalActuatorDelayLowerBound - a_target
+
+      v_target_upper = interp(CP.longitudinalActuatorDelayUpperBound + interp_t, T_IDXS[:CONTROL_N], speeds)
+      a_target_upper = 2 * (v_target_upper - v_target) / CP.longitudinalActuatorDelayUpperBound - a_target
```
sshane, line 69: Think this is correct, essentially moved everything we're comparing up in time, so we should still divide only by the original longitudinal actuator delay.

sshane:
@HaraldSchafer how does it look now?

selfdrive/car/interfaces.py
```diff
@@ -97,6 +97,7 @@ def get_std_params(candidate, fingerprint):
     ret.longitudinalTuning.kpV = [1.]
     ret.longitudinalTuning.kiBP = [0.]
     ret.longitudinalTuning.kiV = [1.]
+    # TODO estimate car specific lag, use .15s for now
```
zorrobyte, line 100: Rav4 TSS2 seems to be:
```
Double the ActuatorDelay lag to .3 so we look into the future just a bit
more, seems a bit more inline with realworld experience.```
https://github.com/krkeegan/openpilot/commit/7000a7147c7be3f5ded6921623a9fcd875c2d8fb#diff-c933b99dbed0471164bc0132e6c5dcb8d32ce3e32051cf1309d60c6474c1c332R242

sshane: Probably will want to make another PR for this, and back up with data 🙂

selfdrive/controls/lib/longcontrol.py
HaraldSchafer: Yeah that sounds right to me

HaraldSchafer:
LGtM

sshane:
Wanted to make sure the sometimes-high differences in output accel make sense, here's some plots: https://imgur.com/a/0Ff2Rm8

When the new interpolated desired speed is rising, the small differences in value between the non-interpped value and interpped value add up for integral, making the new output accel 0.0283 m/s/s higher here when in the first acceleration.

zorrobyte:
Works fine!
```4aa5d874f367adbc|2022-02-18--15-15-10```

PR #23788, opened 2022-02-17
Lateral control: interpolate actuator delay

sshane:
Interpolate the 20hz value we control on to 100hz. Not yet done, as we shouldn't get lat_plan data by the 0th index anymore

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -497,12 +497,12 @@ class Controls:
 
       # Steering PID loop and lateral MPC
       lat_active = self.active and not CS.steerWarning and not CS.steerError and CS.vEgo > self.CP.minSteerSpeed
-      lat_rcv_frames = self.sm.frame - self.sm.rcv_frame['longitudinalPlan']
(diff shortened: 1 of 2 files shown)


neokii:
It seems that 'longitudinalPlan' on line 499 should be 'lateralPlan'.

HaraldSchafer:
LGTM

sshane:
There's another bug, psi is an integral so we can't just use a future value. Subtracting the value at t_since_plan seems to work, but is this correct?

sshane:
New plots from process replay look quite a bit noisier now, but I think I'd doing everything correctly, and sometimes the new plan looks like it's interpolating correctly. Some kind of lateral MPC instability? https://imgur.com/a/zPDx6pG

sshane:
It seems the lateral planner's MPC outputs are much less accurate when compared to the actual future plan in `steerActuatorDelay` seconds. Closing for now

Longitudinal:
Screenshot from 2022-02-25 14-43-30

Lateral:
Screenshot from 2022-02-25 14-31-43

PR #23789, opened 2022-02-18
locationd, paramsd: Check conditions before updating kalman filters

nuwandavek:
When there's a delay in some messages (especially `cameraOdometry`), the yaw rate and gyro bias values are erroneous. leading to bad filter states downstream in `paramsd`. This PR checks that all messages are alive and valid before updating the filters.


selfdrive/locationd/paramsd.py
```diff
@@ -161,7 +161,7 @@ def main(sm=None, pm=None):
   while True:
     sm.update()
     for which in sorted(sm.updated.keys(), key=lambda x: sm.logMonoTime[x]):
-      if sm.updated[which]:
+      if sm.updated[which] and sm.all_alive_and_valid():
```
adeebshihadeh, line 164: do you want to throw away the valid stuff? you can pass `all_alive_and_valid` a list with a subset of messages if you want like this
```suggestion
      if sm.updated[which] and sm.all_alive_and_valid([which]):
```

selfdrive/locationd/locationd.cc
```diff
@@ -462,6 +461,7 @@ kj::ArrayPtr<capnp::byte> Localizer::get_message_bytes(MessageBuilder& msg_build
 {
   cereal::Event::Builder evt = msg_builder.initEvent();
   evt.setLogMonoTime(logMonoTime);
+  evt.setValid(inputsOK);
```
adeebshihadeh, line 464: let's deprecate `inpusOK`

selfdrive/locationd/locationd.cc
```diff
@@ -500,7 +500,7 @@ int Localizer::locationd_thread() {
   while (!do_exit) {
     sm.update();
     for (const char* service : service_list) {
-      if (sm.updated(service) && sm.valid(service)) {
+      if (sm.updated(service) && sm.valid(service) && sm.allAliveAndValid()){
```
adeebshihadeh, line 503: same here, you can check the individual service

selfdrive/locationd/paramsd.py
nuwandavek: I want to update the filter when all the required messages are alive and valid. The current issue is when the sensor messages are valid and alive, but the camera odometry messages take a while (~2 secs) during init. The filter then enters a bad state and recovers slowly.

adeebshihadeh: I would check this outside the loop in that case

selfdrive/locationd/test/test_locationd.py
```diff
@@ -127,43 +130,53 @@ def send_msg(self, msg):
...
+      msg.gpsLocationExternal.longitude = self.lon
+      msg.gpsLocationExternal.altitude = self.alt
+    elif name == 'cameraOdometry':
+      msg = messaging.new_message(name)
       msg.cameraOdometry.rot = [0.0, 0.0, 0.0]
       msg.cameraOdometry.rotStd = [0.0, 0.0, 0.0]
       msg.cameraOdometry.trans = [0.0, 0.0, 0.0]
       msg.cameraOdometry.transStd = [0.0, 0.0, 0.0]
-      self.send_msg(msg)
+    elif name == 'sensorEvents':
```
adeebshihadeh, line 150: I would do something like this at the top:
```
try:
  msg = messaging.new_message(name)
except capnp.lib.capnp.KjException:
  msg = messaging.new_message(name, 0)

if name == "gpsLocationExternal":
  set stuff...
```

```suggestion
    elif name == 'sensorEvents':
```

selfdrive/locationd/test/test_locationd.py
```diff
@@ -127,43 +130,53 @@ def send_msg(self, msg):
...
+      msg = messaging.new_message(name, 0)
+    elif name in ['carState', 'liveCalibration']:
+      msg = messaging.new_message(name)
+    msg.logMonoTime = t
+    return msg
+
+  def test_params_gps(self):
+    # first reset params
+    Params().put('LastGPSPosition', json.dumps({"latitude": 0.0, "longitude": 0.0, "altitude": 0.0}))
```
adeebshihadeh, line 159: you can just remove with Params().delete('LastGPSPosition')

selfdrive/locationd/test/test_locationd.py
```diff
@@ -127,43 +130,53 @@ def send_msg(self, msg):
...
+    self.lon = -70 + (random.random() * 10.0)
+    self.alt = 5 + (random.random() * 10.0)
+    self.fake_duration = 90  # secs
+    # get fake messages at the correct frequency, listed in services.py
+    fake_msgs = [
+      self.get_fake_msg(name, int((i + j / services[name][1]) * 1e9)) for i in range(self.fake_duration)
+      for name in self.LLD_MSGS
+      for j in range(int(services[name][1]))
+    ]
+    for fake_msg in sorted(fake_msgs, key=lambda x: x.logMonoTime):
```
adeebshihadeh, line 171: why sort this? this isn't guaranteed at runtime

selfdrive/locationd/test/test_locationd.py
```diff
@@ -8,6 +8,7 @@
 
 from cereal import log
 import cereal.messaging as messaging
+from cereal.services import services
```
adeebshihadeh, line 11: services isn't supposed to be used directly, it's just for building service_list

nuwandavek: I want the various message frequencies though. Where should I read that from?

selfdrive/locationd/test/test_locationd.py
nuwandavek: Even when I publish messages in incorrect (unsorted logMonoTime) order?

selfdrive/locationd/test/test_locationd.py
```diff
@@ -5,9 +5,11 @@
 import unittest
 import time
 from cffi import FFI
+import capnp
```
adeebshihadeh, line 8: this should go above the from, like this:
```suggestion
import capnp
from cffi import FFI
```

The import convention in openpilot is standard library and third party, then openpilot imports. within those, imports go first and froms after preferably in alphabetical order.

pd0wm:
Controlsd will throw a commIssue when the liveLocationKalman or liveParameters will become invalid. If you want to mark them as invalid you have to remove them from the generic check, and handle the failure of the individual packets explicitly when using then.

nuwandavek:
> Controlsd will throw a commIssue when the liveLocationKalman or liveParameters will become invalid. If you want to mark them as invalid you have to remove them from the generic check, and handle the failure of the individual packets explicitly when using then.

I see. I'm modifying `liveParameters` to only have `valid` as `False` when it's an actual commIssue. But the `liveLocationKalman` case is a commIssue, though. It means that the `cameraOdometry` (likeliest message to cause the issue) message is not alive/valid yet. We don't want to allow people to engage when in this state. I think we want to keep the functionality as is, no? 

Does this sound good?

pd0wm:
sensorEvents on C2 is event driven, and doesn't really have a fixed frequency. You probably want to check that never fails the alive check, since that can now cause a commIssue.

Issue #23666, opened 2022-01-30
CARLA: modeld crashes on startup


maykonpacheco:
### Describe the bug

As CARLA is open and I try to start the cruise by pressing key 1, an error message appears. openpilot Unavailable Interprocess communication problem 

How to reproduce the error.
Open a terminal and run the command
cd tools/sim ./start_carla.sh

Open another terminal and run the command
cd tools/sim ./start_openpilot_docker.sh

Press 1 a few times while focused on bridge.py 

![image](https://user-images.githubusercontent.com/38124066/151686758-2b7c3ea8-9398-4513-9d99-5f463595e8a5.png)
![image](https://user-images.githubusercontent.com/38124066/151686760-7b775450-2c4b-4b9a-a685-722ff3512002.png)

### OS Version

Ubuntu 20.04

### openpilot version or commit

a5844363280ec4bafba9eaf046813f16aba6d6ac

pd0wm:
Looks like modeld crashes.

```
Onnx using  ['CUDAExecutionProvider', 'CPUExecutionProvider']
ready to run onnx model ['input_imgs', 'desire', 'traffic_convention', 'initial_state'] [[1, 12, 128, 256], [1, 8], [1, 2], [1, 512]]
Traceback (most recent call last):
  File "/home/batman/openpilot/selfdrive/modeld/runners/onnx_runner.py", line 60, in <module>
    run_loop(ort_session)
  File "/home/batman/openpilot/selfdrive/modeld/runners/onnx_runner.py", line 34, in run_loop
    inputs.append(read(ts).reshape(shp))
  File "/home/batman/openpilot/selfdrive/modeld/runners/onnx_runner.py", line 17, in read
    assert(len(st) > 0)
AssertionError
```

maykonpacheco:
`
Onnx available providers:  ['CPUExecutionProvider']
Onnx selected provider:  ['CPUExecutionProvider']
Traceback (most recent call last):
  File "/home/maykonpacheco/Develop/openpilot/selfdrive/modeld/runners/onnx_runner.py", line 59, in <module>
    ort_session = ort.InferenceSession(sys.argv[1], options, providers=[provider])
IndexError: list index out of range
`

maykonpacheco:
@pd0wm I still can't figure it out =/

jackhong12:
I guess the reason is the bugs in libzmq. After installing the newest version of libzmq, modeld can work properly.

jackhong12:
[...]

PR #23792, opened 2022-02-18
sim: fix modeld crash #23666

jackhong12:
**Description** [#23666](https://github.com/commaai/openpilot/issues/23666)
modeld will crash on startup in docker. I guess the reason is caused by libzmq. After installing the newest version of libzmq, modeld can work properly.
- same issue: https://github.com/zeromq/libzmq/issues/3313

### DIFF
--- a/tools/sim/Dockerfile.sim
+++ b/tools/sim/Dockerfile.sim
@@ -15,6 +15,10 @@ RUN apt-get update && apt-get install -y --no-install-recommends\
     libx11-6 \
   && rm -rf /var/lib/apt/lists/*
 
+RUN git clone https://github.com/zeromq/libzmq.git /libzmq && \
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
Why do you think this is caused by the ZMQ version?

jackhong12:
I use Valgrind to run model, like `valgrind _modeld`. It showed modeld crashed in libzmq.so. When I run modeld individually, it prints the message `Assertion failed: ok (src/mailbox.cpp:99)` which is also mentioned in the [issue](https://github.com/zeromq/libzmq/issues/3313). So I guess it's caused by this problem. After reinstalling ZMQ, _modeld can work correctly.

adeebshihadeh:
Any idea why this is suddenly an issue?

PR #23794, opened 2022-02-18
camerad: use open_v4l_by_name_and_index on c2 also

geohot:


adeebshihadeh:
> I did this blind, but if we trust CI it should be good to merge. Though I don't see Jenkins there...

Jenkins is down for the next couple hours while the new testing closet goes up.

geohot:
Hmm, jenkins failed, but I don't see an error.

PR #23795, opened 2022-02-18
camerad: fast debayer on c2 cameras

geohot:
HDR debayering requires a 1-D kernel, but non HDR debayering doesn't


geohot:
If this works, it saves 3.5 ms, from 5 -> 1.5

PR #23798, opened 2022-02-19
LatControlAngle: only show saturation alert if understeering

sshane:
Sometimes our angle can actually be overshooting the desired angle due to internal car rate limits, so only show when we're undershooting https://user-images.githubusercontent.com/25857203/154778939-e00621c4-fde6-44b7-a047-f5cdd076955f.png


Verylukyguy:
>https://user-images.githubusercontent.com/25857203/154778939-e00621c4-fde6-44b7-a047-f5cdd076955f.png

This plot looks exactly like a curve near my house.  When the hard curve is to the right, the steering makes in-curve corrections and eventually alerts, and when coming from the other direction and the curve is to the left, it completes the curve without the alert.

pd0wm:
Not sure why overshooting is not an issue. Everything that will make you deviate from the desired path should alert.

PR #23802, opened 2022-02-19
Subaru: add FW for Impreza Limited UDM 2017

martinl:
Adding missing Subaru Impreza Limited 2017 Transmission ecu value for discord user Houston2222


adeebshihadeh:
Which route did this come from?

martinl:
Route: `88c4c8431a0f8d47|2022-02-19--12-15-21`

PR #23808, opened 2022-02-21
Toyota: add Leuxs NX fingerprint

sintman:


vanillagorillaa:
Be sure to include the dongleID for the device these firmware versions were taken from

sintman:
Hi, the dongle ID is 
1303439f4af30ab1

PR #23810, opened 2022-02-21
Toyota: fix learning bad offsets for some models

### LINKED ISSUE #23809: Toyota TSS 2.5: openpilot can learn bad offset if initialized in a turn
### Describe the bug

Something I noticed on my TSS 2.5 Camry Hybrid is that `STEER_TORQUE_SENSOR`'s angle signal is absolute and not relative, meaning it doesn't start at 0 and very closely follows the angle in `STEER_ANGLE_SENSOR` without any offset (unlike TSS2 where `STEER_TORQUE_SENSOR`'s angle is relative.

Meaning if openpilot goes onroad while the car is driving and we're in a turn, a bad offset can be learned. This isn't a problem with torque control, as the desired angle is also offset in this way, however, when testing the LTA message which uses desired angles, the issue is very apparent as the car tried to veer left multiple times after engagement.

https://imgur.com/a/s31NKlQ

comma three

f15e3c37c118e841|2022-02-20--16-53-06--0

### openpilot version

0.8.14

### Additional info

_No response_

sshane:
On most TSS2 cars, the angle measurement in `STEER_TORQUE_SENSOR` is zero when initializing, however for some TSS2.5 cars (at least the 2022 Camry Hybrid), the signal is non-zero until it's ready, so wait for the `INITIALIZING` bit to fall before reading from it.

An example of a bad initial offset: https://imgur.com/a/jCA71Dq

The desired angle was offset using the liveParameters angle offset, so this is why we generally never noticed this while driving, however there should have been some minor hugging for a bit: https://imgur.com/a/JymjExk

Checked 50,000 Toyota segments, the only time `INITIALIZING` is 1 is at the very start of the route (and angle is still zero) or at the very end when the driver is about to shut off their vehicle (but the way the logic is, we never stop reading from the new sensor once we start). Plots: https://imgur.com/a/dH6exmi

### DIFF
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -16,9 +16,9 @@ class CarState(CarStateBase):
     self.shifter_values = can_define.dv["GEAR_PACKET"]["GEAR"]
     self.eps_torque_scale = EPS_SCALE[CP.carFingerprint] / 100.
 
-    # On cars with cp.vl["STEER_TORQUE_SENSOR"]["STEER_ANGLE"]
-    # the signal is zeroed to where the steering angle is at start.
-    # Need to apply an offset as soon as the steering angle measurements are both received
+    # On most cars with cp.vl["STEER_TORQUE_SENSOR"]["STEER_ANGLE"]
+    # the signal is zeroed to where the steering angle starts.
+    # However, on some cars this signal is absolute and requires no offset
     self.accurate_steer_angle_seen = False
     self.prev_accurate_steer_angle = 0.0
     self.angle_offset = FirstOrderFilter(None, 60.0, DT_CTRL, initialized=False)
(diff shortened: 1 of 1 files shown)


selfdrive/car/toyota/carstate.py
```diff
@@ -58,8 +57,8 @@ def update(self, cp, cp_cam):
     ret.steeringAngleDeg = cp.vl["STEER_ANGLE_SENSOR"]["STEER_ANGLE"] + cp.vl["STEER_ANGLE_SENSOR"]["STEER_FRACTION"]
     torque_sensor_angle_deg = cp.vl["STEER_TORQUE_SENSOR"]["STEER_ANGLE"]
 
-    # Some newer models have a more accurate angle measurement in the TORQUE_SENSOR message. Use if non-zero
-    if abs(torque_sensor_angle_deg) > 1e-3:
+    # Wait until an update has been seen to avoid learning an incorrect offset
+    if not bool(cp.vl["STEER_TORQUE_SENSOR"]["INITIALIZING"]):
```
sshane, line 61: Not yet sure what the behavior on other models are with this bit, but TSS-P Prius and TSS2 RAV4 would work with without checking the angle signal for a non-zero value: https://imgur.com/a/WPJSmCP

sshane:
There definitely exists some cars where the initializing bit isn't set to 1, but we don't have the accurate steer angle sensor, so we need to continue checking both: https://imgur.com/a/ssavB7x

PR #23813, opened 2022-02-21
Car Port: Kia EV6

pd0wm:

### DIFF
--- a/launch_openpilot.sh
+++ b/launch_openpilot.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/bash
 
+export FINGERPRINT="KIA EV6 2022"
+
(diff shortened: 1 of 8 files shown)


geohot:
I don't see anything on bus 5. Shouldn't the radar be there?

shinkansan:
I think the radar bus (lcan) is also canfd. If the data speed of bus1 is set to 2000, the value is received.

geohot:
I propose having the OBD-II harness be the default way to use a red panda (replacing the comma power). This way, we don't need to make a different harness.

We should try fingerprinting on both bus 0 and 3 anyway, and then fingerprinting will work in this configuration.

PR #23823, opened 2022-02-22
Added (Ecu.eps) fingerprint for Hyundai Ioniq EV 2017

wallneradam:
I have a HYUNDAI IONIQ EV 2017 in Hungary. I think it is not sold in the USA. It is almost the same as "HYUNDAI IONIQ ELECTRIC LIMITED 2019", only the Ecu.eps was different. After adding this line, everything works very good.


vanillagorillaa:
Be sure to include the dongleID for the device this was taken from

wallneradam:
> Be sure to include the dongleID for the device this was taken from

Where I should include it? Is it OK here?
My dongleID: 434165a5ac343885

PR #23837, opened 2022-02-23
test_onroad: few more fixes

GregorKikelj:


adeebshihadeh:
Brought these changes into #23824 as they need to run on an internal branch to verify the tests still pass with these fixes

PR #23841, opened 2022-02-23
Subaru: remove Impreza CAN fingerprints

adeebshihadeh:
We've had two Ascent users so far report that their cars are mis-fingerprinting on 0.8.13. I have one of their dongle IDs, and there are no FW versions in the logs, which implies their comma power is unplugged. In that case, the CAN fingerprinting fallback identifies them as Imprezas.

@martinl should we remove any other CAN fingerprints or is this the only known overlap? Since we can't trust the CAN fingerprinting, I think we'll need to manually add the FW versions for the Impreza and Ascent.


martinl:
AFAIK, only Impreza and Ascent FPv1 are very similar and have caused identification issues in the past. I think we can also remove the remaining FPv1 fingerprints if you can get and add fw versions of current Subaru users still fingerprinting using FPv1 from logs. Would we also need to provide new FPv2 test routes for each model?

adeebshihadeh:
Nope, no new routes necessary.

RCoaster:
I'm having the same issue- this time the comma power is connected and there's FW versions in my logs.

adeebshihadeh:
> I'm having the same issue- this time the comma power is connected and there's FW versions in my logs.

Great, we'll get your FW versions upstreamed and remove the Impreza fingerprints from the 0.8.13 release so others don't hit the same issue.

PR #23842, opened 2022-02-23
Documented TSS2 LTA steering mode signals

zorrobyte:
Spent a bunch of time this last weekend with @sshane figuring out what these signals do.

LTA steering feels very nice and has a 95 degree steering limit stock, but I was able to get even higher angles with OP. Torque bottoms out at some point.

You can try it out here:
https://github.com/zorrobyte/openpilot/tree/zp-lta

image

image

I'm cleaning up my branch and may attempt a PR at some point. Getting away from torque control and the tuning that comes along with it is incredibly nice. We can also use a steerRateCost of 0.1, which makes OP very responsive. Differences between laneless and laneful are apparent with angle based controls!

Dashcam LTA route:
4aa5d874f367adbc|2022-02-20--22-59-18

“Stock” LTA OP
4aa5d874f367adbc|2022-02-21--18-22-18

"Broken" OP LTA route (would not steer):
4aa5d874f367adbc|2022-02-21--18-37-54

Dashcam with disablements (LTA OEM steer overrides)
4aa5d874f367adbc|2022-02-22--14-47-21

Dadcam mode TSS2.5 stock
f15e3c37c118e841|2022-02-18--13-58-56

### DIFF
--- a/selfdrive/car/toyota/toyotacan.py
+++ b/selfdrive/car/toyota/toyotacan.py
@@ -15,15 +15,22 @@ def create_lta_steer_command(packer, steer, steer_req, raw_cnt):
 
   values = {
     "COUNTER": raw_cnt + 128,
-    "SETME_X1": 1,
-    "SETME_X3": 3,
-    "PERCENTAGE": 100,
-    "SETME_X64": 0x64,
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
This form of documentation is better as a GitHub Discussion, rather than inline comments.

PR #23845, opened 2022-02-23
Use LTA steering mode for TSS2 cars (and possibly TSS2.5 security cars)

zorrobyte:
Toyota has a system called Lane Trace Assist on TSS2.0+ cars, which is their version of L2 Lane Keeping. As such, they have provided us LTA signals in which we can send angles for the car to steer to.

Angle based control prevents us from having to compute and tune PID/LQR/INDI controls, which not only reduces lag, but also enhances the lateral control on these cars. We also take advantage of the EPS motor tuning Toyota has provided for us.

This is a draft PR so that I may collect feedback and clean up the code, panda changes are also required to allow sending the LTA messages. @pd0wm was experimenting with the LTA messages previously.

I have been working with @sshane on this over the weekend and was inspired by branch:
https://github.com/commaai/openpilot/tree/toyota-lta

LTA controls are much smoother than LKA controls and also seemingly have a higher angle limit:
image

Did someone say 🌮 🔔 ?

image

Lastly, I do not believe that Toyota security cars have the security bit on LTA. If so, this will be the controls method used to upstream these cars (and for OP to work at all). I call dibs on the bounty ;)
https://discord.com/channels/469524606043160576/905950538816978974/945496092152303617
https://discord.com/channels/469524606043160576/905950538816978974/945787996278521926

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -48,13 +48,6 @@ class CarController():
     apply_steer = apply_toyota_steer_torque_limits(new_steer, self.last_steer, CS.out.steeringTorqueEps, CarControllerParams)
     self.steer_rate_limited = new_steer != apply_steer
 
-    # Cut steering while we're in a known fault state (2s)
-    if not active or CS.steer_state in (9, 25):
-      apply_steer = 0
-      apply_steer_req = 0
-    else:
-      apply_steer_req = 1
-
     # TODO: probably can delete this. CS.pcm_acc_status uses a different signal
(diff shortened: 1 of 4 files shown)


cydia2020:
is `ACC_CONTROL` key'd? If it's key'd, then op won't be able to control long on those cars. https://discord.com/channels/469524606043160576/905950538816978974/934025110950662165

adeebshihadeh:
Thanks for the testing and PR, but I'm going to close this as it isn't useful to leave a PR open that isn't going to be finished. This will require more work and testing, and we already have our own branch for LTA testing.

Issue #23588, opened 2022-01-20
Disengage on gas toggle


adeebshihadeh:
- [x] add toggle
- [x] update safety model
- [x] panda safety tests - https://github.com/commaai/panda/pull/884
- [x] ensure smooth transition after gas is released - https://github.com/commaai/openpilot/pull/23639
- [x] read back unsafe mode - https://github.com/commaai/openpilot/pull/23563
- [x] match behavior of other systems
  - [x] Hyundai - TODO: verify jerk limits
  - [x] Honda Nidec
  - [x] GM
  - [x] Toyota
- [x] alert that long is disabled while gas is pressed? - https://github.com/commaai/openpilot/pull/24030
- [x] update training guide

pd0wm:
if we decide to keep steering while the gas is pressed, I think we should also change our pre-enable state to also start steering immediately.

geohot:
I think we definitely want to keep steering while gas is pressed. Super Cruise and Autopilot do.

pd0wm:
Yeah, that's what I thought. But currently we don't steer when you engage with the gas pressed. Which could be confusing, since there are two situations where you can press the gas, and only one of them steers. Therefore, I suggest we remove that and also steer when you engage with the gas pressed.

adeebshihadeh:
> Yeah, that's what I thought. But currently we don't steer when you engage with the gas pressed. Which could be confusing, since there are two situations where you can press the gas, and only one of them steers. Therefore, I suggest we remove that and also steer when you engage with the gas pressed.

+1 regardless of this situation. `controlsState.active` vs `controlsState.enabled` is super confusing and has led to a bunch of bugs  in the car controllers.

pd0wm:
I'm not sure we can get rid of the state completely, since if you have disengage on gas enabled you still want to be able to engage with gas. 

But we should able to remove the active vs enabled thing, and just make something generic based gas pressed. The car controller should not need to be aware of the toggle status, only the state machine logic.

pd0wm:
Other TODOs:

- [ ] Ensure the PID controller doesn't wind up when the gas is pressed
- [ ] Ensure we send the correct state for cars with stock longitudinal. Otherwise they might try to maintain zero acceleration, instead of slow down when you press the gas lightly. (https://github.com/commaai/openpilot/issues/22439)

sshane:
[...]

PR #23849, opened 2022-02-24
contolsd: combine active and enabled

sshane:
Refactor of values that describe if openpilot is enabled or not, as `enabled` being all states except `preEnable` (from gas press) is confusing from the perspective of the car controllers. Now they explicitly check for gas pressed when dealing with longitudinal.

Also steers while in pre-enable state so whether openpilot will steer while gas is pressed isn't different based on a toggle

Resolves one of the pre-reqs for https://github.com/commaai/openpilot/issues/23588


adeebshihadeh:
I think we decided that preEnabled will stay as is if disengage on gas is enabled, right @pd0wm?

pd0wm:
@adeebshihadeh That is correct. Panda needs to block steering while gas is pressed if disengage on gas is enabled.

sshane:
This PR is big so I'll make sure no behavior changes here

sshane:
Moved to #23859

PR #23850, opened 2022-02-24
Move gas/brake pedal event logic to controlsd

sshane:
Makes it so we only have to read the disengage on gas param once, and the car interfaces probably don't need to know about the toggle


sshane:
I believe the diffs are just order based, the final alert is whatever is added first if the priority and start frame is the same

selfdrive/controls/controlsd.py
```diff
@@ -192,6 +193,10 @@ def update_events(self, CS):
       self.events.add(EventName.controlsInitializing)
       return
 
+    # Disable on rising edge of gas or brake. Also disable on brake when speed > 0
+    if (CS.gasPressed and self.active) or (CS.brakePressed and not CS.standstill):
```
adeebshihadeh, line 197: this isn't the same as before. now, we don't disengage on the rising edge of brake

selfdrive/controls/controlsd.py
```diff
@@ -192,6 +194,11 @@ def update_events(self, CS):
       self.events.add(EventName.controlsInitializing)
       return
 
+    # Disable on rising edge of gas or brake. Also disable on brake when speed > 0
+    if (CS.gasPressed and self.active) or (CS.brakePressed and (not self.brake_pressed_prev or not CS.standstill)):
+      self.events.add(EventName.pedalPressed)
```
adeebshihadeh, line 199: let's save the previous car state so we can do the same check. it's unclear that `CS.gasPressed and self.active` enforces `Disable on rising edge of gas`

selfdrive/controls/controlsd.py
```diff
@@ -733,6 +740,7 @@ def step(self):
     self.prof.checkpoint("Sent")
 
     self.update_button_timers(CS.buttonEvents)
+    self.CS_prev = CS
```
adeebshihadeh, line 743:
```suggestion
    self.CS_prev = CS.as_reader()
```

sshane: Already a reader, from CarInterface: `self.CS.out = ret.as_reader()`

sshane:
One of the specific routes with differences:

Screenshot from 2022-03-15 19-46-59

Looks okay to me. The active event is the first added event, so now we do the `pedalPressed` event check first and on rising edge of gas it's active for one frame.

PR #23851, opened 2022-02-24
Enable Process Replay to Work with Logs Stored in LFS

krkeegan:
**Description**
Forks are not able to use the process_replay test.  The instructions [here](https://github.com/commaai/openpilot/tree/master/selfdrive/test/process_replay) are incomplete and various configurations need to be changed.

**Changes**
- Add `lfs: true` to the checkout action so that lfs files are downloaded.
- Add the replay log files to `.gitattributes` so the logs get uploaded to lfs
- Exclude the log files from `.gitignore` so they show up as changes to be staged

**Remaining Issue**
I have no idea how I can modify `/.lfsconfig` so that it works for forks and for comma.  For us github users, deleting the file makes everything work.  Similarly, I can change the `url` and `pushurl` in the file to github urls and it works.  But I can't figure out a design that would work across both forks and comma.  Any ideas would be appreciated.  ~If comma has a means of setting the contents of `.lfsconfig` for comma employees as part of its installation tooling, that might help.  I think we can remove `.lfsconfig` from the repo and ignore it.~  Nope that doesn't work, I thought the lfs pointer might contain the url, but no it doesn't.

- Also, if we can get this to work, I can update the README with some additional instructions to help fork maintainers.

**Verification**
I was able to get things to work using my own replay logs on this [testing branch](https://github.com/krkeegan/openpilot/tree/test-lfs).

**Notes**
I am competent with git, but LFS is a whole new thing for me, so I may be making some bad assumptions here.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
 *.dlc filter=lfs diff=lfs merge=lfs -text
 *.onnx filter=lfs diff=lfs merge=lfs -text
+selfdrive/test/process_replay/fakedata|*.bz2 filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/selfdrive_tests.yaml
+++ b/.github/workflows/selfdrive_tests.yaml
@@ -277,6 +277,7 @@ jobs:
     - uses: actions/checkout@v2
       with:
         submodules: true
+        lfs: true
     - name: Cache dependencies
       id: dependency-cache
       uses: actions/cache@v2
--- a/selfdrive/test/.gitignore
+++ b/selfdrive/test/.gitignore
@@ -6,4 +6,5 @@ process_replay/model_diff.txt
 valgrind_logs.txt
 
 *.bz2
+!process_replay/fakedata|*.bz2
 *.hevc


krkeegan:
A friend thought that on github that lfs repositories could be publicly pushable.

I created a gitlab account, I can't push that way either, it says I am not allowed to push code to this project.  So that is a no.

adeebshihadeh:
GitHub LFS no longer works for our purposes, which is why we had to switch to GitLab with the special LFS config. Removing the LFS config will simply fallback to GitHub, which may not always work as we only push our new LFS objects to GitLab and it has a very limited amount of monthly bandwidth. 

We've spent quite a bit of time trying to configure LFS so it works properly for us, as well as forks. It seems there's still a problem when forks try to push commits with an LFS diff, such as rebasing after a new model is released. We're ok with requiring a special script for us to push new LFS objects if it means forks don't run into this stuff, but we've been unable to find a way to do this yet. Let us know if you figure something out.

For this specific purpose, I'd recommend trying to check in your process replay refs directly with git if you don't expect to change them too often.

PR #23857, opened 2022-02-25
interfaces: remove unnecessary brake pedal check

sshane:
We only add `pedalPressed` on rising edge of gas so that we can engage when gas is previously pressed (gasPressed and pedalPressed are both added for one frame, then only gasPressed), but we can always just check brake pressed.

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -156,8 +156,7 @@ class CarInterfaceBase(ABC):
       events.add(EventName.steerUnavailable)
 
     # Disable on rising edge of gas or brake. Also disable on brake when speed > 0.
-    if (cs_out.gasPressed and not self.CS.out.gasPressed) or \
-       (cs_out.brakePressed and (not self.CS.out.brakePressed or not cs_out.standstill)):
+    if (cs_out.gasPressed and not self.CS.out.gasPressed) or (cs_out.brakePressed and not cs_out.standstill):
       events.add(EventName.pedalPressed)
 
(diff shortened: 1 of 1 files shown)


pd0wm:
Not sure what you mean? This PR removes the rising edge check on brake, not on gas.

sshane:
Exactly, for gas press we almost never hit the NO_ENTRY case, and for brake press as well as the PCM disallows engagement (at least on Toyota, maybe not Hyundai). For brake, I think we want to add it all the time

sshane:
Yeah, that was my intention. But I forgot about the standstill case, so maybe we want to define this behavior more clearly, as we add for gas press on rising edge so we can engage while gas pressed (NO_ENTRY is almost never hit), and on brake pressed rising edge while stopped so we can engage behind a lead.

pd0wm:
Would it make sense to create a new event type for these situations? Where we only disengage on creation/rising edge of the event? Then we can keep this logic in one place in the state machine and keep the car code simpler.

@adeebshihadeh any thoughts on this?

adeebshihadeh:
> Would it make sense to create a new event type for these situations? Where we only disengage on creation/rising edge of the event? Then we can keep this logic in one place in the state machine and keep the car code simpler.

This is a bit more complicated than creating a new event type, and I'm not sure what problem it solves. I think you'd want this as a separate property from event type where you can specify level, rising edge, or falling edge triggered. All the events are nicely defined in `selfdrive/controls/lib/events.py` now, and I'm not sure how we'd add these properties to the events.

PR #23859, opened 2022-02-25
carControl: add long and lat active fields

sshane:
Since `preEnable` will have two different behavior sets based on disengage on gas, it might make sense to differentiate which actuators are actually active.


sshane:
Only did Honda for now, but how does this look @pd0wm?

selfdrive/controls/controlsd.py
```diff
@@ -478,25 +474,29 @@ def state_control(self, CS):
     long_plan = self.sm['longitudinalPlan']
 
     actuators = car.CarControl.Actuators.new_message()
+    # Check if actuators are enabled
+    actuators.latActive = self.enabled and not CS.gasPressed
```
sshane, line 478: Disengage on gas logic will be here to keep steering in preEnable/enable

pd0wm:
LGTM

We can probably also clean up some of the long planner conditions using `longActive`: https://github.com/commaai/openpilot/blob/master/selfdrive/controls/lib/longitudinal_planner.py#L70

pd0wm (approved):

selfdrive/controls/controlsd.py
```diff
@@ -643,7 +642,6 @@ def publish_logs(self, CS, start_time, actuators, lac_log):
     controlsState.longitudinalPlanMonoTime = self.sm.logMonoTime['longitudinalPlan']
     controlsState.lateralPlanMonoTime = self.sm.logMonoTime['lateralPlan']
     controlsState.enabled = self.enabled
-    controlsState.active = self.active
```
adeebshihadeh, line 646: no need to remove this

adeebshihadeh: the only diff here should be longActive and latActive

sshane: Should both need to be active for this? If so, when disengage on gas is false, controlsState can be partially active.

selfdrive/controls/controlsd.py
sshane: Previous active logic was to use the pre-enable state, which right now is only for gas press, so we can use state instead. But that's assuming pre-enable will always allow us to start steering if disengage on gas is false.

selfdrive/car/honda/carcontroller.py
```diff
@@ -152,7 +152,7 @@ def update(self, enabled, active, CS, frame, actuators, pcm_cancel_cmd,
     # steer torque is converted back to CAN reference (positive when steering right)
     apply_steer = int(interp(-actuators.steer * P.STEER_MAX, P.STEER_LOOKUP_BP, P.STEER_LOOKUP_V))
 
-    lkas_active = active and not CS.steer_not_allowed
+    lkas_active = actuators.latActive and not CS.steer_not_allowed
```
adeebshihadeh, line 155: let's clean this up at some point. `CS.steer_not_allowed` seems to be a subset of `carState.steerFaultTemporary`

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -54,7 +54,7 @@ def update(self, c, enabled, CS, frame, actuators, pcm_cancel_cmd, visual_alert,
     self.steer_rate_limited = new_steer != apply_steer
 
     # disable when temp fault is active, or below LKA minimum speed
-    lkas_active = c.active and not CS.out.steerFaultTemporary and CS.out.vEgo >= CS.CP.minSteerSpeed
+    lkas_active = actuators.latActive and not CS.out.steerFaultTemporary and CS.out.vEgo >= CS.CP.minSteerSpeed
```
adeebshihadeh, line 57: controlsd already does this

selfdrive/controls/controlsd.py
```diff
@@ -479,25 +479,29 @@ def state_control(self, CS):
     long_plan = self.sm['longitudinalPlan']
 
     actuators = car.CarControl.Actuators.new_message()
+    # Check which actuators are enabled
+    actuators.latActive = self.enabled and not CS.gasPressed
+    actuators.longActive = self.enabled and not CS.gasPressed
```
adeebshihadeh, line 484:
```suggestion
    actuators.latActive = self.active
    actuators.longActive = self.active
```

adeebshihadeh: these states should always be driven by events

selfdrive/controls/controlsd.py
```diff
@@ -479,25 +479,29 @@ def state_control(self, CS):
...
     if not self.joystick_mode:
       # accel PID loop
       pid_accel_limits = self.CI.get_pid_accel_limits(self.CP, CS.vEgo, self.v_cruise_kph * CV.KPH_TO_MS)
       t_since_plan = (self.sm.frame - self.sm.rcv_frame['longitudinalPlan']) * DT_CTRL
-      actuators.accel = self.LoC.update(self.active, CS, self.CP, long_plan, pid_accel_limits, t_since_plan)
+      actuators.accel = self.LoC.update(actuators.longActive, CS, self.CP, long_plan, pid_accel_limits, t_since_plan)
 
       # Steering PID loop and lateral MPC
-      lat_active = self.active and not CS.steerFaultTemporary and not CS.steerFaultPermanent and CS.vEgo > self.CP.minSteerSpeed
+      lat_active = actuators.latActive and not CS.steerFaultTemporary and not CS.steerFaultPermanent and CS.vEgo > self.CP.minSteerSpeed
```
adeebshihadeh, line 504: I would actually check this stuff to set `actuators.latActive` and `actuators.longActive`

adeebshihadeh (commented):
We should also write some tests to ensure the correct behavior from the car controllers. It should be much easier now that we have the actuators output.

selfdrive/car/honda/carcontroller.py
adeebshihadeh: done in #23920

selfdrive/car/toyota/interface.py
```diff
@@ -270,7 +270,7 @@ def update(self, c, can_strings):
   # to be called @ 100hz
   def apply(self, c):
     hud_control = c.hudControl
-    ret = self.CC.update(c.enabled, c.active, self.CS, self.frame,
+    ret = self.CC.update(c, c.enabled, self.CS, self.frame,
```
adeebshihadeh, line 273: remove c.enabled

selfdrive/car/volkswagen/carcontroller.py
```diff
@@ -39,7 +39,7 @@ def update(self, c, enabled, CS, frame, ext_bus, actuators, visual_alert, left_l
       # torque value. Do that anytime we happen to have 0 torque, or failing that,
       # when exceeding ~1/3 the 360 second timer.
 
-      if c.active and CS.out.vEgo > CS.CP.minSteerSpeed and not (CS.out.standstill or CS.out.steerFaultPermanent or CS.out.steerFaultTemporary):
+      if c.latActive and not CS.out.standstill:
```
adeebshihadeh, line 42: move the standstill check to controlsd?

selfdrive/controls/controlsd.py
```diff
@@ -540,7 +547,9 @@ def state_control(self, CS):
         cloudlog.error(f"actuators.{p} not finite {actuators.to_dict()}")
         setattr(actuators, p, 0.0)
 
-    return actuators, lac_log
+    CC.actuators = actuators
```
adeebshihadeh, line 550: do this the other way around instead of creating an actuators message manually, i.e. do this instead of making a new msg
```suggestion
    actuators = CC.actuators
```

selfdrive/car/gm/carcontroller.py
```diff
@@ -67,15 +67,19 @@ def update(self, c, CS, frame, actuators, hud_v_cruise, hud_show_lanes, hud_show
 
       idx = (frame // 4) % 4
 
-      at_full_stop = c.enabled and CS.out.standstill
-      near_stop = c.enabled and (CS.out.vEgo < P.NEAR_STOP_BRAKE_PHASE)
+      # For Openpilot, "enabled" includes pre-enable.
+      # In GM, PCM faults out if ACC command overlaps user gas.
+      acc_enabled = c.enabled and not CS.out.gasPressed
```
sshane, line 72: See if we can replace with `longActive`, copied from interface

selfdrive/car/volkswagen/carcontroller.py
sshane: In another PR I'll replace with the MIN_STEER_SPEED in latcontrol.py, then we can remove lac_log.active as it will be a duplicate of CC.latActive

selfdrive/car/gm/carcontroller.py
```diff
@@ -68,15 +67,19 @@ def update(self, c, enabled, CS, frame, actuators,
 
       idx = (frame // 4) % 4
 
-      at_full_stop = enabled and CS.out.standstill
-      near_stop = enabled and (CS.out.vEgo < P.NEAR_STOP_BRAKE_PHASE)
+      # For Openpilot, "enabled" includes pre-enable.
```
adeebshihadeh, line 70: openpilot is stylized as all lower case

sshane:
After filtering all the diffs with the new active fields + removing the old one, this is the final diff: https://pastebin.com/GTVrXpwX

The actuators is changed because we're not sending any steer at a standstill anymore, as expected:
Screenshot from 2022-03-11 00-11-10

PR #21147, opened 2021-06-05
2020 Ioniq PHEV Support

sshane:
Accompanying PRs: https://github.com/commaai/panda/pull/655, https://github.com/commaai/opendbc/pull/401

No behavior change, sets up support for Hyundai HEVs using the different gas pedal signal.

This new signal is needed and works on:
- 2020 Ioniq PHEV Limited
- 2021 Sonata Hybrid
- 2018 Kia Niro Hybrid

Note: Accel_Pedal_Pos is speed on the above cars

New signal is absent from:
- 2020 Ioniq Electric
- 2020 Kia Niro EV
- 2019 Kona Electric
- 2017 Kia Optima Hybrid (it's non-zero, but doesn't catch all gas presses)

PR #23862, opened 2022-02-25
Hyundai: Add FW for 2022 Ioniq Plug-in Hybrid 

sunnyhaibin:
Add firmware for the 2022 Ioniq Plug-in Hybrid. Expand supported model-year range. Remove `IONIQ_PHEV` from `STEER_MAX = 255` blacklist as the car is able to utilize the full range of `STEER_MAX = 384` without faulting.

Route ID: `fe37d3981ec3749b|2022-02-24--10-34-01`
Harness Type: [Hyundai H](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)

Thanks to community Ioniq Plug-in Hybrid owner cbmhighflier.


adeebshihadeh:
Any reason this is its own platform? Seems like it's in all the same lists as `IONIQ_PHEV`.

sunnyhaibin:
> Any reason this is its own platform instead of adding the FW versions to the existing one? Seems like it's in all the same lists as `IONIQ_PHEV`.

@adeebshihadeh The 2022 Hyundai Ioniq PHEV does not have the `STEER_MAX = 255` limit at the EPS, while `IONIQ_PHEV` is still under the `STEER_MAX` of 255 and unconfirmed if it can steer up to 384 at the moment. If `IONIQ_PHEV` can be confirmed to have a higher limit than `STEER_MAX = 255`, we can combine them both into the same platform in another PR 🙂

adeebshihadeh:
Looks like the existing one was added before the default was 384 in #21147, so let's just add these FW versions to the existing one and move it over to the 384 list.

PR #23863, opened 2022-02-25
HKG: Add FW for 2022 Kia Niro HEV

sunnyhaibin:
Add firmware for the 2022 Kia Niro HEV. Expand supported model-year range.

Route ID: `108ad5e92908c955|2022-02-25--16-35-27`

Thanks to community Niro HEV owner Aryeh95.


sshane:
Can you add a route for where it was taken from?

sunnyhaibin:
> Can you add a route for where it was taken from?

Just added! Was waiting for the route ID before marking this as ready 🙂

PR #23865, opened 2022-02-25
TICI default for PC

haraschax:


pd0wm:
If you change this you also need to fix the CARLA side

HaraldSchafer:
What needs to change in the CARLA side and where is that?

pd0wm:
That code is here: https://github.com/commaai/openpilot/blob/master/tools/sim/bridge.py#L64

We'd need to add a second camera and make sure the vof/resolution match the tici cams.

pd0wm:
This is the proper fix: https://github.com/commaai/openpilot/issues/22574

PR #23869, opened 2022-02-26
temp status yellow is warning

crwusiz:
temp status yellow ( ok -> warning )

temp status color yellow status not ok is warning


adeebshihadeh:
The language here was chosen intentionally, and we'll keep it as is.

PR #23871, opened 2022-02-26
Fix kF value for thermal PI controller

robbederks:
The thermal controller for C3 broke due to this PR: #23752 which set the default feedforward gain to 0


ClockeNessMnstr:
"interp" in feedforward is just "P"

fan_pwr_out = -int(controller.update(setpoint=75, measurement=max_cpu_temp))+30

controller = PIController(k_p=2, k_i=2e-3, neg_limit=-50, pos_limit=30, rate=(1 / DT_TRML))

PR #23873, opened 2022-02-27
Subaru: add FW for 2019 Subaru Impreza

seven121-ym:
I have an Impreza 2019 model in Japan.
It is almost the same as the already registered Impreza 2019.
However, since Ecu.esp and Ecu.eps are different, add them.

dongle id:7f6bb63c061b59cc

### DIFF
--- a/selfdrive/car/subaru/values.py
+++ b/selfdrive/car/subaru/values.py
@@ -89,7 +89,6 @@ FW_VERSIONS = {
       b'z\xc0\x04\x00',
       b'z\xc0\x00\x00',
       b'\x8a\xc0\x10\x00',
-      b'z\xc0\x08\x00',
     ],
     (Ecu.fwdCamera, 0x787, None): [
       b'\x00\x00\x64\xb5\x1f\x40\x20\x0e',
--- a/selfdrive/car/subaru/values.py
+++ b/selfdrive/car/subaru/values.py
@@ -80,6 +80,7 @@ FW_VERSIONS = {
       b'z\x94\f\x90\001',
       b'z\x9c\x19\x80\x01',
       b'z\x94\x08\x90\x00',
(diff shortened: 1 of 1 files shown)


selfdrive/car/subaru/values.py
```diff
@@ -88,6 +89,7 @@ class CAR:
       b'z\xc0\x04\x00',
       b'z\xc0\x00\x00',
       b'\x8a\xc0\x10\x00',
+      b'z\xc0\x08\x00',
```
adeebshihadeh, line 92:
```suggestion
```

seven121-ym: Why is it unnecessary to add a line for Ecu.eps?
"b'z \ xc0 \ x08 \ x00',"
It was present in my car log. Will Openpilot work without this?

adeebshihadeh: It was a duplicate. All of your car's FW versions are now in master.

seven121-ym: もし私が正しく理解できていなかったとしたら、ごめんなさい。
"b'z\xc0\x08\x00',"は"values.py"の中にまだ記載されていないと思う。
似ている"b'z\xc0\x04\x00',"はリストにありますが。

seven121-ym: If I didn't understand it correctly, I'm sorry.
I think "b'z \ xc0 \ x08 \ x00'," is not yet listed in "values.py".
A similar "b'z \ xc0 \ x04 \ x00'," is in the list.

adeebshihadeh: It's equivalent to `b'z\xc0\b\x00'`

seven121-ym: Thank you for answering many times.
That's because of the character code.

PR #23874, opened 2022-02-27
use P instead of interp for tici thermal controller

ClockeNessMnstr:
interp in FF should be replaced by "P" in the controller.


ClockeNessMnstr:
https://github.com/commaai/openpilot/pull/23871

ClockeNessMnstr:
FF would be appropriate for an ambient temperature reading.

robbederks:
Don't think this is true since the feedforward is based on the absolute temperature, while the P-factor works on the error (= setpoint - current_temp)

PR #23875, opened 2022-02-27
Move Feedback from feedforward path in Thermald. Add correct FF function. 

ClockeNessMnstr:
interp in FF should be replaced by "P" in the controller.

reopening https://github.com/commaai/openpilot/pull/23874 base branch was merged. 

@robbederks 

edit: FF updated for dynamic setpoint.

### DIFF
--- a/selfdrive/thermald/fan_controller.py
+++ b/selfdrive/thermald/fan_controller.py
@@ -87,6 +87,8 @@ class TiciFanController(BaseFanController):
     self.target = 75
 
   def update(self, max_cpu_temp: float, ignition: bool) -> int:
+    self.controller.neg_limit = -(80 if ignition else 30)
+    self.controller.pos_limit = -(30 if ignition else 0)
 
     if ignition != self.last_ignition:
       self.controller.reset()
@@ -96,7 +98,6 @@ class TiciFanController(BaseFanController):
                                         measurement=max_cpu_temp,
                                         feedforward=interp(self.target, [60, 100], [-80, 0])
                                         ))
-    fan_pwr_out = max(fan_pwr_out, 30) if ignition else min(fan_pwr_out, 30)
 
     self.last_ignition = ignition
     return fan_pwr_out
--- a/selfdrive/thermald/fan_controller.py
+++ b/selfdrive/thermald/fan_controller.py
@@ -83,20 +83,20 @@ class TiciFanController(BaseFanController):
     cloudlog.info("Setting up TICI fan handler")
 
     self.last_ignition = False
-    self.controller = PIController(k_p=0, k_i=2e-3, k_f=1, neg_limit=-80, pos_limit=0, rate=(1 / DT_TRML))
+    self.controller = PIController(k_p=2, k_i=2e-3, k_f=1, neg_limit=-80, pos_limit=0, rate=(1 / DT_TRML))
+    self.target = 75
 
   def update(self, max_cpu_temp: float, ignition: bool) -> int:
-    self.controller.neg_limit = -(80 if ignition else 30)
-    self.controller.pos_limit = -(30 if ignition else 0)
 
     if ignition != self.last_ignition:
       self.controller.reset()
 
     fan_pwr_out = -int(self.controller.update(
-                      setpoint=75,
-                      measurement=max_cpu_temp,
-                      feedforward=interp(max_cpu_temp, [60.0, 100.0], [0, -80])
-                    ))
+                                        setpoint=self.target,
+                                        measurement=max_cpu_temp,
+                                        feedforward=interp(self.target, [60, 100], [-80, 0])
+                                        ))
+    fan_pwr_out = max(fan_pwr_out, 30) if ignition else min(fan_pwr_out, 30)
 
     self.last_ignition = ignition
     return fan_pwr_out


ClockeNessMnstr:
from robbederks:
 "Don't think this is true since the feedforward is based on the absolute temperature, while the P-factor works on the error (= setpoint - current_temp)

This is correct. 
(max_cpu_temp, [60.0, 100.0], [0, -80])
the current code is implementing P=2 and an offset of 30 in the feedforward using a range/setpoint and the current measurement. 

modified code here should be equivalent.

robbederks:
Again, I don't think this is equivalent at all, and also wrong. The purpose of the feedforward on temperature is due to the fact that the fan speed should not go to 0 when the error becomes 0.

If you read the implementation of the PIController, you'll see :)

robbederks:
Regardless, this tune is probably also not very close to optimal, but that's hard to do since it's not easy to model this thermal behavior offline.

ClockeNessMnstr:
> 

The integrator ensures that the temperature error goes to zero. not the output. Please look at the change. it is equivalent.

robbederks:
While it might mathematically be equivalent, this would make the offset and kP you added be dependent on the setpoint for this controller. I also don't see any benefit from this change.

ClockeNessMnstr:
I updated the FF function so if the setpoint is updated the FF reflects that. 

Previously the measurement was in the feedforward instead of the target. 
Feedforward does not use the measurement.

selfdrive/thermald/thermald.py
```diff
@@ -134,17 +134,16 @@ def handle_fan_uno(controller, max_cpu_temp, fan_speed, ignition):
...
-  controller.neg_limit = -(80 if ignition else 30)
-  controller.pos_limit = -(30 if ignition else 0)
-
   if ignition != last_ignition:
     controller.reset()
 
-  fan_pwr_out = -int(controller.update(
-                     setpoint=75,
-                     measurement=max_cpu_temp,
-                     feedforward=interp(max_cpu_temp, [60.0, 100.0], [0, -80])
```
ClockeNessMnstr, line 146: This line is the problem. 
Specifically "max_cpu_temp" should not be in the feedforward path. This is feedback. 

bonus improvement is that using the target here is correct. however the target is hardcoded so previously I just shifted everything by 30 corresponding to the setpoint.

ClockeNessMnstr:
The feed forward should not have the measurement in it. 

Feed forward would be the set point. So just 30.

ClockeNessMnstr:
Updated with new thermal controller class.

selfdrive/thermald/fan_controller.py
```diff
@@ -83,20 +83,20 @@ def __init__(self) -> None:
...
-    self.controller.neg_limit = -(80 if ignition else 30)
-    self.controller.pos_limit = -(30 if ignition else 0)
 
     if ignition != self.last_ignition:
       self.controller.reset()
 
     fan_pwr_out = -int(self.controller.update(
-                      setpoint=75,
-                      measurement=max_cpu_temp,
-                      feedforward=interp(max_cpu_temp, [60.0, 100.0], [0, -80])
```
ClockeNessMnstr, line 98: This line is the problem. (L98)
Specifically "max_cpu_temp" should not be in the feedforward path. This is feedback.

added improvement is that we can use actual FF here.

selfdrive/thermald/fan_controller.py
```diff
@@ -83,20 +83,20 @@ def __init__(self) -> None:
     cloudlog.info("Setting up TICI fan handler")
 
     self.last_ignition = False
-    self.controller = PIController(k_p=0, k_i=2e-3, k_f=1, neg_limit=-80, pos_limit=0, rate=(1 / DT_TRML))
+    self.controller = PIController(k_p=2, k_i=2e-3, k_f=1, neg_limit=-80, pos_limit=0, rate=(1 / DT_TRML))
+    self.target = 75
 
   def update(self, max_cpu_temp: float, ignition: bool) -> int:
-    self.controller.neg_limit = -(80 if ignition else 30)
-    self.controller.pos_limit = -(30 if ignition else 0)
```
pd0wm, line 90: Why was this removed in favor of the clip after the PID?

ClockeNessMnstr: Just feel cleaner to not change the limits of the PID and handle the ignition logic separately. but not needed.

pd0wm: We do this in controls too. It ensures the integrators can't wind up past the limits. If you revert this change it should be good to merge.

ClockeNessMnstr: All set

selfdrive/thermald/fan_controller.py
```diff
@@ -93,10 +94,10 @@ def update(self, max_cpu_temp: float, ignition: bool) -> int:
       self.controller.reset()
 
     fan_pwr_out = -int(self.controller.update(
-                      setpoint=75,
-                      measurement=max_cpu_temp,
-                      feedforward=interp(max_cpu_temp, [60.0, 100.0], [0, -80])
-                    ))
+                                        setpoint=self.target,
+                                        measurement=max_cpu_temp,
+                                        feedforward=interp(self.target, [60, 100], [-80, 0])
```
pd0wm, line 99: This interp is wrong. To make it match the old controller the result would need to be 30.

We thought about it a bit more, and changing it to an actual PID doesn't make it easier to reason about. It would imply that it's possible to change the setpoint without retuning the values, which is probably not true. If you do that you would need to find the relationship between kp and ff that works at all setpoints.

Since we don't plan on doing that anytime soon we'll just leave the controller as is.

pd0wm, line 99: This interp is wrong. To make it match the old controller the result would need to be 30.

We thought about it a bit more, and changing it to an actual PID doesn't make it easier to reason about. It would imply that it's possible to change the setpoint without retuning the values, which is probably not true. If you do that you would need to find the relationship between kp and ff that works at all setpoints.

Since we don't plan on doing that anytime soon we'll just leave the controller as is.

ClockeNessMnstr: I need to push back on the thought that it's not any easier to reason about. 
The current function in feedforward is a second P-only controller with a setpoint of 60. it provides no FF information.

it is actually helpful to separate the controller behavior and highlight both of the current misunderstandings in this comment: 
> The purpose of the feedforward on temperature is due to the fact that the fan speed should not go to 0 when the error becomes 0. In fact, it's used to describe a decent estimate of the steady-state relationship between temp and fan speed.
> 
> If you read the implementation of the PIController, you'll see :)

This is both not the purpose of feedforward, and the function here is inverse of the relationship that we want to use in FF. 

it would be better to use the PID class correctly so that when reasoning about FF we are actually reasoning about FF. 
I set "feedforward=30" on a new commit. This matches the current behavior. 

Changing the setpoint would be different between the current and new code however the FF increasing with increased setpoint temp would be incorrect and is what would occur currently. 
it would be beneficial to model FF in the form interp(self.target, [ssTemp(fan_max), ssTemp(fan_min)], [-fan_max, fan_min]) 
However, it is not beneficial to do this unless/untill an adaptive setpoint would be desired. but as you pointed out a value of 30 is the current implementation.

ClockeNessMnstr: >It would imply that it's possible to change the setpoint without retuning the values, which is probably not true. 

this is actually true. The controller and FF tuning could be improved but the setpoint does not effect the FF function or the p-gain.

ClockeNessMnstr:
Cleaned up the branch: https://github.com/ClockeNessMnstr/openpilot/commit/bdd8f9cf10ef63ef672f45d37dfe6c89d7df5c1b to only 2 value changes. 

Strictly speaking the pid class / feedforward should not be used as is. 
less strictly speaking, I do think this clarifies the behavior of the fan control scheme.

ClockeNessMnstr:
I'm cleaning up my branches and deleting this one but the implementation is belo 2 lines:     

self.controller = PIController(k_p=**2**, k_i=2e-3, k_f=1, neg_limit=-80, pos_limit=0, rate=(1 / DT_TRML))
                      ......
                      feedforward=**-30**

PR #23885, opened 2022-03-01
juggle.py: default to yes for qlog use

sshane:
It's a bit more convenient to default to yes.


adeebshihadeh:
This is intentional, as it's pretty confusing if you're unexpectedly missing a bunch of messages.

PR #23892, opened 2022-03-02
Toyota: Support Avalon Hybrid 2022 (TSS2.5)

sshane:
Dongle id: eca1080a91720a54

Screenshot from 2022-03-01 23-43-24

Also seems to have no offset for STEER_TORQUE_SENSOR->STEER_ANGLE (just like Camry Hybrid TSS2.5) despite the route starting at 9 degrees:
Screenshot from 2022-03-01 23-45-36


sshane:
Waiting for a test route

PR #23902, opened 2022-03-03
Joystick inactive handoff to controller for injection + Angle mode / axes swap flags

ClockeNessMnstr:
changed the logic of the Joystick debug mode to fallback to the long/lat controller when joystick inputs are at 0

This would allow for the injection test joystick behavior pinned in tuning. ~"hands resting on lap like using OP"

I also added and angle mode to allow the joystick to request a steering angle from the controller instead of overriding the torque. 

last I added a swap axes option to switch the gas/steer. It seems the ps5 controller defaults to lateral being up/down on one of the sticks.


ClockeNessMnstr:
Haven't tested yet so it's a draft until I can test it on mine and dbzx6r.

adeebshihadeh:
These should all be split into separate PRs. Instead of swapping the axes, I'd rather just support the popular controllers directly and add a warning if the controller plugged in doesn't match one of the supported ones.

ClockeNessMnstr:
Sounds good,

PR #23906, opened 2022-03-04
Remove KIA_NIRO_HEV_2021 from STEER_MAX = 255 Blacklist

sunnyhaibin:
Remove `KIA_NIRO_HEV_2021` from `STEER_MAX = 255` blacklist. User reported that the 2022 Kia Niro HEV is able to utilize the full range of `STEER_MAX = 384` without faulting.

Route ID: `108ad5e92908c955|2022-03-03--18-21-16`

Thanks to community Niro HEV owner Aryeh95.


Aryeh95:
Thank you @sunnyhaibin for pointing this out.
I was wondering why the available torque was so much worse than my volt.
After this little change, I didn't get a single fault from the car or a take control warning from OP in the same road where I previously got the take control warning multiple times.

PR #23910, opened 2022-03-04
Hide set speed when main is off

sshane:
Applies to all PCM cars, need to take a look at how the button cars do this.

The process replay difference looks like it's because the main button was turned off in the Toyota route and we now reset to 255 kph

### DIFF
--- a/selfdrive/ui/qt/onroad.cc
+++ b/selfdrive/ui/qt/onroad.cc
@@ -181,7 +181,7 @@ void OnroadHud::updateState(const UIState &s) {
   const auto cs = sm["controlsState"].getControlsState();
 
   float maxspeed = cs.getVCruise();
-  bool cruise_set = maxspeed > 0 && (int)maxspeed != SET_SPEED_NA && sm["controlsState"].getCarState().getCruiseState().getAvailable();
+  bool cruise_set = maxspeed > 0 && (int)maxspeed != SET_SPEED_NA && sm["carState"].getCarState().getCruiseState().getAvailable();
   if (cruise_set && !s.scene.is_metric) {
     maxspeed *= KM_TO_MILE;
   }
--- a/selfdrive/ui/qt/onroad.cc
+++ b/selfdrive/ui/qt/onroad.cc
@@ -181,7 +181,7 @@ void OnroadHud::updateState(const UIState &s) {
   const auto cs = sm["controlsState"].getControlsState();
 
   float maxspeed = cs.getVCruise();
-  bool cruise_set = maxspeed > 0 && (int)maxspeed != SET_SPEED_NA;
+  bool cruise_set = maxspeed > 0 && (int)maxspeed != SET_SPEED_NA && sm["controlsState"].getCarState().getCruiseState().getAvailable();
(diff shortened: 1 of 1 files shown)


sshane:
The process replay difference looks like it's because the main button was turned off in the route and we now reset to 255 kph

jyoung8607:
There is some overlap / alignment opportunity here with `hud_control.speedVisible`.

I use it in my mqb-long development branch to remove the set speed display from the instrument cluster when openpilot isn't engaged. Right now I'm only doing that because openpilot doesn't actually support Resume as-such, so displaying the old setpoint is more of a bug than a feature, and I'd actually like to make openpilot support Resume in the future.

For immediate purposes, it seems cleaner to use `hud_control.speedVisible` at the car port level than comparing to a magic speed value.

sshane:
It looks like the Volkswagen uses button presses to manage ACC state, which supports resume on Hyundai. do you have a branch to look at?

jyoung8607:
> It looks like the Volkswagen interface uses button presses to manage ACC state, and openpilot supports resume on Hyundai. How does openpilot not support it for Volkswagen? Is `accelCruise` != `resumeCruise`?

We have the physical resume button, and read its state, but AFAIK for openpilot longitudinal there's no corresponding controlsd code to resume the prior setpoint. There are only events for `buttonEnable` and `buttonCancel`.

When I started working on openpilot longitudinal for VW, I thought for sure I was missing something, but across the entire openpilot code base there are five references to `resumeCruise`. Four of them are in the Volkswagen port and one in the Tesla port. GM, Honda, and Hyundai enable on the falling edge of `accelCruise` or `decelCruise`.

> I think we still want to show the speed when openpilot is disengaged so you know what speed resume will set you to.

I agree, provided such a capability existed, but at this time it doesn't appear to. I'm in favor of fixing that. If nobody has done so within the next few weeks, I'll take a look at it as part of the VW work.

sshane:
Might've been hard to find, this doesn't work? https://github.com/commaai/openpilot/blob/master/selfdrive/controls/lib/drive_helpers.py#L86

jyoung8607:
> Might've been hard to find, this doesn't work? https://github.com/commaai/openpilot/blob/master/selfdrive/controls/lib/drive_helpers.py#L86

Hmm, so that's why it doesn't work... it keys off `accelCruise`, and my dev/test code [enables on falling edge of `setCruise` / `resumeCruise`](https://github.com/jyoung8607/openpilot/blob/d6590aa6763a8d6bedbe1e24f6a90177892ee362/selfdrive/car/volkswagen/interface.py#L228-L234). On this particular car, those buttons aren't overloaded with `decelCruise` and `accelCruise`.

I'm still not a fan of this implementation. Correct me if I'm wrong, but even if I "fix" this in the VW port (which would produce other wrong behavior) it'll just change the behavior from always-set to always-resume. I have two buttons for a reason, and changing their behavior from stock violates [POLA](https://en.wikipedia.org/wiki/Principle_of_least_astonishment) for no good reason.

adeebshihadeh:
Fixed in controlsd in #24052

PR #23911, opened 2022-03-06
boardd.cc: change fault list size (new exti fault for panda deep sleep mode)

briskspirit:
needed for panda deepsleep mode

merge after cereal and panda bump (or with this PR)

### DIFF
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -338,7 +338,7 @@ void Panda::set_power_saving(bool power_saving) {
   usb_write(0xe7, power_saving, 0);
 }
 
-void Panda::set_deepsleep(void) {
+void Panda::enable_deepsleep(void) {
   usb_write(0xe7, 0, 0);
 }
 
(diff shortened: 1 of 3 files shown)


selfdrive/boardd/panda.cc
```diff
@@ -338,6 +338,10 @@ void Panda::set_power_saving(bool power_saving) {
   usb_write(0xe7, power_saving, 0);
 }
 
+void Panda::set_deepsleep(void) {
```
adeebshihadeh, line 341: I'd call this enable_deepsleep. set sounds like it takes an arg to enable/disable

selfdrive/boardd/panda.h
```diff
@@ -85,6 +85,7 @@ class Panda {
   std::optional<std::vector<uint8_t>> get_firmware_version();
   std::optional<std::string> get_serial();
   void set_power_saving(bool power_saving);
+  void enable_deepsleep(void);
```
adeebshihadeh, line 88: no need for void

PR #23912, opened 2022-03-06
path fixes

geohot:
You shouldn't need "/system/comma" for QT. Also, there's a bug in env_lib_paths if you have 2

### DIFF
--- a/selfdrive/modeld/SConscript
+++ b/selfdrive/modeld/SConscript
@@ -71,7 +71,8 @@ if use_thneed and arch in ("aarch64", "larch64"):
 
   lib_paths = ':'.join(Dir(p).abspath for p in lenv["LIBPATH"])
   kernel_path = os.path.join(Dir('.').abspath, "thneed", "kernels")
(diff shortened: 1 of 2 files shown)


selfdrive/modeld/SConscript
```diff
@@ -71,7 +71,8 @@ if use_thneed and arch in ("aarch64", "larch64"):
 
   lib_paths = ':'.join(Dir(p).abspath for p in lenv["LIBPATH"])
   kernel_path = os.path.join(Dir('.').abspath, "thneed", "kernels")
-  cenv = Environment(ENV={'LD_LIBRARY_PATH': f"{lib_paths}:{lenv['ENV']['LD_LIBRARY_PATH']}", 'KERNEL_PATH': kernel_path})
+  env_lib_paths = ':'.join(lenv['ENV']['LD_LIBRARY_PATH'])
```
geohot, line 74: Actually this bug is probably elsewhere.

PR #23917, opened 2022-03-07
update tools/LogReader doc

xx979xx:
use MultiLogIterator to read the entire route, get timestamps and deal with exceptions


pd0wm:
Maybe add as a separate example?

pd0wm:
@xx979xx If you're seeing exceptions on `wich()` it probably means you cereal repo is older than the driving data. Don't think that's common enough to mention in the examples.

PR #23919, opened 2022-03-07
Add --dbc argument to plotjuggler

robbederks:
It would be nice if you could also parse in a DBC file which isn't in opendbc, but this requires more changes since they are currently built at plotjuggler build time, and not loaded dynamically.

### DIFF
--- a/tools/plotjuggler/juggle.py
+++ b/tools/plotjuggler/juggle.py
@@ -138,7 +138,7 @@ if __name__ == "__main__":
   parser.add_argument("--stream", action="store_true", help="Start PlotJuggler in streaming mode")
   parser.add_argument("--layout", nargs='?', help="Run PlotJuggler with a pre-defined layout")
   parser.add_argument("--install", action="store_true", help="Install or update PlotJuggler + plugins")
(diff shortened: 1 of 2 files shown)


sshane:
The dynamic parsing is done in https://github.com/commaai/openpilot/pull/20750, I'll check it out this week-next and try to get it merged

tools/plotjuggler/juggle.py
```diff
@@ -138,6 +138,7 @@ def juggle_route(route_or_segment_name, segment_count, qlog, can, layout):
   parser.add_argument("--stream", action="store_true", help="Start PlotJuggler in streaming mode")
   parser.add_argument("--layout", nargs='?', help="Run PlotJuggler with a pre-defined layout")
   parser.add_argument("--install", action="store_true", help="Install or update PlotJuggler + plugins")
+  parser.add_argument("--dbc", help="Set the DBC name to load for parsing CAN data")
```
adeebshihadeh, line 141:
```suggestion
  parser.add_argument("--dbc", help="Set the DBC name to load for parsing CAN data. If not set, the DBC will be automatically inferred from the logs.")
```

PR #23929, opened 2022-03-08
WIP: Support for Pixel 3

geohot:
With minor modifications, this all should work on Pixel 3 XL and OnePlus 6T also

Use these instructions set up the userspace:
https://gist.github.com/geohot/569e9e4b20fd41203d8da71c6022be15

Two major issues right now:
* The built binutils segfaults, so it can't build the panda firmware
* <s>SNPE doesn't run. If this isn't fixed by the community, it will have to wait for steelix (SNPE replacement) to be finished</s>
  * was caused by broken libCB on pixel 3, fixed using version from OP6T

Two minor issues from move to Android 9 from 6:
* visionimg.cc no longer works, minor API change required
* eglfs_surfaceflinger needs to be modified. This branch includes a binary, but I think it should just be moved into openpilot.
  * no immediate action required here though

And then:
* camerad must be modified to support Pixel 3
* A bunch of minor fixups for hardware abstraction. It's half a TICI half an EON

Join our discord to discuss:
https://discord.comma.ai/


incognitojam:
I managed to setup the userspace, compile openpilot and launch the UI on a Pixel 3 XL. Using system image `crosshatch-pq3a.190801.002`.

The UI is centered in the screen on the training steps and has black bars, I think because the resolution is set too small for the 3 XL.
black bars around training guide on pixel 3 xl

incognitojam:
I also replicated this on the OnePlus 6T (6GB RAM + 128GB ROM). It's running OxygenOS 9.0.17 (Build: A6013_41_190911).

In order to get touch events to work, you must change `QT_QPA_EVDEV_TOUCHSCREEN_PARAMETERS` in `selfdrive/ui/ui` to `/dev/input/event3`. The touch input is slightly offset, similar to the Pixel 3 XL, too. I assume because of the different screen resolution.


OnePlus 6T next to a Google Pixel 3 XL, both running openpilot

geohot:
To troubleshoot SNPE, you can try compiling the model by hand.

```
cd /data/openpilot/selfdrive/modeld/thneed
LD_LIBRARY_PATH=/data/openpilot/third_party/snpe/aarch64:$LD_LIBRARY_PATH ./compile /data/openpilot/models/supercombo.dlc /data/openpilot/models/supercombo.thneed
```

On the Pixel 3, I get

```
error_code=809; error_message=OpenCL function has returned error. error_code=809; error_message=OpenCL function has returned error. OpenCL Error (-61) CL_INVALID_BUFFER_SIZE; error_component=GPU Runtime; line_no=892; thread_id=535608423664; opencl_error=-61; error_component=GPU Runtime; line_no=271; thread_id=535779222856
```

geohot:
To troubleshoot panda build, you can run arm-none-eabi-as

```
blueline:/data/data/com.termux/files/home # arm-none-eabi-as 
Assembler messages:
Internal error (Segmentation fault).
Please report this bug.
```

If you fix the segfault, panda should build. I believe gcc is correct and just binutils has the issue. It's being built by install.sh

geohot:
The camera issues can be dealt in one of two ways.
* You can add raw support to `camera_qcom2.cc`, this will involve understanding the sensor in the phone and a bit of refactoring. It would be phone specific, but would get high performance with low lag.
* You could also write `camera_android.cc` which uses Android's camera stack, similar to `camera_webcam.cc` This would be more generic, but could possibly have performance issues and issues running multiple cameras at the same time.

nanamiwang:
> * The built binutils segfaults, so it can't build the panda firmware
The root cause of binutils segfault is fopen64 function not found by clang due to undefined `__ANDROID_API__ `macro required by stdio.h.
image
This issue can be fixed by prepending this line to binutils-2.32/bfd/bfdio.c (Adding this to Makefile is better?):
`#define __ANDROID_API__ 28`
Verified that panda/baord can be built successfully on pixel 3 device after this fix applied.
image

geohot:
Awesome fix for binutils!

@adeebshihadeh has been moving the install.sh stuff to termux-packages to build it on the PC. Can we apply this fix to binutils?

0312birdzhang:
In theory, can we use other sdm845 mobiles with Adreno 630 GPU?

0312birdzhang:
launched it on oneplus 6t

image

adeebshihadeh:
I think we can close this in its current form. An AGNOS port seems like the way forward for this kind of thing.

danielkucera:
Hi guys, I am trying to replicate this on Android 12 but can't figure out where this `libqeglfs-surfaceflinger-integration.so` file comes from. Is it something you wrote? Can you share the source?

With the current binary I am stuck on error:
```
: Failed to load EGL device integration "eglfs_surfaceflinger"
: EGL library doesn't support Emulator extensions
```

zorrobyte:
> Hi guys, I am trying to replicate this on Android 12 but can't figure out where this `libqeglfs-surfaceflinger-integration.so` file comes from. Is it something you wrote? Can you share the source?
> 
> * qeglfs_sf_main.cpp
> * qeglfs_sf_integration.cpp
> 
> With the current binary I am stuck on error:
> 
> ```
> : Failed to load EGL device integration "eglfs_surfaceflinger"
> : EGL library doesn't support Emulator extensions
> ```

I'd target Android 9

danielkucera:
> I'd target Android 9, there is also a group of users who have it working in current form over here https://discord.gg/PaSgTgUS8S if you'd like to collaborate

Okay, I'll target Android 9 then. So where is the source of that library?

reekotubbs:
@adeebshihadeh is there anyway to get a copy of the pixel3 branch that was deleted?

danielkucera:
> @adeebshihadeh is there anyway to get a copy of the pixel3 branch that was deleted?

The branch has been deleted but not commits :wink: 
image

PR #23931, opened 2022-03-09
test_models: check openpilot can cancel cruise

sshane:
Checks we can cancel cruise in a timely manner

- [ ] Make sure check is reliable
- [ ] Check for accidental re-enables
- [ ] Move to new test


adeebshihadeh:
Test models only checks things that we can fully reproduce just from the CAN messages. This belongs in FindInteresting.

PR #23936, opened 2022-03-09
thermald: use named thermal zones

geohot:
This fixes the CPU having the wrong zones. It also changes the zones used by the GPU from the "lowf" to the "usr" ones.

### DIFF
--- a/selfdrive/thermald/thermald.py
+++ b/selfdrive/thermald/thermald.py
@@ -67,7 +67,6 @@ def read_tz(x):
   if isinstance(x, str):
     if tz_by_type is None:
       populate_tz_by_type()
-    print(x, tz_by_type[x])
     x = tz_by_type[x]
 
   try:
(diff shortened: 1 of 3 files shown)


geohot:
CI issues?

selfdrive/thermald/thermald.py
```diff
@@ -50,10 +50,25 @@
...
+    if not n.startswith("thermal_zone"):
+      continue
+    with open(os.path.join("/sys/devices/virtual/thermal", n, "type")) as f:
+      tz_by_type[f.read().strip()] = int(n.lstrip("thermal_zone"))
+
 def read_tz(x):
   if x is None:
     return 0
 
+  if isinstance(x, str):
```
adeebshihadeh, line 67: think it's cleaner to do this in `HARDWARE.get_thermal_config`, then it always returns a path you can read from

geohot: I disagree with the way it's currently structured now. The numbers are meaningless and should imo be handled at the last possible moment.

geohot: That's probably better

geohot:
Good to merge?

geohot:
Good to merge?

selfdrive/thermald/thermald.py
pd0wm: Ok, to merge like this. I can do the refactor when I'm back from holiday.

PR #1464, opened 2020-05-04
Add 2020 Honda CRV support

csouers:
Drive > https://my.comma.ai/cabana/?route=243d0daa5d208cdd%7C2020-05-04--13-14-09&exp=1620158889&sig=b2O1Uj7wq5fxDPB7Fi6m5iGkvTm2tznZQxXRjTa1IO0%3D&max=23&url=https%3A%2F%2Fchffrprivate-vzn.azureedge.net%2Fchffrprivate3%2Fv2%2F243d0daa5d208cdd%2F4645f7e8fce3c2c37994cbea4363a9c1_2020-05-04--13-14-09

Looks like normal CRV_5G to me.

PR #23941, opened 2022-03-09
Honda: add missing minimum steer speeds

sshane:
For autogenerating car docs: https://github.com/commaai/openpilot/pull/23762

- [ ] Do todos

CR-V:
- 2020 - Honda CR-V support and fw added: https://github.com/commaai/openpilot/pull/1464/files
- 2021 - no new eps fw added: https://github.com/commaai/openpilot/pull/20341/files

Honda Civic:

- Hatchback always has 12 mph min steer speed, but it fingerprints as sedan/coupe (`CIVIC_BOSCH`) and the fw was added for it in batches, so it's difficult to pin point the specific eps fw. Left it as a TODO.
  Diesel Bosch Civic's are always 12 mph according to the exception in CARS.md, so add simple exception for that
- To sum it up, Nidec Civic (2016-18) is 12 mph, Bosch Civic Sedan/Coupe (2019-20) is 2 mph, Bosch Civic Diesel Sedan/Coupe (2019-20) is 12 mph.

### DIFF
--- a/selfdrive/car/honda/interface.py
+++ b/selfdrive/car/honda/interface.py
@@ -90,6 +90,7 @@ class CarInterface(CarInterfaceBase):
       ret.centerToFront = CivicParams.CENTER_TO_FRONT
       ret.steerRatio = 15.38  # 10.93 is end-to-end spec
       ret.minSteerSpeed = 12. * CV.MPH_TO_MS
+
       if eps_modified:
         # stock request input values:     0x0000, 0x00DE, 0x014D, 0x01EF, 0x0290, 0x0377, 0x0454, 0x0610, 0x06EE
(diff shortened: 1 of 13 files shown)


selfdrive/car/honda/interface.py
```diff
@@ -218,6 +240,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.lateralParams.torqueBP, ret.lateralParams.torqueV = [[0, 1000], [0, 1000]]  # TODO: determine if there is a dead zone at the top end
       tire_stiffness_factor = 0.444
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.8], [0.24]]
+      ret.minSteerSpeed = 12. * CV.MPH_TO_MS
```
adeebshihadeh, line 243: if 12 is the default, no need to set explicitly here

PR #23942, opened 2022-03-09
Hyundai: add missing minimum enable speed

sshane:
For autogenerating car docs: https://github.com/commaai/openpilot/pull/23762

- [ ] Split Optima `minEnableSpeed` by car fw


selfdrive/car/hyundai/interface.py
```diff
@@ -90,6 +90,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[]):  # py
       ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0.], [0.]]
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.25], [0.05]]
       ret.minSteerSpeed = 32 * CV.MPH_TO_MS
```
sshane, line 92: The CARS.md lists this at 32 mph, but it was added 4 years ago and other other 32 mph cars here in interface are 32 mph in CARS.md, so it's possible it was incorrect

PR #23946, opened 2022-03-11
camera_qcom2 refactor

geohot:


geohot:
So as a first step, I think this is worth merging. I know it's a hard diff to read, but it changes no functionality or order of anything. Will be a lot easier to base the next refactor off of this.

geohot:
Much more readable

PR #23950, opened 2022-03-11
Update lat_control at CAN send rate. 

ClockeNessMnstr:
PR is to update the lateral controller at the same rate as the commands are sent. Technically speaking it improves stability but usually this is only a significant impact at  higher frequencies. 

However, I believe this would be important for something like autotune planned. 
With autotune needing to measure the error and adjust gains based on the last control signals and errors, it would seem important to only update the controller at the same rate as the commands are sent to ensure everything is as ideally synced as possible. 
it would also be important to s

draft because only subaru interface.py has a change for the cars put in and I'm not a fan of how the logging conditionals look in controlsd.py. hopefully there's a cleaner way to log at a different rate than the control loop or just repeat the last log.

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -44,6 +44,12 @@ class CarInterfaceBase(ABC):
     if CarController is not None:
       self.CC = CarController(self.cp.dbc_name, CP, self.VM)
 
+  def get_steer_step(self):
+    return 1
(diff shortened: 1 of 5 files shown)


adeebshihadeh:
It's unclear what this is supposed to improve, and it's not useful to keep this PR open in this state.

PR #23951, opened 2022-03-11
locationd: Commissue fix

nuwandavek:
- Related PR https://github.com/commaai/cereal/pull/254
- Only check if all messages are alive and valid at the beginning, before starting the filter. 
- Store instantaneous `sm.allAliveAndValid` in `inputsOK` as before.

### DIFF
--- a/selfdrive/locationd/locationd.cc
+++ b/selfdrive/locationd/locationd.cc
@@ -457,15 +457,16 @@ void Localizer::handle_msg(const cereal::Event::Reader& log) {
 }
 
 kj::ArrayPtr<capnp::byte> Localizer::get_message_bytes(MessageBuilder& msg_builder, uint64_t logMonoTime,
-  bool inputsOK, bool sensorsOK, bool gpsOK)
+  bool inputsOK, bool sensorsOK, bool gpsOK, bool msgValid)
 {
   cereal::Event::Builder evt = msg_builder.initEvent();
   evt.setLogMonoTime(logMonoTime);
-  evt.setValid(inputsOK);
+  evt.setValid(msgValid);
   cereal::LiveLocationKalman::Builder liveLoc = evt.initLiveLocationKalman();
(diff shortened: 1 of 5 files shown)


selfdrive/locationd/locationd.cc
```diff
@@ -497,27 +498,30 @@ int Localizer::locationd_thread() {
...
+    if (filterValid){
       for (const char* service : service_list) {
-        if (sm.updated(service)){
+        if (sm.updated(service) && sm.valid(service)){
           const cereal::Event::Reader log = sm[service];
           this->handle_msg(log);
         }
       }
     }
+    else{
```
adeebshihadeh, line 513:
```suggestion
    } else {
```

selfdrive/locationd/locationd.cc
```diff
@@ -497,27 +498,29 @@ int Localizer::locationd_thread() {
...
   uint64_t cnt = 0;
+  bool filterValid = false;
 
   while (!do_exit) {
     sm.update();
-    if (sm.allAliveAndValid()){
+    if (filterValid){
       for (const char* service : service_list) {
-        if (sm.updated(service)){
+        if (sm.updated(service) && sm.valid(service)){
```
adeebshihadeh, line 507:
```suggestion
        if (sm.updated(service) && sm.allAliveAndValid({service})){
```

selfdrive/locationd/locationd.cc
```diff
@@ -497,27 +498,29 @@ int Localizer::locationd_thread() {
   SubMaster sm(service_list, nullptr, { "gpsLocationExternal" });
 
   uint64_t cnt = 0;
+  bool filterValid = false;
```
adeebshihadeh, line 501: maybe call this filterInitialized instead? took me a second to figure out that's what it was

adeebshihadeh (approved):

PR #23953, opened 2022-03-12
tici: 12 bit camera readout only

geohot:
Before #23887 is ready, this makes the camera use 12 bit mode instead of 10 bit mode. Required for IMX390 bringup. Copied relevant parts from #23887

Tested on desk to work with streamer, not sure what else I should do.

### DIFF
--- a/selfdrive/camerad/cameras/camera_qcom2.cc
+++ b/selfdrive/camerad/cameras/camera_qcom2.cc
@@ -26,7 +26,7 @@ extern ExitHandler do_exit;
 
 const size_t FRAME_WIDTH = 1928;
 const size_t FRAME_HEIGHT = 1208;
-const size_t FRAME_STRIDE = 2416;  // for 10 bit output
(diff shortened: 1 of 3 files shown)


pd0wm:
- Is the output still limited to 10 bits? Otherwise, the scaling value should also change on `selfdrive/camerad/cameras/real_debayer.cl:52`.
 - Check that changing OP_WORD_CLK_DIV does not change the readout time past 50ms. Not sure what it's at now: https://github.com/commaai/thecalbi/pull/192/files#diff-5131e21485d33d4526d452fcc8d9c8c2b6587309fa4e788086a9fdda936f5e43R140

geohot:
I just divide it super early, I don't think this affects AE.

pd0wm:
Looks like changing `OP_WORD_CLK_DIV` does not change the readout timing, so this should be fine. It's currently at 49.9933875 ms, which is why we might be seeing artefacts at long exposure times.

pd0wm (approved):

PR #23966, opened 2022-03-14
camerad: imx390 support

geohot:
Not ready, but will run the cameras.

- [x] Support both types of camera
- [x] Fix WUXGA mode
- [ ] Enable IMX390 HDR
- [x] Make code clean


geohot:
I think we should merge this. While IMX390 support isn't totally there yet, it's a good baseline.

PR #23969, opened 2022-03-15
Support RAV4 Hybrid 2022 with stock longitudinal

sshane:
For certain newer RAV4 vehicles, the radar outputs in a format different from most other Toyota, so use stock longitudinal until full radar parsing support is added. We also need to knock out the radar as it sends the ACC messages

RAV4 fw in PR is from dongle: 825c2be1679633af|2022-04-06--20-24-59

- [x] Safety to block ACC messages that don't contain the cancel signal - https://github.com/commaai/openpilot/pull/23969
- [x] Test models route


alancyau:
Here are a few more 2022 RAV4s with the new fws. I don't recall if they're gas or hybrid models, but I suppose it doesn't matter too much since they're on stock long.

https://github.com/alancyau/openpilot/commit/bedeb1bcd4c176f5b9cd57c107fb7c50a9fae05c
https://github.com/alancyau/openpilot/commit/0c5407475ff5643e471258a2e1c8f100f3db65c6
https://github.com/alancyau/openpilot/commit/926b2b7249524b747506942720ba8ea6239b043a

sshane:
@alancyau definitely need to know if they're hybrids or not so we use the correct gas signal. If you have dongle IDs per each fw version, that'd be a great help

aolin480:
Would this include 2020 RAV4 Hybrid XLR too ? I have one, so if you need anything tested just let me know. I tried testing the branch earlier today but I had an error saying vehicle ID couldn't be found or something like that. Just let me know. Thanks Shane!

sshane:
Your car should already be supported @aolin480, is it not?

aolin480:
Ah okay. Yeah that's what I was wondering. If this is for the newer RAV4 then it's all good, just didn't know if it unlocks something with RAV4's in general. Thanks for clarifying!

sshane:
Yes this is the PR to add support, the website will be updated soon

PR #23971, opened 2022-03-15
Updated Copyright Year in License

GovindKrishnan:
Replaced 2018 with 2022 in the MIT License.


GovindKrishnan:
I believe not. If you look at [Microsoft VS Code](https://github.com/microsoft/vscode) MIT License for instance, it says "2015 - Present".

PR #23974, opened 2022-03-15
Apply EU guideline lateral limits

haraschax:


GregorKikelj:
Would it make sense to consider longitudinal acceleration as well when limiting jerk? Being in a constant curvature turn and accelerating causes an increase in lateral acceleration aka jerk which might then exceed the 5m/s^3 limit.

PR #23982, opened 2022-03-17
Toyota: update Prius V minimum enable speed

sshane:


sshane:
Wait, the Prius V is in the `NO_STOP_TIMER_CAR` list which means "no resume button press required". But a user reported on Discord that it doesn't do full sng.

sshane:
This car can do full sng if you remove the DSU, but it's unclear if we want to change this.

PR #23983, opened 2022-03-17
CARS.md tweaks

sshane:
Footnotes was removed, need it to link to the bottom. Also CX-5 supposedly has good torque. Any other?


adeebshihadeh:
I think you can set it for all Mazda. Only the CX9 and CX5 are actually supported, and I think they both have good torque.

PR #23986, opened 2022-03-17
Update carla script to easier install docker and nvidia-docker

gijskoning:
I couldn't figure out how to use the old behaviour with Install=1

With this change typing "y" will install nvidia-docker2 with its dependencies nvidia-container-toolkit and docker-ce (previously didn't install docker-ce).

Verified branches manually

### DIFF
--- a/tools/sim/start_carla.sh
+++ b/tools/sim/start_carla.sh
@@ -2,8 +2,6 @@
 #echo "Your name is $1"
 # Requires nvidia docker - https://github.com/NVIDIA/nvidia-docker
 if ! $(apt list --installed | grep -q nvidia-container-toolkit); then
-#  if [ "$INSTALL" == "1" ]; then
-#  echo "Nvidia docker is required. Do you want to install it now? (y/n)"# Re-run with "start_carla.sh 1" INSTALL=1 to automatically install."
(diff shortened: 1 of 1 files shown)


tools/sim/start_carla.sh
```diff
@@ -2,16 +2,23 @@
...
     curl -s -L https://nvidia.github.io/nvidia-docker/$distribution/nvidia-docker.list | sudo tee /etc/apt/sources.list.d/nvidia-docker.list
-    sudo apt-get update && sudo apt-get install -y nvidia-container-toolkit
+    sudo apt-get update && sudo apt-get install -y nvidia-docker2 # Also installs docker-ce and nvidia-container-toolkit
+    read -p "Just installed docker? Set permanent docker permission for usergroup. (Run 'sudo systemctl restart docker' afterwards and then rerun start_carla.sh) (y/n)"
+    if [ "${REPLY}" == "y" ]; then
+        # Adding docker to current usergroup
+        sudo groupadd docker
+        sudo usermod -aG docker $USER
+        newgrp docker
+    fi
```
adeebshihadeh, line 18: this is very much up to the user, don't think we should include that in this script

PR #23989, opened 2022-03-17
Add wide camera to Carla

### LINKED ISSUE #23916: CARLA: add support for wide camera
Add flag to CARLA camera to choose between C2 and C3 camera outputs. C3 has one narrow and one wide camera.

gijskoning:
Fixes https://github.com/commaai/openpilot/issues/23916 to send narrow and wide camera images from Carla.

Sending the images seems to work with no errors in bridge.py. Verified images using watch3 and no errors coming from modeld.

Currently fov for wide_camera is set to 163. Higher fov shows unwanted artifacts in Carla. 
Also tried to make a more realistic fish eye lens with the Carla camera attributes without success. 
163 fov:
image
170 fov:
image


common/transformations/camera.py
```diff
@@ -6,8 +6,10 @@
 ## -- hardcoded hardware params --
 eon_f_focal_length = 910.0
 eon_d_focal_length = 650.0
+# tici_f: narrow road cam. tici_e: wide road cam
 tici_f_focal_length = 2648.0
-tici_e_focal_length = tici_d_focal_length = 567.0 # probably wrong? magnification is not consistent across frame
+tici_d_and_e_focal_length = 567.0 # probably wrong? magnification is not consistent across frame
+tici_e_focal_length = tici_d_focal_length = tici_d_and_e_focal_length
```
gijskoning, line 12: Unsure if we want to keep tici_e_focal_length since it isn't used anywhere.

adeebshihadeh: it's used in an internal repo

gijskoning:
Best to wait on the Carla CI pr https://github.com/commaai/openpilot/pull/24009

common/transformations/camera.py
```diff
@@ -6,8 +6,10 @@
 ## -- hardcoded hardware params --
 eon_f_focal_length = 910.0
 eon_d_focal_length = 650.0
+# tici_f: narrow road cam. tici_e: wide road cam
 tici_f_focal_length = 2648.0
-tici_e_focal_length = tici_d_focal_length = 567.0 # probably wrong? magnification is not consistent across frame
+tici_d_and_e_focal_length = 567.0 # probably wrong? magnification is not consistent across frame
```
pd0wm, line 11: Do we need this changes? What's wrong with using `tici_e_focal_length` in carla?

tools/sim/bridge.py
```diff
@@ -36,7 +38,7 @@
 PRINT_DECIMATION = 100
 STEER_RATIO = 15.
 
-pm = messaging.PubMaster(['roadCameraState', 'sensorEvents', 'can', "gpsLocationExternal"])
+pm = messaging.PubMaster(['roadCameraState', "wideRoadCameraState", 'sensorEvents', 'can', "gpsLocationExternal"])
```
pd0wm, line 41:
```suggestion
pm = messaging.PubMaster(['roadCameraState', 'wideRoadCameraState', 'sensorEvents', 'can', "gpsLocationExternal"])
```

common/transformations/camera.py
gijskoning: Not necessary, but thought this would be better for the reader to know that both are the same

tools/sim/bridge.py
```diff
@@ -64,12 +66,16 @@ def steer_rate_limit(old, new):
...
+    self.frame_road_id = 0
+    self.frame_wide_id = 0
     self.vipc_server = VisionIpcServer("camerad")
 
     # TODO: remove RGB buffers once the last RGB vipc subscriber is removed
     self.vipc_server.create_buffers(VisionStreamType.VISION_STREAM_RGB_ROAD, 4, True, W, H)
     self.vipc_server.create_buffers(VisionStreamType.VISION_STREAM_ROAD, 40, False, W, H)
+
+    self.vipc_server.create_buffers(VisionStreamType.VISION_STREAM_RGB_WIDE_ROAD, 4, True, W, H)
+    self.vipc_server.create_buffers(VisionStreamType.VISION_STREAM_WIDE_ROAD, 40, False, W, H)
```
pd0wm, line 78: Creating all these buffers probably allocates a ton op RAM on the GPU, causing issues on lower-end end hardware. Can you try allocating 5 for both fcam and ecam instead of 40? We use more in openpilot to ensure loggerd doesn't drop frames, but we don't care with the simulator.

gijskoning: Buffer of 5 seems to work.

PR #23990, opened 2022-03-17
FPv2: try panda bus if few results

incognitojam:
Currently it looks like we only send FW queries on the OBD-II bus, but I think we could try the Panda bus too if we get few results. It might also let users run OP without comma power attached.

Cherry-picked from #23331 as I didn't get a response from OBD-II on my Ford.

Ideas for discussion:
- Is 3 FW results good enough? Should we look for all ECUs?
- Do we need to remove duplicates between buses?


adeebshihadeh:
Thanks for the PR, but gonna close this one since this isn't quite how we want to do it. This is needed for the comma body and EV6, so I'll be working on this within the next few days.

PR #23991, opened 2022-03-18
Toyota: remove RAV4h and Highlander(H) from pedal sensitivity condition

cydia2020:
Rav4 hybrid, highlander, and highlander hybrid are[ `NO_STOP_TIMER_CAR`](https://github.com/commaai/openpilot/blob/628eecff06595395f9d5166e82680dc1a2c58af7/selfdrive/car/toyota/values.py#L1810).

Although these cars have very sensitive accelerators, they don't need a comma pedal to get out of a full stop, so it doesn't make sense to include them in the condition.


adeebshihadeh:
I'm not sure I understand the point of this. Are you saying that a pedal shouldn't be used anyway with these cars since they don't need a resume press? In that case, we should maintain a list of cars where we ignore the pedal if it's installed instead of just excluding them from these lists.

cydia2020:
> I'm not sure I understand the point of this. Are you saying that a pedal shouldn't be used anyway with these cars since they don't need a resume press? In that case, we should maintain a list of cars where we ignore the pedal if it's installed instead of just excluding them from these lists.

Yea a pedal is unnecessary on these cars, they can SnG without a resume press, so they shouldn't be included here.

The NO_STOP_TIMER_CAR list is pretty much the "ignore standstill/no pedal required" list, these cars are already in that list, yet they're still included in this if condition which doesn't make sense.

adeebshihadeh:
In that case, this change simply makes these cars drive worse with a pedal. Unless the pedal performance is way worse for some reason, I also don't see a reason to ignore it on certain cars.

PR #23998, opened 2022-03-19
Mazda: no steer unavailable alert for cars with full range steering

### LINKED ISSUE #23984: 2022 Mazda CX-5 - Steer Unavailable Below 0 mph
### Describe the bug

When OpenPilot is engaged, the `Steer Unavailable Below 0 mph` message shows up on screen. I verified the car steers down to 0.

image

comma three

2022 Mazda CX-5

N/A

### openpilot version

0.8.13

### Additional info

_No response_

sshane:


selfdrive/car/mazda/carstate.py
```diff
@@ -72,6 +72,8 @@ def update(self, cp, cp_cam):
         self.lkas_allowed_speed = True
       elif speed_kph < LKAS_LIMITS.DISABLE_SPEED:
         self.lkas_allowed_speed = False
+    else:
+      self.lkas_allowed_speed = True
```
adeebshihadeh, line 76: why not just set this to true in the constructor?

sshane: On cars where minSteerSpeed isn't 0, then behavior will be different if openpilot starts up and we're in between the enable speeds. Before it would default to False which would be correct, but now it would be True, incorrectly allowing lkas when we're still under ENABLE_SPEED

sshane:
@Jafaral, it looks like signal `STEER_RATE->LKAS_BLOCK` is a superset of blocking lkas because of the speed, and whether there's a fault. Do you know if there's a bit that describes the fault separately?

selfdrive/car/mazda/carstate.py
```diff
@@ -15,7 +15,7 @@ def __init__(self, CP):
     self.crz_btns_counter = 0
     self.acc_active_last = False
     self.low_speed_alert = False
-    self.lkas_allowed_speed = False
+    self.lkas_allowed_speed = CP.minSteerSpeed <= 0
```
Jafaral, line 18: Not sure this will do it. `minSteerSpeed` is set to 2 by default by OP.  Change it to this:
```
CP.minSteerSpeed <= 2
```

Jafaral (changes requested):

selfdrive/car/mazda/carstate.py
sshane: Where do you see that? It's set in interfaces here: https://github.com/commaai/openpilot/blob/247b66f4a801d3621575edf5a247c54e275d2826/selfdrive/car/interfaces.py#L80

Jafaral: I could be wrong, I haven't been able to drive/test the last couple of months!. 

Last time a checked: 
https://github.com/commaai/openpilot/blob/master/selfdrive/car/mazda/interface.py#L62

I did notice Mazda CX5 2022 min steer speed was coming at 2.  I didn't check well, but I assumed OP doesn't want to steer below 2 but I could be wrong. 

@adeebshihadeh ?

adeebshihadeh: This is right, but this line is confusing. @sshane let's revert back to setting this in the else block in update()

PR #24002, opened 2022-03-19
test_models: support for running on any route

### LINKED ISSUE #23699: Car port experience improvement
### Current pain points

These will be addressed with documentation, tools, and potentially refactors.

- [ ] debugging invalid CANParser checks
 This should show the whole CANParser state for the invalid messages: seen, frequency, checksum, etc.

### Docs
- [ ] fingerprinting
- [ ] model port
- [ ] brand port
- [ ] tuning?

### Scripts 
- [x] make it easy to run `selfdrive/test/test_models.py` on arbitrary routes - https://github.com/commaai/openpilot/pull/24002

sshane:


adeebshihadeh:
Does this work?

selfdrive/car/tests/test_models.py
```diff
@@ -251,5 +257,33 @@ def test_panda_safety_carstate(self):
...
 
+def load_tests(_routes, ci=True):
+  test_suite = unittest.TestSuite()
+  for car_model, test_route in _routes:
+    test_case_args = {"car_model": car_model, "ci": ci}
+    if test_route is not None:
+      test_case_args.update({"route": test_route.route, "segment": test_route.segment})
+
+    # create new test case and discover tests
+    CarModelTestCase = type("CarModelTestCase", (TestCarModel,), test_case_args)
```
sshane, line 268: There seems to be no support I could find to create multiple test cases from one class with different arguments very easily. Creating dynamic subclasses works and was the simplest method I found.

adeebshihadeh:
Not a fan of losing parallel running. Can we make this work with pytest instead of unittest?

sshane:
Should work now without losing any test_models functionality. Made a new file to interface with test_models so there's no difference in behavior. Also made `TestCarModel` a base class, as the parameterize decorator modifies the original class and removes all its functions, so we couldn't make a test case/suite out of it.

selfdrive/car/tests/test_models.py
```diff
@@ -45,12 +46,17 @@
...
-class TestCarModel(unittest.TestCase):
+class TestCarModelBase(unittest.TestCase):
+  car_model = None
+  test_route = None
+  ci = True
 
   @unittest.skipIf(SKIP_ENV_VAR in os.environ, f"Long running test skipped. Unset {SKIP_ENV_VAR} to run")
   @classmethod
   def setUpClass(cls):
+    if cls.__name__ == 'TestCarModel' or cls.__name__.endswith('Base'):
```
sshane, line 57: Skips if base class, and if subclass with default values for car_model, test_route, etc.

PR #24005, opened 2022-03-19
VW MQB: add brake hold

lskov:
**Route**
Route: 7291d51974eb2d1d|2022-03-18--07-48-26--8

### DIFF
--- a/selfdrive/car/volkswagen/carstate.py
+++ b/selfdrive/car/volkswagen/carstate.py
@@ -49,6 +49,7 @@ class CarState(CarStateBase):
     ret.gasPressed = ret.gas > 0
     ret.brake = pt_cp.vl["ESP_05"]["ESP_Bremsdruck"] / 250.0  # FIXME: this is pressure in Bar, not sure what OP expects
     ret.brakePressed = bool(pt_cp.vl["ESP_05"]["ESP_Fahrer_bremst"])
+    ret.brakeHoldActive = bool(pt_cp.vl["ESP_05"]["ESP_Autohold_aktiv"])
     ret.parkingBrake = bool(pt_cp.vl["Kombi_01"]["KBI_Handbremse"])  # FIXME: need to include an EPB check as well
     self.esp_hold_confirmation = pt_cp.vl["ESP_21"]["ESP_Haltebestaetigung"]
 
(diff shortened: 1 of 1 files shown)


selfdrive/car/volkswagen/carstate.py
```diff
@@ -49,6 +49,7 @@ def update(self, pt_cp, cam_cp, ext_cp, trans_type):
     ret.gasPressed = ret.gas > 0
     ret.brake = pt_cp.vl["ESP_05"]["ESP_Bremsdruck"] / 250.0  # FIXME: this is pressure in Bar, not sure what OP expects
     ret.brakePressed = bool(pt_cp.vl["ESP_05"]["ESP_Fahrer_bremst"])
+    ret.brakeHoldActive = bool(pt_cp.vl["ESP_05"]["ESP_Autohold_aktiv"])
```
adeebshihadeh, line 52: you need to add the signal to the can parser below. see where the rest of the ESP_05 signals are added

adeebshihadeh:
@jyoung8607 look good to you?

jyoung8607:
It does not. I think it's probably the wrong signal.

Some cars (mine included) have physical "auto hold" button, which the driver can use to toggle automatic hydraulic hold behavior when the car comes to standstill. It disables the normal automatic transmission "creep" behavior when releasing the brake at a stop. There are other times it kicks in, like anti-rollback for manual transmissions. I think that's what this signal is for. Neither stock ACC nor openpilot care whether this feature is enabled.

My understanding of brakeHoldActive is we want to populate it when the electromechanical parking brake is active, which would prevent us from engaging stock ACC or openpilot longitudinal. The EPB can be manually set, or triggered by an ACC fault, or triggered by standstill exceeding three minutes. We do have appropriate EPB signals.

Next time I rebase mqb-long I'll check the correct EPB signal and send it in a standalone PR. That said, at a quick glance, I don't think brakeHoldActive does anything except under openpilot longitudinal, so it won't matter till mqb-long is ready.

lskov:
@jyoung8607 from what i could see is this value is the one from the car show when then active hold actually holds the car in brake state if the electromechanically auto hold are enabled. E.g. if I break the car manually when coming to a stop and the car holds the car this one is high, and on hills etc. if this is enabled with the auto hold button we have in some cars, and this is not the signal here. 

Regarding this signal, it doesn't look like it will do anything at this state as you mentioned, and only will be used in future versions with e2e long. Anything i could do to help you test mqb_long let me know.

lskov:
@jyoung8607 Sounds good. I did a short test this afternoon when i parked my car and recorded this. 
image
image

This show (ESP_Autohold_aktiv) where the value is the one from the physical "auto hold" button you talked about, and maybe not the one we want here. There is a value called ESP_Autohold_Standby, which is semilar to this one.

Another thing i observed is the KBI_Handbremse 
image

ESP_Status_Bremsdruck might be the one instead?

adeebshihadeh:
I'll close this for now, as it's not useful to keep this PR up as is.

PR #24008, opened 2022-03-20
plotjuggler: data dir arg for local routes

incognitojam:
Support running plot juggler with local routes

```
$ ls /home/incognitojam/realdata
'0000000000000000|2022-03-20--15-08-07--1'
'0000000000000000|2022-03-20--15-08-07--2'
...

$ ./juggle.py --data_dir="/home/incognitojam/realdata" "0000000000000000|2022-03-20--15-08-07"

### DIFF
--- a/tools/plotjuggler/juggle.py
+++ b/tools/plotjuggler/juggle.py
@@ -75,7 +75,7 @@ def start_juggler(fn=None, dbc=None, layout=None):
   subprocess.call(cmd, shell=True, env=env, cwd=juggle_dir)
 
 
-def juggle_route(route_or_segment_name, segment_count, qlog, can, layout, dbc=None):
+def juggle_route(route_or_segment_name, segment_count, data_dir, qlog, can, layout, dbc=None):
   segment_start = 0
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
We use route and segment names in a bunch of different scripts, and I'd like to support this use case without adding this as an argument to all of them. My two thoughts are either an environment variable or prefixing the route name with the path: `/home/incognitojam/realdata/0000000000000000|2022-03-20--15-08-07`. @gregjhogan what do you thihk?

gregjhogan:
I would update the call to `SegmentName(route_or_segment_name, allow_route_name=True)` to handle `route_or_segment_name` being `/home/incognitojam/realdata/0000000000000000|2022-03-20--15-08-07` by adding a new property named `data_dir` (default to `None` if only a route or segment name is passed in).  Then we can pass the `SegmentName.data_dir` value into `Route()` (like you did in the PR, but now it is not a new param).

pd0wm:
Closing in favour of https://github.com/commaai/openpilot/pull/24473.

PR #24009, opened 2022-03-20
Simulator Tests in CI

gijskoning:
Edit:
Add 3 tests to verify bridge.py and the all processes of manager.py correctly startup and keep running for 60 seconds. Each need Carla simulator to run in parallel.
First two test simply verify that connection to Carla is possible and that bridge.py works as expected.
The third runs manager and bridge. Checkes processes are running and that car can be engaged.

Bridge.py looks like a lot of changes but it is mostly creating a Bridge class and putting some functionality in functions

Also speeds up simulation by changing post processes to false


adeebshihadeh:
For this PR, let's focus on just writing a test and remove the CI stuff. We'll probably need to run this on our Jenkins machine with a GPU, which I'll setup after this is merged.

gijskoning:
Thanks for your feedback. Removed the CI stuff. 
Created basic unittests for connecting with Carla without errors and another for running the bridge without errors for 50 seconds.

I moved some functionality in a bridge class which makes it a bit easier to test.

common/text_window.py
```diff
@@ -31,7 +32,8 @@ def close(self):
 
   def wait_for_exit(self):
     if self.text_proc is not None:
-      while True:
+      params = Params()
+      while not params.get_bool('DoShutdown'):
```
gijskoning, line 36: This could halt during an exception

gijskoning: Unsure about this change. Since the CI seemed to crash on the import line.
I moved it to the function and CI is passing, but I think this is just because it isn't executed in any test

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,108 @@
...
+    p.join()
+    # Assert no exceptions
+    self.assertEqual(p.exitcode, 0)
+
+  def assert_processes_running(self, expected_p):
+    running = {p: False for p in expected_p}
+    for p in expected_p:
+      name = managed_processes[p].get_process_state_msg().name
+      for proc in psutil.process_iter():
+        if proc.name().endswith(name):
```
gijskoning, line 52: @adeebshihadeh This assumes the first argument of the processes in selfdrive/manager/process_config.py
are also used at the end of process name. However just noticing that you are breaking that pattern here: https://github.com/commaai/openpilot/pull/24212
I probably need to change the classes in process_config such that they use the same attribute to define the name of the process.

tools/sim/bridge.py
```diff
@@ -231,266 +234,289 @@ def can_function_runner(vs: VehicleState, exit_event: threading.Event):
...
+    vehicle.apply_physics_control(physics_control)
+
+    transform = carla.Transform(carla.Location(x=0.8, z=1.13))
+
+    def create_camera(fov, callback):
+      blueprint = blueprint_library.find('sensor.camera.rgb')
+      blueprint.set_attribute('image_size_x', str(W))
+      blueprint.set_attribute('image_size_y', str(H))
+      blueprint.set_attribute('fov', str(fov))
+      blueprint.set_attribute('enable_postprocess_effects', 'False')
```
pd0wm, line 313: Gate this behind the `low-quality` flag and use that option during the test?

PR #24010, opened 2022-03-20
2020 Silverado steeractuator delay .11 Steer ratio 16.3

solardude2:
Experimented extensively with individual settings to eliminate placebo and figure out what works. Steer delay has a very small window of effectiveness. Any variation below .1 or above .1 beyond .015 is either too much or too little and becomes ineffective. Not sure if the math goes out of wack and is ignored or what but there is a sweet spot. I had issues with hugging inside of curves. .11 fixes the right curve and helps with left curves. If any other vehicles are having hugging issues try this setting to verify it works for all hugging issues. I am experimenting now with very little adjustments from .11 up and down to see if I can get the truck to center curves both directions with just this setting. 
Also raising the steering ratio has very simular effect as raising kdv. It makes the steering very responsive to adjustments. Please put it back to 16.3. I haven't played with reducing ratio below this but have dropped kdv down to .155. It smooths out the steering greatly but can induce wandering at that level. Thanks Jason!

 elif candidate == CAR.SILVERADO_NR:
      ret.minEnableSpeed = -1. # engage speed is decided by pcm
      ret.minSteerSpeed = -1 * CV.MPH_TO_MS
      ret.mass = 2241. + STD_CARGO_KG
      ret.wheelbase = 3.745
      ret.steerRatio = 16.3 # Determined by skip # 16.3 # From a 2019 SILVERADO
      ret.steerRateCost = 0.45
      ret.centerToFront = ret.wheelbase * 0.49
      ret.steerActuatorDelay = 0.11 # Determined by skip # 0.075
      ret.pcmCruise = True # TODO: see if this resolves cruiseMismatch


adeebshihadeh:
This PR has tons of irrelevant changes due to simply being the dev branch of the fork you happen to be running. You'll need to make a clean branch based off commaai/master with only the relevant changes.

PR #24011, opened 2022-03-21
HKG: improved Kona tuning

haram-KONA:
Adjusting the "kf" value of Hyundai KONA hybrid

Update Interface.py  

By adjusting the Interface.py kf value of Hyundai KONA hybrid 2020, it seems desirable to adjust the "kf" value from "0.00005" to "0.00006" for more stable steering wheel operation and improved stability in onroad..

Reference Test Route.

Dongl-ID : 0665dbbc79ef999e

results: 0665dbbc79ef999e|2022-03-21--14-30-57


HaraldSchafer:
The feedforward scaling is not great in openpilot right now. It scales with angle error, while it should scale with lateral acceleration. We are working on fixing this and should be able to use data to tune all the feedfowards from data. Since that is coming soon I am closing this PR. 

https://github.com/commaai/openpilot/issues/24021

PR #24017, opened 2022-03-22
thermald: non-blocking power save configuration

adeebshihadeh:


selfdrive/thermald/thermald.py
```diff
@@ -305,7 +305,10 @@ def thermald_thread(end_event, hw_queue):
 
       params.put_bool("IsEngaged", False)
       engaged_prev = False
-      HARDWARE.set_power_save(not should_start)
+
+      # this can take more than a second
+      t = threading.Thread(target=HARDWARE.set_power_save, args=(not should_start, ))
```
pd0wm, line 310: Shouldn't we wait with going onroad before this is done? Otherwise cores might not be online yet.

adeebshihadeh:
Closing for now as it's unclear how to cleanly resolve the race condition with on-lining the cores. Let's revisit this when we look into going onroad faster.

PR #24018, opened 2022-03-22
calibrationd: add function typing

herman5:
**Description**
- Adds mypy coverage to function signatures in calibrationd

**Verification**
- `$ mypy selfdrive/locationd/calibrationd.py` 
- Pre-commit hooks, CI

### DIFF
--- a/selfdrive/locationd/calibrationd.py
+++ b/selfdrive/locationd/calibrationd.py
@@ -8,8 +8,9 @@ and the image input into the neural network is not corrected for roll.
 
 import gc
 import os
+import capnp
 import numpy as np
-from typing import NoReturn
+from typing import List, NoReturn, Optional
 
 from cereal import log
 import cereal.messaging as messaging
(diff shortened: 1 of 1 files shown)


selfdrive/locationd/calibrationd.py
```diff
@@ -46,20 +47,20 @@ class Calibration:
   INVALID = 2
 
 
-def is_calibration_valid(rpy):
+def is_calibration_valid(rpy: np.ndarray) -> bool:
```
adeebshihadeh, line 50: can we add the ndarray shape to the type annotation?

adeebshihadeh, line 50: can we add the ndarray shape to the type annotation?

herman5: Typing support for ndarray shapes is still a work in progress: https://github.com/numpy/numpy/issues/16544

Once that and PEP 646 support for mypy lands (https://github.com/python/mypy/issues/12280), shape annotations could be added if Python is upgraded to 3.11

adeebshihadeh: ah, cool!

PR #24019, opened 2022-03-22
car port: comma body

briskspirit:


selfdrive/boardd/boardd.cc
```diff
@@ -115,7 +115,7 @@ bool safety_setter_thread(std::vector<Panda *> pandas) {
     return false;
   }
 
-  pandas[0]->set_safety_model(cereal::CarParams::SafetyModel::ELM327);
+  pandas[0]->set_safety_model(cereal::CarParams::SafetyModel::ALL_OUTPUT);
```
adeebshihadeh, line 118: we'll need to make a safety mode for the body since all output isn't in the release panda build. for now, you can just pass all output from the car interface to remove this hack

selfdrive/car/body/carstate.py
```diff
@@ -0,0 +1,58 @@
...
+    ret.seatbeltUnlatched = False
+
+    ret.cruiseState.speed = 0
+    ret.steeringTorque = 0
+
+    ret.steeringPressed = False
+
+    ret.steeringAngleDeg = 0
+
+    ret.doorOpen = False
```
adeebshihadeh, line 30: I would set all the irrelevant fields in a single big block

selfdrive/car/body/values.py
```diff
@@ -0,0 +1,39 @@
...
+class CAR:
+  BODY = "COMMA BODY"
+
+@dataclass
+class CommaCarInfo(CarInfo):
+  package: str = "All"
+  good_torque: bool = True
+
+CAR_INFO: Dict[str, Union[CommaCarInfo, List[CommaCarInfo]]] = {
+  CAR.BODY: CommaCarInfo("Comma Body"),
```
adeebshihadeh, line 26:
```suggestion
  CAR.BODY: CommaCarInfo("comma Body"),
```

selfdrive/car/body/values.py
```diff
@@ -0,0 +1,39 @@
...
+  ANGLE_DELTA_V = [5., .8, .15]     # windup limit
+  ANGLE_DELTA_VU = [5., 3.5, 0.4]   # unwind limit
+  LKAS_MAX_TORQUE = 1               # A value of 1 is easy to overpower
+  STEER_THRESHOLD = 1.0
+
+class CAR:
+  BODY = "COMMA BODY"
+
+@dataclass
+class CommaCarInfo(CarInfo):
```
adeebshihadeh, line 21: no need for the subclass here. it's just for convenience when setting defaults when we support a bunch of models for a given make

selfdrive/car/fw_versions.py
```diff
@@ -92,6 +92,13 @@ def p16(val):
 
 # brand, request, response, response offset
 REQUESTS = [
+  # Body
+  (
+    "body",
+    [TESTER_PRESENT_REQUEST, UDS_VERSION_REQUEST],
+    [TESTER_PRESENT_RESPONSE, UDS_VERSION_RESPONSE],
```
adeebshihadeh, line 99: if you remove the tester present request, it's the same as honda

selfdrive/car/body/carcontroller.py
```diff
@@ -0,0 +1,110 @@
...
+
+  @staticmethod
+  def deadband_filter(torque, deadband):
+      if torque > 0:
+        torque += deadband
+      else:
+        torque -= deadband
+      return torque
+
+  def update(self, CC, CS, frame, actuators, cruise_cancel, hud_alert,
```
sshane, line 40: Try to use the new CarController conventions, like here: https://github.com/commaai/openpilot/commit/7a2f3c023e0054d1a8fd64b004f8185611323de6

selfdrive/car/body/carcontroller.py
```diff
@@ -0,0 +1,110 @@
+import numpy as np
+from selfdrive.car.body import bodycan
+from opendbc.can.packer import CANPacker
+MAX_TORQUE = 500
+MAX_TORQUE_RATE = 50
+MAX_ANGLE_ERROR = 7
```
sshane, line 6: You can put these in a CarControllerParams in values.py

PR #24024, opened 2022-03-23
events: add clutchPressed event

incognitojam:
Currently when you press the clutch in a manual transmission car, openpilot shows an alert for "wrong gear" and beeps a lot. Making this alert silent (or not showing it at all?) when the clutch is pressed and you are changing gear would be chill :smile: 

Note: Cruise control doesn't disengage while changing gear in the stock system

EDIT: I might only be facing this issue due to my car port - maybe other manufacturers "gear signal" doesn't immediately change when the clutch is pressed? Ford's seems to..

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -117,7 +117,9 @@ class CarInterfaceBase(ABC):
       events.add(EventName.doorOpen)
     if cs_out.seatbeltUnlatched:
       events.add(EventName.seatbeltNotLatched)
-    if cs_out.gearShifter != GearShifter.drive and (extra_gears is None or
+    if cs_out.clutchPressed:
+      events.add(EventName.clutchPressed)
+    elif cs_out.gearShifter != GearShifter.drive and (extra_gears is None or
        cs_out.gearShifter not in extra_gears):
       events.add(EventName.wrongGear)
     if cs_out.gearShifter == GearShifter.reverse:
--- a/selfdrive/controls/lib/events.py
+++ b/selfdrive/controls/lib/events.py
@@ -493,6 +493,14 @@ EVENTS: Dict[int, Dict[str, Union[Alert, AlertCallbackType]]] = {
       Priority.LOW, VisualAlert.steerRequired, AudibleAlert.promptRepeat, 1.),
   },
 
+  EventName.clutchPressed: {
+    ET.WARNING: Alert(
+      "Clutch Pressed",
+      "",
+      AlertStatus.userPrompt, AlertSize.small,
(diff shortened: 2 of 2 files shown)


incognitojam:
Just remembered that I'd need to open a PR to cereal as well, but would it be preferred to just not show the wrong gear event while `cs_out.clutchPressed`?

adeebshihadeh:
I'm not sure we want to support manual cars in mainline openpilot. I'm not sure why carState.clutchPressed was merged.

incognitojam:
It looks like it was added to cereal a long time ago (here https://github.com/commaai/cereal/pull/3) but the only other car port using it is VW (added here https://github.com/commaai/openpilot/pull/20253)

jyoung8607:
> I'm not sure we want to support manual cars in mainline openpilot. I'm not sure why carState.clutchPressed was merged.

It costs virtually nothing to support them using stock ACC, and it's trivial for openpilot long (at least for VW). Removing support would eject customers currently running mainline openpilot, for example, @nabeelr (Nabeel of the eponymous day).

nabeelr:
Yes, please don't break support for my car. ❤️

jyoung8607:
> EDIT: I might only be facing this issue due to my car port - maybe other manufacturers "gear signal" doesn't immediately change when the clutch is pressed? Ford's seems to..

For VW manual trans, the only explicit shifter-in-gate signal we have is for Reverse. There are no signals for the rest. Technically the ECU will infer a forward gear from the RPM::wheel speed ratio with the clutch out, but that only works while moving.

My first implementation supported Reverse, Neutral (clutch pressed in), and assumed Drive at all other times. We discovered that sucks because of the same thing you discovered, the non-Drive gear makes openpilot feel sad. So, now we only do Drive and Reverse. This is perfectly sufficient for a good stock ACC experience, and I propose you do something similar.

For openpilot long, we'll need a little bit more logic, but that will depend on the outcome of @sshane's actuator rearrangement in support of the alternate gas press experience. We'll cross that bridge when Shane is done building it.

incognitojam:
> My first implementation supported Reverse, Neutral (clutch pressed in), and assumed Drive at all other times. We discovered that sucks because of the same thing you discovered, the non-Drive gear while shifting makes openpilot feel sad. So, now we only do Drive and Reverse. This is perfectly sufficient for a good stock ACC experience, and I propose you do something similar.

Ok, cool! I'll try and replicate that instead. I'm happy to close this PR now

adeebshihadeh:
Alright, I think we're cool with manuals as long as they don't require much code. It's a similar compromise as a Honda Accord, for example.

PR #24027, opened 2022-03-23
sensord: add support for Quectel raw gps

geohot:
It matches the old Qualcomm raw gps stuff, I just revived it and used the serial port instead of libdiag.


geohot:
CI failure is fake I think. Ready to be merged.

geohot:
0a48944e30871bb9|2022-03-23--20-37-26 and 0a48944e30871bb9|2022-03-24--08-38-36 both have qcomGnss packets.

PR #24028, opened 2022-03-23
Honda Nidec: don't actuate while longitudinal is inactive

sshane:
When openpilot long isn't active but we're enabled (like from gas press), `actuators.accel` becomes 0 as expected, and so does `self.gas` and `self.brake` in Honda's CarController. However the issue is that the variables calculated from those inputs weren't gated on `CC.longActive`, and a zero input can lead to a non-zero output sent to the car.

Issue fixed:
- PCM applies acceleration while user gas is pressed with no disengage on gas, even if car is ahead

Affected Cars:
- Should have only been an issue for Honda Nidecs running openpilot longitudinal without a pedal

Before (testing no disengage on gas):
We're not sending any `PCM_GAS`, however we're still sending `PCM_SPEED` which does part of the acceleration(?) and the car is in a feedback loop as it's calculated from `vEgo`.
Screenshot from 2022-03-23 16-21-27

After:
Now we're setting both command signals to zero.
Screenshot from 2022-03-23 17-07-04


sshane:
The only process replay diff looks correct:

Screenshot from 2022-03-23 17-45-47

PR #24030, opened 2022-03-24
controlsState: add override state with grey border

sshane:
When pre-enabling with gas pressed we change the UI border and alert color to light grey to signify we're not fully enabled (active).

- `State.overriding` can only be entered when engaged or if we're enabling
- `gasPressed` renamed to `gasPressedPreEnable` to match how it affects the state machine.

What it looks like: https://imgur.com/a/zkJYYZc

*This PR adds all code needed for this change, it doesn't change behavior on its own*


selfdrive/ui/ui.cc
```diff
@@ -202,6 +202,8 @@ void UIState::updateStatus() {
       status = STATUS_WARNING;
     } else if (alert_status == cereal::ControlsState::AlertStatus::CRITICAL) {
       status = STATUS_ALERT;
+    } else if (alert_status == cereal::ControlsState::AlertStatus::OVERRIDE) {
```
sshane, line 205: Should we check the OpenpilotState instead like below instead? I could see it either way

selfdrive/controls/controlsd.py
```diff
@@ -459,14 +469,16 @@ def state_transition(self, CS):
         else:
           if self.events.any(ET.PRE_ENABLE):
             self.state = State.preEnabled
+          elif self.events.any(ET.OVERRIDE):
+            self.state = State.overriding
```
sshane, line 473:
```suggestion
          if self.events.any(ET.PRE_ENABLE) and self.disengage_on_gas:
            self.state = State.preEnabled
          elif self.events.any(ET.OVERRIDE) and not self.disengage_on_gas:
            self.state = State.overriding
```

Can also do this and still only have one gasPressed event, but I don't know about assuming `preEnable` and `override` are always related to gas press.

adeebshihadeh (commented):
Let's write some tests for the state machine before merging this.

selfdrive/controls/tests/test_state_machine.py
```diff
@@ -0,0 +1,69 @@
...
+from selfdrive.car.car_helpers import interfaces
+from selfdrive.controls.controlsd import Controls, ENABLED_STATES
+from selfdrive.controls.lib.events import ET
+
+State = log.ControlsState.OpenpilotState
+
+# The event types that maintain the current state
+MAINTAIN_STATES = {State.enabled: None, State.disabled: None, State.overriding: ET.OVERRIDE,
+                   State.softDisabling: ET.SOFT_DISABLE, State.preEnabled: ET.PRE_ENABLE}
+ALL_STATES = (State.disabled, *ENABLED_STATES)
```
adeebshihadeh, line 14: get this list from `log.ControlsState.OpenpilotState`

selfdrive/controls/tests/test_state_machine.py
```diff
@@ -0,0 +1,69 @@
...
+  # Provides identical API for state_transition
+  def __init__(self):
+    self.et = []
+
+  def any(self, event_type):
+    return event_type in self.et
+
+
+class TestStateMachine(unittest.TestCase):
+  CS = car.CarState()
```
adeebshihadeh, line 29: set this in setUp in case any of the tests change it

selfdrive/controls/tests/test_state_machine.py
```diff
@@ -0,0 +1,69 @@
...
+
+  def setUp(self):
+    CarInterface, CarController, CarState = interfaces["mock"]
+    CP = CarInterface.get_params("mock")
+    CI = CarInterface(CP, CarController, CarState)
+
+    self.controlsd = Controls(CI=CI)
+    self.controlsd.events = Events()
+    self.controlsd.soft_disable_timer = 200  # make sure timer never causes state to change
```
adeebshihadeh, line 38: let's add a test for the soft disable timer

selfdrive/controls/controlsd.py
```diff
@@ -471,6 +481,8 @@ def state_transition(self, CS):
...
```
adeebshihadeh, line 485: add override alert types

selfdrive/test/process_replay/ref_commit
```diff
@@ -1 +1 @@
-f6c93d48d14fa7be45961f7325a7957f53126671
\ No newline at end of file
+f6c93d48d14fa7be45961f7325a7957f53126671
```
adeebshihadeh, line 1: revert this

selfdrive/controls/controlsd.py
sshane: I think here we want to show enable first? preEnable doesn't add alert type here

PR #24033, opened 2022-03-25
Training guide updates

sshane:
Update training guide design and copy changes


sshane:
Waiting on final copy updates

Issue #23985, opened 2022-03-17
Openpilot in Carla isn't accelerating: (engaged: True ; throttle:  0.0)


ebadi:
### Describe the bug

There is known issue with modeld crashing because of libzmq library with [the known solution](https://github.com/commaai/openpilot/issues/23666) that **used to work a few weeks ago**.  It seems that recent changes introduced a new issue. Now openpilot, while engaged, is not accelerating in the CARLA simulator (engaged: True ; throttle:  0.0)

shell 1:
![image](https://user-images.githubusercontent.com/1111940/158785042-d263c862-a535-42b7-9435-35130180b013.png)
shell 2:
![image](https://user-images.githubusercontent.com/1111940/158785060-22dfeeb1-85e8-4680-8699-ba8512d9035e.png)

### OS Version

Ubuntu 18.04

### openpilot version or commit

db7b49c71cfc420259d99d3a4a6f0972b4f887e8

ebadi:
It is possible to use AWSD keys to control the car in CARLA from `bridge.py`. So it seem the issue is with openpilot module.

ebadi:
I tried an old **working** version of the repository and here are the _important_ changes that I see when I run `./launch_openpilot.sh`:

```
selfdrive/locationd/locationd.cc: Locationd vs ubloxLocation position difference too large, kalman reset
...
{"event": "commIssue", "invalid": [], "not_alive": ["deviceState", "pandaStates", "peripheralState", "modelV2", "liveCalibration", "longitudinalPlan", "lateralPlan", "liveLocationKalman", "liveParameters", "radarState", "roadCameraState", "driverCameraState"], "can_error": false, "error": true}
...
selfdrive/modeld/modeld.cc: connected with buffer size: 3493536 (1928 x 1208)
...
ready to run onnx model ['input_imgs', 'desire', 'traffic_convention', 'initial_state'] [[1, 12, 128, 256], [1, 8], [1, 2], [1, 512]]
{"event": "commIssue", "invalid": [], "not_alive": ["driverCameraState"], "can_error": true, "error": true}
```

Ming2888:
@ebadi , where did you find the old working version? I have the same trouble running Openpilot in Carla.

ebadi:
@Ming2888 If you only want to get it running, clone this repo and follow the instruction: https://github.com/jackhong12/openpilot

Other relevant posts:
https://github.com/commaai/openpilot/issues/23695
[...]

PR #24034, opened 2022-03-25
CARLA: fix setting camera EOF

jackhong12:
**Description** [issue #23985](https://github.com/commaai/openpilot/issues/23985)

The scalars of the timestamp are different in real-world and simulation, which causes an infinite loop in modeld.

In `selfdrive/modeld/modeld.cc`, it keeps receiving data until the timestamp of the meta_main is 25000000 more than the timestamp of the meta_extra. However, the timestamp value is much smaller than 25000000 in simulation. So the loop below will cause modeld to hang. I guess the timestamp unit here is nanosecond. But the unit in `bridge.py` is second.
 ```c++
    while (get_ts(meta_main) < get_ts(meta_extra) + 25000000ULL) {
      buf_main = vipc_client_main.recv(&meta_main);
      if (buf_main == nullptr)  break;
    }
```

The timestamp is assigned in `tools/sim/bridge.py`.  I think one way to fix this issue is to multiply `eof` by a constant value.
```python
    # convert RGB frame to YUV
    rgb = np.reshape(img, (H, W * 3))
    rgb_cl = cl_array.to_device(self.queue, rgb)
    yuv_cl = cl_array.empty_like(rgb_cl)
    self.krnl(self.queue, (np.int32(self.Wdiv4), np.int32(self.Hdiv4)), None, rgb_cl.data, yuv_cl.data).wait()
    yuv = np.resize(yuv_cl.get(), np.int32(rgb.size / 2))
    eof = self.frame_id * 0.05

    # TODO: remove RGB send once the last RGB vipc subscriber is removed
    self.vipc_server.send(VisionStreamType.VISION_STREAM_RGB_BACK, img.tobytes(), self.frame_id, eof, eof)
    self.vipc_server.send(VisionStreamType.VISION_STREAM_ROAD, yuv.data.tobytes(), self.frame_id, eof, eof)

    dat = messaging.new_message('roadCameraState')
    dat.roadCameraState = {
      "frameId": image.frame,
      "transform": [1.0, 0.0, 0.0,
                    0.0, 1.0, 0.0,
                    0.0, 0.0, 1.0]
    }
    pm.send('roadCameraState', dat)
    self.frame_id += 1
```

After modifying the value of `eof`, Openpilot can accelerate again. I am not sure whether this fix will cause other problems. Let me know if there is a better way to fix this issue. 
```python
    eof = int((time.time() - startTime) * 1000000000)
```

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -39,6 +39,7 @@ STEER_RATIO = 15.
 pm = messaging.PubMaster(['roadCameraState', 'sensorEvents', 'can', "gpsLocationExternal"])
 sm = messaging.SubMaster(['carControl', 'controlsState'])
 
+startTime = time.time()
 
 class VehicleState:
   def __init__(self):
@@ -94,7 +95,7 @@ class Camerad:
     yuv_cl = cl_array.empty_like(rgb_cl)
     self.krnl(self.queue, (np.int32(self.Wdiv4), np.int32(self.Hdiv4)), None, rgb_cl.data, yuv_cl.data).wait()
     yuv = np.resize(yuv_cl.get(), np.int32(rgb.size / 2))
-    eof = self.frame_id * 0.05
+    eof = int((time.time() - startTime) * 1000000000)
 
     # TODO: remove RGB send once the last RGB vipc subscriber is removed
     self.vipc_server.send(VisionStreamType.VISION_STREAM_RGB_ROAD, img.tobytes(), self.frame_id, eof, eof)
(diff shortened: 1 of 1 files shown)


tools/sim/bridge.py
```diff
@@ -94,7 +95,7 @@ def cam_callback(self, image):
     yuv_cl = cl_array.empty_like(rgb_cl)
     self.krnl(self.queue, (np.int32(self.Wdiv4), np.int32(self.Hdiv4)), None, rgb_cl.data, yuv_cl.data).wait()
     yuv = np.resize(yuv_cl.get(), np.int32(rgb.size / 2))
-    eof = self.frame_id * 0.05
+    eof = int((time.time() - startTime) * 1000000000)
```
adeebshihadeh, line 98: I think we still want to set this based on the frame ID. 
```suggestion
    eof = int(self.frame_id * 0.05 * 1e9)
```

adeebshihadeh:
Thanks for the PR!

PR #24038, opened 2022-03-25
ui: repaint toggles after offroadTransition

sshane:
When going offroad in the toggles panel, they become enabled however they're never repainted so they look disabled (I trusted it so much I never tried clicking them to see if they actually were!)

### DIFF
--- a/selfdrive/ui/qt/widgets/controls.h
+++ b/selfdrive/ui/qt/widgets/controls.h
@@ -101,7 +101,7 @@ public:
     QObject::connect(&toggle, &Toggle::stateChanged, this, &ToggleControl::toggleFlipped);
   }
 
-  void setEnabled(bool enabled) { toggle.setEnabled(enabled); toggle.repaint(); }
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/widgets/controls.h
```diff
@@ -101,7 +101,7 @@ class ToggleControl : public AbstractControl {
     QObject::connect(&toggle, &Toggle::stateChanged, this, &ToggleControl::toggleFlipped);
   }
 
-  void setEnabled(bool enabled) { toggle.setEnabled(enabled); }
+  void setEnabled(bool enabled) { toggle.setEnabled(enabled); toggle.repaint(); }
```
adeebshihadeh, line 104: generally, prefer update to repaint. update queues up a repaint whereas repaint forces one
```suggestion
  void setEnabled(bool enabled) { toggle.setEnabled(enabled); toggle.update(); }
```

sshane: Ah I forgot, thanks!

PR #24052, opened 2022-03-27
controlsd: no cruise speed while main off

adeebshihadeh:


selfdrive/controls/controlsd.py
```diff
@@ -403,8 +403,11 @@ def state_transition(self, CS):
     # if stock cruise is completely disabled, then we can use our own set speed logic
     if not self.CP.pcmCruise:
       self.v_cruise_kph = update_v_cruise(self.v_cruise_kph, CS.buttonEvents, self.button_timers, self.enabled, self.is_metric)
-    elif CS.cruiseState.enabled:
-      self.v_cruise_kph = CS.cruiseState.speed * CV.MS_TO_KPH
+    else:
+      if CS.cruiseState.available:
+        self.v_cruise_kph = CS.cruiseState.speed * CV.MS_TO_KPH
+      else:
+        self.v_cruise_kph = 0
```
sshane, line 410: If we set this to 255 instead we can clean up a bunch of magic numbers and only check against `SET_SPEED_NA` here https://github.com/commaai/openpilot/blob/master/selfdrive/ui/qt/onroad.cc#L184

sshane: However now that the `enabled` check is removed, this would be fixed on Toyotas without setting to 0 as the cruise speed from the car resets itself to 0 as well, not sure about other cars

PR #24053, opened 2022-03-28
hyundai: better longitudinal control

gregjhogan:
The following changes improve starting from a stop and make brake control more consistent and lag less:
* when stopped and you want to start moving, make a large accel request until the brakes release
* always tell the vehicle there is a lead 1 meter away

Both of these seem to be required for a consistent, quick start start from stop.  It still takes around a second, but I think it is as fast as stock and no one will complain (any slow movement after you start rolling is the planners fault).

Always telling the vehicle there is a lead 1 meter away seems to make braking way more responsive.  So much that it seems like the lag is only about 0.5 seconds (or less) normally now and we no longer need to add double the error where accel is negative.

This has only been tested on a 2020 palisade and the request values used to get you started when stopped may need to be adjusted for different vehicles (but hopefully not).  I set these values kind of conservative for now because if you set them higher the vehicle does start faster, but it will always go all the way to the requested value and this can cause a harsh jerk when the transition happens back to the plan accel (if it is still small and positive).

I also noticed that sometimes the plan can take a while to reach `vEgoStarting` which caused further delay starting from a stop, so lowering it makes starting from a stop faster and more consistent.  Perhaps we want to do this for other manufacturers, too.

### DIFF
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -1,6 +1,6 @@
 from cereal import car
 from common.realtime import DT_CTRL
-from common.numpy_fast import clip, interp
+from common.numpy_fast import clip
 from common.conversions import Conversions as CV
 from selfdrive.car import apply_std_steer_torque_limits
 from selfdrive.car.hyundai.hyundaican import create_lkas11, create_clu11, create_lfahda_mfc, create_acc_commands, create_acc_opt, create_frt_radar_opt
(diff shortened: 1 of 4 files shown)


pd0wm:
As discussed on slack, let's make some plots of tracking a step response of -2m/s^2 under the following conditions 1) current control with accel interp 2) this PR with lead distance set 3) this PR without lead distance set. Then we can refer back to this PR if we ever look into changing it again.

gregjhogan:
see decel control enabling faster and it doesn't smooth into the target (the latter is probably what used to cause randomly trying to rear-end lead vehicles)

before:
hyundai-long-before-improved3

after:
hyundai-long-after-improved3

that is at high speed (35-40 mph), at low speed the difference is definitely more noticeable

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -88,20 +88,29 @@ def update(self, CC, CS):
...
-      lead_visible = False
+      pid_control = (actuators.longControlState == LongCtrlState.pid)
+      stopping = (actuators.longControlState == LongCtrlState.stopping)
       accel = actuators.accel if CC.longActive else 0
+      # TODO: aEgo lags when jerk is high, use smoothed ACCEL_REF_ACC instead?
+      accel_error = accel - CS.out.aEgo if CC.longActive else 0
 
-      jerk = clip(2.0 * (accel - CS.out.aEgo), -12.7, 12.7)
+      # TODO: jerk upper would probably be better from longitudinal planner desired jerk?
+      jerk_upper = clip(2.0 * accel_error, 0.0, 1.5) # zero when error is negative to keep decel control tight
```
gregjhogan, line 98: the positive jerk clip at 1.5 isn't ideal here, but 12.7 seems crazy high to me, so what should we clip it to?

pd0wm: ISO 15622:2018 has no limits for positive jerk. Guess you can never reach something that high anyway on an ICE car.

pd0wm:
I assume this PR without lead distance set is the the worst of the three. But you can add it for completeness just to make sure we're not missing something.

selfdrive/car/hyundai/carcontroller.py
gregjhogan: That makes sense, but since acceleration is going to always have bad lag at times on gas cars, which leads to big error, which leads to huge jerk allowed, I think it makes sense to clip this at something reasonable to keep driving comfortable and ensure massive overshoot doesn't happen.  Harald and I talked and we think 2 m/s3 is the most we ever really want once you are moving (this PR handles setting positive jerk to initially get moving separately) so I raised it to that.

sshane:
- [ ] Stock sets the jerk limits to 0 when not engaged, need to see if that hurts performance shortly after engaging

selfdrive/car/hyundai/hyundaican.py
```diff
@@ -78,27 +78,27 @@ def create_lfahda_mfc(packer, enabled, hda_set_speed=0):
...
     "VSetDis": set_speed if enabled else 0,
     "AliveCounterACC": idx % 0x10,
-    "ObjValid": 1 if lead_visible else 0,
-    "ACC_ObjStatus": 1 if lead_visible else 0,
+    "ObjValid": 1, # close lead makes controls tighter
+    "ACC_ObjStatus": 1, # close lead makes controls tighter
     "ACC_ObjLatPos": 0,
     "ACC_ObjRelSpd": 0,
-    "ACC_ObjDist": 0,
+    "ACC_ObjDist": 1, # close lead makes controls tighter
```
sshane, line 93: I think someone said this causes a lead to show up on their dash now, is this permanent, and do we care?

gregjhogan: ObjGap below is what causes the lead to be displayed

gregjhogan:
there is still some room for improvement, and I am guessing it is correlated to allowing high positive jerk in SCC14 while the plan transitions from increasing acceleration to decreasing acceleration

`5decac85ab437ed1|2022-04-16--13-17-07--19` is an example of where there was still quite a bit of lag switching from accelerating to decelerating (while the car did stop in time, it wasn't exactly comfortable)
image

this can be improved by lowering max positive jerk for SCC14 (which my last commit raised to 2.0 m/s^3) or by feeding the desired jerk from the plan into the SCC14 jerk limits

currently positive jerk is based on plan error which is not ideal because it can grow much larger than what is desired, which leads to more overshoot and delayed braking

sshane:
You can see here a case where even the stock system's requested acceleration wasn't being met by the car, so I'm not sure if these changes are solely limited to certain cars. Here the car set both the obj status bits to 1, the relative speed and distance signals were filled with accurate values, and the lower jerk was set relatively high:
Screenshot from 2022-05-26 21-43-28

sshane:
Posted some of my findings from this PR as a comment here: https://github.com/commaai/openpilot/issues/24672. Overall, I didn't see any change in almost your exact test (did 2 m/s/s instead of 1 m/s/s which should be worse)

adeebshihadeh:
superseded by #25604

gregjhogan:
we have confirmed that these changes do improve braking responsiveness and starting from a stop in the sonata as well as the palisade

adeebshihadeh:
> we have confirmed that these changes do improve braking responsiveness and starting from a stop in the sonata as well as the palisade

yeah, just closed since Harald opened a new based on this. did you want to also keep this one open?

PR #24054, opened 2022-03-28
Toyota: use driver torque for torque limits

sshane:
Just like all the other cars. Chrysler is the only make left, and if that works well with this common method then we can remove EPS torque limiting entirely (from openpilot and panda), as it sometimes limits torque needlessly when the driver isn't touching the wheel at all.

Unclear if this provides much better control or placebo, need to get a route on the same stretch of road and compare.

Might fault without updating panda torque limiting if you try this branch


selfdrive/car/toyota/values.py
```diff
@@ -20,7 +20,9 @@ class CarControllerParams:
   STEER_MAX = 1500
   STEER_DELTA_UP = 10       # 1.5s time to peak torque
   STEER_DELTA_DOWN = 25     # always lower than 45 otherwise the Rav4 faults (Prius seems ok with 50)
-  STEER_ERROR_MAX = 350     # max delta between torque cmd and torque motor
+  STEER_DRIVER_ALLOWANCE = 50
+  STEER_DRIVER_MULTIPLIER = 1
+  STEER_DRIVER_FACTOR = 10  # TODO: doesn't totally match
```
sshane, line 25: Felt good, however openpilot's steering definitely felt weaker with these values (ie. lower effective steer error).

BrettLynch123:
Tested and working on my 2020 Corolla hybrid. Also not sure if this provides much better control or placebo but its not worse.

sshane:
While testing the PR to kill steer faults, I noticed it sucked at taking turns where we would drift across multiple lanes and take a long time to recover, turns out it's just the openpilot Toyota EPS torque limiting the torque output heavily!

To clarify, this occurs when the wheel falls back to center too quickly so we cut torque intermittenly to prevent any faults. Then the EPS torque slowly drifts away from the commanded torque as there seems to be a small windup after you set the request bit (which we would do every 20 frames) Here's a plot of the EPS torque being relatively far away from the desired causing it to limit the output and not meet the desired angle/accel

Screenshot from 2022-05-08 00-47-57

sshane:
See comments in https://github.com/commaai/panda/pull/936

PR #24062, opened 2022-03-29
controlsState: enter pre-enable when engaging at a standstill with brake

sshane:


selfdrive/car/interfaces.py
```diff
@@ -124,8 +124,8 @@ def create_common_events(self, cs_out, extra_gears=None, pcm_enable=True):
       events.add(EventName.wrongCarMode)
     if cs_out.espDisabled:
       events.add(EventName.espDisabled)
-    if cs_out.gasPressed:
-      events.add(EventName.gasPressed)
+    if cs_out.gasPressed or cs_out.brakePressed:
```
adeebshihadeh, line 127: add the standstill check here

sshane: We only enter pre-enable state if there's no NO_ENTRY events and an ENABLE event, so `pedalPressedPreEnable` only has any affect when `pedalPressed` isn't sent

adeebshihadeh: yeah, the state machine will handle it, but better to be explicit

sshane:
Vehicle-specific issues make this difficult to do (Honda operates on brake pressure, so when you release brake no matter how fast, you move forward)

PR #24063, opened 2022-03-29
openpilot longitudinal: fix long press to increment by 5 mph

### LINKED ISSUE #23997: Hyundai longitudinal: long press to increment by 5 mph inconsistent
Does not work 100% of the time, more likely to work shortly after incrementing by just one?

Route example: 5decac85ab437ed1|2022-03-18--07-53-37

sshane:


selfdrive/controls/lib/drive_helpers.py
```diff
@@ -66,11 +66,12 @@ def update_v_cruise(v_cruise_kph, buttonEvents, button_timers, enabled, metric):
 
   if button_type:
     v_cruise_delta = v_cruise_delta * (5 if long_press else 1)
-    if long_press and v_cruise_kph % v_cruise_delta != 0: # partial interval
+    partial_interval = round(v_cruise_kph / v_cruise_delta, 1) % 1 != 0
+    if long_press and partial_interval:
       v_cruise_kph = CRUISE_NEAREST_FUNC[button_type](v_cruise_kph / v_cruise_delta) * v_cruise_delta
     else:
       v_cruise_kph += v_cruise_delta * CRUISE_INTERVAL_SIGN[button_type]
-    v_cruise_kph = clip(round(v_cruise_kph, 1), V_CRUISE_MIN, V_CRUISE_MAX)
```
sshane, line 73: This is what was mainly causing the inconsistencies, if you round the down then when you do the modulo, it can end up being near v_cruise_delta

gregjhogan: What changed? Pretty sure the opposite was true when I wrote this.

gregjhogan: It was probably the `v_cruise_delta` change in #23962 we may want to revert that or add unit tests for this logic

gregjhogan: Are you sure even with your change it works correctly across all set speeds? I would write a unit test to make sure (and so this doesn't happen again).

gregjhogan: Not surprising, that means `v_cruise_delta` was never intended to have an accurate conversion, and we need to be way more careful merging changes without good tests.  Sounds like we should revert the change and write unit tests that exercise every branch condition of this function and over the reasonable range of speeds.

sshane: adeeb doesn't think we need to (at least all branches), especially if we don't do float modulo and don't use magic numbers (instead check if we're close enough to a full interval)

gregjhogan: looks like it should work now, I guess I didn't think it would be hard to have complete test coverage with unit tests, but maybe I am missing something.

sshane: It's about 100 lines, think it's worth it?: https://github.com/commaai/openpilot/pull/24099/files

gregjhogan: high quality tests should be worth it, copying equations from the function being tested probably doesn't make for a high quality test

spektor56: this is STOCK data I got from someone's Honda civic when I was testing the set speed logic.   Their speed in MPH on right and the value from cabana in KM/H on the left.  I'm really not sure how they do their rounding logic :p

40 = 25
42 = 26
43 = 27
45 = 28
46 = 29
48 = 30
50 = 31
51 = 32
53 = 33
54 = 34
56 = 35
58 = 36
59 = 37
61 = 38
62 = 39
64 = 40
66 = 41
67 = 42
69 = 43
70 = 44
72 = 45
74 = 46
75 = 47
77 = 48
78 = 49
80 = 50
82 = 51
83 = 52
85 = 53
86 = 54
88 = 55
90 = 56
91 = 57
93 = 58
94 = 59
96 = 60
98 = 61
99 = 62
101 = 63
103 = 64
104 = 65
106 = 66
107 = 67
109 = 68
111 = 69
112 = 70
114 = 71
115 = 72
117 = 73
119 = 74
120 = 75
122 = 76
123 = 77
125 = 78
127 = 79
128 = 80
130 = 81
131 = 82
133 = 83
135 = 84
136 = 85
138 = 86
139 = 87
141 = 88
143 = 89
144 = 90

adeebshihadeh:
Reverted the change for now in #24161. This bug was asking to get written with the way it was written, so I added a comment too. This PR even further complicates this code - can we simplify it instead?

sshane:
Will revisit with good tests first

PR #24064, opened 2022-03-29
api.cc: expose error type in requestDone signal

pd0wm:

### DIFF
--- a/selfdrive/ui/qt/api.h
+++ b/selfdrive/ui/qt/api.h
@@ -31,7 +31,7 @@ public:
   bool timeout() const;
 
 signals:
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/api.h
```diff
@@ -31,7 +31,7 @@ class HttpRequest : public QObject {
   bool timeout() const;
 
 signals:
-  void requestDone(const QString &response, bool success);
+  void requestDone(const QString &response, bool success, QNetworkReply::NetworkError error);
```
pd0wm, line 34: extra arguments are safely ignored, so no changes to the other code is needed for this.

PR #24065, opened 2022-03-29
camera error: be verbose

geohot:


geohot:
This isn't even good enough, I want which issue caused it on the second line.

PR #24067, opened 2022-03-29
Toyota: remove 100°/sec steering lockout

sshane:
Will need to look at a lot of different cars to verify the numbers, but in a Toyota Camry Hybrid 2022, steering faults when the steering angle rate from the message `STEER_ANGLE_SENSOR` is above 100 degrees/sec for longer than either 21 or 25 cycles/frames.

Not sure what the purpose is of this, but the signal we check for temporary steering faults (`EPS_STATUS->LKA_STATE`) goes to either 25 or 21 on rising edge of a fault, and it seems to describe to the number of sent steering messages where the wheel has been above 100 deg/s. We can also request zero torque and set the torque request bit to 1, and we still fault under the exact same conditions. So the EPS isn't looking at commanded torque, driver torque, or EPS torque at all for faults.

With this change, the user can override as fast as possible and no faults should occur as we cut steering until the rate falls below the threshold the EPS seems to have internally.

panda PR: https://github.com/commaai/panda/pull/939


BrettLynch:
Tested and working great on my 2020 Corolla hybrid.

cydia2020:
Works well on 18 prius

ntegan1:
works well on 17 corolla.

especially helps out with joystick, making full runs to taco bell much easier.

since you no longer need to stay below the rate limit manually, you can focus on smoother control.

might need to notify the user when it's being rate limited, like a mini steerTempUnavailable

BrettLynch:
Also working well on 2022 rav4 hybrid. Might use it daily as it removes annoying warning: steering unavailable messages when manually steering sharp curves while engaged.

sshane:
Just need panda safety and this should be good to go! Will post notebook results as well. Once we correctly cut torque, you don't notice at all that it cuts torque for a frame at 5 hz

ntegan1:
Did some extensive testing with joystick and no temp/perm faults at all!  
route has many examples of cutting `apply_steer_req`  
b50bf8667197cbdb|2022-05-29--19-57-33  
image

selfdrive/car/toyota/carcontroller.py
```diff
@@ -22,6 +26,8 @@ def __init__(self, dbc_name, CP, VM):
     self.standstill_req = False
     self.steer_rate_limited = False
 
+    self.steer_rate_counter = 0
+
     self.packer = CANPacker(dbc_name)
```
ntegan1, line 31:
```suggestion
    self.standstill_req = False
    self.steer_rate_limited = False
    self.steer_rate_counter = 0

    self.packer = CANPacker(dbc_name)
```
whitespace nitpick maybe?

sshane:
Just wanted to verify the thresholds, so I ran on 685,309 segments (TODO: include dongles and platforms)

Of all the faults (505 unique temporary steering faults, and two permanent faults filtered out), the minimum number of (100hz CAN) frames where the rate is above 100 deg/s is 21, with the mean being around 27 frames until it faults above 100 deg/s. So a threshold of 19 steering message frames should be more than enough to avoid any faults.

The only outlier is a 2020 Highlander Hybrid which faulted in 17 frames, however its steering rate went from 0 to ~140 in one frame, so I suspect something weird happened there (`fe7c7758734362ae|2022-03-21--13-16-00--6`)

And here's just a comparison of using 100Hz CAN frames to count vs. using when the steering rate message was updated (80Hz, doesn't seem to make it more consistent). I'm also not too sure what conditions allow some cars to fault after much longer than 20 frames, but it seems relatively rare according to the plots below (50 faults greater than 30 frames, worth looking into eventually, but even the chosen 19 frames provides great control)

```python
Faults with 0 frames: 0 (2 permanent faults filtered out)
Fault frames vl_all mean: 22.805940594059408, std: 5.340482823542282
Fault frames vl_all min: 17, max: 70
Fault frames vl     mean: 26.766336633663368, std: 5.812998020444156
Fault frames vl     min: 21, max: 85
Max: 85, min: 21
Total valid faults: 505 in 685309 segments
```

Distribution plots: https://imgur.com/a/g77AyYa

selfdrive/car/toyota/carcontroller.py
```diff
@@ -52,11 +57,19 @@ def update(self, CC, CS):
...
+    else:
+      self.steer_rate_counter = 0
+
+    apply_steer_req = 1
     if not CC.latActive:
       apply_steer = 0
       apply_steer_req = 0
-    else:
-      apply_steer_req = 1
+    elif (self.steer_rate_counter - 1) >= MAX_STEER_RATE_FRAMES:
```
adeebshihadeh, line 70:
```suggestion
    elif self.steer_rate_counter > MAX_STEER_RATE_FRAMES:
```

sshane:
Sanity test -- cutting torque every 19 frames, no faults!

Screenshot from 2022-09-06 19-00-36

PR #24068, opened 2022-03-29
Single tuning torque control

haraschax:


khoi:
Been using this for my badly tuned HKG. Works great, I changed kf to 0.1 / MAX_TORQUE which made the steering wheel less jerky in high speed

PR #24070, opened 2022-03-29
Honda: use speed for standstill signal on affected models

sshane:
Until we can find and verify a new wheel moving bit, this is probably the best change for now. The current bit is much too noisy: https://imgur.com/a/FxABcxV

Civic Nidec is in the else block, and HRV is in the block above that, only two blocks that don't already use the `XMISSIONS_SPEED`, both noisy.

### DIFF
--- a/selfdrive/car/honda/carstate.py
+++ b/selfdrive/car/honda/carstate.py
@@ -192,7 +192,7 @@ class CarState(CarStateBase):
       ret.standstill = not cp.vl["STANDSTILL"]["WHEELS_MOVING"]
       ret.doorOpen = bool(cp.vl["SCM_BUTTONS"]["DRIVERS_DOOR_OPEN"])
     else:
-      ret.standstill = not cp.vl["STANDSTILL"]["WHEELS_MOVING"]
+      ret.standstill = cp.vl["ENGINE_DATA"]["XMISSION_SPEED"] < 0.1
       ret.doorOpen = any([cp.vl["DOORS_STATUS"]["DOOR_OPEN_FL"], cp.vl["DOORS_STATUS"]["DOOR_OPEN_FR"],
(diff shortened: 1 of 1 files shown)


selfdrive/car/honda/carstate.py
```diff
@@ -192,7 +192,7 @@ def update(self, cp, cp_cam, cp_body):
       ret.standstill = not cp.vl["STANDSTILL"]["WHEELS_MOVING"]
       ret.doorOpen = bool(cp.vl["SCM_BUTTONS"]["DRIVERS_DOOR_OPEN"])
     else:
-      ret.standstill = not cp.vl["STANDSTILL"]["WHEELS_MOVING"]
+      ret.standstill = cp.vl["ENGINE_DATA"]["XMISSION_SPEED"] < 0.1
```
pd0wm, line 195: This is not a problem on the FREED and HRV?

sshane: Oh I think it is as well, forgot where someone reported that to me. However even the XMISSION_SPEED signal has some crazy high minimum speed it reports (or it's gated on brake pressed), as I was able to creep at at least a few mph and the signal was still 0. I want to look for a new bit for the standstill, just haven't had time

sshane: Used a script to find any bits matching CS.vEgo < (or greater than) 0.1 m/s, however nothing matched as close as using the `XMISSION_SPEED` which was the most accurate and not noisy at all compared to `STANDSTILL->WHEELS_MOVING`. However it does drop to zero at 1 kph, so you can engage up to half a mph with brake pressed.

sshane:
This is causing a portion of the controls mismatches for the MTBF report:

Routes with noisy old standstill signals:
- HRV: 6a1738eb77154557|2022-04-19--08-33-34--0
- Civic Nidec: cfb32f0fb91b173b%7C2022-03-17--08-21-07

sshane:
Moved to https://github.com/commaai/openpilot/pull/24262 to add a standstill check at the same time

PR #24072, opened 2022-03-29
Lock SR within EKF with bool in carParams

ClockeNessMnstr:
SR causes a lot of headaches and personally was never accurate on my vehicle. 
Steering was jerky and cut hard corners very badly or oversteered to the curb. 

Just overriding the value causes it's own issues. namely stiffness factor is not learned effectively. but it's done in  forks because it still performs better in some cases. 

After setting my steering ratio to the stock value Stiffness Factor was re-learned to ~0.665 
I used this value in my interface.py to scale the front/rear stiffness and now have the improved accuracy steering in straights and curves without breaking the learner, so now it can learn stiffness and any others from the dynamic model while locking the SR.

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -77,6 +77,8 @@ class CarInterfaceBase(ABC):
     ret.steerControlType = car.CarParams.SteerControlType.torque
     ret.minSteerSpeed = 0.
     ret.wheelSpeedFactor = 1.0
+    
+    ret.learnSR = True
 
     ret.pcmCruise = True     # openpilot's state is tied to the PCM's cruise state on most cars
     ret.minEnableSpeed = -1. # enable is done by stock ACC, so ignore this
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -78,7 +78,7 @@ class CarInterfaceBase(ABC):
     ret.minSteerSpeed = 0.
     ret.wheelSpeedFactor = 1.0
     
-    ret.learnSR = True
+    ret.lockSR = False
 
     ret.pcmCruise = True     # openpilot's state is tied to the PCM's cruise state on most cars
     ret.minEnableSpeed = -1. # enable is done by stock ACC, so ignore this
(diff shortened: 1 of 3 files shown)


nuwandavek:
This might work in one specific case, but will certainly break in cars where the init `steerRatio` is off. The learner is designed to converge to the correct value, accounting for the vehicle model. Learning the `stiffnessFactor` this way is the right idea and can be used to set the init value for this car model, but we cannot do this in the live learner. 
Could you post a segment (with uploaded rlogs) where you felt like the learnt values were wrong? I can take a look at it and see if there's something wrong.

ClockeNessMnstr:
Grabbing a route now for this and the "save sf" PR. 
That one's largely due to this one if the value can be off significantly between cars if the same model

ClockeNessMnstr:
b51dad8c8187542b|2022-03-30--20-21-50--0
The steering ratio on my vehicle is 13.5, learner learns about 15.5. 
I take a set of turns at the beginning and end of the route. it already learned most of the way before the first set but it gets a bit worse through the second time still.

nuwandavek:
Hey, that route is not on `commaai/openpilot/master`. I can check out if there is an underlying problem if you send me a route where you feel there's an issue, while running openpilot master/release. We can't merge this PR because it will definitely break a lot more cars where the `steer_ratio` might be off. We need to learn this live, as of now. If we want to not learn it, we need to do a deeper analysis on a distribution of cars (from the same model) on our fleet (this is not in our immediate roadmap).

Also remember that the exact values of the `steer_ratio` and `stiffness_factor` don't matter all that much. There are many free variables in that kalman filter to globally optimize for the steering offsets. They just needs to be in the neighborhood.

ClockeNessMnstr:
I ran a route with the steer ratio learning but kept the initial stiffness. it kept the SR to about 14 but is not learning correctly. 

Sorry I can't run master, the tuning on it is not really usable even on straights and it can't make any of the turns in my route. 
I could run a cleaner branch. 

Also, I did structure this change as 0 impact. tests passed including process replay. for cars with a known steering ratio the new "hidden" carparam needs to be set to True.
so it wouldn't have any impact until someone "tuned" the init stiffness value for the car and enabled the lock with the known SR.

PR #24073, opened 2022-03-30
controls: transition into full stopAccel if already at a standstill

sshane:


selfdrive/car/honda/carcontroller.py
```diff
@@ -135,7 +135,10 @@ def update(self, CC, CS):
                                                                            CS.out.vEgo, self.CP.carFingerprint)
 
     # *** rate limit after the enable check ***
-    self.brake_last = rate_limit(pre_limit_brake, self.brake_last, -2., DT_CTRL)
+    if not CS.out.standstill:
```
sshane, line 138: This standstill signal is flaky, so not too useful yet

sshane:
This was to fix the Hondas not braking fast enough when testing no actuation in a pre-enable state, however even with this, it isn't quick enough and can lurch forward if quickly engaging and taking your foot off the brake.

PR #24078, opened 2022-03-30
jenkins: kill potentially leftover processes

adeebshihadeh:


selfdrive/test/setup_device_ci.sh
```diff
@@ -20,6 +20,10 @@ fi
 umount /data/safe_staging/merged/ || true
 sudo umount /data/safe_staging/merged/ || true
 
+pkill -f manager.py
```
pd0wm, line 23:
```suggestion
pkill -9 -f manager.py
```

in case it hangs?

adeebshihadeh: let's see how this works first. want to be careful with un-sigkillable processes like camerad https://github.com/commaai/openpilot/blob/a688811aa8aab10ea0d50a9b53961c8244731092/selfdrive/manager/process_config.py#L10

PR #24081, opened 2022-03-30
tools: verify new car ports and fingerprints

### LINKED ISSUE #23905: Improved procedure for merging car ports and fingerprints
Our current procedure for merging fingerprints is basically “does the user say it works?” and ensuring they copied the values correctly. We need a more rigorous procedure to merge fingerprints to prevent the recent [Lexus NX issue](https://github.com/commaai/openpilot/issues/23870) and [Hyundai cancel issue](https://github.com/commaai/openpilot/issues/23674), especially since this is part of the point of FW fingerprinting. The same procedure should apply to car ports as well. 

Eventually, I think we want a route for each set of fingerprints so that we can ensure the full carState is parsed correctly, where the user opens and closes the doors, seatbelt, etc. For now, let’s make a script that takes a route and ensures some basic functionality:
* fingerprints correctly and fully (i.e. not fuzzy)
* engaged for more than a minute continuously with no gas or steering overrides
* gas cancellation is immediate and reliable
* runs through selfdrive/test/test_models.py

As a bonus, it would be nice if this ran automatically on PRs.

sshane:


sshane:
Will re-open when ready

PR #2416, opened 2020-10-26
Car Port: 2020 Kia Niro EV

nickn17:
***** Template: Car port Kia Niro EV *****

**Checklist**
- [x] added to README
- [x] test route added to [test_car_models](../../selfdrive/test/test_car_models.py)
- [x] route with openpilot: 50c6c9b85fd1ff03|2020-10-26--17-56-06
- [x] route with stock system: 50c6c9b85fd1ff03|2020-10-25--11-25-27

PR #24085, opened 2022-03-30
CarInfo: Add harness information

sshane:
Notes:
- Fixed and added Niro EV harnesses: 2020: https://github.com/commaai/openpilot/pull/2416, 2022: https://github.com/commaai/openpilot/pull/23253


selfdrive/car/chrysler/values.py
```diff
@@ -19,16 +21,22 @@ class CAR:
...
   PACIFICA_2020 = "CHRYSLER PACIFICA 2020"
   JEEP_CHEROKEE = "JEEP GRAND CHEROKEE V6 2018"  # includes 2017 Trailhawk
-  JEEP_CHEROKEE_2019 = "JEEP GRAND CHEROKEE 2019" # includes 2020 Trailhawk
+  JEEP_CHEROKEE_2019 = "JEEP GRAND CHEROKEE 2019"  # includes 2020 Trailhawk
+
+
+@dataclass
+class ChryslerCarInfo(CarInfo):
+  package: str = "Adaptive Cruise"
+  harness: str = "fca"
```
adeebshihadeh, line 30: let's use enums here and the website can use the same definitions

sshane: A global harness enum in car_definitions?

sshane: In that case, we can remove some extra imports and code to fetch the footnotes and actually define an order if we put the footnotes in a global enum again

adeebshihadeh: yeah, think global enum makes the most sense

sshane:
Some platforms of HKG have different car harnesses by year, do we want to show the same car split across multiple lines with nothing different except year, or automatically detect duplicates and parse the years back to what they were before?

We can also change the structuring a bit and make CarInfo have a list of cars that map year to harness

sshane:
will merge once tests pass

selfdrive/car/docs_definitions.py
```diff
@@ -50,6 +50,7 @@ class CarInfo:
   min_steer_speed: Optional[float] = None
   min_enable_speed: Optional[float] = None
   good_torque: bool = False
+  harness: Optional[Enum] = None
```
adeebshihadeh, line 53: don't think you want to set a default here. it shouldn't be easy to forget to specify the harness. defaults make sense for the make subclasses though

sshane: for the makes that specify a default package, we'd need to specify the default harness to also be None, like in Hyundai and Honda, but I can do this. defeats the purpose in those two files

selfdrive/car/gm/values.py
```diff
@@ -62,14 +63,20 @@ class Footnote(Enum):
...
+
+
+CAR_INFO: Dict[str, Union[GMCarInfo, List[GMCarInfo]]] = {
+  CAR.HOLDEN_ASTRA: GMCarInfo("Holden Astra 2017", harness=Harness.custom),
+  CAR.VOLT: GMCarInfo("Chevrolet Volt 2017-18", footnotes=[Footnote.OBD_II], min_enable_speed=0, harness=Harness.custom),
+  CAR.CADILLAC_ATS: GMCarInfo("Cadillac ATS Premium Performance 2018"),
+  CAR.MALIBU: GMCarInfo("Chevrolet Malibu Premier 2017", harness=Harness.custom),
+  CAR.ACADIA: GMCarInfo("GMC Acadia 2018", video_link="https://www.youtube.com/watch?v=0ZN6DdsBUZo", footnotes=[Footnote.OBD_II]),
+  CAR.BUICK_REGAL: GMCarInfo("Buick Regal Essence 2018"),
+  CAR.ESCALADE_ESV: GMCarInfo("Cadillac Escalade ESV 2016", "ACC + LKAS", footnotes=[Footnote.OBD_II]),
```
sshane, line 79: Excluding adding harnesses here as you need two harnesses, so I thought it'd be better if the user read the compatibility page and selected the OBD-II harness and not just see their car and buy without knowing it needs more hardware.

PR #24086, opened 2022-03-30
CARS.md: sort cars in lowercase

sshane:
locked with `pipenv lock --keep-outdated`


jyoung8607:
Škoda is still at the bottom of the Silver list, should be higher.

In case it comes up again, Š does fall after S. We checked the Czech orthography with some native speakers when Subaru got moved above it in an earlier version of this doc.

sshane:
That's difficult to do as python is just sorting by the unicode values of the characters which is correct. We can either hard code it, or include the natsort library which does this nicely for us

jyoung8607:
Škoda gang appreciates your support!

adeebshihadeh (approved):
Next time add the dependency in its own PR

PR #24091, opened 2022-03-31
longitudinal MPC: use reset() function instead of recreating the solver in 

FreyJo:
When the `reset()` function in the longitudinal MPC is called, it is very slow, since the solver is always recreated.
I added a `reset()` functionality in acados and changed the longitudinal MPC in openpilot to use that.

A quick timing comparison of create and reset can be found here:
https://github.com/acados/acados/blob/03af92839caffc5efb4a920a09620dcd2d7cc155/examples/acados_python/timing_example/reset_timing.py
```
# create_time: min 0.2189 ms mean 0.2586 ms max 0.6881 ms over 10000 executions
# reset_time: min 0.0036 ms mean 0.0047 ms max 0.0906 ms over 10000 executions
```

Most relevant changes in acados:
- Resetting 
- I realized that the NaN reset test in openpilot was still failing after adding a `reset()` function to the acados OCP solver https://github.com/acados/acados/pull/796 , while for the one I had added in acados this reset function was enough
- Resetting also the HPIPM memory (implemented here: https://github.com/acados/acados/pull/803) makes the openpilot NaN reset test pass.

I guess all that is left to do is updating the binaries for Comma 2 and Comma 3 devices.


FreyJo:
I found and fixed one more thing related to this:
https://github.com/acados/acados/pull/806
Now it is ready for updating binaries and merging.

adeebshihadeh:
Thanks @FreyJo!

pd0wm:
@adeebshihadeh getting errors when building on Mac. Did you forget to rebuild the macOS binaries?

```
Undefined symbols for architecture arm64:
  "_ocp_nlp_solver_reset_qp_memory", referenced from:
      _long_acados_reset in acados_solver_long.os
ld: symbol(s) not found for architecture arm64
```

PR #24093, opened 2022-03-31
camera_qcom2: target 40ms acquisition time, use register wizard

pd0wm:
Target target 40ms acquisition time (25 fps in register wizard) to leave some slack and timing errors in sync signal from panda

Even though only 2 exposure are used, `FINE_INTEGRATION_TIME3` and `FINE_INTEGRATION_TIME4` are still relavant. If they are not set correctly you will get major artefacts.


selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -64,7 +64,7 @@ const int ANALOG_GAIN_REC_IDX = 0x6; // 0.8x
 const int ANALOG_GAIN_MAX_IDX = 0xD; // 4.0x
 
 const int EXPOSURE_TIME_MIN = 2; // with HDR, fastest ss
-const int EXPOSURE_TIME_MAX = 1904; // with HDR, slowest ss
```
pd0wm, line 67: Seems like this is about ~47ms (1904 * 0x7b9) / 80e6

selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -64,7 +64,7 @@ const int ANALOG_GAIN_REC_IDX = 0x6; // 0.8x
...
+const int EXPOSURE_TIME_MAX = 1618; // with HDR, slowest ss, 40ms
```
pd0wm, line 67: Going over the 40ms of the readout doesn't seem to affect anything. 

Going over 50ms causes camera malfunctions and 10fps

PR #24094, opened 2022-03-31
Changes required to build on MacOS 12.3

mattbsea:
**Description** 

Recently set up OpenPilot dev environment on MacOS 12.3.  I encountered a few issues with setup:

1. Python < 3.8.13 won't compile with latest Apple clang 
https://github.com/pyenv/pyenv/issues/2143#issuecomment-1072032647

1.  Perhaps a similar issue in route.cc  Unable to compile using std::array.  Switch to QStringList

1.  Homebrew ffmpeg now installs version 5 by default.  Need to pin it to `ffmpeg@4` in the mac setup script.

**Verification**

After making these changes, I was able to build for MacOS 12.3 and run replay / UI locally.  Also pushed to my comma 3 and was able to engage and drive as usual.


pd0wm:
Is there an issue with ffmpeg 5? I thought I already fixed all the deprecated api calls in this PR https://github.com/commaai/openpilot/pull/23799.

selfdrive/ui/replay/route.cc
```diff
@@ -95,12 +95,13 @@ void Route::addFileToSegment(int n, const QString &file) {
 
 Segment::Segment(int n, const SegmentFile &files, uint32_t flags) : seg_num(n), flags(flags) {
   // [RoadCam, DriverCam, WideRoadCam, log]. fallback to qcamera/qlog
-  const std::array file_list = {
+  const QStringList file_list = {
```
deanlee, line 98: I think we could fix this by  #include <array>

pd0wm:
Let's wait and see if any backports show up for our current python. Changing python versions requires quite a bit of coordination (workstations, servers, agnos/neos).

Added the array include here: https://github.com/commaai/openpilot/commit/b379fca0ae9d6046658d2a11f71462090aadfcbc

PR #24088, opened 2022-03-31
tici downclock

adeebshihadeh:

PR #24097, opened 2022-03-31
put camera/GPU IRQs on core 3

geohot:
Considering there's nothing pinned to core 3 and it's always online, this should be a strict upgrade from core 0. The USB interrupts, WLAN, cdsp are still on core 0.

I'm not sure you want this on the same core as camerad due to priority inversion issues.


adeebshihadeh:
Do these showup before starting camerad? Some of the camera ones don't IIRC. Also did a few more like the encoders in https://github.com/commaai/openpilot/pull/24088.

geohot:
Closed for #24088

PR #24098, opened 2022-03-31
locationd: body support

adeebshihadeh:


selfdrive/car/body/carcontroller.py
```diff
@@ -6,6 +6,7 @@
 MAX_TORQUE = 500
 MAX_TORQUE_RATE = 50
 MAX_ANGLE_ERROR = 7
+INIT_WAIT_COUNTER = 500
```
adeebshihadeh, line 9: use DT_CTRL and specify in seconds
```suggestion
INIT_WAIT_COUNTER = int(5. / DT_CTRL)
```

PR #24099, opened 2022-03-31
openpilot longitudinal: fix long press to increment by 5 mph

sshane:


sshane:
Pending review of https://github.com/commaai/openpilot/pull/24063/files

PR #24100, opened 2022-04-01
controls: pressing set while gas pressed resets to vEgo

### LINKED ISSUE #24046: Support setting to current speed while gas is pressed via set button
*when not pcmCruise and openpilot is handling button presses. Toyota does this nicely

sshane:
- [x] Only when below speed


adeebshihadeh:
Does this work? If so, should be good to merge.

sshane:
Will get a few testers (tester)

PeteFromGlasgow:
DongleID: `549bb645afc16a95`
Segment ID: `2022-04-07--12-40-23--7` and `2022-04-07--12-40-23--8`

Gave this a try on my Niro EV 2021 and it worked well. There is a minor deviation from stock SCC behaviour. In stock - if the car is going at a higher speed than the set cruise control speed, pressing the set button will always set the speed to the speed the car is going even if the accelerator is not pressed.

sshane:
Thanks for testing! In those two segments, you were always pressing the gas when you increased the set speed, was there another route where it reset to your current speed when not pressing the gas?

tecandrew:
cruise fault occured when testing, but might be caused by something unrelated to these changes :thinking: `eb5a50f9e98f50b9|2022-04-07--15-14-34--1`

also just noticed that speed can't be initially set if starting from 0 mph. set speed defaults to the standard stock SCC lower limit (20 MPH in my case). however, it can be lowered than that if done manually by up/down cruise buttons

PeteFromGlasgow:
@sshane I think I have done that somewhere else in the same route `2022-04-07--12-40-23`, I was also using the fork in another route `2022-04-07--13-13-51`. I'm not if I could pinpoint what segments I used it in.

selfdrive/controls/lib/drive_helpers.py
```diff
@@ -65,6 +65,10 @@ def update_v_cruise(v_cruise_kph, buttonEvents, button_timers, enabled, metric):
         long_press = True
         break
 
+  # If set is pressed while overriding, reset cruise speed to vEgo
+  if enabled and gas_pressed and button_type == car.CarState.ButtonEvent.Type.decelCruise:
```
gijskoning, line 69: Enabled is always true. See line 47
```suggestion
  if gas_pressed and button_type == car.CarState.ButtonEvent.Type.decelCruise:
```

selfdrive/controls/lib/drive_helpers.py
```diff
@@ -65,6 +65,10 @@ def update_v_cruise(v_cruise_kph, buttonEvents, button_timers, enabled, metric):
...
+    return int(round(clip(v_ego * CV.MS_TO_KPH, V_CRUISE_ENABLE_MIN, V_CRUISE_MAX)))
```
gijskoning, line 70: Round() to whole numbers already returns an integerer
```suggestion
    return round(clip(v_ego * CV.MS_TO_KPH, V_CRUISE_ENABLE_MIN, V_CRUISE_MAX))
```

pd0wm: It's a leftover from the python 2 days where it returned a float.

PR #24107, opened 2022-04-02
controlsd: add state machine tests

sshane:
Also handles noEntry events in preEnabled, assuming all softDisable events have a noEntry event type, which should be asserted by a test somewhere

Before this change, if you were in a preEnabled state and caused a SOFT_DISABLE, like taking your seatbelt off, openpilot would stay engaged with no alerts at all. Now we consider noEntry.

### DIFF
--- a/selfdrive/controls/tests/test_state_machine.py
+++ b/selfdrive/controls/tests/test_state_machine.py
@@ -51,6 +51,15 @@ class TestStateMachine(unittest.TestCase):
       self.controlsd.state_transition(self.CS)
       self.assertEqual(State.disabled, self.controlsd.state)
 
+  def test_soft_disable(self):
+    for state in ALL_STATES:
+      if state == State.preEnabled:  # preEnabled considers NO_ENTRY instead
+        continue
+      self.controlsd.events.et = [MAINTAIN_STATES[state], ET.SOFT_DISABLE]
+      self.controlsd.state = state
+      self.controlsd.state_transition(self.CS)
+      self.assertEqual(State.disabled, self.controlsd.state)
+
   def test_no_entry(self):
     # Disabled
     for et in ENABLE_EVENT_TYPES:
@@ -61,7 +70,7 @@ class TestStateMachine(unittest.TestCase):
     # preEnabled also should consider NO_ENTRY
     self.setUp()
     self.controlsd.state = State.preEnabled
-    self.controlsd.events.et = [ET.NO_ENTRY]
+    self.controlsd.events.et = [ET.NO_ENTRY, ET.PRE_ENABLE]
     self.controlsd.state_transition(self.CS)
     self.assertEqual(self.controlsd.state, State.disabled)
 
--- a/selfdrive/controls/tests/test_state_machine.py
+++ b/selfdrive/controls/tests/test_state_machine.py
@@ -70,7 +70,7 @@ class TestStateMachine(unittest.TestCase):
     self.controlsd.state = State.enabled
     self.controlsd.events.add(make_event([ET.SOFT_DISABLE]))
     self.controlsd.state_transition(self.CS)
-    while self.controlsd.soft_disable_timer > 0:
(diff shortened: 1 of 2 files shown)


selfdrive/controls/tests/test_state_machine.py
```diff
@@ -0,0 +1,98 @@
...
+
+# The event types that maintain the current state
+MAINTAIN_STATES = {State.enabled: None, State.disabled: None, State.softDisabling: ET.SOFT_DISABLE,
+                   State.preEnabled: ET.PRE_ENABLE}
+ALL_STATES = (state for state in State.schema.enumerants.values() if state != State.overriding)  # TODO: remove overriding exception
+# The event types checked in DISABLED section of state machine
+ENABLE_EVENT_TYPES = (ET.ENABLE, ET.PRE_ENABLE)
+
+
+class Events:
```
adeebshihadeh, line 20: why can't this use the normal Events class?

adeebshihadeh, line 20: why can't this use the normal Events class?

selfdrive/controls/tests/test_state_machine.py
```diff
@@ -0,0 +1,105 @@
...
+    CI = CarInterface(CP, CarController, CarState)
+
+    self.controlsd = Controls(CI=CI)
+    self.controlsd.events = Events()
+    self.controlsd.soft_disable_timer = int(SOFT_DISABLE_TIME / DT_CTRL)
+    self.CS = car.CarState()
+
+  def test_immediate_disable(self):
+    for state in ALL_STATES:
+      self.controlsd.events.add(make_event([MAINTAIN_STATES[state], ET.IMMEDIATE_DISABLE]))
```
sshane, line 45: Have to modify the global EVENTS dict to make the Events class use our custom alert. Works, however if you comment any lines like this, then re-run the test, it's passing until you comment every _other_ test for some reason.

adeebshihadeh: check out `unittest.mock.patch` https://github.com/commaai/openpilot/blob/b2b6205e252b5291a953dd94c5524715b1f8f444/selfdrive/thermald/tests/test_power_monitoring.py#L3

sshane: Ah my issue was because wrapping a list comprehension with parenthesis yields a generator, not a tuple as I thought, so it was only iterating through ALL_STATES only in the first test.

selfdrive/controls/tests/test_state_machine.py
```diff
@@ -0,0 +1,105 @@
...
+    for state in ALL_STATES:
+      if state == State.preEnabled:  # preEnabled considers NO_ENTRY instead
+        continue
+      self.controlsd.events.add(make_event([MAINTAIN_STATES[state], ET.SOFT_DISABLE]))
+      self.controlsd.state = state
+      self.controlsd.state_transition(self.CS)
+      self.assertEqual(self.controlsd.state, State.disabled if state == State.disabled else State.softDisabling)
+      self.controlsd.events.clear()
+
+    # Test soft disable timer
```
adeebshihadeh, line 69: make a new test case for this. don't call setUp() manually

selfdrive/controls/tests/test_state_machine.py
```diff
@@ -0,0 +1,105 @@
...
+
+  def test_no_entry(self):
+    # disabled with enable events
+    for et in ENABLE_EVENT_TYPES:
+      self.controlsd.events.add(make_event([ET.NO_ENTRY, et]))
+      self.controlsd.state_transition(self.CS)
+      self.assertEqual(self.controlsd.state, State.disabled)
+      self.controlsd.events.clear()
+
+    # preEnabled with preEnabled event
```
adeebshihadeh, line 88: split this out too

sshane:
Tests failed as expected when commenting out various state assignment lines in state_transition (`self.state = State.disabled`)

PR #24108, opened 2022-04-02
Hyundai: remove 90° steering lockout

sshane:
@gregjhogan and I found we can cut torque for two frames before a fault is imminent without any negative effects to steering. Two is required, as we're very slightly out of sync with the EPS's 100hz thread and it [randomly drops our no steer request messages multiple times in a row](https://imgur.com/a/FmU92mH) if we only send one.

Suspected EPS fault counters from Greg:
- 90 degree limit: 95 frames above limit. openpilot will use an angle limit of 85 degrees, and a fault counter of 90 to allow some tolerance.
- Sending torque with `CF_Lkas_ActToi` set to 0: Sonata supposedly only allows 4 frames, Palisade supposedly allows 10 frames. openpilot only needs two to reset 90 degree limit counter

TODO:
- [x] ~Understand what role `CF_Lkas_ToiFlt` has here~. If you don't set this fault bit, we drop torque immediately and the outcome is very similar to an EPS-caused steering fault—a clicking noise and wheel jitter. With it, no noticeable wheel movements or sounds occur. 

Related: Write panda safety to ensure we can never wind up torque with `CF_Lkas_ActToi` bit set to 0 and then set to 1 to effectively bypass panda torque rate limits, while also allowing us to disable `CF_Lkas_ActToi` for two frames while controls allowed.

Applying 255 torque continually for over 10 minutes with no faults at >400 degrees: https://imgur.com/a/1yVAX81


gregjhogan:
The thread in the firmware that controls electronic steering is separate from what reads and parses the CAN messages (the snapshot of the CAN data is passed through global variables).  If we send one message with steering disabled and then the next message with it enabled too close together, the thread controlling electronic steering may never see that we disabled steering for one frame, and the fault counter for being over 90 deg never resets.

I also found in the firmware the fault injection bit `CF_Lkas_ToiFlt` in the same message when set causes wind down from the last request value to zero.  This wind down first holds the last requested torque for two frames, so we can use this to prevent loss of torque while resetting the fault counter for being over 90 deg.

Another approach for ensuring the counter gets reset would be to set `LKAS11.CF_Lkas_ActToi = 0` and `LKAS11.CF_Lkas_ToiFlt = 1` until we see `MDPS12.CF_Mdps_ToiFlt == 1`.  The current approach is susceptible to a bunch of `LKAS11` CAN messages getting flushed all at once due to lagging or other things happening on the CAN bus, but this approach may be slower than we want and steering torque will start to wind down.

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -10,6 +10,12 @@
 VisualAlert = car.CarControl.HUDControl.VisualAlert
 LongCtrlState = car.CarControl.Actuators.LongControlState
 
+# EPS faults if you apply torque while the steering angle is above 90 degrees for more than 1 second
+# All slightly below EPS thresholds to avoid fault
+ANGLE_FAULT_MAX_ANGLE = 85
+ANGLE_FAULT_MAX_FRAMES = 89
+ANGLE_FAULT_CONSECUTIVE_FRAMES = 2
```
adeebshihadeh, line 17: i'd prefix these with `MAX_ANGLE`

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -107,10 +114,23 @@ def update(self, CC, CS):
...
-                                     left_lane_warning, right_lane_warning))
+      # Count up to ANGLE_FAULT_MAX_FRAMES, at which point we need to cut torque to avoid a steering fault
+      if CC.latActive and abs(CS.out.steeringAngleDeg) >= ANGLE_FAULT_MAX_ANGLE:
+        self.angle_limit_counter += 1
+      else:
+        self.angle_limit_counter = 0
+
+      # Cut steer actuation bit for two frames and hold torque with induced temporary fault
+      lat_active = CC.latActive and self.angle_limit_counter <= ANGLE_FAULT_MAX_FRAMES
+      torque_fault = CC.latActive and self.angle_limit_counter > ANGLE_FAULT_MAX_FRAMES
```
adeebshihadeh, line 125:
```suggestion
      torque_fault = CC.latActive and self.angle_limit_counter > ANGLE_FAULT_MAX_FRAMES
      lat_active = CC.latActive and not torque_fault
      
```

adeebshihadeh (approved):

PR #24109, opened 2022-04-02
Optima 2019: allow max torque of 384

sshane:
Optima can apply 384, but faults after 105 seconds

Route: `f15e3c37c118e841|2022-04-02--02-26-05`

Screenshot from 2022-04-02 03-20-42


adeebshihadeh:
@sshane I suspect all can be moved here. Can you work on getting the rest moved over?

sshane:
I'll merge this for now as we know it works, then work on the rest in this PR: https://github.com/commaai/openpilot/pull/24117

PR #24113, opened 2022-04-04
sim: write params to enable in Dockerfile

adeebshihadeh:
The params are written in bridge, but aren't picked up if manager is started first.


ebadi:
I tested this and it works and I therefore close this issue. Thanks @adeebshihadeh. 
p.s. Are you interested if I write a simple unit-test to ensure openpilot can drive inside Carla?  If so, any preference what has to be tested?

PR #24116, opened 2022-04-04
car docs: enforce naming conventions

sshane:
- PHEV should be Plug-in Hybrid
- HEV should be Hybrid
- EV should be Electric


selfdrive/car/tests/test_docs.py
```diff
@@ -17,17 +17,16 @@ def test_generator(self):
...
       for car in cars:
-        tokens = car.model.lower().split(" ")
-        self.assertNotIn("phev", tokens, "Use `Plug-In Hybrid`")
-        self.assertNotIn("hev", tokens, "Use `Hybrid`")
-        self.assertNotIn("ev", tokens, "Use `Electric`")
-        if "plug-in hybrid" in car.model.lower():
-          self.assertIn("Plug-In Hybrid", car.model, "Use correct capitalization")
-        elif "plug-in" in car.model.lower():
-          self.assertIn("Plug-In", car.model, "Use correct capitalization")
+        if car.make in ["Hyundai", "Kia"]:
```
adeebshihadeh, line 23: use carParams.carName for this. in this case, genesis is missing

PR #24118, opened 2022-04-04
Subaru: update carinfo supported model years and good steering torque

martinl:
This PR updates Subaru supported model years (included in initial Subaru FPv2 import) and adds stars for good steering torque

**Checklist**


sshane:
In the future we are going to use data to generate the torque star columns, but I'm fine with manually updating this for now, as long as it has enough torque for comfortable highway driving.

sshane:
Actually can you revert the torque changes and just update the model years, we saw from data that the Impreza has about the same torque as some Hondas. We'll use data from the fleet to fill the torque column out later.

michaelhonan:
Might be worth noting that the Imprezas might not be as limited as previously thought:
https://github.com/commaai/openpilot/pull/23530

sshane:
Yeah, will look into that eventually. For now though, let's use the data for the good torque columns, so it's only Forester for now.

martinl:
Which model year Impreza steering torque did you measure - 2020-21 has lower steering torque limit in openpilot because of eps fault issue but Impreza 2016-19 and Crosstrek 2017-19 should have no issues on highway driving. Would it be ok to split Impreza and Crosstrek and remove steering torque star from 20-21?

sshane:
We looked at Impreza Sport 2020, which had 0.66, 0.71, 0.96 m/s/s lateral acceleration at three speed brake points while Forester had 2.01, 1.67, NA m/s/s

ClockeNessMnstr:
> Which model year Impreza steering torque did you measure - 2020-21 has lower steering torque limit in openpilot because of eps fault issue but Impreza 2016-19 and Crosstrek 2017-19 should have no issues on highway driving. Would it be ok to split Impreza and Crosstrek and remove steering torque star from 20-21?

20-21 look to have more torque not less.

PR #24120, opened 2022-04-04
body: add crc and counter

briskspirit:
Add crc and counter to important CAN messages.


selfdrive/car/body/bodycan.py
```diff
@@ -1,8 +1,18 @@
...
+from selfdrive.car import crc8_pedal
+
+
+def create_control(packer, torque_l, torque_r, idx):
   can_bus = 0
 
   values = {
     "TORQUE_L": torque_l,
     "TORQUE_R": torque_r,
+    "COUNTER": idx & 0xF,
```
adeebshihadeh, line 10: packer should set this

briskspirit: Why do we do the same way for pedal then? in create_gas_interceptor_command()

briskspirit: Also because of the signal name?

adeebshihadeh: not sure, sounds likely though

PR #24128, opened 2022-04-05
body: add gateway ECU for git hash logging

briskspirit:
add gateway ECU for git hash logging


pd0wm:
@briskspirit george wanted a version number instead. Gating on the git hash is annoying since you need to keep updating it while testing.

briskspirit:
@pd0wm that was Adeeb's request to log git hash with non-essential ECU. I can add version to it, but that will be similar to panda - we will forget to bump it and eventually will remove. I will add version to another request similar to what I have now:

*** Results for address 0x720 ***


0xF181 APPLICATION_SOFTWARE_IDENTIFICATION: b'02/27/2022'
0xF18C ECU_SERIAL_NUMBER: b'\x16\x00\t\x80\x0fQ6086'
0xF190 VIN: b'COMMABODYVERSION1'
0xF197 SYSTEM_NAME_OR_ENGINE_TYPE: b'ELECTRIC'


*** Results for address 0x721 ***


0xF181 APPLICATION_SOFTWARE_IDENTIFICATION: b'166bd860'

adeebshihadeh:
Let's make a new ECU type called debug for this purpose and explicitly ignore it for fingerprinting.

pd0wm:
Also skip `debug` ecu's in `match_fw_to_car_exact`

selfdrive/car/body/values.py
```diff
@@ -26,6 +26,9 @@ class CAR:
     (Ecu.engine, 0x720, None): [
       b'02/27/2022'
     ],
+    (Ecu.debug, 0x721, None): [
+      b'166bd860' # git hash of the firmware used
```
pd0wm, line 30: Do we need to specify any versions for it to work? Or can we leave an empty list?

PR #24132, opened 2022-04-05
UI: improved disengage on gas toggle description

spektor56:
Should be depressed as standard in automotive industry.

Pressed implies state change from down -> up, depressed means the state is down

### DIFF
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -68,7 +68,7 @@ TogglesPanel::TogglesPanel(SettingsWindow *parent) : ListWidget(parent) {
     {
       "DisengageOnAccelerator",
       "Disengage On Accelerator Pedal",
-      "When enabled, openpilot will disengage when the accelerator pedal is depressed.",
+      "When enabled, pressing the accelerator pedal will disengage openpilot.",
       "../assets/offroad/icon_disengage_on_accelerator.svg",
     },
 #ifdef ENABLE_MAPS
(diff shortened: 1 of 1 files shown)


pd0wm:
The disengagement happens when the state changes from not pressed to pressed, so not sure I understand why it's wrong.

gregjhogan:
press is an action, depress is a state, and it seems logical to me this sentence tells you what action causes disengagement

spektor56:
> The disengagement happens when the state changes from not pressed to pressed, so not sure I understand why it's wrong.

you mean from not pressed to depressed.  If you "pressed" a key on a keyboard most people would infer that you are no longer pressing the key.  If a key is "depressed" it means it's currently in the down state (which is the only state that this acceleration override is actually active in).

You can see in the Honda owners manual every time they refer to the accelerator or brake pedal they use the terms: depress, depressing, and depressed: http://techinfo.honda.com/rjanisis/pubs/OM/AH/ATLA1717OM/enu/ATLA1717OM.PDF

when they refer to buttons they use "press"

pd0wm:
How about "When enabled, pressing the accelerator pedal will disengage openpilot"? That should get rid of the ambiguity.

selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -68,7 +68,7 @@ TogglesPanel::TogglesPanel(SettingsWindow *parent) : ListWidget(parent) {
     {
       "DisengageOnAccelerator",
       "Disengage On Accelerator Pedal",
-      "When enabled, openpilot will disengage when the accelerator pedal is pressed.",
+      "When enabled, openpilot will disengage when the accelerator pedal is depressed.",
```
adeebshihadeh, line 71:
```suggestion
      "When enabled, pressing the accelerator pedal will disengage openpilot.",
```

PR #24140, opened 2022-04-06
HKG: only allow engagement on user button press

### LINKED ISSUE #23674: Hyundai Elantra Hybrid: disengage failure on gas

sshane:
On rising edge of `carState.cruistState.enabled`, if no user button press is detected in the past 3 frames (about the max from button press to engage), then we don't add the pcmEnable event. controlsd then sets `carControl.cruiseControl.cancel`, cancelling the car's cruise control.

panda TODO:
- [x] panda's `sample_t` struct only supports up to 6 values for torque measurements, but we need 8 openpilot frames/80 ms. However panda only updates measurements on new messages, so re-run the notebook and only update when we have a new CLU11 message (using `CANParser.vl_all`). Should only need 4 values as the message is 50Hz and we do the same thing panda does, so reduce chances for mismatches between openpilot and panda drastically

openpilot TODO:
- [ ] should we only gate on known cars with the pause/resume button? not sure it'll be 100% accurate as some manuals weren't accurate when I checked last time


sshane:
Looked at 6,276 engages from various HKG cars to see how far back we need to consider to avoid both:
- not detecting that a user had the intention to engage (and end up not engaging),
- and avoid allowing the car to re-engage if the gas pedal and cancel button are pressed close together. Car would accidentally re-engage and if we recently saw the pause/resume button (to cancel) we'd think the user meant to re-engage.

Lookback frames are the number of 100Hz cycles we look back on to check if the user caused the engage from a button press, so 1 would only consider the latest button message.

Ideally we only consider the latest message, but apparently that means 1 in every 50 engagements would be essentially ignored, and it only ~doubles as you increase the lookback frame count. Going to re-run on only updating the list when an update from `CLU11` is seen, rather than every 100 ms.

```python
Lookback frames: 1
Total engages: 6276, users caused: 6151 (1.99% not caught)
Missed: 125 total, would falsely cancel every 50.208 user engages
-----
Lookback frames: 2
Total engages: 6276, users caused: 6214 (0.99% not caught)
Missed: 62 total, would falsely cancel every 101.226 user engages
-----
Lookback frames: 3
Total engages: 6276, users caused: 6240 (0.57% not caught)
Missed: 36 total, would falsely cancel every 174.333 user engages
-----
Lookback frames: 4
Total engages: 6276, users caused: 6259 (0.27% not caught)
Missed: 17 total, would falsely cancel every 369.176 user engages
-----
Lookback frames: 5
Total engages: 6276, users caused: 6265 (0.18% not caught)
Missed: 11 total, would falsely cancel every 570.545 user engages
-----
Lookback frames: 6
Total engages: 6276, users caused: 6272 (0.06% not caught)
Missed: 4 total, would falsely cancel every 1569.0 user engages
-----
```

sshane:
Ran on the last 180 days for more accuracy. These are the number of frames back from the current we need to look to allow an engagement using button presses:

```
Platform: All cars
Lookback frames: 1, occurrences: 13348
Lookback frames: 2, occurrences: 128
Lookback frames: 3, occurrences: 59
Lookback frames: 4, occurrences: 39
Lookback frames: 5, occurrences: 6
Lookback frames: 6, occurrences: 7
Lookback frames: 7, occurrences: 3
Lookback frames: 8, occurrences: 1
```

Per-car: https://pastebin.com/wYKkSgDm

The Ioniqs and the Kia Niro EV have a disproportionate amount of engagements with frame counts > ~3, so it's best to use 6-8 so we don't disallow engagement noticeably often.

sshane:
The only two cars at 6 frames that we don't catch, this seems pretty high and should be good:

```python
Platform: HYUNDAI IONIQ ELECTRIC LIMITED 2019
Lookback frames: 6
Total engages: 656, users caused: 655 (0.15% not caught)
Missed: 1 total, would falsely cancel every 656.0 user engages

---
Platform: HYUNDAI IONIQ ELECTRIC 2020
Lookback frames: 6
Total engages: 1586, users caused: 1583 (0.19% not caught)
Missed: 3 total, would falsely cancel every 528.667 user engages
```

sshane:
Using 8 frames, no user-engagement isn't detected properly across 13,591 engages with the buttons from 14 cars we have data for.

sshane:
panda TODO:
- [ ] panda's `sample_t` struct only supports up to 6 values for torque measurements, but we need 8 openpilot frames/80 ms. However panda only updates on new messages, so re-run the notebook and only update when we have a new CLU11 message. Should only need 4 values as the message is 50Hz and we do the same thing panda does, so reduce chances for mismatches drastically

sshane:
Ran one more time on a year of rlogs available (66,072 engagements) across {} different cars. If we process in the same way panda does (updates sliding window only when a CLU11 button message is received), then we only ever need four samples (approx 8 openpilot frames/80ms) to know if the user caused the engage via buttons. Again, a lookback of 1 means we only need to look at the latest CLU11 message received to know if the button was pressed or not for this rising edge of PCM cruise.

```python
Platform: All cars
Lookback samples: 1
Total engages: 16518, users caused: 16187 (2.0% not caught)
Missed: 331 total, would falsely cancel every 49.903 user engages
---
Lookback samples: 2
Total engages: 16518, users caused: 16439 (0.48% not caught)
Missed: 79 total, would falsely cancel every 209.089 user engages
---
Lookback samples: 3
Total engages: 16518, users caused: 16503 (0.09% not caught)
Missed: 15 total, would falsely cancel every 1101.2 user engages
---
Lookback samples: 4
Total engages: 16518, users caused: 16518 (0.0% not caught)
Missed: 0 total, would falsely cancel every N/A user engages
---
All cars: Lookback samples: 1, occurrences: 64748
All cars: Lookback samples: 2, occurrences: 1008
All cars: Lookback samples: 3, occurrences: 256
All cars: Lookback samples: 4, occurrences: 60
```

sshane:
A few plots of this PR in action on real accidental re-engagements: https://imgur.com/a/SiCZPWc

sshane:
Still need prev_cruise_buttons because if we don't get an update from CLU11 right after falling edge, then we can end up sending two falling edge button events:

```python
deque([1.0, 1.0, 1.0, 1.0], maxlen=4)
deque([1.0, 1.0, 1.0, 1.0], maxlen=4)
deque([1.0, 1.0, 1.0, 0.0], maxlen=4)  # sent enable event here (panda updates controlsAllowed here)
deque([1.0, 1.0, 1.0, 0.0], maxlen=4)  # and here
deque([1.0, 1.0, 0.0, 0.0], maxlen=4)
deque([1.0, 1.0, 0.0, 0.0], maxlen=4)
deque([1.0, 0.0, 0.0, 0.0], maxlen=4)
deque([1.0, 0.0, 0.0, 0.0], maxlen=4)
deque([0.0, 0.0, 0.0, 0.0], maxlen=4)
```

Could be fixed by stepping through each CAN message in CarState, or just doing it for the button events here.

sshane:
Ran a replay of CarState (through test models) on this amount of data:

```python
total_segments=674
total_gas_disengagements=1548, total_disengagements=2252
total_engagements=2252
reengage_times=[4.1266198, 4.0869068, 3.8345807, 1.905216]  # frames between disengage and engagement
```

So only 4 accidental re-engagements in this data. Checked to make sure that the code here catches these four occurrences, and nothing else, and in all 674 segments there weren't any extra false positives it caught.

selfdrive/car/hyundai/carstate.py
```diff
@@ -107,9 +113,10 @@ def update(self, cp, cp_cam):
     self.lkas11 = copy.copy(cp_cam.vl["LKAS11"])
     self.clu11 = copy.copy(cp.vl["CLU11"])
     self.steer_state = cp.vl["MDPS12"]["CF_Mdps_ToiActive"]  # 0 NOT ACTIVE, 1 ACTIVE
-    self.brake_error = cp.vl["TCS13"]["ACCEnable"] != 0 # 0 ACC CONTROL ENABLED, 1-3 ACC CONTROL DISABLED
-    self.prev_cruise_buttons = self.cruise_buttons
-    self.cruise_buttons = cp.vl["CLU11"]["CF_Clu_CruiseSwState"]
+    self.brake_error = cp.vl["TCS13"]["ACCEnable"] != 0  # 0 ACC CONTROL ENABLED, 1-3 ACC CONTROL DISABLED
+    self.prev_cruise_buttons = self.cruise_buttons[-1]
+    self.cruise_buttons += cp.vl_all["CLU11"]["CF_Clu_CruiseSwState"]
+    self.main_buttons += cp.vl_all["CLU11"]["CF_Clu_CruiseSwMain"]
```
adeebshihadeh, line 119: unclear that this is list-like, use `extend()` instead

selfdrive/car/hyundai/interface.py
```diff
@@ -315,20 +317,24 @@ def _update(self, c):
     ret = self.CS.update(self.cp, self.cp_cam)
     ret.steeringRateLimited = self.CC.steer_rate_limited if self.CC is not None else False
 
-    events = self.create_common_events(ret, pcm_enable=self.CS.CP.pcmCruise)
+    # On some newer model years, the CANCEL button acts as a pause/resume button based on the PCM state
+    # To avoid re-engaging when openpilot cancels, check user engagement intention via buttons
+    # Main button also can trigger an engagement on these cars
+    allow_enable = any([btn in ENABLE_BUTTONS for btn in self.CS.cruise_buttons]) or any(self.CS.main_buttons)
```
adeebshihadeh, line 323: doesn't matter much in this case, but the list comprehension makes this slower

selfdrive/car/interfaces.py
```diff
@@ -175,8 +175,9 @@ def create_common_events(self, cs_out, extra_gears=None, pcm_enable=True):
       events.add(EventName.steerUnavailable)
 
     # we engage when pcm is active (rising edge)
+    # and if enabling is allowed
```
adeebshihadeh, line 178:
```suggestion
    # we engage when pcm is active (rising edge)
    # enabling can optionally be blocked by the car interface
```

PR #24145, opened 2022-04-06
body: add battery percentage and charger status to signals

briskspirit:


adeebshihadeh:
This only exposes the signals in the CANParser. Where do you want to write them? New fields in CarState?

PR #24157, opened 2022-04-07
camerad: cleanups, support running only one camera

geohot:


pd0wm:
If any of the asserts in camera bringup or AE fails, can we disable that camera and continue with the rest?

geohot:
Yes, but from an application perspective, why would you want that?

pd0wm:
> Yes, but from an application perspective, why would you want that?
> 
> (and actually, sort of not with the strobe conflict)

 - For startup failures it makes it easier for controlsd to alert properly, and to do other debugging
 - I a camera fails while driving we want to soft-disable but keep driving for as long as possible. Our driving models should be robust to a camera dropping out.

geohot:
Hmm actually being robust to the failure we've seen in the field is easy, and will also work for an unplugged sensor.

pd0wm:
There are a few more asserts I think. At least one in the i2c code that's called by auto exposure.

geohot:
Done. Confirmed both:
* starting without sensor attached
* unplugging sensor while running

geohot:
Tried a few more sensor unplugs, you can also sometimes get a kernel panic...but 80% of the time it's i2c

geohot:
Should be good.

geohot:
Hmm, handle_camera_event has some sketchy looking code to handle failure cases. I wonder how it was tested. I added logging to it.

PR #24160, opened 2022-04-07
Make safetyParam unsigned

sshane:
For proper flag support, also doubles the size


sshane:
Need to regen the route before merging

selfdrive/test/process_replay/regen.py
```diff
@@ -217,6 +253,14 @@ def regen_segment(lr, frs=None, outdir=FAKEDATA):
     'camerad': [
       *cam_procs,
     ],
+    'controlsd': [
```
adeebshihadeh, line 256: controlsd is supposed to run live. the only things that should be replayed are hw: CAN, cameras, and sensors

sshane:
Can't regenerate segment until https://github.com/commaai/openpilot/issues/23666 is fixed.

PR #24175, opened 2022-04-09
comma three: correctly set initial EPS bearer settings for AT&T sim cards

adeebshihadeh:


pd0wm:
- Is the modem online at this point? Try setting prebuilt.
- We can check the sim id and only do the changes for AT&T sims.

PR #24177, opened 2022-04-10
loggerd: broadcast encoder data

geohot:
Should be good. Might add a little time and memory usage.


geohot:
I'm fine with merging this. Two things to note:
1. It makes an extra copy of the compressed data. Shouldn't really matter.
2. It recreates the pubmaster on each rotation. Is this okay?

pd0wm:
2. Shouldn't matter from a performance perspective. All readers that aren't caught up will drop some packets though.

PR #24181, opened 2022-04-11
controlsd: add curvature steering

incognitojam:
The Ford power steering control module (PSCM) has a unique controls method.

The CAN message sent from the LKAS camera that we use to actuate the steering, `LateralMotionControl`, has four important signals: curvature, curvature rate, path offset and path angle. The PSCM integrates this information about the path ahead (presumably along with other information about the vehicle such as speed and orientation) to determine the correct steering angle, and then moves towards it according to some parameters (ramp speed and precision, which are additional signals in the message).

LateralMotionControl message

Sending only curvature and path offset, and zero values for curvature rate and path angle, is enough information to drive. The PSCM knows the curvature of the path and our current offset from it and is able to correct for it.

Ideally though we should be able to pull all four actuators from the model or the lateral plan. Path offset is easy to get from the plan in `dPathPoints`, and I visually verified that it resembled the stock system's values well enough using PlotJuggler (in the image below the sign is flipped).

image

**Related PRs**
- https://github.com/commaai/cereal/pull/303

### DIFF
--- a/selfdrive/car/tests/test_car_interfaces.py
+++ b/selfdrive/car/tests/test_car_interfaces.py
@@ -34,7 +34,7 @@ class TestCarInterfaces(unittest.TestCase):
     self.assertGreater(car_params.mass, 1)
     self.assertGreater(car_params.steerRateCost, 1e-3)
 
-    if car_params.steerControlType != car.CarParams.SteerControlType.angle:
+    if car_params.steerControlType not in (car.CarParams.SteerControlType.angle, car.CarParams.SteerControlType.curvature):
       tuning = car_params.lateralTuning.which()
       if tuning == 'pid':
         self.assertTrue(len(car_params.lateralTuning.pid.kpV))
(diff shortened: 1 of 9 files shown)


incognitojam:
Sending only curvature and path offset, and zero values for curvature rate and path angle, is enough to make the system work. The PSCM knows the curvature of the path and our current offset from it and is able to correct for it.

Ideally though we should be able to pull all four actuators from the model or the lateral plan. Path offset is easy to get from the plan in `dPathPoints`, and I visually verified that it resembled the stock system's values well enough using PlotJuggler (in the image below the sign is flipped).

image

incognitojam:
I've realised that curvature and path offset are fine for driving on the highway, but are nowhere near close to being :taco: ready, as the max curvature we can send is `0.02` which corresponds to about a 50m radius curve (if I understand everything right..)

incognitojam:
I'm trying to calculate the path angle using the path positions output from the model - simple `arctan(-positions.y[10]/positions.x[10])` seems to give results with a similar shape to the stock signal
image
image

PR #24182, opened 2022-04-11
parmasd: update linear region

ClockeNessMnstr:
noticed that the large changes to EKF states for params learner were occuring when releasing the wheel at large steering angles. 

I believe this is outside the "linear-region" so these are invalid readings causing issues with the learned values. 

I originally just changed the "if angle < 45deg ~~or not steerPressed~~*" to just check the angle and not enable at higher angles. 
however I still got large swings in values when the learner re-enabled returning to 3 or 4 deg. 

I added logic to lockout the learner until after the steering returns to center after being outside of the "linear region" 
I also updated it based on steering ratio (from car params) for the vehicle to be ~3deg of tire steering instead of 45deg of steering wheel. as that region differs based on vehicle.

### DIFF
--- a/selfdrive/locationd/paramsd.py
+++ b/selfdrive/locationd/paramsd.py
@@ -29,11 +29,13 @@ class ParamsLearner:
     self.kf.filter.set_global("stiffness_rear", CP.tireStiffnessRear)
 
     self.active = False
+    self.lockout = True
 
     self.speed = 0.0
     self.roll = 0.0
     self.steering_pressed = False
     self.steering_angle = 0.0
+    self.steering_ratio = CP.steerRatio
 
     self.valid = True
 
@@ -84,11 +86,12 @@ class ParamsLearner:
 
     elif which == 'carState':
       self.steering_angle = msg.steeringAngleDeg
-      self.steering_pressed = msg.steeringPressed
       self.speed = msg.vEgo
 
-      in_linear_region = abs(self.steering_angle) < 45 or not self.steering_pressed
-      self.active = self.speed > 5 and in_linear_region
+      in_linear_region = abs(self.steering_angle) < 3*self.steering_ratio
+      self.active = self.speed > 5 and in_linear_region and not self.lockout
+      
+      self.lockout = not (self.active or abs(self.steering_angle) < 0.2*self.steering_ratio)
 
       if self.active:
         self.kf.predict_and_observe(t, ObservationKind.STEER_ANGLE, np.array([[math.radians(msg.steeringAngleDeg)]]))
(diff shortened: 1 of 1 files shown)


nuwandavek:
Hey!
Most of the work for a `localizer` or `params_learner` change (except for obvious bugs) is in validation and making sure the change does not have long-tail regressions. So, more than a PR, it would be extremely useful to have an issue, with a specific undesirable behaviour (along with the segment logs uploaded), and any ideas for improvement. We could discuss the extent of the issue and figure out what the regression analysis might look like.

ClockeNessMnstr:
Stock OP doesn't work on my vehicle but I'm hoping this falls under obvious bugs. 

I noticed that the learner was active at high steering angles like 250 deg. is the learner able to learn valid data outside of the 3deg tire steering? 

The effects of this issue was the sudden jumps in values in the learner.  These seemed nearly entirely mitigated when locking it out until hitting 0 deg.

adeebshihadeh:
We're very interested in fixing bugs, but proposed changes like this need to start with an actual, specific issue they're addressing. This process should start with a high quality bug report. What specifically doesn't work on your car? I checked a couple of your branches, but the diff was massive. Looking into non-stock routes is difficult, and most of the time it ends up being a waste of time.

ClockeNessMnstr:
I could make a more straight forward branch / diff. my personal is comma3 atm
most of the diff is probably martinl/subaru-community. 

This is the biggest issue. 
https://github.com/commaai/openpilot/pull/23530
also DM crops the top of my head off now on the C3 and doesn't work. (sent a log a while back)


I wasn't really thinking this part needed to look at any logs though. 
      in_linear_region = abs(self.steering_angle) < 45 or not self.steering_pressed

this line lets the learner make observations at any steering angle.

nuwandavek:
> this line lets the learner make observations at any steering angle.
It only allows the learner to input observations when the angle is < 45 deg or when not `steering_pressed` (i.e op driving), right?
I'm not sure I follow how the filter learns at ~250 degrees. Are you saying this condition breaks sometimes?

ClockeNessMnstr:
Two things, yes, assisting turns over 45deg but also the spikes occur when resuming. Even waiting until 0deg steering

I took out the lockout because it only reduces the jump in values. 
I'm not sure why there's not a spike when leaving the linear region and only on making observations after re entering.

ClockeNessMnstr:
Figured out why there was such large jumps in values. https://github.com/commaai/openpilot/pull/24315
EKF was too confident on estimates for the un-modeled values when not making observations.

ClockeNessMnstr:
Added logs/routes to the other PR showing the issues on stock + stock without "patches"

nuwandavek:
I've made an issue (https://github.com/commaai/openpilot/issues/24566). The solution here is not quite right as we want to learn the parameters even when OP is not engaged. I'll look at the aggregate data and quantify the issue soon. We can see what to do then. This might become very relevant with e2e-long doing high YR banked turns.

PR #24173, opened 2022-04-08
Honda: resolve controls mismatch related to interceptor thresholds

sshane:

PR #24185, opened 2022-04-11
test_models: support for specifying segment

sshane:
Pre-req for PR: https://github.com/commaai/openpilot/pull/24173


selfdrive/car/tests/routes.py
```diff
@@ -36,7 +36,7 @@
   TestRoute("8190c7275a24557b|2020-01-29--08-33-58", CHRYSLER.PACIFICA_2019_HYBRID),
   TestRoute("3d84727705fecd04|2021-05-25--08-38-56", CHRYSLER.PACIFICA_2020),
 
-  #TestRoute("f1b4c567731f4a1b|2018-04-30--10-15-35", FORD.FUSION),
+  # TestRoute("f1b4c567731f4a1b|2018-04-30--10-15-35", FORD.FUSION),
```
adeebshihadeh, line 39: why was this changed? the convention tends to be no spaces for commented out code

sshane: The convention is to always include 1 space after a comment, according to PEP8. Black does this as well

adeebshihadeh: we don't adhere to PEP8

selfdrive/car/tests/test_models.py
```diff
@@ -36,32 +36,37 @@
...
         print(f"Skipping tests for {cls.car_model}: missing route")
         raise unittest.SkipTest
       raise Exception(f"missing test route for {cls.car_model}")
 
     disable_radar = False
-    for seg in (2, 1, 0):
+
+    test_segs = [2, 1, 0]
+    if cls.test_route.segment is not None:
+      test_segs.insert(0, cls.test_route.segment)
```
adeebshihadeh, line 65: we only want to run on the specified segment if one is specified

PR #24190, opened 2022-04-12
Change camera readout to 25ms

pd0wm:
This PR is on hold until we are ready to merge full HDR support. 

The current trigger mode limits the integration time to `FRAME_LENGTH_LINES`. To to increase the roll speed we are forced to lower max exposure time. This can be fixed by changing to a different trigger mode, but that will cause `timestamp_sof` to go out of sync between cameras. Once we have full HDR and can use the same exposure on all cameras so this is not an issue anymore.

Changes clock from 80Mhz to 88Mhz. Need to check effect on GPS.

### DIFF
--- a/selfdrive/camerad/cameras/sensor2_i2c.h
+++ b/selfdrive/camerad/cameras/sensor2_i2c.h
@@ -48,13 +48,16 @@ struct i2c_random_wr_payload init_array_imx390[] = {
 struct i2c_random_wr_payload init_array_ar0231[] = {
   {0x301A, 0x0018}, // RESET_REGISTER
 
-  // CLOCK Settings
-  {0x302A, 0x0006}, // VT_PIX_CLK_DIV
-  {0x302C, 0x0001}, // VT_SYS_CLK_DIV
-  {0x302E, 0x0002}, // PRE_PLL_CLK_DIV
-  {0x3030, 0x0032}, // PLL_MULTIPLIER
-  {0x3036, 0x000C}, // OP_WORD_CLK_DIV
-  {0x3038, 0x0001}, // OP_SYS_CLK_DIV
+  // CLOCK Settings 
+  // Target System Frequency: 88 MHz
+  // Input Clock Frequency: 19.200 MHz
+  // Output data rate (per lane): 58.667 MHz
+  {0x302A, 0x0008}, // VT_PIX_CLK_DIV = 8
+  {0x302C, 0x0001}, // VT_SYS_CLK_DIV = 1
+  {0x302E, 0x0003}, // PRE_PLL_CLK_DIV = 3
+  {0x3030, 0x006E}, // PLL_MULTIPLIER = 110
+  {0x3036, 0x000C}, // OP_PIX_CLK_DIV = 12
+  {0x3038, 0x0001}, // OP_SYS_CLK_DIV = 1
 
   // FORMAT
   {0x3040, 0xC000}, // READ_MODE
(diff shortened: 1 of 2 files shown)


selfdrive/camerad/cameras/sensor2_i2c.h
```diff
@@ -77,7 +80,7 @@ struct i2c_random_wr_payload init_array_ar0231[] = {
 
   // Readout timing
   {0x300C, 0x07B9}, // LINE_LENGTH_PCK
-  {0x300A, 0x0652}, // FRAME_LENGTH_LINES
+  {0x300A, 0x05CB}, // FRAME_LENGTH_LINES
```
geohot, line 83: Why is this still so big? I wouldn't expect it to be much greater than 1208

selfdrive/camerad/cameras/sensor2_i2c.h
```diff
@@ -48,13 +48,16 @@ struct i2c_random_wr_payload init_array_imx390[] = {
...
-  {0x3038, 0x0001}, // OP_SYS_CLK_DIV
+  // CLOCK Settings 
+  // Target System Frequency: 88 MHz
+  // Input Clock Frequency: 19.200 MHz
+  // Output data rate (per lane): 58.667 MHz
+  {0x302A, 0x0008}, // VT_PIX_CLK_DIV = 8
+  {0x302C, 0x0001}, // VT_SYS_CLK_DIV = 1
+  {0x302E, 0x0003}, // PRE_PLL_CLK_DIV = 3
+  {0x3030, 0x006E}, // PLL_MULTIPLIER = 110
+  {0x3036, 0x000C}, // OP_PIX_CLK_DIV = 12
```
geohot, line 59: Why did the name of this reg change?

selfdrive/camerad/cameras/sensor2_i2c.h
```diff
@@ -94,8 +94,8 @@ struct i2c_random_wr_payload init_array_ar0231[] = {
   {0x3348, 0x0111}, // MIPI_F2_VDT_VC
   {0x334C, 0x0211}, // MIPI_F3_VDT_VC
   {0x3350, 0x0311}, // MIPI_F4_VDT_VC
-  {0x31B0, 0x0053}, // FRAME_PREAMBLE
-  {0x31B2, 0x003B}, // LINE_PREAMBLE
+  {0x31B0, 0x0052}, // FRAME_PREAMBLE
```
pd0wm, line 97: This doesn't seem to change anything

pd0wm:
That latest commit makes it run at 40pfs/25ms. The problem was in the fine integration time. Think the register wizards output `FINE_INTEGRATION_TIME4` is supposed to be used for the last exposure that's actually used. In our case that's exposure 2 because for some reason we're using two exposure. I accidentally made this work on master due to a typo: https://github.com/commaai/openpilot/blob/master/selfdrive/camerad/cameras/sensor2_i2c.h#L117 (notice how all the registers are `0x321E`). 

The AE oscillations are fixed by waiting longer to send the i2c commands. From experimenting with different readout times it seems like you need to wait at least 100ms minus whatever your readout time is. This just turned out to work with the 40ms currently on master since I already put in a bit of margin. I don't have a logical explanation for this, except that maybe timestamp_sof is not what we think it is.

pd0wm:
From the datasheet:
> The sensor logic will start the shutter pointer before readout to provide the required coarse integration time (CIT). The required CIT must not be equal or greater than FLL. In HDR mode, the sum of all exposures must be less than FLL.

PR #24198, opened 2022-04-12
2019 Sonata: remove from legacy safety mode

sshane:
Awaiting a tester


adeebshihadeh:
This should be verified with data, not a user.

PR #24200, opened 2022-04-13
fix regen: specify frame timestamps for modeld

sshane:
With the change from PR https://github.com/commaai/openpilot/pull/23901, modeld looks at the frame timestamps instead of frame ids, but regen wasn't migrated along with that, so modeld was just seeing zeros and skipping all frames. Now we just set them to the current `logMonoTime`


sshane:
Wait this doesn't look right, this isn't an issue on C2 as we copy meta_extra below. Let me take another look

PR #24201, opened 2022-04-13
Lateral Planner: Set a minimum heading cost for lateral planner

nuwandavek:
Set a minimum `heading_cost` in the lateral planner where `use_lanelines == False`. This should help reduce pingpoing significantly without harming performance.

image

*ToDo*
- Run lateral report and compare.


HaraldSchafer:
This requires a new model that we have better guarantees around hugging for,

nuwandavek:
Closing this for now, will revisit when we have a new model that I can sanity-check this change on.

ntegan1:
implemented in https://github.com/commaai/openpilot/pull/24521

PR #24202, opened 2022-04-13
Fix camera error description

gijskoning:
Camera error description was changed incorrectly in #24156 possibly due to the order of road, driver, wideroad.
*At least the eventname doesn't match the error description in the code. I haven't verified it in any way yet.

Changed the order to road, wideroad, driver and so fixed the description.

Should the messages have the detailed description of which camera it is? Since this was removed in https://github.com/commaai/openpilot/pull/22943 and then added back in #24156


pd0wm:
Ah, my bad. I used the wrong description.

I also think the mapping in controlsd is wrong (https://github.com/commaai/openpilot/blob/master/selfdrive/controls/controlsd.py#L55). Need to unplug some cameras to verify though.

PR #24204, opened 2022-04-13
reduce updated data usage

pd0wm:
- Set protocol version 2, this changes the `git fetch --dry-run` from 400KB to 18KB. See https://github.blog/2020-03-22-highlights-from-git-2-26/ for some more information
 - Reduce check frequency from once every 1 minute to once every 5 minutes

These two changes combined should reduce the theoretical data usage from 17GB to just 156MB per month for a device with 24/7 uptime.

TODO:
- [x]  Test on C2

### DIFF
--- a/selfdrive/updated.py
+++ b/selfdrive/updated.py
@@ -163,9 +163,13 @@ def setup_git_options(cwd: str) -> None:
 
   # We are using copytree to copy the directory, which also changes
   # inode numbers. Ignore those changes too.
+
+  # Set protocol to the new version (default after git 2.26) to reduce data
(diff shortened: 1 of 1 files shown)


selfdrive/updated.py
```diff
@@ -464,7 +468,7 @@ def main() -> None:
     except Exception:
       cloudlog.exception("uncaught updated exception while setting params, shouldn't happen")
 
-    wait_helper.sleep(60)
+    wait_helper.sleep(300)
```
adeebshihadeh, line 471:
```suggestion
    # TODO: replace this with a good backoff
    wait_helper.sleep(300)
```

PR #24208, opened 2022-04-14
Improve lagging message description

gijskoning:
The current messages don't clearly show which process is lagging.
Change:
"Process-1 lagging by..." -> "Process-1 'bridge' lagging by..."

### DIFF
--- a/common/realtime.py
+++ b/common/realtime.py
@@ -2,8 +2,9 @@
 import gc
 import os
 import time
-import multiprocessing
 from typing import Optional
+
+from setproctitle import getproctitle  # pylint: disable=no-name-in-module
 
 from common.clock import sec_since_boot  # pylint: disable=no-name-in-module, import-error
 from selfdrive.hardware import PC, TICI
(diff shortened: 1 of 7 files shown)


common/realtime.py
```diff
@@ -49,14 +49,14 @@ def config_realtime_process(core: int, priority: int) -> None:
...
-  def __init__(self, rate: int, print_delay_threshold: Optional[float] = 0.0) -> None:
+  def __init__(self, rate: int, process_name: str = "-", print_delay_threshold: Optional[float] = 0.0) -> None:
     """Rate in Hz for ratekeeping. print_delay_threshold must be nonnegative."""
     self._interval = 1. / rate
     self._next_frame_time = sec_since_boot() + self._interval
     self._print_delay_threshold = print_delay_threshold
     self._frame = 0
     self._remaining = 0.0
-    self._process_name = multiprocessing.current_process().name
+    self._process_name = multiprocessing.current_process().name + f" '{process_name}'"
```
adeebshihadeh, line 59: `multiprocessing.current_process().name` doesn't cover this?

adeebshihadeh, line 59: `multiprocessing.current_process().name` doesn't cover this?

gijskoning: The name 'Process-1' didn't seem that useful to me.

adeebshihadeh: instead of passing it in, let's replace `multiprocessing.current_process().name` with `getproctitle`. manager already sets the proc title https://github.com/commaai/openpilot/blob/acd455ed3ae0ffbb82fe36dd6d9a1f4d16279823/selfdrive/manager/process.py#L32

gijskoning: Great! That looks much better.

PR #24209, opened 2022-04-14
Trying to fix visiontest

gijskoning:
Wanted to use the visiontest to see if I can run the models locally. However couldn't get it working.

Did notice that visiontest.mk seems incorrect since util.c doesn't exist. There was a name change from util.c to util.cc a year ago https://github.com/commaai/openpilot/pull/19710 and this file could have been forgotten.

This change doesn't fix it for me yet. I currently get: `error: invalid argument '-std=gnu11' not allowed with 'C++'` Any tips are welcome.


gijskoning:
Mm it seems visiontest is not even used in CI. Is this an old file that is not used anymore?

adeebshihadeh:
Yes, visiontest is an old file. The model should be run with modeld on your PC.

PR #24214, opened 2022-04-15
UI: double click to switch between body and onroad views

adeebshihadeh:


selfdrive/ui/qt/home.cc
```diff
@@ -82,6 +84,18 @@ void HomeWindow::mousePressEvent(QMouseEvent* e) {
   }
 }
 
+void HomeWindow::mouseDoubleClickEvent(QMouseEvent* e) {
+  // switch to the generic robot UI
```
adeebshihadeh, line 88:
```suggestion
```

PR #24221, opened 2022-04-15
EV6 corner radar

pd0wm:
TODO:
- [ ] Find status flags
- [ ] Find relative speed
- [ ] Generate DBC file

route: `d824e27e8c60172c|2022-04-13--13-18-56`

### DIFF
--- a/selfdrive/car/hyundai/radar_interface.py
+++ b/selfdrive/car/hyundai/radar_interface.py
@@ -1,14 +1,30 @@
 #!/usr/bin/env python3
-import math
+from collections import defaultdict
 
-from cereal import car
+from cereal import car, messaging
 from opendbc.can.parser import CANParser
 from selfdrive.car.interfaces import RadarInterfaceBase
 from selfdrive.car.hyundai.values import DBC
 
+import math
+
 RADAR_START_ADDR = 0x500
 RADAR_MSG_COUNT = 32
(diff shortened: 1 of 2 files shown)


selfdrive/car/hyundai/interface.py
```diff
@@ -227,6 +227,18 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[], disabl
       tire_stiffness_factor = 0.5
       ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0.], [0.]]
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.25], [0.05]]
+    elif candidate == CAR.KIA_EV6:
+      ret.mass = 2055 + STD_CARGO_KG
+      ret.wheelbase = 2.9
+      ret.steerRatio = 16.
+      ret.safetyConfigs = [get_safety_config(car.CarParams.SafetyModel.noOutput),
```
gregjhogan, line 234: should probably use `SafetyModel.silent` for private radar bus just to make sure it doesn't interfere with anything?

pd0wm: Boardd already does this if you don't specify a safetyconfig for a panda

PR #24227, opened 2022-04-16
Added fingerprint for Golf MK7

Knudn:
Added fingerprint for my Golf MK7 2019
Dongle ID: 596ffb384692bf3f


jyoung8607:
Can we trouble you to add those sorted in the correct order in the existing list? It makes it a lot easier to find duplicates when adding new entries in the future.

Knudn:
These are the entires where it finds ducplicates

\xf1\x875Q0909144AB\xf1\x891082\xf1\x82\00521A07B05A1
\xf1\x875Q0909144AB\xf1\x891082\xf1\x82\x0521A07B05A1  (the one I added)

They are however not identical, the last segment of the one I added start with x0, while the other starts with 00. 
Can I just remove the one I added and it will work as intended?

jyoung8607:
The octal expression `\005` and hexadecimal expression `\x05` both represent the same underlying value. There are some bugs in underlying layers of openpilot that sometimes cause this data to be displayed inconsistently, but they can be ignored for this purpose. No need to add a second copy.

Knudn:
Perfect, thanks :)

PR #24228, opened 2022-04-16
improve tici thermal control

adeebshihadeh:
Moved the temperature setpoint into the green thermal band. Also think that the controller should be a bit more aggressive, see `5decac85ab437ed1|2022-04-16--13-17-07`

### DIFF
--- a/selfdrive/thermald/fan_controller.py
+++ b/selfdrive/thermald/fan_controller.py
@@ -83,7 +83,7 @@ class TiciFanController(BaseFanController):
     cloudlog.info("Setting up TICI fan handler")
 
     self.last_ignition = False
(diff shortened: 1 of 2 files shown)


selfdrive/thermald/fan_controller.py
```diff
@@ -92,7 +92,7 @@ def update(self, max_cpu_temp: float, ignition: bool) -> int:
     if ignition != self.last_ignition:
       self.controller.reset()
 
-    error = 75 - max_cpu_temp
+    error = (70 if ignition else 75) - max_cpu_temp
```
robbederks, line 95: It might also be fine to lower the offroad setpoint to 70C as well. This is the offroad distribution we get now (so only a small percentage of devices is >70C), and once we lower power consumption it should be even better.
image

robbederks (approved):

gijskoning:
Just wondering, why no proportional gain?

PR #24229, opened 2022-04-16
encoderd: WIP

geohot:
It would be really nice to separate loggerd into loggerd and encoderd.

There's a ton of work to do for this, and you'd have to make sure msgq never drops packets. You'd also have to order loggerd first than encoderd.

Encoderd would be responsible for the rotation. The remuxing logic would move into loggerd from OMXEncoder

### DIFF
--- a/selfdrive/loggerd/.gitignore
+++ b/selfdrive/loggerd/.gitignore
@@ -1,2 +1,3 @@
+encoderd
 loggerd
 tests/test_logger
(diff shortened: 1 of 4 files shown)


geohot:
Hmm, it's possible to not have to restart the encoders with "OMX_BUFFERFLAG_SYNCFRAME"

geohot:
Also looking at the kernel API below OMX. It lets you pass in arbitrary buffers, not just "pre queued" ones. I imagine this could be like thneed and draw almost no CPU.

PR #24239, opened 2022-04-18
AGNOS updater: add download speed indicator

### LINKED ISSUE #24119: AGNOS updater: add download speed indicator

whowe444:
fixes #24119


selfdrive/hardware/tici/agnos.py
```diff
@@ -33,6 +40,13 @@ def read(self, length: int) -> bytes:
         break
       out = self.decompressor.decompress(compressed)
       self.buf += out
+      self.downloaded += len(out)
+      now = time.monotonic()
+      if (now - self.last_read > 1):
```
whowe444, line 45: this is measuring the speed in a pretty naive way since bytes are being read from the buffer in between chunks, but assuming the download takes up the majority of time, it could be somewhat accurate

selfdrive/hardware/tici/agnos.py
```diff
@@ -33,6 +40,13 @@ def read(self, length: int) -> bytes:
...
+        percent_done = round(self.downloaded / self.file_size * 100)
```
pd0wm, line 46: This is now wrong since the data is compressed (and sparsified). The request content length is not equal to the size of the output from the decompressor and unsparsifier.

To measure download speed you need to check the amount of input bytes processed before the decompressor.

pd0wm (changes requested):

adeebshihadeh:
Thanks for the PR, but #24119 also involves showing the download speed in the UI. You're welcome to finish this up and re-open, but no guarantees we'll merge it. We tend to prefer doing updater work ourselves, since most of the work is in validating it.

PR #24241, opened 2022-04-18
Fingerprint for 2022 Hyundai Elantra Hybrid

maxtriano:
Fingerprint for 2022 Elantra Hybrid. Only fwdCamera and eps needed new entries, rest of carFw for the 2022 is the same as the 2021.

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -1034,6 +1034,7 @@ FW_VERSIONS = {
   },
   CAR.ELANTRA_HEV_2021: {
     (Ecu.fwdCamera, 0x7c4, None) : [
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Which route did these come from?

maxtriano:
Route is 786fc028c014be71|2022-04-17--20-06-22

sshane:
Since it's a new model year, I think the plan is to check:
- [x] Run through test models to verify checksum
- [ x Verify gear packet message
- [x] Verify `use_fca`
- [x] Verify `HYBRID_CAR`
- [x] Verify `create_lfahda_mfc` in carcontroller

sshane:
Moved to: https://github.com/commaai/openpilot/pull/24249

PR #24243, opened 2022-04-18
loggerd: prereqs for deanlee loggerd omx encoder

geohot:
Just a refactor to replace "downscale" with "out_width" and "out_height"


geohot:
Replace with #24252

PR #24247, opened 2022-04-18
VW: Add FW for 2019 Volkswagen e-Golf MK7

Knudn:
Added fingerprint for my Golf MK7 2019
Dongle ID: 596ffb384692bf3f


selfdrive/car/volkswagen/values.py
```diff
@@ -255,6 +255,7 @@ class VWCarInfo(CarInfo):
       b'\xf1\x870EA906016F \xf1\x894238',
       b'\xf1\x870EA906016F \xf1\x895002',
       b'\xf1\x870EA906016S \xf1\x897207',
+      b'\xf1\x870EA906016Q \xf1\x895993',
```
jyoung8607, line 258: Should be moved up one, Q suffix firmware falls after F and before S.

jyoung8607 (changes requested):

selfdrive/car/volkswagen/values.py
sshane, line 258:
```suggestion
      b'\xf1\x870EA906016F \xf1\x895002',
      b'\xf1\x870EA906016Q \xf1\x895993',
      b'\xf1\x870EA906016S \xf1\x897207',
```

PR #24248, opened 2022-04-18
remove comma two support

adeebshihadeh:
`commatwo_master` is the place for the remaining comma two work: https://github.com/commaai/openpilot/commits/commatwo_master

The jenkins pipeline there will push to release2-staging instead of master-ci.


selfdrive/controls/controlsd.py
```diff
@@ -221,10 +221,6 @@ def update_events(self, CS):
       self.events.add_from_msg(self.sm['driverMonitoringState'].events)
 
     # Create events for battery, temperature, disk space, and memory
```
jyoung8607, line 223:
```suggestion
    # Create events for temperature, disk space, and memory
```

sshane:
Can remove:
- https://github.com/commaai/openpilot/blob/master/common/file_helpers.py#L84

PR #24253, opened 2022-04-19
Update SNPE to 1.61.0.3358

geohot:


geohot:
weights_fixup.py appears to no longer be needed. we should have deprecated the C2 a while ago

geohot:
Gonna merge this, considering CI passes

PR #24254, opened 2022-04-19
remove weights fixup with new SNPE

geohot:
Let's see what CI does, might require new ref since it's still off by 1e-10


geohot:
It's close. Would require more testing to compare to PC.

pd0wm:
Any update on this?

HaraldSchafer:
Looks the same, red and blue are with and without fixup. Green is in onnx on PC
Screenshot from 2022-06-14 15-26-34
e

PR #24257, opened 2022-04-19
loggerd: Separate writer from encoder

geohot:
The poorly named "raw_logger" should be able to use this too.

Needed to split loggerd and encoderd.

### DIFF
--- a/selfdrive/loggerd/omx_encoder.cc
+++ b/selfdrive/loggerd/omx_encoder.cc
@@ -343,6 +343,16 @@ void OmxEncoder::callback_handler(OmxEncoder *e) {
 
     e->to_write.push(new_buffer);
 
+#ifdef QCOM2
+    if (buffer->nFlags & OMX_BUFFERFLAG_CODECCONFIG) {
+      buffer->nTimeStamp = 0;
+    }
+
+    if (buffer->nFlags & OMX_BUFFERFLAG_EOS) {
+      buffer->nTimeStamp = 0;
+    }
+#endif
+
     if (buffer->nFlags & OMX_BUFFERFLAG_EOS) {
       exit = true;
     }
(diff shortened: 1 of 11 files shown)


geohot:
This is passing. How much do we trust the loggerd tests?

adeebshihadeh:
Loggerd tests should cover this pretty well. I can review as well.

geohot:
Hmm, I suspect there's memory leaks. Do we have valgrind working?

adeebshihadeh:
Valgrind should work on agnos.

selfdrive/loggerd/omx_encoder.cc
```diff
@@ -597,19 +493,8 @@ void OmxEncoder::encoder_close() {
...
-      avformat_free_context(this->ofmt_ctx);
-    } else {
-      if (this->of) {
-        util::safe_fflush(this->of);
-        fclose(this->of);
-        this->of = nullptr;
-      }
-    }
-    unlink(this->lock_path);
+    delete writer;
```
pd0wm, line 496: Why not use a `unique_ptr`?

selfdrive/loggerd/omx_encoder.cc
```diff
@@ -343,16 +343,6 @@ void OmxEncoder::callback_handler(OmxEncoder *e) {
 
     e->to_write.push(new_buffer);
 
-#ifdef QCOM2
```
pd0wm, line 346: Did you verify that this is not needed to properly hand back the buffer to OMX? It was introduced to "Fix random encoder lockups" here: https://github.com/commaai/openpilot/pull/2707. In January we switched to a different Venus firmware, which might have been the real fix: https://commadotai.slack.com/archives/C0155QM7057/p1611277394048100.

Probably best to not have functional changes in this refactor PR, and move this functional change into a separate PR so it's more obvious if we notice any issues later.

geohot: oh, i didn't realize that's why this was done! i thought it was for the ffmpeg remuxer, and i fixed that bug

selfdrive/loggerd/omx_encoder.cc
geohot: done

geohot:
Ran in valgrind, found I was leaking the PubMaster (from old change), switched to unique_ptr to fix. All other valgrind issues are in OMX and GPU driver. (why are we linking the gpu driver?)

geohot:
CI failure seems unrelated. @adeebshihadeh ?

adeebshihadeh:
Both are unrelated: one is an issue with the updated git in the GitHub Actions runners and the other looks like a real athena bug.

PR #24259, opened 2022-04-19
modeld.cc: always use extra client

pd0wm:
@adeebshihadeh how much of this kind of stuff do we want to remove now?


zorrobyte:
Thanks for maintaining comma zero :)

PR #24260, opened 2022-04-19
Torque control with roll

haraschax:
### Adds new controller: latcontrol_torque, that is implemented on TSSP rav4 and TSS2 corolla
This controller is a PI controller on torque with some friction compensation and road roll compensation. It is designed to only use 2 car-specific parameters which is MAX_LAT_ACCEL and FRICTION. Both of which can be read from the data or measured in injection tests.

**MAX_LAT_ACCEL**: this is the lateral acceleration achieved by a car on a flat road when applying max torque in steady state. This value is mostly constant across the speed range, especially speeds above ~20mph.

**FRICTION**: This is the the torque command required to overcome the steering wheel friction. This is also mostly constant across the speed range, but does get higher at low speed. We take the value at high speed (50+mph)

**Benefits of this controller are:**
-  Car specific parameters are measureable, not tuneable, and there are only 2
-  The error domain is lateral acceleration, this is what actually correlates to torque in reality so makes the tuning far more consistent across the speed range
- Has a sensible feedforward for more accurate control
- Compensates for road roll making control more accurate
- Should be able to replace all other torque based controllers over time
- Once the localizer can be used instead of steering angle sensor, this can make the TSSP PRIUS  drive way better.
- The more accurate control means there is no more bang-bang control that saturates the rate limits all the time, which allows us to increase the torque rate limits on toyota and make sharper turns.


selfdrive/car/toyota/values.py
```diff
@@ -18,10 +18,16 @@ class CarControllerParams:
   ACCEL_MIN = -3.5  # m/s2
 
   STEER_MAX = 1500
-  STEER_DELTA_UP = 10       # 1.5s time to peak torque
-  STEER_DELTA_DOWN = 25     # always lower than 45 otherwise the Rav4 faults (Prius seems ok with 50)
   STEER_ERROR_MAX = 350     # max delta between torque cmd and torque motor
 
+  def __init__(self, CP):
+    if CP.lateralTuning.which == 'torque':
```
adeebshihadeh, line 24: use the enum instead of string

PR #24262, opened 2022-04-19
Honda: do same standstill check as panda

sshane:
Notes about Honda fix from old PR: https://github.com/commaai/openpilot/pull/24070.

Essentially STANDSTILL->VEHICLE_MOVING is really noisy near a stop, and it doesn't match the panda check for brake pressed pre-enable which is the XMISSION_SPEED, causing controls mismatches occasionally.

So considering we already use XMISSION_SPEED at low speeds in openpilot for vEgo, and in panda for standstill, this should be safe on all Hondas.

Before this change, if engaging with `0 < XMISSION_SPEED < 0.1` with brake held, it would mean the user gets a controlsMismatch event, now it would yield a pedalPressed no entry


sshane:
All the other makes use < 0.01 or 0.001 in openpilot, and < 1 kph in panda, however with Honda, `XMISSION_SPEED` can have different scaling, so it's simpler to make it compare to 0

adeebshihadeh:
Did you confirm that the new tests would fail on all the segments from the MTBF report and now pass?

sshane:
There's only 8 segments with the standstill mismatch *and* with rlogs, but they all fail with the check before, and all pass with the check after

PR #24268, opened 2022-04-20
Camry TSS2: use torque controller

sshane:
Worked great! Definite improvement over the current PID tune. Not sure if TSS-P Camrys can also use this tune

Route (only difference is I merged in `toyota-steer-faults` which has no effect except in high curvature turns at low speed): f15e3c37c118e841|2022-04-20--00-46-27


sshane:
On second try with the new values, it seems like steering is a bit jerkier in turns and hugs more (with laneless, so could be placebo, on both counts actually). Will look at the data, let me know if anyone tries this

fazeshock3008:
What is The EOD Number for this? I need to Track Down Before it goes into
manufacturing. Thank You

PR #24269, opened 2022-04-20
updated: only run offroad & disable automatic git garbage collection

### LINKED ISSUE #23879: Git high CPU/RAM usage
git garbage collection sometimes uses a lot of memory until it gets OOM killed. This can happen while onroad and cause lags.

- [ ] Ensure `git gc` is only ever started by `updated` and running in the overlay, never triggered by something like`version.py`.
- [ ] Ensure updated is immediately canceling any ongoing git/copy operations when going onroad.

pd0wm:
@adeebshihadeh any thoughts? If we combine this with killing updated when going onroad it should fix https://github.com/commaai/openpilot/issues/23879. We can also run an aggressive `gc`, but not sure if we want to change that behavior.

(disabled jenkins for now to not change the config permanently on those devices in case we don't merge it)

### DIFF
--- a/selfdrive/updated.py
+++ b/selfdrive/updated.py
@@ -171,6 +171,7 @@ def setup_git_options(cwd: str) -> None:
     ("core.checkStat", "minimal"),
     ("protocol.version", "2"),
     ("gc.auto", "0"),
+    ("gc.autoDetach", "false"),
   ]
   for option, value in git_cfg:
     run(["git", "config", option, value], cwd)
--- a/selfdrive/updated.py
+++ b/selfdrive/updated.py
@@ -383,6 +383,7 @@ def main() -> None:
   update_failed_count = 0
 
   # Set initial params for offroad alerts
+  # TODO: set once on startup from manager?
   set_params(False, 0, None)
 
   # Wait for IsOffroad to be set before our first update attempt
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
LGTM. I would also disable `gc.autoDetach` explicitly and cloudlog the gc start and duration.

FYI, this shouldn't have permanently changed anything about jenkins. We make an ephemeral copy of the repo that the tests are run in.

pd0wm:
Shall we also set `pack.windowMemory` and `pack.threads` to something reasonable? Maybe 512M and 4 threads?

pd0wm:
Takes about 40s on master and 4 seconds on release3. Think it's fine to run this every time (and not use `git gc --auto`).

selfdrive/updated.py
```diff
@@ -256,8 +258,19 @@ def finalize_update() -> None:
...
   run(["git", "submodule", "foreach", "--recursive", "git", "reset"], FINALIZED)
 
-  set_consistent_flag(True)
-  cloudlog.info("done finalizing overlay")
+  cloudlog.info("Starting git gc")
+  t = time.monotonic()
+  try:
+    run(["git", "gc"], FINALIZED)
+    cloudlog.info(f"Done git gc, took {time.monotonic() - t:.3f} s")
+  except subprocess.CalledProcessError:
```
pd0wm, line 266: @adeebshihadeh Do we want to block updating if `git gc` fails or is that too risky? Currently this `except` triggers when going onroad during the `git gc`, which is why I have to check the wait helper afterwards to not finalize the overlay in a half-finished gc state.

selfdrive/updated.py
```diff
@@ -367,10 +380,7 @@ def main() -> None:
 
   first_run = True
   last_fetch_time = 0.0
-  update_failed_count = 0
-
-  # Set initial params for offroad alerts
-  set_params(False, 0, None)
```
pd0wm, line 373: This would set `LastUpdateTime` to the current time when starting updated. Was that intentional?

selfdrive/updated.py
adeebshihadeh: hmm, can interruption corrupt the repo? you almost want an overlay for the overlay :P

pd0wm: Can't find anything definitive about corruption. Think it's not a big deal to start over if you interrupt it.

PR #24271, opened 2022-04-20
swaglog: delay creating zmq socket to first use - fix modeld crash in sim

### LINKED ISSUE #23666: CARLA: modeld crashes on startup

gijskoning:
Fixes modeld crash in simulator. The zmq_socket was created at import leading to two sockets in one thread. This pr calls initialize() on the first usage of the logs.
Problem started here: https://github.com/commaai/openpilot/pull/23596

Fixes https://github.com/commaai/openpilot/issues/23666
And prs https://github.com/commaai/openpilot/pull/24203 and https://github.com/commaai/openpilot/pull/23792 can be closed.


selfdrive/common/statlog.cc
```diff
@@ -17,6 +17,8 @@ class StatlogState : public LogState {
 static StatlogState s = {};
 
 static void log(const char* metric_type, const char* metric, const char* fmt, ...) {
+  if (!s.initialized) s.initialize();
```
pd0wm, line 20: You want to take the lock like in `cloudlog_common`, this can cause race conditions if multiple threads call `log` at the same time.

selfdrive/common/util.h
```diff
@@ -182,7 +188,9 @@ class LogState {
     zmq_setsockopt(sock, ZMQ_LINGER, &timeout, sizeof(timeout));
 
     zmq_connect(sock, endpoint);
-  };
+    initialized = true;
+  }
+
   ~LogState() {
     zmq_close(sock);
```
pd0wm, line 195: You can reach this code now with an uninialized `LogState`. Check for `initialized` or you'll try to close an unopened socket.

Also set `zctx` and `sock` to `nullptr` as default values just to be sure.

PR #24274, opened 2022-04-20
Simulator Tests in CI

gijskoning:
Add 3 tests to verify bridge.py and the all processes of manager.py correctly startup and keep running for 60 seconds. Each need Carla simulator to run in parallel.
First two test simply verify that connection to Carla is possible and that bridge.py works as expected.
The third runs manager and bridge. Checks processes are running and that car can be engaged.

Bridge.py looks like a lot of changes but it is mostly creating a Bridge class and putting some functionality in functions

Also speeds up simulation by changing post processes to false

Failed rebasing from https://github.com/commaai/openpilot/pull/24009 so new pr here.


tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+  def test_engage(self):
+    # Startup manager and bridge.py. Check processes are running, then engage and verify.
+    # Set environment vars for manager.py. The same as sim/launch_openpilot.sh
+    os.environ["PASSIVE"] = "0"
+    os.environ["NOBOARD"] = "1"
+    os.environ["SIMULATION"] = "1"
+    os.environ["FINGERPRINT"] = "HONDA CIVIC 2016"
+    os.environ["BLOCK"] = "camerad,loggerd"
+
+    self.subprocess = subprocess.Popen("../../../selfdrive/manager/manager.py")
```
adeebshihadeh, line 55: run `tools/sim/launch_openpilot.sh` instead of copying this

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+    Params().put_bool("DoShutdown", True)
+    # Process could already be closed
+    if type(self.subprocess) is subprocess.Popen:
+      self.subprocess.wait(timeout=10)
+      # Closing gracefully
+      self.assertEqual(self.subprocess.returncode, 0)
+    p_bridge.join()
+
+  def tearDown(self):
+    print("Teardown")
```
adeebshihadeh, line 102: remove this

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+    os.environ["BLOCK"] = "camerad,loggerd"
+
+    self.subprocess = subprocess.Popen("../../../selfdrive/manager/manager.py")
+    sm = messaging.SubMaster(['controlsState', 'carEvents', 'managerState'])
+
+    args = sys.argv[2:]  # Remove test arguments when executing this test
+    args.append('--low_quality')
+    p_bridge, _, q = bridge.main(args, keep_alive=False)
+
+    start_time = time.time()
```
adeebshihadeh, line 62: use time.monotonic, time.time can jump

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+      if sm['controlsState'].active:
+        control_active = True
+        break
+
+      time.sleep(0.1)
+
+    self.assertTrue(control_active)
+
+    # Set shutdown to close manager and bridge processes
+    Params().put_bool("DoShutdown", True)
```
adeebshihadeh, line 93: the test shouldn't really rely on this, send SIGTERM instead

tools/sim/bridge.py
```diff
@@ -231,266 +234,290 @@ def can_function_runner(vs: VehicleState, exit_event: threading.Event):
...
+    gps.destroy()
+    imu.destroy()
+    for c in cameras:
+      c.destroy()
+    vehicle.destroy()
+
+
+def main(add_args=None, keep_alive=True):
+  unblock_stdout()  # Fix error when publishing too many lag message
+  _args = parse_args(add_args)
```
adeebshihadeh, line 500: why is this prefixed with an underscore? usually, this is for private variables in python

tools/sim/bridge.py
```diff
@@ -231,266 +234,290 @@ def can_function_runner(vs: VehicleState, exit_event: threading.Event):
...
+  carla_bridge = CarlaBridge(_args)
+
+  _q: Any = Queue()
```
adeebshihadeh, line 503: same here

tools/sim/bridge.py
```diff
@@ -12,26 +12,19 @@
...
 
-from lib.can import can_function
-
+from selfdrive.manager.helpers import unblock_stdout
+from tools.sim.lib.can import can_function
 import cereal.messaging as messaging
 from cereal import log
-from cereal.visionipc.visionipc_pyx import VisionIpcServer, VisionStreamType  # pylint: disable=no-name-in-module, import-error
+from cereal.visionipc.visionipc_pyx import VisionIpcServer, \
+  VisionStreamType  # pylint: disable=no-name-in-module, import-error
```
adeebshihadeh, line 20: why a new line for this? and if using a new line, indent to the same level as `VisionIpcServer`

tools/sim/bridge.py
```diff
@@ -231,266 +234,290 @@ def can_function_runner(vs: VehicleState, exit_event: threading.Event):
...
+
+    blueprint_library = world.get_blueprint_library()
+
+    world_map = world.get_map()
+
+    vehicle_bp = blueprint_library.filter('vehicle.tesla.*')[1]
+    spawn_points = world_map.get_spawn_points()
+    assert len(spawn_points) > self._args.num_selected_spawn_point, \
+      f'''No spawn point {self._args.num_selected_spawn_point}, try a value between 0 and
+      {len(spawn_points)} for this town.'''
```
adeebshihadeh, line 291: up to you, but we don't enforce an 80 char line limit

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+    # Will raise an error if not connected
+    client.load_world('Town04_Opt')
+
+  def test_run_bridge(self):
+    # Test bridge connect with carla and runs without any errors for 60 seconds
+    test_duration = 60
+
+    params = Params()
+    args = sys.argv[2:]  # Remove test arguments when executing this test
+    args.append('--low_quality')
```
adeebshihadeh, line 35: why are the args from the test being copied? works now because it's just an empty list

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+    p_bridge, _, q = bridge.main(args, keep_alive=False)
+
+    start_time = time.time()
+
+    no_car_events_issues_once = False
+    max_time_per_test = 20
+    while time.time() < start_time + max_time_per_test:
+      sm.update()
+      not_running = {p.name for p in sm['managerState'].processes if not p.running and p.shouldBeRunning}
```
adeebshihadeh, line 68: handle manager crashing too?

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+
+      if len(sm['carEvents']) == 0 and len(not_running) == 0:
```
adeebshihadeh, line 70: not sure this is actually what you want. on some cars, `pcmDisable` is always in the carEvents while not engaged, which is perfectly valid. this just happens to work with the civic that's used in the bridge right now. I would just remove this check, and maybe check the `controlsState.alertXXX` fields.

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+        no_car_events_issues_once = True
+        break
+
+      time.sleep(0.1)
+    self.assertTrue(no_car_events_issues_once)
```
adeebshihadeh, line 75: can we make some nice error messages here since the UI won't be running? let's show which processes crashed and/or which alerts were showing

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+
+      sm.update()
+
+      if sm['controlsState'].active:
+        control_active = True
+        break
+
+      time.sleep(0.1)
+
+    self.assertTrue(control_active)
```
adeebshihadeh, line 90:
```suggestion
    self.assertTrue(control_active, "Simulator never engaged")
```

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # Set shutdown to close manager and bridge processes
+    Params().put_bool("DoShutdown", True)
+    # Process could already be closed
+    if type(self.subprocess) is subprocess.Popen:
+      self.subprocess.wait(timeout=10)
+      # Closing gracefully
```
adeebshihadeh, line 97: inline comments should be added very judiciously. they should really only add something that isn't apparent after reading the code

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+      self.assertEqual(self.subprocess.returncode, 0)
+    p_bridge.join()
+
+  def tearDown(self):
+    print("Teardown")
+    Params().put_bool("DoShutdown", True)
+    manager.manager_cleanup()
```
adeebshihadeh, line 104: does this do anything? manager's running as a subprocess

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+
+      time.sleep(0.1)
+
+    self.assertTrue(control_active)
+
+    # Set shutdown to close manager and bridge processes
+    Params().put_bool("DoShutdown", True)
+    # Process could already be closed
+    if type(self.subprocess) is subprocess.Popen:
+      self.subprocess.wait(timeout=10)
```
adeebshihadeh, line 96: does this actually send a signal to terminate? I suspect it's just hitting the timeout here

tools/sim/bridge.py
```diff
@@ -231,266 +234,290 @@ def can_function_runner(vs: VehicleState, exit_event: threading.Event):
...
+    for t in reversed(threads):
+      t.join()
+    gps.destroy()
+    imu.destroy()
+    for c in cameras:
+      c.destroy()
+    vehicle.destroy()
+
+
+def main(add_args=None, keep_alive=True):
```
adeebshihadeh, line 498: I wouldn't really expect a main function to return anything. instead, I'd suggest passing in the parsed args and queue

tools/sim/test/test_carla_integration.py
```diff
@@ -0,0 +1,110 @@
...
+        break
+
+      time.sleep(0.1)
+
+    self.assertTrue(control_active)
+
+    # Set shutdown to close manager and bridge processes
+    Params().put_bool("DoShutdown", True)
+    # Process could already be closed
+    if type(self.subprocess) is subprocess.Popen:
```
adeebshihadeh, line 95: this kind of check should be done with isinstance, but would remove this check entirely

tools/sim/test/test_carla_integration.py
pd0wm: The idea of the check was to ensure there are no commIssue or other events that prevent engagement. Maybe we should only check events of certain types?

tools/sim/bridge.py
gijskoning: Changed it. Was still there after moving it from the main if statement

tools/sim/test/test_carla_integration.py
gijskoning: This was to execute it in command line, and would fail otherwise. But I will just remove it. 
python -m unittest test_carla_integration.TestCarlaIntegration.test_engage

tools/sim/test/test_carla_integration.py
gijskoning: It doesnt actually. Removed it

tools/sim/test/test_carla_integration.py
gijskoning: I changed it to check if there aren't any messages that contain 'Issue' or 'Error'.

tools/sim/test/test_carla_integration.py
gijskoning: Thanks will keep it in mind. Removed this part completely

tools/sim/test/test_carla_integration.py
gijskoning: Nope can be removed

tools/sim/bridge.py
gijskoning: The reason was to not have to create args and queue in every test. However, have changed it now.

tools/sim/test/test_carla_integration.py
gijskoning: Thanks will keep it in mind for the next time

tools/sim/test/test_carla_integration.py
pd0wm: You can also check for the  `noEntry`, `softDisable` and `immediateDisable` fields

gijskoning: Done!

pd0wm:
Tried running the tests on my computer, but ran into a few issues:
 - The tests wasn't actually waiting for messages to come in. Be careful when using submaster, if no messages have been received yet all struct values will have their default values. Already fixed this by adding a check for `sm.all_alive()`
 - Carla seems to segfault sometimes when one of the test exits. Can we fix this by closing the connection properly?
 - Shutting down the tests always fails with `AssertionError: can only join a child process`, causing the exit code to be negative
 - I'm seeing some `time.sleep()` to wait for things to come online. Those sleeps are usually flaky and cause unnecessary random test failures, and in the best case they make the tests slower than needed. Replace those by checking for the actual condition (e.g. is bridge running correctly), and add a high upper bound (like you do for the engaged checks).

gijskoning:
Thanks for the improvements. I removed some time.sleep calls and all the processes should close safely now.

ebadi:
> **To engage openpilot press 1 a few times while focused on bridge.py to increase the cruise speed.**

I guess this commit broke the keyboard control (Cruise_up).

PR #24279, opened 2022-04-20
paramsd: use real logMonoTime

adeebshihadeh:
Anything rely on this @nuwandavek?


nuwandavek:
No. Is this more correct? Only case I can think of is in the downstream analysis if the carstate logmonotime is very far off from actual logmonotime

adeebshihadeh:
Yes, logMonoTime should be timestamped in the process building/sending the message. For cases where we want to track the source info, we add a field with a frame ID or logMonoTime, e.g. `lateralPlan.modelMonoTime`.

PR #24284, opened 2022-04-21
mypy: enable --warn-return-any

GregorKikelj:
This one took more work, might be worth checking if every change is actually wanted because some changes make the code less readable, it's quite easy to add ` # type: ignore` where needed.


GregorKikelj:
Might also help with https://github.com/commaai/openpilot/pull/23665 which tried to add the same flag

selfdrive/controls/lib/events.py
```diff
@@ -135,8 +135,8 @@ def __str__(self) -> str:
     return f"{self.alert_text_1}/{self.alert_text_2} {self.priority} {self.visual_alert} {self.audible_alert}"
 
   def __gt__(self, alert2) -> bool:
-    return self.priority > alert2.priority
-
+    gt: bool = self.priority > alert2.priority
```
geohot, line 138: Do you really need this variable here? What other type would a compare be?

selfdrive/controls/lib/vehicle_model.py
```diff
@@ -163,10 +163,11 @@ def kin_ss_sol(sa: float, u: float, VM: VehicleModel) -> np.ndarray:
   Returns:
     2x1 matrix with steady state solution
   """
-  K = np.zeros((2, 1))
+  K: np.ndarray = np.zeros((2, 1))
   K[0, 0] = VM.aR / VM.sR / VM.l * u
   K[1, 0] = 1. / VM.sR / VM.l * u
-  return K * sa
+  K_scaled: np.ndarray = K * sa
```
geohot, line 169: Same here, it can't infer?

selfdrive/controls/lib/events.py
GregorKikelj: Oh, could use the same strategy as in bootlog file, will fix. Compare can generally be of type Any if the types aren't specified because __lt__ can be overloaded.

selfdrive/controls/lib/vehicle_model.py
GregorKikelj: It really can't :( Not sure why, it knows K is `np.ndarray` and `sa` is float but can't prove that the product is also an `np.ndarray`.

GregorKikelj:
The submodules needed some type hints as well

- [ ] Cereal
- [ ] Panda

tools/lib/bootlog.py
```diff
@@ -36,9 +36,13 @@ def datetime(self) -> datetime.datetime:
     return timestamp_to_datetime(self._timestamp)
 
   def __eq__(self, b) -> bool:
+    if not isinstance(b, Bootlog):
+      return False
```
adeebshihadeh, line 40:
```suggestion
  def __eq__(self, b: Bootlog) -> bool:
```

tools/lib/bootlog.py
```diff
@@ -36,9 +36,13 @@ def datetime(self) -> datetime.datetime:
...
     return self.datetime == b.datetime
 
   def __lt__(self, b) -> bool:
+    if not isinstance(b, Bootlog):
```
adeebshihadeh, line 44: same here

selfdrive/version.py
```diff
@@ -55,7 +55,7 @@ def get_origin(default: Optional[str] = None) -> Optional[str]:
 
 @cache
 def get_normalized_origin(default: Optional[str] = None) -> Optional[str]:
-  origin = get_origin()
+  origin:Optional[str] = get_origin()
```
adeebshihadeh, line 58: missing space

tools/lib/bootlog.py
GregorKikelj: I can't do that, seems like you can't typehint methods like `__x__`

GregorKikelj:
```
tools/lib/bootlog.py:38: error: Argument 1 of "__eq__" is incompatible with supertype "object"; supertype defines the argument type as "object"
tools/lib/bootlog.py:38: note: This violates the Liskov substitution principle
tools/lib/bootlog.py:38: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides
tools/lib/bootlog.py:38: note: It is recommended for "__eq__" to work with arbitrary objects, for example:
tools/lib/bootlog.py:38: note:     def __eq__(self, other: object) -> bool:
tools/lib/bootlog.py:38: note:         if not isinstance(other, Bootlog):
tools/lib/bootlog.py:38: note:             return NotImplemented
tools/lib/bootlog.py:38: note:         return <logic to compare two Bootlog instances>
```

selfdrive/locationd/calibrationd.py
```diff
@@ -48,15 +48,16 @@ class Calibration:
 
 
 def is_calibration_valid(rpy: np.ndarray) -> bool:
-  return (PITCH_LIMITS[0] < rpy[1] < PITCH_LIMITS[1]) and (YAW_LIMITS[0] < rpy[2] < YAW_LIMITS[1])
-
+  is_valid: bool = (PITCH_LIMITS[0] < rpy[1] < PITCH_LIMITS[1]) and (YAW_LIMITS[0] < rpy[2] < YAW_LIMITS[1])
```
geohot, line 51: Until things like this can be inferred I don't want to merge this

selfdrive/locationd/calibrationd.py
```diff
@@ -165,7 +170,7 @@ def handle_cam_odom(self, trans: List[float], rot: List[float], trans_std: List[
     observed_rpy = np.array([0,
                              -np.arctan2(trans[2], trans[0]),
                              np.arctan2(trans[1], trans[0])])
-    new_rpy = euler_from_rot(rot_from_euler(self.get_smooth_rpy()).dot(rot_from_euler(observed_rpy)))
+    new_rpy: np.ndarray = euler_from_rot(rot_from_euler(self.get_smooth_rpy()).dot(rot_from_euler(observed_rpy)))
```
geohot, line 173: There should be no reason this needs a type, isn't the euler_from_rot function typed?

GregorKikelj:
Removing `https://github.com/numpy/numpy-stubs` from precommit_config mypy additional_dependencies either fixes numpy issues or makes the types less strict. Looking at the repo it says that the dependency is already included in numpy and the repo seems to not really get updates so not sure...

selfdrive/athena/registration.py
```diff
@@ -28,7 +28,7 @@ def register(show_spinner=False) -> str:
 
   IMEI = params.get("IMEI", encoding='utf8')
   HardwareSerial = params.get("HardwareSerial", encoding='utf8')
-  dongle_id = params.get("DongleId", encoding='utf8')
+  dongle_id: str = params.get("DongleId", encoding='utf8')
```
herman5, line 31: In #24379 I made `dongle_id`'s type and `register`'s return type `Optional[str]` - that function makes no guarantee to return `str`

https://github.com/commaai/openpilot/blob/06976a7f94d49ea2b82d7163a9603284cac80be4/selfdrive/athena/registration.py#L32

https://github.com/commaai/openpilot/blob/06976a7f94d49ea2b82d7163a9603284cac80be4/selfdrive/athena/registration.py#L26

selfdrive/controls/lib/vehicle_model.py
herman5: @GregorKikelj re: numpy support https://github.com/commaai/openpilot/pull/24379#issuecomment-1113969757

I was able to get numpy type checking enabled by adding `plugins = numpy.typing.mypy_plugin` to [mypy.ini](https://github.com/commaai/openpilot/blob/master/mypy.ini) and adding `'numpy'` to additional_dependencies:

https://github.com/commaai/openpilot/blob/ec7cd74dc37caad9a281f70fa1bfcb12535efeb9/.pre-commit-config.yaml#L23

Note that typing for shapes is still a WIP (https://github.com/numpy/numpy/issues/16544).

selfdrive/locationd/calibrationd.py
```diff
@@ -48,15 +48,15 @@ class Calibration:
...
 def sanity_clip(rpy: np.ndarray) -> np.ndarray:
   if np.isnan(rpy).any():
     rpy = RPY_INIT
-  return np.array([rpy[0],
-                   np.clip(rpy[1], PITCH_LIMITS[0] - .005, PITCH_LIMITS[1] + .005),
-                   np.clip(rpy[2], YAW_LIMITS[0] - .005, YAW_LIMITS[1] + .005)])
+  clipped: np.ndarray = np.array([rpy[0],
+                          np.clip(rpy[1], PITCH_LIMITS[0] - .005, PITCH_LIMITS[1] + .005),
+                          np.clip(rpy[2], YAW_LIMITS[0] - .005, YAW_LIMITS[1] + .005)])
+  return clipped
```
herman5, line 59: With the numpy mypy plugin enabled, this passes on my machine:

```python
import numpy.typing as npt

RPY_INIT: npt.NDArray[np.float64] = np.array([0.0,0.0,0.0])
PITCH_LIMITS: npt.NDArray[np.float64] = np.array([-0.09074112085129739, 0.14907572052989657])
YAW_LIMITS: npt.NDArray[np.float64] = np.array([-0.06912048084718224, 0.06912048084718235])

def sanity_clip(rpy: npt.NDArray[np.float64]) -> npt.NDArray[np.float64]:
  if np.isnan(rpy).any():
    rpy = RPY_INIT
  return np.array([rpy[0],
                   np.clip(rpy[1], PITCH_LIMITS[0] - .005, PITCH_LIMITS[1] + .005),
                   np.clip(rpy[2], YAW_LIMITS[0] - .005, YAW_LIMITS[1] + .005)])
```

selfdrive/locationd/calibrationd.py
```diff
@@ -145,7 +148,8 @@ def handle_v_ego(self, v_ego: float) -> None:
 
   def get_smooth_rpy(self) -> np.ndarray:
     if self.old_rpy_weight > 0:
-      return self.old_rpy_weight * self.old_rpy + (1.0 - self.old_rpy_weight) * self.rpy
+      smoothed_rpy_weight: np.ndarray = self.old_rpy_weight * self.old_rpy + (1.0 - self.old_rpy_weight) * self.rpy
+      return smoothed_rpy_weight
```
herman5, line 152: Similar to https://github.com/commaai/openpilot/pull/24284/files#r862390202
```python
return self.old_rpy_weight * self.old_rpy + (1.0 - self.old_rpy_weight) * self.rpy
```

selfdrive/controls/lib/vehicle_model.py
GregorKikelj: Will check on monday, need to get to correct PC :) Thanks for the hint.

GregorKikelj:
Let's move to 24379

PR #24287, opened 2022-04-21
mypy: enable --warn-unused-ignores

GregorKikelj:
This kind of conflicts with https://github.com/commaai/openpilot/commit/8aaf00c6a41ac4c9c5e56f8a2dfed8591f153b67 but not sure what that commit fixes
@pd0wm


adeebshihadeh:
looks good, can you open the matching laika PR?

tools/replay/can_replay.py
```diff
@@ -83,7 +83,7 @@ def connect():
 
 if __name__ == "__main__":
   if PandaJungle is None:
-    print("\33[31m", "WARNING: cannot connect to jungles. Clone the jungle library to enable support:", "\033[0m")  # type: ignore
```
pd0wm, line 86: I added this on my computer because it said the print was unreachable. Probably because the `panda_jungle` import actually succeeds for me.

GregorKikelj: Bool seems to fix a problem though I've just made a panda_jungle.py file instead of cloning the whole repo.

GregorKikelj: I'd give it an `Optional[PandaJungle]` type, but PandaJungle doesn't really exist if it can't be imported

pd0wm: Tried to think of something clean too, but couldn't make mypy happy. Guess it's kind of a weird pattern.

I'm ok with this bool since it's just some tools file.

GregorKikelj:
The cleanup diff might break Willem's review

adeebshihadeh:
I couldn't reproduce on my machine, even after re-initializing pre-commit. I imagine the discrepancy is due to some dependencies being a different version. 

@pd0wm does this work on your machine after a `pre-commit clean`?

GregorKikelj:
Yeah can't reproduce the import issue here either after clean...

pd0wm:
@adeebshihadeh getting errors now. It's fixed by removing `~/openpilot/panda_jungle`.

```
batman@workstation-eu-willem:~/openpilot$ pre-commit run mypy --all
mypy.....................................................................Failed
- hook id: mypy
- exit code: 1

tools/replay/can_replay.py:17: error: Cannot assign to a type
tools/replay/can_replay.py:17: error: Incompatible types in assignment (expression has type "None", variable has type "Type[PandaJungle]")
tools/replay/can_replay.py:86: error: Statement is unreachable
Found 3 errors in 1 file (checked 427 source files)
```

adeebshihadeh:
Guess mypy really doesn't support that optional import pattern. Not sure why I didn't see this earlier. Added the bool back in bfeaf35d6dd2c29c65b80a70ccafd9948056731a.

PR #24293, opened 2022-04-22
loggerd: switch to v4l encoder

geohot:
Not ready yet. Advantages:
* simpler code, remove closed source dep
* input is VisionBuf, no need for a copy
* reduced latency, OMX polls! (slowly)
* same underlying kernel API as OMX


geohot:
Hmm, seems to use 70 mW more. My guess is because the qcamera is making the big copy now :)

```
comma@tici:/data/openpilot/selfdrive$ hardware/tici/precise_power_measure.py
measuring for 5 seconds at 123hz 3 times
mean 6171.17 std 433.52
mean 6169.16 std 423.70
mean 6171.89 std 468.98
comma@tici:/data/openpilot/selfdrive$ hardware/tici/precise_power_measure.py
measuring for 5 seconds at 123hz 3 times
mean 6237.49 std 437.15
mean 6247.45 std 409.67
mean 6243.15 std 397.05
```

selfdrive/loggerd/v4l_encoder.cc
```diff
@@ -0,0 +1,350 @@
...
+      edata.setSegmentNum(e->segment_num);
+      edata.setFlags(flags);
+      e->pm->send(e->service_name, msg);
+
+      if (env_debug_encoder) {
+        printf("%20s got %6d bytes in buffer %d with flags %8x and ts %lu lat %.2f ms\n", e->filename, bytesused, index, flags, ts, ((event.getLogMonoTime() / 1000)-ts)/1000.);
+      }
+
+      // TODO: writing should be moved to loggerd
+      if (e->write && bytesused != 0) {
```
geohot, line 129: Gotta move this to a different thread.

geohot:
I expect tests to pass now. This is a net loss of 200 lines and should be functionally identical to the OMX version. I'll do the OMX deletion in another PR, but after review and checking that the files are fine I'm okay with merging this.

geohot:
These tests are so good, found a minor bug in the h264 output size. Fingers crossed it passes now.

geohot:
TESTS PASS!

@adeebshihadeh @pd0wm I'm ready to merge into master, any thoughts?

geohot:
Last test failure seems like noise in modeld power draw...

selfdrive/loggerd/v4l_encoder.cc
```diff
@@ -0,0 +1,380 @@
...
+        (e->remuxing ? event.initQRoadEncodeData() : event.initRoadEncodeData()));
+      edata.setData(kj::arrayPtr<capnp::byte>(buf, bytesused));
+      edata.setTimestampEof(ts);
+      edata.setIdx(idx++);
+      edata.setSegmentNum(e->segment_num);
+      edata.setFlags(flags);
+
+      auto words = new kj::Array<capnp::word>(capnp::messageToFlatArray(msg));
+      auto bytes = words->asBytes();
+      e->pm->send(e->service_name, bytes.begin(), bytes.size());
```
pd0wm, line 157: Any reason you're not using the overloaded version?

```suggestion
      e->pm->send(e->service_name, msg);
```

selfdrive/loggerd/v4l_encoder.cc
```diff
@@ -0,0 +1,380 @@
...
+  }
+}
+
+V4LEncoder::V4LEncoder(
+  const char* filename, CameraType type, int in_width, int in_height,
+  int fps, int bitrate, bool h265, int out_width, int out_height, bool write)
+  : type(type), in_width_(in_width), in_height_(in_height),
+    width(out_width), height(out_height), fps(fps),
+    filename(filename), remuxing(!h265), write(write) {
+  fd = open("/dev/video33", O_RDWR|O_NONBLOCK);
```
pd0wm, line 189: Can we open this by name like we do for cameras?

selfdrive/loggerd/v4l_encoder.cc
```diff
@@ -0,0 +1,380 @@
...
+static int dequeue_buffer(int fd, v4l2_buf_type buf_type, unsigned int *index=NULL, unsigned int *bytesused=NULL, unsigned int *flags=NULL, struct timeval *timestamp=NULL) {
+  v4l2_plane plane = {0};
+  v4l2_buffer v4l_buf = {
+    .type = buf_type,
+    .memory = V4L2_MEMORY_USERPTR,
+    .m = { .planes = &plane, },
+    .length = 1,
+  };
+  int ret = ioctl(fd, VIDIOC_DQBUF, &v4l_buf);
+  if (ret == -1 && errno == 11) return ret;
```
pd0wm, line 31:
```suggestion
  if (ret == -1 && errno == EAGAIN) return ret;
```

selfdrive/loggerd/v4l_encoder.cc
geohot: Because I need the Array and not an ArrayPtr so it doesn't free on out of scope.

selfdrive/loggerd/v4l_encoder.cc
geohot: Done

selfdrive/loggerd/v4l_encoder.cc
```diff
@@ -0,0 +1,363 @@
...
+
+#include "selfdrive/loggerd/v4l_encoder.h"
+#include "selfdrive/common/util.h"
+#include "selfdrive/common/timing.h"
+
+#include "libyuv.h"
+#include "selfdrive/loggerd/include/msm_media_info.h"
+
+// has to be in this order
+#include "selfdrive/loggerd/include/v4l2-controls.h"
```
adeebshihadeh, line 13: think this should go in third_party with the linux structure, i.e. `third_party/linux/v4l2-controls.h`. perhaps we should also fix these headers in AGNOS to actually correspond to our kernel

selfdrive/loggerd/v4l_encoder.cc
```diff
@@ -0,0 +1,363 @@
...
+      auto words = new kj::Array<capnp::word>(capnp::messageToFlatArray(msg));
+      auto bytes = words->asBytes();
+      e->pm->send(e->service_name, bytes.begin(), bytes.size());
+      if (e->write) {
+        e->to_write.push(words);
+      } else {
+        delete words;
+      }
+
+      if (env_debug_encoder) {
```
adeebshihadeh, line 151: swaglog already handles this in a consistent way across the codebase. use `LOGPRINT=<level>` to specify the logging level that should be printed

selfdrive/loggerd/SConscript
```diff
@@ -8,8 +8,7 @@ libs = [common, cereal, messaging, visionipc,
 
 src = ['logger.cc', 'loggerd.cc', 'video_writer.cc']
 if arch == "larch64":
-  src += ['omx_encoder.cc']
-  libs += ['OmxCore', 'gsl', 'CB'] + gpucommon
```
adeebshihadeh, line 12: gpucommon import is unused now

adeebshihadeh (commented):
If this doesn't improve anything, let's push 0.8.14 soon and merge this after. Does this run on PC with some modifications?

selfdrive/loggerd/v4l_encoder.cc
geohot: So I don't want to use swaglog, and I do this in camerad also. There's too much volume for the logs.

geohot:
It doesn't run on PC, there's no encoder hardware. But RawLogger still works.

selfdrive/loggerd/v4l_encoder.cc
adeebshihadeh: no, it's actually quite confusing: >= INFO get uploaded to elasticsearch, all go in rlog, and >= ERROR go in qlog. we can make a level that doesn't go to rlogs though. think it's cleaner than introducing a new pattern in each file

geohot: yea, we should do that, perhaps even two levels. but I think that's a separate PR

geohot:
I would like to merge this very soon, as it's a prereq for:
* encoderd
* NV12

PR #24294, opened 2022-04-22
LatControlTorque: fix deadzone and no steer saturated warning

sshane:
Missing steer saturated due to adding friction after the clip, deadzone possible if friction is maxed out and PID wants to lower torque.

I believe we can't just clip the output after adding friction though, because then if we're at max PID torque (1.0) along with adding some torque from friction, and PID wants to lower torque, then it has to go through the friction component range before it becomes unclipped/unsaturated.

However, the dynamic PID clipping isn't needed if friction isn't adding to the output while PID wants to lower torque, not sure though. @HaraldSchafer what do you think?

https://user-images.githubusercontent.com/25857203/164643380-9eb26dde-eb95-4f69-868d-0c3ba5862098.png


selfdrive/controls/lib/latcontrol_torque.py
```diff
@@ -49,22 +46,27 @@ def update(self, active, CS, CP, VM, params, last_actuators, desired_curvature,
...
 
       setpoint = desired_lateral_accel + LOW_SPEED_FACTOR * desired_curvature
       measurement = actual_lateral_accel + LOW_SPEED_FACTOR * actual_curvature
       error = setpoint - measurement
       pid_log.error = error
 
+      friction_compensation = interp(desired_lateral_jerk, [-JERK_THRESHOLD, JERK_THRESHOLD], [-self.friction, self.friction])
+      # avoid deadzone near saturation
+      self.pid.pos_limit = self.steer_max - friction_compensation
+      self.pid.neg_limit = -self.steer_max - friction_compensation
```
sshane, line 62: This isn't needed if we think the impact of the deadzone at full torque isn't substantial

PR #24296, opened 2022-04-22
Ford: use CADS radar interface

incognitojam:
The Focus Mk4 (and I presume Escape too) seems to use the CADS radar instead of whatever radar was used in the Fusion. I just looked at the messages in cabana and verified that the message IDs for the the Fusion radar weren't there, and that these ones were.

Thanks very much to @ReFil who wrote this :smile: I'm going to test this out today

### DIFF
--- a/selfdrive/car/ford/radar_interface.py
+++ b/selfdrive/car/ford/radar_interface.py
@@ -1,13 +1,12 @@
 #!/usr/bin/env python3
-from math import sin
+from math import cos, sin
 from cereal import car
 from opendbc.can.parser import CANParser
 from selfdrive.car.ford.values import CANBUS, DBC
 from selfdrive.car.interfaces import RadarInterfaceBase
 
-RADAR_MSGS = list(range(0x120, 0x15F + 1))  # 64 points
-LAST_MSG = max(RADAR_MSGS)
-NUM_MSGS = len(RADAR_MSGS)
+RADAR_START_ADDR = 0x120
+RADAR_MSG_COUNT = 64
 
 
 def _create_radar_can_parser(CP):
@@ -17,14 +16,15 @@ def _create_radar_can_parser(CP):
   signals = []
   checks = []
 
-  for ii in range(1, NUM_MSGS + 1):
-    msg = f"MRR_Detection_{ii:03d}"
+  for i in range(1, RADAR_MSG_COUNT + 1):
+    msg = f"MRR_Detection_{i:03d}"
     signals += [
-      (f"CAN_DET_VALID_LEVEL_{ii:02d}", msg),
-      (f"CAN_DET_RANGE_{ii:02d}", msg),
-      (f"CAN_DET_AZIMUTH_{ii:02d}", msg),
-      (f"CAN_DET_RANGE_RATE_{ii:02d}", msg),
-      (f"CAN_DET_AMPLITUDE_{ii:02d}", msg),
+      (f"CAN_DET_VALID_LEVEL_{i:02d}", msg),
+      (f"CAN_DET_AZIMUTH_{i:02d}", msg),
+      (f"CAN_DET_RANGE_{i:02d}", msg),
+      (f"CAN_DET_RANGE_RATE_{i:02d}", msg),
+      (f"CAN_DET_AMPLITUDE_{i:02d}", msg),
+      (f"CAN_SCAN_INDEX_2LSB_{i:02d}", msg),
     ]
     checks += [(msg, 20)]
 
@@ -34,7 +34,7 @@ class RadarInterface(RadarInterfaceBase):
   def __init__(self, CP):
     super().__init__(CP)
     self.updated_messages = set()
-    self.trigger_msg = LAST_MSG
+    self.trigger_msg = RADAR_START_ADDR + RADAR_MSG_COUNT - 1
     self.track_id = 0
 
     self.rcp = _create_radar_can_parser(CP)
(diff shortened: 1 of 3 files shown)


selfdrive/car/ford/radar_interface.py
```diff
@@ -8,27 +8,33 @@
...
-                     RADAR_MSGS * 4,
-                     [0] * msg_n + [0] * msg_n + [0] * msg_n + [0] * msg_n))
-  checks = list(zip(RADAR_MSGS, [20]*msg_n))
+  signals = []
+  checks = []
 
-  return CANParser(DBC[car_fingerprint]['radar'], signals, checks, CANBUS.radar)
+  for addr in RADAR_MSGS:
+    msg = f"MRR_Detection_{addr:03d}"
+    signals += [
```
ReFil, line 20: need to add CAN_DET_AMPLITUDE here

selfdrive/car/ford/radar_interface.py
```diff
@@ -8,27 +8,33 @@
 RADAR_MSGS = list(range(0x120, 0x12F))
```
ReFil, line 8: CADS does more than 16 points, seb and I nerfed it to 16 points in order to try and solve the amplitude issue, I believe it's a 64 point radar but cant remember off the top of my head,

selfdrive/car/ford/radar_interface.py
```diff
@@ -8,27 +8,33 @@
...
+    checks += [(msg, 20)]
+
+  return CANParser(DBC[CP.carFingerprint]['radar'], signals, checks, CANBUS.radar)
 
 class RadarInterface(RadarInterfaceBase):
   def __init__(self, CP):
     super().__init__(CP)
-    self.validCnt = {key: 0 for key in RADAR_MSGS}
+    self.updated_messages = set()
+    self.trigger_msg = 0x12E
```
ReFil, line 34: trigger message cant be 12E because if the radar hadn't got 15 points it'll never trigger, not sure about the best solution for this, need someone on the comma team to weigh in I reckon

selfdrive/car/ford/radar_interface.py
incognitojam: It does look like it yeah. In cabana I see a continuous range of messages from id 120 to 15F

selfdrive/car/ford/radar_interface.py
incognitojam: done thanks

selfdrive/car/ford/radar_interface.py
incognitojam: won't it just send all 64 messages but VALID_LEVEL will be 0?

selfdrive/car/ford/radar_interface.py
```diff
@@ -1,72 +1,96 @@
 #!/usr/bin/env python3
+from math import sin
 from cereal import car
-from common.conversions import Conversions as CV
 from opendbc.can.parser import CANParser
 from selfdrive.car.ford.values import CANBUS, DBC
 from selfdrive.car.interfaces import RadarInterfaceBase
 
-RADAR_MSGS = list(range(0x500, 0x540))
+RADAR_MSGS = list(range(0x120, 0x15F))  # 64 points
```
incognitojam, line 8: oops, maybe the range should be `range(0x120, 0x15F + 1)` to include the last message...

incognitojam:
`86d00e12925f4df7|2022-04-22--20-27-55--4` (looks like segment 3 didn't upload correctly)
Screenshot from 2022-04-23 00-54-59
Screenshot from 2022-04-23 00-58-03

Seems quite good, but can be really noisy with lots of points for different static objects. @ReFil suggested filtering the points using `CAN_DET_AMPLITUDE`

selfdrive/car/ford/radar_interface.py
ReFil: Does the radar continually send all 64 messages? I thought it only sent messages for the points it tracked. I haven't got mine with me otherwise I'd hook it to my can tool and check

ReFil:
https://docbox.etsi.org/ERM/ERMTGSRR/05-CONTRIBUTIONS/2017/ERMTGSRR(17)030005_RCS_of_vehicles_24GHz.pdf 

The RCS values of vehicles are in the range of -12 to 26 dbsm

incognitojam:
Any idea what `CAN_SCAN_INDEX_2LSB` is? I have a feeling it rotates through 0-3 and the "flickering points" actually have a pattern to them

selfdrive/car/ford/radar_interface.py
incognitojam: Screenshot from 2022-04-23 10-33-36
I think it sends all or none? It looks like somehow a couple were dropped but almost all of the messages have the same count.

selfdrive/car/ford/radar_interface.py
```diff
@@ -1,72 +1,99 @@
...
+      if valid:
+        rel_distance = msg[f"CAN_DET_RANGE_{i:02d}"]  # m
+        azimuth = msg[f"CAN_DET_AZIMUTH_{i:02d}"]  # rad
+
+        self.pts[ii].dRel = rel_distance  # m from front of car
+        self.pts[ii].yRel = sin(-azimuth) * rel_distance  # in car frame's y axis, left is positive
+        self.pts[ii].vRel = msg[f"CAN_DET_RANGE_RATE_{i:02d}"]  # m/s relative velocity
+
+        # use aRel for debugging AMPLITUDE (reflection size)
+        self.pts[ii].aRel = msg[f"CAN_DET_AMPLITUDE_{i:02d}"]  # dBsm
```
incognitojam, line 90: I'm not sure where it's gone wrong but `aRel` always seems to be zero when it gets to `tools/replay/ui.py`...

ReFil: There is a limit to its range in cereal

incognitojam:
image

incognitojam:
image
image

Filtering the points with `0 < amplitude <= 15` reduces a lot of the misc points we saw earlier

adeebshihadeh:
Can you keep the old radar parser around too?

incognitojam:
I think this is ready for review now, to make sure I'm supporting both radars in a good way

Is there a better name to use than "fusion radar"?

ReFil:
ford Fusion radar is the Delphi ESR I believe

Sent from Proton Mail mobile

-------- Original Message --------

incognitojam:
renamed them to `DELPHI_ESR` and `DELPHI_MRR` :smile:

adeebshihadeh:
Got a test route for me?

incognitojam:
Sorry I didn't have one ready and I haven't been able to drive all week - hopefully will have one today!

incognitojam:
I think the CI fail was something to do with loading a demo route?

incognitojam:
@adeebshihadeh `86d00e12925f4df7|2022-08-11--16-45-58--28`

PR #24298, opened 2022-04-22
UI: remove RGB buffer usage

jwooning:


BrettLynch:
Hi @jwooning I have been experiencing issues with latest master-ci today. It works fine for quiet awhile but then the whole screen freezes except the current speed and and adjusted speed can change. Everything keeps working fine and op steers and breaks fine just the screen freezes (ahead view doesn't change and neither does lane lines or path). I was wondering if this change might be the cause.

geohot:
Very interested in why this is broken.

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -276,48 +287,10 @@ void CameraViewWidget::vipcThread() {
...
-        wait_fence.reset(new WaitFence());
-
-        // Ensure the fence is in the GPU command queue, or waiting on it might block
-        // https://www.khronos.org/opengl/wiki/Sync_Object#Flushing_and_contexts
-        glFlush();
-
-        latest_texture_id = buf->idx;
-      }
-      // Schedule update. update() will be invoked on the gui thread.
-      QMetaObject::invokeMethod(this, "update");
```
geohot, line 318: Haha, this is the issue

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -209,13 +228,21 @@ void CameraViewWidget::paintGL() {
...
   glBindVertexArray(frame_vao);
-  glActiveTexture(GL_TEXTURE0);
-  glBindTexture(GL_TEXTURE_2D, texture[latest_texture_id]->frame_tex);
 
   glUseProgram(program->programId());
-  glUniform1i(program->uniformLocation("uTexture"), 0);
-  glUniformMatrix4fv(program->uniformLocation("uTransform"), 1, GL_TRUE, frame_mat.v);
+  uint8_t *address[3] = {latest_frame->y, latest_frame->u, latest_frame->v};
+  for (int i = 0; i < 3; ++i) {
+    glActiveTexture(GL_TEXTURE0 + i);
```
geohot, line 235: Bug seems to be subtle GL stuff. Why does the active texture need to change?

BrettLynch123:
Just wanted to let you know it does work fine for about 30 minutes to an hour for me. I'm planning on testing it today with this change reverted to make sure it was the cause.

PR #24300, opened 2022-04-22
more comma two cleanup

adeebshihadeh:


incognitojam:
Does this effectively drop support for uno/black panda?

adeebshihadeh:
No, uno is already unsupported since it doesn't make sense to use one standalone. As for black panda, this just removes the power monitoring support for EONs. Black panda is still supported for multi-panda uses cases.

Issue #24136, opened 2022-04-05
Hyundai: move cars out of legacy safety mode


sshane:
These cars seem to have the correct checksums and counters and can be removed from the [LEGACY_SAFETY_MODE_CAR](https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/values.py#L1132).
- [ ] HYUNDAI SONATA 2019
- [ ] KIA OPTIMA SX 2019 & 2016

adeebshihadeh:
Let's just check all the cars and be more strict about this when merging car ports. Perhaps add it to the car port and fingerprint verification script.

adeebshihadeh:
@sshane is this done?

PR #24306, opened 2022-04-24
remove Kona EV from LEGACY_SAFETY_MODE_CAR Update values.py

Azenzic:
remove the CAR.KONA_EV from the LEGACY_SAFETY_MODE_CAR
616f048bf31258b7|2022-04-23--20-38-39--3
my car is a 2021 Hyundai Kona EV guess we should look at expanding support model years too


Azenzic:
also I need to add DisableRadar_Allow in the params/d folder, but I think that is a safety thing? as using OP long disables my FCW.

adeebshihadeh:
Updated the model years in d9683b63dfd40e1fde3a07df62a380c798cae5b6. In order to move cars into the normal safety mode, we'll need to verify on our end that no cars will regress. This likely works fine on your car, but may not a 2018 Kona EV for example. We're tracking this in  #24136, and in the meantime are adding longitudinal support to the legacy safety mode in https://github.com/commaai/panda/pull/904/.

PR #24307, opened 2022-04-24
Hyundai: remove mismatched video

Azenzic:
update fingerprint to show support for Kona EV 2021 also the Kona EV 2021 uses the Hyundai G harness ( the same as the 2019)
616f048bf31258b7|2022-04-23--20-38-39
616f048bf31258b7|2022-04-23--20-38-39--4
remove improper youtube link that show a Hyundai Elantra driving listed next to a Kona hybrid


selfdrive/car/hyundai/values.py
```diff
@@ -42,7 +42,7 @@ class CAR:
   IONIQ_PHEV_2019 = "HYUNDAI IONIQ PLUG-IN HYBRID 2019"
   IONIQ_PHEV = "HYUNDAI IONIQ PHEV 2020"
   KONA = "HYUNDAI KONA 2020"
-  KONA_EV = "HYUNDAI KONA ELECTRIC 2019"
+  KONA_EV = "HYUNDAI KONA ELECTRIC 2018-2021"
```
sshane, line 45: You can update the car info below, but this should be kept the same

PR #24310, opened 2022-04-24
CI: ensure submodules have the same linter config

adeebshihadeh:


adeebshihadeh:
On second thought, I think this will be more annoying than helpful.

PR #24314, opened 2022-04-25
Discrete controller (discrete PID, no FF, no integrator windup via recalculation/last_actuators)

ClockeNessMnstr:
Uses last actuators to limit the integrator a bit better than https://github.com/commaai/openpilot/pull/23414 without needing any additional value changes / tuning. 
linearizes the error by speed to perform gain scheduling so only a single value is needed for P,I,D
Does not rely on or benefit from the use of FF. Integrator can handle what FF is used for currently and rate limits on the actuators make FF not useful since FF is intended to speed up the controllers reaction to a setpoint change. 
integrator(s) do not wind up they are recalculated based on the last_actuators. My current tuning I = ~1.5x P

Entire PID controller is actually defined in 3 lines. 
    N = 20 # Filter coefficient. corner frequency in rad/s. 20 = ~3.18hz
    Z = [[[1, 1], [2, -2]], [[1], [1]], [[2, -2], [1-2j, 1+2j]]] # Trapezoidal IPD
    T = [[[1, 0], [1    ]], [[1], [1]], [[1    ], [1   , (1/N)*1j]]] # Trapezoidal IPD

This would be PI only by changing those lines:
    Z = [[[1, 1], [2, -2]], [[1], [1]]] # Trapezoidal IP
    T = [[[1, 0], [1     ]], [[1], [1]]] # Trapezoidal IP
    
Useful on it's own as an upgraded PID controller for Lat but also could be a better controller for use in the new torque control to eliminate the badly behaved integrator in the stock system.

### DIFF
--- a/selfdrive/car/tests/test_car_interfaces.py
+++ b/selfdrive/car/tests/test_car_interfaces.py
@@ -42,6 +42,8 @@ class TestCarInterfaces(unittest.TestCase):
         self.assertTrue(car_params.lateralTuning.torque.kf > 0)
       elif tuning == 'indi':
         self.assertTrue(len(car_params.lateralTuning.indi.outerLoopGainV))
+      elif tuning == 'discrete':
+        self.assertTrue(car_params.lateralTuning.discrete)
 
     # Run car interface
     CC = car.CarControl.new_message()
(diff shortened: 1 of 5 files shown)


HaraldSchafer:
I'm really confused what problem this solves. The whole point of latcontrol_torque was to simplify the controllers, remove controllers, increase FF and reduce tuning. This controller appears to be doing the opposite.

If there are issues with integrator behavior in the normal PID we should just fix them. An issue with plotjuggler plots showing any issues would be ideal

PR #24315, opened 2022-04-25
EKF paramsd tuning for convergence.

ClockeNessMnstr:
Currently the EKF/params learner isn't converging. STD's should not be increasing constantly. 
covariance Q values for steering angle, speed, and roll are very low implying that they do not differ from the model, however these three are not modeled at all and should have a much higher co-variance. 
I'm not an expert in tuning these but was able to get good results and stable STD's showing that the EKF was converging on it's own. 

I still need to get stock working but for some reason FPV2 is not working when I tried setting up a branch without all of the subaru changes. I may be stuck until .8.14 is released and mlp has a new branch with merge conflicts sorted. There were a ton of conflicts when I was trying to find out what I needed,  but I will try again to get a cleaner branch for some good before and after logs. 

The issue however is clear in the below comments in the code, both of which should be resolved. 

'# We observe the current stiffness and steer ratio (with a high observation noise) to bound
'# the respective estimate STD. Otherwise the STDs keep increasing, causing rapid changes in the
'# states in longer routes (especially straight stretches).

'# Reset time when stopped so uncertainty doesn't grow

I'm also assuming this would need to go through some extensive testing to make sure nothing breaks as well as to get better values than what I've found for Q/P. But by letting the model be more unsure about where the steering angle, speed, and roll should be after not getting observations the innovation is not creating large swings in the other values to counter the change in value when resuming observations of speed/roll/yaw/steering angle.

### DIFF
--- a/selfdrive/locationd/models/car_kf.py
+++ b/selfdrive/locationd/models/car_kf.py
@@ -63,10 +63,10 @@ class CarKalman(KalmanFilter):
     math.radians(0.02)**2,
     math.radians(0.25)**2,
 
-    .1**2, .01**2,
+    6**2, .03**2,
     math.radians(0.1)**2,
-    math.radians(0.1)**2,
-    math.radians(1)**2,
+    math.radians(12)**2,
+    math.radians(3)**2,
   ])
 
   P_initial = np.diag([
--- a/selfdrive/locationd/models/car_kf.py
+++ b/selfdrive/locationd/models/car_kf.py
@@ -71,7 +71,7 @@ class CarKalman(KalmanFilter):
 
   P_initial = np.diag([
     (2 / 100)**2,
-    .01**2,
+    .1**2,
     math.radians(0.08)**2,
     math.radians(0.25)**2,
 
--- a/selfdrive/locationd/models/car_kf.py
+++ b/selfdrive/locationd/models/car_kf.py
@@ -58,7 +58,7 @@ class CarKalman(KalmanFilter):
 
   # process noise
   Q = np.diag([
-    (.05 / 100)**2,
+    (.005 / 100)**2,
     .01**2,
     math.radians(0.02)**2,
     math.radians(0.25)**2,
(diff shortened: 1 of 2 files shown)


ClockeNessMnstr:
63c983aefcc56c48|2022-04-25--13-41-36 -- "stock"
63c983aefcc56c48|2022-04-25--14-00-05 -- "removed: fake observations + rewind/time reset + stiffness reset"
63c983aefcc56c48|2022-04-25--14-21-35 -- "update with this PR changes to Q / P"


last route is actually learning SR correctly. got to 13.79 instead of floating up above 16 where it started from the previous route. 
will have to keep driving more when I'm free to show where the values are converging to. 

Screenshot from 2022-04-25 14-58-16

ClockeNessMnstr:
Screenshot from 2022-04-27 01-26-33

spec is 13.5 for SR. 
will give it a longer route sometime to make sure the value is really stable but it appears to be getting it about right here.

ClockeNessMnstr:
was learning the SR too low in the rain today. stds were behaving still but may need to learn front and rear stiffness if they need to differ in the model.

pd0wm:
@nuwandavek Any feedback on this?

HaraldSchafer:
Tuning changes to optimizers like vehicle model learner or the localizer take a massive amount of verification work on hundreds of thousands of segments on a diverse set of cars. Edge case failures can completely break driving. And in it's current state the vehicle model performs quite well.

We generally don't make changes to them unless there is a clear problem that affects driving that we are are trying to solve.

ClockeNessMnstr:
The problems are described here in the comments of the current code. 
The first one in particular is describing the EKF not working. 
"observing" the estimated states is not valid. 
and we should be very uncertain of the states like steering angle and velocity

"observing" the estimated states is not valid. 
```
       # We observe the current stiffness and steer ratio (with a high observation noise) to bound
        # the respective estimate STD. Otherwise the STDs keep increasing, causing rapid changes in the
        # states in longer routes (especially straight stretches).
        stiffness = float(self.kf.x[States.STIFFNESS])
        steer_ratio = float(self.kf.x[States.STEER_RATIO])
        self.kf.predict_and_observe(t, ObservationKind.STIFFNESS, np.array([[stiffness]]))
        self.kf.predict_and_observe(t, ObservationKind.STEER_RATIO, np.array([[steer_ratio]]))
```
and we should be very uncertain of the states like steering angle and velocity after not making observations.
being "certain" of those states just pushes innovations into other states we can't observe. 
```
    if not self.active:
      # Reset time when stopped so uncertainty doesn't grow
      self.kf.filter.set_filter_time(t)
      self.kf.filter.reset_rewind()
```

PR #24318, opened 2022-04-25
UI: remove memcpy on frame receive

jwooning:
continuation from reverted #24298


jwooning:
using watch3 code to display 9 camera streams simultaneously the ui freezes very quickly, 3761967d3d24e01c68039dfed534a71fd11a5109 seems to fix this

pd0wm:
The problem with the original connect is that there are only 3 arguments. The version with 4 arguments properly sets `this` as the context for the slot. This influences in which thread the `update()` is run. With the lambda function the `update()` runs from the same thread as `vipcThread`, after the fix it runs in the main gui thread.

pd0wm:
Both the lock and the fence can be removed now too.

PR #24319, opened 2022-04-25
EKF paramsd sign cleanup

ClockeNessMnstr:
This is probably noted already but the signs are backwards from convention in paramsd. 

currently yaw_rate and steering angle are observed opposite of convention and roll is opposite of convention in the model.

This "corrects" the signs and keeps angle offset the same sign. This should align with the main localizer + NED convention. 

Behavior is already correct and not modified here. The params learner is just in it's own mirror universe right now. 
I.e. curently,
locationd: NED
paramsd: NWD
VM dyn_matrices: NEU?


pd0wm:
@nuwandavek can you make a decision on this?

nuwandavek:
I like the idea of having the same axis-conventions, but this needs quite a bit of changes to our internal research codebase for replaying segments and aggregate tests. I'll revisit this at some point (in the near future) and make those changes, but cannot merge this now, sorry!

PR #24320, opened 2022-04-25
Simulator: Fix keyboard input 

gijskoning:
Keyboard input was broken after Simulator CI pr. Thanks Ebadi for noticing it https://github.com/commaai/openpilot/pull/24274#issuecomment-1108531350.

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -12,16 +12,15 @@ import carla  # pylint: disable=import-error
 import numpy as np
 import pyopencl as cl
 import pyopencl.array as cl_array
+from cereal.visionipc.visionipc_pyx import VisionIpcServer, VisionStreamType  # pylint: disable=no-name-in-module, import-error
 
 import cereal.messaging as messaging
 from cereal import log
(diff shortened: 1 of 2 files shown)


tools/sim/bridge.py
```diff
@@ -12,16 +12,15 @@
 import numpy as np
 import pyopencl as cl
 import pyopencl.array as cl_array
+from cereal.visionipc.visionipc_pyx import VisionIpcServer, VisionStreamType  # pylint: disable=no-name-in-module, import-error
```
pd0wm, line 15: why was this moved?

ebadi:
Thanks @gijskoning   for quickly fixing it. I confirm that the keyboard is working now but I still don't see the camera stream in the openpilot window and modeld has not been started.
I don't think your changes caused this but I just wonder how come tests in #24274 actually don't catch this.

gijskoning:
Are you sure you executed all the steps? Running the scripts in tools/sim folder: start_carla.sh, launch_openpilot.sh and bridge.py allows me to run the simulator easily.
You could also try to run bridge in low quality mode using `./bridge.py --low_quality`.
If this doesn't fix your issue it is best to create a new issue with some more details.

PR #24224, opened 2022-04-16
Subaru: remove A/C fw version

martinl:
This PR removes Subaru A/C ECU (0x7c4) fw version added with 3ff14adbd67cb003506e735a1608e44b5f6b275d

0x7c4 is Hyundai fwdCamera ECU address (https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/values.py#L235) which is why it also shows up in subaru carFw as second fwdCamera entry. There similar case with fwdRadar entry (0x7d0). Both entries should be ignored when adding Subaru FW values.

Here is a modified fw_versions.py scan (0xf197) which shows ecu descriptions for 2018 Crosstrek

```
Found ECU descriptions
{
  (Ecu.unknown, 0x7d0, None): [b'\xf1\x00\x00\x00\x02']
  (Ecu.unknown, 0x7e0, None): [b'2.0 DOHC                        ']
  (Ecu.unknown, 0x7c4, None): [b'Air Condition System            ']
  (Ecu.unknown, 0x7e1, None): [b'CVT                             ']
  (Ecu.unknown, 0x746, None): [b'Power Steering System           ']
  (Ecu.unknown, 0x787, None): [b'EyeSight System                 ']
  (Ecu.unknown, 0x7b0, None): [b'VDC/Parking Brake System        ']
  (Ecu.unknown, 0x780, None): [b'Airbag System\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f']
  (Ecu.unknown, 0x735, None): [b'Keyless Entry System            ']
  (Ecu.unknown, 0x747, None): [b'Automatic Headlight Leveling    ']
  (Ecu.unknown, 0x753, None): [b'Tire pressure monitor           ']
  (Ecu.unknown, 0x752, None): [b'\xf1\x00 \x80\x86']
  (Ecu.unknown, 0x783, None): [b'METER                           ']
  (Ecu.unknown, 0x7d5, None): [b'\xf1\x000\x00\x00']
  (Ecu.unknown, 0x7f1, None): [b'Airbag System\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f\x7f']
  (Ecu.unknown, 0x750, 0xec): [b'\xf1\x00\x00\x00\x00']
}
```
[...]

PR #24321, opened 2022-04-25
Subaru: Remove non-adas FPv2 value

ClockeNessMnstr:
This isn't a camera. We don't use these values. We also get something under radar but don't have that either.


sshane:
Duplicate of: https://github.com/commaai/openpilot/pull/24224

ClockeNessMnstr:
👍

PR #24324, opened 2022-04-25
Toyota: fill stockFcw signal

sshane:


cydia2020:
I think this can be extended to all cars with `not self.CP.enableDsu`, are we able to confirm this?

PR #24325, opened 2022-04-25
Remove tag when NVMe drive detected

incognitojam:
Delete the `living-in-the-moment` tag when we detect a drive mounted at `/data/media`

### DIFF
--- a/selfdrive/thermald/thermald.py
+++ b/selfdrive/thermald/thermald.py
@@ -293,7 +293,9 @@ def thermald_thread(end_event, hw_queue):
       media = Path("/data/media")
 
       if living_in_the_moment.is_file() and media.is_mount():
+        os.system("sudo mount -o remount,rw /persist")
         living_in_the_moment.unlink()
+        os.system("sudo mount -o remount /persist")
 
       if not living_in_the_moment.is_file():
         if not media.is_mount():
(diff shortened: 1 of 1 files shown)


incognitojam:
Should we show an offroad alert when we detect a new drive is installed?

adeebshihadeh:
AGNOS mounts `/persist` as read-only, so you'll also need to remount before and after.

incognitojam:
Just went through this on my Comma 3 as I just installed an NVMe drive. I had to run `mkfs.ext4 /dev/nvme0n1` before I could run `mount /data/media` (before that it complained about a bad filesystem). Would we want to format new drives automatically too?

pd0wm:
@adeebshihadeh I assume we don't want to automatically erase?

A system reset should take care of reformatting the name already: https://github.com/commaai/openpilot/blob/master/selfdrive/ui/qt/setup/reset.cc#L16

adeebshihadeh:
No, I don’t think we do. I’ll clean this up and test it. There may also be a permissions issue here.

incognitojam:
I removed the mount/reformat thing, and I suppose indicating to the user whether there is an NVMe drive mounted/formatted can be tackled in another PR

sshane:
@adeebshihadeh did you ever test this?

adeebshihadeh:
After thinking about this some more, I'm not sure it makes much sense. With this change, if you remove the NVMe drive out of your basic edition, you also have to add back a weird file to get the stock experience of the device. Installing an NVMe is an intentional process, and I think it makes more sense to manually remove the file at that time if you want that configuration.

PR #24326, opened 2022-04-25
Don't build OpenCV tests in webcam docker image

incognitojam:
Building opencv tests and apps are [on by default](https://docs.opencv.org/4.x/db/d05/tutorial_config_reference.html#tutorial_config_reference_general_tests), but I'm not sure they are needed.

This saves about 30 minutes when building the webcam docker image!


adeebshihadeh:
Nice find!

PR #24330, opened 2022-04-26
onnxruntime: initialize CUDA provider before predictions

sshane:
The first prediction using the ONNXRuntime takes upwards of 3 seconds, and it does this for each process that utilizes onnx_runner.py, so use 5 seconds to be safe. Also, if `dmonitoringmodeld` is started up later, it pauses modeld! So initialize both at once. This resolves `commIssue`s during startup for regen


sshane:
@mitchellgoffpc can you confirm this doesn't affect any of your stacks?

PR #24332, opened 2022-04-26
regen: quit when fake pandad finishes

sshane:
To avoid CAN data looping around causing unexpected behavior

Tried doing this for all daemons but controlsd never enabled


sshane:
Loop around is fine in the case of #24140, we'll just catch the bad rising edge of engagement

PR #24334, opened 2022-04-26
camerad: don't remap everything every time

geohot:


geohot:
The CI failure is...
```
./camerad                            25.28% (41.00%) using less CPU than normal
```

geohot:
@pd0wm Might be worth merging for 0.8.14, cached the config memory like you suggested, big CPU savings!

PR #24335, opened 2022-04-26
UI: draw camera on same frame from modelV2

sshane:
Makes the path and lines stick to the road much more for bumps and turns. Seeing about 2-3 frames of delay from camera to modelV2 packet on PC, 1 to 2 frames of delay on C3 (that means when we get a frame from camerad, modelV2 is 0.05-0.1 seconds behind).

- [x] Fix reversing in replay (to be fair, modeld isn't robust to reversing either)
- [x] fix restarting camerad


selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -292,8 +295,19 @@ void CameraViewWidget::vipcThread() {
       emit vipcThreadConnected(vipc_client.get());
     }
 
-    if (VisionBuf *buf = vipc_client->recv(nullptr, 1000)) {
+    // Wait until camera frame id is behind modelV2, then update once
+    while (meta_main.frame_id >= (*s->sm)["modelV2"].getModelV2().getFrameId()) {
+      qDebug() << "Sleeping";
+      QThread::msleep(5);
```
geohot, line 301: There has to be a better way than polling.

sshane: yeah not happy with this just yet, will try to use the signals available to the onroad widget. But we still probably want to show the latest camera frames if the model timed out

adeebshihadeh:
Let's get this in for 0.8.14. This should be more generic though. We already subclass this for the onroad widget, so we should be able to handle all the model-specific stuff there.

selfdrive/ui/qt/onroad.cc
```diff
@@ -372,19 +372,21 @@ void NvgWindow::drawLead(QPainter &painter, const cereal::ModelDataV2::LeadDataV
 }
 
 void NvgWindow::paintGL() {
+  UIState *s = uiState();
+  const cereal::ModelDataV2::Reader &model = (*s->sm)["modelV2"].getModelV2();
+  CameraViewWidget::draw_frame_id = model.getFrameId();
```
adeebshihadeh, line 377: make a setter function for this

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -244,7 +253,8 @@ void CameraViewWidget::paintGL() {
 
 void CameraViewWidget::vipcConnected(VisionIpcClient *vipc_client) {
   makeCurrent();
-  latest_frame = nullptr;
+  frames.clear();
+  frame_ids.clear();
```
adeebshihadeh, line 257: clear on vipc connect/disconnect

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -214,14 +217,20 @@ void CameraViewWidget::paintGL() {
   glClearColor(bg.redF(), bg.greenF(), bg.blueF(), bg.alphaF());
   glClear(GL_STENCIL_BUFFER_BIT | GL_COLOR_BUFFER_BIT);
 
-  if (latest_frame == nullptr) return;
+  assert(frames.size() == frame_ids.size());
```
adeebshihadeh, line 220: these should just be one thing. perhaps use an std::pair

adeebshihadeh (commented):
I don't think this should ever jump back in time. If modeld lags, CamaerView should fall back to free-running mode. Once modeld recovers, CameraView can wait a frame or two to get back to drawing the model frame. Let's make a test script to see how lags will look.

selfdrive/ui/qt/widgets/cameraview.h
```diff
@@ -19,11 +19,12 @@ class CameraViewWidget : public QOpenGLWidget, protected QOpenGLFunctions {
   ~CameraViewWidget();
   void setStreamType(VisionStreamType type) { stream_type = type; }
   void setBackgroundColor(const QColor &color) { bg = color; }
+  quint32 draw_frame_id;
```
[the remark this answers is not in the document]
sshane: whoops, I registered as a uint8_t last time I tried, reverting

sshane:
Handles cases where:

- modeld lags and UI hasn't gotten a new modelV2 packet in multiple frames: keeps updating camera frames with no skips (with the same offset, doesn't skip to latest)
- When we get a new modelV2 packet after the above happens:
  - If modelV2 frame id is less than latest camera frame drawn (sometimes with replay), we hold on that camera frame until modelV2 catches up
  - If modelV2 frame id is equal to latest camera frame, there are no camera frame skips or stutters where we hold on a frame.

selfdrive/boardd/boardd.cc
```diff
@@ -489,7 +489,7 @@ void panda_state_thread(PubMaster *pm, std::vector<Panda *> pandas, bool spoofin
 void peripheral_control_thread(Panda *panda) {
   util::set_thread_name("boardd_peripheral_control");
 
-  SubMaster sm({"deviceState", "driverCameraState"});
+  SubMaster sm({"deviceState", "driverCameraState"}, {"deviceState", "driverCameraState"});
```
adeebshihadeh, line 492: if this is necessary, this should be in another PR merged before this one

adeebshihadeh: this really shouldn't be necessary though. let's make it the same API as the python where we default to polling on all

sshane:
Think I want to split the polling on modelV2 out from this PR, as that requires extra complexity making sure the UI is responsive offroad and when model isn't alive. It will also make the perceived model output drawn more closely to the camera frame

selfdrive/ui/qt/onroad.cc
```diff
@@ -372,19 +367,21 @@ void NvgWindow::drawLead(QPainter &painter, const cereal::ModelDataV2::LeadDataV
...
+  CameraViewWidget::setFrameId(model.getFrameId());
   CameraViewWidget::paintGL();
 
   QPainter painter(this);
   painter.setRenderHint(QPainter::Antialiasing);
   painter.setPen(Qt::NoPen);
 
-  UIState *s = uiState();
-  if (s->worldObjectsVisible()) {
+  if (true) {
```
sshane, line 379:
```suggestion
  if (s->worldObjectsVisible()) {
```

PR #24337, opened 2022-04-27
Chrysler: clean up CarController

### LINKED ISSUE #24333: Chrysler ACC fault when openpilot cancels cruise
### Describe the bug

When cruise is set below 39mph, or ACC results in vehicle going below 39mph, and the driver touches the accelerator, the ACC system instantly faults.

Branch: master-ci

I will look into this issue, but wanted to open a ticket to reference. Appears there is a logic error when you mix in the accelerator press.

comma three

2022 Chrysler Pacifica Hybrid 

4395faadf0a373f2|2022-04-26--13-42-38--0

### openpilot version

0.8.14b

### Additional info

_No response_

sshane:


selfdrive/car/chrysler/chryslercan.py
```diff
@@ -52,6 +52,6 @@ def create_wheel_buttons(packer, frame, cancel=False):
   # WHEEL_BUTTONS (571) Message sent to cancel ACC.
   values = {
     "ACC_CANCEL": cancel,
-    "COUNTER": frame % 10
+    "COUNTER": frame % 0x10
```
adeebshihadeh, line 55: 0x10 != 10

sshane: the max range is actually 0x10/15

selfdrive/car/chrysler/interface.py
```diff
@@ -69,4 +69,4 @@ def apply(self, c):
     if (self.CS.frame == -1):
```
sshane, line 69: Is this even used? CS.frame is never -1

sshane: ah yeah -1 used to be the default CANParser value until we got CAN data, but now we wait for initialization

PR #24362, opened 2022-04-28
safety param: make uint16_t

sshane:


selfdrive/boardd/boardd.cc
```diff
@@ -171,10 +171,12 @@ bool safety_setter_thread(std::vector<Panda *> pandas) {
       safety_model = cereal::CarParams::SafetyModel::SILENT;
       safety_param = 0U;
     }
+    // FIXME: first two bytes must be empty
+    assert((safety_param >> 16) == 0U);
```
adeebshihadeh, line 175: this should go in `Panda::set_safety_model`

PR #24367, opened 2022-04-29
Support RAV4 2022 with stock longitudinal

sshane:
fp from dongles: dacb0e4af932211e and a5c341bb250ca2f0

- [x] Test route for test_models.py


RELEASES.md
```diff
@@ -19,6 +19,7 @@ Version 0.8.14 (2022-0X-XX)
  * Toyota Alphard Hybrid 2021 support
  * Toyota Avalon Hybrid 2022 support
  * Toyota RAV4 Hybrid 2022 support
+ * Toyota RAV4 2022 support
```
adeebshihadeh, line 22: swap these

PR #24369, opened 2022-04-29
ui: combine OnroadHud into NvgWindow

pd0wm:
drawing OnroadHUD is very expensive and is therefore only done when values are actually changed. However this causes lots of inconsistency in drawing the camera feed. This dependency on the data also makes it hard to benchmark ui performance. Therefore I suggest moving it back into NvgWindow and drawing it every frame (on the GPU).

This might also reduce our uptick in ui restarts. I think those happen when asking weston to draw too many frames in short succession. Combining the separate draw of hud and camera into one should reduce the load on weston.

### DIFF
--- a/selfdrive/ui/qt/widgets/cameraview.cc
+++ b/selfdrive/ui/qt/widgets/cameraview.cc
@@ -253,7 +253,6 @@ void CameraViewWidget::vipcConnected(VisionIpcClient *vipc_client) {
   stream_width = vipc_client->buffers[0].width;
   stream_height = vipc_client->buffers[0].height;
 
-  glPixelStorei(GL_UNPACK_ALIGNMENT, 1);
   for (int i = 0; i < 3; ++i) {
     glBindTexture(GL_TEXTURE_2D, textures[i]);
     glTexParameteri(GL_TEXTURE_2D, GL_TEXTURE_MIN_FILTER, GL_LINEAR);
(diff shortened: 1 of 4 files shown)


jwooning:
Seems to improve the UI timings quite a bit, this is a plot of time differences between consecutive calls to `CameraViewWidget::paintEvent` 

image

jwooning:
I also looked into the timings between `CameraViewWidget::paintEvent` and `OnroadHud::paintEvent`. After moving the UI to the YUV frame stream the time between these events seem to be smaller more consistent than before. Assuming the recent increase in UI restarts comes from this increase in short consecutive repaints, and PR would combine the repaints, this should solve that issue as well.

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -253,7 +253,6 @@ void CameraViewWidget::vipcConnected(VisionIpcClient *vipc_client) {
   stream_width = vipc_client->buffers[0].width;
   stream_height = vipc_client->buffers[0].height;
 
-  glPixelStorei(GL_UNPACK_ALIGNMENT, 1);
```
pd0wm, line 256: @deanlee this breaks text rendering in the OpenGL widget occasionally. Everything seems to work find without changing the alignment. Do you remember why this was added?

pd0wm:
CPU usage in onroad test went up a little bit, but that's just because the HUD is now always drawn and should more accurately reflect the worst-case scenario.

PR #24378, opened 2022-04-30
Fix KIA_NIRO_HEV in CARS.md

sunnyhaibin:
`KIA_NIRO_HEV` should be `Niro Hybrid 2019` in CARS.md, not `Niro Plug-In Hybrid 2019`.


sshane:
We update the CarInfo in values.py and generate this automatically now

PR #24379, opened 2022-04-30
add mypy check for return-any

herman5:
**Description**
Adds `--return-any` support to mypy. Was less complicated after merging numpy support in separately (https://github.com/commaai/openpilot/pull/24387)

**Verification**
`$ pre-commit run --all`
CI


GregorKikelj:
I think we are tying to solve a similar problem, check https://github.com/commaai/openpilot/pull/24284
If you find a good way to deal with numpy, would be awesome.

PR #24380, opened 2022-04-30
loggerd: switch to v4l encoder try 2

geohot:
Added encoderd in a nondestructive way. It doesn't run by default, but it'll be a short PR to enable it. 

Should be good to merge if 0.8.14 is delayed. PR isn't as long as it first seems.

* Bunch of changes to include files
* remote_encoder and encoderd aren't used (yet), though encoderd does work standalone
* loggerd changes also don't do anything (yet)
* Main change is the new ~400 lines in v4l_encoder.cc


geohot:
Hmm, perhaps we should just use the queue. Think tomorrow about how guaranteed this is.

tools/camerastream/compressed_vipc.py
```diff
@@ -26,23 +23,25 @@ def writer(fn, addr, sock_name):
     msgs = messaging.drain_sock(sock, wait_for_one=True)
     for evt in msgs:
       evta = getattr(evt, evt.which())
-      lat = ((evt.logMonoTime/1e9) - (evta.timestampEof/1e6))*1000
-      print("%2d %4d %.3f %.3f latency %.2f ms" % (len(msgs), evta.idx, evt.logMonoTime/1e9, evta.timestampEof/1e6, lat), len(evta.data), sock_name)
-      if evta.idx != 0 and evta.idx != (last_idx+1):
+      lat = ((evt.logMonoTime/1e9) - (evta.idx.timestampEof/1e9))*1000
+      print("%2d %4d %.3f %.3f latency %.2f ms" % (len(msgs), evta.idx.encodeId, evt.logMonoTime/1e9, evta.idx.timestampEof/1e6, lat), len(evta.data), sock_name)
```
gijskoning, line 27: I think evta.idx.timestampEof/1e6 should become evta.idx.timestampEof/1e9

PR #24383, opened 2022-04-30
Body UI: Show battery percentage in bottom-right corner

kfatehi:
Added battery percentage to the carState which required https://github.com/commaai/cereal/pull/290 

I did not retarget the cereal git submodule in this PR.

I did however implement a battery monitor as per the attached image.

IMG_1449

### DIFF
--- a/selfdrive/car/body/carstate.py
+++ b/selfdrive/car/body/carstate.py
@@ -25,6 +25,8 @@ class CarState(CarStateBase):
     ret.cruiseState.enabled = True
     ret.cruiseState.available = True
 
+    ret.batteryPercent = cp.vl["BODY_DATA"]["BATT_PERCENTAGE"]
+
     return ret
 
   @staticmethod
--- a/selfdrive/ui/qt/body.cc
+++ b/selfdrive/ui/qt/body.cc
@@ -1,20 +1,35 @@
-#include "selfdrive/ui/qt/body.h"
-
 #include <cmath>
 
-BodyWindow::BodyWindow(QWidget *parent) : QLabel(parent) {
+#include <QVBoxLayout>
+
+#include "selfdrive/ui/qt/body.h"
+
+BodyWindow::BodyWindow(QWidget *parent) : QWidget(parent) {
+  layout = new QVBoxLayout(this);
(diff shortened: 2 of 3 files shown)


ErichMoraga:
Consider using a 🔋 instead of a ❤️...

kfatehi:
@ErichMoraga I agree with you. Changed it. Thank you

adeebshihadeh:
Nice, I was going to do this soon! Here's the design we have (ignore the arrow). The whole body UI is on a grid system, but I started with GIFs for the face at first for simplicity. Let's break this up into two PR's:
* implementing the grid system in the UI (with both faces)
* adding the battery indicator from this design

image

geohot:
See my comment on cereal, if we are going to add this to carState it should apply to cars also. Don't privilege the body!

kfatehi:
> Nice, I was going to do this soon! Here's the design we have (ignore the arrow). The whole body UI is on a grid system, but I started with GIFs for the face at first for simplicity. Let's break this up into two PR's:
> 
> * implementing the grid system in the UI (with both faces)
> * adding the battery indicator from this design

@adeebshihadeh thank you. I don't have a lot of time (I go back to work on monday and can't really code on side stuff that much when getting drained during the day coding for my job) nor the knowledge/talent in Qt so I would say since you were already planning to work on it maybe it's best to just close this one out. But maybe tomorrow just for fun and learning I will follow your guidance by opening two new PRs -- I have so many questions though so I think the quality will suck compared to what you had in mind, this is what I have in mind at this moment should I attempt to split this into the 2 PRs:

1. Branching off master, changing the body UI to use QGridLayout with both faces loaded. I didn't quite understand if you are trying to move away from using the GIFs or not, but I'll try something out
2. Branching off [the carstate PR](https://github.com/commaai/openpilot/pull/24386) to bring in the battery value into the QGridLayout per the design that you provided. My naive thought is to try and use color styling in combination with the [UTF-8 center dot character](https://www.compart.com/en/unicode/U+00B7), · · · · ·, if it will look like the design or not I do not know yet. If it doesn't then I guess I could use QPainter and make circles with it.

PR #24386, opened 2022-05-01
body: add fuelGauge to body carState

kfatehi:
Depends on https://github.com/commaai/cereal/pull/292

Partial replacement for https://github.com/commaai/openpilot/pull/24383

To explain why I am dividing by 100 in the change:

https://github.com/commaai/cereal/pull/292 introduces a generic
fuelGauge field and defines it as a Float32 in an effort to make this field
generic and capable of handling more precise reporting.

The Comma body, however, is reporting battery percentage as an integer from
zero to one hundred, so this change is dividing by 100 to get the number
into that range. Is this appropriate?

It also does not perform any error checking to know if it's okay to
do this as I'm making the assumption that it's safe. Is this indeed a safe
assumption?

For verification this was tested on a comma body using code from
https://github.com/commaai/openpilot/pull/24383 which renders a battery
level on the display (doing math as needed):

```
QString::number(sm["carState"].getCarState().getFuelGauge()*100)
```

This works fine. I don't like the fact that I'm doing / 100 and
then * 100 but it's probably worth it for maintaining the generic-ness
of the fuelGauge field.


geohot:
You have to push the new cereal in this PR as well.

kfatehi:
rebased away the extra commits, replay test needs attention but looks out of my control

adeebshihadeh:
Yeah, we need to update the process replay references since they get pushed to an internal azure bucket. No need to worry too much about fixing up your branch history since we squash and merge anyway.

PR #24387, opened 2022-05-02
enable numpy mypy typing

herman5:
**Description**
Minimum config to enable numpy typing in mypy

**Verification**
`$ mypy selfdrive/locationd/calibrationd.py --warn-return-any`


adeebshihadeh:
Does this not work? It would be nice to have numpy typing support.

herman5:
> Does this not work? It would be nice to have numpy typing support.

@adeebshihadeh It does work 😄 - I made these suggestions in https://github.com/commaai/openpilot/pull/24284#discussion_r862389793 but wanted to test in isolation. Can reopen this if we'd like to land numpy typing separately.

adeebshihadeh:
Sure, breaking up changes into smaller ones is generally easier to merge.

.pre-commit-config.yaml
```diff
@@ -20,13 +20,11 @@ repos:
...
         exclude: '^(pyextra/)|(cereal/)|(rednose/)|(panda/)|(laika/)|(opendbc/)|(laika_repo/)|(rednose_repo/)/'
-        additional_dependencies: ['git+https://github.com/numpy/numpy-stubs', 'types-requests', 'types-atomicwrites',
-                                  'types-pycurl', 'types-certifi', 'lxml']
+        additional_dependencies: ['lxml', 'numpy', 'types-atomicwrites', types-Jinja2, 'types-pycurl', 'types-requests', 'types-certifi']
         args:
-        - --warn-unused-ignores
         - --warn-redundant-casts
         - --warn-unreachable
-        #- --html-report=/home/batman/openpilot
+        - --warn-unused-ignores
```
adeebshihadeh, line 27: looks like a bad merge here. new stuff on master is overwritten

common/transformations/orientation.py
```diff
@@ -1,6 +1,7 @@
 # pylint: skip-file
 import numpy as np
 
+from typing import Callable
```
adeebshihadeh, line 4: our convention here is std + third party libs then openpilot imports
```suggestion
import numpy as np
from typing import Callable

```

selfdrive/locationd/calibrationd.py
```diff
@@ -48,8 +48,7 @@ class Calibration:
 
 
 def is_calibration_valid(rpy: np.ndarray) -> bool:
-  return (PITCH_LIMITS[0] < rpy[1] < PITCH_LIMITS[1]) and (YAW_LIMITS[0] < rpy[2] < YAW_LIMITS[1])
-
+  return (PITCH_LIMITS[0] < rpy[1] < PITCH_LIMITS[1]) and (YAW_LIMITS[0] < rpy[2] < YAW_LIMITS[1])  # type: ignore
```
adeebshihadeh, line 51: why is this needed?

herman5: It's not needed until `--warn-return-any` is enabled.

common/window.py
```diff
@@ -54,7 +54,7 @@ def getclick(self):
 if __name__ == "__main__":
   import numpy as np
   win = Window(200, 200, double=True)
-  img = np.zeros((200, 200, 3), np.uint8)
+  img: np.ndarray = np.zeros((200, 200, 3), np.uint8)
```
adeebshihadeh, line 57: why can't this be inferred? we don't want to do this all over the codebase

herman5: The `+=` operator on line 60 throws the below error when unsignted int's are used as the dtype (tried uint16, uint32, and uint64). Regular np.int8, np.int16, etc (and floats for that matter) don't have that issue. We could convert the dtype to np.int16 to keep the same number of character support (albeit some extra memory for the negative number space that isn't being used), keep the type specification, or add an ignore statement to line 60. Personal preference would be to change the dypte since this seems like a valid error.

```
$ mypy common/window.py
common/window.py:60: error: Argument 1 to "__iadd__" of "ndarray" has incompatible type "int"; expected "Union[_SupportsArray[dtype[Union[bool_, unsignedinteger[Any]]]], _NestedSequence[_SupportsArray[dtype[Union[bool_, unsignedinteger[Any]]]]], bool, _NestedSequence[bool]]"
```

Relevant CI build:
https://github.com/commaai/openpilot/runs/6251557005?check_suite_focus=true

PR #24388, opened 2022-05-02
WIP: Refactor body face UI to use grid system

kfatehi:
Partially replaces https://github.com/commaai/openpilot/pull/24383

> Nice, I was going to do this soon! Here's the design we have (ignore the arrow). The whole body UI is on a grid system, but I started with GIFs for the face at first for simplicity. Let's break this up into two PR's:
> 
> * implementing the grid system in the UI (with both faces)
> * adding the battery indicator from this design
> 
> image

This PR attempts to port the current body UI into a grid per some kind of architecture like so (accidentally drew less columns and thus the battery is in the wrong spot but you get the idea of how im thinking about it):

grid_face

this PR wont include the battery, it's just there in the drawing, i will make another PR for the battery (currently i have code for it using center dots and it looks almost perfect, but probably better to use qpainter or graphicsscene). I dont know which to use for drawing but i think because it's so simple, i should use qpainter.

### DIFF
--- a/selfdrive/ui/qt/body.cc
+++ b/selfdrive/ui/qt/body.cc
@@ -1,22 +1,37 @@
-#include "selfdrive/ui/qt/body.h"
-
 #include <cmath>
 
-BodyWindow::BodyWindow(QWidget *parent) : QLabel(parent) {
-  awake = new QMovie("../assets/body/awake.gif");
-  awake->setCacheMode(QMovie::CacheAll);
-  sleep = new QMovie("../assets/body/sleep.gif");
-  sleep->setCacheMode(QMovie::CacheAll);
+#include <QGridLayout>
+#include <QPainter>
 
-  QPalette p(Qt::black);
-  setPalette(p);
-  setAutoFillBackground(true);
+#include "selfdrive/ui/qt/body.h"
 
-  setAlignment(Qt::AlignCenter);
+BodyWindow::BodyWindow(QWidget *parent) : QWidget(parent) {
+  layout = new QGridLayout(this);
+  layout->setMargin(0);
+  layout->setSpacing(0);
 
-  setAttribute(Qt::WA_TransparentForMouseEvents, true);
+  //setAttribute(Qt::WA_TransparentForMouseEvents, true);
+
+  battery = new QLabel();
+  battery->setStyleSheet("QLabel { font-size: 200px; }");
+  battery->setAlignment(Qt::AlignTop | Qt::AlignRight);
+
+  layout->addWidget(battery,0,2);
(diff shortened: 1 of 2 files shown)


kfatehi:
i think i made pretty decent progress on this tonight but didnt finish. @adeebshihadeh looking forward to if you think my design is on the right track. 

programmatically you can see i am using qpainter and rendering each eye based on 4x4 matrix brightness value from 0 to 1

this works well, my only concern being how complex animating might be.

i like this because it seems to me that speech and emotion can be converted to this 4x4 matrix for the eyes and mouth to do things independently... e.g. piping a speech synthesizer to the mouth to produce something sort of visual indicator... emotes like winking, surprise, etc, all come down to this matrix output. not sure how transitions/animations would be handled though or if i might be overcomplicating things. but let me know how this implementation differs from what you had in mind!

i also took care that the sidebar shrinking the layout still renders the dots nice and centered.

anyway back to present reality, the current state of the PR renders the following
 
image

planning to revisit it next weekend at the latest!

adeebshihadeh:
No rush. I merged 061b18805e131aeff6582730d1b6d2994ce7f985 for now, but I still want to move to using the grid system for the face. This is way more complicated than I think this needs to be. I think QPainter is the right way to go, but I would keep everything in the same widget. Perhaps keep the grid in an array of QColors or something then you'd just have to come up with a nice way to set the specific pixels.

kfatehi:
> No rush. I merged [061b188](https://github.com/commaai/openpilot/commit/061b18805e131aeff6582730d1b6d2994ce7f985) for now, but I still want to move to using the grid system for the face. This is way more complicated than I think this needs to be. I think QPainter is the right way to go, but I would keep everything in the same widget. Perhaps keep the grid in an array of QColors or something then you'd just have to come up with a nice way to set the specific pixels.

awesome thank you for implementing that! interesting code... i learn a lot every time i read your code...

regarding the implementation design, I am seemingly again confused about what you mean by grid system. Originally I thought QGridLayout split into 3x5 like in my drawing (forgive the incorrect # of columns) but now I seem to be conflating it in my mind with the "dot matrix" (of which there are 3 on the body's face).

I understand your note to not create classes; I think I did that to conceptualize things, but now that I have decent drawing code I can use functions on Body instead.

I did think about using QColors in the 2d array instead of floats but decided on gray scale, but i like the QColor idea.

adeebshihadeh:
Not sure it makes sense to keep working on this. As we've started adding things to the body UI, we've had trouble integrating them within the grid system. The battery has been moved to a more traditional design rather than the one initially merged that used the grid.

kfatehi:
Ok sounds good!

PR #24392, opened 2022-05-02
delay bz2 compression from logging to uploading

pd0wm:

### DIFF
--- a/selfdrive/athena/athenad.py
+++ b/selfdrive/athena/athenad.py
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 import base64
+import bz2
 import hashlib
 import io
 import json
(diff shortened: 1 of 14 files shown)


pd0wm:
There are no tests in the uploader/athena tests that actually verify the expected data is uploaded. This is probably a good PR to add those.

pd0wm:
There are no tests in the uploader/athena tests that actually verify the expected data is uploaded. This is probably a good PR to add those.

selfdrive/athena/athenad.py
```diff
@@ -64,6 +65,13 @@
 
 cur_upload_items: Dict[int, Any] = {}
 
+
+def strip_bz2_extension(fn):
```
adeebshihadeh, line 69:
```suggestion
def strip_bz2_extension(fn: str) -> str:
```

PR #24398, opened 2022-05-02
Chrysler: add ACC fault signal

### LINKED ISSUE #24361: Chrysler: find ACC faulted signal
Right now we don't set this in openpilot, so controlsd just disengages like normal if ACC gets into a faulted state.

When ACC becomes faulted, it just looks like the PCM disengaged; `CRUISE_STATE` goes to 0 and `ACC_STATUS_2` goes to 1, the same values for when main is off, so find something else.

Preserved routes:
- `4395faadf0a373f2|2022-04-26--15-47-07`
- `4395faadf0a373f2|2022-04-26--15-41-55`
- `4395faadf0a373f2|2022-04-26--13-42-38`
- `4395faadf0a373f2|2022-04-28--15-55-16`
- `4395faadf0a373f2|2022-04-28--16-11-59`
- `4395faadf0a373f2|2022-04-28--16-27-39`

sshane:
Fills the CS.cruiseState.available signal, and add events for ACC faults

Car always starts with `ACC_STATUS_2` at 1, then goes to 3 when ACC is available (white icon). And when a fault occurs, `ACC_STATUS_2` goes to 1 for the rest of the drive, with `ACC_FAULTED` rising.

Seems to be a pretty rare occurrence:
```python
Total routes: 5020, with fault: 34
Total disengagements: 5557, causing fault: 25
```

One notable exception is the `JEEP GRAND CHEROKEE V6 2018`, where in each segment where `ACC_FAULTED` rose after gas disengagement, it recovered after 10 frames and the users were able to re-engage later in the drive. With all other cars the fault was permanent and the bit stayed high

Here's the breakdown per car for faults (all 20 Jeep faults recovered except 1):

```python
CHRYSLER PACIFICA HYBRID 2019: 884 disengagements, causing fault: 10
CHRYSLER PACIFICA HYBRID 2018: 858 disengagements, causing fault: 0
CHRYSLER PACIFICA 2018: 694 disengagements, causing fault: 0
JEEP GRAND CHEROKEE V6 2018: 783 disengagements, causing fault: 19
JEEP GRAND CHEROKEE 2019: 267 disengagements, causing fault: 1
CHRYSLER PACIFICA HYBRID 2017: 1867 disengagements, causing fault: 0
CHRYSLER PACIFICA 2020: 204 disengagements, causing fault: 1
```

Some plots: https://imgur.com/a/k8LpyPR

### DIFF
--- a/selfdrive/car/chrysler/carstate.py
+++ b/selfdrive/car/chrysler/carstate.py
@@ -48,8 +48,8 @@ class CarState(CarStateBase):
     ret.steeringRateDeg = cp.vl["STEERING"]["STEERING_RATE"]
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(cp.vl["GEAR"]["PRNDL"], None))
 
-    ret.cruiseState.enabled = cp.vl["ACC_2"]["ACC_STATUS_2"] == 7  # ACC is green
-    ret.cruiseState.available = cp.vl["ACC_2"]["ACC_STATUS_2"] != 1  # ACC is white
+    ret.cruiseState.enabled = cp.vl["ACC_2"]["ACC_STATUS_2"] == 7  # ACC is green.
+    ret.cruiseState.available = ret.cruiseState.enabled  # FIXME: for now same as enabled
     ret.cruiseState.speed = cp.vl["DASHBOARD"]["ACC_SPEED_CONFIG_KPH"] * CV.KPH_TO_MS
     # CRUISE_STATE is a three bit msg, 0 is off, 1 and 2 are Non-ACC mode, 3 and 4 are ACC mode, find if there are other states too
     ret.cruiseState.nonAdaptive = cp.vl["DASHBOARD"]["CRUISE_STATE"] in (1, 2)
--- a/selfdrive/car/chrysler/carstate.py
+++ b/selfdrive/car/chrysler/carstate.py
@@ -94,6 +94,7 @@ class CarState(CarStateBase):
       ("STEERING_RATE", "STEERING"),
       ("TURN_SIGNALS", "STEERING_LEVERS"),
       ("ACC_STATUS_2", "ACC_2"),
+      ("ACC_FAULTED", "ACC_2"),
       ("HIGH_BEAM_FLASH", "STEERING_LEVERS"),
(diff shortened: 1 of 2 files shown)


selfdrive/car/chrysler/carstate.py
```diff
@@ -69,6 +69,7 @@ def update(self, cp, cp_cam):
     self.lkas_counter = cp_cam.vl["LKAS_COMMAND"]["COUNTER"]
     self.lkas_car_model = cp_cam.vl["LKAS_HUD"]["CAR_MODEL"]
     self.lkas_status_ok = cp_cam.vl["LKAS_HEARTBIT"]["LKAS_STATUS_OK"]
+    self.acc_faulted = cp.vl["ACC_2"]["ACC_FAULTED"] == 1
```
adeebshihadeh, line 72: let's add this to carState

sshane: Will do here: https://github.com/commaai/openpilot/pull/24399

selfdrive/car/chrysler/carstate.py
```diff
@@ -48,11 +48,12 @@ def update(self, cp, cp_cam):
     ret.steeringRateDeg = cp.vl["STEERING"]["STEERING_RATE"]
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(cp.vl["GEAR"]["PRNDL"], None))
 
-    ret.cruiseState.enabled = cp.vl["ACC_2"]["ACC_STATUS_2"] == 7  # ACC is green.
-    ret.cruiseState.available = ret.cruiseState.enabled  # FIXME: for now same as enabled
+    ret.cruiseState.enabled = cp.vl["ACC_2"]["ACC_STATUS_2"] == 7  # ACC is green
+    ret.cruiseState.available = cp.vl["ACC_2"]["ACC_STATUS_2"] != 1  # ACC is white
```
sshane, line 52: Will move the available signal change to another PR once I get this tested by a user

sshane:
Will check one more time with a notebook that this bit is never 1 when engaged and will mergew

sshane:
Out of all the above segments, whenever `ACC_FAULTED` is 1 and `ACC_STATUS_2` is 7, `ACC_STATUS_2` always fell within 10 frames, so this signal never rose while engaged

PR #24399, opened 2022-05-03
Add ACC faulted status to carState

sshane:


selfdrive/car/gm/interface.py
```diff
@@ -190,7 +189,7 @@ def _update(self, c):
       events.add(EventName.belowEngageSpeed)
     if ret.cruiseState.standstill:
       events.add(EventName.resumeRequired)
-    if self.CS.pcm_acc_status == AccState.FAULTED:
+    if ret.accFaulted:
```
adeebshihadeh, line 192: let's move this to the common event creation function

PR #24404, opened 2022-05-03
Test gnssUblox message processing

gijskoning:
Test basic functionality of processing ublox messages.
Assert gps and glonass satellites are read and location fix is created on correct location.


selfdrive/locationd/test/test_ublox_processing.py
```diff
@@ -0,0 +1,77 @@
...
+
+def get_gnss_measurements(log_reader):
+  gnss_measurements = []
+  for msg in log_reader:
+    if msg.which() == "ubloxGnss":
+      ublox_msg = msg.ubloxGnss
+      if ublox_msg.which == 'measurementReport':
+        report = ublox_msg.measurementReport
+        if len(report.measurements) > 0:
+          gnss_measurements += [read_raw_ublox(report)]
```
pd0wm, line 20:
```suggestion
          gnss_measurements.append(read_raw_ublox(report))
```

selfdrive/locationd/test/test_ublox_processing.py
```diff
@@ -0,0 +1,77 @@
...
+  return gnss_measurements
+
+
+class TestUbloxProcessing(unittest.TestCase):
+
+  @classmethod
+  def setUpClass(cls):
+    path = Route("4cf7a6ad03080c90|2021-09-29--13-46-36").log_paths()[0]
```
pd0wm, line 28: Does this work in CI? Other tests use `selfdrive.test.openpilotci.get_url` to build an URL to the CI bucket.

selfdrive/locationd/test/test_ublox_processing.py
```diff
@@ -0,0 +1,77 @@
...
+
+  def test_get_fix(self):
+    dog = AstroDog()
+    position_fix_found = 0
+    count_processed_measurements = 0
+    count_corrected_measurements = 0
+    position_fix_found_after_correcting = 0
+
+    pos_ests = []
+    for measurements in self.gnss_measurements[:10]:
```
pd0wm, line 52:
```suggestion
    for measurements in self.gnss_measurements[:NUM_TEST_MEAS]:
```

Use same constant in assertEqual below.

gijskoning: Thanks, changed it. Did also add a comment to note that processed and corrected measurements doesn't always have to be the same.

PR #24410, opened 2022-05-03
Toyota: fill steerFaultPermanent

sshane:
When the delta of commanded torque and user torque is too high for too long, a permanent LKA fault can occur where you don't get steering for the rest of the drive. The behavior before was to engage but silently not steer.

Also get rid of the code in carcontroller, which is moot code

Examples of `LKA_STATE`=17 faults: https://imgur.com/a/CsrJ9QW

And this is what a fault from a high steer rate looks like (most common and temporary): https://imgur.com/a/xoT5JY1


adeebshihadeh:
This is an immediate disable. Let's check how many immediate disengagements this will now cause before merging.

sshane:
Out of 2,931,274 segments, there's 7 occurrences of this kind of fault

PR #24411, opened 2022-05-03
controlsState: log desired curvature and curvature rate

sshane:
we don't log this anywhere and would need to calculate it offline with `get_lag_adjusted_curvature`

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -562,6 +562,8 @@ class Controls:
     if not CC.longActive:
       self.LoC.reset(v_pid=CS.vEgo)
 
+    controlsState = log.ControlsState.new_message()
+
     if not self.joystick_mode:
       # accel PID loop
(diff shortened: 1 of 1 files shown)


selfdrive/controls/controlsd.py
```diff
@@ -576,8 +576,10 @@ def state_control(self, CS):
       actuators.steer, actuators.steeringAngleDeg, lac_log = self.LaC.update(CC.latActive, CS, self.VM, params,
                                                                              self.last_actuators, desired_curvature,
                                                                              desired_curvature_rate, self.sm['liveLocationKalman'])
+      controlsState = log.ControlsState.new_message(desiredCurvature=desired_curvature, desiredCurvatureRate=desired_curvature_rate)
```
sshane, line 579: is this convention okay? I notice we hardly use it (if at all). we can also do `log.ControlsState(desiredCurvature=...)`

PR #24420, opened 2022-05-04
test for comparing debayer results to ref commit

jwooning:


selfdrive/test/process_replay/test_debayer.py
```diff
@@ -0,0 +1,158 @@
...
+from selfdrive.hardware import PC, TICI
+from selfdrive.test.openpilotci import BASE_URL, get_url
+from selfdrive.version import get_commit
+from tools.lib.logreader import LogReader
+from tools.lib.filereader import FileReader
+
+if TICI:
+  os.environ["PYOPENCL_CTX"] = "0"
+
+try:
```
pd0wm, line 18: Maybe put this in a different script we can call from the Jenkinsfile before running the test. Don't like running this on import time.

jwooning:
It seems that the current debayering isn't consistent. The results 1 pixel away from the edge differ for every run by a small amount, not yet sure why.

PR #24422, opened 2022-05-04
fix model.cc for x86 replay and webcam

zorrobyte:
`Hardware::TICI() && Params().getBool("EnableWideCamera");` is used elsewhere in the codebase, removing such breaks x86/webcam, replay (debug/uiview.py), which also breaks comma CTF challenge.

The sign for `bool use_extra_client` was incorrect as well. `selfdrive/modeld/modeld.cc: vipc_client_extra no frame` is spammed with `!main_wide_camera` bool and the lane line/path is not drawn

Updated PR cc @pd0wm 

Now working as expected with this PR
image
image

### DIFF
--- a/selfdrive/modeld/modeld.cc
+++ b/selfdrive/modeld/modeld.cc
@@ -171,8 +171,8 @@ int main(int argc, char **argv) {
     assert(ret == 0);
   }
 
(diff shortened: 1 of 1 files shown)


pd0wm:
The bool is not wrong. You should update the webcam code to broadcast on the wide frame VIPC socket, not the narrow one.

PR #24424, opened 2022-05-04
warning instead assert

lukaspetersson:
defaultdict bug

print warning instead of  asserting

### DIFF
--- a/tools/latencylogger/latency_logger.py
+++ b/tools/latencylogger/latency_logger.py
@@ -140,7 +140,7 @@ def insert_cloudlogs(lr, timestamps, start_times, end_times):
           else:
             failed_inserts += 1
   if latest_controls_frameid == 0:
-    print("Warning, no timestamp in controlsd")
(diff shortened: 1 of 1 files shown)


tools/latencylogger/latency_logger.py
```diff
@@ -138,8 +139,10 @@ def insert_cloudlogs(lr, timestamps, start_times, end_times):
             timestamps[frame_id][service].append((event, time))
           else:
             failed_inserts += 1
-  assert latest_controls_frameid > 0, "No timestamp in controlsd"
-  assert failed_inserts < len(timestamps), "Too many failed cloudlog inserts"
+  if latest_controls_frameid == 0:
+    print("Warning, no timestamp in controlsd")
+  elif failed_inserts > len(timestamps):
+    print("Warning, many cloudlog inserts failed", failed_inserts)
```
adeebshihadeh, line 145: make these more descriptive. these errors have to do with the implementation details. what should the user actually know and do?

PR #24426, opened 2022-05-04
Add azure-storage-blob to dev packages

sshane:
Source code is this folder: https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/storage/azure-storage-blob/azure/storage/blob

Built azure-storage-blob distribution is ~85 kB

azure-common is ~15 kB build

azure-storage-common is ~50 kB built

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -4,7 +4,7 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
-azure-storage-blob = "==2.1.0"
+azure-storage-blob = "~=2.1"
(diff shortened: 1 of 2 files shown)


Pipfile
```diff
@@ -4,6 +4,7 @@ url = "https://pypi.org/simple"
 verify_ssl = true
 
 [dev-packages]
+azure-storage-blob = "==2.1.0"
```
sshane, line 7: Last version where `BlockBlobService` exists, new version is `BlobServiceClient`

PR #24431, opened 2022-05-05
Toyota: add missing FW versions for 2022 Lexus ES 300h

brandon-zimmerman:
Adding 2022 Lexus ES 300h fingerprint


pd0wm:
What is your dongle ID?

brandon-zimmerman:
My dongle id is eebb9787e66670b1.

PR #24414, opened 2022-05-04
process replay: automatically push refs on fail

sshane:
Idea: if a process replay job fails, and the diff is expected, just copy the printed commit to ref_commit and push! 

Todo:
- [x] ~Add azure-cli to docker~ - only used for token generation, can skip this
- [x] Add azure to python env
- [x] Clean up
- [x] Upload!

Quirks (features?):
- Instead of using the pull request HEAD commit, `actions/checkout` merges the PR into the base branch, so the ref_commit that is used is something you don't know until process_replay finishes, meaning you can't: push a commit that has a diff and then immediately push another commit with the previous commit's hash to update refs

PR #24432, opened 2022-05-05
URLFile: raise exception if remote URL doesn't exist when using cache

sshane:
Right now the behavior is assymetrical when you try to use `URLFile` on a link that doesn't exist when you turn on and off `FILEREADER_CACHE`.

Reproduce with:
```python
import os
from tools.lib.url_file import URLFile
os.environ["FILEREADER_CACHE"] = "1"
url_f = URLFile("https://commadataci.blob.core.windows.net/can_be_anything")
print(url_f.read())
```

No cache:
```python
Exception: Error 400 ['Connection: keep-alive'] (https://commadataci.blob.core.windows.net/can_be_anything): b'\xef\xbb\xbf<?xml version="1.0" encoding="utf-8"?><Error><Code>OutOfRangeInput</Code><Message>One of the request inputs is out of range.\nRequestId:08bb607c-601e-015c-402f-60318c000000\nTime:2022-05-05T03:25:38.9140945Z</Message></Error>'
```

With cache:

```python
b''
```

### DIFF
--- a/tools/lib/url_file.py
+++ b/tools/lib/url_file.py
@@ -121,7 +121,7 @@ class URLFile:
         end = self.get_length() - 1
       else:
         end = min(self._pos + ll, self.get_length()) - 1
-      assert end > 0, "No data from {}".format(self._url)
+      assert end > 0, f"Remote file is empty or doesn't exist: {self._url}"
       if self._pos >= end:
         return b""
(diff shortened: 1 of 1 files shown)


sshane:
This is to fix a problem with https://github.com/commaai/openpilot/pull/24414, so if this would break other things that expect an empty response, I can just handle it there and close this

PR #24435, opened 2022-05-05
adding support for macos x64 acados

LoganxDev:
**Description**
macOS with i386 cpu couldn't build acados as it was looking for the Darwin files which is arm arch. Made new check for i386 cpu to use x64 directory.

**Verification** 
Built on macos with intel cpu

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -61,14 +61,22 @@ if arch == "aarch64" and TICI:
 
 USE_WEBCAM = os.getenv("USE_WEBCAM") is not None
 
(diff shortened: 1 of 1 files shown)


LoganxDev:
Still not building okay.

LoganxDev:
Need the lib folder.

pd0wm:
Closing due to inactivity. Feel free to open a new PR. Preferably we add a build script to build universal binaries for both M1 x64. Currently only M1 is supported.

PR #24437, opened 2022-05-05
debayering: fix uninitialized local memory at both local workgroup edge and image edge

jwooning:


jwooning:
Because the pixels at the outer edges of the image do a early return, they don't fill in the pixels at the local edges of the image. Resulting in the kernel program reading from uninitialized data at the edges of the local workgroup next to where the outer edge of the image is. This results in some small random artifacts in the image.

wrong_debayer

PR #24440, opened 2022-05-05
camerad: 20 bit - 2 exposure HDR

pd0wm:
Start doing HDR without any changes to timing

More minimal follow up of https://github.com/commaai/openpilot/pull/23887


selfdrive/camerad/cameras/real_debayer.cl
```diff
@@ -10,66 +10,66 @@ const __constant half3 color_correction_0 = (half3)(1.82717181, -0.31231438, 0.0
...
   int offset = gx % 2;
-  uint major = (uint)source[start + offset] << 4;
-  uint minor = (source[start + 2] >> (4 * offset)) & 0xf;
-  half pv = (half)((major + minor)/4);
-
-  // normalize
-  pv = max((half)0.0, pv - black_level);
-  pv /= (1024.0f - black_level);
-
-  // correct vignetting
```
pd0wm, line 54: Add back once there is a fast implementation

geohot:
Why is
```
{0x3362, 0x0000}, // DC GAIN
```
still in here?

It's disabling "high conversion gain" for the T1 exposure, which we should have.

selfdrive/camerad/cameras/sensor2_i2c.h
```diff
@@ -110,8 +110,8 @@ struct i2c_random_wr_payload init_array_ar0231[] = {
   {0x31E0, 0x0003},
 
   // HDR Settings
-  {0x3082, 0x0004}, // OPERATION_MODE_CTRL
-  {0x3238, 0x0444}, // EXPOSURE_RATIO
+  {0x3082, 0x0004}, // OPERATION_MODE_CTRL (A) (enable HDR + 2 exposures)
+  {0x3238, 0x0004}, // EXPOSURE_RATIO
```
geohot, line 114: Add comment that this is 16:1 exposure ratio?

geohot:
Overall I like that we are keeping this simple. It'll easily translate to OS04C10 when it's time (2 exposures streamed out where you can set the gain and time of each).

selfdrive/camerad/cameras/sensor2_i2c.h
pd0wm: For now it's just this. Once our tone mapping is good enough to actually use all this dynamic range I'll see if we want to add a gain ratio too.

haraschax:
Most focus is on OXs now which has internal HDR we use, making this PR irrelevant for now.

PR #24445, opened 2022-05-05
Chrysler: reduce ACC faults when disengaging on gas

### LINKED ISSUE #24333: Chrysler ACC fault when openpilot cancels cruise

sshane:
This seems to greatly reduce faults, however the car can still fault if the user is holding a steering wheel button while pressing the gas. Route with no faults: `4395faadf0a373f2|2022-05-06--08-56-29`

### DIFF
--- a/selfdrive/car/chrysler/carcontroller.py
+++ b/selfdrive/car/chrysler/carcontroller.py
@@ -49,8 +49,7 @@ class CarController:
     # *** control msgs ***
 
     if CC.cruiseControl.cancel:
-      # TODO: would be better to start from frame_2b3
-      can_sends.append(create_wheel_buttons(self.packer, self.frame, cancel=True))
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
Great, let's get this merged for 0.8.14.

adeebshihadeh (approved):

sshane:
Going to see if I can get one other tester for this, if not, then good to merge!

PR #24446, opened 2022-05-05
Chrsyler/Stallantis FPV2

realfast:
This should get FPV2 working on most Chrysler/Stellantis vehicles. $F132 is the module part number. This is updated when the software in the module is updated.


pd0wm:
Can you add the new query to the top of the list? This has the smallest risk on breaking other makes.

sshane:
I think we want to merge https://github.com/commaai/openpilot/pull/23311 first, with all the required testing, then we can merge this without worrying about other makes

PR #24450, opened 2022-05-06
Add timeout argument to test_processes

gijskoning:
Add --timeout TIMEOUT_SEC argument to test_processes to increase the expected messages timeout.
Useful when testing laikad messages since currently it takes just a bit more than 15 seconds to load in new data.

Also updated the readme by adding the usages of test_processes


pd0wm:
Can you do this without using globals?

gijskoning:
> Can you do this without using globals?

Yes will do! And also pass on the args in one argument

pd0wm:
Decided to just wait with putting laikad in process replay until it doesn't need internet anymore. It will probably be flaky anyway if you don't set a very large timeout.

PR #24452, opened 2022-05-06
camerad: combine debayering and rgb_to_yuv opencl kernels

jwooning:
this PR combines the dabayering and rgb_to_yuv kernels, this prevents a copy of the frame, which saves cpu and power. 

work done:
- combine kernels 
  - make kernel act on 4x4 pixel blocks
  - no longer send rgb frames
- apply `mad24` where this makes it faster
- optimize local work size


pd0wm:
Looks like this breaks snapshot, and some other CI

adeebshihadeh:
This will also break some provisioning tests.

PR #24455, opened 2022-05-06
Dynamically parse DBCs

sshane:
Now you can edit any DBC file and start PlotJuggler and it should grab your new changes without a complete re-compile of PlotJuggler!

Make sure to re-run `./juggle.py --install` to get the latest plugin


sshane:
This is about 50-70% slower on C3, `hyundai_kia_generic` for example takes consistently `10.5 ms` while on PC it's consistently `6.3 ms`.

For `toyota_nodsu_pt_generated`, it's consistently `3.1-3.2 ms` while on PC it's `1.3-1.5 ms`

PR #24468, opened 2022-05-08
Remove default 0 when indexing a list

herman5:
**Description**
`list_[0:n] == list_[:n]`

If this is approved I'll also update the submodules.

**Verification**
CI


adeebshihadeh:
Generally, we don't merge stylistic PRs like this if we're not going to enforce the style with a linter or static analysis tool.

PR #24473, opened 2022-05-09
Tools: allow using plotjuggler on local data

ntegan1:
https://github.com/commaai/openpilot/pull/24008

example of setting up device data in expected format  
```bash
# on device
cd /data/media/0/realdata; for f in $(ls | grep "^2022"); do echo mv $f "ffffffffffffffff|$f"; done
# after copying to pc or mounting to pc
./tools/plotjuggler/juggle.py '/home/ntegan/data/ffffffffffffffff|2022-04-20--16-20-00'
```

example of putting data back in default format
```bash
cd /data/media/0/realdata; for f in $(ls | grep -E 'f{16}\|'); do echo mv $f $(sed -E 's/^f{16}\|(.*)/\1/g' <<<$f); done
```

example of zero padding segment nums for clean rlog/hevc concattenating
```bash
# this moves a segment like "2022-05-17--15-27-45--9" to "2022-05-17--15-27-45--009"
cd /data/media/0/realdata; for f in $(ls | grep '^2022'); do a=($(sed -E 's/^(.*--.*--)(.*)/\1 \2/g' <<<$f)); echo mv $f "${a[0]}$(printf "%03d" ${a[1]})"; done
```

example of reverting zero padding segment nums
```bash
# this moves a segment like "2022-05-17--15-27-45--009" to "2022-05-17--15-27-45--9"
cd /data/media/0/realdata; for f in $(ls | grep '^2022'); do a=($(sed -E 's/^(.*--.*--)(.*)/\1 \2/g' <<<$f));  echo mv $f "${a[0]}$(sed -E 's/^[0]*(.*)/\1/g' <<<${a[1]})"; done
```

NOTE:
* doesn't support "/" dongle/time delimiter when using data dir, e.g. "/home/Downloads/4cf7a6ad03080c90/2021-09-29--13-46-36/3"

### DIFF
--- a/tools/lib/route.py
+++ b/tools/lib/route.py
@@ -3,6 +3,7 @@ import re
 from urllib.parse import urlparse
 from collections import defaultdict
 from itertools import chain
+from typing import Optional
 
 from tools.lib.auth_config import get_token
 from tools.lib.api import CommaApi
@@ -196,7 +197,11 @@ class SegmentName:
   # TODO: add constructor that takes dongle_id, time_str, segment_num and then create instances
   # of this class instead of manually constructing a segment name (use canonical_name prop instead)
   def __init__(self, name_str: str, allow_route_name=False):
-    self._name_str = name_str
+    data_dir_path_separator_index = name_str.rsplit("|", 1)[0].rfind("/")
+    use_data_dir = (data_dir_path_separator_index != -1) and ("|" in name_str)
+    self._name_str = name_str[data_dir_path_separator_index + 1:] if use_data_dir else name_str
+    self._data_dir = name_str[:data_dir_path_separator_index] if use_data_dir else None
+
     seg_num_delim = "--" if self._name_str.count("--") == 2 else "/"
     name_parts = self._name_str.rsplit(seg_num_delim, 1)
     if allow_route_name and len(name_parts) == 1:
@@ -219,5 +224,8 @@ class SegmentName:
 
   @property
   def route_name(self) -> RouteName: return self._route_name
+
+  @property
+  def data_dir(self) -> Optional[str]: return self._data_dir
 
   def __str__(self) -> str: return self._canonical_name
--- a/tools/lib/tests/test_route_library.py
+++ b/tools/lib/tests/test_route_library.py
@@ -1,0 +1,32 @@
+#!/usr/bin/env python
+import unittest
+from collections import namedtuple
+
+from tools.lib.route import SegmentName
+
+class TestRouteLibrary(unittest.TestCase):
+  def test_segment_name_formats(self):
+    Case = namedtuple('Case', ['input', 'expected_route', 'expected_segment_num', 'expected_data_dir'])
+
+    cases = [ Case("4cf7a6ad03080c90|2021-09-29--13-46-36", "4cf7a6ad03080c90|2021-09-29--13-46-36", -1, None),
+              Case("4cf7a6ad03080c90/2021-09-29--13-46-36--1", "4cf7a6ad03080c90|2021-09-29--13-46-36", 1, None),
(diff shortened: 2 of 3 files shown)


tools/lib/route.py
```diff
@@ -196,7 +197,12 @@ class SegmentName:
   # TODO: add constructor that takes dongle_id, time_str, segment_num and then create instances
   # of this class instead of manually constructing a segment name (use canonical_name prop instead)
   def __init__(self, name_str: str, allow_route_name=False):
-    self._name_str = name_str
+    data_dir_parts = name_str.rsplit("/", 1)
+    use_data_dir = bool(name_str.count("|")) and len(data_dir_parts) == 2
```
ntegan1, line 201:
```suggestion
    use_data_dir = len(data_dir_parts) == 2 and bool(data_dir_parts[1].count("|"))
```

tools/lib/route.py
```diff
@@ -196,7 +197,12 @@ class SegmentName:
   # TODO: add constructor that takes dongle_id, time_str, segment_num and then create instances
   # of this class instead of manually constructing a segment name (use canonical_name prop instead)
   def __init__(self, name_str: str, allow_route_name=False):
-    self._name_str = name_str
+    data_dir_parts = name_str.rsplit("/", 1)
+    use_data_dir = len(data_dir_parts) == 2 and bool(data_dir_parts[1].count("|"))
+    (self._name_str, self._data_dir) = \
+      (data_dir_parts[1], data_dir_parts[0]) if use_data_dir else \
+      (name_str, None)
```
ntegan1, line 204:
```suggestion
    self._name_str = data_dir_parts[1] if use_data_dir else name_str
    self._data_dir = data_dir_parts[0] if use_data_dir else None
```

tools/plotjuggler/juggle.py
```diff
@@ -88,7 +88,8 @@ def juggle_route(route_or_segment_name, segment_count, qlog, can, layout, dbc=No
     if route_or_segment_name.segment_num != -1 and segment_count is None:
       segment_count = 1
 
-    r = Route(route_or_segment_name.route_name.canonical_name)
+    rosn = route_or_segment_name
+    r = Route(rosn.route_name.canonical_name, rosn.data_dir)
```
pd0wm, line 92:
```suggestion
    r = Route(route_or_segment_name.route_name.canonical_name, route_or_segment_name.data_dir)
```

pd0wm:
This doesn't work if you specify a segment number. e.g. `./juggle.py "/home/Downloads/fbbb958f3157f063|2022-05-11--20-37-43/5"`

I also don't think `SegmentName` is the place to handle this. Why not add an argument to plot juggler to specify a directory?

tools/lib/route.py
```diff
@@ -196,7 +197,11 @@ class SegmentName:
   # TODO: add constructor that takes dongle_id, time_str, segment_num and then create instances
   # of this class instead of manually constructing a segment name (use canonical_name prop instead)
   def __init__(self, name_str: str, allow_route_name=False):
-    self._name_str = name_str
+    data_dir_parts = name_str.rsplit("/", 1)
+    use_data_dir = len(data_dir_parts) == 2 and bool(data_dir_parts[1].count("|"))
```
pd0wm, line 201:
```suggestion
    use_data_dir = (len(data_dir_parts) == 2) and ("|" in data_dir_parts[1])

File: tools/lib/tests/test_route_library.py [hunk 5]
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env python
+import unittest
+
+from tools.lib.route import SegmentName
+
+class TestRouteLibrary(unittest.TestCase):
+  def test_segment_name_formats(self):
+    INPUT_IDX = 0
```
pd0wm, line 8: Using a `namedtuple` for this would make things quite a bit cleaner.

PR #24474, opened 2022-05-09
Subaru: Add ES_Distance Cruise_Soft_Disable

martinl:
This PR adds ES_Distance Cruise_Soft_Disable message to carstate (needed for longitudinal control in the future), prerequisite opendbc PR is https://github.com/commaai/opendbc/pull/605


pd0wm:
Why does this need to be added to the list of signals now if they are not used?

pd0wm:
Feel free to reopen once the signal is actually used.

adeebshihadeh:
We need to parse it out so it can be copied.

PR #24476, opened 2022-05-09
clean up gamma variable names

ZwX1616:


pd0wm:
Ah didn't realize this was also tone mapping/gamma correcting!

ZwX1616:
> Ah didn't realize this was also tone mapping/gamma correcting!

my bad. Even took me a while to understand

PR #24477, opened 2022-05-09
tools: add link to jungle repo

zorrobyte:
When using a Panda Jungle for the first time, there is a helpful console message to clone the panda jungle repo, along with a suggested console command. Users may simply run the command and forego adding the udev rules while expecting the jungle to work. If the proper udev rules haven't been added, the script simply sits there and can confuse the user as it appears to be working, when it is not.

### DIFF
--- a/tools/replay/README.md
+++ b/tools/replay/README.md
@@ -61,7 +61,7 @@ cd selfdrive/ui && ./watch3
 
 ## Stream CAN messages to your device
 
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Made it a bit more generic. I'm not sure where it would go, but I think it would also be nice to have some jungle-specific docs in openpilot.

zorrobyte:
> Made it a bit more generic. I'm not sure where it would go, but I think it would also be nice to have some jungle-specific docs in openpilot.

Thank you, adeeb. I'll continue to contribute as I can to the docs and Wiki!

PR #24479, opened 2022-05-09
camerad: simplify debayer and fix clamp

geohot:
Shouldn't change behavior, refactors from #24452

### DIFF
--- a/selfdrive/camerad/cameras/real_debayer.cl
+++ b/selfdrive/camerad/cameras/real_debayer.cl
@@ -13,30 +13,33 @@ const __constant half3 color_correction_2 = (half3)(-0.25277411, -0.05627105, 1.
 // tone mapping params
 const half gamma_k = 0.75;
 const half gamma_b = 0.125;
-const half mp_default = 0.01; // ideally midpoint should be adaptive
+const half mp = 0.01; // ideally midpoint should be adaptive
+const half rk = 9 - 100*mp;
 
-half gamma_apply(half x, half mp) {
+inline half3 gamma_apply(half3 x) {
   // poly approximation for s curve
-  half rk = 9 - 100*mp;
-  if (x > mp) {
-    return (rk * (x-mp) * (1-(gamma_k*mp+gamma_b)) * (1+1/(rk*(1-mp))) / (1+rk*(x-mp))) + gamma_k*mp + gamma_b;
-  } else if (x < mp) {
-    return (rk * (x-mp) * (gamma_k*mp+gamma_b) * (1+1/(rk*mp)) / (1-rk*(x-mp))) + gamma_k*mp + gamma_b;
-  } else {
-    return x;
-  }
+  return (x > mp) ?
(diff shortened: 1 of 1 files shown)


geohot:
Are there debayer tests such that we can trust to merge this?

pd0wm:
There is https://github.com/commaai/openpilot/blob/master/selfdrive/test/process_replay/test_debayer.py which can run on the device.

Install pyopencl with `PYOPENCL_CL_PRETEND_VERSION=2.0 pip install pyopencl`. Once it's in AGNOS we'll add the test to CI.

pd0wm:
The tests shows lots of small changes (0-3).

selfdrive/camerad/cameras/real_debayer.cl
```diff
@@ -13,30 +13,33 @@ const __constant half3 color_correction_2 = (half3)(-0.25277411, -0.05627105, 1.
...
   // poly approximation for s curve
-  half rk = 9 - 100*mp;
-  if (x > mp) {
-    return (rk * (x-mp) * (1-(gamma_k*mp+gamma_b)) * (1+1/(rk*(1-mp))) / (1+rk*(x-mp))) + gamma_k*mp + gamma_b;
-  } else if (x < mp) {
-    return (rk * (x-mp) * (gamma_k*mp+gamma_b) * (1+1/(rk*mp)) / (1-rk*(x-mp))) + gamma_k*mp + gamma_b;
-  } else {
-    return x;
-  }
+  return (x > mp) ?
```
pd0wm, line 21: This is now comparing a half3 to a scalar? Something like `select` sounds more appropriate here.

pd0wm: @jwooning pointed out that the ternary operator with a vector is special in OpenCL and is the same as select: https://www.khronos.org/registry/OpenCL/sdk/2.2/docs/man/html/selectionOperator.html

selfdrive/camerad/cameras/real_debayer.cl
```diff
@@ -13,30 +13,33 @@ const __constant half3 color_correction_2 = (half3)(-0.25277411, -0.05627105, 1.
 // tone mapping params
 const half gamma_k = 0.75;
 const half gamma_b = 0.125;
-const half mp_default = 0.01; // ideally midpoint should be adaptive
+const half mp = 0.01; // ideally midpoint should be adaptive
+const half rk = 9 - 100*mp;
```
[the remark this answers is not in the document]
geohot: Hmm, why? I can't imagine it's this change.

geohot:
You think this is okay to merge?

selfdrive/camerad/cameras/real_debayer.cl
geohot: Yea I just copied that from Joost's PR

selfdrive/camerad/cameras/real_debayer.cl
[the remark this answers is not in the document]
pd0wm: It was really strange, depending on the code that was below removing the `const` qualifier would change the outcome. The change was really small, so it's probably just rounding errors when the compiler computes the `9 - 100 * 0.01` vs when it's done at runtime.

pd0wm:
Think it's good to merge! We can take care of generating new references for test_debayer.

PR #24488, opened 2022-05-10
Locationd: integrate with laika messages

gijskoning:
Managed to make laika messages work with locationd and got a decent working version.
See the image which compares "not using Ublox position messages with it Laika" to regular Ublox position messages.

This change shouldn't affect anything since the GnssMeasurements messages are not sent yet from laikad.

Todo:
- [x] Create Cereal PR

Future steps:
- Use variation data from laikad instead of an incorrect guess.
- Use ephemeris instead of internet data.
- Calculate RMS of segments
- Verify if velocity is really needed to initialize the kalman filter. And if it is in correct format.

image


selfdrive/locationd/locationd.cc
```diff
@@ -440,7 +528,12 @@ void Localizer::handle_msg_bytes(const char *data, const size_t size) {
 
 void Localizer::handle_msg(const cereal::Event::Reader& log) {
   double t = log.getLogMonoTime() * 1e-9;
-  this->time_check(t);
+  if (log.isGnssMeasurements()){
+    t = log.getGnssMeasurements().getUbloxMonoTime() * 1e-9;
+  }
+  else{
+    this->time_check(t);
```
pd0wm, line 535: Why no time check?

selfdrive/locationd/locationd.cc
```diff
@@ -451,6 +544,8 @@ void Localizer::handle_msg(const cereal::Event::Reader& log) {
     this->handle_cam_odo(t, log.getCameraOdometry());
   } else if (log.isLiveCalibration()) {
     this->handle_live_calib(t, log.getLiveCalibration());
+  } else if (log.isGnssMeasurements()) {
```
pd0wm, line 547: Lets disable processing of `gpsLocationExternal` when the first gnssMeasurement is seen.

gijskoning: I want this pr to have the same behaviour as currently in master. Disabling gpslocationExternal would completely stop the working of location in the kalman filter since the messages from laika aren't working when not having internet.

gijskoning: nvm. We want to disable gpslocationExternal such that we can confirm gnssmeasurements are working without it.

selfdrive/locationd/locationd.cc
```diff
@@ -250,16 +250,99 @@ void Localizer::input_fake_gps_observations(double current_time) {
...
+  std::vector<std::vector<double>> satpos_glonassfreqs(size);
+
+  std::vector<VectorXd> pseudo_rate_zs(size);
+  std::vector<MatrixXdr> pseudo_rate_Rs(size);
+  std::vector<std::vector<double>> satpos_satvels(size);
+  auto observation = (c_id == ConstellationId::GPS) ? OBSERVATION_PSEUDORANGE_GPS : OBSERVATION_PSEUDORANGE_GLONASS;
+  auto observation_rate = (c_id == ConstellationId::GPS) ? OBSERVATION_PSEUDORANGE_RATE_GPS : OBSERVATION_PSEUDORANGE_RATE_GLONASS;
+
+  for (int i = 0; i < size; ++i)
+  {
```
pd0wm, line 277: Brackets go on the same line as the function/for/if.

selfdrive/locationd/locationd.cc
```diff
@@ -250,16 +250,99 @@ void Localizer::input_fake_gps_observations(double current_time) {
...
+  std::vector<MatrixXdr> pseudo_Rs(size);
+  std::vector<std::vector<double>> satpos_glonassfreqs(size);
+
+  std::vector<VectorXd> pseudo_rate_zs(size);
+  std::vector<MatrixXdr> pseudo_rate_Rs(size);
+  std::vector<std::vector<double>> satpos_satvels(size);
+  auto observation = (c_id == ConstellationId::GPS) ? OBSERVATION_PSEUDORANGE_GPS : OBSERVATION_PSEUDORANGE_GLONASS;
+  auto observation_rate = (c_id == ConstellationId::GPS) ? OBSERVATION_PSEUDORANGE_RATE_GPS : OBSERVATION_PSEUDORANGE_RATE_GLONASS;
+
+  for (int i = 0; i < size; ++i)
```
pd0wm, line 276:
```suggestion
  for (auto &meas : meas_per_constellation)
```

gijskoning:
Concluded with Harald that we don't want to do any locationd changes yet. We first want to have 'GNSS only' kalman filter running in Laikad that can predict the current position, velocity and heading. Which are needed to fully replace the current GPSLocationExternal messages. This pr will still exist however all the other non locationd changes will move to others pr's.

PR #24491, opened 2022-05-10
Parallel process replay

lukaspetersson:

Branched from PRs: https://github.com/commaai/openpilot/pull/24495 and https://github.com/commaai/openpilot/pull/24496
Uses Cereal from PR: https://github.com/commaai/cereal/pull/299


selfdrive/common/params.cc
```diff
@@ -180,9 +184,30 @@ std::unordered_map<std::string, uint32_t> keys = {
 
 } // namespace
 
+void clean_param_dirs(std::string path){
```
adeebshihadeh, line 187: why clean the directories? I don't think it's params' responsibility to do this

selfdrive/common/params.cc
```diff
@@ -180,9 +184,30 @@ std::unordered_map<std::string, uint32_t> keys = {
...
+    LOGW("File already deleted in a parallel process");
+  }
+}
+
 Params::Params(const std::string &path) {
   static std::string default_param_path = ensure_params_path();
   params_path = path.empty() ? default_param_path : ensure_params_path(path);
+  prefix = std::getenv("OPENPILOIT_PREFIX");
+  clean_param_dirs(getParamPath());
+  std::filesystem::create_directory(getParamPath());
```
adeebshihadeh, line 210: `create_params_path ` already takes care of this

selfdrive/test/process_replay/test_processes.py
```diff
@@ -154,36 +181,43 @@ def format_diff(results, ref_commit):
...
-    results[segment] = {}
-
-    r, n = segment.rsplit("--", 1)
-    lr = LogReader(get_url(r, n))
-
+    if not args.parallel:
+      print(f"***** testing route segment {segment} *****\n")
+      results[segment] = {}
+      r, n = segment.rsplit("--", 1)
+      lr = LogReader(get_url(r, n))
```
adeebshihadeh, line 193: this shouldn't be a completely separate code path. process replay should just take something like a jobs argument

selfdrive/test/process_replay/test_processes.py
```diff
@@ -127,6 +152,8 @@ def format_diff(results, ref_commit):
                       help="Updates reference logs using current commit")
   parser.add_argument("--upload-only", action="store_true",
                       help="Skips testing processes and uploads logs from previous test run")
+  parser.add_argument("--parallel", action="store_true",
+                      help="Run tests in parallel")
```
adeebshihadeh, line 156:
```suggestion
  parser.add_argument("--jobs", action="store_true")
```

selfdrive/test/process_replay/test_processes.py
lukaspetersson: I did it that way initially, but there is then a lot of duplicate code. I will change it back and we can see

PR #24492, opened 2022-05-11
encoderd: enable

geohot:
All functionality is in master, just have to make the tests pass for this. (I think it's good)

encoderd is usually the process driving the rotation unless loggerd has not received any encoder packets in the last 500ms. The log will rotate once all 4 encoders have.

1200 frame based rotation logic remains the same, it's just now in encoderd. The rotate_cv has been replaced with a queue which enqueues the encoder messages for the next segment.


selfdrive/hardware/tici/test_power_draw.py
```diff
@@ -22,7 +22,7 @@ class Proc:
   Proc('camerad', 2.5),
   Proc('modeld', 0.95),
   Proc('dmonitoringmodeld', 0.25),
-  Proc('loggerd', 0.45, warmup=10.),
+  Proc('encoderd', 0.42),
```
geohot, line 25: Half of this power is the NV12 copy!

geohot:
Test failure is real, will investigate tomorrow.

selfdrive/loggerd/encoderd.cc
```diff
@@ -139,8 +140,8 @@ int main() {
     int ret;
     ret = util::set_realtime_priority(52);
     assert(ret == 0);
-    ret = util::set_core_affinity({7});
-    assert(ret == 0);
+    ret = util::set_core_affinity({3});
+    assert(ret == 0 || Params().getBool("IsOffroad"));
```
adeebshihadeh, line 144: this can be removed since core 3 is offline while offroad

geohot:
Woot, jenkins passes!

geohot:
So this is now hitting the qcam size bug.

@adeebshihadeh can you look into this since it's happening on master also? I suspect it's what the fcamera is seeing, and the new devices see different things

geohot:
Be aware, this changes behavior slightly since encoderd is the primary driver of log rotation now. It shouldn't matter in the car since loggerd is started before either encoderd or camerad (just one or the other, with camerad latency this shouldn't ever happen)

But if it does, the first segment won't contain 1200 frames. The first segment also won't contain 1200 frames when you start and kill the logger on the body.

PR #24493, opened 2022-05-11
debayering: change vignetting correction to faster function

jwooning:


jwooning:
This seems to fit pretty well to the current function:
image

The error in the correction factor for every pixel:
image

PR #24495, opened 2022-05-11
params: support OPENPILOT_PREFIX

lukaspetersson:
Add a prefixing ID in the file structure path to params.
Tested by:
* running test_processes
* running on device
* running latency logger on produced logs


selfdrive/common/params.cc
```diff
@@ -4,12 +4,16 @@
...
 #include <csignal>
+#include <filesystem>
+#include <string>
 #include <unordered_map>
 
 #include "selfdrive/common/swaglog.h"
 #include "selfdrive/common/util.h"
 #include "selfdrive/hardware/hw.h"
 
+int MAX_PARAM_DIRS = 400;
```
adeebshihadeh, line 15: why is there a max?

selfdrive/manager/manager.py
```diff
@@ -30,6 +30,8 @@ def manager_init() -> None:
   # update system time from panda
   set_time(cloudlog)
 
+  os.environ['OPENPILOIT_PREFIX'] = "MANAGER"
```
adeebshihadeh, line 33: same comment as msgq PR, manager should use a default

selfdrive/test/process_replay/process_replay.py
```diff
@@ -337,6 +338,7 @@ def ublox_rcv_callback(msg):
 
 
 def replay_process(cfg, lr, fingerprint=None):
+  os.environ['OPENPILOIT_PREFIX'] = str(uuid.uuid4())
```
adeebshihadeh, line 341: this should go in the process replay PR

selfdrive/common/params.cc
```diff
@@ -180,9 +184,31 @@ std::unordered_map<std::string, uint32_t> keys = {
 
 } // namespace
 
+
+void clean_param_dirs(std::string path){
```
adeebshihadeh, line 188: why clean the directories? I don't think it's params' responsibility to do this

selfdrive/common/params.cc
```diff
@@ -180,9 +184,31 @@ std::unordered_map<std::string, uint32_t> keys = {
...
+    LOGW("File already deleted in a parallel process");
+  }
+}
+
 Params::Params(const std::string &path) {
   static std::string default_param_path = ensure_params_path();
   params_path = path.empty() ? default_param_path : ensure_params_path(path);
+  prefix = std::getenv("OPENPILOIT_PREFIX");
+  clean_param_dirs(getParamPath());
+  std::filesystem::create_directory(getParamPath());
```
adeebshihadeh, line 211: `create_params_path` already takes care of this

selfdrive/common/params.h
```diff
@@ -18,7 +18,7 @@ class Params {
   bool checkKey(const std::string &key);
   ParamKeyType getKeyType(const std::string &key);
   inline std::string getParamPath(const std::string &key = {}) {
-    return key.empty() ? params_path + "/d" : params_path + "/d/" + key;
+    return params_path + "/d/" + prefix + (key.empty()? "" : "/" + key);
```
adeebshihadeh, line 21:
```suggestion
    return params_path + "/d/" + prefix + (key.empty() ? "" : "/" + key);
```

adeebshihadeh (commented):
Unlike msgq, we need to be more careful with params since it stores persistent values. Either the default path is the same as the current one or this needs to handle migrating to the new one.

PR #24496, opened 2022-05-11
Prefix messages for process replay

lukaspetersson:
Add a prefixing ID in the file structure path to messages
Tested by:
* running test_processes
* running on device
* running latency logger on produced logs 

Uses Cereal from PR: https://github.com/commaai/cereal/pull/299

### DIFF
--- a/selfdrive/manager/manager.py
+++ b/selfdrive/manager/manager.py
@@ -30,8 +30,6 @@ def manager_init() -> None:
   # update system time from panda
   set_time(cloudlog)
 
-  os.environ['OPENPILOIT_PREFIX'] = "MANAGER"
-
   # save boot log
   subprocess.call("./bootlog", cwd=os.path.join(BASEDIR, "selfdrive/loggerd"))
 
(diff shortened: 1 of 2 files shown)


selfdrive/manager/manager.py
```diff
@@ -30,6 +30,8 @@ def manager_init() -> None:
   # update system time from panda
   set_time(cloudlog)
 
+  os.environ['OPENPILOIT_PREFIX'] = "MANAGER"
```
adeebshihadeh, line 33: manager should use a default prefix. here's one use case this breaks: openpilot's running and you want to run a debug script like dump.py

adeebshihadeh (commented):
Make sure to open PRs for your submodule changes, and ideally link them in the description

selfdrive/test/process_replay/process_replay.py
```diff
@@ -337,6 +338,7 @@ def ublox_rcv_callback(msg):
 
 
 def replay_process(cfg, lr, fingerprint=None):
+  os.environ['OPENPILOIT_PREFIX'] = str(uuid.uuid4())
```
adeebshihadeh, line 341: this should go in the process replay PR

PR #24499, opened 2022-05-12
Toyota: correct weight for the Prius V

cydia2020:
Prius V's weight is incorrect in `interface.py`, correct weight updated from https://www.toyotacertified.com/content/dam/tcuv/sections/brochures/en/prius-v/2017_priusv_ebrochure.pdf

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -46,7 +46,7 @@ class CarInterface(CarInterfaceBase):
       ret.wheelbase = 2.78
       ret.steerRatio = 17.4
       tire_stiffness_factor = 0.5533
-      ret.mass = 3340. * CV.LB_TO_KG + STD_CARGO_KG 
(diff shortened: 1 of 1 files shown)


selfdrive/car/toyota/interface.py
```diff
@@ -46,7 +46,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[], disabl
       ret.wheelbase = 2.78
       ret.steerRatio = 17.4
       tire_stiffness_factor = 0.5533
-      ret.mass = 4387. * CV.LB_TO_KG + STD_CARGO_KG
+      ret.mass = 3340. * CV.LB_TO_KG + STD_CARGO_KG  # from Toyota eBrochures
```
adeebshihadeh, line 49: no need for the comment
```suggestion
      ret.mass = 3340. * CV.LB_TO_KG + STD_CARGO_KG 
```

PR #24503, opened 2022-05-12
Tesla stock AEB forwarding

robbederks:


robbederks:
Doesn't fault my car (have been driving for 5+ hours on this), but also haven't confirmed AEB forwarding actually works, since I had no close encounters with the rear bumper of a lead car.

Think it's still worth to merge as-is (since it definitely has a way higher chance of working than without this PR), and will update my findings the next time I get a bit too close.

PR #24505, opened 2022-05-12
Add desired curvature to controlsState

JMPZ11:
**Description** [] Adds the actual curvature and curvature rate that comes out of "get_lag_adjusted_curvature()" and is fed into and used by the lateral controllers. The fields are referred to in controlsd (and the parameter name in update() in the lateral controllers) as desired_curvature and desired_curvature_rate.

Added the fields to CarControl because we only have access to them in state_control method of controlsd, and CarControl is the most relevant cereal object available in this method.

Depends on cereal PR: https://github.com/commaai/cereal/pull/301

Background

While attempting to diagnose unusual tuning behavior in GM Trucks, I had a need to analyze the curvature, and discovered that the curvature actually used by the lateral controllers was not logged anywhere. I added it temporarily, and was surprised to discover that the output of get_lag_adjusted_curvature was in part responsible for the delayed and jerky steering.

**Verification** [] Doesn't change behavior, just adds 2 fields to be logged

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -138,6 +138,9 @@ class Controls:
     self.LoC = LongControl(self.CP)
     self.VM = VehicleModel(self.CP)
 
+    self.desired_curvature = 0.0
+    self.desired_curvature_rate = 0.0
+
     self.LaC: LatControl
     if self.CP.steerControlType == car.CarParams.SteerControlType.angle:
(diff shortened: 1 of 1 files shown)


sshane:
Ah, this is much cleaner than my PR! https://github.com/commaai/openpilot/pull/24411

sshane:
Reopening as https://github.com/commaai/openpilot/pull/24510 as I can't access opgm

PR #24506, opened 2022-05-12
ui: fix cameraview widget on MacOS after YUV changes

pd0wm:

### DIFF
--- a/selfdrive/ui/qt/widgets/cameraview.cc
+++ b/selfdrive/ui/qt/widgets/cameraview.cc
@@ -13,7 +13,7 @@ namespace {
 
 const char frame_vertex_shader[] =
 #ifdef __APPLE__
-  "#version 150 core\n"
+  "#version 330 core\n"
 #else
   "#version 300 es\n"
 #endif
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -227,7 +227,7 @@ void CameraViewWidget::paintGL() {
     glBindTexture(GL_TEXTURE_2D, textures[i]);
     int width = i == 0 ? stream_width : stream_width / 2;
     int height = i == 0 ? stream_height : stream_height / 2;
-    glTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, width, height, GL_LUMINANCE, GL_UNSIGNED_BYTE, address[i]);
+    glTexSubImage2D(GL_TEXTURE_2D, 0, 0, 0, width, height, GL_RED, GL_UNSIGNED_BYTE, address[i]);
```
pd0wm, line 230: Does this work on PC and C3 or do we need check the system?

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -13,7 +13,7 @@ namespace {
 
 const char frame_vertex_shader[] =
 #ifdef __APPLE__
-  "#version 150 core\n"
+  "#version 330 core\n"
```
pd0wm, line 16: Update to newer version to support location bindings.

selfdrive/ui/qt/widgets/cameraview.cc
pd0wm: Yes, it does.

PR #24507, opened 2022-05-12
debayering: code cleanup

jwooning:
some cleanup from #24452


geohot:
Nice! I like incremental changes like this that are easy to merge.

PR #24508, opened 2022-05-12
hatanaka 2.4

YassineYousfi:


adeebshihadeh:
Any reason this needs to be here instead of the xx Pipfile? I don't see it used in openpilot.

YassineYousfi:
It's only used in laika. @gijskoning do we need it in openpilot?

pd0wm:
It's currently used in laikad.

PR #24510, opened 2022-05-12
Add desired curvature to controlsState

sshane:
cereal PR: https://github.com/commaai/cereal/pull/301


sshane:
first test of cloud update_refs!

PR #24516, opened 2022-05-12
remove MPC integration

ClockeNessMnstr:
MPC state integration seems to cause oscillation.

I currently have been using with jerk_pts from @nuwandavek's MPC turning PR. 
not using heading cost also doesn't make sense. I see @HaraldSchafer has been looking at this change recently in some of the model PR's.

### DIFF
--- a/selfdrive/controls/lib/drive_helpers.py
+++ b/selfdrive/controls/lib/drive_helpers.py
@@ -33,8 +33,7 @@ CRUISE_INTERVAL_SIGN = {
 class MPC_COST_LAT:
   PATH = 1.0
   HEADING = 1.0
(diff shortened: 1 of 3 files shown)


ClockeNessMnstr:
going to keep the branch out there but just split this up when i have time.

ClockeNessMnstr:
still some code from other PR's included here but the integration commit is the improvement.

ClockeNessMnstr:
still some code from other PR's included here but the integration commit is the improvement.

HaraldSchafer:
Unfortunately all PRs that affect optimizers such as this need to be large scale validation. Which makes them unmergeable by default.

PR #24517, opened 2022-05-12
Fix process replay regen

sshane:
Fix regen issues caused by these two PRs: https://github.com/commaai/openpilot/pull/24392 and https://github.com/commaai/openpilot/pull/24416

### DIFF
--- a/selfdrive/test/process_replay/regen.py
+++ b/selfdrive/test/process_replay/regen.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+import bz2
 import os
 import time
 import multiprocessing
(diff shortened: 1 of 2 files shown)


selfdrive/test/process_replay/regen.py
```diff
@@ -268,11 +270,18 @@ def regen_and_save(route, sidx, upload=False, use_route_meta=False):
     lr = LogReader(r.log_paths()[args.seg])
     fr = FrameReader(r.camera_paths()[args.seg])
   else:
-    lr = LogReader(f"cd:/{route.replace('|', '/')}/{sidx}/rlog")
+    lr = LogReader(f"cd:/{route.replace('|', '/')}/{sidx}/rlog.bz2")
```
sshane, line 273: we could also regenerate all the routes, but then we upload and store 40MB per segment instead of 10-13, which also takes much longer to upload

selfdrive/test/process_replay/regen.py
```diff
@@ -268,11 +270,18 @@ def regen_and_save(route, sidx, upload=False, use_route_meta=False):
...
 
-  lr = LogReader(os.path.join(rpath, 'rlog2'))
+  # compress raw rlog before uploading
+  with open(os.path.join(rpath, "rlog"), "rb") as f:
+    data = bz2.compress(f.read())
+  with open(os.path.join(rpath, "rlog.bz2"), "wb") as f:
+    f.write(data)
+  os.remove(os.path.join(rpath, "rlog"))
+
+  lr = LogReader(os.path.join(rpath, 'rlog.bz2'))
```
sshane, line 284: we could instead regenerate all the routes uncompressed, but then we upload and store 40MB per segment instead of 10-13, which also takes much longer to upload

pd0wm: Compressing before uploading is fine!

pd0wm (approved):

sshane:
Should have made a draft, this doesn't quite work yet

PR #24530, opened 2022-05-13
Parallel process replay python

lukaspetersson:

Branched from PR: https://github.com/commaai/openpilot/pull/24495


adeebshihadeh:
Already merged params and MSGQ is almost done, so let's close this and do it all in #24534.

PR #24534, opened 2022-05-14
process replay: support running in parallel

lukaspetersson:

Branched from PRs:  https://github.com/commaai/openpilot/pull/24572


selfdrive/test/process_replay/test_processes.py
```diff
@@ -127,6 +151,7 @@ def format_diff(results, ref_commit):
                       help="Updates reference logs using current commit")
   parser.add_argument("--upload-only", action="store_true",
                       help="Skips testing processes and uploads logs from previous test run")
+  parser.add_argument("--jobs", action="store_true")
```
sshane, line 154: Maybe make an int so you can specify different number of jobs?

sshane, line 154: Maybe make an int so you can specify different number of jobs?

```suggestion
  parser.add_argument("-j", "--jobs", type=int, default=1)
```

lukaspetersson: Depends on the use case. Which interface is preferred by pipeline?

sshane: Doesn't make sense to have an argument named jobs that doesn't take a number of jobs. If you're going to hard code, you should probably at least use `os.cpu_count()`.

adeebshihadeh: yeah, should be an int. 1 is fine as the default, same as make

selfdrive/test/process_replay/test_processes.py
```diff
@@ -154,38 +183,42 @@ def format_diff(results, ref_commit):
     untested = (set(interface_names) - set(excluded_interfaces)) - tested_cars
     assert len(untested) == 0, f"Cars missing routes: {str(untested)}"
 
-  results: Any = {}
+  jobs = 7 if args.jobs else 1
+  pool = NestablePool(jobs)
```
sshane, line 187:
```suggestion
  pool = NestablePool(args.jobs)
```
don't forget this

selfdrive/test/process_replay/process_replay.py
```diff
@@ -336,12 +338,34 @@ def ublox_rcv_callback(msg):
...
 
+def setup_prefix():
+  os.environ['OPENPILOT_PREFIX'] = str(uuid.uuid4())
+  msg_path = '/dev/shm/' + os.environ['OPENPILOT_PREFIX']
+  try:
+    os.mkdir(msg_path)
+  except FileExistsError:
+    pass
+  except PermissionError:
+    print(f"WARNING: failed to make {msg_path}")
```
adeebshihadeh, line 349: we don't want to handle this here. it's only in manager since manager shouldn't really crash
```suggestion
```

selfdrive/test/process_replay/process_replay.py
```diff
@@ -336,12 +338,34 @@ def ublox_rcv_callback(msg):
...
+
+
+def teardown_prefix():
```
adeebshihadeh, line 352: check that OPENPILOT_PREFIX is set and return if it's not

selfdrive/test/process_replay/process_replay.py
```diff
@@ -336,12 +338,34 @@ def ublox_rcv_callback(msg):
...
+  params_path = '/data/params/' if os.environ.get('TICI', 0) else os.environ['HOME'] + '/.comma/params/'
```
adeebshihadeh, line 353: should be able to get this from params directly. Params already has a getParamPath you can use, but it just needs to be exposed in the cython wrapper

selfdrive/test/process_replay/process_replay.py
```diff
@@ -336,12 +338,34 @@ def ublox_rcv_callback(msg):
...
+  symlink_path = params_path + os.environ['OPENPILOT_PREFIX']
```
adeebshihadeh, line 354: this only removes the symlink, not the actual params data

selfdrive/test/process_replay/process_replay.py
```diff
@@ -336,12 +338,34 @@ def ublox_rcv_callback(msg):
...
+
 
 def replay_process(cfg, lr, fingerprint=None):
+  setup_prefix()
   if cfg.fake_pubsubmaster:
-    return python_replay_process(cfg, lr, fingerprint)
+    res = python_replay_process(cfg, lr, fingerprint)
   else:
-    return cpp_replay_process(cfg, lr, fingerprint)
+    res = cpp_replay_process(cfg, lr, fingerprint)
```
adeebshihadeh, line 366: add a finally block to make sure the teardown runs

selfdrive/test/process_replay/test_processes.py
```diff
@@ -154,38 +182,41 @@ def format_diff(results, ref_commit):
...
         os.remove(cur_log_fn)
 
+  results: Any = {}
+  p2 = pool.map_async(run_test_process, pool_args)
+  for tup in p2.get():
+    if tup[0] not in results:
+      results[tup[0]] = {}
+    results[tup[0]][tup[1]] = tup[2]
+  pool.close()
+  pool.join()
```
adeebshihadeh, line 218: have you tried to wrap these in unittest/pytest test cases instead? I think this would have a few benefits:
* nice setup/teardown hooks
* cleaner stdout. everything's jumbled together now
* parallel running for free with pytest-xdist

selfdrive/test/process_replay/process_replay.py
lukaspetersson: https://stackoverflow.com/questions/11700545/how-to-delete-a-symbolic-link-in-python
This post says otherwise.

I can double check

selfdrive/test/process_replay/process_replay.py
lukaspetersson: We have common/basedir.py which handles this for persist. Could be nice to use it for params aswell.

selfdrive/test/process_replay/test_processes.py
lukaspetersson: I will look into this on another branch

selfdrive/test/process_replay/test_processes.py
```diff
@@ -154,38 +182,42 @@ def format_diff(results, ref_commit):
     untested = (set(interface_names) - set(excluded_interfaces)) - tested_cars
     assert len(untested) == 0, f"Cars missing routes: {str(untested)}"
 
-  results: Any = {}
+  pool = NestablePool(args.jobs)
+
+  lreaders: Any = {}
+  p1 = pool.map_async(get_logreader, [seg for car, seg in segments])
```
adeebshihadeh, line 188: we shouldn't need two pools here

selfdrive/test/process_replay/test_processes.py
```diff
@@ -154,38 +182,42 @@ def format_diff(results, ref_commit):
...
         os.remove(cur_log_fn)
 
+  results: Any = {}
+  p2 = pool.map_async(run_test_process, pool_args)
+  for tup in p2.get():
+    if tup[0] not in results:
+      results[tup[0]] = {}
+    results[tup[0]][tup[1]] = tup[2]
+  pool.close()
+  pool.join()
```
adeebshihadeh, line 219: use a context manager instead of doing this manually

selfdrive/test/process_replay/test_processes.py
```diff
@@ -11,6 +12,20 @@
...
+class NoDaemonProcess(multiprocessing.Process):
+  @property # type: ignore
+  def daemon(self):
+    return False
+  @daemon.setter
+  def daemon(self, value):
+    pass
+class NoDaemonContext(type(multiprocessing.get_context())): # type: ignore
+  Process = NoDaemonProcess
+class NestablePool(multiprocessing.pool.Pool): # pylint: disable=abstract-method
```
adeebshihadeh, line 25: shouldn't `concurrent.futures.ProcessPoolExecutor()` do what we want?

selfdrive/test/process_replay/test_processes.py
lukaspetersson: I think we do? We dont want to calculate lr each time, this pool runs less times.

adeebshihadeh (commented):
Progress bars look pretty good now.

adeebshihadeh:
Let's fix updating the refs then should be good to merge.

selfdrive/test/process_replay/test_processes.py
```diff
@@ -53,6 +55,25 @@
 BASE_URL = "https://commadataci.blob.core.windows.net/openpilotci/"
 REF_COMMIT_FN = os.path.join(PROC_REPLAY_DIR, "ref_commit")
 
+def run_test_process(data):
```
sshane, line 58:
```suggestion

def run_test_process(data):
```
I'd just make sure there's two lines before and after function declarations

PR #24535, opened 2022-05-14
Car Port: 2022 Honda Civic

sshane:
Moved from: https://github.com/commaai/openpilot/pull/23307

**Checklist**
- [x] added to README
- [x] route with openpilot: f44aa96ace22f34a|2021-12-22--06-22-31
- [x] route with stock system: f44aa96ace22f34a|2021-12-25--21-08-31
- [x] confirm spam resume works
- [x] ~~figure out possible new LKAS_HUD message~~

Requires: 
- [x] https://github.com/commaai/cereal/pull/234 - will make safety param
- [x] https://github.com/commaai/opendbc/pull/620
- [x] https://github.com/commaai/panda/pull/942

Lateral only for now, long should be pretty easy. There is no radar on this car so openpilot can send ACC commands directly to the car.


sshane:
`ACC_HUD` is on bus 2
`ACC_CONTROL` is on bus 0
`POWERTRAIN_DATA` and all other signals in `get_can_signals` (pt) are on bus 0

kevinharbin:
What's the criteria for "Steer to 0"? I think the 22 civic does that.

adeebshihadeh:
Ignition issue: #24680

> Describe the bug
> The car went onroad/recording with the car off, parked, and no one around.
> There were 29 0 min routes starting at 11:49am then a 48 minute route starting at 11:53 that ended when I got back to the car and unplugged the OBD power.
> 
> I don't know how to reproduce the bug.
> 
> Provide a route where the issue occurs
> f44aa96ace22f34a|2022-05-29--11-49-34--0

sshane:
This user still seems to be having this problem, device turns on and off multiple times for just a few seconds. None of the logs I checked had started or any ignition signals high, so it must be for just a second and it takes a while for loggerd to start

Screenshot from 2022-06-14 11-26-45

selfdrive/car/honda/values.py
```diff
@@ -1392,6 +1394,40 @@ class HondaCarInfo(CarInfo):
       b'57114-TYF-E030\x00\x00'
     ],
   },
+  CAR.CIVIC_2022: {
+    (Ecu.eps, 0x18DA30F1, None): [
+      b'39990-T39-A130\x00\x00',
+      b'39990-T43-J020\x00\x00',  # hatch
```
adeebshihadeh, line 1400: these comments aren’t necessary and probably won’t be maintained

selfdrive/car/honda/values.py
```diff
@@ -75,6 +75,7 @@ class CAR:
   CIVIC = "HONDA CIVIC 2016"
   CIVIC_BOSCH = "HONDA CIVIC (BOSCH) 2019"
   CIVIC_BOSCH_DIESEL = "HONDA CIVIC SEDAN 1.6 DIESEL 2019"
+  CIVIC_2022 = "HONDA CIVIC 2022"
```
adeebshihadeh, line 78: we should probably start moving to using generations like the VW port

selfdrive/car/honda/values.py
```diff
@@ -1429,5 +1466,6 @@ class HondaCarInfo(CarInfo):
 HONDA_NIDEC_ALT_SCM_MESSAGES = {CAR.ACURA_ILX, CAR.ACURA_RDX, CAR.CRV, CAR.CRV_EU, CAR.FIT, CAR.FREED, CAR.HRV, CAR.ODYSSEY_CHN,
                                 CAR.PILOT, CAR.PASSPORT, CAR.RIDGELINE}
 HONDA_BOSCH = {CAR.ACCORD, CAR.ACCORDH, CAR.CIVIC_BOSCH, CAR.CIVIC_BOSCH_DIESEL, CAR.CRV_5G,
-               CAR.CRV_HYBRID, CAR.INSIGHT, CAR.ACURA_RDX_3G, CAR.HONDA_E}
-HONDA_BOSCH_ALT_BRAKE_SIGNAL = {CAR.ACCORD, CAR.CRV_5G, CAR.ACURA_RDX_3G}
+               CAR.CRV_HYBRID, CAR.INSIGHT, CAR.ACURA_RDX_3G, CAR.HONDA_E, CAR.CIVIC_2022}
+HONDA_BOSCH_ALT_BRAKE_SIGNAL = {CAR.ACCORD, CAR.CRV_5G, CAR.ACURA_RDX_3G}  # no radarless car should have this signal
```
adeebshihadeh, line 1470: is this necessarily true or just happens to be true now?

sshane: We assert this in safety right now to avoid accidentally adding a new platform to the list where we haven't seen a car yet with the other brake signal. I'll remove comment though

selfdrive/car/honda/values.py
sshane: I'm fine with making a migration later on if it turns out all new Hondas (/Civics) are like this.

PR #24540, opened 2022-05-15
Toyota: add 2021 RX350 fingerprint

omatzyo:
Adding 2021 RX350


ElonVampire:
收到！

adeebshihadeh:
Which route did this come from?

omatzyo:
> Which route did this come from?

It was pulled from cd /data/openpilot/selfdrive/car/
python fw_versions.py.

Once updated in the local file I was able to use the software. Before this I was stuck in dashcam only mode.

adeebshihadeh:
I see. We'll need your device's dongle ID or a route to verify this new FW version though.

omatzyo:
> I see. We'll need your device's dongle ID or a route to verify this new FW version though.

Absolutely, DongleID 35c4875b3d867d06.

 I’m brand new (just got my comma this week) and before I figured out the fingerprint I was only able to run dragonpilot and I think got my account banned on the comma.so server. While you’re looking at my account could you address that for me please?

sshane:
Your device doesn't look banned. You have no drives with fw fingerprints logged, all the routes are fixed to the Lexus RX 2020. Can you get a route on stock openpilot with no changes?

omatzyo:
Oh that’s good news! Will do. I’ll let you know once its there.

omatzyo:
Ok. I just put comma.so release3 on, added my line to values.py and did a quick drive. It is uploaded now.

Issue #24222, opened 2022-04-15
build_devel.sh: use master commit hash and date in commit body


pd0wm:
This allows us to map master-ci commits to master

PR #24548, opened 2022-05-16
Include master commit and build date in master-ci commit

ntegan1:
https://github.com/commaai/openpilot/issues/24222

this information was previously saved in the `$SOURCE_DIR` `selfdrive/common/version.h` instead of the `$TARGET_DIR`


release/build_devel.sh
```diff
@@ -46,20 +46,23 @@ if [ ! -z "$EXTRA_FILES" ]; then
...
-rm -f panda/board/obj/panda.bin.signed
-
 echo "[-] committing version $VERSION T=$SECONDS"
 git add -f .
 git status
-git commit -a -m "openpilot v$VERSION release"
+git commit -a -m "openpilot v$VERSION release
+
+master commit: $GIT_HASH
+date: $DATETIME"
```
ntegan1, line 65:
```suggestion
git commit -a -m "openpilot v$VERSION release"
```

maybe fixing version.h commit+date info is good enough?

PR #24554, opened 2022-05-16
camerad: remove webcam + replay cameras

adeebshihadeh:


geohot:
Can we also remove the rgb_to_yuv transform?

adeebshihadeh:
Yes, I'm gonna do a big camerad cleanup. I just don't want to remove webcam support without a replacement.

geohot:
Oh I'm fine with just removing it. If people are using it, they can fork or write the simple python one.

JasonJShuler:
Just to clarify - is this breaking the ability to run OP on PCs?

MikeBusuttil:
Hey @adeebshihadeh (or anyone that knows)
- what was the motivation behind removing the webcam?
- where can I find the existing simple Python one that @geohot referenced above?

Context: I plan on tackling #24590 as my first issue

PR #24557, opened 2022-05-17
make debayer faster

geohot:
20ms -> 7.7ms

Edges are wrong, not really sure what desired behavior is here (but tell me what you want if you can't make it fast, use math, not control flow). And vignetting is slightly less precise, one value for the 2x2.

Also saves 120mW, 2.17W -> 2.05W on my device.

Three general OpenCL rules:
* Unless you are doing something very strange or need to accumulate, local caching is a bad idea. The GPU built-in cache does a very good job for normal image reads.
* Memory accesses are slow and should be used very sparingly. Half of the speedup is from using vload8 instead of byte accesses.
* The compiler is smart, it's LLVM. "const", "inline", unrolling loops won't speed things up.

Assuming 20 FLOPS per RGB pixel, this is now getting 57 GFLOPS. Getting more likely requires image inputs instead of pointers, then it uses a different way of accessing memory. But this makes more sense for intermediate stuff like what's in SNPE, where the exact byte locations matter less. 
```
1000/7.7*3*(1920*1280*20*3)
```

### DIFF
--- a/selfdrive/camerad/cameras/camera_common.cc
+++ b/selfdrive/camerad/cameras/camera_common.cc
@@ -38,10 +38,10 @@ public:
              "-cl-fast-relaxed-math -cl-denorms-are-zero "
              "-DFRAME_WIDTH=%d -DFRAME_HEIGHT=%d -DFRAME_STRIDE=%d -DFRAME_OFFSET=%d "
              "-DRGB_WIDTH=%d -DRGB_HEIGHT=%d -DRGB_STRIDE=%d "
-             "-DBAYER_FLIP=%d -DHDR=%d -DCAM_NUM=%d",
+             "-DBAYER_FLIP=%d -DHDR=%d -DCAM_NUM=%d%s",
              ci->frame_width, ci->frame_height, ci->frame_stride, ci->frame_offset,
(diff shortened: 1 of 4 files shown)


geohot:
Ugh, this is why you debayer to 1920x1200. Those 8 pixels so weren't worth it.

geohot:
Meh, the black level should really be done in a kernel define.

geohot:
CI failure is weird and unrelated I think. Still can't merge because the edges are wrong, but maybe Europe can finish this evening.

jwooning:
In total adds about 0.10 ms to fix the edges.

Currently does contain a possible out of bounds read, although the resulting pixels aren't used. It reads two bytes before the start of the pixel data, currently this isn't a direct issue because of the camera register data which starts before the pixel data.

This is not a problem for the bottom-right corner, because of the row alignment, which gives 4 bytes of empty buffer data.

geohot:
You should be able to clamp the X and fix the out of bounds read I think. Then switch to

```
if (gid_x == 0) {
    va.s2 = va.s0;
    vb.s2 = vb.s0;
    vc.s2 = vc.s0;
    vd.s2 = vd.s0;

PR #24353, opened 2022-04-28
camerad: get embedded statistics and data by embedding in pixel data

pd0wm:
follow up of https://github.com/commaai/openpilot/pull/24299

```
0 - frame id: 1638
0 - temp 0 (top): 40.15
0 - temp 1 (bottom): 39.44
1 - frame id: 1638
1 - temp 0 (top): 40.30
1 - temp 1 (bottom): 40.15
2 - frame id: 1639
2 - temp 0 (top): 40.71
2 - temp 1 (bottom): 40.57

```

PR #24564, opened 2022-05-17
camerad: disable registers/stats

pd0wm:
both registers and stats are causing framedrops. Reverting until we figure out how to fix.

out of founds read in debayer kernel must be fixed before merging!

### DIFF
--- a/selfdrive/camerad/cameras/camera_qcom2.cc
+++ b/selfdrive/camerad/cameras/camera_qcom2.cc
@@ -31,8 +31,8 @@ const size_t FRAME_WIDTH = 1928;
 const size_t FRAME_HEIGHT = 1208;
 const size_t FRAME_STRIDE = 2896;  // for 12 bit output. 1928 * 12 / 8 + 4 (alignment)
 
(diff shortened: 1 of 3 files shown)


selfdrive/camerad/cameras/camera_qcom2.cc
```diff
@@ -681,6 +681,18 @@ void CameraState::camera_open() {
...
+
+    if (ci.stats_offset != -1 || ci.registers_offset != -1) {
+      uint16_t smia_test = 0x1802;
+      if (ci.stats_offset != -1) smia_test |= (1 << 7);
+      if (ci.registers_offset != -1) smia_test |= (1 << 8);
+
+      struct i2c_random_wr_payload smia_test_reg[] = {
+        {0x3064, smia_test},
+      };
+      sensors_i2c(smia_test_reg, std::size(smia_test_reg), CAM_SENSOR_PACKET_OPCODE_SENSOR_CONFIG, true);
```
pd0wm, line 693: Should be safe to always send it, but gating behind the env variable to ensure everything is the same as before merging https://github.com/commaai/openpilot/pull/24353

PR #24264, opened 2022-04-20
plannerd: improve lateral accuracy in turns

nuwandavek:
Example: 
image

More details in the report.

PR #24573, opened 2022-05-17
latmpc-imp

ClockeNessMnstr:
PR is to https://github.com/commaai/openpilot/pull/24264 's branch. 

@nuwandavek 

we can set the initial state explicitly. don't need to estimate it since we measure it. separating the MPC from the actual states just drives more error.

### DIFF
--- a/selfdrive/controls/lib/lateral_planner.py
+++ b/selfdrive/controls/lib/lateral_planner.py
@@ -1,5 +1,5 @@
 import numpy as np
-from common.realtime import sec_since_boot, DT_MDL
+from common.realtime import sec_since_boot
 from common.numpy_fast import interp
 from selfdrive.swaglog import cloudlog
 from selfdrive.controls.lib.lateral_mpc_lib.lat_mpc import LateralMpc, X_DIM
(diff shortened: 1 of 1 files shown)


nuwandavek:
No, we cannot use the measured curvature, I think. Separating the MPC from the real state definitely is not ideal. But there's a lot of lag in the system (longer than multiple MPC steps) that needs to be accounted for. We can potentially do something like what's done in the `long_mpc` where the real values are "leaked in" with a decay. But I doubt if that makes a big difference. Speed values are fairly constant over longer horizons than yaw_rate changes.

ClockeNessMnstr:
Closing the loop for anyone who may stumble upon this.

There's some nomenclature between the MPC and Drive_helpers implying that x0[3] (curvature) is the "current" curvature. 
This is compounded by calling curvature_rates[0] the desired curvature_rate. 

Since the MPC's initial curvature state is  "disconnected + integrated" we should actually be getting an estimate of a desired setpoint even though the MPC's free variable is desired rate. And the "rate" is no longer a current desired rate but the desired rate of change in the setpoint regardless of actual current state. 

https://github.com/commaai/openpilot/pull/24701

PR #24585, opened 2022-05-18
Hyundai longitudinal: reduce time from acceleration to deceleration

sshane:
Another PR split out from https://github.com/commaai/openpilot/pull/24053/files. This reduces the minimum positive jerk from 1 to 0, so when acceleration error is negative, we don't allow any pos jerk, hopefully for faster braking after accelerating


sshane:
Unclear if this does anything

PR #24586, opened 2022-05-18
Latency logger: binary search for frame ID

lukaspetersson:
Binary search for frame id interval.
Reduces lookup time from 21s to 9s for large segments.

### DIFF
--- a/tools/latencylogger/latency_logger.py
+++ b/tools/latencylogger/latency_logger.py
@@ -4,6 +4,7 @@ import matplotlib.patches as mpatches
 import matplotlib.pyplot as plt
 import mpld3
 import sys
+from bisect import bisect_left, bisect_right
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
21s -> 9s for the whole runtime of latency logger?

lukaspetersson:
> 21s -> 9s for the whole runtime of latency logger?

No, only the algorithmic part, i.e. excluding logreader + graphing

Total times:
LogReader: 4s
algorithm: 9s
prepare 2 figs: 6s
save 2 figs: 7s
TOTAL: 25s (on this route (60s), will do more tests)

lukaspetersson:
Tested in testing closet

Total times:
Adding new device with full segment: 38 -> 30
Existing device adding full segment: 28 ->23

PR #24587, opened 2022-05-18
camerad: pixclk 88mhz but frame time the same

geohot:
@pd0wm Check this.

It increases PIXCLK to maximum 88 MHz, but it keeps the frame time the same by increasing FRAME_LENGTH_LINES proportionally. Do you see the problematic exposure with this?

With this, we get image roll times of 27.1 ms, close to the minimum for the sensor.

Update: I'm way wrong about the minimum. LINE_LENGTH_PCK can somehow be less than X_OUTPUT_CONTROL.  Here it is at 22.8ms, and seems fine.

Btw, the sof to eof time seems very real, it's the actual image roll time, which makes sense. It starts on first pixel, and it's done when it receives the right number of rows.

### DIFF
--- a/selfdrive/camerad/cameras/sensor2_i2c.h
+++ b/selfdrive/camerad/cameras/sensor2_i2c.h
@@ -52,7 +52,7 @@ struct i2c_random_wr_payload init_array_ar0231[] = {
   // input clock is 19.2 / 2 * 0x37 = 528 MHz
   // pixclk is 528 / 6 = 88 MHz
   // full roll time is 1000/(PIXCLK/(LINE_LENGTH_PCK*FRAME_LENGTH_LINES)) = 39.99 ms
-  // img  roll time is 1000/(PIXCLK/(LINE_LENGTH_PCK*Y_OUTPUT_CONTROL))   = 22.85 ms
+  // img  roll time is 1000/(PIXCLK/(LINE_LENGTH_PCK*Y_OUTPUT_CONTROL))   = 15.18 ms
   {0x302A, 0x0006}, // VT_PIX_CLK_DIV
   {0x302C, 0x0001}, // VT_SYS_CLK_DIV
   {0x302E, 0x0002}, // PRE_PLL_CLK_DIV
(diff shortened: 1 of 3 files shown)


geohot:
Aha!!!

DIGITAL_CTRL (0x30BA) has NUM_EXP_MAX, which determines max LINE_LENGTH_PCK.

It's currently set to 2, but I believe this actually means 3 exposures. When I set it to 1, 0x0452 works as LINE_LENGTH_PCK, and we get 60 FPS.

geohot:
I don't think we actually want to merge this since it messes with the exposure somehow, either through FINE_INTEGRATION_TIME_ or just the numbers being bigger.

But let's get this into the HDR PR. 15.18 ms roll time is great!

geohot:
Hmm, there's an artifact at high exposure time on the right side of the image

geohot:
The wizard says the minimum row time is 15.64 because "TX pulse width failed"

geohot:
Hmm, why is exposure CI failing?

selfdrive/camerad/cameras/sensor2_i2c.h
```diff
@@ -117,8 +121,10 @@ struct i2c_random_wr_payload init_array_ar0231[] = {
   {0x100C, 0x0589}, // FINE_INTEGRATION_TIME2_MIN
   {0x100E, 0x07B1}, // FINE_INTEGRATION_TIME3_MIN
   {0x1010, 0x0139}, // FINE_INTEGRATION_TIME4_MIN
-  {0x3014, 0x08CB}, // FINE_INTEGRATION_TIME_
```
geohot, line 120: I now have a lot of doubts if this 2 exposure HDR is working at all with these parameters.

geohot: Maybe I'm just wrong about DIGITAL_CTRL.

geohot:
What a waste of time, it seems DIGITAL_CTRL is real after all, even though this makes no sense. If you set the num exp to 0 it breaks.

geohot:
From a different datasheet:
"The AR0231AT sensor from ON Semiconductor is capable of
a maximum resolution of 1928 × 1208 at up to 40 fps in HDR mode
and 60 fps in linear mode."

geohot:
The end result is this makes the roll 7ms faster.

PR #24588, opened 2022-05-18
Toyota Corolla SE 2022 missing 'esp' finger print added

480V277:
Daniel Giacoio
05/18/2022

**Description** - Added new esp finger print 'F152612B91' to selfdrive/car/toyota/values.py for Toyota Corolla TSS2. This was a problem for the Corolla SE 2022.

**Verification** [Updated values.py script with the missing esp finger print and is no longer in dashcam mode.]


sshane:
Do you have a route where this fingerprint came from?

DRG11:
> Do you have a route where this fingerprint came from?

Yes, here's a route with no fingerprint: b4b162bb7a7916de|2022-05-15--21-06-54

DRG11:
> Do you have a route where this fingerprint came from?

Yes, here's a route with no fingerprint identified by openpliot: b4b162bb7a7916de|2022-05-15--21-06-54

sshane:
Ah another PR had the exact fw version as yours, merged in here: https://github.com/commaai/openpilot/pull/24598. Thanks for the PR!

Issue #24223, opened 2022-04-16
GM: ACC faults


adeebshihadeh:
Events while engaged on 0.8.13-release
```
CHEVROLET VOLT PREMIER 2017              85a49685933befe2|2022-02-24--15-22-05--5 7
CHEVROLET VOLT PREMIER 2017              1f0e64fd786bf5ed|2022-02-26--15-42-18--1 6
CHEVROLET VOLT PREMIER 2017              e2b25f39da040caa|2022-02-26--16-38-50--10 7
CHEVROLET VOLT PREMIER 2017              1f0e64fd786bf5ed|2022-03-01--14-09-07--2 6
CHEVROLET VOLT PREMIER 2017              2e5410d8378a30c6|2022-03-01--14-42-51--35 6
CADILLAC ESCALADE ESV 2016               495552ac53209990|2022-03-02--17-15-36--34 7
CADILLAC ESCALADE ESV 2016               495552ac53209990|2022-03-02--18-44-06--1 7
CHEVROLET VOLT PREMIER 2017              d2ccdcf33e62cfa0|2022-03-04--13-58-06--8 7
CHEVROLET VOLT PREMIER 2017              3954275d7161f9a8|2022-03-05--11-36-56--17 6
CADILLAC ESCALADE ESV 2016               2e5410d8378a30c6|2022-03-08--18-09-17--20 6
CHEVROLET VOLT PREMIER 2017              2e5410d8378a30c6|2022-03-08--22-47-34--20 6
CHEVROLET VOLT PREMIER 2017              2e5410d8378a30c6|2022-03-10--00-25-40--1 6
CADILLAC ESCALADE ESV 2016               495552ac53209990|2022-03-10--08-38-12--26 7
CADILLAC ESCALADE ESV 2016               495552ac53209990|2022-03-10--08-38-12--27 7
CADILLAC ESCALADE ESV 2016               495552ac53209990|2022-03-10--08-38-12--28 7
CADILLAC ESCALADE ESV 2016               495552ac53209990|2022-03-10--08-38-12--29 7
CHEVROLET VOLT PREMIER 2017              2e5410d8378a30c6|2022-03-11--13-25-56--4 6
CHEVROLET VOLT PREMIER 2017              d2ccdcf33e62cfa0|2022-03-12--19-23-43--2 7
CHEVROLET VOLT PREMIER 2017              9425b0775a4688a1|2022-03-12--11-42-35--74 6
[...]

PR #23712, opened 2022-02-04
GM: Fix noisy brake pressed

sshane:
PR moved from https://github.com/commaai/openpilot/pull/23617

Looking at all available data for the last 6 months, looks like this is a good change: https://imgur.com/a/DIeGHXs

```
Platform: CADILLAC ESCALADE ESV 2016
Checked 10 unique dongles, 119 unique routes, and 513 segments
Mismatches: 58573 frames, 493 segments (any frame)
---
Platform: GMC ACADIA DENALI 2018
Checked 1 unique dongles, 55 unique routes, and 483 segments
Mismatches: 16448 frames, 445 segments (any frame)
---
Platform: CHEVROLET VOLT PREMIER 2017
Checked 66 unique dongles, 2844 unique routes, and 10802 segments
Mismatches: 248608 frames, 7649 segments (any frame)
```

PR #24589, opened 2022-05-18
GM: use vehicle brake-pressed threshold to fix Cruise Fault

hewers:
**Description**

Observed GM vehicle Cruise Fault, or always unable to engage. https://github.com/commaai/openpilot/issues/24223
[Previous debugging](https://github.com/commaai/openpilot/wiki/GM#troubleshooting-open-pilots-cruise-fault-warning-on-volt-brake-pedal-positioning) related this behavior to a sticky brake pedal position on the GM Volt. There is a service bulletin and replacement part.
PlotJuggler shows that when brake position >= 8, in 0.5 seconds ACC will fault. When the value sticks high, then pulling up on the brake pedal will allow engaging, and dropping the brake pedal will cause a cruise fault.

**Verification**

With this change, vehicle correctly disengages when slowly and lightly increasing brake pedal pressure, hovering near this 8-count transition.

### DIFF
--- a/selfdrive/car/gm/carstate.py
+++ b/selfdrive/car/gm/carstate.py
@@ -32,10 +32,8 @@ class CarState(CarStateBase):
     ret.standstill = ret.vEgoRaw < 0.01
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(pt_cp.vl["ECMPRDNL"]["PRNDL"], None))
 
-    # Brake pedal's potentiometer returns near-zero reading even when pedal is not pressed.
-    ret.brake = pt_cp.vl["EBCMBrakePedalPosition"]["BrakePedalPosition"] / 0xd0
-    ret.brakePressed = pt_cp.vl["EBCMBrakePedalPosition"]["BrakePedalPosition"] >= 10
-
+    ret.brake = pt_cp.vl["ECMAcceleratorPos"]["BrakePedalPos"] / 0xd0
+    ret.brakePressed = pt_cp.vl["ECMAcceleratorPos"]["BrakePedalPos"] >= 8
     # Regen braking is braking
     if self.CP.transmissionType == TransmissionType.direct:
       ret.brakePressed = ret.brakePressed or pt_cp.vl["EBCMRegenPaddle"]["RegenPaddle"] != 0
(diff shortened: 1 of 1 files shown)


qadmus:
Fails process replay, needs refs update.

sshane:
How often does your car randomly disengage from this change?

qadmus:
> How often does your car randomly disengage now from this change?

No false disengagements yet. To temporarily resolve the sticky brake pedal sensor, I can repeatedly press the brake pedal to the floor. Each press will lower the stuck value by ~1 count.

adeebshihadeh:
This doesn't seem to actually be just a revert from #23617. The threshold has been slightly lowered, but perhaps this is more correct. @sshane if we don't have time to verify this threshold change before release, perhaps we should just start with the revert.

adeebshihadeh:
This doesn't seem to actually be just a revert from #23617. The threshold has been slightly lowered, but perhaps this is more correct. @sshane if we don't have time to verify this threshold change before release, perhaps we should just start with the revert.

qadmus:
> This doesn't seem to actually be just a revert from #23617. The threshold has been slightly lowered, but perhaps this is more correct. @sshane if we don't have time to verify this threshold change before release, perhaps we should just start with the revert.

Yup, not a revert, change of signal too.

sshane:
Let's strictly revert that PR first (https://github.com/commaai/openpilot/pull/23712/files) as that would be the quickest to merge, however I don't know if that introduced more cruise faults than before. Looking at the route qadmus provided, that old signal would **not** have caught any of his cruise faults:

Screenshot from 2022-05-20 14-41-22

I verified it with qadmus and this new signal is a good change for avoiding cruise faults, however needs some verification with a notebook

sshane:
Looked at some routes from that cruise faults issue, reverting the PR would catch those, not sure why the old signal in qadmus's route was zero for this time

sshane:
Checked this signal vs the current and old one, doesn't seem to catch many more than the old one. Most of the faults are *not* from any thee of the the brake pedal signals:

```python
Total segments with faults: 27
Reverted PR would have caught: 5
Master would have caught:      0
New PR would have caught:      6
```

qadmus:
Back to draft:
- [ ] update panda safety code
- [ ] try and use both signals and both thresholds

sshane:
Just curious, how often are you hitting this?

qadmus:
> Just curious, how often are you hitting this nowadays?

This patch has since been on my daily-driver. I never hit "controls mismatched" due to openpilot remaining engaged when the vehicle platform disengages. I haven't gotten the car fixed yet, so when the issue reoccurs, I stop and fully depress the brake a few times.

I'm fulltime on another project and unable to follow through on my open issues, so I'll close this. Feel free to do anything with it!

PR #24592, opened 2022-05-19
Toyota: fill stockFcw on all cars except those with their DSU unplugged

cydia2020:
Don't know why this was limited to TSS 2.0 cars in the first place, the system only send `create_fcw_command` when a vehicle is not a `NO_DSU_CAR` and DSU is not seen on the CAN Network.


cydia2020:
Closing for now need to look into this a bit more

PR #24594, opened 2022-05-19
curv points

ClockeNessMnstr:
I added curv points to mine and changed "jerk" to "curv_rate" to match the MPC states.
scaling with speed**2  under mpc.run() like the pts (heading *v_ego). 

I also see LAT_ACC was there. this may have already been tested. 

x_pts with path cost might also be significant for harder  turns but I'm not sure how to test that.


ClockeNessMnstr:
Adding these points should just "complete" y_ref. and changing to "curvature" just matches the internal MPC states so that the cost functions don't need to be matched outside of the MPC.py file. 

@nuwandavek just going to close so it's out of the way

PR #24598, opened 2022-05-20
Toyota: add missing esp FW version for 2022 Toyota Corolla

luuuuucas:
Added new esp fingerprint for 2022 Toyota Corolla LE. Comma 3 device remained in dashcam mode before updating values.py.

Route: [6c15cf37ba9e13af|2022-05-19--20-12-09--0]


luuuuucas:
Taken from route [6c15cf37ba9e13af|2022-05-19--20-12-09--0]

PR #24603, opened 2022-05-20
honda bosch: typo affecting longitudinal rate limits

incognitojam:
I don't think the rate limits are being applied correctly to the ACC accel sent to CAN.

The change was introduced [here](https://github.com/commaai/openpilot/commit/4f1eb4278a83a3abebebef67a081961e27594d90#diff-f09b53221a42992af21072f2f671f4234b4137aa083cd1f2a3ff2d94f14abb06R219) and I don't think it was intentional.

EDIT: Originally I noticed that `accel` was sent instead of `self.accel`, but then I also noticed that `accel` was being used to create the value for `self.gas`. I'm less sure about `self.gas` being calculated incorrectly.

EDIT2: reverted the last change

### DIFF
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -222,7 +222,7 @@ class CarController:
         if self.CP.carFingerprint in HONDA_BOSCH:
           self.accel = clip(accel, self.params.BOSCH_ACCEL_MIN, self.params.BOSCH_ACCEL_MAX)
           self.gas = interp(accel, self.params.BOSCH_GAS_LOOKUP_BP, self.params.BOSCH_GAS_LOOKUP_V)
-          can_sends.extend(hondacan.create_acc_commands(self.packer, CC.enabled, CC.longActive, accel, self.gas,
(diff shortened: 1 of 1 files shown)


incognitojam:
Actually, I think there might be another typo? If not then I could be misunderstanding the whole thing..

redacid95:
Unfortunately that did not do the trick, still getting cruise faulted. This time was instant.

The cruise fault is caused by a brake error. In Cabana, I am seeing under Standstill Release > Brake Error 1 > there is a value of 1 when it faults. I am not sure where the cause of this error is coming from though. It appears no accel or gas is being passed when it errors, even thought I see that the res/set button was pressed.

redacid95:
This is why accel is used for self.gas interp on HB:

def compute_gb_honda_bosch(accel, speed):
  # TODO returns 0s, is unused
  return 0.0, 0.0

That is in carcontroller. Will always return 0.0. 

I believe you were correct in changing self.accel, but that seems to be 1 piece to the puzzle.

pd0wm:
Your first change was correct. The lookup for gas actually has to be done on accel.

PR #24608, opened 2022-05-20
latency logger: update demo route

lukaspetersson:
* update demo route to a preserved route

### DIFF
--- a/tools/latencylogger/latency_logger.py
+++ b/tools/latencylogger/latency_logger.py
@@ -214,11 +214,7 @@ if __name__ == "__main__":
   args = parser.parse_args()
 
   r = DEMO_ROUTE if args.demo else args.route_or_segment_name.strip()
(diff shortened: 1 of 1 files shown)


tools/latencylogger/latency_logger.py
```diff
@@ -214,7 +214,11 @@ def get_timestamps(lr):
   args = parser.parse_args()
 
   r = DEMO_ROUTE if args.demo else args.route_or_segment_name.strip()
-  lr = logreader_from_route_or_segment(r, sort_by_time=True)
+  try:
+    lr = logreader_from_route_or_segment(r, sort_by_time=True)
+  except Exception:
+    print("Route or Segment not found")
+    exit()
```
adeebshihadeh, line 221: let's remove this, we don't handle this in any of the other tools

PR #24609, opened 2022-05-20
UI: poll modelV2

sshane:
The issue before was that sm->receive was finishing either too early sometimes, causing the model path drawn to be the same as the previous model packet. This reduces the frames behind camera from 1-2 due to chances of timings to just 1 (50-100ms to 50ms)

Difference is very noticable when going over bumps, path sticks to the road more (as we're removing the inconsistencies).


selfdrive/ui/ui.cc
```diff
@@ -118,7 +118,9 @@ static void update_model(UIState *s, const cereal::ModelDataV2::Reader &model) {
 }
 
 static void update_sockets(UIState *s) {
-  s->sm->update(0);
+  // ensures UI stays responsive when modelV2 is not alive
+  int timeout = s->sm->alive("modelV2") ? 1000 / UI_FREQ : 0;
+  s->sm->update(timeout);
```
sshane, line 123: Kept the QTimer at 50 ms, as that enforces a consistent rate still, then we wait anywhere from 3 to 7 ms on PC replay, and {} to {} ms on C3

selfdrive/ui/ui.cc
```diff
@@ -118,7 +118,12 @@ static void update_model(UIState *s, const cereal::ModelDataV2::Reader &model) {
 }
 
 static void update_sockets(UIState *s) {
-  s->sm->update(0);
+  // ensures UI stays responsive when modelV2 is not alive
+  int timeout = s->sm->alive("modelV2") ? 1000 / UI_FREQ : 0;
```
sshane, line 122: UI was much more jittery when scrolling in any of the views when just setting to 50 when model was dead, even with an instant singleshot timer

selfdrive/ui/ui.cc
```diff
@@ -255,6 +258,9 @@ void UIState::update() {
     watchdog_kick();
   }
   emit uiUpdate(*this);
+  double remaining = next_frame_time - millis_since_boot();
```
sshane, line 261: Think we need to either take the remaining time into account, or just use a static QTimer at 50 ms intervals, either accounts for a long poll time + other processing like drawing

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -265,7 +265,7 @@ void CameraViewWidget::vipcConnected(VisionIpcClient *vipc_client) {
 
 void CameraViewWidget::vipcFrameReceived(VisionBuf *buf) {
   latest_frame = buf;
-  update();
+//  update();  // TODO: fix other places where cameraview is used
```
sshane, line 268: Uncommenting this caused onroad widget to be drawn at 40 fps

sshane:
When there's no map view up and the user isn't touching the screen causing a redraw, model is *always* drawn on the camera frame (at least from C3 being onroad with CAN replay for 45 minutes).

However the problem is that on the second C3 with the map view, it randomly draws the camera 1 frame ahead, and when touching and dragging the map view, camera view lags significantly which doesn't happen on master. Will see what's going on. Any ideas for someone who knows an obvious fix/sees an issue with the PR?

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -216,6 +216,10 @@ void CameraViewWidget::paintGL() {
 
   if (latest_frame == nullptr) return;
 
+  if (latest_frame_id != draw_frame_id) {
+    LOGW("camera and draw frame ids do not match: %d != %d", latest_frame_id, draw_frame_id);
+  }
```
sshane, line 221: This will spam when modeld dies, or on PC replay, which is always 1 frame behind

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -242,6 +243,10 @@ void CameraViewWidget::paintGL() {
   glPixelStorei(GL_UNPACK_ALIGNMENT, 4);
 }
 
+void CameraViewWidget::vipcDisconnected() {
+  latest_frame = nullptr;
+}
```
sshane, line 248: Need this or we get seg faults if we restart replay/camerad with UI onroad, as latest_frame doesn't get reset, and we may try to call paintGL before we get a frame after connect.

sshane:
Has some timing issues, think if we start too late/early we end up polling for the previous model packet on start up. Need to dynamically adjust the update loop so we're consistently getting the latest.

adeebshihadeh:
Closing for now.

PR #24610, opened 2022-05-20
process replay: logreader as bytes

lukaspetersson:
Send logreader as bytes to tasks.

Time with 32 jobs:
branch: 0:45
master: 1:24

Tested by:
* Running test_process with different amount of jobs
* Launching openpilot
* Running logs in latencylogger


tools/lib/logreader.py
```diff
@@ -74,22 +74,23 @@ def reset(self):
 
 
 class LogReader:
-  def __init__(self, fn, canonicalize=True, only_union_types=False, sort_by_time=False):
+  def __init__(self, fn, dat=None, canonicalize=True, only_union_types=False, sort_by_time=False):
```
adeebshihadeh, line 77: move dat to the end in case anything relies on the order here

tools/lib/logreader.py
```diff
@@ -101,6 +102,10 @@ def __init__(self, fn, canonicalize=True, only_union_types=False, sort_by_time=F
     self._ents = list(sorted(_ents, key=lambda x: x.logMonoTime) if sort_by_time else _ents)
     self._ts = [x.logMonoTime for x in self._ents]
 
+  @classmethod
+  def from_bytes(cls, dat):
+    return cls("", dat)
```
adeebshihadeh, line 107:
```suggestion
    return cls("", dat=dat)
```

PR #24612, opened 2022-05-21
GM: Revert noisy brake pressed signal

sshane:
This was causing a few faults that we could have prevented and disengaged, instead of causing a temporary cruise fault. From all the faults here: https://github.com/commaai/openpilot/issues/24223

```python
Total segments with faults: 27, would be able to prevent by disengaging with brake pedal: 6
Reverted PR would have caught: 5  # https://github.com/commaai/openpilot/pull/23712
Master would have caught:      0
New PR would have caught:      6  # https://github.com/commaai/openpilot/pull/24589
```

### DIFF
--- a/selfdrive/car/gm/carstate.py
+++ b/selfdrive/car/gm/carstate.py
@@ -69,7 +69,7 @@ class CarState(CarStateBase):
     ret.brakePressed = ret.brake > 1e-5
     # Regen braking is braking
     if self.car_fingerprint == CAR.VOLT:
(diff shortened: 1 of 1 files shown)


sshane:
Will merge once tests pass

JasonJShuler:
Changing away from a fixed threshold value of 8 is going to be necessary for trucks, possibly other cars.

What if instead of capturing a flawed signal, we used the change in ACC status to trigger a disengagement? This could have the same behavior (unexpected disengagements vs unexpected cruise faults), and keep the brake pressed signal clean.

It looks like we use the cruiseState.enabled for enganging; Does the state change when the "false" brake presses occur?

Are there any public routes I could peek at where this problem was happening?

PR #24614, opened 2022-05-21
Added tests for clip in common/nano_fast.py

Kinvert:
Wrote tests for clip in common/nano_fast.py

I am new to contributing to other's projects. Normally I would make a branch and do a pull request, but in https://docs.comma.ai/CONTRIBUTING.html it says to pull against master.

Please let me know if I'm doing this wrong.

Clip takes in a low value, a high value, and an x value. I wrote a few quick tests to help make sure it's functioning as intended.

Not much, but at least it will help with test coverage.

### DIFF
--- a/common/tests/test_numpy_fast.py
+++ b/common/tests/test_numpy_fast.py
@@ -1,7 +1,27 @@
 import numpy as np
 import unittest
 
-from common.numpy_fast import interp
+from common.numpy_fast import clip, interp
+
+
+class ClipTest(unittest.TestCase):
+  def test_clip_too_low(self):
+    hi = 9
+    lo = 5
+    x = 3
+    assert clip(x, lo, hi) == 5
+
+  def test_clip_between(self):
+    hi = 9
+    lo = 5
+    x = 7
+    assert clip(x, lo, hi) == 7
+
+  def test_clip_too_high(self):
+    hi = 9
+    lo = 5
+    x = 11
+    assert clip(x, lo, hi) == 9
 
 
 class InterpTest(unittest.TestCase):


incognitojam:
I think "against master" just meant that you should target `commaai:master`, which you have done! :slightly_smiling_face:

Kinvert:
> You **can** call your branch whatever you like, but I don't think you can change it now without re-creating the PR.

Emphasis added.

Is this a "can" or "ought to"? Since I already made the PR it's easier to leave as is, on my end. But I don't know what they'd prefer to see here. Will do whichever they'd prefer.

incognitojam:
You would have been right to make a branch. "Against the master branch" didn't refer to the name of your branch, just that your pull request should be targetting comma's master branch.

Kinvert:
Gotcha. Is this something I need to change before they'd potentially merge? Or just do it that way from now on?

incognitojam:
I think it will be fine, it really don't affect anything

pd0wm:
While I appreciate the effort we don't need to add 20 lines of test for a 1 line function that's never touched.

Some small feedback for future PRs:
 - For this kind of reimplementations of an existing function (`np.clip()`) it's usually better to test against the existing implementation with a whole bunch of values.
 -  Use `self.asertEqual` in unittests.

pd0wm:
With never touched I mean it's never changed.

Kinvert:
Understood.

For future PRs:
- Are there areas you're seeking testing coverage?
- Are you guys aiming to be MISRA compliant?
- What areas of the code are easiest for new contributors in your opinion?

Thanks for your time.

pd0wm:
- We could use some more coverage around athena/uploader uploading. Create a new issue for that https://github.com/commaai/openpilot/issues/24624
 - We are aiming to be misra compliant in panda safety code.
 - See https://github.com/commaai/openpilot/issues?q=is%3Aopen+is%3Aissue+label%3A%22good+first+issue%22. Especially the webcam deamon would be very nice!

PR #24628, opened 2022-05-23
process replay: regen in parallel

lukaspetersson:
Run each segment of regen_all in parallel


lukaspetersson:
Some routes (outcommented) are not fingerprinted

lukaspetersson:
Ram usage go from 10 -> 40 % on my machine on master. I therefore run out of memory when I run in parallel with 3 or more jobs.

selfdrive/test/process_replay/regen.py
```diff
@@ -178,6 +179,10 @@ def replay_camera(s, stream, dt, vipc_server, frames, size, use_extra_client):
 
 
 def regen_segment(lr, frs=None, outdir=FAKEDATA):
+  if "OPENPILOT_PREFIX" not in os.environ:
+    global tqdm # pylint: disable=W0602
```
adeebshihadeh, line 183: can you do this in a cleaner way that doesn't require disabling a linter check?

selfdrive/test/process_replay/regen.py
```diff
@@ -24,6 +23,8 @@
 from tools.lib.framereader import FrameReader
 from tools.lib.logreader import LogReader
 
+def tqdm(x):
+  return x
```
adeebshihadeh, line 27: tqdm has a disable arg you can use instead

selfdrive/test/process_replay/regen_all.py
```diff
@@ -1,22 +1,29 @@
...
+
+from selfdrive.test.process_replay.helpers import OpenpilotPrefix
 from selfdrive.test.process_replay.regen import regen_and_save
 from selfdrive.test.process_replay.test_processes import original_segments as segments
 
-if __name__ == "__main__":
-  new_segments = []
-  for segment in segments:
+def regen_job(segment):
+  with OpenpilotPrefix():
```
adeebshihadeh, line 11: how does this work without setting different log directories? did you make sure the logs are valid?

adeebshihadeh: hmm, guess it could work if there's a delay between starting the loggerd instances, though it should still work if they're all started at the same time

lukaspetersson: The segment of each job is unique so why would you need different directories? also, I ran the segments generated in parallel through latency logger and they produce similar looking graphs. Also looked at them in plotjuggler and I think they look valid, but of course I did not look at all datatypes for all routes.

selfdrive/test/process_replay/regen.py
```diff
@@ -178,6 +179,10 @@ def replay_camera(s, stream, dt, vipc_server, frames, size, use_extra_client):
 
 
 def regen_segment(lr, frs=None, outdir=FAKEDATA):
+  if "OPENPILOT_PREFIX" in os.environ:
+    global disable_tqdm
```
adeebshihadeh, line 183: avoid using globals. it's not necessary here

selfdrive/test/process_replay/.gitignore
```diff
@@ -1,2 +1,3 @@
 fakedata/
 debayer_diff.txt
+fakedata*/
```
adeebshihadeh, line 3: remove the original one on line 1?

adeebshihadeh, line 3: remove the original one on line 1?

selfdrive/test/process_replay/regen_all.py
```diff
@@ -1,22 +1,29 @@
...
+def regen_job(segment):
+  with OpenpilotPrefix() as prefix:
     route = segment[1].rsplit('--', 1)[0]
     sidx = int(segment[1].rsplit('--', 1)[1])
-    print("Regen", route, sidx)
-    relr = regen_and_save(route, sidx, upload=True, use_route_meta=False)
-
-    print("\n\n", "*"*30, "\n\n")
-    print("New route:", relr, "\n")
+    relr = regen_and_save(route, sidx, upload=True, use_route_meta=False, outdir=FAKEDATA[:-1]+prefix, disable_tqdm=True)
```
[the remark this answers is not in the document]
adeebshihadeh: also the dongle ID should only be up to 16 chars

selfdrive/test/process_replay/regen.py
```diff
@@ -304,4 +303,4 @@ def regen_and_save(route, sidx, upload=False, use_route_meta=False):
   parser.add_argument("route", type=str, help="The source route")
   parser.add_argument("seg", type=int, help="Segment in source route")
   args = parser.parse_args()
-  regen_and_save(args.route, args.seg, args.upload)
+  regen_and_save(args.route, args.seg, args.upload, outdir=os.path.join(FAKEDATA, "default_dongleid"))
```
adeebshihadeh, line 306:
```suggestion
  regen_and_save(args.route, args.seg, args.upload)
```

PR #24632, opened 2022-05-23
UI: draw map with uiUpdate signal

sshane:
Only draws the map on the global `uiUpdate` signal from `UIState`, just like the onroad window. Should ensure that when we poll on modelV2, map cannot schedule too many events to cause any lag when dragging around in the view. This does however reduce the update rate to only 20Hz when dragging around.


pd0wm:
Think it's fine to sync with the camera during normal driving. During interaction (pan/zoom) I think we should just render as needed to make sure it feels fluid, it's ok to drop a few camera frames when that happens.

selfdrive/ui/qt/maps/map.cc
```diff
@@ -24,16 +24,8 @@ const float MAP_SCALE = 2;
 
 const QString ICON_SUFFIX = ".png";
 
-MapWindow::MapWindow(const QMapboxGLSettings &settings) :
-  m_settings(settings), velocity_filter(0, 10, 0.05) {
-  sm = new SubMaster({"liveLocationKalman", "navInstruction", "navRoute"});
-
-  // Connect now, so any navRoutes sent while the map is initializing are not dropped
-  sm->update(0);
```
adeebshihadeh, line 32: make sure the UIState SubMaster effectively does this too

adeebshihadeh (approved):

selfdrive/ui/qt/maps/map.cc
sshane: It should always, we go onroad from UIState's SubMaster data, and we only called this sm.update() here when we went onroad.

PR #24633, opened 2022-05-23
Register atexit hook to close FrameIterator ffmpeg subprocess

mitchellgoffpc:


geohot:
isn't there a way to set the process as daemon?

gregjhogan:
We should really eliminate using ffmpeg.  I am not convinced that this is any faster than using avcodec directly in a single process/thread.  See c++ framereader for inspiration.

gregjhogan:
It looks pretty straight forward and simple to eliminate the ffmpeg process completely by switching to PyAV (even retaining the ability to specify how many threads are used)
https://github.com/PyAV-Org/PyAV/blob/main/examples/basics/thread_type.py

I think the cuda stuff can just go away, last time I looked I couldn't get better performance with how we are using the video stream.

mitchellgoffpc:
Agreed, I'll close this PR and at some point I'll refactor tools/lib/framereader to use libav instead of ffmpeg.

PR #24634, opened 2022-05-23
acados build script improvements for mac

tecandrew:
**Description**: latest acados rebuilds only build for the arch is on. this adds cmake flags for universal2 libs. 

ref: https://github.com/commaai/openpilot/commit/d9fbec6bf540c06f16f1575df72f3237b5f6a487

- adjust `build.sh`
- adds rust to `mac_setup.sh` for acados rebuilds

**Verification**: check backend arch support using MacOS's `lipo`:

```
~/workspace/openpilot/third_party/acados/Darwin/lib (fix-mac-acados ✗) ls -la
total 12736
drwxr-xr-x  7 atec  staff      224 May 23 15:57 .
drwxr-xr-x  4 atec  staff      128 May 23 15:59 ..
-rwxr-xr-x  1 atec  staff   884288 May 23 15:57 libacados.dylib
-rwxr-xr-x  1 atec  staff  2192144 May 23 15:57 libblasfeo.dylib
-rwxr-xr-x  1 atec  staff  2340992 May 23 15:57 libhpipm.dylib
-rwxr-xr-x  1 atec  staff   546248 May 23 15:57 libqpOASES_e.3.1.dylib
-rwxr-xr-x  1 atec  staff   546248 May 23 15:57 libqpOASES_e.dylib
~/workspace/openpilot/third_party/acados/Darwin/lib (fix-mac-acados ✗) lipo -archs libacados.dylib
x86_64 arm64
```

tested rebuilds on M1 and Intel macs


tools/mac_setup.sh
```diff
@@ -52,6 +53,7 @@ brew "zeromq"
 brew "protobuf"
 brew "protobuf-c"
 brew "swig"
+brew "rustup-init"
```
adeebshihadeh, line 56: rust is pretty big, don't want to include it just for acados since that's checked in

tecandrew: I’ll move the install to the relevant part in build.sh of thirdparty/acados then

adeebshihadeh:
Reverted the binary changes since we like to rebuild those ourselves.

PR #24637, opened 2022-05-24
CI: add build job for latest Ubuntu

adeebshihadeh:


EwoutH:
Actually, while still in beta, the latest Ubuntu virtual environment on GitHub Actions is [Ubuntu 22.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2204-Readme.md).

Biggest change is probably that on Ubuntu 20.04 Clang 11 is the default Clang compiler, while on 22.04 it's Clang 14.

See https://github.com/actions/virtual-environments/issues/5490 for more details (it's slightly outdated).

PR #24640, opened 2022-05-24
test_fw_query_on_routes: make into function

sshane:
Related to https://github.com/commaai/openpilot/issues/23905, so we can run a car port verification script on an rlog and verify it fingerprints offline.

Going to merge once I start using this here: https://github.com/commaai/openpilot/pull/24081 so I can make sure we get everything we need


sshane:
Not really needed, can just call these functions directly

Issue #23644, opened 2022-01-27
Toyota: merge car params and tunes


adeebshihadeh:
Start by merging any params configuration that's common between two cars, e.g. fe5056ffd8dfb6a8ecb1b008a8c77c2b3b2e7511. Then, merge as many tunes as possible in `selfdrive/car/toyota/tunes.py`.

sshane:
https://github.com/commaai/openpilot/pull/24068 will probably close this

adeebshihadeh:
This is for all the params, like mass, wheelbase, etc.

adeebshihadeh:
Params look pretty much merged and tunes will be deleted with torque control.

PR #24641, opened 2022-05-24
Toyota: merge Highlander params

sshane:
Taking a stab at https://github.com/commaai/openpilot/issues/23644. Are these two wheelbases too far apart to merge together? Also a difference of a ~passenger


adeebshihadeh:
@nuwandavek what do you think?

nuwandavek:
I'm assuming the original values are from the car specs, I'd keep it the way it was. I haven't quantified the effect this has (but it does have an effect in the vehicle model), but intuitively seems like any difference in the order of 5-10 cms should be retained. 

I can look into the extent of impact if this is an important change.

sshane:
It's not, but maybe it would be nice to know in general much much of an impact these values make?

PR #24644, opened 2022-05-24
latency logger: add offset to visualize overlap

lukaspetersson:
add offset to visualize overlap


adeebshihadeh:
Hmm, doesn't the lower density doesn't look quite as nice. Let's keep this open and think about it a bit more.

tools/latencylogger/latency_logger.py
```diff
@@ -202,7 +202,10 @@ def graph_timestamps(timestamps, start_times, end_times, relative, title=""):
           points['y'].append(i)
           points['labels'].append(event[0])
     offsets = [[0, -5*j] for j in range(len(service_bars))]
-    ax.broken_barh(service_bars, (i-0.15, 0.3), facecolors=(colors), alpha=0.5, offsets=offsets)
+    if offset_services:
+      ax.broken_barh(service_bars, (i-0.15, 0.3), facecolors=(colors), alpha=0.5, offsets=offsets)
```
adeebshihadeh, line 206: only do this line once and put the if around `offsets`

tools/latencylogger/latency_logger.py
```diff
@@ -226,6 +227,7 @@ def get_timestamps(lr):
   parser.add_argument("--relative", action="store_true", help="Make timestamps relative to the start of each frame")
   parser.add_argument("--demo", action="store_true", help="Use the demo route instead of providing one")
   parser.add_argument("--plot", action="store_true", help="If a plot should be generated")
+  parser.add_argument("--offset", action="store_true", help="Offset service to better visualize overlap")
```
adeebshihadeh, line 230:
```suggestion
  parser.add_argument("--offset", action="store_true", help="Vertically offset service to better visualize overlap")
```

PR #24650, opened 2022-05-25
Add wheelSpeedFactor to Setpoint and Speedometer for Lexus ES Platforms

sunnyhaibin:
On the 2021 Lexus ES300h, the setpoint and speedometer on the openpilot display is lowered by 3.5% when compared to the setpoint and speedometer on the car's dashboard. Adding `wheelSpeedFactor = 1.035` under the Lexus ES platforms and to `cruiseState.speed` in carstate fixes the desynchronization.

Before fix route ID: `36e10531feea61a4|2022-05-24--21-21-37`
After fix route ID: `36e10531feea61a4|2022-05-24--21-00-01`

Below are the pictures of before and after the fix from the cockpit (Left - after, Right - before):

40 MPH setpoint:
image

50 MPH setpoint:
image

60 MPH setpoint:
image

### DIFF
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -93,6 +93,8 @@ class CarState(CarStateBase):
     else:
       ret.cruiseState.available = cp.vl["PCM_CRUISE_2"]["MAIN_ON"] != 0
       ret.cruiseState.speed = cp.vl["PCM_CRUISE_2"]["SET_SPEED"] * CV.KPH_TO_MS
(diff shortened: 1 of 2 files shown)


pd0wm:
The wheel speed factor is not meant to fix the ui. It's the factor between the wheel speeds on CAN and the actual speed from GPS. The HUD applies it's own factor on top of speeds on the CAN bus. 

Some cars have a large error in the wheel speeds on CAN and need that fixes to prevent overshooting the setpoint: https://github.com/commaai/openpilot/issues/2106

PR #24654, opened 2022-05-25
Laikad: Fetch orbit data in thread

gijskoning:
Fetch orbit data in thread such that processing messages is not delayed. 
Only allow Ultra-Rapid orbits. Glonass support will be added in the next pr.

Also update laika

### DIFF
--- a/selfdrive/locationd/laikad.py
+++ b/selfdrive/locationd/laikad.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python3
 import threading
 import time
-from datetime import datetime
 from typing import List
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


selfdrive/locationd/laikad.py
```diff
@@ -23,8 +27,9 @@
 class Laikad:
 
   def __init__(self, use_internet):
-    self.astro_dog = AstroDog(use_internet=use_internet)
+    self.astro_dog = AstroDog(use_internet=use_internet, pull_ephems=['Orbit', 'Nav'])
```
gijskoning, line 30: Will change to enum

selfdrive/locationd/test/test_laikad.py
```diff
@@ -9,6 +10,8 @@
 from selfdrive.test.openpilotci import get_url
 from tools.lib.logreader import LogReader
 
+os.environ["FILEREADER_CACHE"] = "1"
```
pd0wm, line 13: Do not put something that changes the global environment at import time

PR #24658, opened 2022-05-26
Hyundai: longitudinal support for legacy safety

adeebshihadeh:


sshane:
We might need some exceptions in hyundaican, the Optima doesn't have the SCC14 message, which controls things like jerk limits and comfort bands

PR #24661, opened 2022-05-26
process replay: only download used logs

ClockeNessMnstr:
I'm using the whitelist to just run subaru but it re-downloads 13 logs every time. 

is it also possible to just cache these instead based on something? I don't even need to be re-downloading the one I'm using every time I want to do a quick check.


selfdrive/test/process_replay/test_processes.py
```diff
@@ -180,17 +180,16 @@ def format_diff(results, ref_commit):
...
-      p1 = pool.map(get_logreader, [seg for car, seg in segments])
-      for (segment, lr) in tqdm(p1, desc="Getting Logs", total=len(segments)):
-        lreaders[segment] = lr
-
     pool_args: Any = []
     for car_brand, segment in segments:
       if (len(args.whitelist_cars) and car_brand.upper() not in args.whitelist_cars) or \
          (not len(args.whitelist_cars) and car_brand.upper() in args.blacklist_cars):
         continue
+      if not args.upload_only:
```
sshane, line 188: I agree we should only download logs we're testing, but doesn't this just make this single-threaded now?

ClockeNessMnstr: yea, it was quick and dirty. 
updated it now. Hopefully there's also a way to check if the file is already downloaded?

ClockeNessMnstr:
Hopefully there's also a way to check if the file is already downloaded?

selfdrive/test/process_replay/test_processes.py
```diff
@@ -72,9 +72,9 @@ def run_test_process(data):
   return (segment, cfg.proc_name, res)
 
 
-def get_logreader(segment):
+def get_logreader(segment, cache):
```
gijskoning, line 75: This is not needed! Use `os.environ["FILEREADER_CACHE"] = "1"` cache all files downloaded using URLFile which is used by logreader.

I think it would be useful to add these instructions somewhere

ClockeNessMnstr: FILEREADER_CACHE="1" scons -j8 && selfdrive/test/process_replay/test_processes.py --whitelist-cars SUBARU

should this work? 
it's not caching for me

ClockeNessMnstr: Nevermind, i see this doesn't persist so it shouldn't interfere with anything?

ClockeNessMnstr:
Think caching should be good now.

selfdrive/test/process_replay/test_processes.py
```diff
@@ -150,11 +150,15 @@ def format_diff(results, ref_commit):
                       help="Updates reference logs using current commit")
   parser.add_argument("--upload-only", action="store_true",
                       help="Skips testing processes and uploads logs from previous test run")
+  parser.add_argument("--use-cache", action="store_true",
+                      help="Caches log files")
```
adeebshihadeh, line 154: the point of the environment variable is that this arg doesn't need to be added to every script

ClockeNessMnstr: Do you need to set it within a certain folder? 
From openpilot folder:
FILEREADER_CACHE=1 self-drive/test/test_processes....... 
didn't cache

ClockeNessMnstr: Nevermind. 
I ran "FILEREADER_CACHE='1' scons -j8 && selfdrive/test/process_replay/test_processes.py..."
needed  "scons -j8 && FILEREADER_CACHE='1' selfdrive/test/process_replay/test_processes.py..."

PR #24665, opened 2022-05-27
add new fwdRadar for Kia NIRO EV 2022

boli-bohan:
noticed a new fwVersion for the fwdRadar ecu in my 2022 niro ev that was preventing the car from being recognized


sshane:
Do you have a route/dongle id for this car?

BoBowchan:
c906e2a26482bf6a|2022-05-26--22-26-27

PR #24666, opened 2022-05-27
Hyundai longitudinal: set jerk limits to max

sshane:
From my testing on a 2020 Sonata, changing the jerk limits to the maximum value had little effect in both transitioning from 2.0 m/s/s to -3.0 m/s/s instantly, and continually ramping up to 2.0 m/s/s then ramping down to -3.0 m/s/s.

However I think this makes sense conceptually to allow maximum jerk possible, the car should always follow the acceleration request, which with no lag, is already jerk limited from the planner (not including the controller).

Car's jerk request vs our jerk request: https://imgur.com/a/JPSO02U. Will post more results and get testers before merging.

### DIFF
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -104,22 +104,18 @@ class CarController:
             self.last_button_frame = self.frame
 
       if self.frame % 2 == 0 and self.CP.openpilotLongitudinalControl:
-        accel = actuators.accel
-        jerk = 0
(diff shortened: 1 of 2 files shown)


sshane:
Raising the upper jerk doesn't have a good effect while stopping, it allows some slack in the brake controller and we end up creeping forward: `fb34fd9bd05858bc|2022-05-27--18-35-03--5`, https://imgur.com/a/jQYn3WD. Need to think about this

PR #24670, opened 2022-05-27
ui: fix effective low frame rate

sshane:
This also has the added benefit of continuing to display new camera frames when we request a duplicate modeld frame, whereas before we would hang on the current frame for two frames and then skip forward.

No stuttering when using qlogs, and rlogs with skipping around. Adds a tiny bit of complexity, but I think it's better to maintain an index that automatically moves us forward in time for the camera frames and maintains our frame offset when we don't update the model frame, instead of holding on the frame and then jumping to the latest one.


sshane:
The old method would try to draw the exact frame given to it via the model's frame id, however under certain circumstances when UI gets out of sync with the model (or in this case, repeatedly tapping the screen) it can lead to effectively what looks like 10 fps. This is how this PR would handle these cases:

Screenshot from 2022-06-27 23-48-08

sshane:
Will poll on the model to fix

PR #24671, opened 2022-05-28
Hyundai longitudinal: utilize aReqRaw for more responsive control

sshane:
Todo:

- [ ] Run a similar test from https://github.com/commaai/openpilot/issues/24672, but with lower maximum accelerations to verify this improves braking


shahhira:
.github/workflows/selfdrive_tests.yaml

PR #24674, opened 2022-05-28
Hyundai longitudinal: handle mid-press button change

### LINKED ISSUE #24673: Hyundai longitudinal: handle button signal changing mid press
If you hold the accel or decel button and then quickly switch to the other (which is pretty easy actually), the signal for the buttons in CLU11 never goes to 0, so we never add a button event with `decelCruise` or `accelCruise` not being pressed, thus causing its button timer to increment until it's pressed again. It can cause the set speed to jump by 5 mph all the way down to the minimum or maximum allowed cruise speed the next time you engage with any other button.

Example:

```python
buttonEvents: [<car.capnp:CarState.ButtonEvent builder (pressed = true, type = decelCruise)>]
2.0 0.0 deque([2.0, 2.0, 2.0, 0.0], maxlen=4)
buttonEvents: [<car.capnp:CarState.ButtonEvent builder (pressed = false, type = decelCruise)>]
0.0 2.0 deque([0.0, 0.0, 0.0, 2.0], maxlen=4)
buttonEvents: [<car.capnp:CarState.ButtonEvent builder (pressed = true, type = decelCruise)>]
2.0 4.0 deque([2.0, 2.0, 2.0, 4.0], maxlen=4)
buttonEvents: [<car.capnp:CarState.ButtonEvent builder (pressed = true, type = cancel)>]
4.0 0.0 deque([4.0, 4.0, 4.0, 0.0], maxlen=4)
buttonEvents: [<car.capnp:CarState.ButtonEvent builder (pressed = false, type = cancel)>]
0.0 4.0 deque([0.0, 0.0, 0.0, 4.0], maxlen=4)
```

sshane:
Should probably common-ize similar code in GM and Honda


selfdrive/car/hyundai/interface.py
```diff
@@ -14,6 +14,18 @@
 ENABLE_BUTTONS = (Buttons.RES_ACCEL, Buttons.SET_DECEL, Buttons.CANCEL)
 
 
+def get_button_type(but):
+  if but == Buttons.RES_ACCEL:
```
adeebshihadeh, line 18: dict?

sshane: Did that in this PR: https://github.com/commaai/openpilot/pull/24678 so this PR can be a two line change

PR #24677, opened 2022-05-29
Toyota: add missing engine FW version for 2022 RAV4 Hybrid

sshane:
From dongle id e16e4f4ef04a4ab3


sshane:
Ran through test_models, no issues.

PR #24679, opened 2022-05-30
Simulator: Speed up process by removing second camera

gijskoning:
Modeld seems to still work even though it gets the same image for road and wide camera. I think it is fine to do this and not care about a correct wide image. The simulator is just to try out openpilot on PC and that works now while being a bit faster such that other users can easily use it too.


tools/sim/bridge.py
```diff
@@ -522,17 +528,23 @@ def run(self, queue, retries=-1):
...
+      from tools.sim.lib.keyboard_ctrl import keyboard_poll_thread
+
+      keyboard_poll_thread(q)
+    p.join()
 
-    keyboard_poll_thread(q)
-  p.join()
+  finally:
+    # Try cleaning up the wide camera param
+    # in case users wants to use replay after
```
pd0wm, line 549:
```suggestion
    # in case users want to use replay after
```

PR #24683, opened 2022-05-30
carstate cleanup

crwusiz:


adeebshihadeh:
We generally don't merge this kind of stylistic cleanup. These choices are up to the original author, and if necessary, we'll request changes when it's still in a PR.

PR #24460, opened 2022-05-07
Add Chrysler FPv2 requests and logging

realfast:
Other than the ECM and TCM, the Chrylser response offset for most modules is -0x280.

$F132 gives the ECU software part number

PR #24685, opened 2022-05-31
Chrysler: add missing 2022 Pacifica Hybrid fingerprint

TheWizard1328:
Missing FP for 2022 PacHy's

2022 PacHy would consistently FP as a 2019 and fail to even register as a valid vehicle on start up. 80% startups threw a car not detected dash can error.

I have tested this on my 2022 PacHy for the last week now. Priot to these changes I received repeated Car not detected, Dash cam errors on about 80% of my start ups. All I get now are the occasional low speed LKAS warnings. The reason I rem'ed out the 2019 FP (# Based on "8190c7275a24557b|2020-02-24--09-57-23") was because after rigorous testing this was found to be the one conflicting with my FP and printing my van as a 2019 PacHy. Only was to FP as a 2022 was to rem out this one.

**Route**
8a84e5027e568485|2022-05-28--04-30-58--0 - Route prior to fix
8a84e5027e568485|2022-05-31--02-02-18--0 - Route after fix

**Checklist**
- [ ] route with openpilot:


sshane:
I don't see a real difference between the other Pacifica Hybrids and this one, is there a reason you can't just add fw versions for your car?

TheWizard1328:
Without this FP the 2022 PacHy fails to engage or recognize the van on 90% of starts. I almost always get the car not recognized. Dashcam Mode. Some times I had to restart 5 or 6 times, But nit I never get it. If I leave in the 2019 FP and add my FP to the 2019 or as I did in my PR it will always take the existing 2019 FP. But if I rem out the 2019 it works fine.
I have tried FPv1 and FPv2 and only way that worked on getting a FP was FPv1 I believe.

sshane:
Can you try this update I pushed? If it doesn't detect your car let me know with a route please

sshane:
You've already added them, but we need someone with a C3 to test this PR before merging, unless we merge this PR first to move to FW fingerprinting, then we'd probably want someone with a C3 to get FW fingerprints https://github.com/commaai/openpilot/pull/24460

adeebshihadeh:
@sshane let's get this in for 0.8.15. No need for a comma three to verify this

sshane:
Looks like we already support some 2022 (see dongle 4395faadf0a373f2)

sshane:
Not sure why the GH PR diff isn't showing the CARS.md changes, despite them being there on the branch. Will re-open

selfdrive/car/chrysler/values.py
```diff
@@ -89,6 +89,10 @@ class ChryslerCarInfo(CarInfo):
...
...: 8, 672: 8, 678: 8, 680: 8, 683: 8, 701: 8, 703: 8, 704: 8, 705: 8, 706: 8, 709: 8, 710: 8, 711: 8, 719: 8, 720: 6, 729: 5, 736: 8, 737: 8, 738: 8, 746: 5, 752: 2, 754: 8, 760: 8, 764: 8, 766: 8, 770: 8, 773: 8, 779: 8, 782: 8, 784: 8, 792: 8, 793: 8, 794: 8, 795: 8, 796: 8, 797: 8, 798: 8, 799: 8, 800: 8, 801: 8, 802: 8, 803: 8, 804: 8, 805: 8, 807: 8, 808: 8, 816: 8, 817: 8, 820: 8, 825: 2, 826: 8, 832: 8, 838: 2, 848: 8, 853: 8, 856: 4, 860: 6, 863: 8, 878: 8, 882: 8, 886: 8, 897: 8, 906: 8, 908: 8, 924: 8, 926: 3, 937: 8, 938: 8, 939: 8, 940: 8, 941: 8, 942: 8, 943: 8, 947: 8, 948: 8, 958: 8, 959: 8, 962: 8, 969: 4, 973: 8, 974: 5, 979: 8, 980: 8, 981: 8, 982: 8, 983: 8, 984: 8, 992: 8, 993: 7, 995: 8, 996: 8, 1000: 8, 1001: 8, 1002: 8, 1003: 8, 1008: 8, 1009: 8, 1010: 8, 1011: 8, 1012: 8, 1013: 8, 1014: 8, 1015: 8, 1024: 8, 1025: 8, 1026: 8, 1031: 8, 1033: 8, 1050: 8, 1059: 8, 1082: 8, 1083: 8, 1098: 8, 1100: 8, 1216: 8, 1218: 8, 1220: 8, 1225: 8, 1235: 8, 1242: 8, 1246: 8, 1250: 8, 1251: 8, 1252: 8, 1284: 8, 1536: 8, 1568: 8, 1856: 8, 1858: 8, 1860: 8, 1863: 8, 1865: 8, 1875: 8, 1882: 8, 1886: 8, 1890: 8, 1891: 8, 1892: 8, 2016: 8, 2018: 8, 2020: 8, 2024: 8, 2026: 8, 2028: 8
```
sshane, line 95: The only missing address from above is 1536

TheWizard1328: Without the FP from my 2022 PacHy My C2 refuses to detect my PacHy. When it does it keeps failing to connect to systems or disconnects while driving.

sshane: Can you try this change on a clean C2 branch? It should work

PR #24690, opened 2022-05-31
Hyundai Sonata 2020: use torque controller

sshane:
Will get routes of before and after, felt much more confident on the highway with less oscillation in turns. Weirdly, it felt like friction was too low however, I'll see if it's accurately following desired accel

### DIFF
--- a/selfdrive/car/hyundai/interface.py
+++ b/selfdrive/car/hyundai/interface.py
@@ -61,13 +61,16 @@ class CarInterface(CarInterfaceBase):
       ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[9., 22.], [9., 22.]]
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.2, 0.35], [0.05, 0.09]]
     elif candidate in (CAR.SONATA, CAR.SONATA_HYBRID):
-      ret.lateralTuning.pid.kf = 0.00005
       ret.mass = 1513. + STD_CARGO_KG
       ret.wheelbase = 2.84
       ret.steerRatio = 13.27 * 1.15   # 15% higher at the center seems reasonable
       tire_stiffness_factor = 0.65
-      ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0.], [0.]]
-      ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.25], [0.05]]
+      if candidate == CAR.SONATA:
+        set_torque_tune(3.0, 0.01)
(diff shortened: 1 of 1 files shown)


sunnyhaibin:
Test route with rlogs uploaded:
```36e10531feea61a4|2022-05-31--21-12-42```

Test car is 2021 Hyundai Sonata with the N Line trim, which its steering ratio `12.35:1` instead of `13.27:1` from all other trims, so not sure how much that contributed to the oscillation. I didn't edit `steerRatio` in interface.py. I could get another test with `steerRatio` edited for my car if needed.

haram-KONA:
My Car is not a Sonata, but I'm posting my test drive in case it might be of some help.

Hyundai KONA Hybrid: use torque controller
route id : 9a3aa85e86f9c3fc|2022-06-01--11-21-24

Test route with rlogs uploaded: 9a3aa85e86f9c3fc|2022-06-01--11-21-24
(max_lat_accel = 3.15, friction = 0.01)

Test car is 2020 Hyundai KONA Hybrid, which its steering ratio is 13.42, 
I didn't edit steerRatio in interface.py. 

There is almost no oscillation when driving on a straight road, 
and the oscillation is significantly reduced when driving in a curve.

sshane:
@haram-KONA feel free to open a separate PR for that car

speedking456:
I was able to get a decent clip of max lateral in my Kia Seltos here:

image

Max accel seems to be around 2.50 - 2.55 like what Shane found. Increasing the friction made a marginal difference, decreasing from 3 -> 2.55 has been the best improvement so far, feels like it's driving exactly like it should. Still sometimes undershoots sharper curves, especially at 30-40 km/h, but driving straight and around mild bends is buttery.

james5294:
two routes:
first route laneless
second lanelines
seemed better with lanelines
Car: 2022 Sonata Limited

d466fc4afcd0a050|2022-06-02--00-53-49--0
d466fc4afcd0a050|2022-06-02--01-04-08--0

sshane:
Here's what I found from @sunnyhaibin and @james5294's routes https://imgur.com/a/8FB4Yfq

james5294:
re-run same route after change (+ straight highway run added-on)

Car: `2022 Sonata Limited`
route: `d466fc4afcd0a050|2022-06-02--09-00-38--5`

james5294:
route after change

Car: `2022 Sonata Limited`
route: `d466fc4afcd0a050|2022-06-02--19-29-54--0`

sshane:
Seems like different user's Sonatas can have different max lat accels (2.5 to 3.5)

sshane:
Closing until we can figure out if we need multiple params per EPS fw version

PR #24691, opened 2022-06-01
simulator: run simulator test in ci

maxime-desroches:


tools/sim/launch_openpilot.sh
```diff
@@ -5,7 +5,13 @@ export NOBOARD="1"
 export SIMULATION="1"
 export FINGERPRINT="HONDA CIVIC 2016"
 
-export BLOCK="camerad,loggerd,encoderd"
+PROC="camerad,loggerd,encoderd"
+if [[ "$CI" ]]
+then
+  PROC="${PROC},navd,ui"
```
adeebshihadeh, line 11: why doesn't navd work? for UI, I think there's some kind of offscreen thing we can do with Qt

Jenkinsfile
```diff
@@ -75,6 +75,27 @@ pipeline {
       }
 
       stages {
+        stage('simulator') {
+          agent { docker { image 'ghcr.io/commaai/openpilot-base'; args '--user=root' } }
+          steps {
+            sh "git config --global --add safe.directory ${WORKSPACE}"
+            sh "git submodule | awk '{system(\"git config --global --add safe.directory ${WORKSPACE}/\"\$2)}'"
+            sh "git submodule update --init --recursive --force --depth 1"
```
adeebshihadeh, line 83: I setup jenkins to pull the submodules. should be able to remove these
```suggestion
```

Jenkinsfile
```diff
@@ -75,6 +75,27 @@ pipeline {
       }
 
       stages {
+        stage('simulator') {
+          agent { docker { image 'ghcr.io/commaai/openpilot-base'; args '--user=root' } }
```
adeebshihadeh, line 79: any reason this should use openpilot-base? we should use something that already has docker. perhaps we can use GitHub Action's ubuntu runner image if they publish it?

tools/sim/launch_openpilot.sh
```diff
@@ -5,7 +5,13 @@ export NOBOARD="1"
...
 export FINGERPRINT="HONDA CIVIC 2016"
 
-export BLOCK="camerad,loggerd,encoderd"
+PROC="camerad,loggerd,encoderd"
+if [[ "$CI" ]]
+then
+  PROC="${PROC},ui"
+fi
+
+export BLOCK=$PROC
```
adeebshihadeh, line 14:
```suggestion
export BLOCK="camerad,loggerd,encoderd"
if [[ "$CI" ]]; then
  # TODO: offscreen UI should work
  export BLOCK="${BLOCK},ui"
fi
```

tools/sim/start_carla.sh
```diff
@@ -3,7 +3,7 @@
 # Requires nvidia docker - https://github.com/NVIDIA/nvidia-docker
 if ! $(apt list --installed | grep -q nvidia-container-toolkit); then
   read -p "Nvidia docker is required. Do you want to install it now? (y/n)";
-  if [ "${REPLY}" == "y" ]; then
+  if [[ "${REPLY}" == "y" || $CI ]]; then
```
adeebshihadeh, line 6: is this slow? if not, it's nice to test in CI, otherwise we should put this in the docker image we build in

Jenkinsfile
```diff
@@ -75,6 +75,27 @@ pipeline {
...
+          steps {
+            sh "git config --global --add safe.directory ${WORKSPACE}"
+            sh "git submodule | awk '{system(\"git config --global --add safe.directory ${WORKSPACE}/\"\$2)}'"
+            sh "git submodule update --init --recursive --force --depth 1"
+            sh "git lfs pull"
+            sh "curl -fsSL https://get.docker.com -o get-docker.sh && sudo sh get-docker.sh"
+            sh "${WORKSPACE}/tools/sim/build_container.sh"
+            sh "CI=1 ${WORKSPACE}/tools/sim/start_carla.sh"
```
adeebshihadeh, line 87: the test should take care of this

Jenkinsfile
```diff
@@ -75,6 +75,27 @@ pipeline {
...
+            sh "CI=1 ${WORKSPACE}/tools/sim/start_openpilot_docker.sh"
```
adeebshihadeh, line 88: these two should just be replaced by running the test

Jenkinsfile
```diff
@@ -75,6 +75,27 @@ pipeline {
...
+          }
+
+          post {
+            always {
+              sh "docker kill carla_sim || true"
```
adeebshihadeh, line 93: let's make sure the test properly cleans this up if it doesn't already

PR #24695, opened 2022-06-01
Toyota nodsu(radar dsu tss1) longitudinal control support

Smartype:
Enable longitudinal control for CHR, CHR-H, CAMRY, CAMRY-H. Radar Canfilter hardware required. (https://github.com/Smartype/canfilter)

For cars which are low_speed_locked at 28mph, the canfilter board changes the acc_type to 1. But these cars ignore brake requests under 15mph, possibly caused by abs/esc. For those cars, canfilter prevents engagement below 19mph and force disengagement below 15mph(remove lead car bit). 

This continental radar outputs raw tracks(OBJECT_x in dbc) and clusters.

The raw tracks are sent at 15Hz. It detects objects earlier(far away) but it is a bit noise.

The clusters are filtered/optimize clusters, at 10Hz, might be based on both radar and camera data. Each cluster (msg id) has a fixed relative position. front, front of front, left, front of left, right, front of right. So 0x680 is always the front lead car. If it leaves the front position without lost tracking, it will be moved to other msg ids(left, right, ...), we may add a "relative position" attribute to the liveTrack.

- route with openpilot: 2f37c007683e85ba|2022-05-31--20-30-11

- route with stock system: 2f37c007683e85ba|2022-06-01--18-57-04


.gitmodules
```diff
@@ -1,9 +1,11 @@
 [submodule "panda"]
 	path = panda
-	url = ../../commaai/panda.git
+	url = https://github.com/Smartype/panda.git
+	branch = toyota_nodsu_long
```
cydia2020, line 4: This isn't the correct way to do it. You should open a PR in the submodule's repository, in your local openpilot submodule's repo add your remote and checkout the PR's commit, then push it.

selfdrive/car/toyota/values.py
```diff
@@ -1920,7 +1920,8 @@ class ToyotaCarInfo(CarInfo):
             CAR.LEXUS_RX_TSS2, CAR.LEXUS_RXH_TSS2, CAR.HIGHLANDER_TSS2, CAR.HIGHLANDERH_TSS2, CAR.PRIUS_TSS2, CAR.CAMRY_TSS2, CAR.CAMRYH_TSS2,
             CAR.MIRAI, CAR.LEXUS_NX_TSS2, CAR.ALPHARD_TSS2, CAR.AVALON_TSS2, CAR.AVALONH_TSS2, CAR.ALPHARDH_TSS2}
 
-NO_DSU_CAR = TSS2_CAR | {CAR.CHR, CAR.CHRH, CAR.CAMRY, CAR.CAMRYH}
+RADAR_ACC_CAR_TSS1 = {CAR.CHR, CAR.CHRH, CAR.CAMRY, CAR.CAMRYH}
```
sshane, line 1923: Radar ACC generally means here that the radar is sending the ACC messages

selfdrive/car/toyota/values.py
```diff
@@ -1876,10 +1876,10 @@ class ToyotaCarInfo(CarInfo):
   CAR.LEXUS_RXH: dbc_dict('toyota_tnga_k_pt_generated', 'toyota_adas'),
   CAR.LEXUS_RX_TSS2: dbc_dict('toyota_nodsu_pt_generated', 'toyota_tss2_adas'),
   CAR.LEXUS_RXH_TSS2: dbc_dict('toyota_nodsu_pt_generated', 'toyota_tss2_adas'),
-  CAR.CHR: dbc_dict('toyota_nodsu_pt_generated', 'toyota_adas'),
-  CAR.CHRH: dbc_dict('toyota_nodsu_pt_generated', 'toyota_adas'),
-  CAR.CAMRY: dbc_dict('toyota_nodsu_pt_generated', 'toyota_adas'),
-  CAR.CAMRYH: dbc_dict('toyota_nodsu_pt_generated', 'toyota_adas'),
+  CAR.CHR: dbc_dict('toyota_nodsu_pt_generated', 'toyota_nodsu_adas'),
```
cydia2020, line 1879: Did u include this in `openpilot/release/files_common`?

Smartype:
@sshane  can you please with the test error msg? it looks like the test route can not be loaded. But I am able to view it via cabana. It was shared from a Camry owner.

adeebshihadeh:
This is a lot of code to support custom HW. We allocated a couple of lines for the smartDSU, and I think that's an appropriate amount for this too. This should also be split up into longitudinal support and radar support PRs.

PR #24700, opened 2022-06-01
ui: skip texture frame copy

jwooning:


jwooning:
There still seems to be an issue were EGL claims some ownership of the passed in fd, closing it on calling `eglDestroyImage`. Not destroying is also not an option, since there is some limited amount of images that can be created, besides possible memory leaks.

Might need to destroy the images after the fd is closed, or possibly pass in a duplicated fd to `eglCreateImageKHR`.

PR #24701, opened 2022-06-01
update DH names + notes for MPC output curvatures

ClockeNessMnstr:
"current_curvature" is not the correct description of what the MPC is outputting in it's curvature_ego state.
The MPC is integrating it's free variable, curvature_rate, such that curvature[0] is actually the desired_curvature before any delay.
inversely: the curvature_rate_desired is the desired rate of change to the setpoint and not the actual curvature rate. 

If we were to set the initial curvature = measured curvature in the MPC initiation these names would be correct. 
This was possibly how it was initially set up but the nomenclature here is now confusing.

### DIFF
--- a/selfdrive/controls/lib/lateral_planner.py
+++ b/selfdrive/controls/lib/lateral_planner.py
@@ -79,9 +79,9 @@ class LateralPlanner:
                      y_pts,
                      heading_pts)
     # init state for next
-    # mpc.u_sol is the desired curvature rate from x0[3]. 
-    # we integrate here so that x_sol[:,3] is the desired curvature for lat_control
-    # if we set x0[3] = measured_curvature the output would be desired_curvature_rate
+    # mpc.u_sol is the desired curvature rate given x0 curv state. 
+    # with x0[3] = measured_curvature, this would be the actual desired rate.
+    # instead, interpolate x_sol so that x0[3] is the desired curvature for lat_control.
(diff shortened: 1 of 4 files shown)


selfdrive/controls/lib/lateral_planner.py
```diff
@@ -79,6 +79,9 @@ def update(self, sm):
                      y_pts,
                      heading_pts)
     # init state for next
+    # mpc.u_sol is the desired curvature rate from x0[3]. 
+    # we integrate here so that x_sol[:,3] is the desired curvature for lat_control
+    # if we set x0[3] = measured_curvature the output would be desired_curvature_rate
```
HaraldSchafer, line 84: I don't understand this whole comment block. Integrate or Interpolate.

HaraldSchafer:
LGTM besides that comment block

selfdrive/controls/lib/lateral_planner.py
ClockeNessMnstr: Wanted to clarify it's interpolating over the integral of rate, since rate is really the MPC's "output" state. 

This is better maybe:

# mpc.u_sol is the desired curvature rate given x0 curv state. 
# with x0[3] = measured_curvature, this would be the actual desired rate.
# instead, interpolate x_sol so that x0[3] is the desired curvature for lat_control.

HaraldSchafer: Yeah that makes sense!

PR #24704, opened 2022-06-01
Kia Optima: use torque controller

sshane:
Will need to get test route to verify max torque


adeebshihadeh:
Looks like this is already on master

PR #24707, opened 2022-06-02
Hyundai Kona Hybrid 2020: use torque controller

sshane:
User claims the car took this curve on its own with no overriding: https://connect.comma.ai/9a3aa85e86f9c3fc/1654135225953/1654135287942 

Looked at this segment manually and it was able to hit 3.9 m/s/s at 15 m/s:
Screenshot from 2022-06-02 00-16-15

Another route at 3.6 m/s/s, still overshooting (max torque too low): https://connect.comma.ai/9a3aa85e86f9c3fc/1654157401617/1654157440571

Seems like this thing can take turns: https://imgur.com/a/XXDiXSZ (possibly 4.0 m/s/s is still too low, from the fit)

### DIFF
--- a/selfdrive/car/hyundai/interface.py
+++ b/selfdrive/car/hyundai/interface.py
@@ -131,7 +131,7 @@ class CarInterface(CarInterfaceBase):
       ret.steerRatio = 13.42  # Spec
       tire_stiffness_factor = 0.385
       if candidate == CAR.KONA_HEV:
-        set_torque_tune(ret.lateralTuning, 3.6, 0.01)
+        set_torque_tune(ret.lateralTuning, 4.0, 0.01)
       else:
         ret.lateralTuning.pid.kf = 0.00005
         ret.lateralTuning.pid.kiBP, ret.lateralTuning.pid.kpBP = [[0.], [0.]]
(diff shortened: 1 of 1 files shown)


sshane:
Another high lateral acceleration with very little torque example, this thing can take turns: https://connect.comma.ai/9a3aa85e86f9c3fc/1654157371480/1654157441457
Screenshot from 2022-06-02 02-09-19
 
Calculated manually from a few high points from the route, and that script, and the max lateral accel seems to be around 3.8 m/s/s

haram-KONA:
max_lat_accel 4.4 and friction 0.01 is best !! 
route id : 9a3aa85e86f9c3fc|2022-06-03--12-58-55

haram-KONA:
It seems that the case of driving max_lat_accel with 4.5 is more stable than case of 4.4. 
I am adding a driving route-id, so please check it and if there is no problem, 
please set the max_lat_accel to 4.5 and friction 0.01 for Hyundai KONA HYBRID 2020.

route id : 0665dbbc79ef999e|2022-06-09--16-03-20

thank you !!!

sshane:
Done in https://github.com/commaai/openpilot/pull/24818

Issue #23144, opened 2021-12-06
Improve comma three offroad thermals


pd0wm:
- [x] Switch to powersave governor when offroad
- [ ] Look into Quectel power savings
- [x] Lower fan control setpoint

adeebshihadeh:
I think we should also re-evaulate the threshold to go onroad. Currently it's set at 79.5C, but a device that goes onroad at that temperature will either narrowly avoid going into the red band or immediately jump into it. We'll either need to keep the device cooler while it's offroad or build a better experience when we don't immediately go onroad. 

![image](https://user-images.githubusercontent.com/8762862/165426729-4f368722-118d-4011-861f-477805355a1e.png)

jyoung8607:
`3cfdec54aa035f3f|2022-05-29--16-38-08` is an example of the problem Adeeb described, a startup from *just* below the offroad danger temp that goes *just* into the red a short time after, resulting in a very long delay before openpilot was engageable.

Hubblesphere:
Lapse in overheat protection. Off road overheated able to engage temporarily when going onroad before detecting an overheated condition again.

Route:  8d8427a2116eace2|2022-07-22--13-00-54--0

jyoung8607:
Another example: `3cfdec54aa035f3f|2022-08-24--15-21-31`

goodchoom:
When the temperature outside is 70°F or hotter and my car has been sitting outside for an extended period of time, I always remote start my car with the A/C set to LO several minutes in advance so by the time I enter my car it's comfortable for me to drive. Despite the cabin having cooled down, I am sometimes unable to engage openpilot because it went onroad just below the offroad danger temp and then overheated while the cabin was hot. When this happens I have to wait/drive manually for several more minutes until the device cools down enough to be able to engage openpilot again.

Here is a route of this issue occuring: 4f3e4f2f6d94e940|2023-04-20--12-27-08--0

adeebshihadeh:
> When the temperature outside is 70°F or hotter and my car has been sitting outside for an extended period of time, I always remote start my car with the A/C set to LO several minutes in advance so by the time I enter my car it's comfortable for me to drive. Despite the cabin having cooled down, I am sometimes unable to engage openpilot because it went onroad just below the offroad danger temp and then overheated while the cabin was hot. When this happens I have to wait/drive manually for several more minutes until the device cools down enough to be able to engage openpilot again.
> 
[...]

PR #24708, opened 2022-06-02
thermald: consider pmic in component temp management

jyoung8607:
For an offroad C3 sitting in the sun, the PMIC temperature appears to be the highest sensed on the device, but it currently isn't considered for ThermalStatus changes. Add it to the list of monitored components, to give the device a better chance at staying offroad if it's too heat-soaked to startup without going into the red zone.

Doesn't resolve, but may help with #23144.

Route: `3cfdec54aa035f3f|2022-05-29--16-38-08`

image


adeebshihadeh:
Reverted due to some PMICs being consistently far hotter than the rest of the components. These devices generally work fine, but this would make those devices unnecessarily delay startup and spend more time in the red thermal zone.

PR #24710, opened 2022-06-02
correctly parse UDS VIN response for Honda Bosch

### LINKED ISSUE #24705: Boardd: VIN Sanity Check Failed
### Describe the bug

With this change https://github.com/commaai/openpilot/pull/24311 boardd.cc is failing to pass 'assert(value_vin.size() == 17);'

https://gyazo.com/c1e666b6eed9f229305a23fba124947b

I have a 20' Honda Civic Hatchback Sport.

After reverting the change, I was able to start my car without issues. With this failure I am getting "awaiting controls to start", and my C3 fails to go off-road after turning off my car. Interestingly enough, my VIN is being logged right above the failure in the screenshot posted.

f1055ddb93ff2d3b

### openpilot version

c677c6b164fc8e642ca75809213eca0678cf86ac

### Additional info

_No response_

pd0wm:
resolves #24705


sshane:
This can fall back to OBD, would this work for that?

PR #24711, opened 2022-06-02
VW MQB: EA driver inactivity workaround

### LINKED ISSUE #23274: VW MQB: Conflict between openpilot and Emergency Assist
### Describe the bug

When openpilot is engaged, the stock ACC setpoint sometimes decreases to zero without any driver-discernable cause.

For cars that support [Emergency Assist](https://www.youtube.com/watch?v=vEx5zazFLZ8), the camera detects long stretches of low driver input torque and tries to take various actions to rouse the driver, and failing that, bring the car to a safe stop in the safest lane it can identify.

Most of those warnings and actions are ignored because openpilot is replacing the involved messages, but in the case of gateway (J533) integration, the camera and ACC radar are still in direct communication, and the ACC radar will see and and execute an EA request to stop the car. This won't happen for openpilot users integrated at the camera.

comma three, comma two, EON Gold

Any VW MQB with Emergency Assist

0cd0b7f7e31a3853|2021-12-20--03-34-26

### openpilot version

0.8.12

### Additional info

Thanks to @Saber422 for reproducing the issue and supplying not only a route, but an idea for the probable cause!

I'm not sure why we haven't seen more complaints of this nature in the past. That said, EA is only a few years old, and for a given model/year it isn't necessarily enabled for all import regions. There are multiple EA behavior variants, for different markets and legal homologation requirements, and I think only one variant may be causing us problems.

The simplest fix is to placate Emergency Assist; we'll tell it there's driver input torque so EA never kicks in. I plan to do this by filtering LH_EPS_03 and rewriting EPS_Lenkmoment (driver input torque) with the sum of true driver input torque and openpilot's own steering output. This will require a matching Panda safety update.

jyoung8607:
Resolves #23274. We're now filtering and resending LH_EPS_03, so a process_replay delta is expected.

- [x] requires commaai/panda#955
- [x] end user testing


jyoung8607:
An affected end user reports this fix was successful. This PR and the associated Panda PR are ready for review.

Screenshot from 2022-06-03 23-05-30

jyoung8607:
Returning to draft for the moment. Another test user is still managing to set off EA on a long level stretch where openpilot is keeping the car straight with less than 0.5 Nm of control output. Will add an interp to scale up the openpilot part of the augmented driver torque signal in lower ranges and retest.

image

jyoung8607:
My attempt to be clever and sum driver input and openpilot output turns out to do weird things. Inertia combined with high frequency PIF tune control and/or driver hands-on-wheel resulted in opposing (and thus cancelling) outputs and inputs, which then weren't large enough to reset inactivity detection.

We're currently testing a version that stops trying to get fancy, and just sends the greater of actual driver input and doubled openpilot output. The output doubling is because VW's inactivity detection is a little too aggressive, and will sometimes trigger even in stock form with gentle inputs on straight level roads.

jyoung8607:
The EA mitigation in this PR is functional, but I also want to throw up a flag when EA pops off (maybe accFaulted, maybe something new) so that we can rely on comma's MTBF reporting to tell us if the mitigation needs further tuning.

- [ ] Depends on EA signal support from commaai/openpilot#27757

jyoung8607:
@adeebshihadeh once the HCA cleanup is in, I want to add detection of stock EA kicking in. I have a workaround (in this PR) that I *believe* will be effective, but I want to verify it always works for everyone.

The condition to monitor, on bus 2, is: `bool(HCA_01.EA_Ruckfreigabe) or HCA_01.EA_ACC_Sollstatus > 0`.

What's the easiest way to surface that in comma's MTBF reports? Should I just add something to carState (like stockFcw and stockAeb), or do you need a full Event? At this time I don't need openpilot to do anything, I just want to know if you ever see segments with that condition. I considered overloading `accFaulted`, but that would jumble it in with things like the stock ACC radar being occluded by snow or ice.

jyoung8607:
Out of draft, ready for review. If possible, I'd like this to be reviewed in time for 0.9.4. Complaints tracing back to this issue have gone up quite a lot as MY2023 has brought EA to several VW and Audi models in North America.

The core problem is documented in #23274. Summary of this PR:

* When openpilot is engaged, pacify EA by sending simulated driver input toward the stock LKA camera
* Monitor efficacy of the solution by setting `CarState.carFaultedNonCritical` if we detect EA activation

This PR should fix the problem while openpilot is engaged, and has been ad-hoc tested by several users. If we see segments with carFaultedNonCritical while openpilot is engaged, we'll know it needs further tuning.

This PR does NOT attempt to fully restore stock EA functionality while openpilot is disengaged. It's technically possible, but would involve selective pass-through of stock LKA functionality, which is another discussion entirely.

jyoung8607:
Moved to Draft for the moment, but still moving forward. I've asked for feedback from a few more affected drivers. If the current version of this change works for them over the next few days, I'll release for review.

jyoung8607:
User feedback has been good, releasing from Draft. I request this PR be considered for 0.9.6-release. It should be low risk, and this is by far the most frequently reported problem with VW on stock openpilot.

@adeebshihadeh I would like to be notified if you see `carFaultedNonCritical` on VW similar to `steerTimeLimit`.

selfdrive/car/volkswagen/carcontroller.py
```diff
@@ -65,6 +65,15 @@ def update(self, CC, CS, ext_bus, now_nanos):
...
       can_sends.append(self.CCS.create_steering_control(self.packer_pt, CANBUS.pt, apply_steer, hca_enabled))
 
+      if self.CP.flags & VolkswagenFlags.STOCK_HCA_PRESENT:
+        # Pacify VW Emergency Assist driver inactivity detection by changing its view of driver steering input torque
+        # to the greatest of actual driver input or 2x openpilot's output (1x openpilot output is not enough to
+        # consistently reset inactivity detection on straight level roads). See commaai/openpilot#23274 for background.
+        ea_simulated_torque = clip(apply_steer * 2, -self.CCP.STEER_MAX, self.CCP.STEER_MAX)
+        if abs(CS.out.steeringTorque) > abs(ea_simulated_torque):
+          ea_simulated_torque = CS.out.steeringTorque
+        can_sends.append(self.CCS.create_eps_update(self.packer_pt, CANBUS.cam, CS.eps_stock_values, ea_simulated_torque))
```
adeebshihadeh, line 75: you're sure that this message is only read by the camera and only for this feature?

selfdrive/car/volkswagen/carcontroller.py
```diff
@@ -65,6 +65,15 @@ def update(self, CC, CS, ext_bus, now_nanos):
       self.apply_steer_last = apply_steer
       can_sends.append(self.CCS.create_steering_control(self.packer_pt, CANBUS.pt, apply_steer, hca_enabled))
 
+      if self.CP.flags & VolkswagenFlags.STOCK_HCA_PRESENT:
+        # Pacify VW Emergency Assist driver inactivity detection by changing its view of driver steering input torque
+        # to the greatest of actual driver input or 2x openpilot's output (1x openpilot output is not enough to
+        # consistently reset inactivity detection on straight level roads). See commaai/openpilot#23274 for background.
```
adeebshihadeh, line 71: can we also consider controlsd's force_decel?

adeebshihadeh (commented):
Sure, I'll let you know if it pops up.

selfdrive/car/volkswagen/carcontroller.py
jyoung8607: Here on `CANBUS.cam` the camera should be the the only consumer for the EPS driver torque signal. There may be other consumers for the other signals we pass through from the same message.

selfdrive/car/volkswagen/carcontroller.py
jyoung8607: In stock ACC mode, with a gateway harness, unfortunately we're in the wrong bus position to MITM the command that changes the ACC setpoint to zero. Otherwise we'd mitigate the problem that way instead.

In openpilot longitudinal mode it'll be very possible to reimplement the VW Emergency Assist experience keyed off force_decel (among other inputs) but that's an entirely different PR for another day.

jyoung8607: Technically we could just NOT run the torque input mitigation if in force_decel, but EA would only kick in 30-60 seconds after we stop mitigating, if it kicks in at all. I think it would be much too unreliable to set any user/code expectation of it working.

adeebshihadeh: gonna merge this, but I suspect there should be a way to do it for the stock ACC case. what if we only reset it at the minimum interval conditional on attentiveness?

jyoung8607: Right now the chance of EA kicking in (before this mitigation) were kinda random, it hits some people more than others. I suppose we could do it keyed on force_decel and then use the mitigation code to send zero driver torque to increase the probability EA is triggered. But it would only work on the subset of cars configured for EA.

I would rather spend the time on a quality EA experience for openpilot longitudinal. Of course it already works to slow the car, but I'd like to add back the UI alerts, the brake-jerk/seatbelt-jerk functions, and the hazard-light function.

adeebshihadeh: sounds good

andiradulescu:
Instead of continuing work on EA, can’t we just add the option to enable/disable EA in [vw_mqb_config.py](https://github.com/commaai/openpilot/blob/master/selfdrive/debug/vw_mqb_config.py)? I could even try myself to write the code for this, just that I can only test it on Skoda Kodiaq 2018.

PR #24726, opened 2022-06-03
PlotJuggler: add torque control layout

sshane:
Includes accels from `torqueState` and roll-compensated accels from the curvatures which are a bit less noisy than using the localizer, but should be totally accurate and not as delayed.

Screenshot from 2022-06-03 00-03-04


sshane:
Good to merge unless anyone wants certain signals @HaraldSchafer look good?

adeebshihadeh:
Feel free to just merge debug stuff like this. There's no cost to just merging and fixing up as we go.

sshane:
👍 The only thing is this uses the vehicle model when I know Harald says we should ideally use the localizer, however this signal is much less noisy so it's easier to get more accurate measurements from just one or two samples, as apposed to using the localizer and averaging many. Will merge as is now as it's much better than nothing

PR #24737, opened 2022-06-03
message dependencies grapher

lukaspetersson:
simulate_dependencies.py
Solves the problem of ordering processes which rely on each other's outputs. Works on dummy example, but the relationship between the openpilot processes are not as simple as in this graph https://blog.comma.ai/openpilot-in-2021/.

Other attempts were made but without any valuable progress.


adeebshihadeh:
Let's put this in selfdrive/test/process_replay

sshane:
@adeebshihadeh do we still want this?

PR #24748, opened 2022-06-04
HDA2 cars: fix torque limits

sshane:
Our torque limits and steering pressed logic assumes the original Hyundai scaling, but the signal on HDA2 cars is 4x in max value, causing torque to wind down when not touching the wheel.

- [x] Update torque rate limits and steering pressed threshold

https://imgur.com/a/9SfUxMq


sshane:
This PR did this: https://github.com/commaai/openpilot/pull/24901

adeebshihadeh:
Done in 221086857aac770f5d22273119e6175eeb06c729

PR #24756, opened 2022-06-06
ui: fix prime ad storage duration

jyoung8607:
Fix the comma prime signup ad in the UI: the prime cloud storage duration was bumped to a year quite a long time ago.


adeebshihadeh:
Nice catch, though rlogs/full res stuff is only kept for two weeks still. I'll merge this to match the website, and we can figure out how to better communicate the different policies for rlogs and qlogs.

PR #24757, opened 2022-06-06
when fingerprinting fails use last successful fw fingerprint for VIN

gregjhogan:
idea to better handle people taking their car in for service and openpilot doesn't work when they leave to due to updates applied to their vehicle

example: https://github.com/commaai/openpilot/issues/24444

when we process qlogs we should automatically open PRs to add missing fw versions if the fingerprint was not an exact match

needs tests and probably more logging if this is something we actually want to do


adeebshihadeh:
Closing this since I don't think this will be necessary with 100% reliable fuzzy fingerprinting, but noting in #26612.

PR #24760, opened 2022-06-06
Chrysler: carState signals update

realfast:
<!--- ***** Template: Car bug fix *****

These changes are in preparation for upstreaming Ram trucks into Chrysler. 

Changes include:

brakePressed is looking at the exact brake switch signal.

Gas is a more universal signal coming from all Chrysler Engine Control Modules

VEgoRaw is the exact value the EPS looks at for speed. This gives a more accurate signal to turn on and off the control bit. 

cruisestate.enabled is the actual bit

cruisestate.available is the actual bit 

gearShifter is a more universal signal from all Chrysler transmissions.


sshane:
Reasoning behind changes:

<blockquote>
brake pressed is looking at the exact brake switch.

Gas is a more universal signal

VEgoRaw is the exact value the EPS looks at. 

cruisestate.enabled is the actual bit 

cruisestate.available is the actual bit
</blockquote>

realfast (commented):
Wheel speed isn’t as accurate of a signal. The EPS looks at the exact value I shared. Also the wheel speed value isn’t mph it’s in rpm.  If the tire size in a truck is changed, or there is a variance from the factory in tire sizes such as in the Jeep grand cherokees, this value will be off once applying a scaler.

adeebshihadeh:
Does the RAM port require any new signals not used here? If not, I think we'll get the RAM merged with the existing signals then do the signals update after. These updates are good, but I don't want to mix up these two changes.

realfast:
Most of the signals here are already in the ram port.

adeebshihadeh:
Thanks @realfast!

PR #24762, opened 2022-06-07
fullframe DM model

ZwX1616:


ZwX1616:
libyuv is updated to a newer version and needs new mac builds

selfdrive/modeld/models/dmonitoring.cc
```diff
@@ -47,102 +31,73 @@ void dmonitoring_init(DMonitoringModelState* s) {
...
-  }
 
-  //printf("preprocess completed. %d \n", yuv_buf_len);
-  //FILE *dump_yuv_file = fopen("/tmp/rawdump.yuv", "wb");
-  //fwrite(resized_buf, yuv_buf_len, sizeof(uint8_t), dump_yuv_file);
-  //fclose(dump_yuv_file);
+  // snpe UserBufferEncodingUnsigned8Bit doesn't work
+  // fast float conversion instead, also does h crop and scales to 0-1
+  for (int r = 0; r < MODEL_HEIGHT; ++r) {
+    libyuv::ByteToFloat(raw_y_start + r * stride + h_off, net_input_buf + r * MODEL_WIDTH, 0.003921569f, MODEL_WIDTH);
```
geohot, line 88: Oh, its floats! No wonder it's so much power. This is a 5.5MB buffer

adeebshihadeh:
Can you split out the library updates into a separate PR?

ZwX1616:
> Can you split out the library updates into a separate PR? Also looks like it might not be more inconvenient to use a new struct for the DM instead of the existing one with most fields deprecated?

nvm we won't need the new libyuv with no floats

Issue #24763, opened 2022-06-07
Update GM port to use new gear signal with better granularity


JasonJShuler:
GM gear signal ECMPRDNL.PRNDL doesn't report low "L" or manumatic gears.

L mode (aka single pedal mode) changes the behavior of the gas pedal such that when fully released, the vehicle applies braking.
Manumatic mode is present on some large-frame vehicles allowing the driver to specify a max gear.

These two additional gears change the acceleration (and possibly braking) behavior of the vehicle, but with the current signal it is impossible to tell.

The new signal - ECMPRDNL2.PRNDL2 supplies all of the drive modes, including the selected manumatic gear.

adeebshihadeh:
In the future, no need to create issues if you have a PR ready soon.

adeebshihadeh:
Gonna close this since the PR is open.

PR #24764, opened 2022-06-07
GM: parse low and manumatic gears

JMPZ11:
**Description**
Update GM port to use the new ECMPRDNL2 message PRNDL2 and ManualMode signals. Combined, these signals provide much improved granularity for the driver selected gear.

OP PR for issue https://github.com/commaai/openpilot/issues/24763

**Verification**

The OPGM fork has been using this new signal for over a month. The signal is present in all the valid fingerprints.

@sshane offered to assist with regression testing / validation against all routes.


JasonJShuler:
Update: The failed test is from a route that was in L mode - carState.gearShifter is now low instead of drive - which is expected.
Working on updating logs

JasonJShuler:
@sshane I've been digging into the test_replay and I think I need another pair of hemispheres on this.

Running the test locally and checking diff.txt, what I'm seeing doesn't make sense. It claims there should be 6 data points with a value of "park" followed by a ton of "drive". 


```
***** differences for segment fakedata|2022-01-20--18-03-41--0 *****
*** process: controlsd ***
	('change', 'carState.gearShifter', ('park', 'low'))
	('change', 'carState.gearShifter', ('park', 'low'))
	('change', 'carState.gearShifter', ('park', 'low'))
	('change', 'carState.gearShifter', ('park', 'low'))
	('change', 'carState.gearShifter', ('park', 'low'))
	('change', 'carState.gearShifter', ('park', 'low'))
	('change', 'carState.gearShifter', ('drive', 'low'))
	('change', 'carState.gearShifter', ('drive', 'low'))
	('change', 'carState.gearShifter', ('drive', 'low'))
	('change', 'carState.gearShifter', ('drive', 'low'))
...
```

I dumped the rlog to a file (url is https://commadataci.blob.core.windows.net/openpilotci/fakedata/2022-01-20--18-03-41/0/rlog.bz2)

And searched for "gearShifter = park". No results. I then tried "address = 309" - this one isn't as simple because there is also a message id 309 coming from the chassis bus - it has a value of zero, but is completely unrelated.

image


But... the head commit on master passed its tests. I'm not going to mess with updating logs till you have a chance to peek.

selfdrive/car/gm/carstate.py
```diff
@@ -28,7 +28,11 @@ def update(self, pt_cp, loopback_cp):
     ret.vEgoRaw = mean([ret.wheelSpeeds.fl, ret.wheelSpeeds.fr, ret.wheelSpeeds.rl, ret.wheelSpeeds.rr])
     ret.vEgo, ret.aEgo = self.update_speed_kf(ret.vEgoRaw)
     ret.standstill = ret.vEgoRaw < 0.01
-    ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(pt_cp.vl["ECMPRDNL"]["PRNDL"], None))
+    if pt_cp.vl["ECMPRDNL2"]["ManualMode"] == 1:
+      ret.gearShifter = self.parse_gear_shifter("T")
+      ret.manumaticGear = pt_cp.vl["ECMPRDNL2"]["PRNDL2"]
```
adeebshihadeh, line 33: do we really need to add this? it's not used for anything

sshane:
Yeah not really sure where those parks are coming from in test_processes, it's only a few frames at the beginning of the drive. Could make sense if we start logging too early and they're not initialized yet, but they should be, and the default of park (1) doesn't make sense.

selfdrive/car/gm/carstate.py
JasonJShuler: I'm envisioning OP having the ability to adjust the accel/decel behavior as well as the max speed based on the selected manumatic gear. Similar to how we specify a min engage and set speed, while in manumatic mode we might might want to impose appropriate maximums.


The max gear selection changes the acceleration curve.
If it is set to 1 and you floor it, the car will take off fast but then probably top out at 25mph (I would expect the max speed for each gear is provided in the manual)

Whereas if it is 12 (the max on sback's Suburban) there is little difference from standard drive mode.

It sounds like manumatic might also provide engine braking.

sshane: That would be a good fork feature, but let keep this to only required minimum changes for now. Sounds like we want the gear shifter change, as we disallow similar gears in other brands (low, sport, B on Toyota).

JasonJShuler: Sure thing! I will close the cereal PR and update this one

JasonJShuler: Done!

selfdrive/car/gm/interface.py
```diff
@@ -165,7 +166,9 @@ def _update(self, c):
 
       ret.buttonEvents = [be]
 
-    events = self.create_common_events(ret, pcm_enable=self.CP.pcmCruise)
+    events = self.create_common_events(ret, extra_gears=[GearShifter.sport, GearShifter.low,
+                                                         GearShifter.eco, GearShifter.manumatic],
```
sshane, line 170: Before, these all would show up as `GearShifter.drive`?

sshane:
Checked a few random routes:

- Volt (3aef7c4addc53455|2022-08-01--06-12-37): User ended up in drive with the old signal which is now low with the new signal, everything else is the same: https://user-images.githubusercontent.com/25857203/182280259-d8b5aff4-b139-4458-aadd-d057fa378103.png

selfdrive/car/gm/interface.py
twilsonco: Correct

JasonJShuler: I concur with Tim

sshane: Since low and manumatic change the experience of the gas pedal, usually on other cars we block engaging here. Does it change engaged behavior or only disengaged behavior?

PR #24770, opened 2022-06-07
Subaru: Forester Hybrid 2020 dashcam

sshane:
Original work thanks to @martinl!

All Subarus have the `ES_DashStatus` message which is sent by the camera and contains the cruise state signals we need, however there's another higher frequency message sent by the car `CruiseControl` (20 vs 10 Hz) so we currently use that instead. On the Forester Hybrid though, we only have `ES_DashStatus` from the camera.

The plan:
- [ ] Doesn't make sense `CruiseControl` would be missing, there should be a readback from the car somewhere, try to see if it just moved addresses.
- [ ] If not, try to find another common message and signal that all cars can use, and as a fallback, just use the lower-frequency `ES_DashStatus` so no safety params are needed.

Need:
- [ ] Test route
- [ ] Car info


jnewb1:
@sshane applied your suggestions from the crosstrek pr. I'll add a route tomorrow. Can you review this? (cant request a review because it's your PR)

selfdrive/car/tests/routes.py
```diff
@@ -28,6 +28,7 @@
   VOLKSWAGEN.CRAFTER_MK2,  # need a route from an ACC-equipped Crafter
   TOYOTA.RAV4_TSS2_2023,
   TOYOTA.RAV4H_TSS2_2023,
+  SUBARU.FORESTER_HYBRID, # TODO: find a route from martinl?
```
sshane, line 31:
```suggestion
  SUBARU.FORESTER_HYBRID,
```

PR #24774, opened 2022-06-07
GM: interface cleanup

JMPZ11:
**Description**
* Move default longitudinal tune (and other brand defaults) above the model specific config to allow these fields to be customized per model. Previous layout was overwriting longitudinal tunes.
* Remove redundant values (where the value matched global or brand defaults)

I'm guessing the fact that some values typically customized per-model were being set after the model sections was an oversight. It didn't create issues yet as none of the models were overriding them - but some incoming models require non-default longitudinal tuning.

Removal of default values aligns with other ports.

**Verification**
These changes do not alter the result of interface.py as it stands.

### DIFF
--- a/selfdrive/car/gm/interface.py
+++ b/selfdrive/car/gm/interface.py
@@ -81,7 +81,6 @@ class CarInterface(CarInterfaceBase):
       ret.wheelbase = 2.69
       ret.steerRatio = 17.7  # Stock 15.7, LiveParameters
       tire_stiffness_factor = 0.469 # Stock Michelin Energy Saver A/S, LiveParameters
-      ret.steerRatioRear = 0.
       ret.centerToFront = ret.wheelbase * 0.45 # Volt Gen 1, TODO corner weigh
       ret.maxLateralAccel = 1.6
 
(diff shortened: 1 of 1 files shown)


selfdrive/car/gm/interface.py
```diff
@@ -58,22 +58,29 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, disa
...
     ret.steerRateCost = 1.0
     ret.steerActuatorDelay = 0.1  # Default delay, not measured yet
 
+    ret.longitudinalTuning.kpBP = [5., 35.]
+    ret.longitudinalTuning.kpV = [2.4, 1.5]
+    ret.longitudinalTuning.kiBP = [0.]
+    ret.longitudinalTuning.kiV = [0.36]
+
+    ret.steerLimitTimer = 0.4
+    ret.radarTimeStep = 0.0667  # GM radar runs at 15Hz instead of standard 20Hz
```
sshane, line 75:
```suggestion
    ret.steerLimitTimer = 0.4
    ret.minEnableSpeed = 18 * CV.MPH_TO_MS
    ret.radarTimeStep = 0.0667  # GM radar runs at 15Hz instead of standard 20Hz
```

selfdrive/car/gm/interface.py
```diff
@@ -74,9 +74,10 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, disa
     ret.steerLimitTimer = 0.4
     ret.radarTimeStep = 0.0667  # GM radar runs at 15Hz instead of standard 20Hz
 
+    # supports stop and go, but initial engage must be above 18mph (which include conservatism)
```
adeebshihadeh, line 77:
```suggestion
    # supports stop and go, but initial engage must (conservatively) be above 18mph
```

PR #24778, opened 2022-06-08
Mazda:  CX-5 22 FW FP

Jafaral:


sshane:
Are these all from 661621a8442f0688?

Jafaral:
@sshane , no, only one of them. I  can ask the users to provide ids, but that is what I got from them.

adeebshihadeh:
Yes, we'd like to confirm these values with their routes.

Jafaral:
@sshane I dropped the ones I couldn't get dongle  id for.

PR #24782, opened 2022-06-08
Toyota: keep Lane Sway Warning system

cydia2020:
Lets openpilot pass through the lane sway warning system signals sent by the camera.
This retains all features of the SWS including its toggles, so users are able to disable it from the vehicle's MFD if they so wish.
Tested on an 18 Prius.


cydia2020:
I can apply this to the stock LDA as well if that's something that you're is interested in.

selfdrive/car/toyota/carstate.py
```diff
@@ -128,6 +128,19 @@ def update(self, cp, cp_cam):
       ret.leftBlindspot = (cp.vl["BSM"]["L_ADJACENT"] == 1) or (cp.vl["BSM"]["L_APPROACHING"] == 1)
       ret.rightBlindspot = (cp.vl["BSM"]["R_ADJACENT"] == 1) or (cp.vl["BSM"]["R_APPROACHING"] == 1)
 
+    if self.CP.carFingerprint != CAR.PRIUS_V:
```
sshane, line 131: We should just copy the whole message

cydia2020: For the Prius V? This message doesn't get sent by the camera (or exist at all) on the V so I don't think you can copy it.

As for the other cars, openpilot still uses this message for other stuff like LDA and cancel chime masking.

sshane: No I mean create self.lkas_hud and copy the whole dictionary, like the other makes do

sshane:
canTimeout diff is likely from the slower LKAS_HUD signal. It's only True for one frame at the beginning

sshane:
thanks @cydia2020 for the PR! want to open a new one to increase the rate of the LKAS_HUD msg? think 4 or 5 hz is fine

PR #24787, opened 2022-06-09
Honda DBC and carcontroller cleanup

sshane:


sshane:
Moved to https://github.com/commaai/openpilot/pull/24806

PR #24789, opened 2022-06-09
Car interface: set max lateral torque from table

haraschax:


HaraldSchafer:
Screenshot from 2022-06-08 19-16-00
Screenshot from 2022-06-08 19-15-50
Screenshot from 2022-06-08 19-15-41

HaraldSchafer:
MAX_LAT_ACCEL_FIT is just a least squares fit assuming the torque response is linear. MAX_LAT_ACCEL_MEASURED is by actually looking for segments where torque is maxed. In cars with very nonlinear torque response, like hondas they are completely different, and the MEASURED is a better indicator of what it can actually do.

adeebshihadeh:
Not sure this should go in the repo. Shall we put this in a Discussion instead?

HaraldSchafer:
All the values will be read from the table though?

adeebshihadeh:
Ah, I thought this was just for reference.

haram-KONA:
I'm terribly sorry, but I would like to confirm one thing. 
Looking at the torque table, the Friction value of the "Hyundai KONA Hybrid 2020" car seems strange. 
I just tried driving with the value in the table (0.086371), and the oscillation occurs too badly while driving. 
So, when I drive with the friction value of "0.0086371", there is no oscillation and it runs in a very good condition. 
Sorry for the trouble, but please check once again.
I am grateful for all yours.

HaraldSchafer:
@haram-KONA I think the way torque tune works on master may not be optimal, if you want to experiment you can try this branch https://github.com/commaai/openpilot/pull/24801, with the values from this table.

I haven't tried it on Hyundai, so may or may not work well.

ntegan1:
torq_control_fixes works well for me with friction a bit less than half the table value for corolla 2017

echo1001:
There may need to be an additional tune for TSS2 Rav4's with the alternative steering rack. My Rav4 has one and I found my max lat accel is around 3.4.

PR #22863, opened 2021-11-10
Add support for Hyundai Tucson 2021

bluesforte:
This is a working port for the Hyundai Tucson 2021. It has been tested by one of my clients on multiple long drives (3+hours each), and seems to be working well. I didn't do any lateral PID tune due to time constraints, but the default one handles pretty well.

Some notes:

- SteerRatio was determined empirically - 16 "felt" right after a few test drives.
- Car data (wheelbase, weight etc) was determined from the Hyundai website
- The Tucson cruise control appears not to turn on unless above ~20MPH, though does correctly stay on and allows for following cars even in stop-and-go traffic. May be able to be overridden somewhere in openpilot, but minimizing changes and don't want to have any impact on the safety model.
- Wish I had saved more test data while I had access to the car, but wasn't thinking ahead to the pull request :(. If I do get access again, I will grab the route data to add to test_routes.py.

Enjoy!

PR #24791, opened 2022-06-09
Hyundai: Tucson 2021 support

sshane:
Continuation of https://github.com/commaai/openpilot/pull/22863

This is a working port for the Hyundai Tucson 2021. It has been tested by one of my clients on multiple long drives (3+hours each), and seems to be working well. I didn't do any lateral PID tune due to time constraints, but the default one handles pretty well.

Some notes:

- SteerRatio was determined empirically - 16 "felt" right after a few test drives.
- Car data (wheelbase, weight etc) was determined from the Hyundai website
- The Tucson cruise control appears not to turn on unless above ~20MPH, though does correctly stay on and allows for following cars even in stop-and-go traffic. May be able to be overridden somewhere in openpilot, but minimizing changes and don't want to have any impact on the safety model.
- Wish I had saved more test data while I had access to the car, but wasn't thinking ahead to the pull request :(. If I do get access again, I will grab the route data to add to test_routes.py.

Enjoy!


sshane:
Looks like only the top platform (Ultimate) has SCC https://www.hyundaiusa.com/us/en/vehicles/2021-tucson/ultimate

adeebshihadeh (approved):

sshane:
Looks like the min engage speed is 20 mph according to discussion in closed PR

sshane:
In case we need to split these cars (due to torque or different min enable speeds) I made it a separate platform again with the same params. Once we get a user driving with this car I'll add a test route and verify it can enable down to 5 mph and then merge them

PR #24793, opened 2022-06-09
Toyota: Add missing esp fw version for RAV4_TSS2

griff12:


sshane:
Do you have a dongle ID?

griff12:
d11884f83ada3652 is what I am seeing.

Griffin Christenson

PR #24796, opened 2022-06-09
Lexus: NX Hybrid 2020 support

AlexandreSato:
DongleId: 09ae96064ed85a14  | testRoute: 09ae96064ed85a14|2022-01-10--01-57-37


sshane:
Looks like this is a hybrid and would throw can errors since it's looking for the ICE gas pedal signal.

PR #24802, opened 2022-06-09
Remove RHD toggle: determine by driver side instead

ZwX1616:


ZwX1616:
TODO: communicate change

PR #24807, opened 2022-06-10
add liveRestart function for C3

dkiiv:
Idea from @sunnyhaibin#0865 found in Discord

restart openpilot without rebooting (for live tuning, parameter reset, etc.)


adeebshihadeh:
I don't think this script is useful to check into the repo. It doesn't work in the tmux session, and there's no need to start a new tmux session to restart openpilot - just ran `./launch_openpilot.sh`.

PR #24810, opened 2022-06-10
HKG: simplify Kia K5 compatibility

adeebshihadeh:


adeebshihadeh:
@sshane I only checked the K5, so let's double check this, but it seems LKAS/LFA is standard on all. SCC isn't standard on the K5 though.

sshane:
The default is SCC, but sounds good, I'll double check everything

sshane:
This is a 17 Elantra, the owner's manual says "if equipped" by LKAS. Will try to find a better source for trims and availability.

Screenshot from 2022-06-13 23-23-19

adeebshihadeh:
Let's just do this incrementally as we find things out. Want to merge this for just the K5?

PR #1510, opened 2020-05-14
Add 2016 Optima Print

TK211X:
Did I do this right?

PR #24815, opened 2022-06-10
Optima: split into two platforms

### LINKED ISSUE #21078: Kia Optima 2017 missing minSteerSpeed
**Describe the bug**

<When speed below 32mph on Kia Optima 2017 (without MDPS harness), comma device displays master warning (red flash with loud beep) - 'steering temporarily unavailable' and then after 2-3 seconds disables cruise control. This effectively kills stop and go cruise control - this is a new issue in 0.8.4 and onwards. Behaviour in 0.8.3 was that you wouldn't get OP steering below 32mph be cruise control would work as normal (stock ACC) and once speed rose above 32mph, steering from OP would reactivate - I found this to be acceptable.>

<Crossing above and below 32mph in OP >

**Expected behavior**

<For no master warning below 32mph - to work as previous 'steering limit reached, take control' message and for ACC not to be disabled as there is a risk of a collision if the user expects ACC to be on and it disengages because of this>

 - Device: EON Gold
 - Dongle ID: c7f89112bca83879
 - Route: c7f89112bca83879|2021-05-29--13-06-46
 - Timestamp: 14:11
 - Version: 0.8.4 stock
 - Car make/model: Kia Optima 2017

**Additional context**

Thank you for reviewing my bug report. I think this is a step backwards in functionality and has the potential to cause an accident

sshane:
According to https://github.com/commaai/openpilot/pull/1510, 2016 Optima is supported under old platform

Once we get eps fw back this should work (falls back to can fingerprinting as it is without eps)

Checked dongles:  `4f930156368f7830`, `5460c6f74ef5201f`, `fec0d754d02e943e`, `3d96bd05b5513638`, `84269de86f328f98`, `df71aec6e636d7e4`, `ff5d827d501a8851`, `018654717bc93d7d`


sshane:
Once we go to release, I'll update the versions with the brand logging, there's no routes on master with the new logging

sshane:
This is the only dongle id with an Optima with fwdRadar, it seems to be an ecu bug: 4f930156368f7830|2021-12-14--06-26-28. It's also the only 2017 Optima (2018+ have the bug)

sshane:
When radar is missing, these cars fuzzy FP

adeebshihadeh:
Let's start moving towards the generation convention that @jyoung8607 has been using in the VW port. It ends up making a lot more sense when you can group cars by generation, then group those openpilot platforms into the manufacturing platforms. https://github.com/commaai/openpilot/blob/2c9f751616d01a695968150c0edccc57cc2f6d4f/selfdrive/car/volkswagen/values.py#L111

PR #24825, opened 2022-06-11
TSS-P Corolla: update friction lateral tuning parameter

ntegan1:
Screenshot from 2022-06-12 00-00-07

updated friction value from table (~.12) is too high for my car -> see above image.
lowering to around 0.05 eliminates most oscillation

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -31,6 +31,7 @@ class CarInterface(CarInterfaceBase):
     ret.stoppingControl = False  # Toyota starts braking more when it thinks you want to stop
 
     stop_and_go = False
+    torque_params = CarInterfaceBase.get_torque_params(candidate)
 
     if candidate == CAR.PRIUS:
       stop_and_go = True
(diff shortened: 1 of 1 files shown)


HaraldSchafer:
You have the route?

HaraldSchafer:
Was this route with the misc torque control fixes?

ntegan1:
> Was this route with the misc torque control fixes?

yes, this is on master after that was merged

route is
a152dad3c3368320|2022-06-11--19-40-54
difference is more noticeable here because of increased torque values.

route with this updated higher friction (.024->.12) and stock op torque values:
uploaded to this dongle but i'm searching for it in useradmin carparams

sshane:
The route I checked before the torque fixes looked much better than this I feel, did it get worse with that change?

HaraldSchafer:
The PR makes the controller recover from errors quicker, so it could make some type of oscillations worse. 

@ntegan1 I do really want this controller to work well on all cars with the table values, so I'm intrested in fixing this. I did just see that you have the wrong max accel values you though. You should use 'LAT_ACCEL_FACTOR'. Can you try that?

PR #24831, opened 2022-06-12
Laikad: Cache orbit and nav data

gijskoning:
Using Params to save orbits and nav from astro dog.
The cache file is measured to be ~ 3.6MB using sys.getsizeof(json_cache) and will be updated every minute with mostly new nav data. 
Only if no orbits are valid anymore, they will be downloaded and cached. And the orbit prediction file contain at max 24 hours of future ephemeris, but because of the delay from the source in processing these files when downloaded they are valid for around 12-18 hours.

Old ephemeris data is overwritten when new data is added: https://github.com/commaai/laika/pull/91 - Merged

### DIFF
--- a/selfdrive/locationd/laikad.py
+++ b/selfdrive/locationd/laikad.py
@@ -30,7 +30,6 @@ class Laikad:
     self.gnss_kf = GNSSKalman(GENERATED_DIR)
     self.orbit_p: Optional[Process] = None
     self.orbit_q = Queue()
-    self.params = Params()
     self.load_cache()
     self.last_save_t = None
     self.last_fetch_orbits_t = None
--- a/selfdrive/locationd/laikad.py
+++ b/selfdrive/locationd/laikad.py
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 import time
-from multiprocessing import Process, Queue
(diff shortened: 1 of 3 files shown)


selfdrive/locationd/laikad.py
```diff
@@ -20,15 +21,31 @@
...
     self.gnss_kf = GNSSKalman(GENERATED_DIR)
     self.orbit_p: Optional[Process] = None
     self.orbit_q = Queue()
+    self.load_cache()
+    self.last_save_t = None
+    self.last_fetch_orbits_t = None
+
+  def load_cache(self):
+    params = Params()
+    if params.check_key(EPHEMERIS_CACHE) and False: # todo fix get params
```
adeebshihadeh, line 39: `Params::checkKey` only tells you whether this is a valid params key. it doesn't really make sense to use it here

gijskoning: Thanks! Will remove it

gijskoning:
@adeebshihadeh Do you think 4 MB every minute is okay to cache?

adeebshihadeh:
I think that should be fine. You can also try caching on exit.

PR #24832, opened 2022-06-12
Ford: disable radar for now

incognitojam:
The newer Ford vehicles require a different radar parser so I'd like to disable this for now.


incognitojam:
thanks!

PR #24834, opened 2022-06-12
HB: Fix ACC faulted with OP Long enabled

### LINKED ISSUE #24833: Honda ACC faulted with OP Long enabled
### Describe the bug

#24806 is causing ACC faulted. Error message: https://gyazo.com/21938e46df920bd64ee258641163c560

20' Civic Hatchback Sport

f1055ddb93ff2d3b|2022-06-12--09-09-40--0

### openpilot version

master

### Additional info

_No response_

redacid95:
**Description**
Fixes issue: #24833, introduced in #24806.

Based on logic from pervious version from pull mentioned above, I used "enabled" to determine ACC_ON state.

**Verification**
No more error message from issue mentioned above. Tested with OP Long enabled in route below. Verified HUD displayed accurately depending on engaged state. Tested with OP Long disabled as well with no issues.

**Route**
f1055ddb93ff2d3b|2022-06-12--09-54-09--1

### DIFF
--- a/selfdrive/car/honda/hondacan.py
+++ b/selfdrive/car/honda/hondacan.py
@@ -115,7 +115,7 @@ def create_ui_commands(packer, CP, enabled, pcm_speed, hud, is_metric, idx, stoc
     }
 
     if CP.carFingerprint in HONDA_BOSCH:
-      acc_hud_values['ACC_ON'] = hud.car != 0
+      acc_hud_values['ACC_ON'] = enabled != 0
       acc_hud_values['FCM_OFF'] = 1
       acc_hud_values['FCM_OFF_2'] = 1
     else:


adeebshihadeh:
@sshane let's get a honda bosch long route in process replay before merging this. @redacid95 would you like to provide a route for the tests to prevent regressions like this in the future?

sshane:
We should also add this route to safety replay to make sure that doesn't regress again as well

redacid95:
@adeebshihadeh What do I need to do to provide a route for testing? I got no problems with being the HB Guinea pig. I got multiple route of failure and success before and after this change.

adeebshihadeh:
> @adeebshihadeh What do I need to do to provide a route for testing? I got no problems with being the HB Guinea pig. I got multiple routes of failure and success before and after this change.

Any segment will do, but ideally you're engaged for most of it and there's a few engage/disengage events in there.

redacid95:
> > @adeebshihadeh What do I need to do to provide a route for testing? I got no problems with being the HB Guinea pig. I got multiple routes of failure and success before and after this change.
> 
> Any segment will do, but ideally you're engaged for most of it and there's a few engage/disengage events in there.

Alright, I will get a better segment with me engaging and disengaging a few times.... Will provide a little later today.

redacid95:
This should do it: f1055ddb93ff2d3b|2022-06-12--18-47-57--5

I'm having an issue with preserving routes for some reason, don't know if you can do that for me. I have tried from multiple computers. Also does it matter if I'm using my fork?

sshane:
Make sure your device is on WiFi, I requested upload for some logs

redacid95:
> @redacid95 Make sure your device is on WiFi, I requested upload for some logs

At work and my Prime is not working cause I have an old SIM. Will be about 7 hours till I am home. I will let you know when it's connected.

redacid95:
@sshane should be good to do your thing now :)

selfdrive/car/honda/hondacan.py
```diff
@@ -115,7 +115,7 @@ def create_ui_commands(packer, CP, enabled, pcm_speed, hud, is_metric, idx, stoc
     }
 
     if CP.carFingerprint in HONDA_BOSCH:
-      acc_hud_values['ACC_ON'] = hud.car != 0
+      acc_hud_values['ACC_ON'] = enabled != 0
```
sshane, line 118:
```suggestion
      acc_hud_values['ACC_ON'] = enabled != 0 # test
```

sshane:
Moved here as I wasn't able to push to your branch: https://github.com/commaai/openpilot/pull/24850

PR #24836, opened 2022-06-12
Move camerad to system/

adeebshihadeh:


pd0wm:
There is a submaster for driverState, but it's not actually used. That needs to be removed before it can go into system.

PR #24838, opened 2022-06-12
radarState should never be ignored

### LINKED ISSUE #24828: Show bus dropout for radar bus
This user's 2022 Highlander Hybrid has no radar points, thus radard never sends radar packets, and radard doesn't automatically switch to VOACC, causing openpilot to ignore leads: 2aaed1ef9648b871

Unclear if this should be handled in radard or controlsd, but think this is a good idea still.

sshane:
Always sends a radarState, and makes controlsd always consider it (normal sleep, empty ret if no radar from base radar interface)

We can get away without the error change (reduced diff), we'd just send a commIssue event instead of canError


sshane:
@adeebshihadeh parser_pyx handles no CAN messages from boardd, it's what CS.canValid relies on too. Feel free to edit or re-open though

PR #24839, opened 2022-06-12
Ford: implement curvature steering control

incognitojam:
This adds support for the Ford "curvature based" steering control.

In the stock system, the IPMA (MobileEye chip) produces information about the car in the lane (lane curvature, angle of the car to the desired path, offset from the desired path). The PSCM processes this information (presumably along with other data, such as the vehicle speed or roll rate) to calculate a steering angle.

To use this control method for openpilot we can calculate curvature, path deviation and path angle from the model and lateral planner outputs. So far I have found that `model_v2.position.y` is most promising for `pathOffset`, as is `model_v2.orientation.z[0]` for `pathAngle`. The `curvature` signal is trickier since it is calculated from the path the model wants to take, but this includes any corrective action which is already accounted for in the `pathOffset` and `pathAngle`. In the stock system, curvature is more "constant" from the shape of the road or lane, and is not affected by the car's orientation or offset.

The LCA command has signals for "ramp rate" and "precision" which I have set to "fast" and "comfortable" - this might need some more experimentation. Currently there is no rate limitting implemented other than clipping the values to the signal ranges from the DBC.

Demo route: `86d00e12925f4df7|2022-06-10--17-44-23` (this was using my `ford-devel-c3` branch, not this PR)
PlotJuggler layout: TODO (useful for comparing curvature controls state to stock signals)

See also:
- https://github.com/commaai/cereal/pull/303
- https://github.com/commaai/panda/pull/966

### DIFF
--- a/selfdrive/car/ford/carcontroller.py
+++ b/selfdrive/car/ford/carcontroller.py
@@ -1,3 +1,4 @@
+import math
 from cereal import car
 from common.numpy_fast import clip, interp
 from selfdrive.car.ford import fordcan
@@ -5,16 +6,6 @@ from selfdrive.car.ford.values import CarControllerParams
 from opendbc.can.packer import CANPacker
 
 VisualAlert = car.CarControl.HUDControl.VisualAlert
-
-
-def apply_ford_steer_angle_limits(apply_steer, apply_steer_last, vEgo):
-  # rate limit
-  steer_up = apply_steer * apply_steer_last > 0. and abs(apply_steer) > abs(apply_steer_last)
-  rate_limit = CarControllerParams.STEER_RATE_LIMIT_UP if steer_up else CarControllerParams.STEER_RATE_LIMIT_DOWN
-  max_angle_diff = interp(vEgo, rate_limit.speed_points, rate_limit.max_angle_diff_points)
-  apply_steer = clip(apply_steer, (apply_steer_last - max_angle_diff), (apply_steer_last + max_angle_diff))
-
-  return apply_steer
 
 
 class CarController():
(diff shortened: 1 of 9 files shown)


HaraldSchafer:
There's a few issues with the way control is done in the PR now. It breaks some of the abstractions of the openpilot lateral stack.
The latcontrol code only has 1 function, that is to achieve desired_curvature, and to do it quickly.

- You can't use the model in latcontrol, since this bypasses the planner. Latcontrol can not use the outputs of the model directly, they have to go through the planner.
- Curvature is indexed at time 6 and and 15, this bypasses the normal lag compensation the other controllers rely on.
- Desired_curvature needs to be made into useable signals somehow. Preferably by looking for simpler interfaces or somehow finding ways to bypass the complexity. But in the worst cases you should be able to make a function that takes in vEgo and desired curvature and outputs what needs to be in the CAN message.

Not following the normal abstractions will make the behavior inconsistent with other cars and it will break in ways that will be unique to this port and difficult to debug.

incognitojam:
I understand how this makes things inconsistent and I'd like to fix that. We can already control Ford vehicles with just desired curvature (just use the angle controller and convert it back to a curvature and send the command), but I think we can get much better control if there is a way of calculating these other signals.

The DBC only allows for a max curvature of `0.02094 1/m` which I think is only a 50m radius curve. It would be ideal to calculate some path angle from the data too and we could get the radius much lower.

I will take a step back, remove the model from the latcontrol code and look to pull the path angle from the planner. I think `heading_pts` might be a good place to start?

HaraldSchafer:
I think you're better off just making a path angle from the desired curvature than using something else.

incognitojam:
I've adapted this to calculate a path angle from the desired curvature. If it works, we can probably call it "path angle control" instead :stuck_out_tongue:

PR #24848, opened 2022-06-13
Couple more cars to torque tune

haraschax:


haram-KONA:
why not include kona hybrid 2020?

PR #24849, opened 2022-06-13
Move couple toyotas to torque table values

haraschax:


ntegan1:
a152dad3c3368320|2022-06-13--20-29-50 
this table friction value works well for tssp corolla + new e2e long

PR #24850, opened 2022-06-13
regen & process replay: support no disengage on accelerator

sshane:
- [x] Set params from carParams or initData for `DisengageOnAccelerator`
- [ ] Bonus: support disabling radar for HKG and Bosch long (and others)
  - [ ] disable_ecu hangs in test_processes on draining sock


sshane:
This ended up needing more thought than adding a new route, there's issues with controls mismatches, camera malfunction (no wideRoadCameraState, not sure how that worked before), and disable_ecu hanging on the drain_sock function

PR #24851, opened 2022-06-14
Honda Bosch Longitudinal: fix ACC faulted with OP Long enabled

sshane:


sshane:
@redacid95 merged without the route as you were using no disengage on gas and that will require some refactoring to support in process replay. For now, can you get a test route with disengage on gas doing the same thing? Thanks!

PR #24854, opened 2022-06-14
changing source of cereal

philipk19238:


sshane:
What is the purpose of this PR? You generally open a PR against https://github.com/commaai/cereal for cereal changes, not update the openpilot submodule

PR #24856, opened 2022-06-14
Toyota: SnG Support for Stock Long

cydia2020:
As described.

When stock ACC wants to resume (lead vehicle starts to pull away), it sets `RELEASE_STANDSTILL` to 1. This PR allows openpilot to monitor that signal on vehicles with stock ACC and apply a slight accelerator command on the interceptor to allow the vehicle to come out of standstill.

Should work on the 2022 RAV4 and all TSS-P vehicles excluding those that don't support Full Speed Range Dynamic RADAR Cruise Control.

Route:
f9ff7ebb93b72025|2022-06-26--03-01-10 segment 6
f9ff7ebb93b72025|2022-06-26--10-37-54 segment 13


cydia2020:
Not exactly sure why the tests are failing, these cars shouldn't not have `0x343`.

cydia2020:
Works on a 2022 RAV4 as reported by a discord user. I'll ask them to post a route here.

sshane:
Want to rebase using `CC.cruiseControl.resume` and retest? https://github.com/commaai/openpilot/pull/24873

cydia2020:
> Want to rebase using `CC.cruiseControl.resume` and retest? #24873

I actually tried that a few days back, it doesn't seem to work, I will give it another go and rebase if everything works.

selfdrive/car/toyota/carstate.py
```diff
@@ -209,9 +214,14 @@ def get_can_parser(CP):
       ]
       checks.append(("BSM", 1))
 
+    if not CP.openpilotLongitudinalControl and CP.carFingerprint not in (CAR.LEXUS_IS, CAR.LEXUS_RC):
+      signals.append(("RELEASE_STANDSTILL", "ACC_CONTROL"))
```
sshane, line 218: ACC_CONTROL comes from bus 2 for non-radar ACC cars

sshane:
Let me know if you have any logs with it not working

cydia2020:
> Let me know if you have any logs with it not working

Tried it again, still didn't work, I think it's because standstill is ignored on Toyotas if pedal is detected (?)

cydia2020:
> Ah, your other PR to put the car in standstill should fix that. I would try them together to make sure

I think it's because PCM standstill is ignored in car state.

We can change
```python
    if self.CP.carFingerprint in NO_STOP_TIMER_CAR or self.CP.enableGasInterceptor:
      # ignore standstill in hybrid vehicles, since pcm allows to restart without
      # receiving any special command. Also if interceptor is detected
      ret.cruiseState.standstill = False
    else:
      ret.cruiseState.standstill = self.pcm_acc_status == 7
```
to
```python
    if (self.CP.carFingerprint in NO_STOP_TIMER_CAR or self.CP.enableGasInterceptor) and self.CP.openpilotLongitudinalControl:
      # ignore standstill in hybrid vehicles, since pcm allows to restart without
      # receiving any special command. Also if interceptor is detected
      ret.cruiseState.standstill = False
    else:
      ret.cruiseState.standstill = self.pcm_acc_status == 7
```

adeebshihadeh:
This is a lot of code for a minor improvement on a stock feature.

PR #24858, opened 2022-06-14
Pipenv -> poetry

haraschax:
- removes pipenv from openpilot
- add poetry, dependencies defined in `pyproject.toml`
  - `poetry install --no-root` so that "openpilot" isn't installed as a package itself...
- add xx packages to an optional [dependency group](https://python-poetry.org/docs/master/managing-dependencies/)
  - if `xx` dir is present, installs packages to system (`POETRY_VIRTUALENVS_CREATE=false`) and `--with xx`


update_requirements.sh
```diff
@@ -3,7 +3,7 @@ set -e
 
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
 cd $DIR
-
+ls
```
adeebshihadeh, line 6: remove this

Dockerfile.openpilot_base
```diff
@@ -12,19 +12,19 @@ ENV LANG en_US.UTF-8
...
 ENV LC_ALL en_US.UTF-8
 
-ENV PIPENV_SYSTEM=1
+ENV POETRY_SYSTEM=1
 ENV PYENV_VERSION=3.8.10
 ENV PYENV_ROOT="/root/.pyenv"
 ENV PATH="$PYENV_ROOT/bin:$PYENV_ROOT/shims:$PATH"
 
-COPY Pipfile Pipfile.lock .python-version update_requirements.sh /tmp/
+COPY pyproject.toml poetry.lock .python-version update_requirements.sh /tmp/
```
adeebshihadeh, line 20: is the cache still being remove properly here?

update_requirements.sh
```diff
@@ -26,22 +26,22 @@ eval "$(pyenv init --path)"
 
 echo "update pip"
 pip install pip==21.3.1
-pip install pipenv==2021.11.23
+pip install poetry
 
 if [ -d "./xx" ]; then
-  export PIPENV_SYSTEM=1
-  export PIPENV_PIPFILE=./xx/Pipfile
```
adeebshihadeh, line 33: think this will need to be added back before merging?

JMPZ11 (changes requested):
Please also update tools/README.md - I'm still fairly green re: python, and if this changed without the readme, I would be lost... Thanks!

pd0wm:
Poetry 1.2 (with package groups) was just released: https://python-poetry.org/blog/announcing-poetry-1.2.0/

HaraldSchafer:
Cool! might try to revive this.

.pylintrc
```diff
@@ -54,7 +54,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514
+disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514,E1124
```
incognitojam, line 57: Seems that pylint was updated, need to add this otherwise it complains about `video_link` and other keyword args in docs definitions

Dockerfile.openpilot_base
```diff
@@ -12,19 +12,19 @@ ENV LANG en_US.UTF-8
 ENV LANGUAGE en_US:en
 ENV LC_ALL en_US.UTF-8
 
-ENV PIPENV_SYSTEM=1
+ENV POETRY_VIRTUALENVS_CREATE=false
```
incognitojam, line 15: https://python-poetry.org/docs/configuration/#using-environment-variables

Dockerfile.openpilot_base
incognitojam: I think the poetry cache is in `$HOME/.cache` so should be deleted

update_requirements.sh
```diff
@@ -40,25 +40,29 @@ fi
...
-pip install pip==21.3.1
-pip install pipenv==2021.11.23
+pip install pip==22.2.2
 
 if [ -d "./xx" ]; then
   echo "WARNING: using xx Pipfile ******"
-  export PIPENV_SYSTEM=1
   export PIPENV_PIPFILE=./xx/Pipfile
-fi
+  pip install pipenv==2021.11.23
```
incognitojam, line 48: I think we still need to install pipenv for xx?

update_requirements.sh
```diff
@@ -40,25 +40,29 @@ fi
...
 
-if [ -z "$PIPENV_SYSTEM" ]; then
-  echo "PYTHONPATH=${PWD}" > .env
-  RUN="pipenv run"
-else
   RUN=""
+else
+  echo "PYTHONPATH=${PWD}" > .env
+  pip install poetry==1.2.1
+  poetry config virtualenvs.prefer-active-python true --local
```
incognitojam, line 58: since we use pyenv to manage the python installation we should set this
https://python-poetry.org/docs/managing-environments/

adeebshihadeh:
@incognitojam can you finish this up? goal is to get rid of `xx/blob/master/pipenv-merge.sh` in a clean way. if we can't do that, then let's just close this

incognitojam:
@adeebshihadeh we could export the dependencies from the pyproject.toml to generate the xx Pipfile.. or I think adding openpilot as a dependency of xx would achieve the same thing?

docs/docker/Dockerfile
```diff
@@ -4,14 +4,16 @@ ENV PYTHONUNBUFFERED 1
...
 
-COPY Pipfile Pipfile.lock $OPENPILOT_PATH
-RUN pip install --no-cache-dir pipenv==2021.5.29 pip==21.3.1 && \
-    pipenv install --system --deploy --dev --clear && \
-    pip uninstall -y pipenv
+COPY pyproject.toml poetry.lock $OPENPILOT_PATH
+RUN pip install --no-cache-dir poetry==1.2.1 pip==22.2.2 && \
+    poetry install --no-ansi --no-interaction && \
+    pip uninstall -y poetry && \
+    rm -rf /root/.cache
```
adeebshihadeh: hmm, is this even necessary? can we just remove instead? this should already be done in openpilot-base

tools/README.md
```diff
@@ -10,10 +10,9 @@ openpilot is developed and tested on **Ubuntu 20.04**, which is the primary deve
 First, clone openpilot:
 ``` bash
 cd ~
-git clone https://github.com/commaai/openpilot.git
+git clone --recursive https://github.com/commaai/openpilot.git
 
-cd openpilot 
-git submodule update --init
```
adeebshihadeh, line 16: can you keep this and add a comment like "openpilot uses git submodules, update them with:"? we should be better about including all the common workflow stuff here

update_requirements.sh
incognitojam: it also creates a `poetry.toml` file - added to `.gitignore` since it can include custom user configuration

incognitojam:
I've added semver version constraints to all of the `*` packages, using the version from the old `Pipfile.lock`, so they _shouldn't_ have breaking changes.. Also pinned the Python version to `~3.8`.

incognitojam:
if CI passes this should be ready to review!

update_requirements.sh
```diff
@@ -40,25 +40,28 @@ fi
...
+poetry config virtualenvs.prefer-active-python true --local
+
+POETRY_INSTALL_ARGS=""
 if [ -d "./xx" ]; then
-  echo "WARNING: using xx Pipfile ******"
-  export PIPENV_SYSTEM=1
-  export PIPENV_PIPFILE=./xx/Pipfile
+  echo "WARNING: using xx dependency group, installing globally"
+  export POETRY_VIRTUALENVS_CREATE=false
+  POETRY_INSTALL_ARGS="--with xx --sync "
```
adeebshihadeh, line 52:
```suggestion
  POETRY_INSTALL_ARGS="--with xx --sync"
```

update_requirements.sh
```diff
@@ -40,25 +40,28 @@ fi
...
   echo "PYTHONPATH=${PWD}" > .env
-  RUN="pipenv run"
+  RUN="poetry run"
 else
   RUN=""
 fi
 
-echo "pip packages install..."
-pipenv sync --dev
-pipenv --clear
```
adeebshihadeh, line 61: does poetry not have an equivalent of this?

incognitojam: maybe we can do `pip cache purge` and `poetry cache clear --all PyPI`

adeebshihadeh: this makes a big difference in CI, so I'd like that resolved before merging this

incognitojam: I just tested it locally but it always prompted me to type `Y/n` even when I passed the `--no-interaction` option... supposedly fixed in Poetry 1.2.1 https://github.com/python-poetry/poetry/releases/tag/1.2.1

incognitojam: we already do `rm -rf /root/.cache` in the Dockerfile which should clear this anyway?

adeebshihadeh: what about installing with `--no-cache`?

PR #24862, opened 2022-06-15
Try slowdown for turns

haraschax:


HaraldSchafer:
Meh 2.5s is not far enough ahead to be magical

PR #24863, opened 2022-06-15
regen: send wideRoadCameraState

sshane:
Fixes no routes regenerating

This isn't perfect, modeld complains of frames out of sync (it takes 10-20 ms to start the first camera process)

Changed to still send in one process, fixes out of sync issue (mostly)

### DIFF
--- a/selfdrive/test/process_replay/regen.py
+++ b/selfdrive/test/process_replay/regen.py
@@ -156,7 +156,7 @@ def replay_cameras(lr, frs, disable_tqdm=False):
   # init vipc server and cameras
   p = []
   vs = VisionIpcServer("camerad")
-  for s, dt, size, stream, use_extra_client in cameras:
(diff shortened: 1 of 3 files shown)


selfdrive/test/process_replay/regen.py
```diff
@@ -127,7 +127,10 @@ def replay_cameras(lr, frs, disable_tqdm=False):
   ]
 
   def replay_camera(s, stream, dt, vipc_server, frames, size, use_extra_client):
-    pm = messaging.PubMaster([s, ])
+    services = [(s, stream)]
+    if use_extra_client:
+      services.append(("wideRoadCameraState", VisionStreamType.VISION_STREAM_WIDE_ROAD))
+    pm = messaging.PubMaster([s for s, _ in services])
```
sshane, line 133: Not the cleanest, I'd like to fix why it takes so long to start the first cam proc

PR #24871, opened 2022-06-15
navd: handle maxspeed being none

pd0wm:
Can happen on the autobahn according to the docs:
> speed: The maximum speed limit between the coordinates of a segment. The value can be either the legal speed limit or an advisory speed limit where posted, or none if the speed limit is unlimited (for instance, a German Autobahn).

TODO: actually verify this


pd0wm:
The docs are slightly vague, but it's actually encoded as:
```
{'none': True}
```

PR #24873, opened 2022-06-15
Stock longitudinal: spam resume button when lead starts moving

sshane:
- Runs the planner on leads (from vision or radar) when using stock longitudinal or radar off can (so we know when to resume spam)
- Spam resume button only when planner thinks we want to move (old longControlState.starting) logic


selfdrive/controls/controlsd.py
```diff
@@ -643,6 +643,8 @@ def publish_logs(self, CS, start_time, CC, lac_log):
     CC.cruiseControl.cancel = CS.cruiseState.enabled and (not self.enabled or not self.CP.pcmCruise)
     if self.joystick_mode and self.sm.rcv_frame['testJoystick'] > 0 and self.sm['testJoystick'].buttons[0]:
       CC.cruiseControl.cancel = True
+    if len(speeds := self.sm['longitudinalPlan'].speeds) > 1:
```
sshane, line 646: First walrus!

jyoung8607:
I like this, continuous spam on VW has some subtle undesired side effects. Looking forward to trying it.

sunnyhaibin:
This should be done for HKG too. The 2021 Hyundai Sonata N Line has a chime every time the alert to resume appears.

selfdrive/car/honda/carcontroller.py
```diff
@@ -195,7 +195,7 @@ def update(self, CC, CS):
       # If using stock ACC, spam cancel command to kill gas when OP disengages.
       if pcm_cancel_cmd:
         can_sends.append(hondacan.spam_buttons_command(self.packer, CruiseButtons.CANCEL, idx, self.CP.carFingerprint))
-      elif CS.out.cruiseState.standstill:
+      elif CC.cruiseControl.override:
```
sshane, line 198: TODO: add an ordinal or rename override

sshane:
Please provide routes if you try this with one of the affected cars!

selfdrive/car/volkswagen/carcontroller.py
```diff
@@ -93,9 +93,8 @@ def update(self, c, CS, frame, ext_bus, actuators, visual_alert, left_lane_visib
           # Cancel ACC if it's engaged with OP disengaged.
           self.graButtonStatesToSend = BUTTON_STATES.copy()
           self.graButtonStatesToSend["cancel"] = True
-        elif c.enabled and CS.esp_hold_confirmation:
-          # Blip the Resume button if we're engaged at standstill.
-          # FIXME: This is a naive implementation, improve with visiond or radar input.
+        elif c.cruiseControl.override:
```
jyoung8607, line 96:
```suggestion
        elif c.cruiseControl.override and CS.esp_hold_confirmation:
```

I'd like to retain the extra stock-ACC-standstill verification. See #22176 for history.

jyoung8607:
> Please provide routes if you try this with one of the affected cars!

I'm traveling this week but should be able to test on Sunday.

selfdrive/car/volkswagen/carcontroller.py
jyoung8607: On further review, I was not aware of `cruiseState.standstill` back when I started reading the ABS/ESP hold confirmation signal. Probably better to populate that so the check is done in higher level controlsd.

selfdrive/controls/controlsd.py
gijskoning: Ohh, I like it

cydia2020:
we can add this to toyotas with pedal as well, send a very small amount of pedal (in my experience 0.14 to 0.16 is usually enough) when the lead starts moving.

sshane:
Unfortunately the planner is a bit more restricted when it's inactive, so the normal vEgoStarting sometimes can cause us to request resuming a bit later than in the planner was active and we weren't resetting its state constantly to vEgo. Will try to come up with something

jyoung8607:
LGTM on my Golf R: `3cfdec54aa035f3f|2022-06-22--21-17-20` (still uploading)

Isn't spamming at standstill, but still leaves promptly when it's time to do so. I like it!

sshane:
@HaraldSchafer is this a comprehensive planner test? As for the model, are you confident it will be accurate in at least predicting the lead is moving away when they do? If either fails, we won't resume spam.

selfdrive/controls/controlsd.py
```diff
@@ -646,6 +646,8 @@ def publish_logs(self, CS, start_time, CC, lac_log):
     CC.cruiseControl.cancel = CS.cruiseState.enabled and (not self.enabled or not self.CP.pcmCruise)
     if self.joystick_mode and self.sm.rcv_frame['testJoystick'] > 0 and self.sm['testJoystick'].buttons[0]:
       CC.cruiseControl.cancel = True
+    if len(speeds := self.sm['longitudinalPlan'].speeds):
```
adeebshihadeh, line 649: we decided not to use the walrus for now

HaraldSchafer:
The model should be pretty reliable at that.

sshane:
Quickly put the model's outputs through the long planner, for the most part it looks good! https://imgur.com/a/ghOGfMu

Despite the model outputting higher than actual lead distances and velocities, the planner seems to handle it well and not want to resume until it seems reasonable, from those few examples above.

PR #24875, opened 2022-06-16
GM: Bolt EUV port

JMPZ11:
Basic port for vehicles with ACC, but without an ASCM (or accessible radar points)
Bolt EUV and Silverado have been added; Suburban, Tahoe 2022 Bolt LT2, GMC Sierra (after GM VIN query) are now simple ports.

EDIT: Will break down into parts, but wanted to get the final appearance down and get feedback on some choices.

Depends on two unmerged PRs:
- [x] Panda: https://github.com/commaai/panda/pull/962
- [x] OP: https://github.com/commaai/openpilot/pull/24764

**For testing purposes, these PRs have been pre-merged / referenced - This branch builds and passes pre-commit**

Includes Silverado and Bolt EUV - tunes both require review

**Checklist**


adeebshihadeh:
Let's break this up into individual car ports. I'd recommend just removing one from this PR and making a new PR once we get this new platform merged with a single car.

JasonJShuler:
> Let's break this up into individual car ports. I'd recommend just removing one from this PR and making a new PR once we get this new platform merged with a single car.

Sounds like a plan - before I do too much splitting (I'm happy to break it down to more easily testable chunks) I wanted to make sure the approach was acceptable. I'll throw PR comments with things about which I have questions - if that's ok

selfdrive/car/docs_definitions.py
```diff
@@ -146,6 +146,14 @@ class Harness(Enum):
   hyundai_o = "Hyundai O"
   custom = "Developer"
   obd_ii = "OBD-II"
+  gm_cam = "GM Camera"
```
JasonJShuler, line 149: Wasn't sure what to use for harness - developer isn't accurate and would have people ordering wires they don't need - this is why I didn't want to use custom

selfdrive/car/gm/carcontroller.py
```diff
@@ -89,18 +92,18 @@ def update(self, CC, CS):
     time_and_headlights_step = 10
     tt = self.frame * DT_CTRL
 
-    if self.frame % time_and_headlights_step == 0:
+    if CS.CP.openpilotLongitudinalControl and (not CS.CP.radarOffCan) and (self.frame % time_and_headlights_step) == 0:
```
JasonJShuler, line 95: Note: If radar _was_ visible, but OP wasn't doing long, we wouldn't want to be sending. Only when both are true would we want to be sending stuff on the obstacle bus (the ASCM case essentially)

selfdrive/car/gm/carcontroller.py
```diff
@@ -89,18 +92,18 @@ def update(self, CC, CS):
...
 
     speed_and_accelerometer_step = 2
-    if self.frame % speed_and_accelerometer_step == 0:
+    if CS.CP.openpilotLongitudinalControl and (not CS.CP.radarOffCan) and self.frame % speed_and_accelerometer_step == 0:
       idx = (self.frame // speed_and_accelerometer_step) % 4
       can_sends.append(gmcan.create_adas_steering_status(CanBus.OBSTACLE, idx))
       can_sends.append(gmcan.create_adas_accelerometer_speed_status(CanBus.OBSTACLE, CS.out.vEgo, idx))
 
-    if self.frame % self.params.ADAS_KEEPALIVE_STEP == 0:
+    if CS.CP.openpilotLongitudinalControl and CS.CP.networkLocation == NetworkLocation.gateway and (self.frame % self.params.ADAS_KEEPALIVE_STEP) == 0:
```
JasonJShuler, line 106: OP only needs to send keepalives when the factory ECU cannot. Presently only applies in the ASCM case because the entire ASCM is taken off the PT bus.

selfdrive/car/gm/carcontroller.py
```diff
@@ -110,7 +113,8 @@ def update(self, CC, CS):
     lka_active = CS.lkas_status == 1
     lka_critical = lka_active and abs(actuators.steer) > 0.9
     lka_icon_status = (lka_active, lka_critical)
-    if self.frame % self.params.CAMERA_KEEPALIVE_STEP == 0 or lka_icon_status != self.lka_icon_status_last:
+    # SW_GMLAN not yet on cam harness
+    if CS.CP.networkLocation != NetworkLocation.fwdCamera and self.frame % self.params.CAMERA_KEEPALIVE_STEP == 0 or lka_icon_status != self.lka_icon_status_last:
```
JasonJShuler, line 117: Since we did not include this in the Panda list of allowed tx when using the cam harness, OP needs to no try as well

selfdrive/car/gm/carstate.py
```diff
@@ -35,7 +36,7 @@ def update(self, pt_cp, loopback_cp):
     ret.brakePressed = pt_cp.vl["EBCMBrakePedalPosition"]["BrakePedalPosition"] >= 10
 
     # Regen braking is braking
-    if self.car_fingerprint == CAR.VOLT:
+    if self.CP.transmissionType == TransmissionType.direct:
```
JasonJShuler, line 39: Two things.
1. The Bolt EUV is only the first new electric car - rather than trying to hardcode I thought it prudent to use a property of some sort
2. I originally used a set defined in values.py called EV_CAR and included the cars; then I would check if the fingerprint was in the collection. This was before I had even a basic understanding of cereal / capnp.
After noticing that CarParams already contains an attribute to differentiate between EV and Gas cars, I made an assumption - based on the presence of these fields in the CarParams - that we should use them.

If this assumption was wrong, I can redo it using the static values

selfdrive/car/gm/interface.py
```diff
@@ -135,6 +146,46 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, disa
...
+      ret.minEnableSpeed = -1.
+      ret.mass = 2400. + STD_CARGO_KG
+      ret.wheelbase = 3.745
+      ret.steerRatio = 16.3
+      ret.centerToFront = ret.wheelbase * .49
+      ret.networkLocation = NetworkLocation.fwdCamera # Uses Cam Harness
+      ret.radarOffCan = True # No Radar
+      ret.openpilotLongitudinalControl = False # Stock ACC
+      ret.pcmCruise = True # CC is on
+      # Tune
```
JasonJShuler, line 159: Note: This tune needs review. Skip is now using a torque tune with some "fancy" customization for daily driving. Working to simplify still

selfdrive/car/gm/interface.py
```diff
@@ -135,6 +146,46 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, disa
...
+      ret.minEnableSpeed = -1
+      ret.mass = 1669. + STD_CARGO_KG
+      ret.wheelbase = 2.675
+      ret.steerRatio = 16.8 # Not measured
+      ret.centerToFront = ret.wheelbase * 0.4
+      ret.networkLocation = NetworkLocation.fwdCamera # Uses Cam Harness
+      ret.radarOffCan = True # No Radar
+      ret.openpilotLongitudinalControl = False # Stock ACC
+      ret.pcmCruise = True # CC is on
+      # Tune
```
JasonJShuler, line 176: FYI - Trilobyte is our EUV tester, and this may be an old tune. I discovered an error in the params, and I would like to see if the default tune might be acceptable, or if it needs adjustment (centerToFront was about 2x what it should have been)

selfdrive/car/gm/values.py
```diff
@@ -9,8 +9,13 @@
...
-  STEER_MAX = 300  # Safety limit, not LKA max. Trucks use 600.
-  STEER_STEP = 2  # control frames per command
+  # TODO: Different cars have different min steer speed, usually around 3-5 mph
+  # TODO: Delta up and down limits may require adjustment based on excessive rate limiting
+  # TODO: Min steer speed seems to be about 3 MPH; may differ per car
+  # TODO: Unsure of driver values
+  # TODO: MAX_GAS, ZERO_GAS and MAX_BRAKE, MAX_ACC_REGEN are based on a specific volt year...
+  STEER_MAX = 300  # GM LKAS max (input) torque is 3 Nm
+  STEER_STEP = 2  # control frames per command (50 Hz or every 20ms)
   STEER_DELTA_UP = 7
```
JasonJShuler, line 19: Note: skip increased the delta up considerably

selfdrive/car/gm/values.py
```diff
@@ -53,6 +58,8 @@ class CAR:
   ACADIA = "GMC ACADIA DENALI 2018"
   BUICK_REGAL = "BUICK REGAL ESSENCE 2018"
   ESCALADE_ESV = "CADILLAC ESCALADE ESV 2016"
+  SILVERADO = "CHEVROLET SILVERADO 1500 2020"
```
JasonJShuler, line 61: Note: Apparently the "1500" is part of the name: https://media.chevrolet.com/media/us/en/chevrolet/vehicles/silverado/2020.html

selfdrive/car/gm/values.py
```diff
@@ -70,13 +83,15 @@ class GMCarInfo(CarInfo):
...
-  CAR.ACADIA: GMCarInfo("GMC Acadia 2018", video_link="https://www.youtube.com/watch?v=0ZN6DdsBUZo", footnotes=[Footnote.OBD_II]),
-  CAR.BUICK_REGAL: GMCarInfo("Buick Regal Essence 2018"),
-  CAR.ESCALADE_ESV: GMCarInfo("Cadillac Escalade ESV 2016", "ACC + LKAS", footnotes=[Footnote.OBD_II]),
+  CAR.HOLDEN_ASTRA: GMCarInfo("Holden Astra 2017",footnotes=[Footnote.OBD_II], harness=Harness.obd_ii),
+  CAR.VOLT: GMCarInfo("Chevrolet Volt 2017-18", footnotes=[Footnote.OBD_II], min_enable_speed=0, harness=Harness.obd_ii),
+  CAR.CADILLAC_ATS: GMCarInfo("Cadillac ATS Premium Performance 2018", footnotes=[Footnote.OBD_II], harness=Harness.obd_ii), 
+  CAR.MALIBU: GMCarInfo("Chevrolet Malibu Premier 2017", footnotes=[Footnote.OBD_II], harness=Harness.obd_ii),
+  CAR.ACADIA: GMCarInfo("GMC Acadia 2018", video_link="https://www.youtube.com/watch?v=0ZN6DdsBUZo", footnotes=[Footnote.OBD_II], harness=Harness.obd_ii),
+  CAR.BUICK_REGAL: GMCarInfo("Buick Regal Essence 2018",footnotes=[Footnote.OBD_II], harness=Harness.obd_ii),
+  CAR.ESCALADE_ESV: GMCarInfo("Cadillac Escalade ESV 2016", "ACC + LKAS", footnotes=[Footnote.OBD_II], harness=Harness.obd_ii),
```
JasonJShuler, line 92: AFAIK all the existing GM vehicles are attached in the same way - the obd2 harness as well as whatever means the customer chooses to disable the ASCM while keeping radar powered (e.g. L&P Harness). Thought they should all be updated with the same note and switched from custom to the obd_ii harness (since custom maps to developer, it could have let people to buy the wrong thing...)

selfdrive/car/gm/values.py
```diff
@@ -70,13 +83,15 @@ class GMCarInfo(CarInfo):
 
 
 CAR_INFO: Dict[str, Union[GMCarInfo, List[GMCarInfo]]] = {
-  CAR.HOLDEN_ASTRA: GMCarInfo("Holden Astra 2017", harness=Harness.custom),
-  CAR.VOLT: GMCarInfo("Chevrolet Volt 2017-18", footnotes=[Footnote.OBD_II], min_enable_speed=0, harness=Harness.custom),
-  CAR.CADILLAC_ATS: GMCarInfo("Cadillac ATS Premium Performance 2018"),
-  CAR.MALIBU: GMCarInfo("Chevrolet Malibu Premier 2017", harness=Harness.custom),
-  CAR.ACADIA: GMCarInfo("GMC Acadia 2018", video_link="https://www.youtube.com/watch?v=0ZN6DdsBUZo", footnotes=[Footnote.OBD_II]),
-  CAR.BUICK_REGAL: GMCarInfo("Buick Regal Essence 2018"),
-  CAR.ESCALADE_ESV: GMCarInfo("Cadillac Escalade ESV 2016", "ACC + LKAS", footnotes=[Footnote.OBD_II]),
```
JasonJShuler, line 79: Not sure if there is value in looking up the explicit package names for all these... I don't know if package names change from year to year either, or how that should be handled here.

Finally, now that we are adding cars without an ASCM, sometimes both camera-based and radar-based are options; both of which would be supportable, but the ASCM variant can do OP long, while the cam cannot (officially). Again - not sure how this would be handled. More of a future concern...

selfdrive/car/gm/interface.py
```diff
@@ -45,7 +49,10 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, disa
     ret = CarInterfaceBase.get_std_params(candidate, fingerprint)
     ret.carName = "gm"
     ret.safetyConfigs = [get_safety_config(car.CarParams.SafetyModel.gm)]
-    ret.pcmCruise = False  # stock cruise control is kept off
+    ret.pcmCruise = False  # For ASCM, stock cruise control is kept off (but not ACC)
```
JasonJShuler, line 52: The existing comment seemed to need expansion, but I could use help.
* What does PCM stand for / mean?
* What exactly is kept off - standard CC? I have gathered that ACC is kept enabled and the car thinks it is managing the ACC
* Do cars with an ASCM + ACC have separate CC and ACC?

I discovered that flipping this from False to True took care of all the little bits I had been manually commenting out in controlsd... Happy that it works but still a little confused

selfdrive/car/gm/values.py
```diff
@@ -140,8 +155,23 @@ class CanBus:
...
   }],
   CAR.ESCALADE_ESV: [
+  # Note: This fingerprint is incomplete and causes false matches
   {
     309: 1, 848: 8, 849: 8, 850: 8, 851: 8, 852: 8, 853: 8, 854: 3, 1056: 6, 1057: 8, 1058: 8, 1059: 8, 1060: 8, 1061: 8, 1062: 8, 1063: 8, 1064: 8, 1065: 8, 1066: 8, 1067: 8, 1068: 8, 1120: 8, 1121: 8, 1122: 8, 1123: 8, 1124: 8, 1125: 8, 1126: 8, 1127: 8, 1128: 8, 1129: 8, 1130: 8, 1131: 8, 1132: 8, 1133: 8, 1134: 8, 1135: 8, 1136: 8, 1137: 8, 1138: 8, 1139: 8, 1140: 8, 1141: 8, 1142: 8, 1143: 8, 1146: 8, 1147: 8, 1148: 8, 1149: 8, 1150: 8, 1151: 8, 1216: 8, 1217: 8, 1218: 8, 1219: 8, 1220: 8, 1221: 8, 1222: 8, 1223: 8, 1224: 8, 1225: 8, 1226: 8, 1232: 8, 1233: 8, 1234: 8, 1235: 8, 1236: 8, 1237: 8, 1238: 8, 1239: 8, 1240: 8, 1241: 8, 1242: 8, 1787: 8, 1788: 8
```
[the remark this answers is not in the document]
NikitaNekrasov: OPGM has a more complete fingerprint for this car, and at least for the 2017 Escalade ESV, can cause Cruise Faults on stock OP where OPGM would not.

JasonJShuler:
FYI - I have divided this up into several smaller, easier to test PRs. Per @adeebshihadeh 's request, once the prerequisite PRs are merged, I'll force push this into just a simple Silverado port, and then I can create the other PRs for other vehicles - you guys can pull them in at your leisure

selfdrive/car/gm/values.py
```diff
@@ -145,8 +160,23 @@ class CanBus:
...
+  }],
+  CAR.SILVERADO: [
+  # 2020 Chevrolet Silverado 1500 4WD Crew Cab 147" LTZ
+  {
+    190: 6, 193: 8, 197: 8, 201: 8, 208: 8, 209: 7, 211: 2, 241: 6, 249: 8, 257: 8, 288: 5, 289: 8, 298: 8, 304: 3, 309: 8, 311: 8, 313: 8, 320: 4, 322: 7, 328: 1, 352: 5, 381: 8, 384: 4, 386: 8, 388: 8, 413: 8, 451: 8, 452: 8, 453: 6, 455: 7, 460: 5, 463: 3, 479: 3, 481: 7, 485: 8, 489: 8, 497: 8, 500: 6, 501: 8, 528: 5, 532: 6, 560: 8, 562: 8, 563: 5, 565: 5, 608: 8, 609: 6, 610: 6, 611: 6, 612: 8, 613: 8, 707: 8, 715: 8, 717: 5, 761: 7, 789: 5, 800: 6, 810: 8, 840: 5, 842: 5, 844: 8, 848: 4, 869: 4, 880: 6, 977: 8, 1001: 8, 1011: 6, 1017: 8, 1020: 8, 1033: 7, 1034: 7, 1217: 8, 1221: 5, 1233: 8, 1249: 8, 1259: 8, 1261: 7, 1263: 4, 1265: 8, 1267: 1, 1271: 8, 1280: 4, 1296: 4, 1300: 8, 1930: 7
+  }],
+  CAR.BOLT_EUV: [
+  # Chevy Bolt EUV 2019 (Stock VOACC, LKAS)
```
minorOffense, line 177: The year in this comment should probably be 2022

selfdrive/car/gm/values.py
sshane: https://github.com/commaai/openpilot/pull/25356

sshane:
@JasonJShuler rebased (merged) this PR and removed everything already merged into master. Let me know if there's anything else you need to do here or on the panda PR and I can review it

selfdrive/car/gm/carcontroller.py
```diff
@@ -46,18 +49,26 @@ def update(self, CC, CS):
     if CS.lka_steering_cmd_counter != self.lka_steering_cmd_counter_last:
       self.lka_steering_cmd_counter_last = CS.lka_steering_cmd_counter
     elif (self.frame % self.params.STEER_STEP) == 0:
-      lkas_enabled = CC.latActive and CS.out.vEgo > self.params.MIN_STEER_SPEED
+      # GM EPS faults on any gap in received message counters. To handle transient OP/Panda safety sync issues at the
+      # moment of disengaging, increment the counter based on the last message known to pass Panda safety checks.
+      idx = (CS.lka_steering_cmd_counter + 1) % 4
+      # Start counting number of resubmits - this should only happen when the panda drops an LKA frame
+      # If we reach 3 attempts to send the same frame, it's time to switch to inactive
```
sshane, line 56: Want to make this a separate PR for the EPS fault fix?

selfdrive/car/gm/values.py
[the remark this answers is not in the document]
sshane: Feel free to make another PR for this.

selfdrive/car/docs_definitions.py
adeebshihadeh: perhaps we just call this "gm"
```suggestion
  gm = "GM" 
```

selfdrive/car/docs_definitions.py
```diff
@@ -209,6 +209,14 @@ class Harness(Enum):
...
   custom = "Developer"
   obd_ii = "OBD-II"
+  gm_cam = "GM Camera" 
+  # TODO: Submit PR for gm cam harness schematic(s), diagrams and board layout(s) to https://github.com/commaai/neo/tree/master/car_harness
+  # Note: GM cam harness _without low-speed GMLAN support_ is simple and similar to the others
+  #       JJS - I anticipate comma to be able to take over production easily
+  #       It also doesn't make sense for customers to get the Developer harness - they typically don't need the wires
+  # Note 2: JJS - I anticipate designing a camera harness with a low-speed GMLAN gateway (design borrowed from OBD2 harness),
+  #         unless comma plans to do so
+  # Note 3: JJS - I anticipate an eventual
```
adeebshihadeh, line 219: remove these

selfdrive/car/gm/carcontroller.py
JasonJShuler: Yes

selfdrive/car/gm/interface.py
sshane: Do you know if the EUV has similar torque to the other GMs with the same units?

sshane:
- [ ] We need a way to cancel stock cruise control when not using openpilot longitudinal

selfdrive/car/gm/carcontroller.py
```diff
@@ -107,13 +107,23 @@ def update(self, CC, CS):
       if self.CP.networkLocation == NetworkLocation.gateway and self.frame % self.params.ADAS_KEEPALIVE_STEP == 0:
         can_sends += gmcan.create_adas_keepalive(CanBus.POWERTRAIN)
 
+    elif self.CP.networkLocation == NetworkLocation.fwdCamera:
+      # Stock longitudinal, integrated at camera
+      if self.frame % 3 == 0:
```
adeebshihadeh, line 112: we want to send button presses immediately, especially the cancel. also this seems pretty fast - is this rate necessary?

sshane: This is the stock rate of 33hz and it takes 100-200 ms to cancel, we'll try lower though.

sshane: Sending immediately at 10Hz, these are the number of frames to cancel cruise (at 100Hz): 23, 21, 11, 23, 13, 11, 16.

Sometimes it only cancels after we send a single button message, so the best we can do is 10 openpilot frames (100 ms).

sshane:
panda and this PR should be ready for final review!

selfdrive/car/gm/carcontroller.py
```diff
@@ -107,13 +107,22 @@ def update(self, CC, CS):
       if self.CP.networkLocation == NetworkLocation.gateway and self.frame % self.params.ADAS_KEEPALIVE_STEP == 0:
         can_sends += gmcan.create_adas_keepalive(CanBus.POWERTRAIN)
 
+    elif self.CP.networkLocation == NetworkLocation.fwdCamera:
```
adeebshihadeh, line 110: an else seems more appropriate here. fwdCamera just happens to mean "not openpilot long" in this case
```suggestion
    else:
```

selfdrive/car/gm/carstate.py
```diff
@@ -77,9 +79,21 @@ def update(self, pt_cp, loopback_cp):
 
     ret.cruiseState.enabled = pt_cp.vl["AcceleratorPedal2"]["CruiseState"] != AccState.OFF
     ret.cruiseState.standstill = pt_cp.vl["AcceleratorPedal2"]["CruiseState"] == AccState.STANDSTILL
+    if self.CP.networkLocation == NetworkLocation.fwdCamera:
+      ret.cruiseState.speed = (cam_cp.vl["ASCMActiveCruiseControlStatus"]["ACCSpeedSetpoint"] / 16) * CV.KPH_TO_MS
```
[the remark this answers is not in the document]
sshane: There's a strange comment in gmcan.py that doesn't make sense, so I'll probably move it into the DBC in a separate PR

selfdrive/car/gm/values.py
```diff
@@ -87,6 +83,7 @@ class GMCarInfo(CarInfo):
   CAR.ACADIA: GMCarInfo("GMC Acadia 2018", video_link="https://www.youtube.com/watch?v=0ZN6DdsBUZo"),
   CAR.BUICK_REGAL: GMCarInfo("Buick Regal Essence 2018"),
   CAR.ESCALADE_ESV: GMCarInfo("Cadillac Escalade ESV 2016", "Adaptive Cruise Control (ACC) & LKAS"),
+  CAR.BOLT_EUV: GMCarInfo("Chevrolet Bolt EUV 2022", "Chevy Safety Assist", footnotes=[], harness=Harness.none),
```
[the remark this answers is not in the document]
sshane: yep!

adeebshihadeh (approved):

sshane:
Thanks for the PR @JasonJShuler!

PR #24876, opened 2022-06-16
Mazda: fix resume spam at standstill

sshane:
When I made the switch from active and enabled to lat/longActive I accidentally lumped this in with latActive, as before active was True when enabled

### DIFF
--- a/selfdrive/car/mazda/carcontroller.py
+++ b/selfdrive/car/mazda/carcontroller.py
@@ -27,6 +27,7 @@ class CarController():
                                                   CS.out.steeringTorque, CarControllerParams)
       self.steer_rate_limited = new_steer != apply_steer
 
+    if c.enabled:
       if CS.out.standstill and frame % 5 == 0:
         # Mazda Stop and Go requires a RES button (or gas) press if the car stops more than 3 seconds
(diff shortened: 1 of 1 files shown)


selfdrive/car/mazda/carcontroller.py
```diff
@@ -27,11 +27,11 @@ def update(self, c, CS, frame):
                                                   CS.out.steeringTorque, CarControllerParams)
       self.steer_rate_limited = new_steer != apply_steer
 
-      if CS.out.standstill and frame % 5 == 0:
-        # Mazda Stop and Go requires a RES button (or gas) press if the car stops more than 3 seconds
-        # Send Resume button at 20hz if we're engaged at standstill to support full stop and go!
-        # TODO: improve the resume trigger logic by looking at actual radar data
-        can_sends.append(mazdacan.create_button_cmd(self.packer, self.CP.carFingerprint, CS.crz_btns_counter, Buttons.RESUME))
+    if c.enabled and CS.out.standstill and frame % 5 == 0:
```
sshane, line 30: We should find a cruiseState signal for standstill, as we will start using that in this PR: https://github.com/commaai/openpilot/pull/24873

sshane:
@alancyau want to try this? Should fix resuming from a stop

PR #24880, opened 2022-06-16
Honda Bosch Longitudinal: use clipped actuators.accel for gasbrake lookup

hewers:
**Description** Found this while refactoring gasbrake to account for pitch. Changes process replay for one segment. If this is the intended behavior, I'll add a comment in the refactor.

**Verification** a priori


pd0wm:
I wouldn't expect this to make a difference, since the lookup BP don't go past the clipped values. Maybe a rounding error?

qadmus:
Yeah my experiment was tainted, finding another cause for my process replay issue.

PR #24881, opened 2022-06-16
Mazda CX-5 2022: steer from zero

### LINKED ISSUE #23984: 2022 Mazda CX-5 - Steer Unavailable Below 0 mph

sshane:
For all other cars minSteerSpeed will not be zero, so behavior will not be changed. And for the CX5 2022, we will only show unavailable alert when leaving standstill (takes a few seconds for steering to become available again which we don't want to ignore).


selfdrive/car/mazda/interface.py
```diff
@@ -57,7 +57,9 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, disa
       ret.lateralTuning.pid.kpV, ret.lateralTuning.pid.kiV = [[0.19], [0.019]]
       ret.lateralTuning.pid.kf = 0.00006
 
-    if candidate not in (CAR.CX5_2022, ):
+    if candidate in (CAR.CX5_2022, ):
```
sshane, line 60: @Jafaral should this also include CX-9 2021?

Jafaral: No it shouldn't. CX-9 2021 still has the 28mph min steer speed.

sshane:
@alancyau do you mind trying this out?

alancyau:
Tested this branch. The warning went away when the car comes to a stop, but when the lead car starts to accelerate, a `Steering Temporarily Unavailable` warning pops up and OP will disengage. Here are two instances of that happening.

a4af1602d8e668ac|2022-06-17--17-36-07--3
a4af1602d8e668ac|2022-06-17--17-36-07--4

sshane:
That should be normal, to openpilot it sees the car not steering and throws the appropriate warning. It shouldn't disengage though

sshane:
@Jafaral do the Mazdas not steer right away when you engage? Resuming from a stop should be similar, but it takes multiple seconds to start steering again. Can we somehow reset the EPS so it resumes steering immediately?

Jafaral:
@sshane , they do steer right away when you engage, but there is a lag for the EPS to  respond for steer commands  if it was in an LKAS_BLOCK state

sshane:
So I think an alert is fine, and maybe we need a list of cars we know can steer to zero to ignore "faults" and just show a steer unavailable alert

Jafaral:
Yeah, in practice the user will hardly notice that steer   "warm up" when coming out of a full stop. With ACC, the car is already a bit too slow to accelerate anyway.

sshane:
In the example I posted to Discord, the user's car didn't start steering until 20 mph or so.

alancyau:
> Waiting on a stock route from @alancyau in sng to see what the camera does, odd that it doesn't steer for this long.

Which fork did you want a stop and go route for?

sshane:
Just master, as long as dashcam mode is on (openpilot is disabled).

sshane:
Any updates? @alancyau

alancyau:
> Any updates? @alancyau

I should be able to do a test today and send you a log.

alancyau:
@sshane here’s a log from master-ci in dashcam mode. 

a4af1602d8e668ac|2022-06-25--18-10-44--1

sshane:
Did you engage in that route? We need to see an example where you're stopped behind the lead engaged, the car pulls away, and the Mazda starts steering on its own.

alancyau:
I’ll need to do another route. It was only in dashcam mode. Got confused with the requested route.

sshane:
Yes correct, dashcan mode uses your stock car's system and the C3 just records the data for us. If you can get it to resume from a stop and have it steer as fast as possible, that'll be helpful.

alancyau:
@sshane there should be a few SnG in this log running master-ci. 

a4af1602d8e668ac|2022-07-01--14-41-25--13

sshane:
Thanks for that data! But we need a route with the stock system (openpilot turned OFF in the settings) to evaluate how stock handles resuming from a stop, steering wise.

alancyau:
@sshane that makes sense now. I'll get you another route running the factory ACC/LKAS system tomorrow in similar situation. 

Starting with the 2022 CX5, traffic jam assist is offered on the highest Touring trim. I have the lower Select trim. Traffic Jam Assist allows steering below 37mph. I think above 37, Lane Keep Assist takes over. I believe all the trims just run the same EPS, so it's able to accept steering commands below 37.

https://insidemazda.mazdausa.com/press-release/2022-mazda-cx-5-pricing-and-packaging/

sshane:
Oh, so yours wouldn't be able to steer under 37 with the stock system? Dang, that would have been helpful to see what the camera does. I checked your route and it seems very variable, most of the time it waits until ~10 mph to enable steering, while others (when not engaged) it's 2 mph.

alancyau:
I'm not really sure what the stock system does actually. The marketing brochure is a bit confusing on what Traffic Jam Assist actually does. I'll just test the stock system tomorrow and see what the data looks like.

image

sshane:
It actually looks like there's a 3.2 second delay coming from a stop, which even exists when not engaged. So I'll fix this PR up to hide the alert when coming to a stop, I think the alert when resuming is appropriate considering it isn't steering for 3 seconds.


Screenshot from 2022-07-01 18-29-57

alancyau:
Here’s a log in dashcam mode with ACC and LKAS engaged on release3. 

a4af1602d8e668ac|2022-07-02--18-12-53--17

sshane:
Thanks for the route. It's about as expected, LKAS_BLOCK rises at around 19 m/s on stock, so it's not any better (maybe a different trim would shine some light on this though).

sshane:
@alancyau can you try the branch now? I will tune the thresholds through your data and clean the branch up, but this should work

alancyau:
> @alancyau can you try the branch now? I will tune the thresholds through your data and clean the branch up, but this should work

Any specific scenarios you’d like to see?

sshane:
Coming to a stop engaged and resuming from a stop engaged, as many times as possible (so I can verify the time it takes to lose lkas_block)

alancyau:
> Coming to a stop engaged and resuming from a stop engaged, as many times as possible (so I can verify the time it takes to lose lkas_block). Let me know if it ever shows that alert

Just saw this. Which UI alert? 

Here’s a log with 4 full stops while engaged running this branch. 

a4af1602d8e668ac|2022-07-14--08-32-13--19

sshane:
The `Steering Temporarily Unavailable` alert you complained of above is what we're looking to avoid.

alancyau:
Yeah, I was getting the chime sound and orange screen in that previous drive.

sshane:
Since these steer to zero cars consistently take 3.2 seconds to start steering again after resuming, we'll have to keep track of the time resuming from a stop and show an alert if we still can't steer after that time

sshane:
@Jafaral seems to have found that LKAS_BLOCK doesn't mean anything on the CX5 and the stock system applies torque when it's high. @alancyau can you test this branch now to see if your car steers from a stop now? The alert should be gone now, but we want to make sure that it can steer immediately after resuming from a stop.

alancyau:
Yeah I’ll have access to the car in a few days to test.

Jafaral:
Just in case someone reads these comments, This only affects CX-5 2022,. `LKAS_BLOCK` still applies for  all other models/years

Jafaral:
@sshane , can you update the PR title to say CX-5 2022 please?

alancyau:
@sshane I was able to only get one instance of a full stop and go. See a4af1602d8e668ac|2022-08-28--15-20-10--8. It appears the wheel immediately moved when the car started accelerating. Do you want to check the logs to confirm this?

Jafaral (approved):
LGTM!

sshane:
Nice that this just works! Checking the logs now. @Jafaral I see the comment says it can also block for a no-user-torque condition. Can we test that as well? How does it work? Or is the comment outdated?

sshane:
@Jafaral does this look like it was applying torque? Screenshot_22

alancyau:
> @Jafaral does this look like it was applying torque? There was only one example @alancyau can you try to get a few more?
> 
> Screenshot_22

Yeah I’ll get a few more in the next couple of days.

Jafaral:
Interesting. This drive seems to indicate there was a delay in applying torque. Good to see more drives. Can you please share the drives with me @alancyau?

I wonder if  @kravatch's cars is a bit different since he has "official" TJA support. I didn't see any delay in his drives. The next thing would be to look at firmware versions for both the EPS and FSC for both cars. Also look for any difference/bits in the FSC signals.

alancyau:
Here's a log with a few more stop and go testing on @Jafaral's https://github.com/Jafaral/openpilot/tree/mazda-tja branch.

a4af1602d8e668ac|2022-09-02--22-30-21--7

When the lead car starts moving, seems like the steering wheel will move right away for the first second, then appear dead for about 2 seconds, then start moving again.

sshane:
@Jafaral do you know if there is an apply bit for steering control? perhaps the EPS doesn't like us setting it while stopped or LKAS_BLOCKED is true

sshane:
Any of you guys active here?

alancyau:
> Any of you guys with this issue active here?

I don’t have a C3 device anymore, so unable to test this issue.

kravatch:
> > Any of you guys with this issue active here?
> 
> I don’t have a C3 device anymore, so unable to test this issue.

I'm around and able to test. Can you confirm the branch to try and specific scenario to simulate?

sshane:
@kravatch In stop and go, can your car steer from a stop? Or does it fail to steer?

kravatch:
> @kravatch In stop and go, can your car steer from a stop? Or does it fail to steer?

Using OP, my car does fail to steer from a stop. It throws a warning but reengages very quickly once I start up

sshane:
Can you provide a route/segment and let me know at what speed it actually started steering at?

adeebshihadeh:
Closing, unclear if the stock system even does this. We would merge a well tested PR for this though.

PR #24888, opened 2022-06-17
laikad: use non-conflating socket for gnss msgs

gijskoning:


pd0wm:
Let's add to process replay first with the submaster. Subsockets need some special care in process replay.

adeebshihadeh:
@HaraldSchafer want to keep this open?

HaraldSchafer:
Don't understand this change, so closing. @adeebshihadeh Feel free to merge if you want to.

pd0wm:
Using a submaster can drop messages if the main loop takes too long to process. Specifically dropping a nav message would be annoying. Not sure if this is a problem in practice.

adeebshihadeh:
Yeah, we shouldn't really not do this because it's not convenient to test in process replay. I'll reopen and handle making it work in process replay.

PR #24890, opened 2022-06-17
ui: metric wider set speed box

pd0wm:
Before:
2022-06-17-112144_480x600_scrot

After:
2022-06-17-141219_400x497_scrot

Waiting for approval from @jozef-simo


jozef-simo:
Looking great! 👍

PR #24894, opened 2022-06-17
Toyota: longitudinal improvements

sshane:
Probably will split some of this stuff out


sshane:
The start boost and no proportional at a magic number speed isn't the cleanest, and ends up over/undershooting heavily depending on the pitch of the road. Reducing proportional without impacting brake performance is a better solution for now

PR #24895, opened 2022-06-18
Chrysler: use unified signal definitions

sshane:
Just the easiest to verify done so far, will reduce diff from RAM port


sshane:
Angle/rate signals look totally wrong on this route, we'll need to verify it on a lot of cars first it looks like. We can revert that change to merge this now though

sshane:
Actually I believe our signal to be incorrect now looking at the data more closely. The spec steering ratio for a Pacifica is 16.2, but if you check a Pacifica's liveParameters, openpilot is learning ~10.98, that's a 1.47540984x difference.

sshane:
Tested this signal definition on a RAM 1500, looks correct going from 0 to 90 to 180 to 270+

selfdrive/car/chrysler/carstate.py
```diff
@@ -51,12 +51,12 @@ def update(self, cp, cp_cam):
     ret.steeringRateDeg = cp.vl["STEERING"]["STEERING_RATE"]
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(cp.vl["GEAR"]["PRNDL"], None))
 
-    ret.cruiseState.enabled = cp.vl["ACC_2"]["ACC_STATUS_2"] == 7  # ACC is green.
```
sshane, line 54: If it was 7, then available bit needed to be 1

PR #24897, opened 2022-06-18
ui: add more color define

crwusiz:


pd0wm:
Don't think it makes sense to add defines for colors only used once. Will go over all the buttons though and make sure they match.

PR #24899, opened 2022-06-18
New Elantra HEV fingerprinting

maxtriano:
Started getting "Unsupported Car/Dashcam Mode" yesterday in my 2022 Elantra HEV that had been working fine previously, looks like the fingerprint needed updating for some reason.  

Here's a route with openpilot working with the newly added fingerprint: 786fc028c014be71|2022-06-18--11-37-25


maxtriano:
Closing for now until we figure out if this is caused by a bug or if the car actually needs a new fingerprint

Issue #24900, opened 2022-06-18
2022 Elantra HEV no longer recognized as supported car


maxtriano:
### Describe the bug

Seeing this issue as of yesterday, running master-ci. Openpilot had previously worked, but now shows the car as Unsupported and goes into dashcam mode. carFw shows new fwVersion values for engine and transmission, and openpilot works when these are added to values.py. To my knowledge HKG hasn't pushed out an OTA firmware update, but apparently this is happening for other Hyundai owners with different cars, so it's possible. 

Here's a route I drove with the new fingerprint fwVersions added and openpilot working: 786fc028c014be71|2022-06-18--11-37-25

### Which car does this affect?

2022 Hyundai Elantra HEV

### Provide a route where the issue occurs

786fc028c014be71|2022-06-17--22-18-06

### openpilot version

30c3bc07874cbbc57780e777c156f90d710b54bd

dermida:
Afects Hyundai Ioniq EV 2020 to.
Drive  with car unrecognized:
b1078edf825d2db6|2022-06-19--10-41-40

Drive where the car is recognized:
b1078edf825d2db6|2022-06-18--22-30-50

dermida:
Examples where car unrecognized and the day before where is not recognized
I changed the USB cable to a new one to

'carFingerprint': 'mock',
'carFw': [{'address': 2004,
                          'ecu': 'eps',
                          'fwVersion': b'\xf1\x00AE  MDPS C 1.00 1.01 56310/G7310 4APEC101',
                          'responseAddress': 0,
                          'subAddress': 0},
                         {'address': 2000,
                          'ecu': 'fwdRadar',
                          'fwVersion': b'\xf1\x00AEev SCC F-CUP      1.00 1.01 99110-G7000         ',
                          'responseAddress': 0,
                          'subAddress': 0},
                         {'address': 1988,
                          'ecu': 'fwdCamera',
                          'fwVersion': b'\xf1\x00AEE MFC  AT EUR LHD 1.00 1.00 95740-G2600 190730',
                          'responseAddress': 0,
                          'subAddress': 0}],
               'carName': 'hyundai',

'carFingerprint': 'HYUNDAI IONIQ ELECTRIC 2020',
carFw': [{'address': 2004,
                          'ecu': 'eps',
                          'fwVersion': b'\xf1\x00AE  MDPS C 1.00 1.01 56310/G7310 4APEC101',
                          'responseAddress': 0,
                          'subAddress': 0},
[...]

PR #24903, opened 2022-06-19
Hyundai : KONA HYBRID 2020 add CARFW

haram-KONA:
hyundai kona hybrid
add engine, transmission carfw to hyundai kona hybrid
Here's a route with openpilot working with the newly added fingerprint

route : 9a3aa85e86f9c3fc|2022-06-19--09-21-50


haram-KONA:
ISSUES #24900

PR #24904, opened 2022-06-19
Toyota: improve response to acceleration requests

### LINKED ISSUE #1098: Some toyotas dont react on accel from stop well
### Describe the bug

https://github.com/commaai/openpilot/pull/24363

This or something like this seems needed for some toyotas

some TSS2 toyotas

NA

### openpilot version

master

### Additional info

_No response_

sshane:
Toyotas seem reluctant to apply acceleration requests immediately if this bit is set, even if you request a static acceleration of 1.0 m/s/s. There are still issues with the controller lowering or even making acceleration negative if we overshoot the plan while starting which diminishes this improvement, but this actually does make the car respond better to acceleration requests.

Here are two examples with the change showing the worse possible scenario for the car; as soon as we take off we lower the requested acceleration which normally would have the car let off the gas quite a bit, however it continues on to reach the target, albeit with a little bit of slowing that you can feel.

Screenshot from 2022-06-18 23-10-47

Screenshot from 2022-06-18 23-08-19
 
Routes with latest changes for my reference:

f15e3c37c118e841|2022-06-30--00-51-03--0
f15e3c37c118e841|2022-06-30--00-59-09--0
f15e3c37c118e841|2022-06-30--01-01-57--0


selfdrive/car/toyota/toyotacan.py
```diff
@@ -35,7 +35,7 @@ def create_accel_command(packer, accel, pcm_cancel, standstill_req, lead, acc_ty
     "ACC_TYPE": acc_type,
     "DISTANCE": 0,
     "MINI_CAR": lead,
-    "PERMIT_BRAKING": 1,
+    "PERMIT_BRAKING": 1 if accel <= 0 else 0,
```
sshane, line 38: @HaraldSchafer this might not be the best way to do this correctly. If you're on a downward facing hill for example, 0 or even some small positive acceleration means we might want to brake. Should we use the localizer to find the coasting acceleration here?

HaraldSchafer: Can you test it on a hill and look at the plots? Maybe they deal with this internally. Just want to be sure they don't before we start writing logic.

sshane: They do not, it fails to brake no matter how sharp the decline is

sshane:
@HaraldSchafer Are we okay with Toyota overshooting the planner's request with this PR? When starting, nearly every time it will overshoot not only the start, but throughout accelerating. We don't exactly request what the stock system does from a start, so it releases the brakes super quickly rather than ramping off them to avoid overshoot.

We might need some manual profile to fix this, but it might be different per cars (hybrids might get moving faster coasting from a stop like mine).

Screenshot from 2022-07-18 23-51-31

sshane:
Closing as this bit isn't a great way to generically improve starting accuracy (overshoots acceleration at all speed, and can be dangerous if pitch is wrong)

PR #24907, opened 2022-06-19
Chrysler: fill cruiseState.available

### LINKED ISSUE #24906: Chrysler ACC instant faults with latest master-ci build
### Describe the bug

Can't enable ACC with Comma 3 device plugged in. Controls unresponsive on steering wheel until system ultimately faults. Works fine when comma device is unplugged. 

Chrysler Pacifica 2022

4395faadf0a373f2|2022-06-19--15-41-19--0

### openpilot version

0.8.15 Nightly (6/19/2022)

### Additional info

_No response_

sshane:


sshane:
I'll see why process replay didn't fail first before merging

sshane:
The segment in test_processes is only available and engaged so flipping them has no effect. We should soon try to get more comprehensive routes, and test more things like doors, blinkers, gas, etc.

sshane:
This is good to go now, but not sure we want to start using this camel case convention in the DBC, it's not in the reference DBC and was chosen by @realfast I think

PR #24908, opened 2022-06-20
GM: values.py cleanup & minor updates

JMPZ11:
**Description**
Various cleanup / refactorings in GM values.py:
* Update / supplement comments on static limits, and small rearranging for clarity
* All current vehicles use the OBD II harness, not the custom/developer, and all require the same footnote
  * The only change to generated cars.md is that the Acadia and Escalade now have a supported star
* Minor cleanup of top fingerprint to match typical format
* Commented out the problematic, presumably incomplete Escalade ESV fingerprint
  * It is shorter than most, and is missing some common message IDs
  * It was causing incorrect fingerprinting with Volts, Acadias, and various new cars at random
  * If it's removal causes the intended vehicle to stop being identified, a complete FP should be collected
  * Added a new, working fingerprint (may not be for the exact same car)
* Added the virtual CanBus number for messages dropped by the Panda
* Updated the model DBC dictionary to be auto-generated from the values in the CAR class
  * All GM vehicles use the same set of DBC files - this change reduces redundant definitions and simplifies adding new models
  * Includes CUSTOM_DBC (with comment / example) for any models needing non-default dbc dict

**Verification**
Most of the changes are cosmetic / non-impactful
* Validated the generated output to cars.md is corrected
* Escalade ESV fingerprint change may cause the original car to stop fingerprinting, but this can be resolved by collecting a complete FP, and the newly added fingerprint has been working in OPGM for months
* Value of generated DBC constant is identical to the previous explicitly defined version


selfdrive/car/gm/values.py
```diff
@@ -140,17 +143,18 @@ class CanBus:
...
-  CAR.MALIBU: dbc_dict('gm_global_a_powertrain_generated', 'gm_global_a_object', chassis_dbc='gm_global_a_chassis'),
-  CAR.ACADIA: dbc_dict('gm_global_a_powertrain_generated', 'gm_global_a_object', chassis_dbc='gm_global_a_chassis'),
-  CAR.CADILLAC_ATS: dbc_dict('gm_global_a_powertrain_generated', 'gm_global_a_object', chassis_dbc='gm_global_a_chassis'),
-  CAR.BUICK_REGAL: dbc_dict('gm_global_a_powertrain_generated', 'gm_global_a_object', chassis_dbc='gm_global_a_chassis'),
-  CAR.ESCALADE_ESV: dbc_dict('gm_global_a_powertrain_generated', 'gm_global_a_object', chassis_dbc='gm_global_a_chassis'),
+CUSTOM_DBC: Dict[str,Dict[str,str]] = {
+  #e.g.: CAR.EXAMPLE: dbc_dict('custom_powertrain', 'custom_object', chassis_dbc='custom_chassis'),
 }
+DEFAULT_DBC_VAL = dbc_dict('gm_global_a_powertrain_generated', 'gm_global_a_object', chassis_dbc='gm_global_a_chassis')
+DBC = {str(getattr(CAR,car)): DEFAULT_DBC_VAL for car in dir(CAR) if not car.startswith("_") and car not in CUSTOM_DBC}
```
sshane, line 159: You can use a defaultdict

JasonJShuler: Awesome! Done!

selfdrive/car/gm/values.py
```diff
@@ -66,16 +68,16 @@ class Footnote(Enum):
 @dataclass
 class GMCarInfo(CarInfo):
   package: str = "Adaptive Cruise"
-  harness: Enum = Harness.none
+  harness: Enum = Harness.obd_ii
 
 
 CAR_INFO: Dict[str, Union[GMCarInfo, List[GMCarInfo]]] = {
-  CAR.HOLDEN_ASTRA: GMCarInfo("Holden Astra 2017", harness=Harness.custom),
```
sshane, line 73: @vanillagorillaa do we want to start showing these cars in the harness list? It may appear to end users all these cars need is an OBD II harness and all is well. Before they would have needed to read the footnote on the supported cars page first to know to get a custom harness.

selfdrive/car/gm/values.py
```diff
@@ -140,17 +144,15 @@ class CanBus:
...
...7: 8, 199: 4, 201: 8, 208: 8, 209: 7, 211: 2, 241: 6, 249: 8, 288: 5, 289: 1, 290: 1, 298: 8, 300: 1, 304: 8, 309: 8, 311: 8, 313: 8, 320: 8, 328: 1, 352: 5, 368: 8, 369: 8, 381: 5, 384: 4, 386: 5, 388: 8, 393: 7, 398: 8, 407: 7, 413: 8, 417: 7, 419: 1, 422: 4, 426: 7, 431: 8, 442: 8, 451: 8, 452: 8, 453: 6, 454: 8, 455: 7, 463: 3, 479: 3, 481: 7, 485: 8, 487: 8, 489: 8, 497: 8, 499: 3, 500: 6, 501: 8, 508: 8, 510: 8, 512: 3, 530: 8, 532: 6, 534: 2, 554: 3, 560: 8, 562: 8, 563: 5, 564: 5, 573: 1, 577: 8, 578: 8, 579: 8, 587: 8, 608: 8, 609: 6, 610: 6, 611: 6, 612: 8, 613: 8, 647: 6, 707: 8, 715: 8, 717: 5, 719: 5, 761: 7, 789: 5, 800: 6, 801: 5, 803: 8, 804: 3, 805: 8, 810: 8, 821: 4, 823: 7, 832: 8, 840: 5, 842: 5, 844: 8, 866: 4, 869: 4, 880: 6, 961: 8, 967: 4, 969: 8, 977: 8, 979: 8, 985: 5, 1001: 8, 1005: 6, 1009: 8, 1013: 3, 1017: 8, 1019: 2, 1020: 8, 1033: 7, 1034: 7, 1105: 6, 1217: 8, 1221: 5, 1223: 2, 1225: 7, 1233: 8, 1249: 8, 1257: 6, 1265: 8, 1267: 1, 1280: 4, 1296: 4, 1300: 8, 1322: 6, 1323: 4, 1328: 4, 1417: 8, 1601: 8, 1602: 8, 1603: 7, 1611: 8, 1618: 8, 1906: 7, 1907: 7, 1912: 7, 1917: 7, 1918: 7, 1919: 7, 1920: 7, 1930: 7, 2016: 8, 2018: 8, 2019: 8, 2024: 8, 2026: 8
```
sshane, line 152: Let's make this PR a strict cleanup, revert this for another PR, ideally with a new ESCALADE fp.

selfdrive/car/gm/values.py
JasonJShuler: I believe the custom harness is represented as the developer harness, and the developer harness wouod be useless - they are all setup the same as the Volt.

The l&p harness isn't mandatory either, it's just convenient. The same basic instructions at https://zoneos.com/volt/ apply to all ASCM-bearing vehicles

sshane: I think we still want to force people to read the footnote which lets users know they won't have a good time unless they get the ASCM harness (due to faults I believe). Until we are able to pop up with limitations when selecting a car harness, at least

PR #24910, opened 2022-06-20
GM: prep and cleanup for future ports

JMPZ11:
**Description**
Update GM Port to populate and utilize crucial CarParams values that have historically been assumed, unpopulated and/or hard-coded. These changes prepare the port to support vehicles without an ASCM, as well as Electric cars other than the Chevy Volt. Vehicles without an ASCM must use stock longitudinal (either from the camera or a distance-sensing Cruise module) and these changes give OP the ability to permit this.

* Ensure default values are populated for pcmCruise and radarOffCan
* Update calls to common event methods, populating pcm_enable / pcm_cruise parameters with value of pcmCruise. This prevents these methods from throwing errors when factory ACC is in use
* Populate CarParams transmissionType and nertworkLocation
  * transmissionType differentiates between gas and EV/Hybrids
  * networkLocation specifies whether the device is connected at the OBD-II port (presumably in a vehicle with an ASCM) or using the camera harness (in vehicles without an ASCM)
* Update carstate to use transmissionType for determining the use of EV properties, rather than hard-coded CAR.VOLT
* Update carcontroller to only send ACC, keepalive, HUD, Radar and GMLAN messages when the configuration permits
  * ACC & HUD messages only sent when OP is handling longitudinal
  * Radar related messages on the obstacle bus are only send when OP is handling long _and_ radar is present
  * ADAS Keepalives are only sent if OP is doing long and the OBD-II harness is in use (assuming the need to emulate an ASCM)
  * LKA dash indicator messages are only sent when not using the camera harness. (This is used rather than simply when using OBD-2 as _NOT_ sending them should be considered an exceptional case that only applies to the current cam harness design

**Verification**
The basic set of changes have been tested on a Silverado and a GMC Sierra which use camera-based ACC. An older (less restrictive) version has been in use by many in the OPGM forks.
These changes should not impact the behavior of existing models. Until the new vehicles are added, only the default values are used, and OP is expected to operate without change.

I will require assistance updating tests to expect the newly populated parameters, and to ensure OP behaves appropriately under the new scenarios.
Initially there will be two scenarios (Only the second scenario is new):
* ASCM (The configuration of all currently supported cars) - OP handles long, networkLocation is obd2, radarOffCan is False, pcmCruise is False
* **Camera Harness - OP does NOT handle long, networkLocation is fwdCamera, radarOffCan is True, pcmCruise is True**

Any other combination of these 4 parameters isn't yet supported


JasonJShuler:
Failure caused by
```***** differences for segment fakedata|2022-01-20--18-03-41--0 *****
*** process: controlsd ***
	('change', 'carParams.transmissionType', ('unknown', 'direct'))
	('change', 'carParams.networkLocation', ('fwdCamera', 'gateway'))
	('change', 'carParams.transmissionType', ('unknown', 'direct'))
	('change', 'carParams.networkLocation', ('fwdCamera', 'gateway'))```

It seem fwdCamera was the old default?

I will attempt to update the logs for the test

**Comment by JasonJShuler** [comment]
Maybe I'm not actually able to update the logs? I tried using git-lfs to no avail. 

https://github.com/commaai/openpilot/blob/master/selfdrive/test/process_replay/README.md

Maybe I misinterpreted? Anyway I'm stuck.

**Comment by adeebshihadeh** [comment]
@sshane anything blocking this? seems pretty straightforward

File: selfdrive/car/gm/carcontroller.py [hunk 1]
```diff
@@ -61,7 +62,7 @@ def update(self, CC, CS):
       can_sends.append(gmcan.create_steering_control(self.packer_pt, CanBus.POWERTRAIN, apply_steer, idx, lkas_enabled))
 
     # Gas/regen and brakes - all at 25Hz
-    if (self.frame % 4) == 0:
+    if CS.CP.openpilotLongitudinalControl and ((self.frame % 4) == 0):
```
adeebshihadeh, line 65: pass in the CarParams or save them in the constructor

sshane:
Just moved some common checks in the CC if statements out, how does this look @JasonJShuler?

JasonJShuler:
> Just moved some common checks in the CC if statements out, how does this look

Definitely looks cleaner, thanks!

PR #24911, opened 2022-06-20
GM: Set Panda to Cam Harness safety param when needed

JMPZ11:
**Description**
When the CarParam networkLocation is fwdCamera, set the panda safetyparam to FLAG_GM_HW_CAM to enable camera forwarding and support for stock longitudinal

Note: Depends on OP PR #24910 (Add/support CarParams transmissionType...) (added "car." to avoid merge conflict)
Note: Depends on Panda PR [commaai/panda/#962](https://github.com/commaai/panda/pull/962) (Support for cam harness...) (references unmerged branch)

**Verification**

gm-silverado-wip (PR #24875) has been tested on a Silverado and a GMC Sierra and the safetyparam works great!


adeebshihadeh:
Looks like this is done in the Bolt port PR. No need to break out this change.

PR #24916, opened 2022-06-20
FPv2: log all present ECU addresses

sshane:
Adds logging for all ECUs that return a success or fail response so we can run tests offline from data gathered while in master for a few days.


sshane:
Working well so far on these three cars! Didn't see a difference with engine on vs. off.

`Honda Civic matches: defaultdict(<class 'int'>, {'honda': 8})`

`Kia Optima matches: defaultdict(<class 'int'>, {'subaru': 2, 'hyundai': 6, 'toyota': 2, 'volkswagen': 2, 'mazda': 2, 'nissan': 1})`

`Toyota RAV4 matches: defaultdict(<class 'int'>, {'subaru': 2, 'hyundai': 3, 'toyota': 6, 'volkswagen': 1, 'mazda': 1, 'nissan': 1, 'body': 1})`

PR #24921, opened 2022-06-21
Refactor torque stuff

haraschax:


.pre-commit-config.yaml
```diff
@@ -19,7 +19,7 @@ repos:
     rev: v0.931
     hooks:
     -   id: mypy
-        exclude: '^(pyextra/)|(cereal/)|(rednose/)|(panda/)|(laika/)|(opendbc/)|(laika_repo/)|(rednose_repo/)/'
+        exclude: '^(pyextra/)|(cereal/)|(rednose/)|(panda/)|(laika/)|(opendbc/)|(laika_repo/)|(selfdrive/car/torque_data/)|(rednose_repo/)/'
```
adeebshihadeh, line 22: why is this necessary?

selfdrive/car/torque_data/override.yaml
```diff
@@ -0,0 +1,29 @@
...
+FORD FOCUS 4TH GEN: [.nan, 1.5, .nan]
+###
+
+# No steering wheel
+COMMA BODY: [.nan, 1000, .nan]
+
+# Totally new car
+KIA EV6 2022: [3.0, 2.5, 0.05]
+
+# Ideal car
```
adeebshihadeh, line 25: dashcam or fallback car is more accurate

selfdrive/car/interfaces.py
```diff
@@ -109,9 +111,22 @@ def get_std_params(candidate, fingerprint):
 
   @staticmethod
   def get_torque_params(candidate, default=float('NaN')):
+    with open(TORQUE_SUBSTITUTE_PATH) as f:
```
adeebshihadeh, line 114: this should also enforce that each car is only defined once

selfdrive/car/torque_data/override.yaml
HaraldSchafer: for the KIA?

selfdrive/car/interfaces.py
HaraldSchafer: willdo

selfdrive/car/torque_data/override.yaml
adeebshihadeh: for mock

HaraldSchafer: Oh I see

PR #24922, opened 2022-06-21
Toyota: remove longitudinal deadzone

sshane:
Difference between the tunes is super small, and I can't imagine this helps in any situations. If anything we don't want deadzone, as response can be laggy sometimes


adeebshihadeh:
Is this still relevant? We shouldn't really merge changes like this without an explicit goal or validation.

sshane:
Had one tester on a TSS-P Prius report minor oscillations, will close until it makes sense to remove, right now there's a lot of other bigger improvements we can still remove. I've been testing it in my TSS2 with no issues so far, so may need to gate on TSS2.

PR #24925, opened 2022-06-21
Mazda: add missing FW version for CX-5 2022

ymgaq:
Added new engine/transmission firmware versions for Mazda CX-5 2022.
The route below was taken in comma 3 dashcam mode, but I have confirmed that openpilot works correctly when the value.py is changed.

Route: [e9e0cfb5010d48b0|2022-06-20--15-21-16]


sshane:
FW versions look good, I requested upload for the logs for the drive, please keep your device on WiFi so they can upload.

ymgaq:
Okay, it looks like the driving log has been uploaded.

PR #24926, opened 2022-06-21
Chrysler: fix steering angle signals

sshane:
Notes on PR: https://github.com/commaai/openpilot/pull/24895


j-vanetten (commented):
Not sure if they are related, but do you need to change the Jeep's `ret.steerRatio`?  It's currently set to 12.7, but I've seen documentation saying it's closer to 16.something.

sshane:
Random Pacifica 2019 Hybrid segment:

Screenshot from 2022-06-24 19-34-20

adeebshihadeh:
@sshane let's make sure to check each platform and also run on a bunch of mkv segments. can you make a checklist of all the platforms?

sshane:
Some cars don't have a higher than expected difference between the angle signals, let me check it out. "diff between new angle" is calculating the new angle from the old one using the learned steering ratio and angle offset.

Screenshot from 2022-06-24 21-29-56

sshane:
Just comparing the change in the signals, it's as expected: https://user-images.githubusercontent.com/25857203/175758596-5fbce32f-8497-42e1-ad2a-d37813b4fdbc.png

sshane:
That one outlier is just from one car, a Pacifica Hybrid 2018, not AWD: dongle: 5348a2069cb6fbb1. I was confused for a second as paramsd thought the steer ratio was 25 and rising, however this is the device with the camera pointed at the sky recently:
Screenshot from 2022-06-24 22-48-36

sshane:
Also, here are the current average steering ratios and angle offsets for all of the Chrysler vehicles: https://imgur.com/a/aOKCYZs. Definitely seems wrong as it is now, they're supposed to be 16.2 for the Pacificas and ~15 for the Jeeps

sshane:
Will post how steering ratio and offset changes after this change on master, after a few days.

sshane:
Looks like the cars are converging to known correct values!

Screenshot from 2022-07-01 15-20-47
Screenshot from 2022-07-01 15-20-27

PR #24941, opened 2022-06-22
compatibility docs: print diff from PR

sshane:
- [x] Obtain CarInfo Python objects from PR base commit
  - Known issue that you can't use two checkout Actions in the same job with Docker containers due to permission errors, so I gave the runner full permissions to delete and re-create the workspace.
  - Saves base car info list to a pickle file in cache
  - This allows us to get extra info for carFingerprint checking (if model name or years changed it would think it was deleted and added)
- [x] ~~Show diff for model + years changing, not removed and added~~ Will do after years are separate
- [x] Show tier changes
- [x] Show car star changes
- [x] Overwrite/delete last comment
- [x] Replace ref with `${{ github.event.pull_request.base.ref }}`
- [x] Delete comment when no diff

See an example of this in action: https://github.com/commaai/openpilot/pull/25139


sshane:
"""Just a working draft"""

# Car Star Changes

|Make|Model|Supported Package|openpilot ACC|Stop and Go|Steer to 0|Steering Torque|Actively Maintained|
|---|---|---|:---:|:---:|:---:|:---:|:---:|
|comma|body|All|<a href="##"></a>|<a href="##"></a>|<a href="##"></a>|<a href="##"></a>|<a href="##"></a>|
|Genesis|G70 2020 ➡️ G70 2020-2021|SCC + LKAS ➡️ SCC|<a href="##"></a>|<a href="##"></a>|<a href="##"></a>|<a href="##"></a>|<a href="##"></a>|
|Hyundai|Palisade 2020-21|All|<a href="##"></a>|<a href="##"></a>|<a href="##"></a>|<a href="##"></a>|<a href="##"></a>|

PR #24949, opened 2022-06-23
enable panda deep sleep

adeebshihadeh:
also needs CI tests https://github.com/commaai/panda/issues/1005

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -193,9 +193,11 @@ Panda *usb_connect(std::string serial="", uint32_t index=0) {
     return nullptr;
   }
 
+  // common panda config
   if (getenv("BOARDD_LOOPBACK")) {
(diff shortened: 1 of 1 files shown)


brittonx:
I don't know if this is related, but about 3 hours ago or so, I updated my C3 running on master-ci and after the reboot I got "No Panda".  I had to finish my drive, get to the hotel, and reflash the panda with:
cd /data/openpilot/panda ; pkill -f boardd ; PYTHONPATH=.. python -c "from panda import Panda; Panda().flash()"
./recover.sh did not work

haram-KONA:
I updated my C3 running on master-ci and after the reboot I got "No Panda". 
After "C3 Panda flashing", the "NO PANDA" message disappears, update master-ci and when rebooting again, 
it becomes "NO PANDA" again and so on. These actions are repeated over and over again.

PR #24950, opened 2022-06-23
Switch to aggressive fan policy once CAN messages are seen

adeebshihadeh:


robbederks:
My car also has a ton of CAN messages offroad, so not sure if this is something we want

sshane:
I think approaching your car with your key can also cause some CAN messages to be sent around. It could also be just the lux sensor which is worse if light around is changing (car turns inside lights on when approaching).

adeebshihadeh:
Closing, but definitely open to ideas here and want to explore this more.

PR #24958, opened 2022-06-24
Speed up YAML parsing with CSafeLoader

sshane:
yaml.FullLoader is quite slow and not safe by default (can execute code), use CSafeLoader (which uses a C parser, CLoader is equivalent in parsing time)

Time to create docs before: 14 seconds
TIme to create docs after: 2 seconds

Loading a single torque tune takes ~80 ms before this change (on PC), now it's ~10 ms


adeebshihadeh:
Looks good, but a better title would be something like "Speed up YAML parsing with CSafeLoader"

adeebshihadeh:
We can also just as easily make this python code instead of yaml.

sshane:
I think it might be easier to export yaml than it is to automatically build valid Python, but up to @HaraldSchafer

HaraldSchafer:
Doesn't matter to me

PR #24961, opened 2022-06-24
UI: wrap all text for translation

sshane:
This wraps all strings in the Qt tr function, which translates the source text using the context (parent class) and any disambiguation text. If no translations are available it's a noop and returns the text passed in. This is to remove some diff from the main multilang PR: https://github.com/commaai/openpilot/pull/21212

For my reference later: doc about contexts: https://wiki.qt.io/QtInternationalization

Test goes through all QLabel and QPushButtons and makes sure that if the widget's text is not empty, it contains our testing string (which means Qt was able to translate it using the `.ts` translation file).


sshane:
Hmm, if we intend to do dynamic translations in the future, perhaps we don't need to rely on any Qt tools for this, including the `tr` function

sshane:
@deanlee any idea why test_translation refuses to run, even manually? I noticed test_sound also isn't running (as far as I know, I don't see any references to it).

deanlee:
I don't know why, but I'll check into it.

sshane:
Thanks!

sshane:
@deanlee ah nevermind about this PR's issues. Do you want to open a PR to run the test_sound test though?

deanlee:
@sshane There is no sound card on the server, it may not be possible run test_sound in selfdrive_tests ( https://github.com/commaai/openpilot/pull/24990 ) .

pd0wm (approved):
LGTM. You might want to run the UI with `TICI=1` env veriable set during the test. The regulatory widget is behind a `Harware::TICI()` check.\

I see the translation file contains line numbers. Does `lupdate` correct those when the source is changed? Should we add a pre-commit step to make sure we don't forget this?

I assumed to generate the translations it would need to start the actual ui. But seems like it just looks at source code. Is that correct?

sshane:
I'm almost 100% sure those are just for QT Linguist to show the source code to the translator. You can specify a context without a line number and it translates.

sshane:
Sounds like a plan. I don't think it's necessary for this PR, as this is just to test all strings are wrapped, and thus the file is cleared every time. I'll add it to the main PR/the PR that adds a language.

PR #24964, opened 2022-06-25
Toyota: fix BSM detection

sshane:
Fixes failing to find the BSM sometimes for Toyota.

Reverts https://github.com/commaai/openpilot/pull/22803, will try to find another signal, then merge if there's nothing else. We can revert this in the future if ever want to be able to change CarParams after initialization.


pd0wm:
Let's not make any assumptions and use 1 second for all brands.

PR #24971, opened 2022-06-26
move replay from selfdrive/ui/replay to tools/replay

deanlee:


pd0wm:
Can we do this without `no-duplicate-environment`? Can we build the relevant files into a library?

PR #24975, opened 2022-06-27
Log SOM power draw

robbederks:


system/hardware/pc/hardware.py
```diff
@@ -55,6 +55,9 @@ def get_usb_present(self):
 
   def get_current_power_draw(self):
     return 0
+  
+  def get_som_power_draw(self):
```
pd0wm, line 59: You can just let the base implementation return 0. No need to define twice. We need to clean up some of the other ones too.

pd0wm: Will clean this up in a separate PR

PR #24978, opened 2022-06-27
Update rednose: use EKF_sym_pyx

gijskoning:


selfdrive/locationd/models/gnss_kf.py
```diff
@@ -115,12 +120,13 @@ def generate_code(generated_dir):
 
     gen_code(generated_dir, name, f_sym, dt, state_sym, obs_eqs, dim_state, dim_state, maha_test_kinds=maha_test_kinds)
 
-  def __init__(self, generated_dir):
+  def __init__(self, generated_dir, cython=False):
```
pd0wm, line 123: Mesh3D uses some more advanced features of the filter, so can't use the Cython version

pd0wm:
Needs AGNOS updated on the CI devices to pass Jenkins

PR #24979, opened 2022-06-27
Honda Civic 2022: remove LKAS fault reinitialization

sshane:
According to Marco on Discord, his LKAS warning light does not show up with a new official harness (correlated with panda/CAN dropout which is also gone with new harness, think panda wasn't sending our messages causing a fault).

Will get all of our 22 Civic testers with new harnesses to try this out

Before:
Screenshot from 2022-06-27 11-56-49

After:
Screenshot from 2022-06-27 11-56-41


sshane:
Tested by First, Marco, and Kevin with no issues so far on harness without the panda dropout issue.

PR #24981, opened 2022-06-28
Power Monitoring test: fix exceptions

sshane:
This was causing a high rate of printed exceptions in the unit test job in CI (NoneType * int). Looks all other functions had this patch changed from None except this case, and 0 should be equivalent to None here.

Relevant PR: https://github.com/commaai/openpilot/pull/24055

### DIFF
--- a/selfdrive/thermald/tests/test_power_monitoring.py
+++ b/selfdrive/thermald/tests/test_power_monitoring.py
@@ -119,7 +119,7 @@ class TestPowerMonitoring(unittest.TestCase):
   @parameterized.expand(ALL_PANDA_TYPES)
   def test_max_time_offroad(self, hw_type):
     MOCKED_MAX_OFFROAD_TIME = 3600
-    POWER_DRAW = 4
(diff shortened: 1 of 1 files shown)


selfdrive/thermald/tests/test_power_monitoring.py
```diff
@@ -119,7 +119,8 @@ def test_car_battery_integration_lower_limit(self, hw_type):
   @parameterized.expand(ALL_PANDA_TYPES)
   def test_max_time_offroad(self, hw_type):
     MOCKED_MAX_OFFROAD_TIME = 3600
-    with pm_patch("MAX_TIME_OFFROAD_S", MOCKED_MAX_OFFROAD_TIME, constant=True), pm_patch("HARDWARE.get_current_power_draw", None):
+    POWER_DRAW = 4
```
sshane, line 122: @robbederks should this be 0 or 4?

robbederks (approved):
Lgtm!

PR #24987, opened 2022-06-28
CI: fix modeld unit tests not being ran

sshane:


sshane:
Looks like you can't run opencl on GH Actions

PR #24988, opened 2022-06-29
Support asymmetric minimum steering speeds

### LINKED ISSUE #1168: Chrysler: allow for lower min steer speed on the way down
https://github.com/commaai/openpilot/blob/b941b39c56db2d9d7f756cb0949ef6a7cf61774a/selfdrive/car/chrysler/interface.py#L35

controlsd should generically support asymmetric min steer speeds, one for engage and another for disengaging.

sshane:


sshane:
Does this cause problems in any cases? Or does the car tell you to engage at a minimum speed the first attempt?

selfdrive/car/body/interface.py
```diff
@@ -16,7 +16,8 @@ def get_params(candidate, fingerprint=None, car_fw=None, disable_radar=False):
     ret.carName = "body"
     ret.safetyConfigs = [get_safety_config(car.CarParams.SafetyModel.body)]
 
-    ret.minSteerSpeed = -math.inf
+    ret.minSteerEnableSpeed = -math.inf
+    ret.minSteerDisableSpeed = -math.inf
```
sshane, line 20: We can't infer disable from enable (if car enables and disables at one speed) if we only want to set one, we couldn't differentiate between an asymmetric steer speed car and a symmetric one.

selfdrive/car/chrysler/carcontroller.py
```diff
@@ -23,25 +23,13 @@ def __init__(self, dbc_name, CP, VM):
...
-      if CS.out.vEgo < (self.CP.minSteerSpeed - 3.0):
-        lkas_control_bit = False
-    elif self.CP.carFingerprint in RAM_CARS:
-      if CS.out.vEgo < (self.CP.minSteerSpeed - 0.5):
-        lkas_control_bit = False
-
     # EPS faults if LKAS re-enables too quickly
-    lkas_control_bit = lkas_control_bit and (self.frame - self.last_lkas_falling_edge > 200)
-    lkas_active = CC.latActive and self.lkas_control_bit_prev
+    lkas_active = CC.latActive and (self.frame - self.last_lkas_falling_edge > 200)
```
sshane, line 27: @adeebshihadeh safe to get rid of this if latActive considers lower min steer speed? The difference here is it won't send the bit when disabled any more. If you don't know I can try to find a tester

twilsonco:
2017 volt drivers can either set the minimum speed to 0 and ACC will fault if they initially engage too low, or they leave it set at 18mph to avoid the potential for faults and take the convenience hit. Other gm cars might be in a similar boat but not sure.

sshane:
Not sure of the state of this any more. We should support all the needed behavior per car-port (we don't yet), then make it common once we understand the full space of possible behaviors.

PR #24989, opened 2022-06-29
Toyota: Add missing 2021 RAV4 TSS2 esp FW

sshane:
From dongle b97486b552c1f5e3


sshane:
For reference, this user's car responded only to the `[TESTER_PRESENT_REQUEST, DEFAULT_DIAGNOSTIC_REQUEST, EXTENDED_DIAGNOSTIC_REQUEST, UDS_VERSION_REQUEST]` request.

PR #24993, opened 2022-06-29
camerad: reduce cpu usage

deanlee:


pd0wm:
How much does this save?

deanlee:
Haven't tested on device yet. maybe 4～6%(not sure),  there is no reason to do "busy" waiting on try_pop IMO.

pd0wm:
Seems to save about 1-2% of a core

PR #24995, opened 2022-06-29
Add laikadOffline subtest to process replay. 

gijskoning:
To allow for subtests of the same process the reference file is saved as proc_name+subtest_name. Such that each subtest is correctly compared.
Also adds the functionality to add custom os.environ variables per process. All variables form all processes are removed at the start


selfdrive/test/process_replay/process_replay.py
```diff
@@ -418,6 +440,8 @@ def python_replay_process(cfg, lr, fingerprint=None):
     CP = [m for m in lr if m.which() == 'carParams'][0].carParams
     setup_env(CP=CP)
 
+  os.environ.update(cfg.environ)
```
pd0wm, line 443: also add for C++ replay

PR #25014, opened 2022-07-01
UI: fix unable to scroll on 'Regulatory' page

### LINKED ISSUE #24808: ui: unable to scroll on 'Regulatory' page
### Describe the bug

When selecting the VIEW button on the Regulatory option in Settings, user is unable to scroll down to see additional information. This has been a persistent bug for a few versions of OP on the C3 now.

No route

### openpilot version

0.8.15 master

### Additional info

_No response_

deanlee:
grabGesture doesn't always work in dialog's exec() in QT 5.

it seems that the only reliable way for scrolling in QDialog is to use  show() instead of exec()

resolve: https://github.com/commaai/openpilot/issues/24808#issuecomment-1172288970

### DIFF
--- a/selfdrive/car/gm/values.py
+++ b/selfdrive/car/gm/values.py
@@ -1,4 +1,5 @@
-from dataclasses import dataclass
+from collections import defaultdict
+from dataclasses import dataclass, field
 from enum import Enum
 from typing import Dict, List, Union
 
@@ -9,9 +10,9 @@ Ecu = car.CarParams.Ecu
 
 
 class CarControllerParams:
-  STEER_MAX = 300  # Safety limit, not LKA max. Trucks use 600.
-  STEER_STEP = 2  # control frames per command
-  STEER_DELTA_UP = 7
+  STEER_MAX = 300  # GM limit is 3Nm. Used by carcontroller to generate LKA output
+  STEER_STEP = 2  # Control frames per command (50hz)
+  STEER_DELTA_UP = 7  # Delta rates require review due to observed EPS weakness
   STEER_DELTA_DOWN = 17
   MIN_STEER_SPEED = 3.  # m/s
   STEER_DRIVER_ALLOWANCE = 50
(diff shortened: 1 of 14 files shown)


pd0wm:
After this change the main event loop wouldn't be blocked anymore to wait for the result. Is there any way we can wait for the widget to be closed and return the proper status (OK or cancel).

pd0wm:
You can probably just call `QCoreApplication::processEvents()` in a loop until the dialog is done.

deanlee:
> You can probably just call `QCoreApplication::processEvents()` in a loop until the dialog is done.

This will make the bug reproduce like QEventLoop::exec in QDialog::exec()

sshane:
Scrolling works just fine if you remove enough elements to make the Device page non-scrollable. It seems like if your page with the button can scroll like it does now, these are the sequence of events:

```
obj: "MainWindowClassWindow", evt: QEvent::MouseButtonPress
"qt_scrollarea_viewport" QEvent::ScrollPrepare
obj: "MainWindowClassWindow", evt: QEvent::MouseButtonRelease
"" QEvent::MouseButtonPress
"" QEvent::MouseButtonRelease
```

However if the page doesn't scroll (Regulatory scrolling works fine):
obj: "MainWindowClassWindow", evt: QEvent::MouseButtonPress
"qt_scrollarea_viewport" QEvent::ScrollPrepare
"" QEvent::MouseButtonPress
obj: "MainWindowClassWindow", evt: QEvent::MouseButtonRelease
"" QEvent::MouseButtonPress
"" QEvent::MouseButtonRelease
```

```

sshane:
We'll also need to block the event loop, as the new internationalization PR relies getting the result from a scroll view in a dialog box.

deanlee:
> If you click a button on the Software page you can also see it immediately shows pressed style state, while Device takes half a second to enter pressed state. 

I just realized this. It may be this problem that caused the gesture recognizer's error

deanlee:
It seems that setting MousePressEventDelay to a tiny value can also solve this issue https://bugreports.qt.io/browse/QTBUG-67210?focusedCommentId=464719&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-464719 :

deanlee:
lgtm, haven't found any side effects so far

sshane:
Looks good! However I was testing out on the Network page, if you scroll around repeatedly it's somehow easier to accidentally press networks, then it seems like you need to press twice to register a touch to cancel out.

deanlee:
It seems that ElidedLabel should emit clicked  like QPushbutton to work correctly with QScroller, instead of simply detecting mouseReleaseEvent

https://code.woboq.org/qt5/qtbase/src/widgets/widgets/qabstractbutton.cpp.html#_ZN15QAbstractButton14mouseMoveEventEP11QMouseEvent

Achilles308:
Horary for it being fixed!

I don't feel so bad for wasting a few hours trying to figure this issue out to try to get a commit in... turns out it was a decently hard one!

PR #25017, opened 2022-07-01
long planner: always run when not openpilot longitudinal control

sshane:
- Logs reads in radarState if not longitudinal control and/or radarOffcan
  - Checked UI and we now check if openpilotLongitudinalControl to show leads, comment in radard is out of date
- No longitudinal planner reset if not longitudinal control for accurate plan

Pre-req for https://github.com/commaai/openpilot/pull/24873


selfdrive/controls/lib/longitudinal_planner.py
```diff
@@ -69,8 +69,9 @@ def update(self, sm):
     long_control_state = sm['controlsState'].longControlState
     force_slow_decel = sm['controlsState'].forceDecel
 
-    # Reset current state when not engaged, or user is controlling the speed
-    reset_state = long_control_state == LongCtrlState.off
+    # Reset current state when not engaged or user is controlling the speed
+    # No reset if car is not using openpilot longitudinal (for accurate planner predictions)
```
sshane, line 73:
```suggestion
    # No reset if car is not using openpilot longitudinal (for accurate planner logging)
```

HaraldSchafer:
Can we also reset when not engaged and stock ACC? Then it seems more consistent.

PR #25020, opened 2022-07-02
Laneless toggle back

haraschax:


selfdrive/ui/qt/offroad/settings.cc
```diff
@@ -57,6 +57,12 @@ TogglesPanel::TogglesPanel(SettingsWindow *parent) : ListWidget(parent) {
       tr("Upload data from the driver facing camera and help improve the driver monitoring algorithm."),
       "../assets/offroad/icon_monitoring.png",
     },
+    {
+      "EndToEndToggle",
+      tr("\U0001f96c Disable use of lanelines (Alpha) \U0001f96c"),
```
sshane, line 62:
```suggestion
      tr("🥬 Disable use of lanelines (Alpha) 🥬"),
```

sshane: Both should be identical

```python
>>> print("\U0001f96c")
🥬
>>> print("\U0001f96c" == "🥬")
True
```

The file is UTF-8 encoded, and so is the ts file, so I'm not sure why it tries to read the encoding but fails and removes the first 1, converting it to `塞`.

PR #25023, opened 2022-07-02
DOC : Add video link for Hyundai Kona Hybrid 2020

haram-KONA:


sshane:
Thanks for the PR! I wonder if a slightly longer video that shows more driving would be a better choice?

haram-KONA:
Ah~ Then I will delete this PR and re-upload the PR with longer content. thank you

PR #25024, opened 2022-07-03
Toyota: go into standstill if interceptor detected

cydia2020:
Make the car go into standstill if comma pedal is detected while openpilot is running on a `NO_STOP_TIMER_CAR`.

Currently, openpilot doesn't request standstill on some Toyotas to allow stop-and-go, this is a problem on hybrid cars as the pcm keeps sending power to MG2 even if the vehicle is stopped. This PR aims to partially address that by allowing the car to go into standstill when an interceptor is detected.

The user will be able to choose whether they would like an interceptor or not.

Thanks @nelsonjchen for the suggestion: https://discordapp.com/channels/469524606043160576/524327905937850394/990697643976044655


selfdrive/car/toyota/carcontroller.py
```diff
@@ -66,7 +66,7 @@ def update(self, CC, CS):
       pcm_cancel_cmd = 1
 
     # on entering standstill, send standstill request
-    if CS.out.standstill and not self.last_standstill and (self.CP.carFingerprint not in NO_STOP_TIMER_CAR and not self.CP.enableGasInterceptor):
+    if CS.out.standstill and not self.last_standstill and (self.CP.carFingerprint not in NO_STOP_TIMER_CAR or self.CP.enableGasInterceptor):
```
nelsonjchen, line 69: I was wondering about that...

sshane:
Can we get a few testers for this with provided routes with rlogs? Think it's good to merge, just want to see it working for a few users without issues

sshane:
Anyone want to test this and provide a route?

PR #25027, opened 2022-07-04
hyundai: fix FCA11 checksum and counter

gregjhogan:
When using openpilot longitudinal control (at least for 2020 sonata/palisade) before these changes TCS13.CF_VSM_Avail == 3 which means there is an FCA error (preventing the possibility of openpilot using AEB)
image

Looking at the firmware for my palisade I can see there is code for parsing both ways, so the other parsing is probably for some older vehicles or vehicles with a different ECU sending FCA11.  It seems like we would have to split the DBC to support both so I figure for now we should switch to what the 2020 sonata/palisade use.  And after these changes I am able to command AEB on my palisade.

Before merging:
- [x] test on sonata (thanks @sunnyhaibin)
- [x] decide if we should merge this only for specific platforms - some cars probably use a different message layout, but I believe they are broken on master now, too (@adeebshihadeh can you make this decision, adjust this PR accordingly and merge https://github.com/commaai/opendbc/pull/656?) - decided to leave old vehicles broken and add comment about it

### DIFF
--- a/selfdrive/car/hyundai/hyundaican.py
+++ b/selfdrive/car/hyundai/hyundaican.py
@@ -132,6 +132,8 @@ def create_acc_commands(packer, enabled, accel, upper_jerk, idx, lead_visible, s
   }
   commands.append(packer.make_can_msg("SCC14", 0, scc14_values))
 
+  # note that some vehicles most likely have an alternate checksum/counter definition
+  # https://github.com/commaai/opendbc/commit/9ddcdb22c4929baf310295e832668e6e7fcfa602
   fca11_values = {
     "CR_FCA_Alive": idx % 0xF,
     "PAINT1_Status": 1,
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Any reason this isn't merged?

gregjhogan:
* I don't understand the weird opendbc diff
* someone needs to decide if we care about cars using the old checksum
* maybe test on sonata that TCS13.CF_VSM_Avail stays zero after this change

awesomebytes:
In case it matters, in my Kona EV 2022, what was `Supplemental_Counter` is always 0 and the `CR_FCA_Alive` is definitely counting on every message. With these changes I don't think it will make sense for this vehicle in particular.

sunnyhaibin:
Glad to report that this fix is also a success on the 2021 Hyundai Sonata N Line:
- Before: `36e10531feea61a4|2022-08-30--09-46-12`
- After: `36e10531feea61a4|2022-08-30--19-44-47`

Before:
image


After:
image

PR #25034, opened 2022-07-05
Subaru: XV is supported

martinl:
This PR adds Subaru XV as supported model in cars list. Subaru Crosstrek is called XV on EU and AU markets.


selfdrive/car/subaru/values.py
```diff
@@ -45,10 +45,12 @@ class SubaruCarInfo(CarInfo):
   CAR.IMPREZA: [
     SubaruCarInfo("Subaru Impreza 2017-19"),
     SubaruCarInfo("Subaru Crosstrek 2018-19", video_link="https://youtu.be/Agww7oE1k-s?t=26"),
+    SubaruCarInfo("Subaru XV 2018-19", video_link="https://youtu.be/Agww7oE1k-s?t=26"),
```
adeebshihadeh, line 48: should we make all the support model years the same?

sshane: Looks like we support the 17 Crosstrek from some Discord conversations

martinl: I think Impreza went global platform a year earlier, so 2017 for Impreza and 2018 for Crosstrek/XV should be correct https://www.sae.org/news/2017/08/2018-subaru-crosstrek-moves-to-new-stiffer-global-platform

sshane: Hmm here it actually notes the 2016 Impreza and 2017 Crosstrek use the global platform: https://en.wikipedia.org/wiki/Subaru_Global_Platform

martinl: yes, new models usually launch in Japan first, so global platform 2016 Imprezas and 2017 Crosstreks are probably all JDM

martinl: For example, if you look at UDM 2016 vs 2017 Impreza EPS in the parts store then 2016 has separate EPS and 2017 has global platform integrated assembly

https://parts.subaru.com/p/Subaru_2016_Impreza-20L-CVT-4WD-Sport-Premium-Wagon-wEyeSight/Power-Steering-Control-Module/69986136/34710FJ212.html

https://parts.subaru.com/p/Subaru_2017_Impreza-20L-CVT-4WD-Sport-Premium-Wagon-wEyeSight/Rack-and-Pinion/94383198/34110FL00E.html

PR #25035, opened 2022-07-05
Update path for github workflow hardware unit tests

martinl:
**Description** This PR fixes hardware unit tests path in github workflow and changes the included path in release files.

**Verification** All github workflow tests pass

### DIFF
--- a/release/files_common
+++ b/release/files_common
@@ -187,7 +187,7 @@ selfdrive/controls/lib/longitudinal_mpc_lib/.gitignore
 selfdrive/controls/lib/lateral_mpc_lib/*
 selfdrive/controls/lib/longitudinal_mpc_lib/*
 
-selfdrive/hardware
+system/hardware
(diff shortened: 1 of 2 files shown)


release/files_common
```diff
@@ -187,7 +187,7 @@ selfdrive/controls/lib/longitudinal_mpc_lib/.gitignore
 selfdrive/controls/lib/lateral_mpc_lib/*
 selfdrive/controls/lib/longitudinal_mpc_lib/*
 
-selfdrive/hardware
+system/hardware
```
pd0wm, line 190: `selfdrive/hardware` is a symlink we need to keep around. It's used by the updater, so it needs to stay around or a bit to stay backwards compatible with older versions updating to the latest version.

adeebshihadeh, line 190:
```suggestion
selfdrive/hardware
```

PR #25039, opened 2022-07-05
Laikad: improve convergence

gijskoning:
- To fix converging to the right position: If residual between kalman filter and measurements is too high -> move covariance towards initial matrix

- Only use kalman filter for corrections:
Fixes the issue when calculating a pos fix and some measurements have crazy pseudoranges
Reduces complexity
But problem occurs at initalization, filter has a completely wrong position and then the measurements cannot be corrected.
This is fixed using uncorrected measurements if no corrected are available.


gijskoning:
Actually, maybe we want to at least keep updating the pos fix for debugging purposes for the coming release

selfdrive/locationd/laikad.py
```diff
@@ -112,18 +116,27 @@ def process_ublox_msg(self, ublox_msg, ublox_mono_time: int, block=False):
       new_meas = [m for m in new_meas if 1e7 < m.observables['C1C'] < 3e7]
 
       processed_measurements = process_measurements(new_meas, self.astro_dog)
+      self.update_pos_fix(t, processed_measurements)
+      est_pos = self.gnss_kf.x[GStates.ECEF_POS]
 
-      est_pos = self.get_est_pos(t, processed_measurements)
+      allow_incomplete_delay = any(self.gnss_kf.P[GStates.ECEF_POS].diagonal() > 1e3)
```
pd0wm, line 122: Need abs?

gijskoning: Before I wasnt using the diagonal and then needed abs. But I think in all cases it can be removed now.

gijskoning:
https://github.com/commaai/openpilot/blob/159ed0f1e1d3bd174fd6390efab26e176dc41da9/selfdrive/locationd/laikad.py#L191
Should also be removed. The std should be set to the initial value

PR #25049, opened 2022-07-06
V4LEncoder: use std::vector to bind the VisionIpcBufExtra to index

deanlee:
I think it's more safe and readable than pop from a SafeQueue.

`std::vector extras` is thread safe  if there is no access from different threads to the same element.

### DIFF
--- a/selfdrive/loggerd/encoder/encoder.cc
+++ b/selfdrive/loggerd/encoder/encoder.cc
@@ -11,7 +11,7 @@ void VideoEncoder::publisher_init() {
   pm.reset(new PubMaster({service_name}));
 }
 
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
It's not clear why an std::vector is preferable here.

deanlee:
the std::vector may be more robust than the safeQueue, it can guarantees that the  `VisionIpcBufExtra &extra` is exactly the same as the dequeued  frame.
the safeQueue requires the order of enqueue and dequeue matches strictly without any errors during processing.

deanlee:
this pr is buggy and has too many conflicts with master(encoderd has been removed from selfdrive/ to system/). close for now.

PR #25056, opened 2022-07-06
Store ECU types with brand

sshane:
Think this was a bug, `a` would never be in ecu_types, so we'd be overriding any common addr/subaddr combos from different brands.


sshane:
Done in https://github.com/commaai/openpilot/pull/23311

PR #25057, opened 2022-07-07
Add missing 2019 RAV4 Hybrid engine FW version

sshane:


sshane:
Verified with test_car_model and verify_car_port

PR #25059, opened 2022-07-07
Hyundai Kona Hybrid: use torque controller

haram-KONA:


haram-KONA:
In "https://github.com/commaai/openpilot/pull/24921", both kona and kona-hybrid use the values ​​of kona-electric.
This makes for very good driving conditions. so the vehicle model was not separated.

PR #25060, opened 2022-07-07
Laikad: Use clocks for faster fetching orbits

gijskoning:
Using the clocks walltime to receive the time from messages for fetching orbits. It's only used to download orbits at startup faster. 
When ublox messages are received it will ignore the clocks messages.

The last_fetch_time_t is now set to the time of request instead of the time of receiving. This fixes the issue where a wrong walltime could stop fetching orbits for 1 minute.


pd0wm:
Don't forget to add clocks to process replay

PR #25064, opened 2022-07-07
Add Chinese (simplified) translations

sshane:
- Add Chinese (Traditional) language translation file
- Tests:
  - Add test for outdated compiled QM file, seems to be deterministic for the same input .ts file
  - More reliable test by creating a temp directory so as not to change the translation files in the git repo after running test
  - Fix creating new translation files from scratch and not updating the current files


selfdrive/ui/tests/test_translations.py
```diff
@@ -1,49 +1,65 @@
...
+
+  @classmethod
+  def tearDownClass(cls):
+    shutil.rmtree(TMP_TRANSLATIONS_DIR, ignore_errors=True)
+
+  @staticmethod
+  def _read_translation_file(path, file, file_ext):
+    tr_file = os.path.join(path, f"{file}.{file_ext}")
+    with open(tr_file, "rb") as f:
+      # fix relative path depth
```
sshane, line 29:
```suggestion
      # remove relative path depth so comparisons work
```

sshane:
Title was wrong, this is simplified!

sshane:
@gijskoning I don't think XML vs. JSON makes it any more difficult to edit the file, however it's recommended to use QT Linguist, and it's required as that's the format QT uses. Linguist provides a nice UI, with various warnings like leading and trailing punctuation and translation suggestions. `lupdate` (run `./update_translations.py`) also automatically updates the source code line numbers when they're moved, which is only for Linguist

PR #25067, opened 2022-07-07
UI: only show cell settings on non-prime connection

adeebshihadeh:
- [x] merge #25902
- [x] refactor `UIState` to emit signal on prime type change
- [x] hide gsm toggles in `AdvancedNetworking` on `UIState#primeTypeChanged`
- [x] also fixed map settings not being visible after upgrading to prime

### DIFF
--- a/selfdrive/ui/qt/offroad/networking.cc
+++ b/selfdrive/ui/qt/offroad/networking.cc
@@ -8,9 +8,11 @@
 #include <QPainter>
 #include <QScrollBar>
 
+#include "selfdrive/ui/ui.h"
 #include "selfdrive/ui/qt/util.h"
 #include "selfdrive/ui/qt/qt_window.h"
 #include "selfdrive/ui/qt/widgets/controls.h"
+#include "selfdrive/ui/qt/widgets/prime.h"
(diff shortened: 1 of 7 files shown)


selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -148,17 +150,22 @@ AdvancedNetworking::AdvancedNetworking(QWidget* parent, WifiManager* wifi): QWid
   list->addItem(new SshToggle());
   list->addItem(new SshControl());
 
+  int prime_type = uiState()->prime_type;
+  bool gsmVisible = prime_type == PrimeType::NONE || prime_type == PrimeType::LITE;
```
adeebshihadeh, line 154: don't need to do this in the constructor, and you don't want to duplicate the logic anyway. just hide them by default. primeTypeChanged should be emitted shortly after this anyway

incognitojam:
Set them to default to visible since you might need to change the APN to get a network connection. Confirmed with local chffr backend that the UI updates immediately (5 seconds) after prime state changes while offroad, for both the networking settings and nav settings.

PR #25075, opened 2022-07-08
Improve Chinese (Simplified) translations

### LINKED ISSUE #25072: Improve Chinese translations
We'd like to improve our simplified translations and add traditional.

Translation file: https://github.com/commaai/openpilot/blob/master/selfdrive/ui/translations/main_zh.ts

Install QT Linguist with `sudo apt install qtbase5-dev qtchooser qt5-default qttools5-dev-tools` and run `linguist selfdrive/ui/translations/main_zh.ts` to edit the file

hellokitty-666:


sshane:
Do these look good @ZwX1616?

PR #25098, opened 2022-07-09
Display the current language in MultiOptionDialog

deanlee:


sshane:
Nice! Can we just make it so if you select the current language after selecting another it disables the button again?

sshane:
Thanks for the PR!

PR #25101, opened 2022-07-09
UI: changing languages on the fly

### LINKED ISSUE #25092: Multilanguage improvements
- [ ] We would like to dynamically translate without restarting UI
- [ ] Translate alerts (dynamic text requires injecting our own translations with custom contexts into the TS files)
- [x] For the [translations badges](https://github.com/commaai/openpilot/tree/badges), auto-concatenate them together

deanlee:


selfdrive/ui/qt/maps/map_settings.cc
```diff
@@ -152,6 +152,13 @@ MapPanel::MapPanel(QWidget* parent) : QWidget(parent) {
       });
     }
   }
+
+  translateUi();
+}
+
+void MapPanel::translateUi() {
```
sshane, line 159: I think Qt Designer names this function `retranslateUi`. Not a huge deal, but might want to fit in with that convention

deanlee: ok,I'll change it to retranslateUi later

deanlee:
There are still some small problems in `MapPanel` and `Networking`, such as the scanning labels in `Networking`. I think it needs to be refactored to work better with retranslateUi in another PR(e.g. https://github.com/commaai/openpilot/pull/23506)

sshane:
I think we also want to try to do this dynamically, at least largely. You can try to loop through all MainWindow child widgets and then set the text with a widget ID and context. But anything dynamic will likely need a custom `retranslateUi` function like you have here. For everything static though, that would reduce a lot of code.

PR #25102, opened 2022-07-09
mapbox: multilanguage support

sshane:


pd0wm:
We should also be able to get nav instructions in the correct language: https://docs.mapbox.com/api/navigation/directions/#instructions-languages

pd0wm:
Seems to do something! 2022-07-12-174110_1920x1080_scrot

sshane:
Nice!! I'll make sure our language codes match mapbox. We can also remove main,

sshane:
@adeebshihadeh I've verified all of our language codes work and mapbox returns localized strings for directions/instructions, however I haven't been able to test the visual aspect of this PR, so I can't guarantee a lot of strings won't go off the screen. If we think that's fine, this is good to merge

adeebshihadeh:
Hmm, it should just wrap around, so I think this is fine.

PR #25104, opened 2022-07-09
UI: translate alerts

### LINKED ISSUE #25092: Multilanguage improvements

deanlee:
resolve https://github.com/commaai/openpilot/issues/25092


sshane:
We discussed this and we want to try to dynamically inject all the alert strings from events.py into the XML files, then compile

adeebshihadeh:
Closing since as @sshane said, we want to do this dynamically. I think the first step is to extract the static alerts into the translation files; we can do the alert callbacks later.

PR #25105, opened 2022-07-09
Improve Korean translations

crwusiz:
update_translations --release

qm


selfdrive/ui/qt/maps/map_settings.cc
```diff
@@ -104,7 +104,7 @@ MapPanel::MapPanel(QWidget* parent) : QWidget(parent) {
     screenshot->setPixmap(pm.scaledToWidth(1080, Qt::SmoothTransformation));
     no_prime_layout->addWidget(screenshot, 0, Qt::AlignHCenter);
 
-    QLabel *signup = new QLabel(tr("Get turn-by-turn directions displayed and more with a comma \nprime subscription. Sign up now: https://connect.comma.ai"));
+    QLabel *signup = new QLabel(tr("Get turn-by-turn directions displayed and more with a comma\nprime subscription. Sign up now: https://connect.comma.ai"));
```
sshane, line 107: Let's keep this the same so we don't affect other languages

crwusiz: ok..

sshane:
Let's make sure not to touch other languages when the PR is to improve just one, this had some of @cydia2020's changes in his PR: https://github.com/commaai/openpilot/pull/25108

crwusiz:
> Looks like you had a bad merge? I reverted the changes to the other languages

update_translations --release
why your make qm file size different..?

How about improving qm file to update automatically in workflow?

sshane:
Yeah not sure why some contributors are ending up with slightly different files, I'll look into it. I'm using 20.04, lrelease version 5.12.8. Make sure to run ubuntu_setup.sh so it installs the right version for now.

crwusiz:
> Yeah not sure why some contributors are ending up with slightly different files, I'll look into it. I'm using 20.04, lrelease version 5.12.8. Make sure to run ubuntu_setup.sh so it installs the right version for now.

i checked compare file

sshane:
Thanks for the PR!

PR #25108, opened 2022-07-10
Multilang: Improve Japanese translations

cydia2020:
Improving Japanese translation, changed some full-widthed kana to half-width to save some space, fixed some expressions.

Need someone to verify this since I am not native.

~~How do you update the qm files?~~


selfdrive/ui/translations/main_ja.ts
```diff
@@ -408,7 +408,7 @@ location set</source>
         <location filename="../qt/offroad/networking.cc" line="87"/>
         <location filename="../qt/offroad/networking.cc" line="97"/>
         <source>for &quot;</source>
-        <translation>のため &quot;</translation>
+        <translation>&quot; のため</translation>
```
cydia2020, line 411: We should make this an entire sentence in networking.cc, instead of translating each part of the sentence separately, the translation really breaks down here.

selfdrive/ui/translations/main_ja.ts
```diff
@@ -556,17 +556,17 @@ location set</source>
...
         <location filename="../qt/text.cc" line="41"/>
         <source>Exit</source>
-        <translation>退出</translation>
+        <translation>閉じる</translation>
     </message>
     <message>
         <location filename="../qt/util.cc" line="18"/>
         <source>dashcam</source>
-        <translation>ダッシュカム</translation>
+        <translation>ﾄﾞﾗｲﾌﾞﾚｺｰﾀﾞｰ</translation>
```
cydia2020, line 564: "drive recorder" seems to be a more commonly used phrase for dashcam, open to reverting.

selfdrive/ui/translations/main_ja.ts
```diff
@@ -556,17 +556,17 @@ location set</source>
...
     </message>
     <message>
         <location filename="../qt/util.cc" line="18"/>
         <source>openpilot</source>
-        <translation>オープンパイロット</translation>
+        <translation>openpilot</translation>
```
cydia2020, line 569: Shouldn't translate this kind of nouns.

selfdrive/ui/translations/main_ja.ts
sshane: Yeah, you can do things like automatic plural handling, I'll refactor that

sshane:
You can run `./update_translations.py --release`

sshane:
I think the main reason I left it out was because some of the text Google Translate produced was too long for our design. If you can use half-width Katakana to fix that, great then. Let's check how many users we have in Japan before adding to release though

selfdrive/ui/translations/main_ja.ts
```diff
@@ -908,68 +981,68 @@ location set</source>
...
-        <location filename="../qt/offroad/settings.cc" line="230"/>
+        <location filename="../qt/offroad/settings.cc" line="245"/>
         <source>CHECKING</source>
         <translation>確認中</translation>
     </message>
     <message>
-        <location filename="../qt/offroad/settings.cc" line="237"/>
+        <location filename="../qt/offroad/settings.cc" line="252"/>
         <source>Uninstall </source>
         <translation>アンインストール </translation>
```
cydia2020, line 1026: @sshane is it possible to make `getBrand()` an argument? Uninstall "brand" might not be in the same word order in other languages.

For Japanese it's usually "brand" + object particle + uninstall

sshane: Added to other PR

cydia2020:
Don't know why it says qm out of date. Even though I updated the qm file as instructed.

sshane:
Likely the qt version used to compile it is different than the one in the Ubuntu environment. `lrelease version 5.12.8` should work

sshane:
Going to merge without adding to release for now, will have a discussion which languages we want to fully support in the future. Thanks for the PR!

PR #25127, opened 2022-07-12
Replay: Fix --upload-only

gijskoning:
Broke after recent change


adeebshihadeh:
This now ignores errors that are strings [like this](https://github.com/commaai/openpilot/blob/105afee4a21ad008fa741f19c5a62a6d52fdd773/selfdrive/test/process_replay/test_processes.py#L100), and won't show any results for those processes.

gijskoning:
Ah my bad. Should have read your pr better

PR #25147, opened 2022-07-13
Mazda: Support CX-9 2022

Jafaral:
Same as CX-9 2021, no change in LKAS system.

dongle-id: 8c6e0e30decb68f7


adeebshihadeh:
Nice, can you update the year in the CarInfo section too?

Jafaral:
> Nice, can you update the year in the CarInfo section too?

I just  pushed an update but it got rejected, looks like @sshane  beat  me to it!

Jafaral:
BTW,  car info  doesn't look complete. I will submit a PR to fix it.

sshane:
Fixing that comment bug now, it really shouldn't be trying to write it for contributor PRs

sshane:
Thanks for the PR!

PR #25149, opened 2022-07-13
Car docs diff bot: run on fork PRs

sshane:
GitHub Actions automatically skips runs if any Actions files are changed for these for `pull_request_target` targets. Under these events you get all normal tokens, as if a maintainer opened the PR


sshane:
Will need to build in a workflow with read permissions, then pass output to comment to a pull_request_target workflow

PR #25154, opened 2022-07-13
disable_ecu.py: remove duplicate file

sshane:


sshane:
@gregjhogan is it okay if we get more cloudlogs from users debugging with this script? Is there an environment variable we can use to disable logging but still print?

adeebshihadeh:
cloudlogs aren't logged unless `logmessaged` is running.

PR #25159, opened 2022-07-13
Chrysler: never drop control bit on older models

adeebshihadeh:


sshane:
Do you know why we did this? Was the minSteerSpeed too low, and so it caused faults when dropping it? Could be causing some faults on older models when you turn the car off and back on without letting EPS reset.

adeebshihadeh:
The Chrysler port always did this, then I replaced it with simpler logic that worked on our RAM. I believe we had to revert back to this because of faults.

sshane:
Ah I see, possibly always an issue then. Also sounds similar the fault our Ram has, so I'll make sure the fix covers both.

PR #25169, opened 2022-07-14
nav: toggle to show on left side of onroad ui

gregjhogan:
If you mount your device slightly behind your rear-view mirror due to the angle of your windshield, you cannot see the navigation directions.
The map on the left seems better to me because then you can choose what you want on the left side of the screen depending on which view you are on.
I think you could also argue that if you are using navigation you may care more about the directions than the road view, so it should be on the driver side.

### DIFF
--- a/common/params.cc
+++ b/common/params.cc
@@ -140,6 +140,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"LiveParameters", PERSISTENT},
     {"NavDestination", CLEAR_ON_MANAGER_START | CLEAR_ON_IGNITION_OFF},
     {"NavSettingTime24h", PERSISTENT},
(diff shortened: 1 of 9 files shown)


adeebshihadeh:
Discussed with Alex, and decided to make this a toggle.

pd0wm:
@sshane shall we allow unfinished translations in master? Or do we put Google translated strings as placeholder?

sshane:
You can comment out that unit test for now, @adeebshihadeh suggested an embedded percent finished indicator on the translations Readme.

adeebshihadeh:
I think English fallback is better; we'd rather have no translation than potentially nonsensical ones. We'll make sure to make it easy to see the status of missing translations for us and contributors.

PR #25171, opened 2022-07-14
CarInfo: add year sets

sshane:
- Will allow us to detect when years are changed instead of just saying a platform was removed and added (for car docs diff bot)
- For new car compatibility webpage which we want to ship standalone


sshane:
Smaller diff for separating model years for the new website: https://github.com/commaai/openpilot/pull/25208

We can reopen this if we want to explicitly store model years for whatever reason

PR #25178, opened 2022-07-14
Update Corolla Cross Hybrid f/w & doc

ErichMoraga:


selfdrive/car/toyota/values.py
```diff
@@ -124,6 +124,7 @@ class ToyotaCarInfo(CarInfo):
   ],
   CAR.COROLLAH_TSS2: [
     ToyotaCarInfo("Toyota Corolla Hybrid 2020-22"),
+    ToyotaCarInfo("Toyota Corolla Cross Hybrid 2020-22 (Non-US only)", min_enable_speed=7.5),
```
adeebshihadeh, line 127: are you sure about the min enable speed? usually the hybrids have stop and go

ErichMoraga: Yes, I had a lengthy chat w/ `@solin#0617` who had already been using custom forks from Ale and Shane.  He said his min. enable speed was 27 km/h, which is 7.5 m/s... the value I used in the CarInfo.

sshane: Yeah, some non-US models have that low speed lockout we fixed, bypassable by sending 1 shortly after startup

AlexandreSato: Apparently there are countries that receive the corolla cross imported from more than one country. When it comes to Latin American origin, they have this lockout.

PR #25185, opened 2022-07-15
Use agnos.py to check if current AGNOS is valid

pd0wm:
Allows for kernel only updates

### DIFF
--- a/launch_env.sh
+++ b/launch_env.sh
@@ -6,6 +6,7 @@ export NUMEXPR_NUM_THREADS=1
 export OPENBLAS_NUM_THREADS=1
 export VECLIB_MAXIMUM_THREADS=1
 
+# Remove this now?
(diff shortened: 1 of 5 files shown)


pd0wm:
@adeebshihadeh what do you think? Can you think of a cleaner way to check the current `system` version without resorting to this ugly `version_files`?

launch_env.sh
```diff
@@ -6,6 +6,7 @@ export NUMEXPR_NUM_THREADS=1
 export OPENBLAS_NUM_THREADS=1
 export VECLIB_MAXIMUM_THREADS=1
 
+# Remove this now?
```
pd0wm, line 9: We probably want to remove this, however this forces an AGNOS update for everyone. If we merge it at the same time when we actually do an AGNOS update this should be fine.

PR #25186, opened 2022-07-15
Lowbeam

AlexandreSato:
Brightness linked to low beam. This idea come from cydia2020 custom fork


cydia2020:
Relying on low beam alone might be problematic in some conditions especially in countries/jurisdictions where it's required by law to turn on the low beam in rainy conditions as seen in my notes in my Toyota combination meter dimmer repository. I think this would be able to be implemented if we can find the respective dimmer signals for all makes.

AlexandreSato:
I agree with cydia2020 and Jyoung.  Low beam is too simplistic and arbitrary.  Also in this commit it was not my changes in car.capnp

AlexandreSato:
Would it make sense to put a customizable factor that would go into rigor during a time interval?

PR #25122, opened 2022-07-12
VIN: query physical addresses

sshane:
Switches from querying the functional addresses (which don't work on some electric/hybrid Hyundai) to querying the rx addresses from the current functional address requests.

General cars:
- [x] KIA OPTIMA SX 2019 & 2016
- [x] TOYOTA RAV4 2017
- [x] Civic 2016-18
- [x] Civic 2022

Different 29-bit address in logs:
- [ ] HONDA FIT 2018
- [ ] HONDA HRV 2019
- [ ] HONDA CR-V HYBRID 2019
- [ ] HONDA ODYSSEY CHN 2019
- [ ] HONDA ODYSSEY 2018
- [ ] HONDA RIDGELINE 2017
- [ ] HONDA PILOT 2017
- [ ] HONDA CR-V 2016
- [x] HONDA INSIGHT 2019 - f44ac493c3564793|2022-08-28--10-39-49
- [x] HONDA ACCORD HYBRID 2018 - c55c923bcc7cd3fa|2022-08-26--18-38-26

Platforms that have responded to multiple queries:  
From a FW scan on a Ram 1500, the VIN is accessible over identifier 0xF190 on 0x7e0, 0x7e1, 0x7e2, and 0x7E7, which explains why it can randomly respond on three different addresses for these cars. These cars look good.
- [x] RAM 1500 5TH GEN
- [ ] CHRYSLER PACIFICA 2020
- [x] CHRYSLER PACIFICA HYBRID 2019 - ede4a914fcc6fb6d|2022-08-29--15-42-47--1
- [ ] CHRYSLER PACIFICA HYBRID 2018
- [ ] JEEP GRAND CHEROKEE 2019
- [ ] CHRYSLER PACIFICA 2018

Electric Hyundai with bad VINs:
- [ ] HYUNDAI KONA ELECTRIC 2019
- [ ] HYUNDAI IONIQ ELECTRIC 2020
- [x] HYUNDAI IONIQ 5 2022 (ca2ee648cdaf65a6|2022-08-23--19-52-19)
- [x] KIA EV6 2022
- [ ] KIA NIRO EV 2020
- [ ] 2022 Kona Electric in Malaysia does ***NOT*** work (before or after). Checked all ECUs and nothing responded to any standard identifier with his chassis number. User's car also doesn't seemingly have an engine ECU or a VMCU.

PR #24211, opened 2022-04-14
Ford: FPv2 firmware request

incognitojam:
This gets us responses from eps, esp, fwdRadar, fwdCamera and engine ECUs on my Focus, but we may need to verify that we get a response from the transmission too on a vehicle with an automatic transmission... This would be useful for transmission type detection.

PR #25188, opened 2022-07-16
Ford: parse VIN correctly

incognitojam:
Ford ECUs respond to the UDS VIN request with 24 bytes (padded with nulls). This was causing a "Malformed VIN" error in the logs.

Example Route: `86d00e12925f4df7|2022-06-25--12-18-27` (car params in useradmin just shows VIN: 0)

Example from parsing some recordings of UDS traffic:
```
0x7df  ReadDataByIdentifier.Request(didlist=['VIN'])
0x7e8  ReadDataByIdentifier.PositiveResponse(raw=\x57\x46\x30\x4e\x58\x58\x47\x43\x48\x4e\x4a\x52\x36\x39\x37\x31\x32\x00\x00\x00\x00\x00\x00\x00, utf8="WF0NXXGCHNJR69712")
```

### DIFF
--- a/selfdrive/car/vin.py
+++ b/selfdrive/car/vin.py
@@ -28,6 +28,9 @@ def get_vin(logcan, sendcan, bus, timeout=0.1, retry=5, debug=False):
           if vin.startswith(b'\x11'):
             vin = vin[1:18]
 
+          # Ford returns 24 bytes padded with nulls, trim to correct length
+          vin = vin[0:17]
+
           return addr[0], rx_addr, vin.decode()
         print(f"vin query retry ({i+1}) ...")
       except Exception:


sshane:
Can you see what it returns if you query the TX address for the rx address it responds on? Just check errorLogMessage and see what the Vin rx addr is

incognitojam:
Sure, `"vin_rx_addr": 2024`

Example Route: `86d00e12925f4df7|2022-07-16--16-10-52`

Btw I am having issues with fingerprint/detection after merging #24211 but I don't think it's actually related

incognitojam:
I can't see any response if I send the request to `0x7e0`.

sshane:
Did you set functional address to false? What happens if you try this branch? https://github.com/commaai/openpilot/pull/25122/files

incognitojam:
Oh, I think I left functional address set to true! I'll try that branch soon

PR #25190, opened 2022-07-17
simplify compatibility docs

adeebshihadeh:


sshane:
You can rebase once https://github.com/commaai/openpilot/pull/25193 is merged

PR #25199, opened 2022-07-18
check vin validity

adeebshihadeh:


selfdrive/car/vin.py
```diff
@@ -15,6 +16,11 @@
 UDS_VIN_RESPONSE = bytes([uds.SERVICE_TYPE.READ_DATA_BY_IDENTIFIER + 0x40]) + struct.pack("!H", uds.DATA_IDENTIFIER_TYPE.VIN)
 
 VIN_UNKNOWN = "0" * 17
+VIN_RE = "[A-HJ-NPR-Z0-9]{17}"
+
+
+def is_valid_vin(vin: str):
+  return re.fullmatch(VIN_RE, vin) is not None
```
gregjhogan, line 23: maybe this should be case-insensitive?

adeebshihadeh: went through all the logs and didn't find any cases with lowercase letters. would prefer to keep this strict then loosen it as bad VINs come in

PR #25202, opened 2022-07-18
Fix Ford FPv2: Send engine FW request to OBD port

incognitojam:
Follow up to #24211 since the engine FW was being identified as Mazda, as it didn't respond to the request on bus 1.

The engine only seems to respond to requests when they are sent through the OBD port (bus 0), while requests to other ECUs are blocked if they are sent through the OBD port.

Example route: `86d00e12925f4df7|2022-07-18--20-41-05` (this was using the `vin-addrs` branch but I added the extra FW request for testing, can be verified in the git diff)


incognitojam:
I wasn't sure whether or not to use the ECU whitelist for this.. Let me know what you think

sshane:
Whitelisting looks good, it eliminates waiting for the other ecus to respond on the engine query, and vice versa for the engine ecu for the bus=0 query

sshane:
Looks good!

PR #25205, opened 2022-07-18
Hyundai: add missing 2022 Santa Fe FW versions

sshane:
From ec07911a91f863f3


khoi:
worked `ec07911a91f863f3|2022-07-18--11-27-44--0`

PR #25207, opened 2022-07-19
modeld: Move from SNPE to tinygrad

geohot:
~~A bit slower still, 24.5 ms instead of 22.5 ms.~~

Speed is mostly a wash now, switched from exp to native_exp in the ELU and removed some image conversions. There's another straightforward ms to gain with a bit more work too, and a potential 2x-5x speedup if you could maximize GPU usage.

While the replay test isn't perfect, I think it's still correct, and suspect it's actually closer to model on PC. What's our test to confirm this?


adeebshihadeh:
I'm seeing pyopencl caching kernels in /home/, which is a tmpfs. It's not much right now (8MB), but this could become an issue later. Setting [PYOPENCL_NO_CACHE](https://documen.tician.de/pyopencl/runtime_program.html#envvar-PYOPENCL_NO_CACHE) should disable the cache.

geohot:
Ugh, it's slightly non deterministic

geohot:
Non determinism fixed, last bug is I think the first input frame not being black (why onnx replay is failing).

PR #25208, opened 2022-07-19
CarInfo: parse out model years

sshane:
Smaller diff for separating model years for the new website: https://github.com/commaai/openpilot/pull/25171/files

We can just parse the model years out, we don't have to make the huge change right away of switching to sets


selfdrive/car/docs_definitions.py
```diff
@@ -1,12 +1,13 @@
...
 from collections import namedtuple
 from dataclasses import dataclass
 from enum import Enum
-from typing import Dict, List, Optional, Union, no_type_check
+from typing import Dict, List, Optional, Tuple, Union, no_type_check
 
-TACO_TORQUE_THRESHOLD = 2.5  # m/s^2
-GREAT_TORQUE_THRESHOLD = 1.4  # m/s^2
 GOOD_TORQUE_THRESHOLD = 1.0  # m/s^2
+MODEL_YEARS_RE = r"(?<= )((\d{4}-\d{2})|(\d{4}))(,|$)"
```
sshane, line 10: Matches first `XXXX(,|EOL)` year or `XXXX-XX(,|EOL)`

PR #25211, opened 2022-07-19
home screen data and time show with emoji

crwusiz:


pd0wm:
Don't think this in an improvement over the current design.

PR #25215, opened 2022-07-19
Add video for 2019 Highlander Hybrid and let me create an issue for docs/website

davetapley:
I need to create issue for bug on https://comma.ai/vehicles but none of https://github.com/commaai/openpilot/issues/new/choose make sense

### DIFF
--- a/.github/ISSUE_TEMPLATE/docs.md
+++ b/.github/ISSUE_TEMPLATE/docs.md
@@ -1,0 +1,7 @@
+---
+name: Documentation
+about: For issues with documentation and website
+title: ''
+labels: 'docs'
+assignees: ''
+---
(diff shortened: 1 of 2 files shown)


sshane:
What issue is there with the webpage?

sshane:
Is it possible to make this video a timelapse? You can also try to lock the exposure on the road so it's not totally blown out from the sun.

davetapley:
@sshane the footnote for ⬇️ is rendered at the bottom, but not on the rows
https://github.com/commaai/openpilot/blob/26dad2423bd4078aab249ea9db3301a2a8a6c73e/selfdrive/car/toyota/values.py#L89-L91

Video was from a while back, no worries if you don't want to use it.

sshane:
Ah, nice find! I'll fix that

sshane:
Fixed in https://github.com/commaai/openpilot/pull/25219, let me know if you grab a new video!

PR #25223, opened 2022-07-20
Kia Niro EV 2019: add missing FW versions

sshane:


sshane:
@DerGue69 Merging so you can use master/other branches while we debug this, your car seems to be engaging despite the warning

PR #25231, opened 2022-07-21
Multilanguage: Korean improvements/translations

crwusiz:
https://github.com/commaai/openpilot/blob/444de299b6c0dd7081bc23b16b465f80a83d2d70/selfdrive/ui/qt/util.cc#L74-L94


crwusiz:
@sshane how to fix this..?

crwusiz:
I modified it by removing "s" %2 which is not used in translated languages

sshane:
Thanks!

PR #25232, opened 2022-07-21
Honda: match current speed on dash

pd0wm:
Signal is 2Hz on Honda Bosch, actual 10Hz on Nidec

TODO:
- [ ] Signal valid on all Honda cars? Only tested on Honda Bosch


adeebshihadeh:
You can get some coverage on a single segment of all the cars by adding a check [here](https://github.com/commaai/openpilot/blob/c3e95d5e194be9d445f54661d0ae0f0612db6a62/selfdrive/car/tests/test_models.py#L142) locally. If you want more coverage, you can build a route list and put it in `selfdrive/car/tests/routes.py`.

pd0wm:
I also meant we need to check if the signal actually reflects what's shown on the dash. Unfortunately that requires an actual car.

pd0wm:
A few cars don't seem to set `ROUGH_CAR_SPEED_2` (but so set `ROUGH_CAR_SPEED_3`).
 - `HONDA CR-V 2016`
 - `HONDA CR-V EU 2016`
 - `HONDA FIT 2018`
 - `HONDA HRV 2019`
 - `HONDA ACCORD HYBRID 2018`

sshane:
I checked and all the cars that are missing `ROUGH_CAR_SPEED_2` (real UI speed) are also missing `CAR_SPEED`, which I initially thought to use and apply hysteresis as it might be more common. However I'm checking online and these cars with these two missing signals don't actually have digital speedometers (except an RDX with one in the HUD in the top trim).

So `ROUGH_CAR_SPEED_2` should be fine to use. Probably should still detect its presence in case a new model year adds a digital speedometer, or if it's in the HUD on certain trims.

sshane:
Verified that ROUGH_CAR_SPEED_2 is non-zero and relatively close at a steady state to vEgo. I'll have a few users test and should be good to merge!

PR #25234, opened 2022-07-21
Toyota: match speed on dash

pd0wm:
TODO:
- [x] Find imperial bit for set speed
- [ ] Find current speed


selfdrive/car/toyota/carstate.py
```diff
@@ -94,6 +94,9 @@ def update(self, cp, cp_cam):
       ret.cruiseState.available = cp.vl["PCM_CRUISE_2"]["MAIN_ON"] != 0
       ret.cruiseState.speed = cp.vl["PCM_CRUISE_2"]["SET_SPEED"] * CV.KPH_TO_MS
 
+    # TODO: where is the metric/imperial bit?
```
sshane, line 97: `UI_SETTING->UNITS`?

cydia2020: That's potentially linked to the radio, so it might not be present depending on the configuration of the vehicle

sshane: I checked all our test routes in CI (one route per platform) and they all have the UI_SETTING message. Are you sure?

sshane: Ah, the message may exist, but the UNITS signal is 0 on some cars.

sshane: So I found `RSA2->SPDUNT`, and a new undefined signal on address 0x435, msb: 55, size: 3 which not only has the units, but the preference of gas mileage (however this didn't exist on the TSS-P RAV4, just my TSS2.5 Camry Hybrid)

sshane: Wait a minute, I checked the 2018 Camry route, and both `UI_SETTING->UNITS` and `RSA2->SPDUNT` were 0, *however* the new signal I found at 0x435 *does* exist, and it's 3, indicating the user had it set to imperial. So it looks like we can't use RSA2 or UI_SETTING across the board sadly. We might have to detect the presence of each message

sshane: Can't believe I missed this! Found a common units signal

sshane:
- [ ] Verify UI set speed is common and correct
- [ ] Verify UI speed is common and correct
- [x] Verify speed unit is common and correct

selfdrive/car/interfaces.py
```diff
@@ -163,12 +162,10 @@ def update(self, c: car.CarControl, can_strings: List[bytes]) -> car.CarState:
...
 
-    if ret.vEgoCluster == 0.0 and not self.v_ego_cluster_seen:
+    if ret.vEgoCluster == -1:
       ret.vEgoCluster = ret.vEgo
-    else:
-      self.v_ego_cluster_seen = True
 
-    if ret.cruiseState.speedCluster == 0:
+    if ret.cruiseState.speedCluster == -1:
       ret.cruiseState.speedCluster = ret.cruiseState.speed
```
sshane, line 169: Would be nice if we passed in a CarState to the car interface with default values so we could detect if the car interfaces set certain values or not without modifying cereal.

selfdrive/car/toyota/carstate.py
```diff
@@ -94,6 +94,14 @@ def update(self, cp, cp_cam):
       ret.cruiseState.available = cp.vl["PCM_CRUISE_2"]["MAIN_ON"] != 0
       ret.cruiseState.speed = cp.vl["PCM_CRUISE_2"]["SET_SPEED"] * CV.KPH_TO_MS
 
+    is_metric = cp.vl["BODY_CONTROL_STATE_2"]["UNITS"] in (1, 2)
+    conversion_factor = CV.KPH_TO_MS if is_metric else CV.MPH_TO_MS
+    if is_metric:
+      ret.vEgoCluster = cp.vl["BODY_CONTROL_STATE_2"]["UI_SPEED"] * CV.KPH_TO_MS
+    else:
+      ret.vEgoCluster = cp.vl["BODY_CONTROL_STATE_2"]["UI_SPEED"] * CV.KPH_TO_MS * 1.05
```
sshane, line 102: It seems like `UI_SPEED` is not the actual UI speed shown in the dash. For one, it updates visually at a different rate and lags behind dash. Secondly, it cannot show 1 mph/kph (skips from 2 kph to 0 kph) while the dash can. And finally, `UI_SET_SPEED` is actually in the car's unit preference, however `UI_SPEED` stays in metric no matter what (the comment in the DBC was copied over without verifying).

I found similar signals in 0x3cb and 0x361 (CLUTCH), however they didn't exist on all cars and were also always metric. The dash/cluster must being doing this conversion internally, and then slowing down the rate to 2 Hz itself, without publishing the signal anywhere, so looks like we need to do the same here.

sunnyhaibin:
The 2021 Lexus ES300h has an offset of 1.033. This was raised in a previous PR:
- https://github.com/commaai/openpilot/pull/24650

sshane:
These appear to be the true offsets @adeebshihadeh (using ratio of real vs. fake set speed):

```
TOYOTA RAV4 2022                - 1.05
TOYOTA RAV4 HYBRID 2022         - 1.05
TOYOTA MIRAI 2021               - 1.05
TOYOTA RAV4 HYBRID 2019         - 1.045
TOYOTA RAV4 2019                - 1.043
TOYOTA PRIUS 2017               - 1.041
TOYOTA PRIUS TSS2 2021          - 1.041
LEXUS RX 2016                   - 1.039
LEXUS ES 2019                   - 1.038
TOYOTA HIGHLANDER 2017          - 1.037
TOYOTA HIGHLANDER HYBRID 2018   - 1.037
TOYOTA AVALON 2016              - 1.036
TOYOTA COROLLA TSS2 2019        - 1.035
TOYOTA COROLLA 2017             - 1.035
TOYOTA SIENNA 2018              - 1.035
TOYOTA AVALON 2019              - 1.035
TOYOTA HIGHLANDER 2020          - 1.035
TOYOTA CAMRY HYBRID 2021        - 1.035
TOYOTA COROLLA HYBRID TSS2 2019 - 1.035
LEXUS RX 2020                   - 1.035
TOYOTA HIGHLANDER HYBRID 2020   - 1.035
LEXUS RX HYBRID 2017            - 1.035
LEXUS RX HYBRID 2020            - 1.035
LEXUS ES HYBRID 2018            - 1.034
TOYOTA AVALON 2022              - 1.032
LEXUS ES HYBRID 2019            - 1.031
TOYOTA C-HR 2018                - 1.03
TOYOTA CAMRY 2018               - 1.029
TOYOTA CAMRY HYBRID 2018        - 1.029
TOYOTA AVALON HYBRID 2019       - 1.029
TOYOTA CAMRY 2021               - 1.029
TOYOTA RAV4 HYBRID 2017         - 1.024
TOYOTA RAV4 2017                - 1.024
LEXUS NX 2018                   - 1.021
LEXUS NX 2020                   - 1.016
LEXUS NX HYBRID 2018            - 1.016
```

AlexandreSato:
I use 1.09 for corollas and corollacross Brazil made: https://github.com/AlexandreSato/openpilot/commit/05856cd63825947cbe463223da382ccbf8747c79

sshane:
Grouping this way makes much more sense, still increments of 0.005. I found from manual testing that the acceptable range is approximately also 0.005, for example 1.017 to 1.022 worked decently well for my car with hysteresis of 0.5 km/h:

```python
Group: 1.028, group avg: 1.0294, group mean diff: 0.0021
TOYOTA RAV4 HYBRID 2022         - 1.0307460317460317
TOYOTA MIRAI 2021               - 1.0304705882352942
TOYOTA RAV4 2022                - 1.03015
TOYOTA RAV4 2019                - 1.029957142857143
TOYOTA RAV4 HYBRID 2019         - 1.0293478260869566
TOYOTA PRIUS TSS2 2021          - 1.0258405797101449

Group: 1.023, group avg: 1.0221, group mean diff: 0.0009
TOYOTA PRIUS 2017               - 1.0230138888888891
LEXUS ES HYBRID 2018            - 1.0230000000000001
TOYOTA HIGHLANDER HYBRID 2018   - 1.0222173913043477
TOYOTA COROLLA HYBRID TSS2 2019 - 1.0216825396825397
TOYOTA HIGHLANDER 2017          - 1.0208245614035087

Group: 1.018, group avg: 1.0172, group mean diff: 0.0013
LEXUS RX 2020                   - 1.0196923076923077
LEXUS RX HYBRID 2020            - 1.0185
TOYOTA COROLLA TSS2 2019        - 1.0184675324675325
LEXUS ES HYBRID 2019            - 1.0174285714285716
LEXUS RX HYBRID 2017            - 1.0169166666666667
LEXUS ES 2019                   - 1.0167608695652175
TOYOTA CAMRY 2021               - 1.0166190476190478
TOYOTA COROLLA 2017             - 1.0163970588235294
TOYOTA CAMRY HYBRID 2021        - 1.016392857142857
TOYOTA HIGHLANDER 2020          - 1.016029411764706
LEXUS RX 2016                   - 1.015943396226415

Group: 1.013, group avg: 1.0139, group mean diff: 0.0011
TOYOTA HIGHLANDER HYBRID 2020   - 1.0154918032786886
TOYOTA AVALON 2019              - 1.01545
TOYOTA AVALON HYBRID 2019       - 1.0148
TOYOTA AVALON 2016              - 1.0138
TOYOTA CAMRY 2018               - 1.013661764705882
TOYOTA CAMRY HYBRID 2018        - 1.0135952380952378
TOYOTA C-HR 2018                - 1.0131000000000001
TOYOTA SIENNA 2018              - 1.0128214285714285
TOYOTA AVALON 2022              - 1.012
```

sshane:
@adeebshihadeh I think I've gotten the dash speed factors as close as I can. The two troubles I ran into were:

- on Toyota stock ACC, it always travels 1 mph less than the set speed, so to get the correct factors I had to add 1 mph to the set speeds. If you don't the factor would be much too high and the speed wouldn't match.
- on foreign cars that are locked to using metric, their factors seem much higher than their NA counterparts. I verified that it's not a metric vs. imperial difference as the same factor works on the TSS2 Camry Hybrid both in metric and imperial

So we'll have to accept that the C3's speed will never be exact on all Toyotas (unless we learn the factor online), however this PR should be a strict improvement to what's on master.

sshane:
We became a lot closer in PR: https://github.com/commaai/openpilot/pull/26408

PR #25235, opened 2022-07-21
Hyundai: match ego speed on dash

pd0wm:
Assume this is what shown on the dash based on DBC, but untested.

Some cars only set the metric field (`CF_Clu_VehicleSpeed`), not the imperial one (`CF_Clu_VehicleSpeed2`):
```
TestRoute(route='610ebb9faaad6b43|2020-06-13--15-28-36', car_model='HYUNDAI IONIQ ELECTRIC LIMITED 2019', segment=None)`
TestRoute(route='3f29334d6134fcd4|2022-03-30--22-00-50', car_model='HYUNDAI IONIQ PLUG-IN HYBRID 2019', segment=None)
TestRoute(route='b2a38c712dcf90bd|2020-05-18--18-12-48', car_model='HYUNDAI SONATA 2019', segment=None)
```
 
 Some cars populate the imperial field (`CF_Clu_VehicleSpeed2`) with a metric value, but the two fields are not exactly the same. Is this what happens when you set the car to metric and should we always use `CF_Clu_VehicleSpeed2`?
 ```
vEgo: 13.454141616821289, vEgoCluster: 21.45792007446289 - TestRoute(route='efc48acf44b1e64d|2021-05-28--21-05-04', car_model='HYUNDAI KONA ELECTRIC 2019', segment=None)
vEgo: 6.528930187225342, vEgoCluster: 11.623040199279785 - TestRoute(route='49f3c13141b6bc87|2021-07-28--08-05-13', car_model='HYUNDAI KONA HYBRID 2020', segment=None)
vEgo: 19.04660415649414, vEgoCluster: 31.292800903320312 - TestRoute(route='9c917ba0d42ffe78|2020-04-17--12-43-19', car_model='HYUNDAI PALISADE 2020', segment=None)
vEgo: 7.7457804679870605, vEgoCluster: 13.411199569702148 - TestRoute(route='bf43d9df2b660eb0|2021-09-23--14-16-37', car_model='HYUNDAI SANTA FE 2022', segment=None)
vEgo: 6.588061809539795, vEgoCluster: 11.623040199279785 - TestRoute(route='fb3fd42f0baaa2f8|2022-03-30--15-25-05', car_model='HYUNDAI TUCSON 2019', segment=None)
```

### CODE REVIEW

**Comment by sshane** [comment]
`CF_Clu_VehicleSpeed2` works well both for metric and imperial. Seems like with metric, both signals are identical in value

**Comment by sshane** [comment]
Will check with a notebook that the second signal is always filled and should be good to merge

**Comment by sshane** [comment]
Also found these routes to have problems with `CF_Clu_VehicleSpeed2` not set:

`HYUNDAI IONIQ ELECTRIC LIMITED 2019`

`HYUNDAI IONIQ PLUG-IN HYBRID 2019`

`HYUNDAI SONATA 2019`

`KIA NIRO HYBRID 2019`

`KIA OPTIMA SX 2019 & 2016`

`KIA SORENTO GT LINE 2018`

**Comment by sshane** [comment]
Huh, it's fine (very low res, but so is the dash) on our Optima. It must be different per-platform

Screenshot from 2022-08-04 00-07-13

**Comment by sshane** [comment]
Some notes to myself:

These cars do not have `CF_Clu_VehicleSpeed2` set (`(ret.vEgoRaw, cp.vl["CLU15"]["CF_Clu_VehicleSpeed2"], cp.vl["CLU15"]["CF_Clu_VehicleSpeed"] * speed_conv)`):

```python
================================================================================================== short test summary info ==================================================================================================
FAILED test_models.py::TestCarModel_73_HYUNDAI_SONATA_2019::test_panda_safety_carstate - AssertionError: (14.437933921813965, 0.0, 24.14016)
FAILED test_models.py::TestCarModel_63_HYUNDAI_IONIQ_PLUG_IN_HYBRID_2019::test_panda_safety_carstate - AssertionError: (5.679253578186035, 0.0, 8.940800000000001)
FAILED test_models.py::TestCarModel_59_HYUNDAI_IONIQ_ELECTRIC_LIMITED_2019::test_panda_safety_carstate - AssertionError: (3.046875, 0.0, 4.470400000000001)
FAILED test_models.py::TestCarModel_89_KIA_OPTIMA_SX_2019_2016::test_panda_safety_carstate - AssertionError: (14.24696159362793, 0.0, 14.166666666666668)
FAILED test_models.py::TestCarModel_86_KIA_NIRO_HYBRID_2019::test_panda_safety_carstate - AssertionError: (14.555121421813965, 0.0, 23.246080000000003)
FAILED test_models.py::TestCarModel_91_KIA_SORENTO_GT_LINE_2018::test_panda_safety_carstate - AssertionError: (9.37717056274414, 0.0, 10.0)
```

sshane:
It looks like on cars that are locked to metric/metric by default these two signals are actually the same

selfdrive/car/hyundai/carstate.py
```diff
@@ -32,12 +32,18 @@ def __init__(self, CP):
     self.park_brake = False
     self.buttons_counter = 0
 
+    # When available we use CLU15->CF_Clu_VehicleSpeed2 to populate vEgoCluster which is the actual dash speed
+    # However, on some cars this is not always present, so we use the less accurate CF_Clu_VehicleSpeed signal
+    self.accurate_dash_speed_seen = False
+
```
sshane, line 38: The main difference is that `CF_Clu_VehicleSpeed` can sometimes be slightly noisier and the timing of when it increments or decrements speed is slightly different than the dash. However, in a steady state it should match the dash much better than vEgo as it includes the same dash factor as `CF_Clu_VehicleSpeed2`.

sshane: I chose to remove `CF_Clu_VehicleSpeed` from this PR, as it can rapidly oscillate by a full 1 kph sometimes on certain cars which can't be removed with hysteresis, as it seems the signal already has hysteresis (makes steady state inaccurate). `CF_Clu_VehicleSpeed2` affects a lot of cars so I think this PR is good to go as is, once we get a few testers

sshane:
Here's plots for all the cars, you can see some cars have factors and some don't, and the signals can be exceptionally noisy on some cars, despite the signal already being filtered (it is laggy + has hysteresis):

https://imgur.com/a/QSJhr4c

sshane:
Super happy with this now on the Optima (yet to test Sonata, but should match). On Optima the C3 speed matches in a steady state at all speeds, and even when accelerating or decelerating rapidly.

The two things I had to do to make the `CF_Clu_VehicleSpeed` signal match up with the Optima exactly are:

1. Sample the signal at 5 Hz, which is what the dash does. Some cars have `CLU15` messages being sent at 12+ Hz with varying levels of noise in the signal we're interested in, so samping at 5 Hz removes that.
2. Instead of converting and then rounding the always-metric `CF_Clu_VehicleSpeed` signal as normal, I needed to apply the following formula to make the C3 not skip any speeds ([it skips 77 mph, 72 mph, 67 mph, 59 mph, 54 mph, 49 mph, ... 26 mph, 18 mph, 13 mph, 8 mph, 3 mph without this](https://imgur.com/a/7HBayXC)):
  ```python
  math.floor(self.dash_speed_sampled * CV.KPH_TO_MPH + CV.KPH_TO_MPH)
  ```

That's it! Since this signal is set on all cars, *and* since the two signals are the same on metric-only cars, we needed to have these two compensations anyway. Plus it will be better if we're able to use just `CF_Clu_VehicleSpeed` for all cars (if it matches on the other cars, will get testers).

sshane:
Tried filtering on the Optima at 0.2 seconds (was a balance of not being too delayed and getting rid of some oscillation), and it was noticeably more laggy in the UI. Plus it's possible for the signal to oscillate in such a way that even with the filtered signal it will visually oscillate just as fast.

Any time you see the green line thicker than normal, it's oscillating: https://imgur.com/a/DFMBkOd

PR #25250, opened 2022-07-22
Skoda unsupported connector footnote; note J533 alternative isn't available on all cars

belm0:
cc:  @jyoung8607 @vanillagorillaa 

see https://discord.com/channels/469524606043160576/534359517836607488/999917791262031912

PR #25256, opened 2022-07-22
Fix for multiple footnotes in car docs

### LINKED ISSUE #25255: CARS.md generator not handling multiple footnotes
### Describe the bug

I noticed that multiple footnotes is not rendered correctly (only the first footnote is referenced).  E.g.:

```
VWCarInfo("Volkswagen Passat 2015-22", footnotes=[Footnote.VW_HARNESS, Footnote.PASSAT, Footnote.VW_VARIANT], harness=Harness.j533),
```
renders as
Screen Shot 2022-07-22 at 10 26 20 PM

I'd expect it to reference all footnotes (e.g. ⁶𝄒⁷𝄒⁸).

n/a

### openpilot version

n/a

### Additional info

_No response_

incognitojam:
Highlighted in https://github.com/commaai/openpilot/pull/25250#issuecomment-1192910570


selfdrive/car/docs_definitions.py
```diff
@@ -129,9 +129,10 @@ def get_column(self, column: Column, star_icon: str, footnote_tag: str) -> str:
     elif column == Column.MODEL and len(self.years):
       item += f" {self.years}"
 
-    footnote = get_footnote(self.footnotes, column)
-    if footnote is not None:
-      item += footnote_tag.format(self.all_footnotes[footnote])
+    footnotes = get_footnotes(self.footnotes, column)
+    if footnotes is not None:
+      for fn in footnotes:
+        item += footnote_tag.format(self.all_footnotes[fn])
```
sshane, line 135: Can we split on ", " or ","?

incognitojam: Sure

selfdrive/car/docs_definitions.py
```diff
@@ -129,9 +129,9 @@ def get_column(self, column: Column, star_icon: str, footnote_tag: str) -> str:
     elif column == Column.MODEL and len(self.years):
       item += f" {self.years}"
 
-    footnote = get_footnote(self.footnotes, column)
-    if footnote is not None:
-      item += footnote_tag.format(self.all_footnotes[footnote])
+    footnotes = get_footnotes(self.footnotes, column)
+    if footnotes is not None:
+      item += ", ".join(footnote_tag.format(self.all_footnotes[fn]) for fn in footnotes)
```
belm0, line 134: It would look strange to delimit superscript text with a normal comma.  I recommend `𝄒` character as delimeter.

belm0:
I filed bug #25255, please add close reference in description.

selfdrive/car/docs_definitions.py
belm0: or else use Wikipedia style with brackets

Screen Shot 2022-07-22 at 11 26 09 PM

selfdrive/car/docs_definitions.py
sshane: I think we need to put the comma inside the superscript tag, somehow

sshane: How do you like that?

docs/CARS.md
```diff
@@ -148,74 +148,74 @@ A supported vehicle is one that just works when you install a comma device. Ever
...
-|Toyota|RAV4 Hybrid 2017-18|All|<a href="##"><img valign="top" src="assets/icon-star-half.svg" width="22" /></a>[<sup>3</sup>](#footnotes)|<a href="##"><img valign="top" src="assets/icon-star-full.svg" width="22" /></a>|<a href="##"><img valign="top" src="assets/icon-star-full.svg" width="22" /></a>|<a href="##"><img valign="top" src="assets/icon-star-full.svg" width="22" /></a>|
+|Toyota[<sup></sup>](#footnotes)|RAV4 Hybrid 2016[<sup></sup>](#footnotes)|TSS-P[<sup></sup>](#footnotes)|<a href="##"><img valign="top" src="assets/icon-star-half.svg" width="22" /></a>[<sup>3</sup>](#footnotes)|<a href="##"><img valign="top" src="assets/icon-star-full.svg" width="22" /></a>[<sup></sup>](#footnotes)|<a href="##"><img valign="top" src="assets/icon-star-full.svg" width="22" /></a>[<sup></sup>](#footnotes)|<a href="##"><img valign="top" src="assets/icon-star-full.svg" width="22" /></a>[<sup></sup>](#footnotes)|
```
belm0, line 184: empty `<sup>` is being generated, it doesn't seem correct

sshane: Fixed 🙂

selfdrive/car/docs_definitions.py
sshane: We don't link to specific footnotes, so I think commas are fine if they match the font style of the superscript (which they do now)

selfdrive/car/docs_definitions.py
```diff
@@ -129,9 +125,9 @@ def get_column(self, column: Column, star_icon: str, footnote_tag: str) -> str:
     elif column == Column.MODEL and len(self.years):
       item += f" {self.years}"
 
-    footnote = get_footnote(self.footnotes, column)
-    if footnote is not None:
-      item += footnote_tag.format(self.all_footnotes[footnote])
+    footnotes = get_footnotes(self.footnotes, column)
+    if len(footnotes):
```
belm0, line 129: `len()` is not needed here.   (`if footnotes` is idiomadic Python for a list)

selfdrive/car/docs_definitions.py
```diff
@@ -129,9 +125,9 @@ def get_column(self, column: Column, star_icon: str, footnote_tag: str) -> str:
...
+      item += footnote_tag.format(",".join(sorted([str(self.all_footnotes[fn]) for fn in footnotes])))
```
belm0, line 130: if there are more than 9 footnotes, sorting the strings won't be correct

```python
item += footnote_tag.format(",".join([str(self.all_footnotes[fn]) for fn in sorted(footnotes])))
```

belm0, line 130: It's minor, but I think output readability will be improved by enclosing the list in `[]`.  Having the numbers run together with the model year is visually confusing, even if superscript is used.

Screen Shot 2022-07-23 at 6 26 45 AM

`6,7,8` --> `[6,7,8]`

sshane: These are AMA standard, so there's at least a standard for it: https://guides.lib.uw.edu/hsl/ama/intext

selfdrive/car/docs_definitions.py
sshane: We do this countless places around openpilot, so this is sticking with that standard. Plus it's better to be more explicit even if we don't have to be.

selfdrive/car/docs_definitions.py
sshane: Tried it, there's no way I found to center the stars so it didn't look too consistent:
Screenshot from 2022-07-22 23-13-58

PR #25265, opened 2022-07-23
Mazda: use torque controller

adeebshihadeh:
Need to test on CX5 and CX9, rest are dashcam only.


alancyau:
Here’s a log from a 2022 Mazda CX-5 running this branch. 

a4af1602d8e668ac|2022-07-23--23-17-42--0

Jafaral:
CX-9 2021 : 8c521516df3527d8|2022-07-24--18-57-08

adeebshihadeh:
Both of your plots look pretty good. How does it feel compared to the current tuning on master?

Jafaral:
The drive was smooth overall. I did not test current master to make a direct comparison.

alancyau:
Smooth overall. Pretty similar to master. It seems like lane change might be a bit jerky though.

PR #25267, opened 2022-07-24
Ford: fix gear parsing

incognitojam:
`CANDefine` converts all of the signal defined values to uppercase/snake case, so our map from signal to gears should use uppercase too.

Also added Manumatic as an extra drive gear for Ford.

### DIFF
--- a/selfdrive/car/ford/carstate.py
+++ b/selfdrive/car/ford/carstate.py
@@ -1,5 +1,3 @@
-from typing import Dict
-
 from cereal import car
 from common.conversions import Conversions as CV
 from opendbc.can.can_define import CANDefine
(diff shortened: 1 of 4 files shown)


incognitojam:
Just rebased due to conflicts with master

selfdrive/car/ford/carstate.py
```diff
@@ -88,8 +88,8 @@ def update(self, cp, cp_cam):
   @staticmethod
   def parse_gear_shifter(gear: str) -> car.CarState.GearShifter:
     d: Dict[str, car.CarState.GearShifter] = {
-        'Park': GearShifter.park, 'Reverse': GearShifter.reverse, 'Neutral': GearShifter.neutral,
-        'Manual': GearShifter.manumatic, 'Drive': GearShifter.drive,
+        'PARK': GearShifter.park, 'REVERSE': GearShifter.reverse, 'NEUTRAL': GearShifter.neutral,
+        'MANUAL': GearShifter.manumatic, 'DRIVE': GearShifter.drive,
```
adeebshihadeh, line 92: hmm, we can just add this onto the dict in `selfdrive/car/interfaces.py`, then do `d.get(gear.upper(), GearShifter.unknown)`

incognitojam: sounds good :)

PR #25269, opened 2022-07-25
Add Arabic Translation

royjr:


selfdrive/ui/translations/main_ar.ts
```diff
@@ -0,0 +1,1347 @@
...
+    <message numerus="yes">
+        <location filename="../qt/widgets/input.cc" line="168"/>
+        <source>Need at least %n character(s)!</source>
+        <translation type="unfinished">
+            <numerusform></numerusform>
+            <numerusform></numerusform>
+            <numerusform></numerusform>
+            <numerusform></numerusform>
+            <numerusform></numerusform>
+            <numerusform></numerusform>
```
sshane, line 318: Do you mind adding all the plural forms for these translations?

royjr: Done! I also see 6 vanished translations, should we remove them?

sshane: Yep that's fine, you can run `./update_translations.py --vanish`

PR #25274, opened 2022-07-25
cars.md: use markdown images for stars

pd0wm:
Fixes stars on https://docs.comma.ai/CARS.html

https://github.com/commaai/openpilot/blob/markdown-stars/docs/CARS.md


sshane:
svg should not be used anywhere else, LGTM

PR #25280, opened 2022-07-25
Honda Civic 2022: forward fault bit from camera

royjr:
Master branch causes a brake system light to stay on my dash, may require a restart after install for brake system light to come up, then shows up every restart after. The light never shows up again with this fix. @sshane 

Civic22 users with issues
-First (sport, confirmed fix)
-Kevin
-Marco the Dingo (sport touring, experiences issue with sunnypilot, but not on master-ci)

Civic22 users without issues
-Rethkir (sport touring)

### DIFF
--- a/selfdrive/car/honda/carcontroller.py
+++ b/selfdrive/car/honda/carcontroller.py
@@ -237,7 +237,7 @@ class CarController:
     if self.frame % 10 == 0:
       hud = HUDData(int(pcm_accel), int(round(hud_v_cruise)), hud_control.leadVisible,
                     hud_control.lanesVisible, fcw_display, acc_alert, steer_required)
-      can_sends.extend(hondacan.create_ui_commands(self.packer, self.CP, CC.enabled, pcm_speed, hud, CS.is_metric, CS.stock_hud))
+      can_sends.extend(hondacan.create_ui_commands(self.packer, self.CP, CC.enabled, pcm_speed, hud, CS.is_metric, CS.acc_hud, CS.lkas_hud))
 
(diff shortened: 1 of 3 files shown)


sshane:
Do you have a route?

royjr:
master
1a5d045d2c531a6d|2022-07-25--17-05-43
1a5d045d2c531a6d|2022-07-25--17-06-06

civic22_brakesystemlight
1a5d045d2c531a6d|2022-07-25--17-07-18

sshane:
Looks like your car doesn't have the same symptoms as the other users (missing pandaStates or can packets), let me have you try one thing.

It's possible we're initializing and sending 0 for LKAS_PROBLEM too early sometimes, and the car is waiting for a set amount of time before it accepts that value transitioning to 0. So if we just pass it through from the camera, it might work as well.

sshane:
@royjr see if this fixes it

royjr:
Looks good @sshane!

1a5d045d2c531a6d|2022-07-25--18-00-19
1a5d045d2c531a6d|2022-07-25--18-00-51

sshane:
If you drive on this in the next couple of days let me know how it goes. About how often would it remain faulted without these changes on master?

royjr:
@sshane Still looks good, no problems with the fix. Remains faulted on master for at least 1.5hrs, haven't gone on longer rides recently.

sshane:
Any problems since then on this branch or is all good?

royjr:
All good! Wanted to confirm issue persists on master.

PR #25291, opened 2022-07-26
Lateral Acceleration Error alert (Sensor Invalid)

nuwandavek:
Large Lateral Acceleration discrepancies (caused by shaking the device / device falling) trigger this alert.
image


pd0wm (approved):
LGTM

What exactly is the process replay diff?

nuwandavek:
> What exactly is the process replay diff?
This was a bug! I should have checked the ref properly, thanks for pointing out.

The offline paramsd was slightly different from the online one, and there is a sign difference between `locationd` and `paramsd` yaw rates. Fixed that and reverted the ref update. (and also made sure online and offline behaviour is the same).

PR #25292, opened 2022-07-27
params: add put_bool_nonblocking

ZwX1616:


pd0wm:
I don't think this works with the `ensure_bytes`. Why don't you just add a new function `put_bool_nonblocking`?

ZwX1616:
> 

I am confused. `ensure_bytes` here operates on the key no? Was it intended to `ensure_bytes(dat)` ?

ZwX1616:
> I am confused. `ensure_bytes` here operates on the key no? Was it intended to `ensure_bytes(dat)` ?

nvm. we ensure the key here and put() itself ensures the dat. But check_key also ensures the key anyways... (duplicate ensure key in put_nonblocking?)

pd0wm:
Ah you're right. My bad!

ZwX1616:
> Ah you're right. My bad! Still not a fan of the "is_bool" flag. Think a separate function is cleaner.

Yeah I actually prefer that the more I look at it. Thanks for the cleanup

PR #25307, opened 2022-07-29
SPI panda

robbederks:


adeebshihadeh:
Split this up into smaller chunks

PR #25309, opened 2022-07-29
Cart docs diff bot: more resilient to CarInfo structure changes

sshane:
Not sure if we install this package or not in the dev environemnt, let's see


sshane:
Main problem is the `Enum` CarFootnote: `ValueError: CarFootnote(text="28mph for Camry 4CYL L, 4CYL LE and 4CYL SE which don't have Full-Speed Range Dynamic Radar Cruise Control.", column=<Column.FSR_LONGITUDINAL: 'Stop and Go'>, star=None) is not a valid Footnote`

PR #25313, opened 2022-07-30
Subaru: Legacy 2020-22 support

adeebshihadeh:
- [x] test route
- [x] remove forced fingerprint & add 2022 fw versions


adeebshihadeh:
@martinl do we have a test route for this car?

martinl:
> @martinl do we have a test route for this car?

I requested test route from @KingChalupa, he will send one soon

PR #25315, opened 2022-07-30
sim: role_name attribute used by ScenarioRunner for CARLA

jeroenlammersma:
* Added role_name attribute to sim that is used by ScenarioRunner for the CARLA simulator.

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -304,6 +304,7 @@ class CarlaBridge:
     world_map = world.get_map()
 
     vehicle_bp = blueprint_library.filter('vehicle.tesla.*')[1]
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
What exactly does this do?

jeroenlammersma:
This attribute is used by Scenario Runner for CARLA. Without this attribute, Scenario Runner cannot recognize the ego (hero) vehicle.

Info: https://carla.readthedocs.io/en/latest/ts_traffic_simulation_overview/#scenario-runner-and-openscenario
Repo: https://github.com/carla-simulator/scenario_runner

The line of code which retrieves the attribute: https://github.com/carla-simulator/scenario_runner/blob/master/srunner/scenariomanager/carla_data_provider.py#L740

PR #25316, opened 2022-07-31
Mazda Radar Intercept. Work in progress

MoreTore:
This is main part of the mazda radar intercept code.

### DIFF
--- a/selfdrive/car/subaru/carstate.py
+++ b/selfdrive/car/subaru/carstate.py
@@ -111,6 +111,8 @@ class CarState(CarStateBase):
       ("Signal2", "ES_Distance"),
       ("Car_Follow", "ES_Distance"),
       ("Signal3", "ES_Distance"),
+      ("Cruise_Soft_Disable", "ES_Distance"),
+      ("Signal7", "ES_Distance"),
       ("Cruise_Brake_Active", "ES_Distance"),
       ("Distance_Swap", "ES_Distance"),
       ("Cruise_EPB", "ES_Distance"),


adeebshihadeh:
Can you break out the radar interface into its own PR? Do we already have the radar messages on our existing harness or does the interceptor enable this?

MoreTore:
> Do we already have the radar messages on our existing harness or does the interceptor enable this?
Before I go into this I just want to say that I have the code completed and about 10 people are using it.
OpenPilot _could_ see the radar addresses on bus 0 but that is not how it's used. In order to actually use longitudinal, you need to:

1.  Run a cable to the radar canbus and tie it to the camera bus. This isolates the radar signals so we can intercept the data.
2.  The radar signals will be on the camera bus, and we need to pass through some parts of those signals because we are not doing full emulation. So we need to add radar signals and checks to get_cam_can_parser() so they can be passed to create_radar_command(). In mazda_2017.dbc I define the radar track signals as MSG1 (32bits), MSG2 (28bits), and CTR1 (4bits). This allows me to easily pass through all the radar tracks in create_radar_command(cp_cam).
```
if (frame % 10 == 0):
      for addr in range(361,367):
        addr_name = f"RADAR_{addr}"
        msg = CS.cp_cam.vl[addr_name]
        values = {
          "MSGS_1" : int(msg["MSGS_1"]),
          "MSGS_2" : int(msg["MSGS_2"]),
          "CTR"    : int(msg["CTR"])
        } 
        ret.append(packer.make_can_msg(addr_name, 0, values))
```
The actual radar track data is defined in mazda_radar.dbc. These definitions only contain the relevant track data as there is some data in the radar tracks that I haven't figured out.
```
("ANG_OBJ", msg, 0),
("DIST_OBJ", msg, 0),
("RELV_OBJ", msg, 0),
```
4. Update panda safety_mazda.h to block the radar signals from being passed through.
5. Enable longitudinal and call create_radar_command() in carcontroller.py

We cannot use the designated radar bus on the harness because we need mux that bus to the obd port on-road for other reasons.

MoreTore:
> Can you break out the radar interface into its own PR? 

#25321

gregjhogan:
FYI some other cars like this we disable the radar with UDS commands and then openpilot can control longitudinal without needing to run an extra harness. Hyundai we even still get the radar points when the radar is disabled.

Some people do run an extra harness to pass through AEB from the radar, but I don't believe we currently support this setup for any vehicles, so those people would currently be on forks.

MoreTore:
> FYI some other cars like this we disable the radar with UDS commands and then openpilot can control longitudinal without needing to run an extra harness. Hyundai we even still get the radar points when the radar is disabled.
> 
> Some people do run an extra harness to pass through AEB from the radar, but I don't believe we currently support this setup for any vehicles, so those people would currently be on forks.

I heard of that but full emulation is difficult. I have not tried to disable the radar but I would imagine if it's disabled there would be more issues. So what your saying is that if i need to run a harness you won't upstream?

mlouielu:
> FYI some other cars like this we disable the radar with UDS commands and then openpilot can control longitudinal without needing to run an extra harness. Hyundai we even still get the radar points when the radar is disabled.
> 

Is this what `hyundai_enable_radar_points.py` is doing? How do you find out which UDS command can change this behavior? Thanks.

PR #25318, opened 2022-07-31
VW MQB: Early EPS faults are temporary

jyoung8607:
Improve startup behavior for cars without factory Lane Assist.

When openpilot can start up relatively late after the rest of the car, on a car doesn't have factory Lane Assist, nothing is there to send the expected HCA_01 message, and we initially see the EPS in a FAULT state. It recovers quickly when openpilot starts sending, but the openpilot UI briefly shows an unnecessary and disconcerting fault message to the driver.

For the first three seconds after controls start, handle FAULT the same as INITIALIZING, which we legitimately see for a moment if openpilot starts up fast on cars with Lane Assist. This keeps us from showing the driver superfluous UI alerts in most cases, while still letting us alert the driver that we can't steer yet in cases like late startup/reboot with stock ACC already engaged.

### DIFF
--- a/selfdrive/car/volkswagen/carstate.py
+++ b/selfdrive/car/volkswagen/carstate.py
@@ -42,10 +42,10 @@ class CarState(CarStateBase):
     ret.yawRate = pt_cp.vl["ESP_02"]["ESP_Gierrate"] * (1, -1)[int(pt_cp.vl["ESP_02"]["ESP_VZ_Gierrate"])] * CV.DEG_TO_RAD
 
     # Verify EPS readiness to accept steering commands
-    # To handle cars without factory Lane Assist, treat FAULT like INITIALIZING for three seconds after startup
+    # For cars without Lane Assist, treat FAULT like INITIALIZING for worst likely controls init and EPS recovery time
     hca_status = self.hca_status_values.get(pt_cp.vl["LH_EPS_03"]["EPS_HCA_Status"])
-    ret.steerFaultPermanent = hca_status == "DISABLED" or (hca_status == "FAULT" and self.frame >= 300)
-    ret.steerFaultTemporary = hca_status in ("INITIALIZING", "REJECTED") or (hca_status == "FAULT" and self.frame < 300)
+    ret.steerFaultPermanent = hca_status == "DISABLED" or (hca_status == "FAULT" and self.frame >= 600)
+    ret.steerFaultTemporary = hca_status in ("INITIALIZING", "REJECTED") or (hca_status == "FAULT" and self.frame < 600)
 
     # Update gas, brakes, and gearshift.
     ret.gas = pt_cp.vl["Motor_20"]["MO_Fahrpedalrohwert_01"] / 100.0
--- a/selfdrive/car/volkswagen/carstate.py
+++ b/selfdrive/car/volkswagen/carstate.py
@@ -42,6 +42,7 @@ class CarState(CarStateBase):
     ret.yawRate = pt_cp.vl["ESP_02"]["ESP_Gierrate"] * (1, -1)[int(pt_cp.vl["ESP_02"]["ESP_VZ_Gierrate"])] * CV.DEG_TO_RAD
 
     # Verify EPS readiness to accept steering commands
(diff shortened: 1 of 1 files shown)


selfdrive/car/volkswagen/carstate.py
```diff
@@ -41,9 +42,10 @@ def update(self, pt_cp, cam_cp, ext_cp, trans_type):
     ret.yawRate = pt_cp.vl["ESP_02"]["ESP_Gierrate"] * (1, -1)[int(pt_cp.vl["ESP_02"]["ESP_VZ_Gierrate"])] * CV.DEG_TO_RAD
 
     # Verify EPS readiness to accept steering commands
+    # To handle cars without factory Lane Assist, treat FAULT like INITIALIZING for three seconds after startup
     hca_status = self.hca_status_values.get(pt_cp.vl["LH_EPS_03"]["EPS_HCA_Status"])
-    ret.steerFaultPermanent = hca_status in ("DISABLED", "FAULT")
-    ret.steerFaultTemporary = hca_status in ("INITIALIZING", "REJECTED")
+    ret.steerFaultPermanent = hca_status == "DISABLED" or (hca_status == "FAULT" and self.frame >= 300)
```
adeebshihadeh, line 47: this is less than the [controls initializing time](https://github.com/commaai/openpilot/blob/22408781f7d7df5ff0395f80c1b15d6407acd0ff/selfdrive/controls/controlsd.py#L416), perhaps we can consider faults in there too

adeebshihadeh: or you can mark canValid as false

jyoung8607: I slid the timer out to compensate for worst case controls init and commented why. There's no real harm in doing so. I prefer that over the later troubleshooting headache I'd invite by tampering with canValid outside normal paths.

jyoung8607: I looked more closely at your proposal, and as far as I can tell, we don't [signal boardd to set our safety mode](https://github.com/commaai/openpilot/blob/20c140b10eef52b6f6d6b9e142ed944264865bac/selfdrive/controls/controlsd.py#L425) or begin [sending CAN messages from carcontroller](https://github.com/commaai/openpilot/blob/20c140b10eef52b6f6d6b9e142ed944264865bac/selfdrive/controls/controlsd.py#L720-L723) until after CS initialization. For purposes of this PR, the EPS won't start recovering until it starts seeing openpilot CAN traffic, so holding down controls init would be counterproductive.

jyoung8607: Do you care to see the 600 restated as (6. / DT_CTRL)? Seems weird to care about DT_CTRL inside a car port, but you're doing it in comma body...

selfdrive/car/volkswagen/carstate.py
```diff
@@ -41,9 +42,10 @@ def update(self, pt_cp, cam_cp, ext_cp, trans_type):
     ret.yawRate = pt_cp.vl["ESP_02"]["ESP_Gierrate"] * (1, -1)[int(pt_cp.vl["ESP_02"]["ESP_VZ_Gierrate"])] * CV.DEG_TO_RAD
 
     # Verify EPS readiness to accept steering commands
+    # For cars without Lane Assist, treat FAULT like INITIALIZING for worst likely controls init and EPS recovery time
     hca_status = self.hca_status_values.get(pt_cp.vl["LH_EPS_03"]["EPS_HCA_Status"])
-    ret.steerFaultPermanent = hca_status in ("DISABLED", "FAULT")
-    ret.steerFaultTemporary = hca_status in ("INITIALIZING", "REJECTED")
+    ret.steerFaultPermanent = hca_status == "DISABLED" or (hca_status == "FAULT" and self.frame >= 600)
+    ret.steerFaultTemporary = hca_status in ("INITIALIZING", "REJECTED") or (hca_status == "FAULT" and self.frame < 600)
```
adeebshihadeh, line 48: can you also check the falling edge? I can already see this popup in the MTBF report where a user engages in the first 6s with the EPS in the FAULT state

jyoung8607:
Withdrawing based on conversation with @adeebshihadeh, going to solve this in a different way.

PR #25322, opened 2022-08-01
Subaru: Add missing Forester 2021 engine fw

martinl:
**Description** Add missing engine fw version for Forester 2021 from following route

**Route**
Route: 7db92cafc6d7cd6b|2022-07-31--20-13-23

### DIFF
--- a/selfdrive/car/subaru/values.py
+++ b/selfdrive/car/subaru/values.py
@@ -240,6 +240,7 @@ FW_VERSIONS = {
       b'\xcb\"`@\a',
       b'\xcb\"`p\a',
       b'\xf1\x00\xa2\x10\n',
+      b'\xcf"`p\x07',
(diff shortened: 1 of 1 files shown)


selfdrive/car/subaru/values.py
```diff
@@ -240,6 +240,7 @@ class SubaruCarInfo(CarInfo):
       b'\xcb\"`@\a',
       b'\xcb\"`p\a',
       b'\xf1\x00\xa2\x10\n',
```
sshane, line 242: Just a heads up, you're welcome to remove FW versions from other brand's queries for ECUS when adding new FW versions. So for example `b'\xf1\x00\xa2\x10\n',` comes from a Hyundai query. I'll do a big clean up of what I can soon

PR #25334, opened 2022-08-02
Update RDX 2022 min steer speed

sshane:
The 2022 (and some 2021 Odyssey) only can steer to 40 mph, not 3 mph.

On second look-over, it's possible the older RDX's were set to 3 mph with VOACC in mind (different min steer speed). Will check that out


sshane:
None of these cars are in openpilot yet according to @csouers (radar FW `b'36802-TJB-A540\x00\x00'`)

PR #25335, opened 2022-08-02
CarInfo: store if EV/hybrid

sshane:


sshane:
We'll try to use the NHTSA API for this

jyoung8607:
> Add VW

BEV: e-Golf

Hybrids: Arteon eHybrid, Golf GTE, Passat GTE, A3 Sportback e-tron

PR #25336, opened 2022-08-02
Car docs detail sentence: keep mention of brake to zero

sshane:


selfdrive/car/docs_definitions.py
```diff
@@ -152,7 +152,9 @@ def get_detail_sentence(self, CP):
       if self.min_enable_speed > 0:
         acc = f" <strong>while driving above {self.min_enable_speed * CV.MS_TO_MPH:.0f} mph</strong>"
       elif CP.carName not in ("nissan", "subaru", "toyota") or (CP.carName == "toyota" and CP.openpilotLongitudinalControl):
-        acc = " <strong>that automatically resumes from a stop</strong>"
+        acc = " that <strong>automatically resumes from a stop</strong>"
+      else:
+        acc = " <strong>to a stop</strong>"
```
sshane, line 157: down to 0 mph?

down to a stop?

from a stop?

from 0 mph?

sshane:
@adeebshihadeh I think `from a stop` is fine, do you think it implies to resume from a stop, or they can engage at a stop?

adeebshihadeh:
Do we really need this? It seems pretty confusing.

sshane:
I guess not, since all these Toyota happen to say alc at all speeds. Just thought it would be nice to keep in that the ACC would work at all speeds (but not resume from a stop)

PR #25338, opened 2022-08-02
Remove obsolete translations

sshane:


adeebshihadeh:
Can we make CI fail in this case?

sshane:
https://github.com/commaai/openpilot/pull/25342

PR #25342, opened 2022-08-02
Multilang: test for vanished translations

sshane:


sshane:
Eh, this just adds another flag you need to run to make tests pass, and it's nice to keep vanished translations around when working with them. I'll close for now as it doesn't hurt anything keeping them around for a few commits on master

sshane:
Maybe this is only useful for me...

PR #25348, opened 2022-08-03
Toyota: Send ACC_HUD immediately on fcw

cydia2020:
There's a noticeable delay between when fcw triggers on openpilot and when the car starts to display it. `ACC_HUD` should be sent immediately after fcw is triggered.

Only affects TSS-P cars with the DSU unplugged.


sshane:
Do you know what the maximum rate is the car might send this message itself in a FCW event?

cydia2020:
> Do you know what the maximum rate is the car might send this message itself in a FCW event?

Car sends this immediately in an fcw event, it can be sent at 100hz without noticeable issues on my Prius.

sshane:
Was testing sending LKAS_HUD at 100hz recently, it was fine. I actually have a harsh FCW from an old drive, let me see what the car did there

sshane:
Oh haha I see a commit that moved `create_fcw_command` from the same if block as the ui command with `send_ui`, however, `send_ui` was *not* copied to that new if statement. So this was a regression. Thanks for the PR!

PR #25350, opened 2022-08-03
vEgoCluster: fix starting from stop

sshane:
Speed would go from 0 -> 1 -> 2 -> 1 -> 2 -> 3 every time on a Hyundai, as vEgo rises before vEgoCluster does, so it jumps back down from vEgo to vEgoCluster when you get fast enough


sshane:
Look good @pd0wm?

PR #24768, opened 2022-06-07
VW PQ: Volkswagen Passat NMS

jyoung8607:
- [x] commaai/opendbc#630
- [x] Clean up LDW_1 message in opendbc
- [x] Run checksums in opendbc CC packer instead of openpilot
- [x] commaai/opendbc#633
- [x] commaai/panda#964
- [x] Fix stock ACC cancel spamming
- [x] Add support for HUD message (driver takeover request)
- [x] Read vEgoRaw from alternate source, to support NSF later (and match current Panda)
- [x] Read non-driver door and trunk states, if possible
- [x] ~~Read stock FCW/AEB states~~
- [x] Read stock LDW states and forward to SWA for camera integrations
- [x] Test route for Panda safety
- [x] Test route for openpilot
- [x] Set dashcamOnly prior to release from draft

Some of the abstraction and cleanup work (particularly CC button handling) may get done in the MQB port first.

PR #25351, opened 2022-08-03
VW PQ: Cleanup and prep for upstream

jyoung8607:
Null-effect refactor of MQB, reducing diff/installing abstractions in support of #24768.

* volkswagencan -> mqbcan in preparation for pqcan, remove mqb from function naming
* car controller parameters and car controller sender refactor for multiple vehicle CAN APIs
* create_hud_control -> create_lka_hud_control in preparation for longitudinal control, simplify parameters
* harmonize dict update method in create_acc_buttons_control with all the other control functions
* regen CARS.md with updated Passat description in preparation for Passat NMS


selfdrive/car/volkswagen/carcontroller.py
```diff
@@ -13,7 +13,8 @@ def __init__(self, dbc_name, CP, VM):
     self.apply_steer_last = 0
     self.frame = 0
 
-    self.packer_pt = CANPacker(DBC_FILES.mqb)
+    self.packer_pt = CANPacker(dbc_name)
+    self.CCP, self.CCS = MQBCarControllerParams, mqbcan
```
adeebshihadeh, line 17: break these up into two lines

selfdrive/car/volkswagen/values.py
```diff
@@ -96,12 +91,15 @@ class CAR:
...
+DBC: Dict[str, Dict[str, str]] = defaultdict(lambda: dbc_dict("vw_mqb_2010", None))
+
+
 class Footnote(Enum):
   KAMIQ = CarFootnote(
     "Not including the China market Kamiq, which is based on the (currently) unsupported PQ34 platform.",
     Column.MODEL)
   PASSAT = CarFootnote(
-    "Not including the USA/China market Passat, which is based on the (currently) unsupported PQ35/NMS platform.",
+    "Refers only to the MQB-based European B8 Passat, not including the NMS Passat in the USA/China/Mideast markets.",
```
adeebshihadeh, line 102:
```suggestion
    "Refers only to the MQB-based European B8 Passat, not the NMS Passat in the USA/China/Mideast markets.",
```

selfdrive/car/volkswagen/values.py
```diff
@@ -15,54 +15,49 @@
...
-  GRA_ACC_STEP = 3               # GRA_ACC_01 message frequency 33Hz
+  HCA_STEP = 2                   # HCA_01/HCA_1 message frequency 50Hz
+  GRA_ACC_STEP = 3               # GRA_ACC_01/GRA_Neu message frequency 33Hz
 
-  # Observed documented MQB limits: 3.00 Nm max, rate of change 5.00 Nm/sec.
-  # Limiting rate-of-change based on real-world testing and Comma's safety
-  # requirements for minimum time to lane departure.
+  # Documented lateral limits: 3.00 Nm max, rate of change 5.00 Nm/sec.
+  # MQB vs PQ maximums are shared, but rate-of-change limited differently
+  # in subclasses based on lateral accel testing and safety requirements.
```
adeebshihadeh, line 23: i'm fine with the subclass if you really prefer it, but consider just passing in carParams like the other brands. I think it's a bit more readable

selfdrive/car/volkswagen/carcontroller.py
adeebshihadeh: also note that these abstractions only happen to work because MQB and PQ are so similar. if they start to diverge more for any reason, then you'll need to develop this abstraction a bit more

jyoung8607: MQB vs PQ is the very hardest split we have, and are likely to have for the forseeable future.

We have lateral-only control of at least one [NSF](https://en.wikipedia.org/wiki/Volkswagen_Group_New_Small_Family_platform) which turns out to be indistinguishable from PQ except the wheel-speed sensor signal is missing, because (this is not a joke) some of them have no wheel speed sensors. The community A5/S5 port for MLB (pre-Flexray) looks like early/modified MQB on the wire and uses the same J533 harness. MQBevo (Golf Mk8 and friends) will look like MQB with specialsauce. MEB (ID.3/ID.4 and friends), if we can make a lateral-only port work, will look like a MQBevo variant.

That accounts for almost everything that we could conceivably drive except for MLBevo/Flexray.

selfdrive/car/volkswagen/values.py
jyoung8607: I agree it would be somewhat more clear and readable to fully divide the MQB vs PQ CarControllerParams, and will do so. However, I may revisit subclasses if it makes sense for MLB, MQBevo, or MEB down the line.

adeebshihadeh: oh, I meant the opposite. merge them and have an if/else for the MQB/PQ specific stuff

jyoung8607: The merger would look like reverting ac1f51811840415cd64dda7024a988c74f5c7aaa and then figuring out how to feed a different `STEER_DELTA_UP` for PQ into `apply_std_steer_torque_limits()`. That's what made me split CCP.

It's my preference to do it with two different CCP classes. But I'm not the only one that has to live with the code, so if you see a better path, I'm more than happy to get onboard.

adeebshihadeh: see hyundai for example https://github.com/commaai/openpilot/blob/b6910923d18af7332f781a7c83b2601fd7862e3f/selfdrive/car/hyundai/values.py#L24

jyoung8607: Ah! Yeah, I can do that. Reverting to draft till tomorrow.

jyoung8607: See if this most recent update feels more readable to you, I like it.

PR #25353, opened 2022-08-03
CarState: use class car_fingerprint

sshane:


adeebshihadeh:
These kinds of mass changes aren't great for a couple reasons. It's not clear why `self.car_fingeprrint` is preferable over `self.CP.carFingerprint`; I happen to prefer the latter. When it's not clear which pattern is preferable, the one you're trying to remove is gonna naturally pop up again over time. I would leave this as-is until there's a clearly preferable pattern to replace it with. There's more obvious cleanups to do in the car abstraction layer, such as cleaning up the CANParser signal/check convention.

PR #25354, opened 2022-08-03
GM: separate EV/ICE gas/brake lookup tables

twilsonco:
The GM port started with the Volt, a PHEV. The threshold amount of acceleration needed to engage friction brakes assumes -1m/s^2 of regen braking force available, which is correct for Volt at speed, but ICE cars don't have anywhere close to that much engine braking force.  This causes the first 1m/s^2 of desired braking to have essentially no effect in an ICE car, and then the long controller has to correct, causing harsh braking when stopping behind a lead.

This PR adds a separate gas/brake lookup table for ICE GM cars/trucks, using a guessed value of -0.1m/s^2 of engine braking force available. Should result in much smoother stops for ICE cars.


sshane:
Testing with @Verylukyguy, makes a huge difference in following the plan: https://imgur.com/a/WYXmD6m (last two are this PR)

selfdrive/car/gm/interface.py
```diff
@@ -118,6 +118,7 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, disa
       ret.steerRatio = 14.4  # end to end is 13.46
       ret.centerToFront = ret.wheelbase * 0.4
       ret.lateralTuning.pid.kf = 1.  # get_steer_feedforward_acadia()
+      ret.longitudinalActuatorDelayUpperBound = 0.5  # large delay to initially start braking
```
sshane, line 121: It was roughly 1 second in the worse cases, half that should be fine

Verylukyguy:
@Gibby Since you are the only other ICE vehicle with their fingerprint successfully upstreamed into Master, can you load each branch and test the results?  The change worked well for me.

sshane:
Some plots with these changes from @Verylukyguy from route `e38a02cae1ef1e0c|2022-08-06--20-12-04`: https://imgur.com/a/cdNTkaW

Some plots on commatwo_master from `e38a02cae1ef1e0c|2022-08-07--14-58-09`: https://imgur.com/a/J5zIUYt

selfdrive/car/gm/carcontroller.py
```diff
@@ -68,8 +68,13 @@ def update(self, CC, CS):
           self.apply_gas = self.params.MAX_ACC_REGEN
           self.apply_brake = 0
         else:
-          self.apply_gas = int(round(interp(actuators.accel, self.params.GAS_LOOKUP_BP, self.params.GAS_LOOKUP_V)))
-          self.apply_brake = int(round(interp(actuators.accel, self.params.BRAKE_LOOKUP_BP, self.params.BRAKE_LOOKUP_V)))
+          # ICE has less engine braking force, adjust thresholds to remove braking deadzone
```
sshane, line 71:
```suggestion
```

PR #25361, opened 2022-08-04
Toyota: silent disengage

### LINKED ISSUE #1173: [$300 Bounty] Toyota: cancel cruise without chime
The way openpilot cancels cruise control on most Toyotas has the side effect of a chime from the car and a momentary fault on the dash. When openpilot sends the cancellation message, it also sends a message to play a good chime to mask the fault chime that would play, so start by [removing that](https://github.com/commaai/openpilot/blob/fa62b9d3f9882ec33559ceb7377fa991be1d4fb5/selfdrive/car/toyota/toyotacan.py#L82). Once you remove the good chime, you should hear the fault chime, which we're trying to get rid of.

The Lexus IS uses a [different cancellation message](https://github.com/commaai/openpilot/blob/fa62b9d3f9882ec33559ceb7377fa991be1d4fb5/selfdrive/car/toyota/carcontroller.py#L94) than the rest of the Toyotas, so that's probably a good place to start.

$200 for a merged PR that meets this criteria:
1. no chime sound or even momentary fault display
2. reliable, i.e. always immediately cancels cruise
3. generic across all Toyotas

Feel free to open a PR once steps 1 and 2 are done to lock the bounty pending confirmation of 3.

sshane:
This removes the beeps when disengaging. It still shows a visual alert (hands on wheel) however, and I couldn't get it to mask the sound without one.

Funnily enough, the only other signal that could totally mask the faulty beep was `LANE_SWAY_WARNING` at 1 or 2 (3 not ok), so your car would tell you to take a coffee break and think things over every time you disengaged with the gas...an appropriate recommendation for someone disengaging with gas.


sshane:
@cydia2020 @ntegan1 can you guys test this out?

cydia2020:
Ok will do

selfdrive/car/toyota/carcontroller.py
```diff
@@ -63,6 +65,9 @@ def update(self, CC, CS):
     if not CC.enabled and CS.pcm_acc_status:
       pcm_cancel_cmd = 1
 
+    if not pcm_cancel_cmd and self.prev_pcm_cancel:
+      self.disengage_frame = self.frame
```
sshane, line 69: TODO: see how quickly stock cruise cancels, we might be able to get away with no prev_pcm_cancel intermediate variable if we can start our sequence as soon as we start cancelling

sshane:
@adeebshihadeh we can merge with just the toyotacan change and this should be a strict improvement. The rest in carcontroller just ensures we get rid of the bug where you sometimes get the fault chime (was always a problem afaik), and it hides the alert quicker

sshane:
This also fixes the set speed *resetting* on international Toyotas when cancelling (never knew that was a thing) according to Ale Sato.

sshane:
Unfortunately the bad fault sound plays when stopped and you press the brake. The lane sway alert reduces this (more than lda alert), but the message is a little janky on the dash. Made it play the two beep sound when stopped and you press the brake

sshane:
So this still has issues with being inconsistent when cancelling with the gas at speed, and when stopped with the brake. If anyone wants to take this over and fix all inconsistencies, the bounty is yours!

sshane:
Wanted to be reasonably sure this worked...it does!

Screenshot from 2022-09-09 01-21-31

- When pressing the brake at any speed, the silent alert is played for 10 frames longer than the cancel command (if you don't add extra messages, it will fault about 50% of the time)

- On TSS2 cars, when pressing the brake to cancel at a standstill, there is a cooldown period where you cannot re-engage until it's over and the brake pump makes a noise. When the cooldown period is over, the fault sound more often than not plays and you can re-engage again. So here we play the two beep chime which nearly always masks this (silent alert masks this much less often).

- On the TSS-P RAV4 which doesn't normally do ACC down to 0 mph, if you press the brake to cancel at a standstill you can immediately re-engage with no weird cooldown behavior, and consequently, the fault sound is masked fully by the silent alert.

selfdrive/car/toyota/carcontroller.py
```diff
@@ -132,18 +136,26 @@ def update(self, CC, CS):
...
         send_ui = True
         self.alert_active = not self.alert_active
-      elif pcm_cancel_cmd:
-        # forcing the pcm to disengage causes a bad fault sound so play a good sound instead
+
+      # forcing the pcm to disengage causes a bad fault sound so mask with a silent alert.
+      # fully get rid of the fault sound by sending for a few frames after the fault ends
+      elif self.frame - self.pcm_cancel_frame < 10:
         send_ui = True
+        # only play chime for TSS2 when at a stop and pressing brake
```
sshane, line 151:
```suggestion
        # only play chime for TSS2 when pressing brake at a stop
```

selfdrive/car/toyota/carcontroller.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, dbc_name, CP, VM):
     self.last_standstill = False
     self.standstill_req = False
     self.steer_rate_counter = 0
+    self.pcm_cancel_frame = 0
```
sshane, line 31:
```suggestion
    self.pcm_cancel_frame = -EXTRA_CANCEL_FRAMES
```

Make a global constant

selfdrive/car/toyota/carcontroller.py
```diff
@@ -132,18 +136,26 @@ def update(self, CC, CS):
...
       if ((fcw_alert or steer_alert) and not self.alert_active) or \
          (not (fcw_alert or steer_alert) and self.alert_active):
         send_ui = True
         self.alert_active = not self.alert_active
-      elif pcm_cancel_cmd:
-        # forcing the pcm to disengage causes a bad fault sound so play a good sound instead
+
+      # forcing the pcm to disengage causes a bad fault sound so mask with a silent alert.
+      # fully get rid of the fault sound by sending for a few frames after the fault ends
+      elif self.frame - self.pcm_cancel_frame < 10:
```
sshane, line 149:
```suggestion
      elif self.frame - self.pcm_cancel_frame < EXTRA_CANCEL_FRAMES:
```

cydia2020:
> My guess about the TSS2 vs TSS-P differences is that the PCM on TSS2 cars which are designed to do stop and go do not like it if they're forced to disengage at a standstill outside of the normal conditions that cancel (pressing cancel button).

Same with TSS-P full speed DRCC cars, my Prius chimes on brake as well

sshane:
So without the extra frames you get the long single chime like TSS2?

cydia2020:
> So without the extra frames you get the long single chime like TSS2?

Yes

adeebshihadeh:
Closing this for now, wasn't a strict improvement.

PR #25362, opened 2022-08-04
Chrysler: Add legacy FP for 2017 Jeep Grand Cherokee

jyoung8607:
The additional legacy FP time in 0.8.15 has knocked out several users with older Jeep Grand Cherokees. I fished the legacy FP comparison script out of git history, and found the diff against the existing legacy FP was rather extensive, so I've added this user's FP as a second entry for the same CAR.

FP generated from: `c88f65eeaee4003a|2022-08-04--15-37-16`
Tested successfully in: `c88f65eeaee4003a|2022-08-04--17-05-08`

Differential: [fp-diff.txt](https://github.com/commaai/openpilot/files/9261959/fp-diff.txt)


sshane:
Thanks for doing this PR! Was there a script to get a FP from the first few seconds of a log? If not, it'd be nice to update that fingerprint script

jyoung8607:
> Thanks for doing this PR! Strange that we're _missing_ FPs though... Must be that script missing some FP in the beginning. All that matters is we're adding his extra FPs though

Many FPs back in the day were done off live `controlsd` output on missed vehicle ident, which was very timing sensitive. This one I took from the `fingerprint_from_route` script which will bag everything seen in that 60 second segment.

PR #25363, opened 2022-08-04
Chrysler: FPv2 support for Grand Cherokee and Pacifica

jyoung8607:
~~Includes a core fingerprinting code update to consider the engine ECU optional for these models, because a lot of them speak KWP2000 instead of UDS.~~

We chose not to add the security gateway, because it isn't useful in identification or API validation, and also isn't present before MY2018. It could be removed for Ram as well.

| Make and Model | User | Route | Commit |
| --- | --- | --- | --- |
| 2017 Chrysler Pacifica | kdvlr | `111b7e3a9ff2341a\|2022-08-08--12-04-36` | ccc8e81f6cb46227575d9273e77a9281f0a06514 |
| 2017 Chrysler Pacifica | mahlzeit | `Not uploaded` | cca34d0f86bf233522089f19f8355efc25136e37 |
| 2018 Chrysler Pacifica | supa | `f3c237869e58406c\|2022-07-26--12-24-28` | 78efca1cb6ed404a04a463af4450f1065c077a26 |
| 2021 Chrysler Pacifica | randy.BC | `42ccfd732446a8e3\|2022-08-07--22-25-05` | bac74e19ca15b20f57540c46a0be0095bc3951d9 |
| 2017 Jeep Grand Cherokee | morganmastiff | `c88f65eeaee4003a\|2022-08-04--13-30-55` | bb3c3de9739d0d468db133bff1fcccc739e5a235 |
| 2017 Jeep Grand Cherokee | Vladi17 | `9dbe2f39067d71ac\|2022-08-08--15-30-40` | 18b8566e6aea4a3aca9af1f3642b6b5927b59934 |
| 2018 Jeep Grand Cherokee | 4vanetten | `cfbff54da787011a\|2022-08-07--17-35-01` | c0a9bc63ccb3e04287a0deaeb776c32fe4845efb |

Slotting the UDS data in the same CAR that it would normally match, but once Chrysler FPv1 goes away, the current seven CARs should probably be collapsed to two, detecting minimum steer speed from EPS firmware.


jyoung8607:
MY2017 doesn't have a "security gateway" as the recently onboarded Ram does, but MY2018 and forward likely will. That said, it doesn't seem like it adds value to the fingerprint and it sounds like it'll likely be removed from the port.

We aren't getting firmware from the engine ECU. It does send a useless response to the Mazda query but that's it. Data and docs point toward the older ones using KW2000 instead of UDS. Perhaps we can cobble something together like what is used for the Nissan legacy ECUs. Until then, we must NOT add engine firmware from newer UDS supporting ECUs. The engine ECU, if present in the CAR values, is [considered essential](https://github.com/commaai/openpilot/blob/088719c86c050c9a73f45f5a472a88af5e539db1/selfdrive/car/fw_versions.py#L19), so MY2017 would fail fingerprinting as soon as any are added.

We don't have the ACC radar in this set. Ram didn't need it because the DASM combines both the fwdCamera and fwdRadar roles. It would probably communicate if we knew its address, I've asked @realfast and we'll see if we can get it. Missing both the ECU and the ACC radar, we are lacking any confirmed knowledge of longitudinal control APIs and tuning values.

**TL;DR:** We need to get the ACC radar FW, and we need to fix the engine FW or choose to fully exclude it.

jyoung8607:
> **TL;DR:** We need to get the ACC radar FW

It turns out we're getting `fwdRadar` firmware fine; it's mislabeled as `fwdCamera`. 0x753 is the ACC radar on older cars, and DASM (combined ACC/LKA) on newer cars. We think the camera is probably at 0x764. I've sent an update to the branch this end user is running to try collecting LKA camera firmware, and will update this PR with results.

jyoung8607:
I was thinking we'd see UDS engine firmware show up as the cars got newer, but nothing yet. Reverted the fancy footwork in fw_versions.py in favor of a comment for the time being. Haven't seen a PacHy yet, but they shouldn't be very different.

This all seems to work. Releasing from draft, ready for review.

jyoung8607:
@j-vanetten of jvePilot fame rebased his release on 0.8.15 yesterday, and [pulled a snapshot of this PR into jvePilot](https://github.com/j-vanetten/openpilot/commits/jvePilot-release) this evening. That will put the upgraded 0.8.15 UDS code and proposed Chrysler fingerprint schema in front of the right users quickly.

selfdrive/car/chrysler/values.py
```diff
@@ -118,7 +118,101 @@ class ChryslerCarInfo(CarInfo):
   }],
 }
 
+# Don't add engine ECU FW for CARs other than Ram, even if FW is returned on newer vehicles, before resolving the
+# issue with KWP2000 ECUs described in commaai/openpilot#25363
```
adeebshihadeh, line 122: let's add a unit test for this

jyoung8607:
Unit test added, but pulled back to draft. We've gotten a Pacifica Hybrid and it's not returning transmission or ABS/ESP FW, working to find out what's going on there. `ede4a914fcc6fb6d|2022-08-11--14-55-50`

sshane:
Looks like it's not just that one dongle or platform, all hybrids don't return esp as of now on master

jyoung8607:
If you're comfortable with the state of PacHy returns from 0.8.15, and you'd rather ingest all the FW from your notebook workflow than try to reconcile it with this PR, you're welcome to close this PR. In that case it's served its purpose as proof-of-concept/demo.

sshane:
Yeah, it's at least better than we get no esp than spotty esp. I'll look at that route you posted offline to see if I can pull it out

sshane:
Looks like some of your users here have never driven on master/release so we'll want to merge your FW versions as well.

adeebshihadeh:
> Looks like it's not just that one dongle or platform, all hybrids don't return esp as of now on master, so they should fingerprint fine.

It's going to be a problem if we want to add it later. Let's just hold off on this, and get the queries right for the Hybrids this release. No reason to rush this.

sshane:
@jyoung8607 can you have some of the users try this branch to get the engine FW? https://github.com/commaai/openpilot/tree/chrysler-engine-fw

I checked in the logs, and the ESP ECU did not respond at all on the normal 0x8 rx offset (no CAN messages from `0x747 + 0x8` or (`0x747 - 0x280`)). Maybe the offset is different, or the ECU is on a whole new address? But @realfast says that he doesn't see anything different about this ECU on Hybrids, so I'm not really sure.

sshane:
@jyoung8607 can you have your users with no engine FW (blank Mazda request) try this branch? https://github.com/commaai/openpilot/pull/25503

sshane:
It would also be helpful if users with no esp FW to run `panda/examples/query_fw_versions.py` with the esp address (think the offset is -0x280)

sshane:
Some updates:

We've successfully gotten the engine FW for the Chryslers that were missing it, however transmission and esp on the hybrids are still a mystery. realfast believes that the hybrid control module might bake those two ECUs together, like some Ram HD trucks have the TCM (transmission) combined with the engine. Getting a tester to run `query_fw_versions` this weekend for us. We should know what to do by then!

jyoung8607:
Closing this for now. Chrysler FPv2 is clearly possible, and I believe the issue with engine ECU queries was solved in another PR somewhere. This particular PR is just outdated.

wangstein:
Thank you! I actually been using the master branch

From: Jason Young ***@***.***>
Date: Tuesday, March 14, 2023 at 12:07 PM
To: commaai/openpilot ***@***.***>
Cc: Wenni Wang ***@***.***>, Manual ***@***.***>
Subject: Re: [commaai/openpilot] Chrysler: FPv2 support for Grand Cherokee and Pacifica (PR #25363)

Closing this for now. Chrysler FPv2 is clearly possible, and I believe the issue with engine ECU queries was solved in another PR somewhere. This particular PR is just outdated.

PR #25367, opened 2022-08-04
Add instructions for running `ui` replays on environments with non-working hardware rendering setups such as WSL2

nelsonjchen:


nelsonjchen:
ez link to the strange flap:

https://github.com/commaai/openpilot/runs/7681467587?check_suite_focus=true

adeebshihadeh:
This doesn't seem openpilot related at all? Is this just a WSL2 thing?

nelsonjchen:
Always happy to make instructions that enable a greater pool of users to run openpilot tooling. But alas, this is closed. Anyway, hopefully someone can find this on Google then.

sshane:
Nice find! This works perfectly on my WSL2 Windows 11

PR #25370, opened 2022-08-05
plotjuggler: route name in window title

GregorKikelj:
Requires https://github.com/commaai/PlotJuggler/pull/50

### DIFF
--- a/tools/plotjuggler/juggle.py
+++ b/tools/plotjuggler/juggle.py
@@ -78,6 +78,7 @@ def start_juggler(fn=None, dbc=None, layout=None, route_or_segment_name=None):
     extra_args += f" -l {layout}"
   if route_or_segment_name is not None:
     extra_args += f" --window_title \"{route_or_segment_name}\""
+
   cmd = f'{PLOTJUGGLER_BIN} --plugin_folders {INSTALL_DIR}{extra_args}'
   subprocess.call(cmd, shell=True, env=env, cwd=juggle_dir)
 
--- a/tools/plotjuggler/juggle.py
+++ b/tools/plotjuggler/juggle.py
@@ -57,7 +57,7 @@ def load_segment(segment_name):
     return []
 
 
-def start_juggler(fn=None, dbc=None, layout=None):
+def start_juggler(fn=None, dbc=None, layout=None, route_or_segment_name=None):
   env = os.environ.copy()
   env["BASEDIR"] = BASEDIR
   env["PATH"] = f"{INSTALL_DIR}:{os.getenv('PATH', '')}"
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Do old versions of PlotJuggler crash or just ignore the new option? If it crashes, we'll have to detect this.

sshane:
Older versions refuse to start: `Unknown option 'window_title'.`

GregorKikelj:
I can run ./plotjuggler -h and check for the option being listed, if not auto reinstall or emit a warning

sshane:
Maybe we can check PJ version? Assert a required or at least as high as version?

adeebshihadeh:
Let's add a minimum version check and show a prompt when the minimum version isn't met.

GregorKikelj:
Not sure how to check the version

adeebshihadeh:
> Not sure how to check the version

`./plotjuggler -v`

GregorKikelj:
Hmm ok not sure why I completely missed it

tools/plotjuggler/juggle.py
```diff
@@ -46,6 +47,13 @@ def install():
       tar.extractall(path=INSTALL_DIR)
 
 
+def check_plotjuggler_version():
+  proc = subprocess.Popen([PLOTJUGGLER_BIN, "-v"], stdout=subprocess.PIPE)
+  (out, _) = proc.communicate()
+  version = out.decode("utf-8").split(" ")[1]
+  return version >= MINIMUM_PLOTJUGGLER_VERSION
```
adeebshihadeh, line 54: these are strings. it'll work properly if you split into a tuple

GregorKikelj: Works as strings as well, but can do tuples if prefered.

sshane: Not always:

```
"0.8.9" > "0.8.10"
Out[126]: True
```

tools/plotjuggler/juggle.py
```diff
@@ -160,6 +169,9 @@ def juggle_route(route_or_segment_name, segment_count, qlog, can, layout, dbc=No
     print("PlotJuggler is missing. Downloading...")
     install()
 
+  if get_plotjuggler_version() < MINIMUM_PLOTJUGGLER_VERSION:
+    raise Exception(f"PlotJuggler version is too old. Update with './juggle.py --install'")
```
adeebshihadeh, line 173: should offer to update with a prompt

adeebshihadeh: on second thought, better to just install. that's what we do when it's missing

PR #25377, opened 2022-08-06
FW dump from last 2 weeks

sshane:
Adds FW from fuzzy fingerprinting routes (from that brand only). Removes FW from fuzzy fingerprinting routes where brand isn't fingerprinted car.


sshane:
It seems the server openpilot was behind a bit, will need to re-run with it pulled

PR #25393, opened 2022-08-09
GM: resolve resume-related ACC faults

sshane:
Fixes: ACC faults with GM when you enable via resume button without first enabling with set button.

Behavior changes:
- On all non pcmCruise cars, pressing the main button to disable cruise will now reset the set speed in the openpilot UI just like pcmCruise cars (make sure). Think this is a good change to unify this behavior
- On GM, a `buttonEnable` event will not be created when you press resume to enable until you engage at least once with the set button

### DIFF
--- a/selfdrive/car/gm/interface.py
+++ b/selfdrive/car/gm/interface.py
@@ -178,7 +178,7 @@ class CarInterface(CarInterfaceBase):
     if ret.vEgo < self.CP.minSteerSpeed:
       events.add(car.CarEvent.EventName.belowSteerSpeed)
 
-    # The ECM will fault if resume triggers an enable while speed is unset (unset is greater than 70 m/s)
+    # The ECM will fault if resume triggers an enable while speed is unset
     if c.hudControl.setSpeed > 70:
       events.add(car.CarEvent.EventName.resumeBlocked)
 
--- a/selfdrive/car/gm/interface.py
+++ b/selfdrive/car/gm/interface.py
@@ -161,9 +161,13 @@ class CarInterface(CarInterfaceBase):
     if self.CS.cruise_buttons != self.CS.prev_cruise_buttons and self.CS.prev_cruise_buttons != CruiseButtons.INIT:
       be = create_button_event(self.CS.cruise_buttons, self.CS.prev_cruise_buttons, BUTTONS_DICT, CruiseButtons.UNPRESS)
 
-      # Suppress resume button if we're resuming from stop so we don't adjust speed.
-      if be.type == ButtonType.accelCruise and (ret.cruiseState.enabled and ret.standstill):
-        be.type = ButtonType.unknown
+      if be.type == ButtonType.accelCruise:
+        # Suppress resume button if we're resuming from stop so we don't adjust speed.
+        if ret.cruiseState.enabled and ret.standstill:
+          be.type = ButtonType.unknown
+        # The ECM will fault if resume triggers an enable while speed is unset (unset is greater than 70 m/s)
+        elif c.hudControl.setSpeed < 70:
+          be.type = ButtonType.unknown
 
       ret.buttonEvents = [be]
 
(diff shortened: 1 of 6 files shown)


selfdrive/car/gm/interface.py
```diff
@@ -179,7 +179,9 @@ def _update(self, c):
       events.add(car.CarEvent.EventName.belowSteerSpeed)
 
     # handle button presses
-    events.events.extend(create_button_enable_events(ret.buttonEvents, pcm_cruise=self.CP.pcmCruise))
+    # The ECM will fault if resume triggers an enable while speed is unset (unset is greater than 70 m/s)
+    resume_enabled = c.hudControl.setSpeed < 70
+    events.events.extend(create_button_enable_events(ret.buttonEvents, pcm_cruise=self.CP.pcmCruise, resume_enabled=resume_enabled))
```
sshane, line 184: Would like to check == 255, but converting 255 to ms to kph != 255

selfdrive/controls/controlsd.py
```diff
@@ -454,9 +454,13 @@ def state_transition(self, CS):
...
-      self.v_cruise_kph = update_v_cruise(self.v_cruise_kph, CS.vEgo, CS.gasPressed, CS.buttonEvents,
-                                          self.button_timers, self.enabled, self.is_metric)
-      self.v_cruise_cluster_kph = self.v_cruise_kph
+      if CS.cruiseState.available:
+        self.v_cruise_kph = update_v_cruise(self.v_cruise_kph, CS.vEgo, CS.gasPressed, CS.buttonEvents,
+                                            self.button_timers, self.enabled, self.is_metric)
+        self.v_cruise_cluster_kph = self.v_cruise_kph
+      else:
+        self.v_cruise_kph = V_CRUISE_INITIAL
+        self.v_cruise_cluster_kph = V_CRUISE_INITIAL
```
sshane, line 463: Try to clean this up with pcmCruise logic below, can we set it to 255 for pcmCruise cars?

adeebshihadeh:
Let's also show an alert when a resume press is blocked.

sshane:
Was thinking of doing that. I could see either way being okay though, as the stock systems usually all just ignore presses when the main is off, or trying to press resume when first enabling. I'll see how a no entry feels

sshane:
Here's @Verylukyguy trying to resume with no set speed, it correctly ignores the button and doesn't engage/fault.

Screenshot from 2022-08-09 19-29-57

sshane:
- [ ] Need new process replay route

selfdrive/controls/controlsd.py
```diff
@@ -454,9 +454,13 @@ def state_transition(self, CS):
 
     # if stock cruise is completely disabled, then we can use our own set speed logic
     if not self.CP.pcmCruise:
-      self.v_cruise_kph = update_v_cruise(self.v_cruise_kph, CS.vEgo, CS.gasPressed, CS.buttonEvents,
-                                          self.button_timers, self.enabled, self.is_metric)
-      self.v_cruise_cluster_kph = self.v_cruise_kph
+      if CS.cruiseState.available:
```
adeebshihadeh, line 457: does this change cars where we don't need to set first, like Honda Nidec? I'd prefer to leave those as is

selfdrive/controls/lib/events.py
```diff
@@ -635,6 +635,10 @@ def joystick_alert(CP: car.CarParams, CS: car.CarState, sm: messaging.SubMaster,
     ET.NO_ENTRY: wrong_car_mode_alert,
   },
 
+  EventName.resumeBlocked: {
+    ET.NO_ENTRY: NoEntryAlert("Press Set to Engage"),
+  },
```
sshane, line 640: This seemed fine on the Honda, it's very similar to our current no entry alerts (main is off, gear not D, etc)

selfdrive/controls/controlsd.py
sshane: We can try to have the car interfaces handle all the button logic and add a vCruise to CarState. For now added an exception until we have time for this clean up unless you want to do this now in another PR

sshane:
Did this for all cars generically here: https://github.com/commaai/openpilot/pull/25402

PR #25394, opened 2022-08-10
controls: cleanup default set speed

sshane:
Think the 255 default was added for a specific car to send as the default. No reason it should be 255 now, should be handled in the car ports


selfdrive/controls/controlsd.py
```diff
@@ -162,8 +162,8 @@ def __init__(self, sm=None, pm=None, can_sock=None, CI=None):
     self.active = False
     self.can_rcv_error = False
     self.soft_disable_timer = 0
-    self.v_cruise_kph = 255
-    self.v_cruise_cluster_kph = 255
+    self.v_cruise_kph = 0
+    self.v_cruise_cluster_kph = 0
```
sshane, line 166: Need to make sure all the carcontrollers set the correct values when we haven't got a set speed yet, then should be good to merge.

selfdrive/controls/lib/drive_helpers.py
```diff
@@ -95,8 +95,8 @@ def update_v_cruise(v_cruise_kph, v_ego, gas_pressed, buttonEvents, button_timer
 
 def initialize_v_cruise(v_ego, buttonEvents, v_cruise_last):
   for b in buttonEvents:
-    # 250kph or above probably means we never had a set speed
-    if b.type in (ButtonType.accelCruise, ButtonType.resumeCruise) and v_cruise_last < 250:
+    # 0kph means we never had a set speed
+    if b.type in (ButtonType.accelCruise, ButtonType.resumeCruise) and v_cruise_last != 0:
```
adeebshihadeh, line 99: I don't think 0 is a much better magic value

adeebshihadeh:
Looks like this is done in the resume PR?

selfdrive/controls/lib/drive_helpers.py
sshane: Well I liked 0 because we don't have to "estimate" what 255 converted to m/s would be in the car controllers (> 70?) as we only have access to the converted value.

selfdrive/controls/lib/longitudinal_planner.py
```diff
@@ -63,7 +63,7 @@ def update(self, sm):
     v_ego = sm['carState'].vEgo
 
     v_cruise_kph = sm['controlsState'].vCruise
-    v_cruise_kph = min(v_cruise_kph, V_CRUISE_MAX)
+    v_cruise_kph = max(min(v_cruise_kph, V_CRUISE_MAX), V_CRUISE_MIN)
```
sshane, line 66: I wonder if we're always giving a small amount of acceleration on non-pcm cars on first emable (now would be brake)

PR #25402, opened 2022-08-11
controls: block resume if cruise not previously engaged

### LINKED ISSUE #24238: GM: RSOD Cruise Faults when using the Resume button to engage
### Describe the bug

I was getting RSOD Cruise Faults if I used the Resume button to engage the cruise control/OpenPilot when the previous set speed was below the current driving speed.

comma two

GMC Acadia 2018

e38a02cae1ef1e0c|2022-04-17--20-21-26--0

### openpilot version

Master 0.8.14 @ 1fa1c44

### Additional info

May be related to https://github.com/commaai/openpilot/issues/24223

sshane:
This fixes https://github.com/commaai/openpilot/issues/24238 and should fix a large portion of the faults in https://github.com/commaai/openpilot/issues/24223


sshane:
Hmm, this might require panda changes. panda doesn't know to not allow controls in this new case

selfdrive/controls/controlsd.py
```diff
@@ -218,6 +219,10 @@ def update_events(self, CS):
       self.events.add(EventName.controlsInitializing)
       return
 
+    # Block resume if cruise never previously enabled
+    if not self.CP.pcmCruise and self.v_cruise_kph == V_CRUISE_INITIAL:
+      self.events.add(EventName.resumeBlocked)
+
```
sshane, line 225: @adeebshihadeh should this go in SIMULATION or should we get a new process replay segment where set is used first?

adeebshihadeh: think it makes sense to set it here https://github.com/commaai/openpilot/blob/3fc001c945bb5e9092050bac1e55267dab63b9eb/selfdrive/controls/controlsd.py#L422

sshane: The route doesn't start engaged though, they enable with resume 10 seconds in. I don't want to skip the state machine by checking this all the time

common/params.cc
```diff
@@ -93,6 +93,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"CarVin", CLEAR_ON_MANAGER_START | CLEAR_ON_IGNITION_ON},
     {"CompletedTrainingVersion", PERSISTENT},
     {"ControlsReady", CLEAR_ON_MANAGER_START | CLEAR_ON_IGNITION_ON},
+    {"ControlsState", CLEAR_ON_MANAGER_START | CLEAR_ON_IGNITION_ON},
```
sshane, line 96: Python doesn't allow you to set environment variables to bytes, had to make a param

selfdrive/controls/controlsd.py
```diff
@@ -462,9 +466,13 @@ def state_transition(self, CS):
...
-      self.v_cruise_kph = update_v_cruise(self.v_cruise_kph, CS.vEgo, CS.gasPressed, CS.buttonEvents,
-                                          self.button_timers, self.enabled, self.is_metric)
-      self.v_cruise_cluster_kph = self.v_cruise_kph
+      if CS.cruiseState.available:
+        self.v_cruise_kph = update_v_cruise(self.v_cruise_kph, CS.vEgo, CS.gasPressed, CS.buttonEvents,
+                                            self.button_timers, self.enabled, self.is_metric)
+        self.v_cruise_cluster_kph = self.v_cruise_kph
+      else:
+        self.v_cruise_kph = V_CRUISE_INITIAL
+        self.v_cruise_cluster_kph = V_CRUISE_INITIAL
```
adeebshihadeh, line 475: same if/else is done below, but v_cruise is set to 0 instead of V_CRUISE_INITIAL. seems like we can simplify this

adeebshihadeh (approved):

sshane:
Only diff for TOYOTA:

```
('change', 'controlsState.vCruise', (0.0, 255.0))
('change', 'controlsState.vCruiseCluster', (0.0, 255.0))
('change', 'carControl.hudControl.setSpeed', (0.0, 70.83333587646484))
```

sshane:
Had @Verylukyguy test going into P and N and then resuming, it did not fault

PR #25404, opened 2022-08-11
Use steeringAngleDeg for angle control saturation warnings

robbederks:
Segment in which it should have triggered: `82a65289fdc612b7|2022-08-09--16-40-29--62`


HaraldSchafer:
LGTM if you're sure about the signs

PR #25406, opened 2022-08-11
Subaru: new harness is on the shop

adeebshihadeh:


vanillagorillaa:
can you also change subaru to subaru_a?

PR #25408, opened 2022-08-11
Subaru: resume from cruise standstill

adeebshihadeh:


sshane:
You can remove the Subaru exception from docs_definitions for auto-resume now

PR #25415, opened 2022-08-12
Car docs: show more information about cars

### LINKED ISSUE #25333: Car docs: show more information about cars
- [x] Show actual steer and enable speeds
- [x] Remove (useless) stars
- [ ] Show extra information (hidden by default or all the time), like the detail sentence
- [ ] Clean up diff bot to use the lines instead
- [x] Add auto-resume from stop star - https://github.com/commaai/openpilot/pull/25572
- [ ] Link harness to shop and auto-select

sshane:
Reference, old car docs: https://github.com/commaai/openpilot/blob/62bb43b7e279d81d03d1023ea24dc1f845fc3608/docs/CARS.md

Preview: https://github.com/commaai/openpilot/blob/bb-stars/docs/CARS.md


selfdrive/car/CARS_template.md
```diff
@@ -1,30 +1,17 @@
 {% set footnote_tag = '[<sup>{}</sup>](#footnotes)' -%}
-{% set star_icon = '[![star](assets/icon-star-{}.svg)](##)' -%}
 
 <!--- AUTOGENERATED FROM selfdrive/car/CARS_template.md, DO NOT EDIT. --->
 
 # Supported Cars
 
 A supported vehicle is one that just works when you install a comma device. Every car performs differently with openpilot, but all supported cars should provide a better experience than any stock system.
 
-## How We Rate The Cars
```
sshane, line 10: Maybe a short hand-written description for each would be good. For example what steering torque units are, and what it means/is good. What you get with openpilot ACC vs. Stock ACC

adeebshihadeh:
Let's also add a column for the harness name and ideally link to it on the shop.

selfdrive/car/CARS_template.md
```diff
@@ -1,30 +1,17 @@
...
-{% endfor %}
-
-{% endfor %}
-{% endfor %}
-
 # {{all_car_info | length}} Supported Cars
 
 |{{Column | map(attribute='value') | join('|')}}|
-|---|---|---|:---:|:---:|:---:|:---:|
+|---|---|---|:---:|:---:|:---:|:---:|:---:|
```
adeebshihadeh, line 12: this should be done automatically

adeebshihadeh:
Any blocker on this? Looks pretty good to me.

sshane:
Still want the torque stars? If not I'll clean up

adeebshihadeh:
Let's do a star for now.

adeebshihadeh (approved):
Looks great. Let's also add resume from stop at some point; then, we should have all the info the website shows.

PR #25417, opened 2022-08-12
Revert "FPv2: fingerprint on all FW combinations"

sshane:
Reverts the inter-brand fingerprinting part of commaai/openpilot#25204, but still keeps around fingerprinting on multiple FW versions for an ECU from the same brand.

### DIFF
--- a/selfdrive/car/fw_versions.py
+++ b/selfdrive/car/fw_versions.py
@@ -282,6 +282,7 @@ def match_fw_to_car_fuzzy(fw_versions_dict, log=True, exclude=None):
   match_count = 0
   candidate = None
   for addr, versions in fw_versions_dict.items():
+    # Some ECUs respond multiple times to a brand's queries, try matching on all
     for version in versions:
       # All cars that have this FW response on the specified address
       candidates = all_fw_versions[(addr[0], addr[1], version)]
--- a/selfdrive/car/fw_versions.py
+++ b/selfdrive/car/fw_versions.py
@@ -416,9 +416,8 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, debug=Fa
   for brand in sorted(brand_matches, key=lambda b: len(brand_matches[b]), reverse=True):
     car_fw = get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, debug=debug, progress=progress)
     all_car_fw.extend(car_fw)
-
-    # TODO: Until erroneous FW versions are removed, try to fingerprint on all possible combinations so far
-    _, matches = match_fw_to_car(all_car_fw, allow_fuzzy=False)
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Any reason this is a draft?

selfdrive/car/fw_versions.py
```diff
@@ -416,9 +421,8 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, debug=Fa
   for brand in sorted(brand_matches, key=lambda b: len(brand_matches[b]), reverse=True):
     car_fw = get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, debug=debug, progress=progress)
     all_car_fw.extend(car_fw)
-
-    # TODO: Until erroneous FW versions are removed, try to fingerprint on all possible combinations so far
-    _, matches = match_fw_to_car(all_car_fw, allow_fuzzy=False)
+    # Try to match using FW returned from this brand only
+    matches = match_fw_to_car_exact(build_fw_dict(car_fw))
```
sshane, line 425: Tries to match on all FW so far -> matches on FW returned from this branch

selfdrive/car/fw_versions.py
```diff
@@ -337,17 +338,21 @@ def match_fw_to_car_exact(fw_versions_dict):
...
+  brands = get_interface_attr('FW_VERSIONS', ignore_none=True).keys()
   for exact_match, match_func in exact_matches:
-    # TODO: For each brand, attempt to fingerprint using only FW returned from its queries
+    # For each brand, attempt to fingerprint using all FW returned from its queries
     matches = set()
-    fw_versions_dict = build_fw_dict(fw_versions, filter_brand=None)
-    matches |= match_func(fw_versions_dict)
+    for brand in brands:
+      fw_versions_dict = build_fw_dict(fw_versions, filter_brand=brand)
+      matches |= match_func(fw_versions_dict)
```
sshane, line 355: This is responsible for going from matching using all FW, to filtering FW down to each brand's queries and fingerprinting solely on that.

sshane:
There were some good improvements baked into that PR so I only reverted what we need to ensure we only fingerprint using a specific brand's returned versions.

The main improvement this leaves in is making sure we fingerprint on all FW returned from the same ECU using multiple different brand queries (Hyundai is guilty of this).

selfdrive/car/fw_versions.py
```diff
@@ -282,6 +282,7 @@ def match_fw_to_car_fuzzy(fw_versions_dict, log=True, exclude=None):
   match_count = 0
   candidate = None
   for addr, versions in fw_versions_dict.items():
+    # Some ECUs respond multiple times to a brand's queries, try matching on all
```
sshane, line 285:
```suggestion
```

PR #25421, opened 2022-08-12
Car docs: fix required package for Bolt EUV 2022-23

sshane:
Only the Premier trim comes with standard ACC (including the 2023 model), 2LT trim comes with available ACC. https://www.chevrolet.com/electric/bolt-euv


selfdrive/car/gm/values.py
```diff
@@ -83,7 +83,7 @@ class GMCarInfo(CarInfo):
   CAR.ACADIA: GMCarInfo("GMC Acadia 2018", video_link="https://www.youtube.com/watch?v=0ZN6DdsBUZo"),
   CAR.BUICK_REGAL: GMCarInfo("Buick Regal Essence 2018"),
   CAR.ESCALADE_ESV: GMCarInfo("Cadillac Escalade ESV 2016", "Adaptive Cruise Control (ACC) & LKAS"),
-  CAR.BOLT_EUV: GMCarInfo("Chevrolet Bolt EUV 2022-23", "Adaptive Cruise Control (ACC)", footnotes=[], harness=Harness.gm),
+  CAR.BOLT_EUV: GMCarInfo("Chevrolet Bolt EUV 2022-23", "Premier Trim", footnotes=[], harness=Harness.gm),
```
adeebshihadeh, line 86: or premier redline on the 2023

sshane:
We might want new copy for trims on the website. This is how it would appear now:

`openpilot requires the car to come equipped with Premier Trim.`

which seems a bit clunky

PR #25422, opened 2022-08-13
GM: move set speed scaling to DBC

sshane:


selfdrive/car/gm/gmcan.py
```diff
@@ -63,8 +63,7 @@ def create_friction_brake_command(packer, bus, apply_brake, idx, near_stop, at_f
   return packer.make_can_msg("EBCMFrictionBrakeCmd", bus, values)
 
 def create_acc_dashboard_command(packer, bus, acc_engaged, target_speed_kph, lead_car_in_sight, fcw):
-  # Not a bit shift, dash can round up based on low 4 bits.
-  target_speed = int(target_speed_kph * 16) & 0xfff
```
sshane, line 67: This basically was just (`(kph * 16) % 4096)`

sshane:
Waiting on a tester and should be good

twilsonco:
I performed a test on my C3 as follows:
1. engage and set speed to 5mph
2. Pull behind parked car and as op slows, increase set speed to 90mph (I stopped at 85, but same thing). op will still slow and stop behind parked lead
3. Once stopped, you can use the set/decel button to lower speed 1mph at a time and confirm visually that the dash speed matches the onscreen op set speed.

The test showed the numbers agree visually for all speeds between the dash and on-screen set speed, and plots show the same:
image

PR #25429, opened 2022-08-13
GM: Chevy Silverado 2020-21 support

JMPZ11:
**Checklist**
- [x] route with openpilot: 38aa7da107d5d252|2022-08-15--16-01-12


sshane:
Number of `ACCButtons` messages sent at 10 Hz before cruise canceled from gas press: 4, 3, 5, 2, 5, 7, 3, 3, 7, 4, 4, 2.

adeebshihadeh:
> Number of `ACCButtons` messages sent at 10 Hz before cruise canceled from gas press: `4, 3, 5, 2, 5, 7, 3, 3, 7, 4, 4, 2` from a test route. While not always immediate, we can experiment with making it more consistent after merging. The cruise so far has always canceled without fault.

Let's improve this before merging. Some things to try: sending counter, sending checksum, sending multiple messages.

selfdrive/car/gm/gmcan.py
```diff
@@ -1,7 +1,10 @@
 from selfdrive.car import make_can_msg
 
-def create_buttons(packer, bus, button):
-  values = {"ACCButtons": button}
+def create_buttons(packer, bus, idx, button):
+  values = {
+    "ACCButtons": button,
+    # "RollingCounter": idx,
```
sshane, line 6:
```suggestion
    "RollingCounter": idx,
```

selfdrive/car/gm/carcontroller.py
```diff
@@ -112,7 +112,8 @@ def update(self, CC, CS):
       if (self.frame - self.last_button_frame) * DT_CTRL > 0.1:
```
sshane, line 112:
```suggestion
      if (self.frame - self.last_button_frame) * DT_CTRL > 0.04:
```

selfdrive/car/gm/carcontroller.py
```diff
@@ -112,7 +112,8 @@ def update(self, CC, CS):
...
         if CC.cruiseControl.cancel:
           self.last_button_frame = self.frame
-          can_sends.append(gmcan.create_buttons(self.packer_pt, CanBus.CAMERA, CruiseButtons.CANCEL))
+          idx = (CS.buttons_counter + 0) % 4
+          can_sends.append(gmcan.create_buttons(self.packer_pt, CanBus.CAMERA, idx, CruiseButtons.CANCEL))
```
sshane, line 116:
```suggestion
          can_sends.append(gmcan.create_buttons(self.packer_pt, CanBus.CAMERA, CS.buttons_counter, CruiseButtons.CANCEL))
```

selfdrive/car/gm/values.py
```diff
@@ -84,6 +85,9 @@ class GMCarInfo(CarInfo):
   CAR.BUICK_REGAL: GMCarInfo("Buick Regal Essence 2018"),
   CAR.ESCALADE_ESV: GMCarInfo("Cadillac Escalade ESV 2016", "Adaptive Cruise Control (ACC) & LKAS"),
   CAR.BOLT_EUV: GMCarInfo("Chevrolet Bolt EUV 2022-23", "Premier/Premier Redline Trim", video_link="https://youtu.be/xvwzGMUA210", footnotes=[], harness=Harness.gm),
+  # TEMP NOTE: ACC is not standard, only optional on LTZ and High Country Trims if you buy "Safety Package II"
+  # https://cdn.dealereprocess.org/cdn/brochures/chevrolet/2020-silverado1500.pdf
```
adeebshihadeh, line 89: what's the note here for? Safety Package II seems accurate

adeebshihadeh: hmm, the 2022 Silverado ACC package is different. perhaps we should just use "Adaptive Cruise Control" instead. @JasonJShuler do you expect the 2022 silverado will be the same platform or require a new port?

sshane: @JasonJShuler mentioned that the Suburban/Tahoe went Global B in 2022, but not the Silverado. So 2023 likely might need to be a new platform

sshane: Actually I think we support the 2022 as well. From @JasonJShuler: `Yes - If the 2023 is out I think it's still included... the Suburban / Tahoe went Global B in 2022, but no the silverado`

sshane: 2022 Silverado ACC packages: https://www.chevrolet.com/content/dam/chevrolet/na/us/english/index/shopping-tools/download-catalog/09-pdf/2022-chevrolet-new-silverado-1500-ebrochure.pdf

PR #25436, opened 2022-08-14
Lateral planner: Don't prompt/initiate lane changes if lateral control is not active

csouers:
**Description**: Gate lane changes on `latActive` instead of the generic `active` state. Do not initiate lane changes until lateral control becomes active and a lane change has been (re-)initiated by the user.

Cars with steering capabilities limited to that of their factory LKAS (i.e. 2021-22 Honda Odyssey using Stock ACC and 2022 Mazda CX5) are the primary targets for this change. Lane change events are of higher priority than the `steerTempUnavailableSilent` event so the overriding event can communicate to the user that lateral control is active when it is not.

**Verification** Route replay and real world testing on Honda Odyssey 2021+ and Acura RDX 2022 with radar firmware `b'36802-TJB-A540\x00\x00'`.

Todo:
- [ ] Generate & upload new test log data
- [ ] Update ref_commit

### DIFF
--- a/selfdrive/controls/lib/desire_helper.py
+++ b/selfdrive/controls/lib/desire_helper.py
@@ -40,12 +40,12 @@ class DesireHelper:
     self.prev_one_blinker = False
     self.desire = log.LateralPlan.Desire.none
 
-  def update(self, carstate, active, lane_change_prob):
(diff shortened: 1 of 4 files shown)


adeebshihadeh:
Nice catch!

PR #25442, opened 2022-08-15
UI: Device Settings Spacing

Achilles308:
This change crunches the Device settings space between each row so that the Reboot and Shut Off button are always visible (only true if the disable radar option is not shown though).

### DIFF
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -96,7 +96,7 @@ TogglesPanel::TogglesPanel(SettingsWindow *parent) : ListWidget(parent) {
 }
 
 DevicePanel::DevicePanel(SettingsWindow *parent) : ListWidget(parent) {
-  setSpacing(50);
+  setSpacing(7);
   addItem(new LabelControl(tr("Dongle ID"), getDongleId().value_or(tr("N/A"))));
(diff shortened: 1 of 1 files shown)


sshane:
Think it would be better if Reboot and Power Off were on a persistent layout on top of everything else that didn't scroll. But this is a design change, and I think we want to tackle that as a larger, more intentional project.

Achilles308:
> Think it would be better if Reboot and Power Off were on a persistent layout on top of everything else that didn't scroll. But this is a design change, and I think we want to tackle that as a larger, more intentional project.

I would agree persistent buttons would be better.

However, this is a decent compromise since the change language option was added and bumped everything down. Some new users may not even know they exist as there is no scroll wheel to indicate further options are hidden (and important options at that!).

CruelSenpai:
Hold down the Comma logo on the offroad sidebar for 5 secs and it brings up an overlay with reboot and power off options. Then just remove these buttons from Device section of Settings altogether.

adeebshihadeh:
@sshane is right. We want a nicer UI as much as you do, but you don't get there with these random local changes.

Achilles308:
> @sshane is right. We want a nicer UI as much as you do, but you don't get there with these random local changes.

Can you at least add the scroll bar then? So it's clear there are additional rows hidden from view? This is true on the settings page and well as the Software page.

PR #25445, opened 2022-08-16
Common button enable/cancel event creation

sshane:
Any car that sets button events and if not pcmCruise we send a button enable event. If pcmCruise, then we just cancel when cancel is pressed


sshane:
@robbederks the only diff in behavior for this PR is we now send a `buttonCancel` for Teslas when pressing the cancel button, regardless of pcmCruise, like all the other cars we have buttons for. Is this okay?

robbederks:
> @robbederks the only diff in behavior for this PR is we now send a `buttonCancel` for Teslas when pressing the cancel button, regardless of pcmCruise, like all the other cars we have buttons for. Is this okay?

Should be fine!

selfdrive/car/volkswagen/interface.py
```diff
@@ -209,7 +208,7 @@ def _update(self, c):
     ret = self.CS.update(self.cp, self.cp_cam, self.cp_ext, self.CP.transmissionType)
 
     events = self.create_common_events(ret, extra_gears=[GearShifter.eco, GearShifter.sport, GearShifter.manumatic],
-                                       pcm_enable=not self.CS.CP.openpilotLongitudinalControl)
+                                       not self.CS.CP.openpilotLongitudinalControl)
```
sshane, line 211:
```suggestion
                                       pcm_enable=not self.CS.CP.openpilotLongitudinalControl)
```

PR #25453, opened 2022-08-16
GM: dynamic regen acceleration threshold

### LINKED ISSUE #28680: GM sends nonzero gas while preEnabled
### Describe the bug

Chevrolet Volt 2018

### openpilot version

0.9.2

### Additional info

twilsonco:
# Description

GM Chevy Volt has bad longitudinal oscillations when stopping behind a lead. This is partially caused the the threshold value of desired acceleration necessary to actuate friction brakes. 
image
The value of -1m/s^2 reflects the amount of  regen braking force available at city and highway speeds. At lower speeds, however, this value is too high, so when openpilot requests -0.9m/s^2 of acceleration at low speeds, it will not actuate the friction brakes, only lowering the gas command which does not yield the necessary regen at those speeds, causing the long controller integral gain to wind up and correct, delivering harsh, sudden braking, quickly overshooting the desired acceleration and letting off the brakes. This repeats, creating uncomfortable oscillations visible in the image in the verification section below (left side, red path "aEgo").

Some additional context: this was previously found to be an issue for GM ICE cars, which generate far less engine braking force compared to EV regen braking. The solution was to use a threshold gas/brake acceleration value closer to 0m/s^2, implemented by @sshane and myself.

The solution here is to use a speed-based, dynamic gas/brake acceleration threshold based on measurements of the Chevy Volt regen braking force on a flat, smooth road. I performed tests to determine this profile and derived a lookup table used to control the threshold gas/brake acceleration value while on-road.

## Gathering data

To understand how much regen is accessible at different speeds, I drove at 2am to conduct some tests where I drive at highway speeds (with openpilot disengaged) and then let off the gas and allow regen to slow me down as much as it will, to a "stop".

Here are plots of data collected from two runs on the same stretch of fairly flat, smooth freeway, once in each direction.
regen profiles

Here are the two runs overlaid. The pitch-adjusted acceleration is slightly different, but very close. I suspect this is due to an internal controller in the car that is targeting it's own regen braking profile.
regen profiles overlaid w points

## Gas/brake threshold profile

Based on values of vEgo and aEgo at the characteristic points in the aEgo vs time plot indicated above, I derived the following profile, where the middle line is the average between the two runs:
image

While this clearly changes with vehicle pitch, that will be accounted for by [GM longitudinal pitch-compensation](https://github.com/commaai/openpilot/pull/25485).

The positive portion of the plot corresponds to creep acceleration, which necessitates some amount of non-zero brake command in order to maintain velocities below ~1.2m/s. This was determined by, on a flat smooth surface, releasing the brake when stopped in gear and allowing the car to creep forward. The positive acceleration threshold at low speed creates a small delay when resuming from a stop on an incline, but this is only slightly slower to start than before. I've also confirmed this is alleviated by longitudinal pitch-compensation.

Here are the values of the average profile:
```
speed [mph]	regen brake "force" [m/s^2]
1.12		0.68
1.34		0.67
1.57		0.63
1.79		0.56
2.02		0.49
2.24		0.38
2.46		0.30
2.69		0.13
2.89		0.00
3.40		-0.14
3.47		-0.16
3.58		-0.18
3.81		-0.21
4.03		-0.26
4.48		-0.32
4.93		-0.41
5.60		-0.50
12.36		-0.72
21.50		-0.90
45.92		-1.12
52.64		-1.15
78.40		-1.16
```

## Sidenote

Moving forward, this could easily be implemented as an auto-learned profile. If the minimum gas command is being sent with a zero brake command, then the resulting acceleration (pitch-, and perhaps drag-adjusted) can be attributed to engine/regen braking under the current conditions (speed, engine RPM, transmission gear, high voltage battery temperature and state of charge, depending on how thorough you wanted to be). This could also be easily extracted from rlogs.

I expect this type of change could have a significant impact on long control for all makes and models, especially if auto-learned. Learned gas/brake threshold profiles, or those determined server-side from rlogs, could be consolidated and added as the default for a particular car when openpilot is first run, and then auto-learned on each individual car to account for weather, tire pressure/size, differences in suspension, etc. 

# Verification

Preliminary results show a complete elimination of oscillations in aEgo. Here the right set of plots  are with the dynamic lookup table and an adjusted tune, compared to stock on the left. See the complete elimination of acceleration oscillations (red path, 4th plot down), and much better execution of the desired acceleration (green path, 4th plot down). 
I was lucky enough to get rather comparable stops on master vs this branch. 
op-gm-dyamic-gasbrake-with45
op-gm-dyamic-gasbrake-stopped
op-gm-dyamic-gasbrake-gentle
op-gm-dyamic-gasbrake-abrupt

**Route**
Route: c11fcb510a549332|2022-08-23--07-36-40--1


twilsonco:
Sidenote:

Something like this could probably benefit all cars that use a gas brake style control. It would be useful to know additionally either which gear the car is in or just use engine RPMs in order to create a two dimensional look up table that would capture most of the important details.

On EV's, you could really go nuts and make a multi dimensional look up table that includes high-voltage battery temperature, state of charge, and in the case of hybrids, engine RPMs as well.

twilsonco:
converting to draft until new profile derived from better test data

twilsonco:
Updated with new test data.

TODO: these results are from the earlier implementation using a similar, but not the same gas/brake threshold profile. Need to record new test route.

TODO: this enables the use of much lower ki and kp values for the gm long tunes, and I wanted to include a new Volt long tune with this PR

selfdrive/car/gm/values.py
```diff
@@ -49,6 +50,16 @@ class CarControllerParams:
   GAS_LOOKUP_V = [MAX_ACC_REGEN, ZERO_GAS, MAX_GAS]
   BRAKE_LOOKUP_V = [MAX_BRAKE, 0.]
 
+  # determined by letting Volt regen to a stop in L gear from 89mph,
+  # and by letting off gas and allowing car to creep, for determining
+  # the positive threshold values at very low speed
+  EV_GAS_BRAKE_THRESHOLD_BP = [0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1, 1.2, 1.29, 1.52, 1.55, 1.6, 1.7, 1.8, 2.0, 2.2, 2.5, 5.52, 9.6, 20.5, 23.5, 35.0] # [m/s]
+  EV_GAS_BRAKE_THRESHOLD_V = [0.678, 0.67, 0.63, 0.56, 0.49, 0.38, 0.3, 0.13, 0.0, -0.14, -0.16, -0.18, -0.215, -0.255, -0.32, -0.41, -0.5, -0.72, -0.895, -1.125, -1.145, -1.16] # [m/s^s]
```
sshane, line 57: Any chance we can make this a simper/linear piecewise function?

sshane: Is L gear just requesting full regen with no friction brakes?

selfdrive/car/gm/interface.py
```diff
@@ -104,6 +104,13 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, disa
       ret.lateralTuning.pid.kiV = [0.]
       ret.lateralTuning.pid.kf = 1.  # get_steer_feedforward_volt()
       ret.steerActuatorDelay = 0.2
+      
+      ret.longitudinalTuning.kpBP = [5., 15., 35.]
+      ret.longitudinalTuning.kpV = [0.9, 1.0, 0.8]
+      ret.longitudinalTuning.kiBP = [5., 15., 35.]
+      ret.longitudinalTuning.kiV = [0.12, 0.16, 0.13]
+      ret.longitudinalActuatorDelayLowerBound = 0.41
+      ret.longitudinalActuatorDelayUpperBound = 0.41
```
sshane, line 113: Is this necessary?

selfdrive/car/gm/values.py
sshane: @twilsonco can you gather this data again, but instead just request max regen with no brake? Then we can guarantee these values are correct (rather than the L gear giving us a different response as @hewers points out)

sshane: This also doesn't match what you wrote above where you claim 2.89 m/s is where regen accel goes to 0 m/s/s

twilsonco: > This also doesn't match

Correct. As I continue to drive on this on my fork, I found that the positive acceleration portion of the gas/brake profile made the last bit of braking to a stop harsher than preferred. So the current profile in my fork modifies this in two ways; 1) removing the higher positive acceleration values at low speed, and moving the 0m/s^2 threshold to a lower speed. 

I made these changes _without followup tests to generation new profiles_. Rather I made the change based on observation and logs and verified they improved stopping performance anecdotally. https://github.com/twilsonco/openpilot/blob/18ddb5f231b88b5bb86042394ba273adfb1d4000/selfdrive/car/gm/values.py#L71

twilsonco: > @twilsonco can you gather this data again, but instead just request max regen with no brake?

I don't follow. These were recorded in L mode with no brakes. This is the amount of regen accessible to openpilot via gas command when in D or L mode, and the only way to get more regen is with the regen-on-demand paddle which we can't operate from openpilot. Otherwise, the braking system _will_ blend in as much regen as it can when you hit the brakes, but that's also the case with brake commands from openpilot, and it has nothing to do with the amount of regen accessible when the brakes aren't being used.

Now, we did find that a brake command of, say, 300, produces more brake force at 40mph than at 10mph, which seems to be a consequence of the lack of available regen at lower speeds, and can be fixed by increasing the max brake value at lower speeds, then MAX\_BRAKE will always produce the expected -3.5m/s^2 of brake. I do this in my fork already https://github.com/twilsonco/openpilot/blob/18ddb5f231b88b5bb86042394ba273adfb1d4000/selfdrive/car/gm/values.py#L77

selfdrive/car/gm/values.py
```diff
@@ -37,28 +38,42 @@ def __init__(self, CP):
...
-      # lower threshold removes some braking deadzone
-      max_regen_acceleration = -1. if CP.carFingerprint in EV_CAR else -0.1
 
-    self.GAS_LOOKUP_BP = [max_regen_acceleration, 0., self.ACCEL_MAX]
-    self.GAS_LOOKUP_V = [self.MAX_ACC_REGEN, self.ZERO_GAS, self.MAX_GAS]
+      if CP.carFingerprint in EV_CAR:
+        # determined by letting Volt regen to a stop in L gear from 89mph,
+        # and by letting off gas and allowing car to creep, for determining
+        # the positive threshold values at very low speed
+        self.MAX_REGEN_ACCEL_BP = [0, 2.0, 10.]
```
twilsonco, line 61: I think it should also have the CP at 20m/s, where the slope in the regen decal vs speed graph changes. Then the low speed regen acceleration value should be 0.1 rather than 0.5 in order to avoid rough stops that I commented about in response to your review comment @sshane

selfdrive/car/gm/values.py
sshane: I'm just saying you should gather the data again using the regen command, vs the L gear so we know for sure the observations are correct. Did you just find that L gear is similar to the regen command and use that?

sshane: Feel free to update the simpler with your updated observations btw!

selfdrive/car/gm/values.py
```diff
@@ -37,28 +38,42 @@ def __init__(self, CP):
...
+        self.MAX_REGEN_ACCEL_V = [0.5, 0.0, 1.0]
```
sshane, line 62: shouldn't it be flipped sign at very low speed?

sshane: removing for now

sshane:
Made a function to `compute_gb` like we do for Honda, but it splits up the X and Y of the breakpoints, let me know which you found clearer (updating the breakpoints or this).

twilsonco:
I think your version is cleaner and more readable (as usual...), and that will capture 95% of what's going on (though I still plan on doing a NN for long control FF that includes this component of controls). And being framed in the context of desired accel as you've done makes it easier to understand how the gas/brake lookups work conceptually 🤓.

sshane:
@twilsonco @nworb-cire can you guys test this?

Verylukyguy:
> @twilsonco @nworb-cire can you guys test this?
I have tested it on my 2018 GMC Acadia
df51b2195de66a9b|2023-07-11--17-18-07--0
The resume acceleration is still very poor and requires intervention every time.

sshane:
threshold works as expected! just need one Volt to test the new low-speed tuning now

twilsonco:
I can provide test routes this week with a comparison against stock as above.

sshane:
I'll merge a temporary fix to the issue ASAP then, and then revert with this PR

twilsonco:
Finally got around to testing. Recorded two routes, one on master and one on this branch. 
* master: c11fcb510a549332|2023-07-15--18-28-15--0
* gm-volt-long-tune: c11fcb510a549332|2023-07-15--19-39-52--1

-----

Here are some plots comparing the stopping behavior (which is the primary situation this PR addresses) between master and the PR branch currently.
* The subjective experience didn't improve a whole lot for the drives I did compared to the comparison [I originally did above](https://github.com/commaai/openpilot/pull/25453#issue-1340569123). 
* I figure this is because the Volt long tune was done in the context of inaccurate long feedforward due to the shifting gas/brake threshold this PR tries to address. The resulting current Volt long tune has too high kp, which was necessary to stop in time and correct for the inaccurate feedforward. 
* With this change in place, the error—that kp needed to be high to correct—is lessened, and so too can kp be decreased in order to get the (graphically and subjectively) smoother stops as in the original PR.
* **So, a followup adjustment to the Volt tune is recommended.**

I also think **one more breakpoint** would be worth adding to split the error between the current super-simple profile and the one I derived earlier. Here's my proposed revision for the ev profile;
```
bp=[2, 5, 20]
v=[0, -0.75, -1.1]
```
187573929-317f2f67-8e3b-49be-ab72-f77333174351

The corresponding ICE car profile would be 
```
bp=[2, 5, 20]
v=[0, -0.03, -0.1]
```
...or something.

I also think it may be _worth it to use a smoothed profile using quadratic b-splines_, given the dramatic shift in slope that still seems to be making the stops less smooth than they could be.

-----

Despite the subtle change in subjective driver experience, the plots make it clear where and how this is helping.
Recall this change is to make it so that, at lower speeds, the brakes come on sooner (in response to a lower magnitude negative desired accel).
* The right column (with dynamic gas/brake compensation) shows the brake command coming on when actuators.accel is still at -0.28, while in the left column brakes come on at accel = -1.0. 
* The result is that with _master_, there's a pause in deceleration while actuators.accel has to get from 0 to -1.0 before brakes actuate, during which time error begins to grow, and finally as the friction brakes come on, it's in combination with the P corrective response and results in harsh braking, which overshoots and then releases.
* By the time the brake command comes on, you can see error has been growing between aEgo and actuators.accel for 2s.
* Rather than engaging the brakes sooner, you can see it slowly ramping down the gas command (even though at those speeds it's not getting the regen brake force it expects).
* With the dynamic gas/brake threshold, there is no such lapse in brake force, and less error occurs.

| Master | PR branch |
| --------------- | --------------- |
| master-detail | pr-detail |

-----
Here's more plots

| Master | PR branch |
| --------------- | --------------- |
| stop behind stopped lead master1 [video link](https://youtu.be/ToloIZWl3C0) | stop behind stopped lead pr1 [video link](https://youtu.be/mwReH3yRgWQ) |
| stop behind stopping lead master2 [video link](https://youtu.be/1WSMxvEKVRs) | stop for red light pr2 [video link](https://youtu.be/vhZVq1dmIAw) |
| master3 [video link](https://youtu.be/9egaPgICIwQ) | pr3 [video link](https://youtu.be/X9Wjsop_iUk) |
| master4 [video link](https://youtu.be/PZaNptOjJ6k) | ... |

@sshane

Verylukyguy:
Is this ready for merge?

selfdrive/car/gm/values.py
```diff
@@ -53,12 +54,27 @@ def __init__(self, CP):
       # lower threshold removes some braking deadzone
       max_regen_acceleration = -1. if CP.carFingerprint in EV_CAR else -0.1
 
-    self.GAS_LOOKUP_BP = [max_regen_acceleration, 0., self.ACCEL_MAX]
+    # The max amount of deceleration possible using ASCMGasRegenCmd alone.
+    # It is assumed no regen/engine braking is available at low speeds
+    self.MAX_REGEN_ACCEL_BP = [2., 10.]  # m/s
+    self.MAX_REGEN_ACCEL_V = [0., max_regen_acceleration]  # m/s^2
```
sshane, line 60:
```suggestion
    # The max amount of deceleration possible using ASCMGasRegenCmd alone.
    # Some creep acceleration is assumed at low speeds
    self.MAX_REGEN_ACCEL_BP = [2., 10.]  # m/s
    self.MAX_REGEN_ACCEL_V = [0.2, max_regen_acceleration]  # m/s^2
```

@twilsonco I think creep acceleration may be disabled if ACC is active, can you check?

sshane:
I checked your routes and we went from undershooting deceleration to overshooting, you can see integral switching sides. Left is master, right is this branch:

image

twilsonco:
True, but there's consistent error on the left plot that's vanishingly small on the right plot (space between desired and actual speed, also see the magnitude on the p response. It stays at 0.2 on the right and 0.5 on the left). 

However, nworby is making great progress on a better approach to data-based long controls. Might be better to let this sit for a bit until his he ready for comparison.

adeebshihadeh:
If we don't currently expect to merge this PR, let's close. You can re-open if we do end up wanting this version of it.

twilsonco:
Closed this in favor of https://github.com/commaai/openpilot/pull/29659 which is now closed. Please reopen @adeebshihadeh

selfdrive/car/gm/values.py
twilsonco: > I'm just saying you should gather the data again using the regen command, vs the L gear so we know for sure the observations are correct. Did you just find that L gear is similar to the regen command and use that?

Yes, I used L mode under the observation-based assumption that it has access to the same amount of regen as openpilot when using the MAX\_REGEN value. It may be worth it to rerun the tests using a hardcoded MAX\_REGEN command just to be sure, as you suggest.

twilsonco:
I'll update this shortly Routes using a simplified EV lookup table, slightly adjusted to account for the overshoot you pointed out above.

selfdrive/car/gm/interface.py
twilsonco: Neglected to respond to this. Yes, the lookup table alone resulted in the existing long tune gains being too high and it didn't smooth stops at all really. The more accurate (essentially) long feedforward provided by the lookup table decreased error so much that the gains could be lowered considerably. If those same lower gains were used without the lookup table, however, then in many instances the car would probably hit the lead for quicker stops.

PR #25464, opened 2022-08-17
Fix is_rhd param name in map.cc

eFiniLan:
Fix is_rhd param name in map.cc

### DIFF
--- a/selfdrive/ui/qt/maps/map.cc
+++ b/selfdrive/ui/qt/maps/map.cc
@@ -361,7 +361,7 @@ void MapWindow::offroadTransition(bool offroad) {
 }
 
 MapInstructions::MapInstructions(QWidget * parent) : QWidget(parent) {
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
This isn't quite the same since `IsRhdDetected` is only written every 5m. @ZwX1616 what'st the preferred way to do this? perhaps we should just save the param more often or ensure we write it on change

eFiniLan:
> This isn't quite the same since `IsRhdDetected` is only written every 5m. @ZwX1616 what'st the preferred way to do this? perhaps we should just save the param more often or ensure we write it on change

ah got it, I thought IsRhdDetected will write only once (there is no car with a swap-able steering wheel I guess 🤣 )

adeebshihadeh:
Hmm, this widget doesn't update until next startup if you changed the old toggle anyway, so I'll merge as is.

PR #25466, opened 2022-08-17
liveParams pitch

twilsonco:
part of PR stack

### DIFF
--- a/selfdrive/locationd/models/car_kf.py
+++ b/selfdrive/locationd/models/car_kf.py
@@ -39,6 +39,7 @@ class States():
   YAW_RATE = _slice(1)  # [rad/s]
   STEER_ANGLE = _slice(1)  # [rad]
   ROAD_ROLL = _slice(1)  # [rad]
+  ROAD_PITCH = _slice(1)  # [rad]
 
 
 class CarKalman(KalmanFilter):
@@ -51,6 +52,7 @@ class CarKalman(KalmanFilter):
     0.0,
 
     10.0, 0.0,
+    0.0,
     0.0,
     0.0,
     0.0
@@ -67,6 +69,7 @@ class CarKalman(KalmanFilter):
     math.radians(0.1)**2,
     math.radians(0.1)**2,
     math.radians(1)**2,
+    math.radians(1)**2,
   ])
   P_initial = Q.copy()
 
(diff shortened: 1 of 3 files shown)


twilsonco:
c11fcb510a549332|2022-08-17--19-11-34

twilsonco:
Verification from the posted route:
image

sshane:
@nuwandavek do we want this for anything? It could be useful for Honda and Toyota w/ pedal (and maybe using permit braking)

sshane:
You can use `CC.orientationNED[1] * -G` to get a rough pitch. That's basically what this PR is, but without a filter

twilsonco:
> You can use `CC.orientationNED[1]` to get a rough pitch from inside the carcontroller. That's basically what this PR is, but without a smoothing filter. Discussed with @nuwandavek and he says using the future pitch from the model isn't guaranteed to be reliable

That's not the same thing. That give's you pitch correction 1s after you need it. It would be good for keeping long from having insufficient acceleration up steep hills, but using `CC.orientationNED[1]` will not prevent the error from pitch changes like this does.

Furthermore, regarding the model data not being reliable, it's only being used to alter the current live pitch reading slightly to remove delay from smoothing, long actuator delay, and other factors. The model pitch used is capped to be guaranteed within sane values, making this no less safe or reliable than using `CC.orientationNED[1]` directly. The times when the model predicted pitch is unreliable is when the path is unreliable, such as when performing low-speed manual turns where the model has low path confidence hence low pitch confidence. At such low speeds, the model predicted change in pitch 0.8s in the future will be negligibly small. 

I've also been driving on this for some time now, as have my fork users. We've seen no strange behavior. Only smoother long control and better gas mileage. 

Please reconsider.

twilsonco:
Also, why is livePamams roll a thing, @sshane, @HaraldSchafer, if orientationNED[0] is just as good?

HaraldSchafer:
The liveParams learner learns parameters of the lateral motion of the vehicle. Pitch isn't part of that model, so doesn't belong there. 

OrientationNED[1], should be a good estimate of the pitch of the comma3. It will however change with the car's suspension movements, it may be desire-able to filter that out somehow, but not by using a parameter estimate used for a different purpose as a low pass filter. Really the model should one day just also tell you the pitch with respect to the road and then you can subtract the two and get the pitch of the road.

twilsonco:
That makes sense. I had been thinking of liveParams in a different context, as a fancy pants filter that weighted values based on their confidence. 

Thanks

nuwandavek:
> Also, why is livePamams roll a thing, @sshane, @HaraldSchafer, if orientationNED[0] is just as good?

`paramsd` has a model of the car's lateral dynamics. It takes raw sensor/processed inputs such as roll, yaw_rate, vego etc and jointly optimizes for the car's parameters. It is better to use this estimate of the roll (that has been jointly optimized) than the roll directly from orientationNED[0], as it has passed through a fairly sophisticated filter.

If you do not add the parameter to this vehicle model (such as pitch in this case), it is not part of the joint optimization and just acts as a smoother. But as Harald mentioned the biggest issue with pitch is the suspension dynamics, and not necessarily the smoothing.

twilsonco:
Thanks @nuwandavek. I appreciate the explanation. 

As pitch reaches 90 degrees, it too will strongly affect the lateral response 🤓

PR #25467, opened 2022-08-17
liveParams future long-actuator-delayed pitch

twilsonco:
part of PR stack. 

But I don't know how to submit a PR to commaai/openpilot that's comparing two branches in the opgm repo, so the layers of the PR stack will include the commits from the previous.

### DIFF
--- a/selfdrive/locationd/models/car_kf.py
+++ b/selfdrive/locationd/models/car_kf.py
@@ -39,6 +39,7 @@ class States():
   YAW_RATE = _slice(1)  # [rad/s]
   STEER_ANGLE = _slice(1)  # [rad]
   ROAD_ROLL = _slice(1)  # [rad]
+  ROAD_PITCH = _slice(1)  # [rad]
 
 
 class CarKalman(KalmanFilter):
@@ -51,6 +52,7 @@ class CarKalman(KalmanFilter):
     0.0,
 
     10.0, 0.0,
+    0.0,
     0.0,
     0.0,
     0.0
@@ -67,6 +69,7 @@ class CarKalman(KalmanFilter):
     math.radians(0.1)**2,
     math.radians(0.1)**2,
     math.radians(1)**2,
+    math.radians(1)**2,
   ])
   P_initial = Q.copy()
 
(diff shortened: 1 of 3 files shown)


twilsonco:
Route: c11fcb510a549332|2022-08-17--19-03-56

twilsonco:
Verification: from posted route 
image

And here the long actuator delay is set to 0.15s, and the predicted pitch predicts live params pitch by around 0.15s
image

twilsonco:
I verified that this would work with plot juggler beforehand. That is, I found that I could control the time-amount of pitch prediction by using farther values of model.orientation.y[]. A farther predicted pitch (higher index in orientation.y) will shift the green path to the left, predicting the future pitch by an even greater amount of time, and a closer predicted pitch will move the green path to the right.

However, I also looked into doing this for roll compensation (so that it could be applied in advance of steering actuator delay, the way a human responds to expected changes in roll), but for roll, you don't get the same level of predictive capacity. 

Here's a smoothed live-params roll with and without future predicted roll added in (yellow and blue lines respectively), compared to live-params roll (cyan) and calibrated orientationNED pitch (purple). Here I'm using model.orientation.x[10] for the yellow line. That value should be around 1s into the future, yet it only predicts the non-predictive smoothed pitch by around 0.1s, and sometimes not at all. 

image

I speculate this is because the predicted pitch is simply more accurate than the predicted roll. The reason for this is that road pitch changes similarly in rate and magnitude regardless of vehicle. Roll, however, as recorded by a sensor device over a meter above the ground, will vary from vehicle to vehicle based on differences in the vehicle suspension/tire pressure/etc. This makes the data from many different cars much less likely to lead to a consistent result for predicted roll.

Cases where road roll changes gradually and deliberately would be better predicted, but those are also the cases where the difference of 0.2s (steer actuator delay) in actuation is less noticeable to the driver. What I would like to do is make openpilot  turn ahead of time for sudden and short-lived changes in road roll, like when passing an intersection on a crowned road, but it seems that may not be possible with the current model predictions. If they were to improve, this method should work for roll as well as it does now for pitch.

twilsonco:
I'd like to see an example of where this method produces unreliable results @sshane, @nuwandavek. I've looked at tons and tons of log data and seen that all this does is consistently predict future pitch by the desired amount of time, with very high accuracy.

PR #25468, opened 2022-08-17
Compute pitch compensated acceleration

twilsonco:
part of PR stack.

But I don't know how to submit a PR to commaai/openpilot that's comparing two branches in the opgm repo, so the layers of the PR stack will include the commits from the previous.


twilsonco:
route: c11fcb510a549332|2022-08-17--18-58-31

twilsonco:
Verification: from posted route
image

PR #25469, opened 2022-08-17
GM longitudinal pitch compensation

twilsonco:
_Part of PR stack, but I don't know how to submit a PR to commaai/openpilot that's comparing two branches in the opgm repo, so the layers of the PR stack will include the commits from the previous._

# Description

Adds longitudinal pitch compensation to be used in cars that use gas/brake control, where the command specifies an amount of acceleration force rather than an amount of acceleration relative to the road. The implementation is contained in car controller, as a type of passive long feedforward, transparent to the controller. 

To allow this to account for longitudinal actuator delay to allow for smoothing of the pitch signal, future model pitch is used to create a smoothed, predictive pitch.

Though not shown here, another benefit of this improvement is that if more than `ACCEL_MAX` acceleration is required to maintain speed up a steep hill, stock will not be able to maintain speed. With this improvement that is no longer the case.

`actuators.accelPitchCompensated` is used to determine the gas command, which specifies an acceleration force. The brake command on gm specifies acceleration relative to the ground, so the `actuators.accel` value is the correct value for friction brakes, even when going downhill. 
To support this assertion, see that gas command does not match up with aEgo:
image
While brake command does match with aEgo
image
I also verified this extensively (though not with proper documentation) when developing "one pedal mode" in my openpilot fork, which involved artificially sending a constant brake command and observing the car's braking behavior. I found that the same brake command produces far more braking force on a decline, and far less (or none) on an incline.

However, when openpilot requests negative acceleration in order to stop behind a lead _when going uphill_, it would engage the friction brakes sooner (at the same time it would now, without this improvement), despite the negative gravitational acceleration, and resulting in more braking than in necessary, since it's being assisted by gravity.
So you'd like for openpilot to postpone friction brake use when decelerating uphill, but it can become a problem when stopping, as the pitch-compensated acceleration may remain positive even though openpilot would like to stop, resulting in late stopping behind leads.
The solution is to smoothly switch to non-pitch-compensated acceleration for determining brakes at low speed.
I've been using this method on my fork for about a month and it keeps the efficiency and brake-avoidance benefits of pitch-compensation without introducing possible problems when stopping behind leads.

(Any issues caused by the gas and brake signals having different physical meanings will also be minimized by [using the correct gas/brake threshold acceleration value](https://github.com/commaai/openpilot/pull/25453))

**This should also be used on other makes that specify acceleration force with gas or brake commands**

# Verification

The result is a complete elimination of long error due to pitch changes. In the following image the plots on the left are without pitch compensation, and on the right with pitch compensation. 

Screen Shot 2022-08-17 at 3 16 49 PM

See that as I climb up the stepped hill (pitch changing between 0 and ~12% grade), the pitchFutureLone (right side, second plot from top, orange line) predicts the liveParams pitch by ~0.4s, which is the value set for longitudinalActuatorDelayLowerBound (and upper bound), so it's predicting by the correct amount.

While there are large deviations between target and actual acceleration (fourth plot from top, left side) without pitch compensation, the pitch-induced error is completely removed on the right (error stays within ±0.05m/s^2 vs ±0.25m/s^2).
Also the with pitch compensation speed stays within 0±0.1m/s^2 compared to ±0.4m/s^2 (top plot). 

Here's another image showing data where I descend down the same stepped hill. Again, with pitch compensation, the speed and acceleration oscillations due to changing pitch are completely removed.

Screen Shot 2022-08-17 at 3 12 35 PM

This PR includes three other PRs to 1) [add liveParams pitch](https://github.com/commaai/openpilot/pull/25466), 2) [add liveParams future predicted pitch based on long actuator delay](https://github.com/commaai/openpilot/pull/25467), and 3) to [add calculation of pitch-compensated acceleration, saved to actuators.accelPitchCompensated](https://github.com/commaai/openpilot/pull/25468).

Route: c11fcb510a549332|2022-08-17--13-58-51--0

### DIFF
--- a/selfdrive/car/gm/carcontroller.py
+++ b/selfdrive/car/gm/carcontroller.py
@@ -10,7 +10,9 @@ from selfdrive.car.gm.values import DBC, CanBus, CarControllerParams, CruiseButt
 VisualAlert = car.CarControl.HUDControl.VisualAlert
 NetworkLocation = car.CarParams.NetworkLocation
 
-
+# only use pitch-compensated acceleration at 10m/s+
+ACCEL_PITCH_FACTOR_BP = [5., 10.] # [m/s]
+ACCEL_PITCH_FACTOR_V = [0., 1.] # [unitless in [0-1]]
 class CarController:
   def __init__(self, dbc_name, CP, VM):
     self.CP = CP
@@ -59,7 +61,7 @@ class CarController:
       idx = (CS.lka_steering_cmd_counter + 1) % 4
 
       can_sends.append(gmcan.create_steering_control(self.packer_pt, CanBus.POWERTRAIN, apply_steer, idx, CC.latActive))
-
+    
     if self.CP.openpilotLongitudinalControl:
       # Gas/regen, brakes, and UI commands - all at 25Hz
       if self.frame % 4 == 0:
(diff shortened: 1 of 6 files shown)


twilsonco:
Need to record routes on each PR in the stack. Will be ready for review after that

twilsonco:
Routes recorded for other three PRs in the stack (not a real stack), so they're ready for review. Need to record a final route for this PR

sshane:
Generally when proposing a large change, it's best to open a single PR with all your changes so we can see what it would look like. Then if it's a wanted PR and it could be split out into smaller PRs, we'd do that at that time.

twilsonco:
Route: c11fcb510a549332|2022-08-18--07-27-30--5

twilsonco:
As I mention in the related pull requests. Future pitch is _necessary_ in order to account for long actuator and other delays. Humans don't wait for the speed to drop before apply more gas when climbing a hill. They do it proactively, which avoids the error, avoids overcorrecting for the error, and improves efficiency and passenger comfort. 

I've been driving on this logic for close to 1000 miles. It works great, consistently.

twilsonco:
I'll reimplement a standalone carcontroller version without using model pitch.

PR #25470, opened 2022-08-17
controlsd: resume does not increment speed while cruise standstill

### LINKED ISSUE #25432: GM: Runaway set speed increase when no acc buttons pressed
### Describe the bug

While testing for [this PR](https://github.com/commaai/openpilot/pull/25422) using the following procedure, I encountered a bug where the set speed was increasing in increments of 5mph anytime openpilot was engaged, regardless of ACC button presses.

However, after off-road/on-road cycling my C3, a subsequent attempt of the procedure did not reproduce the bug. This is a preliminary report, and I don't yet know how to reproduce the bug consistently. This has only happened once to me. I suspect the bug is due to me having been holding the res/accel button as openpilot came to a stop behind the lead (which involves a change in ACC cruise state), but again, repeating the below procedure didn't reproduce the bug.

Bug description:
Once the bug as been triggered, op increases the set speed as if you're pressing and holding the res/accel button, resulting in 5mph increments to the set speed every half second.
Presses of the set/decel button or the res/accel button will result in 1mph decrements or increments to the set speed, in addition to those cased by the persistent phantom res/accel button hold.

Testing procedure :
1. engage and set speed to 5mph
2. Pull behind parked car and as op slows, increase set speed to 90mph (I stopped at 85, but same thing). op will still slow and stop behind parked lead
3. Once stopped, you can use the set/decel button to lower speed 1mph at a time and confirm visually that the dash speed matches the onscreen op set speed.

GM Chevy Volt

c11fcb510a549332|2022-08-13--10-34-36

### openpilot version

67cd27a9e153eef6359a85575c6fe803f9be602a

### Additional info

Commit has corresponds to latest commit of the [gm-cruise-speed-scaling](https://github.com/commaai/openpilot/compare/gm-cruise-speed-scaling) branch.

sshane:
Should allow you to exit standstill on first resume button press, then increments set speed on any subsequent presses (assuming you exit standstill for a few seconds after pressing resume, like Toyota.). Only really applicable on GM as it's the only make that you need to press resume while being `not pcmCruise`.

We need to store the state of the car when you press the button since GM leaves standstill on rising edge, and we increment speed on falling edge (it would look like we weren't in standstill)

closes: https://github.com/commaai/openpilot/issues/25432


selfdrive/car/gm/interface.py
```diff
@@ -179,7 +179,7 @@ def _update(self, c):
       be = create_button_event(self.CS.cruise_buttons, self.CS.prev_cruise_buttons, BUTTONS_DICT, CruiseButtons.UNPRESS)
 
       # Suppress resume button if we're resuming from stop so we don't adjust speed.
-      if be.type == ButtonType.accelCruise and (ret.cruiseState.enabled and ret.standstill):
+      if be.type == ButtonType.accelCruise and (ret.cruiseState.enabled and ret.cruiseState.standstill) and be.pressed:
```
sshane, line 182: It's better to do this generically, I'll fix it up soon

sshane:
@twilsonco can you stress test this branch?

- [ ] See if you can reproduce runaway on master, then try to reproduce here.
- [ ] Resume/set from a stop
- [ ] Get into a cruise standstill state behind car, then press resume to exit standstill (no set speed change), then press resume again (should increment set speed if you leave standstill for a few seconds)

twilsonco:
>  See if you can reproduce runaway on master, then try to reproduce here
Reproduced it twice on master by pressing and holding the resume button as it transitioned into standstill, then releasing the button. Once in each of these routes.
I wasn't able to reproduce it by, when slowing with the brake pedal, holding resume while coming to a stop and then releasing. 
c11fcb510a549332|2022-08-20--16-19-29
c11fcb510a549332|2022-08-20--16-22-34
Performed the same procedure twice with the fix branch and it did not trigger the runaway set speed. I was able to increment and decrement the set speed while stopped behind a lead
c11fcb510a549332|2022-08-20--16-35-02
>   Resume/set from a stop
Set/resume while stopped with brakes pressed gives a "can't engage under 18" alert. Same when done while creeping 2mph without brakes pressed. Same despite initial or subsequent engagement.
c11fcb510a549332|2022-08-20--16-46-39
>   Get into a cruise standstill state behind car, then press resume to exit standstill (no set speed change), then press resume again (should increment set speed if you leave standstill for a few seconds)
On the fix branch, pressing resume after a lead pulls away (i.e. when the long planner is requesting positive acceleration) exits standstill and openpilot resumes. However, first press exits standstill and increments set speed.
(same route as above)
c11fcb510a549332|2022-08-20--16-46-39

sshane:
Hmm, it looks like your car exited standstill before the pressed=False button event so it treated it as a normal increment of set speed. I'll try to fix that

sshane:
Will re-open with a working branch

sshane:
closes: https://github.com/commaai/openpilot/issues/25432

PR #25471, opened 2022-08-18
Chrysler: FW dump

sshane:
This removes versions not seen fingerprinting, will add those back


adeebshihadeh:
#25363 is also relevant

sshane:
In the notebook I made sure to replace the ECU name on all `(0x753, None)` with "fwdRadar" (for the Ram). Will make sure we correctly set `0x764` to fwdCamera.

jyoung8607:
I released #25363 from draft if you want it, it's good for the JGCs and Pacificas, but there's still the PacHy difference to contend with. Pretty sure they'll identify, but someone should really probe them with panda/examples/query_fw_versions to see what's necessary to fix them up correctly, and I haven't had time to work further with those end users.

sshane:
Will re-open with more versions when ready to merge

PR #25477, opened 2022-08-18
Mazda: CX5 2021 FW

Jafaral:
dongle-id: 82c6d952da5e2122


sshane:
It looks like routes from this dongle are missing the esp FW, causing it to always fuzzy match. Have any idea why it's not getting the esp?

Jafaral:
That is weird. This car is in Taiwan,  and the fw sent to me had 

```
ecu = esp
fwVersion = KBJ5-437K2-B\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00
address = 1888
subAddress = 0
```

which we already have.

sshane:
Hmm, it seems pretty inconsistent. This route has esp: 82c6d952da5e2122|2022-07-30--20-15-01, but the rest after that do not: `82c6d952da5e2122|2022-08-06--14-38-49` for example

Jafaral:
I will let them know. I also don't have access to the routes so can't really tell.

chaly77714:
@sshane 
Hi, I'm the requester for the CX5 2021 fingerprint.
I turned on my wifi for long time, can you see the logs from my device?

sshane:
@chaly77714 can you get a route on master? We added some new logging and would like to see if I can find the issue with your ABS ECU not responding

chaly77714:
> @chaly77714 can you get a route on master? Make sure not to skip FW query. We added some new logging and would like to see if I can find the issue with your ABS ECU not responding

chaly77714:
> @chaly77714 can you get a route on master? Make sure not to skip FW query. We added some new logging and would like to see if I can find the issue with your ABS ECU not responding

Do you mean using the official Openpilot in master branch? because I use dragonpilot  currently.

sshane:
Yes

sshane:
Let me know when you get a route on master @chaly77714

adeebshihadeh:
Any update on this?

chaly77714:
> Any update on this?

Sorry for late reply. I just install master branch and connect to router.
But I don't change anything, so it run on dashmode.

sshane:
Thanks! It appears your abs is responding without any errors the last two attempts on master. I'll go ahead a merge. Please let us know if your car ever fails to fingerprint!

PR #25478, opened 2022-08-18
Honda: 2021-22 Odyssey support

sshane:
Honda Odyssey switched to Bosch hardware with the radar limiting steering to the factory LKAS operating style. It looks like the Acura RDX did too for 2022 but I've only seen one vehicle of this year.

Odyssey is capable of steering down to nearly 0mph with the radar bypassed. As for the RDX, this hasn't been tested.

**Checklist**
- [ ] route with openpilot: wip
- [ ] route with stock system: wip
- [ ] check and alert if the output of the forwarded steering torque message is different to what we've requested. Honda Bosch radar apply minor rate and/or jerk limiting to the torque requests so allow for a bit of mismatch. Look into making this generic across all makes too.
- [ ] Verify with certainty that the Imperial Unit signal in ACC_HUD is set as expected on Honda Bosch. Initial findings, and chats with @vanillagorillaa, show that this is the case. It's included in this PR as I've seen Odyssey 21+ with both set speed units.
- [ ] Is steering consistent with the Nidec Odyssey? I haven't heard any complaints with using it's torque tune on this variant.


sshane:
Temp PR, wasn't able to push to the original one

PR #25450, opened 2022-08-16
Add 2021-22 Honda Odyssey

csouers:
Honda Odyssey switched to Bosch hardware with the radar limiting steering to the factory LKAS operating style. It looks like the Acura RDX did too for 2022 but I've only seen one vehicle of this year.

Odyssey is capable of steering down to nearly 0mph with the radar bypassed. As for the RDX, this hasn't been tested.

**Checklist**
- [x] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [ ] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [ ] route with openpilot: wip
- [ ] route with stock system: wip
- [ ] check and alert if the output of the forwarded steering torque message is different to what we've requested. Honda Bosch radar apply minor rate and/or jerk limiting to the torque requests so allow for a bit of mismatch. Look into making this generic across all makes too.
- [ ] Verify with certainty that the Imperial Unit signal in ACC_HUD is set as expected on Honda Bosch. Initial findings, and chats with @vanillagorillaa, show that this is the case. It's included in this PR as I've seen Odyssey 21+ with both set speed units.
- [ ] Is steering consistent with the Nidec Odyssey? I haven't heard any complaints with using it's torque tune on this variant.

PR #25481, opened 2022-08-18
Add 2022 Acura RDX

csouers:
Similar steering capabilities to 2021+ Honda Odyssey.

Test and example routes are in the works.

**Prereqs**
- https://github.com/commaai/openpilot/pull/25450

**Checklist**
- [ ] route with openpilot:


csouers:
Here's the stock route for this car: `c699fc417b4d6e94|2022-08-21--20-29-14`. The OP route is coming soon.

sshane:
This relies on https://github.com/commaai/openpilot/pull/25450, correct? You can re-open this once that one is in.

PR #25483, opened 2022-08-18
joystick: revert max axes value to 255

sshane:
It's technically possible to get the max value from evdev (run `evtest`), but inputs does not support this yet. 255 seems to be the most common minimum max value, so use that. Then you need to calibrate on any other joystick by moving through the whole range first. Fixes joysticks with ranges under 1023


sshane:
let's try to detect the axes max values first

PR #25485, opened 2022-08-19
GM long: add pitch-compensation

twilsonco:
([This](https://github.com/commaai/openpilot/pull/25469), but without model-predicted pitch, and this is entirely contained in gm/carcontroller)

# Description/justification

This PR adds pitch compensation to longitudinal control for GM cars, resulting in a large decrease in longitudinal error (versus the long plan) when pitch changes or when on an incline/decline.

Adds longitudinal pitch compensation to be used in cars that use gas/brake control, where the command specifies an amount of acceleration force rather than an amount of acceleration relative to the road. The implementation is contained in car controller, as a type of passive long feedforward, transparent to the controller. 

To allow this to account for longitudinal actuator delay to allow for smoothing of the pitch signal, future model pitch is used to create a smoothed, predictive pitch.

Though not shown here, another benefit of this improvement is that if more than `ACCEL_MAX` acceleration is required to maintain speed up a steep hill, stock will not be able to maintain speed. With this improvement that is no longer the case.

`actuators.accelPitchCompensated` is used to determine the gas command, which specifies an acceleration force. The brake command on gm specifies acceleration relative to the ground, so the `actuators.accel` value is the correct value for friction brakes, even when going downhill. 
To support this assertion, see that gas command does not match up with aEgo:
image
While brake command does match with aEgo
image
I also verified this extensively (though not with proper documentation) when developing "one pedal mode" in my openpilot fork, which involved artificially sending a constant brake command and observing the car's braking behavior. I found that the same brake command produces far more braking force on a decline, and far less (or none) on an incline.

However, when openpilot requests negative acceleration in order to stop behind a lead _when going uphill_, it would engage the friction brakes sooner (at the same time it would now, without this improvement), despite the negative gravitational acceleration, and resulting in more braking than in necessary, since it's being assisted by gravity.
So you'd like for openpilot to postpone friction brake use when decelerating uphill, but it can become a problem when stopping, as the pitch-compensated acceleration may remain positive even though openpilot would like to stop, resulting in late stopping behind leads.
The solution is to smoothly switch to non-pitch-compensated acceleration for determining brakes at low speed.
I've been using this method on my fork for about a month and it keeps the efficiency and brake-avoidance benefits of pitch-compensation without introducing possible problems when stopping behind leads.

(Any issues caused by the gas and brake signals having different physical meanings will also be minimized by [using the correct gas/brake threshold acceleration value](https://github.com/commaai/openpilot/pull/25453))

**This should also be used on other makes that specify acceleration force with gas or brake commands**

Route: c11fcb510a549332|2022-08-19--09-03-54--0

### DIFF
--- a/selfdrive/car/gm/carcontroller.py
+++ b/selfdrive/car/gm/carcontroller.py
@@ -1,15 +1,24 @@
 from cereal import car
 from common.conversions import Conversions as CV
-from common.numpy_fast import interp
+from common.filter_simple import FirstOrderFilter
+from common.numpy_fast import clip, interp
 from common.realtime import DT_CTRL
+import math
 from opendbc.can.packer import CANPacker
 from selfdrive.car import apply_std_steer_torque_limits
 from selfdrive.car.gm import gmcan
 from selfdrive.car.gm.values import DBC, CanBus, CarControllerParams, CruiseButtons, EV_CAR
+from selfdrive.controls.lib.drive_helpers import apply_deadzone
+from selfdrive.controls.lib.vehicle_model import ACCELERATION_DUE_TO_GRAVITY
 
 VisualAlert = car.CarControl.HUDControl.VisualAlert
 NetworkLocation = car.CarParams.NetworkLocation
 
+BRAKE_PITCH_FACTOR_BP = [5., 10.] # [m/s] smoothly revert to planned accel at low speeds
+BRAKE_PITCH_FACTOR_V = [0., 1.] # [unitless in [0,1]; don't touch]
+PITCH_DEADZONE = 0.01 # [radians] 0.01 ≈ 1% grade
+PITCH_MAX_DELTA = math.radians(10.0) * DT_CTRL * 4  # 10°/s, checked at 25Hz
+PITCH_MIN, PITCH_MAX = math.radians(-19), math.radians(19) # steepest roads in US are ~18°
 
 class CarController:
   def __init__(self, dbc_name, CP, VM):
@@ -25,6 +34,7 @@ class CarController:
     self.lka_icon_status_last = (False, False)
 
     self.params = CarControllerParams()
+    self.pitch = FirstOrderFilter(0., 0.09 * 4, DT_CTRL * 4) # runs at 25 Hz
 
     self.packer_pt = CANPacker(DBC[self.CP.carFingerprint]['pt'])
     self.packer_obj = CANPacker(DBC[self.CP.carFingerprint]['radar'])
@@ -32,6 +42,7 @@ class CarController:
 
   def update(self, CC, CS):
     actuators = CC.actuators
+    accel = actuators.accel
     hud_control = CC.hudControl
     hud_alert = hud_control.visualAlert
     hud_v_cruise = hud_control.setSpeed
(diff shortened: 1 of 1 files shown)


twilsonco:
Route: c11fcb510a549332|2022-08-19--09-03-54--0

# Verification

Here we compare master with present and future pitch compensation on a couple steep steps of a hill, going up and down the steps.

Pitch compensation cuts speed error in half, and lowers acceleration error by 13%. 
Additionally, this eliminates the problem of `MAX_ACCEL` (2m/s^2) being insufficient for climbing steep hills; with pitch compensation op can accelerate up any hill the car can.

(In the future, when Comma has sufficiently verified the model predicted pitch, I'd like to [include future pitch in order to compensate for long actuator delay](https://github.com/commaai/openpilot/pull/25469), which, as you can see here, decreases speed and acceleration error by an additional 46% compared to when using instantaneous pitch. Compared to master, future pitch compensation reduces speed and acceleration error by 74% and 53% respectively, resulting in greater passenger comfort and improved vehicle efficiency.)

pitch compare

sshane:
@HaraldSchafer would Bolt long be helpful in verifying this? I can try to get that merged

sshane:
It's possible we need something similar to this for Toyotas to undo the accel compensation the PCM does, so we should try to make this as generic as possible.

twilsonco:
So something in between this and the [previous implementation](https://github.com/commaai/openpilot/pull/25469) that was unnecessarily couched inside of liveParams?

adeebshihadeh:
This isn't the large; blocking this on abstracting it out seems like premature optimization.

selfdrive/car/gm/carcontroller.py
```diff
@@ -87,6 +109,9 @@ def update(self, CC, CS):
         send_fcw = hud_alert == VisualAlert.fcw
         can_sends.append(gmcan.create_acc_dashboard_command(self.packer_pt, CanBus.POWERTRAIN, CC.enabled,
                                                             hud_v_cruise * CV.MS_TO_KPH, hud_control.leadVisible, send_fcw))
+      else:
+        accel_g = ACCELERATION_DUE_TO_GRAVITY * apply_deadzone(self.pitch.x, PITCH_DEADZONE) # driving uphill is positive pitch
+        accel += accel_g
```
sshane, line 114: any way you can remove the duplicate calculation?

selfdrive/car/gm/carcontroller.py
```diff
@@ -60,20 +71,31 @@ def update(self, CC, CS):
 
       can_sends.append(gmcan.create_steering_control(self.packer_pt, CanBus.POWERTRAIN, apply_steer, idx, CC.latActive))
 
+
+
+
```
sshane, line 76:
```suggestion
```

selfdrive/car/gm/carcontroller.py
```diff
@@ -60,20 +71,31 @@ def update(self, CC, CS):
...
+    
     if self.CP.openpilotLongitudinalControl:
       # Gas/regen, brakes, and UI commands - all at 25Hz
       if self.frame % 4 == 0:
+        # Pitch compensated acceleration;
+        # TODO: include future pitch (sm['modelDataV2'].orientation.y) to account for long actuator delay
+        pitch = clip(CC.orientationNED[1], self.pitch.x - PITCH_MAX_DELTA, self.pitch.x + PITCH_MAX_DELTA)
+        pitch = clip(pitch, PITCH_MIN, PITCH_MAX)
+        self.pitch.update(pitch)
```
sshane, line 84: Feels like the delta clip and the filter do similar things, any reason for both?

selfdrive/car/gm/carcontroller.py
```diff
@@ -25,13 +34,15 @@ def __init__(self, dbc_name, CP, VM):
...
     self.params = CarControllerParams()
+    self.pitch = FirstOrderFilter(0., 0.09 * 4, DT_CTRL * 4) # runs at 25 Hz
 
     self.packer_pt = CANPacker(DBC[self.CP.carFingerprint]['pt'])
     self.packer_obj = CANPacker(DBC[self.CP.carFingerprint]['radar'])
     self.packer_ch = CANPacker(DBC[self.CP.carFingerprint]['chassis'])
 
   def update(self, CC, CS):
     actuators = CC.actuators
+    accel = actuators.accel
```
sshane, line 45: use class variable so you don't need to set it when not an active control frame. check toyota/carcontroller

selfdrive/car/gm/carcontroller.py
```diff
@@ -25,13 +34,15 @@ def __init__(self, dbc_name, CP, VM):
     self.lka_icon_status_last = (False, False)
 
     self.params = CarControllerParams()
+    self.pitch = FirstOrderFilter(0., 0.09 * 4, DT_CTRL * 4) # runs at 25 Hz
```
sshane, line 37: if we only use the filter, ensure that going from min to max pitch doesn't violate your selected rate limit above

selfdrive/car/gm/carcontroller.py
```diff
@@ -60,20 +71,31 @@ def update(self, CC, CS):
...
+        pitch = clip(pitch, PITCH_MIN, PITCH_MAX)
+        self.pitch.update(pitch)
+        accel_g = ACCELERATION_DUE_TO_GRAVITY * apply_deadzone(self.pitch.x, PITCH_DEADZONE) # driving uphill is positive pitch
+        accel += accel_g
         if not CC.longActive:
           # Stock ECU sends max regen when not enabled
           self.apply_gas = self.params.MAX_ACC_REGEN
           self.apply_brake = 0
         else:
+          brake_accel = actuators.accel + accel_g * interp(CS.out.vEgo, BRAKE_PITCH_FACTOR_BP, BRAKE_PITCH_FACTOR_V)
```
sshane, line 92: if we change brake like this we probably should change gas

selfdrive/car/gm/carcontroller.py
sshane, line 114: use class variable

sshane (changes requested):

simontheflutist:
With a view towards https://github.com/commaai/openpilot/issues/30341, econd-order effects would change the calculation from `pitch` to `sin(pitch) cos(roll)`. I can't say for sure if it's necessary, but freeway ramps, which can have pronounced pitch and roll, come to mind.

twilsonco:
Thanks @simontheflutist.

Oops forgot to close this. This too will be superseded by #29659.

twilsonco:
Closed this in favor of #29659  which is now closed. Reopening

twilsonco:
I'll get Shanes's proposed changes done and a new test route so that this can be merged

sshane:

```bash
PR_NUMBER=33045
```

sshane:
If you would like to re-open, you can run our new [maneuver tool](https://github.com/commaai/openpilot/tree/master/tools/longitudinal_maneuvers) and share the before and after reports. Here are some examples of the report showing a clear and obvious improvement, enabling us to merge a new tune faster and easier:

https://github.com/commaai/opendbc/pull/1248

https://github.com/commaai/opendbc/pull/1260

PR #25490, opened 2022-08-19
Toyota: remove STEERING_LKA frequency check

### LINKED ISSUE #25242: Toyota: enable counter checks in parser
I made the CANParser automatically handle validating counters, which exposed a bug in Toyotas. Goal is to revert https://github.com/commaai/opendbc/commit/4195e8f4c9998c0a1a6084c3cc71499307bbd81e. It seems the counter in 0x2e4 counts up by 2, and 0x283 doesn't seem to come in consistently.

sshane:
No signals of this message are read, it was only for giraffe switch detection


selfdrive/test/process_replay/ref_commit
```diff
@@ -1 +1 @@
-35899d5137e298220e91063f3078109227cc8715
\ No newline at end of file
+93a136f91739847afe1e1a4a1e98bc7c0adb5ec8
```
sshane, line 1: the regen'd routes had a frame or two of canTimeout when the original route did not, now they don't

PR #25495, opened 2022-08-19
test models: check panda standstill

adeebshihadeh:
- [ ] body
- [x] VW
- [x] Hyundai
- [x] Chrysler
- [x] Toyota - https://github.com/commaai/openpilot/pull/25570
- [x] Subaru
- [x] GM


sshane:
closing since we've largely done this in other PRs ([most recently hyundai](https://github.com/commaai/openpilot/pull/29442)) and only the body remains. it still has a TODO

adeebshihadeh:
Let's get the body fixed, I don't like maintaining all these exceptions

PR #25497, opened 2022-08-19
Show CAN error if message counters are invalid

### LINKED ISSUE #25160: Ensure CANParser counter/checksums checks are the same as panda
For example, `SCM_BUTTONS`'s counter started returning garbage values (or issue with a cable) and only panda noticed it, causing a controls mismatch.

We need to change how the checks work in the CANParser, if counter or checksum failed, we don't update the last time seen, so it just causes a TIMEOUT error with the same thresholds as it.

http://data-raw.internal/c390b60ad96d2ff2/2022-07-01--07-40-21/4/rlog.bz2

sshane:
Prevents a possible controls mismatch where panda notices a message's counter go invalid and set controls_allowed to 0, but openpilot stays engaged


selfdrive/car/tests/test_models.py
```diff
@@ -150,7 +150,8 @@ def test_car_interface(self):
       if i > 200:
         can_invalid_cnt += not CS.canValid
 
-    self.assertLess(can_invalid_cnt, 50)
+    print(can_invalid_cnt)
+    self.assertLess(can_invalid_cnt, 1)
```
sshane, line 154:
```suggestion
    self.assertLess(can_invalid_cnt, 50)
```

The one failure in CI was from using the last segment in a drive, msgs timed out. All expected

PR #25501, opened 2022-08-20
test_models: no CAN invalid tolerance

sshane:
- No tolerance for checks (catches bad counters)
- Ensures CAN valid cannot rise, fall, rise within first 2 seconds


sshane:
any reason for this tolerance @adeebshihadeh?

adeebshihadeh:
> any reason for this tolerance @adeebshihadeh?

No good reason to keep it at least. We probably started with it to get some level of test coverage on this.

Issue #24475, opened 2022-05-09
Recover from blocked steering message


adeebshihadeh:
If a steering message has violated any of the limits, panda safety resets the last torque measurements. openpilot is unaware of the blocking, which means the subsequent messages are likely to be blocked as well, eventually resulting in an EPS fault.

Surprisingly, all instances on 0.8.13-release are from a single car:
```
VOLKSWAGEN TIGUAN 2ND GEN                a926bf13019b0627|2022-03-19--09-01-37--79
VOLKSWAGEN TIGUAN 2ND GEN                a926bf13019b0627|2022-03-28--10-59-05--68
VOLKSWAGEN TIGUAN 2ND GEN                a926bf13019b0627|2022-04-05--20-14-06--112
VOLKSWAGEN TIGUAN 2ND GEN                a926bf13019b0627|2022-04-19--10-35-40--46
VOLKSWAGEN TIGUAN 2ND GEN                a926bf13019b0627|2022-04-20--12-47-41--82
VOLKSWAGEN TIGUAN 2ND GEN                a926bf13019b0627|2022-05-01--13-48-04--72
```

adeebshihadeh:
Thinking about this more, I don't think this is necessary.

PR #25502, opened 2022-08-20
GM EPS RT-latch-fault prevention via Panda-OP LKAS state sync

JMPZ11:
**Description**
When the Panda's RT rate limiting is triggered, the Panda resets all it's tracking torque to zero, but OP is not made aware. OP continues attempting to send the high torque it was previously as, and the Panda blocks this because it would technically be too large of a jump - from zero to whatever OP is thinking. This unintentional latching causes most GM EPS units to hard-fault after 200ms.

This code resets OP to 0 / inactive LKA state to match the Panda after 2 dropped frames

**Verification**
Bug fix seems to have largely prevented LKAS faults caused by the RT "latching". It doesn't fix 100% - we are still seeing occasional too-fast frames.

**Route**
Route: [a route with the bug fix]

### DIFF
--- a/selfdrive/car/gm/carcontroller.py
+++ b/selfdrive/car/gm/carcontroller.py
@@ -22,6 +22,8 @@ class CarController:
     self.last_button_frame = 0
 
     self.lka_steering_cmd_counter_last = -1
+    self.lka_last_rc_val = -1
+    self.lka_same_rc_cnt = 0
     self.lka_icon_status_last = (False, False)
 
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
This is a generic issue that all affects all car ports, and as such, I'd prefer to fix this generically in openpilot. #24475 tracks this and lists all the steering faults caused by this on release openpilot. Do you actually see this happening in the field, and if so, got any routes?

PR #25503, opened 2022-08-20
Chrysler: add engine-specific FW request for some cars

sshane:
On our Ram, the transmission responds only 0xF132 (and 0xF112 with different FW), however the engine only gives a valid-looking response on 0xF194 (and 0x192 with different FW) which is `SYSTEM_SUPPLIER_ECU_SOFTWARE_NUMBER`.


selfdrive/car/fw_versions.py
```diff
@@ -97,6 +97,11 @@ def p16(val):
 CHRYSLER_VERSION_RESPONSE = bytes([uds.SERVICE_TYPE.READ_DATA_BY_IDENTIFIER + 0x40]) + \
   p16(0xf132)
 
+CHRYSLER_VERSION_REQUEST_ALT = bytes([uds.SERVICE_TYPE.READ_DATA_BY_IDENTIFIER]) + \
+  p16(uds.DATA_IDENTIFIER_TYPE.SYSTEM_SUPPLIER_ECU_SOFTWARE_NUMBER)
+CHRYSLER_VERSION_RESPONSE_ALT = bytes([uds.SERVICE_TYPE.READ_DATA_BY_IDENTIFIER + 0x40]) + \
+  p16(uds.DATA_IDENTIFIER_TYPE.SYSTEM_SUPPLIER_ECU_SOFTWARE_NUMBER)
+
```
sshane, line 104: @realfast have any clue which of the two identifiers that look most correct, is correct?

Screenshot from 2022-08-20 00-18-59

sshane:
End result: we should either merge this to add engine FW for all cars, or remove all engine FW, as it will cause some Chrysler to *always* fuzzy fingerprint.

selfdrive/car/fw_versions.py
sshane: Should use software, as hardware can be the same across different EPS's despite different software capabilities

sshane:
@realfast suggested we should move all engine FW to use this new query. @adeebshihadeh fine with removing all engine FW and merge this query for data collection? It's fine if it doesn't make it for release, but we should remove engine FW from FW_VERSIONS so we don't have some Rams that fuzzy fingerprint.

realfast:
> @realfast suggested we should move all engine FW to use this new query. @adeebshihadeh fine with removing all engine FW and merge this query for data collection? It's fine if the query doesn't make it for release, but we should remove engine FW from FW_VERSIONS so we don't have some Rams that fuzzy fingerprint.

I haven’t tested this yet myself and can’t for a few days. If you can whip up a branch, I can have a few test it.

sshane:
This one should be good for testing. I left both engine requests in so we can compare the result for cars that do respond with engine FW now

sshane:
On second thought, this should be all fine to wait until after release

sshane:
- Ram HD w/ engine and no transmission FW (no transmission ECU): this gets us nothing (good, since no duplicates) - 8fb5eabf914632ae|2022-08-24--10-09-30
- Ram 1500 w/ transmission and no engine FW: gets us engine FW and no duplicate transmission FW - 2f4452b03ccb98f0|2022-08-25--11-05-05
- Pacifica 2018 ICE w/ neither engine nor transmission FW: gets us both - f3c237869e58406c|2022-08-25--09-15-11

PR #25505, opened 2022-08-20
Add toggle to disable saving road cam

JMPZ11:
**Description**
The high quality video streams quickly fill up the 32-gig version of the C3. I've never used the HQ videos - the rlogs are far more important to me and even they were disappearing as /data filled up.

Toggle is implemented using the same pattern as the RecordFront toggle; defaults to on.

**Verification**

* Verified it prevents the storage of HQ video files
* Verified still able to upload rlogs via connect
* Verified skipping the video streams effectively increases available rlog storage considerably
* Verified video recording resumes when toggle switched back on


adeebshihadeh:
Sounds like a nice fork feature, but we're not interested in this toggle. RecordFront only exists for privacy reasons. 

> Toggle is implemented using the same pattern as the RecordFront toggle; defaults to on.

This isn't true - RecordFront defaults to off. However, we recently added a dialog to select your preference in the Training Guide.

JasonJShuler:
> Sounds like a nice fork feature, but we're not interested in this toggle. RecordFront only exists for privacy reasons.
> 
> > Toggle is implemented using the same pattern as the RecordFront toggle; defaults to on.
> 
> This isn't true - RecordFront defaults to off. However, we recently added a dialog to select your preference in the Training Guide.

Cool beans, thanks for your time!

FYI - I think I used the wrong word - 'pattern' - I was referring to the way it disables the recording (and the option to force it to on). I copied the way RecordFront does it. Mentioned it to clarify that I didn't do anything novel - mostly copy-pasta.

PR #25510, opened 2022-08-21
Hyundai: Show LFA and HDA icons for all CAN-FD HKG

sunnyhaibin:
Currently the LFA and HDA icons only show on non HDA2 CAN-FD HKG when openpilot is engaged. This allows all variant of HDA CAN-FD to show the icons when engaged.

**openpilot disengaged:**
- LFA: off
- HDA: off

**openpilot engaged:**
- LFA: green
- HDA: green

image

### DIFF
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -78,7 +78,7 @@ class CarController:
         can_sends.append(hyundaicanfd.create_cam_0x2a4(self.packer, CS.cam_0x2a4))
 
       # LFA and HDA icons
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Unfortunately, the camera doesn't send this on HDA2 cars. It's likely sent by the ADAS ECU like LFA.

gregjhogan:
FYI, we currently don't show the HDA icon on non CAN-FD cars (just showing the icons doesn't do anything, putting the car in HDA mode decreases the length of time before you need to spam the resume button to something like 30 seconds).

PR #25517, opened 2022-08-22
user event flagging

incognitojam:
- flag events
- show flags in replay tool

### DIFF
--- a/selfdrive/ui/qt/sidebar.cc
+++ b/selfdrive/ui/qt/sidebar.cc
@@ -43,11 +43,16 @@ Sidebar::Sidebar(QWidget *parent) : QFrame(parent) {
   setFixedWidth(300);
 
   QObject::connect(uiState(), &UIState::uiUpdate, this, &Sidebar::updateState);
+
+  pm = std::make_unique<PubMaster, const std::initializer_list<const char *>>({"userFlag"});
 }
 
 void Sidebar::mouseReleaseEvent(QMouseEvent *event) {
   if (home_btn.contains(event->pos())) {
-    // TODO do something
+    // TODO: maybe only create event while car started?
+    MessageBuilder msg;
+    msg.initEvent().initUserFlag();
+    pm->send("userFlag", msg);
   }
   if (settings_btn.contains(event->pos())) {
     emit openSettings();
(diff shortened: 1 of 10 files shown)


incognitojam:
Screenshot from 2022-08-22 15-18-31

selfdrive/ui/qt/sidebar.cc
```diff
@@ -43,9 +43,18 @@ Sidebar::Sidebar(QWidget *parent) : QFrame(parent) {
...
+  pm = std::make_unique<PubMaster, const std::initializer_list<const char *>>({"userTag"});
 }
 
 void Sidebar::mouseReleaseEvent(QMouseEvent *event) {
+  if (home_btn.contains(event->pos())) {
+    // TODO: maybe only create event while car started?
+    MessageBuilder msg;
+    auto userTag = msg.initEvent().initUserTag();
+    userTag.setId(0);
+    pm->send("userTag", msg);
```
incognitojam, line 56: @adeebshihadeh Is it okay to create the publisher in ui? Or should some other process be responsible for this?

adeebshihadeh: yeah, makes sense to do this in ui

adeebshihadeh:
Nice job, especially including the screenshot. Next time make sure that the submodules are on the merged commit on master; we have some additional CI that runs on openpilot master that would've failed in this case. Also would be nice to include a sample route with the new messages too.

selfdrive/ui/qt/sidebar.cc
```diff
@@ -43,9 +43,17 @@ Sidebar::Sidebar(QWidget *parent) : QFrame(parent) {
   setFixedWidth(300);
 
   QObject::connect(uiState(), &UIState::uiUpdate, this, &Sidebar::updateState);
+
+  pm = std::make_unique<PubMaster, const std::initializer_list<const char *>>({"userFlag"});
 }
 
 void Sidebar::mouseReleaseEvent(QMouseEvent *event) {
+  if (home_btn.contains(event->pos())) {
+    // TODO: maybe only create event while car started?
```
adeebshihadeh, line 52: this is alright as is, since loggerd doesn't run offroad anyway. this solution is temporary until we have some real design for event flagging.

incognitojam: ah okay!

PR #25522, opened 2022-08-22
Translations badges: concatenate into one badge

sshane:
No more manually editing README when new language is added/removed

One downsize is we lose linking to each language file from the badges, but trying to see if we can use a tag in svg


adeebshihadeh:
> One downsize is we lose linking to each language file from the badges, but trying to see if we can use a tag in svg

That's fine. Let's just merge as is if this works.

PR #25523, opened 2022-08-22
GMC: Sierra 2020-21 support

sshane:
Tuning looks almost identical (2.0 m/s^2). Car fingerprints as a Silverado. ACC is available on SLT PREMIUM PLUS PACKAGE, AT4 CARBONPRO™ EDITION, and DENALI ULTIMATE PACKAGE trims with Driver Alert
Package II.


adeebshihadeh (changes requested):
Let's also add to the 0.8.16 release notes

PR #25527, opened 2022-08-23
Hyundai: add software version query (for engine)

sshane:
On our Optima, it does not respond to the long or multi query, but does respond to 0x191 (hardware number) and 0x195 (software number, both return same value).

`0xF195 SYSTEM_SUPPLIER_ECU_SOFTWARE_VERSION_NUMBER: b'TJFAJNU06F201H03'`

On our Sonata, most of the ECUs respond to both the "long" and "multi" queries (with the engine relying on the multi query).

This is going to be part of a larger project, but I think we should remove multi if we can verify that the shorter long query is good (verified through notebook that if long FW is the same, there's no meaningful variation in the multi FW). Then we can replace multi with this new query to get the engine, so we don't have three requests for Hyundai, where the multi gets us different FW across the cars. Ideally long gets us all FW except the engine, and the new query gets us the engine


sshane:
EV6 dump: https://pastebin.com/2VhxyH9G

selfdrive/car/fw_versions.py
```diff
@@ -47,6 +47,8 @@ def p16(val):
 
 HYUNDAI_VERSION_REQUEST_LONG = bytes([uds.SERVICE_TYPE.READ_DATA_BY_IDENTIFIER]) + \
   p16(0xf100)  # Long description
+HYUNDAI_VERSION_REQUEST_SOFTWARE = bytes([uds.SERVICE_TYPE.READ_DATA_BY_IDENTIFIER]) + \
+  p16(0xf195)  # Software version number
```
sshane, line 51: 0xf191 returns the same and is what the debug tool uses for engine

PR #25528, opened 2022-08-23
pcmCruise GM: auto-resume

sshane:


twilsonco:
image
Okey dokey!

twilsonco:
Installing. I'll test in the morning. Looks like I'll be pressing the resume button in different scenarios. Stopped, engaging, increase set speed, resume behind lead?

sshane:
This is only set up to work on stock longitudinal GMs right now, it's to press the resume button for the user to resume.

twilsonco:
Ack!
That explains all the `CAMERA` I was seeing!

sshane:
I think the camera must manage the cruise state here (which makes sense why we need to send cancel to the camera), not the PT bus, so we need to spam resume on the camera bus.

sshane:
Will re-open when working

sshane:
So there appears to be some redundancy here. If you block the physical buttons to the camera, neither of the two actions separately cause the car to resume: pressing the resume button physically or sending the resume button to the camera. You need both sides of the relay to see the buttons being pressed in order to go

PR #25530, opened 2022-08-23
GM Port: Add Auto-resume support

JMPZ11:
**Description**
Updated GM carcontroller.py to send the ACC resume button on the PT bus when CC.cruiseControl.resume is true.
Updated the rolling counter logic to increment the RC - without incrementing we were sending the same RC value the car had just sent, and this would probably cause our message to be rejected.

Depends on Panda PR: https://github.com/commaai/panda/pull/1039

TODO: controlsd code that sets cruiseControl.resume may need review to ensure it works correctly with the stock long as well as OP long.

**Verification**
TBD - will need @sshane and / or community assistance to test

### DIFF
--- a/selfdrive/car/gm/carcontroller.py
+++ b/selfdrive/car/gm/carcontroller.py
@@ -109,7 +109,7 @@ class CarController:
 
     else:
       # Stock longitudinal, integrated at camera
-      if (self.frame - self.last_button_frame) * DT_CTRL > 0.04:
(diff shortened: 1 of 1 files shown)


sshane:
Already have PR here for pcmCruise: https://github.com/commaai/openpilot/pull/25528. We can do the rest of the GMs later. We also found the very specific timing values for cancel to optimize consistency, this would change that.

Issue #25493, opened 2022-08-19
sim: launch_openpilot.sh hangs at bootlog


jeroenlammersma:
### Describe the bug

Since commit [aa32ea0: loggerd: log all disk space usage in initData](https://github.com/commaai/openpilot/commit/aa32ea0f64e4035ac929e09ee85d268c3185221c#diff-b60345ac6d952fef3525d70194a17eebccfc84f5ce9c39d6537bfc707a5938e0) launch_openpilot.sh just hangs at (saving) 'bootlog to'. The UI doesn't show and openpilot doesn't start. See provided screenshot.

![launch_openpilot sh](https://user-images.githubusercontent.com/33349469/185657113-8782da64-4098-43e8-88e8-0c9c0a573345.png)

### OS Version

Ubuntu 20.04

### openpilot version or commit

03b074426a7de22b9f5dbfb9bb30f8325ac43b4a

equilibrium-liberalism:
similar issue reproduced in ubuntu 22.04 .

((openpilot) ) user:~/openpilot/tools/sim$     ./launch_openpilot.sh
**selfdrive/loggerd/bootlog.cc: bootlog to /home/hk/.comma/media/0/realdata/boot/2022-08-21--01-45-08**

but the replay by openpilot it self is fine [see](https://rewa-insights.com/t/awesome-automotive-autonomous/381/6?u=rewa-evija)

jeroenlammersma:
> similar issue reproduced in ubuntu 22.04 .
> 
> ((openpilot) ) user:~/openpilot/tools/sim$ ./launch_openpilot.sh **selfdrive/loggerd/bootlog.cc: bootlog to /home/hk/.comma/media/0/realdata/boot/2022-08-21--01-45-08**
> 
> but the replay by openpilot it self is fine [see](https://rewa-insights.com/t/awesome-automotive-autonomous/381/6?u=rewa-evija)

Not sure what the exact issue was, because the page won't load anymore. But it seemed that you are using CARLA 0.9.12, which is outdated. Update to 0.9.13 and it will work again.

jeroenlammersma:
Updated requirements and it works again. Resolved.

jeroenlammersma:
Currently having the exact same issue again:

![launch_openpilot hanging at saving bootlog step](https://user-images.githubusercontent.com/33349469/187771452-96e7492d-013c-4f64-b7c2-d43e9389db27.png)

[...]

Issue #25451, opened 2022-08-16
Simulator: update image output to NV12 from YUV420


jeroenlammersma:
### Describe the bug

I noticed that since commit [ea5b8cd (nv12: encoderd avoids a full frame copy)](https://github.com/commaai/openpilot/commit/ea5b8cdfb10686c972d8d2bc1778e6e6567e45a9) the simulator is displaying in grayscale. Also a black oval shape appears at the bottom of the screen. See the image below.

Is this supposed to happen after these changes, or is there maybe some configuration/installation or graphics driver issue what I might be facing? CARLA itself (/opt/carla-simulator/CarlaUE4.sh) runs in color just fine. It's the image stream coming from openpilot that's in grayscale. I also re-installed Ubuntu 20.04 and openpilot, but the issue was not resolved. I am running the current latest version of openpilot.

If this is the expected side effect, is there a way to get it in color again? And would this be an easy or difficult fix? I am happily willing to look into this issue, but I have no idea where to look or where to even start.

![openpilot-sim-greyscale](https://user-images.githubusercontent.com/33349469/184885506-ef3a0d42-e931-4f3e-95d3-08bfda280985.png)

### OS Version

Ubuntu 20.04

### openpilot version or commit

3fc001c945bb5e9092050bac1e55267dab63b9eb

### Additional info

- CPU: AMD Ryzen 9 5900x
- GPU: NVIDIA RTX 3070
  - Driver Version: 515.65.01

adeebshihadeh:
The CARLA bridge is still outputting images in YUY420; it just needs to be updated to NV12.

adeebshihadeh:
Done in  #26169

PR #25534, opened 2022-08-23
sim: launch_openpilot.sh fix and YUV to NV12 output

jeroenlammersma:
- Fix for openpilot not launching during simulation: #25493 

- Output images in NV12 format during simulation: #25451 

Converting is done from  YUV to NV12, not from RGB to NV12 directly.
I tried changing [rgb_to_yuv.cl](https://github.com/commaai/openpilot/blob/master/system/camerad/transforms/rgb_to_yuv.cl) to [real_debayer.cl](https://github.com/commaai/openpilot/blob/master/system/camerad/cameras/real_debayer.cll), but I am not able to compile the cl program. Also I am not sure if that will work.

Nevertheless, a todo is
- swapping to real_debayer.cl or creating a dedicated rgb_to_nv12.cl program.

This is also something I tried to create, but I had the same issue as mentioned above (see image).

clBuildProgram failed

I know what's going wrong, however I don't know the proper way to fix this in the codebase (I thought SCons would handle building).


selfdrive/loggerd/bootlog.cc
```diff
@@ -24,20 +25,22 @@ static kj::Array<capnp::word> build_boot_log() {
     i++;
   }
 
-  // Gather output of commands
-  std::vector<std::string> bootlog_commands = {
-    "[ -e /dev/nvme0 ] && sudo nvme smart-log --output-format=json /dev/nvme0",
-    "[ -x \"$(command -v journalctl)\" ] && journalctl",
-  };
+  if (!SIMULATION) {
```
adeebshihadeh, line 28: why is this hanging? I don't see why this is simulator specific; it doesn't hang on my machine

tools/sim/bridge.py
```diff
@@ -101,15 +101,22 @@ def _cam_callback(self, image, frame_id, pub_type, yuv_type):
...
     # convert RGB frame to YUV
     rgb = np.reshape(img, (H, W * 3))
     rgb_cl = cl_array.to_device(self.queue, rgb)
     yuv_cl = cl_array.empty_like(rgb_cl)
     self.krnl(self.queue, (np.int32(self.Wdiv4), np.int32(self.Hdiv4)), None, rgb_cl.data, yuv_cl.data).wait()
     yuv = np.resize(yuv_cl.get(), rgb.size // 2)
-    eof = int(frame_id * 0.05 * 1e9)
 
-    self.vipc_server.send(yuv_type, yuv.data.tobytes(), frame_id, eof, eof)
+    # convert YUV to NV12
```
adeebshihadeh, line 112: can you split this out into a separate PR?

jeroenlammersma: Will do.

selfdrive/loggerd/bootlog.cc
jeroenlammersma: I will do some more experimenting, if the issue persists, I will come back to this later.

jeroenlammersma:
updated requirements and now it works again. Case closed.

PR #25540, opened 2022-08-24
Toyota: log FCW on TSS-P

sshane:
- [ ] Add Prius V exception, no ACC_HUD (why is Highlander failing?)


sshane:
not immediately mergable, some cars missing messages

PR #25543, opened 2022-08-24
sim: outputting frames in NV12 format

jeroenlammersma:
Fix for:
- #25451 

Converting is done from YUV to NV12, not from RGB to NV12 directly. I tried changing [rgb_to_yuv.cl](https://github.com/commaai/openpilot/blob/master/system/camerad/transforms/rgb_to_yuv.cl) to [real_debayer.cl](https://github.com/commaai/openpilot/blob/master/system/camerad/cameras/real_debayer.cl), but I am not able to compile the cl program. Also I am not sure if that will work.

Nevertheless, a todo is:
- swapping to [real_debayer.cl](https://github.com/commaai/openpilot/blob/master/system/camerad/cameras/real_debayer.cll) or creating a dedicated rgb_to_nv12.cl program.

The latter is also something I tried to create, but I had the same issue as mentioned above (see image). I know why it fails (include headers), however I don't know the proper way to fix this in the codebase (I thought SCons would handle building).

clBuildProgram failed


jeroenlammersma:
Change too long ago..

PR #25548, opened 2022-08-24
Refactor torque control so PID is in m/ss

haraschax:

### DIFF
--- a/selfdrive/car/interfaces.py
+++ b/selfdrive/car/interfaces.py
@@ -114,11 +114,11 @@ class CarInterfaceBase(ABC):
     ret.maxLateralAccel = get_torque_params(candidate)['MAX_LAT_ACCEL_MEASURED']
 
     ret.pcmCruise = True     # openpilot's state is tied to the PCM's cruise state on most cars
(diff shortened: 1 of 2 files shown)


selfdrive/car/interfaces.py
```diff
@@ -141,10 +141,11 @@ def configure_torque_tune(candidate, tune, steering_angle_deadzone_deg=0.0, use_
 
     tune.init('torque')
     tune.torque.useSteeringAngle = use_steering_angle
-    tune.torque.kp = 1.0 / params['LAT_ACCEL_FACTOR']
-    tune.torque.kf = 1.0 / params['LAT_ACCEL_FACTOR']
-    tune.torque.ki = 0.1 / params['LAT_ACCEL_FACTOR']
+    tune.torque.kp = 1.0
+    tune.torque.kf = 1.0
+    tune.torque.ki = 0.1
```
adeebshihadeh, line 146: want these to be tunable per-car? if not, these should probably just be constants in latcontrol_torque.py

HaraldSchafer: Not sure yet if they need to be

PR #25551, opened 2022-08-25
Pin GitHub Actions 📌

GrantBirki:
# Pin GitHub Actions 📌

This pull request pins all GitHub Actions to an exact SHA. The reasoning for this is to prevent supply chain security attacks in 3rd party Action dependencies

## Learn More 📚 

You can learn more about Actions security hardening at the following links:

- https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions
- https://github.com/mheap/pin-github-action

---

Let me know if you have any questions and hopefully this PR helps to lock down all 3rd party Action dependencies 🔒

### DIFF
--- a/.github/workflows/badges.yaml
+++ b/.github/workflows/badges.yaml
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-20.04
     if: github.repository == 'commaai/openpilot'
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # pin@v3
       with:
         submodules: true
     - uses: ./.github/workflows/setup
--- a/.github/workflows/selfdrive_tests.yaml
+++ b/.github/workflows/selfdrive_tests.yaml
@@ -34,7 +34,7 @@ jobs:
     env:
       STRIPPED_DIR: /tmp/releasepilot
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # pin@v3
       with:
         fetch-depth: 0
         submodules: true
@@ -67,7 +67,7 @@ jobs:
     runs-on: ubuntu-20.04
     timeout-minutes: 50
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # pin@v3
       with:
         submodules: true
     - uses: ./.github/workflows/setup
@@ -85,7 +85,7 @@ jobs:
   #  runs-on: macos-latest
   #  timeout-minutes: 60
   #  steps:
-  #  - uses: actions/checkout@v3
+  #  - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # pin@v3
   #    with:
   #      submodules: true
   #  - name: Determine pre-existing Homebrew packages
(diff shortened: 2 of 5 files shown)


GrantBirki:
Also, did someone run CI on my pull request immediately after opening it? Or is Actions configured to run right away even on forked PRs? If this is the case, it might be worth while for the maintainers of this repo/org to lock down Actions permissions or audit them to ensure CI is only running in a trusted content

.github/workflows/badges.yaml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-20.04
     if: github.repository == 'commaai/openpilot'
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@2541b1294d2704b0964813337f33b291d3f8596b # pin@v3
```
adeebshihadeh, line 18: this seems excessive for GitHub's own actions

.github/workflows/selfdrive_tests.yaml
```diff
@@ -412,22 +412,22 @@ jobs:
...
-        uses: peter-evans/find-comment@1769778a0c5bd330272d749d12c036d65e70d39d
+        uses: peter-evans/find-comment@1769778a0c5bd330272d749d12c036d65e70d39d # pin@v2.0.0
         id: fc
         with:
           issue-number: ${{ github.event.pull_request.number }}
           body-includes: This PR makes changes to
       - name: Update comment
         if: ${{ steps.save_diff.outputs.diff != '' && env.AZURE_TOKEN != '' }}
-        uses: peter-evans/create-or-update-comment@b95e16d2859ad843a14218d1028da5b2c4cbc4b4
+        uses: peter-evans/create-or-update-comment@c9fcb64660bc90ec1cc535646af190c992007c32 # pin@v2.0.0
```
adeebshihadeh, line 422: this seems like a much more likely attack: subtly changing something like this hash in what's presented as a null-effect PR

.github/workflows/badges.yaml
GrantBirki: Regardless of where the Action comes from, it helps to pin Actions. Even if it is from Actions themselves :smiley:

.github/workflows/selfdrive_tests.yaml
GrantBirki: I would have left this one the same since its already pinned but tracking down that SHA without a version comment or something of the sort was tricky. Instead I just used the latest version from `peter-evans`. I use this exact Action myself and have checked the source code as well on this version

adeebshihadeh:
We only use version tags for GitHub published actions. I'm failing to see a **practical** case where this matters, though definitely open to ideas here.

GrantBirki:
> I'm failing to see a practical case where this matters

A practical case where this matters is in the event that a third party Action is compromised which leads to this repository becoming compromised as well.

For example, this repo uses a third party Action from codecov. codecov had a supply chain attack [just last year](https://blog.gitguardian.com/codecov-supply-chain-breach/)..

If the codecov (or any other Action) were to have a compromised release and this project was **not** using an exact pin to a SHA, then it could be vulnerable from that angle.

Even though this may seem like a minor nag to start pinning your Action versions, I do think it is important. People use this software to help them steer their vehicles. The implications of supply chain attacks here could have pretty raw effects on the physical well being of people.

---

TL;DR: I would pin to the exact version of an Action's SHA, everywhere, all the time, and regardless of who created the Action.

---

I hope that helps and I would be happy to discuss further if you still have questions 😃 

@adeebshihadeh

PR #25553, opened 2022-08-25
UI: update transforms in one place

deanlee:


adeebshihadeh:
CameraView, or anything in `qt/widgets/`, shouldn't really know anything about UIState. They're just generic widgets that can, and are, used elsewhere.

deanlee:
I agree with you 100 percent.

PR #25554, opened 2022-08-25
Portuguese: grammar and semantic corrections

guilhermicas:
reply for questions or changes

PR #25558, opened 2022-08-26
fix and improve pt-BR translation

AlexandreSato:


sshane:
Want to change the name of the file to `pt-BR`?

sshane:
I moved your changes over to https://github.com/commaai/openpilot/pull/25554 so the original author can get credit

PR #25560, opened 2022-08-26
Mazda: get blinker state from switch

Jafaral:


adeebshihadeh:
What's the difference? Can you post a route?

sshane:
Hmm, this removes the ability to use the convenience blinker for lane changes.

Screenshot from 2022-08-26 17-42-25

jyoung8607:
If there's benefit to using this signal, it can be wrapped with `update_blinker_from_stalk` instead of the existing `update_blinker_from_lamp`.

Jafaral:
@sshane , not sure where you pulled the data, here is what I see: \
image

Jafaral:
I thought  that is a cleaner signal to use, it is the actual turn signal, and not the lamp signal. Or am I missing something from OP logic here?

Jafaral:
I see the point now. Updated using the  stalk function per @jyoung8607 suggestion. Thanks!

sshane:
What happens if you use convenience blinker for left, then right to cancel the left?

Jafaral:
I will test that scenario.

sshane:
Have a route with the changes?

Jafaral:
@sshane   8c521516df3527d8|2022-08-30--10-05-04--2

There was a bit of a lag before with lamp signals, now it is instantaneous.

adeebshihadeh:
The blinker preference is generally in this order: continuous signal when blinker is activated, lamp state, then stalk state. The lamp state is better than the stalk state since you don't need to worry about things like convenience blinkers.

Jafaral:
What is stake state vs continous signal  @adeebshihadeh ?

Jafaral:
Because what I have above looks like continuous signal to me.

adeebshihadeh:
As far as I understand, this new signal is the position of the blinker stalk, which doesn't capture things like the convenience blinkers.

PR #25564, opened 2022-08-26
ui: fix possible memory leak in EGL

deanlee:
The EGL implementation no longer take ownership of the fd.https://registry.khronos.org/EGL/extensions/EXT/EGL_EXT_image_dma_buf_import.txt:

>     3. Does ownership of the file descriptor pass to the EGL library?

    ANSWER: No, EGL does not take ownership of the file descriptors. It is the
    responsibility of the application to close the file descriptors on success
    and failure.

### DIFF
--- a/selfdrive/ui/qt/widgets/cameraview.cc
+++ b/selfdrive/ui/qt/widgets/cameraview.cc
@@ -296,13 +296,16 @@ void CameraViewWidget::vipcConnected(VisionIpcClient *vipc_client) {
 #ifdef QCOM2
   egl_display = eglGetCurrentDisplay();
 
-  for (auto &pair : egl_images) {
-    eglDestroyImageKHR(egl_display, pair.second);
+  for (auto &[_, s] : egl_images) {
+    eglDestroyImageKHR(egl_display, s.handle);
+    close(s.fd);
   }
   egl_images.clear();
-
   for (int i = 0; i < vipc_client->num_buffers; i++) {  // import buffers into OpenGL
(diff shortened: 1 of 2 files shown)


selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -296,13 +296,16 @@ void CameraViewWidget::vipcConnected(VisionIpcClient *vipc_client) {
...
+    close(s.fd);
   }
   egl_images.clear();
-
   for (int i = 0; i < vipc_client->num_buffers; i++) {  // import buffers into OpenGL
-    int fd = dup(vipc_client->buffers[i].fd);  // eglDestroyImageKHR will close, so duplicate
+    // the EGL_EXT_image_dma_buf_import spec was revised after 2013:
+    // EGL does not take ownership of the file descriptors. It is the
+    // responsibility of the application to close it.
+    int fd = dup(vipc_client->buffers[i].fd);
```
deanlee, line 308: Still using dup to make sure it works under the old spec. It should be harmless to close fd twice.

jwooning:
Although the docs say it doesn't take ownership, it does in fact close the fd. Also, closing an fd twice can be an issue if the fd is reused.

deanlee:
oh I get it. 
another question: if so, should we dup fd for `EGL_DMA_BUF_PLANE0_FD_EXT` and `EGL_DMA_BUF_PLANE1_FD_EXT` respectively?

jwooning:
I would expect it to throw an error if it would try to close the same fd twice, so I guess it doesn't matter, but I didn't test this, might be interesting to see what happens if you test this out.

PR #25565, opened 2022-08-26
ui/main: Dont install event filter on tici

deanlee:
resetInteractiveTimout will be trigged by motion on TICI


adeebshihadeh:
I'm not sure what this is supposed to do? We need the event filter to reset the timeout on any display touch.

deanlee:
my bad

PR #25568, opened 2022-08-26
user flag ratelimit

incognitojam:
Limit creating flags to once per second


deanlee:
IMO, it might be better to use a variable to store the last timestamp. too many QTimers already.

adeebshihadeh:
This doesn't really belong in the UI. This should be handled on the connect side, ideally by fixing the timeline performance.

PR #25570, opened 2022-08-26
Toyota: same standstill check as panda

sshane:


sshane:
Checked 33 platforms over 520 dongles, the minimum wheel speed we ever saw that's != 0 is `0.0027777778450399637` m/s which is exactly `0.01` kph, and above our current threshold of 0.001 m/s which is 0.0036 kph. So checking < 0.001 is the same as == 0

sshane:
panda can't currently be bumped, waiting for a fix

PR #25572, opened 2022-08-27
Car docs: add auto-resume column

sshane:
Need it in car interface (and not car info), because it's conditional on openpilot long control and whether we can spam the buttons

Preview: https://github.com/commaai/openpilot/blob/auto-resume-car-info/docs/CARS.md


sshane:
@adeebshihadeh anything else you want in here?

jyoung8607:
To follow up on my DM conversation with @sshane, I think VW's auto resume can be mostly described by `minEnableSpeed == -1`. We also have a case of not supporting SnG with manual transmissions, but technically it's a bug that we don't encode a minEnableSpeed for those cars with stock ACC (it's in the openpilot long draft PR) so we should just fix that bug instead.

The sticky part is describing FtS cars, the cars that will stop but not *stay* stopped. I don't quite know how to identify model-year combinations with that problem yet. I propose we use the column-specific footnote capability to add a footnote in the new resume-from-stop column saying "provided the car has an electric parking brake", and a similar exception for the comma.ai/vehicles doc generator. I *believe* that makes an accurate statement, or at least more accurate than we make today.

jyoung8607:
On further thought, we'll have to mention manual transmissions in the footnote/caveat. We live-identify those carParams within a model rather than making separate models, any supported VW could theoretically have a manual transmission, and the docs will have to reflect that.

sshane:
@jyoung8607 want to open that PR to add the footnote?

docs/CARS.md
```diff
@@ -6,213 +6,213 @@ A supported vehicle is one that just works when you install a comma three. All s
...
+|Toyota|Camry 2021-22|All|openpilot|0 mph[<sup>4</sup>](#footnotes)|0 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-full.svg)](##)|Toyota|
+|Toyota|Camry Hybrid 2018-20|All|Stock|0 mph|0 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-empty.svg)](##)|Toyota|
+|Toyota|Camry Hybrid 2021-22|All|openpilot|0 mph|0 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-full.svg)](##)|Toyota|
+|Toyota|Corolla 2017-19|All|Stock[<sup>3</sup>](#footnotes)|19 mph|0 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-empty.svg)](##)|Toyota|
+|Toyota|Corolla 2020-22|All|openpilot|0 mph|0 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-full.svg)](##)|Toyota|
+|Toyota|Corolla Cross (Non-US only) 2020-21|All|openpilot|17 mph|0 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-full.svg)](##)|Toyota|
+|Toyota|Corolla Cross Hybrid (Non-US only) 2020-22|All|openpilot|17 mph|0 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-full.svg)](##)|Toyota|
```
yuzisee, line 162: Just a question (I know this is an old PR) but what does it mean for `Resume from stop`=**true** if `No ACC accel below`=**17 mph**

i.e. Isn't "stopped" below 17 mph? So can the Corolla Cross actually resume from stopped? If so, how does it get from stopped to 17 mph?

PR #25574, opened 2022-08-27
test onroad: stricter CPU usage thresholds

adeebshihadeh:


adeebshihadeh:
Unfortunately, it's not consistent enough to do this.

Issue #2395, opened 2020-10-24
HKG resume logic not working on Hyundai Genesis(2015-2016)


xps-genesis:
**Describe the bug**

The new resume logic in Master-Ci for HKG doesn't work on the Hyundai Genesis(2015-2016), This can be an issue with other older HKG cars. The resume logic is triggered when the standstill flag from SCC is set. This often allows the SCC to accept once or twice during a stop but often misses when the lead starts to move. 

**How to reproduce or log data**

ACC activated behind a stopped target vehicle to start the resume logic 

**Expected behavior**

Resume and continue moving when the lead starts to move.

**Device/Version information (please complete the following information):**
 - Device: [comma two]
 - Dongle ID: [6fe86b4e410e4c37]
 - Route: [6fe86b4e410e4c37|2020-10-23--19-14-59]
 - Timestamp: [497 seconds]
 - Version: [master-ci]
 - Car make/model: [Hyundai Genesis 2015-16]

adeebshihadeh:
Is this still an issue?

xps-genesis:
yup still doesn't work on older vehicles

SeanCox:
I tested the PR and left some notes for this https://github.com/commaai/openpilot/pull/25579

SeanCox:
@adeebshihadeh This issue has been fixed.

PR #25579, opened 2022-08-27
Hyundai: 2015-16 Genesis resume from standstill fix

sunnyhaibin:
Allows 2015-16 Hyundai Genesis to accept the resume command from standstill when the lead car drives away.
* route ID
  * `3687c15bdfe09295|2022-08-27--15-54-09`
  * `3687c15bdfe09295|2022-09-02--19-54-13`

**Resolves**
- https://github.com/commaai/openpilot/issues/2395


sshane:
Hmm, the current command doesn't work on that car?

sunnyhaibin:
The current command is inconsistent according to multiple users. This issue has been around throughout the years, and this solution has been in forks for quite some time.

> Hmm, the current command doesn't work on that car?

adeebshihadeh:
I don't really like that there are two code paths here. Let's try to merge them. The main two differences here are incrementing the counter and the max number of messages.

sunnyhaibin:
User tested with the merged version and works for his 2015 Genesis: `3687c15bdfe09295|2022-09-02--19-54-13`

adeebshihadeh:
Glad we got them merged. Has this been tested on cars where the current resume already works? Which change fixes the Genesis: frequency, sending them in a burst, or both?

sunnyhaibin:
> Glad we got them merged. Has this been tested on cars where the current resume already works? Which change fixes the Genesis: frequency, sending them in a burst, or both?

It does work with the Hyundai Sonata, I could give it a go tomorrow to confirm.

Both the frequency and sending them in a burst worked for the Genesis.

sunnyhaibin:
> If it was inconsistent it sounds like a timing issue where we keep sending the message at the wrong time to get the ECU to pick it up. Has anyone tried simply changing the interval to an odd number like 0.09 sec (11.11 Hz) or 0.07 sec (14.29 Hz)?

We've tested with the odd numbers and they all behave the same. Users have been testing with this on my fork for weeks and it has been consistently resuming. I believe this is good to merge.

sshane:
@adeebshihadeh want me to test this on our cars here?

sunnyhaibin:
@adeebshihadeh @sshane Is there anything else needed for this PR to merge? It has been tested by multiple users and is working as expected with no issues.

adeebshihadeh:
I don't doubt that this works, but it still seems like it's unclear what the fix is here. I'd like to know whether it's the sending in bursts, the frequency change, or both.

SeanCox:
I just tested both parts of this code separately. The 20hz didn't seem to fix anything but the extra messages did allow my car to start from a stop.

Here is the branch I used to test: 
https://github.com/SeanCox/openpilot/tree/Hyundai--2015-16-Genesis-resume-from-standstill-fix-extra-messages

Here is a route with it working:
85eab094426c2649|2023-02-02--17-56-18--0

sunnyhaibin:
> I just tested both parts of this code separately. The 20hz didn't seem to fix anything but the extra messages did allow my car to start from a stop.
> 
> Here is the branch I used to test: https://github.com/SeanCox/openpilot/tree/Hyundai--2015-16-Genesis-resume-from-standstill-fix-extra-messages
> 
> Here is a route with it working: 85eab094426c2649|2023-02-02--17-56-18--0

Thanks for testing this. Are you able to also provide a route with you testing the 20Hz change and failed to spam the button? Would be nice to see the comparisons between both of them.

SeanCox:
@sunnyhaibin Here are my notes from testing:

Cruse Control Resume Tests

No changes
Branch:
https://github.com/SeanCox/openpilot 
Route:
85eab094426c2649|2023-02-02--17-14-35--6
Test: Failed

20hz
Branch:
https://github.com/SeanCox/openpilot/tree/Hyundai--2015-16-Genesis-resume-from-standstill-fix-20hz
Route:
85eab094426c2649|2023-02-02--17-41-38–0
Test:
Failed

Extra messages
Branch:
https://github.com/SeanCox/openpilot/tree/Hyundai--2015-16-Genesis-resume-from-standstill-fix-extra-messages
Route:
85eab094426c2649|2023-02-02--17-56-18--0
Test: passed

Both 20hz and Extra Messages:
Branch:
https://github.com/SeanCox/openpilot/tree/Hyundai--2015-16-Genesis-resume-from-standstill-fix
Route:
Not tested
Test: 
Not tested

SeanCox:
@adeebshihadeh What do you think about the state of this PR? Is there anything else needs to be done/tested to get this merged in? Thank you.

adeebshihadeh:
Thanks for testing @SeanCox; your tests were quite helpful. @sshane can you take this one over and test on our sonata and optima?

sshane:
@SeanCox I made a minor change to the counters that should have no impact. Can you stress test this one more time?

sshane:
Realized my refactor made it send once, then wait 10 frames, then send 5 times in a row, rather than the originally intended send 5 times immediately, wait 10 frames, send 5 times...etc.

However this seems to be a good compromise, as sometimes the Optima I tested this one resumed with just one frame, and sometimes it needed the 5 frame burst to get it resuming, so trying the original behavior first and waiting should be fine. The original `resume_count` logic also would have this bug if we exited standstill right before resetting the last button frame.

Screenshot from 2023-02-14 21-54-59
Screenshot from 2023-02-14 21-55-11

SeanCox:
> @SeanCox I made a minor change to the counter that should have no impact. Can you test it out real quick?

I will test this later today.

sshane:
On our Sonata:

```python
Frames to resume:
This PR: 5.804165272815535 frames, 103 runs
Stock: 5.6490840389312975 frames, 131 runs

**Comment by sshane** [comment]
@SeanCox did you test it out

**Comment by sshane** [comment]
I see a few cases from your dongle of this working as expected

File: selfdrive/car/hyundai/carcontroller.py [hunk 1]
```diff
@@ -169,7 +169,8 @@ def update(self, CC, CS, now_nanos):
           if (self.frame - self.last_button_frame) * DT_CTRL > 0.1:
             # send 25 messages at a time to increases the likelihood of resume being accepted
             can_sends.extend([hyundaican.create_clu11(self.packer, self.frame, CS.clu11, Buttons.RES_ACCEL, self.CP.carFingerprint)] * 25)
-            self.last_button_frame = self.frame
+            if (self.frame - self.last_button_frame) * DT_CTRL > 0.15:
```
sshane, line 172:
```suggestion
            if (self.frame - self.last_button_frame) * DT_CTRL >= 0.15:
```

SeanCox:
yeah, I was just about to post this route 85eab094426c2649|2023-02-15--18-42-04--0

PR #25584, opened 2022-08-28
New Translation for Portuguese from Portugal

Noillek:
Used Portuguese-BR as reference... to create a Portuguese-PT version.
My knowledge of programming is super limited but i guess that this would work well. Let me know if can be included.

Thank you

Regards,
Pedro Ferrao (Noillek)


sshane:
Do this in the other PR

PR #25589, opened 2022-08-29
Ram: don't check FW revision

sshane:
According to @realfast, the last two characters are just the FW revision, and don't describe functionality of the EPS.


sshane:
@realfast look good?

PR #25607, opened 2022-08-30
HKG: add electric motor ECU query

sshane:
Hyundai EVs do not appear to have a dedicated engine or transmission ECU (no response whatsoever when querying many data identifiers).

Puts this ECU in data-collection mode. This is also the only ecu that returns the vin, so it likely replaces function of the engine and transmission, will check to make sure

- Checked Hyundai Tech Info, Ioniq 5 does not have a transmission module (TCM) listed, while 2020 Sonata does.
- Of course, no engine ECU at 0x7e0 either


selfdrive/car/fw_versions.py
```diff
@@ -339,12 +339,17 @@ def match_fw_to_car_exact(fw_versions_dict):
       if ecu_type not in ESSENTIAL_ECUS and not len(found_versions):
         continue
 
+      # Ignore ecus in the database for data collection
```
adeebshihadeh, line 342: doesn't have to be in this PR or anything, but we should think about ways to make this more explicit

selfdrive/car/tests/test_fw_fingerprint.py
```diff
@@ -45,15 +45,19 @@ def test_no_duplicate_fw_versions(self):
             self.assertFalse(len(duplicates), f"{car_model}: Duplicate FW versions: Ecu.{ECU_NAME[ecu[0]]}, {duplicates}")
 
   def test_blacklisted_ecus(self):
-    blacklisted_addrs = (0x7c4, 0x7d0)  # includes A/C ecu and an unknown ecu
+    blacklisted_addrs = {
+      "subaru": (0x7c4, 0x7d0),  # includes A/C ecu and an unknown ecu
+      "hyundai": (0x7e2,),  # currently collecting Ecu.vmcu fingerprints
+    }
```
adeebshihadeh, line 51: this is fundamentally the same problem

selfdrive/car/fw_versions.py
sshane: yeah, was thinking the same thing

jyoung8607: +1 I would like to use something like this to gather/prep for adding ABS/ESP and combinationMeter for VW.

sshane:
According to this article from Bosch about VCUs (which is what the parts sites report for the hardware id for the ECU on 0x7e2 for our EV6), the VCU assumes control of the inverter, battery management, engine and transmission control on fully EV cars: https://www.bosch-mobility-solutions.com/en/solutions/vehicle-computer/vehicle-control-unit/

sshane:
Not sure if we need this just yet

Issue #25603, opened 2022-08-29
proclog: missing processes sometimes 


adeebshihadeh:
CI occasionally hits this running `test_onroad.py`, here's an example failure:

```
[...]

PR #25610, opened 2022-08-30
proclog: handle zombie processes

deanlee:


adeebshihadeh:
Is this contributing to #25603? I'm not sure we want to ignore zombie processes.

deanlee:
Other than that, I haven't found any issues yet.

PR #25614, opened 2022-08-30
Ford Explorer 2020-21

adeebshihadeh:
- [ ] steering safety
- [ ] make steering good
  more investigation required - probably need closed loop controller for good steering
- [x] driver "torque blending"
  seems fine already?
- [x] look into FCWs
  disabled radar for now, probably not good yet
- [x] resume from ACC standstill
- [x] random disengagements
  looks like stock TJA was still enabled any maybe interfering/wanted driver wheel touch
- [x] detect non-adaptive cruise
- [x] test route in routes.py
- [x] add to process replay

Test route on 2021 Explorer King Ranch: `62241b0c7fea4589|2022-09-01--15-32-49`
Other firmware is from 2020 Explorer XLT (`59a107f5793d9cc0`)

see also https://github.com/commaai/panda/pull/1052


incognitojam:
I checked those FCW alerts in `62241b0c7fea4589|2022-08-30--13-12-32--4` and it looks like it wasn't the stock FCW

incognitojam:
Test drive on `8ab3348`
Route: `62241b0c7fea4589|2022-08-31--17-01-56`

PR #25615, opened 2022-08-30
modeld: PC Thneed prereqs

geohot:
Makes the tinygrad PR smaller.

Mostly it refactors thneed.cc into thneed_common.cc (used on PC), and thneed_qcom2.cc (Adreno specific stuff). A good refactor regardless.

It also adds passing in the cl_context to the model runners so they can reuse it. On PC this is required.


geohot:
Will merge if tests pass

PR #25617, opened 2022-08-31
controls: enter overriding state for steering override

sshane:
We can make a new state for lateral override, however it's not clear which should take precedence. Making `OpenpilotState.overriding` a superset of both gas override and steering override makes the most sense to me, then you just need to check which event types are active/added.


selfdrive/controls/lib/events.py
```diff
@@ -31,7 +31,8 @@ class Priority(IntEnum):
 class ET:
   ENABLE = 'enable'
   PRE_ENABLE = 'preEnable'
-  OVERRIDE = 'override'
+  OVERRIDE_LAT = 'overrideLat'
+  OVERRIDE_LONG = 'overrideLong'
```
sshane, line 35: Couldn't think of a more generic way to do this. If we keep the one override event type (which I'd prefer), you can't easily use the event types for longActive anymore (`self.events.any(ET.OVERRIDE)` would be true for lat overriding). Aside from doing `and CS.brakePressed` or `and CS.steeringPressed`, but we decided against that when adding longActive for ndog.

adeebshihadeh (approved):
LGTM, I'd just fully spell out lat and long. Also make sure the overrideLong event type field has the same ordinal as the original override type to preserve backwards compatibility with old routes.

PR #25619, opened 2022-08-31
USB power mode cleanup

### LINKED ISSUE #25601: Cleanup USB power modes
Seems like the USB power modes are now just a convoluted way to enable/disable bootkick. Will need to update panda, thermald, boardd, and [AGNOS](https://github.com/commaai/agnos-builder/blob/master/userspace/files/disable_bootkick.py).

adeebshihadeh:
This is all legacy stuff that now just means "disable bootkick" on the comma three. All bootkick logic is now in panda.


robbederks (approved):
lgtm if the red panda ev6 setup will still bootkick!

PR #25620, opened 2022-08-31
Toyota: fix permanent LKAS faults

### LINKED ISSUE #24943: Toyota: EPS fault while overriding
This is a rare steering fault while overriding. There's only one instance on 0.8.14:
```
TOYOTA RAV4 HYBRID 2017                  c6f9dea34ce475de|2022-06-05--20-30-22--8
```

sshane:


BrettLynch123:
Hi I have tested this on a Toyota 2022 Rav 4 Hybrid and works well. Made this short segment of it working 2475fb3eb2ffcc2e|2022-08-31--17-49-11--0

I also tested this on my 2020 Toyota Corolla Hybrid Sedan and made this short segment of it working
2475fb3eb2ffcc2e|2022-08-31--18-31-21--0

Great work :)

sshane:
I don't see `2475fb3eb2ffcc2e|2022-08-31--18-31-21` going above the 500 unit threshold to trigger the fault, you'll need to apply about 2x more torque on the wheel for that one. 

However you were above 500 units on `2475fb3eb2ffcc2e|2022-08-31--17-49-11` for more than 4 seconds without a permanent fault!

BrettLynch123:
Ok I had another try in the corolla and the only way I could think of to add more torque was to go faster as the wheel was turned all the way. So in this one it did disengage but was because stability control kicked in. I tried turning it off then OP wouldn't engage (good safety feature). But I think it was engaged enough to prove it working in the corolla. I also did another test a little slower so stability didn't kick in however It didn't seem to record. Possibly turned car off and on again too quickly not sure what went wrong.
This is the one that recorded 2475fb3eb2ffcc2e|2022-09-01--08-04-17--0

sshane:
Some results from 600,420 segments:

```python
c4f9de5fce000d77|2022-04-30--16-55-29--30, TOYOTA PRIUS TSS2 2021: Frames above 500 torques: 48

59ae3ab324f1ef46|2022-04-03--08-27-04--20, TOYOTA PRIUS TSS2 2021: Frames above 500 torques: 49

9d5c10b6f66f736a|2022-04-17--17-22-13--18, TOYOTA COROLLA TSS2 2019: Frames above 500 torques: 49
```

PR #25624, opened 2022-08-31
Hyundai: enforce ACC main state with controls allowed

sunnyhaibin:
Enforce ACC main state with stock/openpilot longitudinal control in openpilot and panda safety.

Stock ACC: `SCC11|MainMode_ACC`

openpilot ACC: `EMS16|CRUISE_LAMP_M`
- Available on all currently supported openpilot longitudinal HKG cars
- Sync with ACC MAIN button press
- Will be copied and sent with `SCC11|MainMode_ACC` to the car when ACC MAIN is pressed

**Requires**
- https://github.com/commaai/panda/pull/1054


sunnyhaibin:
Upon testing on some older models, the signal does not do anything when the CRUISE MAIN button is pressed. Need to investigate and see if there is another common signal to use for those cars.

sshane:
Is there a signal in TCS that describes this?

sunnyhaibin:
> @sunnyhaibin if we don't expect this to be mergable soon, can we close until it is?

Yes that works for me. I'll open a new PR when it's ready.

PR #25627, opened 2022-08-31
Hyundai CAN-FD: common brake pressed signal

adeebshihadeh:


sshane:
Duplicate of https://github.com/commaai/openpilot/pull/26170

PR #25633, opened 2022-09-01
Add nvme-cli as common requirement

jeroenlammersma:
Since ```nvme``` command is used in: [bootlog.cc](https://github.com/commaai/openpilot/blob/4bb399ba3c13953680522707bba662527fa771b7/selfdrive/loggerd/bootlog.cc#L29)


adeebshihadeh:
Not really necessary when not on AGNOS. It's fine if the command fails on PC.

Issue #25625, opened 2022-08-31
sim: bootlog_commands causes bootlog to hang


jeroenlammersma:
### Describe the bug

The same issue is occurring again, same as:
- #25493 

![launch_openpilot hanging at saving bootlog step](https://user-images.githubusercontent.com/33349469/187779619-f3fd1cf9-5b88-407f-ac69-83b77f33a10e.png)

I checked out on master branch, did requirements update, even installed openpilot on a fresh and clean Ubuntu, but it still hangs at the saving bootlog step.

Do you now also have this issue?

---

I, however, did some more experimenting and found out that line 215 in [util.cc](https://github.com/commaai/openpilot/blob/272b6cc8c3c7ec73ee3a259ee44cfce2c4e85a1f/common/util.cc#L215) in the function 'check_output' is the culprit.

### experiment 1: comment out line 39 in [bootlog.cc](https://github.com/commaai/openpilot/blob/272b6cc8c3c7ec73ee3a259ee44cfce2c4e85a1f/selfdrive/loggerd/bootlog.cc#L39)
**launch_openpilot.sh works**
![code_1](https://user-images.githubusercontent.com/33349469/187779586-811a33f1-c3b7-46dd-97d1-5a735ef88339.png)

### experiment 2: call check_ouput without storing in result in [bootlog.cc](https://github.com/commaai/openpilot/blob/272b6cc8c3c7ec73ee3a259ee44cfce2c4e85a1f/selfdrive/loggerd/bootlog.cc#L39)
**hangs: launch_openpilot.sh does not work**
![code_2](https://user-images.githubusercontent.com/33349469/187780506-2eeed0aa-7bb6-413d-a77e-67b957b5e1a4.png)

### experiment 3: comment out everything, except pipe in check_output function in [util.cc](https://github.com/commaai/openpilot/blob/272b6cc8c3c7ec73ee3a259ee44cfce2c4e85a1f/common/util.cc#L212)
**hangs: launch_openpilot.sh does not work**
![code_3_](https://user-images.githubusercontent.com/33349469/187780514-b62602c1-4374-4fb3-8fc7-61367bc360fd.png)

I do not know what this line is exactly doing, so I was not able to do some further experimentation.

### OS Version

Ubuntu 20.04

### openpilot version or commit

272b6cc8c3c7ec73ee3a259ee44cfce2c4e85a1f

jeroenlammersma:
[...]

PR #25634, opened 2022-09-01
Add nvme bootlog command only when device is TICI

jeroenlammersma:
This fixes issue:
- #25625 

**This works for the simulator, but not sure if removing ```sudo``` from the bootlog command is possible for the comma hardware, so needs validation.**

### DIFF
--- a/selfdrive/loggerd/bootlog.cc
+++ b/selfdrive/loggerd/bootlog.cc
@@ -26,9 +26,12 @@ static kj::Array<capnp::word> build_boot_log() {
 
   // Gather output of commands
   std::vector<std::string> bootlog_commands = {
-    "[ -e /dev/nvme0 ] && nvme smart-log --output-format=json /dev/nvme0",
     "[ -x \"$(command -v journalctl)\" ] && journalctl",
   };
+
+  if (Hardware::TICI()) {
+    bootlog_commands.push_back("[ -e /dev/nvme0 ] && sudo nvme smart-log --output-format=json /dev/nvme0");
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
This is necessary for the comma three. Can you move that command behind an `if (Hardware::TICI)`?

jeroenlammersma:
> This is necessary for the comma three. Can you move that command behind an `if (Hardware::TICI)`?

Sure. Does the order of the commands matter?

jeroenlammersma:
> This is necessary for the comma three. Can you move that command behind an `if (Hardware::TICI)`?

Forgot to mention, next tot the proposed change, the following change is also a possibility to achieve the same result:
``` C++
  // Gather output of commands
  std::vector<std::string> bootlog_commands;
  if (Hardware::TICI())
    bootlog_commands.push_back("[ -e /dev/nvme-1 ] && sudo nvme smart-log --output-format=json /dev/nvme0");
  bootlog_commands.push_back("[ -x \"$(command -v journalctl)\" ] && journalctl");
```
(This change will not change the order of the commands btw).

However, I was not sure which refactor/C++ style you prefer (as comma) and if there are plans to add more bootlog commands in the (near) future.

And if you have some time and it's not to much of an inconvenience, I'd love to hear some feedback about which of the two is considered better coding or is preferred. I still got much to learn.

Cheers,
Jeroen

adeebshihadeh:
Yes, this is how I would have done it. The order of the commands doesn't matter.

PR #25635, opened 2022-09-01
modeld: delete unused SNPE stuff after move to tinygrad

geohot:
Losing over a tinygrad's worth of code!

Removes thneed's CL interceptor, it's saving ability (useless without interceptor), the C model compiler, the SNPE optimizer, and the old conv kernels.


geohot:
@adeebshihadeh If you trust the tests, this should be good to merge :)

I left the tinygrad file list in SConstruct, feel free to change.

PR #25637, opened 2022-09-01
Optima 2019: set LKAS icon correctly

sshane:
I think we assume the cars to have the ability to show individual lane lines, which we do through `CF_Lkas_LdwsSysState` (as well as for showing alerts? this doesn't seem right), however on the Optima, a sys state of 4 for disabled is to show the lines (but Optima chooses green for the color).

Another difference is the `CF_Lkas_SysWarning` signal, it only does anything with `4`, but it's audible, not silent.

Another difference is `CF_Lkas_LdwsOpt_USM` which doesn't seem to do anything, besides if you set it to 0, `CF_Lkas_LdwsSysState` won't work to show any LKAS icon at all.

Another difference is `CF_Lkas_LdwsActivemode` and `CF_Lkas_FcwOpt_USM` don't seem to do anything.

### DIFF
--- a/selfdrive/car/hyundai/hyundaican.py
+++ b/selfdrive/car/hyundai/hyundaican.py
@@ -37,7 +37,7 @@ def create_lkas11(packer, frame, car_fingerprint, apply_steer, steer_req,
     # Note: the warning is hidden while the blinkers are on
     values["CF_Lkas_SysWarning"] = 4 if sys_warning else 0
 
-  # Likely cars without the ability to show individual lane lines in the dash
+  # Likely cars lacking the ability to show individual lane lines in the dash
   elif car_fingerprint in (CAR.KIA_OPTIMA,):
     # SysWarning 4 = keep hands on wheel + beep
     values["CF_Lkas_SysWarning"] = 4 if sys_warning else 0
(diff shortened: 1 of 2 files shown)


selfdrive/car/hyundai/hyundaican.py
```diff
@@ -37,12 +37,26 @@ def create_lkas11(packer, frame, car_fingerprint, apply_steer, steer_req,
     # Note: the warning is hidden while the blinkers are on
     values["CF_Lkas_SysWarning"] = 4 if sys_warning else 0
 
+  # Likely cars without the ability to show individual lane lines in the dash
```
sshane, line 40:
```suggestion
  # Likely cars lacking the ability to show individual lane lines in the dash
```

selfdrive/car/hyundai/hyundaican.py
```diff
@@ -37,12 +37,26 @@ def create_lkas11(packer, frame, car_fingerprint, apply_steer, steer_req,
...
+  # Likely cars lacking the ability to show individual lane lines in the dash
+  elif car_fingerprint in (CAR.KIA_OPTIMA,):
+    # SysWarning 4 = keep hands on wheel + beep
+    values["CF_Lkas_SysWarning"] = 4 if sys_warning else 0
+
+    # SysState 0 = no icons
+    # SysState 1-2 = white car + lanes
+    # SysState 3 = green car + lanes, green steering wheel
+    # SysState 4 = green car + lanes
+    values["CF_Lkas_LdwsSysState"] = 3 if enabled else 1
```
sshane, line 49: checked and this is what we show in the Sonata (green car + lanes, AND green steering wheel) while stock shows neither without lines and active steering

PR #25639, opened 2022-09-02
VW: Add fingerprint for 2015 Audi A3 MK3 (US)

ogiechogie:
Add AUDI A3 MK3 FW values from 2015 Audi A3 2.0T Quattro
* Update values.py


jyoung8607 (approved):
Thank you for the data!

sshane:
Can you provide a dongle id?

ogiechogie:
yeah sure its `2a5a0210bcc9b83b`

PR #25640, opened 2022-09-02
rename esp ECU to abs

incognitojam:
needs cereal https://github.com/commaai/cereal/pull/353


sshane:
any blockers?

incognitojam:
It's called ABS on Ford... want to check every brand?

PR #25650, opened 2022-09-02
Portuguese: add e2e long translations

AlexandreSato:


selfdrive/ui/translations/main_pt-BR.ts
```diff
@@ -1199,12 +1199,12 @@ trabalho definido</translation>
     <message>
         <location filename="../qt/offroad/settings.cc" line="58"/>
         <source>🌮 End-to-end longitudinal (extremely alpha) 🌮</source>
-        <translation type="unfinished"></translation>
+        <translation>🌮 End-to-end longitudinal (experimental) 🌮</translation>
```
sshane, line 1202: no translation?

AlexandreSato: A more explicit translation would be too long and perhaps lose the semiotics. Ex: "solução de ponta a ponta" means "end-to-end". All the others worlds are the same on portuguese from Brazil.

sshane:
thanks for the PR!

PR #25447, opened 2022-08-16
Honda: cleanup minEnableSpeed

csouers:
**Description**: Set the minEnableSpeed parameter more generically. All Honda Bosch and the Nidec Civic can do stop-and-go without the use of a gas pedal interceptor.

**Verification**: Local replay and unit tests

PR #25660, opened 2022-09-04
Toyota minEnableSpeed cleanup

vanillagorillaa:
To go with #25447


adeebshihadeh:
This isn't quite the same. #25447 was generic for all Honda Bosch.

PR #25661, opened 2022-09-04
replace tuples with dicts for better performance

SG5:
**Description** 
dict has better berformance for creating and linear seach

**Verification** [](Explain how you tested the refactor for regressions.)

### DIFF
--- a/selfdrive/athena/athenad.py
+++ b/selfdrive/athena/athenad.py
@@ -214,7 +214,7 @@ def upload_handler(end_event: threading.Event) -> None:
         cloudlog.event("athena.upload_handler.upload_start", fn=fn, sz=sz, network_type=network_type, metered=metered, retry_count=cur_upload_items[tid].retry_count)
         response = _do_upload(cur_upload_items[tid], cb)
 
-        if response.status_code not in (200, 201, 401, 403, 412):
+        if response.status_code not in {200, 201, 401, 403, 412}:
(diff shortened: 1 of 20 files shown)


GregorKikelj:
O(1) is very cool but the quick test for small lists, sets and tuples doesn't seem to suggest that it's worth it. 

```
import timeit

def f1(sample):
    cnt=0
    for x in map(str, range(1000)):
        if x in sample:
            cnt+=1
    return cnt

def list_test():
    return f1(["10"])
def dict_test():
    return f1({"10"})
def tuple_test():
    return f1(("10",))

def large_list_test():
    return f1(["1", "5", "75", "90"])
def large_dict_test():
    return f1({"1", "5", "75", "90"})
def large_tuple_test():
    return f1(("1", "5", "75", "90"))


print(timeit.timeit(list_test, number=10000))
print(timeit.timeit(dict_test, number=10000))
print(timeit.timeit(tuple_test, number=10000))

print()

print(timeit.timeit(large_list_test, number=10000))
print(timeit.timeit(large_dict_test, number=10000))
print(timeit.timeit(large_tuple_test, number=10000))
```

Running with python 3.8.10 gives me
```
1.331106665020343
1.4716969270375557
1.344645882025361

1.5291405050083995
1.517012978962157
1.5398731180466712
```

Should be worth at larger sizes but doesn't seem to help with lookups on small structures.

PR #25668, opened 2022-09-05
ui: multilang training guide

deanlee:
WIP.  

- [x] Draw text, button in paintEvent
- [ ] Make the text area scrollable when necessary

@adeebshihadeh  If you think this PR is ok, I need your help to remove all the text from asserts/training_wide/images.

in addition, I think maybe we can just keep training_wide, remove the boundingRectStandard and asserts/ttraining.what do you think?


adeebshihadeh:
I'm not sure this makes sense with this iteration of the training guide. I think we should focus on making the rest of openpilot that you use every day work with multilang first, such as the onroad alertrs.

adeebshihadeh:
Gonna close until we decide what to do with the training guide. Internally, we had hoped to have new and improved designs by now and didn't want to put effort into this iteration of the training guide.

PR #25672, opened 2022-09-05
hyundai: consistent start from stop

gregjhogan:
there is still a lot of room for improvement (can probably be made twice as fast) but as a starting point this makes it consistent

further improvements:
* make vEgoStarting 0
* request more accel with higher upper jerk limit when planner wants higher jerk

here is an example running this branch after sitting stopped for a long time (the longer you sit stopped the worse the problem is)
start-from-stop-1-zoomed-out
start-from-stop-1-zoomed-in


gregjhogan:
what is on master now seems pretty good now, so at this point I don't think these changes are necessary

PR #25675, opened 2022-09-05
translations tests: ignore line numbers

sshane:
The only difference should be in QT Linguist, you won't be able to see the code context anymore, which was actually pretty helpful.

We should be able to strip the line number tags when checking if files up to date, so I'll try that first before merging

Will also try relative


sshane:
Chose to go with relative locations + ignoring line numbers. Just using relative locations still means you need to update translations, albeit with a likely reduced diff, which kind of defeats the purpose of making this no-fuss to maintain. When reading a ts file, all location lines are omitted, so that they do not influence any of the tests (not important for translating UI, only for the QT Linguist tool)

PR #25677, opened 2022-09-06
Toyota: forward openpilot FCW to cluster when engaged

cydia2020:
Forward fcw alert to the vehicle's dash (makes the vehicle display the blinking yellow fcw icon) if openpilot fcw is triggered while cruise control is engaged and openpilot is controlling longitudinal on a Toyota/Lexus vehicle.

*Note:*
This alert only gets displayed by the dash if ACC is engaged, this alert was originally used for alerting the user that the car might not have enough braking distance while the stock ACC is engaged. 
image

Extensively tested in my own fork, this alert is merely a UI message and does not have priority over the stock system's fcw/aeb in the cluster's logic, so it won't mask any of those signals when it's triggered and it should be safe to use.

Image of the alert: https://github.com/commaai/opendbc/pull/457#issuecomment-951369032


sshane:
How does this compare/impact the current FCW alert we send? Is the only purpose is so we can show on Prius V?

cydia2020:
> How does this compare/impact the current FCW alert we send? Is the only purpose is so we can show on Prius V?

It does not affect the current fcw warning. This new signal is only sent when `enableDsu` if false, and the current FCW signal is only sent when `enableDsu` is true. Without this change, openpilot does not send any fcw alert to the car if the DSU is not unplugged; with this change applied, it sends the cut-in fcw signal (this signal is on 0x343 `ACC_CONTROL` which we are already sending anyway so there won't be any controls conflict) when DSU is still plugged in and openpilot is controlling long (TSS 2.0 or smartDsu).

```python
fcw_alert_acc = (hud_control.visualAlert == VisualAlert.fcw) and not self.CP.enableDsu
```

We can use this to display openpilot fcw on the Prius V, however the vehicle doesn't respond to this signal if ACC is not engaged. Thus the title "... when engaged"

selfdrive/car/toyota/toyotacan.py
```diff
@@ -38,6 +38,7 @@ def create_accel_command(packer, accel, pcm_cancel, standstill_req, lead, acc_ty
     "RELEASE_STANDSTILL": not standstill_req,
     "CANCEL_REQ": pcm_cancel,
     "ALLOW_LONG_PRESS": 1,
+    "ACC_CUT_IN": fcw_alert,
```
adeebshihadeh, line 41: should we rename the signal?

cydia2020: It's used as cut-in warning in the stock system, it's similar to fcw on both the dashboard and HUD, only difference is it's yellow instead of red.

sshane:
I think we should block and forward the ACC_HUD message so we can send the correct FCW message

cydia2020:
> I think we should block and forward the ACC_HUD message so we can send the correct FCW message

That'll unfortunately only work on TSS2.0 camera ACC cars I think. On TSS-P cars I believe this message is sent by the DSU.

sshane:
Where we can do it, I think we should. I'd merge that before this, since the alert is still a bit ambiguous here

cydia2020:
> Where we can do it, I think we should. Either SDSU or TSS2 + not radar ACC (does SDSU block ACC_HUD though?). I'd merge that before this, since the alert is still a bit ambiguous here. Is there anything else in ACC_CONTROL that might show a different alert?

Not that I am aware of, and SDSU only blocks 0x343 (ACC_CONTROL), so this will be limited to camera ACC cars.

selfdrive/car/toyota/toyotacan.py
sshane: `CUT_IN_WARNING`?

sshane:
I tested the alert yesterday and I do like it (beeps and shows lead with orange), but I don't like how it asymmetrically doesn't show when you're not engaged. @adeebshihadeh what do you think?

adeebshihadeh:
lgtm

selfdrive/car/toyota/toyotacan.py
cydia2020: image

In TS it's being named this

PR #25680, opened 2022-09-06
AGNOS setup: multi-lang support

deanlee:


selfdrive/ui/qt/setup/setup.cc
```diff
@@ -363,6 +364,32 @@ void Setup::nextPage() {
...
+  if (std::exchange(initialized, true) == true) return;
+
+  // select language
+  QMap<QString, QString> langs = {
+      {"English", "main_en"},
+      {"Português", "main_pt-BR"},
+      {"中文（繁體）", "main_zh-CHT"},
+      {"中文（简体）", "main_zh-CHS"},
+      {"한국어", "main_ko"},
+      {"日本語", "main_ja"}};
```
sshane, line 380: Can we load from the json file?

deanlee:
the ScrollView has a white border:
dd

deanlee:
tr() doesn't work properly in some places, especially in lambdas
error

deanlee:
found the problem ,The QTranslator needs to exist during the full life time of the QApplication.

selfdrive/ui/qt/setup/setup.cc
```diff
@@ -363,6 +364,25 @@ void Setup::nextPage() {
...
 
 Setup::Setup(QWidget *parent) : QStackedWidget(parent) {
+}
+
+void Setup::showEvent(QShowEvent *event) {
+  static bool initialized = false;
+  if (std::exchange(initialized, true) == true) return;
+
+  // select language
+  QString currentLang = "main_en";
```
sshane, line 374: I think I like no language selected better. If you want English, you have to press Cancel to proceed like this

deanlee: yes i didn't notice the problem

selfdrive/assets/assets.qrc
```diff
@@ -13,5 +13,12 @@
   <file>offroad/icon_wifi_strength_medium.svg</file>
   <file>offroad/icon_wifi_strength_high.svg</file>
   <file>offroad/icon_wifi_strength_full.svg</file>
+  <file alias="languages.json">../ui/translations/languages.json</file>
+  <file alias="main_en">../ui/translations/main_en.qm</file>
+  <file alias="main_pt-BR">../ui/translations/main_pt-BR.qm</file>
+  <file alias="main_zh-CHS">../ui/translations/main_zh-CHS.qm</file>
+  <file alias="main_zh-CHT">../ui/translations/main_zh-CHT.qm</file>
+  <file alias="main_ko">../ui/translations/main_ko.qm</file>
+  <file alias="main_ja">../ui/translations/main_ja.qm</file>
```
sshane, line 22: No way of doing this easily and programmatically? Right now we have everything nicely in one file for languages to show. Any advantages to adding them?

deanlee: Just want to embed the resource into the setup, so that the setup can be released  standalone without the language files.
I'm not sure how to improve it, feel free to modify it in a better way.

adeebshihadeh:
Since we don't make any guarantees about the translation quality, I don't want to actually ship this yet, but we can merge this if you put it behind a flag for now.

adeebshihadeh (changes requested):
Since we don't make any guarantees about the translation quality, I don't want to actually ship this yet, but we can merge this if you put it behind a flag for now.

PR #25688, opened 2022-09-07
single longitudinal toggle for e2e and disable radar

adeebshihadeh:


redacid95:
If I understand this pr correctly, this makes VOACC require e2e long?

adeebshihadeh:
> If I understand this pr correctly, this makes VOACC require e2e long?

yeah

redacid95:
> > If I understand this pr correctly, this makes VOACC require e2e long?
> 
> yeah

I don't know if this will be well received. E2e long needs a lot of work still and I personally use VOACC only and plenty of others do too. That's my 2 cents and I will say no more on the matter if you all still want to proceed with it. I just think that they should remain as 2 different toggles. Maybe force OP Long with the e2e toggle, but keep them different.

adeebshihadeh:
We’re not interested in maintaining two different experimental modes. E2E long should improve pretty quickly now that it’s a toggle.

adeebshihadeh:
@redacid95 we decided to bring back the experimental long toggle #25698

PR #25168, opened 2022-07-14
VW MQB: Cleanup stock ACC button handling

jyoung8607:
Remove my ugly cancel/resume spam implementation and replace in the style of @robbederks's nice Tesla implementation.

Passes tests but haven't driven yet.

PR #25692, opened 2022-09-07
VW: Increase stock ACC button spam reliability

jyoung8607:
The cleanup in #25168 resulted in a regression of cancel/resume spam reliability. Restore some of the timing control (sending just after the car sends) from the prior version, maintaining overall clarity.

LGTM with limited testing, but holding in draft for a few days to get additional test users and testing time.


jyoung8607:
I think this is good. Two more test users also report good results. Releasing from draft.

`3cfdec54aa035f3f|2022-09-07--15-09-49`
`bbc2fa573509b5ff|2022-09-08--17-36-21`

PR #25693, opened 2022-09-07
Ioniq 5: adjust max lateral acceleration

ioniq5-cz:
**Description** 

Currently all IONIQ 5 users need to manually adjust each time they switch a branch these values in order to avoid the not ideal values taken from EV6. There is a lot of pingpong with EV6 values and with this fix the car drives much smoother.

**Verification** 

At least 3 IONIQ 5 users are using these settings since the car is supported. Probably there are better values but @sunnyhaibin proposed those after looking at plotjugger.

**Route**
Route: 2833de0636d26d55|2022-09-07--20-43-50--0


sshane:
@nuwandavek do we have enough data from Ioniq 5's to set this from the data?

ioniq5-cz:
resolved

PR #25695, opened 2022-09-07
pre-commit: test translations

sshane:


.github/workflows/selfdrive_tests.yaml
```diff
@@ -62,7 +62,7 @@ jobs:
         cp .pylintrc $STRIPPED_DIR
         cp mypy.ini $STRIPPED_DIR
         cd $STRIPPED_DIR
-        ${{ env.RUN }} "pre-commit run --all"
+        ${{ env.RUN }} "SKIP=test_translations pre-commit run --all"
```
adeebshihadeh, line 65: why is this skipped?

sshane: Because some ui files are not included in release like setup, and so the translations are expected to be different then.

PR #25699, opened 2022-09-08
FPv2: support collecting versions for specific ecus

sshane:


selfdrive/car/chrysler/values.py
```diff
@@ -129,12 +129,13 @@ class ChryslerCarInfo(CarInfo):
   }],
 }
 
-FW_VERSIONS = {
```
sshane, line 132: If we just ignore ecus with empty version lists, this can cause other cars to mis-fingerprint as we never invalidate this platform since we just `continue` for all two of its ecus.

selfdrive/car/fw_versions.py
```diff
@@ -443,6 +443,11 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, debug=Fa
 
 def get_fw_versions(logcan, sendcan, query_brand=None, extra=None, timeout=0.1, debug=False, progress=False):
   versions = get_interface_attr('FW_VERSIONS', ignore_none=True)
+
+  # Each brand can define extra ECUs to query for data collection
+  for brand, ecus in get_interface_attr('EXTRA_ECUS', ignore_none=True).items():
+    versions[brand]["debug"] = {ecu: [] for ecu in ecus}
+
```
sshane, line 450: Just queried, never considered for fingerprinting

sshane, line 450: Just queried, never considered for fingerprinting

selfdrive/car/chrysler/values.py
```diff
@@ -129,12 +129,13 @@ class ChryslerCarInfo(CarInfo):
...
-  CAR.PACIFICA_2019_HYBRID: {
-    (Ecu.hcp, 0x7e2, None): [],
-    (Ecu.abs, 0x7e4, None): [],
-  },
+# Ecus added for data collection
+EXTRA_ECUS = {
+  (Ecu.hcp, 0x7e2, None),  # manages transmission on hybrids
+  (Ecu.abs, 0x7e4, None),  # alt address for abs on hybrids
+}
```
adeebshihadeh, line 136: let's formalize this a bit more into a generic FPv2 config

adeebshihadeh (changes requested):

selfdrive/car/chrysler/values.py
sshane: How does this look now?

adeebshihadeh:
Let's do the config refactor first in another PR. All the [car-specific stuff](https://github.com/commaai/openpilot/blob/d9ca45ed1b2bc0316aad8ff11470e1a9b96fb09c/selfdrive/car/fw_versions.py#L48) should be moved into its values.py.

sshane:
Sounds good! I was thinking we can do this incrementally though so we can get the Hyundai query in, but you want everything moved over first before this?

adeebshihadeh:
Yes, let's do the refactor first.

sshane:
superceeded by https://github.com/commaai/openpilot/pull/25859

selfdrive/car/fw_versions.py
```diff
@@ -213,6 +213,11 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, debug=Fa
 
 def get_fw_versions(logcan, sendcan, query_brand=None, extra=None, timeout=0.1, debug=False, progress=False):
   versions = VERSIONS.copy()
+
+  # Each brand can define extra ECUs to query for data collection
+  for brand, config in FW_QUERY_CONFIGS.items():
+    versions[brand]["debug"] = {ecu: [] for ecu in config.extra_ecus}
+
```
sshane, line 220: These ECUs will be queried and added to carFw, but since they're not in any specific car, the fingerprinting code will totally ignore them

PR #25702, opened 2022-09-08
ui: add hysteresis to cluster ego speed

sshane:
Seems like Toyota does this (GM might), will check other makes


selfdrive/car/toyota/carstate.py
```diff
@@ -52,6 +53,7 @@ def update(self, cp, cp_cam):
     )
     ret.vEgoRaw = mean([ret.wheelSpeeds.fl, ret.wheelSpeeds.fr, ret.wheelSpeeds.rl, ret.wheelSpeeds.rr])
     ret.vEgo, ret.aEgo = self.update_speed_kf(ret.vEgoRaw)
+    ret.vEgoCluster = self.cluster_speed_steady = apply_hysteresis(ret.vEgo, self.cluster_speed_steady, 0.5)
```
adeebshihadeh, line 56: don't want to do it or all cars? civic dash seems to do this as well

PR #25708, opened 2022-09-08
ui: fix networking crash

sshane:
This code introduced in https://github.com/commaai/openpilot/pull/25698 assumes all items of a `ListWidget` will be only widgets, however in two places (networking and power layout) we add layouts. So when getting the widgets from the inner layout items (`->widget()`) we get null `QWidget` objects causing segfaults when trying to call anything on them.


sshane:
@adeebshihadeh `if (widget->isVisible()) {` always seems to be true for me, even if the widget is pushed down out of view (except for one toggle second to last). Was this to prevent drawing on other settings views? Because I also don't see them calling any paintEvents. Want to just remove?

adeebshihadeh:
Fixed in 613686855779187db9b20ecee34a4f8b1dd5ce67. This is for hidden widgets, not ones out of view.

PR #25713, opened 2022-09-09
Ford: improved steering control

incognitojam:
- refactor rate limits as "angle per second" limits
  - calculate rate limit from command frequency (ford only sends at 20Hz! the limits were 5x too strict before)


incognitojam:
Sometimes the EPS just stops responding to the commands, but it doesn't appear to be faulted - if the driver steers the wheel manually to the desired angle it will recover... I will try limiting the difference between the reported steering angle and the angle we command.
image

PR #25722, opened 2022-09-10
tuning: measure steering accuracy for existing routes

incognitojam:
- adds support for loading existing route

EDIT: I planned to add support for loading logs from a file or URL but this already makes it more usable


jyoung8607:
Since you've got your head in this tool: i always wanted to split out left and right vs combining/averaging them, to look for effects of road crown on self-centering or straight line driving corrections built into the EPS.

incognitojam:
@jyoung8607 I've not had time to come back to this so will have to leave it for another day :smiling_face_with_tear:

PR #25724, opened 2022-09-11
good updater experience

adeebshihadeh:
- [ ] <s>works for existing gitconfig's configured to only pull the originally installed branch</s> will be a follow up PR
- [x] good onroad experience
- [x] branch dropdown
- [x] only pull target branch


Moodkiller:
@adeebshihadeh could AGNOS / OS Version be added back into the Software section? I think a dedicated label below `Current Version` would look the best to accommodate long version and/or branch names?

PR #25725, opened 2022-09-11
Subaru: Add missing 2020 Ascent f/w

bbennett80:
2020 Subaru Ascent required recall service for ecu reflash. New values added.

Dongle ID: `be44434f27731ff8`. Route: `be44434f27731ff8|2022-09-10--18-44-47`.

### DIFF
--- a/selfdrive/car/subaru/values.py
+++ b/selfdrive/car/subaru/values.py
@@ -104,6 +104,7 @@ FW_VERSIONS = {
       b'\000\000e~\037@ \'',
       b'\x00\x00e@\x1f@ $',
       b'\x00\x00d\xb9\x00\x00\x00\x00',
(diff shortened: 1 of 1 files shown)


selfdrive/car/subaru/values.py
```diff
@@ -104,18 +104,21 @@ class SubaruCarInfo(CarInfo):
       b'\000\000e~\037@ \'',
       b'\x00\x00e@\x1f@ $',
       b'\x00\x00d\xb9\x00\x00\x00\x00',
+      b'\x00\x00e@\x00\x00\x00\x00'
```
sshane, line 107:
```suggestion
      b'\x00\x00e@\x00\x00\x00\x00',
```

PR #25728, opened 2022-09-11
camerad:  zero memory before returning from MemoryManager::alloc

deanlee:
non-zeroed memory may result in unexpected behavior in the `camerad`.

### DIFF
--- a/system/camerad/cameras/camera_util.cc
+++ b/system/camerad/cameras/camera_util.cc
@@ -1,5 +1,7 @@
 #include "system/camerad/cameras/camera_util.h"
 
+#include <string.h>
+
 #include <cassert>
 
 #include <sys/ioctl.h>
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
I'm not sure we want this. Perhaps we simply rename alloc -> malloc to make this more clear?

deanlee:
I feel it is safer to  zero memory before return.

for example:
```
struct cam_packet *pkt = (struct cam_packet *)mm.alloc(size, &cam_packet_handle);
  pkt->num_cmd_buf = 0;
  pkt->kmd_cmd_buf_index = -1;
  pkt->header.size = size;
  pkt->header.op_code = CAM_SENSOR_PACKET_OPCODE_SENSOR_NOP;
  pkt->header.request_id = request_id;
```
We have only initialized some of the fields in the structure, other uninitialized fields may contain some "dirty" content .This could potentially cause problems.

ZwX1616:
Cleaning things up. Can reopen if we ever see problems caused by this

PR #25732, opened 2022-09-12
camerad: free memory on failure 

deanlee:


adeebshihadeh:
Perhaps we can refactor to automatically free when it goes out of scope?

deanlee:
created a new pr https://github.com/commaai/openpilot/pull/28807 to refactor to auto-free

PR #25737, opened 2022-09-12
camerad: cleanup CameraBuf::acquire

deanlee:


system/camerad/cameras/camera_common.cc
```diff
@@ -116,41 +116,31 @@ bool CameraBuf::acquire() {
 
   if (camera_bufs_metadata[cur_buf_idx].frame_id == -1) {
     LOGE("no frame data? wtf");
-    release();
```
adeebshihadeh, line 119: wasn't sure why you removed the whole block, so added it back. feel free to open another PR and explain why you removed it

PR #25738, opened 2022-09-12
camerad: dont queue buffer if frame is dropped

deanlee:


adeebshihadeh:
Is this still relevant?

deanlee:
Yes, I think it's better to do a check on error before queue the buffer.

adeebshihadeh:
Is it clear that the dropped frame is still queued up? Did you find some issue on device or is this a hypothetical?

deanlee:
not very sure now. closed

PR #25741, opened 2022-09-12
ui: multilang date

deanlee:
locale


selfdrive/ui/qt/home.cc
```diff
@@ -174,6 +174,7 @@ OffroadHome::OffroadHome(QWidget* parent) : QFrame(parent) {
 }
 
 void OffroadHome::showEvent(QShowEvent *event) {
+  locale_name = QString::fromStdString(Params().get("LanguageSetting")).replace("main_", "");
```
sshane, line 177: Can we put this in the UI state so we just need to read the param once on startup?

sshane:
fyi no need to update translations for changing line numbers any more, but you're free to keep doing it if you want. thanks for the PR!

PR #25742, opened 2022-09-12
ui: always show SetupWidget

deanlee:
If there is no network, the SetupWidget will not be displayed, resulting in a blank space on the right side, which looks very weird.

wrong


selfdrive/ui/qt/widgets/prime.cc
```diff
@@ -299,11 +299,10 @@ SetupWidget::SetupWidget(QWidget* parent) : QFrame(parent) {
 
     QObject::connect(repeater, &RequestRepeater::requestDone, this, &SetupWidget::replyFinished);
   }
-  hide(); // Only show when first request comes back
+  hide();
```
sshane, line 302: Think you meant to delete this line?

deanlee: yes....

PR #25743, opened 2022-09-12
ui: multilang offroad alerts

deanlee:


selfdrive/controls/lib/alerts_offroad.json
```diff
@@ -1,52 +1,40 @@
...
   },
   "Offroad_ConnectivityNeededPrompt": {
-    "text": "Immediately connect to the internet to check for updates. If you do not connect to the internet, openpilot won't engage in ",
     "severity": 0,
     "_comment": "Append the number of days at the end of the text"
   },
   "Offroad_ConnectivityNeeded": {
-    "text": "Connect to internet to check for updates. openpilot won't automatically start until it connects to internet to check for updates.",
     "severity": 1
   },
```
sshane, line 11: I wonder, can we get along without this file now?

deanlee: I agree with you. this file should be mainly used in deprecated openpilot-apks to config & load offroad text. I think it's time to delete it.

adeebshihadeh: hmm, maybe put it in its own header file then? think that's cleaner than directly in the cc file

selfdrive/ui/tests/cycle_offroad_alerts.py
```diff
@@ -11,8 +10,20 @@
...
+    'Offroad_InvalidTime',
+    'Offroad_UnofficialHardware',
+    'Offroad_StorageMissing',
+    'Offroad_BadNvme',
+    'Offroad_CarUnrecognized',
+    'Offroad_ConnectivityNeededPrompt',
+    'Offroad_NoFirmware',
+    'Offroad_IsTakingSnapshot',
+    'Offroad_NeosUpdate',
+  )
```
sshane, line 26: Is there any way to list params and then just take the ones with this prefix?

deanlee: There is no way to get a list of all keys. how about keep it as it is? we can write test in C++ or add a method to return all keys in Params late if necessary

deanlee: add a new method get_all_keys in https://github.com/commaai/openpilot/pull/25779

deanlee:
Cannot put the definition of `offroad_alerts` out of  class. otherwise the string will be initialized after the program starts and before `installTranslator` is called.

sshane:
Is this good for review?

PR #25744, opened 2022-09-12
replay/consoleui: update at 5 HZ

deanlee:


adeebshihadeh:
What's the purpose of this?

PR #25761, opened 2022-09-13
VW: match panda standstill check

sshane:


sshane:
@jyoung8607 do you have any desire of taking this over and finishing it? I think it just needs some verification that the thresholds are correct in openpilot and panda. Ideally we can just check if vEgoRaw == 0, but I'll have to check data to see if VW has any noise wheel speed sensor issues

jyoung8607:
Sure, I've added it to [my list](https://github.com/users/jyoung8607/projects/1/views/1). The thresholds should match *and* the wheels we check should match; right now we're only checking the front two wheels in Panda.

I don't think our wheel speed sensor outputs are noisy. The reason I set standstill at 0.1m/s is that we start seeing EPS faults if we try to steer much closer to zero. If there are other desirable reasons for standstill to reflect exactly vEgoRaw = 0, I could add a separate condition to prevent steer below 0.1m/s, but I don't know of a reason at this time.

sshane:
In the lat controller classes, we have a variable that stops steer at 0.3 m/s, I've been meaning to make that common. It doesn't make sense for the lat controller to output 0 torque and stop steering, but keeping that apply bit high (with latActive) for a little while longer.

jyoung8607:
Your question is timely in light of #26694. I've been pondering the most correct fix. In light of the MIN_STEER_SPEED in latcontrol, I thought it was kind of pointless to track the involved firmware and that I should just go ahead and raise all VW minSteerSpeed to 0.3 m/s. But, I wasn't in love with the idea of VW having unique lateral shutoff handling.

Your proposal to do this globally makes intuitive sense to me. We'll have to go around and suppress unnecessary low speed steering alerts in all car ports, make sure we round down to zero for docs generation, make sure tests don't explode, etc. I'm happy to do it, but I may not be able to get to it until this weekend.

Assuming that change is accepted, I can't immediately think of any reason not to set the VW standstill speed to zero.

sshane:
I had in mind something even easier. Instead of changing the minSteerSpeed on the interface level, change it on the controls level (both are global). That way docs and alerts aren't affected

jyoung8607:
Jinx :) was in the middle of an edit. Agreed that should probably do the job.

sshane:
Merged https://github.com/commaai/openpilot/pull/26805 so this PR should be good to go! @jyoung8607

PR #25772, opened 2022-09-14
Honda: match dash speed

sshane:
Seems like the cluster applies some hysteresis: https://imgur.com/a/ckLBPDz This PR reduces the error a bit and removes steady state mismatching


sshane:
It also looks like the dash applies some delay or filtering. This is with an rc of 0.025 and dt of 1/100 (0.01s). See middle plot vs one above with no filtering, both have hysteresis and are rounded:

Screenshot from 2022-09-14 12-53-45

PR #25773, opened 2022-09-14
inc sensord cpu usage in onroad test

gast04:
Changing sensord to interrupts increased its cpu usage.
Measuring with `top` shows the following increases:
```
lsm, no bmx:
~9% ->  ~(12-13)%

lsm c variant, no bmx:
~12% ->  ~17-18%

lsm, with bmx
~17% -> 21%
```
```
using the test_onroad.py script:
 6.22 ->  9.43%
other values on the CI device have been: 11.2%, 11.9%, 14.13%

so the 12% has been chosen
```

### DIFF
--- a/selfdrive/test/test_onroad.py
+++ b/selfdrive/test/test_onroad.py
@@ -29,7 +29,7 @@ PROCS = {
   "selfdrive.controls.plannerd": 11.7,
   "./_ui": 19.2,
   "selfdrive.locationd.paramsd": 9.0,
-  "./_sensord": 6.17,
+  "./_sensord": 12.0,
   "selfdrive.controls.radard": 4.5,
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Hmm, makes sense that CPU went up, but why does the C variant use more CPU? Perhaps it just happens to be running on different cores (and thus a different cluster).

adeebshihadeh:
Merging so we don't have CI randomly failing, but I'd still like to understand the diff between the different configurations.

gast04:
will retest it, in my todo list

gast04:
retested results (pinned sensord to core 2)
on bmx device:   9.40% -> 14.34%
non bmx device: 6.22% -> 10.62%

PR #25775, opened 2022-09-14
Sensord skip init values, first 500ms

### LINKED ISSUE #22888: sensord: Gyro values are unreasonable sometimes during init
### Describe the bug

When initialized, sometimes the gyro values are unreasonable. We could potentially filter these out in sensord.

image

comma three

fbbb958f3157f063|2021-11-12--11-15-20/0

### openpilot version

c7be73b826b21947197f24c6a60fc5a7098e42af

### Additional info

_No response_

gast04:
resolve https://github.com/commaai/openpilot/issues/22888
by filtering the initial measurements

### DIFF
--- a/selfdrive/sensord/sensors_qcom2.cc
+++ b/selfdrive/sensord/sensors_qcom2.cc
@@ -28,6 +28,10 @@
 ExitHandler do_exit;
 std::mutex pm_mutex;
 
+// filter first values (0.5sec) as those may contain inaccuracies
+uint64_t init_ts = 0;
+constexpr uint64_t init_delay = 500*1e6;
+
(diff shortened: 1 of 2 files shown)


adeebshihadeh (commented):
Did you check if there's some kind of validity flag from the sensor? Does the datasheet explain what's happening here? Does 500ms handle all known cases? Which sensors does this happen for?

gast04:
I thought we agreed on making a generic solution and skip the first values in a 500ms frame?
the diagrams from @nuwandavek shows mainly the lsm, the datasheet doesnt say anything about it,
no validity flag, he also send me one plot from the light sensor, same issue. the bmx states that
its gyro has a 30ms startup time and the bmx magn a 3ms.
need to check the datasheet of the light and magnetometer.

PR #25776, opened 2022-09-14
Sensor tests speedup

gast04:
speed up sensor test case, from ~23s to ~11s

### DIFF
--- a/selfdrive/sensord/tests/test_sensord.py
+++ b/selfdrive/sensord/tests/test_sensord.py
@@ -118,15 +118,14 @@ class TestSensord(unittest.TestCase):
 
     # read initial sensor values every test case can use
     managed_processes["sensord"].start()
-    cls.events = read_sensor_events(3)
+    cls.events = read_sensor_events(5)
     managed_processes["sensord"].stop()
 
   def test_sensors_present(self):
     # verify correct sensors configuration
-    events = self.events
 
     seen = set()
-    for event in events:
+    for event in self.events:
       for measurement in event.sensorEvents:
         # filter unset events (bmx magn)
         if measurement.version == 0:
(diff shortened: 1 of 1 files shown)


selfdrive/sensord/tests/test_sensord.py
```diff
@@ -116,10 +116,14 @@ def setUpClass(cls):
...
-  @with_processes(['sensord'])
+    # read initial sensor values every test case can use
+    managed_processes["sensord"].start()
+    cls.events = read_sensor_events(3)
+    managed_processes["sensord"].stop()
+
   def test_sensors_present(self):
     # verify correct sensors configuration
-    events = read_sensor_events(10)
+    events = self.events
```
adeebshihadeh, line 126: replace all the local events with self.events

selfdrive/sensord/tests/test_sensord.py
```diff
@@ -116,10 +116,14 @@ def setUpClass(cls):
     # make sure gpiochip0 is readable
     HARDWARE.initialize_hardware()
 
-  @with_processes(['sensord'])
+    # read initial sensor values every test case can use
+    managed_processes["sensord"].start()
+    cls.events = read_sensor_events(3)
```
adeebshihadeh, line 121:
```suggestion
    cls.events = read_sensor_events(5)
```

adeebshihadeh (commented):
Why not use for `test_sensor_values_sanity_check`?

selfdrive/sensord/tests/test_sensord.py
gast04: gonna need to do a quick test run with that change, check if it doesnt screw with the values

gast04:
> 

cause if we increase it the miniamal sample check wouldnt make sense anymore, as we would receive more than
twice as much, so I skipped that one for now

PR #25778, opened 2022-09-14
Hyundai: match dash speed

sshane:


sshane:
Hyundai also seems to add hysteresis:

Screenshot from 2022-09-13 22-25-24

I noticed this also with the Civic, but the UI speed seems to lag behind the wheel speeds as well as having hysteresis, which would further explain the discrepancies when accelerating or braking, and also when taking off from a stop (dash lags and says 0 mph while wheel speeds are already 1-2 mph). You can see here how it looks like the dash speed is off by a considerable amount while accelerating, but in steady state matches quite well (Sonata):

Screenshot from 2022-09-13 22-25-11

sshane:
Seems like Hyundai also applies a scaling factor

Issue #22661, opened 2021-10-22
Tesla: implement FW query


adeebshihadeh:
Examples PRs for other brands: https://github.com/commaai/openpilot/pull/1540 https://github.com/commaai/openpilot/pull/20271

sshane:
This should be much easier to merge with https://github.com/commaai/openpilot/pull/25042 in master, @robbederks want to take a stab soon?

robbederks:
Sure, will try it out. Shall I try over the comma power, or do you prefer it on bus 0? Think both might work

PR #25785, opened 2022-09-14
Tesla FW query

robbederks:
To implement #22661


selfdrive/car/tesla/values.py
```diff
@@ -37,6 +35,40 @@ class CAR:
...
+      bus=0,
+    ),
+    Request(
+      [StdQueries.TESTER_PRESENT_REQUEST, StdQueries.UDS_VERSION_REQUEST],
+      [StdQueries.TESTER_PRESENT_RESPONSE, StdQueries.UDS_VERSION_RESPONSE],
+      rx_offset=0x10,
+      bus=0,
+    ),
+  ]
+)
```
sshane, line 53: If you know which ecus respond to which request, let's whitelist them (to avoid waiting to timeout for them, or duplicate logged versions)

robbederks:
@kennedyshead would you be able to try out this query with your AP1 car as well so we can potentially also get rid of the CAN fingerprint for your car?

selfdrive/car/tesla/values.py
kennedyshead: On it

adeebshihadeh:
Let's just merge this for your car, then we can add the AP1 Model S fingerprint in another PR.

Also why not remove the CAN fingerprints? It's fine to break dashcam only cars from fingerprinting.

robbederks:
> Also why not remove the CAN fingerprints? It's fine to break dashcam only cars from fingerprinting.

Ok, was thinking of leaving it in for a bit until we collect some more firmware versions (I'm not sure how many cars are using this port), but we can remove and add ones when they pop up. Will finish this PR in the next few days

robbederks:
Looks like I'm the only one on AP2 uploading routes to our servers, so we can definitely remove the CAN fingerprint!

PR #25787, opened 2022-09-14
sensord: move sensors in lowest power mode on exit/kill

gast04:
Move all sensors in its lowest possible power state on sensord exit/kill.
Its hard to measure a difference but in average a draw about 2mA less an consume about 100mW less.

### DIFF
--- a/selfdrive/sensord/sensors/bmx055_accel.cc
+++ b/selfdrive/sensord/sensors/bmx055_accel.cc
@@ -27,6 +27,8 @@ int BMX055_Accel::init() {
   if (ret < 0) {
     goto fail;
   }
+  // bmx055 accel has a 1.3ms wakeup time from deep suspend mode
+  util::sleep_for(10);
 
   // High bandwidth
   // ret = set_register(BMX055_ACCEL_I2C_REG_HBW, BMX055_ACCEL_HBW_ENABLE);
--- a/selfdrive/sensord/sensors/bmx055_accel.cc
+++ b/selfdrive/sensord/sensors/bmx055_accel.cc
@@ -53,9 +53,7 @@ fail:
 
 int BMX055_Accel::shutdown()  {
   // enter deep suspend mode (lowest power mode)
-
-  int ret = 0;
-  ret = set_register(BMX055_ACCEL_I2C_REG_PMU, BMX055_ACCEL_DEEP_SUSPEND);
+  int ret = set_register(BMX055_ACCEL_I2C_REG_PMU, BMX055_ACCEL_DEEP_SUSPEND);
   if (ret < 0) {
     LOGE("Could not move BMX055 ACCEL in deep suspend mode!")
   }
--- a/selfdrive/sensord/sensors/bmx055_accel.cc
+++ b/selfdrive/sensord/sensors/bmx055_accel.cc
@@ -23,6 +23,11 @@ int BMX055_Accel::init() {
     goto fail;
   }
 
+  ret = set_register(BMX055_ACCEL_I2C_REG_PMU, BMX055_ACCEL_NORMAL_MODE);
+  if (ret < 0) {
+    goto fail;
+  }
+
   // High bandwidth
   // ret = set_register(BMX055_ACCEL_I2C_REG_HBW, BMX055_ACCEL_HBW_ENABLE);
   // if (ret < 0) {
@@ -40,6 +45,18 @@ int BMX055_Accel::init() {
   }
 
 fail:
+  return ret;
+}
+
+int BMX055_Accel::shutdown()  {
+  // enter deep suspend mode (lowest power mode)
+
+  int ret = 0;
+  ret = set_register(BMX055_ACCEL_I2C_REG_PMU, BMX055_ACCEL_DEEP_SUSPEND);
+  if (ret < 0) {
+    LOGE("Could not move BMX055 ACCEL in deep suspend mode!")
+  }
+
   return ret;
 }
 
--- a/selfdrive/sensord/sensors/bmx055_accel.cc
+++ b/selfdrive/sensord/sensors/bmx055_accel.cc
@@ -4,6 +4,7 @@
 
 #include "common/swaglog.h"
 #include "common/timing.h"
+#include "common/util.h"
 
 BMX055_Accel::BMX055_Accel(I2CBus *bus) : I2CSensor(bus) {}
 
--- a/selfdrive/sensord/sensors/bmx055_gyro.cc
+++ b/selfdrive/sensord/sensors/bmx055_gyro.cc
@@ -64,9 +64,7 @@ fail:
 
 int BMX055_Gyro::shutdown()  {
   // enter deep suspend mode (lowest power mode)
-
-  int ret = 0;
-  ret = set_register(BMX055_GYRO_I2C_REG_LPM1, BMX055_GYRO_DEEP_SUSPEND);
+  int ret = set_register(BMX055_GYRO_I2C_REG_LPM1, BMX055_GYRO_DEEP_SUSPEND);
   if (ret < 0) {
     LOGE("Could not move BMX055 GYRO in deep suspend mode!")
   }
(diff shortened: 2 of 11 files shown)


robbederks:
It should also possible to just switch off the regulators for the sensors, but that would probably require a kernel change.

gast04:
hmm, I can check, but not sure if we wanna go down that route

adeebshihadeh:
Nah, this should be sufficient for now.

selfdrive/sensord/tests/test_sensord.py
```diff
@@ -158,6 +158,11 @@ def test_lsm6ds3_100Hz(self):
     data_list.sort()
     tdiffs = np.diff(data_list)
 
+    # with the skip of first 500ms measurements it happens that one value
+    # before the skip is received, this high diffs are filtered as they ruin
+    # the standard dev calculation
+    tdiffs = list(filter(lambda d: d <= 500*10**6, tdiffs))
```
adeebshihadeh, line 164: this shouldn't be necessary anymore, I added a 2s sleep in #25818

selfdrive/sensord/sensors/bmx055_accel.cc
```diff
@@ -43,6 +51,18 @@ int BMX055_Accel::init() {
   return ret;
 }
 
+int BMX055_Accel::shutdown()  {
+  // enter deep suspend mode (lowest power mode)
+
+  int ret = 0;
+  ret = set_register(BMX055_ACCEL_I2C_REG_PMU, BMX055_ACCEL_DEEP_SUSPEND);
```
adeebshihadeh, line 58:
```suggestion
  // enter deep suspend mode (lowest power mode)
  int ret = set_register(BMX055_ACCEL_I2C_REG_PMU, BMX055_ACCEL_DEEP_SUSPEND);
```

selfdrive/sensord/sensors/bmx055_gyro.cc
```diff
@@ -26,6 +27,13 @@ int BMX055_Gyro::init() {
     goto fail;
   }
 
+  ret = set_register(BMX055_GYRO_I2C_REG_LPM1, BMX055_GYRO_NORMAL_MODE);
+  if (ret < 0) {
+    goto fail;
+  }
+  // bmx055 gyro has a 30ms wakeup time from deep suspend mode
+  util::sleep_for(50);
```
adeebshihadeh, line 35: can we not write to the registers during this time?

selfdrive/sensord/sensors/bmx055_gyro.cc
```diff
@@ -54,6 +62,18 @@ int BMX055_Gyro::init() {
   return ret;
 }
 
+int BMX055_Gyro::shutdown()  {
+  // enter deep suspend mode (lowest power mode)
+
+  int ret = 0;
+  ret = set_register(BMX055_GYRO_I2C_REG_LPM1, BMX055_GYRO_DEEP_SUSPEND);
```
adeebshihadeh, line 69:
```suggestion
  // enter deep suspend mode (lowest power mode)
  int ret = set_register(BMX055_GYRO_I2C_REG_LPM1, BMX055_GYRO_DEEP_SUSPEND);
```

selfdrive/sensord/sensors/bmx055_magn.cc
```diff
@@ -155,6 +155,18 @@ int BMX055_Magn::init() {
   return ret;
 }
 
+int BMX055_Magn::shutdown() {
+  int ret = 0;
+
+  // move to suspend mode
+  ret = set_register(BMX055_MAGN_I2C_REG_PWR_0, 0);
```
adeebshihadeh, line 162:
```suggestion
  // move to suspend mode
  int ret = set_register(BMX055_MAGN_I2C_REG_PWR_0, 0);
```

selfdrive/sensord/sensors_qcom2.cc
```diff
@@ -88,6 +92,10 @@ void interrupt_loop(int fd, std::vector<Sensor *>& sensors, PubMaster& pm) {
       events.adoptWithCaveats(i, kj::mv(collected_events[i]));
     }
 
+    if (ts - init_ts < init_delay) {
+      continue;
+    }
```
adeebshihadeh, line 97: perhaps implement this as a generic data valid flag on the base sensor class. the default can be this 500ms init time, but we can then add other cases and override with more specific init checks in the future on a per-sensor basis

selfdrive/sensord/tests/test_sensord.py
gast04: retested, yes you are right

selfdrive/sensord/sensors/bmx055_gyro.cc
gast04: nope, this is behavior is a bit weird, if we dont give the bmx time after setting this bit the change will not be applied and the bmx gyro stays in deep suspend mode

selfdrive/sensord/sensors_qcom2.cc
gast04: I see, will change it

selfdrive/sensord/sensors/bmx055_gyro.cc
adeebshihadeh: is there a register we can read instead?

selfdrive/sensord/sensors_qcom2.cc
gast04: I looked into it, I would like to move this to the sensor split event PR, its just way easier to implement there, in my mind its not worth the effort doing it here, due to the preallocation of sensorEvents and so on

adeebshihadeh: sounds good

selfdrive/sensord/sensors/bmx055_gyro.cc
gast04: unfortunatelly, I havent found anything about this behaviour in the data sheet, except the wakeup time...
we could busy wait if the values got applied to the register?

selfdrive/sensord/sensors_qcom2.cc
```diff
@@ -196,6 +196,10 @@ int sensor_loop() {
     std::this_thread::sleep_for(std::chrono::milliseconds(10) - (end - begin));
   }
 
+  for (int i = 0; i < sensors.size(); i++) {
+    sensors[i]->shutdown();
+  }
```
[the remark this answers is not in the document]
gast04: done

selfdrive/sensord/sensors/bmx055_accel.cc
gast04: done

selfdrive/sensord/sensors/bmx055_gyro.cc
gast04: done

selfdrive/sensord/sensors/bmx055_magn.cc
gast04: done

selfdrive/sensord/sensors/bmx055_gyro.cc
adeebshihadeh: this is fine, we're not gonna use these anyway in a bit

selfdrive/sensord/sensors/lsm6ds3_accel.cc
```diff
@@ -71,12 +71,25 @@ int LSM6DS3_Accel::shutdown() {
...
+  value = 0;
+  ret = read_register(LSM6DS3_ACCEL_I2C_REG_CTRL1_XL, &value, 1);
+  if (ret < 0) {
+    goto fail;
+  }
+
+  value &= 0x0F;
+  ret = set_register(LSM6DS3_ACCEL_I2C_REG_CTRL1_XL, value);
+  if (ret < 0) {
+    LOGE("Could not power-down lsm6ds3 acceleration!")
```
adeebshihadeh, line 88:
```suggestion
    LOGE("Could not power-down lsm6ds3 accelerometer!")
```

gast04: changed

PR #25790, opened 2022-09-15
VIN: query non-OBD bus

sshane:


sshane:
Will re-open when we focus on moving queries over to camera bus as well

PR #25792, opened 2022-09-15
sensord test: stop sensord if interrupts not enabled

gast04:
if the assert triggers the process is not stopped which causes the test to hang indefinitely, 
lets be save and stop it and then raise the assert

### DIFF
--- a/selfdrive/sensord/tests/test_sensord.py
+++ b/selfdrive/sensord/tests/test_sensord.py
@@ -121,6 +121,10 @@ class TestSensord(unittest.TestCase):
     cls.events = read_sensor_events(5)
     managed_processes["sensord"].stop()
 
+  def tearDown(self):
+    # interrupt check might leave sensord running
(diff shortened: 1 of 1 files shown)


selfdrive/sensord/tests/test_sensord.py
```diff
@@ -261,7 +261,10 @@ def test_sensor_verify_no_interrupts_after_stop(self):
     # check if the interrupts are enableds
     with SMBus(SENSOR_BUS, force=True) as bus:
       int1_ctrl_reg = bus.read_byte_data(I2C_ADDR_LSM, 0x0D)
-      assert int1_ctrl_reg == 3, "Interrupts not enabled!"
+
+      if int1_ctrl_reg != 3:
+        managed_processes["sensord"].stop()
+        assert int1_ctrl_reg == 3, "Interrupts not enabled!"
```
adeebshihadeh, line 267: do this in a class tearDown function instead

PR #25795, opened 2022-09-15
2021 Hyundai Elantra Tune and Missing FW

CruiseBrantley:
**Description** Tune improvement for 2021 Hyundai Elantra
There was no data based tune and plotjuggler revealed the value was much too high, @sunnyhaibin helped me find a suitable value and I have tested the tune thoroughly for a couple weeks and verified through plotjuggler it has improved the curve compared to the previous Sonata values.

Here is the linked route segment plot
image

Also added some missing firmware values for fingerprinting 2021 Elantra, I can break this out into a separate PR if required.

**Route**
Route: 5a4405495d2750ef|2022-09-14--19-15-45--31


CruiseBrantley:
Actually have better values, but looking for the best before opening a new PR, @sshane has resolved the fingerprinting issues that were being experienced with my car in another PR.

PR #25797, opened 2022-09-15
Car docs: more consistent package names

sshane:
@jyoung8607 I checked a few brochures and Driver Assistance always included ACC and Lane Assist. Do you know any MYs this isn't the case?

If this isn't right, we can probably also just remove `+ Lane Assist` and make it `Adaptive Cruise Control`


jyoung8607:
Last I touched this (#22689) I sort of threw up my hands at the morass of combinations and wrote what's there today. I feel like there's a reason I didn't go with Driver Assistance last year, IIRC Audi was a little fast-and-loose at times describing lower-end packages with only rear-side radars and emergency braking as having driver assistance features. But your proposal isn't wrong, and matching the rest of VW with "Driver Assistance" is probably good enough for now.

In the medium-term future I'd like to reduce it to just "Adaptive Cruise", but that change would have several other moving parts which require mutual consent and support to move forward. It's a topic for another day.

PR #361, opened 2018-09-11
Adds 2018 Accord Hybrid

energee:
We determined the hybrid had a different fingerprint a few weeks ago but found out recently that user brake is detected like the civic hatchback and not like other boschs. We have two hybrid owners as testers.

PR #25800, opened 2022-09-15
boardd: add CAN health to pandaStates

briskspirit:
after cereal PR #361

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -382,7 +382,7 @@ std::optional<bool> send_panda_states(PubMaster *pm, const std::vector<Panda *> 
     cs.push_back(ps.initCanState1());
     cs.push_back(ps.initCanState2());
 
-    for (uint32_t j = 0; j < 3; j++) {
+    for (uint32_t j = 0; j < PANDA_CAN_CNT; j++) {
       const auto &can_health = pandaCanStates[i][j];
       cs[j].setBusOff((bool)can_health.bus_off);
       cs[j].setBusOffCnt(can_health.bus_off_cnt);
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -302,6 +302,7 @@ std::optional<bool> send_panda_states(PubMaster *pm, const std::vector<Panda *> 
   auto pss = evt.initPandaStates(pandas.size());
 
   std::vector<health_t> pandaStates;
+  std::vector<std::vector<can_health_t>> pandaCanStates;
   for (const auto& panda : pandas){
     auto health_opt = panda->get_state();
     if (!health_opt) {
@@ -318,6 +319,7 @@ std::optional<bool> send_panda_states(PubMaster *pm, const std::vector<Panda *> 
       }
       can_health[i] = *can_health_opt;
     }
+    pandaCanStates.push_back(can_health);
 
     if (spoofing_started) {
       health.ignition_line_pkt = 1;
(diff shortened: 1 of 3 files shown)


selfdrive/boardd/boardd.cc
```diff
@@ -302,6 +302,7 @@ std::optional<bool> send_panda_states(PubMaster *pm, const std::vector<Panda *>
   auto pss = evt.initPandaStates(pandas.size());
 
   std::vector<health_t> pandaStates;
+  std::vector<std::vector<can_health_t>> pandaCanStates;
```
adeebshihadeh, line 305: the size is already known here, use something like std::array instead of vector

selfdrive/boardd/boardd.cc
```diff
@@ -366,6 +377,35 @@ std::optional<bool> send_panda_states(PubMaster *pm, const std::vector<Panda *>
     ps.setInterruptLoad(health.interrupt_load);
     ps.setFanPower(health.fan_power);
 
+    std::vector<cereal::PandaState::PandaCanState::Builder> cs;
+    cs.push_back(ps.initCanState0());
+    cs.push_back(ps.initCanState1());
+    cs.push_back(ps.initCanState2());
```
adeebshihadeh, line 383: same here, or move it into the loop

selfdrive/boardd/boardd.cc
```diff
@@ -310,6 +311,16 @@ std::optional<bool> send_panda_states(PubMaster *pm, const std::vector<Panda *>
 
     health_t health = *health_opt;
 
+    std::vector<can_health_t> can_health;
+    for (uint32_t i = 0; i < 3; i++) {
```
adeebshihadeh, line 315: use PANDA_CAN_CNT

selfdrive/boardd/boardd.cc
```diff
@@ -310,6 +311,16 @@ std::optional<bool> send_panda_states(PubMaster *pm, const std::vector<Panda *>
 
     health_t health = *health_opt;
 
+    std::array<can_health_t, 3> can_health{};
```
adeebshihadeh, line 314: use PANDA_CAN_CNT

selfdrive/boardd/boardd.cc
```diff
@@ -366,6 +377,32 @@ std::optional<bool> send_panda_states(PubMaster *pm, const std::vector<Panda *>
     ps.setInterruptLoad(health.interrupt_load);
     ps.setFanPower(health.fan_power);
 
+    std::array<cereal::PandaState::PandaCanState::Builder, 3> cs = {ps.initCanState0(), ps.initCanState1(), ps.initCanState2()};
```
adeebshihadeh, line 380: same here

selfdrive/boardd/boardd.cc
```diff
@@ -302,6 +302,7 @@ std::optional<bool> send_panda_states(PubMaster *pm, const std::vector<Panda *>
   auto pss = evt.initPandaStates(pandas.size());
 
   std::vector<health_t> pandaStates;
+  std::vector<std::array<can_health_t, 3>> pandaCanStates;
```
adeebshihadeh, line 305: also don't need the outer vector, can replace the health one too

briskspirit: confused why... vector of can_heath_t array objects for multiple pandas.

adeebshihadeh: I meant you can make it an array too

briskspirit: it is unknown size same as pandaStates.

adeebshihadeh: it's `pandas.size()`

PR #25723, opened 2022-09-10
Hyundai: support HDA1 EV6 and Ioniq 5

hoomoose:
**Simple code fix to make OP work with hda1 ev6 & I5 - no full port required!**

**Description** 
OP does not work with HDA1 versions of Kia EV6 or Hyundai Ioniq5 giving CAN Errors. Changing all instances of the ACCELERATOR signal to be based on "alt_buttons" flag rather than "hda2" flag fix this issue. This replaces the need for the Car Port PR https://github.com/commaai/openpilot/pull/25629 that has subsequently been closed.

Fixes:
carstate.py
1) The ACCELERATOR signal on the hda1 EV6/I5 is the same as the hda2 EV6/I5, therefore I changed the logic in carstate.py to pick the ACCELERATOR notation to use based on the "alt_buttons" flag rather than "hda2" flag
2) The Accelerator check at the bottom of carstate.py was made to have check based on alt_buttons flag rather than hda2 flag.

values.py
1) I updated notation so that hda1 and hda2 have separate harness requirements
<s>2) Max steer needed to be 384 to make basic turns - the attached route demonstrates this. Without this adjustment, unnecessary "max steering" alerts occur.</s>

<s>override.yml
1) The EV6 required a lower Lateral Accel of 2.5 in order to perfectly complete the turns in the route provided when the Max steer was adjusted to 384</s>

**Prerequisite Minor Additional Panda Fix Required**
https://github.com/commaai/panda/pull/1063
1) The Accelerator check signal needed to be on bus 0 for hda1 EV6/I5.
2) The Accelerator RX check was changed to be based on "alt_buttons" flag rather than "hda2" flag
<s>3) Max steer was adjusted to 384</s>

**Verification** 
This branch has been installed on my C3 (with the Panda changes made locally), force fingerprinted to be an EV6, and tested on my hda1 EV6 (Wind trim). The dramatic S-curve performs perfectly with this branch! You can check "torque request" data as well as "lat accell" data to see that the "steer max" and "lat accel" changes were required to perform this short route with an S-curve.

**Route**
Route w/ 270 Max Steer and failure: 68d6a96e703c00c9|2022-09-10--16-09-39--0
Route w/ 384 Max Steer and perfection: 68d6a96e703c00c9|2022-09-10--16-25-04--0

PR #25801, opened 2022-09-16
hyundai: ev6 - use good steering angle sensor

gregjhogan:
Our EV6 has a rack mounted power steering motor, and we have signals for two different steering angles.  I have a theory that one sensor is on the steering column and the other is on the steering rack, and we are currently using the one on the steering column which looks like it may be a much worse signal (see data from `d824e27e8c60172c|2022-09-15--09-57-55--6` below).

image

Not sure if we need to worry about some CAN-FD cars having a rack mounted power steering motor and other cars having a column mounted power steering motor.

Before merging:
- [x] Decide if it is OK the logged angle rate is for a different sensor (@HaraldSchafer says that is fine)
- [x] Determine if `LAT_ACCEL_FACTOR` should change (@HaraldSchafer can you take a look?)
- [ ] Determine what platforms we should apply this change to (@adeebshihadeh can you look at data for other cars this change would currently affect and exclude them if necessary?)


gregjhogan:
Do we care if the steering angle rate is for a different sensor than the steering angle?

sshane:
Do you see a real difference between the acceleration derived from the steering angle and what the localizer says?

gregjhogan:
> Do you see a real difference between the acceleration derived from the steering angle and what the localizer says?

I haven't looked at the data for that, but In the image of the above segment I see at one point the angle in one signal changing about 0.1 deg and the other changing almost 2 deg when they diverge and come back together.

hoomoose:
I have an hda1 ev6 and can check if I have two steering angles like your ev6 if that is helpful at all. Let me know what I should look for or post.

gregjhogan:
The localizer seems to confirm this change is an improvement
image

mattwebbio:
Saw @hoomoose comment in the Discord that these changes completely eliminated ping pong in his HDA1 EV6 Wind, so I cherry-picked these two commits (https://github.com/commaai/openpilot/pull/25801/commits/4cde6afd36e22cb15d3544ec3d2410e34e0892a8, https://github.com/commaai/openpilot/pull/25801/commits/508c00fd0642a30251ab678a8daf6b249cca1604) on to my local copy of [his HDA1 EV6 branch](https://github.com/commaai/openpilot/pull/25723). Interestingly, I had the exact opposite experience - in my HDA1 EV6 Light, these two commits on top of that branch result in extreme oversteering. Are there more recent changes in master that this PR depends on?

ioniq5-cz:
I also cherry picked those changes into my IONIQ 5 and drive IS now super smooth. Definitely an improvement!

sshane:
From a Tucson Hybrid route, it appears `MDPS->STEERING_ANGLE` is always zero, however `MDPS->STEERING_ANGLE_2` *is* set. It also looks nearly identical to the current `STEERING_SENSORS->STEERING_ANGLE` signal.

Screenshot from 2022-09-27 19-07-44
Screenshot from 2022-09-27 19-10-21

gregjhogan:
On the ev6 `STEERING_SENSORS->STEERING_ANGLE` also looked extremely similar a lot of the time. It does seems like the [Tuscon has a rack mounted](https://www.hyundai.news/eu/models/suv/tucson/press-kit/the-all-new-hyundai-tucson-a-smart-tech-hero-with-a-standout-design.html#:~:text=The%20all%2Dnew%20Tucson%E2%80%99s%20Rack%2Dmounted%20Motor%20Driven%20Power%20Steering%20(R%2DMDPS)) EPS (R-MDPS).

sshane:
It looks like it can vary per-platform as well:

```python
set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=1526
platform='KIA EV6 2022', segment='22de8111a8c5463c|2022-07-29--13-34-49--6', len(new_steering_angle)=5999, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=1377
platform='HYUNDAI TUCSON HYBRID 4TH GEN', segment='36e10531feea61a4|2022-08-16--18-14-13--0', len(new_steering_angle)=5558, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=650
platform='HYUNDAI TUCSON HYBRID 4TH GEN', segment='36e10531feea61a4|2022-08-16--18-14-13--7', len(new_steering_angle)=6000, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=1814
platform='KIA EV6 2022', segment='22de8111a8c5463c|2022-07-29--13-34-49--1', len(new_steering_angle)=5999, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=1257
platform='KIA EV6 2022', segment='22de8111a8c5463c|2022-07-29--13-34-49--0', len(new_steering_angle)=5930, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=1369
platform='KIA EV6 2022', segment='22de8111a8c5463c|2022-07-29--13-52-45--2', len(new_steering_angle)=6000, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=140
platform='HYUNDAI TUCSON HYBRID 4TH GEN', segment='36e10531feea61a4|2022-08-16--18-14-13--4', len(new_steering_angle)=5999, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=72
platform='HYUNDAI TUCSON HYBRID 4TH GEN', segment='36e10531feea61a4|2022-08-16--18-14-13--5', len(new_steering_angle)=5999, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=846
platform='KIA EV6 2022', segment='22de8111a8c5463c|2022-07-29--13-34-49--5', len(new_steering_angle)=5999, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=97
platform='HYUNDAI TUCSON HYBRID 4TH GEN', segment='36e10531feea61a4|2022-08-16--18-14-13--3', len(new_steering_angle)=6000, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=52
platform='HYUNDAI TUCSON HYBRID 4TH GEN', segment='36e10531feea61a4|2022-08-16--18-14-13--6', len(new_steering_angle)=5999, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=362
platform='HYUNDAI TUCSON HYBRID 4TH GEN', segment='36e10531feea61a4|2022-08-16--18-14-13--2', len(new_steering_angle)=5999, new_steering_angle == new_steering_angle_2=False

set(new_steering_angle)={-0.0}, len(set(new_steering_angle))=1, len(set(new_steering_angle_2))=1280
platform='HYUNDAI TUCSON HYBRID 4TH GEN', segment='36e10531feea61a4|2022-08-16--18-14-13--1', len(new_steering_angle)=5999, new_steering_angle == new_steering_angle_2=False
```

We might need live detection like we do for Toyota

sshane:
It looks like 2023 EV6 has a R-MDPS, perhaps this user has a 2023 but labeled 2022 as they're foreign. Yes means the two signals in MDPS match and the current signal is not precise, no means they don't match and the current signal is as precise as the signal in MDPS: `{'KIA EV6 2022': {'yes': 154, 'no': 5}, 'HYUNDAI IONIQ 5 2022': {'yes': 40}, 'HYUNDAI TUCSON HYBRID 4TH GEN': {'no': 8}}`

sshane:
There are four different situations here:

- Tucson Hybrid: `MDPS->STEERING_ANGLE` is not set, and `MDPS->STEERING_ANGLE_2` is nearly identical to `STEERING_SENSORS->STEERING_ANGLE` (~0.05-0.1 degree angle error across 8 segments). We can use `MDPS->STEERING_ANGLE_2` here.
- EV6: `MDPS->STEERING_ANGLE` equals `MDPS->STEERING_ANGLE_2` and has an average of 0.9 degrees of error across 154 segments (max segment error was 2.32 deg, min 0.18 deg).
- IONIQ 5 (40 segments): `MDPS->STEERING_ANGLE` equals `MDPS->STEERING_ANGLE_2` and has an average of 0.80 degrees of error across 40 segments (max segment error was 1.75 deg, min 0.25 deg).
- IONIQ 5 (5 segments): All from one user, but this IONIQ 5 is similar to the TUCSON in that `MDPS->STEERING_ANGLE` is not set. Average angle error with `STEERING_SENSORS->STEERING_ANGLE` is 0.75 deg (max segment error was 1.1 deg, min 0.62 deg) The difference between the Tucson though is that the two signals have a higher error, and precision looks to be about the same for both. Toss up on whether either signal is more accurate to the localizer's accel so it's fine to use the new signal here: https://imgur.com/a/8Sk6TcS

sshane:
It looks like this new signal can be more noisy and diverge from the localizer accel compared to the previous signal. Since this isn't easy to merge anymore, going to leave this open for now.

EV6: https://imgur.com/a/V7lmbKU

gregjhogan:
When I zoom in on a random place where the MDPS signal is noisy, the noise seems to have the same frequency as the motor torque output.  I feel like the localizer may filter this level of noise out, and things like suspension dynamics may also hide changes in steering angle from the localizer at the level we are looking here, but maybe I am wrong.
image

mattwebbio:
Is there any chance you (Comma) would consider merging the torque values separate from the switch to MDPS? Using the MDPS steering angle sensor makes OP unusable for me on the EV6 Light (maybe I don't have it?), but without the torque settings in this PR I have pretty strong understeer.

gregjhogan:
recent example where ping pong was reported that looks similar `5e31c7585bf1a946|2022-11-04--00-00-39--9`
image

adeebshihadeh:
@sshane what's the verdict on this?

@nuwandavek is it worth switching the signal for a single platform?

sshane:
I found that the offset between the current and this signal roughly matches what `paramsd` finds using lateral acceleration, I need some time to find why I thought the new signal was the one that diverged from the localizer accel.

sunnyhaibin:
Looked at the data from the 2022 Hyundai Tucson Hybrid and this signal has much less noise than the current one used in upstream.

PR #25802, opened 2022-09-16
Multilang: add missing Chinese (Traditional) translations

eFiniLan:
added missing cht translation
@sshane :)


sshane:
Added some new translations I didn't know we were missing, can you open one more PR?

eFiniLan:
> Added some new translations I didn't know we were missing, can you open one more PR? [64c2d4b](https://github.com/commaai/openpilot/commit/64c2d4b30f7d4be1e1726a5668c548ecf8d6e32e)

Please see https://github.com/commaai/openpilot/pull/25807 :)

PR #25810, opened 2022-09-16
GM: update minimum accel

sshane:
This never made sense to me, if `ACCEL_MIN` was lowered from the normal limit of -3.5 m/s^2 to achieve more braking, this change did the opposite. It only moved the breakpoint further away.

For example, if we wanted to achieve -3.5 m/s^2 the current code would only apply 305 units (equivalent to -3.05 m/s^2). ***Then***, longcontrol clips the output of actuators.accel to -3.5, so we would never be able to brake at -3.5, much less -4 m/s^2!

I see in the original repo, compute_gb used to divide accel by 3 when positive and 6 when negative which seemed even worse to me. It was then changed to be the average of the two at 4 in a clean up PR. Do you know why this was like this @HaraldSchafer?


sshane:
Bolt is also -3.5

Screenshot from 2022-09-14 00-07-49

HaraldSchafer:
I don't think what you are saying is true. Longcontrol applies the limits set in the carcontroller, so that would be -4? Nothing universally limits actuators.accel to -3.5? The discrepancy here seems to be that the max brake doesn't actually get you -4m/ss? But the comments are still valid, if you want to achieve -3.5 on a downhill you will need to apply an amount of brakes that would do more than -3.5 on a flat road, like -4 for example. So to me it seems like the max brake value should just be increased to make everything function the way the comments intended.

sshane:
Ah my bad, you're right. I was looking at `ACCEL_MIN_ISO` for feedforward clipping which was removed here: https://github.com/commaai/openpilot/commit/210a6163ac9a8ccc425114fd722e864befa77966

Yep, sounds like we should lower `MAX_BRAKE` to match the min accel though! I'll update safety and test on a few cars to make sure the car accepts this value (some community members have already changed the min brake and are running with it as they claim you get less regen at low speeds).

sshane:
A Volt tried this, they got -4.0 to -4.1 m/s^2 across the full speed range at 400 brake units: c11fcb510a549332|2022-10-03--14-43-17--0. No faults.

Same with Bolt EUV: -4.0 to -4.1 m/s^2 from 30 m/s to 0: https://connect.comma.ai/376bf99325883932/1664866516475/1664866539087

Screenshot from 2022-10-04 00-16-28

selfdrive/car/gm/values.py
```diff
@@ -23,11 +23,12 @@ class CarControllerParams:
   ADAS_KEEPALIVE_STEP = 100
   CAMERA_KEEPALIVE_STEP = 100
 
-  # Volt gasbrake lookups
-  # TODO: These values should be confirmed on non-Volt vehicles
+  # Volt gas/brake lookups
+  # TODO: These values should be confirmed on non-Volt vehicles.
+  # TODO: MAX_GAS should achieve 2 m/s^2 and MAX_BRAKE with regen should achieve -4.0 m/s^2
```
sshane, line 28:
```suggestion
  # MAX_GAS should achieve 2 m/s^2 and MAX_BRAKE with regen should achieve -4.0 m/s^2
```

PR #25811, opened 2022-09-16
Multilang: revise some Japanese translations

ambientocclusion:
**Description** 
Improve some Japanese translations.

@sshane 
let me know which ones are too long.


selfdrive/ui/translations/main_ja.ts
```diff
@@ -1169,7 +1169,7 @@ location set</source>
     <message>
         <location line="+11"/>
         <source>🌮 End-to-end longitudinal (extremely alpha) 🌮</source>
-        <translation>🌮 エンドツーエンドのアクセル制御 (超アルファ版) 🌮</translation>
+        <translation>🌮 エンドツーエンドのアクセル制御 (超α版) 🌮</translation>
```
sshane, line 1172: Nice! This was one of the long ones

sshane:
Screenshot from 2022-09-16 01-32-27

ambientocclusion:
> Screenshot from 2022-09-16 01-32-27
> 
> I'll check onroad tomorrow

Hmm there are two ways I can think of.

- One way is 実行 (Proceed).
 I feel like it's suitable here, but It's not direct word translation for "uninstall", so if same translation is used somewhere else,  might be a problem. I don't fully understand Qt translation system, but it seems like it doesn't specify which of the translation source is used if same word is used.

- Other way is ｱﾝｲﾝｽﾄｰﾙ as some of the other translations using half Katakana, but half Katakana rarely used in general like emphasizing something unusual.

sshane:
Translations are usually limited to their immediate class, or context. In this case it's the `SoftwarePanel` class/context. Proceed sounds good to me, we just need to be careful not to re-use it in that class, then we could switch to half-width Katakana

ambientocclusion:
@sshane Ok, changed "アンインストール" to "実行". Let me know if any other ones are too long.

selfdrive/ui/translations/main_ja.ts
```diff
@@ -309,7 +309,7 @@
     <name>MapETA</name>
     <message>
         <source>eta</source>
-        <translation>予定到着時間</translation>
+        <translation>到着予定時間</translation>
```
sshane, line 312: This might be too long for nav, but let's fix in another PR. this isn't a regression

sshane:
thanks for the PR!

PR #25812, opened 2022-09-16
Multilang: update pt-BR translations

AlexandreSato:
I translated the last terms, but for some reason I'm not able to compile the view: 

image

I'll wait/solve to confirm the PR


AlexandreSato:
I reduced some texts that were very close to the corners preserving understanding.

sshane:
Looks good! Thanks for the PR

PR #25819, opened 2022-09-16
RPv2: fix data length check

briskspirit:


selfdrive/boardd/panda.cc
```diff
@@ -389,7 +389,8 @@ void Panda::pack_can_buffer(const capnp::List<cereal::CanData>::Reader &can_data
     }
     auto can_data = cmsg.getDat();
     uint8_t data_len_code = len_to_dlc(can_data.size());
-    assert(can_data.size() <= ((hw_type == cereal::PandaState::PandaType::RED_PANDA) ? 64 : 8));
+    assert(can_data.size() <= ((hw_type == cereal::PandaState::PandaType::RED_PANDA ||
+                                hw_type == cereal::PandaState::PandaType::RED_PANDA_V2) ? 64 : 8));
```
adeebshihadeh, line 393: think it's fine to just always check <= 64

briskspirit: What if somebody adds canfd message for black panda when porting?

PR #25823, opened 2022-09-17
Subaru: match ego speed on dash

sshane:
Not sure if there's any factor here (C3 seemed like it updated earlier when accelerating and about the same when decelerating), but seems like dash applies both hysteresis and delay (filtering?).

- [x] ~Find a signal first before merging~ (couldn't find one from the car or camera)
- [ ] Test some other Subarus

### DIFF
--- a/selfdrive/car/subaru/carstate.py
+++ b/selfdrive/car/subaru/carstate.py
@@ -13,6 +13,11 @@ class CarState(CarStateBase):
     can_define = CANDefine(DBC[CP.carFingerprint]["pt"])
     self.shifter_values = can_define.dv["Transmission"]["Gear"]
 
+    # TODO: cluster is super delayed when updating, could it be just hysteresis or also a lower update rate?
+    # "felt" like 1-2 Hz, try to measure
+    self.cluster_speed_hyst_gap = CV.KPH_TO_MS
+    self.cluster_min_speed = CV.KPH_TO_MS
+
   def update(self, cp, cp_cam, cp_body):
     ret = car.CarState.new_message()
 


selfdrive/car/subaru/carstate.py
```diff
@@ -13,6 +13,11 @@ def __init__(self, CP):
     can_define = CANDefine(DBC[CP.carFingerprint]["pt"])
     self.shifter_values = can_define.dv["Transmission"]["Gear"]
 
+    # TODO: cluster is super delayed when updating, could it be just hysteresis or also a lower update rate?
+    # "felt" like 1-2 Hz, try to measure
+    self.cluster_speed_hyst_gap = CV.KPH_TO_MS
+    self.cluster_min_speed = CV.KPH_TO_MS
```
sshane, line 19:
```suggestion
    self.cluster_speed_hyst_gap = 0.15
    self.cluster_min_speed = CV.KPH_TO_MS
```

sshane:
@michaelhonan hey do you mind testing this branch? Let me know how much above or below the C3 is vs. your dash, and if possible a video would be a great help! On our Outback it's sometimes above by 1 mph and sometimes matches, so it's likely a very small factor and some hysteresis + low sample rate.

michaelhonan:
Merged branch into my fork (containing only tuning values / increased torque: https://github.com/michaelhonan/openpilot/commit/ac44ef7a4c0489b9bca3a5a6c56fd29eddbaff76)

Saw a difference of 1-3kph compared to the dash value.

**Video:**
https://www.youtube.com/watch?v=1T5_0EU0t4E

**Segment from video:**
dc30ef2ad294a732|2022-09-17--19-51-40--2

PR #25827, opened 2022-09-17
camerad: refactor struct MultiCameraState into class CameraServer

deanlee:


system/camerad/cameras/camera_server.h
```diff
@@ -0,0 +1,36 @@
+#pragma once
```
adeebshihadeh, line 1: let's move the camera server files out of cameras/, that dir should probably just be for the image sensor files

PR #25830, opened 2022-09-17
sensord: add helper function verify_chip_id to I2CSensor

deanlee:


selfdrive/sensord/sensors/i2c_sensor.h
```diff
@@ -33,4 +34,21 @@ class I2CSensor : public Sensor {
   virtual int init() = 0;
   virtual bool get_event(cereal::SensorEventData::Builder &event) = 0;
   virtual int shutdown() = 0;
+
+  template <class... Args>
+  std::optional<uint8_t> verify_chip_id(uint8_t address, Args... args) {
```
adeebshihadeh, line 39: let's keep this simple - no need for an std::optional or template. for the cases where it can be multiple chip id's, either call the function twice or take a vector instead

PR #25714, opened 2022-09-09
Sensor events splitup

gast04:
Split up `sensorEvents` into one message per sensor, and immediately publish once the measurement is ready.

- [ ] update xx
- [x] update provisioning test

PR #25832, opened 2022-09-17
sensord: build message without orphans

deanlee:
the orphan is slower and more complicated than calling `initSensorEvents` directly, and it will leave memory holes in the serialized message. the only way to remove these extra spaces is to copy the message to a new MessageBuilder, since only the reachable objects will be copied.


gast04:
yea, the orphan's are not ideal, we have currently a refactor going on, which splits the sensor events and also removes the 
orphans (https://github.com/commaai/openpilot/pull/25714), so I would like to close this one

PR #25833, opened 2022-09-17
ui/DriveStats: small changes

deanlee:
1. read 'IsMetric' in one place.
2. call updateStats on doc changed


sshane:
Is this ready?

deanlee:
yes.forgot to open it

adeebshihadeh:
Thanks, but I don't think the drive stats will be around much longer. We're thinking of putting something else there.

PR #25837, opened 2022-09-19
boardd: new helper function usb_open_if

deanlee:


adeebshihadeh:
We don't need to saves lines like this, especially using anything that looks like `template <class Predicate>`.

deanlee:
It's not just about saving code. There are still some bugs in the original code.

I will submit code to fix these bugs first. But there are too many goto statements,make it hard to understand and even harder to maintain.

I think it's better to refactor it instead of patching



But I will still submit another version without the template. Rewrite the current code. Too much goto code now

deanlee:
It's not just about saving code. There are still some bugs in the original code.

I will submit pr to fix these bugs first.

PR #25844, opened 2022-09-19
UI: draw path from planner

adeebshihadeh:


selfdrive/controls/lib/longitudinal_planner.py
```diff
@@ -153,6 +157,9 @@ def publish(self, sm, pm):
     longitudinalPlan.modelMonoTime = sm.logMonoTime['modelV2']
     longitudinalPlan.processingDelay = (plan_send.logMonoTime / 1e9) - sm.logMonoTime['modelV2']
 
+    longitudinalPlan.xs = self.x_desired_trajectory.tolist()
+    longitudinalPlan.zs = self.z_model.tolist()
+
```
sshane, line 162: Want to call these `distances` and `heights` to be consistent with the other names?

sshane:
On master, update_model takes on average `0.093` ms to execute with a standard deviation of `0.065` ms.

On this branch:

`update_model()` takes on average `0.078 ms` to execute with a standard deviation of `0.033 ms`.

`update_plan()` takes on average `0.013 ms` to execute with a standard deviation of `0.006 ms`.

So together, they average `0.091 ms`, nearly identical to the time taken before for the combined function.

adeebshihadeh:
Closing for now. 2.5s isn't enough at low speeds.

PR #25849, opened 2022-09-20
sensord: move bmx accel,gyro to deep suspend on startup, deprecate bmx temp and light

gast04:
bmx sensor data is not used anymore, and will not be used in the future, therfore the bmx gyro and accel is moved into a low power state on startup, the light sensor is discontinued. And also the bmx tempererature will not be used anymore


adeebshihadeh:
Decided to keep these since we want to keep checking them in provisioning.

gast04:
:smiling_face_with_tear:

PR #25852, opened 2022-09-20
Kia: update Optima platform name

sshane:
From: https://en.wikipedia.org/wiki/Kia_K5 and https://carbuzz.com/cars/kia/optima-generations#4th-gen-jf

2015-2018 is 4th gen, 2019-2020 is 4th gen facelift


sshane:
@jyoung8607 what is your convention for a facelift? Add a `.1`?

crwusiz:
optima(k5)
2016~2018 (JF) 2th gen ~ 2.5th gen
2019~ (DL3) 3th gen

jyoung8607:
> @jyoung8607 what is your convention for a facelift? Add a `.1`?

In VW's convention, usually `.5` (or `.1` for Transporter family). My 2018 Golf R is a Mk7.5. We don't really note most facelifts because they don't matter to us (except for Transporter, where T6 doesn't have EPS but T6.1 does).

As long as you find some nice logical grouping other than model-year (because that works really badly, especially for global markets and manufacturers slow-rolling refreshes/recycling model names) I think you can label those groups in whatever manner best fits that manufacturer. As an example, both Hyundai and Chrysler seem to have two or three character platform identifiers for their openpilot-relevant generations, we're free to use those if they make sense.

sshane:
> optima(k5)
> 
> 2016~2018 (JF) (wiki is 4th gen) ( be called 2th gen ~ 2.5th gen (fase lift) )
> 
> 2019~ (DL3) (wiki is 5th gen) ( be called 3th gen )
> 
> I think it's easy to understand how to classify it by vehicle code name

2021 model year is 5th gen not 4th

jyoung8607:
Speaking of per-manufacturer naming conventions...

  `CAR.KIA_OPTIMA_MK4`

The MkX generation nomenclature is definitely a VW culture thing. If you were looking to standardize on generations, I'd consider doing `_G4` instead of `_MK4`. I would be fine with aligning VW to that, or to the 2-3 character platform IDs.

gregjhogan:
FYI, there was already some precedence for this, and the letter and number are flipped, for example:
https://github.com/commaai/openpilot/blob/754b44e7ceb2b6b029cf511b821f4328b90364f9/selfdrive/car/honda/values.py#L129
https://github.com/commaai/openpilot/blob/754b44e7ceb2b6b029cf511b821f4328b90364f9/selfdrive/car/honda/values.py#L138

sshane:
Variable name is easy enough to change. Platform name looks like they use years here still

PR #25855, opened 2022-09-21
sensord: lsm self test

gast04:
Add lsm6ds4 self test on sensord startup


selfdrive/sensord/sensors/lsm6ds3_accel.cc
```diff
@@ -28,11 +144,29 @@ int LSM6DS3_Accel::init() {
     source = cereal::SensorEventData::SensorSource::LSM6DS3TRC;
   }
 
+  ret = perform_self_test(LSM6DS3_ACCEL_POSITIVE_TEST);
```
adeebshihadeh, line 147: let’s gate the self test behind an environment variable for this release. we should always run the test and log the result, but only assert if the environment variable is set. make sure to set it in the unit tests

selfdrive/sensord/sensors/lsm6ds3_accel.h
```diff
@@ -10,21 +10,37 @@
...
+#define LSM6DS3_ACCEL_NEGATIVE_TEST   0b10
+#define LSM6DS3_ACCEL_MIN_ST_LIMIT_mg 90.0f
+#define LSM6DS3_ACCEL_MAX_ST_LIMIT_mg 1700.0f
 
 class LSM6DS3_Accel : public I2CSensor {
   uint8_t get_device_address() {return LSM6DS3_ACCEL_I2C_ADDR;}
   cereal::SensorEventData::SensorSource source = cereal::SensorEventData::SensorSource::LSM6DS3;
+
+  // self test functions
+  int perform_self_test(int test_type);
```
adeebshihadeh, line 41: i’d just call this selftest

selfdrive/sensord/sensors/lsm6ds3_gyro.cc
```diff
@@ -5,11 +5,113 @@
...
 LSM6DS3_Gyro::LSM6DS3_Gyro(I2CBus *bus, int gpio_nr, bool shared_gpio) : I2CSensor(bus, gpio_nr, shared_gpio) {}
 
+void LSM6DS3_Gyro::wait_for_data_ready() {
+  uint8_t drdy = 0;
+  uint8_t buffer[6];
+
+  do {
+    read_register(LSM6DS3_GYRO_I2C_REG_STAT_REG, &drdy, sizeof(drdy));
+    drdy &= LSM6DS3_GYRO_DRDY_GDA;
+  } while(drdy == 0);
```
adeebshihadeh, line 21: make sure to be consistent with style
 ```suggestion
  } while (drdy == 0);
```

File: selfdrive/sensord/sensors/lsm6ds3_gyro.cc [hunk 4]
```diff
@@ -5,11 +5,113 @@
...
+
+  // read first values and discard
```
adeebshihadeh, line 23: no needs for these kinds of comments. comments should explain things that aren’t immediately clear from reading the code

selfdrive/sensord/sensors/lsm6ds3_gyro.cc
```diff
@@ -5,11 +5,113 @@
...
+        (test_val[i] > LSM6DS3_GYRO_MAX_ST_LIMIT_mdps)) {
+      test_result = false;
+    }
+  }
+
+  if (!test_result) {
+    ret = -1;
+  }
+
+fail:
```
adeebshihadeh, line 111: usually goto’s are reserved for cases where you want to do some cleanup. in this case, i’d just return directly

selfdrive/sensord/sensors/lsm6ds3_accel.h
gast04: ok, will rename sticked to the one in bmx magn its called `perform_self_test` there as well

PR #25859, opened 2022-09-21
FPv2: move ecus to query to config

sshane:
Supports querying specific ecus for data collection

So we can start collecting data on the Optima's missing engine and eps without worry, and EV CAN-FD cars' VMU which handles transmission and engine-like capabilities


selfdrive/car/honda/values.py
```diff
@@ -447,7 +451,7 @@ class HondaCarInfo(CarInfo):
       b'78109-TED-Q510\x00\x00',
       b'78109-TEG-A310\x00\x00',
     ],
-    (Ecu.fwdCamera, 0x18dab0f1, None): [
+    (Ecu.fwdRadar, 0x18dab0f1, None): [
```
sshane, line 454: Weirdly, this address was labeled as both fwdCamera and fwdRadar, causing issues with labeling this ecu currently. Could it have been a typo/wrong information?

selfdrive/car/chrysler/values.py
```diff
@@ -166,14 +166,14 @@ class ChryslerCarInfo(CarInfo):
       bus=0,
     ),
   ],
+  ecus={(0x742, None): Ecu.combinationMeter, (0x744, None): Ecu.srs, (0x747, None): Ecu.abs,
+        (0x753, None): Ecu.fwdRadar, (0x75a, None): Ecu.eps, (0x7e0, None): Ecu.engine,
+        (0x7e1, None): Ecu.transmission, (0x761, None): Ecu.eps},
+  # Ecus present in hybrids, currently in data collection mode
+  extra_ecus={(0x7e2, None): Ecu.hcp, (0x7e4, None): Ecu.abs},
```
sshane, line 173: If we don't really care about being too careful, we can just add these ecus on to the main ecu dictionary. I don't feel too strongly one way or the other (together with a comment separating them, or another field like `extra_ecus` and have a tests for all this)

@adeebshihadeh thoughts?

sshane:
So we can start collecting data on the Optima's missing engine and eps without worry, and EV CAN-FD cars' VMU which handles transmission and engine-like capabilities

sshane:
Not sure if I like specifying all ECUs in config without also replacing addr/subaddr with Ecu enums everywhere. Will make a PR to just add extra ECUs to query and re-visit

PR #25860, opened 2022-09-21
Hyundai: match stock FW queries

sshane:
Opened a PR just to preserve this data, some stuff will likely be split out. This is what diagnostic software requests for these ECUs, avoids the multi query's issue with changing bytes appended to the end of the versions, and removes a lot of redundant information


sshane:
Optima:
- engine - 0x22 0xf1 0x91 returns the same and is what the debug tool uses for engine (vehicle manufact. ecu hardware number)
- eps - `0x1a 0x80` seems like this is completely manufacturer-specific, only fields are stuff like `MDPS Type`, nothing generic (also puts it in a weird diagnostic mode first: `0x10 0x90`)
- transmission - 0x22 0xf1 0x81 (application software number)
- abs - 0x22 0xf1 0x00 (vehicle name, system name, variant code, hw ver, sw ver, release date, hmc/kmc part config ecu serial number)
- fwdRadar 0x22 f1 0x00

EV6:
- eps - 0x22 0xf1 0x00
- abs - 0x22 0xd1 0x00
- camera - same
- radar - same
- vcu (ev engine & trans) - 0x22 0xf1 0x81 (application s/w id)

sshane:
Just a reference

PR #25865, opened 2022-09-21
loggerd: refactor RemoteEncoder from struct to class

deanlee:
The original code is brilliant but it's not easy to understand and maintain. so I'm trying to refactor `RemoteEncoder` from struct to class, with the following main changes:

1. split `handle_encoder_msg` into small functions.
2. remove recursive call to `handle_encoder_msg`.
3. use function pointers to make code more concise.
4. make it easy to move VideoWriter  to threads in the future.(or write video in encoderd)


adeebshihadeh:
Glanced over this and looks cleaner, but this is way too big as a single PR. Can you split this up into smaller chunks of ~50 lines max?

PR #25866, opened 2022-09-21
HKG: Remove Kia Stinger from 255 torque blacklist

kmhan1221:


kmhan1221:
**Description** 
Remove Kia Stinger 2018 from 255 torque blacklist to verify that ECU can command the 384 steer max

**Verification** 
Test drove vehicle through a very tight highway exit, verifying in cabana that 384 is being commanded.
image
MicrosoftTeams-image (11)



**Route**
Route: 8c809cffb861be71|2022-09-21--16-10-17--1

gregjhogan:
You also want to check if the motor actually outputs more torque. The original reason this was limited to 255 was not because the vehicle wouldn't accept higher requests (if it is just a deadzone at the top it will actually make steering worse in some cases).

kmhan1221:
@gregjhogan Could you point me to which CAN message you're looking for? I'm having a hard time finding a message representing actual torque applied

incognitojam:
> I'm having a hard time finding a message representing actual torque applied
Try this one https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/carstate.py#L84

kmhan1221:
Screenshot_20220921_223536_Chrome
Here's the MDPS_OutTq graph.

kmhan1221:
The Genesis G70 shares the same Rack and Pinion EPS with the Stinger and if it has been confirmed already that the G70 EPS can output 3 Nm of torque (G70 is not in the blacklist), my assumption is that the Stinger EPS can as well. 
image
Here's a (Korean) blogpost discussing the different types of Hyundai Mobis EPS systems, as well as a short interview with one of the chief engineers, fyi.

gregjhogan:
The limits are all software (the max you get from lane keep assist is way below the assist you get manually steering), so you still need to verify.  Should be easy looking at the motor torque output considering this should be adding ~50% more torque!

kmhan1221:
@gregjhogan I'll record some drives with the 255 torque and some with 384 and get a comparison of the OutTq value!

sshane:
Hey! Got any routes from your tests?

kmhan1221:
@sshane I have a solid route for the 255 torque trying its best to take a turn, I still need to go through the same curve sometime with the 384 torque at the same speed. Which Cabana message displays the vehicle's speed?

255 route: 8c809cffb861be71|2022-09-25--07-47-47--14
384 route: coming soon

kmhan1221:
(I understand that ideally I want to test with exact same conditions on the same branch/fork, this is an unofficial report from personal testing tonight with a different fork. I simply wanted to verify that the hardware can actually output ~15 torque )

Here's a route running Sunnypilot with 384 torque. 
8c809cffb861be71|2022-09-27--19-28-46--18
Screenshot_20220927_213626_Chrome


You can see that the output torque reaches 15 without driver input, which is a 50% increase from the 255 torque which could only output up to 10 in similar curvature/speed 


Screenshot_20220925_223448_Chrome

As an anecdotal side note, I have not run into any issues in lateral control running the 384 torque on both stock and custom forks in highway and local road environments, and in fact I feel more confident that the car will handle a tight turn than when I'm running the 255 torque. 

It's tough testing curves here in Phoenix when highways are so straight and full of traffic :( but I'll keep trying getting some better data points on stock.

PR #25867, opened 2022-09-22
pin sensord and lsm interrupt to core 1, for better timing

gast04:
Pin sensord and lsm gpio interrupt to core1, this gives better timings, the high outlier values  basically vanished (only saw one in 150k samples)

before: (this image is slightly cropped, otherwise the outliers ruin the comparison)
hist_nopin_interrupts_26k_accel

after:
hist_pin_interrupts_26k_accel_3

same result obtained from the lsm gyro. as can be seen the values are very consistent in three piles, all within 150us)

### DIFF
--- a/selfdrive/sensord/sensors_qcom2.cc
+++ b/selfdrive/sensord/sensors_qcom2.cc
@@ -218,5 +218,11 @@ int sensor_loop() {
 
 int main(int argc, char *argv[]) {
   setpriority(PRIO_PROCESS, 0, -18);
+
+  cpu_set_t mask;
+  CPU_ZERO(&mask);
+  CPU_SET(1, &mask);
+  sched_setaffinity(0, sizeof(mask), &mask);
+
   return sensor_loop();
 }
--- a/selfdrive/sensord/sensors_qcom2.cc
+++ b/selfdrive/sensord/sensors_qcom2.cc
@@ -174,7 +174,9 @@ int sensor_loop() {
     return -1;
   }
 
-  // pin interrupt to core 1
+  // increase interrupt quality by pinning interrupt and process to core 1
+  setpriority(PRIO_PROCESS, 0, -18);
+  util::set_core_affinity({1});
   std::system("sudo su -c 'echo 1 > /proc/irq/336/smp_affinity_list'");
 
   PubMaster pm({"sensorEvents"});
@@ -220,7 +222,5 @@ int sensor_loop() {
 }
 
 int main(int argc, char *argv[]) {
-  setpriority(PRIO_PROCESS, 0, -18);
-  util::set_core_affinity({1});
-  return sensor_loop();
+ return sensor_loop();
 }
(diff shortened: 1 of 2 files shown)


selfdrive/sensord/sensors_qcom2.cc
```diff
@@ -218,5 +218,11 @@ int sensor_loop() {
 
 int main(int argc, char *argv[]) {
   setpriority(PRIO_PROCESS, 0, -18);
+
+  cpu_set_t mask;
+  CPU_ZERO(&mask);
+  CPU_SET(1, &mask);
+  sched_setaffinity(0, sizeof(mask), &mask);
```
adeebshihadeh, line 225: there's a helper for this in common/util.cc

system/hardware/tici/hardware.py
```diff
@@ -440,6 +440,7 @@ def initialize_hardware(self):
     affine_irq(1, 7)    # msm_drm
     affine_irq(1, 250)  # msm_vidc
     affine_irq(1, 8)    # i2c_geni (sensord)
+    affine_irq(1, 336)    # lsm gpio (sensord)
```
adeebshihadeh, line 443: need to do this after setting up the gpio interface since the interrupt isn't setup until then

adeebshihadeh (commented):
Nice. Can you post before/after plots with all of openpilot running? Let's also try moving to core 2 since there's a bunch of other stuff on core 1. I'd also be curious if it's any better on one of the perf cluster cores (4-7). Make sure to also move the I2C IRQ affinity when testing those.

selfdrive/sensord/sensors_qcom2.cc
gast04: done

system/hardware/tici/hardware.py
gast04: done

gast04:
movint to core2 didnt show any differences:

(accelerometer core1, with openpilot)
hist_pin_interrupts_29k_accel_op_core1

(accelerometer core2, with openpilot)
hist_pin_interrupts_29k_accel_op_core2

however moving to core4 showed an improvement of the timings, all values are about 50us apart compared to the 150us before (tested without openpilot running)
(accelerometer core4, no openpilot)
hist_pin_interrupts_29k_accel_core4

with all of openpilot running the values the values/timings become a bit blur/fuzzy around the previously more exact points
(accelerometer core4, with openpilot)
hist_pin_interrupts_26k_accel_op_core4

comparision, without pinning (core1):
hist_nopin_interrupts_29k_accel_op

we see a way broader histogram, but no three piles, this is probably cause even thought the irq affinity is set, when
moved to another core, core0 still receives interrupts too, these could not be turned off

selfdrive/sensord/sensors_qcom2.cc
```diff
@@ -218,5 +221,6 @@ int sensor_loop() {
 
 int main(int argc, char *argv[]) {
   setpriority(PRIO_PROCESS, 0, -18);
+  util::set_core_affinity({1});
```
adeebshihadeh, line 224: let’s move this and the set priority near the IRQ affinity setting

PR #25869, opened 2022-09-22
loggerd: c++ LoggerState

deanlee:


adeebshihadeh:
Can you rebase this?

adeebshihadeh:
A lot of the line reduction here is just from removing the thread safety, which is unrelated to the rest of the refactor. That should really be done in its own PR.

adeebshihadeh:
Merged anyway since we good tests around this now and it was pretty straightforward. Next time let's break it up though.

deanlee:
>  Next time let's break it up though.

ok!

PR #25877, opened 2022-09-22
athenad: more types

incognitojam:


adeebshihadeh:
Can you split this up into smaller, self-contained PRs? For example, adding type hints is a nice easy to merge PR. Let's also split this by athena/uploader unless there's a reason not to.

selfdrive/athena/athenad.py
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env python3
+from __future__ import annotations
```
adeebshihadeh, line 2: is this necessary?

selfdrive/athena/athenad.py
```diff
@@ -90,15 +103,16 @@ def initialize(upload_queue):
       cloudlog.exception("athena.UploadQueueCache.initialize.exception")
 
   @staticmethod
-  def cache(upload_queue):
+  def cache(upload_queue: Queue[UploadItem]) -> None:
     try:
-      items = [i._asdict() for i in upload_queue.queue if i.id not in cancelled_uploads]
+      queue: List[Optional[UploadItem]] = list(upload_queue.queue)
+      items = [i._asdict() for i in queue if isinstance(i, UploadItem) and (i.id not in cancelled_uploads)]
```
adeebshihadeh, line 109:
```suggestion
      items = [i._asdict() for i in queue if i is not None and (i.id not in cancelled_uploads)]
```

selfdrive/athena/tests/helpers.py
```diff
@@ -1,23 +1,26 @@
...
 from functools import wraps
 from multiprocessing import Process
+from typing import Dict, Optional, Union
 
 from common.timeout import Timeout
 
 
 class MockResponse:
-  def __init__(self, json, status_code):
+  def __init__(self, json: str, status_code: int):
```
adeebshihadeh, line 16: don't think it's necessary to add type hints to tests like this

selfdrive/athena/athenad.py
```diff
@@ -663,15 +681,15 @@ def ws_proxy_send(ws, local_sock, signal_sock, end_event):
...
   last_ping = int(sec_since_boot() * 1e9)
   while not end_event.is_set():
     try:
       opcode, data = ws.recv_data(control_frame=True)
       if opcode in (ABNF.OPCODE_TEXT, ABNF.OPCODE_BINARY):
         if opcode == ABNF.OPCODE_TEXT:
-          data = data.decode("utf-8")
-        recv_queue.put_nowait(data)
+          data = cast(bytes, data).decode("utf-8")
+        recv_queue.put_nowait(cast(str, data))
```
adeebshihadeh, line 692: why is the cast necessary? are we missing the types for an external package?

selfdrive/athena/athenad.py
incognitojam: Yeah, the type hints for Queue don't work correctly without it

selfdrive/athena/athenad.py
```diff
@@ -428,7 +443,7 @@ def setBandwithLimit(upload_speed_kbps, download_speed_kbps):
     return {"success": 0, "error": "failed to set limit", "stdout": e.stdout, "stderr": e.stderr}
 
 
-def startLocalProxy(global_end_event, remote_ws_uri, local_port):
+def startLocalProxy(global_end_event: threading.Event, remote_ws_uri: str, local_port: int) -> Dict[str, Any]:
```
adeebshihadeh, line 446: does this not work?
```suggestion
def startLocalProxy(global_end_event: threading.Event, remote_ws_uri: str, local_port: int) -> Dict[str, int]:
```

selfdrive/athena/athenad.py
```diff
@@ -412,12 +427,12 @@ def cancelUpload(upload_id):
...
 
 @dispatcher.add_method
-def primeActivated(activated):
+def primeActivated(activated: bool):
   return {"success": 1}
 
 
 @dispatcher.add_method
-def setBandwithLimit(upload_speed_kbps, download_speed_kbps):
+def setBandwithLimit(upload_speed_kbps: int, download_speed_kbps: int):
```
adeebshihadeh, line 435: return type?

selfdrive/athena/athenad.py
```diff
@@ -412,12 +427,12 @@ def cancelUpload(upload_id):
 
 
 @dispatcher.add_method
-def primeActivated(activated):
+def primeActivated(activated: bool):
```
adeebshihadeh, line 430: return type?

selfdrive/athena/athenad.py
```diff
@@ -350,7 +365,7 @@ def uploadFileToUrl(fn, url, headers):
 
 
 @dispatcher.add_method
-def uploadFilesToUrls(files_data):
+def uploadFilesToUrls(files_data: List[Dict[str, Any]]):
```
adeebshihadeh, line 368: is Any necessary here?

selfdrive/athena/athenad.py
```diff
@@ -278,7 +293,7 @@ def getMessage(service=None, timeout=1000):
 
 
 @dispatcher.add_method
-def getVersion() -> Dict[str, str]:
+def getVersion():
```
adeebshihadeh, line 296: why remove this?

selfdrive/athena/athenad.py
incognitojam: I guess `Dict[str, Union[str, bool, Dict[str, str]]]` is more exact but much more complicated

selfdrive/athena/athenad.py
incognitojam: I removed all of the dispatcher method types when I realised the hardware ones would be Any, adding them back now though!

selfdrive/athena/athenad.py
adeebshihadeh: all the more reason to do it

selfdrive/athena/athenad.py
```diff
@@ -340,84 +381,89 @@ def do_reboot():
...
 @dispatcher.add_method
-def uploadFileToUrl(fn, url, headers):
-  return uploadFilesToUrls([{
+def uploadFileToUrl(fn: str, url: str, headers: Dict[str, str]) -> UploadFilesToUrlResponse:
+  response: UploadFilesToUrlResponse = uploadFilesToUrls([{
     "fn": fn,
     "url": url,
     "headers": headers,
   }])
+  return response
```
incognitojam, line 394: unfortunately mypy thinks this function is returning `Any` unless it is spelled out like this
```
selfdrive/athena/athenad.py:389: error: Returning Any from function declared to return "Dict[str, Union[int, List[Dict[str, Union[str, bool, int, float, Dict[str, str]]]], List[str]]]"
```

selfdrive/athena/athenad.py
```diff
@@ -340,84 +381,89 @@ def do_reboot():
...
 
 
+UploadFilesToUrlResponse = Dict[str, Union[int, List[UploadItemDict], List[str]]]
+
+
 @dispatcher.add_method
-def uploadFileToUrl(fn, url, headers):
-  return uploadFilesToUrls([{
+def uploadFileToUrl(fn: str, url: str, headers: Dict[str, str]) -> UploadFilesToUrlResponse:
+  response: UploadFilesToUrlResponse = uploadFilesToUrls([{
```
incognitojam, line 389: unfortunately mypy thinks this function is returning `Any` unless it is spelled out like this
```
selfdrive/athena/athenad.py:389: error: Returning Any from function declared to return "Dict[str, Union[int, List[Dict[str, Union[str, bool, int, float, Dict[str, str]]]], List[str]]]"
```

selfdrive/athena/athenad.py
```diff
@@ -340,84 +381,89 @@ def do_reboot():
...
-def cancelUpload(upload_id):
+def cancelUpload(upload_id: Union[str, List[str]]) -> Dict[str, Any]:
   if not isinstance(upload_id, list):
     upload_id = [upload_id]
 
   uploading_ids = {item.id for item in list(upload_queue.queue)}
   cancelled_ids = uploading_ids.intersection(upload_id)
   if len(cancelled_ids) == 0:
-    return 404
+    return {"success": 0, "error": "not found"}
```
incognitojam, line 454: this was the only place where we returned an int error code and not a json dict

selfdrive/athena/athenad.py
```diff
@@ -274,7 +315,7 @@ def getMessage(service=None, timeout=1000):
   if ret is None:
     raise TimeoutError
 
-  return ret.to_dict()
+  return cast(Dict[str, Any], ret.to_dict())
```
incognitojam, line 318: need to check whether or not cast was actually required

incognitojam: I think it's required, would need to update `capnp.lib.capnp._DynamicStructReader` typing

selfdrive/athena/athenad.py
incognitojam: I think this might be due to `jsonrpc.dispatcher` not having typing information, so Python doesn't know for sure that `uploadFilesToUrls` returns the same

PR #25888, opened 2022-09-25
updated to new msgq_poll

deanlee:
msgq_poll can work as expected after https://github.com/commaai/cereal/pull/366


deanlee:
don't know why the test_rotation failed.I'm pretty sure it's not the cereal's problem.

deanlee:
don't know why the test_rotation failed.I'm pretty sure it's not the cereal's problem.

PR #25892, opened 2022-09-26
configure mypy outside of pre-commit environment

incognitojam:
- [x] matching configuration in `mypy.ini`
- [x] add types to pipenv dev dependencies (e.g. `types-PyYAML`, `types-pycurl`...)

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -54,7 +54,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514
+disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514,E1124
 
 
 # Enable the message, report, category or checker with the given id(s). You can
--- a/.pylintrc
+++ b/.pylintrc
@@ -54,7 +54,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514,E1124
+disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514
 
 
 # Enable the message, report, category or checker with the given id(s). You can
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,7 +2,7 @@
 python_version = 3.8
 plugins = numpy.typing.mypy_plugin
 files = body, common, docs, scripts, selfdrive, site_scons, system, tools
-; exclude = pyextra, cereal, panda, opendbc, laika, laika_repo, rednose, rednose_repo, tinygrad, tinygrad_repo
+exclude = pyextra, cereal, panda, opendbc, laika, laika_repo, rednose, rednose_repo, tinygrad, tinygrad_repo
 
 ; third-party packages
 ignore_missing_imports = True
--- a/mypy.ini
+++ b/mypy.ini
@@ -2,7 +2,7 @@
 python_version = 3.8
 plugins = numpy.typing.mypy_plugin
 files = body, common, docs, scripts, selfdrive, site_scons, system, tools
-exclude = pyextra, cereal, panda, opendbc, laika, laika_repo, rednose, rednose_repo, tinygrad, tinygrad_repo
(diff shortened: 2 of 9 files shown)


adeebshihadeh:
Why add the dependencies to the Pipfile? It's not great to have them defined in two places.

incognitojam:
If you configure mypy linting inside an editor it doesn't have the dependencies

adeebshihadeh:
Then let's figure out how to define them once and use in both places.

mypy.ini
```diff
@@ -1,4 +1,14 @@
 [mypy]
 python_version = 3.8
-ignore_missing_imports = True
 plugins = numpy.typing.mypy_plugin
+
+; third-party packages
+ignore_missing_imports = True
```
adeebshihadeh, line 6: is there necessary outside of pre-commit?

mypy.ini
```diff
@@ -1,4 +1,14 @@
 [mypy]
```
adeebshihadeh, line 1: make sure to remove all the configuration in pre-commit and point it to this config file if necessary

incognitojam: we might have to change the pre-commit hook to run the local mypy instead of installing in a new virtualenv

mypy.ini
incognitojam: Actually might not need this after we add the types

incognitojam: nvm, there are lots of packages missing types, will have to do another PR in future to add more types packages

.pylintrc
```diff
@@ -54,7 +54,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514
+disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514,E1224
```
incognitojam, line 57: pylint started complaining about unnecessary keyword args (e.g. `video_link=`  in car docs)

.pylintrc
```diff
@@ -54,7 +54,7 @@ confidence=
 # --enable=similarities". If you want to run only the classes checker, but have
 # no Warning level messages displayed, use"--disable=all --enable=classes
 # --disable=W"
-disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514
+disable=C,R,W0613,W0511,W0212,W0201,W0106,W0603,W0621,W0703,W1201,W1203,E1136,W1514,E1124
```
incognitojam, line 57: pylint started complaining about unnecessary keyword args (e.g. `video_link=` in car docs)

adeebshihadeh: these should just be fixed

mypy.ini
adeebshihadeh: kk, but would like to add those soon. that's free coverage

selfdrive/debug/check_freq.py
```diff
@@ -43,7 +43,7 @@
     if t - prev_print > 1:
       print()
       for name in socket_names:
-        dts = np.diff(rcv_times[name])
+        dts = np.diff(rcv_times[name])  # type: ignore
```
adeebshihadeh, line 46: why is this needed?

selfdrive/debug/check_timings.py
```diff
@@ -20,6 +20,6 @@
         ts[s].append(m.logMonoTime / 1e6)
 
       if len(ts[s]):
-        d = np.diff(ts[s])
+        d = np.diff(ts[s])  # type: ignore
```
adeebshihadeh, line 23: and this one?

selfdrive/debug/check_freq.py
incognitojam: I haven't figured out why, but now mypy doesn't think `Deque[float]` is a `_SupportsArray`, but `MutableSequence` works fine

.pylintrc
incognitojam: #25899

mypy.ini
```diff
@@ -1,4 +1,16 @@
...
 python_version = 3.8
-ignore_missing_imports = True
 plugins = numpy.typing.mypy_plugin
+files = body, common, docs, scripts, selfdrive, site_scons, system, tools
```
adeebshihadeh, line 4: does this work? docs say it's supposed to be actual file paths, but glob patters are supported

adeebshihadeh (approved):
LGTM once the pylint exception addition is reverted and the files thing is checked out

mypy.ini
incognitojam: Seems to work fine!
```
((openpilot) ) batman@z840-openpilot4:~/openpilot$ mypy .
Success: no issues found in 469 source files
((openpilot) ) batman@z840-openpilot4:~/openpilot$ mypy
Success: no issues found in 469 source files
```

PR #25894, opened 2022-09-26
updated to new msgq_poll

deanlee:
msgq_poll can work as expected after https://github.com/commaai/cereal/pull/366


adeebshihadeh:
Closing the openpilot side since this has been in a draft for a while. We can re-open once the cereal side is ready.

PR #25897, opened 2022-09-27
IsoTpParallelQuery: extend timeout for each ISO-TP frame

sshane:
Previous behavior:
- After we receive all the iso-tp frames for a response, we then extend the timeout for that address in IsoTpParallelQuery. This means that we would extend the timeout for an address once with this query (once after the full tester present response comes back):
    ```
    Request(
      [StdQueries.SHORT_TESTER_PRESENT_REQUEST, TOYOTA_VERSION_REQUEST],
      [StdQueries.SHORT_TESTER_PRESENT_RESPONSE, TOYOTA_VERSION_RESPONSE],
      bus=0,
    ),
    ```

New behavior:
- As we receive iso-tp frames for a response, we extend the timeout for that address. This means that responses that need many frames will not time out after a total max timeout, it will continually be extended as we build the response.

Instead of timing out and re-querying ECUs that just take longer than 100 ms to complete (lots of frames, or high interval time), we fully wait for the response to finish or truly time out before attempting the next request, reducing the number of invalid consecutive frame instances.

This now allows us to receive longer FW versions like some Hyundai transmission or VMCU FW versions.


selfdrive/car/isotp_parallel_query.py
```diff
@@ -107,21 +106,23 @@ def get_data(self, timeout, total_timeout=60.):
...
+        if updated:
+          response_timeouts[tx_addr] = time.monotonic() + timeout
+
         if not dat:
           continue
 
         counter = request_counter[tx_addr]
         expected_response = self.response[counter]
-        response_valid = dat[:len(expected_response)] == expected_response
+        response_valid = expected_response is None or dat[:len(expected_response)] == expected_response
```
sshane, line 123:
```suggestion
        response_valid = dat[:len(expected_response)] == expected_response
```

gregjhogan (approved):

PR #25900, opened 2022-09-27
ci: don't trigger on pull request sync event

incognitojam:
By default the `pull_request` trigger causes the workflows to run whenever a PR is `opened`, `reopened` or `synchronized` (a commit is pushed), but since we have the `push` trigger it causes the workflows to run twice for every push to a branch with an active PR.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

### DIFF
--- a/.github/workflows/selfdrive_tests.yaml
+++ b/.github/workflows/selfdrive_tests.yaml
@@ -4,6 +4,7 @@ on:
     branches-ignore:
       - 'testing-closet*'
   pull_request:
+    types: [opened, reopened]
 
 env:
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
This is intentional. One is for your branch and the other is your branch merged into master.

adeebshihadeh:
Re-opening since we're now saturating our concurrency limits pretty often.

incognitojam:
turns out that the process replay and car docs diff comment only run on pull request checks...

PR #25902, opened 2022-09-27
networking: add unmetered cellular toggle

incognitojam:
- [x] add toggle
- [x] wifi manager: write metered property to connection settings

```sh
$ cat /data/params/d/GsmUnmetered && echo
0
$ nmcli -f connection.metered connection show lte
connection.metered:                     unknown

# hit toggle (force unmetered)
$ cat /data/params/d/GsmUnmetered && echo
1
$ nmcli -f connection.metered connection show lte
connection.metered:                     no
```

Screenshot from 2022-09-29 19-31-05

Related #23731


adeebshihadeh:
>   * [ ]  reset metered and APN when prime is detected

This shouldn't be necessary since each prime type has its own NetworkManager connection. The `lte` connection is a fallback connection for non-prime SIMs, and changes should only be made for that connection.

incognitojam:
Screenshot from 2022-09-29 19-31-05

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -169,12 +168,21 @@ AdvancedNetworking::AdvancedNetworking(QWidget* parent, WifiManager* wifi): QWid
...
       params.put("GsmApn", apn.toStdString());
     }
-    wifi->updateGsmSettings(roamingEnabled, apn);
+    wifi->updateGsmSettings(params.getBool("GsmRoaming"), apn, params.getBool("GsmUnmetered"));
   });
   list->addItem(editApnButton);
 
+  // Metered toggle
+  const bool unmeteredOverride = params.getBool("GsmUnmetered");
+  ToggleControl *meteredToggle = new ToggleControl(tr("Force Unmetered Cellular"), "", "", !unmeteredOverride);
```
adeebshihadeh, line 177: this is slightly confusing; I would simplify the toggle to just say "cellular metered" (but keep using no and unknown behind the hood)

adeebshihadeh: let's also add a description

selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -345,7 +345,7 @@ NetworkType WifiManager::currentNetworkType() {
   return NetworkType::NONE;
 }
 
-void WifiManager::updateGsmSettings(bool roaming, QString apn) {
+void WifiManager::updateGsmSettings(bool roaming, QString apn, bool unmeteredOverride) {
```
adeebshihadeh, line 348: unmeteredOverride is confusing too. in the theme of simplification, let's also internally just pass around a "metered state".

adeebshihadeh (approved):

selfdrive/ui/qt/offroad/networking.cc
incognitojam: how is `Prevent large data uploads when on a metered connection`?

incognitojam:
Screenshot from 2022-09-30 16-01-03

incognitojam:
An issue I noticed is when you tap on the toggle to read the description, it appears offscreen, so you can't read it unless you try to scroll down again

selfdrive/ui/qt/offroad/networking.cc
adeebshihadeh: lgtm. let's make sure it also defaults to metered/unknown

sshane: does this only apply to cellular? sounds like it might apply to mobile hotspots

incognitojam: oh, what about `Prevent large data uploads on cellular connection`?

incognitojam:
Screenshot from 2022-09-30 16-11-12

incognitojam:
Confirmed after installing the fork the "cellular metered" defaults to on

PR #25910, opened 2022-09-28
params: catch signals with sigtimedwait in blocking read 

deanlee:
catch signals with std::signal and restore it to the origin one before return is not safe.
the previous handler cannot receive the signals during block read, and If two blocking reads occur at the same time, it may set prev_handler_sigint to params_sig_handler, which will cause the original handler to be lost and never be called.

- [x] catch signals with sigtimedwiat
- [ ] test case


deanlee:
need a test case for this

PR #25912, opened 2022-09-28
params: safe and efficient async writing parameters

deanlee:
The original version of put_nonblocking may create many threads(which are not joined) waiting on the FileLock . 
If the process exits while the thread is runing, these pending write operations will fail with exception.


common/params_pyx.pyx
```diff
@@ -105,7 +117,7 @@ cdef class Params:
     return self.p.allKeys()
 
 def put_nonblocking(key, val, d=""):
-  threading.Thread(target=lambda: Params(d).put(key, val)).start()
+  Params(d).put_non_blocking(key, val);
```
incognitojam, line 120: should semicolon be here?

deanlee: lol,my bad.

adeebshihadeh:
Can you write a good test for this?

common/params_pyx.pyx
```diff
@@ -105,7 +117,7 @@ cdef class Params:
     return self.p.allKeys()
 
 def put_nonblocking(key, val, d=""):
-  threading.Thread(target=lambda: Params(d).put(key, val)).start()
+  Params(d).put_non_blocking(key, val)
 
 def put_bool_nonblocking(key, bool val, d=""):
-  threading.Thread(target=lambda: Params(d).put_bool(key, val)).start()
+  Params(d).put_bool_non_blocking(key, val)
```
adeebshihadeh, line 123: shall we also remove these helpers?

common/params.cc
```diff
@@ -326,3 +328,29 @@ void Params::clearAll(ParamKeyType key_type) {
...
+      Params(path).put(key, value);
+    }
+  }
+
+  std::future<void> f;
+  SafeQueue<std::tuple<std::string, std::string, std::string>> q;
+};
+
+void Params::putNonBlocking(const std::string &key, const std::string &val) {
+  static AsyncWriter async_writer;
```
adeebshihadeh, line 354: why static instead of part of the class?

adeebshihadeh (commented):
This would be great to get in. Can you also write a test case for this that actually checks the async functionality?

common/params_pyx.pyx
deanlee: test_power_monitor.py still needs this to mock it to a block writing.

common/params.cc
deanlee: There may be many instances of Params. Only one asycwriter is needed per process.

common/params_pyx.pyx
adeebshihadeh: fixed that in https://github.com/commaai/openpilot/pull/29788

adeebshihadeh:
@deanlee seems there's still one test failure
```
=================================== FAILURES ===================================
_________________ TestPowerMonitoring.test_pandaState_present __________________

self = <selfdrive.thermald.tests.test_power_monitoring.TestPowerMonitoring testMethod=test_pandaState_present>

    def test_pandaState_present(self):
      pm = PowerMonitoring()
      for _ in range(10):
        pm.calculate(None, None)
      self.assertEqual(pm.get_power_used(), 0)
>     self.assertEqual(pm.get_car_battery_capacity(), (CAR_BATTERY_CAPACITY_uWh / 10))
E     AssertionError: 30000000 != 3000000.0

selfdrive/thermald/tests/test_power_monitoring.py:38: AssertionError
```

deanlee:
checking

deanlee:
fixed. it's a bug.the `std::future<T>::~future`  does not always block waiting for the thread to finish when running in python/pytest

fix: do not reply on ~future(), explicit call `wait` in dctor.

deanlee:
improved test case to catch this issue.

adeebshihadeh:
Nice cleanup!

adeebshihadeh:
Had a seemingly related failure on master: https://github.com/commaai/openpilot/actions/runs/6102074880/job/16559742075

deanlee:
I'll check again,  I think this issue may not be related to asyc write

deanlee:
@adeebshihadeh : think we still need to mock the nonblocking put to blocking put

the writing thread of `static AsyncWriter` will continue to run if the queue is not empty after the params is destroyed/re-create in `TestPowerMonitoring.setUp(self)`. 
this  causes  `test_pandaState_present` to fail if 'CarBatteryCapacity' is written by asyncWriter before init car_battery_capacity_uWh from `self.params.get("CarBatteryCapacity")`

deanlee:
reopend https://github.com/commaai/openpilot/pull/29808
pass param prefix  to asyWriter to fix test issues https://github.com/commaai/openpilot/pull/29808/commits/3fb13e429093507458513512d50628b454173a83

PR #25918, opened 2022-09-28
fingerprinting: send tester present to functional addresses first

sshane:
~~We should switch back to querying the functional address and just select the VIN more carefully (not exiting on the first response).~~ New EVs respond very flakily to the UDS query on functional address. Dealer software queries 0x7e2 directly.

This fixes consecutive frame errors for a user's Elantra 2021. Still looking into it, but it appears if this address is not queried in some way, messages can be dropped from the OBD-II bus. Even sending tester present to this address fixes VIN and FW queries skipping frames (which we might want to do instead? yes).

This PR:
- Adds a IsoTpParallelQuery before all other iso-tp queries where we send a tester present request to the functional addresses (0x7DF and 0x18DB33F1) with a timeout of 0.1s to process all responses (processing ensures these responses don't flow over to the VIN query which is done right after).
- Adds a sleep of 0.2s before the use of the sendcan PubSocket to allow boardd to connect (TODO: can we use 0.1s? what is reliable?)


sshane:
What the Kia/Hyundai J2534 sotware does:

Regardless of car, it always sends a UDS VIN query to 0x7e2 (VMCU), then if it gets no response it sends an OBD VIN query on functional address 0x7DF.

sshane:
Confirmed that this also fixes skipped iso-tp frames on the Sonata and EV6

selfdrive/car/car_helpers.py
```diff
@@ -97,6 +98,12 @@ def fingerprint(logcan, sendcan):
       car_fw = list(cached_params.carFw)
     else:
       cloudlog.warning("Getting VIN & FW versions")
+      # Give some time to let sendcan connect to boardd
```
sshane, line 101:
```suggestion
      # Allow some time to let sendcan connect to boardd
```

sshane:
Update: We may be able to query 0x7DF for the VIN again, but that would rely on PR: https://github.com/commaai/openpilot/pull/25930

sshane:
Fixed with https://github.com/commaai/openpilot/pull/25933

PR #25919, opened 2022-09-29
Gps test Setup, PoC

gast04:
Initial gps testing commit


tools/gpstest/src/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 OSQZSS
```
adeebshihadeh, line 3: gitignore this repo and add a script to clone it and do any other relevant setup

tools/gpstest/SConstruct
```diff
@@ -0,0 +1,8 @@
+env = Environment(
+        CFLAGS = ["-O2"],
+        CC='gcc',
+    )
+
+src = ["src/limegps.c", "src/gpssim.c"]
+libs = ["LimeSuite", "pthread", "m"]
+env.Program("bin/LimeGPS", src, LIBS=libs, LIBPATH=['lib'])
```
adeebshihadeh, line 8: can't we just run `make`?

tools/gpstest/src/LICENSE
adeebshihadeh: here's an example of something like this: https://github.com/commaai/openpilot/blob/master/selfdrive/debug/profiling/palanteer/setup.sh

tools/gpstest/SConstruct
gast04: sure, but we use SConscript everywhere else^^

adeebshihadeh: that's only for our stuff. for example, what happens if gpssim.c gets renamed to gps_sim.c? `make` will still work, but this scons stuff will have to be updated

tools/gpstest/gpstest.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+LimeGPS_BIN=LimeGPS/LimeGPS
+if test -f "$LimeGPS_BIN"; then
+    LD_PRELOAD=lib/libLimeSuite.so $LimeGPS_BIN $@
+else
+    echo "LimeGPS binary not found, run setup.sh first"
```
adeebshihadeh, line 7: always indent with two spaces

tools/gpstest/LimeUtil.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/bash
+LD_PRELOAD=lib/libLimeSuite.so bin/LimeUtil $@
```
adeebshihadeh, line 2: why are these checked in? we generally don't check in binaries unless there's a good reason. do we just need to clone and build a separate project?

gast04: the LimeSuite repo is bigger, so I thought makes sense to prebuilt it once, but changed

PR #25920, opened 2022-09-29
locationd: Fix GPS sensor times with fixed offsets

haraschax:


HaraldSchafer:
Also remove all the clocks code if its unused?

PR #25925, opened 2022-09-29
panda: check connected in mutex

deanlee:
make sure to return immediately after entering mutex if other thread has set connected to false in lock


adeebshihadeh:
Is this still relevant?

deanlee:
closed

PR #25930, opened 2022-09-29
IsoTpParallelQuery: process all functional responses

sshane:
On some cars (namely Hyundai), sending a VIN query to the functional 0x7DF address leads to multiple ECUs sending back "valid" responses (any response that can be decoded with UTF-8). And the way it's set up now, IsoTpParallelQuery only uses one IsoTpMessage object for the tx_addr of 0x7DF, so we will stop processing responses after the first. This means the returned VIN can vary between a few different values randomly.

Changes:
- Returns all responses from a functional query, with correct rx_addrs, and physical tx_addrs (should actually be 0x7DF, but the user of `IsoTpParallelQuery` knows they've done a functional addr query)
- Instead of having special logic in `IsoTpParallelQuery`'s `rx()` function for functional vs. physical querying, we just create a `IsoTpMessage` object for each possible tx/rx_addr and handle them like physical address querying. This way, we also handle timing out properly (considering the full range of addresses that can possibly respond)
- When `functional_addr` param is set to True, we assert you can only add the two functional addresses.


selfdrive/car/isotp_parallel_query.py
```diff
@@ -16,17 +16,20 @@ def __init__(self, sendcan, logcan, bus, addrs, request, response, response_offs
...
-      else:
-        self.real_addrs.append((a, None))
+    if functional_addr:
+      assert all([a in FUNCTIONAL_ADDRS for a in addrs]), "Non-functional addresses in addrs"
+      real_addrs = []
+      if 0x7DF in addrs:
+        real_addrs.extend([(0x7e0 + i, None) for i in range(8)])
+      if 0x18DB33F1 in addrs:
+        # TODO: is this correct?
+        real_addrs.extend([(0x18da00f1 + (i << 8), None) for i in range(256)])
```
sshane, line 28: @gregjhogan do you have a preference how we handle processing all responses when functional querying? I personally like making the rest of the library agnostic to whether we've functional or physical queried, so we can re-use the same logic for physical querying.

Also, are the ranges correct? It might be nice to put them in uds.py, either as a global variable, or make `FUNCTIONAL_ADDRS` a dict

selfdrive/car/isotp_parallel_query.py
```diff
@@ -16,17 +16,20 @@ def __init__(self, sendcan, logcan, bus, addrs, request, response, response_offs
...
-        self.real_addrs.append(a)
-      else:
-        self.real_addrs.append((a, None))
+    if functional_addr:
+      assert all([a in FUNCTIONAL_ADDRS for a in addrs]), "Non-functional addresses in addrs"
+      real_addrs = []
+      if 0x7DF in addrs:
+        real_addrs.extend([(0x7e0 + i, None) for i in range(8)])
+      if 0x18DB33F1 in addrs:
+        # TODO: is this correct?
```
sshane, line 27:
```suggestion
```

selfdrive/car/isotp_parallel_query.py
```diff
@@ -16,17 +16,19 @@ def __init__(self, sendcan, logcan, bus, addrs, request, response, response_offs
...
-        self.real_addrs.append((a, None))
+    if functional_addr:
+      assert all([a in FUNCTIONAL_ADDRS for a in addrs]), "Non-functional addresses in addrs"
+      real_addrs = []
+      if 0x7DF in addrs:
+        real_addrs.extend([(0x7e0 + i, None) for i in range(8)])
+      if 0x18DB33F1 in addrs:
+        real_addrs.extend([(0x18da00f1 + (i << 8), None) for i in range(256)])
+    else:
+      real_addrs = [a if isinstance(a, tuple) else (a, None) for a in addrs]
```
sshane, line 29:
```suggestion
      if 0x7DF in addrs:
        real_addrs.extend([(0x7E0 + i, None) for i in range(8)])
      if 0x18DB33F1 in addrs:
        real_addrs.extend([(0x18DA00F1 + (i << 8), None) for i in range(256)])
    else:
      real_addrs = [a if isinstance(a, tuple) else (a, None) for a in addrs]
```

selfdrive/car/isotp_parallel_query.py
gregjhogan: Looks good to me.  I think you correctly identified the standard addresses, and we can just add more if we find any additional non-standard addrs we need in the future.

PR #25932, opened 2022-09-30
ecu_addrs: use IsoTpParallelQuery

sshane:
still a draft


sshane:
hmm, this is already twice as slow even without querying each request individually (0.33 on master to 0.7 seconds for the Sonata)

PR #25938, opened 2022-09-30
Replay: handle deprecated event

deanlee:


adeebshihadeh:
any particular reason to migrate the sensor events? we only did it for panda to keep the UI working

deanlee:
ok i closed

PR #25939, opened 2022-09-30
IsoTpParallelQuery: process all functional responses

sshane:
Re-do of https://github.com/commaai/openpilot/pull/25930 as that PR queried physical addresses, not the functional address. This PR correctly sends the first request (really the first/single frame) to the functional addresses, then hands off communication to each individual server/physical address, as you're supposed to do.

This PR aims to query in the following fashion:
- We send first query to 0x7DF (or the 11-bit functional address).
- We listen for first frame responses on range 0x7E0 to 0x7E7
- We then send flow control continue to any physical address that respond (0x7E0 for example). No more messages are sent to 0x7DF


selfdrive/car/isotp_parallel_query.py
```diff
@@ -79,22 +78,19 @@ def get_data(self, timeout, total_timeout=60.):
...
-
-      msg = IsoTpMessage(can_client, timeout=0, max_len=max_len, debug=self.debug)
-      msg.send(self.request[0])
-
-      msgs[tx_addr] = msg
+      msgs[tx_addr] = self._create_isotp_msg(*tx_addr, rx_addr)
       request_counter[tx_addr] = 0
       request_done[tx_addr] = False
 
+    # Send first request to functional addrs, subsequent responses are handled by physical addrs
```
sshane, line 85:
```suggestion
    # Send first request to functional addrs, subsequent responses are handled on physical addrs
```

PR #25946, opened 2022-10-01
c++ cabana:  Initial version 

deanlee:
It's a very draft version at the moment. The layout is basically the same as the web version.

updated:

cabana


adeebshihadeh:
Looks great! I'm happy to merge a draft and do this very incrementally like we did with the replay tool if you want.

deanlee:
@adeebshihadeh sounds good. I'll set it to "ready for review" after I think it can be merged as an initial version

deanlee:
usage:
`cabana <route>`

run  demo with qcamera:
`cabana --demo  --qcam`

deanlee:
usage:
`cabana <route>`

run  demo with qcamera:
`cabana --demo  --qcam`

tools/cabana/messageswidget.cc
```diff
@@ -0,0 +1,92 @@
...
+  int i = 0;
+  const QString filter_str = filter->text();
+  for (auto &[address, list] : parser->items) {
+    if (list.empty()) continue;
+
+    QString name = tr("untitled");
+    if (auto msg = parser->getMsg(address)) {
+      name = msg->name.c_str();
+    }
+    if (!filter_str.isEmpty() && !name.contains(filter_str)) continue;
```
sshane, line 81:
```suggestion
    if (!filter_str.isEmpty() && !name.toLower().contains(filter_str.toLower())) continue;
```

tools/cabana/messageswidget.cc
```diff
@@ -0,0 +1,92 @@
...
+    if (!item) {
+      item = new QTableWidgetItem();
+      table_widget->setItem(row, col, item);
+    }
+    return item;
+  };
+
+  table_widget->setRowCount(parser->items.size());
+  int i = 0;
+  const QString filter_str = filter->text();
```
sshane, line 73:
```suggestion
  const QString filter_str = filter->text().toLower();
```

sshane:
Great work! Just a note; it seems like it doesn't support showing the same signal multiple times for the buses. `ACC_CONTROL` for example gets one entry and rapidly switches between the buses (128 and 2) and data.

deanlee:
> `` for example gets one entry and rapidly switches between the buses (128 and 2) and data.

yes, I'll fix it soon.

deanlee:
> `` for example gets one entry and rapidly switches between the buses (128 and 2) and data.
fixed.

@adeebshihadeh : I think it can be merged as an initial version now.

adeebshihadeh:
Thanks @deanlee; looks like a great first step!

PR #25951, opened 2022-10-03
HKG: Add FW and Remove from dashcamOnly for 2019 Elantra GT i30

sunnyhaibin:
Add firmware for the 2019 Hyundai Elantra GT i30. Remove the platform from the `dashcamOnly` flag.

Route ID:
- stock longitudinal: `734ef96182ddf940|2022-10-02--16-41-44`
- openpilot longitudinal:

Thanks to community Elantra GT i30 owner Breadjarmang.


adeebshihadeh:
@sshane [this comment](https://github.com/sunnyhaibin/openpilot-1/blob/3f11edbde408cfec6ac4aea8a81fce904c07c283/selfdrive/car/hyundai/values.py#L112) says it's the same as the Elantra. any reason why it has a different openpilot platform?

sshane:
Not that I'm aware of, this was before my time. I'll have a look at merging them soon

selfdrive/car/hyundai/values.py
```diff
@@ -34,7 +34,6 @@ def __init__(self, CP):
 
     # To determine the limit for your car, find the maximum value that the stock LKAS will request.
     # If the max stock LKAS request is <384, add your car to this list.
-    elif CP.carFingerprint in (CAR.GENESIS_G80, CAR.GENESIS_G90, CAR.ELANTRA, CAR.HYUNDAI_GENESIS, CAR.ELANTRA_GT_I30, CAR.IONIQ,
                                CAR.IONIQ_EV_LTD, CAR.SANTA_FE_PHEV_2022, CAR.SONATA_LF, CAR.KIA_FORTE, CAR.KIA_NIRO_PHEV,
```
sshane, line 37:
```suggestion
    elif CP.carFingerprint in (CAR.GENESIS_G80, CAR.GENESIS_G90, CAR.ELANTRA, CAR.HYUNDAI_GENESIS, CAR.ELANTRA_GT_I30, CAR.IONIQ,
                               CAR.IONIQ_EV_LTD, CAR.SANTA_FE_PHEV_2022, CAR.SONATA_LF, CAR.KIA_FORTE, CAR.KIA_NIRO_PHEV,
```

selfdrive/car/hyundai/values.py
```diff
@@ -34,7 +34,6 @@ def __init__(self, CP):
 
     # To determine the limit for your car, find the maximum value that the stock LKAS will request.
     # If the max stock LKAS request is <384, add your car to this list.
```
sshane, line 36:
```suggestion
    # If the max stock LKAS request is <384, add your car to this list.
    elif CP.carFingerprint in (CAR.GENESIS_G80, CAR.GENESIS_G90, CAR.ELANTRA, CAR.HYUNDAI_GENESIS, CAR.ELANTRA_GT_I30, CAR.IONIQ,
```

sshane:
Do you know what harness this user's car uses? I don't see it in any of our old databases

sshane:
This is actually the same platform as the Ceed with the same harness. Should we tend to combine these cases into single platforms (with two CarInfos) @adeebshihadeh?

adeebshihadeh:
> This is actually the same platform as the Ceed with the same harness. Should we tend to combine these cases into single platforms (with two CarInfos) @adeebshihadeh?
> 
> https://en.wikipedia.org/wiki/Hyundai_i30

Yes, if they're the same as far as openpilot is concerned.

sshane:
So these two marketed cars (i30 and Elantra GT) have SCC with stop and go, whereas the Elantra 2019 does not. And the transmission signal is slightly different than the Ceed, so we can't combine it there. Going to keep it separate as it's relatively different enough

sshane:
Let's test longitudinal, this car wasn't in legacy safety mode cars before

sunnyhaibin:
> Let's test longitudinal first too, this car wasn't in legacy safety mode cars before

@sshane Requested the user to test and will report back when tested.

sshane:
@sunnyhaibin any updates?

sshane:
Feel free to remove from legacy safety mode if you confirm longitudinal works

PR #25962, opened 2022-10-04
GM: don't consider soft disabling as ACC fault

sshane:
***In dashcam mode,*** if stock ACC is braking for a car and the user presses the cancel button (taps or holds), ACC stays engaged for up to 2.8 seconds (observed) and a warning sound and lights are played (FCW severity) while the car tries to bring you to a stop before it disengages.

The entire time, the cruise state is "faulted," however we usually take faulted to mean we caused an issue with the car. In this instance though, faulted is a state where the car is soft disabling and continuing longitudinal actuation as normal.

### DIFF
--- a/selfdrive/car/gm/carstate.py
+++ b/selfdrive/car/gm/carstate.py
@@ -17,6 +17,7 @@ class CarState(CarStateBase):
     self.shifter_values = can_define.dv["ECMPRDNL2"]["PRNDL2"]
     self.lka_steering_cmd_counter = 0
     self.buttons_counter = 0
+    self.frames_acc_faulted = 0
 
   def update(self, pt_cp, cam_cp, loopback_cp):
     ret = car.CarState.new_message()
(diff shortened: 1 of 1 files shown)


selfdrive/car/gm/carstate.py
```diff
@@ -77,7 +78,15 @@ def update(self, pt_cp, cam_cp, loopback_cp):
...
-    ret.accFaulted = pt_cp.vl["AcceleratorPedal2"]["CruiseState"] == AccState.FAULTED
+
+    if pt_cp.vl["AcceleratorPedal2"]["CruiseState"] == AccState.FAULTED:
+      self.frames_acc_faulted += 1
+    else:
+      # GM uses the FAULTED state in AcceleratorPedal2->CruiseState for genuine ACC faults
+      # as well as soft disables with full longitudinal actuation (tries to brake to a stop).
+      # Set accFaulted if FAULTED less than 10 frames (immediate disable)
+      ret.accFaulted = 0 < self.frames_acc_faulted <= 10
+      self.frames_acc_faulted = 0
```
sshane, line 89: Need to check from data where this frame threshold is

selfdrive/car/gm/carstate.py
```diff
@@ -77,7 +78,15 @@ def update(self, pt_cp, cam_cp, loopback_cp):
...
     ret.cruiseState.available = pt_cp.vl["ECMEngineStatus"]["CruiseMainOn"] != 0
     ret.espDisabled = pt_cp.vl["ESPStatus"]["TractionControlOn"] != 1
-    ret.accFaulted = pt_cp.vl["AcceleratorPedal2"]["CruiseState"] == AccState.FAULTED
+
+    if pt_cp.vl["AcceleratorPedal2"]["CruiseState"] == AccState.FAULTED:
+      self.acc_faulted_frames += 1
+    else:
+      # GM uses the FAULTED state in AcceleratorPedal2->CruiseState for genuine ACC faults
+      # as well as soft disables with full longitudinal actuation (tries to brake to a stop).
+      # Set accFaulted if FAULTED less than 10 frames (immediate disable)
```
sshane, line 87:
```suggestion
      # Set accFaulted if state is FAULTED less than 10 frames (immediate disable)
```

sshane:
If the data checks out, does this sound like an appropriate fix to you @adeebshihadeh?

The alert from the car is another issue I think, and that involves finding a different cancel command (or looking into what threshold user brake disengages ACC, is it really adjusted by how much the car is currently braking?)

selfdrive/car/gm/carstate.py
sshane: Also should consider accFaulted if in this state for longer than max soft disable time

sshane:
These aren't considered in the MTBF report

PR #25965, opened 2022-10-04
canbana: complete basic functions

deanlee:
all functions are basically available after this PR, Except writing DBC to file.


adeebshihadeh:
For stuff like show and hide, let's try to use icons when we can. Does Qt have any built in?

deanlee:
This PR has been completed. reday for review&merge.

It is basically a usable version with many optimizations, the CPU usage is less than 40% on my laptop (without CUDA)

llll

PR #25966, opened 2022-10-04
Ford: handle VIN

sshane:
It says length will be 24 bytes and appends 7 null bytes. Only happens with functional VIN query


incognitojam (approved):

selfdrive/car/vin.py
```diff
@@ -27,6 +27,9 @@ def get_vin(logcan, sendcan, bus, timeout=0.1, retry=5, debug=False):
         for addr in valid_vin_addrs:
           vin = results.get((addr, None))
           if vin is not None:
+            # Ford pads with null bytes
+            vin = vin.replace(b'\x00', b'')
```
adeebshihadeh, line 31: this should handle the ford case more explicitly, e.g. check that the length is 24 before doing this

PR #25975, opened 2022-10-05
Car docs: test no duplicate years

sshane:
Per `(make, model)` key (okay to duplicate years in a platform if it's a differently marketed car)


sshane:
@adeebshihadeh this what you wanted?

adeebshihadeh:
Yes

PR #25916, opened 2022-09-28
updated: configure branch upstream

adeebshihadeh:

PR #25976, opened 2022-10-05
updater: update git config for new target branch

incognitojam:
- [x] don't break not-single-branch setups

Related:
- https://github.com/commaai/openpilot/pull/25916


adeebshihadeh:
Hmm, I'd rather revert that for now than touch the git config.

incognitojam:
When is the updater running where it isn't single branch?

PR #25980, opened 2022-10-06
regen: add arg for migrating sensorEvents with old timestamps

nuwandavek:
@gast04 Was there a reason not to add the same log time while migrating events? I cannot reproduce `locationd` on old data if the timings are not right.


adeebshihadeh:
For the regen use case, you want a new monotime. This is only correct for your use case.

nuwandavek:
> For the regen use case, you want a new monotime. This is only correct for your use case.

better?

PR #25982, opened 2022-10-06
cabana: use QFormLayout

deanlee:


deanlee:
I just remembered there is this class. :hear_no_evil: it makes the code cleaner

PR #25984, opened 2022-10-06
cabana: Fix the incorrect Y axis

deanlee:
sorry, I forgot to set the Y axis at the first loading,it's always 0-1.0 if don't drag the selection range


adeebshihadeh:
This is looking really good @deanlee!

deanlee:
I'm glad you liked it :D

PR #25986, opened 2022-10-06
joystickd: add controller mapping

jp-solutionz:
Current config maps steering to right trigger (ABS_RZ) when using a default xinput controller: https://inputs.readthedocs.io/en/latest/user/hardwaresupport.html

This results in default full left steering angle (1) requiring right trigger to return to centre (0) or right (-1).

It appears the intended mapping for steering is right thumbstick (ABS_RX).

Cancel button is also non-existent on default xinput controller. May be X button (BTN_NORTH) or Right Trigger (ABS_RZ).

Tested on Xbox One Controller via USB Cable, Logitech F710 and GameSir T4 Pro.

### DIFF
--- a/tools/joystick/joystickd.py
+++ b/tools/joystick/joystickd.py
@@ -38,13 +38,20 @@ class Keyboard:
 
 
 class Joystick:
-  def __init__(self):
+  def __init__(self, gamepad=False):
     # TODO: find a way to get this from API, perhaps "inputs" doesn't support it
-    self.min_axis_value = {'ABS_Y': 0., 'ABS_RX': 0.}
-    self.max_axis_value = {'ABS_Y': 255., 'ABS_RX': 255.}
-    self.cancel_button = 'BTN_NORTH'  # (BTN_NORTH=X, ABS_RZ=Right Trigger)
-    self.axes_values = {'ABS_Y': 0., 'ABS_RX': 0.}  # gb, steer
-    self.axes_order = ['ABS_Y', 'ABS_RX']
+    if gamepad:
+      self.cancel_button = 'BTN_NORTH'  # (BTN_NORTH=X, ABS_RZ=Right Trigger)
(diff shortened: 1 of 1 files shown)


incognitojam:
I think the current behaviour is optimised for joysticks like this https://www.logitechg.com/en-us/products/space/extreme-3d-pro-joystick.963290-0403.html

jp-solutionz:
> 

Joystick [README](https://github.com/commaai/openpilot/tree/master/tools/joystick) mentions "many common gamepads and joysticks" without referencing any specific model(s), however it links to [inputs Doc](https://inputs.readthedocs.io/en/latest/user/hardwaresupport.html) which lists the following Linux support so it's implied they are the default target;
- Xbox 360 Controller via USB cable
- Xbox One Controller via USB cable
- PS4 Controller via USB cable
- PS3 Controller via USB cable
- Pi-Hut SNES Style USB Gamepad
- Wii controller

The default behaviour of full left steer when using the above is not "ideal" (safe) particularly as the issue doesn't present until OP is active (for most vehicles = moving) and only after trigger is initially pressed.

Ideally joystickd should include an initialisation / mapping feature.

In the meantime mapping should default to standard xinput gamepad, or the docs updated to specify default compatible joystick(s).

incognitojam:
A potential solution might be to add an argument (`--gamepad`) to joystickd to use the new configuraiton you've created for gamepads, or to use the old configuration for joysticks. We wouldn't want to remove the joystick config as many people including comma use them for testing vehicles too.

incognitojam:
@jp-solutionz want to test this with your controller(s)?

jp-solutionz:
> 

Nice work. --gamepad works as expected and without it joystick mapping appears unaltered.

jp-solutionz (commented):
Tested with xinput --gamepad works as expected

PR #25989, opened 2022-10-06
Fingerprinting: log if using cache

sshane:
So we can know if ecu_responses is expected to be empty, and to avoid looking into these routes further (going to have no error logs from iso-tp library)

The `fingerprinted` event already has:
- source (if fixed, fw, or can)
- fuzzy
- fw_count (can know if we're missing versions from something like skipped frames, not if query returns different version however)
- ecu_responses (only set when not cached)
- vin_rx_addr (only set when not cached)

### DIFF
--- a/selfdrive/car/car_helpers.py
+++ b/selfdrive/car/car_helpers.py
@@ -92,11 +92,11 @@ def fingerprint(logcan, sendcan):
         cached_params = None
 
     if cached_params is not None and len(cached_params.carFw) > 0 and cached_params.carVin is not VIN_UNKNOWN:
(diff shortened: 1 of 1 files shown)


selfdrive/car/car_helpers.py
```diff
@@ -95,16 +95,19 @@ def fingerprint(logcan, sendcan):
       cloudlog.warning("Using cached CarParams")
```
sshane, line 95: Could have made these errors, but that would mean we'd have to figure out which log messages correspond to each other. Easy from the logs, not so sure about in Kibana

PR #25992, opened 2022-10-06
updated: initial test setup

incognitojam:


release/build_release.sh
```diff
@@ -100,6 +100,7 @@ TEST_FILES="tools/"
 cd $SOURCE_DIR
 cp -pR -n --parents $TEST_FILES $BUILD_DIR/
 cd $BUILD_DIR
+selfdrive/test/test_updated.py
```
adeebshihadeh, line 103: add to the jenkinsfile instead

incognitojam:
@adeebshihadeh `/tmp` runs out of space :confused: can we make it bigger? or just pick somewhere else in `/data`
https://github.com/commaai/agnos-builder/blob/972feac8d03671a2344841cfe8c11b7e03232b4c/userspace/files/fstab#L7

adeebshihadeh:
Yeah, just use something on /data. tempfile.TemporaryDirectory can take a prefix

incognitojam:
It seems like it fails in CI because the repos/modules only initially get cloned with a depth of 5, and the updater is trying to fetch objects which are older. Guessing it works on PC because it's not a shallow repo... Might want to fix in updater instead to not fetch unnecessary objects.

PR #25997, opened 2022-10-07
Added updateZoom function to map_renderer, plus custom style.json

mitchellgoffpc:


selfdrive/navd/map_renderer.cc
```diff
@@ -210,6 +222,11 @@ extern "C" {
     return new MapRenderer(settings, false);
   }
 
+  void map_renderer_update_zoom(MapRenderer *inst, float zoom) {
+    inst->updateZoom(zoom);
+    QApplication::processEvents();
```
adeebshihadeh, line 227: what’s this for? calling this manually is generally a bad practice

mitchellgoffpc: I copied the logic from map_renderer_update_position, but perhaps the processEvents call isn't necessary in this case.

mitchellgoffpc: I'm going to go ahead and merge as-is since this logic was added by willem and appears to date back to https://github.com/commaai/openpilot/pull/22800/commits/0072b641abdcd8fbf4824cb89c5800744a97afc1. Once I merge this plus a separate xx PR, I can add pipeline refs to check if replacing processEvents with update/repaint alters the behavior in any way. If those tests don't show any difference, I'll make a separate PR to switch them over.

PR #26000, opened 2022-10-07
params: remove all files if call clearAll with the ParamKeyType.ALL key.

### LINKED ISSUE #25994: params: remove params dir if all keys are cleared
Leftover params are causing some CI failures since they're no longer valid keys, even though Params::clearAll is called. If we call clearAll with the ALL key type, then we should just remove everything in the directory, regardless if it's a known key.

@deanlee want to try this one?

deanlee:
resolve https://github.com/commaai/openpilot/issues/25994


deanlee:
I'm concerned that rm the directory may have other side effects. It may be safer to just delete all files.

PR #26002, opened 2022-10-07
IsoTpParallelQuery: support delays between requests

sshane:
Might want to make this a generic parameter instead in `FwQueryConfig`, which would in turn be a generic parameter for `IsoTpParallelQuery`. No real point to only have delays between certain requests.

This has the side effect of fixing FW queries on CAN-FD. We can either retry the query right after if we see skipped iso-tp frames, or do it more proactively by querying each ECU with something safe, then waiting 200ms just like the dealer software does (it also requests a diagnostic mode `b"\x10\x03`, but the timing matters the most).

WIP


selfdrive/car/car_helpers.py
```diff
@@ -91,7 +91,7 @@ def fingerprint(logcan, sendcan):
       if cached_params.carName == "mock":
         cached_params = None
 
-    if cached_params is not None and len(cached_params.carFw) > 0 and cached_params.carVin is not VIN_UNKNOWN:
+    if False and cached_params is not None and len(cached_params.carFw) > 0 and cached_params.carVin is not VIN_UNKNOWN:
```
calledit, line 94: Seams strange to have a "if False" statement in master.

sshane:
This is done much simpler in https://github.com/commaai/openpilot/pull/26040, and more correctly too. As this PR would only cease queries to individual ECUs, it would still be possible to continually be querying through the gateway here, based on ECU response times

PR #26006, opened 2022-10-07
VW MNB: Volkswagen Crafter Mk2

jyoung8607:
Adding support for the Crafter platform/family of light commercial trucks and selected badge-engineered cousins. This platform is technically known as MNB, but from an openpilot perspective, doesn't seem any different from MQB. In fact, certain controllers and other parts are lifted directly from the Mk7 Golf.

**Models**: VW Crafter, VW e-Crafter, VW Grand California, MAN TGE, MAN eTGE. I really wanted to add the [Torsus Terrastorm](https://torsus.eu/en/terrastorm) line of apocalypse buses but their site won't option the underlying Crafter with ACC. Maybe if we call...

- [ ] ~~route with openpilot:~~ (deferred for now per DM conversation with adeeb)
- [x] added to CARS
- [x] added to RELEASES

Thanks to community VW Crafter owner lost_from_the_tribe!


GregorKikelj:
Oh cool, MAN support. BTW these are vans not trucks(I mean I guess you could configure them as sort of wanna be trucks.... But still not trucks)

jyoung8607:
The end user here is having to have a custom windshield mount printed, hoping to have a test drive this weekend.

adeebshihadeh:
Any update on this?

jyoung8607:
As of our last contact on 11/18 he's still waiting on a mount. Not sure what's going on. I'll ping him.

jyoung8607:
image

adeebshihadeh:
are we just waiting on a route? @sshane should be able to find and pull one if so

jyoung8607:
Out of draft, ready for review.

adeebshihadeh:
To recap our discussion:
* we're reasonably confident this code (including fingerprint) will work with an ACC-equipped Crafter
* non-ACC cars won't work on stock openpilot, but we'll make this more explicit with the `wrongCruiseMode` event

PR #26010, opened 2022-10-08
GM camera ACC: hide take steering alert

sshane:


sshane:
@adeebshihadeh The camera needs to see this bit rise for multiple frames in a row, and we can't guarantee that sending over the EPS' message. I tried quite a few different timing options (as well as sending multiple) and the camera always picks out the true EPS message over ours, after we send our message for a few minutes.

However, forwarding through openpilot is reliable (5+ minutes with no warning). 

I made sure that if we fault the camera somehow (it sees no or bad PSCMStatus messages), ACC and FCW from the camera is maintained, as well as LKAS from openpilot.

It's also a temporary fault if we block this message, so the second it sees PSCMStatus again, the camera gets out of its LKAS faulted state (which has no effect).

sshane:
AEB and FCW are retained even if the camera doesn't receive PSCMStatus: 376bf99325883932|2022-10-19--22-40-30

selfdrive/car/gm/carstate.py
```diff
@@ -26,6 +26,7 @@ def update(self, pt_cp, cam_cp, loopback_cp):
     self.prev_cruise_buttons = self.cruise_buttons
     self.cruise_buttons = pt_cp.vl["ASCMSteeringButton"]["ACCButtons"]
     self.buttons_counter = pt_cp.vl["ASCMSteeringButton"]["RollingCounter"]
+    self.pscm_status = pt_cp.vl["PSCMStatus"]
```
adeebshihadeh, line 29: why did you remove the copy? this overwrites the dict now

sshane: It's fine if we do, but I can add it back so it's explicit

PR #26011, opened 2022-10-08
Add ip and port arguments to /tools/sim/bridge.py

wjxjmj:
Since Carla requires too many resources, running the Carla client on a PC with a high performance video card while working on openpilot on a laptops is a better configuration.

Currently, bridge.py in the latest docker images assigns only a local IP address as 127.0.0.1 in function connect_carla_client(). Therefore, each time after executing /tools/sim/start_openpilot_docker.sh, users have to manually modify the very function in file bridge.py.

To make it easier to set the ip address and port number of the Carla client, I added the ip and port arguments to bridge.py. Also, I have updated the descriptions of the new arguments in README.md.


incognitojam:
thanks!

PR #26013, opened 2022-10-08
GM camera ACC: reduce LKAS faults on startup

sshane:
Previously our first counter sent for `ASCMLKASteeringCmd` would only ever be 1, meaning we'd very often fault if the last counter from the camera was also 1. I've seen our Bolt be totally fine while skipping a single or multiple counter values, so I believe it's only when we duplicate a counter value that it faults (1 out of 4 times).

Will test on our Bolt to see what reliable causes a fault and prevents it before merging

WIP


sshane:
Need to merge https://github.com/commaai/openpilot/pull/26024 first

PR #26014, opened 2022-10-08
GM: add checksum to button message

JMPZ11:
**Description**
It appears some components reject button messages without the checksum. This PR adds the checksum to button spam messages

**Verification**
As there are a limited number of possible values, it was possible to compare the output of this formula with all of them. Tested the checksum with different button values and all the rolling counters.

- [ ] Deal with bit 24. It only appears to be "1" in cars with ACC. The checksum formula needed to be changed based on a non-ascm car since we don't currently send this value. I would like to have a single formula that works in both scenarios


selfdrive/car/gm/gmcan.py
```diff
@@ -4,6 +4,11 @@ def create_buttons(packer, bus, idx, button):
   values = {
     "ACCButtons": button,
     "RollingCounter": idx,
+    # When bit 24 is 1 (eg Acadia)
+    #"SteeringButtonChecksum": 0x1000 - ((button & 0x7) << 4) - 0x24 - (0x4ef * (3-idx)),
+    # When bit 24 is zero (Bolt EV no ACC)
+    # TODO: Try to merge the two formulae
+    "SteeringButtonChecksum": 0x1000 - ((button & 0x7) << 4) - 0x330 - (0x3f0 * (3-idx)),
```
adeebshihadeh, line 11:
```suggestion
    # TODO: merge into one
    # When bit 24 is 1 (e.g. Acadia),
    #"SteeringButtonChecksum": 0x1000 - ((button & 0x7) << 4) - 0x24 - (0x4ef * (3-idx)),
    # When bit 24 is zero (e.g. Bolt EV no ACC),
    #"SteeringButtonChecksum": 0x1000 - ((button & 0x7) << 4) - 0x330 - (0x3f0 * (3-idx)),
```

sshane: merged into one

sshane:
@JasonJShuler please feel free to clean up what I did in a future PR! Thanks for opening this 🙂

PR #26015, opened 2022-10-08
GM: disabling checks on loopback bus

JMPZ11:
**Description**

Because the loopback buses are based on messages sent by OP, timing checks are not relevant. They cause false fault events and serve no purpose. The alert itself is hidden behind the untested-branch alert and goes away as soon as the relay opens. This fix disables all the checks for the one message in the loopback bus.

### Example from Connect of the error at startup

Shane tells me an orange section at startup is normal so this may not be the best test - presumably this change should shorten the time in fault at startup.
image

**Verification**

I will need @sshane 's help testing.

- [ ] Test on supported car and ensure error alerts are gone

### DIFF
--- a/selfdrive/car/gm/carstate.py
+++ b/selfdrive/car/gm/carstate.py
@@ -160,9 +160,7 @@ class CarState(CarStateBase):
     ]
 
     checks = [
-      ("ASCMLKASteeringCmd", 10), # 10 Hz is the stock inactive rate (every 100ms).
-      #                             While active 50 Hz (every 20 ms) is normal
-      #                             EPS will tolerate around 200ms when active before faulting
+      ("ASCMLKASteeringCmd", 0), # As a loopback bus, timing enforcement is unnecessary and causes superfluous faults
     ]
 
-    return CANParser(DBC[CP.carFingerprint]["pt"], signals, checks, CanBus.LOOPBACK)
+    return CANParser(DBC[CP.carFingerprint]["pt"], signals, checks, CanBus.LOOPBACK, enforce_checks = False)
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
LGTM, @sshane?

sshane:
Yep, don't see a reason.

selfdrive/car/gm/carstate.py
```diff
@@ -160,9 +160,7 @@ def get_loopback_can_parser(CP):
     ]
 
     checks = [
-      ("ASCMLKASteeringCmd", 10), # 10 Hz is the stock inactive rate (every 100ms).
-      #                             While active 50 Hz (every 20 ms) is normal
-      #                             EPS will tolerate around 200ms when active before faulting
+      ("ASCMLKASteeringCmd", 0), # As a loopback bus, timing enforcement is unnecessary and causes superfluous faults
```
sshane, line 163:
```suggestion
      ("ASCMLKASteeringCmd", 0),
```

selfdrive/car/gm/carstate.py
```diff
@@ -160,9 +160,7 @@ def get_loopback_can_parser(CP):
...
     ]
 
-    return CANParser(DBC[CP.carFingerprint]["pt"], signals, checks, CanBus.LOOPBACK)
+    return CANParser(DBC[CP.carFingerprint]["pt"], signals, checks, CanBus.LOOPBACK, enforce_checks = False)
```
sshane, line 166:
```suggestion
    return CANParser(DBC[CP.carFingerprint]["pt"], signals, checks, CanBus.LOOPBACK, enforce_checks=False)
```

PR #26019, opened 2022-10-08
cabana: increase replay's segment cache limit & add setting dialog 

deanlee:
replay caches 3 segments by default, it's enough for normal playback.  but this will cause the cabana can only display 3 minutes of data.
cabana only subscribes to “can” and “roadEncodeIdx” with new function `replay::installEventFilter`. the overhead of replay is very small. so we can safely increase the segment cache limit to display more data.

default cache limit: (3m):
cach_3

cache 10m:

cach_10


deanlee:
added a setting dialog
settings

PR #26024, opened 2022-10-09
GM: refactor steer fault avoidance code

sshane:
Just seems a bit easier to understand for me, plus fixes a bug where if the loopback bus lags, we can send the same counter multiple times

Diff: https://imgur.com/a/iRZnlic


selfdrive/car/gm/carcontroller.py
```diff
@@ -42,23 +42,21 @@ def update(self, CC, CS):
     can_sends = []
 
     # Steering (50Hz)
-    # Avoid GM EPS faults when transmitting messages too close together: skip this transmit if we just received the
-    # next Panda loopback confirmation in the current CS frame.
-    if CS.lka_steering_cmd_counter != self.lka_steering_cmd_counter_last:
-      self.lka_steering_cmd_counter_last = CS.lka_steering_cmd_counter
```
sshane, line 48: What this used to assume was that one message missing on the loopback bus was due to lag, and two+ messages blocked. We however don't know either for sure without a blocked/192 bus. If we block we want to re-transmit, and if we lag we want to increment counter as normally, but skip sending.

Perhaps we should do the panda block recovery with this

sshane:
And it looks fairly common for all brands to have single blocked messages on falling edge of controls allowed, so we can't refactor this code without maintaining it's blocking handling -- which again was purely accidental, but prevents a ***lot*** of faults.

Screenshot from 2022-10-11 17-09-27

sshane:
Cleaned up a bit in https://github.com/commaai/openpilot/pull/26039

PR #26027, opened 2022-10-09
Recover from blocked steering messages

sshane:
To recover from blocked messages, think we can do this incrementally


adeebshihadeh:
We can certainly do this incrementally, but I'd really like to do this generically from the start. This is a problem with all ports, and it'll quickly get out of hand if we do it like this for all of them.

sshane:
proof of concept, but did you mean something like this where controlsd passes in all blocked addrs?

gregjhogan:
This is a big change, but perhaps we should break apart generating the car specific state and sending CAN messages at specific frequencies.  Like have carcontroller generate the complete car specific state at a high frequency and something else manage outputting the CAN messages each at the appropriate frequency (e.g. register functions with associated frequencies).  Then we can detect blocked messages and immediately send a new message generically as fast as possible.  We could probably find a way to do it one car at a time if we thought it was a good long-term abstraction.

sshane:
will come up with something final and re-open

PR #26033, opened 2022-10-11
Ford: fix button press and cleanup

incognitojam:
- Send fake button messages to the main bus
  - Fixes resume from stop
  - Don't send TJA toggle message immediately after resume/cancel (elif branch)
- Cleanup steering control code
  - Remove EPS roll compensation compensation
  - Set ramp rate based on steering angle change (this helps to reduce ping pong and effectively creates a small deadzone)
- Car state: use `Veh_V_ActlBrk` signal for vehicle speed
  - it is more noisy but it appears `Veh_V_ActlEng` was just a moving average of `Veh_V_ActlBrk`, we already have a filter in carstate
image
- Add more notes on TJA/LCA command

Related:
- https://github.com/commaai/panda/pull/1085

### DIFF
--- a/selfdrive/car/ford/carcontroller.py
+++ b/selfdrive/car/ford/carcontroller.py
@@ -3,7 +3,7 @@ from cereal import car
 from common.numpy_fast import clip, interp
 from opendbc.can.packer import CANPacker
 from selfdrive.car.ford import fordcan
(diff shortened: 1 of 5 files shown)


adeebshihadeh:
Good to merge?

incognitojam:
Yep!

PR #26039, opened 2022-10-12
GM camera ACC: reduce LKAS faults on startup

sshane:
A version of https://github.com/commaai/openpilot/pull/26013/files that should be mergeable


selfdrive/car/gm/carcontroller.py
```diff
@@ -44,20 +45,22 @@ def update(self, CC, CS):
...
-    if CS.lka_steering_cmd_counter != self.lka_steering_cmd_counter_last:
-      self.lka_steering_cmd_counter_last = CS.lka_steering_cmd_counter
+    if CS.loopback_lka_steering_cmd_updated:
+      self.lka_steering_cmd_counter += 1
+      self.sent_lka_steering_cmd = True
     elif (self.frame % self.params.STEER_STEP) == 0:
+      # Initialize ASCMLKASteeringCmd counter using the camera
+      if not self.sent_lka_steering_cmd and self.CP.networkLocation == NetworkLocation.fwdCamera:
+        self.lka_steering_cmd_counter = CS.camera_lka_steering_cmd_counter + 1
+
```
sshane, line 55: The first few messages are nearly always blocked, this ensures we always send the correct counter (verified by checking this message on bus 0)

sshane:
This is needed to reduce the LKAS faults on start up for camera ACC cars. I discovered that the EPS allows two instances of frame-tomfoolery before it faults. For example, you can skip a counter value, or send a duplicate counter value a total of two times, and more and the EPS instantly faults.

Since the counters can be often misaligned when the relay closes, at worse, this PR guarantees no faults a minimum of 3 starts and stops (assuming counter is misaligned every time the relay closes, we'd use our allotment of 2 of these events every time the car turns off).

sshane:
Diff is from previously getting the counter from the loopback bus (isn't first segment, and it didn't happen to start at 0), vs now just incrementing a counter, which both are the same in live use.

PR #26040, opened 2022-10-12
Hyundai CAN-FD: add delay between requests

### LINKED ISSUE #25710: Certain ECUs are skipping iso-tp CAN packets on bus 1
Moved from PR https://github.com/commaai/openpilot/pull/25687 which isn't a real solution since we drop VIN CAN packets too. Some ECUs on Hyundai skip iso-tp frames occasionally, causing that FW version to be thrown out. However if you check the CAN data, those packets aren't skipping on the other buses. Is this a panda bug or ECU bug? Why do they show up on other buses when we request on OBD-II bus?

sshane:
I believe the gateway needs to see no traffic from the client after a request in order to do some sort of initialization where it doesn't skip any more iso-tp frames. Not sure if this is the best way to go about this yet.

Dealer software sends some diagnostic commands, then tester present 3-4 times in quick succession, then waits 200ms to do FW queries.


sshane:
Going to query from the camera https://github.com/commaai/openpilot/pull/26063

PR #26045, opened 2022-10-12
Hyundai CAN-FD: remove 90° steering lockout

sshane:
panda PR: https://github.com/commaai/panda/pull/1098

TODO:
- [ ] Decide if we need torque winddown bit.
      Certain cars are not affected immediately by dropping torque actuation bit and maintain full torque


ioniq5-cz:
I just passed the torque_fault along from carstate.py to hyundaicanfd.py. we definitely need torque winddown bit still since else we get a slapping noise on the EPS. I tried this a few weeks back and now
I did the same with the branch here. See video for reference: https://discord.com/channels/469524606043160576/538741329799413760/1019205700251029554

I searched for days for the correct winddown but without success. 

    "STEER_REQ": 1 if lat_active and not torque_fault else 0,

sshane:
Tried out 1 frame multiple times a second vs. 2 frames once a second, and we indeed need to send two invalid messages consecutively to make this fault-free: https://imgur.com/a/DM8xwh6

sshane:
@ioniq5-cz do you have HDA2 or 1? Does openpilot send LFA or LKAS for your car? I noticed tonight that LFA does the winddown for LKAS, so if we can send LFA we improve the loss of torque a bit. Sending LKAS and setting the bit low, the Driving ECU actually ramps our requested torque down and up each time

ioniq5-cz:
HDA2

sshane:
Ah, so make sure openpilot longitudinal is enabled and let me know how it goes when you try it again!

PR #26048, opened 2022-10-12
all_known_cars considers cars in DBC

sshane:
We weren't testing the Odyssey CHN car in test_models which turns out has a lot of CAN valid errors


adeebshihadeh:
This is kind of intentional if a car isn't in either of the fingerprinting dicts.

sshane:
Yeah okay, this also doesn't catch defaultdicts

PR #26049, opened 2022-10-12
regen: Refactor log migrate functions to avoid needing azure keys

nuwandavek:


adeebshihadeh:
This isn't really the right fix, as there's nothing guaranteeing this won't happen again. A potential, more correct fix would be to move [these lines](https://github.com/commaai/openpilot/blob/23e78da6a49b8aa43c7d60226aa466a8599be700/selfdrive/test/update_ci_routes.py#L17) into a function so they don't run at import time.

selfdrive/test/process_replay/helpers.py
```diff
@@ -24,3 +25,58 @@ def __exit__(self, exc_type, exc_obj, exc_tb):
     shutil.rmtree(self.msgq_path, ignore_errors=True)
     del os.environ['OPENPILOT_PREFIX']
     return False
+
+
+def migrate_carparams(lr):
```
adeebshihadeh, line 30: can you move these back into regen.py?

selfdrive/test/update_ci_routes.py
```diff
@@ -14,11 +14,16 @@
   (_DATA_ACCOUNT_CI, "commadataci"),
 ]
 
-DEST_KEY = azureutil.get_user_token(_DATA_ACCOUNT_CI, "openpilotci")
-SOURCE_KEYS = [azureutil.get_user_token(account, bucket) for account, bucket in SOURCES]
-SERVICE = BlockBlobService(_DATA_ACCOUNT_CI, sas_token=DEST_KEY)
+
+def initialize_azure_keys():
```
adeebshihadeh, line 18:
```suggestion
def get_azure_keys():
```

PR #26050, opened 2022-10-12
Update RELEASES.md

sshane:


adeebshihadeh:
I don't think this is worth adding to the release notes, but let's definitely discuss it in the release blog post.

PR #26059, opened 2022-10-13
Updated translation main_zh-CHS

OxygenLiu:
100% completed


sshane:
Can you only update the missing translations in this PR, and open a new one to revise translations?

OxygenLiu:
> Can you only add the missing translations in this PR and revert everything else, and open a new one to revise translations?

https://github.com/commaai/openpilot/pull/26492

PR #26063, opened 2022-10-13
CAN-FD HKG: query FW versions from camera

sshane:


hoomoose:
This worked for me.
Route: 68d6a96e703c00c9%7C2022-10-13--10-04-35

selfdrive/car/hyundai/values.py
```diff
@@ -292,6 +292,19 @@ class Buttons:
       [HYUNDAI_VERSION_REQUEST_MULTI],
       [HYUNDAI_VERSION_RESPONSE],
     ),
+    # CAN-FD queries
+    Request(
+      [HYUNDAI_VERSION_REQUEST_LONG],
+      [HYUNDAI_VERSION_RESPONSE],
+      # whitelist_ecus=[Ecu.fwdRadar],
```
sshane, line 299:
```suggestion
      whitelist_ecus=[Ecu.fwdRadar],
```

selfdrive/car/hyundai/values.py
```diff
@@ -292,6 +292,19 @@ class Buttons:
...
+      bus=4,
+    ),
+    Request(
+      [HYUNDAI_VERSION_REQUEST_LONG],
+      [HYUNDAI_VERSION_RESPONSE],
+      # whitelist_ecus=[Ecu.fwdCamera, Ecu.adas],
```
sshane, line 305:
```suggestion
      whitelist_ecus=[Ecu.fwdCamera, Ecu.adas],
```

selfdrive/car/hyundai/values.py
```diff
@@ -292,6 +292,19 @@ class Buttons:
       [HYUNDAI_VERSION_REQUEST_MULTI],
       [HYUNDAI_VERSION_RESPONSE],
     ),
+    # CAN-FD queries
+    Request(
+      [HYUNDAI_VERSION_REQUEST_LONG],
+      [HYUNDAI_VERSION_RESPONSE],
+      whitelist_ecus=[Ecu.fwdRadar],
+      bus=4,
```
sshane, line 300: E-CAN is bus 5, and from the schematics the radar is not on E-CAN, so this checks out why it was skipping on bus 5, but not A-CAN (4/6).

It also looks like the ADAS Unit is on E-CAN as well

selfdrive/car/hyundai/values.py
```diff
@@ -1327,6 +1333,7 @@ class Buttons:
       b'\xf1\x00CV1 MFC  AT USA LHD 1.00 1.05 99210-CV000 211027',
       b'\xf1\x00CV1 MFC  AT EUR LHD 1.00 1.05 99210-CV000 211027',
     ],
+    (Ecu.adas, 0x730, None): [],
```
adeebshihadeh, line 1336: make sure this isn't an essential ECU. only HDA2 cars have an ADAS ECU, though perhaps the camera will respond to this query

selfdrive/car/hyundai/values.py
```diff
@@ -292,6 +292,19 @@ class Buttons:
...
+      [HYUNDAI_VERSION_REQUEST_LONG],
+      [HYUNDAI_VERSION_RESPONSE],
+      whitelist_ecus=[Ecu.fwdRadar],
+      bus=4,
+    ),
+    Request(
+      [HYUNDAI_VERSION_REQUEST_LONG],
+      [HYUNDAI_VERSION_RESPONSE],
+      whitelist_ecus=[Ecu.fwdCamera, Ecu.adas],
+      bus=5,
```
adeebshihadeh, line 306: what happens if you don't have an aux panda? does it just wait for the timeout?

selfdrive/car/hyundai/values.py
```diff
@@ -292,6 +292,19 @@ class Buttons:
       [HYUNDAI_VERSION_REQUEST_MULTI],
       [HYUNDAI_VERSION_RESPONSE],
     ),
+    # CAN-FD queries
+    Request(
+      [HYUNDAI_VERSION_REQUEST_LONG],
+      [HYUNDAI_VERSION_RESPONSE],
+      whitelist_ecus=[Ecu.fwdRadar],
```
sshane, line 299:
```suggestion
      whitelist_ecus=[Ecu.fwdRadar, Ecu.adas],
```

selfdrive/car/hyundai/values.py
```diff
@@ -292,6 +292,19 @@ class Buttons:
...
+      bus=4,
+    ),
+    Request(
+      [HYUNDAI_VERSION_REQUEST_LONG],
+      [HYUNDAI_VERSION_RESPONSE],
+      whitelist_ecus=[Ecu.fwdCamera, Ecu.adas],
```
sshane, line 305:
```suggestion
      whitelist_ecus=[Ecu.fwdCamera],
```

selfdrive/car/hyundai/values.py
sshane: Yep, checked and it still exact matches without

selfdrive/car/hyundai/values.py
sshane: I believe so. I was thinking we could maybe have some generic way to determine if we have an aux panda and only query with those requests, if specified in the config. But not sure how complex that would be

selfdrive/car/hyundai/values.py
sshane: The ADAS ECU (Driving) is not available on bus 4, but it doesn't skip on bus 5 (A-CAN), so using that.

selfdrive/car/hyundai/values.py
sshane: This adds around 0.2-0.4 seconds of extra delay

sshane:
d545129f3ca90f28|2022-10-13--16-21-25

sshane:
I also need to handle the whitelisting properly, I'll fix it soon

selfdrive/car/car_helpers.py
```diff
@@ -76,7 +76,7 @@ def _get_interface_names() -> Dict[str, List[str]]:
 
 
 # **** for use live only ****
-def fingerprint(logcan, sendcan):
+def fingerprint(logcan, sendcan, aux_panda):
```
adeebshihadeh, line 79: pass in the number of pandas instead

selfdrive/car/fw_versions.py
```diff
@@ -252,6 +252,11 @@ def get_fw_versions(logcan, sendcan, query_brand=None, extra=None, timeout=0.1,
   for addr in tqdm(addrs, disable=not progress):
     for addr_chunk in chunks(addr):
       for brand, r in requests:
+        # Skip queries based off presence of external panda
+        aux_panda_query = r.bus > 3
+        if aux_panda_query != aux_panda:
+          continue
```
adeebshihadeh, line 258: make this generic for any number of pandas

sshane: Will change behavior to querying up to n pandas. So now it will run the internal panda queries as well since we don't know if we want to run queries on panda 3 or 1 and 2 as well. Unless we can always asume the last panda will be the one fingerprinting which I don't think we can?

sshane: limited the queries to the buses available with all pandas

selfdrive/car/hyundai/values.py
```diff
@@ -1369,16 +1363,6 @@ class Buttons:
     (Ecu.fwdCamera, 0x7c4, None): [
       b'\xf1\x00NX4 FR_CMR AT USA LHD 1.00 1.00 99211-N9240 14Q',
     ],
-    (Ecu.eps, 0x7d4, None): [
```
adeebshihadeh, line 1372: do we have a test that prevents these from being included if they're not in the whitelist?

selfdrive/controls/controlsd.py
```diff
@@ -104,9 +93,24 @@ def __init__(self, sm=None, pm=None, can_sock=None, CI=None):
...
+
+      num_pandas = len(messaging.recv_one_retry(self.sm.sock['pandaStates']).pandaStates)
+      self.CI, self.CP = get_car(self.can_sock, self.pm.sock['sendcan'], num_pandas)
+    else:
+      self.CI, self.CP = CI, CI.CP
+
+    self.joystick_mode = params.get_bool("JoystickDebugMode") or (self.CP.notCar and sm is None)
+    if not self.joystick_mode:
+      self.sm.ignore_alive.append('testJoystick')
+      self.sm.ignore_average_freq.append('testJoystick')
```
adeebshihadeh, line 112: these are essentially private variables of SubMaster. it's fine to just always ignore the joystick for now since it doesn't have a frequency specified anyway

selfdrive/car/hyundai/values.py
sshane: We have no whitelist for specific cars. We'd need to check CANFD_CAR

selfdrive/car/hyundai/tests/test_hyundai.py
```diff
@@ -0,0 +1,29 @@
...
+from selfdrive.car.hyundai.values import CANFD_CAR
+
+Ecu = car.CarParams.Ecu
+
+ECU_NAME = {v: k for k, v in Ecu.schema.enumerants.items()}
+VERSIONS = get_interface_attr("FW_VERSIONS", ignore_none=True)
+
+
+class TestHyundaiFingerprint(unittest.TestCase):
+  def test_auxiliary_requests_ecu_whitelist(self):
```
sshane, line 16:
```suggestion
  def test_auxiliary_request_ecu_whitelist(self):
```

selfdrive/car/hyundai/values.py
sshane: made a hyundai specific test file for this

Issue #25771, opened 2022-09-13
Hyundai:  HDA-1 CAN-FD resume from ACC standstill 


hoomoose:
### Describe the bug

Resume from standstill via button spamming does not work for hkg HDA-1 CAN-FD cars. My case is with an HDA 1 EV6 although I have been told this is the case with all HDA1 CAN-FD cars.
 
This is a known bug where CAN-FD hkg cars with HDA1 can send button presses, but the car does not respond to them. The HDA1 EV6 uses the 0x1cf cruise button signal on Panda bus 0 (send_can bus 4). The route provided at the 100s mark shows the car not resuming on Master (aea1a0d5e684daf7682703174233390b0cf79c83) with the following required fixes for hda1 EV6 (https://github.com/commaai/openpilot/pull/25723). 

Other HDA1 CAN-FD cars such as the Tucson use an alt_cruise_button signal for button presses but my understanding is that there is no response from the car in that case either.

Attempted solutions:
1. I have tried changing the code in hyundaicanfd.py to allow for the 0x1cf Cruise_buttons signal to be sent on bus 4 (as required for the HDA1 EV6) rather than bus 5 (as required for the HDA2 EV6), but this did not help. After this change, I can see button presses are being sent to the car by looking at send_can bus 4 in plotjuggler, but there is no response from the car.

### Provide a route where the issue occurs

68d6a96e703c00c9|2022-09-10--16-25-04--0

### openpilot version

https://github.com/commaai/openpilot/pull/25723

### Additional info

Is it possible to make a can message for an accelerator press at an extremely minor level (not even enough to move a car), but just enough to activate the resume from standstill?

hoomoose:
SOLVED!!!
However I am not sure how this needs to be coded. 
in Hyundaicanfd.py, the create_buttons msg needs to be sent on bus 6, and in safety_hyundai_canfd.h L22, 0x1cf needs to be on bus 2. That is it.

I just need help writing the If statement in hyundaicanfd.py that says if HDA2, bus 5, else bus 6. for the create_buttons msg. 

With this issue solved and the HDA1 fix PR, you can have many HDA1 HKG cars fully compatible!!!!! The rest will just be a matter of finding out if they have different signals to add to the dbc file. 

Low hanging fruit!

PR #26069, opened 2022-10-13
HKG: resume from standstill for HDA1 CAN-FD

hoomoose:
**Description**
Resume from Standstill does not work for HKG Can-FD cars with HDA1. This is fixed by putting the can send for Cruise Buttons on bus 6 rather than bus 5. The fix requires an IF statement, which I have not written since I am not sure how to do so.
This fix only solves the problem for HDA1 cars with the 0x1cf signal like the EV6 and I5.
Solves issue: https://github.com/commaai/openpilot/issues/25771

**Prerequisite**
This requires a simple Panda change on L39 of safety_hyundai_canfd.h
Bus 0 needs to be bus 2. That is it.
see: https://github.com/commaai/panda/pull/1102

**Verification** 
This fix has been employed in a fork hoomoose:hda1ev6i5 written specifically for hda1 ev6 and I5 cars and has been successfully tested/used by many!


mattwebbio:
Also interested in seeing this get merged!

selfdrive/car/hyundai/hyundaicanfd.py
```diff
@@ -45,7 +45,8 @@ def create_buttons(packer, cnt, btn):
     "SET_ME_1": 1,
     "CRUISE_BUTTONS": btn,
   }
-  return packer.make_can_msg("CRUISE_BUTTONS", 5, values)
+  #this needs to be bus 6 for hda1 cars and bus 5 for hda2
+  return packer.make_can_msg("CRUISE_BUTTONS", 6, values)
```
sshane, line 49:
```suggestion
  return packer.make_can_msg("CRUISE_BUTTONS", get_e_can_bus(CP), values)
```

selfdrive/car/hyundai/hyundaicanfd.py
```diff
@@ -39,13 +39,22 @@ def create_cam_0x2a4(packer, camera_values):
...
     "CRUISE_BUTTONS": btn,
   }
-  return packer.make_can_msg("CRUISE_BUTTONS", 5, values)
+  
+  if CP.flags & HyundaiFlags.CANFD_HDA2:
+    ret.append(packer.make_can_msg("CRUISE_BUTTONS", 5, values))
+  else:
+    ret.append(packer.make_can_msg("CRUISE_BUTTONS", 6, values))
+  
+  return ret
```
sshane, line 57:
```suggestion
  if CP.flags & HyundaiFlags.CANFD_HDA2:
    return packer.make_can_msg("CRUISE_BUTTONS", 5, values)
  else:
    return packer.make_can_msg("CRUISE_BUTTONS", 6, values)
```

no need to make it a list, I'd also see if we might want a helper function to identify this bus like we do for e can

selfdrive/car/hyundai/hyundaicanfd.py
```diff
@@ -39,13 +39,22 @@ def create_cam_0x2a4(packer, camera_values):
   })
   return packer.make_can_msg("CAM_0x2a4", 4, camera_values)
 
-def create_buttons(packer, cnt, btn):
+def create_buttons(packer, CP, cnt, btn):
+
+  ret = []
+
```
sshane, line 45:
```suggestion
def create_buttons(packer, CP, cnt, btn):
```

selfdrive/car/hyundai/hyundaicanfd.py
hoomoose: My code didn't work anyways... It gives an error. I will try your code and see if that works for me.

hoomoose:
Current code works!

PR #26070, opened 2022-10-13
HKG: Can-fd resume from standstill improvement

hoomoose:
**Description** 
Resume from standstill fails to work 1/5 times in a given drive in my hda1 EV6 (and others as well). To solve this problem I have employed the same technique as used in non-canfd hkg cars. I propose sending 5 "resume" button presses at a time. I attempted sending 25 button presses as is currently done with non-canfd cars, but this resulted in a communications error. 5 button presses appears to improve the resume from standstill (no failures in 3 days of driving) and does not give the communications error. 

**Verification** 
This fix has been employed for a few days now by a few users in a fork created specifically for hda1 ev6 and I5 cars (hoomoose:hda1ev6i5) since master does not have the prerequisite resume from standstill fix needed for hda1 ev6 and I5. In three days of driving with this fix, I have not had a failure of resume from standstill (yet...). However, this fix would affect all can-fd cars with the cruise_buttons signal and can/should be tested by hda1 and hda2 cars.

**Prerequisite**
For testing on an HDA1 car, the hda1 fix is a prerequisite:
https://github.com/commaai/openpilot/pull/26069

### DIFF
--- a/selfdrive/monitoring/driver_monitor.py
+++ b/selfdrive/monitoring/driver_monitor.py
@@ -32,7 +32,8 @@ class DRIVER_MONITOR_SETTINGS():
     self._BLINK_THRESHOLD = 0.895
 
     self._EE_THRESH11 = 0.275
-    self._EE_THRESH12 = 3.0
+    self._EE_THRESH12 = 5.5
+    self._EE_MAX_OFFSET1 = 0.06
     self._EE_THRESH21 = 0.01
     self._EE_THRESH22 = 0.35
 
@@ -204,7 +205,7 @@ class DriverStatus():
       distracted_types.append(DistractedType.DISTRACTED_BLINK)
 
     if self.ee1_calibrated:
-      ee1_dist = self.eev1 > self.ee1_offseter.filtered_stat.M * self.settings._EE_THRESH12
+      ee1_dist = self.eev1 > min(self.ee1_offseter.filtered_stat.M, self.settings._EE_MAX_OFFSET1) * self.settings._EE_THRESH12
     else:
       ee1_dist = self.eev1 > self.settings._EE_THRESH11
     # if self.ee2_calibrated:


hoomoose:
I had my first failure with this method, so it is not perfect. Just thought I would report this.

sunnyhaibin:
@hoomoose Have you tried something similar to this PR?
* https://github.com/commaai/openpilot/pull/25579

hoomoose:
I just tested out the solution offered by Sunnyhaibin, and have committed the changes.
In one day of driving through rush hour traffic, I had no errors and 100% resume rate. I would estimate success was 14/14. I do not think a conclusion was reached whether to merge the previous PR, but it certainly seems to cause no negative issues with an EV6 and would appear (with limited data from one can-fd car) to offer success. What would be a next step - have more cars try this out?

sunnyhaibin:
> I just tested out the solution offered by Sunnyhaibin, and have committed the changes. In one day of driving through rush hour traffic, I had no errors and 100% resume rate. I would estimate success was 14/14. I do not think a conclusion was reached whether to merge the previous PR, but it certainly seems to cause no negative issues with an EV6 and would appear (with limited data from one can-fd car) to offer success. What would be a next step - have more cars try this out?

Provide the route(s) with the resume tests should be sufficient for review.

sunnyhaibin:
> I just tested out the solution offered by Sunnyhaibin, and have committed the changes. In one day of driving through rush hour traffic, I had no errors and 100% resume rate. I would estimate success was 14/14. I do not think a conclusion was reached whether to merge the previous PR, but it certainly seems to cause no negative issues with an EV6 and would appear (with limited data from one can-fd car) to offer success. What would be a next step - have more cars try this out?

@hoomoose Provide the route(s) with the resume tests should be sufficient for review.

hoomoose:
Installing as I type - thanks!

hoomoose:
In driving through downtown to and from my rec center on the new Master with no modifications from me (which is awesome by the way!), I had the following results.
To gym 4/6 successful resumes (Route 68d6a96e703c00c9|2022-10-19--16-28-36)
From Gym 7/8 successful resumes (Route 68d6a96e703c00c9|2022-10-19--17-49-42)
(and 1 false positive where the car resumed but the car in front wasn't moving, but that's a success in my book.)
Overall successful resume rate 11/14.
I am happy to keep driving as is to get more data, but I would say this has been typical of my experience all along before trying to do improvements. Others who post (can-fd with hda1 and hda2) seem to have similar experiences of the ~15 -25% failure rate. Do you have a similar experience with your EV6?  I will also try this fix again after re-basing with current master to see if I have the same success rate as I previously posted and I will post routes.

PR #26071, opened 2022-10-13
HKG: add missing FW versions for 2022 Kia EV6 Light

mattwebbio:
Route: 859ffc6fd4e4168c|2022-10-13--11-01-34


sshane:
@adeebshihadeh we should support all EV6 and Ioniq 5 now, right?

mattwebbio:
> @adeebshihadeh we should support all EV6 and Ioniq 5 now, right?

I know you're not asking me, but yes you do! https://github.com/commaai/openpilot/pull/25723 was merged this morning 😄

adeebshihadeh:
> @adeebshihadeh we should support all EV6 and Ioniq 5 now, right?

yes

PR #26072, opened 2022-10-13
don't build cabana unless extras

incognitojam:
this caused CI to fail for xx since qt libs aren't installed in CI docker


adeebshihadeh:
This isn’t the right fix. Cabana and the tools should build by default on PC.

incognitojam:
we could add an option for "no qt"?

adeebshihadeh:
Sure, or add the dependencies to the xx Docker image. I think the latter is preferable.

PR #26073, opened 2022-10-13
Honda: rename outlier ECUs

sshane:
Needed for https://github.com/commaai/openpilot/pull/25859

Not sure if these are really different, or a mistake when adding them. It's strange that on most cars this address is fwdRadar, but on these it's fwdCamera, and we're missing radar? Also, why are we missing a lot of camera FW versions?


sshane:
Actually, doesn't need to be done just yet. Not sure if I like specifying all ECUs in config without also replacing addresses with Ecu enums everywhere

PR #26074, opened 2022-10-14
docs: fix model name/year formatting

incognitojam:


sshane:
We should remove this todo soon so this can't happen https://github.com/commaai/openpilot/blob/master/selfdrive/car/docs_definitions.py#L79

PR #26076, opened 2022-10-14
HKG: EV6 - improved LAT_ACCEL_FACTOR and FRICTION values 

mattwebbio:
**Description**

This PR improves understeering in the 2022 Kia EV6. This is done simply by cherry-picking a commit from https://github.com/commaai/openpilot/pull/25801, per my comment: https://github.com/commaai/openpilot/pull/25801#issuecomment-1279170301.

**Verification**

@hoomoose and I have found that in both of our EV6s, these torque values perform significantly better. Without them, stock OP can only maintain centering on straightaways.

**Route**

without: 859ffc6fd4e4168c|2022-10-13--17-49-56
with: 859ffc6fd4e4168c|2022-10-07--20-13-11

### DIFF
--- a/selfdrive/car/torque_data/override.yaml
+++ b/selfdrive/car/torque_data/override.yaml
@@ -21,7 +21,7 @@ FORD FOCUS 4TH GEN: [.nan, 1.5, .nan]
 COMMA BODY: [.nan, 1000, .nan]
 
 # Totally new cars
-KIA EV6 2022: [3.5, 3.0, 0.0]
+KIA EV6 2022: [3.0, 3.0, 0.05]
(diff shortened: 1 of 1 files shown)


sshane:
@nuwandavek do we want to wait for all the live torque params we get back to fix all the data?

nuwandavek:
@sshane I'm going to retune all values next week (same with the ionic PR) once the low speed controls are stable on good torque cars (toyotas mostly).

adeebshihadeh:
#26261 has new values that work really well on our EV6. It will also enable the live torque parameter learner on the EV6. Let us know if you still have issues after that's merged.

PR #26077, opened 2022-10-14
loggerd: fix length of ArrayPtr in handle_encoder_msg

jakepoz:
**Description** 
I was just reading the code for loggerd.cc and it appears that the ArrayPtr created from the received cereal message has the wrong size. It should be in 64bit capnp-words, not bytes.

**Verification** 
I don't own a Comma 3, but I'm using cereal and some of this code in my own robotics project and I was getting crashes.

### DIFF
--- a/selfdrive/loggerd/loggerd.cc
+++ b/selfdrive/loggerd/loggerd.cc
@@ -55,7 +55,7 @@ int handle_encoder_msg(LoggerdState *s, Message *msg, std::string &name, struct 
   int bytes_count = 0;
 
   // extract the message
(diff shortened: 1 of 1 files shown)


deanlee:
It is true that the correct size should be `msg->getSize() / sizeof(capnp::word))`.  but I suspect that your crash is because the data (`msg->getData()`) is not word aligned at all. 
FlatArrayMessageReader will read the first few bytes to determine the length of the message to read, if the start address  is word aligned and the length data in table is correct, sending a larger size to it should  not cause a crash.

jakepoz:
Yeah, I believe the crash was only showing up because there was corrupt data in the message.

PR #26079, opened 2022-10-14
GM: prevent fault when engaging at a stop

sshane:
This fixes a fault where you're pressing the brake at a standstill (< 0.5 m/s) and try to engage ACC (stock or OP-long).

The camera and ECM both have different brake pressed thresholds on the `ECMAcceleratorPos->BrakePedalPos` signal, and it looks like the camera's is lower (so that while engaged, the camera hits its threshold earlier and disengages without a fault. however GM didn't think of when you want to engage at a stop, we have the opposite problem).

When you engage at a stop, the camera thinks you're pressing the brake enough to start sending active ACC messages, but the ECM still wants you to press the brake further, causing a fault.


selfdrive/car/gm/interface.py
```diff
@@ -212,7 +212,9 @@ def _update(self, c):
     if ret.cruiseState.standstill:
       events.add(EventName.resumeRequired)
     if ret.vEgo < self.CP.minSteerSpeed:
-      events.add(car.CarEvent.EventName.belowSteerSpeed)
+      events.add(EventName.belowSteerSpeed)
+    if ret.vEgoRaw <= 0.5 and ret.brake <= 20:
+      events.add(EventName.depressBrakePedal)
```
sshane, line 217: The true condition is `ECMVehicleSpeed->VehicleSpeed` != 0, but it becomes non-zero around when vEgoRaw is 0.5 m/s.

The brake pressed threshold is also either 19 or 20. Faults at < 0.5 m/s slowly increasing pressure until you pass this threshold

selfdrive/car/gm/interface.py
```diff
@@ -212,7 +212,14 @@ def _update(self, c):
...
     if ret.vEgo < self.CP.minSteerSpeed:
-      events.add(car.CarEvent.EventName.belowSteerSpeed)
+      events.add(EventName.belowSteerSpeed)
+
+    if self.CP.networkLocation == NetworkLocation.fwdCamera:
+      # The ECM has a higher brake pressed threshold than the camera, causing an
+      # ACC fault when you engage at a stop with your foot partially on the brake
+      # TODO: use ECM's standstill threshold
+      if ret.vEgoRaw <= 0.5 and ret.brake <= 20:
+        events.add(EventName.depressBrakePedal)
```
sshane, line 222: Can't use brakePressed since it includes regen paddle which doesn't count as brake pressed to ECM at a standstill

sshane:
Pressed the brake at varying levels of pressure while at a stop and tried to engage, I never got an ACC fault from the brake threshold issue. Camera also was fine with our cancelling repeatedly.

selfdrive/car/gm/interface.py
```diff
@@ -212,7 +212,14 @@ def _update(self, c):
...
       events.add(EventName.resumeRequired)
     if ret.vEgo < self.CP.minSteerSpeed:
-      events.add(car.CarEvent.EventName.belowSteerSpeed)
+      events.add(EventName.belowSteerSpeed)
+
+    if self.CP.networkLocation == NetworkLocation.fwdCamera:
+      # The ECM has a higher brake pressed threshold than the camera, causing an
+      # ACC fault when you engage at a stop with your foot partially on the brake
+      # TODO: use ECM's standstill threshold
+      if ret.vEgoRaw <= 0.5 and ret.brake <= 20:
```
sshane, line 221:
```suggestion
      if ret.vEgoRaw <= 0.1 and ret.brake <= 20:
```

sshane: The wheel speeds are always 0.311 kph as the max speed that will allow engagement with brake pressed.

sshane:
@adeebshihadeh just to show how clear this fault is: https://imgur.com/merb9x2

sshane:
Showing the max speed with brake that it will engage at (0.311 kph on all the wheel speed sensors): https://imgur.com/a/VAxB55L

sshane:
Discovered another ACC fault! When in ***dashcam mode*** and reproducing this fault, randomly (only happened twice), the camera will show an "ACC Unavailable" alert on the dash and refuse to engage ACC. Sometimes you need to press the gas again, or wait multiple minutes, but it eventually resolves itself.

This fault is not from the ECM, in fact it will accept ACC messages fine (tried with OP long). This is the camera faulting and refusing to enable ACC, so this specific new fault cannot be fixed until we use OP long all the time.

selfdrive/car/gm/interface.py
```diff
@@ -212,7 +212,14 @@ def _update(self, c):
...
       events.add(EventName.resumeRequired)
     if ret.vEgo < self.CP.minSteerSpeed:
-      events.add(car.CarEvent.EventName.belowSteerSpeed)
+      events.add(EventName.belowSteerSpeed)
+
+    if self.CP.networkLocation == NetworkLocation.fwdCamera:
+      # The ECM has a higher brake pressed threshold than the camera, causing an
+      # ACC fault when you engage at a stop with your foot partially on the brake
+      # TODO: use ECM's standstill threshold
+      if ret.vEgoRaw < 0.1 and ret.brake <= 20:
```
sshane, line 221:
```suggestion
      if ret.vEgoRaw < 0.1 and ret.brake < 20:
```

selfdrive/car/gm/interface.py
```diff
@@ -212,7 +212,14 @@ def _update(self, c):
     if ret.cruiseState.standstill:
       events.add(EventName.resumeRequired)
     if ret.vEgo < self.CP.minSteerSpeed:
-      events.add(car.CarEvent.EventName.belowSteerSpeed)
+      events.add(EventName.belowSteerSpeed)
+
+    if self.CP.networkLocation == NetworkLocation.fwdCamera:
+      # The ECM has a higher brake pressed threshold than the camera, causing an
+      # ACC fault when you engage at a stop with your foot partially on the brake
+      # TODO: use ECM's standstill threshold
```
sshane, line 220:
```suggestion
```

selfdrive/car/gm/interface.py
```diff
@@ -212,7 +212,13 @@ def _update(self, c):
...
       events.add(EventName.resumeRequired)
     if ret.vEgo < self.CP.minSteerSpeed:
-      events.add(car.CarEvent.EventName.belowSteerSpeed)
+      events.add(EventName.belowSteerSpeed)
+
+    if self.CP.networkLocation == NetworkLocation.fwdCamera:
+      # The ECM has a higher brake pressed threshold than the camera, causing an
+      # ACC fault when you engage at a stop with your foot partially on the brake
+      if ret.vEgoRaw < 0.1 and ret.brake < 20:
+        events.add(EventName.gmAccFaultedTemp)
```
sshane, line 221: @adeebshihadeh maybe we can add a CS regenBraking field as well for stuff like this, so brakePressed is only the brake

sshane:
It actually appears the ECM's brake threshold is 20 on enable, and 11 on disable. So we'll need some logic for this (faults on OP long now)

PR #26081, opened 2022-10-14
GM camera ACC: raise brake pressed threshold

sshane:
Raises brake pressed threshold to what the ECM's is at a stop

At a stop:
- Camera starts longitudinal actuation at 11 units
- ECM stops faulting when it hits 20

Fixes: cruise faulted when pressing brake to disengage while car is braking (it thinks you pressed cancel)


selfdrive/car/gm/carstate.py
```diff
@@ -51,7 +51,11 @@ def update(self, pt_cp, cam_cp, loopback_cp):
     # To avoid a cruise fault we need to match the ECM's brake pressed signal and threshold
     # https://static.nhtsa.gov/odi/tsbs/2017/MC-10137629-9999.pdf
     ret.brake = pt_cp.vl["ECMAcceleratorPos"]["BrakePedalPos"]
-    ret.brakePressed = ret.brake >= 8
+    if self.CP.networkLocation != NetworkLocation.fwdCamera:
+      ret.brakePressed = ret.brake >= 8
+    else:
+      # Match the ECM's brake pressed threshold at a stop
+      ret.brakePressed = ret.brake >= 20
```
adeebshihadeh, line 58: can you add a sentence or two to the comment here explaining why this threshold is necessary?

adeebshihadeh (approved):

sshane:
Looks like we'll need two thresholds. On enabling at a stop it needs to be 20, on disabling ECM faults at 12 if you don't disable at 11

PR #26082, opened 2022-10-14
GM: match ECM standstill check

sshane:
Useful for avoiding faults both for OP long and without. As well as allowing engagement where the ECM allows it (sometimes up to 0.5 m/s, but as far as ECM and camera are concerned, we're at a standstill).

Also matches panda standstill now.

- [ ] Should check to make sure this isn't noisy at all on ASCM cars


selfdrive/car/gm/carstate.py
```diff
@@ -162,6 +164,7 @@ def get_can_parser(CP):
       ("ECMEngineStatus", 100),
       ("PSCMSteeringAngle", 100),
       ("ECMAcceleratorPos", 80),
+      ("ECMVehicleSpeed", 80),
```
sshane, line 167: TODO: find the rate

sshane:
This doesn't seem right. The camera's standstill check is 0.311 kph, but the ECM's is above 0.2 m/s

PR #26087, opened 2022-10-15
GM: allow enabling at a standstill

sshane:
expands to all cars (need to test on 17 Volt as it's said to fault if first enable is under 18 mph)


sshane:
Waiting for a 17 Volt tester

sshane:
For the 17 Volt (and others?), we need to properly handle allow enable under 18 after first engage first

PR #26090, opened 2022-10-15
Hyundai: Longitudinal support for CAN-based Camera SCC cars

sunnyhaibin:
openpilot longitudinal control support for CAN-based HKG cars that have the following messages send from the car on bus 2:
- `SCC11`
- `SCC12`
- `SCC13`
- `SCC14`
- `FCA11`
- `FCA12`

Currently with 2022 Hyundai Kona EV, without radar disable. As the messages send from the car on bus 2, we block the messages and send our own longitudinal commands to bus 0.

route:

Supported platforms:
- 2022 Hyundai Kona
- 2022 Hyundai Kona EV

**Requires**
* https://github.com/commaai/panda/pull/1107

Thanks to community Kona EV 2022 owner Zack1010OP.


selfdrive/car/hyundai/carstate.py
```diff
@@ -143,6 +143,8 @@ def update(self, cp, cp_cam):
     # save the entire LKAS11 and CLU11
     self.lkas11 = copy.copy(cp_cam.vl["LKAS11"])
     self.clu11 = copy.copy(cp.vl["CLU11"])
+    self.fca11 = copy.copy(cp_cruise.vl["FCA11"])
+    self.fca12 = copy.copy(cp_cruise.vl["FCA12"])
```
sshane, line 147: Can we not just forward these messages through panda without any modification?

sunnyhaibin: I believe we can. We only modify those messages to show the amber FCW icon on the dashboard at the moment. Since this does not disable radar, we probably could just forward these messages. @adeebshihadeh thoughts?

adeebshihadeh: until we confirm that AEB is unaffected, let's just treat them the same as a disabled radar, i.e. block in fwd hook and send our own (no copy)

sunnyhaibin:
> Will this enable longitudinal support for Sonata 2018? @sunnyhaibin

No, it does not. The Supported Platforms section in the description of this PR mentioned this.

sunnyhaibin:
@sshane Can we please add `ignore-stale` tag to this PR?

sunnyhaibin:
@sshane What else is needed for this PR? Multiple users have been using openpilot longitudinal with this change on sunnypilot and working as we expected.

sshane:

```bash
PR_NUMBER=33045
```

PR #26096, opened 2022-10-16
CI speedup

adeebshihadeh:


adeebshihadeh:
@deanlee the replay test is one of the slower tests  in the "unit test" job. want to look into speed it up a bit?

`./tools/replay/tests/test_replay -d yes` will show the times of each test case

PR #26097, opened 2022-10-16
Cabana: complete edit functions

deanlee:
1.complete edit functions
2.select cells in binary view to create new signal:

[Kazam_screencast_00004.webm](https://user-images.githubusercontent.com/27770/196039453-6f25a3c9-3a40-4f40-9934-cba097d61800.webm)

3.Fix some bugs that may cause crashes
4.cleanup layouts
5.synchronize the data displayed by the chart after editing dbc msg and signal

PR #26103, opened 2022-10-16
Cabana: show MSB/LSB of signals in the BinaryView

deanlee:
Use a custom delegate to draw cells

abc1


deanlee:
a custom SelectionModel has beed added to https://github.com/commaai/openpilot/pull/26097.  I recommend merging 26097 first. so that this PR can be done better base on 26097.

adeebshihadeh:
Can we also give the vertical header the same padding as the cells? Looks a bit weird with the no padding.

deanlee:
almost done.may need to add some bounds checks for array
cc3

deanlee:
Fixed a crash caused  by different data size for the same address from different bus

PR #26104, opened 2022-10-16
ui brightness: fix exposure scale factor

pd0wm:
Screen looked brighter than usual at night. I think the scale factor of 6x got lost in this refactor: https://github.com/commaai/openpilot/pull/25915/files#diff-b4c3a1489c8e0ba7f36583bd4088a422537e9098c4e82b102b22d57e72c03aceL185. 

This scale factor accounts for that fact that even in very dark scenes the exposure will never reach 100%, and depends on the sensor. Left the scale factor for the OX03C10 at 1. Not sure if you want the scale factor in ui or camerad.


adeebshihadeh:
Thanks :)

PR #26109, opened 2022-10-17
Move Qt moc files to scons cache directory

### LINKED ISSUE #26093: Move Qt moc files to another dir
perhaps put them in the scons cache directory? @deanlee interested in this one?

deanlee:
resolve https://github.com/commaai/openpilot/issues/26093

The disadvantage of set QT_MOCHPREFIX in the root is cannot have files with the same name. currently their filenames are all unique.so I only set QT_MOCHPREFIX here once.

To ensure that there is no problem with the same name, we need to set QT_MOCHPREFIX in the sub-sconstruct to a different directory.

@adeebshihadeh what do you think ? Do we need to set the moc directories separately ?


adeebshihadeh:
Nice, this should be enough

PR #26113, opened 2022-10-17
networking: fix metered setting

incognitojam:
- When the metered toggle is enabled (default) set the `connection.metered` setting to `NM_METERED_UNKNOWN` (allow network manager to decide?)
- When the toggle is disabled, set to `NM_METERED_NO`

Bug introduced in #25902

### DIFF
--- a/selfdrive/ui/qt/offroad/wifiManager.cc
+++ b/selfdrive/ui/qt/offroad/wifiManager.cc
@@ -368,7 +368,7 @@ void WifiManager::updateGsmSettings(bool roaming, QString apn, bool metered) {
       changes = true;
     }
 
-    int meteredInt = metered ? NM_METERED_NO : NM_METERED_UNKNOWN;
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/offroad/wifiManager.cc
```diff
@@ -368,7 +368,7 @@ void WifiManager::updateGsmSettings(bool roaming, QString apn, bool metered) {
       changes = true;
     }
 
-    int meteredInt = metered ? NM_METERED_NO : NM_METERED_UNKNOWN;
+    int meteredInt = metered ? NM_METERED_UNKNOWN : NM_METERED_NO;
```
sshane, line 371: is `NM_METERED_UNKNOWN` the default for new connections (lets NM decide)?

sshane: Looks like it https://developer-old.gnome.org/libnm/stable/libnm-nm-dbus-interface.html

sshane (approved):

PR #26129, opened 2022-10-18
HKG: add red panda footnote

sshane:
can possibly replace hardcoded footnotes in web2 with this + a bool field for the shop page


sshane:
not sure if we want this?

PR #26130, opened 2022-10-18
Hyundai HDA2 detection: use ADAS ECU

sshane:


adeebshihadeh (approved):

sshane:
process replay doesn't support HKG long, need a new route

PR #26135, opened 2022-10-18
Cabana: restore last session on startup

deanlee:
restore last session on startup.(e.g. route, dbc file, msg, charts, spliter sizes, video position, zoom range)


deanlee:
too many conflicts with master ,closed and moved to https://github.com/commaai/openpilot/pull/26186

Issue #24121, opened 2022-04-04
Set missing car minimum enable/steer speeds in CarParams


sshane:
Some cars don't have any minimum enable or steer speeds defined in their CarParams as we expect the car to handle this for us if we're `pcmCruise`. We should find the correct speeds and add them to CarParams so we can eventually remove these exceptions for the car docs: https://github.com/commaai/openpilot/blob/e40c36f22ba93200a5fc83211005b28fca3fa728/selfdrive/car/docs_definitions.py#L55

Draft PRs to work off of:

- https://github.com/commaai/openpilot/pull/23942
- https://github.com/commaai/openpilot/pull/23941

abesisu:
High there @sshane, I'm a student working on a class at Iowa State University where we choose an open-source project to analyze and contribute to. If I were to try and contribute to this issue, how might I find the correct missing minimum enable/steer speeds that need to be put in?

abesisu:
@sshane is there anything I'm doing wrong in [this PR](https://github.com/commaai/openpilot/pull/26136) that would lead these checks to fail? I haven't changed CarInfo and I've only added to CarParams. I originally tried to remove the min_steer_speed from values.py, but the test I'm failing said the minSteerSpeed should be in both CarParams and CarInfo.

jyoung8607:
> @sshane is there anything I'm doing wrong in [this PR](https://github.com/commaai/openpilot/pull/26136) that would lead these checks to fail? I haven't changed CarInfo and I've only added to CarParams. I originally tried to remove the min_steer_speed from values.py, but the test I'm failing said the minSteerSpeed should be in both CarParams and CarInfo.

The unit test is complaining _because_ both are set; they shouldn't be. The CarInfo setting should only be used if there's some reason they aren't (or can't) be set in CarParams; now that you have set them, they should be removed from CarInfo.

The process replay test is mainly a speed bump designed to make sure openpilot behaves the same before and after any given change, and if it doesn't, that someone examines the difference and confirms that it's expected and desired. In your case, it detects the CarParams change. Assuming the change has no ill effects in real world testing on those cars, a comma employee would review that difference and update the replay baseline to include them as part of merging the PR.

abesisu:
@jyoung8607 do you know of a way that I can get the unit test to pass? If what I've done is correct? https://github.com/commaai/openpilot/pull/26136

jyoung8607:
[...]

PR #26136, opened 2022-10-18
Honda: move common minSteerSpeeds to interface

abesisu:
This change utilizes interface.py for Honda cars better. Common concrete minSteerSpeed values in values.py are moved into the interface.

These changes are part of what @sshane describes in [this comment](https://github.com/commaai/openpilot/issues/24121#issuecomment-1265925494) on #24121


sshane:
I merged https://github.com/commaai/openpilot/pull/26138 so you can move the speeds into the interface. Can you rebase?

abesisu:
@sshane, @ incognitojam is there anything I should change or add to this PR?

adeebshihadeh:
We added the docs override field since we needed to verify that the minimum steer speeds are real and don't vary within a platform. It was a temporary addition to get the docs building programmatically without changing any behavior in case the information was incorrect. Do you actually know that these are correct and don't vary within a platform?

abesisu:
> We added the docs override field since we needed to verify that the minimum steer speeds are real and don't vary within a platform. It was a temporary addition to get the docs building programmatically without changing any behavior in case the information was incorrect. Do you actually know that these are correct and don't vary within a platform?

The changes in this PR are simply a refactor. None of the values change for the specific cars. Whatever values were there before are still used.

Issue #25996, opened 2022-10-07
Mobile web joystick performance improvement


ntegan1:
use websockets for communicating with webserver instead of xhr

maybe turn into an installable pwa as well, wss:// sslcert

pwa allows for joystick/controller attached to iphone/android via bluetooth or lightning/usb adapter,  
until someone gets bluetooth on c3

sxsmg:
Hey, can I take this up?

adeebshihadeh:
Closing sine this isn't an issue, but feel free to make a PR.

PR #26137, opened 2022-10-18
Tools: Send web joystick axes via WebSocket instead of XHR

ntegan1:
https://github.com/commaai/openpilot/issues/25996

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -19,3 +19,6 @@
 [submodule "tinygrad"]
   path = tinygrad_repo
   url = ../../commaai/tinygrad.git
+[submodule "third_party/websockets"]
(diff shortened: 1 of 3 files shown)


incognitojam:
Why not add websockets as a pipenv dependency?

ntegan1:
> Why not add websockets as a pipenv dependency?

I couldn't find where python packages get installed on device, was guessing it's in Agnos somewhere

incognitojam:
Oh, it would be in https://github.com/commaai/agnos-builder/blob/master/userspace.

incognitojam:
I forgot this might have been good in the new AGNOS update

PR #26140, opened 2022-10-18
mypy: disallow explicit Any in type annotations

incognitojam:


adeebshihadeh:
I don't think this is necessary, let's just not add any new ones.

PR #26141, opened 2022-10-18
athena: stricter socket timeout when device awake

incognitojam:
when device is awake, reduce the `TCP_KEEPIDLE` delay. also set a `TCP_USER_TIMEOUT` according to recommendation in https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/

```
# current websocket sockopts
SO_KEEPALIVE: 1
TCP_KEEPIDLE: 30     # change to 5 seconds when awake?
TCP_KEEPINTVL: 10
TCP_KEEPCNT: 3
TCP_USER_TIMEOUT: 0  # set to TCP_KEEPIDLE + TCP_KEEPINTVL * TCP_KEEPCNT?
```

https://man7.org/linux/man-pages/man7/tcp.7.html#:~:text=TCP_KEEPIDLE

### DIFF
--- a/common/params.cc
+++ b/common/params.cc
@@ -124,6 +124,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"HasAcceptedTerms", PERSISTENT},
     {"IMEI", PERSISTENT},
     {"InstallDate", PERSISTENT},
+    {"IsDeviceAwake", CLEAR_ON_MANAGER_START},
     {"IsDriverViewEnabled", CLEAR_ON_MANAGER_START},
     {"IsEngaged", PERSISTENT},
     {"IsLdwEnabled", PERSISTENT},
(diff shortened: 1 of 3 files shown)


gregjhogan:
I am pretty sure websockets are intended to have TCP keepalive disabled since they have a replacement mechanism ping/pong which they claim is better. 

https://websockets.readthedocs.io/en/stable/topics/timeouts.html

gregjhogan:
I guess it isn't clear to me if you can have HTTP keepalive active (so the server doesn't close the connection immediately) but disable TCP keepalive.  It seems best for the websocket to decide when the connection is dead and explicitly close and open a new connection if the websocket library has this built in.

gregjhogan:
Originally the goal was to only have the modem wake up once per minute when there is no activity to minimize power draw, which would require we disable TCP keepalive or set the interval to a very high value.  At the same time, you don't want to have to wait for a minute for a navigation request to get to the device.  I like the idea of being more aggressive with re-establishing the connection when on-road since we shouldn't have to worry about power draw.  I am not sure if we should be using TCP keepalive or ping/pong requests.  I have seen multi-second round-trip latency on cellular, which probably increases the difficulty here.

incognitojam:
https://hpbn.co/optimizing-for-mobile-networks/#eliminate-unnecessary-application-keepalives

I will double check the current socket configuration since it did look like we have TCP keepalive enabled, which might mean the modem doesn't get to idle much.

PR #26142, opened 2022-10-19
sim: outputting frames in NV12 format

jeroenlammersma:
Fix for:
- #25451 

Converting is done from YUV:I420 to YUV:NV12, using numpy. The conversion from I420 to NV12 is luckily done rather quickly. This addition will work fine for us, experimenting with openpilot using CARLA in colour again.

Although this works fine, the most optimal solution would be using a dedicated RGB to YUV:NV12 converted OpenCL program, I tried changing [rgb_to_yuv.cl](https://github.com/commaai/openpilot/blob/master/system/camerad/transforms/rgb_to_yuv.cl) to [real_debayer.cl](https://github.com/commaai/openpilot/blob/master/system/camerad/cameras/real_debayer.cl), but I am not able to compile the cl program. Also I am not sure if that will work.

So, although this fixes the issue, a nice extra would be:
- swapping to [real_debayer.cl](https://github.com/commaai/openpilot/blob/master/system/camerad/cameras/real_debayer.cll) or creating a dedicated rgb_to_nv12.cl program.

### DIFF
--- a/tools/sim/bridge.py
+++ b/tools/sim/bridge.py
@@ -103,15 +103,22 @@ class Camerad:
     img = np.reshape(img, (H, W, 4))
     img = img[:, :, [0, 1, 2]].copy()
 
-    # convert RGB frame to YUV
+    # TODO: use real_debayer.cl or dedicated rgb_to_nv12.cl program to convert from RGB to NV12 directly
+    # convert RGB frame to YUV I420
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Thanks for the PR; closing in favor of #26169 which does this in the CL kernel.

PR #26143, opened 2022-10-19
Hyundai: add corner radar FW query

adeebshihadeh:


selfdrive/car/hyundai/values.py
```diff
@@ -315,7 +315,8 @@ class Buttons:
     ),
   ],
   extra_ecus=[
-    (Ecu.adas, 0x730, None),  # ADAS Driving ECU on HDA2 platforms
+    (Ecu.adas, 0x730, None),         # ADAS Driving ECU on HDA2 platforms
+    (Ecu.cornerRadar, 0x7B7, None),
```
sshane, line 319: You'll also want to whitelist this ECU in at least one request (preferably just one that's common across all cars)

PR #26147, opened 2022-10-19
Toyota: parse radar for radar-ACC TSS-P cars

Smartype:
Test route: https://connect.comma.ai/2f37c007683e85ba/1674339187978/1674340344411


selfdrive/controls/radard.py
```diff
@@ -11,7 +11,7 @@
 from selfdrive.controls.lib.cluster.fastcluster_py import cluster_points_centroid
 from selfdrive.controls.lib.radar_helpers import Cluster, Track, RADAR_TO_CAMERA
 from system.swaglog import cloudlog
-
+from selfdrive.car.toyota.values import RADAR_ACC_CAR_TSSP
```
sshane, line 14: your car-specific code shouldn't touch radard like this

Smartype: ok, then we have to update RadarPoint to support absolute speed.

Smartype: @sshane Shall we add a union to RadarPoint or simple add a field named v and set vRel to nan when we want to use v instead?

sshane: You could pass in vEgo to RI.update() and set vRel appropriately with that information. Or add v to RadarPoint in cereal and have radar use whichever is set (or add a bool). I'd try the first option since that will be easier without any cereal changes

sshane:
Is PR just to add radar parsing support? No longcontrol is in the works?

Smartype:
I have a hardware similar to sdsu. I was told to separate the radar support and long control stuffs. 

> Is PR just to add radar parsing support? No longcontrol is in the works? If so, it sounds like when we crack disabling an ECU, we can get longcontrol for these cars (but radar will likely not be publishing any track data, so not sure if this is worth it)

Smartype:
0x1CF COUNTER FAIL #2 -- 14 -> 1

looks like the case fails in hyundai_canfd.dbc. I don't think this is related to this PR.

hyundai_canfd.dbc:BO_ 463 CRUISE_BUTTONS: 8 XXX

sshane:
The diff is from changing stuff unrelated to your PR (aLeadTau)

selfdrive/controls/radard.py
```diff
@@ -134,7 +129,7 @@ def update(self, sm, rr):
 
       # create the track if it doesn't exist or it's a new track
       if ids not in self.tracks:
-        self.tracks[ids] = Track(v_lead, self.kalman_params, 0.81 if self.radar_acc_tssp else 0.9)
+        self.tracks[ids] = Track(v_lead, self.kalman_params, math.pow(0.9, self.radar_ts / 0.05))
```
Smartype, line 132: Hi @sshane, I'd like do this, as this radar report at a lower frequency. The aLead TAU will be less sensitive. But this will fail the replay tests.
We could use fixed 0.9 for others, and 0.81 for this one. What's your suggestion? Thanks.

Smartype:
@sshane can we get this merged? If not, what else shall I do? Thanks!

williefeb19:
Is someone able to push this through or give some feedback? We have been patiently waiting.

sshane:
Can you rebase on master and clean up the PR?

Smartype:
@sshane on my private fork, I added a vAbs to RadarPoint. So that we can keep the changes in Toyota only. 

    vAbs @7 :Float32 = nan; # m/s

But this requires cereal updates. Which one you prefer?

Smartype:
@sshane  This is the PR for cereal

https://github.com/commaai/cereal/pull/414

sshane:
I'm fine with passing in v_ego for now. Did you rebase the PR?

Smartype:
> I'm fine with passing in v_ego for now. Did you accidentally close the PR?

This might be caused by force pushing...
I have created a new PR, but this one depends on cereal PR

https://github.com/commaai/openpilot/pull/27486

If you can merge the cereal PR, I can ask someone upload a route based on my branch.

PR #26148, opened 2022-10-19
GM camera ACC: show under enable speed alert

sshane:
A more generic fix for: https://github.com/commaai/openpilot/pull/26079.

plus allow engagement under 3 mph (replaces temp cruise fault alert)

this also fixes a fault with OP long (we need to disallow enabling under 3 except standstill + brake)


selfdrive/car/gm/carstate.py
```diff
@@ -54,11 +55,12 @@ def update(self, pt_cp, cam_cp, loopback_cp):
     # https://static.nhtsa.gov/odi/tsbs/2017/MC-10137629-9999.pdf
     ret.brake = pt_cp.vl["ECMAcceleratorPos"]["BrakePedalPos"]
     if self.CP.networkLocation != NetworkLocation.fwdCamera:
+      # TODO: verify 17 Volt can enable for the first time at a stop and set brake_pressed
       ret.brakePressed = ret.brake >= 8
```
sshane, line 59: 17 Volt faults if enabled under 18 mph the first time. We either need to handle it properly, or enabling at 0 mph doesn't fault. Left out ASCM GM to avoid allowing any faults with a TODO

selfdrive/car/gm/interface.py
```diff
@@ -207,19 +204,16 @@ def _update(self, c):
                                                          GearShifter.eco, GearShifter.manumatic],
                                        pcm_enable=self.CP.pcmCruise)
 
-    if ret.vEgo < self.CP.minEnableSpeed:
+    # Enabling at a standstill with brake is allowed
+    # TODO: verify 17 Volt can enable for the first time at a stop and allow for all GMs
+    if ret.vEgo < self.CP.minEnableSpeed and not (ret.standstill and ret.brake >= 20 and
```
sshane, line 209: @adeebshihadeh how do you feel about adding a regen field to carState? That way we can properly differentiate between regen brake and brake pedal? Also can check against each in test_models

PR #26151, opened 2022-10-19
FPv2 tests: test all addrs map to one ECU

sshane:
There's only one offender in all of openpilot, the address `0x18dab0f1` is labeled as both fwdRadar and fwdCamera for Honda (breaking future ECU address abstraction/clean up). Does anyone know which is correct?


sshane:
Going to assume that `0x18dab0f1` is actually fwdRadar (like the rest) on these cars, rather than two ECUs having the same address, which seems unlikely.

PR #26155, opened 2022-10-19
Cabana: no popup when adding a new signal.

deanlee:


deanlee:
Done. added some duplicate name and bounds checks

PR #26158, opened 2022-10-19
Cabana: load DBC from car fingerprint

deanlee:
stole code from original cabana.  :satisfied:


tools/cabana/SConscript
```diff
@@ -12,5 +12,6 @@ else:
 
 qt_libs = ['qt_util', 'Qt5Charts'] + base_libs
 cabana_libs = [widgets, cereal, messaging, visionipc, replay_lib, opendbc,'avutil', 'avcodec', 'avformat', 'bz2', 'curl', 'yuv'] + qt_libs
+qt_env.Execute('./generate_dbc_json.py --out car_fingerprint_to_dbc.json')
```
incognitojam, line 15: this script depends on params which might not be built at this point.
```
./generate_dbc_json.py --out car_fingerprint_to_dbc.json
  File "./generate_dbc_json.py", line 5, in <module>
    from selfdrive.car.car_helpers import get_interface_attr
  File "/home/batman/openpilot/selfdrive/car/car_helpers.py", line 5, in <module>
  File "/home/batman/openpilot/common/params.py", line 1, in <module>
    from common.params_pyx import Params, ParamKeyType, UnknownKeyName, put_nonblocking, put_bool_nonblocking # pylint: disable=no-name-in-module, import-error
ModuleNotFoundError: No module named 'common.params_pyx'
scons: *** Error 1
```
Is there a way to "sconsify" this instead of running as an external script, to make sure that dependencies are built first?

deanlee: execute script after building cabana should fix this issue https://github.com/commaai/openpilot/pull/26620

PR #26161, opened 2022-10-19
cabana: improve signal highlight on hover

deanlee:
draw border on hover. I think it's much clearer than before

[Kazam_screencast_00015.webm](https://user-images.githubusercontent.com/27770/196793963-aeca6221-2641-4223-a9dc-4e1b892eef4b.webm)


adeebshihadeh:
@deanlee loving the small, self-contained PRs :) 

Can we make it a little more subtle? Also, looks like the highlight border is rounded but nothing else is.

adeebshihadeh:
Hmm, how about just the white text? I think both is still pretty distracting.

deanlee:
emm.simple is better

PR #26164, opened 2022-10-19
Hyundai: pass actuation torque to safety

sshane:
Think we should split this up in a few PRs:
- Move car controller params into CarParams, so we can start to abstract car control code (base class, cleanup common code, reduce imports to values, etc)
- Then pass steerMax to panda safety

Since we only have two bytes to work with (one for flags and one for torque), we need to compress the max torque somehow, either with a separate set of bit flags, or just by scaling it down. @adeebshihadeh have any preference?


adeebshihadeh:
Scaling down sounds good.

PR #26178, opened 2022-10-20
use upstream tinygrad

YassineYousfi:


Pipfile
```diff
@@ -95,6 +95,7 @@ utm = "*"
 websocket_client = "*"
 hatanaka = "==2.4"
 PySide2 = "*"
+networkx = "~=2.3"
```
adeebshihadeh, line 98: what's this for? is it also needed on agnos?

YassineYousfi: for thneed, yeah can we get that in?

adeebshihadeh:
Only jenkins failed due to 38d6ce20726fee16339d0a30afee51759e94148d. @YassineYousfi would it be helpful to add an explicit a pc-thneed build to CI?

PR #26182, opened 2022-10-21
Mazda: Add CX-5 2023 to supported cars

Jafaral:
Confirmed working by a couple of users.


sshane:
Do you have a dongle id for a user? Are their fingerprints already in openpilot?

Jafaral:
@sshane 

e36424ef0cb55df0

adeebshihadeh:
Confirmed with VIN that it's a 2023. It was missing a camera FW version.

Jafaral:
Ah,  make sense sense OP falls  back to fuzzy matching and new users may not know what that means. Thanks @adeebshihadeh

PR #26183, opened 2022-10-21
CI: add PC thneed build

adeebshihadeh:


adeebshihadeh:
The build fails without much useful info Closing for now, but I can fix it up if it would be useful @YassineYousfi.
```
Segmentation fault (core dumped)
scons: *** [selfdrive/modeld/models/supercombo.thneed] Error 139
scons: building terminated because of errors.
```

PR #26184, opened 2022-10-21
GM: inactive gas/regen safety

sshane:
A minor pre-req for camera ACC OP long + adds safety around max regen

ASCM sends max regen, but camera ACC doesn't

### DIFF
--- a/selfdrive/car/gm/carcontroller.py
+++ b/selfdrive/car/gm/carcontroller.py
@@ -67,7 +67,7 @@ class CarController:
       # Gas/regen, brakes, and UI commands - all at 25Hz
       if self.frame % 4 == 0:
         if not CC.longActive:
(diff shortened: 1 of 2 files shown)


selfdrive/car/gm/carcontroller.py
```diff
@@ -67,8 +67,8 @@ def update(self, CC, CS):
       # Gas/regen, brakes, and UI commands - all at 25Hz
       if self.frame % 4 == 0:
         if not CC.longActive:
-          # Stock ECU sends max regen when not enabled
-          self.apply_gas = self.params.MAX_ACC_REGEN
+          # ASCM sends max regen, camera ACC sends slightly different value
```
sshane, line 70:
```suggestion
          # ASCM sends max regen when not enabled
```

PR #26196, opened 2022-10-21
GM: reduce LKAS faults while inactive

sshane:
This sends at the inactive rate when not actuating torque, as the stock camera/ASCM does. Sending at 10Hz when not active means we decrease our chances a lag will cause an LKAS fault under these circumstances

- [x] Make sure we don't send too close together when switching rates

We can also fix a fault when the relay closes on camera cars, but that will require some extra work to delay sending on occasion while inactive to sync up with the camera, so doing that in the next PR


Verylukyguy:
I have gotten these LKAS Faults while not engaged and sitting at idle in park.

sshane:
Send em over and I can check if timing was the issue

Verylukyguy:
> Send em over and I can check if timing was the issue

They were on my Comma 2 (e38a02cae1ef1e0c).  I preserved some of the routes, but they are months old, some of them are showing as no longer available.

sshane:
Let me know if you get any more then

sshane:
This regresses opening relay. Before we would try to send a message at 50 Hz, where the first few would get blocked by the relay not being opened. However now we only try at 10Hz, so it's possible we will wait too long to send our first counter, and fault the EPS much more often.

adeebshihadeh:
> @adeebshihadeh is it expected that the safety mode is not fully set by the time we initialize controls and are trying to send messages?

This isn't really guaranteed anywhere; it's always just been fast enough to not be a problem. I wouldn't be surprised if it took up to 100ms.

sshane:
Fixed above issue by setting the rate to active rate when initializing (not sent), preserves original opening-relay behavior now.

Unrelated to this, but I noticed a bug with how the logic works now. We can be sort of unlucky and try to send a message just before the relay opens, blocking it, then reading an updated counter from the camera (that doesn't reach the car), and end up skipping a counter. It's rare, but still a possibility.

Will be tackled in another PR, solution is to only read bus 0 to initialize the counter

sshane:
How this initialization looks: https://imgur.com/a/qsVWJda

Stress tested the transition from 10-50Hz a bit: 376bf99325883932|2022-10-24--23-42-57--7

Screenshot from 2022-10-24 22-49-50

sshane:
The diff is from a process replay bug, not sure why the previous refs were like this. It was spending all this time stuck in `if CS.loopback_lka_steering_cmd_updated:`. New ref looks correct at least:

Screenshot from 2022-10-24 23-15-37

PR #26197, opened 2022-10-21
GM camera ACC: cancel after delay

### LINKED ISSUE #26185: GM camera ACC: understand brake thresholds
What we know so far:

- ECM will fault if `ECMAcceleratorPos->BrakePedalPos` < 20 while engaging at a stop (camera tries to enable at >=11).
- Camera cancels ACC while moving at `ECMAcceleratorPos->BrakePedalPos`=11 if brake is pressed slowly, but is inconsistent when pressing brake faster

What we don't know:
- ECM-enforced: If there is a time/message count component to the brake pressed threshold
- Camera-related: If there is a component in commanded brake in how long it takes to cancel

We still are having problems where the camera cancels ACC 0.15 seconds after the brake is pressed to what we thought the threshold was while moving. So even with our brake pressed threshold of 20, we can still occasionally cancel earlier than the camera, causing that annoying FCW and fake cruise faulted alert.

sshane:
- [x] Find the max time the ECM allows before it faults using OP long (camera will always cancel before it faults)

Reverts https://github.com/commaai/openpilot/pull/26081

The camera (and likely ECM) has a set delay from pressing brake above its threshold to cancelling cruise, so if the user rapidly depresses brake, the threshold can rise well above ours and we start to cancel, causing a FCW and (fake) cruise fault. Raising the brake threshold was not the correct choice here, as it will constantly fault when using OP long.

Screenshot from 2022-10-20 22-51-08
Screenshot from 2022-10-20 22-55-07

### DIFF
--- a/selfdrive/car/gm/carcontroller.py
+++ b/selfdrive/car/gm/carcontroller.py
@@ -20,6 +20,7 @@ class CarController:
     self.apply_brake = 0
     self.frame = 0
     self.last_button_frame = 0
+    self.cancel_counter = 0
 
     self.lka_steering_cmd_counter = 0
     self.sent_lka_steering_cmd = False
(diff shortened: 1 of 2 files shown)


sshane:
Went and gathered data of how long it takes the camera to cancel cruise:

```python
Took camera 0.080633675 s to cancel
Took camera 0.091127553 s to cancel
Took camera 0.06993979900000001 s to cancel
Took camera 0.060120651000000004 s to cancel
Took camera 0.089879444 s to cancel
Took camera 0.07975477900000001 s to cancel
Took camera 0.07994696400000001 s to cancel
Took camera 0.079749311 s to cancel
Took camera 0.09012001400000001 s to cancel
Took camera 0.08046362400000001 s to cancel
Took camera 0.039980539 s to cancel
Took camera 0.010068415 s to cancel
Took camera 0.0 s to cancel
Took camera 0.010380286 s to cancel
Took camera 0.020240162000000003 s to cancel
Took camera 0.03019936 s to cancel
Took camera 0.080115608 s to cancel
Took camera 0.030638626000000002 s to cancel
Took camera 0.009931125 s to cancel
Took camera 0.0 s to cancel
Took camera 0.059962209 s to cancel
Took camera 0.079799353 s to cancel
```

sshane:
No more FCW and cruise faults with the old brake threshold!

PR #26200, opened 2022-10-21
updated: add default target branch

incognitojam:


adeebshihadeh:
What’s this for? We don’t want to automatically change your branch.

incognitojam:
Right now updated crashes if you checkout a commit without a branch

adeebshihadeh:
This isn't really the right way to fix this. We should catch these cases and just surface them in the UI.

PR #26203, opened 2022-10-22
UI: fade into wide camera

haraschax:

### DIFF
--- a/selfdrive/locationd/calibrationd.py
+++ b/selfdrive/locationd/calibrationd.py
@@ -66,7 +66,6 @@ class Calibrator:
     # Read saved calibration
     params = Params()
     calibration_params = params.get("CalibrationParams")
-    self.wide_camera = params.get_bool('WideCameraOnly')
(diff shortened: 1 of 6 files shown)


common/params.cc
```diff
@@ -179,7 +179,6 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"UpdaterNewReleaseNotes", CLEAR_ON_MANAGER_START},
     {"Version", PERSISTENT},
     {"VisionRadarToggle", PERSISTENT},
-    {"WideCameraOnly", PERSISTENT},
```
adeebshihadeh, line 182: This param really means "single camera mode", which we still want to support

HaraldSchafer: We want to support this on a C3 or for devices that only have 1 camera? Like webcam and sim? If the latter I think we should set it up differently

adeebshihadeh: we mostly care about it for the latter, but I don't see why C3 should be treated differently

PR #26209, opened 2022-10-23
Cabana: use deque::resize() instead of pop_back in loop

deanlee:


tools/cabana/canmessages.cc
```diff
@@ -71,8 +71,8 @@ void CANMessages::process(QHash<QString, std::deque<CanData>> *messages) {
       msgs = std::move(new_msgs);
     } else {
       msgs.insert(msgs.begin(), std::make_move_iterator(new_msgs.begin()), std::make_move_iterator(new_msgs.end()));
-      while (msgs.size() >= settings.can_msg_log_size) {
```
deanlee, line 74: also fixed a problem here, the correct condition should be (msgs.size() > settings.can_msg_log_size)

PR #26210, opened 2022-10-23
Cabana: Added ability to highlight and resize overlapped signals

deanlee:
Highlight the complete area of ​​a signal on mouse over, even if part of the area are overlapped by other signals.
Overlapped signals can also be dragged to resize after this Pr.

[Kazam_screencast_00005.webm](https://user-images.githubusercontent.com/27770/197387391-04992217-af84-412d-a4b2-ea68f6c09349.webm)


adeebshihadeh:
It would be nice if we could also show a warning for overlapping signals, similar to how we show one for wrong messages size.

PR #26213, opened 2022-10-23
Cabana: limit X-axis range from settings

deanlee:
display max [n] minutes data in chart, default is 3 minutes.

resolve https://github.com/commaai/openpilot/discussions/26091:
>  separate setting for the time range on plots instead of just using the cache size. for example, I have lots of RAM, so I want to keep tons of segments in memory, but this makes the plots unusable without zooming. 180s is a good default for this I think

Screenshot from 2022-10-24 22-19-09


deanlee:
done with the following changes:
1.limit X-axis range from settings.
2.fixed some display issues (e.g, background not refreshed after the line marked moved). 
3.refactored chartview to make it easier to support multiple signals in one chart later.

PR #26227, opened 2022-10-25
quectel no sim GPS support

### LINKED ISSUE #26084: Quectel GNSS: ensure it works without a SIM

gast04:
enable quectel GPS support if no sim present

### DIFF
--- a/selfdrive/sensord/rawgps/rawgpsd.py
+++ b/selfdrive/sensord/rawgps/rawgpsd.py
@@ -5,7 +5,7 @@ import signal
 import itertools
 import math
 import time
-import subprocess as sp
+import subprocess
 from typing import NoReturn
 from struct import unpack_from, calcsize, pack
 
@@ -87,12 +87,12 @@ def try_setup_logs(diag, log_types):
   else:
     raise Exception(f"setup logs failed, {log_types=}")
 
-def mmcli(cmd: str) -> None:
+def at_cmd(cmd: str) -> None:
   for _ in range(5):
     try:
-      sp.check_call(f"mmcli -m any --timeout 30 {cmd}", shell=True)
+      subprocess.check_call(f"mmcli -m any --timeout 30 --command='{cmd}'", shell=True)
       break
-    except sp.CalledProcessError:
+    except subprocess.CalledProcessError:
       cloudlog.exception("rawgps.mmcli_command_failed")
   else:
     raise Exception(f"failed to execute mmcli command {cmd=}")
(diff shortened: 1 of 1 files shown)


selfdrive/sensord/rawgps/rawgpsd.py
```diff
@@ -5,7 +5,7 @@
 import itertools
 import math
 import time
-import subprocess
+import subprocess as sp
```
adeebshihadeh, line 8: I haven't really seen this done in other projects, can you revert this to be consistent with the rest of openpilot? also found this cool site: https://import-as.github.io

selfdrive/sensord/rawgps/rawgpsd.py
```diff
@@ -112,7 +120,11 @@ def setup_quectel(diag: ModemDiag):
   mmcli("--command='AT+QGPSCFG=\"dpoenable\",0'")
   # don't automatically turn on GNSS on powerup
   mmcli("--command='AT+QGPSCFG=\"autogps\",0'")
-  mmcli("--location-enable-gps-raw --location-enable-gps-nmea")
+  mmcli("--command='AT+QGPSSUPLURL=\"supl.google.com:7275\"'")
```
adeebshihadeh, line 123: if we're only doing AT commands now, let's change `mmcli` to take an AT command directly and rename to `at_cmd`

selfdrive/sensord/rawgps/rawgpsd.py
```diff
@@ -90,13 +90,21 @@ def try_setup_logs(diag, log_types):
...
-    except subprocess.CalledProcessError:
+    except sp.CalledProcessError:
       cloudlog.exception("rawgps.mmcli_command_failed")
   else:
     raise Exception(f"failed to execute mmcli command {cmd=}")
 
+
+def gps_enabled() -> bool:
+  try:
+    p = sp.run("mmcli -m any --command=\"AT+QGPS?\"", stdout=sp.PIPE, check=True, shell=True)
```
adeebshihadeh, line 103: use check_output instead

adeebshihadeh (commented):
Is this just a pre-req to simless support or does this already work without a sim? I suspected (and would prefer) we'd have to do this entirely outside of ModemManager, but if this works, this is fine too. We should just confirm it works in all of the following scenarios:
* booted up no sim
* booted up with sim, sim removed while rawgpsd running
* previous scenario, but sim reinserted

selfdrive/sensord/rawgps/rawgpsd.py
```diff
@@ -112,7 +120,11 @@ def setup_quectel(diag: ModemDiag):
   mmcli("--command='AT+QGPSCFG=\"dpoenable\",0'")
   # don't automatically turn on GNSS on powerup
   mmcli("--command='AT+QGPSCFG=\"autogps\",0'")
-  mmcli("--location-enable-gps-raw --location-enable-gps-nmea")
+  mmcli("--command='AT+QGPSSUPLURL=\"supl.google.com:7275\"'")
+
+  mmcli("--command='AT+QGPSCFG=\"outport\",\"usbnmea\"'")
+  if not gps_enabled():
+    mmcli("--command='AT+QGPS=1'")
```
adeebshihadeh, line 127: instead of this, we should check if enabled as the first step of the setup and turn off at that point. I remember something about not taking any configuration until GPS is off in the docs

selfdrive/sensord/rawgps/rawgpsd.py
gast04: my bad, I am just used to use it this way, reverted

gast04:
I have tested the three scenarios, all of those work, what I noticed and what slowed testing down, it seems to be more sensitive to the signal, usually I get valid GPS signal on my desk, not without the sim, only works next to the window

PR #26228, opened 2022-10-25
controlsd: update is_metric live

gregjhogan:
I noticed changing units while driving causes all the numbers displayed to update appropriately, but the speed increment for short/long button press does not change.

This also fixes alerts that show the speed if the units are changed while driving.

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -856,6 +856,8 @@ class Controls:
     start_time = sec_since_boot()
     self.prof.checkpoint("Ratekeeper", ignore=True)
 
+    self.is_metric = Params().get_bool("IsMetric")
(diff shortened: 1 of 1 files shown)


selfdrive/controls/controlsd.py
```diff
@@ -856,6 +856,8 @@ def step(self):
     start_time = sec_since_boot()
     self.prof.checkpoint("Ratekeeper", ignore=True)
 
+    self.is_metric = Params().get_bool("IsMetric")
```
gregjhogan, line 859: Is it a bad idea to read a param at this frequency?

adeebshihadeh: should be fine, but this re-creates the Params object every time

PR #26229, opened 2022-10-25
Hyundai: CAN-FD blindspot signals

adeebshihadeh:


selfdrive/car/hyundai/interface.py
```diff
@@ -311,6 +311,13 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[], experi
     ret.vEgoStarting = 0.1
     ret.startAccel = 2.0
 
+    # *** feature detection ***
+    if candidate in CANFD_CAR:
+      bus = 5 if ret.flags & HyundaiFlags.CANFD_HDA2 else 4
+      ret.enableBsm = 0x1e5 in fingerprint[bus]
```
sshane, line 317: You can check if `cornerRadar` is in the carFw

adeebshihadeh: i'd rather check the addr directly for now

PR #26242, opened 2022-10-26
GM: common torque control feedforward

sshane:


twilsonco:
I wonder how many other makes would benefit from this, which is essentially increasing the FF slope at lower speeds. Here's Sonata, for example

morrislee:
this will be very nice addition, feedforward from tim before mitigated a lot of the inconsistencies between speeds in curves! i am willing to assist in any driving data needed for this on the sierra

sshane:
new PR: https://github.com/commaai/openpilot/pull/26371

PR #26245, opened 2022-10-26
Hyundai : konas (kona, kona_ev, kona_hev) Torque Tuning 

haram-KONA:
Hyundai konas (kona, kona_ev, kona_hev) Torque control parameters tuning & steermax tuning (revert)

When driving on the highway with the current Kona settings and tuning values, 
Undershoot occurs on a sharp curve that appears suddenly, and the torque of the steering wheel is sharp decline. 
In a sharp curve, the torque of the steering wheel is insufficient, and this causes the vehicle to momentarily bounce off the curve.  If this happens at high speed on the highway, it seems quite dangerous.

It seems that tuning is needed to get a more stable and safe driving of the Konas.

Attach the route segments drived after tuning is applied

route segments : 
9a3aa85e86f9c3fc|2022-10-26--13-38-48--13
9a3aa85e86f9c3fc|2022-10-26--13-38-48--14
9a3aa85e86f9c3fc|2022-10-26--13-38-48--15

9a3aa85e86f9c3fc|2022-10-26--13-04-28--23
9a3aa85e86f9c3fc|2022-10-26--13-04-28--24
9a3aa85e86f9c3fc|2022-10-26--13-04-28--25


sshane:
We limited the torque of the Konas as they have much higher torque than most Hyundais in their platform, as the comment describes. We are working to fix the lack of a torque alert soon in this PR: https://github.com/commaai/openpilot/pull/26164

PR #26246, opened 2022-10-26
2021 Hyundai Elantra Torque Values

CruiseBrantley:
**Description**
`selfdrive/car/torque_data/params.yaml` has no data value for the 2021 Hyundai Elantra, instead it is drawn from `substitute.yaml` based on the 2020 Sonata.
Live Torque values have been great, but has reset a few times (unsure of the cause, maybe other fork installation at times) but it takes a long time after reset to reach these values again. As an interim fix until data is added to `params.yaml` I'd like to update the defaults with tested values which are much closer to where live torque takes it.

**Verification**
I've tested these on my car manually and they feel much better then stock until live torque has enough bucket points to enable. They're not exact but they're nice round values in the swing that I've seen from live torque plots. They're definitely closer than where the Sonata has them starting from, but I would take other recommended values as well!
My longest route with 12k totalBucketPoints
`5a4405495d2750ef|2022-10-15--21-30-27`
image


CruiseBrantley:
@nuwandavek Bump! Since you mentioned it may be merged last Friday 😅

nuwandavek:
Hey! I looked into it, the values between your vehicle and some other elantras are pretty different. So I wanted to investigate the differences before merging. Will keep this pr updated!

CruiseBrantley:
Definitely appreciate the closer look! I'm fine with other values, the Sonata ones just aren't the best fit

nuwandavek:
I've set the values to this for now. The route you've posted here is the only route with such low values. Other vehicles from the same platform and your own values from other routes are in the range of what I have set now - which is strange. Try this out, I will also keep an eye out and analyse it as more logs come in.

https://github.com/commaai/openpilot/pull/26344

CruiseBrantley:
> I've set the values to this for now. The route you've posted here is the only route with such low values. Other vehicles from the same platform and your own values from other routes are in the range of what I have set now - which is strange. Try this out, I will also keep an eye out and analyse it as more logs come in.
> 
> #26344

@nuwandavek is my car still an outlier or do they just take a lot of time / totalBucketPoints to get to their eventual goal? My recent routes just look to be *slowly* converging back to these values, although I won't know until much more drive time at this rate. Hopefully it doesn't reset soon as I won't know at what point it flattens out. I'm fine with the values going where they like but resets are the frustration as you lose all the progress on vastly divergent vehicles and have a less optimal tune for a long time.
`5a4405495d2750ef|2022-11-05--16-27-07`
image
The recent routes have felt "fine" but kind of loose steering wise, could be I'm more used to it overcorrecting though
`5a4405495d2750ef|2022-11-05--16-27-07--33`
image

I could update again if I don't hit any resets and the curves start to flatten for a while if it's valuable to you. The drive feels ok, and I'm not complaining - I just prefer as perfect as I can get.

CruiseBrantley:
@nuwandavek Just to follow the thread with a bit more data
`5a4405495d2750ef|2022-11-13--15-25-33`
image
They feel really rock solid here, but wondering if my car's still quite an outlier or if other's just hadn't driven enough for solid data perhaps.

CruiseBrantley:
@nuwandavek Just a bit more data with a lot more driving under it
image
I believe latAccel around 2.75 and friction around .10?

PR #26256, opened 2022-10-27
add gps tests to CI

gast04:
add ublox gps tests to CI


gast04:
GPS setup with limeSDR is too unstable to put in CI, this will stay a manual test

PR #26258, opened 2022-10-27
Chrysler: dynamic min steer speed for LKAS bit to avoid fault

sshane:
Applicable for Ram 1500s with special EPS FW. When booted onroad, if the LKAS bit is set above 32 mph and the truck slows below that, a fault occurs.

What's safe to do:
- Above 32 mph, set and unset the LKAS bit as many times as we want
- Setting the LKAS bit under 32 mph, then keep it high forever

What will cause a fault:
- Setting the bit above 32 mph, then slowing below 32 mph with it high
- Unsetting the bit under 32 mph once it's been set under 32 mph

Therefore, we need a dynamic minimum steering speed. This handles the following cases:
- C3 boots up and truck is > 32 mph, user enables:
  - Truck falls below 32 mph
  - The truck stops steering at 32 mph
  - The user disables under 32 mph
  - LKAS bit rises, steering speed is now 6 mph for the rest of the drive
- C3 boots up and truck is > 32 mph, user does not enable:
  - Truck falls below 32 mph
  - LKAS bit rises, steering speed is now 6 mph for the rest of the drive
- C3 boots up and truck is < 32 mph:
  - LKAS bit rises, steering speed is now 6 mph for the rest of the drive

### DIFF
--- a/selfdrive/car/chrysler/carcontroller.py
+++ b/selfdrive/car/chrysler/carcontroller.py
@@ -1,5 +1,6 @@
 from opendbc.can.packer import CANPacker
 from common.realtime import DT_CTRL
+from common.conversions import Conversions as CV
 from selfdrive.car import apply_toyota_steer_torque_limits
 from selfdrive.car.chrysler.chryslercan import create_lkas_hud, create_lkas_command, create_cruise_buttons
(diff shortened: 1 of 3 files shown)


sshane:
Screenshot from 2022-10-26 22-22-57

PR #26259, opened 2022-10-27
QuectelGPS: add supl server

gast04:
improve gps time to first fix by presetting a supl server


adeebshihadeh (approved):
LGTM if you tested this well.

gast04:
the only thing noticeable is, when used without sim card it sometimes takes a while until it gets a fix, ~2-4min

gregjhogan:
what is the slowest you have observed the time to first fix (from cold start) when you do have internet via prime sim with this change?

gast04:
a fix after a full (forced) reset with `QGPSDEL`, will still take about 5-10min, it strongly depends on the location, a device with assistance data in warm start will deliver measurements < 10s

gast04:
quickly tested in the stress test rack, and always got a fix in max 1min

PR #26261, opened 2022-10-27
[torqued] Update offline values

nuwandavek:
- Use same code path while running live and offline on qlogs
- Update offline (from qlog) values for platforms with enough diverse data


selfdrive/car/torque_data/override.yaml
```diff
@@ -21,7 +21,7 @@ FORD FOCUS 4TH GEN: [.nan, 1.5, .nan]
 COMMA BODY: [.nan, 1000, .nan]
 
 # Totally new cars
-KIA EV6 2022: [3.5, 3.0, 0.0]
```
adeebshihadeh, line 24: should be moved to params.yaml now

selfdrive/car/torque_data/params.yaml
```diff
@@ -43,13 +43,13 @@ KIA K5 2021: [2.405339728085138, 1.460032270828705, 0.11650989850813716]
 KIA NIRO EV 2020: [2.9215954981365337, 2.1500583840260044, 0.09236802474810267]
 KIA SORENTO GT LINE 2018: [2.464854685101844, 1.5335274218367956, 0.12056170567599558]
 KIA STINGER GT2 2018: [2.7499043387418967, 1.849652021986449, 0.12048334239559202]
-LEXUS ES 2019: [2.0203086922726112, 2.134803912579666, 0.12757526789308554]
```
adeebshihadeh, line 46: not a big deal now, but we'll eventually want something to update the max lat accel. it's used in the docs

selfdrive/locationd/torqued.py
```diff
@@ -16,7 +16,9 @@
 HISTORY = 5  # secs
 POINTS_PER_BUCKET = 1500
 MIN_POINTS_TOTAL = 4000
+MIN_POINTS_TOTAL_QLOG = 800
```
adeebshihadeh, line 19: is this a factor derived from the decimation? if so, should use the decimation from services.py

selfdrive/locationd/torqued.py
```diff
@@ -27,6 +29,7 @@
 LAT_ACC_THRESHOLD = 1
 STEER_BUCKET_BOUNDS = [(-0.5, -0.3), (-0.3, -0.2), (-0.2, -0.1), (-0.1, 0), (0, 0.1), (0.1, 0.2), (0.2, 0.3), (0.3, 0.5)]
 MIN_BUCKET_POINTS = [100, 300, 500, 500, 500, 500, 300, 100]
+MIN_BUCKET_POINTS_QLOG = [10, 30, 50, 50, 50, 50, 30, 10]
```
adeebshihadeh, line 32: same here. if not, `MIN_BUCKET_POINTS / 10` is cleaner

selfdrive/car/torque_data/params.yaml
nuwandavek: Yeah, I'll do it in another PR. Ideally it should be the same as the expected max (via max torque).

selfdrive/locationd/torqued.py
nuwandavek: Not really, it was tuned to be good enough for an approximate estimate of the values.

selfdrive/locationd/torqued.py
```diff
@@ -90,9 +93,18 @@ def load_points(self, points):
 
 
 class TorqueEstimator:
-  def __init__(self, CP):
+  def __init__(self, CP, decimated=False):
     self.hist_len = int(HISTORY / DT_MDL)
     self.lag = CP.steerActuatorDelay + .2   # from controlsd
+    self.decimated = decimated
```
adeebshihadeh, line 99: this is unused

adeebshihadeh (approved):

adeebshihadeh:
Do we not have enough data for the rest of the cars? Or is this only from rlogs and we're still running it on the qlogs?

nuwandavek:
> Do we not have enough data for the rest of the cars? Or is this only from rlogs and we're still running it on the qlogs?

No, the updated values were on the qlogs. Other platforms do not have enough  long routes from enough unique dongle ids (running op release/master, not a fork). I am slightly hesitant to change the values drastically based on 1/2 dongle ids. If the values were close enough to begin with, it wouldn't matter anyway.

PR #26263, opened 2022-10-27
UI: show finish registration widget if not paired

### LINKED ISSUE #26262: No barcode two pair a new device
### Describe the bug

Hi everyone I'm having problem with my comma , too is not showing barcode in order to pair the device. Any help or advice would be much appreciated

When trying to install

### openpilot version

0.8.13.1

### Additional info

Hi everyone I'm having problem with my comma , too is not showing barcode in order to pair the device. Any help or advice would be much appreciated

deanlee:
resolve https://github.com/commaai/openpilot/issues/26262

### DIFF
--- a/common/params.cc
+++ b/common/params.cc
@@ -100,6 +100,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"ControlsReady", CLEAR_ON_MANAGER_START | CLEAR_ON_IGNITION_ON},
     {"CurrentRoute", CLEAR_ON_MANAGER_START | CLEAR_ON_IGNITION_ON},
     {"DashcamOverride", PERSISTENT},
+    {"DevicePaired", PERSISTENT},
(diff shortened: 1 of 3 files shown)


incognitojam:
what case did that user hit where the registration didn't show?

deanlee:
> do you know how to reproduce the case where the registration didn't show?

no network or "NOT FOUND" returned from url '/v1/devices/dongleid/owner'

Jameselectronis:
I have bought it used from ebay.  Don't know anything about previous owners  registration

Jameselectronis:
> resolve #26262

I am not about how to do that. Any tips or instructions would much appreciated

Jameselectronis:
> 

I have connected to my wifi if that's the case. but still showing that only. no bar code!!

incognitojam:
I think if we did set a param for `DevicePaired` it wouldn't have helped in the no network scenario as the device would just remember that it used to be paired. Closing this now though as it seems it was user error not a bug.

PR #26266, opened 2022-10-27
UI: reduce frame drops.

deanlee:
The drawing freq of UI is not stable compared with the vipc receive Freq. lot's of frames will be dropped If we only paint the last one.
use  `deque<> frames` as a buffer to sync between UI thread and vipc thread.  this can reduce many frame drops.


deanlee:
if the update is called without new frames(this may occur if update() is not triggered by vipc thread), paintGL will the  previouse uploaded texture to avoid black screen

deanlee:
lgtm,no frame drops in my testing for 15 minutes.

PR #26267, opened 2022-10-27
Silent below steer speed alert

sshane:


selfdrive/car/chrysler/carcontroller.py
```diff
@@ -57,6 +57,8 @@ def update(self, CC, CS):
       elif self.CP.carFingerprint in RAM_CARS:
         if CS.out.vEgo < (self.CP.minSteerSpeed - 0.5):
           lkas_control_bit = False
+      if CS.out.steerFaultTemporary or CS.out.steerFaultPermanent:
+        lkas_control_bit = False
```
sshane, line 61:
```suggestion
```

selfdrive/controls/lib/events.py
```diff
@@ -238,7 +238,7 @@ def below_steer_speed_alert(CP: car.CarParams, CS: car.CarState, sm: messaging.S
     f"Steer Unavailable Below {get_display_speed(CP.minSteerSpeed, metric)}",
     "",
     AlertStatus.userPrompt, AlertSize.small,
-    Priority.MID, VisualAlert.steerRequired, AudibleAlert.prompt, 0.4)
+    Priority.MID, VisualAlert.steerRequired, AudibleAlert.none, 0.4)
```
sshane, line 241: Might want to only mute sub-ten as discussed in Discord

adeebshihadeh:
I agree this needs to be fixed, but I'm not really sure this is better. One of the biggest complaints with the EV6 stock system is that the steering cuts out without warning. Let's keep this open for now, and I'll think about another solution while working on some of the planned UI updates.

sshane:
Think an issue is more appropriate for this

PR #26268, opened 2022-10-27
Ram 1500: dynamic min steer speed

sshane:
Avoids a fault when OP boots while Ram 1500 with steer to zero is above 32 mph by setting min steer speed to 32 mph


sshane:
@adeebshihadeh should I work on being able to alter the car params from the car state so this isn't as messy? Do we have a need for that anywhere else?

adeebshihadeh:
> @adeebshihadeh should I work on being able to alter the car params from the car state so this isn't as messy? Do we have a need for that anywhere else?

No, carParams shouldn't be dynamically updated.

sshane:
should be done generically first

PR #26269, opened 2022-10-27
controlsd: steer in old preEnable state

adeebshihadeh:
preEnabled is now completely replaced by the overriding state.


sshane:
Might want to leave the preEnable state in case we want to enforce no acceleration actuation at a stop

adeebshihadeh:
> Might want to leave the preEnable state and event types in case we want to enforce no acceleration actuation at a stop, but that would be the last remaining use

Yeah, let's keep it for that. Let's do that soon too.

PR #26273, opened 2022-10-28
controlsd: preEnable on brake at a standstill

sshane:
diff is adding a pre enable event at a standstill 
Screenshot from 2022-11-29 15-06-49


selfdrive/controls/controlsd.py
```diff
@@ -253,6 +253,9 @@ def update_events(self, CS):
       (CS.brakePressed and (not self.CS_prev.brakePressed or not CS.standstill)):
       self.events.add(EventName.pedalPressed)
 
+    if CS.brakePressed and CS.standstill:
+      self.events.add(EventName.pedalPressedPreEnable)
```
adeebshihadeh, line 257: let's change the name to something like preEnableStandstill

sshane:
Has a minor problem with the Bolt. It allows enabling at a standstill with brake, despite the min enable speed of 3 mph without brake. However the noEntry upon releasing the brake after enabling takes precendence over transitioning from preEnabled to enabled and we disengage

adeebshihadeh:
> either need to know the state of OP in the car interface or don't consider noEntries after we're preEnabled. Which overriding already does, so that might make the most sense. Cruise is already activated on most cars in this state, and noEntry is really for first enabling. @adeebshihadeh that sound good?

yeah that makes more sense

selfdrive/controls/controlsd.py
```diff
@@ -579,7 +582,7 @@ def state_control(self, CS):
     # Check which actuators can be enabled
     CC.latActive = self.active and not CS.steerFaultTemporary and not CS.steerFaultPermanent and \
                    CS.vEgo > self.CP.minSteerSpeed and not CS.standstill
-    CC.longActive = self.active and not self.events.any(ET.OVERRIDE_LONGITUDINAL) and self.CP.openpilotLongitudinalControl
+    CC.longActive = self.enabled and not self.events.any(ET.OVERRIDE_LONGITUDINAL) and self.CP.openpilotLongitudinalControl
```
sshane, line 585: changed to enabled (which includes preEnabled) so that we can apply brake while pre-enabling at a standstill to keep the vehicle stopped

adeebshihadeh:
Should we also remove the noTarget event?

sshane:
Was going to do in another PR as it has some implications we might want to think about. I think we restricted to only enabling behind a lead (or now when e2e wants to stay stopped) because it maybe makes it harder to do unsafe things?

PR #26274, opened 2022-10-28
Toyota: lower stopping speed

sshane:
the current one means that we often are stuck stopped for too long when the lead slowly inches up in traffic, and a harsh stop on some occasions


TKassis:
Is this related to very slow acceleration (Toyota Corolla) when the lead car accelerates?

krkeegan:
> Is this related to very slow acceleration (Toyota Corolla) when the lead car accelerates?

In my experience playing with these values, it might improve that just slightly, but it may not be enough for you to even notice.

sshane:
Yeah this isn't a fix for any acceleration issues, just abrupt stopping and a delay to move when the lead is creeping

selfdrive/car/toyota/interface.py
```diff
@@ -216,7 +216,9 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[], experi
       tune.kiBP = [0., 5., 12., 20., 27.]
       tune.kiV = [.35, .23, .20, .17, .1]
       if candidate in TSS2_CAR:
-        ret.stoppingDecelRate = 0.3 # reach stopping target smoothly
+        ret.vEgoStopping = 0.1
+        ret.vEgoStarting = 0.1
```
sshane, line 220:
```suggestion
        ret.vEgoStopping = 0.25
        ret.vEgoStarting = 0.25
```

PR #26276, opened 2022-10-28
definition changes - removing unused signal

ioniq5-cz:
open another PR to correspond with the definition changes in this PR https://github.com/commaai/opendbc/pull/729 , specifically removing CRUISE_INFO | NEW_SIGNAL_4 in carstate.py.


adeebshihadeh:
Thanks, but already done in #26277

PR #26282, opened 2022-10-30
Cabana: add save DBC dialog

deanlee:
Screenshot from 2022-10-30 14-31-28


adeebshihadeh:
I think adding a "copy to clipboard" button and a "save as" button to this dialog popup can complete the save DBC feature.

deanlee:
updated:
Screenshot from 2022-10-30 15-26-13

deanlee:
need test case for this function

deanlee:
fixed precision problem:
> tools/cabana/tests/test_cabana.cc:30: FAILED:
  REQUIRE( sig[j].factor == new_sig[j].factor )
with expansion:
  0.0049015947 == 0.004901595

PR #26286, opened 2022-10-30
Cabana: support multiple instances

deanlee:


tools/cabana/cabana.cc
```diff
@@ -22,12 +24,22 @@ int main(int argc, char *argv[]) {
     cmd_parser.showHelp();
   }
 
+  QString uuid =  QUuid::createUuid().toString(QUuid::WithoutBraces);
```
adeebshihadeh, line 27: let's put this helper in `common/`

deanlee: do you mean cabana/common or somewhere else such as /ui/qt/util.cc?

adeebshihadeh: the root common/

deanlee:
I'm not sure if it's appropriate to put it in qtutil.h like this

tools/cabana/cabana.cc
deanlee: I have reverted https://github.com/commaai/openpilot/pull/26286/commits/19ef28945e8647724bbfc6c8469a664a58c09f16 , how about keep this pr as it is ,and move to common next time?

adeebshihadeh: sure, but I meant make a common function to setup a prefix environment

PR #26226, opened 2022-10-25
test_models: run on route list

sshane:
Includes one segment from every dongle id.

Catching some bugs:

```python
FAILED run_test_models.py::TestCarModel_11_NISSAN_LEAF_2018::test_panda_safety_carstate - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'standstill': 14}
FAILED run_test_models.py::TestCarModel_26_HYUNDAI_SONATA_2020::test_panda_safety_carstate - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'cruiseState': 2}
FAILED run_test_models.py::TestCarModel_100_TOYOTA_HIGHLANDER_2017::test_panda_safety_carstate - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'cruiseState': 1}
FAILED run_test_models.py::TestCarModel_174_KIA_EV6_2022::test_car_interface - AssertionError: 5798 != 0
FAILED run_test_models.py::TestCarModel_201_TOYOTA_PRIUS_2017::test_panda_safety_carstate - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'cruiseState': 3}
FAILED run_test_models.py::TestCarModel_283_NISSAN_ROGUE_2019::test_panda_safety_rx_valid - AssertionError: False is not true
FAILED run_test_models.py::TestCarModel_283_NISSAN_ROGUE_2019::test_car_interface - AssertionError: 348 != 0
FAILED run_test_models.py::TestCarModel_321_HONDA_CIVIC_2016::test_panda_safety_carstate - AssertionError: 1 is not false : panda safety doesn't agree with openpilot: {'controlsAllowed': 1}
FAILED run_test_models.py::TestCarModel_620_TOYOTA_PRIUS_2017::test_car_interface - AssertionError: 1881 != 0
FAILED run_test_models.py::TestCarModel_620_TOYOTA_PRIUS_2017::test_panda_safety_rx_valid - AssertionError: False is not true
FAILED run_test_models.py::TestCarModel_633_HONDA_ODYSSEY_2018::test_car_interface - AssertionError: 5799 != 0
FAILED run_test_models.py::TestCarModel_633_HONDA_ODYSSEY_2018::test_panda_safety_carstate - AssertionError: 2 is not false : panda safety doesn't agree with openpilot: {'brakePressed': 1, 'mainOn': 544}
[...]

PR #26296, opened 2022-10-31
test_models: no fuzzy fingerprinting routes

sshane:
Some of the routes in https://github.com/commaai/openpilot/pull/26226 are fuzzy fingerprinting as other cars causing fake addr check failures.

`7e4c14db18b441c3|2022-07-27--08-48-26` for example which is a Chinese Odyssey, but fingerprints as a Fit and Odyssey on a few occasions with dirty=False


sshane:
doing this in the route list script, a lot of our routes are fixed or fuzzy. we should just be more careful about adding these routes

PR #26297, opened 2022-11-01
Map: save gps bearing

gast04:
store last GPS bearing in params, to get better experience on map start


adeebshihadeh:
Let's get the magnetometer working instead.

PR #26304, opened 2022-11-01
test_models: speed up test

sshane:
Speeds up 214 runs through this function from 73 seconds to 61 seconds. With a baseline of ~30 seconds to do the setUp and tearDown, this now takes about 70% of the previous time taken for just this function

- 214 function runs (`FILEREADER_CACHE=1`, `-j50`)
  - Before: 73s
  - After: 61ss
  - Baseline (only setUpClass, setUp, tearDown): 30s
  - Speedup: ~27%
- 691 function runs (`FILEREADER_CACHE=1`, `-j50`)
  - Before: 190s
  - After: 153s
  - Baseline (only setUpClass, setUp, tearDown): 70s
  - Speedup: ~30%

### DIFF
--- a/selfdrive/car/tests/test_models.py
+++ b/selfdrive/car/tests/test_models.py
@@ -214,10 +214,10 @@ class TestCarModelBase(unittest.TestCase):
 
     # warm up pass, as initial states may be different
     for can in self.can_msgs[:300]:
+      self.CI.update(CC, (can.as_builder().to_bytes(), ))
       for msg in can_capnp_to_can_list(can.can, src_filter=range(64)):
         to_send = package_can_msg(msg)
         self.safety.safety_rx_hook(to_send)
-        self.CI.update(CC, (can_list_to_can_capnp([msg, ]), ))
 
     if not self.CP.pcmCruise:
       self.safety.set_controls_allowed(0)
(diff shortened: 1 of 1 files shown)


selfdrive/car/tests/test_models.py
```diff
@@ -214,10 +214,10 @@ def test_panda_safety_carstate(self):
 
     # warm up pass, as initial states may be different
     for can in self.can_msgs[:300]:
+      self.CI.update(CC, (can.as_builder().to_bytes(), ))
```
sshane, line 217: @adeebshihadeh do you remember why we passed in a single CAN message at a time here?

selfdrive/car/tests/test_models.py
```diff
@@ -214,10 +214,10 @@ def test_panda_safety_carstate(self):
...
       for msg in can_capnp_to_can_list(can.can, src_filter=range(64)):
         to_send = package_can_msg(msg)
         self.safety.safety_rx_hook(to_send)
-        self.CI.update(CC, (can_list_to_can_capnp([msg, ]), ))
```
sshane, line 220: @adeebshihadeh do you remember why we passed in a single CAN message at a time here?

adeebshihadeh: don't think it's important for the warm up pass

PR #26308, opened 2022-11-01
joystick: use long controller

sshane:
Puts more longcontrol behavior in the loop, such as PID gains, long control states like stopping/starting, prevent_overshoot

Bolt with joystick before this is smooth as we can modulate the accel smoothly, however the PID gains make it much more unstable which we should hit with joystick mode

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -16,7 +16,7 @@ from system.version import is_tested_branch, get_short_branch
 from selfdrive.boardd.boardd import can_list_to_can_capnp
 from selfdrive.car.car_helpers import get_car, get_startup_event, get_one_can
 from selfdrive.controls.lib.lateral_planner import CAMERA_OFFSET
-from selfdrive.controls.lib.drive_helpers import V_CRUISE_INITIAL, update_v_cruise, initialize_v_cruise, get_lag_adjusted_curvature, CONTROL_N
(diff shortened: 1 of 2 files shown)


selfdrive/controls/controlsd.py
```diff
@@ -616,12 +616,20 @@ def state_control(self, CS):
...
       self.LoC.reset(v_pid=CS.vEgo)
 
-    if not self.joystick_mode:
-      # accel PID loop
-      pid_accel_limits = self.CI.get_pid_accel_limits(self.CP, CS.vEgo, self.v_cruise_kph * CV.KPH_TO_MS)
+    # accel PID loop
+    if self.joystick_mode:
+      joystick_accel = 4.0 * clip(self.sm['testJoystick'].axes[0], -1, 1) if self.sm.rcv_frame['testJoystick'] > 0 else 0.0
+      long_plan = log.LongitudinalPlan(accels=[joystick_accel] * CONTROL_N,
+                                       speeds=interpolate_mpc_to_traj(LongitudinalMpc.extrapolate_lead(0, CS.vEgo, joystick_accel, 0).T[1]))
```
gast04, line 623:
```suggestion
                                       speeds = interpolate_mpc_to_traj(LongitudinalMpc.extrapolate_lead(0, CS.vEgo, joystick_accel, 0).T[1]))
```

sshane: usually keyword args have no space

gast04: didnt know :+1:

sshane:
mostly feels like the old joystick now with the missing long deadzone added. not sure what the point is any more

PR #26319, opened 2022-11-02
manager: watchdog restart is non-blocking

sshane:
If a watched process dies right as we want to go offroad, we can sit onroad while we wait for the process to be stopped, logging can errors that can extend across segments


sshane:
the start in this case is now blocking, need to fix that

sshane:
not clear how to fix this cleanly, closing for now

PR #26320, opened 2022-11-02
Toyota: only set gas pressed when seen

sshane:
Since this signal checks a signal is false, we don't know the difference between false and default. If no PCM_CRUISE message comes in for the very first compare and we compare panda and openpilot's gas state, openpilot will think the gas is pressed while panda doesn't.


sshane:
Not sure if we've ever hit this without a CAN errors. Find this and re-open if this is ever an issue

adeebshihadeh:
This is also what the initializing time is for when running live. For test models, we have the warmup time.

sshane:
In the case I saw, there were no PCM messages in the warmup period

PR #26324, opened 2022-11-02
PJ: add longitudinal layout

sshane:
just a first draft, feel free to improve upon it. i just find myself making this layout over and over when looking at long stuff


sshane:
Screenshot from 2022-11-02 01-54-00

PR #26326, opened 2022-11-02
Cabana: add y-axis padding

deanlee:
resolve https://github.com/commaai/openpilot/discussions/26091#discussioncomment-4025302

**No padding:**
Screenshot from 2022-11-02 20-27-05

**After add padding:** 
Screenshot from 2022-11-02 20-26-18


sshane:
thanks @deanlee!

PR #26338, opened 2022-11-03
[torqued] Fix high speed oscillations

nuwandavek:
Changes
- Reduce low speed factor during high speeds (scales down the error)
image

- Friction: Currently full friction is applied at `0.2` error (~4% of instances). Increase to `0.3` (~2% of cases) to reduce oscillations at higher errors.


rjsmith1999:
FWIW, this change significantly improves torque control performance for my Honda Passport. 
Significantly less oscillation. 

Looking forward to seeing torqued rolled out to more brands!

PR #25434, opened 2022-08-14
Hyundai CAN-FD: support 2023 Kia Sportage & 2022 Hyundai Santa Cruz

sunnyhaibin:
**Checklist**
- [x] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)

**2022 Hyundai Santa Cruz**
- [x] route with openpilot:
  - stock longitudinal: `f0709d2bc6ca451f|2022-10-15--08-13-54`
  - openpilot longitudinal: `f0709d2bc6ca451f|2022-10-15--10-39-32`
- [x] harness type: [Hyundai N](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)

**2023 Kia Sportage**
- [x] route with openpilot:
  - stock longitudinal: `192283cdbb7a58c2|2022-10-15--00-57-57`
  - openpilot longitudinal: `192283cdbb7a58c2|2022-10-15--01-43-18`, `965a780ebeeed150|2022-10-16--20-50-38`
- [x] harness type: [Hyundai N](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)

**2022 Genesis GV70**
- [x] route with openpilot: `ca4de5b12321bd98|2022-10-18--21-15-59`
- [x] harness type: [Hyundai L](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)

**Physical setup**
- Top port behind comma three
  - comma Power <-> First Harness Box <-> comma three USB C/Right-Angle C <-> comma three
- Bottom port behind comma three
  - Camera Harness <-> Second Harness Box <-> USB C/C 3.1 <-> Red Panda <-> USB A/A <-> USB A to C **OTG Adapter** <-> comma three

**openpilot Longitudinal Control**
* openpilot longitudinal control support for 2022 Hyundai Santa Cruz (ICE) and 2023 Kia Sportage (ICE), without radar disable. Message `0x1a0 | CRUISE_INFO` is on bus 6, we block the message from bus 6 and send our own longitudinal commands to `0x1a0 | CRUISE_INFO` bus 4.

**Prerequisites**
- https://github.com/commaai/openpilot/pull/25276
- https://github.com/commaai/panda/pull/999

**Requires**
[...]

PR #26345, opened 2022-11-03
Hyundai: longitudinal support for all CAN-FD EV and Hybrids

hoomoose:
**Description**
OP long (and e2e) is currently enabled for HDA2 Can-fd HKG cars but not HDA1. <s>I have picked the modifications from https://github.com/commaai/openpilot/pull/25434 (thank you Sunnyhaibin!) that only relate to enabling OP Long. This code should not affect HDA2 variants of cars that have both HDA1 and HDA2 variants because the flag for labelling a car as "CANFD_CAMERA_SCC" is placed after an "Else" statement following cars flagged as HDA2. Therefore, an HDA2 car can never receive the "CANFD_CAMERA_SCC" flag. I have added EV6 and Ioniq 5 to the CAMERA_SCC_CAR car list because I can verify this with my HDA1 EV6, but the referenced PR shows that other Can-fd cars have this tested as well.</s>

11/8/22 update:
Based on latest code, all that is changed is 1) set OP long to true for can-fd cars and 2) and "if hda2" in front of sending adrv messages, 3) two commits excluding cars flagged as CANFD_CAMERA_SCC from having radar disabled.

**Prerequisite**
This requires a Panda update that does the following: 
1. add in ecu detection for HDA1 cars with SCC info on bus 0
2. add an hda2 requirement for checking for ADAS ecu
3. add 0x1a0 through the fwd hook if not hda2
4. remove hda2 requirement for OP long

see: https://github.com/commaai/panda/pull/1128

**Verification**
HDA1 EV6
Route: 68d6a96e703c00c9|2022-11-03--13-35-26


hoomoose:
With the HKG GV70 PR now committed, I have added two commits excluding  cars flagged as CANFD_CAMERA_SCC from having radar disabled.

hoomoose:
With latest change.
Route:  68d6a96e703c00c9|2022-11-08--16-46-18

adeebshihadeh:
Thanks, but this is blocked on refactoring the panda safety address checks. Once that's done, it'll be easy to support HDA1 long.

adeebshihadeh:
Looks like only ICE CAN-FD will be blocked on that refactor. We can get this in the meantime.

ebg1223:
@adeebshihadeh just curious, is ICE CAN-FD blocking on these PRs: #26090 #1107 ? or is this a separate issue? Just trying to see if I can contribute, but want to make sure i'm looking in the right direction.

adeebshihadeh:
> @adeebshihadeh just curious, is ICE CAN-FD blocking on these PRs: #26090 [commaai/panda#1107](https://github.com/commaai/panda/pull/1107) ? or is this a separate issue? Just trying to see if I can contribute, but want to make sure i'm looking in the right direction. For ICE CAN-FD cars using radar cruise, it would be the same as the hybrid / evs correct?

No, we don't support long for any CAN-FD models yet. Unfortunately, there's not much to contribute in this case. We just need to do the refactor ourselves.

PR #26347, opened 2022-11-03
CI: set OX frame sync tolerance

ZwX1616:


system/camerad/test/test_camerad.py
```diff
@@ -4,12 +4,14 @@
...
 import cereal.messaging as messaging
+from cereal import log
 from cereal.services import service_list
 from selfdrive.manager.process_config import managed_processes
 from system.hardware import TICI
 
 TEST_TIMESPAN = 30
-LAG_FRAME_TOLERANCE = 0.5 # ms
+LAG_FRAME_TOLERANCE = {log.FrameData.ImageSensor.ar0321: 0.5, # ms
+                                              log.FrameData.ImageSensor.ox03c10: 1.0} # more tolerance due to passive sync
```
adeebshihadeh, line 14: can we make this extra tolerance more specific? we expect them to be synced much tighter except when it's in phase with the internal sensor sync

ZwX1616: updated comment to better explain how syncing works on each

PR #26352, opened 2022-11-03
locationd: add gps sanity check for quectel gps

gast04:
quectel gps verticalAccuracy is clipped to 500, and flags is always 1, so invalid values would pass, causing location
jumps

also the quectel gps returns values in a 1hz rate which causes gpsOK flickering due to the time limit of 1s,
as was seen in: `be13974ffe78aa72|2022-11-02--19-02-26--18`

### DIFF
--- a/selfdrive/locationd/locationd.cc
+++ b/selfdrive/locationd/locationd.cc
@@ -267,19 +267,23 @@ void Localizer::handle_gps(double current_time, const cereal::GpsLocationData::R
   bool gps_invalid_flag = (log.getFlags() % 2 == 0);
   bool gps_unreasonable = (Vector2d(log.getAccuracy(), log.getVerticalAccuracy()).norm() >= SANE_GPS_UNCERTAINTY);
   bool gps_accuracy_insane = ((log.getVerticalAccuracy() <= 0) || (log.getSpeedAccuracy() <= 0) || (log.getBearingAccuracyDeg() <= 0));
-  bool gps_accuracy_insane_quectel = log.getVerticalAccuracy() == 500;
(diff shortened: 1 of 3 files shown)


selfdrive/locationd/locationd.cc
```diff
@@ -270,15 +270,20 @@ void Localizer::handle_gps(double current_time, const cereal::GpsLocationData::R
   bool gps_lat_lng_alt_insane = ((std::abs(log.getLatitude()) > 90) || (std::abs(log.getLongitude()) > 180) || (std::abs(log.getAltitude()) > ALTITUDE_SANITY_CHECK));
   bool gps_vel_insane = (floatlist2vector(log.getVNED()).norm() > TRANS_SANITY_CHECK);
 
-  if (gps_invalid_flag || gps_unreasonable || gps_accuracy_insane || gps_lat_lng_alt_insane || gps_vel_insane) {
+  // quectel gps verticalAccuracy is clipped to 500
+  bool gps_accuracy_insane_quectel = false;
+  if (!Params().getBool("UbloxAvailable", true)) {
```
adeebshihadeh, line 275: this should only be read once on startup

selfdrive/locationd/locationd.cc
```diff
@@ -498,8 +504,10 @@ void Localizer::determine_gps_mode(double current_time) {
 }
 
 int Localizer::locationd_thread() {
+
+  ublox_available = Params().getBool("UbloxAvailable", true);
```
sshane, line 508:
```suggestion
  ublox_available = Params().getBool("UbloxAvailable", true);

```

PR #26354, opened 2022-11-04
ui: fix disappearing path in wide cam mode

sshane:
Before/after: https://imgur.com/a/pFcE23I

The function that converts the 3D xyz points into xy screen coords has a check to make sure all points fall within the screen boundaries + some margin (so that there's no gaps on the bottom or top).

https://github.com/commaai/openpilot/blob/9a8c7f2453bcb917eb428704c1733f28e5d42d11/selfdrive/ui/ui.cc#L21-L36

In narrow mode, this filters out points that are behind the camera (model predicts x positions upwards of -2 meters sometimes) as the screen coordinates are really high. However, in wide cam mode, these coords are not so high, causing the first few polygon xy points to be above the screen.

Then since `allow_invert=False` with the path, we skip most (if not all) points, as the points will of course all be below the top of the screen.

https://github.com/commaai/openpilot/blob/9a8c7f2453bcb917eb428704c1733f28e5d42d11/selfdrive/ui/ui.cc#L61-L75

Now we clip the incoming model x to under the camera. In wide cam mode, xyz points aren't within the screen boundaries + margin until around 1.3 m, and narrow is even further.

### DIFF
--- a/selfdrive/ui/qt/onroad.cc
+++ b/selfdrive/ui/qt/onroad.cc
@@ -173,9 +173,8 @@ void OnroadAlerts::paintEvent(QPaintEvent *event) {
 AnnotatedCameraWidget::AnnotatedCameraWidget(VisionStreamType type, QWidget* parent) : fps_filter(UI_FREQ, 3, 1. / UI_FREQ), CameraWidget("camerad", type, true, parent) {
   pm = std::make_unique<PubMaster, const std::initializer_list<const char *>>({"uiDebug"});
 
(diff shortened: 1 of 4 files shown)


sshane:
For example, here's the first point drawn in one of these frames where the path disappears: https://pastebin.com/RBJGxu7Z

x is -1.784, but the screen point is `QPointF(1673.79,-339.519)` which is within the margin, so it's drawn. Then the subsequent points are of course greater than the first point's x and are skipped

PR #26359, opened 2022-11-04
VW MQB: Enable experimental long

jyoung8607:
Throw the switch!


selfdrive/car/volkswagen/interface.py
```diff
@@ -27,12 +27,14 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, expe
     ret.carName = "volkswagen"
     ret.radarOffCan = True
 
+    use_off_car_defaults = len(fingerprint[0]) == 0  # Pick sensible carParams during offline doc generation/CI jobs
```
adeebshihadeh, line 30: @sshane let's add something to make this more explicit

PR #26145, opened 2022-10-19
Hyundai: Add FW for 2023 Ioniq 5 HDA2

sunnyhaibin:
Add firmware for the 2023 Hyundai Ioniq 5 HDA2. Expand supported model-year range.

Route ID: `8a53aa70d88313f1|2022-11-01--15-17-19`

Thanks to community Ioniq 5 owner jdtram.

PR #26376, opened 2022-11-05
Hyundai: update Ioniq 5 HDA I supported model years

sunnyhaibin:
Add firmware for the 2023 Hyundai Ioniq 5 SEL (HDA II). Expand supported model-year range.

Route ID: `eda20e76a555a83d|2022-11-01--15-40-00`

Thanks to community Ioniq 5 HDA II CAN-FD owner Slavka89.


selfdrive/car/hyundai/values.py
```diff
@@ -143,7 +143,7 @@ class HyundaiCarInfo(CarInfo):
   CAR.SONATA_HYBRID: HyundaiCarInfo("Hyundai Sonata Hybrid 2020-22", "All", harness=Harness.hyundai_a),
   CAR.IONIQ_5: [
     HyundaiCarInfo("Hyundai Ioniq 5 (without HDA II) 2022" , "Highway Driving Assist", harness=Harness.hyundai_k),
-    HyundaiCarInfo("Hyundai Ioniq 5 (with HDA II) 2022", "Highway Driving Assist II", harness=Harness.hyundai_q),
+    HyundaiCarInfo("Hyundai Ioniq 5 (with HDA II) 2022-23", "Highway Driving Assist II", harness=Harness.hyundai_q),
```
adeebshihadeh, line 146: I think it's safe to assume the compatibility is the same for HDA2 and non-HDA2, so let's bump that one too

sshane:
You already merged this FW in this PR: https://github.com/commaai/openpilot/pull/26145 so this is only a docs update now

PR #26378, opened 2022-11-05
onroad ui: fix opacity affecting other drawn icons

sshane:
If you call `drawIcon` with a non-100% opacity, future calls will inherit that opacity. We only set opacity on the painter drawing the last icon so this wasn't an issue before

### DIFF
--- a/selfdrive/ui/qt/onroad.cc
+++ b/selfdrive/ui/qt/onroad.cc
@@ -404,11 +404,13 @@ void AnnotatedCameraWidget::drawText(QPainter &p, int x, int y, const QString &t
 }
 
 void AnnotatedCameraWidget::drawIcon(QPainter &p, int x, int y, QPixmap &img, QBrush bg, float opacity) {
+  p.save();
   p.setPen(Qt::NoPen);
   p.setBrush(bg);
   p.drawEllipse(x - radius / 2, y - radius / 2, radius, radius);
   p.setOpacity(opacity);
   p.drawPixmap(x - img_size / 2, y - img_size / 2, img);
+  p.restore();
 }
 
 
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/onroad.cc
```diff
@@ -404,11 +404,13 @@ void AnnotatedCameraWidget::drawText(QPainter &p, int x, int y, const QString &t
 }
 
 void AnnotatedCameraWidget::drawIcon(QPainter &p, int x, int y, QPixmap &img, QBrush bg, float opacity) {
+  p.save();
   p.setPen(Qt::NoPen);
   p.setBrush(bg);
   p.drawEllipse(x - radius / 2, y - radius / 2, radius, radius);
   p.setOpacity(opacity);
   p.drawPixmap(x - img_size / 2, y - img_size / 2, img);
+  p.restore();
```
sshane, line 413: @adeebshihadeh setting opacity back to 100% will also solve this, not sure which is preferred?

adeebshihadeh: both solutions depend on the state of the painter that's passed in. instead, the opacity should be set before drawEllipse

selfdrive/ui/qt/onroad.cc
```diff
@@ -404,6 +404,7 @@ void AnnotatedCameraWidget::drawText(QPainter &p, int x, int y, const QString &t
 }
 
 void AnnotatedCameraWidget::drawIcon(QPainter &p, int x, int y, QPixmap &img, QBrush bg, float opacity) {
+  p.setOpacity(1.0);  // bg dictates opacity
```
sshane, line 407:
```suggestion
  p.setOpacity(1.0);  // bg dictates opacity of ellipse
```

PR #26346, opened 2022-11-03
UI: split lat / long icons

adeebshihadeh:

PR #26379, opened 2022-11-05
ui: onroad e2e long toggle

sshane:
WIP, just to see what it's like onroad. Includes code from https://github.com/commaai/openpilot/pull/26346


sshane:
Closing to preserve history https://github.com/commaai/openpilot/pull/26405

PR #26381, opened 2022-11-05
Subaru: add footnote to docs for international models

jprous:
Updated Subaru required specs in Cars.md

Although all Foresters in the USA may be fitted with Eyesight & LKAS per default, this is not the case in the rest of the world. 

Also, even if the Forester is fitted with EyeSight, it may NOT be fitted with LKAS, which will be restrictive for OpenPilot to steer.

Regards,
Jean-Pierre


incognitojam:
You should change the package in [`values.py`](https://github.com/commaai/openpilot/blob/master/selfdrive/car/subaru/values.py#L55) and run `./selfdrive/car/docs.py` to update `CARS.md`

adeebshihadeh (changes requested):
The documentation is in a bit of a confusing state right now since we primarily target the US market, but it's not exactly clear when we don't. In order to make it as unambiguous for US market users, I'd still like this to show "All" for them, but we should do something to clear this up. For now, I propose utilizing the existing footnote infra to communicate this. Opened #26384 to track a more complete solution.

jprous:
> The documentation is in a bit of a confusing state right now since we primarily target the US market, but it's not exactly clear when we don't. In order to make it as unambiguous for US market users, I'd still like this to show "All" for them, but we should do something to clear this up. For now, I propose utilizing the existing footnote infra to communicate this. Opened #26384 to track a more complete solution.

Thanks for the feedback Adeeb! I checked the comma website again, I cannot see that comma states anywhere that the list of compatible vehicles is USA based only. Remember, you have a huge intl following, of which a substantial part is people from developing countries, where cars such as Tesla's & Chevy bolts are simply not for sale. So your product/project draws attention and gives hope! In the Subaru community, we know of at least 2 cases where people invested in OP hardware based on the info on the site and ended up hitting a wall, both cases outside US, but in fairly well developed developing countries.

Suggestion: Instead of a footnote (which feels like a scheme-y tactic as a get out of jail card), rather have a extra tag, or extra column for INTL. Eg: US:All; INT:ES w LKAS. 
This will not only be applicable to Subaru, but to almost all the Japanese and Korean cars on the list.

Thanks again for looking into this. Don't under-estimate the support and admiration you guys have outside the USA! 

Greetings,
JP

martinl:
For Subaru, there are currently 2 known markets (Singapore, South Africa) where LKAS is disabled for Eyesight. For South Africa, jprous found a newer model year Forester that has LKAS enabled, so it does not apply to all model years. 

There are also differences in supported model years since Subaru introduces new models or platforms in Japan market one year earlier than rest of the world.

Maybe add a general note to CARS.md that supported cars list applies primarily to US market models and other markets models enabled features for same trims and model years may be different.

adeebshihadeh:
@sshane can you just make a generic footnote for all Subaru's that says this?

sshane:
@jprous can you allow pushing to your branch?

sshane:
@martinl do you know if this language matches what you're seeing for the non-US markets?

sshane:
@jprous can you allow me to push to your branch?

sshane:
thanks!

PR #26397, opened 2022-11-07
HKG: Car Port for 2022 Kia Stinger

sunnyhaibin:
**Checklist**
- [x] test route added to [test_routes.py](https://github.com/sunnyhaibin/openpilot-1/blob/d0050a2996be8e40777bf5ca987898803c2a7e29/selfdrive/test/test_routes.py#L74)
- [x] route with openpilot: `5b50b883a4259afb|2022-11-09--15-00-42`
- [x] Harness Type: [Hyundai K](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)

Thanks to community Stinger owner jyou0012.


adeebshihadeh:
Why is this broken out into a different platform? Anything different about this Stinger? Is it a new generation?

sunnyhaibin:
> Why is this broken out into a different platform? Anything different about this Stinger? Is it a new generation?

The older Stinger does not have LFAHDA_MFC message, while the newer Stinger do. This is similar to some other HKG models as well.

PR #26398, opened 2022-11-07
Kia EV6: Add EU fwdCamera firmware

alenl2:
Just added EU fw version for the fwdCamera.


sshane:
Do you have a route/dongle id where this came from?

alenl2:
dongle: 11300f9f578742eb

PR #26399, opened 2022-11-07
micd: scale sound volume with ambient noise level

### LINKED ISSUE #23084: soundd: scale volume with ambient noise instead of speed

incognitojam:
- adds `sounddevice` pypi package (depends on `libportaudio2` added in agnos 6.2)
- micd calculates `filteredSoundPressureDb`, published in `microphone` msg

Needs:
- https://github.com/commaai/cereal/pull/382
- https://github.com/commaai/cereal/pull/391

Closes #23084


system/micd.py
```diff
@@ -0,0 +1,43 @@
...
+    self.noise_level = np.linalg.norm(indata)
+
+  def micd_thread(self):
+    with sd.InputStream(callback=self.calculate_volume):
+      while True:
+        self.update()
+
+
+def main(pm=None):
+  config_realtime_process(5, Priority.CTRL_LOW)
```
incognitojam, line 33: not sure if this is necessary

incognitojam:
```
sounddevice.query_devices(kind='input')
{'name': 'default', 'index': 36, 'hostapi': 0, 'max_input_channels': 32, 'max_output_channels': 32, 'default_low_input_latency': 0.008684807256235827, 'default_low_output_latency': 0.008684807256235827, 'default_high_input_latency': 0.034807256235827665, 'default_high_output_latency': 0.034807256235827665, 'default_samplerate': 44100.0}
```

system/hardware/tici/hardware.py
```diff
@@ -96,6 +96,9 @@ def get_sound_card_online(self):
     return (os.path.isfile('/proc/asound/card0/state') and
             open('/proc/asound/card0/state').read().strip() == 'ONLINE')
 
+  def get_sound_input_device(self):
+    return "(hw:0,0)"
```
adeebshihadeh, line 100: is this necessary? looks like this relies on some sounddevice naming scheme?

adeebshihadeh (commented):
Looks good. Before merging, let's
- [ ] write a small test that plays some sounds and ensures that's reflected in the noise level
- [ ] test on at least five different cars with the device that's already in the car. sound varies from car to car, as well as device to device

system/hardware/tici/hardware.py
incognitojam: I think this is from ALSA. I remember it crashed when I didn't specify a device but will test again without

system/micd.py
```diff
@@ -0,0 +1,76 @@
...
+    self.measurements = np.array([])
+    self.filter = FirstOrderFilter(1, 7, DT_MIC)
+    self.last_alert_time = 0
+
+  def update(self):
+    self.sm.update(0)
+
+    if self.sm.updated['controlsState']:
+      if self.sm['controlsState'].alertSound > 0:
+        self.last_alert_time = time.time()
```
adeebshihadeh, line 33: hmm, things in system/ shouldn't subscribe to anything in selfdrive/. can we do this more generically? if not, perhaps move to selfdrive/ for now

system/micd.py
```diff
@@ -0,0 +1,67 @@
...
+    self.measurements = np.empty(0)
+    self.spl_filter = FirstOrderFilter(0, 4, DT_MIC, initialized=False)
+
+  def update(self):
+    # self.measurements contains amplitudes from -1 to 1 which we use to
+    # calculate an uncalibrated sound pressure level
+    if len(self.measurements) > 0:
+      # https://www.engineeringtoolbox.com/sound-pressure-d_711.html
+      sound_pressure = np.sqrt(np.mean(self.measurements ** 2))  # RMS of amplitudes
+      sound_pressure_level = 20 * np.log10(sound_pressure / REFERENCE_SPL) if sound_pressure > 0 else 0  # dB
```
sshane, line 29: maybe clip sound_pressure (in the log) to the min of reference_spl?

sshane, line 29:
```suggestion
      sound_pressure_level = 20 * np.log10(max(sound_pressure, REFERENCE_SPL) / REFERENCE_SPL)  # dB
```

sshane: not sure which is better, what do you think?

incognitojam: I think it's okay if dB goes negative

PR #26400, opened 2022-11-07
GM Camera: use ECM brake pressed bit

sshane:
It appears different GM camera platforms use different `ECMAcceleratorPos->BrakePedalPos` to cancel. Our Bolt for example is around 8, however the Silverado is around 20.

- Our Bolt (threshold of 8): https://pastebin.com/hAxDqcZS https://user-images.githubusercontent.com/25857203/200418683-eb2953c0-bf39-45bb-ba02-e6acf0ab5a0e.png
- Silverado 1500 (threshold of 8): https://pastebin.com/ULftqEw7 https://user-images.githubusercontent.com/25857203/200418699-7ef601f5-f714-4ec7-8df2-b8e4b95f52e8.png

Some times to cancel for the Silverado are upwards of a second! This means we'd try to cancel before the camera did, and cause a FCW/fake cruise fault.

If we use a threshold of 20 for the Silverado, the timings look much more similar to our Bolt:
- https://pastebin.com/TtN9bmkS

Since this will be cumbersome to maintain for all the different possible thresholds in the same platform of cars (Camera-ACC), we can use a brake pressed bit which captures this threshold.

We couldn't previously do this for the older platform since the Volts can have a faulty brake pedal housing and the bits did not capture what the position signal did (https://imgur.com/a/tC7sk3c). My guess is the bits are calibrated to the brake pedal's floor reading, however the ACC system also does a sanity check on the uncalibrated position signal, which seems like a bug.

If we use the `EBCMBrakePedalPosition->BrakePressed` bit, it closely matches our Bolt's current threshold, as well as the Silverado's threshold of ~20:

- Bolt w/ bit: https://pastebin.com/BKP9X9tQ
- Silverado w/ bit: https://pastebin.com/mdh3x2Fe

Also, the ECM brake bit isn't what the camera looks at like we thought, it's much too delayed: https://user-images.githubusercontent.com/25857203/200418296-81556eef-6a7b-4bb6-a5ec-30e23d92200f.png

`EBCMBrakePedalPosition->BrakePressed` actually looks correct.

Edit: Using the `ECMEngineStatus` bit as it's delayed and reduces chances of hitting a fault when you quickly engage after pressing the brake.


sshane:
This might not catch everything, you can fault if you engage shortly after releasing the brake pedal

sshane:
Turns out the fault I hit is a pre-existing fault. Even if the brake pos is zero, and all bits are zero, if the user recently braked hard and then enabled, the ECM faults... So no regression there. The ECM bit is actually more delayed than the other bits (and brake pos threshold), so this means we reduce the chances of this fault

selfdrive/car/gm/carstate.py
```diff
@@ -51,12 +51,15 @@ def update(self, pt_cp, cam_cp, loopback_cp):
...
-    # To avoid a cruise fault we need to match the ECM's brake pressed signal and threshold
-    # https://static.nhtsa.gov/odi/tsbs/2017/MC-10137629-9999.pdf
     ret.brake = pt_cp.vl["ECMAcceleratorPos"]["BrakePedalPos"]
-    ret.brakePressed = ret.brake >= 8
+    if self.CP.networkLocation == NetworkLocation.fwdCamera:
+      ret.brakePressed = pt_cp.vl["ECMEngineStatus"]["BrakePressed"] != 0
+    else:
+      # Some Volt 2016-17 have loose brake pedal push rod retainers which causes the ECM to believe
+      # that the brake is being intermittently pressed without user interaction.
+      # To avoid a cruise fault we need to match the ECM's brake pressed signal and threshold
```
sshane, line 60:
```suggestion
      # To avoid a cruise fault we need to use a conservative brake position threshold
```

sshane:
The new times of the Silverado using the ECM engine status bit. Everything looks good! All under 0.1 seconds:

```python
Took camera 0.08 s to cancel (10.400260981s)
Took camera 0.08 s to cancel (198.639840926s)
Took camera 0.05 s to cancel (206.16028593000001s)
Took camera 0.08 s to cancel (208.039599551s)
Took camera 0.06 s to cancel (209.930031813s)
Took camera 0.1 s to cancel (213.250302209s)
Took camera 0.08 s to cancel (217.93009203300002s)
Took camera 0.1 s to cancel (229.370370435s)
Took camera 0.1 s to cancel (262.590533403s)
Took camera 0.08 s to cancel (266.90982264s)
Took camera 0.08 s to cancel (281.30991355500004s)
Took camera 0.06 s to cancel (307.89028925400004s)
Took camera 0.07 s to cancel (372.059810973s)
Took camera 0.06 s to cancel (375.90025203600004s)
Took camera 0.06 s to cancel (497.989887306s)
Took camera 0.09 s to cancel (501.98990536800005s)
Took camera 0.08 s to cancel (505.709933501s)
Took camera 0.06 s to cancel (513.349860224s)
Took camera 0.08 s to cancel (516.710127745s)
Took camera 0.1 s to cancel (547.810230044s)
```

PR #26403, opened 2022-11-07
[controlsd] Revert Rav4 TSS2 to PID control

nuwandavek:
Torque control on Rav4 TSS2 has some issues (the multiple eps versions already needed separate PID tunes). I have identified some of the issues, but it might still be a regression if not fully fixed. Reverting back to the old PID tunes for now, will continue working on the issues in a branch.

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -119,7 +119,6 @@ class CarInterface(CarInterfaceBase):
       # See https://github.com/commaai/openpilot/pull/21429#issuecomment-873652891
       for fw in car_fw:
         if fw.ecu == "eps" and (fw.fwVersion.startswith(b'\x02') or fw.fwVersion in [b'8965B42181\x00\x00\x00\x00\x00\x00']):
-          ret.lateralTuning.init('pid')
           ret.lateralTuning.pid.kpV = [0.15]
           ret.lateralTuning.pid.kiV = [0.05]
           ret.lateralTuning.pid.kf = 0.00004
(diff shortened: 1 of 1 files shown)


selfdrive/car/toyota/interface.py
```diff
@@ -110,6 +110,19 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=[], experi
...
       ret.mass = 3585. * CV.LB_TO_KG + STD_CARGO_KG  # Average between ICE and Hybrid
+      ret.lateralTuning.init('pid')
+      ret.lateralTuning.pid.kpV = [0.6]
+      ret.lateralTuning.pid.kiV = [0.1]
+      ret.lateralTuning.pid.kf = 0.00007818594
+
+      # 2019+ RAV4 TSS2 uses two different steering racks and specific tuning seems to be necessary.
+      # See https://github.com/commaai/openpilot/pull/21429#issuecomment-873652891
+      for fw in car_fw:
+        if fw.ecu == "eps" and (fw.fwVersion.startswith(b'\x02') or fw.fwVersion in [b'8965B42181\x00\x00\x00\x00\x00\x00']):
```
sshane, line 121: Heads up, the single byte prefix on some FW versions is simply a length, so it may not describe what you want (though I still haven't figured out why some cars have different lengths). For example, new RAV4s not using torque have a length 2 EPS FW, as well as the radar ACC RAV4s. Have we looked into the 2022s to make sure they need this different tuning? @nuwandavek

PR #26408, opened 2022-11-08
Toyota: factor on cluster speed to match dash

sshane:
Works pretty well with hysteresis for the 22 CamryH which is actually 1.08


sshane:
Screenshot from 2022-11-08 20-53-50

PR #26420, opened 2022-11-09
Multilang: update pt-BR

AlexandreSato:
Include translations for "Experimental Mode" toggle.


sshane:
thanks for the translations!

PR #26428, opened 2022-11-10
Hyundai: add missing Sonata 2022 FW versions

sshane:


humza2000:
I have some questions. I added the VW fw version because there is already a Volkswagon Engine Ecu in CAR.SONATA on line 472:
b'\xf1\x870\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xf1\x82DNDWN5TMDCXXXJ1A',
Is this a mistake? 
You removed a Hyundai response for transmission ecu from my PR i.e 
b'\xf1\x87954A02N060\x00\x00\x00\x00\x00\xf1\x81T02832A1  \xf1\x00T02601BL  T02832A1  VDN8T25XXX832NS8G\x0e\xfeE' , along with the Volkswagen fw. Do we only add the first Hyundai response for each type of Fw?
 
Also, sorry I forgot to click the allow edit button.

humza2000:
Got it. Thanks!

PR #26442, opened 2022-11-10
GM: match stock enabling behavior

sshane:
Stock GM (ASCM and CAM platform) enables on falling edge of set, but rising edge of resume. Since openpilot enables on falling edge of both, you can fault the ECM if you hold the resume button longer than ~0.5 seconds before letting go, every time.

You can also fault if you depress resume just before you finish braking, then release (all under 0.5 seconds) as the ECM doesn't respect this attempt to enable at all.

It appears the max time from rising edge of pressing the resume button is 48 frames/480 ms (not sure why not 50/500ms).
Screenshot from 2022-11-10 00-09-13

The simplest and least confusing way to solve this is to match the stock system's behavior and ECM's safety checks by enabling on rising edge of resume.

If we silently noEntry, or even with an alert, that seems much more confusing to me.


sshane:
This is also a fault on ASCM GM! df51b2195de66a9b|2022-11-10--07-45-24--0 - from Verylukyguy

sshane:
This will catch 10/44 faults from 0.8.16 (applicable on both ASCM and camera platforms)

sshane:
diff is expected: 
Screenshot from 2022-11-10 20-57-05
Screenshot from 2022-11-10 20-57-16

selfdrive/car/interfaces.py
```diff
@@ -250,8 +250,8 @@ def create_common_events(self, cs_out, extra_gears=None, pcm_enable=True, allow_
       # Enable OP long on falling edge of enable buttons (defaults to accelCruise and decelCruise, overridable per-port)
       if not self.CP.pcmCruise and (b.type in enable_buttons and not b.pressed):
         events.add(EventName.buttonEnable)
-      # Disable on rising edge of cancel for both stock and OP long
-      if b.type == ButtonType.cancel and b.pressed:
+      # Disable on rising and falling edge of cancel for both stock and OP long
+      if b.type == ButtonType.cancel:
```
sshane, line 254: this is to prevent a fault on GM where you can go from cancel to resume (which enables on resume rising edge now now). Cancel still has precedence. Weirdly, this causes no fault at a standstill, only while moving

selfdrive/car/gm/interface.py
```diff
@@ -218,9 +218,14 @@ def _update(self, c):
 
       ret.buttonEvents = buttonEvents
 
+    # The ECM allows enabling on falling edge of set, but only rising edge of resume.
+    # Override standard behavior and enable on rising edge of resume
```
sshane, line 222:
```suggestion
```

selfdrive/car/gm/interface.py
```diff
@@ -218,9 +218,14 @@ def _update(self, c):
 
       ret.buttonEvents = buttonEvents
 
+    # The ECM allows enabling on falling edge of set, but only rising edge of resume.
```
sshane, line 221:
```suggestion
    # The ECM allows enabling on falling edge of set, but only rising edge of resume
```

PR #26450, opened 2022-11-10
GM: Detect unplugged or missing radar

nworb-cire:
Most GM models have different trim levels, only some of which work with openpilot. The highest trim level of Volt and Malibu have stock ACC and are supported, but LT or non-ACC Premier trims do not. These cars can be made to work with openpilot through firmware modification, but presently it is still required to retrofit a firmware module. This pull request removes the requirement for hardware modifications to these vehicles, i.e. compatibility can be added with vehicle firmware modifications only.

Among GM vehicles that use radar, fingerprint analysis (of those found both in stock and elsewhere, e.g. OPGM) shows that a common denominator for distinguishing whether radar is present is the `0x2cb == 715` CAN message. This is seen most clearly [here](https://raw.githubusercontent.com/opgm/openpilot/develop/selfdrive/car/gm/values.py) where one observes that all Volt and Malibu models without radar are missing `715` in the fingerprint, but models with radar present have it.

There is a possible false negative for Escalade ESV, so that is hardcoded for now. It is possible that stock radar was turned off during fingerprinting of that vehicle, so it will need to be double checked. Since this is only confirmed to affect Volt and Malibu at this time, the check could be moved into their model-specific sections if desired.


sshane:
So you're claiming that these lower trim ASCM GMs that don't have a radar will accept our longitudinal messages? Or is this to enable a lateral only port for these cars, where nothing does longitudinal?

nworb-cire:
This concerns only vehicles whose ECU can already accept longitudinal messages. The precise usefulness is for non-ASCM cars with modified firmware, which can accept longitudinal messages but are prevented from engaging openpilot due to absent radar causing radard to fail. Currently the solution for such vehicles is either to retrofit radar or alter the code to set `radarOffCan = True` to silence radard errors. This essentially does the latter dynamically.

Unmodded non-ASCM cars have different fingerprints than the ones currently in `values.py`. Adding lateral-only control for them is outside the scope of this PR.

adeebshihadeh (changes requested):
This is fine, with some caveats:
* `dashcamOnly` until the vision only policy is good enough
* not listed in the compatibility docs. wiki/discord is a better place for this kind of thing
* no special code paths to support re-flashed cars (it's fine the way this is done, but nothing more than this)

sshane:
@adeebshihadeh why not experimental long for these cars? We now treat missing radar differently from no radar cars

selfdrive/car/gm/interface.py
```diff
@@ -86,7 +87,9 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, expe
     else:  # ASCM, OBD-II harness
       ret.openpilotLongitudinalControl = True
       ret.networkLocation = NetworkLocation.gateway
-      ret.radarOffCan = False
+      radar_whitelist = {CAR.ESCALADE_ESV}  # 0x2cb missing from fingerprint even though it has radar
+      radar_missing = 0x2cb not in fingerprint[0] and candidate not in radar_whitelist
+      ret.radarOffCan = radar_missing
```
sshane, line 92: Think it's safe if you remove `radar_missing`, and only set `radarOffCan`. Then below, dashcam any car that's not integrated at the fwdCamera and radarOffCan. Removes the need for a local variable

nworb-cire: Better to set `radarOffCan = False` in the Escalade sublogic and move dashcam stuff to the bottom, or to leave `radar_whitelist` where it is?

nworb-cire:
I have just been informed (in the Discord GM channel) that there were previous suspicions about the Escalade fingerprint being incomplete. This wouldn't be the first time it had issues. Assuming that the fingerprint *is* incorrect, we should be able to take out the special logic once it is redone.

selfdrive/car/gm/interface.py
```diff
@@ -98,7 +98,8 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, expe
     # These cars have been put into dashcam only due to both a lack of users and test coverage.
     # These cars likely still work fine. Once a user confirms each car works and a test route is
     # added to selfdrive/car/tests/routes.py, we can remove it from this list.
-    ret.dashcamOnly = candidate in {CAR.CADILLAC_ATS, CAR.HOLDEN_ASTRA, CAR.MALIBU, CAR.BUICK_REGAL, CAR.EQUINOX, CAR.BOLT_EV}
+    ret.dashcamOnly = (candidate in {CAR.CADILLAC_ATS, CAR.HOLDEN_ASTRA, CAR.MALIBU, CAR.BUICK_REGAL, CAR.EQUINOX, CAR.BOLT_EV}
+                       or (ret.radarOffCan and candidate not in CAMERA_ACC_CAR))  # Until vision only policy is good enough
```
sshane, line 102:
```suggestion
                       or (ret.radarOffCan and ret.networkLocation == NetworkLocation.gateway))  # Until vision only policy is good enough
```

selfdrive/car/gm/interface.py
```diff
@@ -161,6 +162,8 @@ def get_params(candidate, fingerprint=gen_empty_fingerprint(), car_fw=None, expe
       ret.centerToFront = ret.wheelbase * 0.5
 
     elif candidate == CAR.ESCALADE_ESV:
+      ret.radarOffCan = False  # 0x2cb not in Escalade fingerprint even though it has radar. Likely incomplete fingerprint.
+      ret.dashcamOnly = False  # TODO: Remove this special logic if Escalade is re-fingerprinted
```
sshane, line 166: you should be able to eventually get rid of this, I can update the fingerprint in another PR if it's incorrect

selfdrive/car/gm/radar_interface.py
```diff
@@ -38,7 +38,10 @@ class RadarInterface(RadarInterfaceBase):
   def __init__(self, CP):
     super().__init__(CP)
 
-    self.rcp = create_radar_can_parser(CP.carFingerprint)
+    if CP.radarOffCan:
+      self.rcp = None
+    else:
+      self.rcp = create_radar_can_parser(CP.carFingerprint)
```
nworb-cire, line 44: With this change, we can likely eliminate the logic on lines 19-20 as well

sshane: Agree, just check radarOffCan like some of the other car ports do here

selfdrive/car/gm/values.py
```diff
@@ -143,6 +143,10 @@ class CanBus:
...
+  # Volt Premier 2018 w/ flashed firmware, no radar
+  {
+    170: 8, 171: 8, 189: 7, 190: 6, 192: 5, 193: 8, 197: 8, 199: 4, 201: 6, 209: 7, 211: 2, 241: 6, 288: 5, 289: 1, 290: 1, 298: 2, 304: 1, 308: 4, 309: 8, 311: 8, 313: 8, 320: 3, 328: 1, 352: 5, 368: 8, 381: 2, 384: 8, 386: 5, 388: 8, 389: 2, 390: 7, 417: 7, 419: 1, 426: 7, 451: 8, 452: 8, 453: 6, 454: 8, 456: 8, 458: 8, 479: 3, 481: 7, 485: 8, 489: 5, 493: 8, 495: 4, 497: 8, 499: 3, 500: 6, 501: 3, 508: 8, 512: 3, 528: 4, 530: 8, 532: 6, 537: 5, 539: 8, 542: 7, 546: 7, 550: 8, 554: 3, 558: 8, 560: 6, 562: 4, 563: 5, 564: 5, 565: 5, 566: 5, 567: 3, 568: 1, 573: 1, 608: 8, 609: 6, 610: 6, 611: 6, 612: 8, 613: 8, 647: 3, 707: 8, 711: 6, 761: 7, 810: 8, 821: 4, 823: 7, 832: 8, 840: 5, 842: 5, 844: 8, 853: 8, 866: 4, 961: 8, 967: 4, 969: 8, 977: 8, 979: 7, 988: 6, 989: 8, 995: 7, 1001: 5, 1003: 5, 1005: 6, 1009: 8, 1017: 8, 1019: 2, 1020: 8, 1033: 7, 1034: 7, 1105: 6, 1187: 4, 1217: 8, 1221: 5, 1223: 3, 1225: 7, 1227: 4, 1233: 8, 1249: 8, 1257: 6, 1265: 8, 1267: 1, 1273: 3, 1275: 3, 1280: 4, 1300: 8, 1322: 6, 1323: 4, 1328: 4, 1417: 8, 1905: 7, 1906: 7, 1907: 7, 1910: 7, 1912: 7, 1922: 7, 1927: 7, 2024: 8, 2025: 8, 2028: 8
```
sshane, line 149: Any need for this? You can leave 0x2cb in the FP. It only eliminates platforms if an address is NOT in this list

sshane:
715 is the gas/regen command message which is sent by the ASCM. Won't it also be missing when the ASCM is unplugged?

morrislee:
would this prevent longitudinal control via OP VOACC on GM vehicles that strictly uses vision for ACC that is not equipped with radar? I applied a factor to my Sierra 1500 for the vlead speed on voacc, seems to work more realistic now. stops behind lead pretty reliably so far

sshane:
@morrislee this PR is unrelated to your car

nworb-cire:
> 715 is the gas/regen command message which is sent by the ASCM. Won't it also be missing when the ASCM is unplugged/using the L&P harness?
Is there a complete list of codes and their meanings?

Also @adeebshihadeh can we gate this behind the new experimental long toggle instead of setting it to dashcam-only by default?

nworb-cire:
[George has indicated](https://discord.com/channels/469524606043160576/524328425415245827/1060355961832865852) that radarless (i.e. vision only) policy is just about ready, so closing this pending the necessary changes being merged to master, then I will refactor them as necessary.

PR #26460, opened 2022-11-10
[locationd] Add input checks

nuwandavek:
There are currently a lot of checks before observations are input into the kalman filter. This can lead to silent degradation of the filter if observations from some sensors are not used for a while (this is currently not tracked). Also, there are checks inside the filter (EKF) implementation to ignore observations older than some time buffer. We want to make sure we have visibility into all observations that are ignored, and act accordingly.

There are two types of checks
- Value based checks: Failures here are acceptable if under some frequency
- Time based checks: Failures here are not acceptable

Todo:
- [x] Test on ~100k segments and compare to master
- [x] Test on synthetic segments with errors
- [x] Test on actual segments with error
- [x] Make sure process replay segments are actually testing locationd


selfdrive/locationd/locationd.cc
```diff
@@ -209,10 +220,15 @@ void Localizer::handle_sensor(double current_time, const cereal::SensorEventData
...
   double sensor_time = 1e-9 * log.getTimestamp();
-
+  
   // sensor time and log time should be close
   if (std::abs(current_time - sensor_time) > 0.1) {
     LOGE("Sensor reading ignored, sensor timestamp more than 100ms off from log time");
+    this->observation_timings_invalid = true;
+    return;
+  }
+  else if (!this->is_timestamp_valid(sensor_time)) {
```
adeebshihadeh, line 230:
```suggestion
  } else if (!this->is_timestamp_valid(sensor_time)) {
```

selfdrive/locationd/locationd.cc
```diff
@@ -209,10 +220,15 @@ void Localizer::handle_sensor(double current_time, const cereal::SensorEventData
   }
 
   double sensor_time = 1e-9 * log.getTimestamp();
-
+
```
adeebshihadeh, line 223:
```suggestion

```

adeebshihadeh (commented):
I'll merge this since it's already been validated, but the tracking should be done generically. It's too easy now to write a bug where the tracking silently doesn't work for a particular service. We also have a nice filter class in util/common.cc that should be used.

selfdrive/locationd/locationd.cc
```diff
@@ -200,6 +203,14 @@ VectorXd Localizer::get_stdev() {
...
 
+bool Localizer::are_inputs_ok() {
+  return this->critical_services_valid(this->observation_values_invalid) && !this->observation_timings_invalid;
+}
+
+void Localizer::observation_timings_invalid_reset(){
+  this->observation_timings_invalid = false;
+}
+
 void Localizer::handle_sensor(double current_time, const cereal::SensorEventData::Reader& log) {
```
adeebshihadeh, line 214: the handle_xxx functions should return whether the data was used or not, then handle_msg can do all the validity tracking

selfdrive/locationd/locationd.cc
```diff
@@ -521,10 +579,15 @@ int Localizer::locationd_thread() {
 
   uint64_t cnt = 0;
   bool filterInitialized = false;
+  const std::vector<std::string> critical_input_services = {"cameraOdometry", "liveCalibration", "accelerometer", "gyroscope"};
```
adeebshihadeh, line 582: does it make more sense to blacklist non-critical services? that way new services will be checked by default

PR #26463, opened 2022-11-11
car interfaces: handle buttons changing state without going unpressed

sshane:
When you transition from resume to set or vice versa, without going to unpressed, we should send two events (unpress of previous and press of current). Right now we don't send the unpress of previous, so controlsd keep incrementing the cruise speed forever on non-`pcmCruise` cars.

Hyundai and GM already handle this. This de-duplicates the logic, and adds it to Honda.

Honda panda PR required to avoid a controls mismatch: https://github.com/commaai/panda/pull/1607


sshane:
this will expose controls mismatches in the other car ports, need to handle properly first

sshane:
This doesn't expose more controls mismatches, but we should fix them regardless. GM also handles this correctly, think Honda is the last that uses this function to not. Need to do something like this for Honda: https://github.com/commaai/panda/pull/1152

sshane:
@jyoung8607 if you're interested in a refactor, want to unify the two button event functions? Looks like the Tesla and VW one already handles what this PR does for the other function? It might be tricky since VW has signals for single bits, while other cars have one signal with multiple values.

jyoung8607:
> @jyoung8607 if you're interested in a refactor, want to unify the two button event functions? Looks like the Tesla and VW function already handles what this PR fixes for the other function.

It's a good goal, but for the next 60-90 days I don't have much time available for openpilot projects, and I need to spend what time I do have on my car port backlog. I'll add it to my long-term TODOs but someone else is welcome to tackle it before then.

PR #26478, opened 2022-11-12
controlsd: add tests around cruise speed

sshane:
Adds tests for:
- Speed adjusts on falling edge of buttons
- Cruise standstill exception added recently (https://github.com/commaai/openpilot/pull/25470)
- Enforcing limits on speed when enabling with SET

Only not pcmCruise tests for now as that mode is the most fragile, will make more complete in future PRs.


gast04:
please rebase, for CI ;)

sshane:
come onnnnnnnnnnnnnnnnnn

PR #26482, opened 2022-11-12
Multilang: add missing Japanese translations

ambientocclusion:
Description
Add missing Japanese translations.

Verification
Translated by a native.


sshane:
Thanks for the PR!

sshane:
@ambientocclusion do you mind opening another PR to translate the updated strings? I just made it so when we alter a string you won't have to re-translate the whole paragraph again!

ambientocclusion:
@sshane will do in weekend.

PR #26492, opened 2022-11-14
Update missing items in main_zh-CHS.ts

OxygenLiu:


sshane:
thanks for the pr!

sshane:
@OxygenLiu do you mind opening another PR to translate the updated strings? I just made it so when we alter a string you won't have to re-translate the whole paragraph again!

OxygenLiu:
@sshane no problem. https://github.com/commaai/openpilot/pull/26524

PR #26500, opened 2022-11-15
Rate limit honda steer

haraschax:


selfdrive/car/honda/carcontroller.py
```diff
@@ -99,6 +99,12 @@ def process_hud_alert(hud_alert):
                       "lanes_visible", "fcw", "acc_alert", "steer_required"])
 
 
+def rate_limit_steer(new_steer, last_steer):
```
sshane, line 102: You can use the current rate limiting code, and just pas in 0 user/eps torque so we don't duplicate code

PR #26515, opened 2022-11-16
WIP: switch locationd to use laikad

gast04:
test drive done (15.11.2022)
TODOs: 
* bearing takes too long to apply after a turn
* poor GPS signal warning appeared when driven into a covered parking lot


adeebshihadeh:
Once this is done, let's make sure to update provisioning too.

PR #26521, opened 2022-11-16
cabana: default factor to 1

pd0wm:
Qt cabana is very usable, and so much faster :)


deanlee:
I'm happy you like it :)

adeebshihadeh:
If you've got any feedbck: https://github.com/commaai/openpilot/discussions/26091

PR #26524, opened 2022-11-16
Updated main_zh-CHS.ts

OxygenLiu:


sshane:
thanks for all your work!

PR #26540, opened 2022-11-18
Add FPv2: 2022 Subaru Outback Wilderness

martinl:
**Description** Add missing fw values for 2022 Subaru Outback Wilderness

**Route**
Route: 7e051e2bb863d0b6|2022-11-17--17-45-50--49


sshane:
Thanks for the PR! Think it's worth having the wilderness subbrand as its own CarInfo? Do people assume that they're the same car?

Issue #26547, opened 2022-11-19
body: fingerprinting broken after setting separation time


adeebshihadeh:

sshane:
@briskspirit is this fixed?

adeebshihadeh:
No, we still need to ship the body FW update.

PR #26548, opened 2022-11-19
body: add can fingerprint

adeebshihadeh:
Minimal, temporary fix for #26547. The real fix is fixing the UDS implementation in the body ECU.

### DIFF
--- a/selfdrive/car/body/values.py
+++ b/selfdrive/car/body/values.py
@@ -26,6 +26,12 @@ CAR_INFO: Dict[str, CarInfo] = {
   CAR.BODY: CarInfo("comma body", package="All"),
 }
 
+FINGERPRINTS = {
+  CAR.BODY: [{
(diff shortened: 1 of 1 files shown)


selfdrive/car/body/values.py
```diff
@@ -26,6 +26,12 @@ class CAR:
   CAR.BODY: CarInfo("comma body", package="All"),
 }
 
+FINGERPRINTS = {
+  CAR.BODY: [{
+    513: 8, 516: 8, 514: 3, 515: 4,
```
sshane, line 31: The full fingerprint appears to be `{513: 8, 516: 8, 514: 3, 515: 4, 1833: 8, 1832: 8}`

sshane: Though those two higher addrs are only sent three times, but close to the start of the route.

PR #26555, opened 2022-11-20
Add missing FW versions for Audi S3/A3

MikeBurnsDev:
Added new fingerprint from 2016 Audi S3 for: engine, transmission, srs, eps, and fwdradar. Tested working in vehicle.


selfdrive/car/volkswagen/values.py
```diff
@@ -844,7 +848,7 @@ class VWCarInfo(CarInfo):
       b'\xf1\x875Q0909144T \xf1\x891072\xf1\x82\00521G00807A1',
     ],
     (Ecu.fwdRadar, 0x757, None): [
-      b'\xf1\x875Q0907567N \xf1\x890400\xf1\x82\00101',
+      b'\xf1\x875Q0907567N \xf1\x890400\xf1\x82\x0101',
```
jyoung8607, line 851:
```suggestion
      b'\xf1\x875Q0907567N \xf1\x890400\xf1\x82\00101',
```

Since this radar firmware was the same from your car (only difference being the octal vs hex representation) let's keep the original instead of changing it.

jyoung8607 (changes requested):
LGTM except for the superfluous radar change.

sshane:
Do you have a dongle ID for this car?

sshane:
Any updates @MikeBurnsDev?

MikeBurnsDev:
Dongle id is: 76b48b05df10f05e

PR #26556, opened 2022-11-20
Multilang: add missing Japanese translations

ambientocclusion:
Description
Add missing Japanese translations.
@sshane 

Verification
Translated by a native.


sshane:
thanks again for all your work!

PR #26565, opened 2022-11-21
Car docs: add a make-specific init function

sshane:
@jyoung8607 does this satisfy your VW footnote needs?

We can also use this for the website docs for CAN-FD harnesses, I noticed we were actually missing some cars since it's not super easy to keep up to date, nor is it obvious when it's wrong. With this approach, it could  automatically use the CAN-FD list.

Same for Honda's harnesses where we can just use the list we use in the car interface to avoid any mistakes


selfdrive/car/docs_definitions.py
```diff
@@ -173,6 +173,10 @@ def init(self, CP: car.CarParams, all_footnotes: Dict[Enum, int]):
 
     return self
 
+  def init_make(self, CP: car.CarParams):
+    """CarInfo subclasses can add make-specific logic for footnotes or similar."""
```
sshane, line 177:
```suggestion
    """CarInfo subclasses can add make-specific logic for harness selection, footnotes, etc."""
```

jyoung8607:
> @jyoung8607 does this satisfy your VW footnote needs?

It does, thank you!

PR #26573, opened 2022-11-22
Mazda Port 2019+

MoreTore:
- [ ] support older mazda longitudinal control with radar intercept or unplug radar or just use radar points for OP learning stack.
- [ ] route with openpilot:
- [x] route with stock system: 164080f7933651c4|2022-12-01--16-19-34
- [ ] route with limit tests.
- [ ] update panda safety. https://github.com/commaai/panda/pull/1164
- [ ] update DBC file. https://github.com/commaai/opendbc/pull/749
- [ ] update cereal safetyModel or use safety param instead. https://github.com/commaai/cereal/pull/386
- [ ] test on Comma 3 device. Waiting for order delivery.
- [ ] test on multiple Mazda models
- [ ] find unique and consistent can ignition signal. 
- [ ] use better initial tuning values and steering ratios
- [ ] Longitudinal tuning


MoreTore:
Still need to get some harnesses out to people to get more drives uploaded.

MoreTore:
@sshane I added the opendbc bump to this PR. I'm not sure if that complicates things. Maybe you want it in a separate PR?

sshane:
Yes, bumping opendbc on this PR is appropriate. Can you try to merge --no-ff master into your PR? Force pushing makes it tough to review as you don't know what was changed, or if it was a simple rebase on master

MoreTore:
> Yes, bumping opendbc on this PR is appropriate. Can you try to merge --no-ff master into your PR? Force pushing makes it tough to review as you don't know what was changed, or if it was a simple rebase on master

Ah sorry too late. Already have the sperate PR https://github.com/commaai/openpilot/pull/26608.

It would be a simple rebase https://github.com/commaai/openpilot/pull/26608

common/realtime.py
```diff
@@ -94,3 +94,91 @@ def monitor_time(self) -> bool:
     self._frame += 1
     self._remaining = remaining
     return lagged
+
+class DurationTimer:
```
sshane, line 98: Are these really nessesary?

selfdrive/car/mazda/carstate.py
```diff
@@ -18,118 +18,169 @@ def __init__(self, CP):
     self.lkas_allowed_speed = False
     self.lkas_disabled = False
 
-  def update(self, cp, cp_cam):
+  def update(self, cp, cp_cam, cp_body):
 
     ret = car.CarState.new_message()
-    ret.wheelSpeeds = self.get_wheel_speeds(
```
sshane, line 24: Let's try to make the signal names match so we don't need to split this, but if we can't let's split the update function into different gens like Hyundai CAN-FD does.

selfdrive/car/mazda/carstate.py
```diff
@@ -168,6 +219,55 @@ def get_can_parser(CP):
...
+        ("STATIC_0", "CRZ_BTNS",0),
+        ("STATIC_1", "CRZ_BTNS",0),
+        ("STATIC_2", "CRZ_BTNS",0),
+        ("CTR", "CRZ_BTNS", 0),
+        ("LEFT_BLINK", "BLINK_INFO", 0),
+        ("RIGHT_BLINK", "BLINK_INFO", 0),
+        ("ACCEL_CMD", "ACC", 0),
+        ("HOLD", "ACC", 0),
+        ("RESUME", "ACC", 0),
+        ("NEW_SIGNAL_1", "ACC", 0),
```
sshane, line 245: Can we understand what these signals are so we can set them appropriately? Or is there a reason to pass them through?

selfdrive/car/mazda/interface.py
```diff
@@ -40,8 +40,26 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long):
       ret.mass = 3443 * CV.LB_TO_KG + STD_CARGO_KG
       ret.wheelbase = 2.83
       ret.steerRatio = 15.5
+    elif candidate in CAR.MAZDA3_2019:
+      ret.openpilotLongitudinalControl = True
+      ret.stopAccel = 0.0
+      ret.longitudinalTuning.kpBP = [0., 5., 35.]
+      ret.longitudinalTuning.kpV = [1.2, 0.8, 0.5]
+      ret.longitudinalTuning.kiBP = [0., 35.]
+      ret.longitudinalTuning.kiV = [0.18, 0.12]
```
sshane, line 49: Let's do longitudinal in a subsequent PR so this one is more easily mergable

selfdrive/car/mazda/mazdacan.py
```diff
@@ -117,3 +126,96 @@ def create_button_cmd(packer, car_fingerprint, counter, button):
...
+        "DISTANCE_SETTING" : int(cp_cam.vl["CRZ_CTRL"]["DISTANCE_SETTING"]),
+        "ACC_ACTIVE_2"     : int(cc.longActive),
+        "DISABLE_TIMER_1"  : 0,
+        "DISABLE_TIMER_2"  : 0,
+        "NEW_SIGNAL_1"     : int(cp_cam.vl["CRZ_CTRL"]["NEW_SIGNAL_1"]),
+        "NEW_SIGNAL_2"     : int(cp_cam.vl["CRZ_CTRL"]["NEW_SIGNAL_2"]),
+        "NEW_SIGNAL_3"     : int(cp_cam.vl["CRZ_CTRL"]["NEW_SIGNAL_3"]),
+        "NEW_SIGNAL_4"     : int(cp_cam.vl["CRZ_CTRL"]["NEW_SIGNAL_4"]),
+        "NEW_SIGNAL_5"     : int(cp_cam.vl["CRZ_CTRL"]["NEW_SIGNAL_5"]),
+        "NEW_SIGNAL_6"     : int(cp_cam.vl["CRZ_CTRL"]["NEW_SIGNAL_6"]),
```
sshane, line 169: Usually it's standard practice to copy the whole message in CarState, then start the dictionary at the copied values.

Like: https://github.com/commaai/openpilot/blob/master/selfdrive/car/hyundai/carstate.py#L145

selfdrive/car/mazda/values.py
```diff
@@ -13,13 +13,24 @@
...
+    if CP.carFingerprint in GEN1:
+      self.STEER_MAX = 800                # theoretical max_steer 2047
+      self.STEER_DELTA_UP = 10             # torque increase per refresh
+      self.STEER_DELTA_DOWN = 25           # torque decrease per refresh
+      self.STEER_DRIVER_ALLOWANCE = 15     # allowed driver torque before start limiting
+      self.STEER_DRIVER_MULTIPLIER = 1     # weight driver torque
+      self.STEER_DRIVER_FACTOR = 1         # from dbc
+      self.STEER_ERROR_MAX = 350           # max delta between torque cmd and torque motor
+    if CP.carFingerprint in GEN2:
+      self.STEER_MAX = 8000
```
sshane, line 26: Is the new signal def really supposed to be 10x higher?

selfdrive/car/mazda/values.py
```diff
@@ -44,11 +61,18 @@ class MazdaCarInfo(CarInfo):
...
+  CAR.CX_60: MazdaCarInfo("Mazda CX-60 unreleased"),
+  CAR.CX_70: MazdaCarInfo("Mazda CX-70 unreleased"),
+  CAR.CX_80: MazdaCarInfo("Mazda CX-80 unreleased"),
+  CAR.CX_90: MazdaCarInfo("Mazda CX-90 unreleased"),
 }
 
 
 class LKAS_LIMITS:
-  STEER_THRESHOLD = 15
+  STEER_THRESHOLD = 1500
```
sshane, line 75: Same

selfdrive/controls/controlsd.py
```diff
@@ -723,6 +723,7 @@ def publish_logs(self, CS, start_time, CC, lac_log):
       self.pm.send('sendcan', can_list_to_can_capnp(can_sends, msgtype='sendcan', valid=CS.canValid))
       CC.actuatorsOutput = self.last_actuators
       self.steer_limited = abs(CC.actuators.steer - CC.actuatorsOutput.steer) > 1e-2
+      ControlsTimer.tick()
```
sshane, line 726: What's the purpose of this?

common/realtime.py
MoreTore: No, but I  think it looks cleaner to use this class for the control frame counter. You can call the `ControlsTimer.tick()` in controlsd instead of doing `self.frame += 1` in every brand's carcontroller update(). Then you can create an arbitrary amount of ControlsTimer objects to have independent duration timers. For example, in mazda/carcontroller.py I need to keep the "HOLD" signal high for 6 seconds so I just create a ControlsTimer object with a duration of 6 seconds and check if it's active(). Then I call the objects reset() to zero the timer. Another use is to call the ControlsTimer.interval(x) which replaces the ugly looking `if self.frame % x == 0`.

So no, it's not necessary but I think it's a nicer interface to use.

selfdrive/car/mazda/carstate.py
MoreTore: I haven't figured out what they do. All I know is that I have seen them change.

selfdrive/car/mazda/mazdacan.py
MoreTore: Thank you. I will clean that up.

selfdrive/car/mazda/values.py
MoreTore: Yes. The max can go even higher. I may end up reducing it further.

selfdrive/car/mazda/values.py
MoreTore: I will separate the THRESHOLD's based on GEN1, GEN2. 1500 is a good threshold for GEN2

selfdrive/controls/controlsd.py
MoreTore: This increments the self.frame used in all carcontroller.py 's. I will separate this out to another PR later.

common/realtime.py
sshane: It might be nice to clean up the car controllers in this way, but let's make that a separate PR

selfdrive/car/mazda/interface.py
MoreTore: Ok sounds good.

common/realtime.py
MoreTore: Roger that

MoreTore:
I think I was ban from the discord. No longer working on OP.

PR #26581, opened 2022-11-22
GM Cam: log when cruise not adaptive

### LINKED ISSUE #26578: GM: log nonAdaptive cruise mode

sshane:
On GM camera cars, nothing on the powertrain side of the bus changes when you go from adaptive to non-adaptive. The ECM cruise state is the same and the camera uses the same regen/accel and friction brake commands. The only difference is the camera changes its source of speed from lead + set speed to just set speed and changes the UI on the dash to reflect this.

So when OP controls long, nothing changes when adaptive CC is off as we send that signal and control gas/brake, ie. no need to check it there


sshane:
diff is from the can parser being initialized to 0 for first frame

PR #26582, opened 2022-11-22
Kia Sorento: fix dash speed matching

### LINKED ISSUE #26564: [‘19 Kia Sorento] Large difference in cluster and C3 speeds
### Describe the bug

The speed on my dash doesn’t match the speed on my C3 by sometimes up to 25+ MPH. In the photo I attached you can see the cluster says 68 MPH but the C3 says 42 MPH.

2019 Kia Sorento

b78a19b557b991ab|2022-11-20--12-38-48--2

### openpilot version

0.9.0

### Additional info

3FE64716-97F5-4DE0-B0C2-BB84F9725D02

sshane:
Only platform in our local data that sets `CF_Clu_VehicleSpeed` to the units of `CF_Clu_SPEED_UNIT`. On all other platforms, `CF_Clu_VehicleSpeed` is metric and `CF_Clu_VehicleSpeed2` is the one that is supposed to follow the unit bit (mph if the bit is 1, for example).

Some other cars also only set `CF_Clu_VehicleSpeed` and leave `CF_Clu_VehicleSpeed2` at zero, however `CF_Clu_VehicleSpeed` is still always metric (kph), this car is the exception.

This is true for the two Sorento dongles in our data in the last 120 days:
b78a19b557b991ab, 7eb2a0fcfa5547f8


sshane:
I used my bit finder script on the fleet, and made `bit_to_find = platform == "KIA SORENTO GT LINE 2018"` across 17k segments, with 100 of them being Sorento routes with the weird speed scaling.

The two closest are
```python
hex(msg)='0x52a', bit_mismatches=115507 of 9580626 (1.21%), byt_idx=7, bit_idx=7
hex(msg)='0x410', bit_mismatches=0 of 5351700 (0.0%), byt_idx=3, bit_idx=7
```

Where 0x410 is `CGW_USM1->CF_Gway_DrLockSoundRValue`. and 0x52a is a bit in CLU15 that is far from perfect (zero on my Sorento test route and some Sonata/Optima control routes). Interestingly, that 0x410 signal matches exactly, but the name makes it seem totally unrelated. Since we only have one car where this applicable across the fleet, I'm going to hard code it to a platform for now. If the future, we can switch to this signal, if it holds true on one more platform.

sshane:
TL;DR: Only the three Sorento dongles set `CGW_USM1->CF_Gway_DrLockSoundRValue` to anything differently than 17k random segments from other platforms explored. But since I'm not confident the signal is actually what we want (look at the name), I'll add a TODO until we see two platforms with the gateway signal ***and*** the speed mismatch.

PR #26586, opened 2022-11-23
Toyota: always log cruise standstill

sshane:
AFAIK cruiseState.standstill isn't used anywhere in Toyota-specific code, so this makes no sense to not log under certain conditions -- these "NO_STOP_TIMER" cars definitely can be put into standstill if engaged at a stop

Only thing that changes is the long controller doesn't try to apply accel when at a stop in standstill to no effect (undefined situation from the car's perspective, but it happens to handle this)

### DIFF
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -6,7 +6,7 @@ from common.realtime import DT_CTRL
 from opendbc.can.can_define import CANDefine
 from opendbc.can.parser import CANParser
 from selfdrive.car.interfaces import CarStateBase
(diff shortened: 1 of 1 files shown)


ntegan1:
does this make sense for tssp interceptor cars?

I seem to have stopped resuming from red lights yesterday, and haven't had time to see if my pedal has broken.

a152dad3c3368320|2022-11-23--22-44-01--4 stopped after going ~30mph, and before going ~45mph. (~250 sec into route?)  
pressed gas because car didn't want to seem to go.

I'll look into it more later but couldn't see how my car would exit standstill, when reading through longcontrol.py

sshane:
https://github.com/commaai/openpilot/pull/26597

PR #26591, opened 2022-11-23
Add ECU fw for 2016 Golf GTI

uppercaseVar:
This pull request is to add fw versions for a 2016 Golf GTI. This file has been tested on a vehicle and prevents the unrecognized vehicle error from showing.


vanillagorillaa:
Be sure to include the dongleID of the device this was taken from

uppercaseVar:
The dongle ID is 756a4b5cfe2fa939.

PR #26592, opened 2022-11-24
Fix typo in msm_kgsl.h

eltociear:
fucntion -> function


incognitojam:
This is an include from third party code, no need to fix typos

PR #26606, opened 2022-11-28
tools: Plotjuggler VideoViewer

ntegan1:
Add support in juggle.py for VideoViewer


ntegan1:
looks like this will be ubuntu only until we can include qtav mac in tools/setup.sh

adeebshihadeh:
Thanks for implementing this, but I think what we really want is Foxglove (see #30272) for this more advanced stuff.

PR #26608, opened 2022-11-28
Bump opendbc for new mazda port

MoreTore:


MoreTore:
@sshane Here is the PR.

adeebshihadeh:
We'll bump it here once the opendbc PR is merged. Just bump it in your own openpilot branch until then.

PR #26610, opened 2022-11-28
CarParams: log car control params

sshane:
Starting at only logging the max allowed control torque for the upcoming lat control refactor to map the accel factor to the absolute torque. Eventual plan is to add and fill these fields out and remove the fragmented CarControllerParams classes wherever possible: https://github.com/commaai/cereal/commit/e4f04c063f16cacc1ce64b24090a98f1c8f84f20


selfdrive/car/interfaces.py
```diff
@@ -98,6 +98,12 @@ def get_params(cls, candidate: str, fingerprint: Optional[Dict[int, Dict[int, in
     ret = CarInterfaceBase.get_std_params(candidate)
     ret = cls._get_params(ret, candidate, fingerprint, car_fw, experimental_long)
 
+    # Set maximum allowed steering torque
+    # TODO: the interfaces should set carControlParams
+    CarControllerParams = get_interface_attr("CarControllerParams")[ret.carName]
+    if CarControllerParams is not None and ret.steerControlType != car.CarParams.SteerControlType.angle:
+      ret.carControlParams.steerMax = CarControllerParams(ret).STEER_MAX
```
sshane, line 105: @adeebshihadeh are you fine with this cereal structure? Plan is to replace the classes with carParams eventually

adeebshihadeh:
Superseded by #26743

PR #26613, opened 2022-11-29
Car interface: set common params after port

sshane:
adds a wrapper function for `get_params`.

- removes all the pylint exceptions for the default mutable list and dict parameter values
- ability to set params after make-specific car interface fills out params (like rotationalInertia, soon to add stiffness)


adeebshihadeh (approved):
Nice cleanup!

selfdrive/car/interfaces.py
```diff
@@ -87,10 +87,27 @@ def __init__(self, CP, CarController, CarState):
   def get_pid_accel_limits(CP, current_speed, cruise_speed):
     return ACCEL_MIN, ACCEL_MAX
 
+  @classmethod
+  def get_params(cls, candidate: str, fingerprint: Optional[Dict[int, Dict[int, int]]] = None, car_fw: Optional[List[car.CarParams.CarFw]] = None, experimental_long: bool = False):
+    if fingerprint is None:
+      fingerprint = gen_empty_fingerprint()
+    if car_fw is None:
+      car_fw = list()
```
sshane, line 95:
```suggestion
    if fingerprint is None:
      fingerprint = gen_empty_fingerprint()

    if car_fw is None:
      car_fw = list()
```

PR #26614, opened 2022-11-29
mock: remove steering angle

sshane:
We only `self.yawRate`, not `self.yaw_rate`, so we always used 0.


adeebshihadeh:
Not sure why mock needs the steering angle. Can we just remove?

PR #26640, opened 2022-11-30
Increase Minimum Offroad Battery Voltage

EdwardApollo:
Updated minimum offroad battery voltage to reduce wear on car batteries and updated max offroad time to reflect the increased power consumption at idle based on EPHOT hardware power draw.

Most vehicles make use of lead acid batteries which during normal use shouldn't ever decrease below ~12V (different sources give a range of 11.9-12.2V). Increasing this limit will prevent premature battery failure by preventing excessive sulfation of the cells and by extension prematurely reduce the remaining life of the car battery.

Offroad Battery Voltage Distribution (Last 7 Days):
image-1

image
image


EdwardApollo:
I can also see the merit to leaving max off-road time at 30h to mitigate the need for a fresh boot after 24h assuming good battery conditions. 

The primary change here is the battery voltage.

HaraldSchafer:
3.7w sounds way too high. The problem with increasing the 11v limit is that it doesn't account for measurement error or voltage drop. We found the current measurement we are using now to be the best shutoff mechanism. The 30hr limit was decided because it's the minimum to have the device always on when you get in the car if you drive daily.

AlexandreSato:
> Updated minimum offroad battery voltage to reduce wear on car batteries and updated max offroad time to reflect the increased power consumption at idle based on EPHOT hardware power draw.
> 
> Most vehicles make use of lead acid batteries which during normal use shouldn't ever decrease below ~12V (different sources give a range of 11.8-12.2V). Increasing this limit will prevent premature battery failure by preventing excessive sulfation of the cells and by extension prematurely reduce the remaining life of the car battery.
> 
> image image

there is 0.16V voltage drop in the wire to the comma 3

EdwardApollo:
> 3.7w sounds way too high. The problem with increasing the 11v limit is that it doesn't account for measurement error or voltage drop. We found the current measurement we are using now to be the best shutoff mechanism. The 30hr limit was decided because it's the minimum to have the device always on when you get in the car if you drive daily.

@HaraldSchafer  30 hour limit makes total sense, as other metrics will catch it in the meantime if needed

Has this error ever been measured ~1V? Otherwise 11.0V still seems much too low.
>...
> there is 0.16V voltage drop in the wire to the comma 3

@AlexandreSato this is pretty close to what I'm seeing comparing the value in comma connect and actually taking a reading on my battery directly. 

On connect it is currently showing 12.3V and 12.2V. Conservatively assuming the larger of the 2 voltage drops, I'm seeing a 0.14V difference, so yes 0.16V seems reasonable.

So instead of 12V, 11.9 or 11.8V as a cutoff seems more reasonable, and assuming this difference doesn't grow much further than that 11.0V still seems very low.

PXL_20221130_210339771

jyoung8607:
You don't want to shutdown after 10 hours. That would require a cold boot every morning for a daily work commute. Remember there exist cars/integrations where you'll have faults and need to restart the car if openpilot goes onroad too late.

Also, that 3.7W idle usage does NOT sound right. I recall something in the 1W range when playing with it during EPHOT but I haven't measured lately. It should actually have gotten better with Panda deep sleep mode.

The battery voltage floor is indeed too low IMO, but before dramatically raising it:

Account for steady-state voltage drop between C3 and the battery. It's not huge but it's not zero. From memory, last time this was discussed, it was in the 0.2-0.3 volt range on some tested cars. The chart you posted is valid, but only when measuring open circuit voltage at the battery. You are measuring from some distance away, across a circuit with a load on it.

Account for instantaneous drop when owners unlock their cars and get in, before starting the engine. C3 is likely to see voltages in the mid-11s even with the battery in a decent state of charge. Again, that chart is showing OCV measurements, and from the moment an owner hits unlock on their key fob, the car is going to be very "awake" and dragging that way down.

The current voltage threshold for shutdown is < 11.0V for 5 seconds. I think something like <11.8V for 5 *minutes* (which is more like 12.0 at the battery) could be more appropriate.

Hubblesphere:
> The current voltage threshold for shutdown is < 11.0V for 5 seconds. I think something like <11.8V for 5 minutes (which is more like 12.0 at the battery) could be more appropriate.

This was what I was thinking as well. Need it to be a longer time frame. I also don't want power on time to be shortened.

EdwardApollo:
Reverted max offroad time to 30h, changed voltage cutoff to 11.8V

EdwardApollo:
> You don't want to shutdown after 10 hours. That would require a cold boot every morning for a daily work commute. Remember there exist cars/integrations where you'll have faults and need to restart the car if openpilot goes onroad too late.

Reverted to 30h

> Also, that 3.7W idle usage does NOT sound right. I recall something in the 1W range when playing with it during EPHOT but I haven't measured lately. It should actually have gotten better with Panda deep sleep mode.

Removed the comment entirely, as it pertained to C2 anyway

> The battery voltage floor is indeed too low IMO, but before dramatically raising it:
> 
> Account for steady-state voltage drop between C3 and the battery. It's not huge but it's not zero. From memory, last time this was discussed, it was in the 0.2-0.3 volt range on some tested cars. The chart you posted is valid, but only when measuring open circuit voltage at the battery. You are measuring from some distance away, across a circuit with a load on it.
> 
> Account for instantaneous drop when owners unlock their cars and get in, before starting the engine. C3 is likely to see voltages in the mid-11s even with the battery in a decent state of charge. Again, that chart is showing OCV measurements, and from the moment an owner hits unlock on their key fob, the car is going to be very "awake" and dragging that way down.
> 
> The current voltage threshold for shutdown is < 11.0V for 5 seconds. I think something like <11.8V for 5 _minutes_ (which is more like 12.0 at the battery) could be more appropriate.

Changed to voltage floor to 11.8V instead of 12 per my earlier comment comparing voltage at the battery and the value displayed in connect

Good point regarding timing, definitely a decent point

HaraldSchafer:
We'd have to check across the fleet weather we ever have big measurement errors or voltage drop, before we can make changes to this. 

But besides that I just went through some of our previous discussions on this. All the cases we've seen a dead battery the battery was already so degraded that even with a much higher voltage cutoff the car likely wouldn't have started anyway. There's no case I am aware of where the C3 killed a car with a healthy battery.

I think the best thing to do is to have some kind of battery-health tracking so that when your battery is running out of capacity you are somehow warned.

EdwardApollo:
> @HaraldSchafer 
> We'd have to check across the fleet weather we ever have big measurement errors or voltage drop, before we can make changes to this.
> 
> But besides that I just went through some of our previous discussions on this. All the cases we've seen a dead battery the battery was already so degraded that even with a much higher voltage cutoff the car likely wouldn't have started anyway. There's no case I am aware of where the C3 killed a car with a healthy battery.
> 
> I think the best thing to do is to have some kind of battery-health tracking so that when your battery is running out of capacity you are somehow warned.

So yes, if a car's battery is so degraded that the car can't start anyway, the C3 won't have an affect in that

*However*, part of lengthening a car batteries life is not allowing it to get in that voltage range in which the car can still start, but you are significantly reducing the life of it.

While technically, a battery will likely start a car at any voltage above 11.6V, allowing it to get to 11.9V or below will cause the battery to degrade *much* quicker.

AlexandreSato:
This pull request is important to preserve the comma image, something is degrading the batteries faster

EdwardApollo:
@robbederks took a stab at adding an IsOffroad parameter check to ensure a shutdown doesn't happen per your suggestion. There is a good chance I did it incorrectly, so here be dragons.

EdwardApollo:
Testing this commit now. Will report back if I see any issues.

EdwardApollo:
This commit boots and seems to work fine on my comma 3. 

Tau increased 12s->45s to achieve a more effective LPF.

HaraldSchafer:
Is the offroad flag needed? Does the ignition check not cover this?

EdwardApollo:
> Is the offroad flag needed? Does the ignition check not cover this?

Good point. If ignition on doesn't include accessory mode then I guess not.

incognitojam:
> Good point. If ignition on doesn't include accessory mode then I guess not.
There is no distinction between accessory mode and ignition in openpilot - if using the 12V line to detect ignition it usually goes high regardless of accessory mode or engine on.

robbederks:
> @robbederks took a stab at adding an IsOffroad parameter check to ensure a shutdown doesn't happen per your suggestion. There is a good chance I did it incorrectly, so here be dragons.
> 
> Lowpass tau is still at 12s. Maybe this should be increased also

Yeah, IsOffroad is basically the same as the ignition line, so that's not a useful check

EdwardApollo:
> @HaraldSchafer 
> Is the offroad flag needed? Does the ignition check not cover this?


> @incognitojam 
> There is no distinction between accessory mode and ignition in openpilot - if using the 12V line to detect ignition it usually goes high regardless of accessory mode or engine on.

> @robbederks 
> Yeah, IsOffroad is basically the same as the ignition line, so that's not a useful check

Reverted this change. Being that the primary concern was that the device may power off while the car is I'm accessory mode, I think we're good.

EdwardApollo:
image-1

Adding Offroad Battery Voltage Distribution from the last 7 days for context

EdwardApollo:
Thinking about this a bit further, if the goal is to prevent the battery from ever seeing <12v, we should set 11.9V as the limit, using the steady state voltage drop between the battery and the comma device as a buffer to ensure that those who don't drive daily don't repeatedly enter the sub 12V range, as even entering that range will accelerate battery degradation.

Once the standby C3 power draw is improved, it could be raised to 12V with minimal impact on the majority of drivers, really only affecting those who leave their vehicle idle for multiple days.

EdwardApollo:
For ease of review, the only changes made via this PR are:

CAR_VOLTAGE_LOW_PASS_K = 0.011 # LPF gain for 45s tau (dt/tau / (dt/tau + 1))
and
VBATT_PAUSE_CHARGING = 11.9

HaraldSchafer:
Please put the value back to 11.8V

The voltage cut-off is not intended to prevent the battery from draining too much to prevent degradation. That is what the charge/drain tracking is for. That is the primary battery-management and is very reliable, much more reliable than voltage cut-off can be. All the voltage cut-off can do is prevent deterioration when the charge/drain tracking fails which I have only seen happen in cases where the battery was so deteriorated it had less than 10% charge left. If there any more examples of the charge/drain tracking failing for any reason, we want to look into them.

Just to reiterate the C3 shuts down after 30Wh of power draw (and only if it thinks battery has recently been charged), **that's less than 10% of even a small prius battery**

EdwardApollo:
> Please put the value back to 11.8V
> 
> The voltage cut-off is not intended to prevent the battery from draining too much to prevent degradation. That is what the charge/drain tracking is for. That is the primary battery-management and is very reliable, much more reliable than voltage cut-off can be. All the voltage cut-off can do is prevent deterioration when the charge/drain tracking fails which I have only seen happen in cases where the battery was so deteriorated it had less than 10% charge left. If there any examples of the charge/drain tracking failing for any reason, we want to look into them.
> 
> Just to reiterate the C3 shuts down after 30Wh of power draw (and only if it thinks battery has recently been charged), **that's less than 10% of even a small prius battery**

Reverted

HaraldSchafer:
Ready to merge when the diff is clean, seems to be merged weirdly or something

EdwardApollo:
> Ready to merge when the diff is clean, seems to be merged weirdly or something

Accidentally merged master into this branch, then reverted. Unsure how to fix.

EdwardApollo:
@HaraldSchafer  I could try git reset --hard c00fb9a8ce26d2d77543d643910c5dcd8fefc440 and then force push, but feel free to advise or modify yourself if there's a better method

EdwardApollo:
New PR created to clean up commit. See https://github.com/commaai/openpilot/pull/26661

PR #26642, opened 2022-11-30
GM: fix gas/regen signal

sshane:
The signal was not properly defined, and was wrapping around to 0 (looking at a stock route) when it requested more than 4095 units. I'm actually not sure if this definition is incorrect, it may keep going a few more bits back, since those are always 1. I don't know why GM chose these arbitrary values for this signal


sshane:
not really needed until we want to use more than the full range available now

PR #26645, opened 2022-12-01
Honda Nidec: put max PCM gas into CarControllerParams

sshane:
panda safety in master


selfdrive/car/honda/values.py
```diff
@@ -27,6 +27,7 @@ class CarControllerParams:
   NIDEC_MAX_ACCEL_V = [0.5, 2.4, 1.4, 0.6]
   NIDEC_MAX_ACCEL_BP = [0.0, 4.0, 10., 20.]
 
+  NIDEC_GAS_MAX = 198
```
sshane, line 30:
```suggestion
  NIDEC_GAS_MAX = 198  # 0xc6
```

PR #26651, opened 2022-12-01
VW MQB: Restore lead car indicator in instrument cluster

jyoung8607:
**Description**

Display a lead car in the instrument cluster when openpilot sees one. openpilot itself doesn't expose anything for safe-time-gap distance yet, so the displayed car is fixed at the middle of the range.

There is a frame > 100 check to ensure we've had a chance to see `Kombi_03.KBI_Variante` if it's present. On older cars that didn't have the option for digital instrument clusters, that message may not be forwarded to Extended CAN, so we can't put a frequency check in the CAN parser or delay controls startup waiting for it.

**Verification**

Drive tested with leads in and out of range, worked as expected.

On a related note, `hudControl.leadVisible` could use a debounce timer in higher level controls.

**Route**

Route: `3cfdec54aa035f3f|2022-11-30--23-25-39`

### DIFF
--- a/selfdrive/car/volkswagen/carstate.py
+++ b/selfdrive/car/volkswagen/carstate.py
@@ -13,6 +13,7 @@ class CarState(CarStateBase):
     self.CCP = CarControllerParams(CP)
     self.button_states = {button.event_type: False for button in self.CCP.BUTTONS}
     self.esp_hold_confirmation = False
+    self.upscale_lead_car_signal = False
 
   def create_button_events(self, pt_cp, buttons):
     button_events = []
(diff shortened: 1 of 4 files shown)


selfdrive/car/volkswagen/carstate.py
```diff
@@ -312,6 +317,7 @@ def get_can_parser(CP):
       ("Airbag_02", 5),     # From J234 Airbag control module
       ("Kombi_01", 2),      # From J285 Instrument cluster
       ("Blinkmodi_02", 1),  # From J519 BCM (sent at 1Hz when no lights active, 50Hz when active)
+      ("Kombi_03", 0),      # From J285 instrument cluster (not present on older cars, 1Hz when present)
```
adeebshihadeh, line 320: any way to detect this? not a huge fan of just ignoring it

jyoung8607: It's the best of the possible alternatives. If the message is genuinely missing, the involved code would fail in a safe, albeit cosmetically annoying way (lead car showing too close in the instrument cluster).

We could write off any future possibility of making OP long work with camera-side integrations; Kombi_03 appears to be guaranteed present at the CAN gateway on bus 1. I don't want to do that, it writes off supporting OP long with certain newer cars where it's hard to make a gateway harness.

I considered adding the instrument clusters to the fingerprint, but enumerating the digital clusters within that would become an annoying task with ongoing maintenance required as new part numbers are introduced. I checked and the software part numbers don't lend themselves to parsing to differentiate digital vs analog.

Also considered a standalone active UDS data identifier query to the cluster in the car port init(). Haven't searched for a discriminator yet but the possibility is good I could find something. But I already know I can only do that across the OBD bus. Based on our DM conversation some months ago, I don't want to build-in a new dependency on something I might not be able to accomplish from can0/can2 alone.

PR #26652, opened 2022-12-01
micd: don't update filtered sound level if playing sound

incognitojam:


sshane:
This worked perfectly! Paused the filter and restarted super fast, and it didn't affect the filtered value at all

Issue #23085, opened 2021-12-01
soundd: look into ramping up immediate disengage sound


adeebshihadeh:

adeebshihadeh:
@deanlee want to try this one out? I'm thinking we add another field to `sound_list` where you can specify how many loops it should take a sound to ramp up to the full volume.

PR #26662, opened 2022-12-01
soundd: ramp up from lower volume over a few seconds

incognitojam:
related:
- https://github.com/commaai/openpilot/issues/23085

just a proof of concept

### DIFF
--- a/selfdrive/ui/soundd/sound.cc
+++ b/selfdrive/ui/soundd/sound.cc
@@ -46,11 +46,10 @@ void Sound::update() {
     return;
   }
 
-  // scale volume with speed
+  // scale volume with measured sound level
   if (sm.updated("microphone")) {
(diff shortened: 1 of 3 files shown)


system/hardware/pc/hardware.h
```diff
@@ -15,7 +15,7 @@ class HardwarePC : public HardwareNone {
     volume = util::map_val(volume, 0.f, 1.f, MIN_VOLUME, MAX_VOLUME);
 
     char volume_str[6];
-    snprintf(volume_str, sizeof(volume_str), "%.3f", volume);
-    std::system(("pactl set-sink-volume @DEFAULT_SINK@ " + std::string(volume_str)).c_str());
+    snprintf(volume_str, sizeof(volume_str), "%.0f", volume * 100);
+    std::system(("pactl set-sink-volume @DEFAULT_SINK@ " + std::string(volume_str) + "%").c_str());
```
incognitojam, line 19: the volume will be scaled "naturally" if the argument is a percentage - let pulseaudio do this here and remove the scaling from soundd?

PR #26665, opened 2022-12-02
navmodeld

mitchellgoffpc:


selfdrive/modeld/models/nav.h
```diff
@@ -0,0 +1,75 @@
+#pragma once
+
+#include "cereal/messaging/messaging.h"
+#include "cereal/visionipc/visionipc_client.h"
+#include "common/util.h"
+#include "common/modeldata.h"
+#include "selfdrive/modeld/models/commonmodel.h"
+#include "selfdrive/modeld/runners/run.h"
+
+#define NAV // TODO: This should probably be defined in the sconscript or something
```
adeebshihadeh, line 10: will we ever want a build without this defined?

selfdrive/modeld/models/nav.cc
```diff
@@ -0,0 +1,73 @@
+#include "selfdrive/modeld/models/nav.h"
+
+#include <cstdio>
+#include <cstring>
+
+#include "common/mat.h"
+#include "common/modeldata.h"
+#include "common/params.h"
```
adeebshihadeh, line 8: unused

selfdrive/modeld/models/nav.cc
```diff
@@ -0,0 +1,73 @@
...
+#include "common/timing.h"
+#include "system/hardware/hw.h"
```
adeebshihadeh, line 10: unused

selfdrive/modeld/navmodeld.cc
```diff
@@ -0,0 +1,57 @@
...
+
+int main(int argc, char **argv) {
+  setpriority(PRIO_PROCESS, 0, -15);
+
+  // init the models
+  NavModelState model;
+  navmodel_init(&model);
+  LOGW("models loaded, navmodeld starting");
+
+  VisionIpcClient vipc_client = VisionIpcClient("navd", VISION_STREAM_MAP, true);
```
adeebshihadeh, line 44: with the conflate enabled and the way the decimation is done, you're open to frame drops. is this intentional?

selfdrive/modeld/models/nav.cc
```diff
@@ -0,0 +1,73 @@
...
+
+
+template<class T, size_t size>
+constexpr const kj::ArrayPtr<const T> to_kj_array_ptr(const std::array<T, size> &arr) {
```
adeebshihadeh, line 14: looks like this is copy and pasted from models/driving.cc?

mitchellgoffpc: It is, not sure where a good place to put it would be?

selfdrive/modeld/navmodeld.cc
mitchellgoffpc: No, I didn't realize frame_ids could be skipped. I can change to a counter instead.

selfdrive/modeld/models/nav.h
mitchellgoffpc: Removed for now, this flag really belongs in driving.h with the other model input gates

selfdrive/modeld/navmodeld.cc
adeebshihadeh: if we care about it actually being every 10th frame, just change that true to a false. otherwise, change the decimation to use a counter

PR #26668, opened 2022-12-02
micd: apply A-weighting to the sound pressure level

sshane:
Apply [A-weighting](https://en.wikipedia.org/wiki/A-weighting) on the microphone audio signal so that our sound pressure level calculation is more representative of what a human perceives the loudness to be (reduces lower frequencies).

https://user-images.githubusercontent.com/25857203/205243275-f2859457-d01e-4f2e-9c07-bfebb0039888.mp4


system/micd.py
```diff
@@ -10,7 +11,35 @@
...
+  # Calculate the frequency axis for the signal
+  freqs = np.fft.fftfreq(measurements_windowed.size, d=1 / SAMPLE_RATE)
+
+  # Calculate the A-weighting filter
+  # https://en.wikipedia.org/wiki/A-weighting
+  A = 12194 ** 2 * freqs ** 4 / ((freqs ** 2 + 20.6 ** 2) * (freqs ** 2 + 12194 ** 2) * np.sqrt((freqs ** 2 + 107.7 ** 2) * (freqs ** 2 + 737.9 ** 2)))
+  A /= np.max(A)  # Normalize the filter
+
+  # Apply the A-weighting filter to the signal
+  return np.fft.ifft(np.fft.fft(measurements_windowed) * A)
```
sshane, line 42:
```suggestion
  return np.fft.ifft(np.abs(np.fft.fft(measurements_windowed)) * A)
```

@robbederks is this the correct way to cleanly get rid of the imaginary warning?

selfdrive/ui/soundd/sound.cc
```diff
@@ -48,7 +48,7 @@ void Sound::update() {
 
   // scale volume with speed
   if (sm.updated("microphone")) {
-    float volume = util::map_val(sm["microphone"].getMicrophone().getFilteredSoundPressureDb(), 58.f, 77.f, 0.f, 1.f);
+    float volume = util::map_val(sm["microphone"].getMicrophone().getFilteredSoundPressureWeightedDb(), 30.f, 52.f, 0.f, 1.f);
```
sshane, line 51:
```suggestion
    float volume = util::map_val(sm["microphone"].getMicrophone().getFilteredSoundPressureWeightedDb(), 30.f, 55.f, 0.f, 1.f);
```

PR #26671, opened 2022-12-02
Toyota: Don't Set Standstill Bit on TSS-P Prius

cydia2020:
I think this is an acceptable change now openpilot has #25024

This PR enables full SnG (no pedal or user intervention required) on all TSS-P Toyota Prius. It does so by not setting the standstill bit when the vehicle comes to a complete stop. This may waste some energy and wear the motors out faster but it will be up to the user to decide whether they would like to use a pedal.


sshane:
@cydia2020 do you want to re-open so I can verify on our Prius before re-merging?

PR #26672, opened 2022-12-02
micd: revert check playing sound (high cpu usage)

sshane:
This command is causing micd to use ~27% CPU onroad. @incognitojam we should try to call at a lower frequency at the least


incognitojam:
maybe we want something like the [`pulsectl`](https://github.com/mk-fg/python-pulse-control) python package (or do it from C++ with `libpulse`)

adeebshihadeh:
Le'ts also add micd to the CPU usage tests.

incognitojam:
using `pactl` python package
```
is_sound_playing took 0.000s
is_sound_playing took 0.000s
is_sound_playing took 0.000s
is_sound_playing took 0.000s
```

calling pactl using subprocess (!)
```
is_sound_playing took 0.031s
is_sound_playing took 0.034s
is_sound_playing took 0.033s
is_sound_playing took 0.033s
```

PR #26674, opened 2022-12-02
Micd: update sound levels in callback

sshane:
So we can ensure we pass in a number of samples that are a power of 2 to the fft


system/micd.py
```diff
@@ -45,41 +46,46 @@ def __init__(self, pm):
...
-      self.spl_filter_weighted.update(sound_pressure_level_weighted)
-    else:
-      sound_pressure = 0
-      sound_pressure_weighted = 0
-      sound_pressure_level_weighted = 0
+    self.spl_filter_weighted = FirstOrderFilter(0, 2.5, DT_MIC, initialized=False)
 
-    self.measurements = np.empty(0)
+  def update(self):
+    if self.sound_pressure_level_weighted != 0:
```
sshane, line 57:
```suggestion
    if self.sound_pressure_level_weighted != 0 and not self.spl_filter_weighted.initialized:
```

robbederks (approved):

PR #26678, opened 2022-12-02
Subaru: add missing fw values for 2023 Crosstrek Limited

martinl:
**Description** This PR adds missing fw values for 2023 Crosstrek Limited

**Route**
Route: `f900fccd457f495a|2022-12-02--14-13-03`


sshane:
@martinl our docs say we only support the crosstrek 2020-21, do we support 2022 and now 2023?

martinl:
I have not seen any 2022 but looks like it, since 2023 has not changed

PR #26683, opened 2022-12-03
tools: add script to find similar bits

sshane:
Used to find these undefined bits within minutes: https://github.com/commaai/opendbc/commit/4e3899cbba7188116e5919923a7dc24264fa3d38

You just need to find a good proxy for what you want to find. In my case, I saw in PJ that the blocked cnt increased right around where the stock system faulted, so I filtered for that.

TODO:
- [ ] Clean up
- [ ] Somehow make a CLI interface? It might be hard to do everything you want over CLI. Sometimes you want `bit_to_find` to be latching, sometimes you want to compare cereal fields, sometimes you want to compare other bits/signals on CAN

### DIFF
--- a/selfdrive/debug/find_similar_bits.py
+++ b/selfdrive/debug/find_similar_bits.py
@@ -1,12 +1,14 @@
 import os
 os.environ["FILEREADER_CACHE"] = "1"
+from collections import defaultdict
+from tqdm import tqdm
+import matplotlib.pyplot as plt
+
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Added in cabana. Still want this in?

sshane:
It's still useful to compare to openpilot signals, so I'll still use this script. I'll see about adding that functionality to cabana in the future

PR #26684, opened 2022-12-03
fix dependencies for macOS

incognitojam:
- skip on macOS:
  - carla
  - spidev
  - opencv-python-headless

macOS CI will remain disabled as there are still build issues


incognitojam:
Poetry's support for dependencies with multiple version constraints doesn't seem very mature. It will either attempt to install the `opencv-python-headless` package from the linux wheel on macOS, or install the package from pypi on linux (ignoring the wheel), depending on whether the package is already installed when you run the script.

For now we can just skip installing `opencv-python-headless` on macOS, until someone can figure out what is wrong (maybe there is a bug in Poetry?)

MaherJendoubi:
Please Merge this PR

PR #26685, opened 2022-12-03
Hyundai: setup alt steering limits

adeebshihadeh:


sshane:
we should also adjust the torque params, these cars are probably now undershooting a lot

PR #26687, opened 2022-12-03
Toyota: log EPS fault from LKA cmd message drop out

sshane:
The LKA_STATE code for the EPS failing to receive LKA msgs is 3, and on rising edge of the fault ***while engaged*** it goes to 9 for ~1.2 seconds, then 11 for 0.8 seconds, then rests at 3 (recoverable if EPS gets steering msgs again).

If openpilot were ever to stop sending LKA messages and the EPS status was at 3, openpilot would enable without any steering assist or warnings (not even from car).

f15e3c37c118e841|2022-12-02--23-47-57

Screenshot from 2022-12-03 01-02-12


sshane:
Running on the fleet, it's catching valid faults!

- Prius 2017, first segment, we drop a bunch of messages and would create a noEntry: https://user-images.githubusercontent.com/25857203/205433514-ea85e830-9044-4331-b055-995e29e1ccd2.png

PR #26690, opened 2022-12-03
soundd: lower a little more the max volume breakpoint

AlexandreSato:
Listening to loud music I was not able to hear the sound of comma3, now I'm getting it.

Test-route
Before: `075b133b6181e058|2022-12-02--21-02-53--42`
image

After: `075b133b6181e058|2022-12-03--12-26-29`
Video: https://youtu.be/RedCnd37Tkg

### DIFF
--- a/selfdrive/ui/soundd/sound.cc
+++ b/selfdrive/ui/soundd/sound.cc
@@ -48,7 +48,7 @@ void Sound::update() {
 
   // scale volume with speed
   if (sm.updated("microphone")) {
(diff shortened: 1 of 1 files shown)


sshane:
52 is not that loud for my car, leading to sounds that are too loud. Perhaps we need some kind of smart calibration. Have any ideas @incognitojam?

incognitojam:
Without being sure how loud "loud music" is, the comma three speaker is probably not able to always overpower your car's sound system :slightly_smiling_face: 

We are of course looking to get lots of feedback to help tune this, but I think #openpilot-experience on Discord is a better place for feedback instead of making hot fixes in pull requests. Thanks for sharing the plotjuggler data/segments though!

PR #26693, opened 2022-12-04
Added new fingerprint for Toyota Corolla Cross GR Sport

Iamz:
Test route: 08d7726c5efb3caa|2022-12-04--13-09-02--6


AlexandreSato:
your car have Stop And Go? just for curious

Iamz:
> your car have Stop And Go? just for curious

With stock ACC, it can slow down to 0 then give the driver 3 seconds to step on the brake because it doesn't have auto brake hold.

AlexandreSato:
After this 3sec begin a "beep beep" alarm?

Iamz:
> After this 3sec begin a "beep beep" alarm?

It beeps right after it stops. After 3 seconds, it disengage and the car creeps forward.

AlexandreSato:
>> It beeps right after it stops. After 3 seconds, it disengage and the car creeps forward.

Interesting, the LATAN made do not have Stop and Go (or ACC full range speed) and when I install my ACCType1Module they start to have this behavior in the video below.  Whistle but keep the car braked  Even if I move the gear lever from position D to N the brakes are still required by PCM.  And ours do not have an electronic park brake or auto hold
https://m.youtube.com/watch?t=3m53s&v=UYiZXbu8gPg&feature=youtu.be

Iamz:
@incognitojam Is there anything else I have to do to get this PR merged?

sshane:
Can you provide a test route where the car comes to a stop with openpilot engaged? I'd like to see it disengage, and how we handle that

Iamz:
> Can you provide a test route where the car comes to a stop with openpilot engaged? I'd like to see it disengage, and how we handle that

This is one of the segments where the car came to a stop with openpilot engaged. 
`08d7726c5efb3caa|2022-12-19--15-12-44--6`

Iamz:
@sshane Have you had a chance to take a look at the route?

sshane:
@Iamz interesting, I see you came to a stop, but the PCM did not disengage! So are you saying it only disengages when using the stock cruise control? Can you provide a route where you are using the stock cruise control and come to a stop where it disengages?

Iamz:
> @Iamz interesting, I see you came to a stop, but the PCM did not disengage! So are you saying it only disengages when using the stock cruise control?
> 
> Can you provide a route where you are using the stock cruise control and come to a stop where it disengages on its own? I would like to see how the camera cancels ACC, since it seems like the PCM is okay staying engaged at a stop when openpilot is controlling longitudinal.

@sshane Yes, only the stock cruise control disengages. Op does not. Here a segment using stock cruise control and c3 in dashcam mode. The offset is 450 seconds.
`08d7726c5efb3caa|2023-01-14--09-05-50--7`

Stock cruise control brings the car to a stop, then beeps at me to press the brake pedal. After 3 seconds, it disengages so I step on the brake pedal.

Issue #24623, opened 2022-05-23
track mypy coverage


pd0wm:
> It would also be nice if there was some way to verify that our coverage isn't going down over time. It seems like subtle changes can drastically affect the mypy coverage silently.
https://github.com/commaai/openpilot/issues/24277

JacobO1994:
Any thoughts on how the team prefers this to be done? I wouldn't mind trying to work this out.

devhsal:
Not sure if this is what you are looking for but [coveralls.io](https://coveralls.io/) is a service that provides real-time code-coverage statistics on all your repositories, total percentages covered and lines covered.

It's free for open-source projects. 

Here is an example from 2016 of the team at Zulip using the service to [achieve 100% mypy code coverage](https://blog.zulip.com/2016/10/13/static-types-in-python-oh-mypy/).

adeebshihadeh:
We just need stricter enforcement (#29458), this is a waste of time.

PR #26695, opened 2022-12-04
MyPy tracking script

natefrank48:
Changes made for this issue #24623

**Description** This scripting tool will allow MyPy coverage to be tracked in the project.

**Verification** Ran the script and verified that the output was correct.


adeebshihadeh:
Thanks for the PR, however I don't want to maintain a complicated custom script to do this. Feel free to open a new PR if you can simplify this.

PR #26704, opened 2022-12-05
Mazda: Add 2023 CX9 fw

alancyau:
DongleID: 629c4aced073d15a
Discord User: bsk#7841


bhajneet:
This branch worked for my CX-9. Thank you for doing this Alan! I highly recommend updating the fingerprinting wiki page to default to this method. Ultimately you want people to open the PR anyways, so creating the fork and skipping all the SSH stuff seems the best way to test fingerprinting.

alancyau:
@bhajneet glad it works. Re-opening this PR.

sshane:
Do you know if the min steering speed still 28 mph?

bhajneet:
> Do you know if the min steering speed still 28 mph?

How can I check for you? As far as I know comma is hardcoded to show that message. So I'm wondering if comma just doesn't even try to steer below that vs the car is unable to steer below that.

sshane:
That's correct, it doesn't try to steer below that speed so we don't know yet. You can try to comment out these two lines (or delete) to have it attempt to steer to zero: https://github.com/commaai/openpilot/blob/47977b2c0ed9f46d79e4a839ed8d83bace885504/selfdrive/car/mazda/interface.py#L44-L45

alancyau:
@bhajneet I just commented those 2 lines on the same branch. You can let the C3 pull the new update and test.
https://github.com/alancyau/openpilot/commit/89ce319f38d33ba2dc0bda60d924b608cddce29f

bhajneet:
Okay, this is my report:

- Below 32 mph, OpenPilot does not use any steering.
- After OpenPilot is engaged (green border / message), it will steer down to 28 mph.
- After going below 28 mph it will warn the driver to take control of the steering.
- If disengaged below 32 mph, OpenPilot continues to give the orange caution that there is no steering available.

I hope this helps. I'm not sure if this is a hard limitation from the Mazda hardware or something that still needs to be developed in the OpenPilot software. If it is possible to get low speed steering with some software changes, I would like to help if possible, since it will benefit my driving experience!

alancyau:
That seems to align with the current CX9. Also 23 eps fw is still the same, so I didn’t expect any change on the eps side. 

Unfortunately this is a software limitation on the eps firmware side and requires reflashing. I’ve heard people swapping eps racks before so maybe a 22 cx5 may work.

PR #26706, opened 2022-12-05
Sensord: refactor sensors_qcom2

deanlee:
1. moved sensor related data into `struct SensorState`
2.  removed sensor_service map (There is no need to look up the socket name from the map)
3.  removed the hardcoded pubsock names(build names from function `get_pubsock_names`)

### DIFF
--- a/selfdrive/sensord/sensors_qcom2.cc
+++ b/selfdrive/sensord/sensors_qcom2.cc
@@ -1,9 +1,9 @@
 #include <sys/resource.h>
 
 #include <chrono>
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
I don't really want to touch sensord without a purpose. Besides minor ones, the only refactor I think is worth doing here is trying out [libgpiod](https://kernel.googlesource.com/pub/scm/libs/libgpiod/libgpiod/+/v0.2.x/README.md). I'm not sure if it'll be cleaner though.

PR #26711, opened 2022-12-06
Chrysler: fix LKAS fault for 2017 Pacifica

### LINKED ISSUE #26701: Chrysler Pacifica 2017: LKAS error
### Describe the bug

Installed comma 3 and from the beginning I keep getting lkas fault code. It happens when I’m accelerating up to 15mph to calibrate the unit. I have worked with individuals on discord but have had no success. Also, leaving on a road trip on the 16th and would love to use the unit. Thanks

`48a73729944822df|2022-12-03--00-16-14`

### openpilot version

Release 3

### Additional info

```
'carFingerprint': 'CHRYSLER PACIFICA 2018',
```

sshane:
fritzie29#1365 on Discord has a 2017 Pacifica with newer 2020 EPS firmware, meaning that the dealer likely reflashed their EPS while taking it in for service, raising their min steer speed unfortunately. We can't gate this on platform any more, since it can be different based on the EPS firmware.

I quickly went through all our EPS firmwares, and this seems right, but will do a second check.

EDIT: False alarm, this user reports that he had his entire rack and pinion replaced, and his dongle is the only one with this updated firmware, so it's unlikely this will become a widespread phenomenon. PR now only includes his FW and abstracts the higher steering speed limit into a flag

EDIT 2: Actually I have a report of another user with a 2017 having their EPS flashed by a dealer only to find out they can't use LKAS under 30 any more. They got the dealer to flash them back and now can use at 9 mph again. The dealer told them to avoid having their EPS flashed. So it is possible for dealers to flash old EPS's, just exceedingly rare it appears.

Screenshot from 2022-12-05 22-03-21

### DIFF
--- a/docs/CARS.md
+++ b/docs/CARS.md
@@ -22,10 +22,10 @@ A supported vehicle is one that just works when you install a comma three. All s
 |Chevrolet|Bolt EV 2022-23|2LT Trim with Adaptive Cruise Control Package|openpilot available[<sup>1</sup>](#footnotes)|3 mph|6 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-empty.svg)](##)|GM|
 |Chevrolet|Silverado 1500 2020-21|Safety Package II|openpilot available[<sup>1</sup>](#footnotes)|3 mph|6 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-empty.svg)](##)|GM|
 |Chevrolet|Volt 2017-18[<sup>3</sup>](#footnotes)|Adaptive Cruise Control (ACC)|openpilot|0 mph|6 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-empty.svg)](##)|OBD-II|
-|Chrysler|Pacifica 2017-18|Adaptive Cruise Control (ACC)|Stock|0 mph|39 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-full.svg)](##)|FCA|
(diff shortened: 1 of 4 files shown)


docs/CARS.md
```diff
@@ -22,10 +22,10 @@ A supported vehicle is one that just works when you install a comma three. All s
 |Chevrolet|Bolt EV 2022-23|2LT Trim with Adaptive Cruise Control Package|openpilot available[<sup>1</sup>](#footnotes)|3 mph|6 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-empty.svg)](##)|GM|
 |Chevrolet|Silverado 1500 2020-21|Safety Package II|openpilot available[<sup>1</sup>](#footnotes)|3 mph|6 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-empty.svg)](##)|GM|
 |Chevrolet|Volt 2017-18[<sup>3</sup>](#footnotes)|Adaptive Cruise Control (ACC)|openpilot|0 mph|6 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-empty.svg)](##)|OBD-II|
-|Chrysler|Pacifica 2017-18|Adaptive Cruise Control (ACC)|Stock|0 mph|9 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-full.svg)](##)|FCA|
+|Chrysler|Pacifica 2017-18|Adaptive Cruise Control (ACC)|Stock|0 mph|39 mph|[![star](assets/icon-star-full.svg)](##)|[![star](assets/icon-star-full.svg)](##)|FCA|
```
sshane, line 25: older cars are starting to get flashed to newer firmware, and might eventually outnumber the old FW cars. this now under promises, over delivers for users rather than the opposite

sshane:
False alarm, this user reports that he had his entire rack and pinion replaced, and his dongle is the only one with this updated firmware, so it's unlikely this will become a widespread phenomenon.

PR #26714, opened 2022-12-06
Toyota: Don't Set Standstill Bit on TSS-P Prius #26671

cydia2020:
Reopening #26671 as per request.


sshane:
I discovered we already have a roll-back problem in master. If you are stopped on a steep hill in standstill behind a car, and press the gas or press resume, the car will leave standstill and start to roll back indefinitely as we never enter standstill again, and -2.0 ms/s/s isn't enough brake to stay stopped on all hills (-3.5 m/s/s worked nicely/similar to standstill)

PR #26721, opened 2022-12-07
Toyota: match stock standstill entrance behavior

sshane:
At the moment, we only enter standstill once -- when coming to a stop. If the user presses the resume button, we never enter standstill again -- despite that car **NOT** being in the `NO_STOP_TIMER_CAR` list, which is a divergence from the stock system, and half-defeats the purpose of the list.

This matches stock:
- User presses resume when car (openpilot) doesn't want to move
- Car leaves standstill for a second and goes back into standstill

Cars like the Prius still might momentarily roll back, but this is better than rolling back in a steady state.

Whenever the car wants to move, we allow releasing standstill.

TODO: 
- [ ] need to test if we need a timer like stock

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -27,7 +27,7 @@ class CarController:
     self.last_steer = 0
     self.alert_active = False
     self.last_standstill = False
-    self.standstill_req = False
+    # self.standstill_req = False
     self.steer_rate_counter = 0
 
     self.packer = CANPacker(dbc_name)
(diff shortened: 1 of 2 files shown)


sshane:
Prius's PCM does not respect our `RELEASE_STANDSTILL` (can't be used to put the car into a standstill), so this PR is useless. The real solution is requesting max brake, as -2.0 is not enough to keep the car stopped.

cydia2020:
> Prius's PCM does not respect our `RELEASE_STANDSTILL`

I think it does, mine doesn't go into standstill if this bit is untouched when the vehicle comes to a stop.

My car behaves like:

1. When this bit is 0, pcm will go into waiting mode (cruise "up" button locked) when coming to a stop and keep the car in that mode.
2. When the bit is always 1, then the car won't go into standstill, and will be able to roll forward and accelerate once openpilot starts requesting acceleration.
3. When the bit is set to 0 when coming to a stop, and then set to 1 when the lead car moves, the car exits "waiting mode", goes into "press cruise lever or accelerator to resume" mode, during which time the driver can use the "up" button to resume cruise control. (Stock behaviour)

sshane:
Yeah I got that wrong but didn't update my comment, that's how TSS2 operates too.

sshane:
Have the Prius this week, will try to get some form of this merged soon!

PR #26733, opened 2022-12-09
Toyota: don't log standstill for NO_STOP_TIMER_CAR

### LINKED ISSUE #26723: Standstill message causing car not to continue anymore
### Describe the bug

RAV4 with a SDSU never had this issue of hitting the gas or resume to accelerate after stopped. Now the standstill message shows up and car does not continue when the lead car goes. Running master ci but someone in discord still said to report issue here. 

2018 Toyota rav4 hybrid

0fd770798a15a939|2022-12-07--13-30-43--0

### openpilot version

0.9.1/master-ci/c5a0289/dec 06

### Additional info

Experimental mode and disengage on gas were off.

sshane:
Fixes a regression introduced in https://github.com/commaai/openpilot/pull/26586 where we started always logging standstill. The Toyotas in this list behave differently than most other cars with standstill, as they require no special resume spam to exit — it's entirely automatic with an acceleration cmd which we code to avoid in longcontrol for wind-up reasons.

TODO:
- [ ] Add standstill logging back but don't regress auto-start in a future PR

### DIFF
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -124,7 +124,7 @@ class CarState(CarStateBase):
       self.low_speed_lockout = cp.vl["PCM_CRUISE_2"]["LOW_SPEED_LOCKOUT"] == 2
 
     self.pcm_acc_status = cp.vl["PCM_CRUISE"]["CRUISE_STATE"]
-    if self.CP.carFingerprint not in NO_STOP_TIMER_CAR:
+    if self.CP.carFingerprint not in (NO_STOP_TIMER_CAR - TSS2_CAR):
       # ignore standstill state in certain vehicles, since pcm allows to restart with just an acceleration request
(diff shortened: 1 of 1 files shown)


selfdrive/car/toyota/carstate.py
```diff
@@ -124,7 +124,9 @@ def update(self, cp, cp_cam):
       self.low_speed_lockout = cp.vl["PCM_CRUISE_2"]["LOW_SPEED_LOCKOUT"] == 2
 
     self.pcm_acc_status = cp.vl["PCM_CRUISE"]["CRUISE_STATE"]
-    ret.cruiseState.standstill = self.pcm_acc_status == 7
+    if self.CP.carFingerprint in NO_STOP_TIMER_CAR:
```
sshane, line 127:
```suggestion
    if self.CP.carFingerprint not in NO_STOP_TIMER_CAR:
```

PR #26734, opened 2022-12-09
nav: add mapRenderState

adeebshihadeh:
Analogous to the cameraState packets for debugging nav with qlogs.

### DIFF
--- a/selfdrive/navd/map_renderer.cc
+++ b/selfdrive/navd/map_renderer.cc
@@ -178,7 +178,7 @@ void MapRenderer::publish(const double render_time) {
   // Send state msg
   MessageBuilder msg;
   auto state = msg.initEvent().initMapRenderState();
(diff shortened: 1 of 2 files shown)


selfdrive/navd/map_renderer.cc
```diff
@@ -139,7 +141,7 @@ void MapRenderer::sendVipc() {
   VisionBuf* buf = vipc_server->get_buffer(VisionStreamType::VISION_STREAM_MAP);
   VisionIpcBufExtra extra = {
     .frame_id = frame_id,
-    .timestamp_sof = ts,
+    .timestamp_sof = sm->rcv_time("liveLocationKalman"),
```
adeebshihadeh, line 144: @mitchellgoffpc let's use this in `modeld` to detect stale data

PR #26755, opened 2022-12-10
Hyundai: Add FW Versions for Southeast Asia 2022 Ioniq 5 HDA1

sunnyhaibin:
Add firmware for the Malaysia (probably most Southeast Asia ones) 2022 Hyundai Ioniq 5 HDA1.

This Ioniq 5 HDA1 (Malaysia) CAN traffic is the same as US/EUR's Ioniq 5 HDA2. No `Ecu.adas` was queried. Changing `CANFD_HDA2` detection to if `0x50 | LKAS` is on bus 6.

Harness used: Hyundai Q (Hyundai K with flipped CAN bus pairs)

Route ID: `6837b3afa7250239|2022-12-11--03-17-43`

Thanks to community Malaysian 2022 Hyundai Ioniq 5 HDA1 owner `NinjaBoss
#8617 (Discord)`.


sunnyhaibin:
User mentioned that there is a discrepancy between the car cluster's set speed to openpilot's "MAX" set speed when using stock longitudinal:
- Car cluster: 50 km/h
- openpilot "MAX": 80 km/h

selfdrive/car/hyundai/interface.py
```diff
@@ -29,7 +29,7 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long):
 
     if candidate in CANFD_CAR:
       # detect HDA2 with ADAS Driving ECU
-      if Ecu.adas in [fw.ecu for fw in car_fw]:
+      if 0x50 in fingerprint[6]:
```
sshane, line 32:
```suggestion
      if Ecu.adas in [fw.ecu for fw in car_fw]:
```

We verified the car does actually return an adas ecu, we just need to make sure it's consistent (failed the first time without any errors logged)

sshane:
Can we add a footnote, or perhaps update the titles to reflect international cars needing the harness for HDA2?

selfdrive/car/hyundai/values.py
```diff
@@ -172,6 +172,7 @@ def init_make(self, CP: car.CarParams):
   CAR.VELOSTER: HyundaiCarInfo("Hyundai Veloster 2019-20", min_enable_speed=5. * CV.MPH_TO_MS, harness=Harness.hyundai_e),
   CAR.SONATA_HYBRID: HyundaiCarInfo("Hyundai Sonata Hybrid 2020-22", "All", harness=Harness.hyundai_a),
   CAR.IONIQ_5: [
+    HyundaiCarInfo("Hyundai Ioniq 5 (Southeast Asia only) 2022-23", "All", harness=Harness.hyundai_q),
```
sshane, line 175: Likely applies to Malaysia, Singapore, Indonesia and Thailand

sshane:
@adeebshihadeh are you fine with merging as is? a footnote on one of the entries would be a little confusing to me.

PR #26756, opened 2022-12-10
[WIP] UI: add replay capabilities

### LINKED ISSUE #26810: replay: make it work on device
An error in cameraview.cc is thrown on this line: https://github.com/commaai/openpilot/blob/08b4cff484940b394fa64d89dad65330bcb51a4a/selfdrive/ui/qt/widgets/cameraview.cc#L328

deanlee:
much still remains to be done to run it on device, such as:
- [ ] the `SubMaster::update_msgs` is used to update events from replay when in replay mode. but we still need to call sm->update(0) in ui/update_sockets to monitor the events from other modules (e.g. stop replay if ignition is on).
- [ ] fix  #26810

Screenshot from 2022-12-11 03-29-10

Screenshot from 2022-12-11 02-40-22


ntegan1:
Why?

PR #26758, opened 2022-12-11
HKG CAN-FD: common distance unit signal

sunnyhaibin:
Identified in https://github.com/commaai/openpilot/pull/26755 and https://github.com/commaai/openpilot/pull/26635, `DISTANCE_UNIT` from `0x4D8 | CLUSTER_INFO` is not used on all CAN-FD cars, causing some CAN-FD cars to convert the set speed to an incorrect value.

On the Malaysian Ioniq 5 (https://github.com/commaai/openpilot/pull/26635) and comma's EV6 HDA2, @sshane and I verified that the 2 cars can use the following message to correctly identify the distance unit the car is using:
- `DISTANCE_UNIT` from `0x1AA | CRUISE_BUTTONS_ALT` (from https://github.com/commaai/opendbc/pull/753)


sshane:
Verified with 32 unique dongles and segments from these platforms:

```python
{'HYUNDAI IONIQ 5 2022': 890, 'KIA EV6 2022': 565, 'KIA SPORTAGE HYBRID 5TH GEN': 143, 'HYUNDAI SANTA CRUZ 1ST GEN': 4, 'KIA SPORTAGE 5TH GEN': 37, 'HYUNDAI TUCSON HYBRID 4TH GEN': 67, 'GENESIS GV70 1ST GEN': 59}
```

The new signal matches the old, except for the Sorento which changes from metric to imperial, and one EV6 in Taiwan (`7c4c77e442583f42`) which changes from imperial to metric.

sshane:
Diff is from the higher freq signal giving us an update earlier

Screenshot from 2022-12-12 19-46-03

sshane:
thanks for opening the PR!

PR #26759, opened 2022-12-11
SANTA_FE_2022: add missing FW versions

chrihart:

**Checklist**
- [ ] route with openpilot:


incognitojam:
Since it seems like you're just adding new fingerprints for an existing car, you don't need to provide a test route or anything, just share your dongle ID so that we can verify that it's correct

chrihart:
> As you're just adding new fingerprints for an existing car, you don't need to provide a test route or anything. Can you post your dongle ID?

Sure thing - db56bda387c66164

PR #26772, opened 2022-12-12
EV6: switch blinker signal

adeebshihadeh:
The lamp signal previously used naturally includes openpilot-commanded blinkers. In the future, we'll probably want to split blinker user command from the lamp state in carState.


adeebshihadeh:
Looks like this messages isn't all on the cars.

PR #26779, opened 2022-12-13
Cabana: add signal value filters to the logs view

deanlee:
Remove "Find signal value" button and dialog from `DetailView`. add dynamic value filters to the logsView to filtering in all CAN events in real time:

[Kazam_screencast_00002.webm](https://user-images.githubusercontent.com/27770/207235947-d796783b-1be2-48b1-a969-c15c17ee94bc.webm)


adeebshihadeh:
Nice, can we also add a "!="?

deanlee:
> Nice, can we also add a "!="?

added: https://github.com/commaai/openpilot/pull/26780

PR #26794, opened 2022-12-14
carla sim: fix accelerometer sensor data

ebadi:
Fix incorrect value for accelerometer.type introduced in 29d3ed2ce63a65f793dc0ecb553180a0d0fba03e


adeebshihadeh:
Thanks!

Issue #26503, opened 2022-11-15
Carla simulation not available in Openpilot UI


franco126:
### Describe the bug

When I run the simulation on PC, Carla "interface" in Openpilot UI is not shown. 
So I can't see car or road.
![Schermata del 2022-11-15 13-21-13](https://user-images.githubusercontent.com/51539908/201918862-e3574c17-a060-4c33-bc22-927366841b56.png)
When I press 1 in bridge.py shell, it says "Openpilot available press set to engage. 
In launch_openpilot.sh shell seems to be everything ok.
Even if I run in docker container, I get the same issue
Here some screenshot...
![Schermata del 2022-11-15 13-26-40](https://user-images.githubusercontent.com/51539908/201919587-1305f1f3-e65e-424c-aa54-10ed1035e908.png)
![Schermata del 2022-11-15 13-27-01](https://user-images.githubusercontent.com/51539908/201919591-b51260bf-6736-4302-a99f-462dc932f65f.png)

### OS Version

Ubuntu 20.04

### Steps to reproduce

- Terminal 1: run CARLA (not docker version)
- Terminal 2: poetry shell - cd /tools/sim - ./launch_openpilot.sh
- Terminal 3: poetry shell - cd tools/sim - ./bridge.py

Ruishenl:
I ran into the exact same issue. I am on Ubuntu22

eli-davis:
same issue, ubuntu 20, followed all directions...doesnt show up on ui

jbeck9:
Same. Openpilot seems to run normally, just nothing on the UI. This is running Carla and OpenPilot through the Docker images for me.

I'm guessing it is related to #26169, but that is just a guess.

didpurwanto:
Same here. Carla is working fine but the bridge in 0.9.0 version has nothing.

joeking11829:
Same issue here.

fatsheep2020:
same issue here

didpurwanto:
Any updates about this?
In my case rn, I can see the future trajectory, but cannot load the images from Carla.

ebadi:
> Same. Openpilot seems to run normally, just nothing on the UI. This is running Carla and OpenPilot through the Docker images for me.

> same issue, ubuntu 20, followed all directions...doesnt show up on ui

@jbeck9 and @eli-davis, how do you know that openpilot is working?

For me OP seems not to be **fully** engaged after I press `1`. 
[...]

Issue #26595, opened 2022-11-24
Carla Openpilot Not Working with the latest docker image


Ruishenl:
### Describe the bug

I am following the instruction to get the carla-openpilot integration working, but the dashboard doesn't seem to render the carla world. I am able to pass my keyboard input to engage openpilot but got Openpilot Unavailable message:
![Screenshot from 2022-11-22 16-31-08](https://user-images.githubusercontent.com/22068468/203853198-c31db90d-be7a-4a3f-963d-862b6456466d.png)

Carla seems to be having memory error as well but no sure if it's related : 
`Signal 11 caught.
Malloc Size=65538 LargeMemoryPoolOffset=65554 
CommonUnixCrashHandler: Signal=11
Engine crash handling finished; re-raising signal 11 for the default handler. Good bye.
./CarlaUE4.sh: line 5:    13 Segmentation fault      (core dumped) "$UE4_PROJECT_ROOT/CarlaUE4/Binaries/Linux/CarlaUE4-Linux-Shipping" CarlaUE4 "$@`

Steps to reproduce:
Terminal 1
./start_carla.sh

Terminal 2 
./start_openpilot_docker.sh

Terminal 3:
docker exec -it OPENPILOT_CONTAINER_ID /bin/bash
./launch_openpilot.sh

Terminal 4
docker exec -it OPENPILOT_CONTAINER_ID /bin/bash
./bridge.py

### OS Version

Pop!_OS 22.04

### openpilot version or commit

b15018387c774b9f470e88f22b96861d62917656

### Additional info

I also have a warning message showing the current branch is not tested. Am i supposed to use a released tag instead ?
Also tried running the integration_tests in the container but failed

fatsheep2020:
@incognitojam  do it simulator issue? anddo you know how to fix it, as i knew, it work well before....
thanks a lot!

incognitojam:
This is the same issue as #26503

zeroyang0809:
hello do you fix this problem? could you show solutions to fix it? 
thanks

Issue #24689, opened 2022-05-31
Remove WideCameraOnly param


pd0wm:
VisionIPC should communicate which streams are available when connecting.

adeebshihadeh:
@deanlee want to take care of this one? we just have to add something to VIPC to know which streams are available, then use that in the UI to replace the WideCameraOnly param.

deanlee:
how about keeping the existing interface and only add a static function `getAvailableStreams` to vipc_client?  https://github.com/commaai/cereal/pull/400

PR #26796, opened 2022-12-14
carla sim: fix missing camera

ebadi:
Passing ` --high_quality --dual_camera` to `bridge.py` solved the camera issue for me. 

image

Closes:
#26503 #26595

### DIFF
--- a/tools/sim/tmux_script.sh
+++ b/tools/sim/tmux_script.sh
@@ -2,5 +2,5 @@
 tmux new -d -s carla-sim
 tmux send-keys "./launch_openpilot.sh" ENTER
 tmux neww
-tmux send-keys "./bridge.py $*" ENTER
+tmux send-keys "./bridge.py --high_quality --dual_camera $*" ENTER
 tmux a -t carla-sim


incognitojam:
is --high_quality required for this to work?

ebadi:
@incognitojam  `--high_quality` is *not* needed, but I think `UI` should run in `--high_quality` . 

Reducing simulator quality is introduced for CI purpose ([Simulator: Fix CI and set low_quality default](https://github.com/commaai/openpilot/commit/ee433dfa57a60364423a222a7f18dc137b77f532)). 

For CI, OP still run in low quality by default:
https://github.com/commaai/openpilot/blob/master/tools/sim/tests/test_carla_integration.py#L45

jeroenlammersma:
@ebadi correct, `--high_quality` is indeed not needed. Just now I tried for myself and passing the `--dual_camera` flag does the trick.

But unfortunately, there is currently an issue with steering as well.

ebadi:
> @ebadi correct, `--high_quality` is indeed not needed. Just now I tried for myself and passing the `--dual_camera` flag does the trick.
> 
> But unfortunately, there seems to be an issue with steering currently as well.

@jeroenlammersma  Great that it works for you.   Here (#26829) I am also working on refactoring `bridge.py` . You might be interested to try it out. If this one also doesn't work, please open a separate issue.

jeroenlammersma:
> @jeroenlammersma Great that it works for you. Here (#26829) I am also working on refactoring `bridge.py` . You might be interested to try it out. If this one also doesn't work, please open a separate issue.

Looks good, great addition! Currently, I have a vacation, but when I am back I will try it out ;-).

Furthermore, perhaps interesting for you, at the moment I am supervising two student groups at the Hanze University of Applied Sciences that are improving driving with openpilot using the (CARLA) simulator as well. Their assignments are:
- Improve manual driving by implementing [manual_control.py](https://github.com/carla-simulator/carla/blob/master/PythonAPI/examples/manual_control.py) (and if time allows, [manual_control_steeringwheel.py](https://github.com/carla-simulator/carla/blob/master/PythonAPI/examples/manual_control_steeringwheel.py)) from CARLA in the bridge. This allows multiple key presses at once, as well as more controls (like blinkers, lights, etc). 
- Make Driving Monitoring (DM) (optionally) available in the bridge and make use of a connected webcam.

Because openpilot is (currently) illegal to use in the Netherlands, our aim is to experiment in the simulator. Next to that, we want to see if we can use the comma three and openpilot as base for the development of a virtual driving coach. So, that's why we need DM to work in a simulated environment.

If you want, I can keep you updated when their projects have some working protoypes.

Cheers, Jeroen

ebadi:
> > @jeroenlammersma Great that it works for you. Here (#26829) I am also working on refactoring `bridge.py` . You might be interested to try it out. If this one also doesn't work, please open a separate issue.
> 
> Looks good, great addition! Currently, I have a vacation, but when I am back I will try it out ;-).
> 
> Furthermore, perhaps interesting for you, at the moment I am supervising two student groups at the Hanze University of Applied Sciences that are improving driving with openpilot using the (CARLA) simulator as well. Their assignments are:
> - Improve manual driving by implementing [manual_control.py](https://github.com/carla-simulator/carla/blob/master/PythonAPI/examples/manual_control.py) (and if time allows, [manual_control_steeringwheel.py](https://github.com/carla-simulator/carla/blob/master/PythonAPI/examples/manual_control_steeringwheel.py)) from CARLA in the bridge. This allows multiple key presses at once, as well as more controls (like blinkers, lights, etc). 
> - Make Driving Monitoring (DM) (optionally) available in the bridge and make use of a connected webcam.
> 
> Because openpilot is (currently) illegal to use in the Netherlands, our aim is to experiment in the simulator. Next to that, we want to see if we can use the comma three and openpilot as base for the development of a virtual driving coach. So, that's why we need DM to work in a simulated environment.
> 
> If you want, I can keep you updated when their projects have some working protoypes.
> 
> Cheers, Jeroen
> 

Interesting 🤔. I am on vacation until 4-5th January and until then i have no access to my laptop (using mobile phone now).  When I am back i will share interesting resources/pointers with you.

adeebshihadeh:
Both flags, `--high_quality --dual_camera`, were added since the CARLA system requirements are much higher than openpilot's. Closing since I think leaving both off is still the best default. #24689 describes the real fix.

> But unfortunately, there seems to be an issue with steering currently as well.

We have a CI test to ensure we don't entirely break the simulator, and eventually we want to have a test for the driving quality.

zeroyang0809:
hello i meet the problem "openpilot unavailable waiting for controls to start"
this problem is familiar with your problem
do you know how to fix it ?
thx

Issue #24220, opened 2022-04-15
Add comma API to athenad


pd0wm:
Respond to a subset of the comma api (routes, segments, etc) to make our tools work. Can be used by setting `API_HOST` to your device's IP. This can later be expanded by using athena to discover the ip automatically, and start the server on demand.

PR #26804, opened 2022-12-15
on device video server

ntegan1:
TODO

¿add button to download full concatenated video thing for route? [here](https://github.com/ntegan1/apilot/commit/e3fa90a5c19bce45b92bdbe2afa0a30af89d7079)
¿crop dcam to steering wheel and overlay on ecam? [here](https://github.com/ntegan1/apilot/commit/3adfedb991d41150c478034f6b1029813ce558a1)

Thanks royjr for the simplification and clean html rewrite!


ntegan1:
https://user-images.githubusercontent.com/3820605/207763655-30d02287-3311-4600-b9ec-53e694874f60.mp4

viewing in PC
this browser doesn't support h265 so only qcamera here

ntegan1:
https://user-images.githubusercontent.com/3820605/207764241-9d4d1ea9-c314-4f2a-9bf8-bb5c7eeffb30.mp4

works better on ios and android

royjr:
I love this.. but seems a but too convoluted. I tweaked into a (maybe too simple) version of this [here](https://github.com/commaai/openpilot/compare/master...royjr:openpilot:vidserver). Definitely should be under tools.

adeebshihadeh:
Thanks for the PR, but the real solution here is directly serving up the videos to connect from the device (#24220).

PR #26805, opened 2022-12-15
controlsd: set latActive with max minimum steer speed

### LINKED ISSUE #26694: Steering bug: Steering stops working at very low speeds in experimental mode.
### Describe the bug

2022 Arteon in experimental mode:

When at very low speeds or often stopping for a stop sign or another car, Openpilot displays a warning "Steering temporarily unavailable" and won't start working again even at higher speeds until the device is rebooted. At the same time, my car displays the message "Take control of the wheel" like it used to with my stock lane keeping. 

`200d2fe234e1fef3|2022-12-03--18-50-25`

### openpilot version

0.9.0

```
'carFingerprint': 'VOLKSWAGEN ARTEON 1ST GEN',

'gitBranch': 'mqb-fw-dukajuke',
'gitCommit': 'c1524b4b082de8bf9091193c5682b45280ae3105',
'gitRemote': 'https://github.com/jyoung8607/openpilot.git',
```

### Additional info

_No response_

jyoung8607:
Don't set CC.latActive (and by extension, the car port LKAS torque active control bits) if we're not truly steering.

Per discussion in #25761
Resolves #26694


selfdrive/controls/controlsd.py
```diff
@@ -569,9 +569,11 @@ def state_control(self, CS):
 
     CC = car.CarControl.new_message()
     CC.enabled = self.enabled
+
     # Check which actuators can be enabled
+    min_lat_active_speed = self.CP.minSteerSpeed if self.joystick_mode else max(self.CP.minSteerSpeed, MIN_LATERAL_CONTROL_SPEED)
     CC.latActive = self.active and not CS.steerFaultTemporary and not CS.steerFaultPermanent and \
-                   CS.vEgo > self.CP.minSteerSpeed and not CS.standstill
+                   CS.vEgo > min_lat_active_speed and not CS.standstill
```
sshane, line 576: joystick never really supported steering at 0, since 0 is not > 0, but I get it can be useful. Revert the joystick exception in this PR and do separately?

jyoung8607: Ah, I thought `minSteerSpeed` defaulted to -1 like `minEnableSpeed`.

sshane: Oh true true, the standstill check actually enforces this

jyoung8607:
No apparent problems on a test drive, ready for review.

Route: `3cfdec54aa035f3f|2022-12-15--19-01-14`

sshane:
hmm I had forgotten about the body

jyoung8607:
Want to gate the change on `not notCar` (heh) just like I was previously gating on `self.joystick_mode`?

sshane:
No I think you were right, checking joystick mode makes more sense, since this magic speed comes from the latcontrol class

PR #26807, opened 2022-12-15
GM: factor on cluster speed to match dash

morrislee:
This factor seems to match GMC Sierra 1500 cluster speeds pretty good. GM may be using some filtering for cluster speed, it does not fluctuate as much as OP's displayed speeds

### DIFF
--- a/selfdrive/car/gm/carstate.py
+++ b/selfdrive/car/gm/carstate.py
@@ -43,6 +43,7 @@ class CarState(CarStateBase):
     )
     ret.vEgoRaw = mean([ret.wheelSpeeds.fl, ret.wheelSpeeds.fr, ret.wheelSpeeds.rl, ret.wheelSpeeds.rr])
     ret.vEgo, ret.aEgo = self.update_speed_kf(ret.vEgoRaw)
+    ret.vEgoCluster = ret.vEgo * 1.00675
     # sample rear wheel speeds, standstill=True if ECM allows engagement with brake
     ret.standstill = ret.wheelSpeeds.rl <= STANDSTILL_THRESHOLD and ret.wheelSpeeds.rr <= STANDSTILL_THRESHOLD
 


morrislee:
I have got to admit, I am not sure if this is the right approach to fix OP showing slightly a slower speeds than factory cluster, I am open to suggestions

Verylukyguy:
I have found that using a GPS device on my 2019 Chevy Malibu, that the cluster is almost always 1 MPH slower.  Matching the speed for conformity probably isn’t necessary.

morrislee:
> I have found that using a GPS device on my 2019 Chevy Malibu, that the cluster is almost always 1 MPH slower. Matching the speed for conformity probably isn’t necessary.

I appreciate your input, I have checked my cluster with multiple GPS speeds, it is pretty much on the dot. It was slower than it says it is until I gone with slightly bigger tires. Now its a perfect match. It just bothers my OCD that OP reports a slightly different speed than the cluster

sshane:
btw hysteresis support is enabled by setting these two variables to appropriate values: https://github.com/commaai/openpilot/blob/08b4cff484940b394fa64d89dad65330bcb51a4a/selfdrive/car/toyota/carstate.py#L20

sshane:
I'll check on our Bolt EUV and LGTM. Maybe only for GM Cam, unless @Verylukyguy can try to find what his is.

I'd also like to make sure that the dash doesn't already publish the UI speed on CAN somewhere first.

sshane:
Think I found the cluster signal the dash uses: https://github.com/commaai/opendbc/pull/780. May close this PR if it works out

sshane:
Unfortunately that signal is noisy in some situations, and half the rate of the dash. It also looks like some cars have different factors. This is another Silverado, but the factor is less than 1:

Screenshot from 2023-02-10 13-15-02

adeebshihadeh:
> Unfortunately that signal is noisy in some situations, and half the rate of the dash. It also looks like some cars have different factors. This is another Silverado, but the factor is less than 1:
> 
> Screenshot from 2023-02-10 13-15-02

Can we just use that to determine the factor?

sshane:
Possibly, I'll have to make sure that the noisiness doesn't affect anything.

sshane:
0.5 kph hysteresis gap was around the sweet spot to improve all the GM camera ACC platforms from a mean speed diff 0.34 -> 0.31 m/s after hysteresis.

- Silverado: `Mean speed diff 0.27 -> 0.24 m/s after hysteresis`
- Bolt: `Mean speed diff 0.35 -> 0.33 m/s after hysteresis`

With a 1.01 factor + hysteresis:

- Silverado: `Mean speed diff 0.27 -> 0.26 m/s after hysteresis`
- Bolt: `Mean speed diff 0.35 -> 0.22 m/s after hysteresis`
So while a factor improves Bolts significantly, there are some Silverados with factors of 1 or less, so we can't use a static factor across the board.

More plots: https://imgur.com/a/xnGL2Rg

sshane:
Adding hysteresis in this PR since I can't edit the base branch: https://github.com/commaai/openpilot/pull/27301

PR #26812, opened 2022-12-16
HKG CAN-FD: set cruise state available with car signal

sunnyhaibin:
Matching HKG CAN platform's `ret.cruiseState.available` state check when stock longitudinal is used.


adeebshihadeh:
I'm not sure we want to do this for these cars, since the main button also engages cruise along with "main enable". We probably want a lower level signal from something like the TCS that says whether cruise will engage if the main button is pressed.

sshane:
We have a signal from TCS that tells us if there are any faults preventing engagement, it's what we log for accFaulted. We also have main-enable cars on the CAN platform, so we'd need to differentiate those ideally if we want to do this. TCS allowing engagement doesn't tell us if the SCC system will allow engagement (foot on brake, below engage speed)

adeebshihadeh:
> We have a signal from TCS that tells us if there are any faults preventing engagement, it's what we log for accFaulted. We also have main-enable cars on the CAN platform, so we'd need to differentiate those ideally if we want to do this. TCS allowing engagement doesn't tell us if the SCC system will allow engagement too (below engage speed)
> 
> To me, showing available when set or resume do not work might be confusing

Agree with everything here, except the combination of openpilot no entries + TCS allowed should be a pretty good indicator. If it isn't, let's fix what's missing.

sshane:
There might be an SCC available signal, I'll check it out

adeebshihadeh:
Sounds good, I'll convert to draft.
- [ ] CAN platform
- [ ] CAN-FD platform

sshane:
@adeebshihadeh I don't like conflating a good PR like this with extra stuff. I'm going to merge to bring CAN FD up to the current standard we use with pause/resume cars, then open an issue for the behavior change.

sshane:
Behavior changes:
- Available:
  - OP long:
    - No change (always available assuming no faults)
  - Stock long:
    - Always -> Main On
- Enabled:
  - OP long:
    - Never -> when enabled
  - Stock long:
    - No change

PR #26822, opened 2022-12-18
New Mazda Gen 4 Port

MoreTore:
**Checklist**
- [x] route with openpilot: 164080f7933651c4|2022-12-02--08-05-29--37
- [x] route with stock system: 164080f7933651c4|2022-12-01--16-19-34


MoreTore:
@sshane what are the odds we get this port added to release 9.1?

sshane:
Is this port ready for review? Everything should be in a working state? What actual users do you have and for which cars?

MoreTore:
> ?

Only thing is there is an issue with the can_ignition signal. The issue is that the signal I'm using stays on for a few seconds after the car is off, so in tmux I can see missing signals and when I start the car shortly after shutting it off I get a can bus disconnected event on the screen. I think the solution would be to use another signal that goes silent when the car is off. 

I'm going to push one more commit for tuning and then it's good.

MoreTore:
> ?

Just wondering what the deadline is to get it in 9.1. it's basically ready just need to finish panda safety limits.

MoreTore:
@sshane  maybe you can figure out what im doing wrong with my panda safety. 

const SteeringLimits MAZDA_2019_STEERING_LIMITS = {
  .max_steer = 8000,
  .max_rate_up = 45,
  .max_rate_down = 80,
  .max_rt_delta = 3500,
  .max_rt_interval = 250000,
  .driver_torque_factor = 1,
  .driver_torque_allowance = 1500,
  .type = TorqueDriverLimited,
};

if (steer_torque_cmd_checks(desired_torque, -1, MAZDA_2019_STEERING_LIMITS)) 

is failing.

sshane:
I can check it out when I'm back. Nice work on the port so far!

MoreTore:
Thank you thank you!

MoreTore:
Once this is merged I will add the auto-resume logic because the stock system requires a button press. I already have long control working but I feel it needs some tweaking and testing. Namely the following distance, stopping distance, and the AEB signal needs to be tested somehow.

Side note, I am hoping you guys will add more variables in cereal to dynamically change following distance and comfort-acceleration settings.  Another big issue for me is when accelerating from a stop when the roads are slippery the tires lose traction and OP just floors it 💯 !

MoreTore:
I figured out my issue with the panda safety. It was a type casting issue.

MoreTore:
I am having a lot of trouble with the torque tuning. It works pretty good at high speeds with some very minor ping pong, but during a low speed turn it will jerk the wheel and I need to take over. I tried playing with the low_speed_factor table but that just makes it not steer enough. It's like the self centering force is high because as soon as the steer command backs off the steering wheel straightens out fast.

adeebshihadeh:
Does this Mazda platform have a steering lockout?

MoreTore:
> Does this Mazda platform have a steering lockout?
Yes if you turn the wheel beyond a certain angle the cruise control shut off. The angle is quite high; it is only an issue when making low-speed 90-degree turns.

MoreTore:
I found a way around the steering angle limit but it requires openpilot to handle the cruise state. I would think that's not allowed for safety reasons.

MoreTore:
I found a good simple solution to the non-linear steering torque problem.  I created a new function torque_from_lateral_accel_non_linear(). Very simple modification and now I have perfect a steering tune.

sshane:
@MoreTore Are you on Discord? Do these cars have any LKAS lockouts?

MoreTore:
No you guys ban me lol! Yea there is a lockout timer. Personally I have bypassed them with a ti. Just leave the car in dash cam mode for now.

sshane:
Have you found a way around the lockout, what does it do?

adeebshihadeh:
The author has been banned for advertising a torque interceptor; see [this post](https://blog.comma.ai/safer-control-of-steering/#torque-interceptor-safety-critical-hw-required) for some context. If you or someone else would like to pick up this work, feel free to fork and open a new PR.

PR #26825, opened 2022-12-20
Nav: allow engage without a nav route

royjr:


adeebshihadeh:
Thanks for the PR, but taco2 is frozen and not really good enough for daily driving anyway.

PR #26826, opened 2022-12-20
Add rlog CAN to csv script

mlouielu:
**Description** 

This is a script to convert rlog CAN message to csv file.

Output can be used by `panda/examples/can_bit_transition.py`.


adeebshihadeh:
Thanks for the PR, but I don't really want to further this pattern of two different log types. Instead, I'd merge a PR that moves those panda scripts to openpilot and makes them take rlogs.

PR #26830, opened 2022-12-22
Toyota: make LKAS message 5hz

cydia2020:
As per request in #24782, 5 hz does introduce some delays when the user operates the steering wheel buttons. To mitigate this, I have this rate at 20 hz on my device, it also worked fine on my Prius when I set it to 100 Hz.


adeebshihadeh:
@sshane this seems straightforward. any reason hasn't been merged?

PR #26834, opened 2022-12-23
Cabana: add tool to find similar bits

deanlee:
menu->tools->find similar bits:

[Kazam_screencast_00013.webm](https://user-images.githubusercontent.com/27770/209356267-1d4e160a-d5f0-48d7-955a-a4fbdc58f483.webm)

Merry Christmas To You All :)


deanlee:
@sshane would this be OK?

adeebshihadeh:
> Merry Christmas To You All :)

Merry Christmas :)


I like the addition of the tools directory - I expect we'll want to add quite a few more of these tools.

sshane:
@deanlee how do you select a bit? I just see bit 0 and 1 for a message.

sshane:
@deanlee I think there was some small misunderstanding, my script compared any bit you specified against all others, not just a static 1 or 0. Can we simply add current message bit selection now? That should complete this :)

PR #26835, opened 2022-12-23
cabana: rename memu items

deanlee:


adeebshihadeh:
I think the current names are more correct.

PR #26837, opened 2022-12-25
cabana: align y axis correctly

deanlee:
align plots by finding the longest y-axis label. remove hard coded `aligned_pos`,


nelsonjchen:
@mlouielu were you looking for something like this?

mlouielu:
It is, this works perfectly!

PR #26850, opened 2022-12-31
Laikad: set active

gast04:
use laikad as main gps source


.github/workflows/tools_tests.yaml
```diff
@@ -38,7 +38,7 @@ jobs:
     - name: Build Docker image
       run: eval "$BUILD"
     - name: Unit test
-      timeout-minutes: 2
+      timeout-minutes: 6
```
adeebshihadeh, line 41: why was this changed?

adeebshihadeh: I reverted it in https://github.com/commaai/openpilot/commit/50f62a7adc011c753aa846807a8eabfdbdf93996. please be careful updating these and make sure it's absolutely necessary

PR #26855, opened 2023-01-01
add cpplint

adeebshihadeh:


adeebshihadeh:
I want to check a bunch more stuff, but let's get this merged for now. `build/include_what_you_use` in particular would be nice

adeebshihadeh:
@deanlee feel free to play around and add options if you think we should have them

PR #26872, opened 2023-01-03
Added new fingerprint for Toyota Corolla Hatchback 6MT

bscholer:
AFAIK this engine fingerprint is unique to the 6MT version of the COROLLA_TSS2.

It is worth noting that adding the fingerprint is only one of the steps to get 6MT working. The other is a change to [`carstate.py`](https://github.com/commaai/openpilot/wiki/Toyota-Lexus#manual-transmission-6mt), but this would mess with every other car.

### DIFF
--- a/selfdrive/car/toyota/values.py
+++ b/selfdrive/car/toyota/values.py
@@ -122,6 +122,7 @@ CAR_INFO: Dict[str, Union[ToyotaCarInfo, List[ToyotaCarInfo]]] = {
     ToyotaCarInfo("Toyota Corolla 2020-22", video_link="https://www.youtube.com/watch?v=_66pXk0CBYA"),
     ToyotaCarInfo("Toyota Corolla Cross (Non-US only) 2020-23", min_enable_speed=7.5),
     ToyotaCarInfo("Toyota Corolla Hatchback 2019-22", video_link="https://www.youtube.com/watch?v=_66pXk0CBYA"),
+    ToyotaCarInfo("Toyota Corolla Hatchback 6MT 2019-22", video_link="https://youtu.be/ITxknEpp9R8"),
   ],
   CAR.COROLLAH_TSS2: [
     ToyotaCarInfo("Toyota Corolla Hybrid 2020-22"),
@@ -750,6 +751,7 @@ FW_VERSIONS = {
       b'\x018966312S7000\x00\x00\x00\x00',
       b'\x018966312W3000\x00\x00\x00\x00',
       b'\x018966312W9000\x00\x00\x00\x00',
+      b'\x018966312X0000\x00\x00\x00\x00',
       b'\x01896637644000\x00\x00\x00\x00',
     ],
     (Ecu.engine, 0x7e0, None): [
--- a/selfdrive/car/toyota/values.py
+++ b/selfdrive/car/toyota/values.py
@@ -122,7 +122,6 @@ CAR_INFO: Dict[str, Union[ToyotaCarInfo, List[ToyotaCarInfo]]] = {
     ToyotaCarInfo("Toyota Corolla 2020-22", video_link="https://www.youtube.com/watch?v=_66pXk0CBYA"),
     ToyotaCarInfo("Toyota Corolla Cross (Non-US only) 2020-23", min_enable_speed=7.5),
     ToyotaCarInfo("Toyota Corolla Hatchback 2019-22", video_link="https://www.youtube.com/watch?v=_66pXk0CBYA"),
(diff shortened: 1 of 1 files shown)


bscholer:
not really sure why `selfdirve / unit tests` fails, I didn't make any changes to the file it's complaining about.

jyoung8607:
You added what amounts to a new supported car but didn't regenerate CARS.md to include your new ToyotaCarInfo entry. That's done with `selfdrive/car/docs.py`.

That said:

1) I don't think it's appropriate to add another entry just to mention manual transmission support, so it's better to just drop that. Other openpilot brand ports (for example, VW family) support manual transmissions in any supported car without duplicating the supported car doc entries.

2) Upstreamed fingerprints should describe working cars. It sounds like this car wouldn't actually work without additional openpilot changes, so it's not appropriate to merge yet. New MT owners might plug in and experience confusing bugs and errors instead of the more-correct unrecognized car message.

If you're not sure how to make the `carstate` change such that it's conditional on auto-detecting a manual transmission, share what you have and we may be able to help.

vanillagorillaa:
Please provide the dongleID for the device this was gathered from

bscholer:
Fair points on both @jyoung8607. The two approaches I could see maybe working are:
1. Checking if the `CLUTCH_PRESSED` message is present, but I'm fairly certain the message will *be there* for all Toyotas, just never change. Maybe I'd need to wait for the clutch to be pressed and then change the `LOW_SPEED_LOCKOUT` value to `False` if that's the case. I'm not sure if this can be done after the fact though.
  I saw that VW figures out the trans type first, and then handles the clutch, however, I have not been able to find anything indicating the trans type or gear. The only thing that changes in `GEAR_PACKET` is `GEAR`, but only for reverse.
1. Checking for this engine fingerprint, and setting it that way, but hard-coding a fingerprint feels really jank and like something that wouldn't get upstreamed.

This is a route that shows the fingerprint: [`0a7a1e3d3dc09b9d|2022-06-27--20-46-01--0`](https://cabana.comma.ai/?route=0a7a1e3d3dc09b9d%7C2022-06-27--20-46-01&exp=1704348156&sig=SFvlPvk4%2FpO86qYNNkpR4OLph5kM%2B90PD2VYn5ETwhE%3D&max=2&url=https%3A%2F%2Fchffrprivate.azureedge.net%2Fchffrprivate3%2Fv2%2F0a7a1e3d3dc09b9d%2Fce4124d7187e683396394e8313c6a78c_2022-06-27--20-46-01)

bscholer:
@jyoung8607 here's what I've got so far, which should add `clutchPressed`. the issue is that every Toyota will receive `clutchPressed`. I'm not seeing a way to do this check (below) within `carstate::update()`. the best I can figure is that it would need a car port of its own.

if you have ideas, I'm all ears.
https://github.com/commaai/openpilot/blob/0155ade9976674ac946417e98b3b477f32381639/selfdrive/car/toyota/interface.py#L134-L137

sshane:
Do you still need help?

bscholer:
> Do you still need help?

Yeah, I'm not really sure what the best way forward is. Any ideas?

sshane:
Hmm yeah our low speed lockout check doesn't nicely play with your car. If the acc type is 1, then we assume that the PCM will allow braking to a stop, and if acc type is not 1, then we are okay with the low speed lockout, as the car is telling us it's not supported. However your car is reporting an acc type of 1, so with our definition, we expect the car to brake to zero, but the PCM is telling us it's not able to.

Is there a way to detect manual? We can add that to the check to solve this

bscholer:
> Is there a way to detect the transmission in a way that's clean and generic?

I've gone back and forth with @jyoung8607 on this some, and I don't think there's a good way of doing it really. Toyota doesn't really seem to send anything (besides the [clutch signal](https://github.com/commaai/opendbc/pull/606) and FW version) over the CAN that could differentiate in 6MT vs auto. 

The thing about the clutch is, in the case of turning the car on without starting, the clutch may never be pressed throughout the entire duration of the route.

That gets me thinking though, @sshane would it be feasible to watch for a clutch press, and if it ever happens, save a param (I believe that's the term for saved settings) that makes the vehicle always a manual? Not sure if that would even work, but its a shot in the dark.

sshane:
@bscholer can you drive on master? I added a query to get the automatic transmission, you can provide a route and I can tell if it exists on your car or not.

adeebshihadeh:
Closing due to inactivity, feel free to open a new one when you're ready to pick this back up.

PR #26873, opened 2023-01-04
Navigation: add destination marker

gast04:
marker_sample

current marker is the comma logo, leave the final one to the design team
(ignore the wrong blue route, it was the only sample I found with nav)


selfdrive/ui/qt/maps/map.h
```diff
@@ -111,6 +112,7 @@ class MapWindow : public QOpenGLWidget {
   MapETA* map_eta;
 
   void clearRoute();
+  void update_destination_marker();
```
adeebshihadeh, line 115: use camel case to match the rest of the class

adeebshihadeh (approved):

PR #26879, opened 2023-01-05
longcontrol: enter stopping state immediately

sshane:
Possibly fixes a bug with GMs where we command accel for one frame on engaging at a stop when openpilot wants to go.

Screenshot from 2023-01-04 21-27-37

### DIFF
--- a/selfdrive/controls/lib/longcontrol.py
+++ b/selfdrive/controls/lib/longcontrol.py
@@ -32,6 +32,8 @@ def long_control_state_trans(CP, active, long_control_state, v_ego, v_target,
   else:
     if long_control_state == LongCtrlState.off:
       long_control_state = LongCtrlState.pid
+      if stopping_condition:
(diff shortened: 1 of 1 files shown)


selfdrive/controls/lib/longcontrol.py
```diff
@@ -32,6 +32,8 @@ def long_control_state_trans(CP, active, long_control_state, v_ego, v_target,
   else:
     if long_control_state == LongCtrlState.off:
       long_control_state = LongCtrlState.pid
+      if stopping_condition:
+        long_control_state = LongCtrlState.stopping
 
     elif long_control_state == LongCtrlState.pid:
```
sshane, line 38:
```suggestion
    if long_control_state in (LongCtrlState.off, LongCtrlState.pid):
      long_control_state = LongCtrlState.pid
```

sshane:
@HaraldSchafer any reason not to go into stopping state one frame early?

PR #26880, opened 2023-01-05
Hyundai: add FW versions for Sonata 2023

BBBmau:
**Route**
stock openpilot Route:  10b8dcf2fb80bf56|2023-01-03--13-15-34--5
Experimental Route: 10b8dcf2fb80bf56|2023-01-04--14-14-33--22


BBBmau:
@sshane how exactly were you able to get the values for `values.py` and engine? Was it within the logs from the route that i provided?

sshane:
Yep, I checked the CarParams of your first segment and pulled the Hyundai FW versions that were missing

PR #528, opened 2019-02-09
Add 18 Buick Regal TourX Support

HOYS:
Adds support for 2018 Buick Regal TourX w/ ACC package.

[Drive Link](https://community.comma.ai/cabana/?route=aa20e335f61ba898%7C2019-01-28--11-12-28&max=47&url=https%3A%2F%2Fchffrprivate.blob.core.windows.net%2Fchffrprivate3%2Fv2%2Faa20e335f61ba898%2Fccfad38ceac8e8f634022b1a6effb418_2019-01-28--11-12-28)

PR #26885, opened 2023-01-05
GM: check radarOffCan in radar_interface

nworb-cire:
GM currently uses a hardcoded list of vehicle models to determine whether to use radar. This change brings the GM radar interface in line with other makes in using `CP.radarOffCan` to determine whether to create a radar CAN parser. (Compare Honda's `radar_interface.py` implementation.)

### DIFF
--- a/selfdrive/car/gm/radar_interface.py
+++ b/selfdrive/car/gm/radar_interface.py
@@ -3,7 +3,7 @@ import math
 from cereal import car
 from common.conversions import Conversions as CV
 from opendbc.can.parser import CANParser
-from selfdrive.car.gm.values import DBC, CAR, CanBus
+from selfdrive.car.gm.values import DBC, CanBus
 from selfdrive.car.interfaces import RadarInterfaceBase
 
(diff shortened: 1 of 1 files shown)


selfdrive/car/gm/radar_interface.py
```diff
@@ -16,9 +16,6 @@
 
 
 def create_radar_can_parser(car_fingerprint):
-  if car_fingerprint not in (CAR.VOLT, CAR.MALIBU, CAR.HOLDEN_ASTRA, CAR.ACADIA, CAR.CADILLAC_ATS, CAR.ESCALADE_ESV):
```
sshane, line 19: The Buick Regal is not in this list, do we know if that was a mistake? If so, this is a good change. I wouldn't imagine it wouldn't have a radar, but we should check why it was left out.

sshane: Think it was mistakenly forgotten to be added when we did the internal merge: https://github.com/commaai/openpilot/pull/528/files

PR #26889, opened 2023-01-06
Multilang: add French translations

croumegous:
Added French language support


sshane:
thanks, but we added French here: https://github.com/commaai/openpilot/pull/29260

PR #26902, opened 2023-01-08
Allow Honda Nidec vehicles with comma pedal to drive at max speed.

gadjex:
**Description**
Without this fix Honda Nidec vehicles with the comma pedal will drive around 2 mph under the max speed. Current code does not take into account if your are using a comma pedal with a Nidec vehicle and doesn't allow acceleration near cruise_speed. This fix removes those limits if you are using a comma pedal.

**Verification**
Drove around for weeks with this fix. Temporarily reverted this fix to check if this was indeed the culprit.

**Route**
Route: f1aabec98068b718|2023-01-08--15-28-36--2


VirtuallyChris:
This has bugged me to no end. So excited!

sshane:
@VirtuallyChris let me know with a route if you can test it!

VirtuallyChris:
@sshane sent on Discord!

PR #26921, opened 2023-01-12
Hyundai CAN-FD: set available from TCS ACC available

sshane:
`ACCEnable == 0` means that ACC is available and there are no faults. Since CAN FD cars always enable on the main button press (if main off), it makes a little more sense to check if there are any TCS faults preventing engagement, rather than SCC available (which would only be true after the first engage)


sshane:
@adeebshihadeh doing CAN FD here

sshane:
Behavior changes:

- Enabled:
  - OP long:
    - No change (available if no faults)
  - Stock long:
    - SCC available -> TCS available (true on startup)

adeebshihadeh:
Yes, can you explain this more completely in the comment?

PR #26926, opened 2023-01-12
Tall vehicles: calibrate vlead from model error

morrislee:
**Description**

When using vision only longitudinal control from OpenPilot (more prominent on taller vehicles), it estimates the world objects moving too slow overall including itself [#26799](https://github.com/commaai/openpilot/discussions/26799). This is verified from speed reported from the model is off from wheel speed, we used this info to calibrate e2e before, now we can use this concept to calibrate vlead and lead acceleration too. 

**Verification**

I have been driving a few days with another proof of concept branch and has been working really well on my GMC Sierra 1500 AT4 (2 inch taller than regular 1500's)

**Route**
Route: https://connect.comma.ai/fe80a4e1cedec853/1673461336764/1673461574279

As shown in the route, it maintained the distance with little to no overshooting or undershooting, also experience a cut-in that was handled perfectly with OP VOACC, I have driven a few hundred kms with this implemented so far

### DIFF
--- a/selfdrive/controls/lib/radar_helpers.py
+++ b/selfdrive/controls/lib/radar_helpers.py
@@ -132,12 +132,10 @@ class Cluster():
 
   def get_RadarState_from_vision(self, lead_msg, v_ego, vision_v_ego):
     # Learn vision model velocity error to correct vlead
+    corrected_v_lead = lead_msg.v[0]
     if v_ego > 0 and vision_v_ego > 0:
       vision_velocity_factor = v_ego / vision_v_ego
       corrected_v_lead = lead_msg.v[0] * vision_velocity_factor
-      #if no v_ego or vision_v_ego set, fallback to uncalibrated values
-    else: 
-      corrected_v_lead = lead_msg.v[0]
     return {
       "dRel": float(lead_msg.x[0] - RADAR_TO_CAMERA),
       "yRel": float(-lead_msg.y[0]),
--- a/selfdrive/controls/lib/radar_helpers.py
+++ b/selfdrive/controls/lib/radar_helpers.py
@@ -130,14 +130,21 @@ class Cluster():
       "aLeadTau": float(self.aLeadTau)
     }
 
-  def get_RadarState_from_vision(self, lead_msg, v_ego):
+  def get_RadarState_from_vision(self, lead_msg, v_ego, vision_v_ego):
+    # Learn vision model velocity error to correct vlead
+    if v_ego > 0 and vision_v_ego > 0:
+      vision_velocity_factor = v_ego / vision_v_ego
+      corrected_v_lead = lead_msg.v[0] * vision_velocity_factor
+      #if no v_ego or vision_v_ego set, fallback to uncalibrated values
+    else: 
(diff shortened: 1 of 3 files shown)


HaraldSchafer:
I think it's best not to correct lead_a, can we just do v?

selfdrive/controls/lib/radar_helpers.py
```diff
@@ -130,14 +130,21 @@ def get_RadarState(self, model_prob=0.0):
...
 
-  def get_RadarState_from_vision(self, lead_msg, v_ego):
+  def get_RadarState_from_vision(self, lead_msg, v_ego, vision_v_ego):
+    # Learn vision model velocity error to correct vlead
+    if v_ego > 0 and vision_v_ego > 0:
+      vision_velocity_factor = v_ego / vision_v_ego
+      corrected_v_lead = lead_msg.v[0] * vision_velocity_factor
+      #if no v_ego or vision_v_ego set, fallback to uncalibrated values
+    else: 
+      corrected_v_lead = lead_msg.v[0]
```
sshane, line 140: I would set v_lead above before the if statement, then override if possible

selfdrive/controls/lib/radar_helpers.py
```diff
@@ -130,14 +130,19 @@ def get_RadarState(self, model_prob=0.0):
       "aLeadTau": float(self.aLeadTau)
     }
 
-  def get_RadarState_from_vision(self, lead_msg, v_ego):
+  def get_RadarState_from_vision(self, lead_msg, v_ego, vision_v_ego):
+    # Learn vision model velocity error to correct vlead
+    corrected_v_lead = lead_msg.v[0]
+    if v_ego > 0 and vision_v_ego > 0:
+      vision_velocity_factor = v_ego / vision_v_ego
+      corrected_v_lead = lead_msg.v[0] * vision_velocity_factor
```
sshane, line 138: Might make more sense to copy what longitudinal_planner does, which is simply get an error to subtract

selfdrive/controls/lib/radar_helpers.py
morrislee: How does it look now? I understand that this is cleaner

selfdrive/controls/lib/radar_helpers.py
morrislee: should be a scale rather than a subtract, if there is an error, then it might apply to a static lead no?

Verylukyguy:
My 2018 Acadia is considered a Mid-Sized SUV and the camera height is 53" or 1.35 meters.  This is only slightly higher that the standard 1.22 meters.  I've heard that increasing the height might mitigate some of the lane hugging.  Is this true?

Verylukyguy:
Interesting.  Skip did a lot of trial-and-error testing when the Silverado was first ported.  His camera height was 1.62 meters and it seemed to cause wandering and hugging.

morrislee:
been driving with harold's PR #27493 seems to be very similar in performance, will close this PR as harold's approach is a bit cleaner and safer

PR #26928, opened 2023-01-12
cabana: fix chart glitches

deanlee:
fixed https://github.com/commaai/openpilot/discussions/26091#discussioncomment-4661712:

> The graphs glitch out if you zoom in and out a few times. They end up becoming blank after a while:


sshane:
@deanlee now it's very difficult to select sections on the timeline, it resets halfway through selection, similar to web Cabana

deanlee:
> @deanlee now it's very difficult to select sections on the timeline, it often does nothing when clicking and dragging, similar to It is not easy to align qt charts on y-axis. I need some time to think about how to better solve these issues.

PR #26936, opened 2023-01-13
Add missing COROLLA_TSS2 engine f/w

ErichMoraga:
`@wyu#6910`  2020 Corolla (ICE)  DongleID/route 98d3152d30060750|2023-01-12--18-41-03


sshane:
None of the ECUs responded correctly, looking into it. You added a response from an we don't log for Toyota (Tesla)

sshane:
The user had a use local diff, all his other routes look reasonable

PR #26938, opened 2023-01-13
cabana: use Qt::Window flag for FindSimilarBitsDlg

deanlee:
resolve https://github.com/commaai/openpilot/discussions/26091#discussioncomment-4673849


sshane:
this tool is getting seriously good!

PR #26946, opened 2023-01-13
Cabana: support live streaming

deanlee:
support live streaming from a panda or c3. 

cabana will only cache `settings.cached_segment_limit * 60` seconds data in the memory to prevent memory exhaustion in living mode. you can change this setting in the option dialog : `menu->files->settings->Cached segment limit`

**usage:**
receive msgq messages:
`cabana --stream`
receive zmq messages:
`cabana --stream --zmq="127.0.0.1"`

**Currently, there are some restrictions in the live streaming mode.**
1. no video
2. can't disable dynamic mode in historylog (the "dynamic mode" checkbox is disabled)
3. can't zoom-in/zoom-out charts
4. can't pause, or seekto a positon

I'll try to remove some of these restrictions in the future.

[Kazam_screencast_00048.webm](https://user-images.githubusercontent.com/27770/212420081-4634e216-91f2-41f6-9585-0ceb0ea6cf20.webm)


sshane:
This will be amazing for new car ports or general debugging, thanks deanlee!!

deanlee:
Conflict with https://github.com/commaai/openpilot/pull/26970. temporarily convert to draft to wait  for  https://github.com/commaai/openpilot/pull/26970 to be merged first.

pd0wm:
Looks like fading of the colors broke in this PR. In https://github.com/commaai/openpilot/pull/26970 the red/cyan slowly fade out.

pd0wm:
That looks correct!

pd0wm:
Probably best to keep these extra color changes separate from the streaming PR. There is no need to abstract it into a class to be able to merge this.

deanlee:
ok,I'll remove these changes

deanlee:
@adeebshihadeh this pr is ready. many later works depends on this. It's getting harder to fix it's conflicts with the master.

adeebshihadeh:
@sshane can you test a bit and merge?

pd0wm:
I also have code for an `AbstractStream` implementation that connects to a panda directly. Can submit a PR if you're interested (and this PR is merged).

deanlee:
that's great!!

adeebshihadeh:
> I also have code for an `AbstractStream` implementation that connects to a panda directly. Can submit a PR if you're interested (and this PR is merged).

Any reason not to use boardd? I think sendcan/can are already a nice abstraction if people want to use other CAN tools.

pd0wm:
I'm trying to get cabana to run on Windows as well, not requiring any messaging makes the setup much simpler 😅. It uses the same `Panda` class as boardd, so it's only a few lines different from @deanlee's livestream.

No worries if you're not interested in maintaining it!

adeebshihadeh:
> I'm trying to get cabana to run on Windows as well, not requiring any messaging makes the setup much simpler sweat_smile. It uses the same `Panda` class as boardd, so it's only a few lines different from @deanlee's livestream.

Should be fine if it's a small change. Maybe <20 lines?

sshane:
Seems to be working well on my desk with a C3!

sshane:
Thanks @deanlee! I'll do some more testing in a car with a real use and lyk if I see any issues. Pausing the way PlotJuggler handles it with live streaming would be nice 😄

PR #26969, opened 2023-01-15
cabana: add icons

deanlee:
current unicode characters does not display properly on some platforms.
icons is from https://github.com/tabler/tabler-icons


adeebshihadeh:
Adding the [Bootstrap icons](https://icons.getbootstrap.com/) in #26980

PR #26973, opened 2023-01-15
cabana: MacOS fixes

pd0wm:
- `OpenpilotPrefix` is msgq specific, which isn't used on MacOS
- OpenGL needs to be disabled on the graphs or it will draw the video. There might be a more proper fix for this. https://github.com/commaai/openpilot/discussions/26091#discussioncomment-4634932
- `Qt5Charts` needs to be added as a framework instead of library

Still draft until I have a chance to test on Linux

### DIFF
--- a/tools/cabana/cabana.cc
+++ b/tools/cabana/cabana.cc
@@ -33,9 +33,11 @@ int main(int argc, char *argv[]) {
     replay_flags |= REPLAY_FLAG_QCAMERA;
   }
 
+  // TODO: Remove when OpenpilotPrefix supports ZMQ
 #ifndef __APPLE__
   OpenpilotPrefix op_prefix;
 #endif
+
   CANMessages p(&app);
   int ret = 0;
   if (p.loadRoute(route, cmd_parser.value("data_dir"), replay_flags)) {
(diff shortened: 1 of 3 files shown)


tools/cabana/cabana.cc
```diff
@@ -33,7 +33,9 @@ int main(int argc, char *argv[]) {
     replay_flags |= REPLAY_FLAG_QCAMERA;
   }
 
+#ifndef __APPLE__
   OpenpilotPrefix op_prefix;
+#endif
```
adeebshihadeh, line 38: we should just make the prefix work with ZMQ. @deanlee want to try to implement that?

tools/cabana/chartswidget.cc
```diff
@@ -267,7 +267,10 @@ void ChartView::setPlotAreaLeftPosition(int pos) {
 
 void ChartView::addSeries(const QString &msg_id, const Signal *sig) {
   QLineSeries *series = new QLineSeries(this);
+
+#ifndef __APPLE__
```
adeebshihadeh, line 271: can you put a comment or TODO here?

deanlee:
> OpenGL needs to be disabled on the graphs or it will draw the video. There might be a more proper fix for this. new cabana feedback #26091 (comment)

I suspect it's a bug in CameraWidget

tools/cabana/cabana.cc
deanlee: I'll try it later, let's keep the #ifndef for now.

pd0wm:
Added comments and tested on Linux

PR #26977, opened 2023-01-16
Honda Nidec: HUD set cruise speed doesn't always match UI

gadjex:
Fix bug - Affects only Honda Nidec with comma pedal: HUD set cruise speed doesn't _always_ match UI max cruise speed when units is set to mph.

**Description**
Set cruise speed in KPH is rounded before sent to the HUD resulting in the HUD's conversion to MPH showing +-1 when the HUD is set to MPH. This happens at only specific set cruise speeds, such as 49 and 54mph are displayed on the HUD as 48 and 53 respectively. This only affects Honda Nidec vehicles with the comma pedal. Converting the speed to a MPH rounded integer and back to KPH before rounding the KPH and sending it over CAN ensures the HUD's conversion from KPH to MPH is correct. This fix takes into account if the units are set to metric or imperial.

**Verification**
I applied this fix and tested that the vehicle's HUD from 25mph to 70mph is matching comma's UI.


VirtuallyChris:
Will test this shortly. Has also bugged me to no end, thank you for doing these fixes @gadjex ! Up next, make the pedal actually accurate to the longitudinal plan. Currently it's terribly slow from a stop ;)

VirtuallyChris:
This doesn't work on my Nidec Honda. The set speed shown on the dash gets offset by 1 or 2 mph at all times, and is worse than before this PR. There still is a rounding issue on certain speeds like 84mph which you can reproduce in the comma Civic. It's probably a slightly off DBC scaling factor.

selfdrive/car/honda/carcontroller.py
```diff
@@ -1,18 +1,28 @@
...
+V_CRUISE_FACTOR = IMPERIAL_INCREMENT / CV.KPH_TO_MPH
 VisualAlert = car.CarControl.HUDControl.VisualAlert
 LongCtrlState = car.CarControl.Actuators.LongControlState
 
 
+def round_hud_set_speed(v_cruise_kph, is_metric):
+  if is_metric:
+    return round(v_cruise_kph)
+  cruise_mph = round(v_cruise_kph * CV.KPH_TO_MPH)
+  return math.floor(cruise_mph * CV.MPH_TO_KPH * V_CRUISE_FACTOR)
```
sshane, line 23: the factor fixes the very high end where 90 is rounded to 91

sshane:
Closing since it's reported to not work on a Civic

PR #26981, opened 2023-01-16
cabana: use bootstrap icons

deanlee:
`QSvgRenderer` can't parse the `symbol` tag in `bootstrap-icons.svg`, so I  wrote a custom parser.


selfdrive/ui/qt/util.cc
```diff
@@ -218,3 +221,37 @@ QColor interpColor(float xv, std::vector<float> xp, std::vector<QColor> fp) {
     );
   }
 }
+
+static QHash<QString, QByteArray> load_bootstrap_icons() {
+  QHash<QString, QByteArray> icons;
+
+  QFile f("../../third_party/bootstrap/bootstrap-icons.svg");
```
pd0wm, line 228: Maybe use qrc to include the file in the binary so it can be run from anywhere?

deanlee: done in [6627abe](https://github.com/commaai/openpilot/pull/26981/commits/6627abe9e5c44f9155481ec2f1c1d368399a921a)

PR #26987, opened 2023-01-17
Add low Gnss laikad test

gast04:


gast04:
merge after https://github.com/commaai/laika/tree/fix_trop_delay

selfdrive/locationd/laikad.py
```diff
@@ -391,13 +391,16 @@ def process_msg(laikad, gnss_msg, mono_time, block=False):
 
 
 def clear_tmp_cache():
+  if "CI" in os.environ or "REPLAY" in os.environ:
+    return
```
adeebshihadeh, line 395: why does CI rely on this?

gast04: replay uses CI env flag, and runs it in parallel, thats doesnt work

adeebshihadeh: hmm, can you add a comment to remove this once we're not writing this cache to disk? CI exceptions should be used extremely sparingly and only when absolutely necessary

PR #26996, opened 2023-01-18
cabana: sort bus:id numerically instead of alphabetically

pd0wm:
Before `0:4a` would be sorted after `0:3aa`

### DIFF
--- a/tools/cabana/canmessages.h
+++ b/tools/cabana/canmessages.h
@@ -11,6 +11,8 @@
 
 struct CanData {
   double ts = 0.;
+  uint32_t src = 0;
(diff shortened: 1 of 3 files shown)


tools/cabana/canmessages.h
```diff
@@ -11,6 +11,8 @@
 
 struct CanData {
   double ts = 0.;
+  uint32_t src = 0;
```
deanlee, line 14: cabana use uint8_t src in all other places.

pd0wm, line 14:
```suggestion
  uint8_t src = 0;
```

PR #26997, opened 2023-01-18
cabana: refactor and improve signal view

deanlee:
refactor to use standard TreeView/Model architecture, with many new features and bug fixes:

Screenshot from 2023-01-18 23-05-02

dark theme:

light theme:


pd0wm:
Like the tree structure a lot! Will make it much easier to add multiplexed signals (https://pisnoop.s3.amazonaws.com/snoop_help_dbc.htm, https://github.com/commaai/opendbc/blob/master/vw_golf_mk4.dbc#L1412)

sshane:
Can we make the fields' height similar to before? I feel like it's a little hard to click around and edit fields now

deanlee:
It uses the system default height. how about keep it as it for now? we can change it later if really need.

sshane:
Since we can use the keyboard to navigate the tree view and expand stuff, it would be nice if Enter allowed you to edit stuff/toggle check boxes! Doesn't need to be in this PR

sshane:
Since we can use the keyboard to navigate the tree view and expand stuff, it would be nice if Enter allowed you to edit stuff/toggle check boxes! Doesn't need to be in this PR. Is this PR good to go?

deanlee:
qt use F2 to trigger the editor: https://doc.qt.io/qt-5/qtreeview.html#key-bindings
we can improve the key bindings in another pr.I think this pr is ok to be merged

pd0wm:
Just a suggestion, and might be better to do in a different PR. Is it possible to show the latest signal value next to the signal name? Switching back and forth between the logs can be a bit annoying.

deanlee:
it should be easy to show signal value in column 1 by emit mode->dataChange (signal name is col 0). 
maybe add a helper function in util.h to reuse it in log model and signal model to read the signal values.
I think about how to make it cleaner

deanlee:
@pd0wm  added in [de81678](https://github.com/commaai/openpilot/pull/26997/commits/de816784276348537410f6b43a7acf5fe91dd67b)

[Kazam_screencast_00066.webm](https://user-images.githubusercontent.com/27770/215151810-878fb4c0-e6a2-4ab5-8171-1a2aad70e528.webm)

deanlee:
added a splitter between signal view & binary view:

[Kazam_screencast_00067.webm](https://user-images.githubusercontent.com/27770/215155578-b4f0e315-1446-44f4-b7cd-cc4a69037a84.webm)

deanlee:
> Since we can use the keyboard to navigate the tree view and expand stuff, it would be nice if Enter allowed you to edit stuff/toggle check boxes! Doesn't need to be in this PR. Is this PR good to go?

@sshane I think this PR is ok for merge after [de81678](https://github.com/commaai/openpilot/pull/26997/commits/de816784276348537410f6b43a7acf5fe91dd67b). the new added code is worth it. although it may introduce some new issues, it should be easy to fix.

pd0wm:
Looks so much better :)

Would be nice if the position of the divider between the bytes and signal list stayed in place when opening new signals. With CAN-FD you end up constantly making the bytes view larger.

PR #27004, opened 2023-01-19
Remove old gpsLocation usage

gast04:


adeebshihadeh:
Let's close until laikad has been out of shadow mode for a bit.

PR #27005, opened 2023-01-19
cabana: add a slider to adjust chart range from 1 sec to max cached seconds.

deanlee:
add slider to set the chart range in toolbar (drag slider or  press key: left/right arrow, page up/page down), removed `Chart range (minutes)` from setting dialog.

[Kazam_screencast_00058.webm](https://user-images.githubusercontent.com/27770/213380766-10231006-397b-4f18-9dba-59b680a00e4f.webm)


pd0wm:
If this works too during live mode a lower bound of 1 second or 10 seconds would be nice. 1 minute is way too long to identify fast moving things like counters.

deanlee:
it's works in live mode too.I'll change the step value and the minimum value.

deanlee:
the minimum range is set to 1 sec.
the single step is set to 1s ( left/right arrow key)
the page step is set to 1m (page up / page down)

[Kazam_screencast_00059.webm](https://user-images.githubusercontent.com/27770/213457791-f8d18686-f80b-4ffb-9740-b53be282fa5c.webm)

deanlee:
lgtm, work in both live and replay mode.

deanlee:
fixed a few bugs in previous code

PR #27007, opened 2023-01-19
cabana: improve open/save functions

pd0wm:

### DIFF
--- a/tools/cabana/mainwin.cc
+++ b/tools/cabana/mainwin.cc
@@ -68,13 +68,14 @@ MainWindow::MainWindow() : QMainWindow() {
 
 void MainWindow::createActions() {
   QMenu *file_menu = menuBar()->addMenu(tr("&File"));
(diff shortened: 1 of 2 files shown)


tools/cabana/mainwin.cc
```diff
@@ -213,18 +214,24 @@ void MainWindow::loadDBCFromFingerprint() {
...
-                                                   QDir::cleanPath(settings.last_dir + "/untitled.dbc"), tr("DBC (*.dbc)"));
-  if (!file_name.isEmpty()) {
+  if (file_name.isEmpty()) {
+    saveAsDBCToFile();
+  } else {
     settings.last_dir = QFileInfo(file_name).absolutePath();
     QFile file(file_name);
     if (file.open(QIODevice::WriteOnly)) {
       file.write(dbc()->generateDBC().toUtf8());
-      detail_widget->undo_stack->clear();
```
deanlee, line 223: any reason to remove this line? I think we should clear the undo history after save.

PR #27009, opened 2023-01-19
UI: update translations during scons build

adeebshihadeh:


sshane:
Diff is because we don't include a lot of the source files in the release version. This is why I had to exclude the translation test from pre-commit in release CI

adeebshihadeh:
> Diff is because we don't include a lot of the source files in the release version. This is why I had to exclude the translation test from pre-commit in release CI

Can you fix that? The environments shouldn't be different; we can either add them to the release files or only run the translations for the main UI. I'll wait on this PR until that's in.

sshane:
Oh for some reason I thought we didn't include any UI source files. Yep, I can make sure we include all relevant

release/check-dirty.sh
```diff
@@ -7,5 +7,6 @@ cd $DIR
 if [ ! -z "$(git status --porcelain)" ]; then
   echo "Dirty working tree after build:"
   git status --porcelain
+  git diff
```
sshane, line 10:
```suggestion
```

PR #27029, opened 2023-01-21
Toyota: reliable fuzzy fingerprinting

sshane:
After taking a look at the Toyota FW versions, it looks like there's some sort of pattern here, so there may be potential for doing some heuristics on just the existing FW versions alone.

```python
* COMMON FW PREFIXES - ALL CARS
  * ('abs', '0x7b0'):        F1526 (220 FW)
  * ('dsu', '0x791'):        88 (81 FW)
  * ('eps', '0x7a1'):        8965B (147 FW)
  * ('engine', '0x7e0'):     3 (78 FW)
  * ('fwdRadar', '0x750'):   8821F (132 FW)
  * ('fwdCamera', '0x750'):  8646F (168 FW)
  * ('engine', '0x700'):     89663 (364 FW)
  * ('abs', '0x7d1'):        898A36203000 (ONLY ONE FW)
```


sshane:
Superseded by https://github.com/commaai/openpilot/pull/28641

PR #27030, opened 2023-01-21
cabana: fix flicker during align charts

deanlee:


tools/cabana/chartswidget.cc
```diff
@@ -260,9 +260,11 @@ void ChartsWidget::alignCharts() {
   for (auto c : charts) {
     plot_left = qMax((qreal)plot_left, c->getYAsixLabelWidth());
   }
+  setUpdateEnabled(false);
```
sshane, line 263:
```suggestion
  setUpdatesEnabled(false);
```
?

deanlee:
closed,I have a better solution, to improve the chart alignment and remove almost all flickers and glitches

PR #27033, opened 2023-01-21
HKG: Update Genesis GV60 harnesses

sshane:
User reports advanced needs Hyundai A (from https://github.com/commaai/openpilot/pull/27011). Compared to original car port PR dongle, neither have an ADAS ecu


selfdrive/car/hyundai/values.py
```diff
@@ -225,7 +225,10 @@ def init_make(self, CP: car.CarParams):
   ],
 
   # Genesis
-  CAR.GENESIS_GV60_EV_1ST_GEN: HyundaiCarInfo("Genesis GV60 2023", "All", harness=Harness.hyundai_k),
+  CAR.GENESIS_GV60_EV_1ST_GEN: [
+    HyundaiCarInfo("Genesis GV60 (Advanced Trim) 2023", "All", harness=Harness.hyundai_a),
+    HyundaiCarInfo("Genesis GV60 (Performance Trim) 2023", "All", harness=Harness.hyundai_k),
```
adeebshihadeh, line 230: difference isn't HDA?

sshane: See nothing about it in the brochure, and neither dongles have an ADAS ECU

PR #27040, opened 2023-01-22
Update longcontrol.py

apache2046:
The PID's v error input should consider CP.longitudinalActuatorDelay too, so use the v_target  here is more reasonable


sshane:
I think the reasoning we use current velocity for feedback is because we don't want to control on a setpoint that we expect there will always be error with. v_target_now should actually match vEgo, since our feedforward *is* lag compensated, so the P and I components should always be 0 in an ideal system.

@HaraldSchafer might have some more input

HaraldSchafer:
Yes, it is designed to work as @sshane  explained.

apache2046:
Thanks for the explaination

PR #27058, opened 2023-01-23
Update dump_pll.c

MiChaelinzo:
Here are the main changes I made:

- I added error handling for the `open()` and `mmap()` functions. If either of these functions fail, the code will print an error message and terminate the program. This will provide more useful information for debugging if the program fails to run correctly.

I changed the while(1) loop to include a return statement at the end of the program. This is to avoid infinite looping and to ensure that the program terminates correctly. 
I added a call to close(fd) at the end of the program to close the file descriptor for /dev/mem and release the resources associated with it. 
I added a return 0 statement at the end of the main function to indicate successful termination of the program. These changes improve the overall robustness and readability of the code.

### DIFF
--- a/scripts/dump_pll.c
+++ b/scripts/dump_pll.c
@@ -38,22 +38,51 @@ void hexdump(uint32_t *d, int l) {
 #define PSCTL_OFFSET 0x164
 
 int main() {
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Thanks, but this is a one-off debugging script.

PR #27063, opened 2023-01-24
Honda: reliable fuzzy fingerprinting

sshane:
this branch is stemmed from the first hkg branch lol, will clean up


sshane:
Honda is in pretty good shape currently

image

PR #27064, opened 2023-01-24
ui: tap experimental mode icon to toggle

### LINKED ISSUE #27002: Make experimental mode icon a toggle
Hit box should be a circle a bit bigger than the real icon; mis-taps are especially annoying because of the sidebar/nav/fullscreen cycling.

incognitojam:
After experimental mode has been enabled for the first time (confirmed), it can be toggled by tapping the engage-ability/experimental mode icon in the upper right.

Closes #27002

### DIFF
--- a/selfdrive/ui/qt/onroad.cc
+++ b/selfdrive/ui/qt/onroad.cc
@@ -175,13 +175,65 @@ void OnroadAlerts::paintEvent(QPaintEvent *event) {
 }
 
 
+ExperimentalButton::ExperimentalButton(QWidget *parent) : QPushButton(parent) {
+  setCheckable(true);
+  setFixedSize(144, 144);
+
+  params = Params();
+  engage_img = loadPixmap("../assets/img_chffr_wheel.png", size());
+  experimental_img = loadPixmap("../assets/img_experimental.svg", size());
+
+  QObject::connect(this, &QPushButton::toggled, [=](bool checked) {
+    params.putBool("ExperimentalMode", checked);
+  });
+}
+
+void ExperimentalButton::updateState(const UIState &s) {
+  const SubMaster &sm = *(s.sm);
+
+  // button is "checked" if experimental mode is enabled
+  setChecked(sm["controlsState"].getControlsState().getExperimentalMode());
+
+  // engageable status, shown when unchecked
+  const auto cs = sm["controlsState"].getControlsState();
+  setProperty("engageable", cs.getEngageable() || cs.getEnabled());
+
+  // disable button when experimental mode is not available, or has not been confirmed for the first time
+  const auto cp = sm["carParams"].getCarParams();
+  const bool experimental_mode_available = cp.getExperimentalLongitudinalAvailable() ? params.getBool("ExperimentalLongitudinalEnabled") : cp.getOpenpilotLongitudinalControl();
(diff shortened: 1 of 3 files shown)


selfdrive/ui/qt/onroad.cc
```diff
@@ -402,6 +402,21 @@ void AnnotatedCameraWidget::drawHud(QPainter &p) {
 // Window that shows camera view and variety of
 // info drawn on top
 
+void AnnotatedCameraWidget::mousePressEvent(QMouseEvent* e) {
+  const int exp_btn_w = radius + bdr_s * 4;
+  const int exp_btn_h = radius + bdr_s * 3;
+
+  if (engageable) {
+    auto params = Params();
```
adeebshihadeh, line 410: keep this around as a class variable

selfdrive/ui/qt/onroad.cc
```diff
@@ -402,6 +402,21 @@ void AnnotatedCameraWidget::drawHud(QPainter &p) {
...
+    if (params.getBool("ExperimentalModeConfirmed") && e->x() > rect().right() - exp_btn_w && e->y() < exp_btn_h) {
```
adeebshihadeh, line 411: not sure the rect is going to be intuitive when you mis-tap, think a circle would be better

selfdrive/controls/controlsd.py
```diff
@@ -144,6 +144,7 @@ def __init__(self, sm=None, pm=None, can_sock=None, CI=None):
       self.params.remove("ExperimentalLongitudinalEnabled")
     if not self.CP.openpilotLongitudinalControl:
       self.params.remove("ExperimentalMode")
+      self.params.remove("ExperimentalModeConfirmed")
```
adeebshihadeh, line 147: why was this added?

incognitojam: Because I'm using `ExperimentalModeConfirmed` to determine whether experimental mode is available (not checking OP long/experimental long car params) If you moved from OP long to stock long you would still be able to toggle experimental mode onroad, unless this is cleared

adeebshihadeh: it should do the same checks as here: https://github.com/commaai/openpilot/blob/690999a2a6138368a93e818cbcebc984c7e58b91/selfdrive/ui/qt/offroad/settings.cc#L149

adeebshihadeh:
This should also have pressed states, so probably best to just make this a real button.

selfdrive/ui/qt/onroad.cc
sshane: A circular tap area? I don't think you'd notice that much in the car.

incognitojam:
Re-written using QPushButton with checked state when experimental long is enabled. Setting the opacity of the images works well as a pressed state

incognitojam:
```
------------------------------------------------
------------ UI Draw Timing: master ------------
------------------------------------------------
min  1.89ms
max  19.87ms
std  1.47ms
mean 4.94ms
------------------------------------------------


------------------------------------------------
------------ UI Draw Timing: branch ------------
------------------------------------------------
min  2.04ms
max  16.12ms
std  1.09ms
mean 4.89ms
------------------------------------------------
```

PR #27067, opened 2023-01-24
tinygrad: update to new version

geohot:
First commit is just fixing the files and confirming test pass with OPENCL backend (don't merge). Then will switch to GPU backend, and if tests pass, let's merge it and I can delete the OPENCL backend from tinygrad!

GPU backend is 4% faster, contains no hand coded kernels, and makes 1 less copy of the input. It's also mainline tinygrad, so improvements will trickle down. This is faster even while using the same kernel tunings, overchecking some validity stuff, and overfetching weights in the depthwise convs.There's at least 20% more to gain if we ever need it.


selfdrive/modeld/SConscript
```diff
@@ -87,10 +87,12 @@ if use_thneed and arch == "larch64" or GetOption('pc_thneed'):
     "#tinygrad_repo/extra/utils.py",
     "#tinygrad_repo/tinygrad/llops/ops_gpu.py",
     "#tinygrad_repo/tinygrad/llops/ops_opencl.py",
+    "#tinygrad_repo/tinygrad/graph.py",
```
adeebshihadeh, line 90: can replace this list with scons's `env.Glob` or read this list from `release/files_common`

geohot: Ah yea I should probably just read it from release

geohot:
So model replay failed (as expected). Looks like small differences. We still have the stuff to confirm if the model is good? What would it take to get this in CI?

geohot:
If test passes, this should be good to merge.

adeebshihadeh:
> So model replay failed (as expected). Looks like small differences. We still have the stuff to confirm if the model is good? (hoping compare_runtime.py still works) What would it take to get this in CI?
> 
> `python release_tests/driving/compare_runtime.py -vm d8501d20-bb59-4193-aa82-82b2737dedd6/449 -pm 609d90f3-65e6-4617-a60c-d6d99eead408/700 -tc 823c54552a2f861a7ad`
> 
> Timings are 19.5 ms now down from 20.3! We have room to more than 2x the model size, and that's without more tuning.
> 
> 2023-01-24-134837_641x2037_scrot

We can get the compare_runtime report in CI, but does it have pass/fail thresholds?

geohot:
Hmm, so compare_runtime is using the pipeline. The only annoying step was having to find the model names (there's a typo in the last commit to supercombo.onnx). Otherwise not high priority.

PR #27070, opened 2023-01-24
ui: dmoji

ZwX1616:


selfdrive/ui/ui.h
```diff
@@ -25,6 +25,17 @@ typedef cereal::CarControl::HUDControl::AudibleAlert AudibleAlert;
 
 const mat3 DEFAULT_CALIBRATION = {{ 0.0, 1.0, 0.0, 0.0, 0.0, 1.0, 1.0, 0.0, 0.0 }};
 
+const int FACE_KPTS_SIZE = 31;
```
adeebshihadeh, line 28: this shouldn't be hardcoded

selfdrive/ui/ui.cc
```diff
@@ -226,6 +259,12 @@ UIState::UIState(QObject *parent) : QObject(parent) {
   timer = new QTimer(this);
   QObject::connect(timer, &QTimer::timeout, this, &UIState::update);
   timer->start(1000 / UI_FREQ);
+
+  // initial dm draw
```
adeebshihadeh, line 263: what's this for? don't we hide the icon until the DM packet is seen?

selfdrive/ui/qt/onroad.cc
```diff
@@ -273,9 +273,11 @@ void AnnotatedCameraWidget::updateState(const UIState &s) {
   setProperty("is_cruise_set", cruise_set);
   setProperty("is_metric", s.scene.is_metric);
   setProperty("speed", cur_speed);
+  setProperty("isStandstill", sm["carState"].getCarState().getStandstill());
```
adeebshihadeh, line 276: unused?

selfdrive/ui/qt/onroad.cc
```diff
@@ -273,9 +273,11 @@ void AnnotatedCameraWidget::updateState(const UIState &s) {
...
   setProperty("setSpeed", set_speed);
   setProperty("speedUnit", s.scene.is_metric ? tr("km/h") : tr("mph"));
   setProperty("hideDM", cs.getAlertSize() != cereal::ControlsState::AlertSize::NONE);
+  setProperty("dm_fade_state", fmax(0.0, fmin(1.0, dm_fade_state+0.2*(0.5-(float)(dmActive)))));
```
adeebshihadeh, line 280: why use a property here? seems unnecessary

selfdrive/ui/ui.cc
ZwX1616: no? we only hide dm on alerts

selfdrive/ui/qt/onroad.cc
ZwX1616: you're right. fixing

selfdrive/ui/qt/onroad.cc
ZwX1616: how else do you set it? scene is const?

selfdrive/ui/ui.h
ZwX1616: yeah actually this was for something else, guess can remove

selfdrive/ui/ui.cc
ZwX1616: made a dm_rendered flag instead. This way the icon never shows if dm model packet for whatever reason doesn't show up

PR #27073, opened 2023-01-25
Civic 2022: add camera versions

sshane:
From 5 dongles over the last month:

```
{b'36161-T20-A060\x00\x00': {'35c004509e3057f2', '6e12554fbd20d6d1'}, b'36161-T47-A070\x00\x00': {'259b1607c209f5e2'}, b'36161-T20-A080\x00\x00': {'be8440c6657a7ef8'}, b'36161-T20-A070\x00\x00': {'9fc499d4cc3a269b'}}
```


selfdrive/car/honda/values.py
```diff
@@ -1441,6 +1441,12 @@ def init_make(self, CP: car.CarParams):
       b'78108-T21-A230\x00\x00',
       b'78108-T22-A020\x00\x00',
     ],
+    (Ecu.fwdRadar, 0x18dab0f1, None): [
```
sshane, line 1444: This is actually the fwdCamera, since this car doesn't have a radar. The problem is this address is used for both radar and camera in different vehicles, and we have a test to assert consistent addrs to ecus

PR #27074, opened 2023-01-25
Honda Civic 2022: gather FW versions for unknown ECU

sshane:
This and the fwdCamera is the only ECU that responds to a query from the camera to the PT bus. When we do FW fingerprinting from the camera only, this ECU might be important

### DIFF
--- a/selfdrive/car/honda/values.py
+++ b/selfdrive/car/honda/values.py
@@ -164,6 +164,10 @@ FW_QUERY_CONFIG = FwQueryConfig(
       bus=0,
     ),
   ],
+  extra_ecus=[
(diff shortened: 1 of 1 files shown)


selfdrive/car/honda/values.py
```diff
@@ -164,6 +164,10 @@ def init_make(self, CP: car.CarParams):
       bus=0,
     ),
   ],
+  extra_ecus=[
+    # Only other ECU on PT bus accessible by camera on radarless Civic
```
sshane, line 168:
```suggestion
    # The only other ECU on PT bus accessible by camera on radarless Civic
```

PR #27082, opened 2023-01-25
Raise Ram DT torque limits to 350

realfast:
**Description** [ Raise the torque limit to 350 for Ram DT. This allows us to go from 2.0m/s2 of lateral accel to 2.5m/s2 of lateral accel.]

**Verification** [We have been running this for over a year at these values. Anything over 350 will cause an eps fault.]


selfdrive/car/torque_data/override.yaml
```diff
@@ -21,7 +21,7 @@ FORD FOCUS 4TH GEN: [.nan, 1.5, .nan]
 COMMA BODY: [.nan, 1000, .nan]
 
 # Totally new cars
-RAM 1500 5TH GEN: [2.0, 2.0, 0.0]
+RAM 1500 5TH GEN: [2.0, 2.5, 0.0]
```
sshane, line 24: you'll also want to raise the factor -- the first variable. the second is just for documentation purposes

adeebshihadeh:
Merged the increase to panda (https://github.com/commaai/panda/pull/1284), so you can run it on your fork if you want. However, we need to improve the tuning before bumping up the limit in openpilot. Converting to draft until that happens.

vincentw56:
> adeebshihadeh

What can I do to help with tunning?

selfdrive/car/torque_data/override.yaml
```diff
@@ -23,7 +23,7 @@ FORD MAVERICK 1ST GEN: [.nan, 1.5, .nan]
...
```
sshane, line 26:
```suggestion
RAM 1500 5TH GEN: [2.5, 2.5, 0.0]
```

adeebshihadeh:
Closing until we have a new tune to go along with it.

PR #27086, opened 2023-01-25
Ram 1500: more generic EPS steer to zero check

sshane:
@realfast what do you think about this? I'd like to make the check more generic to catch new cars without a specific revision of the FW. It looks like that there's a clear pattern with the FW versions. Older major revisions allow a lower min steering speed. Same with the Chrysler check above. We might be able to unify them in the future when we see more Ram FW.


realfast:
I agree this is a great way to handle this.

PR #27091, opened 2023-01-26
Ford: use curvature signal for lateral control

sshane:
Curvature allows for better control instead of path angle signal. Made it use the common steer angle rate limit function instead of a new curvature one with duplicated logic

Naming might be confusing, added a TODO to address in the future


selfdrive/car/ford/values.py
```diff
@@ -22,12 +22,13 @@ class CarControllerParams:
...
-  LKAS_STEER_RATIO = 2.75       # Approximate ratio between LatCtlPath_An_Actl and steering angle in radians
-                                # TODO: remove this once we understand how the EPS calculates the steering angle better
-  STEER_DRIVER_ALLOWANCE = 0.8  # Driver intervention threshold in Nm
+  CURVATURE_MAX = 0.02  # Max curvature for steering command, m^-1
+  STEER_DRIVER_ALLOWANCE = 0.8  # Driver intervention threshold, Nm
 
-  RATE_LIMIT_UP = AngleRateLimit(speed_points=[0., 5., 15.], max_angle_diff_points=[5., .8, .15])
-  RATE_LIMIT_DOWN = AngleRateLimit(speed_points=[0., 5., 15.], max_angle_diff_points=[5., 3.5, 0.4])
+  # Curvature rate limits
+  # TODO: unify field names used by curvature and angle control cars.
```
sshane, line 29:
```suggestion
  # TODO: unify field names used by curvature and angle control cars
```

incognitojam (approved):

PR #27095, opened 2023-01-26
Ford: add Lincoln Aviator 2021 to Ford Explorer platform

incognitojam:
The Aviator has very similar firmware versions to the Ford Explorer. The only difference with the other Fords is the gear shifter signal `TrnGear_D_RqDrv` is not constantly set when in D (it only goes "high" for 1 frame).

`TrnRng_D_RqGsm` is a different signal we can use which is common with other Fords. I think this better represents the state of the gear shift module, not just any momentary driver input.

FW `f0709d2bc6ca451f|2022-12-10--12-36-59--0`
VIN: `5LM5J7XC8MGL09541`


selfdrive/car/interfaces.py
```diff
@@ -401,15 +401,17 @@ def parse_gear_shifter(gear: Optional[str]) -> car.CarState.GearShifter:
...
+      'R': GearShifter.reverse, 'REVERSE': GearShifter.reverse,
+      'N': GearShifter.neutral, 'NEUTRAL': GearShifter.neutral,
+      'E': GearShifter.eco, 'ECO': GearShifter.eco,
+      'T': GearShifter.manumatic, 'MANUAL': GearShifter.manumatic,
+      '_1': GearShifter.manumatic, '_2': GearShifter.manumatic, '_3': GearShifter.manumatic,
+      '_4': GearShifter.manumatic, '_5': GearShifter.manumatic, '_6': GearShifter.manumatic,
+      'D': GearShifter.drive, 'DRIVE': GearShifter.drive,
+      'S': GearShifter.sport, 'SPORT': GearShifter.sport, 'SPORT_DRIVESPORT': GearShifter.sport,
+      'L': GearShifter.low, 'LOW': GearShifter.low,
+      'B': GearShifter.brake, 'BRAKE': GearShifter.brake,
```
sshane, line 414: Why is `P`, `R`, `N`, etc here? I don't see them in the DBC

selfdrive/car/interfaces.py
```diff
@@ -401,15 +401,15 @@ def parse_gear_shifter(gear: Optional[str]) -> car.CarState.GearShifter:
...
-        'S': GearShifter.sport, 'SPORT': GearShifter.sport,
-        'L': GearShifter.low, 'LOW': GearShifter.low,
-        'B': GearShifter.brake, 'BRAKE': GearShifter.brake,
+      'P': GearShifter.park, 'PARK': GearShifter.park,
+      'R': GearShifter.reverse, 'REVERSE': GearShifter.reverse,
+      'N': GearShifter.neutral, 'NEUTRAL': GearShifter.neutral,
+      'E': GearShifter.eco, 'ECO': GearShifter.eco,
+      'T': GearShifter.manumatic, 'MANUAL': GearShifter.manumatic,
+      'D': GearShifter.drive, 'DRIVE': GearShifter.drive,
+      'S': GearShifter.sport, 'SPORT': GearShifter.sport, 'SPORT_DRIVESPORT': GearShifter.sport,
```
incognitojam, line 410: could also change the DBC if that's preferred...

PR #27105, opened 2023-01-27
paramsd: Check if roll from the localizer is actually valid

nuwandavek:
Currently, the `paramsd` filter does not use roll if `liveLocationKalman/orientationNED` is invalid. That signal is always invalid when there is no GPS, as there is no bearing (yaw) information. But roll and pitch are derived from the accelerometer, and should be valid.
Changes:
- `roll_valid` only checks if the roll/roll_std are within acceptable bounds
- When roll is not valid, op enters the `vehicleModelInvalid` event
---
ToDo
- [x] Verify by running paramsd offline on non GPS segments
- [x] Quantify how many more events would have been triggered
- [x] Quantify the improvement in performance


lushisong:
@nuwandavek Thank you for improving paramsd continually! I've had a question for a long time that why steerRatio the learner figure out tends to be 10%-20% higher than factory spec value? It doesn't matter when SteeringAngle is small(e.g. below 15°) but it usually appears constant oversteering when big SteeringAngle(e.g. 45° or more), it's obvious when driving on countryroad with big curves and the corresponding car speed is about 20-40kph. And I'm not sure the exact causes of these oversteerings, but definitely the learner's steerRatio is significantly higher.
For my car(a SUV with very low tireStiffnessFactor of 0.35), the factory spec steerRatio is 15.7, and the learner value is about 17.5~18.5, and I've noticed that Rav4 and Palisade have the same problem(learner's value 10%-20% higher, seeing https://github.com/commaai/openpilot/issues/19672#issuecomment-855327642).
Thank you again!

PR #27107, opened 2023-01-27
replay: fix hang if started with a special segment and there is no INIT_DATA in events

deanlee:
fix issue: https://github.com/commaai/openpilot/discussions/26091#discussioncomment-4791434

replay need read INIT_DATA to get the route's start ts (nano_since_boot).

### DIFF
--- a/tools/replay/replay.cc
+++ b/tools/replay/replay.cc
@@ -28,7 +28,7 @@ Replay::Replay(QString route, QStringList allow, QStringList block, SubMaster *s
 
   if (!allow_list.empty()) {
     // the following events are needed for replay to work properly.
-    allow_list.insert(cereal::Event::Which::PANDA_STATE_D_E_P_R_E_C_A_T_E_D);
+    allow_list.insert(cereal::Event::Which::INIT_DATA);
     allow_list.insert(cereal::Event::Which::CAR_PARAMS);
   }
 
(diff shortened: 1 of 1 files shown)


deanlee:
closed. 
@sshane  I think this is the real fix for your issue  : https://github.com/commaai/openpilot/pull/27111

tools/replay/replay.cc
```diff
@@ -25,6 +25,13 @@ Replay::Replay(QString route, QStringList allow, QStringList block, SubMaster *s
       s.push_back(it.name);
     }
   }
+
+  if (!allow_list.empty()) {
+    // the following events are needed for replay to work properly.
+    allow_list.insert(cereal::Event::Which::INIT_DATA);
```
deanlee, line 31: these added events will not pushed to stream(because they are not in the push socket_ array)
it's only used internally by replay.

tools/replay/replay.cc
```diff
@@ -25,6 +25,13 @@ Replay::Replay(QString route, QStringList allow, QStringList block, SubMaster *s
...
+    allow_list.insert(cereal::Event::Which::CAR_PARAMS);
+  }
+
```
deanlee, line 34: these added events will not pushed to stream(because they are not in the push socket_ array)
it's only used internally by replay.

deanlee:
these added events will not pushed to stream(because they are not in the push socket_ array)
it's only used internally by replay.
fixed in https://github.com/commaai/openpilot/pull/27107

sshane:
Seems like I'm hitting a similar bug when you only allow one service (I did `driverStateV2`), says loading forever.

deanlee:
looks like a display issue in consoleUI.

PR #27109, opened 2023-01-27
Cabana: add window icon

sshane:
@deanlee is there a better way to do this?

Screenshot from 2023-01-26 20-42-58


deanlee:
it's good.:D

adeebshihadeh:
Lol, let's at least make this a bit nicer. Maybe just bigger comma and make it a circle?

tools/cabana/cabana.cc
```diff
@@ -13,6 +13,7 @@ int main(int argc, char *argv[]) {
   initApp(argc, argv);
   QApplication app(argc, argv);
   app.setApplicationDisplayName("Cabana");
+  app.setWindowIcon(QIcon("./images/cabana.png"));
```
pd0wm, line 16: You want to use qrc to load the file, otherwise it won't work if your current working directory is wrong.

sshane:
@deanlee do you want to make a PR to add support for a tools/cabana qrc file? That's the only thing left for this :)

deanlee:
@sshane  added qrc resource file in https://github.com/commaai/openpilot/pull/27386

sshane:
@deanlee away from my desktop and cabana doesn't yet work on Win11, can you test this out?

PR #27116, opened 2023-01-27
feat: Optimize the order_nodes method

Sanix-Darker:
## WHAT and HOW

- Replacing the auto_array_ptr with a stack of pairs of integers. This eliminates the need for dynamic memory allocation and indexing into the stack.

- I have also used C++11's std::tie function to unpack the pair and assign it to the variables parent and pos. This eliminates the need for a separate struct for holding the pair.

- I also use std::stack::empty() instead of idx > 0 to check if the stack is empty.

Bonus: I also changed the naming of the variables from camelCase to snake_case, which is more common in C++.


adeebshihadeh:
This is in a third party library. The PR should be made against their repo instead.

PR #27118, opened 2023-01-27
process replay: test routes in dashcam

sshane:
process replay stopped sending can msgs, due to process replay expecting `sendcan` after 2000 frames


sshane:
@adeebshihadeh I noticed we commented out some routes at the bottom of the segment list because they are dashcammed. I wanted to get the Bronco route in, but that's also dashcam. Is this change fine with you, then we can also test the other two segments also in dashcam?

https://github.com/commaai/openpilot/blob/master/selfdrive/test/process_replay/test_processes.py#L36

adeebshihadeh:
This is fine, but I don't think there's a need to enable the other dashcam routes.

PR #27119, opened 2023-01-27
Glonass support, ubloxd add ephemeris parsing

gast04:
add glonass parsing and publishing to ubloxd, not enabled yet
diff is so large cause gps ephemeris parsing got moved to a new function


release/files_common
```diff
@@ -229,6 +229,8 @@ selfdrive/locationd/generated/ubx.cpp
 selfdrive/locationd/generated/ubx.h
 selfdrive/locationd/generated/gps.cpp
 selfdrive/locationd/generated/gps.h
+selfdrive/locationd/generated/glonass.cpp
+selfdrive/locationd/generated/glonass.h
```
adeebshihadeh, line 233: you can replace all the generated/ lines here with
```suggestion
selfdrive/locationd/generated/*.cpp
selfdrive/locationd/generated/*.h
```

adeebshihadeh:
> diff is so large cause gps ephemeris parsing got moved to a new function

can you first split that out into its own preparation PR to reduce the diff here?

HaraldSchafer:
LGTM

release/files_common
gast04: added in https://github.com/commaai/openpilot/pull/27125
which needs to be merged first

gast04:
split up, merge oder
* https://github.com/commaai/openpilot/pull/27122
* https://github.com/commaai/openpilot/pull/27125

then  I will rebase this one

adeebshihadeh (approved):
NIce job, it's much cleaner after splitting out the null-effect GPS ephemeris parsing changes!

selfdrive/locationd/ublox_msg.cc
```diff
@@ -246,11 +264,143 @@ kj::Array<capnp::word> UbloxMsgParser::parse_gps_ephemeris(ubx_t::rxm_sfrbx_t *m
...
+    eph.setSvType(data->m());
+  }
+
+  // string number 5
+  {
+    kaitai::kstream stream(glonass_strings[msg->sv_id()][5]);
+    glonass_t gl_stream(&stream);
+    glonass_t::string_5_t* data = static_cast<glonass_t::string_5_t*>(gl_stream.data());
+
+    // TODO: remove parsing string5, its only needed for the year
```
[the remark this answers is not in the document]
gast04: not sure if ever tbh, in theorie we can get the year later in python using datetime, but than we need to account leap year and rollovers and this was just not worth it in my mind to get the ephemeris 2seconds faster (time of receiving one string)

PR #27123, opened 2023-01-28
Ford: filter steering pressed

sshane:
Driver torque is a pretty noisy signal. On a harsh bumpy road, driver torque goes to 0.875 a max of 3 frames, with most being 2 frame occurrences, and one 1.0625 occurrence. The 0.8 threshold is super light on the wheel, so we may want to bump it a bit though.

54827bf84c38b14f|2023-01-27--16-11-27--0


incognitojam:
bumping it up to 1.0 Nm is probably fine

YassineYousfi:
@sshane this works much better than master on the bronco, lgtm

sshane:
@incognitojam can you quickly check that 1.0 is a good threshold and is close enough to our other cars?

PR #27124, opened 2023-01-28
Ford: driver torque safety

sshane:
Not really blending :(, just winds down torque as soon as driver touches wheel.

If you set just actuation bit to 0 but keep curvature, it will silently fault and fully wind down and back up which takes a long time. Lane changes are impossible.

If you set both signals to 0, then the ramp back up is the same as enabling, which isn't too bad for lane changes. A minor jerk is felt in curves though


sshane:
Superseded by https://github.com/commaai/openpilot/pull/27415

PR #27126, opened 2023-01-28
Add support for Malaysian Civic 2022

jayvin100:
dongle ID: 8fcafab4167b8c6c


sshane:
Your radar isn't giving a valid response so you're fuzzy FPing for now

PR #27131, opened 2023-01-28
cabana: add button to suppress highlighted bytes

pd0wm:
Add button to suppress highlighting for all bytes that are currently highlighted. Useful for finding new signals.

Renamed `HexColors` to `ChangeTracker`. Will be used too for bit level tracking in an upcoming PR.

Screenshot 2023-01-28 at 17 37 58

Screenshot 2023-01-28 at 17 38 01


deanlee:
how about move to the right of msg filters? add a stretch after lineedit

pd0wm:
Think without a stretch looks better:
Screenshot 2023-01-28 at 17 58 54

Downsides are that it looks a little cramped on smaller screens, and the buttons move around when the text on the clear button changes.

deanlee:
I have no idea now:D   the final decision is yours.

pd0wm:
Let's leave the buttons at the bottoms. I'd like to add filtering per column at some point anyway.

PR #27132, opened 2023-01-28
cabana: add --dbc command line argument

pd0wm:
Useful during development to quickly start cabana in a ready to go state.

### DIFF
--- a/tools/cabana/mainwin.cc
+++ b/tools/cabana/mainwin.cc
@@ -243,6 +243,11 @@ void MainWindow::loadDBCFromClipboard() {
 }
 
 void MainWindow::loadDBCFromFingerprint() {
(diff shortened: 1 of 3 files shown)


tools/cabana/mainwin.cc
```diff
@@ -246,6 +246,11 @@ void MainWindow::loadDBCFromClipboard() {
 }
 
 void MainWindow::loadDBCFromFingerprint() {
+  // Don't overwrite already loaded DBC
```
pd0wm, line 249: Need to test if this still works under Linux. on MacOS loading from fingerprint is broken.

pd0wm: Still works fine!

PR #27135, opened 2023-01-28
2020 Hyundai Santa Fe fingerprint update

pfeiferj:
**Description**
 Add a new transmission fw version to the 2019-2020 santa fe to allow fingerprinting.

**Verification**
Properly fingerprints my 2020 Santa Fe and enables open pilot.


incognitojam:
can you share a route ID or your dongle ID?

pfeiferj:
> can you share a route ID or your dongle ID?

Sorry for the delay getting back to you.

Route:
6cc716ffa2eb1f32|2023-02-02--14-20-45--0

PR #27138, opened 2023-01-29
GM: disallow engage shortly after pressing brake

sshane:
Just got sent a segment where the ECM will fault if you try to engage on the frame brakePressed falls: https://connect.comma.ai/fe80a4e1cedec853/1674777037139/1674777042711

- [ ] Reproduce how long exactly it takes to not fault with our current signal
- [ ] Decide if we should make this generic?
- [ ] Panda

### DIFF
--- a/selfdrive/car/gm/carstate.py
+++ b/selfdrive/car/gm/carstate.py
@@ -20,7 +20,7 @@ class CarState(CarStateBase):
     self.loopback_lka_steering_cmd_updated = False
     self.camera_lka_steering_cmd_counter = 0
     self.buttons_counter = 0
-    self.prev_brake_pressed = 0
+    self.prev_brake_pressed = False
 
   def update(self, pt_cp, cam_cp, loopback_cp):
     ret = car.CarState.new_message()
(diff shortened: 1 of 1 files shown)


selfdrive/car/gm/carstate.py
```diff
@@ -20,6 +20,7 @@ def __init__(self, CP):
     self.loopback_lka_steering_cmd_updated = False
     self.camera_lka_steering_cmd_counter = 0
     self.buttons_counter = 0
+    self.prev_brake_pressed = 0
```
sshane, line 23:
```suggestion
    self.prev_brake_pressed = False
```

Verylukyguy:
I may have had this happen on this route:  df51b2195de66a9b|2023-02-10--08-35-39--0

I backed out of my driveway, shifted into Drive, and tried to engage and got a Steering error.

sshane:
Very strange, it looks like you got a few temporary errors whenever it tried steering the first two attempts, then recovered. I've seen this be triggered in the Bolt when you go over bumps, but no where else. Please keep sending more segments if you get this again

sshane:
Unclear if this is a real bug, closing for now.

PR #27140, opened 2023-01-29
cabana: small improvements to charts

pd0wm:
Small changes. Mostly improves the layout when working on laptops where horizontal space is at a premium.

- Replace zoom out icon by magnifying glass - instead of counterclockwise arrow. Noticed the previous icon wasn't intuitive for new users.
- Remove label for current zoom window. This should be obvious from the video slider or the charts x axis. This label takes up quite a bit of space that pushes the zoom out button into the qtoolbar extension which makes it hard to find.
- Only show column count when there is space for more than one column. This also helps with not pushing the zoom out button into an extension.
- Fix computation of number of columns to properly account for layout size and spacing, preventing part of the plot to fall off.

Screenshot 2023-01-29 at 10 40 33
Screenshot 2023-01-29 at 10 40 50

### DIFF
--- a/tools/cabana/chartswidget.cc
+++ b/tools/cabana/chartswidget.cc
@@ -30,8 +30,8 @@ ChartsWidget::ChartsWidget(QWidget *parent) : QWidget(parent) {
   title_label->setContentsMargins(0, 0, 12, 0);
   columns_cb = new QComboBox(this);
   columns_cb->addItems({"1", "2", "3", "4"});
-  toolbar->addWidget(new QLabel(tr("Columns:")));
-  toolbar->addWidget(columns_cb);
+  columns_lb_action = toolbar->addWidget(new QLabel(tr("Columns:")));
+  columns_cb_action = toolbar->addWidget(columns_cb);
 
   QLabel *stretch_label = new QLabel(this);
   stretch_label->setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Preferred);
@@ -46,8 +46,7 @@ ChartsWidget::ChartsWidget(QWidget *parent) : QWidget(parent) {
   range_slider->setPageStep(60);  // 1 min
   toolbar->addWidget(range_slider);
 
-  toolbar->addWidget(zoom_range_lb = new QLabel());
-  reset_zoom_btn = toolbar->addAction(bootstrapPixmap("arrow-counterclockwise"), "");
+  reset_zoom_btn = toolbar->addAction(bootstrapPixmap("zoom-out"), "");
   reset_zoom_btn->setToolTip(tr("Reset zoom (drag on chart to zoom X-Axis)"));
   remove_all_btn = toolbar->addAction(bootstrapPixmap("x"), "");
   remove_all_btn->setToolTip(tr("Remove all charts"));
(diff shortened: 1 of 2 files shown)


tools/cabana/chartswidget.cc
```diff
@@ -166,8 +165,11 @@ void ChartsWidget::setMaxChartRange(int value) {
 }
 
 void ChartsWidget::updateToolBar() {
+  if (docking) setColumnCount(0);
```
deanlee, line 168: I think it would be better not to set it to 0 and allow multiple columns when docking,  let the charts automatically adjust according to the settings,it's a good feature for wide screen:

[Kazam_screencast_00068.webm](https://user-images.githubusercontent.com/27770/215329793-8f79d6e8-2476-4112-9aff-d615437b8624.webm)

tools/cabana/chartswidget.cc
```diff
@@ -166,8 +165,11 @@ void ChartsWidget::setMaxChartRange(int value) {
...
+  columns_lb_action->setVisible(!docking);
+  columns_cb_action->setVisible(!docking);
```
deanlee, line 170: how about call updateToolbar in resizeEvent, and set them to visible if rect.width() > CHART_MIN_WIDTH * 2?

pd0wm: That's a good solution! I was worried about the dropdown disappearing but still showing multiple columns.

tools/cabana/chartswidget.cc
```diff
@@ -230,7 +232,7 @@ void ChartsWidget::setColumnCount(int n) {
 void ChartsWidget::updateLayout() {
   int n = column_count;
   for (; n > 1; --n) {
-    if ((n * (CHART_MIN_WIDTH + charts_layout->spacing())) < rect().width()) break;
+    if ((n * CHART_MIN_WIDTH) < rect().width()) break;
```
pd0wm, line 235: charts_layout->spacing() returns `-1` for some reason. Might need a different way to compute the padding between the charts so nothing falls of.

pd0wm: image

deanlee: yes,let's move it for now.

deanlee: it returns -1 if use default system setting. we need call setSpacing to set it to a value.

pd0wm: Think there is more spacing that we need to consider. I can make more pixels disappear from the right than the spacing between plots is wide.

deanlee: looks like we need setspacing in constructor. Otherwise the chart will be out of bounds：

Screenshot from 2023-01-29 23-14-39

pd0wm: This is with `setSpacing(10)` called. Much better, but not yet perfect.

image

deanlee: the correct width should be (n*CHART_MIN_WIDTH) + (n - 1) * charts_layout->spacing()

deanlee: > This is with `setSpacing(10)` called. Much better, but not yet perfect.
> 
> image

not sure what we missed

pd0wm: Looks like I need to use `charts_layout->geometry().width()` to compare to. But that value is not valid from the resize event.

pd0wm: I'll just update the visibility in `updateLayout` since that's where the decision about the number of columns is made anyway. Looks like that fixes everything.

deanlee: great!

PR #27141, opened 2023-01-29
cabana: do not round time when seeking from chartswidget

pd0wm:
Rounding the time makes it seek to a different place then was clicked. Obvious when seeking when very zoomed in.

Also prevent zooming past the end of the route, as that causes a segfault when the start and end of the rubberband are both past the end.


pd0wm:
@deanlee Maybe we want to redo the logic to decide between zoom and click as well? Maybe this should be decided on the distance in x pixels of the rectangle, not in time so the behaviour is consistent between zoom levels.

PR #27143, opened 2023-01-29
MacOS: build fixes

martinl:
Some suggested build fixes for macOS:
 - exclude common/gpio.cc
 - exclude spi for boardd
 - exclude v4l for loggerd
 - update libdbc.dylib relative path for cabana

I tested some car porting tools (fw fingerprinting test, plotjuggler, cabana) and they are usable now on macOS 13.1

### DIFF
--- a/selfdrive/boardd/boardd.cc
+++ b/selfdrive/boardd/boardd.cc
@@ -7,6 +7,7 @@
 #include <unistd.h>
 
 #include <algorithm>
+#include <array>
 #include <atomic>
 #include <bitset>
 #include <cassert>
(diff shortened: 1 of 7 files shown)


selfdrive/boardd/boardd.cc
```diff
@@ -18,6 +18,9 @@
 #include <cstdlib>
 #include <future>
 #include <thread>
+#ifdef __APPLE__
+#include <array>
```
pd0wm, line 22: Should be no problem with always including this, right?

martinl: yes, removed the #ifdef

incognitojam: I think these should be alphabetical now :)

PR #27145, opened 2023-01-29
cabana: show dots when zoomed far into a signal

pd0wm:
Makes it more clear that values are interpolated when comparing low frequency signals to high frequency ones. Dots appear when there is more than 20 pixels of space between samples.

Screenshot 2023-01-29 at 15 45 18

### DIFF
--- a/tools/cabana/chartswidget.cc
+++ b/tools/cabana/chartswidget.cc
@@ -469,6 +469,16 @@ void ChartView::updatePlot(double cur, double min, double max) {
     axis_x->setRange(min, max);
     updateAxisY();
   }
+
+  // Show points when zoomed in enough
+  for (auto &s : sigs) {
+    auto begin = std::lower_bound(s.vals.begin(), s.vals.end(), axis_x->min(), [](auto &p, double x) { return p.x() < x; });
+    auto end = std::lower_bound(s.vals.begin(), s.vals.end(), axis_x->max(), [](auto &p, double x) { return p.x() < x; });
+
+    int pixels_per_point = width() / (end - begin);
+    s.series->setPointsVisible(pixels_per_point > 20);
+  }
+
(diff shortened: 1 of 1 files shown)


deanlee:
cool!

tools/cabana/chartswidget.cc
```diff
@@ -469,6 +469,16 @@ void ChartView::updatePlot(double cur, double min, double max) {
     axis_x->setRange(min, max);
     updateAxisY();
   }
+
+  // Show points when zoomed in enough
+  for (auto &s : sigs) {
+    auto begin = std::lower_bound(s.vals.begin(), s.vals.end(), axis_x->min(), [](auto &p, double x) { return p.x() < x; });
+    auto end = std::lower_bound(s.vals.begin(), s.vals.end(), axis_x->max(), [](auto &p, double x) { return p.x() < x; });
+
+    int pixels_per_point = width() / (end - begin);
```
deanlee, line 478: Do we need to make sure not to divide by 0 ?

pd0wm: Looks like `pixels_per_point` will just be 0 when end==begin, but will add a check for completeness.

gregjhogan:
Thanks, this checks the first item off my feature request list https://github.com/commaai/openpilot/discussions/26091#discussioncomment-4163472!

pd0wm:
@deanlee are there any performance issues if we increase the chart range to more than 5 minutes? Maybe we can detect if there are a lot of points and apply some kind of decimation on larger ranges?

deanlee:
> @deanlee are there any performance issues if we increase the chart range to more than 5 minutes? From some limited testing it only seems to slow down if you don't have OpenGL enabled. Maybe we can fix OpenGL on MacOS and increase the range?

yes, the bottleneck is drawing plot in gui thread without opengl.

PR #27149, opened 2023-01-30
cabana: syc the last messages after seeking

deanlee:
issue: the last message in historyLogs occasionally out of sync with the last message in messageList.
fixed:  fetch data with lastMessage's timestamp instead of can->currentSec.

before:

after:


pd0wm:
Is it possible to make the values sync/update when you seek while paused?

deanlee:
> Is it possible to make the values sync/update when you seek while paused?

I'll do it in another pr

deanlee:
added a function `updateLastMsgsTo` to update the last messages after seeking. all states are set to the correct time. the messageList, BinaryView, HistoryLog and signaleView are also updated to the correct last message.

`updateLastMsgsTo` runs very fast. it takes about 50ms for 11 minutes of data on an old laptop.

PR #27151, opened 2023-01-30
panda.cc: fix possible heap overflow on wrong checksum

pd0wm:
When the checksum computation fails `size`/`receive_buffer_size` is not decremented. If I understand the code correctly this broken message would stay in the buffer while all newly received data starts overflowing `receive_buffer`.

Setting `size = 0` is not guaranteed to get back in sync with the packets. So maybe you want to implement a better recovery mechanism.

(untested on device)

### DIFF
--- a/selfdrive/boardd/panda.cc
+++ b/selfdrive/boardd/panda.cc
@@ -236,6 +236,9 @@ void Panda::can_send(capnp::List<cereal::CanData>::Reader can_data_list) {
 }
 
 bool Panda::can_receive(std::vector<can_frame>& out_vec) {
+  // Check if enough space left in buffer to store RECV_SIZE data
+  assert(receive_buffer_size + RECV_SIZE < sizeof(receive_buffer));
+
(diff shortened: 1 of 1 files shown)


robbederks:
Good catch!

pd0wm:
Ah I see, boardd immediately aborts when this happens once. I was using the panda abstraction layer directly from cabana, and the process kept crashing when the checksum computation failed 😅.

robbederks:
I see. Yes, this is an additional check we added after getting rid of the per-chunk counters, to make it obvious if the whole chunking and syncing logic goes off the rails. No more weird random CAN messages on the rx / tx side of things when that happens.

Anything we can merge to make this code portable to windows? not sure what kind of bitfields you're talking about, but I assume it's because of some assumptions we're making somewhere?

pd0wm:
Actually no changes to the code where necessary. Turns out MSVC packs structs with bitfields differently than GCC. So when compiling for windows mingw gcc emulates this microsoft style packing unless you set the `-mno-ms-bitfields` flag.

Issue #26016, opened 2022-10-08
GM camera ACC: LKAS faults


sshane:
- [x] Fault when relay opens (we send counter of 1, should copy camera's counter) - https://github.com/commaai/openpilot/pull/26039
- [x] Fault when relay opens with above fix - We can be sort of unlucky and try to send a message just before the relay opens, blocking it, then reading an updated counter from the camera (that doesn't reach the car), and end up skipping a counter. It's rare, but still a possibility. - https://github.com/commaai/openpilot/pull/27163
- [x] Random LKAS faults -> perm faults - https://github.com/commaai/openpilot/pull/27250
- [x] Fault when relay closes (our counter doesn't sync up with camera's) - https://github.com/commaai/openpilot/pull/27164

ncoop23:
0def4a390f6fe5c0|2022-10-25--19-17-40--0

Just had this happen today driving along straight and level near the end of the drive. Also had a little steering tune issue as well.

dominicdoty:
I'm in a 2017 Bolt on an opgm/newcombo variant, but in case it helps your debugging, I had an LKAS fault while driving today on the segments:
`82a0516f27e12844\|2022-11-04--18-03-04--15`
`82a0516f27e12844\|2022-11-04--18-03-04--16`

abueide:
@sshane we ran into this on our bolt euv and it scared the crap out of us xD. Is there anything we can do to help fix this? We're pretty new to openpilot development so any advice would be appreciated!

minorOffense:
Just happened to me today on the bolt-torque branch (0.9.1) on a 2022 Bolt EUV

dominicdoty:
Had another LKAS fault recently, my branch which is synced to comma master LKAS sync code.

I've had LKAS faults on this exact segment of road under similar conditions 3 or 4 times, leading me to think there's some reproducible issue there.

Its a long left hand sweeping on ramp with a meter light at the end. OP disengaged, heavy traffic, stop and go at 0-10 mph.

`82a0516f27e12844|2022-12-08--17-07-17--22`

ncoop23:
[...]

PR #27163, opened 2023-01-31
GM camera ACC: reliable relay open init

sshane:
Simply changes from reading counter from bus 2 to bus 0, which is a reliable way to know the last counter the EPS has seen, so we don't skip anything. I tested turning car on 15 times without a skipped counter, whereas previously, it was easy to hit within 3 starts!

[This this change, you are now guaranteed 3 starts without a fault.](https://github.com/commaai/openpilot/pull/26039#issuecomment-1275582164)

This is the issue that is described in https://github.com/commaai/openpilot/issues/26016 as the second bullet point, which this solution catches now. See an example from `master`:

Screenshot from 2023-01-30 21-15-58

### DIFF
--- a/selfdrive/car/gm/carcontroller.py
+++ b/selfdrive/car/gm/carcontroller.py
@@ -63,7 +63,7 @@ class CarController:
     elif (self.frame - self.last_steer_frame) >= steer_step:
       # Initialize ASCMLKASteeringCmd counter using the camera until we get a msg on the bus
       if init_lka_counter:
-        self.lka_steering_cmd_counter = CS.camera_lka_steering_cmd_counter + 1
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
> Simply changes from reading counter from bus 2 to bus 0, which is a reliable way to know the last counter the EPS has seen, so we don't skip anything. I tested turning car on 15 times without a skipped counter, whereas previously, it was easy to hit within 3 starts!

Let's run this in a loop hundreds of times to ensure this is fixed.

sshane:
Tested with relay closing fix PR as well, since we would fault more often without it. We were able to start the car and open the relay 32 times before the EPS faulted from skipped counters, which were actually caused from closing the relay at the inopportune time. We never opened on the wrong counter.

6e997623cb7cdaa6|2023-01-31--00-23-41

PR #27167, opened 2023-01-31
cabana: dynamically switch between OpenGL

pd0wm:
Continuation of https://github.com/commaai/openpilot/pull/27145

Actually, it looks like this whole points feature doesn't work if you have OpenGL enabled.

https://doc.qt.io/qt-6.2/qabstractseries.html#useOpenGL-prop

> The OpenGL acceleration of series drawing is meant for use cases that need fast drawing of large numbers of points. It is optimized for efficiency, and therefore the series using it lack support for many features available to non-accelerated series:
> 
>  - Pen styles, marker shapes and light markers are ignored for accelerated series. Only solid lines and plain scatter dots are supported. The scatter dots may be circular or rectangular, depending on the underlying graphics hardware and drivers.

Switching dynamically between OpenGL enabled/disabled seems to work decent, as long as you multiply the pen width by the device pixel ratio when OpenGL is enabled.

@deanlee any thoughts?

### DIFF
--- a/tools/cabana/chartswidget.cc
+++ b/tools/cabana/chartswidget.cc
@@ -471,32 +471,32 @@ void ChartView::updatePlot(double cur, double min, double max) {
   if (min != axis_x->min() || max != axis_x->max()) {
     axis_x->setRange(min, max);
     updateAxisY();
-  }
 
-  // Show points when zoomed in enough
-  for (auto &s : sigs) {
-    auto begin = std::lower_bound(s.vals.begin(), s.vals.end(), axis_x->min(), [](auto &p, double x) { return p.x() < x; });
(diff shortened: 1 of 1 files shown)


deanlee:
lgtm

pd0wm:
I also noticed that dragging over a signal to zoom doesn't work while in OpenGL mode. I assume that has always been broken?

deanlee:
It looks like we only need to update them when the x-axis changes. how about  putting them in 

`if (min != axis_x->min() || max != axis_x->max())`

deanlee:
> I also noticed that dragging over a signal to zoom doesn't work while in OpenGL mode. I assume that has always been broken?

that's one of the qchart's opengl limitation too

deanlee:
@pd0wm I improved this function slightly, paint points in drawForeground instead of swithing between opengl.  https://github.com/commaai/openpilot/pull/27239

PR #27127, opened 2023-01-28
Fix 2017 Genesis G80 car harness and shop link in docs

SeanCox:
**Description** [](A description of the bug and the fix. Also link any relevant issues.)

The 2017 Genesis G80 Harness is different than the 2018-19 Harnesses. The 2017 also has the limitations of not being useable under 19mph for long and 37mph for lat. 

Here are some eBay links showing the difference in the lane keep assist cameras between the 2017 and 2018 year models:

2017:
https://www.ebay.com/itm/314008788332?chn=ps&_trkparms=ispr%3D1&amdata=enc%3A1qV1mvaERRaexn3OSGvsi5Q27&norover=1&mkevt=1&mkrid=711-117182-37290-0&mkcid=2&mkscid=101&itemid=314008788332&targetid=1644837434603&device=c&mktype=&googleloc=9011539&poi=&campaignid=16743749222&mkgroupid=138744546207&rlsatarget=pla-1644837434603&abcId=9300842&merchantid=7866994&gclid=Cj0KCQiAq5meBhCyARIsAJrtdr7XJYQKoOQFCMd-X7fcQNxz7Dph92aNE9KbcpgsDUjzqk6XCNASbgkaAnQCEALw_wcB

2018:
https://www.ebay.com/itm/265208107107?chn=ps&norover=1&mkevt=1&mkrid=711-117182-37290-0&mkcid=2&mkscid=101&itemid=265208107107&targetid=1645685073288&device=c&mktype=&googleloc=9011539&poi=&campaignid=16730423415&mkgroupid=135815925780&rlsatarget=pla-1645685073288&abcId=9300841&merchantid=101677473&gclid=Cj0KCQiAq5meBhCyARIsAJrtdr58z_q80cN2pr-GUm8nSa6bkIQg_9QX88OO96E6Imf5CJ1G2U58LvYaAgWVEALw_wcB

**Verification** [](Explain how you tested this bug fix.)

I bought a 2017 Genesis G80 and the comma 3 with the H harness only to find out I needed a J harness and the limitations my car has.

I have also talked to 2 other discord users that have 2017 G80s that used the J harness.

PR #27170, opened 2023-01-31
Genesis G80 2017: group with Hyundai Genesis

sshane:
The 2017 model year American G80 is actually more similar to a Hyundai Genesis 2016, and it fingerprints as such. Same minimum steering speed, enable speed, and harness, just different package

According to https://en.wikipedia.org/wiki/Genesis_G80, the G80 started production 2016, selling as the 2017 model year, so it's likely the platforms are identical, which seems to be the case. See relevant PR: https://github.com/commaai/openpilot/pull/27127


sshane:
The FW from this 2017 Genesis G80 indeed matches the Hyundai Genesis's format: https://github.com/commaai/openpilot/pull/27171

PR #27171, opened 2023-01-31
Hyundai: add FW for Genesis and Genesis G80

sshane:
Taken from only C3s with good logging:

```bash
./dongles_for_car.py "HYUNDAI GENESIS 2015-2016" 30
06353521b1f136dd
190555490a82ee55
1bc85e3b0b53e1ad
27a580d124055206
31aadcc0d26cc7f9
3687c15bdfe09295
37738bb1fb0ec3b7
445c144c95eb7409
61b3de3cce404316
85eab094426c2649
8ee0ef535a86fd75
9c803cede3069e78
ab58672499c7e756
bc1f17f055ceffc6
de5967e116d5fb68
15 total
```

```bash
./dongles_for_car.py "GENESIS G80 2017" 30
7f8f479cfa6f392a
8ac49fa47993c396
9b17f69eefc0356a
a204ef4b8bc8b99d
bcf7d921bfdfe671
e934d5bcb61dacdd
6 total
```

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -367,6 +367,13 @@ FW_QUERY_CONFIG = FwQueryConfig(
 )
 
 FW_VERSIONS = {
+  CAR.HYUNDAI_GENESIS: {
+    (Ecu.fwdCamera, 0x7c4, None): [
+      b'\xf1\x00DH LKAS 1.1 -150210',
+      b'\xf1\x00DH LKAS 1.4 -140110',
+      b'\xf1\x00DH LKAS 1.5 -140425',
+    ],
+  },
   CAR.IONIQ: {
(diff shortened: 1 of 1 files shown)


sshane:
Interesting for reliable fuzzy FP. We'll have to detect old generations with the presence of a two letter code, or it it's a version. I wonder what other differences are hiding behind having no CAN fingerprints!

@adeebshihadeh I went and got FW from C3 users last 30 days, at what point should we remove CAN fingerprints? Is there an expectation of never any regressions to fingerprint these users? I can go back further then

adeebshihadeh:
> @adeebshihadeh I went and got FW from C3 users last 30 days, at what point should we remove CAN fingerprints? Is there an expectation of never regressing to fingerprint these users? I can go back further then

Yes, no regressions! I was planning on just removing them in a subsequent release after we know fuzzy FP is solid.

PR #27173, opened 2023-02-01
glonass support laikad

gast04:
can be merged after https://github.com/commaai/laika/pull/128 and https://github.com/commaai/openpilot/pull/27119


selfdrive/locationd/laikad.py
```diff
@@ -136,10 +136,11 @@ def read_report(self, gnss_msg):
     return week, tow, new_meas
 
   def is_ephemeris(self, gnss_msg):
+    mwhich = gnss_msg.which()
     if self.use_qcom:
-      return gnss_msg.which() == 'drSvPoly'
+      return mwhich == 'drSvPoly'
     else:
-      return gnss_msg.which() == 'ephemeris'
+      return mwhich == 'ephemeris' or mwhich == 'glonassEphemeris'
```
adeebshihadeh, line 143: `in` tends to be cleaner than or's
```suggestion
      return gnss_msg.which() in ('ephemeris', 'glonassEphemeris')
```

selfdrive/locationd/laikad.py
```diff
@@ -136,10 +136,11 @@ def read_report(self, gnss_msg):
...
 
   def read_ephemeris(self, gnss_msg):
     # TODO this only works on GLONASS
```
adeebshihadeh, line 146: does this comment have to be updated? not sure what's it's referring to

selfdrive/locationd/laikad.py
```diff
@@ -149,7 +150,10 @@ def read_ephemeris(self, gnss_msg):
         return
       ephem = parse_qcom_ephem(gnss_msg.drSvPoly, self.gps_week)
     else:
-      ephem = convert_ublox_ephem(gnss_msg.ephemeris)
+      if gnss_msg.which() == 'ephemeris':
+        ephem = convert_ublox_gps_ephem(gnss_msg.ephemeris)
+      else:
+        ephem = convert_ublox_glonass_ephem(gnss_msg.glonassEphemeris)
```
adeebshihadeh, line 156: i'd do an else if here and raise something like unknown msg type in the else block

adeebshihadeh (approved):

selfdrive/locationd/laikad.py
gast04: yea, it doesnt make sense, removed it

PR #27176, opened 2023-02-01
Hyundai: remove HYUNDAI_GENESIS from legacy steer max blacklist

SeanCox:
**Description** [](A description of the bug and the fix. Also link any relevant issues.
Tested removing HYUNDAI_GENESIS from legacy

Route with steering limit reached

**Route**
Route: [a route with the bug fix]
85eab094426c2649|2023-02-01--14-01-32--11


sshane:
Did you check there's deadzone? @adeebshihadeh we also need to change the torque params for these changes

adeebshihadeh:
> Did you check there's deadzone? @adeebshihadeh we also need to change the torque params for these changes or there will be incorrect tuning

Yes. Good point on the tuning, @nuwandavek can we refactor the torque params to avoid this?

PR #27180, opened 2023-02-02
Honda Bosch: gather available ECUs from camera

### LINKED ISSUE #27075: Honda Bosch: query PT bus (1) from the camera for data collection
- [ ] Pretty much just need to support this in openpilot. Right now we never set the safety param, so bus `1` is always the OBD port.

sshane:
Nidec PR: https://github.com/commaai/openpilot/pull/27071

Closes: https://github.com/commaai/openpilot/issues/27075

Will test and merge tomorrow!


selfdrive/car/fw_versions.py
```diff
@@ -203,19 +204,29 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
...
+      matched_brand = brand
       break
 
+  # Do non-OBD queries for matched brand, or all if no match is found
+  Params().put_bool("FirmwareObdQueryDone", True)
+
+  for brand, config in FW_QUERY_CONFIGS.items():
+    if brand == matched_brand or matched_brand is None:
+      car_fw = get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, num_pandas=num_pandas, obd_multiplexing=False, debug=debug, progress=progress)
+      all_car_fw.extend(car_fw)
```
sshane, line 224:
```suggestion
      all_car_fw.extend(get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, num_pandas=num_pandas, obd_multiplexing=False, debug=debug, progress=progress))
```

selfdrive/car/fw_versions.py
```diff
@@ -203,19 +204,29 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
...
     car_fw = get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, num_pandas=num_pandas, debug=debug, progress=progress)
     all_car_fw.extend(car_fw)
     # Try to match using FW returned from this brand only
     matches = match_fw_to_car_exact(build_fw_dict(car_fw))
     if len(matches) == 1:
+      matched_brand = brand
       break
 
+  # Do non-OBD queries for matched brand, or all if no match is found
+  Params().put_bool("FirmwareObdQueryDone", True)
```
sshane, line 219: @adeebshihadeh is there any way to know when boardd has put the panda in this safety mode? There's a small delay, so this often ends up not working

selfdrive/car/fw_versions.py
```diff
@@ -216,7 +217,10 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
       break
 
   # Do non-OBD queries for matched brand, or all if no match is found
-  Params().put_bool("FirmwareObdQueryDone", True)
+  params = Params()
+  params.put_bool("FirmwareObdQueryDone", True)
+  while not params.get_bool("ObdMultiplexingFinished"):
```
adeebshihadeh, line 222: you can just specify block=True

selfdrive/car/fw_versions.py
```diff
@@ -203,19 +204,30 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
...
     # Try to match using FW returned from this brand only
     matches = match_fw_to_car_exact(build_fw_dict(car_fw))
     if len(matches) == 1:
+      matched_brand = brand
       break
 
+  # Do non-OBD queries for matched brand, or all if no match is found
+  params = Params()
+  params.put_bool("FirmwareObdQueryDone", True)
+  params.get_bool("ObdMultiplexingDisabled", block=True)
```
adeebshihadeh, line 221: let's add a cloudlog here that we're waiting for this param and another one when we get it

selfdrive/car/fw_versions.py
```diff
@@ -203,19 +204,30 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
...
+
+  for brand in FW_QUERY_CONFIGS.keys():
+    if brand == matched_brand or matched_brand is None:
```
adeebshihadeh, line 224: why do it if we don't have a match?

sshane: If the Honda Bosch fuzzy matches, I'd like to still query it

selfdrive/car/fw_versions.py
sshane: warnings?

selfdrive/car/fw_versions.py
sshane: I think we might get less data, but not required. I have a suspicion a lot of Hondas fuzzy FP because they have so many ECUs

adeebshihadeh: let's just check the data. this is cloudlogged, so you can just check kibana quickly

selfdrive/car/fw_versions.py
adeebshihadeh: yeah that works

selfdrive/car/fw_versions.py
sshane: Last 90 days:

```python
169 dongles are mostly fuzzy FP, out of 208 total dongles
15 platform are mostly fuzzy FP, out of 17 total platforms
```

PR #27184, opened 2023-02-02
navd: check duration_typical before using

gast04:
fix (happend twice this year)
```
  File "/data/openpilot/selfdrive/manager/process.py", line 43, in launcher
    getattr(mod, 'main')()
  File "/data/openpilot/selfdrive/navd/navd.py", line 329, in main
    route_engine.update()
  File "/data/openpilot/selfdrive/navd/navd.py", line 78, in update
    self.send_instruction()
  File "/data/openpilot/selfdrive/navd/navd.py", line 229, in send_instruction
    total_time_typical = step['duration_typical'] * remaining
TypeError: unsupported operand type(s) for *: 'NoneType' and 'float'
```
this seems not always be available at navigation start
it should give the time of the route under typical conditions and we use it set the ETA color, so setting it to the
total_time seems like a reasonable fallback

### DIFF
--- a/selfdrive/navd/navd.py
+++ b/selfdrive/navd/navd.py
@@ -226,7 +226,11 @@ class RouteEngine:
     remaining = 1.0 - along_geometry / max(step['distance'], 1)
     total_distance = step['distance'] * remaining
     total_time = step['duration'] * remaining
-    total_time_typical = step['duration_typical'] * remaining
+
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Do the API docs say anything about this?

gast04:
nope, havent found anything, docs only state the description of the field

PR #27199, opened 2023-02-03
cabana: fix incorrect point size after adding new series to chart or changing series type

deanlee:
**Issue**: the point size is incorrect after adding new series/changing series type:

[Kazam_screencast_00078.webm](https://user-images.githubusercontent.com/27770/216548226-152b09ba-f3bb-4f7b-b01c-241e9020fca0.webm)

**Fixed**: move point size related code to function `updateSeriesPoints`, call 'updateSeriesPoints' to update point size after create serries:

[Kazam_screencast_00077.webm](https://user-images.githubusercontent.com/27770/216548239-f37ec30e-f816-4955-8ffa-fac9df8253c7.webm)


adeebshihadeh:
Can you revert the opendbc change? Looks like that was added by mistake.

deanlee:
reverted

PR #27201, opened 2023-02-03
GM: reduce steering message rate

sshane:
I've gotten the Bolt EUV to fault, sitting in park, sending an inactive steering command message at 50Hz. I'm fairly confident that the rate is the only factor at play here, but I'm waiting for more reproductions to back up my claims. `6e997623cb7cdaa6|2023-02-02--16-30-16`

So far I've ran this 50Hz test 4 or 5 times, but only once has it faulted. The other times it's ran for 30-50 minutes with no fault, so that's the only concerning thing.

I've had a user with a Bolt run this branch for the past 5 days without any faults, and he has racked up 8 20 mile drives so far. On release3, he was getting faults every other day with similar drives (commute).


gregjhogan:
Have you tried increasing the frequency to see if faults happen more often? Or improved the logging to understand if messages are being sent closer together when it faults? Maybe something simple using just a panda can reliably reproduce the issue if it is just timing?

sshane:
Yes, I've sent at 100Hz and it either faults instantly, or lasts for a few minutes. I'd like to make the results more consistent though.

sshane:
Fault at 50Hz inactive:

- 6e997623cb7cdaa6|2023-02-03--12-21-40
- 6e997623cb7cdaa6|2023-02-02--16-30-16

sshane:
We're going to enforce at least 15ms from receiving the 128 readback message and adding a new message to `sendcan`, so this means 33Hz will be required, else we'll very inconsistently switch sending at 50Hz and 33Hz. Going to merge this and then follow it up with the timing enforcement here: https://github.com/commaai/openpilot/pull/27250/files

sshane:
process replay shows the new steer step is about 25Hz, but I believe that is because of the CAN 128 readback msgs in the logs fixed to 50Hz. Will be fixed when we get new routes.

PR #26372, opened 2022-11-05
cabana: handling extra DBC info

deanlee:
Screenshot from 2022-11-05 11-37-27

PR #27203, opened 2023-02-03
cabana: add support for load&save extra dbc info

deanlee:
closed https://github.com/commaai/opendbc/pull/739, https://github.com/commaai/openpilot/pull/26372, https://github.com/commaai/openpilot/pull/26377. load/save extra info in dbcManager.
 

Screenshot from 2023-02-04 03-10-00


deanlee:
refactor dbcmanager to better support extra dbc info and Multiplexed Signals (https://github.com/commaai/opendbc/pull/777) in the future.

pd0wm:
I was working on this a bit as well, and represented the values as an `std::vector<std::pair<int, std::string>> val` in opendbc. This should make working with the values later much easier.

deanlee:
got a coredump after https://github.com/commaai/openpilot/pull/27258#event-8473003462. I'm fixing it

pd0wm:
Hmm, must have missed something when rebasing my changes on latest master. Where do you need to check for sig not being a nullptr?

deanlee:
normally we don't need to check if sig==null, the only exception is in updateSeries,  the sig may be null (for read all signals), we need get signal pointer from s.sig

pd0wm:
What do you think of storing the description as a list of key value pairs? That allows us to show the correct description in the signal widget, and implement a table based editor.

deanlee:
good idea

deanlee:
view/edit signal value description in table:


Screenshot from 2023-02-17 21-07-54

PR #27204, opened 2023-02-03
micd: make sure function always returns >=0 db

GregorKikelj:


system/micd.py
```diff
@@ -16,7 +16,7 @@
 def calculate_spl(measurements):
   # https://www.engineeringtoolbox.com/sound-pressure-d_711.html
   sound_pressure = np.sqrt(np.mean(measurements ** 2))  # RMS of amplitudes
-  if sound_pressure > 0:
+  if sound_pressure >= REFERENCE_SPL:
```
sshane, line 19:
```suggestion
  if sound_pressure > REFERENCE_SPL:
```

if equal, it will be 0 anyway

GregorKikelj: True, we could also try 
```suggestion
  sound_pressure_level = 20 * np.log10(max(1, sound_pressure / REFERENCE_SPL))
```
to keep it short

system/micd.py
```diff
@@ -16,10 +16,8 @@
 def calculate_spl(measurements):
   # https://www.engineeringtoolbox.com/sound-pressure-d_711.html
   sound_pressure = np.sqrt(np.mean(measurements ** 2))  # RMS of amplitudes
-  if sound_pressure > 0:
-    sound_pressure_level = 20 * np.log10(sound_pressure / REFERENCE_SPL)  # dB
-  else:
-    sound_pressure_level = 0
+  sound_pressure = max(REFERENCE_SPL, sound_pressure) # Make sure measurements being 0 doesn't cause an issue
```
sshane, line 19:
```suggestion
  sound_pressure = max(REFERENCE_SPL, sound_pressure)
```

system/micd.py
```diff
@@ -16,10 +16,8 @@
...
+  sound_pressure_level = 20 * np.log10(sound_pressure / REFERENCE_SPL)  # dB
```
sshane, line 20: Actually we don't want to modify sound_pressure since it's returned

system/micd.py
```diff
@@ -16,10 +16,8 @@
 def calculate_spl(measurements):
   # https://www.engineeringtoolbox.com/sound-pressure-d_711.html
   sound_pressure = np.sqrt(np.mean(measurements ** 2))  # RMS of amplitudes
-  if sound_pressure > 0:
-    sound_pressure_level = 20 * np.log10(sound_pressure / REFERENCE_SPL)  # dB
-  else:
-    sound_pressure_level = 0
+  sound_pressure_clamped = max(REFERENCE_SPL, sound_pressure) # Make sure measurements being 0 doesn't cause an issue
+  sound_pressure_level = 20 * np.log10(sound_pressure_clamped / REFERENCE_SPL)  # dB
```
sshane, line 20: the spl from this is only used in one place, maybe we should just clip it when publishing? not sure if it makes sense to clip this function

sshane: actually it does make sense for the sound pressure to be less than the reference, so sound pressure level being negative also is okay. think it's safe to close

GregorKikelj:
Not an ideal fix, closing

PR #27207, opened 2023-02-03
laikad: add laikad startup tests

gast04:


tools/gpstest/test_laikad.py
```diff
@@ -0,0 +1,105 @@
+#!/usr/bin/env python3
+import os
```
adeebshihadeh, line 2: a unit test for laikad should go in the `tests/` directory where laikad is. `tools/gpstset` is just for the infra we have here to spoof GPS

gast04: hmm, I dont fully agree for this cause it needs the gps infra to be tested, thats why I would like to keep the startup timing tests in this directory

PR #27212, opened 2023-02-04
Toyota: Car Port for Toyota C-HR 2021 (TSS2)

eFiniLan:
**Checklist**
  - [x] route with openpilot (stock logitudinal) - ea8fbe72b96a185c|2023-02-08--15-11-46
  - [x] route with stock system - ea8fbe72b96a185c|2023-02-08--15-01-09


ErichMoraga:
You understand this PR is effectively more than just a car port, since you're also sneaking in the CAN filter modification, right?

selfdrive/car/toyota/values.py
```diff
@@ -2067,15 +2087,15 @@ class ToyotaCarInfo(CarInfo):
...
+            CAR.MIRAI, CAR.LEXUS_NX_TSS2, CAR.LEXUS_NXH_TSS2, CAR.ALPHARD_TSS2, CAR.AVALON_TSS2, CAR.AVALONH_TSS2, CAR.ALPHARDH_TSS2, CAR.CHR_TSS2}
 
 NO_DSU_CAR = TSS2_CAR | {CAR.CHR, CAR.CHRH, CAR.CAMRY, CAR.CAMRYH}
 
 # the DSU uses the AEB message for longitudinal on these cars
 UNSUPPORTED_DSU_CAR = {CAR.LEXUS_IS, CAR.LEXUS_RC}
 
 # these cars have a radar which sends ACC messages instead of the camera
-RADAR_ACC_CAR = {CAR.RAV4H_TSS2_2022, CAR.RAV4_TSS2_2022}
+RADAR_ACC_CAR = {CAR.RAV4H_TSS2_2022, CAR.RAV4_TSS2_2022, CAR.CHR_TSS2}
```
sshane, line 2098: Erich's right, this PR should strictly be to add a lateral-only port for the C-HR

eFiniLan: Cool, I will have a separate PR for smart dsu value availability 👍

sshane:
Is this PR ready for review? The car drives just fine? Can you also add a lateral only route? The can data might not pass our tests and I'd prefer no custom hardware.

eFiniLan:
> Is this PR ready for review? The car drives just fine as a TSS2 platform car? Can you also add a lateral only route? The can data might not pass our tests and I'd prefer no custom hardware.

hi Shane, yes I have reverted back to stock longitudinal (so no radar interceptor installed), the route in routes.py is op lat + stock long.

I'm still uploading logs I guess that's why it's failing, do I need to make the log public and preserved?

sshane:
Thanks! I need to run a script on my end to make the car model test pass, I'll do that tomorrow.

Do you know why the docs said we already supported the 2021? I'm worried that this might be a regional thing.

eFiniLan:
I think it was added because the vehicle was mentioned and confirmed working in #toyota back in 2021 when I tried to port it (back then we didn't have 2022 RAV4 w/ Conti radar merged yet so I didn't manage to submit the PR).

sshane:
I see, sounds good! https://github.com/commaai/openpilot/pull/23078

sshane:
thanks @eFiniLan!

PR #27218, opened 2023-02-04
cabana: fix incorrect clamp

deanlee:
min &max is the seconds from routeStartTime.  this incorrect clamp makes the chart  unzoomable


pd0wm:
My bad, must have only tested with streams where the start time was zero 🙁

deanlee:
I also often get confused between relative time and absolute time. :D

PR #27236, opened 2023-02-06
setup: replace software selection widget with input dialog

### LINKED ISSUE #105: Remove default options from setup, only URL

incognitojam:
remove the "Choose Software to Install" screen and jump straight to inputting an installer URL after confirming wi-fi network.

|New|Old|
|-|-|
|Removed|Choose Software to Install (custom)|

closes https://github.com/commaai/agnos-builder/issues/105

### DIFF
--- a/selfdrive/ui/qt/setup/setup.cc
+++ b/selfdrive/ui/qt/setup/setup.cc
@@ -170,7 +170,16 @@ QWidget * Setup::network_setup() {
   QPushButton *cont = new QPushButton();
   cont->setObjectName("navBtn");
   cont->setProperty("primary", true);
-  QObject::connect(cont, &QPushButton::clicked, this, &Setup::nextPage);
+  QObject::connect(cont, &QPushButton::clicked, [=]() {
+    QString url = InputDialog::getText(tr("Enter installer URL"), this, "");
+    if (url.isEmpty()) {
+      return;
+    }
+    setCurrentWidget(downloading_widget);
+    QTimer::singleShot(1000, this, [=]() {
+      download(url);
+    });
+  });
   blayout->addWidget(cont);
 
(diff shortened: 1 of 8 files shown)


selfdrive/ui/qt/setup/setup.cc
```diff
@@ -170,7 +170,16 @@ QWidget * Setup::network_setup() {
   QPushButton *cont = new QPushButton();
   cont->setObjectName("navBtn");
   cont->setProperty("primary", true);
-  QObject::connect(cont, &QPushButton::clicked, this, &Setup::nextPage);
+  QObject::connect(cont, &QPushButton::clicked, [=]() {
+    QString url = InputDialog::getText(tr("Enter installer URL"), this, "");
```
adeebshihadeh, line 174: let's put back the old prompt here
```
Enter URL
for Custom Software
```

selfdrive/ui/qt/setup/setup.cc
```diff
@@ -170,7 +170,16 @@ QWidget * Setup::network_setup() {
...
+    if (url.isEmpty()) {
+      return;
+    }
+    setCurrentWidget(downloading_widget);
```
adeebshihadeh, line 178: looks like this is slightly different. the logic here was carefully done to avoid glitches when opening/closing the input popup. did you test that?

incognitojam: my bad, I see the problem

adeebshihadeh: cool, setup specifically has to be super well tested since we don't use it everyday

adeebshihadeh:
Can you open an agnos-builder PR to update the setup? You can just copy the one from your device in.

PR #27237, opened 2023-02-06
setup: add serial to request headers

incognitojam:
```
* Connected to nightly.comma.ai (20.188.93.37) port 80 (#0)
> GET / HTTP/1.1
Host: nightly.comma.ai
User-Agent: AGNOSSetup-
Accept: */*
X-openpilot-serial: cccccc
...
```

### DIFF
--- a/system/hardware/tici/hardware.h
+++ b/system/hardware/tici/hardware.h
@@ -7,19 +7,6 @@
 #include "common/util.h"
 #include "system/hardware/base.h"
 
-std::string get_device_serial() {
-  std::ifstream stream("/proc/cmdline");
-  std::string cmdline;
-  std::getline(stream, cmdline);
-
-  auto start = cmdline.find("serialno=");
-  if (start != std::string::npos) {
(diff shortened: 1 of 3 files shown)


system/hardware/tici/hardware.h
```diff
@@ -21,6 +21,23 @@ class HardwareTici : public HardwareNone {
   static int get_voltage() { return std::atoi(util::read_file("/sys/class/hwmon/hwmon1/in1_input").c_str()); };
   static int get_current() { return std::atoi(util::read_file("/sys/class/hwmon/hwmon1/curr1_input").c_str()); };
 
+  static std::string get_serial() {
+    static std::string serial("");
```
adeebshihadeh, line 25: why is this static?

adeebshihadeh (approved):

system/hardware/tici/hardware.h
incognitojam: I thought it would be slow so this is to cache it

PR #27241, opened 2023-02-07
setup: check installer is elf

incognitojam:
Setup will show the "Download failed" screen if a non-elf file is downloaded

Enter Custom Software URL (1)
Download Failure


adeebshihadeh:
We probably want to show some more specific text in this case.

PR #27250, opened 2023-02-08
GM: enforce steering command message timing

sshane:


selfdrive/car/body/carcontroller.py
```diff
@@ -35,7 +35,7 @@ def deadband_filter(torque, deadband):
       torque -= deadband
     return torque
 
-  def update(self, CC, CS):
+  def update(self, CC, CS, now_nanos):
```
sshane, line 38: `log_mono_time`?

selfdrive/controls/controlsd.py
```diff
@@ -731,7 +731,7 @@ def publish_logs(self, CS, start_time, CC, lac_log):
 
     if not self.read_only and self.initialized:
       # send car controls over can
-      self.last_actuators, can_sends = self.CI.apply(CC)
+      self.last_actuators, can_sends = self.CI.apply(CC, int(sec_since_boot() * 1e9))
```
sshane, line 734: we can also make a messaging helper function to get log mono time, copied what it does now

sshane:
We can also get rid of `self.sent_lka_steering_cmd` if we check the CAN parser timestamp, I'll do that cleanup in another PR

adeebshihadeh (approved):

sshane:
Working as expected. With and without at 50Hz:

Screenshot from 2023-02-09 15-20-02
Screenshot from 2023-02-09 15-21-57

PR #27258, opened 2023-02-08
cabana: extend color palette, make chart colors match signal view

pd0wm:
Generate "random" but nice looking colors. I'm open to a different color generation function, but this seems to produce acceptable results and is deterministic based on signal properties.

2023-02-08-123824_493x800_scrot
2023-02-08-123804_814x715_scrot

### DIFF
--- a/tools/cabana/binaryview.cc
+++ b/tools/cabana/binaryview.cc
@@ -173,7 +173,6 @@ void BinaryViewModel::refresh() {
   if ((dbc_msg = dbc()->msg(msg_id))) {
     row_count = dbc_msg->size;
     items.resize(row_count * column_count);
-    int i = 0;
     for (auto sig : dbc_msg->getSignals()) {
       auto [start, end] = getSignalRange(sig);
       for (int j = start; j <= end; ++j) {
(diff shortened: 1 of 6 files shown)


deanlee:
Wooooooow! awesome!

deanlee:
found some ui issues in your screenshot (not related to this PR). but the cell spacing and button size looks good on my laptop.

pd0wm:
I'm using i3, which probably has a very minimal theme with no extra spacing/padding.

deanlee:
> the most notable thing is the lack of border to resize the messages/graph window. There only some very minimal dots to show where to drag.

yes i was bothered by it too. I can't even see the dots.

adeebshihadeh:
Looks great!

Side note: @pd0wm anything you're still missing from web cabana? I'd like to deprecate web cabana soon.

pd0wm:
Think usability is way past web cabana :)

My whish list:
- A nice list of drives pulled from the API if you don't start with a command line argument would be nice. I've been experimenting with a startup wizard to choose a source.
- A prebuilt (single statically linked binary) cabana built by github actions that you can just download and run.

PR #27259, opened 2023-02-08
cabana: bit level change frequency highlighting

pd0wm:
Makes it much easier to spot location of potential signals. Fast moving signals will also show a nice gradient from LSB to MSB.

2023-02-08-111851_674x462_scrot
2023-02-08-111827_684x309_scrot

2023-02-08-112452_671x597_scrot
2023-02-08-112448_674x377_scrot

### DIFF
--- a/tools/cabana/binaryview.cc
+++ b/tools/cabana/binaryview.cc
@@ -7,6 +7,8 @@
 #include <QScrollBar>
 #include <QToolTip>
 
+#include <cmath>
+
 #include "tools/cabana/commands.h"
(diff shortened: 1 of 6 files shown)


gregjhogan:
Seems harder to see where each signal starts and ends now, maybe we should add a solid border of the darkest color for the signal around the set of bits for each signal?

pd0wm:
Will look into it!

pd0wm:
@gregjhogan This should fix it: https://github.com/commaai/openpilot/pull/27293

PR #27260, opened 2023-02-08
cabana: set correct geometry for legend

deanlee:
Screenshot from 2023-02-08 20-35-04


pd0wm:
Looks better on my machine
2023-02-08-190407_688x925_scrot

deanlee:
It seems that there is still a problem, the color marker is not aligned at the v-center.

deanlee:
looks good after remove margins and setGeometry in resizeEvent:

Screenshot from 2023-02-09 02-26-34

PR #27267, opened 2023-02-09
HKG CAN FD: detect hybrid/EV

sshane:
Detects the accelerator signal (what we really care about with ICE/hybrid/EV distinction) so we can merge a lot of platforms!

This is mainly needed to fix an issue where a CAN-FD Tucson Hybrid is fingerprinting as a Tucson, since we only have access to the camera and radar FW versions which are identical here.

Doing CAN-FD first here, but may just do everything at once. Draft until I clean this up


sshane:
don't see a need yet

PR #27268, opened 2023-02-09
Hyundai CAN-FD: use common PT bus function

sunnyhaibin:
Utilizes the helper function to detect PT bus on HKG CAN-FD platforms. This PR benefits:
- https://github.com/commaai/openpilot/pull/27267

### DIFF
--- a/selfdrive/car/hyundai/hyundaicanfd.py
+++ b/selfdrive/car/hyundai/hyundaicanfd.py
@@ -1,11 +1,11 @@
 from common.numpy_fast import clip
 from selfdrive.car.hyundai.values import HyundaiFlags
 
-def get_e_can_bus(flags):
+def get_e_can_bus(CP):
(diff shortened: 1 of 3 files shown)


selfdrive/car/hyundai/interface.py
```diff
@@ -2,6 +2,7 @@
 from cereal import car
 from panda import Panda
 from common.conversions import Conversions as CV
+from selfdrive.car.hyundai.hyundaicanfd import get_e_can_bus
```
sshane, line 5:
```suggestion
```

selfdrive/car/hyundai/hyundaicanfd.py
```diff
@@ -1,11 +1,11 @@
 from common.numpy_fast import clip
 from selfdrive.car.hyundai.values import HyundaiFlags
 
-def get_e_can_bus(CP):
+def get_e_can_bus(flags):
```
sshane, line 4: need to change usage in hyundaicanfd.py

PR #27270, opened 2023-02-09
cabana: set series color in createSeries

deanlee:
Fixed the issue that the  series color will be reset to default color after switching type:

[Kazam_screencast_00016.webm](https://user-images.githubusercontent.com/27770/217867690-88a9eca8-d490-4a5b-ae5a-37b76dd03ca5.webm)

after:

[Kazam_screencast_00014.webm](https://user-images.githubusercontent.com/27770/217868168-7565082c-7ce6-4b2c-aee3-5fa730a5561d.webm)


pd0wm:
I think this breaks updating the color when the signal color changes.

deanlee:
Y, I didn't notice that. still need to update color in updateseries, It should work now.

PR #27278, opened 2023-02-09
Add several missing CHR_TSS2 firmwares

ErichMoraga:
`@Dk#3633` 2021 C-HR  DongleID/route  4930c054f086c108|2023-02-09--18-39-48 Verified working great, per user.

Hot off the heels of @eFiniLan's https://github.com/commaai/openpilot/pull/27212 :wink:


sshane:
Nice! US 2021 has the same radar as the RAV4 2022

PR #27279, opened 2023-02-10
controlsd: higher default set speed in experimental mode

adeebshihadeh:


selfdrive/controls/lib/drive_helpers.py
```diff
@@ -8,10 +8,12 @@
...
 #          model predictions above this speed can be unpredictable
-V_CRUISE_MAX = 145  # kph
-V_CRUISE_MIN = 8  # kph
-V_CRUISE_ENABLE_MIN = 40  # kph
-V_CRUISE_INITIAL = 255  # kph
+# V_CRUISE's are in kph
+V_CRUISE_MIN = 8
+V_CRUISE_MAX = 145
+V_CRUISE_UNSET = 255
+V_CRUISE_INITIAL = 40
```
adeebshihadeh, line 15: @sshane I renamed `V_CRUISE_INITIAL` to `V_CRUISE_UNSET` since it didn't seem to make sense

sshane: Sounds good. Can we add a test case?

PR #27280, opened 2023-02-10
GM: remove redundant CarController class variable

sshane:
If we trust the CANParser, `CS.loopback_lka_steering_cmd_ts_nanos` should be 0 if it has never seen that message.


selfdrive/car/gm/carcontroller.py
```diff
@@ -30,7 +30,6 @@ def __init__(self, dbc_name, CP, VM):
     self.cancel_counter = 0
 
     self.lka_steering_cmd_counter = 0
-    self.sent_lka_steering_cmd = False
```
adeebshihadeh, line 33: this is not a good assumption. you can add valid/alive flag to the can parser if you want instead

adeebshihadeh: you can also just check canValid

sshane: the old assumption wasn't good or this one?

sshane: made it more explicit by only updating CS's ts_nanos on explicit update of the message

PR #27286, opened 2023-02-10
GM camera ACC: display speed on cluster

sshane:
Message appears to exist on some ASCM cars, but is all zeros, except the first byte at `0xc8`. Quickly checked and didn't see any other similar message for the Volt.

opendbc PR: https://github.com/commaai/opendbc/pull/780

TODO:
- [ ] Check units
- [ ] Check it exists on all camera ACC cars


sshane:
This signal is lower rate and can be noisy sometimes. Going to just add hysteresis here: https://github.com/commaai/openpilot/pull/27301

PR #27289, opened 2023-02-10
cabana: use different surface format for chart and cameraview

deanlee:
@pd0wm does this fix the opengl issue on mac?

### DIFF
--- a/tools/cabana/cabana.cc
+++ b/tools/cabana/cabana.cc
@@ -8,12 +8,20 @@
 #include "tools/cabana/streams/livestream.h"
 #include "tools/cabana/streams/replaystream.h"
 
+void hand_term(int s) {
+  std::signal(s, SIG_DFL);
+  qApp->quit();
+}
+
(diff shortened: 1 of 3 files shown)


pd0wm:
It segfaults on startup:
```
tools/cabana/_cabana --demo
services  std::vector(can, roadEncodeIdx, carParams, wideRoadEncodeIdx)
loading route  "4cf7a6ad03080c90|2021-09-29--13-46-36"
load route 4cf7a6ad03080c90|2021-09-29--13-46-36 with 11 valid segments
seeking to 0 s, segment 0
loading segment 0...
connecting to stream 0, was connected to 0
UNSUPPORTED (log once): POSSIBLE ISSUE: unit 0 GLD_TEXTURE_INDEX_2D is unloadable and bound to sampler type (Float) - using zero texture because texture unloadable
Starting listener for: camerad
[1]    87891 segmentation fault  tools/cabana/_cabana --demo
```

PR #27293, opened 2023-02-10
cabana: draw borders around signals

pd0wm:
After adding bit level highlighting it was a bit harder to see where the signal edges where. This should make things more clear. Had to move LSB/MSB indicators up slightly to make everything fit.

Screenshot 2023-02-10 at 10 34 24

As you can see if the shape is concave very small gaps appear at the two inside corners. If you want I can add a bunch more code to handle this case, or we leave it like this.
Screenshot 2023-02-10 at 10 34 47

Before:
Screenshot 2023-02-10 at 10 40 13

### DIFF
--- a/tools/cabana/binaryview.cc
+++ b/tools/cabana/binaryview.cc
@@ -266,10 +266,28 @@ BinaryItemDelegate::BinaryItemDelegate(QObject *parent) : QStyledItemDelegate(pa
   hex_font.setBold(true);
 }
 
+bool BinaryItemDelegate::isSameColor(const QModelIndex &index, int dx, int dy) const {
+  QModelIndex index2 = index.sibling(index.row() + dy, index.column() + dx);
+  if (!index2.isValid()) {
+    return false;
+  }
+
+  auto item1 = (const BinaryViewModel::Item *)index.internalPointer();
+  auto item2 = (const BinaryViewModel::Item *)index2.internalPointer();
+
+  QColor color1 = item1->bg_color;
+  QColor color2 = item2->bg_color;
+
+  // Ignore alpha
+  return (color1.red() == color2.red()) && (color2.green() == color2.green()) && (color1.blue() == color2.blue());
+}
+
 void BinaryItemDelegate::paint(QPainter *painter, const QStyleOptionViewItem &option, const QModelIndex &index) const {
   auto item = (const BinaryViewModel::Item *)index.internalPointer();
   BinaryView *bin_view = (BinaryView *)parent();
   painter->save();
+
+  const int border_width = 2;
 
   if (index.column() == 8) {
     painter->setFont(hex_font);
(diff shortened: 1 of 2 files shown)


deanlee:
this is the result that paint border with 1px width, Qt::DashLine style:

Screenshot from 2023-02-10 20-57-57
Screenshot from 2023-02-10 20-58-11
Screenshot from 2023-02-10 20-57-44

gregjhogan:
I guess that doesn't look great, but maybe it is OK.  An alternative I can think of instead of the border would be a solid color line between MSB and LSB or only border on the bottom (arrow at end showing wrap direction when multiple lines? maybe that is overkill)

adeebshihadeh:
I kind of like the dotted lines.

pd0wm:
I tried the dotted lines, but I don't like how they don't match up on the boundaries between bits. But you like dash/dots more I'm happy to change it :)

deanlee:
y, the dotted line is not good look.

I googled for some hexview designs, and I feel that it will look better if add some padding on the left and right:

HexEditor11
de

deanlee:
It should look better if make binaryView more compact with small fonts.

pd0wm:
Rough draft. If you'd like to move forward with this I can clean up the gaps and ensuring the background doesn't draw beyond the outlines.
Screenshot 2023-02-11 at 11 19 40

deanlee:
y,It looks better than the origin version. what do you think? If you think it's okay too, then I'll  make binview more compact based on it.

pd0wm:
Think it looks better with the padding. I won't have time till next week to clean it up though. So if you want you can also merge it as is and me or @deanlee can submit a PR later to add the paddings.

deanlee:
It can be merged as it as. I'll get the rest of the work done if you're busy.

deanlee:
looks better after  apply this to a smaller binView:

Screenshot from 2023-02-11 20-27-50
Screenshot from 2023-02-11 20-26-13
Screenshot from 2023-02-11 20-25-46
Screenshot from 2023-02-11 20-25-32

pd0wm:
Are you sure you want to make the binary view more compact? This is the main piece of the UI that the user interacts with when they are adding new signals. You're sitting in the car, interacting with something, and in the meantime you try to see which bits are changing. I also noticed that at the current size the UI is pretty usable on a touchscreen.

Maybe you can make a checkbox to toggle between large/compact view? I imagine the compact view might be more desirable with CAN-FD.

Maybe this PR is not a suitable place to discuss this though.

deanlee:
Sorry I didn't take that into consideration, you are right. I will undo this part of the changes.

deanlee:
the origin size:

Screenshot from 2023-02-11 21-30-24

pd0wm:
2px looks good. Feel free to take my code and open a new PR if you want, then I'll close this one.

gregjhogan:
This is what I was trying to describe, I think it looks pretty nice.
image

adeebshihadeh:
That looks great, and agree that it would be nice to get the MSB/LSB visually "attached" to the underline.

gregjhogan:
https://github.com/commaai/openpilot/pull/27308

deanlee:
looks good, but I'm worry about areas like this, if their color is similar to others, it's hard to distinguish quickly

gregjhogan:
Only showing MSB for these single bit signals is weird I think.  Maybe we can switch it to just the colored bar about the same width as the MSB text (and remove the MSB text)?

deanlee:
I made a table and compared the two modes, the border seems clearer and consistent.

| line  | border |
| Screenshot from 2023-02-12 02-55-17  | Screenshot from 2023-02-12 02-57-41  |
| Screenshot from 2023-02-12 02-54-10  | Screenshot from 2023-02-12 02-59-08 |
| Screenshot from 2023-02-12 02-52-57 | Screenshot from 2023-02-12 03-01-54 |
|Screenshot from 2023-02-12 02-52-42 | Screenshot from 2023-02-12 03-02-55 |

gregjhogan:
I really like seeing the wrap direction that the color line gives you when the signal is multiple rows.  I think the information might still be easier to quickly understand with the single bit fields changed to something better and the MSB/LSB integrated into the color line with border. I do like the better separation your changes have for when adjacent signals have similar colors.

deanlee:
yes, I looooved that line too when I first saw it. but since all the color blocks are still mixed together, IMO, it didn't solve the most important problem we were trying to solve in the first place: distinguishing different signal and non-signal regions at a glance.

deanlee:
Maybe we can add a button to toggle between the two modes? they each have pros and cons

sshane:
The borders are definitely easier to understand on the phone screen I'm using right now

PR #27295, opened 2023-02-10
cabana: improve SignalItemDelegate::paint

deanlee:
1. smaller color label
2. no bold ( There is already a color label and tree indicator for signals, the bold should only be used for the most important items)

Screenshot from 2023-02-11 03-20-45

before:

Screenshot from 2023-02-11 03-21-57


adeebshihadeh:
The before and after pictures are super helpful! Can you also put them in a table so they show up side-by-side?

deanlee:
ok

PR #27304, opened 2023-02-11
events: reword ESP off alert

sshane:
Not every one knows what ESP stands for or means. Though I'm not sure about using `Electronic Stability Program`, or `Electronic Stability Control`, or any other variant out there.

### DIFF
--- a/selfdrive/controls/lib/events.py
+++ b/selfdrive/controls/lib/events.py
@@ -729,8 +729,8 @@ EVENTS: Dict[int, Dict[str, Union[Alert, AlertCallbackType]]] = {
   },
 
   EventName.espDisabled: {
-    ET.SOFT_DISABLE: soft_disable_alert("ESP Off"),
-    ET.NO_ENTRY: NoEntryAlert("ESP Off"),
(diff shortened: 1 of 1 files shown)


selfdrive/controls/lib/events.py
```diff
@@ -729,8 +729,8 @@ def joystick_alert(CP: car.CarParams, CS: car.CarState, sm: messaging.SubMaster,
   },
 
   EventName.espDisabled: {
-    ET.SOFT_DISABLE: soft_disable_alert("ESP Off"),
-    ET.NO_ENTRY: NoEntryAlert("ESP Off"),
+    ET.SOFT_DISABLE: soft_disable_alert("Electronic Stability Program Off"),
+    ET.NO_ENTRY: NoEntryAlert("Electronic Stability Program Off"),
```
adeebshihadeh, line 733: I'd also change off -> disabled

adeebshihadeh: also think ESC is more general than ESP. I believe ESP refers to bosch's ESC

PR #27308, opened 2023-02-11
cabana: colored line between MSB and LSB

gregjhogan:
image

### DIFF
--- a/tools/cabana/binaryview.cc
+++ b/tools/cabana/binaryview.cc
@@ -291,10 +291,18 @@ void BinaryItemDelegate::paint(QPainter *painter, const QStyleOptionViewItem &op
   }
 
 
+  QColor bg = item->bg_color;
+  bg.setAlpha(255);
   painter->drawText(option.rect, Qt::AlignCenter, item->val);
   if (item->is_msb || item->is_lsb) {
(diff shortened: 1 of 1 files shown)


gregjhogan:
Would probably look even better with a gray border around the MSB/LSB and colored line with rounded corners.  @deanlee maybe you can do that in a separate PR?

gregjhogan:
The border around this color bar is really needed for when the signal is changing at the max frequency since then the entire cell is the same color as the color bar.

gregjhogan:
I added a color line for single bit fields, I think this makes them easier to read (and the MSB/LSB doesn't really apply anyway)
image

gregjhogan:
I thought it would look best replacing MSB/LSB with the colored line being an arrow the appropriate direction, but I can't seem to get the border to draw nicely at a diagonal.
image

deanlee:
much better after apply style line.

gregjhogan:
Updated with the best I could come up with since the arrow thing looked like garbage.  We can always improve it more in the future (more padding between signals, something indicating endianness for signals not spanning two rows, etc...).
image

adeebshihadeh:
Merged the border, which I think looks pretty good. Do you still prefer this?

PR #27245, opened 2023-02-07
GM: detect unplugged or missing radar

nworb-cire:
Force pushed to the branch so github won't let me reopen #26450

PR #27311, opened 2023-02-12
Car docs: pass docs flag to car interfaces

sshane:


sshane:
@adeebshihadeh Do you want to keep the car interfaces agnostic of docs, or is this fine? Volkswagen already detects docs by checking the CAN fingerprint, and so will GM here: https://github.com/commaai/openpilot/pull/27245

PR #27316, opened 2023-02-12
cabana: add name validation when editing message

deanlee:
1. check for duplicate message name
2. can't  use 'untitled' as message name


tools/cabana/detailwidget.cc
```diff
@@ -205,12 +206,28 @@ EditMessageDialog::EditMessageDialog(const QString &msg_id, const QString &title
...
+  connect(btn_box, &QDialogButtonBox::accepted, this, &QDialog::accept);
+  connect(btn_box, &QDialogButtonBox::rejected, this, &QDialog::reject);
+}
 
-  connect(buttonBox, &QDialogButtonBox::accepted, this, &QDialog::accept);
-  connect(buttonBox, &QDialogButtonBox::rejected, this, &QDialog::reject);
+void EditMessageDialog::validateName(const QString &text) {
+  bool valid = false;
+  error_label->setVisible(false);
+  if (!text.isEmpty() && text != original_name && text != "untitled") {
```
adeebshihadeh, line 222: let's pull out `untitled` into a constant

adeebshihadeh (changes requested):

PR #27331, opened 2023-02-14
 GM: update steering limits

nuwandavek:
GM platforms (verified on `Bolt`, `Silverado`) have an internal rate limit on the `actuatorsOutput/steer` that we request. If the requested rate is higher than the internal rate, it has some integrator-wind-up like effects with the steering wheel not moving proportional the requested steer torque, for sometime before resuming.

Associated panda PR: https://github.com/commaai/panda/pull/1239


sshane:
I'd check what the Volt and Acadia's limits are first. If they're identical, it looks good!

PR #27333, opened 2023-02-14
process replay: don't upload if diff doesn't exist

sshane:
Likely process replay crashed instead of failed, and not all logs are available. No point in wasting container space uploading logs that won't be used


Silvrbckw (commented):
Merge

sshane:
That's correct

PR #27323, opened 2023-02-13
Hyundai: detect LFA HDA message

sshane:
Similar to https://github.com/commaai/openpilot/pull/27322, but this should be guaranteed. I'll verify on data and merge!

Haven't checked extensively, but I believe we can merge a few platforms with no other differences (KIA_STINGER for example), which improves fuzzy FP.

PR #27335, opened 2023-02-14
HKG: detect LKAS message format

sshane:
A lot of the platforms in this list we got rid of here: https://github.com/commaai/openpilot/pull/27323 are also in this list, so we can't merge many platforms until we auto-detect this list as well. It's not a 1-to-1 with the SEND_LFA cars, and I think that's correct (sadly).

Checked a few cars which didn't have the LFA message, but did have this different LKAS format.

Also checked the fwdCamera FW versions, thinking that the prescense of `LKAS` or `MFC` described this, but still nope. The Konas have `LKAS` in their FW versions, but use the new format. Then checked for all the cars with `MFC` in their FW version, but some use the old format.


sshane:
Looks like this is the last list of HKG that's holding us up on merging platforms. Once this is merged, a lot more than https://github.com/commaai/openpilot/pull/27370 can be merged! Just need to figure out how to detect this.

sshane:
Checked a few thousand random segments and of the common addresses for all cars, there are no bits on any bus to describe the format exactly, so I'll try to determine if we can combine the platforms another way.

sshane:
There are two dongles of 80k segments that don't have this message when they should: 26e73a0e32642dc4 with `HYUNDAI IONIQ ELECTRIC 2020` and 6bc5e4e66135831a with a `HYUNDAI SONATA 2020`. Might be valid exceptions, looking into it.

sshane:
I'm also looking back at the LKAS11 message since I don't trust that every car for the platforms in the list is correct (we found some international cars didn't have the LFA message for example in the last PR). For each dongle id, these are the signals that change, so we shouldn't use them to detect this format:

```python
# find constant signal values for each dongle id
non_constant_sigs = set()
for dongle, sigvals in sig_vals_per_dongle.items():
  for sig_name, vals in sigvals.items():
    if len(vals) > 1:
      non_constant_sigs.add(sig_name)

print(non_constant_sigs)
>>> {'CF_Lkas_LdwsRHWarning', 'CF_Lkas_FcwOpt_USM', 'CF_Lkas_HbaSysState', 'CF_Lkas_HbaLamp', 'CF_Lkas_SysWarning', 'CF_Lkas_FusionState', 'CF_Lkas_Unknown1', 'CF_Lkas_LdwsLHWarning', 'CF_Lkas_HbaOpt', 'CF_Lkas_LdwsOpt_USM', 'CF_Lkas_ActToi', 'CF_Lkas_LdwsSysState', 'CF_Lkas_FcwOpt', 'CF_Lkas_FcwSysState', 'CF_Lkas_LdwsActivemode'}
```

adeebshihadeh:
Still relevant?

sshane:
Yes, but I haven't figured out how to detect it yet. Closing until I do

PR #27338, opened 2023-02-14
cabana: add keyboard shortcuts to binary view

pd0wm:
- Delete signal: x, backspace, delete
- Toggle endianness: e
- Toggle signedness: s
- Open chart: c, p, g (Chart, Plot, Graph)

Any thoughts on how to explain this to the user?

### DIFF
--- a/tools/cabana/binaryview.cc
+++ b/tools/cabana/binaryview.cc
@@ -7,9 +7,11 @@
 #include <QMouseEvent>
 #include <QPainter>
 #include <QScrollBar>
+#include <QShortcut>
 #include <QToolTip>
 
 #include "tools/cabana/commands.h"
+#include "tools/cabana/signaledit.h"
 #include "tools/cabana/streams/abstractstream.h"
 
 // BinaryView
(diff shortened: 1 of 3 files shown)


pd0wm:
@deanlee any ideas on how to cleanly set the edited/new signal to be the new `hovered_sig`?

deanlee:
there is a signals in SIgnalView:

> signals:
  void highlight(const Signal *sig);

`emit highlight(sig)` can notify bin view to set `hovered_sig` to `sig` . Is this what you need?

deanlee:
call BinaryView::highlight(signal) or `emit highlight(sig)` in SignalView. Do these two methods meet your needs?

pd0wm:
How do I get a pointer to the new signal?

deanlee:
the old pointer stay valid.

tools/cabana/binaryview.cc
```diff
@@ -62,19 +62,27 @@ void BinaryView::addShortcuts() {
   QShortcut *shortcut_endian = new QShortcut(QKeySequence(Qt::Key_E), this);
   QObject::connect(shortcut_endian, &QShortcut::activated, [=]{
     if (hovered_sig != nullptr) {
+      const Signal *hovered_sig_prev = hovered_sig;
       Signal s = *hovered_sig;
       s.is_little_endian = !s.is_little_endian;
       emit editSignal(hovered_sig, s);
+
+      hovered_sig = nullptr;
+      highlight(hovered_sig_prev);
```
deanlee, line 71: can call highlight(hovered_sig) here, hovered_sig will not changed after refresh, no need to save it to a temp variable hovered_sig_prev.

deanlee:
> Any thoughts on how to explain this to the user?

added an online-help: https://github.com/commaai/openpilot/pull/27349 .

PR #27339, opened 2023-02-14
locationd: adjust GPS alive check for QCOM GPS

adeebshihadeh:
This check was added in 88423e25df8bfbd8e6a3275eb086f468cf9a59f1 and broke QCOM GPS. This is just used for an alert and the reset tracker. Does the reset tracker need to be retuned with this change?


nuwandavek (approved):
@gast04 I'm actually not sure why this was changed. This particular PR looks good. But unless there is a reason, I'd rather we know instantly when there is no gps on the locationd `GpsOk` flag.

gast04:
this was also the first logic we had, I changed it back cause gps ok always returned true when no more locations are received, that caused issues when driving into a covered parking lot

PR #27343, opened 2023-02-14
Expand lateral MPC to 10s

haraschax:


selfdrive/controls/lib/lateral_mpc_lib/SConscript
```diff
@@ -46,7 +46,7 @@ generated_files = [
 acados_dir = '#third_party/acados'
 acados_templates_dir = '#third_party/acados/acados_template/c_templates_tera'
 
-source_list = ['lat_mpc.py',
```
adeebshihadeh, line 49: we should fix this up, it's too easy to write this bug. I don't remember why we don't just do the code generation every time plannerd runs? if we definitely need the compilation step, I can write something that avoids manually managing this list

selfdrive/controls/lib/lateral_mpc_lib/SConscript
```diff
@@ -46,7 +46,7 @@ generated_files = [
...
+source_list = ['lat_mpc.py', '../../../modeld/constants.py',
```
adeebshihadeh, line 49:
```suggestion
source_list = ['lat_mpc.py',
  #selfdrive/modeld/modeld/constants.py',
```

PR #27345, opened 2023-02-14
FPv2: log responses from data collection queries

sshane:
Marks responses to queries added for logging, so we don't fingerprint on them later. Fixes a bug where this Civic 2022 always fuzzy fingerprints due to data collection response replacing missing real response (not supported for some reason by the radar on this specific car): https://github.com/commaai/openpilot/pull/27126

### DIFF
--- a/selfdrive/car/fw_versions.py
+++ b/selfdrive/car/fw_versions.py
@@ -233,18 +233,9 @@ def get_fw_versions(logcan, sendcan, query_brand=None, extra=None, timeout=0.1, 
   parallel_addrs = []
   logging_addrs = []
   ecu_types = {}
-  # logging_addrs = defaultdict(list)
 
   for brand, brand_versions in versions.items():
-    # print(brand, brand_versions.keys())
-    # print(brand, brand_versions["debug"])
     for candidate, ecu in brand_versions.items():
-      # if brand not in logging_addrs:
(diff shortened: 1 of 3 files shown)


sshane:
This PR will add the `logging` field to all `CarFw` items in the `carFw` list that:

- Are returned from a data-collection-only request
- Correspond to an `extra_ecu` (ECUs that are added to just gather their FW versions)

selfdrive/car/fw_versions.py
```diff
@@ -232,15 +231,19 @@ def get_fw_versions(logcan, sendcan, query_brand=None, extra=None, timeout=0.1,
...
-      for ecu_type, addr, sub_addr in c.keys():
+    for candidate, ecu in brand_versions.items():
+      for ecu_type, addr, sub_addr in ecu.keys():
         a = (brand, addr, sub_addr)
         if a not in ecu_types:
           ecu_types[a] = ecu_type
 
+        if a not in logging_addrs and candidate == "debug":
+          logging_addrs.append(a)
+
```
sshane, line 246: continues an old convention, want to refactor eventually

sshane:
@Ktrtkd are you still experiencing issues?

Ktrtkd:
> @Ktrtkd are you still experiencing issues?

For this topic no.

PR #27346, opened 2023-02-15
Honda: HR-V 2023 radarless support

Ktrtkd:
**Description** Adds Brazilian HRV/Vezel 2023Model
**Verification** Fingerprint
**Route**
Test Route: 147613502316e718|2023-02-14--19-59-09

<!--- ***** Template: Car port *****

**Checklist**
- [ ] route with openpilot:

### DIFF
--- a/selfdrive/car/honda/values.py
+++ b/selfdrive/car/honda/values.py
@@ -95,6 +95,7 @@ class CAR:
   RIDGELINE = "HONDA RIDGELINE 2017"
   INSIGHT = "HONDA INSIGHT 2019"
   HONDA_E = "HONDA E 2020"
+  HRV_2023 = "HONDA HRV 2023 NON US"
 
 
 class Footnote(Enum):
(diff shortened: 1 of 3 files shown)


selfdrive/car/honda/values.py
```diff
@@ -1517,6 +1545,7 @@ def init_make(self, CP: car.CarParams):
   CAR.INSIGHT: dbc_dict('honda_insight_ex_2019_can_generated', None),
   CAR.HONDA_E: dbc_dict('acura_rdx_2020_can_generated', None),
   CAR.CIVIC_2022: dbc_dict('honda_civic_ex_2022_can_generated', None),
+  CAR.HRV_2023: dbc_dict('honda_hrv_2023_can_generated', None),
```
sshane, line 1548: Where is this DBC? Can you use the radarless Civic one?

sshane:
Thanks for the PR! Do you know if the radarless aspect is different in the US? If not, we can remove that part in the car docs

selfdrive/car/honda/values.py
Ktrtkd: Probably still not sure. This HR-V is based on Honda Vezel. Not the US model.  Japan/Thailand/Brazil. This data comes from a Brazillian HR-V

Ktrtkd:
> Thanks for the PR! Do you know if the radarless aspect is different in the US? If not, we can remove that part in the car docs

I think its similar, I might made a mistake to pull this here. Still need to test properly in my local branch.

selfdrive/car/honda/values.py
Ktrtkd: I will try the civic2022 dbc.

sshane:
Any reason you closed the PR?

Ktrtkd:
> 

Yup. Re-doing everthing from scratch by adding under Civic 22 in another repo and doing proper tests. 
But having the same results. Not sure how to handle this, commented on https://github.com/commaai/openpilot/pull/27345

PR #27352, opened 2023-02-15
cabana: refactor message id from QString to struct

pd0wm:
Packing and unpacking QString to get the bus and address didn't seem like the cleanest solution. Was running into issues when experimenting with supporting virtual messages derived from can data (e.g. parsing UDS) that needed extra identifiers. With this struct in place it should be easy to add other identifying fields to `MessageId`. This can also be used for things like automotive ethernet where the identifier contains an IP address.

Switched to `std::optional<MessageId>` when empty strings were used to denote that no message id was set.

@deanlee Any thoughts? Any specific reason you started with using QString to represent the source, addr pair?

### DIFF
--- a/tools/cabana/messageswidget.cc
+++ b/tools/cabana/messageswidget.cc
@@ -151,30 +151,27 @@ void MessageListModel::setFilterString(const QString &string) {
 }
 
 void MessageListModel::sortMessages() {
-  // TODO: Add proper comparison operators to MessageId
   beginResetModel();
   if (sort_column == 0) {
     std::sort(msgs.begin(), msgs.end(), [this](auto &l, auto &r) {
-      auto ll = std::pair{msgName(l), l.source};
-      auto rr = std::pair{msgName(r), r.source};
+      auto ll = std::pair{msgName(l), l};
+      auto rr = std::pair{msgName(r), r};
       return sort_order == Qt::AscendingOrder ? ll < rr : ll > rr;
     });
   } else if (sort_column == 1) {
     std::sort(msgs.begin(), msgs.end(), [this](auto &l, auto &r) {
-      auto ll = std::pair{l.source, l.address};
-      auto rr = std::pair{r.source, r.address};
-      return sort_order == Qt::AscendingOrder ? ll < rr : ll > rr;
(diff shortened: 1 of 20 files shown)


tools/cabana/messageswidget.cc
```diff
@@ -106,7 +106,7 @@ QVariant MessageListModel::data(const QModelIndex &index, int role) const {
   if (role == Qt::DisplayRole) {
     switch (index.column()) {
       case 0: return msgName(id);
-      case 1: return id;
+      case 1: return id.toString(); // TODO: put source and address in separate columns
```
pd0wm, line 109: Noticed the "bus:addr" format is confusing for new users. But functional changes are for a separate PR.

deanlee:
no special reason. is just for initial quick development.  In fact, I have long wanted to change it to this.

deanlee:
how about use MessageId for `updateLastMsgsTo` too?
https://github.com/commaai/openpilot/blob/7f6085805690eace8d19068a319aea2d1fffea27/tools/cabana/streams/abstractstream.cc#L61

pd0wm:
Good to know. Hopefully this can be merged quickly, otherwise we'll have a bunch of conflicts.

adeebshihadeh:
Looks like this has conflicts already, must've been something I merged before seeing this. I'll get this merged as soon as they're fixed!

adeebshihadeh:
Until we've deprecated web cabana and consider the new one stable, I'm also fine with one of you guys managing your own master branch and periodically opening a PR to sync up. Should even be easy enough to have GH Actions do it on a regular schedule.

pd0wm:
@adeebshihadeh Just rebased!

I'm worried the PR will get too big or we get out of sync with upstream changes. If you don't mind quickly looking over the PRs I'm happy with the current situation.

adeebshihadeh:
Don't mind at all, just an offer if that works better for you guys :)

jon-chuang:
Hi, I'm getting e.g.
```
./tools/cabana/dbcmanager.h:33:1: error: unknown type name 'Q_DECLARE_METATYPE'
```

error after this PR. Any ideas on how to fix?

jon-chuang:
Turns out you need `#include <QVariant>` for Qt5

PR #27355, opened 2023-02-15
CAN fingerprint script improvements

sshane:
Matches online better

### DIFF
--- a/selfdrive/debug/get_fingerprint.py
+++ b/selfdrive/debug/get_fingerprint.py
@@ -22,7 +22,7 @@ while True:
   for c in lc.can:
     # read also msgs sent by EON on CAN bus 0x80 and filter out the
     # addr with more than 11 bits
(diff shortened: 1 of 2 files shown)


selfdrive/debug/get_fingerprint.py
```diff
@@ -22,7 +22,7 @@
   for c in lc.can:
     # read also msgs sent by EON on CAN bus 0x80 and filter out the
     # addr with more than 11 bits
-    if c.src in [0, 2] and c.address < 0x800:
+    if c.src % 0x80 == 0 and c.address < 0x800 and c.address not in (0x7df, 0x7e0, 0x7e8):
```
sshane, line 25: 2 can be an entirely different bus on some cars (ASCM GM) with different lengths for the same addrs. Instead of checking 0 and 2, check 0 and 128 which includes:

relay open:
- messages originating from PT side and cam side (except ASCM GM where 2 isn't the same bus)

relay closed:
- messages originating from PT side
- messages forwarded to PT side from cam side
- messages sent to PT side from OP

PR #27276, opened 2023-02-09
GM: Escalade 2017 support

rickygilleland:
2017 Escalade port. Also adds support for fingerprinting v2 to GM vehicles.

- [x] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot: ef8f2185104d862e|2023-02-09--18-37-13
- [x] route with stock system: ef8f2185104d862e|2023-02-09--18-56-47

PR #27357, opened 2023-02-16
GM: common longitudinal actuator delay

sshane:
This helped on the Acadia (https://github.com/commaai/openpilot/pull/25354), as well as the new Bolt EV/EUV where both cars have a large delay to initially start braking (similar to lateral friction we must overcome). The Escalade merged here also seems to have this same delay: https://github.com/commaai/openpilot/pull/27276

This PR adds this delay to all GMs.


sshane:
Our Bolt: Screenshot from 2023-02-15 15-50-11

sshane:
Waiting for a Silverado (camera ACC) and Escalade (ASCM) to try this and will merge.

sshane:
Escalade: https://github.com/commaai/openpilot/pull/27406

sshane:
Silverado also has similar lag: fe80a4e1cedec853|2023-03-20--20-24-23

sshane:
Moving the Volt over as well, as the regen that is supposedly non-laggy isn't uniform across all speeds, and non-existent at low speeds, so the brakes are also reportedly laggy there.

PR #27362, opened 2023-02-16
cabana: fix updating tabbar_ids on "Close Other Tabs"

pd0wm:
Missed keeping the tab bar in sync with the ids on the "close other tabs" action. Would be great if you could attach extra data to the tabbar like on comboboxes...


deanlee:
we need to clear message_ids in  removeAll

y, it's better to set data to tabData.

deanlee:
fixed in https://github.com/commaai/openpilot/pull/27368

PR #27368, opened 2023-02-17
cabana: attach messageid to tabdata

deanlee:
save messageid to tabdata, remove the variable 'message_ids' used to hold the ids. 
also fixed a issues after this change:  setMessage will be called twice on first click


pd0wm:
Ah great that this is possible after all. I looked in the docs but couldn't find it.

PR #27369, opened 2023-02-17
SANTA_FE_2022: add missing transmission FW

chrihart:

**Checklist**
- [ x] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [ ] route with openpilot:


sshane:
Can you add your dongle id?

chrihart:
Dongle ID:  1ad7728ce54a161d

PR #27380, opened 2023-02-18
MPC path in UI

haraschax:


selfdrive/ui/ui.cc
```diff
@@ -79,9 +79,11 @@ void update_line_data(const UIState *s, const cereal::ModelDataV2::XYZTData::Rea
...
 
-void update_model(UIState *s, const cereal::ModelDataV2::Reader &model) {
+void update_model(UIState *s, 
+                  const cereal::ModelDataV2::Reader &model,
+                  const cereal::UiPlan::Reader &plan) {
   UIScene &scene = s->scene;
-  auto model_position = model.getPosition();
+  auto model_position = plan.getPosition();
   float max_distance = std::clamp(model_position.getX()[TRAJECTORY_SIZE - 1],
                                   MIN_DRAW_DISTANCE, MAX_DRAW_DISTANCE);
```
sshane, line 88: This segfaults on old routes since the list will be empty. Not sure if we care about replaying old routes though. We might want to check that `uiPlan` is updated in the calling function, or check the length of `position` here.

HaraldSchafer: willdo

selfdrive/ui/ui.cc
```diff
@@ -79,10 +79,15 @@ void update_line_data(const UIState *s, const cereal::ModelDataV2::XYZTData::Rea
...
+                  const cereal::ModelDataV2::Reader &model,
+                  const cereal::UiPlan::Reader &plan) {
   UIScene &scene = s->scene;
-  auto model_position = model.getPosition();
-  float max_distance = std::clamp(model_position.getX()[TRAJECTORY_SIZE - 1],
+  auto plan_position = plan.getPosition();
+  if (plan_position.getX().size() > 0){
+    plan_position = model.getPosition();
+  }
+  float max_distance = std::clamp(plan_position.getX()[TRAJECTORY_SIZE - 1],
```
sshane, line 90: probably want to flip this

HaraldSchafer: Why does it matter?

PR #27391, opened 2023-02-19
ui: more granular accel gradient

### LINKED ISSUE #27377: UI: improve acceleration visualization in exp mode path
Hard to distinguish between the default yellow color and accel green. Make the default color white or silver?

sshane:


selfdrive/ui/qt/onroad.cc
```diff
@@ -511,24 +521,77 @@ void AnnotatedCameraWidget::drawLaneLines(QPainter &painter, const UIState *s) {
...
+      qDebug() << "max_gradient_point:" << max_gradient_point;
     }
-    start_hue = 60;
-    // speed up: 120, slow down: 0
-    end_hue = fmax(fmin(start_hue + acceleration_future * 45, 148), 0);
+//    float gradient_height = bg.finalStop().y();
+    for (int i = 0; i < right_points.length(); i++) {
+      const auto &acceleration = sm["uiPlan"].getUiPlan().getAccel();
+      float acceleration_future = 0;
+      if (i > acceleration.size()) {
```
sshane, line 543:
```suggestion
      if (i >= acceleration.size()) {
```

selfdrive/controls/plannerd.py
```diff
@@ -22,6 +22,7 @@ def publish_ui_plan(sm, pm, lateral_planner, longitudinal_planner):
   uiPlan.position.x = np.interp(plan_odo, model_odo, lateral_planner.lat_mpc.x_sol[:,0]).tolist()
   uiPlan.position.y = np.interp(plan_odo, model_odo, lateral_planner.lat_mpc.x_sol[:,1]).tolist()
   uiPlan.position.z = np.interp(plan_odo, model_odo, lateral_planner.path_xyz[:,2]).tolist()
+  uiPlan.accel = longitudinal_planner.a_desired_trajectory_full.tolist()
```
adeebshihadeh, line 25: let's split the uiPlan and planner changes into their own PR

sshane: already asked @HaraldSchafer to merge this separately

sshane:
@deanlee we'd like to get this in for 0.9.2, but there may be some performance issues. I couldn't find any reason for why rounding certain QColor values caused better performance, but I may be measuring it wrong and it may not matter. For example, the time to draw the path's gradient seems to go up, but when taking a look at the total frame draw time, it doesn't seem to be much different at all.

Do you mind taking a look at the performance to decide if there is an issue or not and how to improve it?

deanlee:
I'll testing it on pc and try to find the bottleneck. although there may be some differences between running it on a device.

sshane:
@deanlee did you find anything?

deanlee:
Not much progress:(. 
there are too many brush switches and save/restore state in `paintGL`. they all results in a loss of performance.  from the long term perspective,  it might be better to change onroad widget to `qgraphicview', set viewport to CameraWidget,  split the hud  to multiple `QGraphicsItem`, and cache them carefully in opengl texture.  change to QGraphicsView should improve performance quite a bit.

We can also try tweaking `setSamples()` in `setQtSurfaceFormat`, reduce it from 16 to 8 or 11, this might help a little bit.(not very sure)

adeebshihadeh:
>  when taking a look at the total frame draw time, it doesn't seem to be much different at all.

I didn't realize _overall time_ hasn't changed. If we're confident in that, then this is good to merge.

selfdrive/ui/qt/onroad.cc
```diff
@@ -511,24 +517,54 @@ void AnnotatedCameraWidget::drawLaneLines(QPainter &painter, const UIState *s) {
...
+    std::copy_if(right_points.begin(), right_points.end(), std::back_inserter(valid_points),
+                 [&](const QPointF &point) { return point.y() >= 0 && point.y() <= height(); });
+    qDebug() << "valid_points len:" << valid_points.length() << "right_points len:" << right_points.length();
+
+    int drawn = 0;
+    int final_idx = 0;
+    if (!valid_points.isEmpty()) {
+      for (int p = 0; p < MAX_UI_POINTS; p++) {
+        int i = (p / (MAX_UI_POINTS - 1)) * (valid_points.length() - 1);
+        final_idx = i;
```
sshane: this now ensures we: 1. always draw the last visible point, 2. always draws max 8 points, but think we can simplify this a bit and sacrifice those two by not enforcing them.

plus, when the path is very close to the vehicle, there will still be many points that are pretty much the same color, so we should enforce point distances instead of indexes.

sshane:
By just drawing half the points, we can significantly reduce the frame time mean and standard deviation.

Here are the frame draw times on the old ui_accel_gradient branch:

```python
Experimental mode: True
drawTimeMillis - mean, std: 8.272704999089378 3.167139487730116
Samples: 561553 (468 segments)
Experimental mode: False
drawTimeMillis - mean, std: 5.821154543572167 1.7535524668115001
Samples: 56654 (47 segments)
```

And here are the frame draw times on the most recent ui_accel_gradient branch drawing half-points:

```python
Experimental mode: True
drawTimeMillis - mean, std: 4.529638814409807 1.2606483899192586
Samples: 91195 (76 segments)
Experimental mode: False
drawTimeMillis - mean, std: 4.270750440440242 1.241070306852941
Samples: 10477 (9 segments)
```

And here is data from the fleet on this branch (old version drawing all points):

```python
branch,            mean time,          std time,          ~segments
release3         : 5.6188635471500845, 1.9414304008117007, 965894.365
master           : 5.513534462997544, 1.767649620346747, 105920.11
master-ci        : 5.590394380513677, 1.7636259702959756, 123464.92083333334
ui_accel_gradient: 7.913238576212197, 3.0676772445545213, 422.79833333333335
```

selfdrive/ui/qt/onroad.cc
sshane: too complicated

selfdrive/ui/qt/onroad.cc
```diff
@@ -492,6 +492,12 @@ void AnnotatedCameraWidget::updateFrameMat() {
       .translate(-intrinsic_matrix.v[2], -intrinsic_matrix.v[5]);
 }
 
+float interp1d(float value, float start_min, float start_max, float end_min, float end_max) {
+  value = std::max(start_min, std::min(start_max, value));
+  float factor = (value - start_min) / (start_max - start_min);
+  return end_min + factor * (end_max - end_min);
+}
+
```
sshane, line 500: todo: turn into equations and clamp until we have a common interp function. @adeebshihadeh do we have a need for this anywhere else?

adeebshihadeh: if it's generic, it should go in util.h, regardless if there's any current use cases

sshane: 🤦‍♂️ there's already a function in util that does exactly this!

sshane:
Drawing with half points produces no perceivable difference in the final drawn path.

selfdrive/ui/qt/onroad.cc
```diff
@@ -511,24 +511,34 @@ void AnnotatedCameraWidget::drawLaneLines(QPainter &painter, const UIState *s) {
...
+    const QVector<QPointF> right_points = scene.track_vertices.mid(0, scene.track_vertices.length() / 2);
+
+    for (int i = 0; i < right_points.length(); i++) {
+      const auto &acceleration = sm["uiPlan"].getUiPlan().getAccel();
+      if (i >= acceleration.size()) break;
+
+      // Some points are out of frame
+      if (right_points[i].y() < 0 || right_points[i].y() > height()) continue;
+
+      // flip so 0 is bottom of frame
```
sshane, line 525:
```suggestion
      // Flip so 0 is bottom of frame
```

PR #27394, opened 2023-02-19
fix(cabana): include #qvariant

jon-chuang:
Building with Qt5 requires this header for the file to compile. 
See https://github.com/commaai/openpilot/pull/27352#issuecomment-1435941500

### DIFF
--- a/tools/cabana/dbcmanager.h
+++ b/tools/cabana/dbcmanager.h
@@ -3,6 +3,7 @@
 #include <map>
 #include <QList>
 #include <QString>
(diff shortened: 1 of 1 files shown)


jon-chuang:
Overlap with: https://github.com/commaai/openpilot/pull/27390. Not sure which is more complete.

sshane:
looks like QVariant happens to include the two includes in https://github.com/commaai/openpilot/pull/27390, so using that as it seems more explicit.

PR #27395, opened 2023-02-19
Added FW for 2023 Hyundai Santa Cruz Limited

mattcrum:
dongle id: 7fad154b0e7243fb
fwdCamera and fwdRadar fingerprint added


sunnyhaibin:
Make sure you update [docs/CARS.md](https://github.com/commaai/openpilot/blob/master/docs/CARS.md) when you bump the model year.

PR #27397, opened 2023-02-20
Hyundai CAN-FD: common PT bus function cleanup

sunnyhaibin:
**Cleanup of**
- https://github.com/commaai/openpilot/pull/27268


selfdrive/car/hyundai/hyundaicanfd.py
```diff
@@ -1,6 +1,7 @@
 from common.numpy_fast import clip
 from selfdrive.car.hyundai.values import HyundaiFlags
 
+
```
sshane, line 4:
```suggestion
```

PR #27403, opened 2023-02-20
GM: revert custom Bolt torque tune 

adeebshihadeh:
Some Bolt users reported this tuning was a significant regression.


jettoblack:
I have a Bolt EUV 2022 and the current 0.9.1 master seems much improved over 0.9.0 release.  I hope you aren't totally reverting it.

adeebshihadeh:
> I have a Bolt EUV 2022 and the current 0.9.1 master seems much improved over 0.9.0 release. I hope you aren't totally reverting it.

Unfortunately, it wasn't an improvement for all users. We're reverting it until we're able to ship a tune that improves all Bolts. For now, you can run #27414 if you'd like this tune.

PR #27404, opened 2023-02-20
laikad: use new cache param

gast04:
add compatibility if laikad cache is from latest master but a branch is checked out before the laikad cache update, old laikad would crash on with new cache file

### DIFF
--- a/common/params.cc
+++ b/common/params.cc
@@ -139,6 +139,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"IsUpdateAvailable", CLEAR_ON_MANAGER_START},
     {"JoystickDebugMode", CLEAR_ON_MANAGER_START | CLEAR_ON_IGNITION_OFF},
     {"LaikadEphemeris", PERSISTENT | DONT_LOG},
+    {"LaikadEphemerisV2", PERSISTENT | DONT_LOG},
     {"LanguageSetting", PERSISTENT},
     {"LastAthenaPingTime", CLEAR_ON_MANAGER_START},
     {"LastGPSPosition", PERSISTENT},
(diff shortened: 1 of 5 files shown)


selfdrive/locationd/laikad.py
```diff
@@ -90,7 +90,9 @@ def load_cache(self):
   def cache_ephemeris(self, t: GPSTime):
     if self.save_ephemeris and (self.last_cached_t is None or t - self.last_cached_t > SECS_IN_MIN):
       put_nonblocking(EPHEMERIS_CACHE, json.dumps(
-        {'version': CACHE_VERSION, 'last_fetch_navs_t': self.last_fetch_navs_t, 'navs': self.astro_dog.navs},
+        {'version': CACHE_VERSION, 'last_fetch_navs_t': self.last_fetch_navs_t, 'navs': self.astro_dog.navs,
```
adeebshihadeh, line 93: do we need backwards compatibility? often, it's difficult to guarantee backwards compatibility, and it doesn't give you much anyway. usually it's easier to have robust detection for backwards incompatibility and bust the cache

gast04: shane run into this, so it definitely might happen

i agree it doesnt give much,
what you mean with robust detection?
we have cache versioning but wasnt used back then, not sure how to handle differently

common/params.cc
```diff
@@ -139,6 +139,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"IsUpdateAvailable", CLEAR_ON_MANAGER_START},
     {"JoystickDebugMode", CLEAR_ON_MANAGER_START | CLEAR_ON_IGNITION_OFF},
     {"LaikadEphemeris", PERSISTENT | DONT_LOG},
```
adeebshihadeh, line 141: let's remove the old one

adeebshihadeh (approved):
I'd still like to improve the cache restore robustness, but that's a big change to get in for 0.9.1. Let's merge this as-is, then fix that up in another PR.

PR #27405, opened 2023-02-20
improved fan malfunction detection

adeebshihadeh:
previous false negative: `2dfdef71f033f267|2023-02-13--11-24-25`

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -280,8 +280,9 @@ class Controls:
 
     # Alert if fan isn't spinning for 5 seconds
     if self.sm['peripheralState'].pandaType != log.PandaState.PandaType.unknown:
-      if self.sm['peripheralState'].fanSpeedRpm == 0 and self.sm['deviceState'].fanSpeedPercentDesired > 50:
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
@koch-cf I just read your support ticket, and looks like we may just need to update the fan stall thresholds for your device. Your device is in a weird zone right now because it spins just enough to not trigger the malfunction alert and _right before_ the stall check (see panda/board/drivers/fan.h). Both of those should be addressed in this PR, can you give it a try?

koch-cf:
Hello,
I’m sorry but I have already processed a return for the device and have purchased a replacement. I’ll try the new device with the new PR and let you know if I have any issues.
Thank you,Chris Koch


Sent from Yahoo Mail for iPhone


On Monday, February 20, 2023, 6:51 PM, Adeeb Shihadeh ***@***.***> wrote:




@koch-cf I just read your support ticket, and looks like we may just need to update the fan stall thresholds for your device. Your device is in a weird zone right now because it spins just enough to not trigger the malfunction alert and right before the stall check (see panda/board/drivers/fan.h). Both of those should be addressed in this PR, can you give it a try?

adeebshihadeh:
No problem, I'll check it out when it gets back here.

robbederks (approved):

PR #27409, opened 2023-02-21
cabana: static link to opendbc

deanlee:
depends on https://github.com/commaai/opendbc/pull/786


adeebshihadeh:
opendbc side has been merged

PR #27415, opened 2023-02-21
Ford: driver torque limiting

sshane:
Not an explicit driver torque limit, more similar to certain torque cars where we ensure the requested torque is near the current torque, which gives us two things:
- No wind up if EPS is faulty and not applying much torque
- Handles driver overriding


selfdrive/car/ford/values.py
```diff
@@ -27,10 +27,12 @@ class CarControllerParams:
...
-  # TODO: unify field names used by curvature and angle control cars
-  # ~2 m/s^3 up, ~-3 m/s^3 down
-  ANGLE_RATE_LIMIT_UP = AngleRateLimit(speed_bp=[5, 15, 25], angle_v=[0.004, 0.00044, 0.00016])
-  ANGLE_RATE_LIMIT_DOWN = AngleRateLimit(speed_bp=[5, 15, 25], angle_v=[0.006, 0.00066, 0.00024])
+  # The curvature signal is limited to 0.003 to 0.009 m^-1/sec by the EPS depending on speed and direction
+  # Limit to ~2 m/s^3 up, ~3 m/s^3 down at 75 mph
+  # Worst case, the low speed limits will allow 4.3 m/s^3 up, 4.9 m/s^3 down at 75 mph
+  ANGLE_RATE_LIMIT_UP = AngleRateLimit(speed_bp=[5, 25], angle_v=[0.0002, 0.0001])
+  ANGLE_RATE_LIMIT_DOWN = AngleRateLimit(speed_bp=[5, 25], angle_v=[0.000225, 0.00015])
+  CURVATURE_ERROR = 0.002  # ~6 degrees at 10 m/s, ~10 degrees at 35 m/s
```
sshane, line 35: This is from https://github.com/commaai/openpilot/pull/27446, not exclusive to this PR

sshane:
We actually can't use `apply_dist_to_meas_limits`, as it assumes the val is a torque. When the val is positive, and the meas is negative, it only goes to a minimum of 0 + steer error. So in the following scenario, the limit the function would output would be massively far from the meas and cause wind up, since we're limiting in curvature space, not torque:

```python
apply_dist_to_meas_limits(val=1000, val_last=1000, val_meas=-1000, STEER_DELTA_UP=1000*2, STEER_DELTA_DOWN=1000*2, STEER_ERROR_MAX=200, STEER_MAX=2000)
200.0
```

PR #27420, opened 2023-02-22
Update values.py fw Toyota Prius Plug-In MY2022

HNR001:
Add missing FW for Toyota Prius Plug-In EUDM MY2022


sshane:
Can you add your dongle id? It will be in device settings

sshane:
Any updates?

HNR001:
The dongle id is: dfc4b737182810c0

sshane:
sorry for the delay!

PR #27435, opened 2023-02-23
cabana: add support for multiple DBC files

pd0wm:
Add option to open separate DBC file  for each bus. Still possible to open a DBC that applies to all busses.

Technically you should also be able to add multiple DBC files for the same bus (in cases where each node has it's own DBC), but will consider that out of scope for this PR.

Tried to keep the changes as minimal as possible, but it still got quite big.

Currently supported:
- Opening a .dbc file that applies to all busses
- Opening a .dbc file that applies to one ore more specific busses
- Having multiple .dbc files open at the same time
- Opening a file from opendbc that applies to all busses
- Opening a recent file that applies to all busses

Not (yet) supported, todo in separate PRs:
 - Save as with more than one file open
 - Save to clipboard with more than one file open
 - Creating a new DBC file for a specific bus only
 - Closing a DBC file
 - Opening a DBC from opendbc or recent files that applies to a specific bus (instead of all)
 - An advanced DBC manager dialog where you can do things like change bus assignments or load multiple DBC files for the same bus.
 - Add radar bus DBC to `loadDBCFromFingerprint` to load both files automatically.

image

image

### DIFF
--- a/tools/cabana/SConscript
+++ b/tools/cabana/SConscript
@@ -28,7 +28,7 @@ cabana_env.Depends(assets, Glob('/assets/*', exclude=[assets, assets_src, "asset
 
 prev_moc_path = cabana_env['QT_MOCHPREFIX']
 cabana_env['QT_MOCHPREFIX'] = os.path.dirname(prev_moc_path) + '/cabana/moc_'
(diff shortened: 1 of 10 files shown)


pd0wm:
@adeebshihadeh @deanlee This PR is ready for review!

After it's merged I'll start working on the features that need this PR as a prerequisite.

deanlee:
looks like missing indentation at https://github.com/commaai/openpilot/pull/27435/files#diff-f15a7381948cf7bab301af3b25f4e044c552d509b9de78b404df01136efe3443R44
can't comment in file view now, looks like there are some issues with github.

PR #27412, opened 2023-02-21
Honda HRV Non US model - Bosch Radarless

Ktrtkd:
Adds Brazilian HRV/Vezel 2023 Model - Bosch Radarless
[https://saladeimprensa.honda.com.br/releases/honda-apresenta-o-new-hr-v-ao-mercado-brasileiro](url)
Might be simmilar to other markets (Thailand, Philiphines and so), Based on Civic 2022
Verification: Fingerprint

Test Route: 147613502316e718|2023-02-21--18-09-57

PR #27438, opened 2023-02-23
port Brazilian Honda HR-V 2023

AlexandreSato:
i'm helping Alex Takeda in #27412

**Checklist**
- [x] route with openpilot: `147613502316e718|2023-02-24--13-03-54`
- [ ] lateral tunning


RELEASES.md
```diff
@@ -1,6 +1,7 @@
 Version 0.9.2 (2023-03-XX)
 ========================
 * Draw MPC path instead of model predicted path, this is a more accurate representation of what the car will do.
+* Honda HR-V Vezel 2023 non US support thanks to Alex Takeda!
```
sshane, line 4: I would imagine the US version will be similar, so let's remove the non-US for now unless we know that they differ. You can treat this car as if location doesn't matter (rename in docs, platform, etc)

sshane:
Which PR should we keep open?

PR #27446, opened 2023-02-24
Ford: simple rate limits

sshane:
If you actually look at the data, the EPS is imposing a rate limit in curvature, rather than lateral acceleration. So the assumptions that we could achieve 2 m/s^3 at low speeds were totally wrong (with only curvature as a command).

I spot checked some high speed injection test routes, and low speed normal driving routes, and concluded that the curvature rate limit (across 1 second) for up is anywhere between 0.0015 to 0.0035, depending on speed. Definitely not different by an order of magnitude that you'd see if the EPS jerk limits were implemented as lateral acceleration

Injection test route: 54827bf84c38b14f|2023-02-22--15-16-51


sshane:
From the injection test route above:

```python
Samples: 14 injection events
Mean curvature rate (1s): 0.0030784814249013187 1/m
Std curvature rate (1s): 0.0010435255384705557 1/m
```

- You can obviously see that the curvature rates stay the same across speeds

Screenshot from 2023-02-24 00-43-51

- While the lateral jerk increases ~linearly with speed (should be speed^2)

Screenshot from 2023-02-24 00-43-58

sshane:
Screenshot from 2023-02-24 15-32-13
Screenshot from 2023-02-24 15-32-18
Screenshot from 2023-02-24 15-33-47
Screenshot from 2023-02-24 15-34-07

sshane:
Jerk data:

Screenshot from 2023-02-24 15-06-49
Screenshot from 2023-02-24 15-07-00

sshane:
This doesn't regress steering that much at all, and no reason to keep open, so merging. We can revisit later once the port is merged.

PR #27452, opened 2023-02-26
update to python 3.11.4

adeebshihadeh:


incognitojam:
https://www.phoronix.com/review/python-311-performance#:~:text=Python%203.11%20performance%20is%20showing%20huge%20uplift

incognitojam:
https://github.com/PyAV-Org/PyAV/blob/main/CHANGELOG.rst
looks like `pyav` only added support for python 3.11 in 10.0.0, but they also dropped support for ffmpeg < 4.3 in that release (my workstation uses that). think it affects tools/camerastream.

incognitojam:
`onnx` added support for 3.11 in `1.13`, required protobuf>=3.20.2 so bumped too 3.20.3 (final 3.x release)

incognitojam:
hm... this needs new onnxruntime and opencv-python

.python-version
```diff
@@ -1 +1 @@
-3.8.10
+3.11.2
```
adeebshihadeh, line 1: looks like 3.11.4 is out

selfdrive/athena/athenad.py
```diff
@@ -790,8 +790,6 @@ def main():
     except (ConnectionError, TimeoutError, WebSocketException):
       conn_retries += 1
       params.remove("LastAthenaPingTime")
-    except socket.timeout:
-      params.remove("LastAthenaPingTime")
```
adeebshihadeh, line 794: why was this removed?

pyproject.toml
```diff
@@ -10,13 +10,13 @@ documentation = "https://docs.comma.ai"
 
 
 [tool.poetry.dependencies]
-python = "~3.8"
+python = "~3.11"
```
adeebshihadeh, line 13: let's split out the package updates into another PR

selfdrive/athena/athenad.py
maxime-desroches: https://docs.python.org/3/whatsnew/3.10.html#socket

socket.timeout and TimeoutError are now the same exception and pylint throws https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/duplicate-except.html

adeebshihadeh: we'll have to silent it for now, since C3's will still be running 3.8 for a while after this is merged. can you revert this, silence the pylint warning, and add a good comment explaining this?

adeebshihadeh:
FYI, the poetry lock is out of date on this branch. Probably want to rebase on master.

adeebshihadeh:
@maxime-desroches can you rebase on master and clean this up? I'd like to merge this today

PR #27457, opened 2023-02-28
ui: longer screen timeout for driver view

### LINKED ISSUE #27454: UI: longer screen timeout for driver view

deanlee:
resolve https://github.com/commaai/openpilot/issues/27454


selfdrive/ui/qt/offroad/driverview.cc
```diff
@@ -32,11 +32,11 @@ DriverViewScene::DriverViewScene(QWidget* parent) : sm({"driverStateV2"}), QWidg
 
 void DriverViewScene::showEvent(QShowEvent* event) {
   frame_updated = false;
-  params.putBool("IsDriverViewEnabled", true);
+  uiState()->setDriverViewEnabled(true);
```
adeebshihadeh, line 35: the interactive timeout also needs to be reset

adeebshihadeh (changes requested):

selfdrive/ui/ui.cc
```diff
@@ -300,7 +307,8 @@ void Device::setAwake(bool on) {
 }
 
 void Device::resetInteractiveTimout() {
```
adeebshihadeh, line 309: this is a net +12 lines for a super simple change. I think it would be cleaner to be able to call `Device::resetInteractiveTimout` from the driver view code and pass it a timeout. this way it's more generic and the Device class doesn't need to know anything about driver view

adeebshihadeh (changes requested):

deanlee:
how about the new version ? it is much simpler than the previous

deanlee:
@adeebshihadeh do you think this is okay?

deanlee:
How about adding a close button in the driverview? This way the screen can be clicked to refresh the timeout and we don't need to specify a fixed timeout for it

selfdrive/ui/ui.h
```diff
@@ -206,7 +206,7 @@ class Device : public QObject {
   void interactiveTimeout();
 
 public slots:
-  void resetInteractiveTimeout();
+  void resetInteractiveTimeout(int timeout = 0);
```
sshane, line 209:
```suggestion
  void resetInteractiveTimeout(int timeout = -1);
```
?

selfdrive/ui/ui.cc
```diff
@@ -294,8 +294,11 @@ void Device::setAwake(bool on) {
   }
 }
 
-void Device::resetInteractiveTimeout() {
-  interactive_timeout = (ignition_on ? 10 : 30) * UI_FREQ;
+void Device::resetInteractiveTimeout(int timeout) {
+   if (timeout == 0) {
```
sshane, line 298:
```suggestion
   if (timeout == -1) {
```

selfdrive/ui/ui.cc
```diff
@@ -294,8 +294,11 @@ void Device::setAwake(bool on) {
   }
 }
 
-void Device::resetInteractiveTimeout() {
-  interactive_timeout = (ignition_on ? 10 : 30) * UI_FREQ;
+void Device::resetInteractiveTimeout(int timeout) {
+   if (timeout == -1) {
```
sshane, line 298:
```suggestion

  if (timeout == -1) {
```

selfdrive/ui/ui.cc
```diff
@@ -294,8 +294,11 @@ void Device::setAwake(bool on) {
...
+    timeout = (ignition_on ? 10 : 30);
+  }
+  interactive_timeout = timeout * UI_FREQ;
```
sshane, line 301:
```suggestion
void Device::resetInteractiveTimeout(int timeout) {
  if (timeout == -1) {
    timeout = (ignition_on ? 10 : 30);
  }
  interactive_timeout = timeout * UI_FREQ;
```

selfdrive/ui/qt/window.cc
```diff
@@ -21,6 +21,7 @@ MainWindow::MainWindow(QWidget *parent) : QWidget(parent) {
     main_layout->setCurrentWidget(onboardingWindow);
   });
   QObject::connect(settingsWindow, &SettingsWindow::showDriverView, [=] {
+    device()->resetInteractiveTimeout(60);
```
sshane, line 24: can we set back to normal timeout when exiting by tapping?

sshane:
@deanlee this looks good, but I think after https://github.com/commaai/openpilot/pull/28656, when it does time out, the driver view stays up and tapping once doesn't close it. If there's a signal that gets emitted when the timeout happens, we should close the driver view then.

PR #27459, opened 2023-02-28
ui: draw model on matching frame

### LINKED ISSUE #27353: UI: draw model on matching frame

deanlee:
resolve https://github.com/commaai/openpilot/issues/27353 

| Before  | After |
| Screenshot from 2023-02-28 20-09-59  | Screenshot from 2023-02-28 20-10-26  |

video:

1. before:

[Kazam_screencast_00086.webm](https://user-images.githubusercontent.com/27770/221849737-d0e9c0cc-3d02-4446-af4d-6a07f761a774.webm)

2. after:

[Kazam_screencast_00087.webm](https://user-images.githubusercontent.com/27770/221849647-68243469-f494-43cf-b3cf-88de14fe2756.webm)

@adeebshihadeh feel free to close it if you are already working on this.


selfdrive/ui/qt/widgets/cameraview.h
```diff
@@ -21,9 +21,6 @@
 #include "system/camerad/cameras/camera_common.h"
 #include "selfdrive/ui/ui.h"
 
-const int FRAME_BUFFER_SIZE = 5;
-static_assert(FRAME_BUFFER_SIZE <= YUV_BUFFER_COUNT);
```
adeebshihadeh, line 25: this was done to prevent drawing really old frames. i'd still like to keep this

sshane:
@deanlee you didn't notice that the UI's loop can get out of sync with the model causing many frame skips in a row for a period of time?

deanlee:
> I believe the exact problem was we'd sometimes get a new camera frame before the model published the corresponding frame id, so it would skip to the default (end or start of the frame buffer), then skip back once we were getting model frame ids for the current camera frame.

no, it will not skip to the end or start of the frame buffer by default, it will draw the frame get from last `sm["uiPlan"].getUiPlan().getFrameId()`.

paintGL will find `sm["uiPlan"].getUiPlan().getFrameId()`  in cache and paint it. it only skip to the last frame read from camera(end of the frame buffer) when the `sm["uiPlan"].getUiPlan().getFrameId()` is not already present in the cache.  and this may only occur when the camera is initially started.

  
```

auto frame_it = frames.rbegin();
  if (draw_frame_id > 0) {
    auto it = std::find_if(frames.rbegin(), frames.rend(), [=](auto &f) { return f.first == draw_frame_id; });
    if (it != frames.rend()) frame_it = it;
  }
  auto [frame_id, frame] = *frame_it;
```

sshane:
Run replay with qlogs and you can see it skip and hang every time it updates the path, though I'm not sure if that translates to on-device issues.

deanlee:
> Run replay with qlogs and you can see it skip and hang every time it updates the path, though I'm not sure if that translates to on-device issues.

Is it possible that this is caused by the low frequency of the video frames pushed from qlogs:

service.h:
`{ "roadEncodeIdx", 8015, false, 20, 1 }`

deanlee:
> Run replay with qlogs and you can see it skip and hang every time it updates the path, though I'm not sure if that translates to on-device issues.

Is it possible that this issue is caused by the frequency settings in the qlogs?

service.h:
```
{ "roadEncodeIdx", 8015, false, 20, 1 },
{ "uiPlan", 8057, true, 20, 40 },
```

sshane:
All of the `roadEncodeIdx` should be there to accompany the frames, but the problem is that uiPlan times out from ui's perspective. It gets an updated uiPlan and draws the camera frame for it, but then it doesn't get any more updates so it keeps drawing that frame until we don't have any more frames for that frame id, then it skips to the front (or back) of the buffer.

deanlee:
I'm concerned that this could be a case of back-frame playback. it could be more bad than duplicate frames.

```
ui_plan: 1      4       6
ui     : 1 2 3 4 5 4 5 6 7 6 7

```

sshane:
yeah exactly, except it would be more like this, right?

```
# 4 frame buffer
ui_plan: 1...      4         6
ui     : 1 1 1 1 5 4 4 4 4 7 6 6...
```

deanlee:
Currently, my understanding is (which may be wrong), that this issue may only occur when replaying qlogs.It should not happen when playing normal logs or on a c3 device, unless controlsd crashes.If that's the case,  then skipping synchronization during Qlog replay may be a simpler solution.(looks like we would need to use more complex code and cache at least 40 frames of VIPC buffer in both the replay and CameraView, in order to sync with uiPlan and prevent duplicate frame warning when playing qlogs.)

sshane:
I think this is what can happen on device, leading to duplicate and skipped frames being drawn fairly often, though it's been a while since I've looked at it.
```
# 4 frame buffer
ui_plan: 1   2 4 5
ui     : 1 1 2 4 5
```

adeebshihadeh:
@deanlee along the lines of this, want to try to make the UI poll and update on recv of `modelV2` instead of a new frame over vipc? we want to do this to improve the GPU scheduling (camera debayer -> modeld -> ui); POC in #28363

sshane:
We might want to poll on uiPlan now that we draw the path using it

deanlee:
> @deanlee along the lines of this, want to try to make the UI poll and update on recv of `modelV2` instead of a new frame over vipc? we want to do this to improve the GPU scheduling (camera debayer -> modeld -> ui); POC in #28363

The current implementation of msgq_poll cannot be truly used for synchronously messages in this case. It simply sleeps for a fixed time (timeout) and then checks if there are new messages. 
submitted a pr https://github.com/commaai/cereal/pull/458 to make msgq_poll work properly like zqm_poll.

deanlee:
@adeebshihadeh : poll on the `uiPlan` could potentially block the main thread. so I have used a signal `uiPlanUpdated` to synchronize the uiplan events with the drawing of `AnnotatedCameraWidget`

deanlee:
> @deanlee along the lines of this, want to try to make the UI poll and update on recv of `modelV2` instead of a new frame over vipc? we want to do this to improve the GPU scheduling (camera debayer -> modeld -> ui); POC in #28363

submitted a new pr https://github.com/commaai/openpilot/pull/28370 for it.

sshane:
@deanlee are there any plans to merge this soon? If not we can close until you're ready.

deanlee:
closed, need more time to come up with a better solution.

PR #27462, opened 2023-02-28
Add China HONDA ACCORD HYBRID 2018 finger

javenzy:
Add finger to support China HONDA ACCORD HYBRID 2018,use ACCORDH model.

Have tested on my car,it can work fine.


sshane:
Thanks for the PR!

PR #27476, opened 2023-03-02
Adding the 2023 US HR-V

galegozi:
Please note that the pull request is not complete; however, I am opening it for feedback. Please do not close this, as I intend to merge this PR once it is complete & ready to go.

This PR adds the 2023 US HR-V. This vehicle is different from other HR-Vs (such as the Brazillian), and is _significantly_ different from US HR-Vs in previous years. In some markets, this is also known as the ZR-V.

**Checklist**
- [ ] route with openpilot:


Ktrtkd:
Basically this branch is just a test for HR-V USA Model (ZR-V). It's still under Civic 2022 for testing purposes. Might need adjustments.

RELEASES.md
```diff
@@ -18,6 +18,7 @@ Version 0.9.1 (2023-02-23)
 * Toyota C-HR 2021 support thanks to eFiniLan!
 * Toyota C-HR Hybrid 2022 support thanks to Korben00!
 * Volkswagen Crafter and MAN TGE 2017-23 support thanks to jyoung8607!
+* Added Honda HR-Vezel non US
```
sshane, line 21: You don't need to call it non-US, especially since your car is North American. I checked and it says the ZR-V is released as the 2023 HR-V: https://en.wikipedia.org/wiki/Honda_ZR-V

sshane:
What is your dongle id?

RELEASES.md
galegozi: The reason it is stated as such is because it references the non us 2023 hr-v (the us model is different from the rest of the world).
I think for the code, the 2023 us hrv should be listed as a 2023 zrv, where we specify somewhere the naming convention...

sshane:
Added your car here: https://github.com/commaai/openpilot/pull/27731

PR #27485, opened 2023-03-03
UI: clear old frames on recv timeout

deanlee:


adeebshihadeh:
What's the purpose of this? I think the logic in the camera widget is more generic.

deanlee:
I may be overdoing it. just think it's not good to check QWidget::isVisible in a different thread.
```
if (!isVisible()) {
  vipc_client->connected = false;
}
```

adeebshihadeh:
Sure, timeout may be cleaner, but I'd then be worried about a race condition. Unless, there's specifically a bug that this addresses or it's clearly an improvement, then it's hard to merge this PR.

PR #27486, opened 2023-03-03
Toyota: RADAR_ACC_TSS-P cars radar support

Smartype:
Please review.
This depends on: https://github.com/commaai/cereal/pull/414, I will update the submodule when it is accepted.
Or you can accept this one: https://github.com/commaai/openpilot/pull/27487, which has no depends on cereal.


selfdrive/car/toyota/interface.py
```diff
@@ -195,11 +195,13 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long):
 
     ret.enableBsm = 0x3F6 in fingerprint[0] and candidate in TSS2_CAR
     # Detect smartDSU, which intercepts ACC_CMD from the DSU allowing openpilot to send it
-    smartDsu = 0x2FF in fingerprint[0]
+    smartDsu = 0x2FF in fingerprint[0] or 0x2AA in fingerprint[0]
```
sshane, line 198: I understand this is to enable longitudinal support on radar-controlled TSS-P cars. Instead of working on the radar parsing first (which isn't required for OP longitudinal), maybe it's a better idea to just open a PR to add longitudinal support. Would it just be this one line here?

Smartype: Thanks for your suggestion. I have opened another PR to detect the device as smartDsu.
https://github.com/commaai/openpilot/pull/28387

Btw, is it possible to merge the radar parser? This radar is pretty reliable.

PR #27487, opened 2023-03-03
Toyota: radar_acc_tssp_car radar parser (v_ego)

Smartype:
Please review.
Another PR which requires less file changed. https://github.com/commaai/openpilot/pull/27486

### DIFF
--- a/release/files_common
+++ b/release/files_common
@@ -569,6 +569,7 @@ opendbc/toyota_new_mc_pt_generated.dbc
 opendbc/toyota_nodsu_pt_generated.dbc
 opendbc/toyota_adas.dbc
 opendbc/toyota_tss2_adas.dbc
+opendbc/toyota_radar_dsu_tssp.dbc
 
 opendbc/vw_golf_mk4.dbc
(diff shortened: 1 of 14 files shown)


Smartype:
You can accept this one, or accept the cereal PR. If you accept the cereal PR. I will work on another radar ACC PR.

cereal PR: https://github.com/commaai/cereal/pull/414
another radar ACC PR: https://github.com/commaai/openpilot/pull/27486

cydia2020:
Can this be merged? I've this in my fork and folks are reporting it works well, it also doesn't affect the functionality of other platforms.

chadgauth:
When reviewing the original I thought the vRel from radars providing absolute velocity should be calculated in the radar interface, but I can put my finger on exactly the best way to implement this. I could be available to clean this up some perhaps tomorrow.

williefeb19:
I can confirm that cydia2020 fork does work well. I have been hoping for OP support for Smartype's firmware/hardware for a while.

sshane:
Closing this in favor of https://github.com/commaai/openpilot/pull/27487 since they're both very similar.

PR #27489, opened 2023-03-03
VW MQB: Add FW for 2022 Skoda Kodiaq 1.5TSI

Saber422:
`2ca49a03c8084514`

carVin': 'TMBJB9NS3N8051692'


Saber422:
route name : 0bbe367c98fa1538|2023-03-03--13-54-43

sshane:
@jyoung8607 know of any differences with this car?

Saber422:
With the addition of capacitive steering wheel, the same 2Q0 camera and Continental radar, a false warning of ‘cruise faults’ appears when the turn on ignition before engine starts. Other conditions are similar to the old kodiaqs. The owner has already uploaded and shared it with you. Thanks for your time to review.

jyoung8607:
> @jyoung8607 know of any differences with this car?

It should be fine. I already expanded the supported model-year range in #27532.

> With the addition of capacitive steering wheel, the same 2Q0 camera and Continental radar, a false warning of ‘cruise faults’ appears when turn on ignition before engine starts. Other conditions are similar to the old kodiaqs.

I looked at the route you supplied earlier in Discord. The drivetrain coordinator does indeed indicate a temp-fault condition before you actually start the vehicle. Other vehicles handle this with a separate "initializing" state. It seems to be cosmetic only. At some point, I'll see if there's a good way to suppress this until after the vehicle is actually started.

Separately, I'd like to use the capacitive wheel touch state as a DM reset criteria, but that would be an entirely separate PR and we don't need to worry about it for now. It's not unique to Kodiaq, there are lots of other newer VW MQB with that wheel and we already support them.

PR #27493, opened 2023-03-03
Radard: just get relative speed from model

haraschax:


jyoung8607:
Yesterday I did a little bit of city driving and a lot of highway driving on this code. I haven't driven with 

I was really impressed with how it felt around-town, and in stop-and-go traffic. Apparently the static aLeadTau can work. That's subjective, I haven't looked at the data for the around-town stuff.

On the highway, it finds mostly stable but very long following distances. For this, I did look at the data. At least on the highway, it looks like the model has a pretty substantial negative bias on the lead speed, to the tune of -1 to -2 m/s.

Looking at a long stretch where I followed the same car, the model predicts the lead going slower than the ego vehicle for the entire time period. Obviously this isn't possible. It makes the planner want to slow down, when in reality we're not getting any closer to the lead, hence very long follow distance. The vEgo-to-model-speed compensation actually makes this mildly worse.

@HaraldSchafer, may I ask why the model outputs absolute lead speeds rather than relative? It seems to do a good-enough (albeit noisy) job inferring lead distance, from which relative speed and then acceleration and absolute speed could be derived. We shouldn't even need a fantastically accurate lead speed, we just need to know if the lead is moving toward us or away from us, and a very rough idea of how fast. That's how humans handle it.

`3cfdec54aa035f3f|2023-03-07--18-04-43`

image

morrislee:
After using the new 2023-driving model, the difference between calibrated to non-calibrated is almost hard to tell. City driving is fine, highway causes oscillation behind lead like rubber banding effect. I ran the plotjuggler on non-calibrated data on my tall vehicle, looks like the vlead data is fairly noisy, but the model leadv3 is pretty smooth. Is the vlead data contaminated somehow?

image

morrislee:
I have used the scale difference from vego to temporalPose/trans/0 and applied to leadsV3/0/v/0 and came up with more consistent result here. Using stock ACC to follow a lead on cruise control shows a very steady negative bias, maybe about 3 to 5% slower than actual speed on top of the noisy leadsV3

image

jyoung8607:
> I ran the plotjuggler on non-calibrated data on my tall vehicle, looks like the vlead data is fairly noisy, but the model leadv3 is pretty smooth. Is the vlead data contaminated somehow?

You're probably graphing with just qlogs. They are decimated to save space and bandwidth. With the qlogs, you're seeing radarState at 5Hz and modelV2 at 0.5Hz, and that's what's giving you the appearance of discontinuity. If you upload rlogs, you can see both radarState and modelV2 at 20Hz.

morrislee:
Are we no longer needing to calibrate the lead speed?

haraschax:
This should have the same effect, but is cleaner.

PR #27496, opened 2023-03-04
test_onroad: test uncompressed qlog size

sshane:


selfdrive/test/test_onroad.py
```diff
@@ -187,6 +188,10 @@ def test_cloudlog_size(self):
     big_logs = [f for f, n in cnt.most_common(3) if n / sum(cnt.values()) > 30.]
     self.assertEqual(len(big_logs), 0, f"Log spam: {big_logs}")
 
+  def test_qlog_size_sane(self):
```
adeebshihadeh, line 191: we want to test the compressed size. we also don't just want a single limit, otherwise it's going to be hard to make this not flaky and catch regressions quickly

adeebshihadeh (changes requested):

PR #27498, opened 2023-03-04
cabana:  use a segment tree to solve range queries in O(log n) time complexity

deanlee:
issue: 
function `updateAxisY`(O(N)) is heavily  called in ui thread,  this could lead to lagging when there are multiple charts with a large number of events. 

resolve: 
using a segment tree to improve the performance of range query for min and max values.

[Kazam_screencast_00088.webm](https://user-images.githubusercontent.com/27770/222913871-fcb97d0f-9dcb-46b0-b392-a1923c935a38.webm)

### DIFF
--- a/tools/cabana/chartswidget.cc
+++ b/tools/cabana/chartswidget.cc
@@ -562,6 +562,9 @@ void ChartView::updateSeries(const Signal *sig, const std::vector<Event *> *even
       if (events->size()) {
         s.last_value_mono_time = events->back()->mono_time;
       }
+      if (!can->liveStreaming()) {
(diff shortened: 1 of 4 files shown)


s-nandi:
Just curious, would it make sense to use a sparse table instead? Sparse tables provide O(1) queries and are easier to grok imo, and while they don't support point updates, it looks like that isn't necessary for this PR

deanlee:
@s-nandi : we may need to use segment tree in live mode later. therefore, supporting updates is necessary. The  O(log n) is good enough for the current dataset.

PR #27500, opened 2023-03-05
Subaru: Add 2019 Forester FPv2

martinl:
Route: `23f027ded69bd8be|2023-03-05--18-07-47`


selfdrive/car/subaru/values.py
```diff
@@ -273,6 +273,7 @@ class SubaruCarInfo(CarInfo):
       b'\xa3 \031\024\000',
       b'\xa3  \x14\x01',
       b'\xf1\x00\xbb\r\x05',
+      b'\xa3 \x18&\x00',
     ],
     (Ecu.eps, 0x746, None): [
       b'\x8d\xc0\x04\x00',
```
sshane, line 279:
```suggestion
      b'\x8d\xc0\x04\x00',
      b'\x8d\xc0\x00\x00',
```
think you're missing this one too

PR #27503, opened 2023-03-06
Tesla: half the DAS_steeringControl message rate

robbederks:
Seems like this actually only gets sent at 50Hz by the car, not 100Hz.


sshane:
any blockers?

robbederks:
> any blockers?

Yes, haven't been able to test this PR yet, autopilot is disabled again on my car. Will merge ASAP

PR #27508, opened 2023-03-06
added missing fingerprint jeep 2020

Miniz199922:
1239: 8, 1256: 8, 1879: 8


Miniz199922:
Found another one 1897:8

PR #27515, opened 2023-03-07
loggerd: fixed the message queue growing infinitely when the incoming encoders less than max_waiting.

### LINKED ISSUE #27512: PC: loggerd memory leak
### Describe the bug

running loggerd the amount of memory used keeps going up until it runs out

### OS Version

Ubuntu 20.04

_No response_

### Additional info

_No response_

deanlee:
Issue: `offset_segment_num > s->rotate_segment` will be always true if the incoming encoder types less than the `max_waiting`:

https://github.com/commaai/openpilot/blob/ebd25b8a7de29542be7c6a213c578bc363999857/selfdrive/loggerd/loggerd.cc#L155-L165

Set the max_waiting according to the incoming encoders to fix this issue, resolve #27512

### DIFF
--- a/selfdrive/loggerd/loggerd.cc
+++ b/selfdrive/loggerd/loggerd.cc
@@ -25,7 +25,7 @@ void logger_rotate(LoggerdState *s) {
 
 void rotate_if_needed(LoggerdState *s) {
   // all encoders ready, trigger rotation
-  bool all_ready = s->ready_to_rotate == s->max_waiting;
+  bool all_ready = s->max_waiting > 0 && s->ready_to_rotate == s->max_waiting;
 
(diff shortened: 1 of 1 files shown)


bleufa:
idk if this fixes the issue, I can still see the memory usage increasing in top

deanlee:
That's weird. I ran it for an hour and it's lgtm.

deanlee:
Probably a small memory leak caused by another issue

bleufa:
i'm not running encoderd at all, just in offroad I can see the memory usage increasing after running for a few minutes. also see possibly lost bytes in valgrind

bleufa:
can you see the memory usage increasing?

deanlee:
Not yet. I can't see any place where memory leaks may occur If the encoderd is not running.

PR #27518, opened 2023-03-07
 VW MQB: Add FW for 2015 VOLKSWAGEN GOLF 7TH GEN

Saber422:


jyoung8607:
Thank you for the data! comma will need the dongle ID/route in order to verify and merge.

sshane:
@Saber422 any update? You just need to post what it says for dongle id in Device settings.

Saber422:
I forgot to save the route file for this car, so now I have no data in my cabana. The car owner had installed window tint on the windshild that blocked GPS signals, resulting in a trip with no GPS information. Therefore, I suggested that he should not pair his account before replacing the window tint to avoid his account being banned by server. After I meet with the car owner, I will provide the Route ID of this car.

Saber422:
I have time to meet this car, and setup my C3 on this car.(I can't get GPS since his windows tint is with metal)
@sshane route name ：0bbe367c98fa1538|2023-04-02--09-02-57--0

PR #27520, opened 2023-03-07
VW MQB: Škoda Fabia Mk4

jyoung8607:
Adding support for the fourth-generation Škoda Fabia.

- [x] route with openpilot: `0bbe367c98fa1538|2023-03-04--17-46-11`
- [x] added to CARS
- [x] added to RELEASES

Thanks once again to @Saber422 and the fantastic work he does with his local openpilot community!


jyoung8607:
Note: I added the MQB_A0 footnote warning of possible problems with harness availability, since at least some MY2022+ MQB-A0 seem to have gone toward a combined BCM-CAN gateway that we don't have a harness for, but Saber422 was able to use a normal gateway harness on the test vehicle which is MY2022 by VIN.

PR #27549, opened 2023-03-09
Raised the Ram DT max steer to 350

vincentw56:
Description [ Raise the torque limit to 350 for Ram DT. This allows us to go from 2.0m/s2 of lateral accel to 2.5m/s2 of lateral accel. This PR replaces the older PR#27082]

Verification [We have been running this for over a year at these values. Anything over 350 will cause an eps fault.]


sshane:
Thanks, but we have https://github.com/commaai/openpilot/pull/27082 already open

PR #27540, opened 2023-03-09
cabana: DBCManager only expose methods/signals through MessageId

pd0wm:
Prerequisite for https://github.com/commaai/openpilot/pull/27435

Emits message changed signals separately for each source to prepare for multiple DBC files.

PR #27552, opened 2023-03-10
cabana: refactor Msg and Signal into dbc.cc

pd0wm:
Added explicit `cabana::Signal` and `cabana::Msg` types instead of relying on include hacks to prevent overlap with types from `common_dbc.h`.

Let's hold off on merging this PR until https://github.com/commaai/openpilot/pull/27540 is merged.


deanlee:
Great. that's actually what I wanted to do before. the only reason I used include hacks was because I worried that changing the code too much would make it difficult to review.

adeebshihadeh:
I like the namespacing!

PR #27554, opened 2023-03-10
cabana: use ID as NEW_MSG_ suffix

pd0wm:
Eg. `NEW_MSG_A12` for id 0xa12


tools/cabana/signaledit.cc
```diff
@@ -236,7 +236,8 @@ bool SignalModel::saveSignal(const Signal *origin_s, Signal &s) {
 void SignalModel::addSignal(int start_bit, int size, bool little_endian) {
   auto msg = dbc()->msg(msg_id);
   for (int i = 1; !msg; ++i) {
```
deanlee, line 238: I think we can use i =0 && if (i > 0) for now

PR #27555, opened 2023-03-10
cabana: extend max chart range to 30 minutes

pd0wm:
Change range slider to a log scale, otherwise selecting smaller ranges is almost impossible.


tools/cabana/chartswidget.cc
```diff
@@ -163,7 +163,7 @@ void ChartsWidget::updateState() {
 }
 
 void ChartsWidget::setMaxChartRange(int value) {
-  max_chart_range = settings.chart_range = value;
+  max_chart_range = settings.chart_range = range_slider->value();
```
pd0wm, line 166: Unsure on how to make `QSlider::valueChanged` go through the new mapping.

PR #27556, opened 2023-03-10
cabana: show signal values with consistent number of digits

pd0wm:
Show consistent number of decimal places in signal and history view to prevent numbers moving around. Make historyview align right.

Merge after https://github.com/commaai/openpilot/pull/27552 to prevent conflicts.

Before:
Screenshot 2023-03-10 at 11 39 03

After:
Screenshot 2023-03-10 at 11 37 45

Before:
Screenshot 2023-03-10 at 11 39 22
After:
Screenshot 2023-03-10 at 11 37 34

### DIFF
--- a/tools/cabana/dbc.cc
+++ b/tools/cabana/dbc.cc
@@ -1,4 +1,5 @@
 #include "tools/cabana/dbc.h"
+#include "tools/cabana/util.h"
 
 uint qHash(const MessageId &item) {
   return qHash(item.source) ^ qHash(item.address);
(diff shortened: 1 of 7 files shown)


deanlee:
How about moving it into a member function of  class Signal ? Furthermore, we could consider caching the precision to a member variable and updating it when necessary.

pd0wm:
The functions fetching the precision are all const, so can't update the cached value. I'll refactor to make sure the precision is updated when the values are set/changed.

adeebshihadeh:
> Merge after #27552 to prevent conflicts.

Looks like there's conflicts anyway :(

PR #27565, opened 2023-03-12
Add missing FW for 2020 Lexus NX Hybrid

klaw73:


sshane:
Can you provide the dongle ID these came from?

klaw73:
3a4c8750b7b461a0

PR #27590, opened 2023-03-15
Add missing FW for 2022(2023MY) Hyundai IONIQ 5

krvista:
Added Korean F/W ID for 2022 (2023MY) Hyundai IONIQ 5 4WD Korean specification.
And it works perfectly fine with mine.

Route ID : 801497025234177|2023-03-12--04-44-26


sshane:
@krvista that doesn't look like a valid dongle id. Can you post what you see in Device settings?

krvista:
image

PR #27591, opened 2023-03-15
LAIKAD: TTFF, and ephemeris statuses in logs

haraschax:


sshane:
@HaraldSchafer there's a process replay diff

PR #27602, opened 2023-03-16
VW MQB: Add FW for 2022 Škoda Kodiaq

Saber422:
@jyoung8607  I try to do this better. I will get other car someday when it's available.
This owner get a C3. And I setup my C3 on this car to get information to merge this FPV2.
The route name is 0bbe367c98fa1538|2023-03-16--11-41-19--2


sshane:
The routes are for a Polo, is this the right car/route?

jyoung8607:
The route `0bbe367c98fa1538|2023-03-16--11-41-19` he gave is indeed from a Kodiaq:
```
               'carVin': 'TMBLE9NS9N8057416',
               'fingerprintSource': 'fixed',
```

He does a lot of in-person demos for people on their cars with his personal C3. I guess forcing FP to Polo (his personal car) makes it simple to start up on other VW family vehicles he's demonstrating on. We can mostly get away with that on VW, the only problem is the wheelbase will be wrong with potential downstream effects on steer ratio learning and lateral control accuracy.

PR #27609, opened 2023-03-17
Toyota: add 2023 RAV4/RAV4 Hybrid as dashcam

sshane:
These cars will be dashcam, and if dashcam is changed in the code, then the car will not steer (we enforce an angle cmd message with no actuation).

WIP, pending clean up

FW from: https://github.com/commaai/openpilot/pull/27494 (hybrid) and eb4ea4c8dc99c8bb (ICE)


selfdrive/car/toyota/carcontroller.py
```diff
@@ -76,6 +76,11 @@ def update(self, CC, CS, now_nanos):
       apply_steer_req = 0
       self.steer_rate_counter = 0
 
+    # Never actuate with LKA on cars that only work LTA
```
sshane, line 79:
```suggestion
    # Never actuate with LKA on cars that only support LTA
```

selfdrive/car/toyota/tests/test_toyota.py
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env python3
+import unittest
+
+from selfdrive.car.toyota.values import TSS2_CAR, ANGLE_CONTROL_CAR
+
+
+class TestToyotaInterfaces(unittest.TestCase):
+  def test_angle_tss2_sets(self):
```
sshane, line 8:
```suggestion
  def test_angle_car_set(self):
```

PR #27613, opened 2023-03-17
controls: check steer control type

sshane:
Adds steer control type checking everywhere lateralTuning is checked, since if the steer control type is angle, lateralTuning can be anything, depending on how the car interfaces set it up (pid, angle, lqr).


sshane:
Handling in the car port for now

PR #27586, opened 2023-03-15
VW MLB: Audi A4/S4

jyoung8607:
**Checklist**

- [x] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [ ] route with openpilot:
- [ ] route with stock system:
- [x] car harness used (if comma doesn't sell it, put N/A): VW J533

**Additional Tasks**

- [ ] commaai/opendbc#793
- [ ] commaai/panda#1289
- [ ] #27624
- [ ] upcoming PR to handle EPS timer resets
- [ ] Clean up abstractions, don't love the current structure
- [x] Migrate to lateral accel torque controller
- [ ] Find a TSK main switch signal
- [ ] Solve issue with door-open signals
- [ ] Solve issue with gearshift position signals
- [x] Verify BSM blocks lane changes
- [ ] Verify blinker comfort vs instantaneous

PR #27624, opened 2023-03-19
VW: Prep for MLB, EPS timer cleanup

jyoung8607:
First of two PRs to prepare for #27586: Cleanup some of my old camelCase-spaghetti and use named constants. The ~2 minute case should never have been hit in the real world.

Add a soft disabling CarEvent as a backstop, so that if this logic ever fails to reset the timer, we alert the user to take action before steering actually stops working, and we also surface the problem in comma's MTBF reports.

I don't anticipate the CarEvent to get any real-world use on MQB, but [the next PR](https://github.com/jyoung8607/openpilot/compare/vw-mqb-eps-timer-cleanup...jyoung8607:openpilot:vw-pq-mlb-opp-reset?expand=1) will layer on some logic we've successfully tested to reset the EPS timebomb for PQ/MLB, and that will need to be watched more closely.

- [x] commaai/cereal#422.


selfdrive/car/volkswagen/carcontroller.py
```diff
@@ -38,36 +39,31 @@ def update(self, CC, CS, ext_bus, now_nanos):
       #   * Don't send > 3.00 Newton-meters torque
       #   * Don't send the same torque for > 6 seconds
       #   * Don't send uninterrupted steering for > 360 seconds
-      # One frame of HCA disabled is enough to reset the timer, without zeroing the
-      # torque value. Do that anytime we happen to have 0 torque, or failing that,
-      # when exceeding ~1/3 the 360 second timer.
+      # MQB racks reset the uninterrupted steering timer after a single frame
+      # of HCA disabled; this is done whenever output happens to be zero.
```
sshane, line 43: This sounds very similar to the workarounds for Toyota and Hyundai. Can you not also set the bit to 0 while torque != 0? Does it fault? Wind down torque and take a while to back up? Any other problems?

jyoung8607: I think I can get away with that on MQB, or at least that's what my three-year-old code/comments say. But I decided to discard that case because:

1) It adds complexity for a case I don't think we ever hit in the real world; we get resets at zero very frequently
2) We gain validation for that frequent zero assumption with the new Event, which we didn't have before
3) You've recently added LKA bit enabled validation in Panda, and I'd like to include VW in that
4) The upcoming PR to mitigate the PQ/MLB 6 minute timeout definitely needs to hold LKA *and* torque at zero

PR #25719, opened 2022-09-10
camerad: abstract cameras

deanlee:
resolve https://github.com/commaai/openpilot/issues/25718

PR #27627, opened 2023-03-20
camerad: move camera related parameters to CameraInfo

deanlee:
split from https://github.com/commaai/openpilot/pull/25719, move camera related parameters to `struct CameraInfo` and accessing them through `std::uniqueptr<const CameraInfo>`.
this will improve code readability and prevent accidental modifications to variables in `CameraInfo`. additionally, it will make it easier to encapsulate all camera-related code into their own classes in the future.


deanlee:
moved to https://github.com/commaai/openpilot/pull/30639

PR #27629, opened 2023-03-20
UI: easier nav destination entry

incognitojam:
- Add map panel to offroad home, replacing drive stats
- Replace map panel with prime ad when user is not subscribed
- Includes #28269

||New|Old|
|-|-|-|
|Unpaired|unpaired|TODO|
|No prime||
|With prime<br>(work in progress)|with prime|


actuallylemoncurd:
RIP comma points? 😭

james5294:
I have noticed that If you replace driver stats with the Navigation Pane and you then completely remove navigation from settings we lose the ability to quickly launch favorite navigation locations while on-road. The car has to be an off-road to see the new Navigation Pane. This leaves us only to input navigation addresses by phone (not as quick/not as easy). It would be nice to still somehow be able to access favorites/recents while on-road.

PR #27631, opened 2023-03-20
remove steerLimitTimer

adeebshihadeh:


sshane:
@adeebshihadeh do you want to merge or should we close? I will say that on my Camry with LTA, because the general response is quite delayed, I get alerts for deviations that I feel aren't warranted, since the angle will reach the desired angle right after the default 0.4 seconds. 0.8 seconds for that control scheme is much better.

PR #27640, opened 2023-03-21
automatically detect available camera streams

deanlee:
related pr: https://github.com/commaai/cereal/pull/424


selfdrive/controls/controlsd.py
```diff
@@ -86,8 +87,11 @@ def __init__(self, sm=None, pm=None, can_sock=None, CI=None):
       ignore = ['testJoystick']
       if SIMULATION:
         ignore += ['driverCameraState', 'managerState']
-      if self.params.get_bool('WideCameraOnly'):
-        ignore += ['roadCameraState']
+        # block until get the available streams
+        available_streams = VisionIpcClient.available_streams("camerad", block=True)
```
adeebshihadeh, line 91: hmm, this can't block too long or else controlsd won't run in case of camera malfunctions. maybe we can fit it in during the initializing state? https://github.com/commaai/openpilot/blob/a86730a2c601188c480285c45999efa00c3b3d4a/selfdrive/controls/controlsd.py#L434

adeebshihadeh (changes requested):

deanlee:
moved to set_initial_state.

tools/sim/bridge.py
```diff
@@ -562,6 +561,4 @@ def run(self, queue, retries=-1):
     p.join()
 
   finally:
-    # Try cleaning up the wide camera param
-    # in case users want to use replay after
-    Params().remove("WideCameraOnly")
+    pass
```
adeebshihadeh, line 564: the whole try block can be removed now

selfdrive/controls/controlsd.py
```diff
@@ -217,6 +214,15 @@ def set_initial_state(self):
       if any(ps.controlsAllowed for ps in self.sm['pandaStates']):
         self.state = State.enabled
 
+    if SIMULATION:
+      ignore = ['driverCameraState', 'managerState']
+      # block until get the available streams
+      available_streams = VisionIpcClient.available_streams("camerad", block=True)
```
adeebshihadeh, line 220: does this block until the vipc server is up? controlsd shouldn't hang if camerad doesn't start

selfdrive/modeld/modeld.cc
```diff
@@ -191,6 +188,18 @@ int main(int argc, char **argv) {
   model_init(&model, device_id, context);
   LOGW("models loaded, modeld starting");
 
+  bool main_wide_camera = false;
+  while (!do_exit) {
+    auto streams = VisionIpcClient::getAvailableStreams("camerad", false);
+    if (!streams.empty()) {
+      main_wide_camera = (streams.size() == 1 && streams.count(VISION_STREAM_WIDE_ROAD) == 1);
```
adeebshihadeh, line 195: this should also handle the case of a driver cam + road cam, so the check for size == 1 doesn't work

selfdrive/controls/controlsd.py
```diff
@@ -212,6 +209,14 @@ def set_initial_state(self):
       if any(ps.controlsAllowed for ps in self.sm['pandaStates']):
         self.state = State.enabled
 
+    if SIMULATION:
+      ignore = ['driverCameraState', 'managerState']
+      available_streams = VisionIpcClient.available_streams("camerad", block=False)
+      if len(available_streams) == 1 and VisionStreamType.VISION_STREAM_WIDE_ROAD in available_streams:
+        ignore += ['roadCameraState']
+      self.sm.ignore_alive += ignore
```
adeebshihadeh, line 217: this block shouldn't go in set_initial_state, it can just go directly in the body where set_initial_state is called

selfdrive/controls/controlsd.py
```diff
@@ -212,6 +209,14 @@ def set_initial_state(self):
       if any(ps.controlsAllowed for ps in self.sm['pandaStates']):
         self.state = State.enabled
 
+    if SIMULATION:
```
adeebshihadeh, line 212: this should always run, not just in SIMULATION mode

adeebshihadeh (changes requested):

PR #27648, opened 2023-03-21
Update values.py

Miniz199922:
More fingerprints added. its crazy i keep finding more. Sorry for multiple pulls


sshane:
Moving your car over to FPv2, you shouldn't have to worry about fingerprinting anymore! https://github.com/commaai/openpilot/pull/27649

PR #27649, opened 2023-03-22
Chrysler: add FW versions for 2019 Jeep Grand Cherokee

sshane:
From 5c77528ab6bf019a


Miniz199922:
thank you for doing this.

PR #27650, opened 2023-03-22
Subaru: show alert for when stock ACC disengages

### LINKED ISSUE #1163: Subaru: EyeSight unexpected ACC disengagement
Due to the rain? There's a couple instances in this route: `977a870d85685a57|2023-03-15--11-00-35`. This may be one of them: https://connect.comma.ai/977a870d85685a57/1678904273720/1678904290463

sshane:
As discussed earlier, we don't want to continue using the `cruiseDisabled` event in its current form (immediate disable without no entry), so I'll think about how we want to do this.

Todos:
- [ ] The car is in a no entry state, add button presses and show alert (make `cruiseDisabled` a `noEntry`)
- [ ] What happens if `carState.cruiseState.enabled` falls before the soft disable signal, or we see it that way? Do we not alert? Need to see how that works.

Screenshot_53

### DIFF
--- a/selfdrive/car/subaru/carstate.py
+++ b/selfdrive/car/subaru/carstate.py
@@ -231,6 +231,7 @@ class CarState(CarStateBase):
         ("Distance_Swap", "ES_Distance"),
         ("Standstill", "ES_Distance"),
         ("Signal3", "ES_Distance"),
+        ("Cruise_Soft_Disable", "ES_Distance"),
         ("Close_Distance", "ES_Distance"),
         ("Signal4", "ES_Distance"),
         ("Standstill_2", "ES_Distance"),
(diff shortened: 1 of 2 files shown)


sshane:
@adeebshihadeh did you have an idea on how to clean up the Honda usage of this signal? It's constantly added while disengaged to catch the radar disabling. Do we want to add one frame of it on falling edge of enabled or something similar?

sshane:
@jnewb1 perhaps you'd like to take this over when you get a chance? We need to make sure we handle soft disables from the car generically, I believe the cruise state can go low and disable openpilot before the soft disable signal rises, so we don't alert on occasion.

sshane:
Need to find a way to show soft disable type alerts after we've already disabled, as it's not a guarantee that the cars set the bits first, then disable cruise.

PR #27652, opened 2023-03-22
camerad: remove exp_lock, setting exposure parameters in set_camera_exposure

deanlee:
we don't need this lock any more.


adeebshihadeh:
Why don't we need this? It's still accessed by different threads.

deanlee:
Yes,my bad. closed. I forgot why I did it. It's been too long.

PR #27657, opened 2023-03-23
FPv2: consider bus for brand matching

sshane:
Minor improvement for cases where a car's diagnostic address is accessible on one bus, and both buses are present in the database.

For instance, when querying a Subaru's engine (0x7e0) on bus 1 (OBD), we add a match to all brands with an engine on 0x7e0 (such as Toyota, Chrysler, etc.). However with this PR, since Toyota queries engine on bus 0 and Subaru on bus 1, the bus 1 response won't mistakenly add a match for Toyota.


sshane:
Unclear if this provides any benefit, or if we even need it yet.

PR #27662, opened 2023-03-24
Hyundai CAN FD: log ADAS parking ECU FW version

sshane:
These ECUs describe whether a CAN FD platform Tucson Hybrid is a hybrid, so adding a request and the addrs to query for all CAN FD cars.

Our EV6 (HDA2):
- Responds to 0x7b1 with these ids that give important info: 0xF110 (`b'CV1  ADAS_PRK ANL 1.00 1.04 99910-CV070'`)
- Responds to 0x7b3 with these ids that give important info: 0xF187 (`97255CV830`)

Tucson Hybrid (non-HDA2):
- Responds to 0x7b1 with these ids that give important info: 0xF110 (`b'NX4H ADAS_PRK ANL 1.00 1.02 99910-P0800'`)
- Responds to 0x7b3 with these ids that give important info: 0xF100 (`b"NX4   97255-P0031CONTROL ASS'Y-DATC  1.05 NX4HEV ATC 0.7 V1.01"`)

So we're adding a new query for just these ECUs to query 0xF187, 0xF110, and 0xF100 in a multi query.


selfdrive/car/hyundai/values.py
```diff
@@ -355,22 +360,37 @@ class Buttons:
...
+      [HYUNDAI_VERSION_REQUEST_ALT],
+      [HYUNDAI_VERSION_RESPONSE],
+      whitelist_ecus=[Ecu.parking],
       bus=5,
     ),
   ],
   extra_ecus=[
     (Ecu.adas, 0x730, None),         # ADAS Driving ECU on HDA2 platforms
+    (Ecu.parking, 0x7b1, None),      # ADAS Parking ECU (may exist on all platforms)
+    (Ecu.hvac, 0x7b3, None),         # HVAC Control Assembly
```
sshane, line 393: hvac supported 0xF100, but parking ecu would return negatively to it, and if you stack requests into a multi query, parking ecu would only return a response to a shorter id without the necessary hybrid information, and hvac would then return negatively, so they needed to be split into two requests.

PR #27668, opened 2023-03-24
Hyundai CAN FD: log hvac FW version

sshane:
https://github.com/commaai/cereal/pull/427


sshane:
hopefully this is all we need!

PR #27676, opened 2023-03-24
HKG: add missing 2019 Genesis G70 FW 

erotisch:


sshane:
Can you provide a dongle id for where these came from? It will be in Device settings on your comma three

erotisch:
His Shane,

The dongle ID is aea0e22d415dfe02

Cheers,
Adam

sshane:
@erotisch do you have the 2021 model year?

erotisch:
No, it's a 2019

PR #27677, opened 2023-03-24
Update values.py

z3bby:
Toyota C-HR Hybrid 2021 - openpilot was unable to identify your car

Route 8b7530f5a63c87f7|2023-03-24--15-59-51


sshane:
Moved to  since you based the branch off master-ci. It looks like you're setting longitudinal to true, but you have a radar ACC car. I fixed that and set your car to stock long, since you were getting control collisions (the stock system and the OP's longitudinal messages were both being actuated on).

PR #27685, opened 2023-03-25
boardd: fix uninitialized can_header

briskspirit:


adeebshihadeh:
I tried enabling some more of clang's diagnostic flags, but none caught this.

PR #27687, opened 2023-03-25
Toyota: add missing FW versions for C-HR Hybrid 2021

sshane:
8b7530f5a63c87f7


sshane:
@z3bby can you try this branch out and post a route with rlogs uploaded? I could only check the sparse logs to see if your car was a radar ACC car, so this would confirm it. Then we can merge.

z3bby:
of course @sshane !
Here it is: 8b7530f5a63c87f7|2023-03-27--16-04-47.
Thank you.

sshane:
It looks like it indeed is a radar ACC car. Thanks for the route!

PR #27688, opened 2023-03-25
FPv2: match brand against extra ecus

sshane:
- [x] Queries these ecus with tester present
- [x] Matches against them

Improves number of matches for brands that use extra ecus (like CAN FD hyundai)


sshane:
Analyzing the `get_brand_addrs` function, which returns addresses for each brand to do brand matching in `get_brand_ecu_matches`:

```python
for brand in brand_addrs:
  print(brand, brand_addrs[brand] - prev_brand_addrs[brand])
  
body set()
chrysler {(2020, None), (2018, None)}
ford set()
honda {(416986097, None)}
hyundai {(1975, None), (1840, None), (1971, None), (1969, None)}
mazda set()
nissan set()
subaru set()
tesla set()
toyota set()
volkswagen set()
```

PR #27691, opened 2023-03-25
FW Request: return addrs

sshane:
Not sure if I like how big the diff is, it was supposed to make things simpler, but it turned into a *change*.

This gives me ideas of putting the FW versions inside the config though, then we can set up info we use later in the `__post_init__` function, removing duplicate code everywhere. Think: `config.ecu_types[(addr, sub_addr)]`, `config.get_addrs(extra_ecus=False, obd_multiplexing=True)`, etc.

Might close and re-open with that unification


sshane:
re-thinking

PR #27697, opened 2023-03-26
FPv2: only query brands with matching present ECUs

sshane:
Speeds up fingerprinting for CAN cars, where we do CAN fingerprinting after trying FW fingerprinting. Trying all the brands' requests takes 5+ seconds and shows that controls did not start alert


sshane:
Need to verify with at least 10,000 segs this doesn't break down and should be good to merge

sshane:
Looks good, checked this data: https://pastebin.com/LQqKnaVn

Only two segments were flagged, and had ECU address scan exceptions that were fixed in https://github.com/commaai/openpilot/pull/27956

PR #27702, opened 2023-03-27
cabana: go to previous zoom window

pd0wm:
Quickly go back to the previous zoom window using the mouse back button or toolbar action. This solves the problem of having to start over if you zoom in too far.

MacOS generates a fake gesture when pressing the mouse back button, even without grabbing this specific gesture. ~Need to check what happens on Linux and Windows before this can be merged.~ Linux generates a proper mouse pressed event

image

### DIFF
--- a/tools/cabana/chartswidget.cc
+++ b/tools/cabana/chartswidget.cc
@@ -56,6 +56,9 @@ ChartsWidget::ChartsWidget(QWidget *parent) : align_timer(this), QFrame(parent) 
   range_slider->setPageStep(60);  // 1 min
   range_slider_action = toolbar->addWidget(range_slider);
 
+  undo_zoom_action = toolbar->addAction(utils::icon("arrow-counterclockwise"), tr("Previous zoom"));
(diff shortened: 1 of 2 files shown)


deanlee:
Currently clicking the right mouse button will reset the room.
https://github.com/commaai/openpilot/blob/28cd8fc402bf47b119f960bd2bb511468f43d230/tools/cabana/chartswidget.cc#L681-L683

How about go to previous zoom with right mouse? and click the button in toolbar to reset room.

deanlee:
Currently clicking the right mouse button will reset the room.
https://github.com/commaai/openpilot/blob/28cd8fc402bf47b119f960bd2bb511468f43d230/tools/cabana/chartswidget.cc#L681-L683

How about go to previous zoom with right mouse? user still can click the button in toolbar to reset room.

pd0wm:
> Currently clicking the right mouse button will reset the room.
> 
> https://github.com/commaai/openpilot/blob/28cd8fc402bf47b119f960bd2bb511468f43d230/tools/cabana/chartswidget.cc#L681-L683
> 
> How about go to previous zoom with right mouse? user still can click the "zoomin" button in toolbar to reset room.

Sounds good to me! Will change once GitHub is no longer down.

PR #27706, opened 2023-03-27
Lexus: ES 2017-18 does not support stop and go

sshane:


sshane:
Will wait for verification from a user, not too many logs

PR #27707, opened 2023-03-27
Toyota: show true min enable speed for DSU cars

sshane:
Docs weren't accurate, this has been a confusion for people for a while

Leaving Prius V and Sienna alone, since I still haven't confirmed their features (sng/speeds may be incorrect).

Simply replaces 19 mph with 28 mph, which is more correct.


adeebshihadeh:
Shouldn't this go in carParams?

sshane:
I didn't want to regress user's minimum engage speed if we got it wrong for a platform (since I'm not so sure about the Prius V, and other less-common cars). Will try to go over every car to make sure the speeds are correct, then merge to CarParams.

sshane:
It's unclear whether we should show the best possible params in the docs, or what you get by default (for example we show experimental long limits for the Silverado which allows lower minEnableSpeed). We should clean up min acceleration and enable speeds first

PR #27716, opened 2023-03-28
cabana: clear zoom stack after reset zoom

deanlee:
fixed the zoom stack not cleared after user resets zoom via toolbar button.


pd0wm:
I guess it's more intuitive to clear the stack at that point. My reasoning was that you would be able to undo the reset zoom.

deanlee:
> My reasoning was that you would be able to undo the reset zoom.

y, but I'm worried that this stack is getting pretty messy and growing all the time, if the user keeps click the button to reset after zooming,

PR #27720, opened 2023-03-28
remove accFaultedTemp, add steerTimeLimit

jyoung8607:
Was intended to be used for GM but didn't end up being necessary. Garbage collecting per discussion with @sshane. Replace it with a new event we're using from #27624.

The corresponding cereal carEvents are updated in commaai/cereal#422.


adeebshihadeh:
Merged the cereal PR. We have a test that ensures all non-deprecated have alerts defined, so you'll have to add an alert definition in this PR for the new event.

PR #27722, opened 2023-03-28
Honda Civic 2022: match stock Auto Idle Stop behavior

royjr:
allow idlestop once stopped and counter reached to match stock behavior


sshane:
Auto Idle Stop seems to only exist on Civic 2022 atm (not the HRV which will be the same platform): https://www.hondainfocenter.com/-/media/Honda-Sales-Tool-Media-Folder/Images/2023-HR-V/PDFs/2023-HR-V-Features-by-trim.ashx

selfdrive/car/honda/hondacan.py
```diff
@@ -72,7 +72,7 @@ def create_acc_commands(packer, enabled, active, accel, gas, stopping, car_finge
   if car_fingerprint in HONDA_BOSCH_RADARLESS:
     acc_control_values.update({
       "CONTROL_ON": enabled,
-      "IDLESTOP_ALLOW": 0,  # disallows idle stop
+      "IDLESTOP_ALLOW": stopping_counter > 200,  # allow idle-stop after 4 seconds (50 Hz)
```
sshane, line 75:
```suggestion
      "IDLESTOP_ALLOW": stopping_counter > 200,  # allow idle stop after 4 seconds (50 Hz)
```

sshane:
@royjr can you provide a route of the stock system in action coming to a stop and turning off the engine after some time?

royjr:
I no longer have this vehicle unfortunately.

PR #27724, opened 2023-03-28
Honda: Match stock hide speed on gas override behavior

royjr:
hide speed on gas override to match stock behavior


selfdrive/car/honda/carcontroller.py
```diff
@@ -127,7 +127,7 @@ def update(self, CC, CS, now_nanos):
     actuators = CC.actuators
     hud_control = CC.hudControl
     conversion = hondacan.get_cruise_speed_conversion(self.CP.carFingerprint, CS.is_metric)
-    hud_v_cruise = hud_control.setSpeed / conversion if hud_control.speedVisible else 255
+    hud_v_cruise = hud_control.setSpeed / conversion if hud_control.speedVisible and not CC.cruiseControl.override else 255
```
sshane, line 130: Do you know if all Honda platforms hide set speed on gas press? What about all Honda Bosch?

royjr: Not so sure

selfdrive/car/honda/carcontroller.py
```diff
@@ -127,7 +127,10 @@ def update(self, CC, CS, now_nanos):
     actuators = CC.actuators
     hud_control = CC.hudControl
     conversion = hondacan.get_cruise_speed_conversion(self.CP.carFingerprint, CS.is_metric)
-    hud_v_cruise = hud_control.setSpeed / conversion if hud_control.speedVisible else 255
+    if self.CP.carFingerprint in HONDA_BOSCH_RADARLESS:
+      hud_v_cruise = hud_control.setSpeed / conversion if hud_control.speedVisible and not CC.cruiseControl.override else 255
+    else:
+      hud_v_cruise = hud_control.setSpeed / conversion if hud_control.speedVisible else 255
```
royjr, line 133: Alternative one liner:
`hud_v_cruise = hud_control.setSpeed / conversion if hud_control.speedVisible and ((self.CP.carFingerprint in HONDA_BOSCH_RADARLESS and not CC.cruiseControl.override) or self.CP.carFingerprint not in HONDA_BOSCH_RADARLESS) else 255`

sshane: I would pass override into the ACC function like you did before, that's a bit cleaner

selfdrive/car/honda/carcontroller.py
rjsmith1999: I think the behavior you’re describing also occurs on my nidec Honda. 

There are some exceptions though (the speed shows up again temporarily if you adjust the set speed while overriding).

royjr: Interesting, so the behavior occurs across all Hondas then. I did see that exception, I don’t care much for it, unless you think it’s important?

rjsmith1999: Well, it will probably make the code more complicated which is disappointing.  (unless the cluster implements this logic for us? I assume it doesn’t because you’re asking about it)

 But, with the set speed hidden it it’s hard to know what you’ve adjusted the set speed to. You can still see it in the openpilot ui, but I find myself looking at the cluster most of the time. 

If you ask me, I would trade the extra code complexity for the ability to see the set speed adjustments.

sshane:
Nidec also hides the speed permanently with foot on gas.

selfdrive/car/honda/carcontroller.py
sshane: @rjsmith1999 can you send a route with your car blinking the set speed?

sshane:
hiding the speed while pressing the gas doesn't seem like an improvement to me, and adding the code to temporarily show it when changing the speed would be a lot of code for this. closing, but feel free to experiment. for eg. perhaps blinking the speed with the gas like Hyundai does

PR #27725, opened 2023-03-28
ui: experimental -> alpha OP long toggle

royjr:
I find these two toggles to be a little confusing. You can't enable "Experimental Mode" without having "Experimental openpilot Longitudinal Control" enabled, so the order should be flipped. Both toggles include the word experimental which might be confusing to new users.

```
Experimental Mode
Experimental openpilot Longitudinal Control
```
to
```
openpilot Longitudinal Control
Experimental Mode
```
CF47AC18-6E7C-4421-A5E0-B0729767A047_1_201_a


adeebshihadeh:
Thanks for the PR! I reverted the change to the ordering to ensure the expanded version of the Experimental mode toggle shows without having to scroll.

PR #27726, opened 2023-03-29
updated: Allow version information poped up when DisableUpdates is enabled

eFiniLan:
Tested on master branch on C3.

When you have DisableUpdates enabled, those version information will not pop up.

This change will make those information available.

### DIFF
--- a/selfdrive/updated.py
+++ b/selfdrive/updated.py
@@ -393,7 +393,12 @@ class Updater:
 def main() -> None:
   params = Params()
 
+  updater = Updater()
+  update_failed_count = 0  # TODO: Load from param?
+  exception = None
+
   if params.get_bool("DisableUpdates"):
+    updater.set_params(update_failed_count, exception)
     cloudlog.warning("updates are disabled by the DisableUpdates param")
     exit(0)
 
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Any reason you keep updates disabled? I like that this makes it obvious that updates are disabled, otherwise they look silently broken.

eFiniLan:
> Any reason you keep updates disabled? I like that the current behavior makes it obvious that updates are disabled, otherwise they look silently broken.

Hi Adeeb,

I disabled it to keep these changes: https://github.com/commaai/openpilot/pull/27417
(speaking of which, I'd like to get some comments about it too if it's something you would accept)

I run the master branch on my C3 and have radar acc filter installed for openpilot longitudinal.

If you believe that the current behavior is intended, please close the PR.

royjr:
Maybe we can make it more obvious by coloring the version information red?

adeebshihadeh:
> I disabled it to keep these changes: #27417 (speaking of which, I'd like to get some comments about it too if it's something you would accept)

Cleaned up and merged. Thanks for including a test route!


As for this PR, I'm going to close this for two reasons: we want to encourage updating frequently and it's important to keep the updater code as simple as possible.

PR #27728, opened 2023-03-29
Honda: HR-V 2023 port

sshane:
From I got my honda back#0542 on Discord


sshane:
https://github.com/commaai/openpilot/pull/27731

PR #27741, opened 2023-03-30
LongitudinalPlanner: update max accel in turn to use vehicle model

pfeiferj:
**Description**
Uses the vehicle model to calculate the lateral acceleration when determining the maximum allowed acceleration in a turn. This improves the accuracy and consistency of accelerating during turns.

**Verification**
Have used for the last couple weeks in my 2020 Santa Fe. Have verified it limits acceleration.


pfeiferj:
@sshane Assuming you think this looks like it does what you meant in your FIXME comment, how should I go about determining the values for the lookup table?

sshane:
@HaraldSchafer we had something similar on the taco2 branch, do we want to merge this any time soon?

pfeiferj:
I've verified this as working, so going to go ahead and remove the draft flag. If changes are required for the lookup table I'd need some guidance on how to choose the values.

haraschax:
I'm going to close this PR. Technically this is orrect. It's not clear this improves or degrades driving behavior. And really we should get rid of this acceleration prevention anyway.

PR #27742, opened 2023-03-30
cabana: fix parts of chart widget not being redrawn

pd0wm:
Was changed in https://github.com/commaai/openpilot/pull/27727/files but has some side effects.

Screenshot 2023-03-30 at 09 37 57
(value marker is also not at correct mouse position)

### DIFF
--- a/tools/cabana/chartswidget.cc
+++ b/tools/cabana/chartswidget.cc
@@ -528,7 +528,7 @@ void ChartView::updatePlot(double cur, double min, double max) {
     updateAxisY();
     updateSeriesPoints();
   }
-  update();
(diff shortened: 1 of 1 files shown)


deanlee:
my bad, does this happen on mac with opengl disabled?
it's better to revert all of the `update` back to `scene()->invalidate({}, QGraphicsScene::ForegroundLayer)`. what do you think?

pd0wm:
It seems to only happen when openGL is disabled. Do you want to check for opengl to decide to call `update` or `invalidate`, or just revert everything back to how it was?

deanlee:
Revert them all is fine, it's not necessary to complicate the code for this little cpu saving.

PR #27744, opened 2023-03-30
cabana: show min and max values in chart tooltip

pd0wm:
Useful for computing the correct scalars when comparing the values to a different source (e.g. a scantool).
Screenshot 2023-03-30 at 13 48 04


pd0wm:
Rebased

PR #27746, opened 2023-03-30
cabana: fix drawing sparkline at half width on hi-dpi screen

pd0wm:
See https://github.com/commaai/openpilot/discussions/26091#discussioncomment-5477772

@deanlee why did you add this `* option.widget->devicePixelRatioF()`? I think there are some openGL functions that are not pixel ratio aware, but doesn't seem necessary here.


deanlee:
My mistake. I haven't tested it on high dpi screen

PR #27748, opened 2023-03-30
cabana: allow zooming to smaller ranges

pd0wm:
Use width of rubber instead of absolute time. Otherwise it's impossible to zoom beyond 0.5 seconds.


tools/cabana/chartswidget.cc
```diff
@@ -713,7 +713,7 @@ void ChartView::mouseReleaseEvent(QMouseEvent *event) {
     if (rubber->width() <= 0) {
       // no rubber dragged, seek to mouse position
       can->seekTo(min);
-    } else if ((max_rounded - min_rounded) >= 0.5) {
+    } else if (rubber->width() > 10) {
       // zoom in if selected range is greater than 0.5s
```
deanlee, line 717: looks like this comment is not needed

PR #27749, opened 2023-03-30
cabana: display signal value in all charts on mouse hover

deanlee:
related discuss:  https://github.com/commaai/openpilot/discussions/26091#discussioncomment-5477990

Screenshot 2023-03-30 23:36:43


deanlee:
rebased

Issue #27361, opened 2023-02-16
IsoTpParallelQuery hits total timeout on Mazda


MoreTore:
### Describe the bug

Finger printing has been slow when using 9.0 for months. I thought it was just my car but it seems to work fine on my C2 with 0.8.13. I have been putting off reporting this issue. This issue only happens after clearing the cached fingerprint. 

Here is the cleaned up tmux output.

```
Waiting for wayland
wayland ready
adjusting time from '2022-08-31 10:27:49.994168' to '2023-02-16 14:50:10'
Thu 16 Feb 2023 02:50:10 PM UTC
INFO:root:System time valid
fatal: no upstream configured for branch 'new_mazda_C3_stage'
Using Wayland-EGL
gbm_create_device(156): Info: backend name is: msm_drm
qt.qpa.wayland: "wl-shell" is a deprecated shell extension, prefer using "xdg-shell-v6" or "xdg-shell" if supported by the compositor by se
tting the environment variable QT_WAYLAND_SHELL_INTEGRATION
Using the 'wl-shell' shell integration
qt.qpa.wayland: Wayland does not support QWindow::requestActivate()
scons: Reading SConscript files ...
Git commit hash for gitversion.h: dc9270f6
scons: warning: Calling missing SConscript without error is deprecated.
Transition by adding must_exist=False to SConscript calls.
Missing SConscript 'selfdrive/navd/SConscript'
File "/data/openpilot/SConstruct", line 436, in <module>
scons: done reading SConscript files.
scons: Building targets ...
scons: `.' is up to date.
scons: done building targets.
corrupted double-linked list
selfdrive/loggerd/bootlog.cc: bootlog to /data/media/0/realdata/boot/2023-02-16--08-50-29
fatal: no upstream configured for branch 'new_mazda_C3_stage'
gbm_create_device(156): Info: backend name is: msm_drm
: QPixmap::scaleWidth: Pixmap is a null pixmap
: "wl-shell" is a deprecated shell extension, prefer using "xdg-shell-v6" or "xdg-shell" if supported by the compositor by setting the envi
ronment variable QT_WAYLAND_SHELL_INTEGRATION
: Wayland does not support QWindow::requestActivate()
NVME not mounted
Panda 1b0023001850335332313620 connected, version: DEV-dc9270f6-DEBUG, signature ab79f3c5faf4c5c0, expected ab79f3c5faf4c5c0
selfdrive/boardd/main.cc: starting boardd
[...]

PR #27762, opened 2023-03-31
IsoTpParallelQuery: extend timeouts when receiving consecutive frames

sshane:
This fixes a rare bug introduced in https://github.com/commaai/openpilot/pull/25897, which originally aimed to fix long responses (in bytes and time) timing out. This new behavior should now strictly be correct.

Previous behavior:
- On receipt of any message from the rx address (including msgs from a broadcast address), return `updated=True` and extend the address timer

  **Problem**: if an address existed on the car broadcasting junk (not a diagnostic address), `IsoTpParallelQuery` would think that the ECU was building a response and wait for the total timeout of 1 minute to time out.

New behavior:
- On receipt of a valid consecutive frame and rx is not done, return `rx_in_progress=True` and extend the address timer
- If rx has completed, only if the response is valid and we need to send the next request do we extend the address timer
- Don't extend timeout for the first iso-tp frame if there are multiple, only subsequent frames

This would have handled https://github.com/commaai/openpilot/issues/27361 by not extending the timeout due to those messages being interpreted as single frames. Consecutive frames also have two checks that would fail in nearly all cases where it is a broadcast address.


sshane:
These dongles would be affected by 0x730 + 0x8 broadcasting on their cars (we assume 0x738 is the RX addr for the ADAS driving ECU)"

```
{('23940cdb2383e639', 'HYUNDAI SONATA 2019'), ('a9bb30d693e9008d', 'HYUNDAI KONA ELECTRIC 2019'), ('734ef96182ddf940', 'HYUNDAI ELANTRA 2017'), ('0665dbbc79ef999e', 'HYUNDAI KONA HYBRID 2020'), ('a3669d37d7fe28c2', 'KIA OPTIMA 4TH GEN FACELIFT'), ('7ae1c131629d96e5', 'HYUNDAI SONATA 2019'), ('3aaf08d1932fa7d1', 'KIA OPTIMA 4TH GEN FACELIFT'), ('c9508b688d2009f6', 'HYUNDAI KONA ELECTRIC 2019'), ('6b111e8e45cc2d07', 'KIA SORENTO GT LINE 2018'), ('32f16ba9460d4e9d', 'KIA OPTIMA 4TH GEN FACELIFT'), ('9a3aa85e86f9c3fc', 'HYUNDAI KONA HYBRID 2020'), ('3d96bd05b5513638', 'KIA OPTIMA 4TH GEN FACELIFT'), ('7b81314388240dbb', 'KIA OPTIMA 4TH GEN FACELIFT'), ('235090a72cae0a02', 'KIA OPTIMA 4TH GEN FACELIFT'), ('c1fc13c1b806e536', 'HYUNDAI SONATA 2019'), ('b78a19b557b991ab', 'KIA SORENTO GT LINE 2018'), ('37b6542f3211019a', 'KIA OPTIMA 4TH GEN FACELIFT'), ('700d37ccd12315cf', 'HYUNDAI SONATA 2019'), ('39ff090a42469a72', 'HYUNDAI KONA ELECTRIC 2019')}
```

PR #27769, opened 2023-03-31
cabana: show min and max values in sparkline on hover

deanlee:
Screenshot 2023-04-01 03:41:21

Screenshot 2023-04-01 03:49:08

### DIFF
--- a/tools/cabana/signalview.cc
+++ b/tools/cabana/signalview.cc
@@ -344,6 +344,17 @@ bool SignalItemDelegate::helpEvent(QHelpEvent *event, QAbstractItemView *view, c
   return QStyledItemDelegate::helpEvent(event, view, option, index);
 }
 
+void SignalItemDelegate::updateEditorGeometry(QWidget *editor, const QStyleOptionViewItem &option, const QModelIndex &index) const {
+  auto item = (SignalModel::Item *)index.internalPointer();
+  if (editor && item->type == SignalModel::Item::Sig && index.column() == 1) {
+    QRect geom = option.widget->style()->subElementRect(QStyle::SE_ItemViewItemText, &option);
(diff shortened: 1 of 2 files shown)


deanlee:
fyi:
the buttons's gemotry is set in `SignalItemDelegate::updateEditorGeometry`. maybe something is wrong here

deanlee:
the origin code use `h->addStretch(0);` to put the widgets on the right. but this will cause the entire column to be covered by the widget. the mouse event on column 1 will not be sent to SignalView. 

So I removed the h->addStretch(0), and set widget position to the right in  function `updateEditorGeometry`.

pd0wm:
Are you ok with replacing `option.widget->style()->subElementRect(QStyle::SE_ItemViewItemText, &option);` by `option.rect`? That seems to do the trick as well.

deanlee:
that's ok!   the reason I use `subElementRect` is that it can correctly add margins to option.rect(QStyle::PM_FocusFrameVMargin&QStyle::PM_FocusFrameHMargin)

pd0wm:
That does not fix it. For completeness, I changed the function to:

```c++
void SignalItemDelegate::updateEditorGeometry(QWidget *editor, const QStyleOptionViewItem &option, const QModelIndex &index) const {
  auto item = (SignalModel::Item *)index.internalPointer();
  if (editor && item->type == SignalModel::Item::Sig && index.column() == 1) {
    QStyleOptionViewItem opt = option;
    initStyleOption(&opt, index);
    QRect geom = option.widget->style()->subElementRect(QStyle::SE_ItemViewItemText, &opt);
    geom.setLeft(geom.right() - editor->sizeHint().width());
    editor->setGeometry(geom);
    return;
  }
  QStyledItemDelegate::updateEditorGeometry(editor, option, index);
}
```

PR #27773, opened 2023-04-01
Car interfaces: remove hard-coded signal definitions

### LINKED ISSUE #27695: CANParser: parse all signals from specified messages
Many of the lines in a car port are spent on whitelisting signals to be parsed out. The CANParser should be fast enough now that we can simply whitelist messages and parse out all their signals.

- [x] remove the signals arg from the CANParser input, see https://github.com/commaai/opendbc/pull/610 as a starting point
- [ ] add a test case to the [car unit tests](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/test_models.py) to ensure parsing time is reasonable
- [ ] if the test fails, optimize the CANParser until it doesn't

sshane:


sshane:
@deanlee the car interfaces would look much cleaner once we start parsing all the signals :). Do you think performance is good enough to use now?

deanlee:
the new benchmark looks good:

https://github.com/commaai/opendbc/pull/795#issuecomment-1492864190

need to continue improve the code and do more testing, but surely, I think it's ok to remove them all after https://github.com/commaai/opendbc/pull/795  is ready.

jyoung8607:
> [and vw! (sorry comments)](https://github.com/commaai/openpilot/pull/27773/commits/913aad74c25e9b5209a87b8f3753dd50aa5fbe02)

If you strike down my comments, I shall become more verbose than you can possibly imagine.

...

j/k :) this is a good change.

adeebshihadeh:
@sshane before merging this, I'd like a prereq PR that explicitly copies out any fields that . otherwise, it's easy to merge a DBC update that silently changes behavior, especially since those strings won't be grep-able.

sshane:
@adeebshihadeh I'll look over everything tomorrow and should be good if your opendbc PR is ready.

selfdrive/car/chrysler/radar_interface.py
```diff
@@ -89,4 +83,4 @@ def update(self, can_strings):
     ret.points = [x for x in self.pts.values() if x.dRel != 0]
 
     self.updated_messages.clear()
-    return ret
\ No newline at end of file
+    return ret
```
sshane, line 86:
```suggestion
    return ret```

**Comment by sshane** [comment]
Ran test_models locally a few times on master and this PR to test the time to parse CAN messages in each platform's interface and carstate. The two tests ran were `test_car_interface` and `test_radar_interface` with 4 jobs, all rlogs cached.

`master`:
```
430 tests
03:26
03:24
03:22
```

`what-if-no-signals`:
```
430 tests
03:40
03:42
```

sshane:
@jyoung8607 @incognitojam if there's any comments this removes for signals, make sure they're in the DBC. Then you can add to where they're used if how they're used isn't obvious

PR #27777, opened 2023-04-01
fingerprint for KIA STINGER 2022

mansunsang:
fingerprint for KIA Stinger 2022 sold in South Korea


sshane:
Can you provide the dongle id that these came from? It will be in Device settings.

PR #27783, opened 2023-04-02
Toyota: missing Corolla Cross Hybrid FW

kyoma22:
add corolla cross HV


sshane:
Can you provide the dongle that these came from?

sshane:
Any updates?

kyoma22:
dongle id is c7400d04331ac3ee

PR #27786, opened 2023-04-02
cabana: significantly reduce CPU usage by caching chart to pixmap

deanlee:
override `QChartView::paintEvent` to improve the pool real-time performance of `QtChart` by caching chart to pixmap.
The default implementation (https://codebrowser.dev/qt5/qtbase/src/widgets/graphicsview/qgraphicsview.cpp.html#_ZN13QGraphicsView10paintEventEP11QPaintEvent) redraws all items in graphics scene every time, which can lead to very high CPU usage on updating.

The test results after refactor: （settings.fps = 10）:
-  One chart:
  opengl enabled: save ~6% of a core
  opengl disabled: save ~12% of a core 
- Four charts:
  opengl enabled: save ~22% of a core
  opengl disabled: save ~50% of a core

### DIFF
--- a/tools/cabana/chartswidget.cc
+++ b/tools/cabana/chartswidget.cc
@@ -393,10 +393,14 @@ ChartView::ChartView(QWidget *parent) : tip_label(this), QChartView(nullptr, par
   setChart(chart);
 
   createToolButtons();
-  setRenderHint(QPainter::Antialiasing);
   // TODO: enable zoomIn/seekTo in live streaming mode.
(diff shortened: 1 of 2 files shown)


pd0wm:
This draws the graph at 2x the size when using OpenGL and a hi-dpi monitor on MacOS. Is that also true on Linux? Setting the pixel ratio to 1 when using OpenGL seems to fix it.

deanlee:
It is quite possible to have the same issue on linux with a hi-dpi monitor.

PR #27793, opened 2023-04-03
cabana: fix chart background issue in live mode

deanlee:
before:
Screenshot from 2023-04-03 17-13-45

after:
Screenshot from 2023-04-03 17-13-04


tools/cabana/chartswidget.cc
```diff
@@ -881,6 +882,9 @@ void ChartView::paintEvent(QPaintEvent *event) {
     QChartView::paintEvent(event);
   }
 }
+void ChartView::drawBackground(QPainter *painter, const QRectF &rect) {
```
sshane, line 885:
```suggestion

void ChartView::drawBackground(QPainter *painter, const QRectF &rect) {
```

PR #27794, opened 2023-04-03
Amplifier I2C retry

robbederks:
This is necessary, since a collision on the multi-master I2C bus will throw an OSError on the ioctl.


adeebshihadeh:
Already started in #27735, but wanted to get a CI test in before merging it.

PR #27803, opened 2023-04-04
VW PQ: Miscellaneous longitudinal bug fixes

jyoung8607:
**Description**

Fix three separate issues with openpilot longitudinal control on VW PQ:

1) Fix the instrument cluster getting "stuck" displaying the ACC screen with lead car indicator. With the fixes from commaai/opendbc#792 we can now correctly set the lower display priority.

2) Fix the engine auto start-stop inhibit signal. It was inverted.

3) Set the accel request valid field when enabled. Not sure how we got away without this in the past, for example on the 2018 NMS Passat I tested last year, but some older ECUs are ignoring accel requests without it.

**Verification**

The fixes have all been road-tested on an older NMS (USA) Passat and a European VW Sharan.

**Route**

Route: `e12eb71c201af95a|2023-04-05--07-26-02`

**Additional Info**

The combination of `ACA_PrioDisp` and `ACA_AnzDisplay` are supposed to be used to temporarily pop the ACC page to the foreground and increase brightness when activity happens: engagement, follow distance adjustment, or a lead car comes into range. Eventually we should emulate that stock behavior. However, openpilot doesn't allow adjustment of the follow distance, and its lead car data is so low-value right now (presence only, no distance, noisy/not debounced) that it's not yet worth implementing. The driver can still switch to the ACC data screen manually if they want to see it.

### DIFF
--- a/selfdrive/car/volkswagen/pqcan.py
+++ b/selfdrive/car/volkswagen/pqcan.py
@@ -85,8 +85,8 @@ def create_acc_hud_control(packer, bus, acc_hud_status, set_speed, lead_distance
     "ACA_V_Wunsch": set_speed,
     "ACA_gemZeitl": lead_distance,
     "ACA_PrioDisp": 3,
-    # TODO: ACA_ID_StaACC, ACA_AnzDisplay, ACA_kmh_mph, ACA_Aend_Zeitluecke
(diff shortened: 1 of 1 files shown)


jyoung8607:
Testing was successful, ready for review.

PR #27806, opened 2023-04-05
cabana: cheat engine search

jnewb1:
**Description**
A proof of concept version of a "cheat-engine" like search feature in Cabana. To try it out, clone the branch and try doing a scan (Tools->Search)

Things I need to add:

- undo scan (needs the full scan history which is going to be annoying)
- combining similar bit ranges together or displaying them differently. right now if there is some padding between two messages it will display the same bit range with a bit of padding like 100 times
- float/signed value support
- big endian
- more testing
- add to "watch" (set of signals you can watch more closely, aside from the giant list)
- add to signals button

Looking for feedback @deanlee


tools/cabana/tools/search.h
```diff
@@ -0,0 +1,194 @@
...
+
+        MessageId messageID;
+        size_t offset;
+        size_t size;
+
+        uint64_t previousValue;
+    
+        uint64_t getValue(){
+            auto msg = can->last_msgs[messageID];
+            uint64_t* data = (uint64_t*)(msg.dat.data());
```
deanlee, line 48: can't assume the data size is 64 bytes. you need to get the real size from msg.dat.size()

tools/cabana/tools/search.cc
```diff
@@ -0,0 +1,255 @@
...
+    }
+    if(selectedScanType == DecreasedValue){
+        return new DecreasedValueSignalFilter();
+    }
+
+    return nullptr;
+}
+
+void SearchDlg::nextScan(){
+    auto filterer = getCurrentFilterer();
```
deanlee, line 246: looks like a memory leak here.the filterer not deleted after use

tools/cabana/tools/search.h
```diff
@@ -0,0 +1,194 @@
...
+            return getBitValue(*data, offset, size);
+        }
+};
+
+class SignalFilterer {
```
deanlee, line 53: Is it possible that these filter classes can be replaced by std library, e.g. std::greater, std::less, that might saves a lot of code.  (like what we did in void LogsWidget::setFilter())

tools/cabana/videowidget.cc
```diff
@@ -188,15 +188,36 @@ void Slider::sliderChange(QAbstractSlider::SliderChange change) {
   }
 }
 
+bool sortTimelineBasedOnEventPriority(std::tuple<int, int, TimelineType> left, std::tuple<int, int, TimelineType> right){
```
deanlee, line 191: Think it's better to move video related changes to another pr.

SConstruct
```diff
@@ -175,8 +175,8 @@ env = Environment(
   CCFLAGS=[
     "-g",
     "-fPIC",
-    "-O2",
-    "-Wunused",
+    "-O0",
```
deanlee, line 178: it's better not to change flags now

tools/cabana/tools/search.cc
```diff
@@ -0,0 +1,255 @@
...
+
+    for(auto &sig : filteredSignals){
+        sig.previousValue = sig.getValue();
+    }
+}
+
+std::vector<Sig> getAllPossibleSignals(int bits_min, int bits_max){
+    std::vector<Sig> ret;
+    
+    for(auto msg_id : can->last_msgs.keys()) {
```
deanlee, line 165: the last_msgs is updated at settings.fps (default is 10hz).  it's mainly used for UI to display the latest signal data. you probably want to use can->rawEvents or can->events to scan through all signals.

tools/cabana/tools/search.h
jnewb1: some of then probably could, but others probably couldn't ("increased value by...", "value between"). We could probably replace it with simple functions instead of classes though.

jnewb1:
- Added undo ability
- Used a "basesignal", which is shared between "signal" and "searchsignal"
- added "add to signals" button

first few scans are now kinda slow, probably because we are storing a lot more data for each signal now. probably some stuff we can do to optimize it

jnewb1:
@deanlee this is ready for another review

deanlee:
1. Is there any particular reason you want to change the underlying structure of cabana::signal? I don't see a need for that. Let's change only the file `tools/search.cc[h]` if possible and not make any unnecessary changes in other files.
2. The current code implementation has performance issues. you might want to consider a more efficient approach. 
3. also be careful to avoid the data copying. For example, code like `auto events = can->events(messageID);`  should be changed to `const auto &events = can->events(messageID);`

jnewb1:
> * Is there any particular reason you want to change the underlying structure of cabana::signal? I don't see a need for that. Let's change only the file `tools/search.cc[h]` if possible and not make any unnecessary changes in other files.
> * The current code implementation has performance issues. you might want to consider a more efficient approach.
> * also be careful to avoid the data copying. For example, code like `auto events = can->events(messageID);`  should be changed to `const auto &events = can->events(messageID);`
> * The total number of lines of code in cabana is currently about 4000 lines. 600+ lines (more than 15%) is a bit too much to implement this feature. It would be nice if this could be done in fewer lines.

1. I wanted to reuse other code if possible (such as getting the value of a signal), but I didn’t want to have to store the other info related to a signal, like the signal name, so I thought a base class made sense. Do you have another suggestion?
2. It’s not ideal but it only takes 5-10 seconds for full scan. I can definitely multithread it but I’ve used it quite a bit myself without much issue.
3. I will take another look through my code and find data copying.
4. I can look at trimming it down a bit, but I would expect I can only remove maybe 100-150 lines

sshane:
Regarding the number of lines, @jnewb1 is there a minimal set of changes here to get the infrastructure merged? Like maybe starting with only supporting a few scan types.

jnewb1:
> 1. Is there any particular reason you want to change the underlying structure of cabana::signal? I don't see a need for that. Let's change only the file `tools/search.cc[h]` if possible and not make any unnecessary changes in other files.
> 2. The current code implementation has performance issues. you might want to consider a more efficient approach.
> 3. also be careful to avoid the data copying. For example, code like `auto events = can->events(messageID);`  should be changed to `const auto &events = can->events(messageID);`
> 4. The total number of lines of code in cabana is currently about 4000 lines. 600+ lines (more than 15%) is a bit too much to implement this feature. It would be nice if this could be done in fewer lines.

1. Search is using functions like:

getSignalRange
updateSigSizeParamsFromRange
get_raw_value

etc..., which would need an duplicate function in search.c/search.h if we don't share a base class.

jnewb1:
> 1. Is there any particular reason you want to change the underlying structure of cabana::signal? I don't see a need for that. Let's change only the file `tools/search.cc[h]` if possible and not make any unnecessary changes in other files.
> 2. The current code implementation has performance issues. you might want to consider a more efficient approach.
> 3. also be careful to avoid the data copying. For example, code like `auto events = can->events(messageID);`  should be changed to `const auto &events = can->events(messageID);`
> 4. The total number of lines of code in cabana is currently about 4000 lines. 600+ lines (more than 15%) is a bit too much to implement this feature. It would be nice if this could be done in fewer lines.

2/3. removed more data copying, and performance is essentially instantaneous now, even on the early scan with hundreds of thousands of signals

jnewb1:
resynced with master, and reduced line count to 477 by refactoring using functions rather than classes. IMO a little less pretty, but at least its smaller

deanlee:
1. don't change the underlying signal data structure just because some of fields are not needed in the search.think of it as a read only library header file instead of changing it arbitrarily. design your own data structure on this basis.
2. it's ok to copy and create you own get_raw_value in search.cc. 
3. There still some problems in the current code need to be fixed before opening it for review, such as repeatedly assigning a value to start_bit, incorrectly using lower_bound to search for the next message, and not checking the return value correctly, the first line of the tableview should be put in the header, etc.

deanlee:
@jnewb1 Thank you very much for your time and this fantastic PR. However, this PR has been ongoing for a long time, and continuously reviewing and changing it may waste a lot of time. Therefore, I have submitted a new PR to implement this feature. https://github.com/commaai/openpilot/pull/28182  any suggestion? does it looks good to you?

adeebshihadeh:
Thanks @jnewb1, merged @deanlee's refactored version.

jnewb1:
😆

jnewb1:
> @jnewb1 Thank you very much for your time and this fantastic PR. However, this PR has been ongoing for a long time, and continuously reviewing and changing it may take up more time. Therefore, I have submitted a new PR to implement this feature. #28182, does it looks good to you?

The more useful options are “changed” and “unchanged”.

deanlee:
> The more useful options are “changed” and “unchanged”. Most of the time you won’t know the actual value

There are already a well-implemented activity colors module that used to track `changed/non-changed/increasing/decreasing/constant changing` bits and display theme with different colors.

jnewb1:
> > The more useful options are “changed” and “unchanged”. Most of the time you won’t know the actual value
> 
> There is already a well-implemented activity colors module that used to track `changed/non-changed/increasing/decreasing/constant-changing` bits and display theme with different colors.

Well sure, but you can only look at one message at a time. Let me give you an example. I was looking for a signal on my Subaru, that was the “infotainment status” (a message that popped up on my infotainment screen that showed when LKAS was turning off due to not touching the steering wheel). What I did was whenever I saw the signal change while driving, I marked it with a user flag on the C3. Then when I got home, I did a search where I did “changed value” on all the flags, and “unchanged” in between the flags. I didn’t know the actual value that was used to store the message at all, I just knew it was changing. I was able to search the entire cars can bus that way. It turns out it wasn’t even a defined message yet. I don’t think the color activity monitor and the “changed/unchanged” are even comparable for this use case.

deanlee:
This requires comparing the snapshots of data from different times. (This is also what you did in this pr). but it's not good for other types of searching: all intermediate information between mouse clicks is lost.

The new findSignalDlg is not based on snapshot, but searches through all received messages. So for the time being it is not possible to provide this kind of data comparison. I think we need to implement such snapshot comparison separately rather than merging them together.

jnewb1:
> This requires comparing the snapshots of data from different times. (This is also what you did in this pr). but it's not good for other types of searching: all can events between clicks is ignored.
> 
> The new findSignalDlg is not based on snapshot, but searches through all cached messages. So for the time being it is not possible to provide this kind of data comparison in it. I think we may need to implement such snapshot comparison separately rather than merging them together.

Cheat engine uses snapshot based searching, and I think this use case is very similar to cheat engine. You won't know the "factor" or "offset" of a signal that you are looking for, or if its enum-like you won't know the enum values.

PR #27807, opened 2023-04-05
change prev_accel_constraint condition & set initial prev_a value

ajouatom:
**Description** 

Synchronize aEgo while the PID is off and apply prev_accel_constraint to ensure a continuous acceleration output after an override due to the gas pedal.

**Verification** 

After the gas pedal override, I confirmed through PlotJuggler that the accels of MPC are dispersed from aEgo.

### DIFF
--- a/selfdrive/controls/lib/longitudinal_planner.py
+++ b/selfdrive/controls/lib/longitudinal_planner.py
@@ -9,7 +9,7 @@ from common.filter_simple import FirstOrderFilter
 from common.realtime import DT_MDL
 from selfdrive.modeld.constants import T_IDXS
 from selfdrive.controls.lib.longcontrol import LongCtrlState
-from selfdrive.controls.lib.longitudinal_mpc_lib.long_mpc import LongitudinalMpc, MIN_ACCEL, MAX_ACCEL
+from selfdrive.controls.lib.longitudinal_mpc_lib.long_mpc import LongitudinalMpc, MIN_ACCEL, MAX_ACCEL, N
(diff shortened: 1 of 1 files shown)


ajouatom:
Before
image

After
image

sshane:
What does this aim to do?

ajouatom:
> What does this aim to do? Make sure we don't start too close to aEgo for hard acceleration overrides, or does this improve smoothness after overriding?

Yes!, improve smoothness after overriding.

When attempting to change lanes while approaching a leading vehicle, drivers typically engage gas override. During this time, the MPC is creating a deceleration profile, and if the gas override ends and the PID control begins, a discontinuity point can occur, causing a significant shock to the vehicle.

Therefore, synchronizing the prev_a value of the MPC with aEgo while PID is off can result in a very smooth cruise control.

rchybicki:
Shouldn't this be merged? 
I've been using testing this change since April, and recently lost it during a merge - I instantly felt the difference on the first day. It smooths out a scenario very common for me: 
manual accel from slow speeds -> engage -> let go of the gas. There is a 1-2s pause before OP takes over (using OP Long).
Without this change, the jerk after engaging like that is much worse.

ajouatom:
> Shouldn't this be merged? I've been using testing this change since April, and recently lost it during a merge - I instantly felt the difference on the first day. It smooths out a scenario very common for me: manual accel from slow speeds -> engage -> let go of the gas. There is a 1-2s pause before OP takes over (using OP Long). Without this change, the jerk after engaging like that is much worse.

It seems like this request hasn't been merged, probably because other Openpilot users don't press the gas pedal the same way we do.

PR #27812, opened 2023-04-05
Honda: Added fingerprints for a 2022 Acura RDX

royjr:
5ec737f0b399336e swag0#8418


sshane:
Wow, all ECUs were missing! Nice job, you at least enabled many future cars to now fuzzy fingerprint by the sheer number of different FW versions Honda has

PR #27814, opened 2023-04-05
Honda Bosch Radarless:  check cruise faults

royjr:
Dependent on https://github.com/commaai/opendbc/pull/807


sshane:
Only found two instances of this signal being high in the field out of 1k segments (all we have locally for radarless Civic), and they look like by products of some CAN issue. CAN valid also drops while this is happening, and POWERTRAIN_DATA is spotty. Both from one dongle. Looks like the signal is good (and radarless cars don't fault often)!


Screenshot from 2023-04-14 02-50-00

sshane:
thanks for starting the PR @royjr!

PR #27815, opened 2023-04-05
Cleanup: sort Honda values

royjr:
fixed sorting and missing commas


adeebshihadeh:
Thanks for the PR, but we can't merge this. It's non-trivial to review with quite a bit of risk.

PR #27817, opened 2023-04-06
rawgpsd: log time to first fix

sshane:


sshane:
Possibly moving to locationd

PR #27819, opened 2023-04-06
quectel: move vertical accuracy sanity check

sshane:
moves the gps-specific check into rawgpsd, added here: https://github.com/commaai/openpilot/pull/26352


system/sensord/rawgps/rawgpsd.py
```diff
@@ -283,6 +282,8 @@ def cleanup(sig, frame):
       gps.verticalAccuracy = report["q_FltVdop"]
       gps.bearingAccuracyDeg = report["q_FltHeadingUncRad"] * 180/math.pi
       gps.speedAccuracy = math.sqrt(sum([x**2 for x in vNEDsigma]))
+      # quectel gps verticalAccuracy is clipped to 500
```
sshane, line 285:
```suggestion
      # quectel gps verticalAccuracy is clipped to 500, invalid if so
```

PR #27820, opened 2023-04-06
cabana: improve chart drag and drop

deanlee:
1. transparent drag and drop overlay
2. faster merge charts
3. draw drop indicator

4. auto scroll while dragging over scroll area:

[Kazam_screencast_00026.webm](https://user-images.githubusercontent.com/27770/230290536-3c0753a4-3a44-465e-be79-1786438f25ce.webm)


pd0wm:
It would be very nice if you can also drag a series out of a combined graph back into it's own graph. No you have to go into a menu to unmerge them. Maybe by dragging from the legend entry?

deanlee:
I'll try to implement it after we can change chart order by drag&drop(another PR).

PR #27822, opened 2023-04-06
VW MQB: add missing 2019 Polo MK6 FW

Saber422:
Route No1 ：0bbe367c98fa1538|2023-04-06--13-28-30--0 Route No2 ：0bbe367c98fa1538|2023-04-06--13-35-01--0


sshane:
@jyoung8607 the 2019 isn't listed for any of the Polos. Is this a market difference (Taiwan), or can we extend the MY to 2019?

jyoung8607:
We can extend back to 2018. No earlier, the pre-2018 MQB-A0 steering racks don't support Lane Assist. I'll do it in another PR.

PR #27829, opened 2023-04-07
Subaru: infotainment status intercept and rewrite

jnewb1:
*Description*

Fixes the error message that pops on the Subaru infotainment screen when eyesight lateral is running in the background, and it gets disabled because of no steering wheel input.

*Verification*

Tested on my Subaru legacy 2022. Will need more testing on other platforms. Easiest way is probably to enable eyesight lateral while driving with openpilot, and make a bookmark whenever something changes on the infotainment screen.

These two PR’s are required first:
https://github.com/commaai/panda/pull/1333
https://github.com/commaai/opendbc/pull/809


.gitignore
```diff
@@ -83,3 +83,6 @@ build/
 !**/.gitkeep
 
 poetry.toml
+
+sandbox/
+settings
```
adeebshihadeh, line 88: revert these

selfdrive/car/subaru/subarucan.py
```diff
@@ -76,6 +76,19 @@ def create_es_dashstatus(packer, dashstatus_msg):
 
   return packer.make_can_msg("ES_DashStatus", 0, values)
 
+def create_infotainmentstatus(packer, infotainmentstatus_msg, visual_alert):
+  values = copy.copy(infotainmentstatus_msg)
```
adeebshihadeh, line 80: already copied in carstate.py

selfdrive/car/subaru/subarucan.py
```diff
@@ -76,6 +76,19 @@ def create_es_dashstatus(packer, dashstatus_msg):
...
+
+  # Filter stock LKAS disabled and Keep hands on steering wheel OFF alerts
+  if values["LKAS_State_Infotainment"] in [3,4]:
```
adeebshihadeh, line 83:
```suggestion
  if values["LKAS_State_Infotainment"] in (3, 4):
```

selfdrive/car/subaru/subarucan.py
```diff
@@ -76,6 +76,19 @@ def create_es_dashstatus(packer, dashstatus_msg):
...
+    values["LKAS_State_Infotainment"] = 0
+  
+  # Show Keep hands on wheel alert for openpilot steerRequired alert
```
adeebshihadeh, line 86: let's also show the obstacle detected alert when `visual_alert == VisualAlert.fcw`

.gitignore
jnewb1: done

selfdrive/car/subaru/subarucan.py
jnewb1: done. also removed the copy from dashstatus

selfdrive/car/subaru/subarucan.py
jnewb1: done, did same in dashstatus

selfdrive/car/subaru/subarucan.py
jnewb1: done. _needs verification_

adeebshihadeh:
NIce job! Merged the DBC updates and the panda side looks good, but we'll have to wait on merging that until this is ready to merge. @sshane can you test on our Outback and merge if it looks good?

adeebshihadeh (approved):
NIce job! Merged the DBC updates and the panda side looks good, but we'll have to wait on merging that until this is ready to merge. @sshane can you test on our Outback and merge if it looks good?

jnewb1:
There was an issue since “values” was no longer defined. I switched it to use the parameter value.

Also tested that “2” was “obstacle detected”. “1” also seemed to be “obstacle detected”

sshane:
All the definitions match on our Outback!

jnewb1:
> All the definitions match on our Outback! Yep, 1 and 2 were both obstacle detected for this car as well

awesome. legacy and outback are the same platform though, so it might be good to get a community route or have someone test it

sshane:
Changed the panda branch to a temporary one so tests would run. It looks like this infotainment message does not exist on non-GEN2 cars, so it should be safe to gate behind that and merge. @jnewb1 can you change to only GEN2?

```python
FAILED test_models.py::TestCarModel_158_SUBARU_OUTBACK_2015_2017::test_car_interface - KeyError: 'INFOTAINMENT_STATUS'
FAILED test_models.py::TestCarModel_152_SUBARU_FORESTER_2017_2018::test_car_interface - KeyError: 'INFOTAINMENT_STATUS'
FAILED test_models.py::TestCarModel_151_SUBARU_ASCENT_LIMITED_2019::test_car_interface - AssertionError: 5799 != 0
FAILED test_models.py::TestCarModel_159_SUBARU_OUTBACK_2018_2019::test_car_interface - KeyError: 'INFOTAINMENT_STATUS'
FAILED test_models.py::TestCarModel_156_SUBARU_LEGACY_2015_2018::test_car_interface - KeyError: 'INFOTAINMENT_STATUS'
FAILED test_models.py::TestCarModel_154_SUBARU_IMPREZA_LIMITED_2019::test_car_interface - AssertionError: 10480 != 0
```

sshane:
Changed the panda branch to a temporary one so tests would run. It looks like this infotainment message does not exist on a few platforms:
```python
FAILED test_models.py::TestCarModel_151_SUBARU_ASCENT_LIMITED_2019::test_car_interface - AssertionError: 5799 != 0
FAILED test_models.py::TestCarModel_154_SUBARU_IMPREZA_LIMITED_2019::test_car_interface - AssertionError: 10480 != 0
```

@adeebshihadeh should we detect the presence of this message and set a `carParams.flags`?

jnewb1:
> Changed the panda branch to a temporary one so tests would run. It looks like this infotainment message does not exist on a few platforms:
> 
> ```python
> FAILED test_models.py::TestCarModel_151_SUBARU_ASCENT_LIMITED_2019::test_car_interface - AssertionError: 5799 != 0
> FAILED test_models.py::TestCarModel_154_SUBARU_IMPREZA_LIMITED_2019::test_car_interface - AssertionError: 10480 != 0
> ```
> 
> @adeebshihadeh should we detect the presence of this message and set `carParams.flags` since the message exists on only some global cars?
> 
> I actually found a Forester with the message and a Forester without the message. The one that has it is 2021 while the other is 2019 (both "global"), so this can even vary per platform. Let's merge as global gen 2 only for now, then detect this in a new PR. @jnewb1 you want to do that?
> 
> ffb24d475c36bab4|2023-04-12--16-20-59 f3b229479041308c|2023-04-12--17-06-26

Sure. So you mean add a panda safety flag for infotainment status, and add it to the panda once we verify this message has been sent at least once from the camera? Then overwrite it from then on?

adeebshihadeh:
> @adeebshihadeh should we detect the presence of this message and set `carParams.flags` since the message exists on only some global cars?

Yes, and the panda side can just always block right?

> I actually found a Forester with the message and a Forester without the message. The one that has it is 2021 while the other is 2019 (both "global"), so this can even vary per platform. Let's merge as global gen 2 only for now, then detect this in a new PR. @jnewb1 you want to do that?

Any reason to wait on this? I'd rather do them all now.

sshane:
Times to seeing this CAN message from the first CAN message, looks reasonable:

Screenshot from 2023-04-14 23-24-32

Here are the cars where this message exists:

```python
{'SUBARU FORESTER 2019': {True: 4071}, 'SUBARU OUTBACK 6TH GEN': {True: 2453, False: 1}, 'SUBARU IMPREZA LIMITED 2019': {False: 1694}, 'SUBARU ASCENT LIMITED 2019': {False: 2847, True: 50}, 'SUBARU IMPREZA SPORT 2020': {True: 1052}, 'SUBARU LEGACY 7TH GEN': {True: 5}, 'SUBARU OUTBACK 2018 - 2019': {False: 3}}
```

martinl:
I tested the PR with 2018 Crosstrek (does not have the infotainment alert message) - everything works as expected

Route: `8de015561e1ea4a0|2023-04-16--17-12-54`

sshane:
There appears to be some bits that some cars set (and actively change), I'll figure out if they're important to forward.

Forester: 
image

Outback:
image

sshane:
So the static signal appeared to not change any visual UI elements, but the one that I've seen change can turn the lane lines blue on the dash. However from what I've seen, the stock system doesn't use this and only opts to show a green steering wheel.

No LKAS (stock system usually never shows this): PXL_20230417_222254957 MP

LKAS + lines visible: PXL_20230417_222310100

LKAS + no lines visible: PXL_20230417_222332844

LKAS active + lines visible: PXL_20230417_222407423

Lines blue (changes signal at `2`): PXL_20230417_222429802

Lines blue + LKAS active: PXL_20230417_222450061

PR #27833, opened 2023-04-07
cabana: support multiple tabs in ChartsView

deanlee:
click "new tab" button in toolbar to create new tab

Screenshot from 2023-04-08 02-52-35


sshane:
Can you put the new tab button on the right? Think that feels better. It's also different than your screenshot which is odd

Screenshot from 2023-04-07 14-13-16

deanlee:
done

PR #27834, opened 2023-04-07
Hyundai: rename GEARS feature list and add test

sshane:
Saw someone accidentally put their CAN FD car in a gear list. Also FEATURES is no longer features! Just gears


adeebshihadeh:
This test doesn’t seem to have any implications whether is passes or fails. If that’s the case, this mistake should just be refactored out instead of adding a test.

sshane:
I plan on detecting most of these feature lists, which we'll want to do to improve fuzzy FP (since the platform codes are shared between openpilot platforms). So they'll eventually be gone and we can remove the test, but were you looking for a shorter term refactor?

PR #27841, opened 2023-04-08
Multiplying by one doesn't do anything

rockerrock137:
Save a bit of cpu

**Verification** [](Explain how you tested the refactor for regressions.)

x * 1 = x

### DIFF
--- a/selfdrive/controls/lib/vehicle_model.py
+++ b/selfdrive/controls/lib/vehicle_model.py
@@ -103,7 +103,7 @@ class VehicleModel:
       Steering wheel angle [rad]
     """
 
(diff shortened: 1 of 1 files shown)


rockerrock137:
In dyn_ss_sol, may also consider using numpy.linalg.lstsq instead of numpy.linalg.solve to solve the linear equation. lstsq is more efficient and numerically stable, especially when dealing with ill-conditioned matrices.

deanlee:
> **Verification** [](Explain how you tested the refactor for regressions.)
> 
> x * 1 = x

I'm guessing `x*0.1` is used to force floating point arithmetic. The correct verification should be:

`x * 1 / n = x *1.0 / n`

PR #27847, opened 2023-04-08
CAR.IONIQ_5 update for Ecu.fwdRadar values

markalan020:
missing firmware version for car. Including in here. From dongle id 5608f42aaece1463. Confirmed working on fork installed to comma 3


sshane:
Your FW versions are already in openpilot, this is just a different version of the response.

PR #27854, opened 2023-04-09
Add FPv2: Subaru Forester 2020

martinl:
This PR adds missing FW values for Subaru Forester 2020

Route: `cf42f4c081ad79b1|2023-04-09--11-29-48`


sshane:
This is very suspicious, on the route you posted the fwdCamera returned `b'\x00\x00e^\x1f@ !'`, but on recent routes it's returning  `b'\x00\x00e^\x00\x00\x00\x00'`.

PR #27858, opened 2023-04-10
cabana: Hack for WSL2 QDockWidget floating bug

MoreTore:
This is a bug fix/workaround for an undocumented QT bug when using QDockWidget where floating widgets do not respond to mouse QEvents. 

https://github.com/commaai/openpilot/discussions/26091#discussioncomment-5563395

Note that this will remove the undock button but still allows the user to drag the QDockWidget title bar to undock. Once the widget is floating the dock button is shown. This is done because pressing the floatButton will lead to the behavior of the widget not responding to any QEvents.

### DIFF
--- a/tools/cabana/mainwin.cc
+++ b/tools/cabana/mainwin.cc
@@ -193,6 +193,27 @@ void MainWindow::createDockWindows() {
   video_dock->setFeatures(QDockWidget::DockWidgetMovable | QDockWidget::DockWidgetFloatable);
   video_dock->setWidget(video_splitter);
   addDockWidget(Qt::RightDockWidgetArea, video_dock);
+
+#ifdef WSL2
+  dock->setFeatures(QDockWidget::DockWidgetMovable);
+  video_dock->setFeatures(QDockWidget::DockWidgetMovable);
+  QObject::connect(dock, &QDockWidget::topLevelChanged, this, &MainWindow::onDockWidgetTopLevelChanged);
+  QObject::connect(video_dock, &QDockWidget::topLevelChanged, this, &MainWindow::onDockWidgetTopLevelChanged);
+#endif
+}
+
+// This is a workaround for a bug in WSL2
+void MainWindow::onDockWidgetTopLevelChanged(bool topLevel) {
+  QDockWidget *dockWidget = qobject_cast<QDockWidget *>(sender());
+  if (dockWidget) {
+    if (topLevel) {
+      // the floating button breaks mouse tracking when docked so I remove it when docked and add it back when floating
+      dockWidget->setFeatures(QDockWidget::DockWidgetMovable | QDockWidget::DockWidgetFloatable); 
+      dockWidget->setTitleBarWidget(nullptr); // this fixes the mouse tracking issue when floating
+    } else {
+      dockWidget->setFeatures(QDockWidget::DockWidgetMovable);
+    }
+  }
 }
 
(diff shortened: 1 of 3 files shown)


deanlee:
It's not good if the undock button disappears by default.

I suggest to use a conditional inclusion, only do this in wsl2 environment.

MoreTore:
> It's not good if the undock button disappears by default.
> 
> I suggest to use a conditional inclusion, only do this in wsl2 environment.

Kk no problem.

tools/cabana/mainwin.cc
```diff
@@ -193,6 +193,27 @@ void MainWindow::createDockWindows() {
...
+}
+
+// This is a workaround for a bug in WSL2
+void MainWindow::onDockWidgetTopLevelChanged(bool topLevel) {
+  QDockWidget *dockWidget = qobject_cast<QDockWidget *>(sender());
+  if (dockWidget) {
+    if (topLevel) {
+      // the floating button breaks mouse tracking when docked so I remove it when docked and add it back when floating
+      dockWidget->setFeatures(QDockWidget::DockWidgetMovable | QDockWidget::DockWidgetFloatable); 
+      dockWidget->setTitleBarWidget(nullptr); // this fixes the mouse tracking issue when floating
```
deanlee, line 212: 1. I'm curious why setting `setTitleBarWidget(nullptr)` does the trick. the origin `titlebarWidget` is already nullptr, Calling this function just causes `QDockWidget` to reset window flags internally. am I right?
2. if it works, Can we delete other codes like set different dock features on WSL2, Just set titlebar widget to null to let qdockwidget reset window flags after toplevelchanged?

deanlee: If this issue only occurs after restarting cabana, the problem may be restoring state before showing the main window.  would a delayed call to `restoreState(settings.window_state);` fix it? : `QTimer::singleShot(0, [this](){restoreState(settings.window_state);});`

deanlee:
If this issue only occurs after restarting cabana, the problem may be restoring state before showing the main window. would a delayed call to restoreState(settings.window_state); fix it? : QTimer::singleShot(0, [this](){restoreState(settings.window_state);});

MoreTore:
> If this issue only occurs after restarting cabana, the real problem might be restoring state before showing the main window. would a delayed call to `restoreState(settings.window_state)` fix it ? : `QTimer::singleShot(0, [this](){restoreState(settings.window_state);});`

No, what I meant is that if I undock the widget, the widget is permanently broken even after restarting Cabana. The only way to get the widget to re-dock is to delete the settings file.

tools/cabana/mainwin.cc
MoreTore: setting setTitleBarWidget(nullptr); fixes the issue when dragging the titlebar to undock/re-dock but if I press the float button it causes the same old behaviour of not responding to events.
> This is done because pressing the floatButton will lead to the behavior of the widget not responding to any QEvents.

MoreTore: the way I figured this workaround out was by looking at the qt QDockWidget and i read that there was some special behavior with the title bar. 

from https://doc.qt.io/qt-5/qdockwidget.html#QDockWidget

> If a title bar widget is set, [QDockWidget](https://doc.qt.io/qt-5/qdockwidget.html) will not use native window decorations when it is floated.

PR #27859, opened 2023-04-10
debug: print live torque params

MoreTore:
the only way to view live torque params over tmux

### DIFF
--- a/selfdrive/debug/print_live_torque_params.py
+++ b/selfdrive/debug/print_live_torque_params.py
@@ -1,0 +1,26 @@
+from common.params import Params
+from cereal import log, messaging
+  
+def get_cached_torque_params():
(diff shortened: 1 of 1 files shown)


MoreTore:
The checks seem to be broken?

sshane:
https://github.com/commaai/openpilot/blob/master/selfdrive/debug/dump.py should be what you want

MoreTore:
> https://github.com/commaai/openpilot/blob/master/selfdrive/debug/dump.py should be what you want, but we recommend using PlotJuggler live streaming for this

yea I do use pj but others have trouble getting the whole openpilot env setup. I mainly made this for someone to get the cached params while the car is off and logging is disabled. dump.py  can't do that from what I can see.

PR #27860, opened 2023-04-10
Delay Shutdown: delay shutdown for 5 minutes after ignition turned off.

### LINKED ISSUE #27856: Delay shutdown for 5 minutes after ignition is turned off.
When the car is started and the C3 starts up, the device should not shut off so quickly after turning the car off. I was making some changes to my branch and when I tried to check for updates my device shut down since the car had not been run for long enough. 

Disable low battery shut down for 5 minutes after the ignition is turned off.

MoreTore:
resolve https://github.com/commaai/openpilot/issues/27856

- [x] Test

### DIFF
--- a/selfdrive/thermald/tests/test_power_monitoring.py
+++ b/selfdrive/thermald/tests/test_power_monitoring.py
@@ -15,7 +15,7 @@ def mock_sec_since_boot():
 with patch("common.realtime.sec_since_boot", new=mock_sec_since_boot):
   with patch("common.params.put_nonblocking", new=params.put):
     from selfdrive.thermald.power_monitoring import PowerMonitoring, CAR_BATTERY_CAPACITY_uWh, \
-                                                    CAR_CHARGING_RATE_W, VBATT_PAUSE_CHARGING
+                                                    CAR_CHARGING_RATE_W, VBATT_PAUSE_CHARGING, DELAY_SHUTDOWN_TIME_S
 
 TEST_DURATION_S = 50
 GOOD_VOLTAGE = 12 * 1e3
@@ -116,10 +116,9 @@ class TestPowerMonitoring(unittest.TestCase):
           self.assertFalse(pm.should_shutdown(ignition, True, start_time, False))
       self.assertTrue(pm.should_shutdown(ignition, True, start_time, False))
 
-  # Test to check policy of stopping charging when the car voltage is too low
   def test_car_voltage(self):
     POWER_DRAW = 0 # To stop shutting down for other reasons
-    TEST_TIME = 100
+    TEST_TIME = 350
     VOLTAGE_SHUTDOWN_MIN_OFFROAD_TIME_S = 50
     with pm_patch("VOLTAGE_SHUTDOWN_MIN_OFFROAD_TIME_S", VOLTAGE_SHUTDOWN_MIN_OFFROAD_TIME_S, constant=True), pm_patch("HARDWARE.get_current_power_draw", POWER_DRAW):
       pm = PowerMonitoring()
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Seems reasonable to me, but this should go in the power monitor itself. @robbederks any objections?

MoreTore:
> Seems reasonable to me, but this should go in the power monitor itself. @robbederks any objections?

Do you want me to change it?

robbederks:
Yes let's put it in the should shutdown function, other than that it looks good to me.

Let's maybe also add a unit test for this

selfdrive/thermald/tests/test_power_monitoring.py
```diff
@@ -116,11 +116,12 @@ def test_max_time_offroad(self):
           self.assertFalse(pm.should_shutdown(ignition, True, start_time, False))
       self.assertTrue(pm.should_shutdown(ignition, True, start_time, False))
 
-  # Test to check policy of stopping charging when the car voltage is too low
   def test_car_voltage(self):
-    POWER_DRAW = 0 # To stop shutting down for other reasons
-    TEST_TIME = 100
+    DELAY_SHUTDOWN_TIME_S = 300
```
robbederks, line 120: Let's not redefine that constant here, but rather just import it

selfdrive/thermald/tests/test_power_monitoring.py
```diff
@@ -177,6 +179,31 @@ def test_harness_connection(self):
         if i % 10 == 0:
           self.assertFalse(pm.should_shutdown(ignition, False, ssb, False))
       self.assertFalse(pm.should_shutdown(ignition, False, ssb, False))
+      
+  def test_delay_shutdown_time(self):
+    DELAY_SHUTDOWN_TIME_S = 300
+    with pm_patch("HARDWARE.get_current_power_draw", 0), \
+         pm_patch("DELAY_SHUTDOWN_TIME_S", DELAY_SHUTDOWN_TIME_S, constant=True):
```
robbederks, line 186: No need to patch this I assume

selfdrive/thermald/tests/test_power_monitoring.py
```diff
@@ -177,6 +179,31 @@ def test_harness_connection(self):
...
+      ignition = False
+      in_car = True
+      start_time = ssb  
+      offroad_timestamp = start_time
+      started_seen = False
+  
+      while start_time <= offroad_timestamp + DELAY_SHUTDOWN_TIME_S:
+        pm.calculate(VOLTAGE_BELOW_PAUSE_CHARGING, ignition)
+        should_shutdown = pm.should_shutdown(ignition, in_car, offroad_timestamp, started_seen)
+        self.assertFalse(should_shutdown, f"Should not shutdown before {DELAY_SHUTDOWN_TIME_S} seconds offroad time")
```
robbederks, line 199:
```suggestion
        self.assertFalse(pm.should_shutdown(ignition, in_car, offroad_timestamp, started_seen), f"Should not shutdown before {DELAY_SHUTDOWN_TIME_S} seconds offroad time")
```

selfdrive/thermald/tests/test_power_monitoring.py
```diff
@@ -177,6 +179,31 @@ def test_harness_connection(self):
...
+        start_time += 1  # Simulate the passage of time
+      started_seen = True
```
robbederks, line 201: `started_seen` should be high for the whole test period, otherwise the shutdown delay logic is completely masked by it!

MoreTore: Yea I realized this afterwards.

MoreTore:
@robbederks Everything should be in order now.

MoreTore:
@robbederks any chance this can be added to master-ci soon? This issue is extremely annoying. Why am I the only one complaining about this? https://youtube.com/shorts/dYZdq0esOzw?feature=share

robbederks:
Make sure that the test you wrote actually runs. Just looking at the code I don't see how the while loop would ever complete, so I'm relatively sure it wouldn't pass.

From what I can tell on mobile it looks like the whole test function is defined at the wrong indentation?

MoreTore:
> Make sure that the test you wrote actually runs. Just looking at the code I don't see how the while loop would ever complete, so I'm relatively sure it wouldn't pass.
> 
> From what I can tell on mobile it looks like the whole test function is defined at the wrong indentation?

It runs and passes as it is.

robbederks:
Ah yes, forgot how this test suite worked with the mocked SSB.

Lgtm then.

Issue #27703, opened 2023-03-27
ui: ignore steering icon double taps


royjr:
### Describe the bug

When tapping the steering icon in the upper right corner to switch between chill and experimental modes, a second tap is sometimes registered causing the mode to change twice rapidly which results in the mode basically staying the same.

This is most likely to happen while driving while the device and user are in motion. A small button tap delay such as 0.5s may help with this.

### Provide a route where the issue occurs

1a5d045d2c531a6d

### openpilot version

master

sshane:
Not sure exactly how the touch event works, but you might be able to override the event handler and do nothing if a double tap is registered

MoreTore:
I was just working on something similar in another project. I will make a pr to fix this.

PR #27861, opened 2023-04-10
ignore double tap on experimental button

MoreTore:
https://github.com/commaai/openpilot/issues/27703


deanlee:
Is there any special reason you need to `installeventFilter` to self  instead of just overloading `event()`?
In addition, this depends on the time interval of db click, this is probably too short and out of our control.

MoreTore:
> Is there any special reason you need to `installeventFilter` to self instead of just overloading `event()`? 

your right. I changed it

>In addition, this depends on the time interval of db click, it is probably too short and out of our control.

So far it seems to have fix the issue.

deanlee:
how about use a timer? https://github.com/commaai/openpilot/pull/27865.

MoreTore:
> how about use a timer? #27865.

that does seem better.

PR #27862, opened 2023-04-10
add steering angle and vego to torque controller

MoreTore:
add steering angle and vego to torque controller. This is required for some unofficially supported vehicles and makes it easier to merge commits.


MoreTore:
@adeebshihadeh there are over 300 Openpilot customers who use this. Why are you closing this

PR #27865, opened 2023-04-10
ui: prevent constant clicking on experimental button

### LINKED ISSUE #27703: ui: ignore steering icon double taps

deanlee:
resolve https://github.com/commaai/openpilot/issues/27703


selfdrive/ui/qt/onroad.cc
```diff
@@ -180,12 +180,13 @@ ExperimentalButton::ExperimentalButton(QWidget *parent) : QPushButton(parent) {
...
-  params = Params();
   engage_img = loadPixmap("../assets/img_chffr_wheel.png", {img_size, img_size});
   experimental_img = loadPixmap("../assets/img_experimental.svg", {img_size, img_size});
 
-  QObject::connect(this, &QPushButton::toggled, [=](bool checked) {
-    params.putBool("ExperimentalMode", checked);
+  QObject::connect(this, &QPushButton::toggled, [this](bool checked) {
+    setEnabled(false);
+    Params().putBool("ExperimentalMode", checked);
+    QTimer::singleShot(1000, [this]() { setEnabled(true); });
```
sshane, line 189: How about 100ms? I feel like the double clicks are tens of ms apart, not on the order of hundres

sshane: yep, a second button press is registered about 20 ms after, I've reproduced this a few times now.

deanlee: > Though, it would be even better to debounce all UI buttons generically, can we do that?
y, think we can do it in another pr.

sshane: there's a race condition with the onroad ExperimentalButton. the button signal `toggled` causes a param write, which controlsd reads and sends in `controlsState`, which `ExperimentalButton` reads to set checked...which writes the param again. So if you get a toggled signal, and then an old controlsState packet it will write twice!

you can  also write the param through another mean and ExperimentalButton will write it again (the `setChecked` call sends a `toggled` signal)

deanlee: Where is the parameter ExperimentalModeConfirmed set?

sshane: controls.h `bool confirmed = store_confirm && params.getBool(key + "Confirmed");`

deanlee:
still use 1000ms to prevent check ExperimentalMode in updatestate before timeout,  @sshane Is  one second timeout enough for this? should we increase the timeout to make sure the mode is switched before timeout?

selfdrive/ui/qt/onroad.cc
```diff
@@ -180,12 +180,14 @@ ExperimentalButton::ExperimentalButton(QWidget *parent) : QPushButton(parent) {
...
   engage_img = loadPixmap("../assets/img_chffr_wheel.png", {img_size, img_size});
   experimental_img = loadPixmap("../assets/img_experimental.svg", {img_size, img_size});
 
-  QObject::connect(this, &QPushButton::toggled, [=](bool checked) {
-    params.putBool("ExperimentalMode", checked);
+  QObject::connect(this, &QPushButton::clicked, [this](bool checked) {
+    setEnabled(false);
+    Params().putBool("ExperimentalMode", checked);
+    // prevent double click and skip check for experimental mode in updateState before timeout.
+    QTimer::singleShot(1000, [this]() { setEnabled(true); });
```
sshane, line 190:
```suggestion
    QTimer::singleShot(100, [this]() { setEnabled(true); });
```

yes 100ms sounds good. the fact that this rarely happens means that the param must be getting read in one control frame most of the time

selfdrive/ui/qt/onroad.cc
```diff
@@ -180,12 +180,14 @@ ExperimentalButton::ExperimentalButton(QWidget *parent) : QPushButton(parent) {
   setFixedSize(btn_size, btn_size);
   setCheckable(true);
 
-  params = Params();
```
sshane, line 183: why delete?

deanlee: sorry my bad

selfdrive/ui/qt/onroad.cc
```diff
@@ -197,7 +200,9 @@ void ExperimentalButton::updateState(const UIState &s) {
   setVisible(cs.getEngageable() || cs.getEnabled());
 
   // button is "checked" if experimental mode is enabled
-  setChecked(sm["controlsState"].getControlsState().getExperimentalMode());
+  if (isEnabled()) {
+    setChecked(sm["controlsState"].getControlsState().getExperimentalMode());
+  }
```
sshane, line 205: What if we always update from controlsState and disallow clicking to change the checked state? Then we make no assumptions about how long the round trip from setting param to controlsd reading it, to it sending it out with cereal?

deanlee: In my opinion, this experimental button may not belong here at all. All other icons in the HUD cannot be clicked, only this one can, and there is no visual difference. This is a very strange thing, and users who are not familiar with it would not understand its purpose. Perhaps such a button should only exist in the settings interface.

deanlee:
The new commit made the following changes:
1. removed the checkable state of the button, use a member variable `experimental_mode` to save the current mode, this greatly simplifies the logic for handling clicks.
2. read params in changeMode to determine whether it is possible to switch modes, instead of doing two slow I/O operations every time 'updateState' is called.
3. the timer is no longer used to prevent users from repeatedly clicking because even if the user clicks repeatedly, the value write to param 'ExperimentalMode' will not change.

selfdrive/ui/qt/onroad.cc
```diff
@@ -174,43 +174,41 @@ void OnroadAlerts::paintEvent(QPaintEvent *event) {
...
-    params.putBool("ExperimentalMode", checked);
-  });
+void ExperimentalButton::changeMode() {
+  auto params = Params();
+  const auto cp = (*uiState()->sm)["carParams"].getCarParams();
+  const bool experimental_mode_available = cp.getExperimentalLongitudinalAvailable() ? params.getBool("ExperimentalLongitudinalEnabled") : cp.getOpenpilotLongitudinalControl();
+  bool can_change = experimental_mode_available && params.getBool("ExperimentalModeConfirmed");
+  if (can_change) {
+    params.putBool("ExperimentalMode", !experimental_mode);
+  }
```
sshane, line 194: Can we make a global variable or helper function to consolidate all of this experimental/alpha longitudinal logic? I think we have very similar logic in settings.cc as well as here which is pretty confusing to keep track of

deanlee: How about implementing it in another PR? 
Think it would be better to make a more comprehensive implementation instead of just targeting this specific functionality.For example, adding a member variable "struct settings" to `uistate` to track all changes in settings. This way, most calls to the `param.read` in the UI can be eliminated. It also allows for real-time switching of states, such as between miles and kilometers.

selfdrive/ui/qt/onroad.cc
```diff
@@ -174,43 +174,41 @@ void OnroadAlerts::paintEvent(QPaintEvent *event) {
...
   engage_img = loadPixmap("../assets/img_chffr_wheel.png", {img_size, img_size});
   experimental_img = loadPixmap("../assets/img_experimental.svg", {img_size, img_size});
+  QObject::connect(this, &QPushButton::clicked, this, &ExperimentalButton::changeMode);
+}
 
-  QObject::connect(this, &QPushButton::toggled, [=](bool checked) {
-    params.putBool("ExperimentalMode", checked);
-  });
+void ExperimentalButton::changeMode() {
+  auto params = Params();
```
sshane, line 188: any reason to create params every time?

sshane:
looks mostly good to me, @incognitojam can you check it out and merge? Only thing is the duplicate logic, but that's not new with this PR and we can clean it up soon in another

PR #27867, opened 2023-04-11
hyundai-can long control improvements

pfeiferj:
**Description**
Calculate jerk from the acceleration plan and use as the lower limit value when sending accel commands to hyundai can vehicles.

Using small min values for the jerk increases the resolution of the acceleration that the car provides. This allows the car to better fit the acceleration to the plan by making smaller adjustments to the acceleration to maintain/hit the target acceleration.

However, with too small of values for either the min or max jerk the car responds to changes in acceleration very sluggishly. Thus when we are requesting rapid changes to acceleration we need larger jerk values.

To achieve a good balance of resolution to responsiveness we can determine what the jerk in the planned acceleration is between the last acc command and the current command and use that as the basis of the values we request from the car. This allows a dynamic resolution that matches the current demands of the plan.

**Verification**
I've been driving using this change for the last month or so and find the acceleration to be noticeably smoother. However, I'm marking as a draft for now as I'm not entirely confident in the behavior when stopping and accelerating from a stop. I believe it's similar or better in behavior at stopping speeds as the current method, but most of my driving does not have stop and go driving to fully test with. Some tuning may still be necessary at very low speeds.

I also suspect that using this method would improve the feel of acceleration on can-fd vehicles, but do not have one to test with so the change currently is only on the can controls.

### DIFF
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -173,9 +173,15 @@ class CarController:
               self.last_button_frame = self.frame
 
       if self.frame % 2 == 0 and self.CP.openpilotLongitudinalControl:
-        # TODO: unclear if this is needed
-        jerk = 3.0 if actuators.longControlState == LongCtrlState.pid else 1.0
-        can_sends.extend(hyundaican.create_acc_commands(self.packer, CC.enabled, accel, jerk, int(self.frame / 2),
+        # calculate jerk from plan, give a small offset for the upper limit for the cars ecu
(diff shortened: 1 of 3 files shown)


james5294:
I have been testing this for the last few weeks and can confirm a noticeable improvements on my 2022 Hyundai Sonata with both acceleration from a stop and braking.. Both being smoother.

pfeiferj:
Makes sense, never really got proper validation done anyways. It probably would also be a bit more complexity than even this added to get it to be truly correct as well

PR #27868, opened 2023-04-11
Subaru: lower steering pressed threshold

sshane:


sshane:
will re-open once ready to merge

PR #27870, opened 2023-04-11
[Car Port] 2022 Honda HR-V e:HEV

eFiniLan:
co-work with @somkanae

opendbc PR: https://github.com/commaai/opendbc/pull/812

**Checklist**
- [x] route with openpilot: 1e4baee1aa2687a0|2023-04-11--11-42-42
- [x] route with stock system: 1e4baee1aa2687a0|2023-04-11--12-01-15


adeebshihadeh:
@sshane anything blocking this? looks like a straightforward car port

sshane:
The user of this Thai car reports they get "Hill Descent Error" that I haven't had time to look into at all. The Brazilian HR-V also has issues in experimental mode (it fails to respond to our brake command at low speed), so I didn't want to merge any global HR-V yet until I looked into it more.

adeebshihadeh:
We shouldn’t block merging ports on experimental longitudinal control. Let’s just blacklist them from that and include a good comment.

sshane:
I'll try to find a way to detect it, they also are slightly different in another way I forgot about, so probably a new platform will work.

sshane:
thanks for the PR, but looks like this is also in https://github.com/commaai/openpilot/pull/27915, with proper detection

PR #27872, opened 2023-04-11
tiguan 2021 ehybrid fingerprint

jonjander:
tiguan 2021 ehybrid fingerprint


selfdrive/car/volkswagen/values.py
```diff
@@ -793,6 +797,7 @@ def init_make(self, CP: car.CarParams):
       b'\xf1\x872Q0907572Q \xf1\x890342',
       b'\xf1\x872Q0907572R \xf1\x890372',
       b'\xf1\x872Q0907572T \xf1\x890383',
+      b'\xf1\x872Q0907572T \xf1\x890383',
```
jonjander, line 800: Duplicated row

jonjander:
> can you provide the dongle id that these came from? it will be in device settings on your comma three

ff0d015e712fd2a4

selfdrive/car/volkswagen/values.py
sshane, line 800:
```suggestion
```

sshane, line 800:
```suggestion
```

PR #27873, opened 2023-04-11
cabana: speed up the sparkline rendering with opengl viewport

deanlee:
before (raster based painting. we turned off anti-aliasing when drawing Sparkline due to performance issues. )
   

after (use opengl with Antialiasing,  it's10x~20x faster than before)
 Screenshot from 2023-04-11 18-28-00


deanlee:
closed.  https://github.com/commaai/openpilot/pull/27925 is a more perfect solution.

PR #27875, opened 2023-04-11
VW MQB: add missing 2021 TRANSPORTER T6.1 FW

Saber422:
route name: f9fa13485e6df75c|2023-04-10--08-21-09--0


selfdrive/car/volkswagen/values.py
```diff
@@ -825,9 +826,11 @@ def init_make(self, CP: car.CarParams):
     ],
     (Ecu.srs, 0x715, None): [
       b'\xf1\x872Q0959655AE\xf1\x890506\xf1\x82\02316170411110411--04041704161611152S1411',
```
sshane, line 828:
```suggestion
```

PR #27880, opened 2023-04-12
fingerprinting: log FW query time

sshane:
logs total VIN time, present ECU addrs, and FW query. Optima is ~2 seconds for all that

from our tests the theoretical maximum value is 0.7s FW + 1.0s VIN + 0.8s present ECUs = 2.5s


sshane:
@adeebshihadeh do you know of any other way to derive the total fingerprinting time?

adeebshihadeh:
Let's just add to carParams.

sshane:
Ah yeah, we don't read the logMessage service in the pipeline yet which we'd need for this without it being in the carParams. We also will want `cached` in the carParams as well since we need to filter those out for the metrics.

selfdrive/car/car_helpers.py
```diff
@@ -111,6 +113,8 @@ def fingerprint(logcan, sendcan, num_pandas):
     exact_fw_match, fw_candidates, car_fw = True, set(), []
     cached = False
 
+  fingerprinting_time = sec_since_boot() - start_time
```
sshane, line 116: TODO: decide if we want to log VIN/FW/CAN fingerprinting separately

sshane:
Just want to make sure since we're adding cereal fields. Think we need `cached`, since we don't parse logMessage in the pipeline. Should we start to move everything into carParams though? Or perhaps a struct related to fingerprinting so it doesn't get too messy?

adeebshihadeh:
Does this really need to be in the qlog? I'd just start cloudlogging it.

sshane:
> Let's just add to carParams.

That's what I wanted to do, but I realized we don't parse `logMessage` in the pipeline, so can't be added to tsdb metrics.

sshane:
merging without cereal changes, useful for manual debugging

PR #27908, opened 2023-04-14
cabana: fix message view performance issues after #27901

deanlee:
Issues:
1. the CPU usage **increased by 20%** after #27901.  
2. The new messagelist is not compatible with the state saved by the previous settings.

fixed:
1. Optimize the sizehint() and paint() , the cpu usage has been reduced by about 20%, basically the same as the previous version.
2. rename setting key from 'message_header_state' to 'msg_header_state' to avoid the issues with incompatible settings file.

[updated]
merged another performance related pr: remove copies in BinaryViewModel::UpdateState

### DIFF
--- a/tools/cabana/settings.cc
+++ b/tools/cabana/settings.cc
@@ -30,7 +30,7 @@ void Settings::save() {
   s.setValue("geometry", geometry);
   s.setValue("video_splitter_state", video_splitter_state);
   s.setValue("recent_files", recent_files);
-  s.setValue("msg_header_state", message_header_state);
+  s.setValue("message_header_state_v2", message_header_state);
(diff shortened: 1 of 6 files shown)


tools/cabana/settings.cc
```diff
@@ -30,7 +30,7 @@ void Settings::save() {
   s.setValue("geometry", geometry);
   s.setValue("video_splitter_state", video_splitter_state);
   s.setValue("recent_files", recent_files);
-  s.setValue("message_header_state", message_header_state);
+  s.setValue("msg_header_state", message_header_state);
```
sshane, line 33: `message_header_state_v2`?

PR #27910, opened 2023-04-14
cabana: choose camera type in open route dialog

deanlee:
added a combobox to choose camera type:

fallback to `->roadCamera->qCamera` If the specified video is not found in route.

Screenshot from 2023-04-14 15-05-48
Screenshot from 2023-04-14 15-05-26

### DIFF
--- a/tools/cabana/route.cc
+++ b/tools/cabana/route.cc
@@ -5,7 +5,6 @@
 #include <QLabel>
 #include <QMessageBox>
 #include <QPushButton>
-#include <QRadioButton>
 
 #include "tools/cabana/streams/replaystream.h"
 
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Can we make this a live dropdown?

adeebshihadeh:
I was thinking around the video player somewhere so you can switch at any time.

deanlee:
reverted changes.

> I was thinking around the video player somewhere so you can switch at any time.

It is not easy for the current implementation to support live switch. 
replay does not support dynamic video switching. we can only load all videos and push theme in replay to allow the cameraview  switch between different videos. this will causes high cpu usage. maybe we can implement this feature later after we have a better solution?

adeebshihadeh:
Sounds good!

PR #27896, opened 2023-04-13
cabana: fix chart pixmap size on Hi-DPI screens

pd0wm:
TODO:
- [ ] Fix chart button border being the wrong size
- [ ] Track down segfault when dragging to different screen
- [ ] Fix drag/drop pixmap

image

PR #27912, opened 2023-04-14
cabana: fix chart high-dpi issues

deanlee:
related discussions: 
https://github.com/commaai/openpilot/pull/27896#issuecomment-1507262606

### DIFF
--- a/tools/cabana/chart/chart.cc
+++ b/tools/cabana/chart/chart.cc
@@ -405,7 +405,7 @@ QPixmap getBlankShadowPixmap(const QPixmap &px, int radius) {
   shadow.setDevicePixelRatio(dpr);
   shadow.fill(Qt::transparent);
   QPainter p(&shadow);
-  scene.render(&p, {QPoint(0, 0), shadow.size() / dpr}, item.boundingRect().adjusted(-radius, -radius, radius, radius));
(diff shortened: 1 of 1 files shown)


pd0wm:
Looks good on my macbook with hi-dpi now! Thanks for looking into it.

I noticed you didn't bring over the changes to invalidate the cache when dragging to a different monitor (in case you have a hi-dpi and non hi-dpi) monitor. When testing that I had a crash once, but wasn't able to reproduce it. Do you think we need a lock around the pixmap?

deanlee:
does this line fixed the mixed hi-dpi, non hi-dpi monitors issue?  https://github.com/commaai/openpilot/pull/27896/files#diff-75828cd93d8908bdfa27fe5ee42a91043738815e9945fc73fdf83b9f943b9321R45 :

`QObject::connect(window()->windowHandle(), &QWindow::screenChanged, [this]() { resetChartCache(); });`

pd0wm:
Yes &QWindow::screenChanged is emitted when dragging to a new monitor.

deanlee:
> edit: I also noticed now the button pixmaps are not hi-dpi, but this has always been the case. Maybe something we want to address as well?

I'll fix it in another pr.

PR #27919, opened 2023-04-15
Mazda: Fix BSM detection with blinkers

Jafaral:
depends: https://github.com/commaai/opendbc/pull/815


Jafaral:
@sshane , I got a couple if hits here: 8c521516df3527d8|2023-05-09--08-32-12--9

PR #24742, opened 2022-06-03
Driving Personality setting (also changes follow distance)

haraschax:

PR #27921, opened 2023-04-15
Add sliders to adjust openpilot behavior

MoreTore:
Example with dummy values.
Screenshot from 2023-04-15 16-29-36


adeebshihadeh:
We're working on something like this in #24742

MoreTore:
> We're working on something like this in #24742

My idea was more to be able to adjust how hard the car brakes and accelerates without a lead. For example in winter, my car spins out because OP tries to accelerate too much. It also brakes way too hard when lowering the crisis setpoint, I just want to to coast. This should be separate from following distance which should be adjusted through car buttons.

PR #27924, opened 2023-04-16
Ford: show FCW alert on IPC

incognitojam:
Handle fcw visual alert by showing Ford fcw alert on IPC (without sound, it's controlled by a separate CAN signal)

Demo with fcw_alert set to True:

https://user-images.githubusercontent.com/4038174/232290838-9bc0eb74-7e86-4f0f-a5e3-259870f2cf42.mp4


adeebshihadeh:
Is this what the car shows for a stock FCW? The text there seems to imply that it's actually assisting instead of just alerting.

incognitojam:
I'll see if I can find a different one

sshane:
The stock system definitely shows this alert even for just FCW, no AEB, quite confusing actually.

selfdrive/car/ford/fordcan.py
```diff
@@ -147,6 +147,7 @@ def create_acc_ui_msg(packer, main_on: bool, enabled: bool, hud_control, stock_v
   values = {
     **stock_values,
     "Tja_D_Stat": status,
+    "FcwVisblWarn_B_Rq": 1 if fcw_alert else 0,
```
sshane, line 150: Stock system also sets `FcwAudioWarn_B_Rq`, we want that?

sshane (approved):

incognitojam:
I think this same alert is used, even when the car won't brake autonomously, to pre-charge the brakes (so the driver is still "assisted")
Screenshot from 2023-05-18 20-19-24

sshane:
It looks like it treats FCW as pre-FCA, so it just has one alert. It shows when no pre-charging is done 
image

sshane:
Can we merge this?

sshane:
made it forward FCW from camera if it has one, our FCW is only an OR

PR #27932, opened 2023-04-17
cabana: add setting to choose preferred drag direction for new signals

pd0wm:
Got user feedback that default drag direction felt backwards to them, so made it possible to always start at the LSB. Also added option to always create BE/LE signals, since on most cars all signals are the same endianness.

Default setting is the same as current behavior.

2023-04-17-124030_322x447_scrot


deanlee:
how about add a link to the info label, click link to open setting dialog to let user know where to change the drag direction?
change
Screenshot from 2023-04-17 20-16-12
to something like this? "Drag-Select to create new signal. [change the drag direction](url)"

PR #27936, opened 2023-04-17
cabana: support direct streaming from panda over USB

pd0wm:
Refactor livestream class to more easily add new live stream types. Added panda stream by using abstraction from `selfdrive/boardd`.

Adding support for socketcan should be very simple too now, as requested in discussions https://github.com/commaai/openpilot/discussions/26091#discussioncomment-5514607.

TODO:
- [x] Testing
- [x] Set serial from cmd line arguments
- [x] Move logging into `LiveStream` base class


selfdrive/boardd/panda.cc
```diff
@@ -21,9 +21,6 @@ Panda::Panda(std::string serial, uint32_t bus_offset) : bus_offset(bus_offset) {
 
   hw_type = get_hw_type();
 
-  assert((hw_type != cereal::PandaState::PandaType::WHITE_PANDA) &&
```
pd0wm, line 24: Moved this check to boardd, as far as I'm aware this is not a limitation of the panda communication class.

PR #27938, opened 2023-04-17
cabana: startup stream chooser dialog

deanlee:
If the user run `./cabana` without any command-line parameters, popup a dialog to let the user choose the stream type.

todo: 
  - [x]  choose a dbc file to open

1. device (msg/zmq) :
Screenshot from 2023-04-18 03-49-13
1. panda (depends on https://github.com/commaai/openpilot/pull/27936) :
Screenshot from 2023-04-18 03-49-25
3. route (remote/local) :
Screenshot from 2023-04-18 03-49-34


pd0wm:
Nice! Played around with wizards for startup, but this is quite a bit cleaner.

It might be nice if the stream specific code lives in the various AbstractStream implementations. Then all the stream specific code lives in a single class/file.

tools/cabana/streams/abstractstream.h
```diff
@@ -87,5 +87,16 @@ class AbstractStream : public QObject {
   uint64_t last_event_ts = 0;
 };
 
+class AbstractOpenStreamWidget : public QWidget {
```
pd0wm, line 90: You could also add a static method to `AbstractStream` to build the wdiget. E.g. `static AbstractOpenStreamWidget* getOpenStreamWidget()`. Then the rest of the code only needs to know about a single stream class for each implementation.

pd0wm, line 90: You could also add a static method to `AbstractStream` to build the wdiget. E.g. `static AbstractOpenStreamWidget* getOpenStreamWidget()`. Then the rest of the code only needs to know about a single stream class for each implementation.

pd0wm:
I have some code already written to choose baud rates and CAN-FD settings for the panda stream. I'll make a PR once this is merged :)

PR #27941, opened 2023-04-17
cabana/cameraview: Scale the glViewPort by devicePixelRatio()

projectgus:
Fix for an issue I ran into using Cabana on a "hidpi" Linux system with Wayland+Plasma (Qt 5.15.9) where [devicePixelRatio() != 1.0](https://doc.qt.io/qt-5/qwindow.html#devicePixelRatio) and the video doesn't take up the full widget area.

Screenshot of the demo route showing the issue:

image

On the recommended Ubuntu 20.04 install I wasn't able to reproduce. It seems devicePixelRatio() is equal to 1.0 there even with scaling at 200% or 300%. It might be different if "Fractional Scaling" is enabled in GNOME (I couldn't make that option work in my test WM for some reason.)

I was going to enable this just for Linux, but it appears to also be recommended for Retina macOS:

> [However, in OpenGL pixels are always device pixels. For example, geometry passed to glViewport() needs to be scaled by devicePixelRatio().](https://doc.qt.io/qt-5/scalability.html#high-dpi-scaling-on-macos-and-io)

So I've turned it on unconditionally, with the assumption that on many systems `devicePixelRatio() == 1.0` so it won't change anything. I don't have a Mac or WSL system to verify if this either fixes or creates any issues there, though.

Thanks for developing Cabana and the comma.ai ecosystem, I really like this tool!

### DIFF
--- a/selfdrive/ui/qt/widgets/cameraview.cc
+++ b/selfdrive/ui/qt/widgets/cameraview.cc
@@ -114,6 +114,18 @@ CameraWidget::~CameraWidget() {
   doneCurrent();
 }
 
+// Qt uses device-independent pixels, depending on platform this may be
+// different to what OpenGL uses
+int CameraWidget::glWidth()
+{
+    return width() * devicePixelRatio();
+}
+
+int CameraWidget::glHeight()
+{
+  return height() * devicePixelRatio();
+}
+
 void CameraWidget::initializeGL() {
   initializeOpenGLFunctions();
 
@@ -188,7 +200,7 @@ void CameraWidget::availableStreamsUpdated(std::set<VisionStreamType> streams) {
 }
 
 void CameraWidget::updateFrameMat() {
-  int w = width(), h = height();
+  int w = glWidth(), h = glHeight();
 
   if (zoomed_view) {
     if (active_stream_type == VISION_STREAM_DRIVER) {
@@ -266,7 +278,7 @@ void CameraWidget::paintGL() {
 
   updateFrameMat();
 
-  glViewport(0, 0, width(), height());
+  glViewport(0, 0, glWidth(), glHeight());
   glBindVertexArray(frame_vao);
   glUseProgram(program->programId());
   glPixelStorei(GL_UNPACK_ALIGNMENT, 1);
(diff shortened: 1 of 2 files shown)


projectgus:
Failing benchmark seems to be fully unrelated, but I don't seem to have permission to retry the job.

pd0wm:
Ah good fix! I was running into this on MacOS as well. See https://github.com/commaai/openpilot/discussions/26091#discussioncomment-5646096

projectgus:
> Is everything still ok if you force a ratio with eg QT_SCALE_FACTOR=1 on a high-dpi display?

Just tried it out on Wayland and X11 (Ubuntu 20.04) and it seems to work as expected, the UI widgets all scale up or down in size and the video draws correctly in the available area.

> Did you test the openpilot ui as well?

Sorry, I did not... actually I don't own any comma.ai hardware, nor an Nvidia graphics card to run the simulator. My interest is using Cabana to analyse some different CAN logs for a totally different purpose.

pd0wm:
You can test the openpilot ui (`selfdrive/ui/ui`) on replay as well! No need for a simulator.

martinl:
current state on macOS:

macOS hidpi display (m1 MBP) `QT_SCALE_FACTOR=1.0 ./tools/cabana/cabana --demo` - too small to be usable

Screenshot 2023-04-18 at 21 46 49

macOS hidpi display (m1 MBP) `QT_SCALE_FACTOR=0.5 ./tools/cabana/cabana --demo` - usable

Screenshot 2023-04-18 at 21 40 11

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -114,6 +114,18 @@ CameraWidget::~CameraWidget() {
   doneCurrent();
 }
 
+// Qt uses device-independent pixels, depending on platform this may be
+// different to what OpenGL uses
+int CameraWidget::glWidth()
+{
```
adeebshihadeh, line 120:
```suggestion
int CameraWidget::glWidth() {
```

selfdrive/ui/qt/widgets/cameraview.cc
```diff
@@ -114,6 +114,18 @@ CameraWidget::~CameraWidget() {
...
+    return width() * devicePixelRatio();
+}
+
+int CameraWidget::glHeight()
+{
```
adeebshihadeh, line 125:
```suggestion
int CameraWidget::glHeight() {
```

PR #27942, opened 2023-04-17
replay framereader: Log in some unlikely failure paths

projectgus:
Very minor patch to add logging in some failure paths.

Context: I was debugging an unrelated behavior running Cabana on my system and added logging to these in case they were silently failing (they weren't!)

Thought it might be helpful to submit as a patch, anyhow.


deanlee:
nice

PR #27949, opened 2023-04-18
boardd: fix scons warning "Two different environments were specified"

deanlee:
This warning appeared after https://github.com/commaai/openpilot/pull/27936


adeebshihadeh:
It would be great if there was some way to make this fail in CI. We already set PYTHONWARNINGS=error, but it doesn’t seem to catch this.

PR #27952, opened 2023-04-18
cabana: socketcan support

pd0wm:
Test on a virtual socketcan device using.
```bash
sudo modprobe vcan
sudo ip link add dev vcan0 type vcan
sudo ip link set up vcan0
cangen vcan0 -g 1 -L 8 # Generate some fake data
```

Start cabana using `./cabana --socketcan vcan0` or use open stream dialog.

2023-08-23-070822_613x280_scrot

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -282,7 +282,7 @@ Export('envCython')
 
 # Qt build environment
 qt_env = env.Clone()
-qt_modules = ["Widgets", "Gui", "Core", "Network", "Concurrent", "Multimedia", "Quick", "Qml", "QuickWidgets", "Location", "Positioning", "DBus", "Xml"]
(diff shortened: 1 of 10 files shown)


martinl:
what would be needed to add support for USB connected black/red pandas on macOS (that does not have socketcan) - I see that Qt supports some CAN adapters https://doc.qt.io/qt-5/qtcanbus-backends.html and there is https://www.mac-can.com that also has plugin interface

martinl:
nvm, I already found that panda is supported via usb serial interface. I'll go test if it works with macOS

pd0wm:
Those are already supported! Just plug it in and select from the startup dialog.

Screenshot 2023-05-08 at 10 03 14

pd0wm:
Tested build on Linux, MacOS and C3.

@adeebshihadeh This is ready for review. I assume the failing CI is due to the changes in `install_ubuntu_dependencies.h`?

@deanlee Any thoughts?

deanlee:
lgtm!

PR #27953, opened 2023-04-18
cabana: fix displaying zero length messages

pd0wm:


pd0wm:
@deanlee any reason you didn't add support for zero length messages? There could be a variety of reasons why those are on the bus (e.g. wakeup, trigger response from different ECU).

PR #27954, opened 2023-04-18
scons: set warning as exception

deanlee:


adeebshihadeh:
Moved to #28004, thanks for getting the ball rolling on this one!

PR #27955, opened 2023-04-18
navd: fix scons warning ""Two different environments were specified"

deanlee:
fixed scons warning:

`scons: warning: Two different environments were specified for target /tmp/scons_cache/moc_files/moc_map_renderer.cc,`

### DIFF
--- a/selfdrive/navd/SConscript
+++ b/selfdrive/navd/SConscript
@@ -1,4 +1,4 @@
-Import('qt_env', 'arch', 'common', 'messaging', 'visionipc', 'cereal', 'transformations')
+Import('qt_env', 'arch', 'cache_dir', 'common', 'messaging', 'visionipc', 'cereal', 'transformations')
 
 base_libs = [common, messaging, cereal, visionipc, transformations, 'zmq',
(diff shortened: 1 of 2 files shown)


selfdrive/navd/SConscript
```diff
@@ -18,4 +18,6 @@ if arch in ['larch64', 'x86_64']:
   nav_src = ["main.cc", "map_renderer.cc"]
   qt_env.Program("map_renderer", nav_src, LIBS=qt_libs + ['common', 'json11'])
 
-  qt_env.SharedLibrary("map_renderer", ["map_renderer.cc"], LIBS=qt_libs + ['common', 'messaging'])
+  shared_env = qt_env.Clone()
+  shared_env['QT_MOCHPREFIX'] = cache_dir + '/moc_files/moc_shared_'
+  shared_env.SharedLibrary("map_renderer", ['map_renderer.cc'], LIBS=qt_libs + ['common', 'messaging'])
```
adeebshihadeh, line 23: i'd prefer to fix this more generally or rename the lib to something like `libmaprender`

deanlee: It seems that renaming the lib is not working. this warning is caused by the moc file. besides using a different moc file, I can't think of a better solution at the moment.

adeebshihadeh:
#28004

PR #27956, opened 2023-04-18
ecu addrs: check msg length

sshane:
Looks like this was hit once while testing on a branch, good to catch though. Needed for https://github.com/commaai/openpilot/pull/27697 to not regress fingerprinting if a remote frame happens to be on the bus


sshane:
This actually has happened quite a few times in the wild, removing my other filters. 88 times, mostly across a few dongles and mostly the Honda Civic (Bosch and Nidec).

```python
addr_scan_exc_affected_dongles {'ecca667110ff6812', '5cfa1ad8f35779d0', 'a3f025079c99dde7', 'b9a21c12423207c5', 'a9009cfb3ac0c3c3', '7f6cd31ea33923bd'}
addr_scan_exc_affected_platforms {'HONDA CIVIC (BOSCH) 2019', 'HYUNDAI SONATA 2020', 'HONDA ACCORD 2018', 'HONDA CIVIC 2016'}
```

Issue #27665, opened 2023-03-24
FPv2 improvements


sshane:
- [x] Consider extra ecus for present ecu brand matching - https://github.com/commaai/openpilot/pull/27688
  For CAN FD for example, only the two ECUs in the DB are being matched against the hyundai brand, when we could match more (like cornerRadar, adas, etc)
- [x] Put FW versions inside the config for lots of nice helpers in one place - https://github.com/commaai/openpilot/pull/27691 - started with https://github.com/commaai/openpilot/pull/31227
- [x] Consider bus when matching brands - https://github.com/commaai/openpilot/pull/27657
- [ ] Sort requests by OBD multiplexing mode, so brands with auto-generated auxiliary requests don't switch unnecessarily because of bad sorting - not really needed at this moment - https://github.com/commaai/openpilot/pull/31263
- [ ] Group first subaddress query into parallel query group (reduces total queries by 1) - https://github.com/commaai/openpilot/pull/28675
- [x] Only update OBD multiplexing mode when on bus % 1 == 1 - https://github.com/commaai/openpilot/pull/27669
- [x] Test that measures the time to query FW versions in the worse case. For each brand and overall. Fail when there is a large number of queries or OBD multiplexing mode switches (soon to be optimized) - mostly this: https://github.com/commaai/openpilot/pull/27753
- [ ] Batch tester present for finding present ecus (if required). We now send tester present to 79 addresses at once
- [x] ~Filter out addresses that are already on the bus before we start to query, they are likely broadcast addrs and we shouldn't extend our timeouts for them~
- [x] Break out of IsoTpParallelQuery loop once either all addresses are timed out or done. Right now, We break if all addrs are timed out or done. For example, we wouldn't break if addr A is timed out and not done, and addr B is done and not timed out - https://github.com/commaai/openpilot/pull/27764
- [x] Only extend addr timeout in IsoTpParallelQuery if we got a valid consecutive frame - https://github.com/commaai/openpilot/pull/27762
- [x] #28615
- [x] IsoTpParallelQuery: extend timeout for every frame type (re-do of https://github.com/commaai/openpilot/pull/29654)

More related to fingerprint matching:
- [ ] https://github.com/commaai/openpilot/issues/28483

sshane:
Can make a project if we don't this cluttering up issues

MoreTore:
[...]

PR #27958, opened 2023-04-18
FPv2 timing test: add present ECUs and VIN queries

sshane:


selfdrive/car/tests/test_fw_fingerprint.py
```diff
@@ -134,13 +137,27 @@ def _benchmark(brand, num_pandas, n):
...
+      - Present ECU queries
+      - FW queries for each brand
+      - Total time taken for a car which doesn't match to any brand (CAN fp)
+    """
+
+    # Includes all FW queries + queries in functions
+    total_ref_time = 6.2
+    func_ref_times = {
+      'get_vin': 1.05,
+      'get_present_ecus': 0.5,
```
sshane, line 157: realized we can group the first parallel Toyota address query into the parallel queries to reduce the fingerprinting time by 0.1s for all cars, and 0.2 for Toyota, added to https://github.com/commaai/openpilot/issues/27665

sshane:
ha forgot about this, added in https://github.com/commaai/openpilot/pull/28676

PR #27760, opened 2023-03-31
present ecus: handle Chrysler abs

sshane:
It seems like on the `CHRYSLER PACIFICA 2018` and `CHRYSLER PACIFICA 2020`, this ECU responds with a non standard ISO-TP CAN frame of only 3 bytes. I checked 300k rlogs we have for the last 90 days, and this ECU and platform are the only ones to ever send back non-eight-byte messages. It also reverts back to 8 bytes for the FW query.

`1223` is the ABS response address:

```python
47542224824 rxaddr=1863, bus=2, 420.96 ms, 0x023e000000000000, b'\x02>\x00\x00\x00\x00\x00\x00', len(can.dat)=8
47542224824 rxaddr=1863, bus=128, 420.96 ms, 0x023e000000000000, b'\x02>\x00\x00\x00\x00\x00\x00', len(can.dat)=8
47542224824 rxaddr=1223, bus=2, 420.96 ms, 0x027e00, b'\x02~\x00', len(can.dat)=3
47542224824 rxaddr=1223, bus=0, 420.96 ms, 0x027e00, b'\x02~\x00', len(can.dat)=3
47542224824 rxaddr=1223, bus=1, 420.96 ms, 0x027e00, b'\x02~\x00', len(can.dat)=3
47861919355 rxaddr=1863, bus=2, 740.66 ms, 0x0322f13200000000, b'\x03"\xf12\x00\x00\x00\x00', len(can.dat)=8
47861919355 rxaddr=1863, bus=128, 740.66 ms, 0x0322f13200000000, b'\x03"\xf12\x00\x00\x00\x00', len(can.dat)=8
47861919355 rxaddr=1223, bus=0, 740.66 ms, 0x100d62f132363835, b'\x10\rb\xf12685', len(can.dat)=8
47861919355 rxaddr=1223, bus=2, 740.66 ms, 0x100d62f132363835, b'\x10\rb\xf12685', len(can.dat)=8
47861919355 rxaddr=1223, bus=1, 740.66 ms, 0x100d62f132363835, b'\x10\rb\xf12685', len(can.dat)=8
```

PR #27963, opened 2023-04-19
is_tester_present: don't assert length 8

sshane:
CAN frame data optimization may be performed on single frames, flow control, and last consecutive frames.

- [ ] clean up and integrate with panda uds library fixes (also asserts the same thing)

a case this would fix: https://github.com/commaai/openpilot/pull/27760


sshane:
not ready to do yet, should handle in panda first

PR #27967, opened 2023-04-19
FPv2: Allow excluded types as last resort

MoreTore:
Alternative commit with type checking and comments. (Seems too complicated for what it is) https://github.com/commaai/openpilot/compare/master...MoreTore:openpilot:fw_version_super_fuzz

Fixes an issue that causes the car to be unrecognized when only 1 essential ECU firmware is detected but there are at least 3 matching excluded ECUs.

This commit refactors match_fw_to_car_fuzzy
to allow excluded types as last resort.
This is done by keeping track of the excluded
and included matches rather than leaving out
excluded types. If there is only one included
match, it then checks that the excluded match
candidate is the same as the included match candidate.

https://github.com/commaai/openpilot/issues/27665#issuecomment-1501959513

I have done some testing using a public segment, my recognized car and 1 unrecognized car that is now recognized with these changes.

- [ ] test all fingerprints


MoreTore:
https://github.com/commaai/openpilot/issues/27665#issuecomment-1515536235

PR #27970, opened 2023-04-19
cabana: use QSortFilterProxyModel to handle sorting/filtering

pd0wm:

### DIFF
--- a/tools/cabana/messageswidget.cc
+++ b/tools/cabana/messageswidget.cc
@@ -24,9 +24,14 @@ MessagesWidget::MessagesWidget(QWidget *parent) : QWidget(parent) {
   // message table
   view = new MessageView(this);
   model = new MessageListModel(this);
+
+  proxy_model = new QSortFilterProxyModel(this);
+  proxy_model->setSourceModel(model);
+  proxy_model->setFilterCaseSensitivity(Qt::CaseInsensitive);
+
   auto delegate = new MessageBytesDelegate(view, settings.multiple_lines_bytes);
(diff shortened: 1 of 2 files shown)


deanlee:
I have used QSortFilterProxyModel in the earliest version, but due to the overhead of the mapping layer it adds, the refresh  performance is poor. Although messageswidget has undergone a series of refactoring now, I am still concerned that it will affect the performance of each refresh.

deanlee:
the header's state has been restored in MainWidonw by restoreHeaderState() after startup

After modifying any property of the header, we need to rename the key in the setting file for the changes to take effect. For example, changing "message_header_state_v2" to "message_header_state_v3".

deanlee:
> the header's state has been restored in MainWidonw by restoreHeaderState() after startup
> 
> After modifying any property of the header in the code, we need to rename the key in the setting file for the changes to take effect. For example, changing the key "message_header_state_v2" in settings.cc to "message_header_state_v3". after set sections movable to true in code
> 
> This is a compatibility issue with the state, and currently I haven't thought of a good way to solve it except for changing the key in the setting file.

maybe move the restoreHeaderState() from MainWin into MessageWidget::MessagesWIdget(), and call it before view->header()->setxxxx() can fix this issue?

deanlee:
> I was experimenting with it in the hopes that view->header()->setSectionsMovable(true) would start working, but it doesn't.. Do you have any idea why that doesn't work?

I'm working on this. It seems that there would be a size issue if the hex column is moved to another position.

PR #27971, opened 2023-04-19
cabana: make head section movable

deanlee:
[Kazam_screencast_00050.webm](https://user-images.githubusercontent.com/27770/233111161-460cdca6-b5a5-4758-a96f-9cbcdf9ad07c.webm)

@pd0wm Do you think this is okay?


pd0wm:
yes, this is what I had in mind!

Maybe also put a comment on line of the `restoreHeaderState` that it's put first to make sure it's not overriding anything that comes after.

PR #27973, opened 2023-04-19
UI: cleanup old training guide assets

adeebshihadeh:


selfdrive/ui/qt/offroad/onboarding.cc
```diff
@@ -40,7 +40,6 @@ void TrainingGuide::mouseReleaseEvent(QMouseEvent *e) {
 }
 
 void TrainingGuide::showEvent(QShowEvent *event) {
-  img_path = width() == WIDE_WIDTH ? "../assets/training_wide/" : "../assets/training/";
   boundingRect = width() == WIDE_WIDTH ? boundingRectWide : boundingRectStandard;
```
incognitojam, line 43: shouldn't we always use the wide bounding rect coords now too?

PR #27976, opened 2023-04-19
cabana: fix shadowed variable in abstractstream.cc

pd0wm:
I don't mind making these PRs, but maybe we should set up some CI with a more recent clang than the one from ubuntu 20.04 to catch this. It happens to me on MacOS and on arch linux with latest clang. Maybe we can catch it too with an Ubuntu 22.04 build?


adeebshihadeh:
Definitely interested in better CI to catch this, but we don't have much time to invest in this at the moment. @deanlee perhaps you'd be interested in trying to get some 22.04 CI up? It shouldn't be too hard.

PR #27977, opened 2023-04-19
cabana: fix crash on MacOS when combining graphs

pd0wm:
Introduced in https://github.com/commaai/openpilot/pull/27842

@deanlee Do you understand why this fixes it? Does `QDrag::exec` delete the `ChartView` object?


deanlee:
omg, it looks like the same issue as the first drag/drop bug on Mac before. 
chart will call a single shot time(time out is 0) to delete it self in the next event loop. what I expect is that this will be executed after the mousePressEvent has finished running. But it seems that on macOS, QDialog::exec will process all events in the queue (including the TimerEvent used to delete chart) before returning.

PR #27981, opened 2023-04-20
Ford: driver torque limiting v2

sshane:
Moving work from https://github.com/commaai/openpilot/pull/27415 to preserve its history. It also has some notes, so I'd go check it out


sshane:
Some plots from internal discussion to preserve history.

Max lateral accels & jerks comparison (left two are sending at 100Hz):
image

Lateral accels from the fleet (461 rlogs, last year):
image (6)

Max diff between ABS and PCM speeds (for 2nd speed check) from fleet (461 rlogs, last year).
image (7)

Diffs are mainly from bumps/ABS activating since PCM speed is a filtered version of ABS speed.
image (8)

PR #27990, opened 2023-04-21
locationd: log time to first fix

sshane:
instead of logging ttff for ublox and quectel, we decided that a locationd ttff would be better for now


selfdrive/locationd/locationd.cc
```diff
@@ -529,6 +535,9 @@ void Localizer::time_check(double current_time) {
   if (std::isnan(this->last_reset_time)) {
     this->last_reset_time = current_time;
   }
+  if (std::isnan(this->start_time)) {
+    this->start_time = current_time;
+  }
```
sshane, line 540: @HaraldSchafer do you want to measure from first gpsLocation message, or from first of any of the messages?

selfdrive/locationd/locationd.cc
```diff
@@ -180,6 +180,7 @@ void Localizer::build_live_location(cereal::LiveLocationKalman::Builder& fix) {
   fix.setPosenetOK(!(std_spike && this->car_speed > 5.0));
   fix.setDeviceStable(!this->device_fell);
   fix.setExcessiveResets(this->reset_tracker > MAX_RESET_TRACKER);
+  fix.setTimeToFirstFix(std::isnan(this->ttff) ? -1 : this->ttff);
```
sshane, line 183:
```suggestion
  fix.setTimeToFirstFix(std::isnan(this->ttff) ? -1. : this->ttff);
```

sshane:
We don't patch nanos_since_boot in the C++ version of SubMaster, so `sm.rcv_time(trigger_msg) / 1e9` always returns the current process replay run time. Changed to the logMonoTime of the received trigger message which should be very close.

sshane:
Also, this check takes a few seconds in process replay, so it doesn't set the start time immediately: https://github.com/commaai/openpilot/blob/e12fd7184948fc4d9a5014a745fe03fe5a2e55dd/selfdrive/locationd/locationd.cc#L688

PR #28015, opened 2023-04-24
cabana: fix wrong total seconds in replay mode

deanlee:
related to 2 issue:
1. https://github.com/commaai/openpilot/discussions/26091#discussioncomment-5712934
2. https://github.com/commaai/openpilot/pull/28009

[update]

- fixed the order issue of all_events when start cabana from a specified segment.
- replace deque with vector  to improve performance

### DIFF
--- a/tools/cabana/streams/livestream.cc
+++ b/tools/cabana/streams/livestream.cc
@@ -82,8 +82,8 @@ void LiveStream::updateEvents() {
   }
 
   uint64_t last_ts = post_last_event && speed_ == 1.0
-                         ? all_events_.back()->mono_time
-                         : first_event_ts + (nanos_since_boot() - first_update_ts) * speed_;
+                       ? all_events_.back()->mono_time
+                       : first_event_ts + (nanos_since_boot() - first_update_ts) * speed_;
   auto first = std::upper_bound(all_events_.cbegin(), all_events_.cend(), current_event_ts, [](uint64_t ts, auto e) {
     return ts < e->mono_time;
   });
--- a/tools/cabana/streams/livestream.cc
+++ b/tools/cabana/streams/livestream.cc
@@ -58,7 +58,6 @@ void LiveStream::timerEvent(QTimerEvent *event) {
     }
     if (!all_events_.empty()) {
       begin_event_ts = all_events_.front()->mono_time;
-      total_sec = (all_events_.back()->mono_time - all_events_.front()->mono_time) / 1e9;
       updateEvents();
       return;
     }
@@ -68,7 +67,6 @@ void LiveStream::timerEvent(QTimerEvent *event) {
 
 void LiveStream::updateEvents() {
   static double prev_speed = 1.0;
-  static uint64_t prev_newest_event_ts = all_events_.back()->mono_time;
 
   if (first_update_ts == 0) {
     first_update_ts = nanos_since_boot();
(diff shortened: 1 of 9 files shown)


sshane:
Great! That fixed it getting into a weird state when scrubbing outside the timeline. But there's still some weirdness, it can get stuck at the end of the graph timeline and not update.

https://user-images.githubusercontent.com/25857203/234127018-bf49bc00-82d7-419b-a87a-07321b328d2c.mp4

deanlee:
> Great! That fixed it getting into a weird state when scrubbing outside the timeline. But there's still some weirdness, it can get stuck at the end of the graph timeline and not update. Do you want to wait to merge or shall we merge this fix now?
> cabana.not.scrubbing.properly.mp4

This is a old issue that has been present for a long time due to the loop playback mechanism in tools/replay/replay.cc. I think it would be better to fix it with another PR. not in this one

deanlee:
> But there's still some weirdness, it can get stuck at the end of the graph timeline and not update.
fixed

sshane:
Much better! Now it looks like the max time it prints "33:00" is invalid and won't let me seek to it since the last rlog is missing (segment 33), which I'm fine with if we can either allow seeking to it using the qlog, or seek to it but don't jump around like it sometimes does now. That said, this PR seems like it addresses the other issues I've had, so should we merge now?

https://user-images.githubusercontent.com/25857203/234380987-202ad5c7-444e-4a38-8f67-7006fef9e76c.mp4

deanlee:
This should be an old bug that has been around for a while (we use the number of segments to calculate the total time(60s*n)  in replay instead of the actual time in logs). Think it's better to fix it in another pr.

deanlee:
Changed the maximum value of the slider from `segment count * 60s`  to the actual time of the loaded events to fix this issue.

sshane:
Can we do that in another PR to think about it more? I like to see the total route length when you open a route at the beginning, now it only shows current time + 3 segments/minutes

deanlee:
We have no way to  know the actual time of the logs until all segments are loaded. It is indeed difficult to display the time quickly and accurately. 
Another compromise solution is to still display the time based on `segment count * 60` and make adjustments after all segments are loaded. pop up a prompt if user clicks on a time period without logs .
Either way, it does take more time to think about this issue. find a not too complicated way to solve this headache problem.

deanlee:
The code has been changed to set the maximum time to "segment count * 60s" after startup, and then correct it by loading the last valid qlog to get the actual time in the background thread .
The only issue with this solution is that if there is no valid qlogs, the maximum time will still be displayed as "segment count * 60s". But for now, I think this is acceptable.

tools/cabana/streams/livestream.cc
```diff
@@ -82,16 +81,13 @@ void LiveStream::updateEvents() {
...
 
-  uint64_t last_event_ts = all_events_.back()->mono_time;
-  bool at_the_end = current_event_ts == prev_newest_event_ts;
-  if (!at_the_end) {
-    last_event_ts = first_event_ts + (nanos_since_boot() - first_update_ts) * speed_;
-  }
-
+  uint64_t last_ts = post_last_event && speed_ == 1.0
+                         ? all_events_.back()->mono_time
+                         : first_event_ts + (nanos_since_boot() - first_update_ts) * speed_;
```
sshane, line 86:
```suggestion
  uint64_t last_ts = post_last_event && speed_ == 1.0
                       ? all_events_.back()->mono_time
                       : first_event_ts + (nanos_since_boot() - first_update_ts) * speed_;
```

PR #28019, opened 2023-04-25
regen: sync fcamera and ecamera feeds

fredyshox:
Stops modeld from spamming: `frames out of sync!` error message

### DIFF
--- a/selfdrive/test/process_replay/regen.py
+++ b/selfdrive/test/process_replay/regen.py
@@ -117,48 +117,71 @@ def replay_service(s, msgs):
       rk.keep_time()
 
 
-def replay_cameras(lr, frs, disable_tqdm=False):
+def replay_single_camera(s, stream, dt, vipc_server, frames, size):
+  services = [(s, stream)]
+  pm = messaging.PubMaster([s for s, _ in services])
+  rk = Ratekeeper(1 / dt, print_delay_threshold=None)
+
+  img = b"\x00" * int(size[0] * size[1] * 3 / 2)
+  while True:
+    if frames is not None:
+      img = frames[rk.frame % len(frames)]
+
+    rk.keep_time()
+
+    m = messaging.new_message(s)
+    msg = getattr(m, s)
+    msg.frameId = rk.frame
+    msg.timestampSof = m.logMonoTime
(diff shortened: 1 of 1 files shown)


selfdrive/test/process_replay/regen.py
```diff
@@ -117,48 +117,100 @@ def replay_service(s, msgs):
...
+    msg.timestampSof = m.logMonoTime
+    msg.timestampEof = m.logMonoTime
+    pm.send(s, m)
+
+    vipc_server.send(stream, img, msg.frameId, msg.timestampSof, msg.timestampEof)
+
+  
+def replay_dual_camera(keys, streams, frames, sizes, dt, vipc_server):
+  if len(keys) != len(streams) != len(frames) != len(sizes) != 2:
+    raise ValueError("keys, streams, frames, and sizes must all be length 2")
```
adeebshihadeh, line 144: why is this a different function than `replay_single_camera`?

selfdrive/test/process_replay/regen.py
```diff
@@ -117,48 +117,100 @@ def replay_service(s, msgs):
...
+      msg = getattr(m, s)
+      msg.frameId = rk.frame
+      msg.timestampSof = m.logMonoTime
+      msg.timestampEof = m.logMonoTime
+      pm.send(s, m)
+
+      vipc_server.send(stream, img, msg.frameId, msg.timestampSof, msg.timestampEof)
+
+
+def replay_eon_cameras(frs, disable_tqdm=False):
```
adeebshihadeh, line 176: same here

selfdrive/test/process_replay/regen.py
fredyshox: To synchronize sending of fcam and ecam messages. I figured that if they share the same timestamps and are sent sequentially in the same process, the modeld receives them in order and is actually using matching frames for eval. Otherwise you get a lot of messages about frames being out of sync.

selfdrive/test/process_replay/regen.py
fredyshox: Since the process of simulating camerad, differs so much on eon vs tici (with the new dual cam replay), I thought it would be cleaner to make separate methods for each.

PR #28020, opened 2023-04-25
Hyundai: explicitly copy signals to forward

sshane:
Copied all signals defined in carstate for CANParser that are copied right now


sshane:
@adeebshihadeh is this what you had in mind?

adeebshihadeh:
Yes, but I'd keep the lists inside the functions themselves.

sshane:
Kept `ES_DISTANCE_SIGNALS` as a global since it's used in multiple places, rest are in functions.

selfdrive/car/volkswagen/mqbcan.py
```diff
@@ -10,8 +13,16 @@ def create_steering_control(packer, bus, apply_steer, lkas_enabled):
   return packer.make_can_msg("HCA_01", bus, values)
 
 
-def create_lka_hud_control(packer, bus, ldw_stock_values, enabled, steering_pressed, hud_alert, hud_control):
-  values = ldw_stock_values.copy()
+def create_lka_hud_control(packer, bus, CP, ldw_stock_values, enabled, steering_pressed, hud_alert, hud_control):
+  values = {}
+  if CP.networkLocation == car.CarParams.NetworkLocation.fwdCamera:
```
jyoung8607, line 18: For both MQB and PQ, rather than passing down and checking carParams again, I think it might be cleaner to just check for ldw_stock_values not being empty. `!= {}`, or `len > 0`, or change the `{}` to `None` for [MQB](https://github.com/commaai/openpilot/blob/b4d4920fa46e7ff5c06a61c4f1c92849b5be7197/selfdrive/car/volkswagen/carstate.py#L101) and [PQ](https://github.com/commaai/openpilot/blob/b4d4920fa46e7ff5c06a61c4f1c92849b5be7197/selfdrive/car/volkswagen/carstate.py#L214) in carstate, or whatever style suits you.

sshane: I just realized `COUNTER` is implicitely forwarded every where. @jyoung8607 is it important for the VW port to forward the counter exactly?

jyoung8607: For the ACC button control messages, there's COUNTER increment logic in carcontroller and it's overriding your copy. I want to keep that logic; making button spam work consistently was a multi-year battle I have no wish to re-fight. If you feel it would be cleaner to push that increment down from carcontroller into [mqb|pq]can.py, I'm amenable to that.

For the LDW messages, I wasn't bothering to pass the COUNTER value before, so opendbc was generating one. I don't think the car is likely to care if we pass through or regenerate the counter in this case. If you're trying to drive process replay diffs down to zero, then I'd continue not copying it as you are now.

sshane: @jyoung8607 feel free to do your own PR if you'd like for VW!

jyoung8607: @sshane #28179

sshane: nice! @jnewb1 do you want to do one for Subaru?

PR #28022, opened 2023-04-25
HKG: Add FW Versions for 2023 Kia Forte

sunnyhaibin:
Add firmware for the 2023 Kia Forte. Expand supported model-year range up to 2023 and harness type.

Route ID: `c105c687eb108b77|2023-04-25--19-22-17`

Thanks to community 2023 Kia Forte owner `mrb0908#9440` (Discord).


selfdrive/car/hyundai/values.py
```diff
@@ -1156,21 +1159,29 @@ class Buttons:
       b'\xf1\x00BD  MDPS C 1.00 1.02 56310-XX000 4BD2C102',
       b'\xf1\x00BD  MDPS C 1.00 1.08 56310/M6300 4BDDC108',
       b'\xf1\x00BD  MDPS C 1.00 1.08 56310M6300\x00 4BDDC108',
+      b'\xf1\x00BDm MDPS C A.01 1.03 56310M7800\x00 4BPMC103',
+      b'\xf1\x8756310M7800\x00\xf1\x00BDm MDPS C A.01 1.03 56310M7800\x00 4BPMC103',
```
sshane, line 1163: You only need to add the shortest response (usually), these are just responses from two slightly different FW queries. Perhaps we should add some brand-specific infrastructure to deal with cars that respond to both.

selfdrive/car/hyundai/values.py
```diff
@@ -195,7 +195,10 @@ def init_make(self, CP: car.CarParams):
   CAR.SANTA_CRUZ_1ST_GEN: HyundaiCarInfo("Hyundai Santa Cruz 2022-23", harness=Harness.hyundai_n),
 
   # Kia
-  CAR.KIA_FORTE: HyundaiCarInfo("Kia Forte 2019-21", harness=Harness.hyundai_g),
+  CAR.KIA_FORTE: [
+    HyundaiCarInfo("Kia Forte 2019-21", harness=Harness.hyundai_g),
+    HyundaiCarInfo("Kia Forte 2023", harness=Harness.hyundai_e),
```
sshane, line 200: still not standard 😢

selfdrive/car/hyundai/values.py
sunnyhaibin: Gotcha, that would clear up some confusions for new users. I will remove the longer response in the meantime.

selfdrive/car/hyundai/values.py
sunnyhaibin: It's 2023!

sshane:
weird, his abs responded with the FW query

sshane:
Can you tell the user to get a route where they start the car after the comma three being off and upload rlogs?

PR #28034, opened 2023-04-26
cabana: add filter per column

pd0wm:
Screenshot 2023-05-04 at 10 33 22
Screenshot 2023-05-04 at 10 33 37

- [x] Change top header to contain msg/signal count, clear all filters and multiple lines checkbox
- [x] Test behavior during horizontal scrolling
- [x] Test behavior with hidden columns
- [x] Add range/wildcard filtering for numbers (e.g. 100-200, 100-)
- [x] Add regex filter
- [x] Re-evaluate filters when new data comes in (needed for count, frequency and data)

### DIFF
--- a/tools/cabana/messageswidget.cc
+++ b/tools/cabana/messageswidget.cc
@@ -180,34 +180,34 @@ void MessageListModel::setFilterStrings(const QMap<int, QString> &filters) {
   fetchData();
 }
 
-void MessageListModel::sortMessages(QList<MessageId> &new_msgs) {
+void MessageListModel::sortMessages(Qt::SortOrder sort_order, int sort_column, QList<MessageId> &new_msgs) {
   // QList<MessageId> new_msgs = msgs;
   if (sort_column == Column::NAME) {
-    std::sort(new_msgs.begin(), new_msgs.end(), [this](auto &l, auto &r) {
+    std::sort(new_msgs.begin(), new_msgs.end(), [=](auto &l, auto &r) {
       auto ll = std::pair{msgName(l), l};
       auto rr = std::pair{msgName(r), r};
       return sort_order == Qt::AscendingOrder ? ll < rr : ll > rr;
     });
   } else if (sort_column == Column::SOURCE) {
-    std::sort(new_msgs.begin(), new_msgs.end(), [this](auto &l, auto &r) {
+    std::sort(new_msgs.begin(), new_msgs.end(), [=](auto &l, auto &r) {
       auto ll = std::pair{l.source, l};
       auto rr = std::pair{r.source, r};
       return sort_order == Qt::AscendingOrder ? ll < rr : ll > rr;
     });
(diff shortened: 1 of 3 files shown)


tools/cabana/messageswidget.cc
```diff
@@ -47,21 +38,26 @@ MessagesWidget::MessagesWidget(QWidget *parent) : QWidget(parent) {
...
 
   // signals/slots
-  QObject::connect(filter, &QLineEdit::textEdited, model, &MessageListModel::setFilterString);
+  QObject::connect(header, &MessageViewHeader::filtersUpdated, model, &MessageListModel::setFilterStrings);
   QObject::connect(multiple_lines_bytes, &QCheckBox::stateChanged, [=](int state) {
     settings.multiple_lines_bytes = (state == Qt::Checked);
     delegate->setMultipleLines(settings.multiple_lines_bytes);
     view->setUniformRowHeights(!settings.multiple_lines_bytes);
-    model->sortMessages();
+    model->fetchData(); // Why?
```
deanlee, line 54: The reason I call `sortMessages` here is to reset the model to force the view to recalculate the size of the bytes column. 
This call should be necessary for messages that have not been updated by updateState（or cabana is in pause mode)

pd0wm: Shall we just call begin/endresetModel then?

deanlee: Yes, it's fine to just reset the model here . This way the view will recalculate the size and  display bytes column correctly.

pd0wm: Do I need to use a CAN FD route for multiline bytes? Is there a public one you've been using for testing?

deanlee: Frankly, I just added some temporary code to `MessageListModel::data` for testing:

```
 } else if (role == BytesRole && index.column() == 5) {
    if (QSet<int>({1, 3, 6, 7}).contains(index.row())) {
      QByteArray dat(64, 0);
      memcpy(dat.data(), can_data.dat.data(), can_data.dat.size());
      return dat;
    }
    return can_data.dat;
  }
```

deanlee:
There doesn't seem to be a simple way to reset filters. would it be better to place a "Reset Filters" button above the header?

pd0wm:
> 

Sounds good. I was looking to add the number of signals somewhere as well, since we like that too as a metric of RE progress.

pd0wm:
@adeebshihadeh This is ready for review

@deanlee Happy with the new logic on when to call `resetModel()`? `fetchData` always filters and sorts the messages, and then compares if the list of IDs is different, if so it initiates the reset.

deanlee:
It's very nice. I'll implement the features we discussed in https://github.com/commaai/openpilot/pull/28093#issuecomment-1534215763  after this is merged, with some minor optimizations if necessary.

PR #28037, opened 2023-04-26
ubloxd: enable satReport parsing

fredyshox:


adeebshihadeh:
Are we sure we still want this?

fredyshox:
Yes, I've talked with @haraschax earlier today about this, and we concluded that it's some leftover from debugging and should be enabled. Also tested ubloxd on couple thousand recent segments and it seems to be stable.

PR #28045, opened 2023-04-27
cabana: dispaly CPU & memory usage on the status bar

deanlee:
Display Cabana's current CPU and memory usage in the status bar, makes it convenient for users to check the current resource utilization.


adeebshihadeh:
cabana is getting quite big, and I don't think this is worth 40 lines.

deanlee:
yes. 40 lines for an additional small feature is indeed too much.

PR #28055, opened 2023-04-28
Nissan: set LKA active with latActive

sshane:


sshane:
@avolmensky this is my proposal

PR #28064, opened 2023-04-29
Subaru: stop and go autoresume

martinl:
Subaru Stop and Go autoresume support for global gen1 and pre-global models.

Subaru stock ACC stops behind stopped cars but does not resume automatically. We send Throttle:Throttle_Pedal signal from openpilot to ACC when car in front starts moving to resume from ACC Hold state (mimicking stock ACC resume using gas press behaviour). It works for cars with electric parking brake and does not change stock ACC behaviour for cars with manual parking brake (which disengage 3 seconds after stopping).

For Global models, we also allow setting ACC hold manually when stopped with no lead car (eg at intersection/traffic light) and resume from manual hold by using ACC resume button press, retaining stock ACC behaviour.

We tried to resume by sending ACC resume button press signals but that did not work since ACC buttons are directly wired to Eyesight and not accepted as input from CAN messages.

Prerequisites:
- https://github.com/commaai/panda/pull/1379

I will post test routes for both global and pre-global and then PR should be ready for review


martinl:
I tested spamming the Throttle message without panda fwd blocking and it did not work, so I reverted back to original version

jnewb1:
> I tested spamming the Throttle message without panda fwd blocking and it did not work, so I reverted back to original version

When you tried spamming, did you do COUNTER+1?, like we do for ES_Distance? I can get it to work about 70% of the time by spamming throttle on CAN1 with COUNTER +1 (the only option, since we can’t intercept CAN1).

martinl:
I did not try to use counter+1. There is another option for gen1 using brake message speed signal that would avoid using auto hold and would also work for manual parking brake cars https://github.com/martinl/openpilot/blob/subaru-community/selfdrive/car/subaru/carcontroller.py#L96

martinl:
I tested sending throttle message with counter+1 and I can confirm it also works about 60% of the time on gen1. Here is a route using jnewb1 subaru-gen2-sng-throttle branch `8de015561e1ea4a0|2023-07-17--16-59-49`

martinl:
Would it make a difference if gen1 used conditional blocking for Throttle message in panda safety only when openpilot needs to send the throttle message (eg by sending custom can message with one signal bit to start/stop the blocking)?

sshane:
Isn't mergable, pending AEB tests I think. @martinl let us know if you have any conclusion about that!

To summarize, sending fake gas pedal messages to the camera, which does AEB, might change AEB behavior with stock long, in release, which isn't acceptable.

PR #25345, opened 2022-08-03
Subaru: POC for longitudinal control

martinl:
**Description** POC for Subaru Global longitudinal control. For acceleration, there are 2 signals - cruise_throttle and cruise_rpm that are basically throttle and rpm target requests. Scaling values used in carcontroller are roughly modeled by plotting and observing stock acc signals and testing. Brake request is single linear signal with 0-400 scale. 

Stock AEB/PCB is passed through. 

The general idea for logitudinal control is that eyesight is not engaged by rewriting the feedback signals from car. openpilot generates the required signals for lateral and longitudinal control and dash display.

POC is usable and I have used it for testing the openpilot vision only performance from time to time

**Verification** Tested on 2018 Crosstrek

PR #28066, opened 2023-04-29
Subaru: frequency based messaging

jnewb1:
**Description**
Changes Subaru messaging to use frequency rather than relying on the counter variable from eyesight. This paradigm change will be required for gen2 long, where we disable or unplug the eyesight camera.

**Verification**
Tested on my Subaru Legacy 22

**Route**
d9df6f87e8feff94|2023-04-29--10-08-40--0


selfdrive/car/subaru/carcontroller.py
```diff
@@ -10,16 +10,23 @@ def __init__(self, dbc_name, CP, VM):
...
-    self.es_distance_cnt = -1
-    self.es_dashstatus_cnt = -1
-    self.infotainmentstatus_cnt = -1
     self.cruise_button_prev = 0
     self.last_cancel_frame = 0
 
     self.p = CarControllerParams(CP)
     self.packer = CANPacker(DBC[CP.carFingerprint]['pt'])
 
+  def should_send_message(self, message_name):
```
sshane, line 19: let's do what other car ports do and just check the freq in the if statement, we can do global clean ups like this later

jnewb1:
we should probably merge https://github.com/commaai/openpilot/pull/25345 first, then ill add all the long frequency stuff to this PR

martinl:
Looks like all three 10Hz messages can be behind one `if self.frame % 10 == 0:`

martinl:
Also, about original implementation with message frequencies array - message frequencies are already defined it carstate get_can_parser checks, is there an easy way to use them from carcontroller in future cleanup PR for all brands?

martinl:
Also tested and working with global gen1: `8de015561e1ea4a0|2023-04-30--20-34-11`

sshane:
@jnewb1 can you open a PR to make the counters use openpilot's instead of the forwarded messages'?

PR #28073, opened 2023-05-01
calibrationd: add offroad alert when remount detected

adeebshihadeh:

### DIFF
--- a/common/params.cc
+++ b/common/params.cc
@@ -202,6 +202,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"Offroad_TemperatureTooHigh", CLEAR_ON_MANAGER_START},
     {"Offroad_UnofficialHardware", CLEAR_ON_MANAGER_START},
     {"Offroad_UpdateFailed", CLEAR_ON_MANAGER_START},
+    {"Offroad_Recalibration", CLEAR_ON_ONROAD_TRANSITION},
 };
 
 } // namespace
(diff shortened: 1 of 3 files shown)


common/params.cc
```diff
@@ -202,6 +202,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"Offroad_TemperatureTooHigh", CLEAR_ON_MANAGER_START},
     {"Offroad_UnofficialHardware", CLEAR_ON_MANAGER_START},
     {"Offroad_UpdateFailed", CLEAR_ON_MANAGER_START},
+    {"Offroad_Recalibration", CLEAR_ON_ONROAD_TRANSITION},
```
adeebshihadeh, line 205:
```suggestion
    {"Offroad_Recalibration", CLEAR_ON_MANAGER_START | CLEAR_ON_ONROAD_TRANSITION},
```

common/params.cc
```diff
@@ -202,7 +202,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"Offroad_TemperatureTooHigh", CLEAR_ON_MANAGER_START},
     {"Offroad_UnofficialHardware", CLEAR_ON_MANAGER_START},
     {"Offroad_UpdateFailed", CLEAR_ON_MANAGER_START},
-    {"Offroad_Recalibration", CLEAR_ON_ONROAD_TRANSITION},
+    {"Offroad_Recalibration", CLEAR_ON_MANAGER_START | CLEAR_ON_ONROAD_TRANSITION},
```
sshane, line 205: Want me to add a test to assert `CLEAR_ON_MANAGER_START` if `CLEAR_ON_ONROAD_TRANSITION`?  Since things can break if you reboot the device onroad now that we don't clear in boardd.

adeebshihadeh: no, we should just also clear them at the same time

sshane: https://github.com/commaai/openpilot/pull/28198

PR #28081, opened 2023-05-02
cabana: supports switching streams on the fly

deanlee:
Currently, you can only switch between streams or routes by restarting cabana, includes re-choosing the DBC files. 
This PR adds a dynamic switching feature, so you no longer need to restart and can use the current DBC files to open a new stream/route/device


pd0wm:
Would be nice if you could also switch to no stream at all (go offline). E.g. in case when you're connected to the panda, but want to temporarily use the panda with a different program/script.

deanlee:
Yes, this feature can be implemented in future PR

PR #28083, opened 2023-05-03
Laikad: blocking raw_gnss receive

haraschax:
Saves 5-10% cpu

### DIFF
--- a/selfdrive/locationd/laikad.py
+++ b/selfdrive/locationd/laikad.py
@@ -5,7 +5,6 @@ import time
 import shutil
 from collections import defaultdict
 from concurrent.futures import Future, ProcessPoolExecutor
-from datetime import datetime
 from enum import IntEnum
 from typing import List, Optional, Dict, Any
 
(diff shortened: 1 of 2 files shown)


selfdrive/locationd/laikad.py
```diff
@@ -449,7 +449,7 @@ def main(sm=None, pm=None):
   laikad = Laikad(save_ephemeris=not replay, auto_fetch_navs=use_internet, use_qcom=use_qcom)
 
   while True:
-    for in_msg in messaging.drain_sock(raw_gnss_sock):
+    for in_msg in messaging.drain_sock(raw_gnss_sock, wait_for_one=True):
```
adeebshihadeh, line 452: this makes it such that the clocks block below will never be hit. for this case, you probably just want to do a single blocking receive

HaraldSchafer: You mean because of this
```
 laikad.got_first_gnss_msg
 ```
 ?

adeebshihadeh: yeah

HaraldSchafer: Not fully sure why the clocks block is there, seems redundant

HaraldSchafer: Let me just remove

PR #28093, opened 2023-05-03
cabana: supports multiple dbc files, either globally or for a specified bus.

deanlee:
Currently, Cabana only supports one global DBC file and can only specify one DBC file for a bus.

This PR adds support for multiple DBC files, and updates Messages and Signals in different files upon saving.

1. open multiple dbc files globally:
    
    
2. open multiple dbc files for a bus:

### DIFF
--- a/tools/cabana/cabana.cc
+++ b/tools/cabana/cabana.cc
@@ -92,7 +92,7 @@ int main(int argc, char *argv[]) {
 
   // Load DBC
   if (!dbc_file.isEmpty()) {
-    w.loadFile(dbc_file);
+    w.loadFile({dbc_file});
   }
 
   w.show();
--- a/tools/cabana/dbc/dbc.h
+++ b/tools/cabana/dbc/dbc.h
@@ -1,9 +1,6 @@
 #pragma once
 
-#include <map>
 #include <QList>
-#include <QMetaType>
-#include <QObject>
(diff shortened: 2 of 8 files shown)


deanlee:
@pd0wm any thoughts? Are there any other issues that I have overlooked?

pd0wm:
How are you handling the case where a message/id is present in multiple DBC files? You have to consider the use case where you're trying to combine multiple DBC files for related cars into a single DBC file for the route you're looking at.

My thought was to add a tab to the message editor for each DBC. Then it's explicit which DBC file you're editing. This should als have some options to move/copy a message definition from one DBC file to the other.

deanlee:
> How are you handling the case where a message/id is present in multiple DBC files?

If there are multiple files with the same messge id, the first one of them will be used for now

> Unrelated to this PR, but related to this use case. We should show all messages that are present in the DBC file, even if they are not in the route. Just show with count = 0, and no data.

I'll implement this feature after https://github.com/commaai/openpilot/pull/28034 is merged

pd0wm:
> > How are you handling the case where a message/id is present in multiple DBC files?
> 
> If there are multiple files with the same messge id, the first one of them will be used for now
> 

If you don't know which DBC file a message is coming from this could be very confusing I think.

PR #28100, opened 2023-05-04
HKG: Add new fw for Kia Sportage 5th Gen

BrandonHacks:
Update values.py

New radar fw

If a route is a must then I’ll have to submit one tomorrow.


incognitojam:
Just need a dongle ID

BrandonHacks:
> Just need a dongle ID

09559f1fcaed4704

PR #28107, opened 2023-05-04
cabana: transmit

pd0wm:
Useful for figuring out what certain dashbaord messages mean, or scaling of things like wheel speed by observing shown value in the car.

 - Transmit at a fixed rate, on button press, or on receive on that message (useful to override the state of a certain signal).
 - Allow overriding signals, otherwise the received value is copied to be transmitted.

Future work for separate PRs:
- Setting a "function" as TX value, such as a counter or checksum
- Flooding the bus with a msg of certain ID to block everything with a higher ID
- In binary view, show status of each bit. (e.g. copied from original msg, overriding but same as original, overriding but different than original).
- Set transmit value by editing binary view directly. E.g. click on bits to toggle them, or editing the byte value.
- Decouple transmit timing from FPS

### DIFF
--- a/tools/cabana/dbc/dbc.cc
+++ b/tools/cabana/dbc/dbc.cc
@@ -64,6 +64,10 @@ double get_raw_value(const uint8_t *data, size_t data_size, const cabana::Signal
   return val * sig.factor + sig.offset;
 }
 
+void set_raw_value(const uint8_t *data, size_t data_size, const cabana::Signal &sig, double value) {
+
+}
+
 bool cabana::operator==(const cabana::Signal &l, const cabana::Signal &r) {
   return l.name == r.name && l.size == r.size &&
          l.start_bit == r.start_bit &&
--- a/tools/cabana/dbc/dbc.cc
+++ b/tools/cabana/dbc/dbc.cc
@@ -64,8 +64,31 @@ double get_raw_value(const uint8_t *data, size_t data_size, const cabana::Signal
   return val * sig.factor + sig.offset;
 }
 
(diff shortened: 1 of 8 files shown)


pd0wm:
@deanlee Any thoughts on how to decouple this from the `fps` setting? `SignalView::updateState` is only called at the rate of the fps.

Ideally I'd like to react to CAN messages as soon as they come in. I'd like to copy over the message, override some values and immediately send it out again. The only place I can think of would be `LiveStream::handleEvent`. But this would also require using dbc related functions, which I guess we want to avoid from stream code.

deanlee:
If you want to process the package and refresh the ui immediately after streamThread receiving it, all I can think of at the moment is to move the code in LiveStream::timerEvent to a slot function, then emit a signal after handleevent and let the main thread call the slot function to process and send the data.

pd0wm:
I don't need to modify the packets, or update the ui. I just want some kind of signal/event with a list of message from the last batch.

Maybe we can convert into `struct CanEvent` in the `handleEvent` function instead of later, and then also emit a signal in case other code is interested?

deanlee:
sounds good!

dibble9012:
Maybe I'm misunderstanding how this should be setup.... transmit is greyed out when I run cabana over ZMQ to my C3 w/red panda for CANFD. I'm just trying to test dashboard messages right now. Panda safety on the C3 is set to allOutput. Any help would be great, thanks!

pd0wm:
This PR is not done yet

adeebshihadeh:
Still relevant or shall we close for now?

PR #28108, opened 2023-05-04
cabana: show and edit all messages present in DBC files

deanlee:
Show and edit all messages in the DBC file, even if they are not in the data stream. (related discuss: https://github.com/commaai/openpilot/pull/28093#issuecomment-1534318700)

- If the message does not have any data, the relevant column will be displayed as `N/A`:

  
- Edit a DBC message without any data:
  

The following improvements have also been added:

1. re-`fetchData` only when needed. avoid calling expensive `sort_messages` for each update (especially the default sort by name) . (related discuss: https://github.com/commaai/openpilot/pull/28034#issuecomment-1534306748)
2. Disable the "reset filters" button when there is no filter.
3. add shortcut key to "reset filters" and "multiple line bytes"
4. Fixed the wrong grid line issue when dragging the last column:
 
     [Kazam_screencast_00109.webm](https://user-images.githubusercontent.com/27770/236400735-8a601f7d-4093-41f7-820b-8739cf3ace11.webm)

### DIFF
--- a/tools/cabana/dbc/dbc.h
+++ b/tools/cabana/dbc/dbc.h
@@ -61,6 +61,7 @@ namespace cabana {
   };
 
   struct Msg {
+    uint32_t address;
     QString name;
     uint32_t size;
     QList<cabana::Signal> sigs;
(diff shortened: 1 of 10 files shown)


pd0wm:
Does filtering on data/count/freq still work after this? You need to re-evaluate the filters every time new messages come in.

deanlee:
These are a couple of issues I hadn't considered -_-b. it would consume >9% cpu on my laptop If recalculated and sorted every time. may have to change the way of optimization.

deanlee:
Now it can filter data/count/freq properly

PR #28113, opened 2023-05-04
Ford: add Escape test route and remove dashcam only

incognitojam:


AlexF150-2015:
I'm crazy wanting to have a Comma 3 but I don't know if my F150 2015 Platinum  will work, I can't find videos of people with F150 only 1 video on youtube and it's very old,
On the website it says it is not compatible but I have ACC + LKAS
I just don't have lane centering

PR #28114, opened 2023-05-05
Ford: improved ACC UI for OP long

incognitojam:
- [x] Show "stopped" status message on IPC when in standstill, similar to stock
- [x] Prevent the distance control UI from appearing on button press (as this does not influence OP long).
- [x] OP controlled lead indicator
- [x] Only override stock values when using OP long


sshane:
you can merge this if it all works

PR #28115, opened 2023-05-05
process_replay: universal replay mechanism using cereal fake sockets

fredyshox:


cereal
sshane, line 1: heads up, you bumped cereal to a commit not pushedyet, it causes issues with the updater. can you push the cereal commit to a branch?

adeebshihadeh: this should no longer be true

selfdrive/test/process_replay/compare_logs.py
```diff
@@ -70,7 +70,10 @@ def compare_logs(log1, log2, ignore_fields=None, ignore_msgs=None, tolerance=Non
     field_tolerances = {}
   default_tolerance = EPSILON if tolerance is None else tolerance
 
-  log1, log2 = (list(filter(lambda m: m.which() not in ignore_msgs, log)) for log in (log1, log2))
+  log1, log2 = (
+    sorted(list(filter(lambda m: m.which() not in ignore_msgs, log)), key=lambda m: (m.logMonoTime, m.which()))
+    for log in (log1, log2)
+  )
```
adeebshihadeh, line 76: let's split this out into its own PR

selfdrive/test/process_replay/process_replay.py
```diff
@@ -29,175 +24,165 @@
...
+        for s in pubs_with_events
+      }
+
+    return self
+
+  def __exit__(self, exc_type, exc_obj, exc_tb):
+    del self.recv_called_events
+    del self.recv_ready_events
+    del self.poll_called_event
+    del self.poll_ready_event
```
adeebshihadeh, line 62: what are these dels for?

selfdrive/test/process_replay/process_replay.py
```diff
@@ -206,204 +191,291 @@ def get_car_params(msgs, fsm, can_sock, fingerprint):
...
+
+    if cfg.config_callback is not None:
+      params = Params()
+      cfg.pubs, cfg.drained_pub = cfg.config_callback(params, cfg)
+      cfg.polled_pubs = list(set(cfg.polled_pubs) & set(cfg.pubs))
+
+    all_msgs = sorted(lr, key=lambda msg: msg.logMonoTime)
+    pub_msgs = [msg for msg in all_msgs if msg.which() in set(cfg.pubs)]
+
+    with ReplayContext(cfg) as rc:
```
adeebshihadeh, line 420: the context managers can be chained, so the top can just be `with OpenpilotPrefix(), with ReplayContext(cfg) as rc:`

selfdrive/test/process_replay/process_replay.py
```diff
@@ -206,204 +191,291 @@ def get_car_params(msgs, fsm, can_sock, fingerprint):
...
+        # Wait for process to startup
+        with Timeout(10, error_msg=f"timed out waiting for process to start: {repr(cfg.proc_name)}"):
+          while not all(pm.all_readers_updated(s) for s in cfg.pubs):
+            time.sleep(0)
+
+        for s in sockets.values():
+          messaging.recv_one_or_none(s)
+
+        # TODO wait for sockets to reconnect, for now lets just sleep
+        time.sleep(1)
```
adeebshihadeh, line 442: can't we directly check this?

selfdrive/test/process_replay/process_replay.py
```diff
@@ -206,204 +191,291 @@ def get_car_params(msgs, fsm, can_sock, fingerprint):
...
+        cnt = 0
+        for msg in pub_msgs:
+          with Timeout(cfg.timeout, error_msg=f"timed out testing process {repr(cfg.proc_name)}, {cnt}/{len(pub_msgs)} msgs done"):
+            resp_sockets, end_of_cycle = cfg.subs, True
+            if cfg.should_recv_callback is not None:
+              end_of_cycle = cfg.should_recv_callback(msg, CP, cfg, cnt)
+
+            if len(log_msgs) == 0 and len(resp_sockets) > 0:
+              for s in sockets.values():
+                messaging.recv_one_or_none(s)
```
adeebshihadeh, line 454: then this can be removed

selfdrive/test/process_replay/process_replay.py
```diff
@@ -29,175 +24,165 @@
...
+    assert(len(self.non_polled_pubs) != 0 or len(self.polled_pubs) != 0 or self.drained_pub is not None)
+  
+  def __enter__(self):
+    messaging.toggle_fake_events(True)
+
+    if len(self.polled_pubs) > 0 and self.drained_pub is None:
+      self.poll_called_event = messaging.fake_event("", messaging.FAKE_EVENT_POLL_CALLED)
+      self.poll_ready_event = messaging.fake_event("", messaging.FAKE_EVENT_POLL_READY)
+      self.recv_called_events = {}
+      self.recv_ready_events = {}
```
adeebshihadeh, line 42: do we need to handle polled sockets differently? shouldn't hooking recv be enough?

selfdrive/test/process_replay/compare_logs.py
fredyshox: https://github.com/commaai/openpilot/pull/28201

selfdrive/test/process_replay/process_replay.py
fredyshox: Done

selfdrive/test/process_replay/process_replay.py
fredyshox: __dealloc__ is used to close event fds

selfdrive/test/process_replay/process_replay.py
fredyshox: I've found a way to handle that differently, by calling receive on sub-sockets for the first time, when I'm sure the process is ready (recv_called set for the first time)

selfdrive/test/process_replay/process_replay.py
fredyshox: with the new FakePoller implementation, these are not needed anymore

selfdrive/test/process_replay/process_replay.py
fredyshox: Ok, it must be the way it was, because the config_callback must be run before ReplayContext is created, because it changes the `pubs` and `drained_pub` of the process - laikad uses that to switch between qcom and ublox

PR #28118, opened 2023-05-06
Calculate average desired curvature using planned distance instead of vego

pfeiferj:
### The problem:
The psis used when calculating the average desired curvature in get_lag_adjusted_curvature appear to be the psi based on the planned acceleration. In get_lag_adjusted_curvature we try to remove the distance from the psi to get the curvature. To determine the distance the function takes the current velocity times the lag. However, when the model is planning on having a positive or negative acceleration in the curve there will be a difference in distance the model plans on traveling vs the distance that would be traveled at the current velocity. This results in poor planner performance in the curve because the planner is using a falsely high or low curvature depending on negative or positive acceleration in the model.

### Solution:
This change gets a simple average of the first planned velocity and the velocity interpolated at the lag time to then use in the distance calculation. This results in more accurate curve planning by the planner.

### Graphs:
**Without change:**
Screenshot 2023-05-05 at 7 51 32 PM
Notice how the planner curvature (green second row) flattens out at the apex of the curve whenever we begin accelerating.

**With change:**
Screenshot 2023-05-05 at 7 22 41 PM
In this screenshot note how much smoother the planner curvature lowers and rises. Also note that this was even hitting the curve 10 mph faster than in the previous graph without the change, which traditionally would have caused poorer performance.

### Personal Observation:
Control throughout curves feels noticeably smoother and more planted. Fewer odd jerks of the steering wheel. Seems not to cut the corner as bad either.

### DIFF
--- a/selfdrive/controls/controlsd.py
+++ b/selfdrive/controls/controlsd.py
@@ -614,7 +614,8 @@ class Controls:
       self.desired_curvature, self.desired_curvature_rate = get_lag_adjusted_curvature(self.CP, CS.vEgo,
                                                                                        lat_plan.psis,
                                                                                        lat_plan.curvatures,
-                                                                                       lat_plan.curvatureRates)
+                                                                                       lat_plan.curvatureRates,
+                                                                                       long_plan.distances)
       actuators.steer, actuators.steeringAngleDeg, lac_log = self.LaC.update(CC.latActive, CS, self.VM, lp,
                                                                              self.last_actuators, self.steer_limited, self.desired_curvature,
                                                                              self.desired_curvature_rate, self.sm['liveLocationKalman'])
--- a/selfdrive/controls/lib/drive_helpers.py
+++ b/selfdrive/controls/lib/drive_helpers.py
@@ -17,6 +17,7 @@ V_CRUISE_INITIAL_EXPERIMENTAL_MODE = 105
 IMPERIAL_INCREMENT = 1.6  # should be CV.MPH_TO_KPH, but this causes rounding errors
 
 MIN_SPEED = 1.0
+MIN_DIST = 0.001
 CONTROL_N = 17
 CAR_ROTATION_RADIUS = 0.0
 
@@ -163,11 +164,12 @@ def rate_limit(new_value, last_value, dw_step, up_step):
   return clip(new_value, last_value + dw_step, last_value + up_step)
 
 
-def get_lag_adjusted_curvature(CP, v_ego, psis, curvatures, curvature_rates):
-  if len(psis) != CONTROL_N:
+def get_lag_adjusted_curvature(CP, v_ego, psis, curvatures, curvature_rates, distances):
+  if len(psis) != CONTROL_N or len(distances) != CONTROL_N:
     psis = [0.0]*CONTROL_N
     curvatures = [0.0]*CONTROL_N
     curvature_rates = [0.0]*CONTROL_N
+    distances = [0.0]*CONTROL_N
   v_ego = max(MIN_SPEED, v_ego)
 
   # TODO this needs more thought, use .2s extra for now to estimate other delays
@@ -178,7 +180,11 @@ def get_lag_adjusted_curvature(CP, v_ego, psis, curvatures, curvature_rates):
   # psi to calculate a simple linearization of desired curvature
   current_curvature_desired = curvatures[0]
   psi = interp(delay, T_IDXS[:CONTROL_N], psis)
-  average_curvature_desired = psi / (v_ego * delay)
+  distance = interp(delay, T_IDXS[:CONTROL_N], distances)
+  distance = max(MIN_DIST, distance)
+
+  average_curvature_desired = psi / distance
+
   desired_curvature = 2 * average_curvature_desired - current_curvature_desired
 
   # This is the "desired rate of the setpoint" not an actual desired rate
--- a/selfdrive/controls/lib/lateral_planner.py
+++ b/selfdrive/controls/lib/lateral_planner.py
@@ -119,18 +119,25 @@ class LateralPlanner:
     else:
       self.solution_invalid_cnt = 0
 
-  def publish(self, sm, pm):
+  def publish(self, sm, pm, longitudinal_planner):
     plan_solution_valid = self.solution_invalid_cnt < 2
+    if len(longitudinal_planner.v_desired_trajectory) < CONTROL_N:
+      return
+
     plan_send = messaging.new_message('lateralPlan')
     plan_send.valid = sm.all_checks(service_list=['carState', 'controlsState', 'modelV2'])
+
+    velocities = np.maximum(0.01, np.array(longitudinal_planner.v_desired_trajectory[0:CONTROL_N]))
+    curvatures_rates = np.array(self.lat_mpc.x_sol[0:CONTROL_N, 3])
 
     lateralPlan = plan_send.lateralPlan
     lateralPlan.modelMonoTime = sm.logMonoTime['modelV2']
     lateralPlan.dPathPoints = self.y_pts.tolist()
     lateralPlan.psis = self.lat_mpc.x_sol[0:CONTROL_N, 2].tolist()
 
-    lateralPlan.curvatures = (self.lat_mpc.x_sol[0:CONTROL_N, 3]/self.v_ego).tolist()
-    lateralPlan.curvatureRates = [float(x/self.v_ego) for x in self.lat_mpc.u_sol[0:CONTROL_N - 1]] + [0.0]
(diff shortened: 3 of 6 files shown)


pfeiferj:
Just made another observation from the graphs. Despite hitting the curve 10 mph faster there was still less roll than when I hit the curve without the change. This lines up with my impression of how it feels in the curves.

rchybicki:
Not sure if this helps, but I tested the code out, and it feels better in many turns with speeds above 40kph, but there is one short, low speed turn on the way to my house that it's not making anymore when I apply these changes, like it didn't have enough torque - but there's also no torque limit warning.

Here's a clip and route before applying:
https://connect.comma.ai/36921882a3bf0f19/clips/f16a5ed8fadb456a8be06452f17471ff
36921882a3bf0f19|2023-05-08--15-20-48--10
https://connect.comma.ai/36921882a3bf0f19/1683552048292/1683552839093

and after applying these changes, with me taking over to not hit the curb:
https://connect.comma.ai/36921882a3bf0f19/clips/9db1a9a705684e1e943419b610f559be
36921882a3bf0f19|2023-05-08--08-33-44--9
https://connect.comma.ai/36921882a3bf0f19/1683527624477/1683528386889

I'm driving through that curve very often, I did multiple tests with and without the patch with the same results each time. I'm running a fork of the latest sunnypilot dev branch, with some changes on top but I think none that would affect lateral.

pfeiferj:
> Not sure if this helps, but I tested the code out, and it feels better in many turns with speeds above 40kph, but there is one short, low speed turn on the way to my house that it's not making anymore when I apply these changes, like it didn't have enough torque - but there's also no torque limit warning.
> 
> Here's a clip and route before applying: https://connect.comma.ai/36921882a3bf0f19/clips/f16a5ed8fadb456a8be06452f17471ff 36921882a3bf0f19|2023-05-08--15-20-48--10 https://connect.comma.ai/36921882a3bf0f19/1683552048292/1683552839093
> 
> and after applying these changes, with me taking over to not hit the curb: https://connect.comma.ai/36921882a3bf0f19/clips/9db1a9a705684e1e943419b610f559be 36921882a3bf0f19|2023-05-08--08-33-44--9 https://connect.comma.ai/36921882a3bf0f19/1683527624477/1683528386889
> 
> I'm driving through that curve very often, I did multiple tests with and without the patch with the same results each time. I'm running a fork of the latest sunnypilot dev branch, with some changes on top but I think none that would affect lateral.

Interesting, i'll take a look over the data in plotjuggler here in a bit. there's a chance that the model struggles with that curve and the way that get_lag_adjusted_curvature worked before caused it to request to turn more than the model actually desired, but i wouldn't think it would be that severe at those speed. I know there's still some amount of error in the calculations even with this change because it actually ends up requesting the curvature up to a point just before where we would expect to be distance wise. There's some things that could be done to deal with that last bit of error but my assumption was it was low enough of an error to not be worth the increase in complexity to fix it. I'll have to take a look at the data though to see if it seems to be caused by the error in the distance we're requesting.

pfeiferj:
> Not sure if this helps, but I tested the code out, and it feels better in many turns with speeds above 40kph, but there is one short, low speed turn on the way to my house that it's not making anymore when I apply these changes, like it didn't have enough torque - but there's also no torque limit warning.
> 
> Here's a clip and route before applying: https://connect.comma.ai/36921882a3bf0f19/clips/f16a5ed8fadb456a8be06452f17471ff 36921882a3bf0f19|2023-05-08--15-20-48--10 https://connect.comma.ai/36921882a3bf0f19/1683552048292/1683552839093
> 
> and after applying these changes, with me taking over to not hit the curb: https://connect.comma.ai/36921882a3bf0f19/clips/9db1a9a705684e1e943419b610f559be 36921882a3bf0f19|2023-05-08--08-33-44--9 https://connect.comma.ai/36921882a3bf0f19/1683527624477/1683528386889
> 
> I'm driving through that curve very often, I did multiple tests with and without the patch with the same results each time. I'm running a fork of the latest sunnypilot dev branch, with some changes on top but I think none that would affect lateral.

One other thought is that it could be related to the STEERING_RATE_COST in the lateral planner. It would stand to reason that the amount of extra curvature that get_lag_adjusted_curvature was requesting overcame the high steering rate cost that's set in the lateral planner currently, but that once it starts requesting a more accurate curvature the steering rate cost overcomes the curvature the model is requesting and causes that feeling of not having enough torque at low speeds

rchybicki:
> > Not sure if this helps, but I tested the code out, and it feels better in many turns with speeds above 40kph, but there is one short, low speed turn on the way to my house that it's not making anymore when I apply these changes, like it didn't have enough torque - but there's also no torque limit warning.
> > Here's a clip and route before applying: https://connect.comma.ai/36921882a3bf0f19/clips/f16a5ed8fadb456a8be06452f17471ff 36921882a3bf0f19|2023-05-08--15-20-48--10 https://connect.comma.ai/36921882a3bf0f19/1683552048292/1683552839093
> > and after applying these changes, with me taking over to not hit the curb: https://connect.comma.ai/36921882a3bf0f19/clips/9db1a9a705684e1e943419b610f559be 36921882a3bf0f19|2023-05-08--08-33-44--9 https://connect.comma.ai/36921882a3bf0f19/1683527624477/1683528386889
> > I'm driving through that curve very often, I did multiple tests with and without the patch with the same results each time. I'm running a fork of the latest sunnypilot dev branch, with some changes on top but I think none that would affect lateral.
> 
> One other thought is that it could be related to the STEERING_RATE_COST in the lateral planner. It would stand to reason that the amount of extra curvature that get_lag_adjusted_curvature was requesting overcame the high steering rate cost that's set in the lateral planner currently, but that once it starts requesting a more accurate curvature the steering rate cost overcomes the curvature the model is requesting and causes that feeling of not having enough torque at low speeds

Ok, let me know if I can help in any way or if you could use more routes.

YassineYousfi:
why is the `(v + velocities[0]) / 2)` needed?

pfeiferj:
> why is the `(v + velocities[0]) / 2)` needed?

it's a rough approximation of the average velocity between now and the point we're calculating the curvature at. average velocity * delay would be the distance the model thinks you're going to be at for the psi value. If we use only the velocity at the delay then we would have a similar issue as using v_ego, the final velocity * distance will either overshoot or undershoot based off of if the starting velocity was higher or lower than the final velocity.

pfeiferj:
another thought i had since i made this PR was potentially exposing the x solution from the longitudinal mpc in the longitudinalPlan. Grabbing the x solution position at the delay would likely be more accurate than this rough average.

YassineYousfi:
That makes more sense. Or you can integrate the speeds up to t=delay using np.cumsum if you want to quickly experiment without changing the messages.

pfeiferj:
Switched over to mpc distance. Initial impression is good, but i'll be getting in more driving today.

pfeiferj:
Pretty happy with how it's performing. I've made a route public to use as an example: https://connect.comma.ai/6cc716ffa2eb1f32/1684276696320/1684277240226

pfeiferj:
@YassineYousfi any thoughts on this in its current state? Or anything you'd like me to do for validation?

james5294:
This single-handedly removes crossing the yellow line on every other corner for me. Without this modification op is almost unusable on the winding side roads where I live. I have tried it on master. I have tried it on other forks, no matter where I place this commit the effects are immediate.

22 sonata limited

james5294:
I still experienced cutting corner, with both new models, including the newest, hot coffee.

pfeiferj:
> I still experienced cutting corner, with both new models, including the newest, hot coffee.

Do you mean with this change? Rchybicki is saying that with this change most of his cutting is gone

rchybicki:
> I still experienced cutting corner, with both new models, including the newest, hot coffee.

On stock Nicki me too, with this PR and Nicki nearly no more corner cutting.

YassineYousfi:
We are looking into this. Some turn cutting comes from higher in the planner stack which we are also looking into.

pfeiferj:
hmm, not sure how but i managed to have my cereal and panda out of sync with master. updated to pull in latest master

pfeiferj:
squashed commits and rebased

pfeiferj:
Big thanks to @nworb-cire for pointing some things out in the lateral planner.

@YassineYousfi pushed up a couple more changes that seem to improve things.

1.) when calculating the v_plan in the lateral_planner only use the velocity on the x axis instead of the normalized velocity across the axes. This makes the usage of velocity more consistent as the calculations for curvature rates and lateral accel depend on the tangential velocity. 

2.) Use the longitudinal planner velocities when calculating the curvatures and curvature rates for the published plan. This helps align the calculations to the model expected state at each position thereby creating less error between the model and the plan.

PR #28119, opened 2023-05-06
Lane Departure Warning: fix desire index bug

sshane:
As found by @ZwX1616, we were checking the wrong desire predictions from the vision model.

Changed in this PR which forgot about controlsd 😢 https://github.com/commaai/openpilot/pull/2452


sshane:
image

image

sshane:
image

image

PR #28127, opened 2023-05-07
seldrive: use common sig handler

sshane:
It appears sometimes soundd hits this timeout here when `Ctrl+C`ing manager, but I haven't been able to reliably reproduce yet to debug. This is now just a clean up until the next PR

https://github.com/commaai/openpilot/blob/758a9d7c24e4736dd94cfbab127ab1414b3423a0/selfdrive/manager/process.py#L132-L149


sshane:
Hmm, it's still hanging when running many times in a row

PR #27656, opened 2023-03-23
boardd: ability to switch between ELM safety params

sshane:

PR #28128, opened 2023-05-07
boardd: reset safety on onroad transition

### LINKED ISSUE #28050: controlsd: can get stuck if ignition is quickly cycled
I assume it's something to do with the blocking OBD multiplexing setting. Fixed by restarting car again

think it's this route https://connect.comma.ai/9b25e8c1484a1b67/1682583458867/1682583515939

sshane:
This bug always existed, it just wasn't as apparent before https://github.com/commaai/openpilot/pull/27656 since openpilot now waits for boardd to respond in the current startup step.

The issue is that it takes up to 4 seconds after ignition falling for manager to kill controlsd, so you can quickly restart ignition and let controlsd set `FirmwareQueryDone` before dying. When controlsd gets started up again a few seconds later, boardd is at the next step and does not write the read back parameter for switching OBD multiplexing modes resulting in controlsd hanging.

Before the change above a similar thing could happen where boardd switches to the safety mode before controlsd dies the first time, then the relay is open and the car is faulted until controlsd finally starts up again the second time.


sshane:
Without speeding up the power save mode setting in thermald.py to reduce the window where this is possible, I think the only two solutions are resetting the safety thread on a parameter write, so either controlsd exiting or starting. But I know we intentionally don't want to close the relay when restarting the device, so not sure if this is acceptable.

sshane:
Behavior table:

| Case | Previous Behavior | New Behavior |
| ------------- | ------------- | ------------- |
| Ignition Rising Edge | Start safety setter thread, clear ignition params | Clear ignition params |
| Ignition Falling Edge | Kill safety setter thread, clear ignition off params, set safety mode to `noOutput` | Kill safety setter thread, clear ignition off params, set safety mode to `noOutput` |
| IsOnroad rising edge | No change | Start safety setter thread |
| IsOnroad falling edge | No change | No change |

sshane:
@adeebshihadeh here's the flow explaining the current problem:

start ignition -> controlsd starts -> stop ignition -> start ignition -> controlsd writes `FirmwareQueryDone` and/or `CarParams` -> `deviceState.started` is finally false -> openpilot goes offroad -> openpilot goes onroad -> boardd not in correct step.

Even if you just have a param that controlsd writes to start the safety setter thread, the params controlsd previously wrote like `FirmwareQueryDone` and `CarParams` will not be cleared since they happened after the ignition on transition. I think the true fix here is to replace `everything` with `IsOnroad`, not even just the safety setter thread

sshane:
new PR: https://github.com/commaai/openpilot/pull/28151

PR #28134, opened 2023-05-08
camerad: write static AE settings only once

ZwX1616:


system/camerad/cameras/sensor2_i2c.h
```diff
@@ -752,6 +752,12 @@ struct i2c_random_wr_payload init_array_ox03c10[] = {
   {0x5486, 0x08}, {0x5487, 0xDE},
   {0x5686, 0x08}, {0x5687, 0xDE},
   {0x5886, 0x08}, {0x5887, 0xDE},
+
+  // fixed gains
+  {0x3588, 0x01}, {0x3589, 0x00},
```
adeebshihadeh, line 757: sure you want to hardcode these? can still use `ox03c10_analog_gains_reg`

ZwX1616: think it's fine, we already hardcode all kinds of things here that we are sure about

PR #28140, opened 2023-05-09
Ford: add alert for steering wheel angle not ready

sshane:
`Vehicle Sensors Calibrating` should be different enough from `Sensor Data Invalid` and `Calibration Invalid` so users don't confuse them when talking to support or similar.

Test route: 54827bf84c38b14f|2023-05-08--14-58-40--0


selfdrive/car/ford/carstate.py
```diff
@@ -16,6 +16,8 @@ def __init__(self, CP):
     if CP.transmissionType == TransmissionType.automatic:
       self.shifter_values = can_define.dv["Gear_Shift_by_Wire_FD1"]["TrnRng_D_RqGsm"]
 
+    self.vehicle_sensors_valid = False
```
sshane, line 19: we can make this a carState field either now or in the future when we start to check other quality flags

selfdrive/controls/lib/events.py
```diff
@@ -945,4 +945,10 @@ def joystick_alert(CP: car.CarParams, CS: car.CarState, sm: messaging.SubMaster,
     ET.NO_ENTRY: NoEntryAlert("LKAS Disabled"),
   },
 
+  EventName.vehicleSensorsInvalid: {
+    ET.PERMANENT: soft_disable_alert("LKAS Unavailable: Vehicle Sensors Calibrating"),
+    ET.IMMEDIATE_DISABLE: ImmediateDisableAlert("LKAS Unavailable: Vehicle Sensors Invalid"),
+    ET.NO_ENTRY: NoEntryAlert("LKAS Unavailable: Vehicle Sensors Calibrating"),
+  },
```
sshane, line 952: I want to let the user know there is an eventual end to the alert, but `Calibrating` might not make sense for all the cases we'll use this alert for.

selfdrive/car/ford/carstate.py
```diff
@@ -106,6 +109,7 @@ def get_can_parser(CP):
       ("AccStopMde_D_Rq", "EngBrakeData"),                   # PCM ACC standstill
       ("AccEnbl_B_RqDrv", "Cluster_Info1_FD1"),              # PCM ACC enable
       ("StePinComp_An_Est", "SteeringPinion_Data"),          # PSCM estimated steering angle (deg)
+      ("StePinCompAnEst_D_Qf", "SteeringPinion_Data"),       # PSCM estimated steering angle (quality flag)
```
sshane, line 112: This signal sometimes remains at 0 while the ABS is calibrating the steering pinion offset (`BrakeSnData_6->StePinOffst_An_Est`), and so the steering angle is a static 0 -- we should not engage here.

adeebshihadeh: this should be a comment in the code

selfdrive/controls/lib/events.py
adeebshihadeh: let's make it a good alert for the current case. we can think about other cases when they pop up

sshane: I think it's okay as is now then

sshane: look good?

selfdrive/controls/lib/events.py
```diff
@@ -945,4 +945,10 @@ def joystick_alert(CP: car.CarParams, CS: car.CarState, sm: messaging.SubMaster,
     ET.NO_ENTRY: NoEntryAlert("LKAS Disabled"),
   },
 
+  EventName.vehicleSensorsInvalid: {
+    ET.IMMEDIATE_DISABLE: ImmediateDisableAlert("Vehicle Sensors Invalid"),
+    ET.PERMANENT: NormalPermanentAlert("Vehicle Sensors Calibrating", "Drive to Calibrate"),
```
sshane, line 950: It recovers in ~30 seconds in a few routes I looked at, but it did not recover when I tested it today as I kept the car at a standstill due to flat tire. So it likely requires the vehicle to be driven.

PR #28144, opened 2023-05-09
ui/AnnotatedCameraWidget: only save and restore the changed pen and brush

deanlee:
save() and restore() is a convenient way to save and restore the entire painter state. it's expensive to repeatedly calling them in every drawing function.


adeebshihadeh:
Does this actually improve performance?

deanlee:
The initial test showed that it could save 0.5% to 1% of CPU usage. But I tested it again today and it seems that there is almost no improvement. So I closed this pull request.

PR #28146, opened 2023-05-09
VW MQB: Add FW for 2023 Volkswagen Touran

Saber422:
@jyoung8607 this is 2023 touran, we can extend the model year of VW Touran
0bbe367c98fa1538|2023-05-09--15-06-11--0


sshane:
@jyoung8607 reminder if you want to update the model years

PR #28151, opened 2023-05-10
onroad transition: clear params and start safety setter thread

### LINKED ISSUE #28050: controlsd: can get stuck if ignition is quickly cycled

sshane:
- manager clears params, writes `IsOnroad`, `IsOffroad`, and starts processes when going onroad
- boardd starts safety setter thread on `IsOnroad` rising edge, still keeps ignition as falling edge


sshane:
moves param clearing behavior out of https://github.com/commaai/openpilot/pull/28128 into manager

sshane:
An intermediate PR will be needed to have manager write the onroad params to avoid a race condition with boardd when it starts using `IsOnroad`. Not going to do here as we can cleanly split it out cleanly

sshane:
Actually can't merge this standalone as-is, since params like `FirmwareQueryDone` will still be around as we'll start the safety setter thread before we clear params. I'll merge a separate PR that just adds ignition off flags to `FirmwareQueryDone` and `ControlsReady` as either works.

PR #28168, opened 2023-05-11
ui/drawDriverState: don't draw circle bg twice

deanlee:
It looks like there is no need to draw the circle background again after  `drawIcon`. which not only adds some extra calls, but also makes the pixmap darker.

| Before  | After |
|   |  |


ZwX1616:
lgtm

PR #28170, opened 2023-05-11
Ford: catch camera ACC faults

sshane:
Catch cases where camera is faulted, but not PCM.

Just opened to verify with data later

If the IPMA gets bad quality flags from the EPS (any counter & checksum fine) it will raise this signal, for example


selfdrive/car/ford/carstate.py
```diff
@@ -55,6 +55,8 @@ def update(self, cp, cp_cam):
     ret.cruiseState.nonAdaptive = cp.vl["Cluster_Info1_FD1"]["AccEnbl_B_RqDrv"] == 0
     ret.cruiseState.standstill = cp.vl["EngBrakeData"]["AccStopMde_D_Rq"] == 3
     ret.accFaulted = cp.vl["EngBrakeData"]["CcStat_D_Actl"] in (1, 2)
+    if ret.openpilotLongitudinalControl:
```
sshane, line 58:
```suggestion
    if not ret.openpilotLongitudinalControl:
```

sshane:
looks good to me, IPMA ACC is never active when this is set, and there's only two real cases in the wild. one is caused by some blocked messages on a maverick, and one is unknown, but resolves in the next drives

PR #28173, opened 2023-05-12
UI/CameraWidget: remove deprecated background color

deanlee:
now `CameraWidget` is not responsible for drawing the color border. we no longer need to fill the background with `bg`.


adeebshihadeh:
Isn't this still needed to ensure a consistent background when there's no frame yet?

deanlee:
Yes... Clearing the background to black may look inconsisten.

PR #28175, opened 2023-05-12
Add fingerprint for 2022 Honda Passport

ecaslak:
Added fingerprint and changed docs to include support for 2022 Honda Passport.


sshane:
Can you add your dongle id?

ecaslak:
Dongle fff9af2c280cebe7

sshane:
thanks!

PR #28201, opened 2023-05-15
process_replay: sort logs before comparing them

fredyshox:


selfdrive/test/process_replay/compare_logs.py
```diff
@@ -70,7 +70,10 @@ def compare_logs(log1, log2, ignore_fields=None, ignore_msgs=None, tolerance=Non
     field_tolerances = {}
   default_tolerance = EPSILON if tolerance is None else tolerance
 
-  log1, log2 = (list(filter(lambda m: m.which() not in ignore_msgs, log)) for log in (log1, log2))
+  log1, log2 = (
+    sorted(list(filter(lambda m: m.which() not in ignore_msgs, log)), key=lambda m: (m.logMonoTime, m.which()))
```
adeebshihadeh, line 74: can you do the sort and filter at the same time?

fredyshox: should be simpler now

PR #28204, opened 2023-05-15
Subaru: rename legacy to preglobal

jnewb1:
**Description** "Legacy" is a confusing name because there is a car called the "Subaru Legacy", "Preglobal" is a better name.

**Verification** Should be functionally equivalent, just name changes

https://github.com/commaai/cereal/pull/445
https://github.com/commaai/panda/pull/1415

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "panda"]
   path = panda
-  url = ../../commaai/panda.git
+  url = ../../jnewb1/panda.git
 [submodule "opendbc"]
   path = opendbc
   url = ../../commaai/opendbc.git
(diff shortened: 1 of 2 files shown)


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "panda"]
   path = panda
-  url = ../../commaai/panda.git
+  url = ../../jnewb1/panda.git
```
adeebshihadeh, line 3: merged the other PRs, can you revert this?

adeebshihadeh (changes requested):

.gitmodules
jnewb1: done

adeebshihadeh:
The submodules will have to be updated to the versions on master.

jnewb1:
> The submodules will have to be updated to the versions on master.

done

adeebshihadeh:
Nice cleanup!

PR #28208, opened 2023-05-16
Nissan: use common steering pressed filtering

sshane:
this approach might be much better than what we do now. currently, no matter how hard you press, if it's under the time period it will not register, which doesn't work very well for makes like GM which have driver torque signals at 10Hz


sshane:
undecided if better

PR #28209, opened 2023-05-16
GM: steering pressed filtering

sshane:
This is actually quite bad, it activates all the time on minor bumps

### DIFF
--- a/selfdrive/car/gm/carstate.py
+++ b/selfdrive/car/gm/carstate.py
@@ -1,5 +1,6 @@
 import copy
 from cereal import car
+from common.filter_simple import FirstOrderFilter
 from common.conversions import Conversions as CV
 from common.numpy_fast import mean
 from opendbc.can.can_define import CANDefine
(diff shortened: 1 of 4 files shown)


twilsonco:
FWIW, gm driver torque data doesn't look noisy except when OP has the wheel.

Driver only torque: 
Screen Shot 2023-05-19 at 8 05 34 AM
Screen Shot 2023-05-19 at 8 08 28 AM


And now while OP is steering:
Screen Shot 2023-05-19 at 8 09 23 AM
Screen Shot 2023-05-19 at 8 09 57 AM

The driver torque oscillations are identical to those in the delivered LKA torque. 

If planned steer torque could be smoother with respect to time, so too would the driver torque that registers.

sshane:
The plots you compared are on two different scales (both x axis and y axis), so you can expect to see some more noise.

twilsonco:
Oh I see what you mean. Different from what I had in mind. Thanks.

sshane:
not ready yet

PR #28212, opened 2023-05-16
car docs: list all needed hardware

maxime-desroches:


selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,68 @@ class Star(Enum):
...
-  nissan_a = "Nissan A"
-  nissan_b = "Nissan B"
-  mazda = "Mazda"
-  ford_q3 = "Ford Q3"
-  ford_q4 = "Ford Q4"
-  none = "None"
-
-
-class HarnessPart(Enum):
+class CarPart(Enum):
```
sshane, line 33: can you separate the part categories with comments? like `# harnesses`, `# cables`, etc

sshane, line 33: can you separate the part categories with comments? like `# harnesses`, `# cables`, `# mounts` etc

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,68 @@ class Star(Enum):
...
+DEFAULT_CAR_PARTS: List[CarPart] = [CarPart.harness_box, CarPart.comma_power_v2, CarPart.rj45_cable_7ft]
 
 
 @dataclass
-class HarnessKit:
-  connector: Harness = Harness.none
-  parts: List[HarnessPart] = field(default_factory=lambda: copy.copy(DEFAULT_HARNESS_PARTS))
+class CarParts:
+  def __init__(self, parts: List[CarPart] = None):
+    self.parts = parts or []
```
sshane, line 90: for typing, I would just add `parts` initialized to an empty list and remove custom init

selfdrive/car/docs_definitions.py
```diff
@@ -154,8 +161,8 @@ class CarInfo:
   min_steer_speed: Optional[float] = None
   min_enable_speed: Optional[float] = None
 
-  # harness connectors + all the parts needed
-  harness_kit: HarnessKit = HarnessKit()
+  # all the parts needed to use this car
```
sshane, line 164:
```suggestion
  # all the parts needed for the supported car
```

selfdrive/car/docs_definitions.py
```diff
@@ -184,13 +191,13 @@ def init(self, CP: car.CarParams, all_footnotes: Dict[Enum, int]):
...
+    # hardware column
+    hardware_col = "None"
+    if self.car_parts.parts:
       model_years = self.model + (' ' + self.years if self.years else '')
-      harness_connector = f'- 1 <a href="https://comma.ai/shop/comma-three.html?make={self.make}&model={model_years}">{harness_col} connector</a>'
-      harness_parts = '<br>'.join([f"- {self.harness_kit.parts.count(part)} {part.value}" for part in sorted(set(self.harness_kit.parts), key=lambda part: part.value)])
-      harness_col = f'<details><summary>View</summary><sub>{harness_connector}<br>{harness_parts}</sub></details>'
+      buy_link = f'<a href="https://comma.ai/shop/comma-three.html?make={self.make}&model={model_years}">Buy Here</a>'
+      parts = '<br>'.join([f"- {self.car_parts.parts.count(part)} {part.value}" for part in sorted(set(self.car_parts.parts), key=lambda part: part.value)])
+      hardware_col = f'<details><summary>View</summary><sub>{parts}</sub><br><br>{buy_link}</details>'
```
sshane, line 200: try buy link inside the sup?

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,68 @@ class Star(Enum):
...
-class HarnessKit:
-  connector: Harness = Harness.none
-  parts: List[HarnessPart] = field(default_factory=lambda: copy.copy(DEFAULT_HARNESS_PARTS))
+class CarParts:
+  def __init__(self, parts: List[CarPart] = None):
+    self.parts = parts or []
+
+  @classmethod
+  def default(cls, parts: List[CarPart], default: List[CarPart] = None):
+    return cls(parts + (default or DEFAULT_CAR_PARTS))
```
sshane, line 94: either here or as a test, we should make sure that a connector part exists for all cars with a parts list

selfdrive/car/ford/values.py
```diff
@@ -60,15 +60,14 @@ class RADAR:
...
 @dataclass
 class FordCarInfo(CarInfo):
   package: str = "Co-Pilot360 Assist+"
-  harness_kit: HarnessKit = HarnessKit(Harness.ford_q3)
+  car_parts: CarParts = CarParts.default([CarPart.ford_q3_connector])
 
 
 CAR_INFO: Dict[str, Union[CarInfo, List[CarInfo]]] = {
-  CAR.BRONCO_SPORT_MK1: FordCarInfo("Ford Bronco Sport 2021-22"),
+  CAR.BRONCO_SPORT_MK1: FordCarInfo("Ford Bronco Sport 2021-22", car_parts=CarParts.default([CarPart.ford_q3_connector, CarPart.angled_mount])),
```
sshane, line 70: we need to add a normal mount for all other cars

sshane (changes requested):

selfdrive/car/docs_definitions.py
```diff
@@ -67,31 +68,46 @@ class CarPart(Enum):
...
+    p.parts = [part for part in parts + (default or DEFAULT_CAR_PARTS) if part not in (remove or [])]
+    return p
 
   @classmethod
-  def default(cls, parts: List[CarPart], default: List[CarPart] = None):
-    return cls(parts + (default or DEFAULT_CAR_PARTS))
+  def custom(cls, parts: List[CarPart]):
+    p = CarParts()
+    p.parts = parts.copy()
+    return p
```
sshane, line 110: why is this needed

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,84 @@ class Star(Enum):
...
-  connector: Harness = Harness.none
-  parts: List[HarnessPart] = field(default_factory=lambda: copy.copy(DEFAULT_HARNESS_PARTS))
+class CarParts:
+  parts: List[CarPart] = field(default_factory=list)
+
+  @classmethod
+  def default(cls, parts: List[CarPart], default: List[CarPart] = None, remove: List[CarPart] = None):
+    p = CarParts()
+    p.parts = [part for part in parts + (default or DEFAULT_CAR_PARTS) if part not in (remove or [])]
+    return p
```
sshane, line 104: if we don't use the `default` argument anywhere, let's remove it and maybe change `parts` to `add`

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,84 @@ class Star(Enum):
   EMPTY = "empty"
 
 
-class Harness(Enum):
```
adeebshihadeh, line 34: I think we still want the harness connectors to have their own enum, perhaps subclass instead

sshane: subclass should be fine, typing was a problem if it was a separate enum

selfdrive/car/tests/test_docs.py
```diff
@@ -74,8 +74,8 @@ def test_harnesses(self):
         if car.name == "comma body":
           raise unittest.SkipTest
 
-        self.assertNotIn(car.harness_kit.connector, [None, Harness.none], f"Need to specify car harness: {car.name}")
-        self.assertTrue(car.harness_kit.parts, f"Need to specify harness parts: {car.name} with {car.harness_kit.connector.value} connector")
+        self.assertTrue(car.car_parts.parts, f"Need to specify car parts: {car.name}")
```
sshane, line 77:
```suggestion
        self.assertTrue(len(car.car_parts.parts) > 0, f"Need to specify car parts: {car.name}")
```

i like being more clear here

sshane, line 77: can you think of any more good tests? like you need at least one mount type, etc.

maxime-desroches: The only overlapping prerequisite for all cars is that they need a mount, a connector and some cables. We currently test for all three. Do we want to enforce anything else? We could enforce stuff like a single connector, but that might not be true forever.

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,86 @@ class Star(Enum):
...
+class PartType(Enum):
+  harness_connector = "Harness Connector"
+  generic_part = "Generic Part"
+  mount = "Mount"
+
+
+Part = namedtuple("Part", ["value","type"], defaults=("",PartType.generic_part))
+
+
+class CarPart(Enum):
```
sshane, line 42: think this would look cleaner if you used multiple Enums that subclassed from a base

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,83 @@ class Star(Enum):
...
+  obd_c_cable_1_5ft = "OBD-C cable (1.5 ft)"
 
 
-DEFAULT_HARNESS_PARTS: List[HarnessPart] = [HarnessPart.harness_box, HarnessPart.comma_power_v2, HarnessPart.rj45_cable]
+class Device(CarPart):
+  comma_3 = "comma 3"
+  red_panda = "red panda"
+
+
+DEFAULT_CAR_PARTS: List[CarPart] = [HarnessAccessory.harness_box, HarnessAccessory.comma_power_v2, Cable.rj45_cable_7ft, Mount.mount]
```
sshane, line 99: think every comma three comes with a 1.5ft OBD C cable and is needed (except for some cars like Nissan? I would check)

adeebshihadeh: yes they all do. some need longer cables, but that's done with a coupler and long comma two cable

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,83 @@ class Star(Enum):
...
-  ford_q4 = "Ford Q4"
-  none = "None"
-
-
-class HarnessPart(Enum):
+class CarPart(Enum):
+  pass
+
+
+class HarnessConnector(CarPart):
```
sshane, line 37: of the two options, enums and sub-enums vs. one CarPart class with a type field, which do you think will be easier to maintain?

maxime-desroches: The only one enum CarPart option would allow to easily work with jinja for the website (eg: get the connector in all the parts for the vehicles_template.json and do special restrictions for diff part types). Otherwise, doing this will be very dirty. 

It would also allow to add other fields (links to the shop for each parts) and have ALL the parts in a single import and single enum.

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,83 @@ class Star(Enum):
   EMPTY = "empty"
 
 
-class Harness(Enum):
```
sshane, line 34: How about you have dataclasses for each part type, and one big enum to store everything in?

something like:

```python
@dataclass
class BasePart:
  name: str

  def __repr__(self):
    return self.name

  def __str__(self):
    return self.name

  @property
  def type(self) -> str:
    raise NotImplementedError


@dataclass
class Connector(BasePart):
  @property
  def type(self) -> str:
    return 'connector'


@dataclass
class Mount(BasePart):
  @property
  def type(self) -> str:
    return 'mount'


class CarPart(Enum):
  nidec = Connector("Honda Nidec connector")
  ...
  mount = Mount("mount")
  ...
```

selfdrive/car/ford/values.py
```diff
@@ -62,12 +62,14 @@ class RADAR:
...
 
-
 @dataclass
 class FordCarInfo(CarInfo):
   package: str = "Co-Pilot360 Assist+"
-  harness_kit: HarnessKit = HarnessKit(Harness.ford_q3)
+  car_parts: CarParts = CarParts.default([CarPart.ford_q3])
 
+  def init_make(self, CP: car.CarParams):
+    if self in (CAR_INFO.get(CAR.BRONCO_SPORT_MK1), CAR_INFO.get(CAR.MAVERICK_MK1)):
```
sshane, line 71:
```suggestion
    if CP.carFingerprint in (CAR.BRONCO_SPORT_MK1, CAR.MAVERICK_MK1):
```

selfdrive/car/ford/values.py
```diff
@@ -62,12 +62,14 @@ class RADAR:
...
-
 @dataclass
 class FordCarInfo(CarInfo):
   package: str = "Co-Pilot360 Assist+"
-  harness_kit: HarnessKit = HarnessKit(Harness.ford_q3)
+  car_parts: CarParts = CarParts.default([CarPart.ford_q3])
 
+  def init_make(self, CP: car.CarParams):
+    if CP.carFingerprint in (CAR.BRONCO_SPORT_MK1, CAR.MAVERICK_MK1):
+      self.car_parts = CarParts.default([CarPart.ford_q3, CarPart.angled_mount], remove=[CarPart.mount])
```
sshane, line 72:
```suggestion
      self.car_parts = CarParts.default([CarPart.ford_q3, CarPart.angled_mount], remove=[CarPart.mount])

```

selfdrive/car/tests/test_docs.py
```diff
@@ -74,8 +74,10 @@ def test_harnesses(self):
         if car.name == "comma body":
           raise unittest.SkipTest
 
-        self.assertNotIn(car.harness_kit.connector, [None, Harness.none], f"Need to specify car harness: {car.name}")
-        self.assertTrue(car.harness_kit.parts, f"Need to specify harness parts: {car.name} with {car.harness_kit.connector.value} connector")
+        self.assertTrue(len(car.car_parts.parts) > 0, f"Need to specify car parts: {car.name}")
+        self.assertTrue(len(list(filter((lambda p: p.value.type is PartType.connector), car.car_parts.parts))) == 1, f"Need to specify exactly ONE harness connector: {car.name}")
+        self.assertTrue(len(list(filter((lambda p: p.value.type is PartType.mount), car.car_parts.parts))) == 1, f"Need to specify exactly ONE mount: {car.name}")
+        self.assertTrue(len(list(filter((lambda p: p is CarPart.right_angle_obd_c_cable_1_5ft), car.car_parts.parts))) > 0, f"Need to specify a right angle OBD-C cable (1.5ft): {car.name}")
```
sshane, line 80:
```suggestion
        car_part_types = [p.value.type for p in car.car_parts.parts]
        self.assertTrue(len(car.car_parts.parts) > 0, f"Need to specify car parts: {car.name}")
        self.assertTrue(car_part_types.count(PartType.connector), f"Need to specify exactly ONE harness connector: {car.name}")
        self.assertTrue(car_part_types.count(PartType.mount) == 1, f"Need to specify exactly ONE mount: {car.name}")
        self.assertTrue(CarPart.right_angle_obd_c_cable_1_5ft in car.car_parts.parts, f"Need to specify a right angle OBD-C cable (1.5ft): {car.name}")
```

this is a bit cleaner

selfdrive/car/ford/values.py
maxime-desroches: I don't see a change between those two lines?

maxime-desroches: nvm

selfdrive/car/nissan/values.py
```diff
@@ -30,21 +30,21 @@ class CAR:
...
 
-NISSAN_HARNESS_PARTS = [HarnessPart.harness_box, HarnessPart.rj45_cable, HarnessPart.long_obdc_cable, HarnessPart.usbc_coupler]
+NISSAN_PARTS = [CarPart.harness_box, CarPart.rj45_cable_7ft, CarPart.long_obdc_cable, CarPart.usbc_coupler, CarPart.mount, CarPart.right_angle_obd_c_cable_1_5ft]
 
 
 @dataclass
 class NissanCarInfo(CarInfo):
   package: str = "ProPILOT Assist"
-  harness_kit: HarnessKit = HarnessKit(Harness.nissan_a, parts=NISSAN_HARNESS_PARTS)
+  car_parts: CarParts = CarParts.default([CarPart.nissan_a], default=NISSAN_PARTS)
```
sshane, line 39: now you can get rid of default argument and do something like
```suggestion
  car_parts: CarParts = CarParts([CarPart.nissan_a] + NISSAN_PARTS)
```

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,120 @@ class Star(Enum):
...
+
+DEFAULT_CAR_PARTS: List[CarPart] = [CarPart.harness_box, CarPart.comma_power_v2, CarPart.rj45_cable_7ft, CarPart.mount, CarPart.right_angle_obd_c_cable_1_5ft]
+
+
+@dataclass
+class CarParts:
+  parts: List[CarPart] = field(default_factory=list)
+
+  @classmethod
+  def default(cls, add: List[CarPart] = None, default: List[CarPart] = None, remove: List[CarPart] = None):
```
sshane, line 144: remove default argument

sshane:
It would be nice if the car docs diff bot worked on these kinds of PRs! But I couldn't get it to save the structure in the pickle file like it says is possible. If you have extra time, I think we want to get rid of all the custom pickle stuff and just parse the markdown file directly, which should be much simpler. https://github.com/commaai/openpilot/blob/master/selfdrive/debug/dump_car_info.py https://github.com/commaai/openpilot/blob/master/selfdrive/debug/print_docs_diff.py

maxime-desroches:
> It would be nice if the car docs diff bot worked on these kinds of PRs! But I couldn't get it to save the structure in the pickle file like it says is possible. If you have extra time in the future, I think we want to get rid of all the custom pickle stuff and just parse the markdown file directly, which should be simpler. https://github.com/commaai/openpilot/blob/master/selfdrive/debug/dump_car_info.py https://github.com/commaai/openpilot/blob/master/selfdrive/debug/print_docs_diff.py

I was wondering why the bot never showed up. I will take a look

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,120 @@ class Star(Enum):
...
+
+DEFAULT_CAR_PARTS: List[CarPart] = [CarPart.harness_box, CarPart.comma_power_v2, CarPart.rj45_cable_7ft, CarPart.mount, CarPart.right_angle_obd_c_cable_1_5ft]
+
+
+@dataclass
+class CarParts:
+  parts: List[CarPart] = field(default_factory=list)
+
+  @classmethod
+  def default(cls, add: List[CarPart] = None, remove: List[CarPart] = None):
```
sshane, line 144: maybe `common` would be a better name, your choice

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,120 @@ class Star(Enum):
...
 
 
 @dataclass
-class HarnessKit:
-  connector: Harness = Harness.none
-  parts: List[HarnessPart] = field(default_factory=lambda: copy.copy(DEFAULT_HARNESS_PARTS))
+class Part:
+  name: str
+
+  @property
```
sshane, line 45: maybe use abstractmethod decorator and test that mypy/pylint catches subclass not implementing it

selfdrive/car/ford/values.py
```diff
@@ -62,11 +62,14 @@ class RADAR:
 
 DBC: Dict[str, Dict[str, str]] = defaultdict(lambda: dbc_dict("ford_lincoln_base_pt", RADAR.DELPHI_MRR))
```
sshane, line 64:
```suggestion


```

selfdrive/car/gm/values.py
```diff
@@ -89,9 +89,9 @@ class GMCarInfo(CarInfo):
 
   def init_make(self, CP: car.CarParams):
     if CP.networkLocation == car.CarParams.NetworkLocation.fwdCamera:
-      self.harness_kit = HarnessKit(Harness.gm)
+      self.car_parts = CarParts.default([CarPart.gm])
     else:
-      self.harness_kit = HarnessKit(Harness.obd_ii, parts=[HarnessPart.long_obdc_cable, HarnessPart.usbc_coupler])
+      self.car_parts = CarParts([CarPart.obd_ii, CarPart.long_obdc_cable, CarPart.usbc_coupler, CarPart.mount, CarPart.right_angle_obd_c_cable_1_5ft])
```
sshane, line 94: use .default/common and just add long USB C and coupler?

selfdrive/car/nissan/values.py
```diff
@@ -30,21 +30,21 @@ class CAR:
   ALTIMA = "NISSAN ALTIMA 2020"
 
 
-NISSAN_HARNESS_PARTS = [HarnessPart.harness_box, HarnessPart.rj45_cable, HarnessPart.long_obdc_cable, HarnessPart.usbc_coupler]
+NISSAN_PARTS: List[CarPart] = [CarPart.harness_box, CarPart.rj45_cable_7ft, CarPart.long_obdc_cable, CarPart.usbc_coupler, CarPart.mount, CarPart.right_angle_obd_c_cable_1_5ft]
```
sshane, line 33: same, we can just add long usb c, coupler (car docs bot should catch changing parts lists). unless you think it's better to list everything to be explicit

selfdrive/car/tests/test_docs.py
```diff
@@ -74,8 +74,11 @@ def test_harnesses(self):
         if car.name == "comma body":
           raise unittest.SkipTest
 
-        self.assertNotIn(car.harness_kit.connector, [None, Harness.none], f"Need to specify car harness: {car.name}")
-        self.assertTrue(car.harness_kit.parts, f"Need to specify harness parts: {car.name} with {car.harness_kit.connector.value} connector")
+        car_part_type = [p.value.type for p in car.car_parts.parts]
+        self.assertTrue(len(car.car_parts.parts) > 0, f"Need to specify car parts: {car.name}")
+        self.assertTrue(car_part_type.count(PartType.connector) == 1, f"Need to specify one harness connector: {car.name}")
+        self.assertTrue(car_part_type.count(PartType.mount) == 1, f"Need to specify one mount: {car.name}")
+        self.assertTrue(CarPart.right_angle_obd_c_cable_1_5ft in car.car_parts.parts, f"Need to specify a right angle OBD-C cable (1.5ft): {car.name}")
```
sshane, line 81: if there's any other common parts you can check them here too, else just close this

selfdrive/car/volkswagen/values.py
```diff
@@ -160,13 +160,13 @@ class Footnote(Enum):
...
     "in software, but doesn't yet have a harness available from the comma store.",
-    Column.HARNESS)
+    Column.HARDWARE)
 
 
 @dataclass
 class VWCarInfo(CarInfo):
   package: str = "Adaptive Cruise Control (ACC) & Lane Assist"
-  harness_kit: HarnessKit = HarnessKit(Harness.j533, parts=[HarnessPart.harness_box, HarnessPart.long_obdc_cable, HarnessPart.usbc_coupler])
+  car_parts: CarParts = CarParts([CarPart.j533, CarPart.harness_box, CarPart.long_obdc_cable, CarPart.usbc_coupler, CarPart.mount, CarPart.right_angle_obd_c_cable_1_5ft])
```
sshane, line 169: same as other two comments

sshane (approved):

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +30,120 @@ class Star(Enum):
...
+
+
+@dataclass
+class CarParts:
+  parts: List[CarPart] = field(default_factory=list)
+
+  @classmethod
+  def default(cls, add: List[CarPart] = None, remove: List[CarPart] = None):
+    p = [part for part in (add or []) + DEFAULT_CAR_PARTS if part not in (remove or [])]
+    return cls(p)
```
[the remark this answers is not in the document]
maxime-desroches: I will trust your sens of English nomenclature over mine

selfdrive/car/nissan/values.py
maxime-desroches: since we have so few exceptions (OBD II, J533 and Nissan), I think it is worth being explicit

selfdrive/car/tests/test_docs.py
maxime-desroches: the intersection for the moment is only a mount and a connector (unless we add the comma3 or a panda in the parts)

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +31,117 @@ class Star(Enum):
...
 @dataclass
-class HarnessKit:
-  connector: Harness = Harness.none
-  parts: List[HarnessPart] = field(default_factory=lambda: copy.copy(DEFAULT_HARNESS_PARTS))
+class AbstractDataClass:
+  name: str
+
+
+class Part(ABC, AbstractDataClass):
+  @abstractmethod
```
sshane, line 48: you need both property and abstract

selfdrive/car/docs_definitions.py
```diff
@@ -31,60 +31,117 @@ class Star(Enum):
...
+  accessory = "Accessory"
+  mount = "Mount"
 
 
 @dataclass
-class HarnessKit:
-  connector: Harness = Harness.none
-  parts: List[HarnessPart] = field(default_factory=lambda: copy.copy(DEFAULT_HARNESS_PARTS))
+class AbstractDataClass:
+  name: str
```
sshane, line 44: why both this and `Part`?

selfdrive/car/tests/test_docs.py
sshane: should add comma three too (can do it in the next PR)

selfdrive/car/docs_definitions.py
sshane: Why no dataclass? Can you not make a base abstract dataclass?

maxime-desroches: mypy does not support a class being both a dataclass and having abstractmethod ([see this issue](https://github.com/python/mypy/issues/5374)).

We could have done something like this to make it work:
```
@dataclass
class PartDataClass:
  name: str


class PartAbstractClass(ABC):
  @property
  @abstractmethod
  def type(self) -> PartType:
    raise NotImplementedError


class Part(PartAbstractClass, PartDataClass):
  pass


class Connector(Part):
  @property
   def type(self) -> PartType:
     return ....

```

but I prefered just removing the dataclass from `Part` and just adding an `__init__` which gives the same result but is cleaner imo
```
class Part(ABC):
  def __init__(self, name: str):
    self.name = name

  @property
  @abstractmethod
  def type(self) -> PartType:
    raise NotImplementedError
```

PR #28237, opened 2023-05-19
Test mapsd

mitchellgoffpc:


selfdrive/navd/map_renderer.cc
```diff
@@ -160,6 +160,13 @@ void MapRenderer::update() {
   gl_functions->glFlush();
   double end_t = millis_since_boot();
 
+  while (TEST_MODE && !loaded()) {
```
adeebshihadeh, line 163: why is this only needed in test mode? seems like this is useful to have while running live

selfdrive/test/process_replay/test_processes.py
```diff
@@ -40,23 +40,23 @@
...
-  ("CHRYSLER", "aregen99B094E1E2E|2023-05-10--14-41-40--0"), 
-  ("RAM", "aregen5C2487E1EEB|2023-05-10--14-44-09--0"), 
-  ("SUBARU", "aregen98D277B792E|2023-05-10--14-46-46--0"), 
-  ("GM", "aregen377BA28D848|2023-05-10--14-48-28--0"), 
-  ("GM2", "aregen7CA0CC0F0C2|2023-05-10--14-51-00--0"), 
-  ("NISSAN", "aregen7097BF01563|2023-05-10--14-52-43--0"), 
-  ("VOLKSWAGEN", "aregen765AF3D2CB5|2023-05-10--14-54-23--0"), 
-  ("MAZDA", "aregen3053762FF2E|2023-05-10--14-56-53--0"), 
-  ("FORD", "aregenDDE0F89FA1E|2023-05-10--14-59-26--0"), 
+  ("BODY", "aregenECF15D9E559|2023-05-10--14-26-40--0"),
```
adeebshihadeh, line 43: can you revert the changes to this file?

selfdrive/navd/map_renderer.cc
mitchellgoffpc: might be best to have this running live as well, just need to make sure the loop isn't too tight

.github/workflows/selfdrive_tests.yaml
```diff
@@ -250,6 +250,8 @@ jobs:
       run: |
         ${{ env.RUN }} "CI=1 coverage run selfdrive/test/process_replay/test_processes.py -j$(nproc) && \
                         coverage xml"
+      env:
+        MAPBOX_TOKEN: ${{ secrets.MAPBOX_TOKEN }}
```
adeebshihadeh, line 254: this has to be passed into the docker container

.github/workflows/selfdrive_tests.yaml
```diff
@@ -248,7 +248,7 @@ jobs:
         ${{ env.RUN }} "scons -j$(nproc)"
     - name: Run replay
       run: |
-        ${{ env.RUN }} "CI=1 coverage run selfdrive/test/process_replay/test_processes.py -j$(nproc) && \
+        ${{ env.RUN }} -e MAPBOX_TOKEN=${{ secrets.MAPBOX_TOKEN }} "CI=1 coverage run selfdrive/test/process_replay/test_processes.py -j$(nproc) && \
```
adeebshihadeh, line 251: this should be added to the default run command

PR #28245, opened 2023-05-21
GM: signed vehicle speed

sshane:
instead of https://github.com/commaai/openpilot/pull/28214

bit more work to verify, but this should be the correct way to handle it


sshane:
- [ ] Verify wherever vEgo is used, it doesn't abs() it, or assume it's always positive

PR #28255, opened 2023-05-22
Calibrationd: allow more pitch angle

haraschax:


incognitojam:
Should the description be updated in the settings here? https://github.com/commaai/openpilot/blob/master/selfdrive/ui/qt/offroad/settings.cc#L268

PR #28256, opened 2023-05-22
car docs: parts use dataclass

sshane:


sshane:
eh not that bad, will revisit if parts class grows to needing a dataclass

PR #28257, opened 2023-05-22
params: specify path using PARAMS_ROOT

fredyshox:
Needed for OP in the pipeline in order to reduce disk use. 

CC: @gregjhogan

### DIFF
--- a/system/hardware/hw.h
+++ b/system/hardware/hw.h
@@ -19,10 +19,7 @@ inline std::string log_root() {
   return Hardware::PC() ? util::getenv("HOME") + "/.comma/media/0/realdata" : "/data/media/0/realdata";
 }
 inline std::string params() {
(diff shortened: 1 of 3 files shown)


system/hardware/hw.h
```diff
@@ -19,6 +19,9 @@ inline std::string log_root() {
   return Hardware::PC() ? util::getenv("HOME") + "/.comma/media/0/realdata" : "/data/media/0/realdata";
 }
 inline std::string params() {
+  if (const char *root = getenv("PARAMS_ROOT")) {
+    return root;
+  }
   return Hardware::PC() ? util::getenv("HOME") + "/.comma/params" : "/data/params";
```
adeebshihadeh, line 25: i'd use util::getenv with a default

fredyshox: Done. Also changed getenv signature to have matching types for return value and `default_val`

PR #28260, opened 2023-05-23
car docs: pass all car parts + types to template

maxime-desroches:


selfdrive/car/docs.py
```diff
@@ -61,7 +61,8 @@ def generate_cars_md(all_car_info: List[CarInfo], template_fn: str) -> str:
     template = jinja2.Template(f.read(), trim_blocks=True, lstrip_blocks=True)
 
   footnotes = [fn.value.text for fn in get_all_footnotes()]
-  cars_md: str = template.render(all_car_info=all_car_info, group_by_make=group_by_make,
+  cars_md: str = template.render(all_car_info=all_car_info, car_parts=CarPart,
+                                 part_types=PartType, group_by_make=group_by_make,
```
sshane, line 65: oh I would keep the variable name in the template the same, so `PartType=PartType` so you can easily search from the template to find the source variable

PR #28264, opened 2023-05-23
car docs: allow auto_resume override

maxime-desroches:


sshane:
Can you use the row down below? `elif CP.autoResumeSng:`

PR #28266, opened 2023-05-23
car docs: fix diff bot

sshane:
opened to test, merged to fix.

pickle couldn't figure out which Part object in the CarPart enum belonged to which Part when depickle-ing it since we didn't implement repr, str, or eq. use dataclass which does this for us

### DIFF
--- a/selfdrive/car/docs_definitions.py
+++ b/selfdrive/car/docs_definitions.py
@@ -1,5 +1,4 @@
 import re
-from abc import ABC, abstractmethod
 from collections import namedtuple
 from dataclasses import dataclass, field
 from enum import Enum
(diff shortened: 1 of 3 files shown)


selfdrive/car/docs_definitions.py
```diff
@@ -43,6 +43,15 @@ class Part(ABC):
...
 
+  def __str__(self):
+    return self.name + self.type.value
+
+  def __repr__(self):
+    return self.name + self.type.value
+
+  def __eq__(self, other):
+    return self.name == other.name and self.type.value == other.type.value
+
```
sshane, line 54: it's definitely worth it making this a dataclass so we don't need to manually override all this. running docs in CI should catch the abstractmethod (instead of using ABC due to that mypy)

sshane:
@maxime-desroches any simple test to catch this would be great

PR #28269, opened 2023-05-23
UI: move experimental toggle to right

incognitojam:
- Move the experimental toggle button to the right column to make more space for https://github.com/commaai/openpilot/pull/27629
- Move prime ad to the left when no prime

||New|Old|
|-|-|-|
|Unpaired|||
|No prime|||
|Prime|||


deanlee:
It looks like the margin above the brand in the original UI looks too large. Is it necessary to make an adjustment?

PR #28276, opened 2023-05-24
ui/MainWindow: use switch in eventFilter

deanlee:
Because we have installed this event filter for the entire app, it would be better to return as soon as possible.  https://stackoverflow.com/questions/6860525/c-what-is-faster-lookup-in-hashmap-or-switch-statement

Also, I think using a switch statement here may be more readable.


Silvrbckw:
Merge

PR #28278, opened 2023-05-24
Assisted GPS qcom

haraschax:

### DIFF
--- a/system/sensord/rawgps/rawgpsd.py
+++ b/system/sensord/rawgps/rawgpsd.py
@@ -109,6 +109,31 @@ def gps_enabled() -> bool:
   except subprocess.CalledProcessError as exc:
     raise Exception("failed to execute QGPS mmcli command") from exc
 
+def download_and_inject_assistance():
+  assist_data_file = '/tmp/xtra3grc.bin' 
+  assistance_url = 'http://xtrapath3.izatcloud.net/xtra3grc.bin'
+  try:
+    c = pycurl.Curl() 
+    c.setopt(c.URL, 'http://www.alfe.de')
+    c.setopt(c.NOBODY, 1)
+    c.perform()
+    print(c.getinfo(c.CONTENT_LENGTH_DOWNLOAD))
+    with open(assist_data_file, "wb") as fp:
+      curl = pycurl.Curl()
+      curl.setopt(pycurl.URL, assistance_url)
+      curl.setopt(pycurl.CONNECTTIMEOUT, 5)
+
+      curl.setopt(pycurl.WRITEDATA, fp)
+      curl.perform()
+      curl.close()
+  except pycurl.error as e:
+    cloudlog.exception(f'Failed to download assistance file with error: {e}')
+  if os.path.isfile(assist_data_file):
+    try:
+      subprocess.check_call(f"mmcli -m any --timeout 30 --location-inject-assistance-data={assist_data_file}", shell=True)
+    except subprocess.CalledProcessError:
+      cloudlog.exception("rawgps.mmcli_command_failed")
+
 def setup_quectel(diag: ModemDiag):
   # enable OEMDRE in the NV
   # TODO: it has to reboot for this to take effect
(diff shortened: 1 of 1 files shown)


system/sensord/rawgps/rawgpsd.py
```diff
@@ -120,13 +122,37 @@ def setup_quectel(diag: ModemDiag):
...
   # disable DPO power savings for more accuracy
   at_cmd("AT+QGPSCFG=\"dpoenable\",0")
   # don't automatically turn on GNSS on powerup
   at_cmd("AT+QGPSCFG=\"autogps\",0")
 
-  at_cmd("AT+QGPSSUPLURL=\"supl.google.com:7275\"")
+  # Do internet assistance
+  at_cmd("AT+QGPSXTRA=1")
+  assist_data_file = '/tmp/xtra3grc.bin'
+  assistance_url = 'http://xtrapath3.izatcloud.net/xtra3grc.bin'
```
adeebshihadeh, line 135: some things to consider here:
* file size sanity check before downloading. `/tmp/` is a tmpfs, so need to be careful about filling it up
* remove the file once it's uploaded to the module
* probably want some retry logic (e.g. going from wifi -> lte), and potentially put it in a thread to not block startup

system/sensord/rawgps/rawgpsd.py
```diff
@@ -120,13 +122,37 @@ def setup_quectel(diag: ModemDiag):
...
+  except pycurl.error as e:
+    cloudlog.exception(f'Failed to download assistance file with error: {e}')
+  if os.path.isfile(assist_data_file):
+    try:
+      subprocess.check_call(f"mmcli -m any --timeout 30 --location-inject-assistance-data={assist_data_file}", shell=True)
+    except subprocess.CalledProcessError:
+      cloudlog.exception("rawgps.mmcli_command_failed")
+  #at_cmd("AT+QGPSXTRADATA?")
+  time_str = datetime.utcnow().strftime("%Y/%m/%d,%H:%M:%S")
+  at_cmd(f"AT+QGPSXTRATIME=0,\"{time_str}\",1,1,1000")
```
adeebshihadeh, line 154: docs say time has to go first. perhaps ModemManager is already doing this for us in `--location-inject-assistance-data`?

system/sensord/rawgps/rawgpsd.py
```diff
@@ -120,13 +122,37 @@ def setup_quectel(diag: ModemDiag):
...
+  #at_cmd("AT+QGPSDEL=0")
 
   # disable DPO power savings for more accuracy
   at_cmd("AT+QGPSCFG=\"dpoenable\",0")
   # don't automatically turn on GNSS on powerup
   at_cmd("AT+QGPSCFG=\"autogps\",0")
 
-  at_cmd("AT+QGPSSUPLURL=\"supl.google.com:7275\"")
+  # Do internet assistance
+  at_cmd("AT+QGPSXTRA=1")
```
adeebshihadeh, line 133: should we also remove the SUPL configuration? might make it difficult to debug with multiple assistance sources

system/sensord/rawgps/rawgpsd.py
haraschax: Can’t block or put in thread since this is only possible with the gps engine off. If it fails it fails, that’s ok.

will add file size check and delete

system/sensord/rawgps/rawgpsd.py
haraschax: I tested both ways seemed the same. And no modemmanager does not not do it for us

system/sensord/rawgps/rawgpsd.py
haraschax: Yes that’s why I removed the url. I never got it to work anyway, and the other approach appears to be the standard.

adeebshihadeh: ah missed that it was deleted

PR #28286, opened 2023-05-24
process_replay: fix crash on segments without carParams

fredyshox:


selfdrive/test/process_replay/process_replay.py
```diff
@@ -347,10 +347,10 @@ def replay_process(cfg, lr, fingerprint=None):
 
       assert controlsState is not None and initialized, "controlsState never initialized"
 
-    CP = [m for m in lr if m.which() == 'carParams'][0].carParams
     if fingerprint is not None:
       setup_env(cfg=cfg, controlsState=controlsState, lr=lr, fingerprint=fingerprint)
     else:
+      CP = next((m.carParams for m in lr if m.which() == 'carParams'), None)
```
adeebshihadeh, line 353: some processes need carParams. can you add an assert with a good error message for that case?

fredyshox: done

PR #28290, opened 2023-05-25
car docs: store parts as dataclasses

sshane:
This is my best attempt at satisfying everything we want from the docs:
- One big enum we can import and reference (instead of importing CarHarness, CarMount, CarAccessory...)
- Ability to reference by attribute (CarPart.mount, not CarPart['mount'])
- Static so we maintain typing
- Ability to set attributes on a usage-basis of each part (what this PR adds)

Uses the name to sort parts, I'll move that out


sshane:
I don't think this is very extensible, if you want to specify the parent parts/what it comes with you have to do it for every part, duplicated every line... https://github.com/commaai/openpilot/pull/28292 might be better

PR #28292, opened 2023-05-25
car docs: document parts of parts

sshane:
Includes old and new code (why it's only a green diff)

This aims to make it easy to document and maintain the parts that come in the shop and the parts required to drive the car.

- [x] clean up old parts code
- [x] use new code for all cars
- [x] ensure we can make a template using this for the site
- [ ] (another PR) display all openpilot hardware and their sub parts somewhere in a md file on GitHub, then only list top level parts on CARS.md and site (comma three, Hyundai A connector) with links?


selfdrive/car/docs_definitions.py
```diff
@@ -38,6 +38,130 @@ class PartType(Enum):
...
+    for part in getattr(self, _parts):
+      parts.extend(part.value.all_parts(required))
+
+    return parts
+
+
+class EnumBase(Enum):
+  @classmethod
+  def type(cls):
+    return PartTypeNew(cls)
```
sshane, line 69: this allows similar type checking in the Jinja template as before (no isinstance required)

sshane:
This allows you to specify sub-parts recursively for any part. So you can separate car harness, red panda kit, and comma three parts for the shop. You can also specify to only show the required parts on the docs and then have all parts in the shop.

It still may be a bit complicated, but I don't know if the alternative with the flat parts enum will be any easier to maintain. I tried to encapsulate all of the specifics of the part shop into one place, instead of having logic spread across openpilot, jinja website templates, and javascript which eventually will get out of hand.

@maxime-desroches what do you think about something like this? It's also possible we don't need this complexity just yet.

maxime-desroches:
Yes this is pretty much exactly what I was about to write but with kits. It does add complexity + redundant/duplicate code, but at least it is in only one place and is in Python which make it much more manageable.

sshane:
feel free to take this over and make changes as you wish then

selfdrive/car/docs_definitions.py
```diff
@@ -30,120 +30,144 @@ class Star(Enum):
...
-
+# A part + its comprised parts
 @dataclass
-class Part:
+class BasePart:
   name: str
+  # A list of parts that come with this part in the shop
+  parts: List[Enum] = field(default_factory=list)
+  # If there are extra parts included, you can override required_parts
+  required_parts: Optional[List[Enum]] = None
```
sshane, line 40: I thought we decided to keep everything extra/non-required in web3, @adeebshihadeh have any preference for required parts vs. what the shop includes?

selfdrive/car/docs_definitions.py
```diff
@@ -214,7 +238,7 @@ class CarInfo:
   auto_resume: Optional[bool] = None
 
   # all the parts needed for the supported car
-  car_parts: CarParts = CarParts()
+  car_parts: Optional[CarParts] = None
```
sshane, line 241:
```suggestion
  car_parts: CarParts = CarParts()
```
?

selfdrive/car/docs_definitions.py
adeebshihadeh: yeah extra stuff shop includes should go in web3

sshane: @maxime-desroches let's remove the required stuff then. I added to bring everything into openpilot to see what it would look like, but it makes more sense in web3

PR #28293, opened 2023-05-25
car docs: sort parts by shown name

sshane:


maxime-desroches (changes requested):
This won't pass mypy. I tried to do this exact change in the past.

edit: yeah you can see in the github actions: 
`selfdrive/car/docs_definitions.py:254: error: Returning Any from function declared to return "Union[SupportsDunderLT, SupportsDunderGT]"`

PR #28302, opened 2023-05-26
VW MQB: Angled mounts for Transporter and Crafter

jyoung8607:
Thanks to Ford for getting comma to make these so we can use them too! @sshane @maxime-desroches does this directly control order fulfillment, or do I need to request store purchase flow changes separately?


maxime-desroches:
Angled mounts are currently out of stock, so once this is merged we will update the website to temporary list those cars as "available soon" like we currently do for the Ford Bronco.

incognitojam:
Do you know what angle is ideal for these cars? The angled mount we stock is the same as adding an 8 degree wedge

jyoung8607:
I don't think we've had anyone do an actual measurement to find the ideal angle, but the affected owners have been getting by just fine with locally printed versions of the 8 degree mount from the neo repo. So, it'll do.

PR #28306, opened 2023-05-26
Tesla: explicitly copy signals to forward

sshane:
split from https://github.com/commaai/openpilot/pull/28020


sshane:
new tesla label!

PR #28307, opened 2023-05-26
Nissan: use generated DBCs

vanillagorillaa:
Requires https://github.com/commaai/opendbc/pull/841


sshane:
also need to fix the panda safety test now that this is merged

vanillagorillaa:
> also need to fix the panda safety test now that this is merged

https://github.com/commaai/panda/pull/1438

PR #28309, opened 2023-05-26
cabana: support multiplexed signals

deanlee:
~depends on https://github.com/commaai/opendbc/pull/777~   No longer dependent on it. Cabana parses multiplexed signals on its own.

- [x]  parse & edit multiplexed signals
- [x] Group multiplexed signals in tree view 
- [x] don't display overlapping bits warning for multiplexed signals
- [x] add test case for multiplexed signals.
- [x] Display the frequency of each multiplexed signal separately.
   

other changes:

1. cache signal color in member variable
2. remove sources from dbcManager& cleanup code
3. many small code improvements


deanlee:
@adeebshihadeh  this PR is ready.

jnewb1:
Does this support extended multiplexing?

> But CAN DBC files also support extended multiplexing, where a multiplexed signal (in this case S1) can also be a multiplexor and thus control the interpretation of other parts of the data payload.

deanlee:
> Does this support extended multiplexing?
> 
> > But CAN DBC files also support extended multiplexing, where a multiplexed signal (in this case S1) can also be a multiplexor and thus control the interpretation of other parts of the data payload.

Currently not supported. but it can be considered to add in the future.

adeebshihadeh:
@deanlee Any reason this is done in cabana? I think this should go in opendbc's CANParser. @pd0wm might have some thoughts too

deanlee:
@pd0wm  what do you think? should the entire DBC be parsed in Cabana, or just a part of it and put the multiplexed related stuffs in OpenDBC?

pd0wm:
I think it's not trivial to come with a good API for handling multiplexed signals in the CANPacker/Parser, so maybe for now leave it in cabana? There is already a bunch of cabana specific stuff in `DBCFile::parseExtraInfo`.

PR #28311, opened 2023-05-26
offroad alerts: change email reference to support page

sshane:


sshane:
It's also referenced in https://github.com/commaai/openpilot/blob/support-email/selfdrive/assets/offroad/tc.html

PR #28313, opened 2023-05-26
pre-commit: validate poetry lock

fredyshox:


.github/workflows/selfdrive_tests.yaml
```diff
@@ -157,6 +157,17 @@ jobs:
...
+  validate_lock:
+    name: validate poetry lock
+    runs-on: ubuntu-20.04
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: true
+    - uses: ./.github/workflows/setup
+    - name: Check poetry lock file
+      run: ${{ env.RUN }} "poetry lock --check"
```
adeebshihadeh, line 169: let’s just add this to the static analysis job

fredyshox: added that to pre-commit hooks instead

.pre-commit-config.yaml
```diff
@@ -83,3 +83,8 @@ repos:
     rev: '1.5.0'
     hooks:
     -   id: poetry-check
+    -   id: poetry-lock
+        name: validate poetry lock
+        files: '^(pyproject.toml)|(poetry.lock)$'
```
adeebshihadeh, line 88: just need to add the poetry.lock file to the release files

fredyshox: ok, added lock and toml to common release files - both are needed for the check

adeebshihadeh: is this files line still needed?

fredyshox: they are not i can delete them

fredyshox: btw the build_release job handles files required by those checks differently, it copies them over, instead of having them in release files:
```
    - name: pre-commit
      run: |
        cd $GITHUB_WORKSPACE
        cp .pre-commit-config.yaml $STRIPPED_DIR
        cp .pylintrc $STRIPPED_DIR
        cp mypy.ini $STRIPPED_DIR
        cd $STRIPPED_DIR
        ${{ env.RUN }} "pre-commit run --all"
```
should i align?

adeebshihadeh: yeah let's do that, good catch

PR #28315, opened 2023-05-26
car docs: fix docs diff on github

maxime-desroches:


sshane:
think it might be worth it to just write a simple unit test that uses dump and print on the same commit to make sure there's no diff, that would explicitly test this

.github/workflows/selfdrive_tests.yaml
```diff
@@ -370,7 +370,7 @@ jobs:
         id: save_diff
         run: |
           ${{ env.RUN }} "scons -j$(nproc)"
-          output=$(${{ env.RUN }} "python selfdrive/debug/print_docs_diff.py --path /tmp/openpilot_cache/base_car_info") || true
+          output=$(${{ env.RUN }} "python selfdrive/debug/print_docs_diff.py --github --path /tmp/openpilot_cache/base_car_info")
```
sshane, line 373: this can be separate from this PR

sshane, line 373:
```suggestion
          output=$(${{ env.RUN }} "python selfdrive/debug/print_docs_diff.py --github --path /tmp/openpilot_cache/base_car_info") || true
```

selfdrive/debug/print_docs_diff.py
```diff
@@ -107,11 +109,15 @@ def print_car_info_diff(path):
           markdown_builder.append(COLUMN_HEADER)
         markdown_builder.extend(changes[category])
 
-    print("\n".join(markdown_builder))
+    diff = "\n".join(markdown_builder)
+    if github:
+      diff = diff[:MAX_GITHUB_COMMENT_SIZE] + '\n...\nTo see the full diff, run ``` python selfdrive/debug/print_docs_diff.py ```'
```
sshane, line 114: does `MAX_GITHUB_COMMENT_SIZE` take into account the extra message added?

sshane (approved):

sshane:
@maxime-desroches as long as it fails when failing to post the comment, we can close this for now

selfdrive/debug/print_docs_diff.py
maxime-desroches: yes it does

sshane:
Probably going to add more complexity to format it properly, want to reduce complexity so going to close for now

PR #28317, opened 2023-05-27
setup: show downloading when ready

sshane:
doesn't show "downloading" when going back


sshane:
https://github.com/commaai/openpilot/pull/27236/commits/3997d4f43cc557a88c008fb8074cd335d8ed5188

PR #28318, opened 2023-05-27
Hyundai: blacklist bad camera FW

sshane:
these cars have different tuning we need to handle. related to https://github.com/commaai/opendbc/issues/1115

accidentally added back because there was nothing stopping that from happening

will do part number test since that will most likely catch other cases


sshane:
should look into the problem more, it's possible that the santa cruz is fine and tucson is not (it has a much higher lat acc)

selfdrive/car/hyundai/tests/test_hyundai.py
```diff
@@ -52,6 +52,20 @@ def test_auxiliary_request_ecu_whitelist(self):
       self.assertEqual(len(ecus_not_in_whitelist), 0,
                        f"{car_model}: Car model has ECUs not in auxiliary request whitelists: {ecu_strings}")
 
+  def test_blacklisted_parts(self):
+    # Asserts no ECUs known to be shared across platforms exist in the database
```
sshane, line 56:
```suggestion
    # Asserts no ECUs known to be shared across platforms exist in the database.
```

PR #28325, opened 2023-05-28
IsoTpParallelQuery: log all timeouts

sshane:
Currently, we will only log timeouts on queries with more than 1 request (not bytes, or frames, actual tx and rx cycle). So hyundai with one request will never have any timeouts logged

### DIFF
--- a/selfdrive/car/isotp_parallel_query.py
+++ b/selfdrive/car/isotp_parallel_query.py
@@ -120,7 +120,6 @@ class IsoTpParallelQuery:
 
         counter = request_counter[tx_addr]
         expected_response = self.response[counter]
-        # response_valid = dat[:len(expected_response)] == expected_response
         response_valid = dat.startswith(expected_response)
 
         if response_valid:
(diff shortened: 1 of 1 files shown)


selfdrive/car/isotp_parallel_query.py
```diff
@@ -140,8 +142,13 @@ def get_data(self, timeout, total_timeout=60.):
...
         if cur_time - response_timeouts[tx_addr] > 0:
-          if request_counter[tx_addr] > 0 and not request_done[tx_addr]:
-            cloudlog.error(f"iso-tp query timeout after receiving response: {tx_addr}")
+          if not request_done[tx_addr]:
+            if request_counter[tx_addr] > 0:
+              cloudlog.error(f"iso-tp query timeout after receiving partial response: {tx_addr}")
+            elif tx_addr in addrs_ever_responded:
+              cloudlog.error(f"iso-tp query timeout while receiving response: {tx_addr}")
+            else:
+              cloudlog.error(f"iso-tp query timeout with no response: {tx_addr}")
```
sshane, line 151: Previous: only logged if we successfully rx'd response for one request and still had more to send (brands which tx a tester present first for example).

Now: log each case separately: 1) above, 2) still on first request or only one request and we have received frames but not final frame, and no iso-tp errors, and 3) no valid response frames at all (ECU does not exist or a cable/hardware issue).

Issue #28308, opened 2023-05-26
car docs: better hardware footnotes


sshane:
![image](https://github.com/commaai/openpilot/assets/25857203/66ffccfc-6193-4ee2-8878-0f7ec7dbe7c0)

@maxime-desroches can you think of a better place for these, or fix the formatting?

adeebshihadeh:
Unclear what this issue is for. Seems fine to me.

sshane:
it's for the inconsistent superscript link, but I agree it's very minor. not worth the additional 40 lines to fix at the moment

PR #28329, opened 2023-05-28
car docs: fix hardware footnotes formating

maxime-desroches:
https://github.com/commaai/openpilot/issues/28308

We could also put the footnotes on the Buy Here link, but that would be less visible


selfdrive/car/docs_definitions.py
```diff
@@ -325,9 +332,8 @@ def get_column(self, column: Column, star_icon: str, video_icon: str, footnote_t
     elif column == Column.VIDEO and len(item) > 0:
       item = video_icon.format(item)
 
-    footnotes = get_footnotes(self.footnotes, column)
-    if len(footnotes):
-      sups = sorted([self.all_footnotes[fn] for fn in footnotes])
-      item += footnote_tag.format(f'{",".join(map(str, sups))}')
+    # hardware footnotes need special formatting
+    if column != Column.HARDWARE:
+      item += self.get_ordered_footnotes(footnote_tag, column)
```
sshane, line 337: Hmm, I'd like if we could keep this as generic as possible and still add the footnotes after. Perhaps this get_column can instead do the markdown spoiler wrapping generically? Would that be cleaner?

sshane:
I don't think the extra class is worth it, I just quickly glanced. Any simpler way we can fix?

PR #28331, opened 2023-05-28
GM camera ACC: add camera diagnostic address to fingerprints

morrislee:
**Description** 

Sometimes, silverado/sierra will not fingerprint on startup, found that two of the addresses in these instances

**Verification**

This happens in a blue moon, will need to test it for a while


sshane:
these two are the tx and rx addrs for iso-tp communication for the fwdCamera (0x24b and 0x64b), maybe we should handle this by adding to all cars (or camera ACC cars)

morrislee:
Looks like these two are indeed repeating every time when the vehicle does not fingerprint, this should fix the once in a blue moon miss fingerprints

PR #28336, opened 2023-05-30
Nissan: add counter signal to forwarded message

sshane:
I was not aware that dbc.cc only adds the counter type to signals where we explicitly say to track them for that dbc


sshane:
@adeebshihadeh I also wasn't aware that the CAN parser doesn't fill out the vl dictionary if it doesn't get any messages, so there can be key errors if a message is totally missing

sshane:
I also wasn't aware that the CAN parser doesn't fill out the vl dictionary if it doesn't get any messages, so there can be key errors if a message is totally missing, I'll fix by getting with a default, as the behavior before was.

sshane:
Actually it does fill out all values as zeros, strange how I wasn't seeing this on my WSL2 laptop last night...

Issue #27961, opened 2023-04-18
openpilot did not engage after braking to zero caused a acc fault


rjsmith1999:
### Describe the bug

Issue seemed to start after I let openpilot brake to zero. I temporarily saw a warning on my comma 3, not sure what it was as I took over pretty quickly. 

Everything seemed fine after that, but the “acc” indicator in my dash turned orange. 

I drove for a while and then re-engaged openpilot. All seemed fine for a while. Some point later openpilot gave an emergency warning: cruse not enabled I think. 

I took over, and continued driving. acc indicator was still orange, but I could no longer engage openpilot (nothing happened when I pressed set) (pcm stopped sending enable messages?)

At a stoplight, I restarted the car a few times. acc remained orange. A few other messages showed up, but went away quickly, I think after my comma 3 went onroad. 
Those messages were “RDM unavailable” or something, “acc unavailable” and a ”brake system” orange indicator.. 

After driving for a while, the acc indicator turned green and I could use openpilot like normal. 

### Which car does this affect?

Honda Passport Elite 2021

### Provide a route where the issue occurs

Will add routes when they finish uploading…

### openpilot version

6f35d23c062ab000af90f5a0d7704eab8682479d

### Additional info

This drive was on my own repo that’s one (personal) commit ahead of the hash I pasted above. Sorry if that complicates things. Can try to reproduce on master. 

****
My take on the situation, which could be way off. It seems like braking to zero caused a fault related to the abs brake request. Maybe we can we can fix this, maybe it’s unavoidable…

At the very least, I think openpilot’s warning behavior could be improved in this situation. 
- It seems like openpilot was hiding the brake system, acc and rdm warning messages on the gauge cluster
- There was an acc fault, but openpilot didn’t show a no entry alert 

Also, maybe if we’re in a fault situation like this where the pcm isn’t sending enable messages, we could read the button messages and send an openpilot unavailable alert (I think that’s possible?)

also.. no idea why I could engage once after the acc fault, but not subsequent times. 🤷

sshane:
Stock AEB activated so we blocked our own braking messages and started forwarding the camera's. I searched for a fault bit when you weren't able to enable after this, but I found nothing that matched very well. There's a bug however, openpilot didn't consider it AEB and disengage as the AEB bit was set, but it was not requesting any braking. panda checks the bit and if the camera command is >= the openpilot command, which 0 >= 0 is true.

[...]

PR #28338, opened 2023-05-30
cars: log non-critical car faults

sshane:
adds field to carstate and an event to log when some portion of the car's ADAS systems are faulted, but openpilot can still engage. for example, the camera on Nidec (and many others) wants to display a faulted signal but we block and send our own. this should only be used to log superficial faults, nothing that prevents any kind of control


sshane:
this would have exposed the cause behind https://github.com/commaai/openpilot/issues/27961 earlier!

PR #28339, opened 2023-05-30
Honda: log CS accFaulted field

sshane:
makes it easier to see in the qlogs when there's a Honda ACC fault vs. looking at the carEvents. removes last usage to brakeUnavailable


sshane:
@adeebshihadeh should we deprecate brakeUnavailable? I think we should have used it when we made the switch to accFaulted, but now it's not used anymore

adeebshihadeh:
Yes, deprecate if unused.

PR #28349, opened 2023-05-31
replay: lock uiDebug and userFlag messages behind --all flag

fredyshox:
Creating PubSockets for `uiDebug` and `userFlag` services conflicts with pub sockets used on selfdrive/ui. It crashes when using with ZMQ, and causes undefined behaviour with msgq. And using them together is major use case.

This PR solves this by locking this services by default, with ability to unlock them via `--ui-debug` flag. IMO it's cleaner than putting them into blacklist by default, and requiring user to clear the blacklist by themselves if needed.


tools/replay/main.cc
```diff
@@ -16,6 +16,7 @@ int main(int argc, char *argv[]) {
       {"qcam", REPLAY_FLAG_QCAMERA, "load qcamera"},
       {"no-hw-decoder", REPLAY_FLAG_NO_HW_DECODER, "disable HW video decoding"},
       {"no-vipc", REPLAY_FLAG_NO_VIPC, "do not output video"},
+      {"ui-debug", REPLAY_FLAG_UI_DEBUG, "do output uiDebug and userFlag messages. may causes issues when used along with UI"}
```
adeebshihadeh, line 19: I think this use case can just use the `--allow` flag and they're blocked by default

fredyshox: you're right, that's basically the same thing

fredyshox: but if you want to replay everything, then you need to specify every service in `--allow`

adeebshihadeh: that's true, but this should still be a bit more generic. maybe make this flag something like `--all`?

fredyshox: you're right, that seems nice

tools/replay/replay.cc
```diff
@@ -16,8 +16,14 @@ Replay::Replay(QString route, QStringList allow, QStringList block, SubMaster *s
   auto event_struct = capnp::Schema::from<cereal::Event>().asStruct();
   sockets_.resize(event_struct.getUnionFields().size());
   for (const auto &it : services) {
+    uint16_t which = event_struct.getFieldByName(it.name).getProto().getDiscriminantValue();
+    if ((which == cereal::Event::Which::UI_DEBUG || which == cereal::Event::Which::USER_FLAG) &&
+        !(flags & REPLAY_FLAG_ALL_SERVICES) &&
```
adeebshihadeh, line 21: maybe make a set somewhere with the blacklisted ones and then it can be included in the help dialog

fredyshox: Added `base_blacklist` which is passed to Replay. Then the help message for `--all` looks like this:
```
  --all                  do output all messages including uiDebug, userFlag.
                         this may causes issues when used along with UI
```

Issue #21483, opened 2021-07-05
Tool to generate screenshots of UI in all states


adeebshihadeh:
It would be nice to have a tool that generates screenshots of UI in all the possible states for testing design changes. We could also use this as a regression test to ensure refactors don't change design.

Examples of possible states:
* home screen
  * not paired
  * paired with prime
  * paired with no prime
* onroad UI
  * 0/1/2 leads
  * DM and engage-ability icons on and off

maykonpacheco:
Hi, could you give us more details on how this would work and how it would help the user experience?

Edgars01:
Hello. Do you still need this? I can try to make it as my first contribution.

PR #28350, opened 2023-05-31
UI: snapshot testing

incognitojam:
Current states:
- no prime
- prime
- experimental mode

It seems as though you must set `QT_QPA_PLATFORM="offscreen"` when running the `ui_snapshot` tool otherwise it won't match in the tests.

It uses imagemagick to compare the images. We could make it slightly more lenient if there are differences between machines/Qt versions. Imagemagick can also output the diff to a file so we could display it or use it in a github actions script in the future.

Related:
- https://github.com/commaai/openpilot/pull/28356
- https://github.com/commaai/openpilot/issues/21483


incognitojam:
Unfortunately this generates screenshots like this right now:
image

I believe because the assets path is incorrect - we should use the ":/" prefix everywhere

adeebshihadeh (commented):
Very cool! Let's get the tool itself merged first. I think it's best to write the test as a python unit test; it'll be easier that way once it's testing the whole UI.

PR #28351, opened 2023-05-31
Deprecate brakeUnavailable event

sshane:
https://github.com/commaai/cereal/pull/455


selfdrive/controls/lib/events.py
```diff
@@ -724,7 +724,7 @@ def joystick_alert(CP: car.CarParams, CS: car.CarState, sm: messaging.SubMaster,
     ET.SOFT_DISABLE: soft_disable_alert("Calibration Incomplete"),
     ET.NO_ENTRY: NoEntryAlert("Calibration in Progress"),
   },
-  
+
```
sshane, line 727:
```suggestion
  
```

PR #28363, opened 2023-06-01
schedule ui drawing

adeebshihadeh:


selfdrive/ui/qt/onroad.cc
```diff
@@ -75,6 +77,13 @@ void OnroadWindow::updateState(const UIState &s) {
     bg = bgColor;
     update();
   }
+
+  // force paint now
+  double tsm = (float)(nanos_since_boot() - sm["modelV2"].getLogMonoTime()) / 1e6;
```
deanlee, line 82: This line may have problems in replay mode.

selfdrive/ui/qt/onroad.cc
```diff
@@ -51,9 +51,11 @@ OnroadWindow::OnroadWindow(QWidget *parent) : QWidget(parent) {
 }
 
 void OnroadWindow::updateState(const UIState &s) {
+  const SubMaster &sm = *(s.sm);
```
deanlee, line 54: Would it be better to return immediately if onroadWindow is not visible

selfdrive/ui/qt/onroad.cc
adeebshihadeh: we want the best timestamp for when the model actually ran, so needs to be logMonoTime. I added another check to handle replay

PR #28370, opened 2023-06-02
ui/AnnotatedCameraWidget: paint on uiPlan event

deanlee:
poll 'uiPlan' in worker thread to avoid blocking main thread.


adeebshihadeh:
Thanks, but was able to write it a bit simpler in #28363

PR #28371, opened 2023-06-02
Added new Ram firmwares

vincentw56:
**Description** These are new firmwares that we have collected and verified working.

**Verification** Ram users who have used the Realfast fork provided these firmware additions. After adding them to that fork, the Ram users were able to use OP.


sshane:
Can you provide the dongle ids of the users? I can do a FW dump eventually but that will take some time as I don't have time right now.

selfdrive/car/chrysler/values.py
```diff
@@ -353,6 +356,12 @@ class ChryslerCarInfo(CarInfo):
       b'M2370131MB',
       b'M2421132MB',
     ],
+    (Ecu.gateway, 0x18DACBF1, None): [
+      b'68450806AB',
+      b'68488419AB',
+      b'68535476AB',
+      b'68535476AC',
+    ],
```
sshane, line 364: @realfast do you know why we didn't add gateway before? Do not all trucks have it?

vincentw56: We have them, but they are commented out on the 1500s. I am not sure why. It hasn't impacted us.

vincentw56:
> Can you provide the dongle ids of the users? I can do a FW dump eventually but that will take some time as I don't have time right now.

Hmm, I will have to see since I didn't collect those. Didn't know those were needed.

selfdrive/car/chrysler/values.py
realfast: > @realfast do you know why we didn't add gateway before? Do not all trucks have it?
@sshane because it wasn't needed at all to identify the truck. I will work with @vincentw56 to figure out the issue.

sshane:
Yeah, we generally like to know where FW versions come from before adding to openpilot, even veteran users sometimes accidentally add incorrect FW versions on occasion, and it's good to have info to look back on. Like I said, I can look it up from our data, but that will take some time.

sshane:
Any updates?

vincentw56:
I won't be able to provide Dongles since I didn't record those. But I did have 3 PRs after this that had those.

incognitojam:
Please contact those users so that we can check their logs and verify the FW individually

PR #28372, opened 2023-06-02
IsoTpParallelQuery: remove spammy log

sshane:
I originally only wanted to add the timeout while receiving response, so this wasn't needed. Making it a warning


sshane:
just going to comment out since it doesn't handle functional address requests very well (times out on all possible response addresses)

PR #28373, opened 2023-06-02
CI: never cancel concurrent on master

adeebshihadeh:


.github/workflows/selfdrive_tests.yaml
```diff
@@ -8,7 +8,7 @@ on:
 
 concurrency:
   group: ${{ github.event_name == 'push' && github.ref == 'refs/heads/master' && github.run_id || github.head_ref || github.ref }}-${{ github.workflow }}-${{ github.event_name }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.ref != 'refs/heads/master' }}
```
sshane, line 11: like `group`, `cancel-in-progress` is also evaluated dynamically for each run on events, so the output will change to the new branch ref when pushed to another branch. does not work :( https://github.com/commaai/openpilot/commit/0129da088ba19f3022c857e9bb7dac37b0f9f352

PR #28378, opened 2023-06-03
Test power usage for mapsd and navmodeld

mitchellgoffpc:


system/hardware/tici/tests/test_power_draw.py
```diff
@@ -46,12 +49,21 @@ def tearDown(self):
...
+    msg.liveLocationKalman.status = 'valid'
 
     prev = baseline
     used = {}
     for proc in PROCS:
       managed_processes[proc.name].start()
-      time.sleep(proc.warmup)
+
+      # Send liveLocationKalman at 20hz
+      for _ in range(int(20*proc.warmup)):
```
adeebshihadeh, line 64: does this work? map renderer should only use the latest stuff

adeebshihadeh (approved):
LGTM if it's passed a bunch of times in a loop

PR #28389, opened 2023-06-03
Cabana: Improve contrast when using dark themes

projectgus:
Change some hard-coded colors in the Cabana UI that didn't always contrast well when using different system color themes:

* Signal view text/value editor background
* Signal log view heading text
* Binaryview signal box color (when not selected or highlighted)
* Chart title & message text

Before (automatic system theme with dark colors)
before_automatic_dark_theme

Before (manually configured dark theme)
before_manual_darktheme

After (automatic system theme with dark colors)
after_automatic_dark

After (manually configured dark theme)
after_manual_darktheme

I've only tested this on Linux, but I think it should produce either the same results on systems where the Base colour is White or improved contrast on other systems.

### DIFF
--- a/tools/cabana/chart/chart.cc
+++ b/tools/cabana/chart/chart.cc
@@ -213,9 +213,19 @@ void ChartView::updateTitle() {
   for (QLegendMarker *marker : chart()->legend()->markers()) {
     QObject::connect(marker, &QLegendMarker::clicked, this, &ChartView::handleMarkerClicked, Qt::UniqueConnection);
   }
+
+  // Use CSS to draw titles in the WindowText color
+  auto tmp = palette().color(QPalette::WindowText);
+  auto titleColorCss = tmp.name(QColor::HexArgb);
+  // Draw message details in similar color, but slightly fade it to the background
+  tmp.setAlpha(180);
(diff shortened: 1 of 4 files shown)


incognitojam:
Did you test this with light theme too?

tools/cabana/chart/chart.cc
```diff
@@ -215,7 +215,13 @@ void ChartView::updateTitle() {
   }
   for (auto &s : sigs) {
     auto decoration = s.series->isVisible() ? "none" : "line-through";
-    s.series->setName(QString("<span style=\"text-decoration:%1\"><b>%2</b> <font color=\"gray\">%3 %4</font></span>").arg(decoration, s.sig->name, msgName(s.msg_id), s.msg_id.toString()));
+    QColor windowText = palette().color(QPalette::WindowText);
+    auto titleColor = windowText.name(QColor::HexArgb);
+    windowText.setAlpha(180);
+    auto msgColor = windowText.name(QColor::HexArgb);
```
incognitojam, line 221: does it make sense to move these declarations out of the loop?

projectgus: @incognitojam Good idea, have done!

PR #28390, opened 2023-06-03
cabana: Add menu item that resets the window layout

projectgus:
This is useful if you've broken out a docked column  and then closed the broken out window, for example.

The workaround is to find and delete the settings file, but this is a little easier. :)

### DIFF
--- a/tools/cabana/mainwin.cc
+++ b/tools/cabana/mainwin.cc
@@ -32,6 +32,9 @@ MainWindow::MainWindow() : QMainWindow() {
   createStatusBar();
   createShortcuts();
 
+  // save default window state to allow resetting it
+  default_state = saveState();
+
   // restore states
   restoreGeometry(settings.geometry);
   if (isMaximized()) {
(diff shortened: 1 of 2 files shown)


tools/cabana/mainwin.cc
```diff
@@ -135,6 +138,10 @@ void MainWindow::createActions() {
   commands_act->setDefaultWidget(new QUndoView(UndoStack::instance()));
   commands_menu->addAction(commands_act);
 
+  edit_menu->addSeparator();
+  edit_menu->addAction(tr("Reset Window Layout"),
+                       [=]() { this->restoreState(this->default_state); });
```
deanlee, line 143: how about change it to `[this]() { restoreState(default_state); });` ?
we don't use `this` pointer explicitly If there is no particular reason.

projectgus: @deanlee Done, thanks! I think I had a weird feeling "oh, I'm in a lambda scope so I should be really explicit about what I'm binding". But it's not necessary at all!

PR #28391, opened 2023-06-03
Subaru: DBC Update for gen2 long

jnewb1:
**Description** 
When ES is disabled, ES_DashStatus wont be available for reference anymore. I found matching signal for set speed elsewhere.

**Verification**
Tested on my Legacy 2022

**Route**

**Depends**
https://github.com/commaai/opendbc/pull/847


jnewb1:
ill just combine this into a single PR

PR #28396, opened 2023-06-04
Toyota: enable BSM on C-HR

cydia2020:
Enable BSM on TSS-P Toyota C-Hr, users on discord are saying that this is working for them. The signals are flipped on this car, other than that it's the same as any other TSS2.0 toyotas.

Route: to be provided


selfdrive/car/toyota/carstate.py
```diff
@@ -144,6 +144,11 @@ def update(self, cp, cp_cam):
       ret.leftBlindspot = (cp.vl["BSM"]["L_ADJACENT"] == 1) or (cp.vl["BSM"]["L_APPROACHING"] == 1)
       ret.rightBlindspot = (cp.vl["BSM"]["R_ADJACENT"] == 1) or (cp.vl["BSM"]["R_APPROACHING"] == 1)
```
incognitojam, line 145: If you are reading the same signals I think you simplify it like this
```suggestion
      ret.rightBlindspot = (cp.vl["BSM"]["R_ADJACENT"] == 1) or (cp.vl["BSM"]["R_APPROACHING"] == 1)
      # TSS-P C-HR has flipped BSM signal
      if self.CP.carFingerprint in (CAR.CHR, CAR.CHRH):
        ret.leftBlindspot, ret.rightBlindspot = ret.rightBlindspot, ret.leftBlindspot
```

selfdrive/car/toyota/carstate.py
```diff
@@ -144,6 +144,11 @@ def update(self, cp, cp_cam):
...
```
incognitojam, line 146: If you are reading the same signals I think you simplify it like this
```suggestion
      # TSS-P C-HR has flipped BSM signal
      if self.CP.carFingerprint in (CAR.CHR, CAR.CHRH):
        ret.leftBlindspot, ret.rightBlindspot = ret.rightBlindspot, ret.leftBlindspot

```

selfdrive/car/toyota/carstate.py
```diff
@@ -226,7 +229,7 @@ def get_can_parser(CP):
       signals.append(("INTERCEPTOR_GAS2", "GAS_SENSOR"))
       checks.append(("GAS_SENSOR", 50))
 
-    if CP.enableBsm:
+    if CP.enableBsm or (CP.enableBsm and CP.carFingerprint in (CAR.CHR, CAR.CHRH)):
```
incognitojam, line 232: I don't think this does anything

cydia2020: Okay yea just realised it as well, that's an error on my part, will fix

sshane:
Do you have a route? Maybe there's a generic bit in the message that signifies that they are flipped.

cydia2020:
> Do you have a route? Maybe there's a generic bit in the message that signifies that they are flipped.

I'm just waiting for a route from another user as I don't own this particular vehicle.

cydia2020:
> Looks like there might be something useful in byte 2 of the BSM message
> 
> image

There could be, however bsm is known to not work with the TSS-P Prius, is it indicated by a specific bit?

sshane:
Why does it work on the TSSP Prius? Does it not have BSM or are they just always 0?

cydia2020:
> Why does it work on the TSSP Prius? Does it not have BSM or are they just always 0? Do you know if any of the other TSSPs in this list also don't have BSM so I can figure out the bits?

On the TSS-P Prius, all the bits that op reads are 0 even though the vehicle does have bsm, only available bits are on/off, buzzer, and fault iirc.

sshane:
Are you confident this doesn't vary by region or should I check the data?

cydia2020:
> By the way, are you confident this doesn't vary by region or should I check the data?

I'm not, please do check, we've only tested on 2 vehicles.

adeebshihadeh:
Are we still waiting on routes for this? If so, @sshane we should close until we have them.

sshane:
Just checked, it seems region dependent. A CHR from the US looks normal, but a CHR Hybrid from Thailand looks flipped

adeebshihadeh:
Only for the CHR? Other Toyota models are consistent across regions? If so, this is a good candidate for merging as a comment.

sshane:
Only checked two dongles for the CHR since I don't know of an automatic way to check this.

adeebshihadeh:
Let's merge as a comment. @cydia2020 at the moment, we don't have the resources to validate this to the degree necessary, but if you can get the initial data to understand this with some certainty, we can put in a bit of effort to ensure it matches across the rest of the fleet.

sshane:
Some notes:

- 3d270376633fcfdb|2022-10-29--19-38-06 - car in Germany, drives on right, IS FLIPPED, dat[1]=0xDA, dat[2]=0x00
- 482ce47555b39e90|2023-08-23--19-07-26 - car in US, drives on right, IS NOT FLIPPED, dat[1]=0xDA, dat[2]=0x00
- 29c9bda9ab685a97|2023-08-23--18-35-57 - car in Thailand, drives on left, IS FLIPPED, dat[1]=0xEA, dat[2]=0x68
- Our TSS2 Corolla - IS NOT FLIPPED, dat[1]=0x88, dat[2]=0x6E

So it appears that this bit isn't in the BSM message since it's the same for the German and US car

cydia2020:
> Some notes:
> 
> * 3d270376633fcfdb|2022-10-29--19-38-06 - car in Germany, drives on right, IS FLIPPED, dat[1]=0xDA, dat[2]=0x00
> * 482ce47555b39e90|2023-08-23--19-07-26 - car in US, drives on right, IS NOT FLIPPED, dat[1]=0xDA, dat[2]=0x00
> * 29c9bda9ab685a97|2023-08-23--18-35-57 - car in Thailand, drives on left, IS FLIPPED, dat[1]=0xEA, dat[2]=0x68
> * Our TSS2 Corolla - IS NOT FLIPPED, dat[1]=0x88, dat[2]=0x6E
> 
> So it appears that this bit isn't in the BSM message since it's the same for the German and US car

Can openpilot query fw for the BSM radars? Maybe it's fw related

sshane:
On our Corolla TSS2, I was able to on a sub-address of 0x750, I just didn't label it. You can probably look up the first four characters to find the part.

```python
comma@tici:/data/openpilot/panda/examples$ ./query_fw_versions.py --bus 0 --addr 0x750 --nonstandard --subaddr scan
querying addresses ...
0x750: 100%|█████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:25<00:00, 25.36s/it]
*** Results for address 0x750, sub-address 0x41 ***
0x8801 KWP: b'8816F1201101    '
*** Results for address 0x750, sub-address 0x42 ***
0x8801 KWP: b'8816F1201102    '
*** Results for address 0x750, sub-address 0x5F ***
0x8801 KWP: b'8911F1202003    '
```

cydia2020:
For the Japanese market, PNC# 88162 and 88163 are for the RH BSM modules, 88172 and 88173 are for LH modules, PNC# is universal across the board regardless of market and model. In Techstream different modules are marked as Master/Slave, maybe it's related to which one is master and which one is slave? Still need a way to determine that.

PR #28399, opened 2023-06-04
cabana dbc: Sort signals by name if start bits are equal

projectgus:
This is particularly significant for VECTOR__INDEPENDENT_SIG_MSG where a lot of the signals share the same start bit.

Without this change, the order is undefined for most of these signals and results in a lot of unnecessary churn in the DBC file contents between saves.

### DIFF
--- a/tools/cabana/dbc/dbc.cc
+++ b/tools/cabana/dbc/dbc.cc
@@ -9,7 +9,13 @@ std::vector<const cabana::Signal*> cabana::Msg::getSignals() const {
   std::vector<const Signal*> ret;
   ret.reserve(sigs.size());
   for (auto &sig : sigs) ret.push_back(&sig);
-  std::sort(ret.begin(), ret.end(), [](auto l, auto r) { return l->start_bit < r->start_bit; });
+  std::sort(ret.begin(), ret.end(), [](auto l, auto r) {
+    if (l->start_bit != r->start_bit) {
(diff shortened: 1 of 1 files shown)


tools/cabana/dbc/dbc.cc
```diff
@@ -9,7 +9,13 @@ std::vector<const cabana::Signal*> cabana::Msg::getSignals() const {
   std::vector<const Signal*> ret;
   ret.reserve(sigs.size());
   for (auto &sig : sigs) ret.push_back(&sig);
-  std::sort(ret.begin(), ret.end(), [](auto l, auto r) { return l->start_bit < r->start_bit; });
+  std::sort(ret.begin(), ret.end(), [](auto l, auto r) {
+    if (l->start_bit != r->start_bit) {
+      return l->start_bit < r->start_bit;
+    }
+    // For VECTOR__INDEPENDENT_SIG_MSG, many signals have same start bit
+    return QString::compare(l->name, r->name, Qt::CaseSensitive) < 0;
```
deanlee, line 17: this is same as `return l->name < r->name;` 
https://codebrowser.dev/qt5/qtbase/src/corelib/text/qstring.cpp.html#_ZNK7QStringltE13QLatin1String

projectgus: @deanlee Oops, thanks! I actually looked for this overload but didn't see it somehow. Have updated.

PR #28400, opened 2023-06-04
Silverado/Sierra Tune

morrislee:
Silverado/Sierra Tune

**Description** 
With the tune on master, vehicle ping pongs in straights a lot and is poor in centering in area such as center lane. I used sunnypilot to force torqued and let it find the best fit for majority of the speed. This is going to be good enough before non-linear tune like the bolt is implemented.

**Verification** 
Lateral control seems to have drastic improvement when I increased the steering delay to 0.5 as found from desired vs actual lateral acceleration. This is probably not the right approach but it works. I am not experiencing turn cutting, ping pong on straights now.

The default stopping parameters makes a harsh stop behind lead, I have found these parameters stops behind lead as close to stock smoothness as I feel comfortable. It has been very reliable to test stopping behind lead since hc model and the relative lead speed in op voacc.

Curb weight of the Silverado and Sierra in this class that actually supports Openpilot with the right equipment has shown to have vehicle weight specified as built per unit in the door jam has been hovering high ~2495 to just over 2500. This is should be corrected to match these vehicles rather than the base spec'ed vehicles typically that is not equipped so.

**Route**
Route: fe80a4e1cedec853|2023-06-03--20-52-48--0
Lateral:
image
Longitudinal:
image


selfdrive/car/gm/interface.py
```diff
@@ -219,11 +219,16 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
       CarInterfaceBase.configure_torque_tune(candidate, ret.lateralTuning)
 
     elif candidate == CAR.SILVERADO:
-      ret.mass = 2200. + STD_CARGO_KG
+      ret.mass = 2500. + STD_CARGO_KG
       ret.wheelbase = 3.75
       ret.steerRatio = 16.3
       ret.centerToFront = ret.wheelbase * 0.5
       tire_stiffness_factor = 1.0
+      ret.steerActuatorDelay = 0.5
```
sshane, line 227: @nuwandavek do you know if the two lateral changes are good changes? it really has this much delay and low lat accel?

selfdrive/car/gm/interface.py
```diff
@@ -219,11 +219,16 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
       CarInterfaceBase.configure_torque_tune(candidate, ret.lateralTuning)
 
     elif candidate == CAR.SILVERADO:
-      ret.mass = 2200. + STD_CARGO_KG
+      ret.mass = 2500. + STD_CARGO_KG
```
sshane, line 222: can you find the average curb weight of all the trims and update this in a new PR?

selfdrive/car/gm/interface.py
```diff
@@ -219,11 +219,16 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
...
       ret.wheelbase = 3.75
       ret.steerRatio = 16.3
       ret.centerToFront = ret.wheelbase * 0.5
       tire_stiffness_factor = 1.0
+      ret.steerActuatorDelay = 0.5
+      ret.vEgoStopping = 0.1
+      ret.stopAccel = -0.25
+      ret.longitudinalTuning.kpV = [2.4, 1.2]
+      ret.longitudinalTuning.kiV = [0.16]
```
sshane, line 231: I would make sure to match the stock system first, if possible. Watch a few stock routes coming to a stop (or provide some)

selfdrive/car/gm/interface.py
morrislee: average from as built of 4 samples I have know to be 2516

selfdrive/car/gm/interface.py
morrislee: This seems to be not needed when using tim's NNFF, so this may not be the correct approach

nuwandavek: yeah, I'd split out the necessary changes (mass) from the delay. 0.5 seconds is a lot of delay - maybe there are other effects.

selfdrive/car/gm/interface.py
Verylukyguy: image

Verylukyguy: GMC
image

morrislee: these are guide lines, but GM actually provide a sticker showing the actual weight of each vehicle built at door jam. I dont believe every denali has adaptive cruise and lane keep assists (I have seen ones without windshield camera). I believe manually sampling the values is more accurate on what weight we need to go by. Trucks has too many variations in length, accessories, tires/rim combo can all heavily affect the weight. Some trucks comes with carbon fiber too that dramatically reduces the weight. I have sampled the few individuals with OpenPilot running on their vehicle with the following weights:

2495
2509
2521
2539

Avg: 2516

These are mixed with 5.3, 6.2 and the 3.0 Diesel equipped vehicles, the weight is actually a lot closer to each other.

Verylukyguy:
The Silverado/Sierra mass increase was merged on PR https://github.com/commaai/openpilot/pull/29355

morrislee:
Not sure what I am looking for, is it the dip at very end? Because the value from master drops a lot more, it is a very harsh stop @sshane 

OP Long Master Stop
fe80a4e1cedec853|2023-08-19--14-05-40--12
https://connect.comma.ai/fe80a4e1cedec853/1692476263358/1692476321696
Master Long Stopping


Op Long Tuned Stop
fe80a4e1cedec853|2023-08-19--14-31-10--1
https://connect.comma.ai/fe80a4e1cedec853/1692477133191/1692477167696
OP Tuned Stopping


Stock Long Stop
fe80a4e1cedec853|2023-08-19--14-18-52--1
https://connect.comma.ai/fe80a4e1cedec853/1692476398949/1692476438451
Stock Long Stopping

sshane:
can you rebase?

Verylukyguy:
Silverado/Sierra non-linear lateral tune was merged on https://github.com/commaai/openpilot/pull/29540

PR #28403, opened 2023-06-05
Subaru: POC gen2 long

jnewb1:
**Description**
Proof of concept Subaru gen2 long

General Idea:
- ES is UDS disabled, and unlocked with a secret key
- Read buttons through UDS, and pass them to cruise with ES_Distance
- Create all ES_* signals from scratch
- The rest is the same as gen1 long, except on different busses

TODO:

- [x] Verify that the ECU unlock can be done universally on gen2 cars (need 2-3 routes from gen2 cars with [jnewb1/subaru-gen2-uds-test](https://github.com/jnewb1/openpilot/tree/subaru-gen2-uds-test)).
- [x] Figure out what we need to do with the three unknown signals that I'm currently sending a constant value for. I checked 2-3 routes and they appear to only have constant values, but we will need to verify that.
- [x] Fix a few issues with the dashstatus
- [ ] Get gen1 merged first behind a toggle? Or try and do both at once? https://github.com/commaai/openpilot/pull/25345
- [ ] If multiple EYESIGHT_UDS_RESPONSE messages come in, they might overwrite each other. Need DBC multiplexing. https://github.com/commaai/opendbc/pull/854
- [ ] <del>Read AEB through UDS, and passthrough?</del> (for experimental long, this is not required - shane)

**Route**
d9df6f87e8feff94|2023-06-04--19-03-42--0

https://github.com/commaai/panda/pull/1443
https://github.com/commaai/opendbc/pull/848


jnewb1:
> Verify that the ECU unlock can be done universally on gen2 cars (need 2-3 routes from gen2 cars with [jnewb1/subaru-gen2-uds-test](https://github.com/jnewb1/openpilot/tree/subaru-gen2-uds-test)).

Verified on my 22 legacy, and an eyesight module from a 21 legacy

jnewb1:
image
key works on 2020 outback too

jnewb1:
0x22a appears to be Subaru's High Beam Assist. Since we won't be replicating that, we can just send a constant value for that, or try and read it through UDS to keep that functionality.

Other two signals are constant for all routes that I've checked.

jnewb1:
key isn't required at all if you read the button signal directly instead of creating the dynamically defined message. going to close this and make a new PR after the gen1 long is merged

sshane:
Did we ever merge this?

jnewb1:
> Did we ever merge this?

no. it requires reading the buttons over UDS. If we think that's fine then I'll revive this

jnewb1:
any we need this https://github.com/commaai/opendbc/pull/854

sshane:
There's `ReadDataByPeriodicIdentifier` but it probably is more sketchy than repeatedly querying, since any non period query will stop the broadcast (according to KWP2000 docs)

jnewb1:
I think i tried `ReadDataByPeriodicIdentifier` and it gave a negative response

PR #28410, opened 2023-06-05
Hyundai: add missing FW for Santa Fe Hybrid 2022

Uptrend-john:
Added new EPS data for SANTA FE HYBRID 2022

added this to the values.py everything else was correct. When I plugged it in the car was recognized 

 b'\xf1\x00TM  MDPS C 1.00 1.02 56310-CLEC0 4TSHC102'


haraschax:
Can you give this PR a proper name?

PR #28412, opened 2023-06-05
Ford: detect auxiliary panda

incognitojam:
depends on https://github.com/commaai/openpilot/pull/28489

### DIFF
--- a/selfdrive/car/__init__.py
+++ b/selfdrive/car/__init__.py
@@ -1,10 +1,12 @@
 # functions common among cars
-import capnp
+import math
 from collections import namedtuple
+from typing import Dict, Optional
+
+import capnp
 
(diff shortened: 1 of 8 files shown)


sshane (changes requested):
Use the helper adeeb wrote for hkg

incognitojam:
@sshane It's quite Hyundai specific with ACAN and ECAN... will make a simple version

incognitojam:
actually I don't think the helper is useful in this case

incognitojam:
Running `./test_models.py TestCarModel_96_KIA_EV6_2022` doesn't work, but if I add `print("CANBUS", self._a, self._e, self._cam)` to `CanBus` it works...

PR #28415, opened 2023-06-05
regen: use WideCameraOnly

fredyshox:


adeebshihadeh:
Want to get #27640 merged instead?

fredyshox:
Ok, that seams reasonable. Much more seamless than settings the param yourself. Also why is the wide road camera stream assumed to be default when using single camera, instead of the standard road one?

adeebshihadeh:
> Ok, that seams reasonable. Much more seamless than settings the param yourself. Also why is the wide road camera stream assumed to be default when using single camera, instead of the standard road one?

No good reason, it was originally a toggle in the UI during comma three development to completely switch to the wide cam for testing.

adeebshihadeh:
Merged #27640, is this still necessary?

fredyshox:
Gonna see how it works on the regen side, and if we need some changes there.

PR #28425, opened 2023-06-06
ui: refactor ButtonParamControl

deanlee:
1. remove hard-coded widths to adapt to different widths of multiple languages.
2. smaller button font (Reduced from 45 to 40）
3. use checkable button and button groups to simplify the code.
4. use `atoi` instead of `std::stoi` to avoid throwing exceptions.
5. remove the extra semicolon in controls.h.

Screenshot from 2023-06-06 15-32-49

### DIFF
--- a/selfdrive/ui/qt/widgets/controls.h
+++ b/selfdrive/ui/qt/widgets/controls.h
@@ -234,8 +234,10 @@ public:
       button_gorup->addButton(button, i);
     }
 
-    QObject::connect(button_gorup, QOverload<int>::of(&QButtonGroup::buttonClicked), [=](int id) {
-      params.put(key, std::to_string(id));
+    QObject::connect(button_gorup, QOverload<int, bool>::of(&QButtonGroup::buttonToggled), [=](int id, bool checked) {
+      if (checked) {
+        params.put(key, std::to_string(id));
+      }
     });
   }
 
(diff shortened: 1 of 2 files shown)


sshane:
nice! can we add pressed states too for the buttons?

selfdrive/ui/qt/widgets/controls.h
```diff
@@ -188,89 +189,61 @@ class ParamControl : public ToggleControl {
...
-      buttons[i]->setStyleSheet(unselect_style);
-      if (param_value == i) {
-        buttons[i]->setStyleSheet(select_style);
-      }
-    }
-  };
-  void showEvent(QShowEvent *event) override {
-    refresh();
-  };
+    QObject::connect(button_gorup, QOverload<int>::of(&QButtonGroup::buttonClicked), [=](int id) {
```
sshane, line 237: should be `buttonToggled` so we don't keep writing if the user taps the current button?

selfdrive/ui/qt/widgets/controls.h
```diff
@@ -188,89 +189,63 @@ class ParamControl : public ToggleControl {
...
+        background-color: #4a4a4a;
+      }
+      QPushButton:checked {
+        background-color: #33Ab4C;
+      }
+    )";
     key = param.toStdString();
+    int value = atoi(params.get(key).c_str());
+
+    QButtonGroup *button_gorup = new QButtonGroup(this);
```
sshane, line 226:
```suggestion
    QButtonGroup *button_group = new QButtonGroup(this);
```

sshane:
can we do something similar on the planner side? it also reads the param and tries to convert to int, but doesn't check the data first

PR #28430, opened 2023-06-06
Update zh-CHT translation

protonchang:
* Add new translation


protonchang:
Sure, but it seems that @eFiniLan is already working on this.
I'll just close my PR then.
His translation also looks good :)

PR #28387, opened 2023-06-03
detect toyota tss-p canfilter device as smartDsu to enable long

Smartype:

PR #28440, opened 2023-06-07
Toyota: experimental longitudinal when using smartDSU on radar-ACC TSS2 car

eFiniLan:
for other models such as VAG / Honda, op longitudinal is allowed only when the experimental_long is enabled.

so in Toyota, we should implement such logic on radar-less car (RADAR_ACC_CAR) w/ can filter (smart dsu) installed.

case 1, no can filter installed - use stock long
case 2, can filter installed + experimental mode is off - use stock long
case 3, can filter installed + experimental mode is on - use op long


sshane:
does this rely on https://github.com/commaai/openpilot/pull/28387?

eFiniLan:
> does this rely on #28387?

not directly, it's related to this implementation: https://github.com/commaai/openpilot/pull/27417 (just use the same same sdsu firmware)

I know a couple of ppl using this on Camry/CHR

by doing this my 2021 C-HR can run vision only op long.

that's why I think confirmation is required :)

selfdrive/car/toyota/interface.py
```diff
@@ -214,6 +214,14 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
 
     # if the smartDSU is detected, openpilot can send ACC_CMD (and the smartDSU will block it from the DSU) or not (the DSU is "connected")
     ret.openpilotLongitudinalControl = bool(ret.flags & ToyotaFlags.SMART_DSU) or ret.enableDsu or candidate in (TSS2_CAR - RADAR_ACC_CAR)
+
+    # toyota radar acc car should consider radarUnavailable = True
+    # and when radarUnavailable, openpilot longitudinal control is allowed when smart dsu is installed and experimental_long is on
+    ret.radarUnavailable = candidate in RADAR_ACC_CAR
+    if ret.radarUnavailable:
+      ret.experimentalLongitudinalAvailable = bool(ret.flags & ToyotaFlags.SMART_DSU)
+      ret.openpilotLongitudinalControl = experimental_long and ret.experimentalLongitudinalAvailable
```
adeebshihadeh, line 223: this should only be set once. can you combine with the logic above?

adeebshihadeh:
@sshane can you get this merged soon?

sshane:
How does this now relate to https://github.com/commaai/openpilot/pull/28387? This detects the SDSU, but that one detects a different message from a device *not* based on a panda.

sshane:
Cleaned up the logic a bit and fixed a bug (it still checked if the SDSU was there and enabled longitudinal regardless of toggle `or bool(ret.flags & ToyotaFlags.SMART_DSU)`). Please let me know if this accomplishes what you were trying to do.

This also only requires experimental toggle for the TSS2 radar-ACC cars, do we have any TSS-P radar-ACC cars which can be used with the SDSU?

eFiniLan:
> @eFiniLan How does this now relate to #28387? That detects a different address.
> 
> Can you tell me if these assumptions are correct?:
> 
> The original SDSU and firmware for SDSU works for TSS2 radar-based ACC cars, because the radar essentially performs the same functions as the DSU (sends ACC_CONTROL which SDSU blocks conditionally when we send our own).
> 
> And the "CAN filter" device in that PR above is an SDSU alternative that always blocks the ACC_CONTROL message.

Yes :)

so SDSU basically blocks the ACC_CONTROL message coming out from DSU.

and with those RADAR_ACC_CARs, they have radar sending ACC_CONTROL.

so the CAN filter above simply blocks the ACC_CONTROL msg just like what SDSU does.
(we use panda hardware + same BUS mapping as SDSU = we can use SDSU firmware unmodified)

selfdrive/car/toyota/interface.py
```diff
@@ -222,8 +222,20 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
...
+    use_sdsu = bool(ret.flags & ToyotaFlags.SMART_DSU)
+    if candidate in RADAR_ACC_CAR:
+      ret.experimentalLongitudinalAvailable = use_sdsu
+      use_sdsu = use_sdsu and experimental_long
+
+    # openpilot longitudinal enabled by default:
+    #  - non-(TSS2 radar ACC cars) w/ smartDSU installed
+    #  - cars w/ DSU disconnected
+    #  - TSS2 cars with camera sending ACC_CONTROL where we can block it
+    # openpilot longitudinal enabled when experimental long is toggled:
```
sshane, line 236:
```suggestion
    # openpilot longitudinal behind experimental long toggle:
```

sshane:
can you open a PR to put longitudinal behind the toggle for NODSU TSS-P cars too?

PR #28441, opened 2023-06-07
FPv2: assert non-empty FW lists

sshane:


incognitojam (approved):

sshane:
We sometimes check that the length of the lists is not empty to skip it (in at least 1 test), so we can get rid of that if we assume the lists to never be empty.

PR #28443, opened 2023-06-07
ui: disable longitudinal personality settings when using stock long

### LINKED ISSUE #28438: Gray out longitudinal personality setting when using stock long.
Longitudinal personality setting doesn't make sense for cars with stock long. Should be grayed out for those cars. @deanlee you want do this :D?

deanlee:
resolve https://github.com/commaai/openpilot/issues/28438

Screenshot from 2023-06-08 02-29-50


adeebshihadeh (commented):
Can you make all buttons grey if it's disabled?

deanlee:
disable title button:
Screenshot from 2023-06-09 05-24-29

adeebshihadeh:
Oh, I just meant the selected green style shouldn't be used when it's disabled.

deanlee:
Screenshot from 2023-06-09 05-38-59

sshane:
@deanlee experimental mode is not greyed out, but driving personality is when you first boot openpilot without carparams. I think we want to do what experimental mode does?

PR #28447, opened 2023-06-07
process_replay: progress bar support

fredyshox:

### DIFF
--- a/selfdrive/test/process_replay/process_replay.py
+++ b/selfdrive/test/process_replay/process_replay.py
@@ -6,6 +6,7 @@ import platform
 from collections import OrderedDict
 from dataclasses import dataclass, field
 from typing import Dict, List, Optional, Callable
+from tqdm import tqdm
 
 import cereal.messaging as messaging
 from cereal import car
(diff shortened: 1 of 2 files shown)


selfdrive/test/process_replay/process_replay.py
```diff
@@ -338,9 +339,9 @@ def replay_process_with_name(name, lr, *args, **kwargs):
   return replay_process(cfg, lr, *args, **kwargs)
 
 
-def replay_process(cfg, lr, fingerprint=None, return_all_logs=False):
+def replay_process(cfg, lr, fingerprint=None, return_all_logs=False, disable_pbar=False):
```
adeebshihadeh, line 342: didn't know what a pbar was
```suggestion
def replay_process(cfg, lr, fingerprint=None, return_all_logs=False, display_progress=False):
```

adeebshihadeh (commented):
These two changes are unrelated and should be split up.

fredyshox:
Ok, I've force pushed progress bar commit and I will create new PR for trigger_empty_recv

PR #28450, opened 2023-06-07
Multilang: Fix text overlap in pt-BR translation

AlexandreSato:
something in my PC environment had distorted the UI aspect ratio and didn't demonstrate the overlap of texts. Now I double-checked on comma3 itself

image

Before:
image


sshane:
make sure to rebase, think you're based on an old commit

AlexandreSato:
> make sure to rebase, think you're based on an old commit

Yes, you got it right, now I've synced with master and re checked again, thanks!

PR #28453, opened 2023-06-08
Subaru: use latActive for steer request bit

sshane:
Old logic made the bit rapidly cycle on/off when desired torque is near 0

image


sshane:
@martinl want to check one your stock routes to make sure the stock system keeps the bit high when torque is zero but it's still steering?

martinl:
yes, I will test

sshane:
Expected on gen 2 Outback:
image

sshane:
Looks good!

image

image

image

image

image

sshane:
Expected diff 
image

PR #28458, opened 2023-06-08
Honda: set autoResumeSng for docs

sshane:


vanillagorillaa:
Do we want to add a footnote that auto resume can be obtained by adding a pedal? Otherwise LGTM

sshane:
We used to have that footnote for the Toyotas but removed it intentionally iirc, @adeebshihadeh right?

Issue #28267, opened 2023-05-23
Car docs diff bot: unit tests


sshane:
Some kind of test that at least test it doesn't crash, right now it silently fails if the CarInfo structure changes. If the easiest solution to this issue is to make it fail the CI when it's expected to fail in PRs, that's fine too.

sshane:
Yep, it's silently failing from recent changes, very easy to accidentally do the way it's set up: https://github.com/commaai/openpilot/actions/runs/5054044656/jobs/9068524632?pr=28266

sshane:
https://github.com/commaai/openpilot/pull/28459

PR #28459, opened 2023-06-09
car docs: test print_docs_diff

maxime-desroches:
https://github.com/commaai/openpilot/issues/28267

### DIFF
--- a/selfdrive/car/tests/test_docs.py
+++ b/selfdrive/car/tests/test_docs.py
@@ -1,12 +1,17 @@
 #!/usr/bin/env python3
 from collections import defaultdict
+from contextlib import redirect_stdout
+import os
 import re
 import unittest
 
+from common.basedir import BASEDIR
 from selfdrive.car.car_helpers import interfaces, get_interface_attr
 from selfdrive.car.docs import CARS_MD_OUT, CARS_MD_TEMPLATE, generate_cars_md, get_all_car_info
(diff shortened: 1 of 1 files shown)


selfdrive/car/tests/test_docs.py
```diff
@@ -22,6 +27,13 @@ def test_generator(self):
     self.assertEqual(generated_cars_md, current_cars_md,
                      "Run selfdrive/car/docs.py to update the compatibility documentation")
 
+  def test_docs_diff(self):
+    with redirect_stdout(None):
```
sshane, line 31: is this to just get rid of `Dumping car info to...`?

selfdrive/car/tests/test_docs.py
```diff
@@ -22,6 +27,13 @@ def test_generator(self):
...
+      dump_path = os.path.join(BASEDIR, "selfdrive", "car", "tests", "cars_dump")
```
sshane, line 32:
```suggestion
      dump_path = os.path.join(BASEDIR, "selfdrive/car/tests/cars_dump")
```

selfdrive/car/tests/test_docs.py
maxime-desroches: Also to get rid of the potential output of print_car_info_diff. If we don't mind the output, I can remove the line

sshane: the diff should always be none, so should be fine

PR #28463, opened 2023-06-09
process_replay: modeld and dmonitoringmodeld support

fredyshox:


selfdrive/test/process_replay/process_replay.py
```diff
@@ -447,6 +538,24 @@ def _replay_single_process(cfg, lr, fingerprint, disable_progress):
       return log_msgs
 
 
+def setup_vision_ipc(cfg, lr):
+  assert len(cfg.vision_pubs) != 0
+
+  device_type = next(msg.initData.deviceType for msg in lr if msg.which() == "initData")
+  is_eon = device_type == "eon"
```
adeebshihadeh, line 545: we should setup a dict or helper function for this

selfdrive/test/process_replay/process_replay.py
```diff
@@ -447,6 +538,24 @@ def _replay_single_process(cfg, lr, fingerprint, disable_progress):
...
+
+  vipc_server = VisionIpcServer("camerad")
+  if "roadCameraState" in cfg.vision_pubs:
+    vipc_server.create_buffers(VisionStreamType.VISION_STREAM_ROAD, 40, False, *(eon_f_frame_size if is_eon else tici_f_frame_size))
+  if "driverCameraState" in cfg.vision_pubs:
+    vipc_server.create_buffers(VisionStreamType.VISION_STREAM_DRIVER, 40, False, *(eon_d_frame_size if is_eon else tici_d_frame_size))
+  if device_type == "tici" and "wideRoadCameraState" in cfg.vision_pubs:
```
adeebshihadeh, line 552: is the `device_type == "tici"` condition necessary?

selfdrive/test/process_replay/process_replay.py
```diff
@@ -447,6 +538,24 @@ def _replay_single_process(cfg, lr, fingerprint, disable_progress):
...
 
+def setup_vision_ipc(cfg, lr):
+  assert len(cfg.vision_pubs) != 0
+
+  device_type = next(msg.initData.deviceType for msg in lr if msg.which() == "initData")
+  is_eon = device_type == "eon"
+
+  vipc_server = VisionIpcServer("camerad")
+  if "roadCameraState" in cfg.vision_pubs:
+    vipc_server.create_buffers(VisionStreamType.VISION_STREAM_ROAD, 40, False, *(eon_f_frame_size if is_eon else tici_f_frame_size))
```
adeebshihadeh, line 549: no need for 40 buffers here, it'll just use more memory unnecessarily

selfdrive/test/process_replay/process_replay.py
```diff
@@ -25,25 +27,27 @@
 TIMEOUT = 15
 PROC_REPLAY_DIR = os.path.dirname(os.path.abspath(__file__))
 FAKEDATA = os.path.join(PROC_REPLAY_DIR, "fakedata/")
-
+VIPC_STREAM = {"roadCameraState": VisionStreamType.VISION_STREAM_ROAD, "driverCameraState": VisionStreamType.VISION_STREAM_DRIVER,
```
adeebshihadeh, line 30: this dict is in a bunch of places, we should move it somewhere common

selfdrive/test/process_replay/process_replay.py
fredyshox: well probably not, now I see that it'd be more versatile to check if messages if this type are present in provided logs

selfdrive/test/process_replay/process_replay.py
fredyshox: now it's 5

selfdrive/test/process_replay/process_replay.py
fredyshox: I've added helpers for handling vision related metadata (camera states, streams, dt, frame sizes) to `vision_meta.py`.

selfdrive/test/process_replay/process_replay.py
fredyshox: no longer need this

selfdrive/test/process_replay/process_replay.py
adeebshihadeh: should only need one or maybe two since everything's blocking

fredyshox:
Updated the model_replay_ref_commit - the outputs are the same as before, just the log order changed (verified on c3).

selfdrive/test/process_replay/process_replay.py
```diff
@@ -11,40 +11,43 @@
...
 from selfdrive.manager.process_config import managed_processes
 from selfdrive.test.process_replay.helpers import OpenpilotPrefix
+from selfdrive.test.process_replay.vision_meta import meta_from_camera_state, available_streams
 from selfdrive.test.process_replay.migration import migrate_all
 
 # Numpy gives different results based on CPU features after version 19
 NUMPY_TOLERANCE = 1e-7
 CI = "CI" in os.environ
 TIMEOUT = 15
+MAX_FRAMES_PER_SEGMENT = 1200
```
haraschax, line 29: Why is this needed?

selfdrive/test/process_replay/process_replay.py
```diff
@@ -327,6 +376,32 @@ def locationd_config_pubsub_callback(params, cfg):
...
+    main_pub_drained=False,
+    vision_pubs=["roadCameraState", "wideRoadCameraState"],
+  ),
+  ProcessConfig(
+    proc_name="dmonitoringmodeld",
+    pubs=["liveCalibration", "driverCameraState"],
+    subs=["driverStateV2"],
+    ignore=["logMonoTime", "driverStateV2.modelExecutionTime", "driverStateV2.dspExecutionTime"],
+    config_callback=None,
+    init_callback=None,
```
haraschax, line 398: None should be default

selfdrive/test/process_replay/model_replay.py
```diff
@@ -27,20 +24,26 @@
...
 def get_log_fn(ref_commit, test_route):
   return f"{test_route}_model_tici_{ref_commit}.bz2"
 
 
-def replace_calib(msg, calib):
-  msg = msg.as_builder()
-  if calib is not None:
-    msg.liveCalibration.rpyCalib = calib.tolist()
-  return msg
+def trim_logs_to_max_frames(logs, max_frames, frs_types):
```
haraschax, line 34: Can this be removed with the encodeIdx garantuees?

fredyshox: this is still used when its run on PC (so GH CI) to trim logs to 100 frames

selfdrive/test/process_replay/process_replay.py
```diff
@@ -186,6 +187,54 @@ def torqued_rcv_callback(msg, cfg, frame):
...
+    if msg.which() == "initData":
+      self.is_dual_camera = msg.initData.deviceType in ["tici", "tizi"]
+      return False
+    elif msg.which() not in ["roadCameraState", "wideRoadCameraState"]:
+      return False
+    
+    camera_state = getattr(msg, msg.which())
+    if self.min_frame_id is None:
+      self.min_frame_id = camera_state.frameId
+    if (camera_state.frameId - self.min_frame_id) >= MAX_FRAMES_PER_SEGMENT:
```
haraschax, line 211: All this can also be deleted?

selfdrive/test/process_replay/vision_meta.py
```diff
@@ -0,0 +1,46 @@
+from collections import namedtuple
+from cereal.visionipc import VisionStreamType
+from common.realtime import DT_MDL, DT_DMON
+from common.transformations.camera import tici_f_frame_size, tici_d_frame_size, tici_e_frame_size, eon_f_frame_size, eon_d_frame_size
+
+class CameraFrameSizes:
```
adeebshihadeh, line 6: this is just a dict?

fredyshox: it mimics the behavior of a dict - that way i don't have to list sizes for every possible device type

adeebshihadeh: why not just use a dict?

fredyshox: it would be much larger and would have duplicate entries for tizi and tici

fredyshox: but it's the same thing, i can make separate declarations for each camera type

selfdrive/test/process_replay/process_replay.py
```diff
@@ -186,6 +191,48 @@ def torqued_rcv_callback(msg, cfg, frame):
...
+  return msg.which() == "driverCameraState"
+
+
+class ModeldCameraSyncRcvCallback:
+  def __init__(self):
+    self.road_present = False
+    self.wide_road_present = False
+    self.is_dual_camera = True
+
+  def __call__(self, msg, cfg, frame):
```
haraschax, line 204: I find functions written like this very confusing, there's a lot of returns and goes back and forth between state management and returning.

I would write it like this:

```
class ModeldCameraSyncRcvCallback:
  def __init__(self):
    self.road_present = False
    self.wide_road_present = False
    self.is_dual_camera = True

  def __call__(self, msg, cfg, frame):
    if msg.which() == "initData":
      self.is_dual_camera = msg.initData.deviceType in ["tici", "tizi"]
    elif msg.which() == "roadCameraState":
      self.road_present = True
    elif msg.which() == "wideRoadCameraState":
      self.wide_road_present = True

    if self.road_present and self.wide_road_present
      self.road_present, self.wide_road_present = False, False
      return True
    elif self.road_present and not self.dual_camera:
      self.road_present = False
      return True
    else:
      return False
```

fredyshox: it was probably refactored N times, before ended up looking like this... your snippet looks much cleaner :+1:

PR #28471, opened 2023-06-09
cabana: display signal details in tooltip 

deanlee:


sshane:
maybe only show minimal but important info on the signal tooltip up top?

deanlee:
How about this?

sshane:
looks good!

PR #28477, opened 2023-06-10
UI: split out driving toggles

incognitojam:
Screenshot from 2023-06-09 17-07-52
Screenshot from 2023-06-09 17-07-54


adeebshihadeh:
Not clear that this is a strict improvement, closing for now.

PR #28475, opened 2023-06-10
FwQueryConfig: add platform code support and tests

sshane:
Adds ability to add a function to get platform codes from a set of FW versions to the FW query config as well as generic and Hyundai-specific tests. Split from and prep for https://github.com/commaai/openpilot/pull/26939

PR #28478, opened 2023-06-10
env: add dateutil types

sshane:
for https://github.com/commaai/openpilot/pull/28475

needed since dateutil is used in a typed function


incognitojam:
MGMT

incognitojam (approved):

PR #26939, opened 2023-01-13
HKG fingerprinting: add matches from platform codes

sshane:
Still a draft. There are some cars that use the same platform codes, but are fundamentally different as far as openpilot is concerned. One big example are the older Ioniqs; some have radars, some use FCA or SCC for AEB, some are legacy, some send the LFA message, some support different torques, etc.

For these cars, we just ignore the multiple matches. @adeebshihadeh I am thinking we can track the cars we won't support with this fuzzy fingerprinting and work to combine the platforms over time (with good detection for the different features) 

All Hyundai:

```python
HYUNDAI IONIQ HYBRID 2017-2019      : radar: {b'AEhe'}, camera: {b'AEH'}
HYUNDAI IONIQ PLUG-IN HYBRID 2019   : radar: {b'AEhe'}, camera: {b'AEP'}
HYUNDAI IONIQ PHEV 2020             : radar: {b'AEhe'}, camera: {b'AEP'}
HYUNDAI IONIQ ELECTRIC 2020         : radar: {b'AEev'}, camera: {b'AEE'}
HYUNDAI IONIQ ELECTRIC LIMITED 2019 : radar: {b'AEev'}, camera: {b'AEE'}
HYUNDAI IONIQ HYBRID 2020-2022      : radar: {b'AEhe'}, camera: {b'AEH'}
HYUNDAI SONATA 2020                 : radar: {b'DN89', b'DN8'}, camera: {b'DN8'}
HYUNDAI SONATA 2019                 : radar: {b'LF'}, camera: {b'LFF'}
HYUNDAI TUCSON 2019                 : radar: {b'TL'}, camera: {b'TL'}
HYUNDAI SANTA FE 2019               : radar: {b'TM'}, camera: {b'TM'}
HYUNDAI SANTA FE 2022               : radar: {b'TM'}, camera: {b'TMA', b'TM'}
HYUNDAI SANTA FE HYBRID 2022        : radar: {b'TMhe'}, camera: {b'TMH'}
HYUNDAI SANTA FE PlUG-IN HYBRID 2022: radar: {b'TMhe'}, camera: {b'TMP'}
KIA STINGER GT2 2018                : radar: {b'CK'}, camera: {b'CK'}
KIA STINGER 2022                    : radar: {b'CK'}, camera: {b'CK'}
HYUNDAI PALISADE 2020               : radar: {b'ON', b'LX2'}, camera: {b'ON', b'LX2'}
[...]

PR #28481, opened 2023-06-10
fuzzy fingerprinting: ensure all ECUs exist before matching

### LINKED ISSUE #1095: fuzzy fingerprinting: false positive matches with missing critical ECUs
For example, if the radar or camera failed to respond or does not exist (or any other important ECU), but we get enough matches with other ECUs, we would match when we really shouldn't.

this is already masking a fingerprinting/car/hardware bug with a 2022 Civic where the camera fails to respond, but we match any way: https://github.com/commaai/openpilot/pull/27126

sshane:
related to https://github.com/commaai/openpilot/pull/26939 and resolves https://github.com/commaai/opendbc/issues/1095

Very similar to what exact matching does where it invalidates candidates if the ECU is missing on the car


selfdrive/car/fw_versions.py
```diff
@@ -91,14 +93,15 @@ def match_fw_to_car_fuzzy(fw_versions_dict, log=True, exclude=None):
...
-  # if there are enough matches. FIXME: parameterize this or require all ECUs to exist like exact matching
   if len(matched_ecus) >= 2:
-    if log:
-      cloudlog.error(f"Fingerprinted {candidate} using fuzzy match. {len(matched_ecus)} matching ECUs")
-    return {candidate}
-  else:
-    return set()
+    # TODO: verify that this is safe and makes sense by looking at data
+    # Ensure all ECUs in database have responded to FW queries before matching
+    candidate_ecus = {(addr[1], addr[2]) for addr in FW_VERSIONS[candidate]}
```
sshane, line 99: TODO: use the config and be make-specific

selfdrive/car/fw_versions.py
```diff
@@ -91,14 +93,15 @@ def match_fw_to_car_fuzzy(fw_versions_dict, log=True, exclude=None):
...
+    if len(candidate_ecus - seen_ecus) > 0:
```
sshane, line 100:
```suggestion
    if len(candidate_ecus - seen_ecus) == 0:
```

sshane:
We currently rely on this behavior for some cars. TODO: come back with what those cars are (running on data) and re-evaluate

(Civic 2022 is one of them)

sshane:
Fixed the spotty Civic query that was blocking this! https://github.com/commaai/openpilot/pull/31963

sshane:
There's many Toyotas with one or two routes where fwdCamera, fwdRadar, eps, or abs are missing. Still looking for rlogs to verify the reason. May be fixed by https://github.com/commaai/openpilot/issues/32242

sshane:

```bash
PR_NUMBER=33045
```

PR #28490, opened 2023-06-11
ui: hide sidebar when map shows

sshane:


sshane:
@deanlee can you think of a cleaner way to do this? We can check for this case every on every update but thought it might be be cleaner to use signals (eg. `if (sidebar->isVisible() && onroad->isMapVisible()) sidebar->hide();`)

deanlee:
can't think of a better way right now,  this might be the most straightforward approach.

sshane:
@incognitojam is this good to merge?

PR #28510, opened 2023-06-12
laikad: move 'from_qcom_source' into try block

deanlee:
the `NotImplementedError` thrown by `from_qcom_source` was not caught properly.


haraschax:
Thanks!

PR #28511, opened 2023-06-12
Panda logs

robbederks:


selfdrive/boardd/pandad.py
```diff
@@ -125,6 +126,16 @@ def main() -> NoReturn:
...
 
+      # store last panda logs
+      LOG_AMOUNT = 100
+      logs = {}
+      for panda in pandas:
+        try:
+          logs[panda.get_usb_serial()] = panda.get_logs(True)[-LOG_AMOUNT:]
+        except Exception:
+          cloudlog.exception(f"Error getting logs for {panda.get_usb_serial()}")
+      params.put("PandaLogs", json.dumps(logs, default=str))
```
adeebshihadeh, line 137: the param doesn't seem quite right since you don't get the logs if you don't go onroad. why not just cloudlog? would also add a log if it exceeds 100

robbederks:
This works, but adds about an extra two seconds on SPI pandas since it tries to get all available logs. Should be optimized more

adeebshihadeh:
> This works, but adds about an extra two seconds on SPI pandas since it tries to get all available logs. Should be optimized more

Not sure why this would only affect SPI pandas? And this doesn't seem to be the case. We've got good tests for the pandad startup time - https://github.com/commaai/openpilot/blob/master/selfdrive/boardd/tests/test_pandad.py

robbederks:
Oh I fixed that issue since that comment, seems like I forgot to merge this PR, I thought I did!

PR #28527, opened 2023-06-13
Hyundai: remove old Sonata FW

sshane:


sshane:
first commit used to belong to `2ac344325fa49bac`

PR #28530, opened 2023-06-13
test: car interface fuzzy testing

maxime-desroches:
Found an index out of bounds error for the body carcontroller https://github.com/commaai/openpilot/pull/28539

need to add hypothesis to agnos


selfdrive/car/tests/test_car_interfaces.py
```diff
@@ -9,11 +11,36 @@
 from selfdrive.car.car_helpers import interfaces
 from selfdrive.car.fingerprints import _FINGERPRINTS as FINGERPRINTS, all_known_cars
 
+def get_random_car_control():
```
adeebshihadeh, line 14: use the existing stuff and extend it if necessary https://github.com/commaai/openpilot/blob/master/selfdrive/test/process_replay/test_fuzzy.py

adeebshihadeh (commented):
Can you make a separate PR to fix the body bug?

selfdrive/test/process_replay/fuzzy_generation.py
```diff
@@ -0,0 +1,69 @@
+import hypothesis.strategies as st
+import random
```
adeebshihadeh, line 2: let's move this to selfdrive/test/ and need to add to release/files_common

adeebshihadeh (approved):

selfdrive/test/process_replay/fuzzy_generation.py
maxime-desroches: I will make a separate PR for the test_car_interface stuff since it needs an agnos update

PR #28531, opened 2023-06-13
HKG: use platform codes to fuzzy fingerprint

sshane:
v2 of https://github.com/commaai/openpilot/pull/26939

Uses combination of platform codes, part numbers, and ECU FW dates to fuzzy match. Think of it as a less-specific FW version. Should catch cases where a car uses a new part number we haven't seen (and requires different tuning), as well as new model years we don't support yet (with possible different architectures like the new Palisades).

- [x] first tries standard fuzzy fingerprinting, then if no matches tries hyundai fuzzy fingerprinting
- [x] invalidate all ICE and hybrid CAN FD platforms as they often do not have specifiers to tell the difference
  - [ ] can remove once we add hvac ECU which tells us this!
- [x] check date of FW version to prevent newer model years we haven't seen from fuzzy matching
- [x] check part number to catch cases like the Tucson using a Santa Cruz EPS

PR #28532, opened 2023-06-13
HKG: split up platform code parsing

sshane:
split out regex split up from https://github.com/commaai/openpilot/pull/28531. makes parsing FW versions more modular. the idea is that we end up with the FW versions with unnecessary specific information masked out (like left/right hand drive, country of origin, random characters, etc.)

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -349,7 +349,7 @@ FINGERPRINTS = {
 
 
 def get_platform_codes(fw_versions: List[bytes]) -> Set[bytes]:
-  codes: DefaultDict[bytes, Set[bytes]] = defaultdict(set)
+  codes: DefaultDict[bytes, Set[Optional[bytes]]] = defaultdict(set)
   for fw in fw_versions:
     code_match, date_match = (PLATFORM_CODE_FW_PATTERN.search(fw),
                               DATE_FW_PATTERN.search(fw))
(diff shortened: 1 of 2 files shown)


selfdrive/car/hyundai/values.py
```diff
@@ -392,7 +394,8 @@ def get_platform_codes(fw_versions: List[bytes]) -> Set[bytes]:
 
 # Regex patterns for parsing platform code, FW date, and part number from FW versions
 PLATFORM_CODE_FW_PATTERN = re.compile(b'((?<=' + HYUNDAI_VERSION_REQUEST_LONG[1:] +
-                                      b')[A-Z]{2}[A-Za-z0-9]{0,2})(?:.*([0-9]{6}))?')
+                                      b')[A-Z]{2}[A-Za-z0-9]{0,2})')
+DATE_FW_PATTERN = re.compile(b'(?<=[ -])([0-9]{6}$)')
 PART_NUMBER_FW_PATTERN = re.compile(b'(?<=[0-9][.,][0-9]{2} )([0-9]{5}[-/]?[A-Z][A-Z0-9]{3}[0-9])')
```
sshane, line 399: Now checks that date string is at the end of the FW version as well as is preceded by a space or dash (Hyundai Genesis coming in yet again with the exception)

selfdrive/car/hyundai/values.py
```diff
@@ -365,14 +367,14 @@ def get_platform_codes(fw_versions: List[bytes]) -> Set[bytes]:
       continue
 
     try:
-      parsed = {datetime.strptime(date.decode()[:4], '%y%m') for date in dates}
+      parsed = {datetime.strptime(cast(bytes, date).decode()[:4], '%y%m') for date in dates}
```
sshane, line 370: this is guaranteed to be bytes, but mypy doesn't handle it well

PR #28533, opened 2023-06-13
locationd: initial value for gps_std_factor

fredyshox:
This was uninitialized when using the default constructor causing undefined behaviour.

### DIFF
--- a/selfdrive/locationd/locationd.cc
+++ b/selfdrive/locationd/locationd.cc
@@ -673,8 +673,8 @@ void Localizer::configure_gnss_source(LocalizerGnssSource source) {
 }
 
 int Localizer::locationd_thread() {
-  bool ublox_available = Params().getBool("UbloxAvailable", true);
-  this->configure_gnss_source(ublox_available ? LocalizerGnssSource::UBLOX : LocalizerGnssSource::QCOM);
+  LocalizerGnssSource source = Params().getBool("UbloxAvailable", true) ? LocalizerGnssSource::UBLOX : LocalizerGnssSource::QCOM;
+  this->configure_gnss_source(source);
   const char* gps_location_socket;
   if (this->gnss_source == LocalizerGnssSource::UBLOX) {
     gps_location_socket = "gpsLocationExternal";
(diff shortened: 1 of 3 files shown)


haraschax:
`this->ublox_available` is a bit confusing, can you make an enum called something like gnss_source>

And then you can do
```
if (Params().getBool("UbloxAvailable", true)){
  gnss_source = UBLOX
} else {
  gnss_source = QCOM 
}
```
And then replace setup_gps with:
```
this->configure_gnss_source(gnss_source);
```

fredyshox:
Added `LocalizerGnssSource` and `configure_gnss_source`

selfdrive/locationd/locationd.cc
```diff
@@ -84,10 +84,10 @@ Localizer::Localizer() {
 
   VectorXd ecef_pos = this->kf->get_x().segment<STATE_ECEF_POS_LEN>(STATE_ECEF_POS_START);
   this->converter = std::make_unique<LocalCoord>((ECEF) { .x = ecef_pos[0], .y = ecef_pos[1], .z = ecef_pos[2] });
+  this->gnss_source = gnss_source;
```
haraschax, line 87: Move this to configure_gnss_source?

fredyshox: you're right

selfdrive/locationd/locationd.cc
```diff
@@ -326,7 +326,7 @@ void Localizer::handle_gps(double current_time, const cereal::GpsLocationData::R
   VectorXd ecef_pos = this->converter->ned2ecef({ 0.0, 0.0, 0.0 }).to_vector();
   VectorXd ecef_vel = this->converter->ned2ecef({ log.getVNED()[0], log.getVNED()[1], log.getVNED()[2] }).to_vector() - ecef_pos;
   float ecef_pos_std;
-  if (ublox_available) {
+  if (gnss_source == LocalizerGnssSource::UBLOX) {
```
haraschax, line 328: Actually while you're at it, fix this with configure_gnss_source too? You can define a variance_factor and vertical_variance factor, and set them to 0,1,3 appropriately.

selfdrive/locationd/locationd.cc
```diff
@@ -372,7 +372,7 @@ void Localizer::handle_gnss(double current_time, const cereal::GnssMeasurements:
   }
 
   double sensor_time = log.getMeasTime() * 1e-9;
-  if (ublox_available)
+  if (gnss_source == LocalizerGnssSource::UBLOX)
```
haraschax, line 374: Same here, just define a gnss_time_offset in  configure_gnss_source

haraschax:
LGTM if tests pass

PR #28539, opened 2023-06-14
Fix index out of bound error for body carcontroller

maxime-desroches:


selfdrive/car/body/carcontroller.py
```diff
@@ -40,7 +40,7 @@ def update(self, CC, CS, now_nanos):
     torque_l = 0
     torque_r = 0
 
-    llk_valid = len(CC.orientationNED) > 0 and len(CC.angularVelocity) > 0
+    llk_valid = len(CC.orientationNED) > 1 and len(CC.angularVelocity) > 1
```
adeebshihadeh, line 43: @haraschax longer term, how do you feel about changing a bunch of these lists to structs?

haraschax: Seems like a good idea

PR #28548, opened 2023-06-14
Added new Volt Fingerprint

prodigz:
Added Volt with latest ECM from ACDelco


morrislee:
is this camera based ACC as well? if so, we might need to add the missing address like PR #28331 

These addresses randomly shows up on in a blue moon, perhaps due to onstar from time to time

prodigz:
I added a radar but no stock acc.  The fingerprint passed one day of testing but I can test further before you merge if you like.

sshane:
What is your dongle id? To confirm, this car originally didn't have adaptive cruise control and you flashed ECUs to get it? What about lane keeping? Can it actuate brakes to bring you to a stop or is this regen only?

prodigz:
Can we abandon this PR for now?  After working with it more I think the fingerprint may not be correct.  With the modifications it requires that I unplug my original LKA camera to work as expected.  After doing so it seems that my finger print matches an existing one an this may not be needed.

PR #28554, opened 2023-06-15
cabana: use raw signal value to match value descriptions

deanlee:
the value in value description is a signal raw value transferred on the bus.  we should use raw value to match it.

fixed the issue: https://github.com/commaai/openpilot/discussions/26091#discussioncomment-6179534


incognitojam:
awesome thanks!

PR #28557, opened 2023-06-15
Toyota: log when AEB is disabled unexpectedly

sshane:
We can label this specific case as an ACC fault, but the PCM isn't actually faulted here, and neither is the camera. It's just telling the car that it won't actuate PSC/AEB. There's a few different reasons it will lock out too: radar alignment, camera visibility poor, over temperature, etc. This catches all of those as well as the user disabling AEB (do we want that though?)

### DIFF
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -118,6 +118,7 @@ class CarState(CarStateBase):
       if not (self.CP.flags & ToyotaFlags.SMART_DSU.value):
         self.acc_type = cp_acc.vl["ACC_CONTROL"]["ACC_TYPE"]
       ret.stockFcw = bool(cp_acc.vl["ACC_HUD"]["FCW"])
+      ret.aebDisabled = cp_acc.vl["ACC_HUD"]["PCS_INDICATOR"] != 0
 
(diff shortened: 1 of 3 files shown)


selfdrive/car/toyota/carstate.py
```diff
@@ -118,6 +118,7 @@ def update(self, cp, cp_cam):
       if not (self.CP.flags & ToyotaFlags.SMART_DSU.value):
         self.acc_type = cp_acc.vl["ACC_CONTROL"]["ACC_TYPE"]
       ret.stockFcw = bool(cp_acc.vl["ACC_HUD"]["FCW"])
+      ret.aebDisabled = cp_acc.vl["ACC_HUD"]["PCS_INDICATOR"] != 0
```
sshane, line 121: 1 is user disabled (or lockout), 2 is fault.

sshane:
@adeebshihadeh do we want to generically enforce AEB is turned on (if possible), or should we just add lockouts we're aware of into `accFaulted`? Since we can likely engage if openpilotLongitudinalControl, we might also want to log as a non-critical fault.

sshane:
If anything, this should log overheated camera as ACC faults (only when stock long). Need to revisit looking at data in future

PR #28558, opened 2023-06-15
cabana: Prevent dialog that load autosaved file from blocking replay->start()

deanlee:
fixed issue:  https://github.com/commaai/openpilot/discussions/26091#discussioncomment-6184481


sshane:
ty!

Issue #25756, opened 2022-09-12
Mazda: query FW versions on non-OBD bus


sshane:

sshane:
@Jafaral want to take a look into this? We'd like to be able to query on bus 0 essentially. First step would be to just see if it works!

Jafaral:
@sshane will do. I'm away from my car for a couple of weeks at least.

Jafaral:
@sshane, how exactly should I run this test?

sshane:
@Jafaral add a field called `bus=0`: https://github.com/commaai/openpilot/blob/master/selfdrive/car/mazda/values.py#L69

sshane:
moved to https://github.com/commaai/openpilot/issues/26643

PR #28569, opened 2023-06-16
Mazda: query and log FW on bus 0

sshane:
prep for https://github.com/commaai/openpilot/issues/25756


sshane:
@Jafaral can you try this relatively soon to see if it returns anything? If so we can merge it for the next release to gather data

sshane:
We do get some stuff on bus 0, worth merging as-is

Issue #25754, opened 2022-09-12
Subaru: query FW versions on non-OBD bus


sshane:

sshane:
moved to https://github.com/commaai/openpilot/issues/26643

jnewb1:
tried on the outback. bus 0 you only get the camera, bus 1 you get powertrain ecus (transmission, engine, abs). Couldn't get eps on any bus

sshane:
Great, let's add bus 1 (no OBD) and 0 then. Can you also check later with panda's query_fw_versions.py + `--nonstandard` with a good scan of each bus to see if there's another ECU we're not currently querying that we might want to? I saw 0x730 returns something, that's the adas ECU on Hyundai.

sshane:
Ah, we don't for gen 1:

```
               'carFw': [{'address': 1862,
                          'brand': 'subaru',
                          'bus': 1,
                          'ecu': 'eps',
                          'fwVersion': b'\n\xc0\x04\x00',
                          'logging': False,
                          'obdMultiplexing': True,
                          'request': [b'>\x00', b'"\xf1\x82'],
                          'responseAddress': 1870,
                          'subAddress': 0},
```

PR #28571, opened 2023-06-16
Subaru: query and log FW on bus 0

sshane:
prep for https://github.com/commaai/openpilot/issues/25754

will test on outback tomorrow to make sure we get something


sshane:
@jnewb1 can you test this? Our Outback is on a road trip!

jnewb1:
> @jnewb1 can you test this? Our Outback is on a road trip!

d9df6f87e8feff94|2023-06-16--21-02-09

i just started the car, do you need anything else?

sshane:
Thanks for the test, looks like we don't get much -- just the camera. No need to query for now

PR #28585, opened 2023-06-17
Nissan FPv2: log different diagnostic session types

sshane:
Need to test on an X-Trail (which mainly uses the KWP queries). On the Bronco sending `b"\x10\xc0"` to the engine causes it to enter neutral tow mode (and it doesn't respond with a positive or negative code, just times out). No other subfunction 0x00 to 0xFF causes this behavior.

I also tried prepending the diagnostic entrance command with tester present, another diagnostic entrance command with different subfunction, appending the KWP `STOP_DIAGNOSTIC_SESSION` (0x20) service, etc. which did not make the alert go away.


sshane:
@avolmensky can you try this when you get the chance?

avolmensky:
> @avolmensky can you try this when you get the chance?

I am overseas until the end of July, will not be able to test this one at the moment

PR #28588, opened 2023-06-17
Subaru: read LKAS buttons

jnewb1:
**Description**
Starting to break up my [long PR](https://github.com/commaai/openpilot/pull/28403/files) into smaller pieces. This will log button presses on Subaru.

**Verification**
~~Should not modify any current behavior, will just be useful later.~~
I also cleaned up some preglobal stuff, so these changes will need to be verified.

depends: https://github.com/commaai/opendbc/pull/861


selfdrive/car/subaru/carstate.py
```diff
@@ -1,10 +1,36 @@
...
 from opendbc.can.can_define import CANDefine
 from common.conversions import Conversions as CV
 from selfdrive.car.interfaces import CarStateBase
 from opendbc.can.parser import CANParser
-from selfdrive.car.subaru.values import DBC, CAR, GLOBAL_GEN2, PREGLOBAL_CARS, SubaruFlags
+from selfdrive.car.subaru.values import DBC, CAR, GLOBAL_GEN2, PREGLOBAL_CARS, Buttons, SubaruFlags
+
+PREV_BUTTON_SAMPLES = 8
+
+def cruise_buttons_conversion(cruise_buttons):
```
sshane, line 12: Think the other ports use a dict constant in values.py or similar

selfdrive/car/subaru/carstate.py
```diff
@@ -13,6 +39,8 @@ def __init__(self, CP):
     can_define = CANDefine(DBC[CP.carFingerprint]["pt"])
     self.shifter_values = can_define.dv["Transmission"]["Gear"]
 
+    self.cruise_buttons = deque([Buttons.NONE] * PREV_BUTTON_SAMPLES, maxlen=PREV_BUTTON_SAMPLES)
```
sshane, line 42: do you need a deque?

jnewb1: was copying hyundai, but i think we only need 1 sample really https://github.com/commaai/openpilot/blob/2ebd7ab088ade61bbf661c140483fc477d444bc2/selfdrive/car/hyundai/carstate.py#L22

selfdrive/car/subaru/carstate.py
jnewb1: if we merge "Main", "Resume", and "Set" into a single DBC field then we could do that (like preglobal has right now)

PR #28589, opened 2023-06-17
Subaru: Use Cruise_Status for set speed instead of dashstatus

jnewb1:
**Description**
DashStatus isn't available when ES is disabled. I found the same set speed signal in Cruise_Status, at least for global.

**Verification**
Been using this change on my long PR without issues.

https://github.com/commaai/opendbc/pull/859

### DIFF
--- a/selfdrive/car/subaru/carstate.py
+++ b/selfdrive/car/subaru/carstate.py
@@ -56,7 +56,11 @@ class CarState(CarStateBase):
     cp_cruise = cp_body if self.car_fingerprint in GLOBAL_GEN2 else cp
     ret.cruiseState.enabled = cp_cruise.vl["CruiseControl"]["Cruise_Activated"] != 0
     ret.cruiseState.available = cp_cruise.vl["CruiseControl"]["Cruise_On"] != 0
-    ret.cruiseState.speed = cp_cruise.vl["Cruise_Status"]["Cruise_Set_Speed"] * CV.KPH_TO_MS
+
+    if self.car_fingerprint in PREGLOBAL_CARS:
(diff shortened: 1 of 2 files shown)


sshane:
If we're able to read buttons and send our own cruise speed (it's not tied to the PCM), we can use the pcmCruise flag instead.

jnewb1:
> If we're able to read buttons and send our own cruise speed to the dash (it's not tied to the PCM), we can use the pcmCruise flag instead (false) when openpilotLongitudinalControl=true. Besides, I think some newer Subaru we don't yet support don't have CruiseControl. Does this message still exist on them? Check out Hyundai or GM camera ACC for similar implementations of switching pcmCruise gated on longitudinal

I had that working on my original long POC before I realized we could just pass the buttons through ES_Distance to keep the stock CC status. If you think that's better long term I can switch to that. Only thing is you will need different logic on the panda for gen2, because the buttons will be read through UDS instead of Cruise_Buttons.

jnewb1:
Cruise_Status is available on a 23 forester

sshane:
So the PCM and camera independently keep track of the set speed with the buttons? We can't change/control it with a signal? What happens if you request gas while above the PCM set speed?

On Toyota, the PCM keeps track of the set speed with the hard-wired buttons, and you cannot change it or request accel above it. On Hyundai, the ABS module (which forwards gas commands to the engine) doesn't care whatever set speed is being displayed in the UI and you can change it however you like.

jnewb1:
> 

I think PCM controls set speed, and eyesight just reads that speed. But eyesight gives the PCM the button state through ES_Distance. I'd have to check what happens if you just ignore the PCM set speed and set your own.

PR #28595, opened 2023-06-19
Use the cluster speed when limiting speed

pfeiferj:
The Problem:

Openpilot displays the current speed based off of the speed that is shown in the car's cluster. This is done to prevent a disparity between the car and the information that openpilot is giving, that disparity can lead to less confidence from the user that openpilot is correct. However, the internal openpilot speed is used to limit speed. This leads to openpilot sometimes showing a speed above the set max speed which reduces user confidence that openpilot is correctly limiting based off of the users' max speed setting.

The Solution:
By calculating the difference between the cluster speed and vEgo we can determine the maximum speed that will result in the cluster speed matching the speed the user has set.

### DIFF
--- a/selfdrive/controls/lib/longitudinal_planner.py
+++ b/selfdrive/controls/lib/longitudinal_planner.py
@@ -95,8 +95,11 @@ class LongitudinalPlanner:
     self.mpc.mode = 'blended' if sm['controlsState'].experimentalMode else 'acc'
 
     v_ego = sm['carState'].vEgo
+    v_ego_raw = sm['carState'].vEgoRaw
+    v_ego_cluster = sm['carState'].vEgoCluster
+    v_ego_diff = v_ego_raw - v_ego_cluster if v_ego_cluster > 0 else 0
     v_cruise_kph = min(sm['controlsState'].vCruise, V_CRUISE_MAX)
-    v_cruise = v_cruise_kph * CV.KPH_TO_MS
+    v_cruise = v_cruise_kph * CV.KPH_TO_MS + v_ego_diff
 
     long_control_off = sm['controlsState'].longControlState == LongCtrlState.off
     force_slow_decel = sm['controlsState'].forceDecel


pfeiferj:
Draft as i have not tested the code/theory yet.

pfeiferj:
validated working as expected. This may resolve https://github.com/commaai/openpilot/issues/28446. @sshane let me know if you have any thoughts.

sshane:
I believe https://github.com/commaai/openpilot/pull/29011 should also fix this, can you test? What car do you have?

pfeiferj:
> I believe #29011 should also fix this, can you test? What car do you have?

Sorry for the late response, was on vacation. I'm in a hyundai santa fe 2020 (2019 in op). I'll give that other branch a try

pfeiferj:
@sshane ok, so after testing the other branch I realized these are doing two separate things. Turns out I slightly misunderstood what the issue was asking. So https://github.com/commaai/openpilot/pull/29011 causes the max speed to be set to the currently displayed speed, which seems to be what the ticket is about. This PR, however, makes op use the pcm cruise speed when actively limiting the speed.

An example of the difference could be explained as follows:
With https://github.com/commaai/openpilot/pull/29011 if op is displaying the current speed as 60 mph and I set the max speed to my current speed the max speed will be set to 60 mph. However, OP will then _speed up_ about 1-2 mph and start displaying 61-62 mph while the limit is still set at 60 mph. This is because op is using the vEgo speed instead of the cluster speed when limiting so, if it sees a 60 mph limit it will raise vEgo to 60 mph which is equivalent to 61-62 mph cluster speed.

This PR uses the offset between vEgo and the cluster speed to ensure that when the max says 60 mph, the speed that op is displaying as the current speed is the speed the 60 mph limit is compared against.

adeebshihadeh:
What car is this for? The "cluster" speeds are only meant to match what the car's showing so it doesn't mismatch openpilot. All control is meant to use the real speeds.

pfeiferj:
> What car is this for? The "cluster" speeds are only meant to match what the car's showing so it doesn't mismatch openpilot. All control is meant to use the real speeds.

Any car that the cluster speed and the wheel speed are different. This is so that if I set a max speed the displayed speed/cluster speed doesn't go above the max speed that is set. 

Currently with hkg, for example, if I set a max speed of 60 mph openpilot will end up going a speed that displays as 62 mph. This is confusing to a user as they see openpilot say it's going 2 mph above the max speed they set

PR #28604, opened 2023-06-20
Hyundai: explicitly copy signals to forward (#28020)

sshane:
Reverts commaai/openpilot#28406

let's see if tests catch!


maxime-desroches:
It did fail for some Hyundai by generating these carControl msg: 

https://github.com/commaai/openpilot/actions/runs/5319882482/jobs/9632915697?pr=28604#step:5:10073

PR #28526, opened 2023-06-13
Hyundai: remove old FW

sshane:
{'5c9daa3b70d6f706', '61a9419605869f7a'} these two dongles were one car in 2021 with ` b'\xf1\x8756310AA050\x00\xf1\x00CN7 MDPS C 1.00 1.06 56310AA050\x00 4CNDC106\xf1\xa01.06',` and now have `b'\xf1\x00CN7 MDPS C 1.00 1.06 56310AA050\x00 4CNDC106',` with the new queries

PR #28607, opened 2023-06-20
Toyota: remove bad FW

sshane:
Similar to https://github.com/commaai/openpilot/pull/28526

Added in https://github.com/commaai/openpilot/commit/bfa2d030d2927cb9cfdb01f8f1e614d3790d134b as part of a FW dump, likely a mistake. This FW should only be on the TSS-P Corolla.

Noticed when looking at shared part numbers, not having the length code (which usually means it's from the KWP query) is not common for TSS2 vehicles.


sshane:
There's some EPS FW versions that don't have a length code in TSS2 vehicles, but looks those those are real, and EPS parts are shared across model years.

PR #28612, opened 2023-06-21
Toyota FPv2: log electronic parking brake ECU

sshane:
```python
*** Results for address 0x750 (0x2C) ***

0xF100 IDENTIFICATION_OPTION_VEHICLE_MANUFACTURER_SPECIFIC_DATA_IDENTIFIER: b'\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x84\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'
0xF181 APPLICATION_SOFTWARE_IDENTIFICATION: b'\x01895DF3309100\x00\x00\x00\x00'
0xF186 ACTIVE_DIAGNOSTIC_SESSION: b'\x03'
```

image


sshane:
Added enhancement to https://github.com/commaai/openpilot/issues/27665. We don't need to split out the first sub-address we're querying into its own group, we can query it with the rest of the parallel addrs

selfdrive/car/tests/test_fw_fingerprint.py
```diff
@@ -198,7 +198,7 @@ def test_fw_query_timing(self):
         'nissan': 0.9,
         'subaru': 0.1,
         'tesla': 0.2,
-        'toyota': 0.7,
+        'toyota': 1.0,
```
sshane, line 201: we can bring this down to only 0.1 seconds, but I just want to see a few examples with all queries on this ECU

PR #28616, opened 2023-06-21
Toyota: remove extra ECUs from OBD query

sshane:
Pretty sure only engine uses this, but want to get a few routes before we do this

For recent PRs merged: https://github.com/commaai/openpilot/pull/28613, https://github.com/commaai/openpilot/pull/28612, https://github.com/commaai/openpilot/pull/28628, and https://github.com/commaai/openpilot/pull/28633


sshane:
nope, hybrid ECU uses this, think it's a good idea to query all the new ECUs with this just in case

PR #28625, opened 2023-06-22
Modeld: send confidence class

ZwX1616:


selfdrive/modeld/models/driving.cc
```diff
@@ -199,6 +199,44 @@ void fill_meta(cereal::ModelDataV2::MetaData::Builder meta, const ModelOutputMet
   meta.setHardBrakePredicted(above_fcw_threshold);
 }
 
+void fill_confidence(ModelState* s, cereal::ModelDataV2::Builder &framed) {
```
adeebshihadeh, line 202: perhaps this should go in another packet like the way driverState and driverMonitoringState are split up?

ZwX1616: modelV2 is for "parsed" model output no? Moreover, I think we are already something similar for the model FCW

PR #28626, opened 2023-06-22
car_helpers: add option to disable fingerprinting cache

sshane:
Add environment variable to disable cache.

Good for stress testing fingerprinting


selfdrive/car/car_helpers.py
```diff
@@ -79,14 +79,15 @@ def _get_interface_names() -> Dict[str, List[str]]:
 def fingerprint(logcan, sendcan, num_pandas):
   fixed_fingerprint = os.environ.get('FINGERPRINT', "")
   skip_fw_query = os.environ.get('SKIP_FW_QUERY', False)
+  no_cached_fingerprint = os.environ.get('NO_FINGERPRINT_CACHE', False)
```
sshane, line 82:
```suggestion
  disable_fw_cache = os.environ.get('DISABLE_FW_CACHE', False)
```

PR #28634, opened 2023-06-22
Nissan FPv2: use default diagnostic session

### LINKED ISSUE #28582: Ford: "Park not available" alert
One of the Nissan KWP2000 queries puts the Bronco Sport into a neutral tow mode state, but it doesn't have neutral tow mode...

Summary of issue: Dash says park not available because it thinks we're in neutral tow mode, but exits that mode and alert goes away as soon as you put it in park. Parking pawl activates as normal, it just applies the parking brake on its own.

sshane:
The default KWP2000 diagnostic session type works to get the camera and EPS for Nissan! :partying_face: This should fix the park not available alert on Ford when it fails to fingerprint.

For context, the diagnostic mode was going to be left out of the first Nissan FPv2 PR, but was added back last minute because they realized that certain ECUs didn't respond without it (camera and EPS on this Nissan Leaf). It seemed to have been copied from a Nissan diagnostic program? It seems like that the standard default diagnostic session 0x81 works as well as the undocumented 0xc0 we were using before this. https://github.com/commaai/openpilot/pull/22655#issuecomment-961615172

A blessed Nissan Leaf user driving on master (the only one out of 13)!

Last route without the changes in https://github.com/commaai/openpilot/pull/28585:

<details>
<summary>289d318fded3221a|2023-06-17--13-10-22</summary>

image

</details>

First route with those changes:

<details>
<summary>289d318fded3221a|2023-06-18--09-36-51</summary>

image

</details>

---
@adeebshihadeh 0x81 perfectly replaces the two ECUs normally associated with 0xc0 on this Leaf, you think this is safe to switch, only seeing one car? On an X-TRAIL, abs and engine also rely on the KWP queries (different than this Leaf)


adeebshihadeh:
No, we should check each platform.

sshane:
Should be safe to merge given we still have and rely on CAN fingerprints for many cars:

- Rogue: 0xc0 diagnostic mode ECUs: `[Ecu.engine, Ecu.abs, Ecu.eps. Ecu.combinationMeter]`
  - No route on master to test, but we 100% rely on CAN fingerprinting for this car at the moment.
- Altima: 0xc0 diagnostic mode ECUs: `[Ecu.eps]`
  - 62b0735d47268ec7|2023-07-15--17-28-55 (gets EPS perfectly fine, which used to use mode 0xc0)
- Leaf (non-IC): 0xc0 diagnostic mode ECUs: `[Ecu.eps, Ecu.fwdCamera]`
  - 289d318fded3221a|2023-06-18--09-36-51 (gets EPS and fwdCamera also fine with 0x81)
- Leaf (IC): 0xc0 diagnostic mode ECUs: `[Ecu.eps, Ecu.fwdCamera]`
  - No routes, but should be similar to Leaf (non-IC)
- X-Trail: 0xc0 diagnostic mode ECUs: `[Ecu.engine, Ecu.abs, Ecu.eps, Ecu.combinationMeter]`
  - No routes, but there's only 1 or two users last year + has CAN fingerprint

sshane:
Works for all ECUs on an X-Trail! User mentions that there's no longer any warning lights on startup, was 0xc0 a programming mode? fa9224be239005c7|2023-08-04--12-57-25

PR #28636, opened 2023-06-22
Multilang: Chinese (Traditional) offroad alerts translation

eFiniLan:
CHT translation for offroad alerts + minor changes to punctuation marks


adeebshihadeh:
Thanks!

PR #28638, opened 2023-06-22
update mapbox style

incognitojam:
updated map components and dataset


incognitojam:
Looks like colours have changed slightly

PR #28096, opened 2023-05-03
Toyota: Add missing HIGHLANDER_TSS2 fw

dmcanavan16:

PR #353, opened 2018-09-06
Tried native Stop and Go on Highlander ICE with no comma pedal: Works!

zeeexsixare:
I edited Highlander interface.py to try out stop and go like on the Hybrid.  To my surprise it works out of the box!  Video of operation recorded here: https://comma.slack.com/files/UBHSVCMSS/FCPBN09LP/vid_20180906_075803.mp4

PR #28641, opened 2023-06-22
Toyota: parse FW versions for more reliable fuzzy fingerprinting

sshane:
v2 of https://github.com/commaai/openpilot/pull/27029 so its state is preserved

Todos:
- [x] Figure out a nice way to blacklist
- [x] Do final verification on data (exclude segment's own FW, test no false positives, etc)


sshane:
Ported the Hyundai test to figure out which cars this won't work on at the moment, and most of the failures are because we don't have the hybrid ECU logged yet (getting matches for hybrid and ICE variants). Once we add that, our number of platforms that won't work with this new system goes from 20 to 13, out of 51.

Blacklisted platforms for being too similar with their major platform identifier code (two characters):

```python
AssertionError: Items in the first set but not the second:
'TOYOTA C-HR 2018'
'TOYOTA RAV4 HYBRID 2023'
'TOYOTA CAMRY HYBRID 2021'
'TOYOTA RAV4 HYBRID 2022'
'TOYOTA C-HR HYBRID 2018'
'TOYOTA HIGHLANDER 2020'
'LEXUS RX 2020'
'LEXUS ES HYBRID 2019'
'TOYOTA CAMRY 2021'
'LEXUS ES 2019'
'TOYOTA RAV4 2022'
'TOYOTA HIGHLANDER HYBRID 2020'
'TOYOTA RAV4 2023' : (13, 51)
```

However, we also have minor versions that we can also expand on using in the future (I see a pattern we can use, but I've also seen some overlap, so will take some verification and we can do that after).

sshane:
Here you can see how Toyota increments their software versions for ECU updates, Toyota likes to split features/types on the major version (Y6, Y8, Y7, U1, etc). This is the engine ECU that we don't consider due to its variability, however it is very unique.

image

https://static.nhtsa.gov/odi/tsbs/2018/MC-10149998-9999.pdf

Most other ECUs have a few shared FW versions between MY and even ICE/hybrid (EPS and ABS with the Corolla for example) with just minor version increments, however for bigger generational changes like the 2022 RAV4 with the radar ACC, or the 2023 RAV4 with the move to LTA for lane keeping, those are described in the major versions.

|    | fwdRadar | eps |
| ------------- | ------------- | ----------- |
| RAV4 TSS2  | `{b'8821F-33-01': {b'100', b'200', b'300', b'400'}}`  | `{b'8965B-42-1': {b'70', b'71', b'80', b'81'}, b'8965B-0R-01': {b'200', b'300', b'400'}}` (0R vs. 42 is manufacture location. JP vs. NA) | 
| RAV4 TSS2 2022 (first radar ACC MY)  | `{b'8821F-0R-01': {b'100'}}`  | `{b'8965B-0R-01': {b'500'}}` (no major version change) | 
| RAV4 TSS 2023 (LTA) | `{b'8821F-0R-03': {b'100'}}` | `{b'8965B-0R-11': {b'000'}}` |

sshane:
Wrote a notebook to parse the VINs of our fleet to verify our understanding of the calibration IDs (FW versions), and there seems to be a clear pattern now with the changes in this PR.

Here are a subset of the Avalon dongles for qlogs we have here. The Avalon in openpilot is split into these three platforms: 2016-18 (TSS-P), 2019-21 (TSS-P, has stop and go), and 2021-22 (TSS2).

Each of the ECUs we will look at for fuzzy FP are below comparing their platform codes with their VIN model years and trims (all built in Kentucky). The EPS and radar are shared for all the TSS-P variants (except the 2019), but the camera is nicely split to describe that SNG difference noted above, as it needs to know that for operation:

```python
Checked dongles: 19
Ecu.eps:
  b'8965B-41-0':
    ('2016', '')
    ('2017', 'XLE')
    ('2018', 'XLE')
    ('2020', 'XSE')
    ('2021', 'XLE')
  b'8965B-07-0':
    ('2019', 'XLE')
  b'8965B-41-1':
    ('2022', '')
    ('2022', 'Limited')
    ('2022', 'Touring')
    ('2022', 'XLE')
Ecu.fwdCamera:
  b'8646F-07-01':
    ('2016', '')  # TSS-P is not standard
  b'8646F-07-03':
    ('2017', 'XLE')  # TSS-P is standard
    ('2018', 'XLE')
  b'8646F-07-02':
    ('2019', 'XLE')  # has SNG
    ('2020', 'XSE')
    ('2021', 'XLE')
  b'8646F-41-04':
    ('2022', '')  # TSS2
    ('2022', 'Limited')
    ('2022', 'Touring')
    ('2022', 'XLE')
Ecu.fwdRadar:
  b'8821F-47-02':
    ('2016', '')
    ('2017', 'XLE')
    ('2018', 'XLE')
    ('2019', 'XLE')
    ('2020', 'XSE')
    ('2021', 'XLE')
  b'8821F-62-01':
    ('2022', '')
    ('2022', 'Limited')
    ('2022', 'Touring')
    ('2022', 'XLE')
```

sshane:
So far no mismatches when excluding FW of the car we're testing, working on testing all Toyota dongles:

```python
Match type: original, matches: 685 dongles, non-matches: 73 dongles
Match type: toyota, matches: 751 dongles, non-matches: 7 dongles
Match type: toyota (excluded self FW from DB), matches: 472 dongles, non-matches: 286 dongles
```

adeebshihadeh:
> Common reason for not all dongle routes to fingerprint was a missing ECU (not fuzzy FP's issue).

How common is this? Why does it happen?

sshane:
I saved a list of these routes to check later. Haven't found a route with rlogs yet. One case the camera FW was missing, but it responded to tester present. Another case EPS was missing, but so were EPS messages causing a CAN error. Another case ABS was missing but it never responded to tester present, and the user could engage. All of these cases the normal fuzzy fingerprinting function matched.

This comment about ABS is suspicious:

https://github.com/commaai/openpilot/blob/fb89a2a6dd123077ebf6ecb250ec6848c4bc1e7e/selfdrive/car/toyota/values.py#L342-L346

sshane:
Looks like a few fwdCamera cases from TSS-P Camrys might be because we query too quickly after the previous query for the radar fails (since they're sub-addressed, the gateway or handling ECU might need a timeout). This usually isn't a problem since we wait on other ECUs within a request, but when querying sub-addressed ECUs they're the only ECU we're querying at a time so we can continue immediately.

Going to run a test on our Toyotas in a loop to see if I can reproduce.

sshane:
Ran a notebook on rlog data only:

```python
Checked routes: 6121, dongles: 973 platforms: {'TOYOTA RAV4 HYBRID 2019': 668, 'TOYOTA CAMRY HYBRID 2021': 402, 'TOYOTA COROLLA TSS2 2019': 532, 'LEXUS RX 2016': 96, 'TOYOTA HIGHLANDER HYBRID 2020': 312, 'TOYOTA HIGHLANDER 2020': 325, 'TOYOTA PRIUS 2017': 737, 'TOYOTA CAMRY HYBRID 2018': 88, 'TOYOTA RAV4 2019': 464, 'TOYOTA PRIUS TSS2 2021': 285, 'TOYOTA RAV4 HYBRID 2017': 115, 'TOYOTA COROLLA HYBRID TSS2 2019': 222, 'TOYOTA SIENNA 2018': 153, 'TOYOTA AVALON HYBRID 2022': 33, 'TOYOTA RAV4 2017': 172, 'TOYOTA COROLLA 2017': 174, 'TOYOTA HIGHLANDER HYBRID 2018': 89, 'TOYOTA RAV4 HYBRID 2022': 231, 'LEXUS IS 2018': 27, 'TOYOTA HIGHLANDER 2017': 84, 'TOYOTA C-HR HYBRID 2022': 7, 'TOYOTA MIRAI 2021': 34, 'TOYOTA CAMRY 2018': 236, 'LEXUS RX 2020': 64, 'TOYOTA AVALON HYBRID 2019': 19, 'LEXUS ES 2019': 52, 'TOYOTA RAV4 2022': 118, 'LEXUS NX 2020': 25, 'TOYOTA C-HR 2018': 16, 'TOYOTA CAMRY 2021': 92, 'TOYOTA AVALON 2019': 21, 'LEXUS ES HYBRID 2019': 66, 'TOYOTA RAV4 HYBRID 2023': 6, 'TOYOTA C-HR HYBRID 2018': 1, 'LEXUS RX HYBRID 2020': 27, 'TOYOTA AVALON 2016': 15, 'LEXUS RX HYBRID 2017': 51, 'LEXUS NX HYBRID 2018': 27, 'LEXUS NX 2018': 7, 'TOYOTA AVALON 2022': 18, 'TOYOTA ALPHARD 2020': 2, 'LEXUS RC 2020': 6, 'TOYOTA PRIUS v 2017': 1, 'LEXUS NX HYBRID 2020': 1}


With pandaStates errors on canState0/2:
Missing ECU routes: 40, dongles: 19

With out pandaStates errors:
Missing ECU routes: 2, dongles: 2
All missing ECU platform (routes): {'TOYOTA CAMRY 2018': 13, 'TOYOTA HIGHLANDER 2020': 18, 'TOYOTA RAV4 2022': 1, 'TOYOTA CAMRY 2021': 1, 'TOYOTA COROLLA TSS2 2019': 2, 'TOYOTA RAV4 2019': 1, 'TOYOTA COROLLA 2017': 1, 'TOYOTA HIGHLANDER 2017': 2, 'TOYOTA RAV4 HYBRID 2019': 1, 'TOYOTA CAMRY HYBRID 2018': 1, 'TOYOTA HIGHLANDER HYBRID 2018': 1}
```

sshane:
Last 10 FW version PRs are below. This PR would have matched 3/10 cases. 4/10 failed due to a counterpart hybrid platform that is too similar to match. And the final 3/10 failures were for legitimate reasons—2 unseen model years and 1 API difference.

Using the hybrid ECU to detect if it's a hybrid, it eliminates all 4 of those failures, bringing the matches up to 7/10.

| Failure/Success | Reason                                   | Pull Request                                                     |
|-----------------|------------------------------------------|-----------------------------------------------------------------------|
| Failure         | New model year                           | Lexus IS 2022 - https://github.com/commaai/openpilot/pull/30032       |
| Failure         | In exclude list for hybrid similarity   | Lexus NX 2020 - https://github.com/commaai/openpilot/pull/30031       |
| Success         |                                          | Camry Hybrid 2023 - https://github.com/commaai/openpilot/pull/29508   |
| Failure         | Hybrid similarity                        | C-HR Hybrid 2020 - https://github.com/commaai/openpilot/pull/29223    |
| Failure         | Hybrid similarity                        | Lexus RX Hybrid 2022 - https://github.com/commaai/openpilot/pull/29060 |
| Failure         | Japanese EPS version that supports LKA  | RAV4 Hybrid 2023 - https://github.com/commaai/openpilot/pull/28891    |
| Failure         | New model year                           | UX Hybrid 2023 - https://github.com/commaai/openpilot/pull/28766      |
| Success         | Recall                                   | Camry TSS-P - https://github.com/commaai/openpilot/pull/28679         |
| Failure         | Hybrid similarity                        | Lexus NX 2021 - https://github.com/commaai/openpilot/pull/28678       |
| Success         |                                          | Highlander TSS2 - https://github.com/commaai/openpilot/pull/28096     |

adeebshihadeh:
3/10 is a start, but we definitely have to do better. What's this hybrid similarity? Can we fix it?

sshane:
That's the ECUs that are shared that I talked about that will be fixed by using the hybrid ECU.

adeebshihadeh (approved):

sshane:
There are still a few minor API differences with hybrid/ICE. For example, only hybrid variant of RAV4 and Lexus ES are in `NO_STOP_TIMER_CAR`. Plus, the DSU messages dict has hybrid differences, specifically Highlander.

I believe that these differences likely aren't important and can still be merged, but I do not want to do too many things at once, so I'll remove the gas pedal logging and only merge platforms that I can from the fuzzy exclude list (which should be a lot, but not all)

adeebshihadeh:
I would like to look into that first before merging any of the hybrids. Otherwise, these differences will be difficult to find on new ports.

sshane:
The differences are only TSS-P, so there shouldn't be any confusion with new TSS2 models:
- TSS-P `stop_and_go`:
    It looks like certain [TSS-P hybrids have the ability to respond to acceleration commands below 19 mph](https://github.com/commaai/one/commit/b1b74054e949476e26afdec50f3167f00571f605#diff-c933b99dbed0471164bc0132e6c5dcb8d32ce3e32051cf1309d60c6474c1c332R148), despite the stock system not using the ability (our RAV4 ICE does not respond, but hybrid does). This mostly an undocumented feature/glitch, not totally related to hybrid/ICE as CHR ICE and [Highlander ICE](https://github.com/commaai/openpilot/pull/353) TSS-P can do it. RAV4H can do it, but not RAV4.
- TSS-P `NO_STOP_TIMER_CAR`:
    It looks like the meaning of this list changed over the years. [We used to enter standstill on all cars](https://github.com/commaai/one/blob/8a55a81cf77911af9cec3ac668de4de8d3c2be7d/selfdrive/car/toyota/carcontroller.py#L176-L186), and ignore the fact that the cruise state was in standstill on these cars which resumed regardless. [Now we stopped entering standstill altogether](https://github.com/commaai/openpilot/blob/7f945f10c65b445f630b970ddf4b245d917d5448/selfdrive/car/toyota/carcontroller.py#L117-L124) in https://github.com/commaai/openpilot/pull/2716.
- DSU static messages
    Certain platforms don't have both hybrid and ICE in the same lists, however I think that's a mistake. Take Lexus RX TSS-P for example (`f69ee4a1b79ac089|2023-09-27--14-36-02`). The ICE is missing from both msg 0x128 entries but it exists on the bus. Same with 0x2e6, 0x2e7, 0x33e, 0x365, 0x470. Likely a mistake.

So rule of thumb going forward for new TSS-P car ports: if it can do stop and go and the default doesn't match, make a new platform for ICE/hybrid. If the DSU messages don't fully match or don't work, make a new platform.

sshane:
Re-ran the notebook and still no false positives with excluded self FW, and no mismatches with online or stock fuzzy fingerprinting function.

Issue #27682, opened 2023-03-24
ui: crash when scrolling nav destinations


sshane:
While onroad, while a destination is set

sshane:
It actually hangs, not crash

```bash
[Switching to thread 1 (Thread 0x7fb0c31b70 (LWP 63984))]
#0  __lll_lock_wait (futex=futex@entry=0x16470f8, private=0) at lowlevellock.c:52
52      in lowlevellock.c
(gdb) bt
#0  __lll_lock_wait (futex=futex@entry=0x16470f8, private=0) at lowlevellock.c:52
#1  0x0000007fb7a52cec in __GI___pthread_mutex_lock (mutex=0x16470f8) at pthread_mutex_lock.c:115
#2  0x0000007faf6af880 in EglWaylandWlWindowSurface::DequeueBuffer() () from /lib/aarch64-linux-gnu/libeglSubDriverWayland.so
#3  0x0000007faf6af914 in EglWaylandWlWindowSurface::GetBuffer(EglSubResource*, EglMemoryDesc*) () from /lib/aarch64-linux-gnu/libeglSubDriverWayland.so
#4  0x0000007fb4f52e48 in ?? () from /lib/aarch64-linux-gnu/libGLESv2_adreno.so
#5  0x0000007fb4f54904 in ?? () from /lib/aarch64-linux-gnu/libGLESv2_adreno.so
#6  0x0000007fb4f0b030 in ?? () from /lib/aarch64-linux-gnu/libGLESv2_adreno.so
#7  0x0000007fb4f0a51c in ?? () from /lib/aarch64-linux-gnu/libGLESv2_adreno.so
#8  0x0000007fb4e69d1c in ?? () from /lib/aarch64-linux-gnu/libGLESv2_adreno.so
#9  0x0000007fb7204ca4 in QPlatformBackingStore::composeAndFlush(QWindow*, QRegion const&, QPoint const&, QPlatformTextureList*, bool) ()
   from /lib/aarch64-linux-gnu/libQt5Gui.so.5
#10 0x0000007fb7533430 in ?? () from /lib/aarch64-linux-gnu/libQt5Widgets.so.5
#11 0x0000007fb75350a4 in ?? () from /lib/aarch64-linux-gnu/libQt5Widgets.so.5
#12 0x0000007fb7536154 in ?? () from /lib/aarch64-linux-gnu/libQt5Widgets.so.5
[...]

PR #28642, opened 2023-06-22
ui/map_settings: remove repaint 

deanlee:
not sure if it's related to https://github.com/commaai/openpilot/issues/27682 

https://github.com/qt/qt/blob/0a2f2382541424726168804be2c90b91381608c6/src/gui/painting/qbackingstore.cpp#L381-L387

but there doesn't seem to be any need to call repaint here

### DIFF
--- a/selfdrive/ui/qt/maps/map_settings.cc
+++ b/selfdrive/ui/qt/maps/map_settings.cc
@@ -218,7 +218,6 @@ void MapSettings::refresh() {
   }
 
   destinations_layout->addStretch();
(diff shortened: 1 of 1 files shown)


pd0wm:
Wait so the random hangs are not a bug in `libeglSubDriverWayland.so` but something in Qt itself that's causing a deadlock?

deanlee:
no, I don't know exactly what the bug is, I'm just looking at the qt side to see if there are unnecessary and possibly problematic calls.

PR #28656, opened 2023-06-23
ui: ignore tap to wake device 

### LINKED ISSUE #28655: ui: ignore tap to wake device
Now that we have nav entry on home screen, you often press a random destination every time you tap to wake. @deanlee intetested in this?

deanlee:
resolve https://github.com/commaai/openpilot/issues/28655

### DIFF
--- a/selfdrive/ui/qt/window.cc
+++ b/selfdrive/ui/qt/window.cc
@@ -79,16 +79,20 @@ void MainWindow::closeSettings() {
 }
 
 bool MainWindow::eventFilter(QObject *obj, QEvent *event) {
+  bool ignore = false;
   switch (event->type()) {
     case QEvent::TouchBegin:
     case QEvent::TouchUpdate:
     case QEvent::TouchEnd:
     case QEvent::MouseButtonPress:
-    case QEvent::MouseMove:
+    case QEvent::MouseMove: {
+      // ignore events when device is awakened by resetInteractiveTimout
+      ignore = !uiState()->awake;
       device.resetInteractiveTimout();
       break;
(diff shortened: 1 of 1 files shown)


sshane:
What happens if you cause a `TouchBegin` while the screen is off, then release causing a `TouchEnd` while the screen is on? Would it click the button under it or does it need both `TouchBegin` and `TouchEnd`?

deanlee:
This is indeed a problem. Especially the handling of mouse press/release events.

sshane:
the sidebar uses manual touch events it looks like

deanlee:
reverted to the first vision: "ignore events before wake up", looks like it's the right solution.

> What happens if you cause a TouchBegin while the screen is off, then release causing a TouchEnd while the screen is on? Would it click the button under it or does it need both TouchBegin and TouchEnd?

After reading the qt source code, I'm sure that the widget will not receive  `TouchEnd` or `MouseButtonRelease` events if `TouchBegin`/`MouseButtonPressed` is filtered.


[qt_button_down](https://codebrowser.dev/qt5/qtbase/src/widgets/kernel/qwidgetwindow.cpp.html#qt_button_down) will be nullptr if `TouchBegin`/`MouseButtonPressed` is filtered. this will prevent subsequent touch/mouse events from being sent to any window:
https://codebrowser.dev/qt5/qtbase/src/widgets/kernel/qwidgetwindow.cpp.html#_ZN13QWidgetWindow16handleMouseEventEP11QMouseEvent

sshane:
works on PC, thanks!

PR #28665, opened 2023-06-24
UI: Indicate when Nav on OP is enabled

mitchellgoffpc:


selfdrive/ui/qt/onroad.cc
```diff
@@ -64,9 +64,20 @@ void OnroadWindow::updateState(const UIState &s) {
...
+                        !(*s.sm)["modelV2"].getModelV2().getNavEnabled();
+  if (navDisabled != navDisabledNow) {
+    split->setSpacing(navDisabledNow ? bdr_s * 2 : 0);
+    if (map) {
+      map->setFixedWidth(topWidget(this)->width() / 2 - bdr_s * (navDisabledNow ? 2 : 1));
+    }
+  }
+
+  // repaint border
+  if (bg != bgColor || navDisabled != navDisabledNow) {
```
mitchellgoffpc, line 78: @incognitojam I think we may also have the repaint the background after an alert

incognitojam: I couldn't reproduce any issues with alerts, and I think this would only preserve the current behaviour?

mitchellgoffpc: Nvm, I had an issue before where the alerts seemed to leave a "ghost" image on the border if I didn't repaint, but I can't reproduce now.

PR #28669, opened 2023-06-24
car: use default_factory for mutable fields in dataclasses

maxime-desroches:
This is a prerequisite to switch to python 3.11:
https://docs.python.org/3/whatsnew/3.11.html#dataclasses

### DIFF
--- a/selfdrive/car/chrysler/values.py
+++ b/selfdrive/car/chrysler/values.py
@@ -1,5 +1,5 @@
 from enum import IntFlag
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 from typing import Dict, List, Optional, Union
 
 from cereal import car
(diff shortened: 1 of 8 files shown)


selfdrive/car/chrysler/values.py
```diff
@@ -60,7 +60,7 @@ def __init__(self, CP):
 @dataclass
 class ChryslerCarInfo(CarInfo):
   package: str = "Adaptive Cruise Control (ACC)"
-  car_parts: CarParts = CarParts.common([CarHarness.fca])
+  car_parts: CarParts = default_mutable(CarParts.common([CarHarness.fca]))
```
sshane, line 63: Does it make sense to make CarParts callable like?

```python
  def __call__(self):
    return self
```

PR #28684, opened 2023-06-24
GM ASCM: gate missing radar behind OP long toggle

nworb-cire:


selfdrive/car/gm/interface.py
```diff
@@ -123,7 +127,7 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
     # These cars likely still work fine. Once a user confirms each car works and a test route is
     # added to selfdrive/car/tests/routes.py, we can remove it from this list.
     ret.dashcamOnly = candidate in {CAR.CADILLAC_ATS, CAR.HOLDEN_ASTRA, CAR.MALIBU, CAR.BUICK_REGAL, CAR.EQUINOX} or \
-                      (ret.networkLocation == NetworkLocation.gateway and ret.radarUnavailable)
+                      (ret.networkLocation == NetworkLocation.gateway and ret.radarUnavailable and not experimental_long)
```
adeebshihadeh, line 130: why was this changed?

nworb-cire: When the radarless setup was initially implemented, there was a bit of debate on whether it should be gated behind the OP long toggle or the dashcamOverride param. The override param has since been removed entirely, so this allows the setup to still run out of dashcam mode when experimental long is enabled.

sshane: I don't think we want to remove the car from dashcam mode if the toggle is set, close @adeebshihadeh?

adeebshihadeh: yeah, that's right

nworb-cire: Currently, the logic operates in such a way that GM ASCM platform vehicles are entirely barred from use with openpilot if radar tracks are unavailable. This does not align with how other makes handle absence of radar tracks, where vision-only longitudinal is gated behind the OP long toggle. This PR would cause all makes to treat OP long identically in this respect.

I also personally would like to be able to run stock and submit bug reports without being grilled about dirty git 🥲

sshane: The difference here is that we don't want to natively support cars without an ASCM, where the car's ECUs would fault if openpilot is not sending on the bus.

Perhaps we should switch from detecting the radar to detecting ASCM, but in what case does an ASCM car not have a radar? If it has LKAS only? Then you need to reflash ECUs, which means we don't want to support it in release just yet still.

sshane: In other brands with missing radar tracks, there exists a radar and we don't parse it, or it is vision only. That is the difference.

sshane:
Closing for now.

Issue #25382, opened 2022-08-07
Toyota: `NO_STOP_TIMER_CAR` hybrids consume power when stopped


cydia2020:
### Describe the bug

It has come to my attention thanks to a Discord user's tip-off that Toyota `NO_STOP_TIMER_CAR` hybrids may unnecessarily waste power when the vehicle is at a standstill with openpilot longitudinal engaged. The user reported the car uses around 400 watts of extra power when stopped with openpilot engaged, but only wastes around 100 watts with the stock system under the same conditions.

I believe this is due to openpilot's implementation of the `NO_STOP_TIMER_CAR` logic, where some hybrid cars won't go into standstill when *openpilot longitudinal is stopping behind a lead vehicle* (this issue does not apply when the user is manually stopping behind a lead car and enabling openpilot after they have came to a complete stop).

I have been procrastinating to report this issue due to lack of access to proper equipment, my Chinese clone of the XHorse cable was unstable at best, it recently kicked the bucket, so I have decided to purchase an OBDLink LX. With that piece of equipment in hand, I have been able to personally verify this with my MY2018 Prius.

I will try to conduct some tests on this issue, and document my findings here. Please, if you have access to the proper hardware, test this on your vehicle as well, the more data the better.

### Which car does this affect?

Presumably all toyota hybrids marked as no stop timer car

### Provide a route where the issue occurs

NA

### openpilot version

all

cydia2020:
Test 1:

Testing procedure:

Please note, this test may be unreliable, but it gives us some insights as to what is going on. I will conduct more tests in the coming days with better logging.

Data logged:

Motor (MG2) Current (A)
Motor (MG2) RPM
HV Battery Voltage (V)
HV Battery Current (A)

The test:

I engaged openpilot with the vehicle stopped shortly after logging started on the OBDLink app, I waited a few seconds for the graphs to stabilise, and looked at the data recorded by the OBDLink app.

I then slightly stepped on the accelerator pedal, enough to get the vehicle moving, but not sufficient for the speed sensor to register, I observed the data again.

Findings:

![image](https://user-images.githubusercontent.com/12470297/183293883-375c7200-6a0f-43c4-86e5-30992752f038.png)
Graph 1 - the vehicle is in standstill, ready to be resumed with a tap of the accelerator or the resume button, around 1 amp of current is flowing from the HV battery to the rest of the vehicle. The HV Battery’s voltage sits at roughly 220V, and is stable.

[...]

PR #28689, opened 2023-06-25
Toyota: skip entering standstill with smartDSU

sunnyhaibin:
Models that have smartDSU installed and detected do not need to set `standstill_req` when in standstill state. Currently, platforms such as `TOYOTA PRIUS 2017` with smartDSU installed (emits `0x2FF`) requires manual user input to resume from standstill without this change.

image

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -6,7 +6,7 @@ from selfdrive.car.toyota.toyotacan import create_steer_command, create_ui_comma
                                            create_fcw_command, create_lta_steer_command
 from selfdrive.car.toyota.values import CAR, STATIC_DSU_MSGS, NO_STOP_TIMER_CAR, TSS2_CAR, \
                                         MIN_ACC_SPEED, PEDAL_TRANSITION, CarControllerParams, \
-                                        UNSUPPORTED_DSU_CAR
+                                        UNSUPPORTED_DSU_CAR, ToyotaFlags
 from opendbc.can.packer import CANPacker
 
 VisualAlert = car.CarControl.HUDControl.VisualAlert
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Why is the smartDSU relevant here?

wired4sound:
> Why is the smartDSU relevant here?

We typically commented out a few lines in code or forks like Stock Additions and DragonPilot would have a special SnG toggle for the user to turn on if they had certain Toyotas.   Without the toggle on in certain forks or commenting out a few lines of code in vanilla, then any time the car comes to a complete stop, you have to hit resume in order for openpilot to continue its operations.

Is this intended behavior even though comma has SDSU detection?

wired4sound:
> > Why is the smartDSU relevant here?
> 
> We typically commented out a few lines in code or forks like Stock Additions and DragonPilot would have a special SnG toggle for the user to turn on if they had certain Toyotas. Without the toggle on in certain forks or commenting out a few lines of code in vanilla, then any time the car comes to a complete stop, you have to hit resume in order for openpilot to continue its operations.
> 
> Is this intended behavior even though comma has SDSU detection?

Here is the two lines I was referencing to:

https://github.com/commaai/openpilot/issues/25382#issuecomment-1208041401

wired4sound:
Did this get reverted?

https://github.com/commaai/openpilot/pull/26671

wired4sound:
https://github.com/commaai/openpilot/pull/26709

sshane:
I think Adeeb is suggesting this is a Prius-specific change disguised as an SDSU-specific change. We have https://github.com/commaai/openpilot/pull/26721 which covers this case. The problem is that we haven't verified that the car doesn't roll back on hills when trying to resume, so we put it in standstill state to prevent that.

However, very similar cars are already in the `NO_STOP_TIMER_CAR`, so I don't know why they would be different, that decision was made a long time ago. Once we verify similar behavior to TSS2 on hills with our Prius we will merge that PR above.

PR #28691, opened 2023-06-25
Added optional sw-render flag to UI

ljkeller:
WSL can have render issues without the LIBGL_ALWAYS_SOFTWARE=1 variable.

This new --sw-render flag is a simple way for WSL based developers to run the UI without additions to the FAQ about this issue.

History: https://github.com/commaai/openpilot/discussions/26091#discussioncomment-4258846

### DIFF
--- a/system/loggerd/loggerd.cc
+++ b/system/loggerd/loggerd.cc
@@ -58,7 +58,7 @@ struct RemoteEncoder {
   bool seen_first_packet = false;
 };
 
-int handle_encoder_msg(LoggerdState *s, Message *msg, std::string &name, struct RemoteEncoder &re, EncoderInfo encoder_info) {
+int handle_encoder_msg(LoggerdState *s, Message *msg, std::string &name, struct RemoteEncoder &re, const EncoderInfo &encoder_info) {
   int bytes_count = 0;
 
   // extract the message


adeebshihadeh:
Thanks for the PR! A few things about this:
* the `ui` bash script wrapper should really be removed, so I don't want to add to it. (it was just needed on the comma two with NEOS since it didn't support `rpath`)
* can we detect when this is necessary and apply automatically?
* is this a bug in WSL or our UI?

ljkeller:
> Thanks for the PR! A few things about this:
> 
> * the `ui` bash script wrapper should really be removed, so I don't want to add to it. (it was just needed on the comma two with NEOS since it didn't support `rpath`)
> * can we detect when this is necessary and apply automatically?
> * is this a bug in WSL or our UI?

Of course! I want to get some contributions in, but bash is not my strength.

1. I understand not wanting to add a tumor. **Then I think the next best option is probably a section in the developer FAQ/WIKI**.
2. There is no perfect way to auto-detect this render issue AFAIK. I could auto-detect WSL (imperfectly, but pretty well), but I don't think _all_ WSL users see this bug.
3. The render issue appears to be WSL side. Here are some examples with different drivers: (https://github.com/microsoft/wslg/issues/799 , https://github.com/microsoft/wslg/issues/317 , https://github.com/microsoft/wslg/issues/469) 

Do you have any suggestions? I know I'm not the only one to see this issue, and I was segfaulting selfdrive/ui/ui until I used LIBGL_ALWAYS_SOFTWARE=1. I just want to make sure this is tracked somewhere. There is two okay spots I see to put it in wiki: https://github.com/ljkeller/openpilot/tree/master/tools under "Windows" or in the developer wiki somewhere.

Infodump for future devs:

- I'm running WSL 2 Ubuntu 20.04.06 & have a Nvidia 2080 Ti on my windows machine.
- `selfdrive/ui/ui` does not seg fault until it starts getting data from running `tools/replay/replay '0c7f0c7f0c7f0c7f|2021-10-13--13-00-00' --dcam --ecam`. In other words, the GUI is running until it switches to the road view.

adeebshihadeh:
Let's document it in the Windows section of the tools README for now. If we can detect this reliably in the future, it would be nice to set automatically.

PR #28693, opened 2023-06-25
camerad: remove unused frame_length from `struct FrameMetadata`

deanlee:


adeebshihadeh:
Can you also deprecate the field in cereal?

PR #28696, opened 2023-06-25
Document WSL bug workaround

ljkeller:
WSL developers may run into an issue while running GUIs, where GUIs will segfault or crash due to hardware rendering issues.

Use LIBGL_ALWAYS_SOFTWARE=1 if you see segfaults or other strange GUI failures even after setting up a fresh WSL.

History: https://github.com/commaai/openpilot/discussions/26091#discussioncomment-4258846 .

Followed @adeebshihadeh advice to put this in README and not UI wrapper from (https://github.com/commaai/openpilot/pull/28691)


TheSecurityDev:
It seems this fix is no longer required. The link is no longer valid, but here's the archive for documentation purposes. https://web.archive.org/web/20231113143431/https://github.com/commaai/openpilot/discussions/26091#discussioncomment-4258428

PR #28699, opened 2023-06-26
GM ASCM: don't apply gas when stopping

### LINKED ISSUE #28680: GM sends nonzero gas while preEnabled

sshane:
Fixes a GM cruise fault on some ASCM vehicles where we apply gas under the "regen accel" while pre-enabled. Note that the Acadia is unaffected, likely only a Volt fault.

An Acadia:

image

A Volt (with fault):

image

I couldn't find many ASCM rlogs where the user enables at a stop (we don't have a lot of GM segments)


sshane:
Real fix is https://github.com/commaai/openpilot/pull/25453, it improves tuning, brakes immediately when stopping, or pre-enabling

PR #28701, opened 2023-06-26
ui: fix memory leak in `MapSettings`

deanlee:


selfdrive/ui/qt/maps/map_settings.cc
```diff
@@ -130,12 +130,12 @@ void MapSettings::updateCurrentRoute() {
       qWarning() << "JSON Parse failed on NavDestination" << dest;
       return;
     }
-    auto destination = new NavDestination(doc.object());
+    auto destination = std::make_unique<NavDestination>(doc.object());
```
ljkeller, line 133: I see the memory leak here in `updateCurrentRoute` and also with the vector in `refresh`; that's a good change.

What about the other allocations happening in this file relating to buttons/layouts? Like in the constructor `auto *frame = new QVBoxLayout(this);` for example, `frame`  is allocated and used, but the memory doesn't seem to be managed. I'm new to this repo and it's dependencies so I'm probably missing something obvious.

ljkeller: QT manages memory with its own mechanisms as long as you play by the rules https://doc.qt.io/qt-6/objecttrees.html so the other allocations of QT stuff are probably good.

NVM

deanlee: Yes, qt will take care of these memory releases

PR #28703, opened 2023-06-26
GM: Resolve preenable cruise fault

### LINKED ISSUE #28680: GM sends nonzero gas while preEnabled

nworb-cire:
Proof of concept currently. Allows pre-enable and removal of user brake into autohold brake mode, but faults once the gas command is sent. Still need to figure that out. 

Should resolve https://github.com/commaai/openpilot/issues/28680 as well, since that seems to be one of the many causes of this fault.

### DIFF
--- a/selfdrive/car/gm/gmcan.py
+++ b/selfdrive/car/gm/gmcan.py
@@ -72,23 +72,7 @@ def create_gas_regen_command(packer, bus, throttle, idx, enabled, at_full_stop):
   return packer.make_can_msg("ASCMGasRegenCmd", bus, values)
 
 
-def create_friction_brake_command(packer, bus, apply_brake, idx, enabled, near_stop, at_full_stop, CP):
-  mode = 0x1
-
-  # TODO: Understand this better. Volts and ICE Camera ACC cars are 0x1 when enabled with no brake
-  if enabled and CP.carFingerprint in (CAR.BOLT_EUV,):
-    mode = 0x9
-
-  if apply_brake > 0:
-    mode = 0xa
-    if at_full_stop:
-      mode = 0xd
-
-    # TODO: this is to have GM bringing the car to complete stop,
-    # but currently it conflicts with OP controls, so turned off. Not set by all cars
-    #elif near_stop:
-    #  mode = 0xb
-
+def create_friction_brake_command(packer, bus, apply_brake, idx, enabled, near_stop, at_full_stop, CP, mode):
   brake = (0x1000 - apply_brake) & 0xfff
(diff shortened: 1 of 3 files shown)


selfdrive/car/gm/gmcan.py
```diff
@@ -81,7 +81,9 @@ def create_friction_brake_command(packer, bus, apply_brake, idx, enabled, near_s
 
   if apply_brake > 0:
     mode = 0xa
-    if at_full_stop:
+    if pre_enabled:
+      mode = 0xb
+    elif at_full_stop:
       mode = 0xd
```
sshane, line 87: Can you try to send brake without making changes to the mode?

nworb-cire: I don't have an ASCM/stock ACC, but I could probably ask one of the other guys to record a route.

0xb by itself will allow pre-enable, then holds brakes for roughly half a second after user brake is released. After that, the system faults and releases the brake hold.

0xd by itself causes a fault while user brake is pressed IIRC.

sshane: I'm surprised this isn't more widely reported then, are you sure it's not due to your car being non-stock? @Verylukyguy can you enable at a stop with your foot on the brake? What happens when you do so?

Verylukyguy: > 

I can enable at a stop with my foot on the brake.  I can provide a test route, but not until later tonight.

sshane:
Not being able to auto-resume shouldn't be related to this fault, is it?

nworb-cire:
I suppose they are similar issues-- both in preenable and at ACC standstill while already enabled, we would like the car to respond to ACC gas commands, but it either faults or does not respond.

They are not the same because there appear to be three subtly different cruise faults:
1. a cruise fault occurs when nonzero gas is sent while user brake is pressed (https://github.com/commaai/openpilot/issues/28680)
2. a cruise fault occurs when enabling/re-enabling at a perfect standstill (that's what this PR aims to address). No cruise fault occurs if vEgo >= some epsilon > 0
3. a cruise fault occurs when ACC is already enabled, at ACC standstill, and we send a gas command. Some preliminary work has been done to spam resume, but with no results as of yet. (We know we can get the ECU to respond to resume spamming in some cases, see OPGM CC long, aka redneck ACC)

Cases 2 and 3 are slightly different faults because of the vehicle's internal ACC state. In case 3, we can press resume to begin sending gas commands. In case 2, we cannot.

nworb-cire:
Closing pending further investigation.

PR #28704, opened 2023-06-26
encoder: reduce memory allocations and copying

deanlee:
reduce two allocations and one copying for video data by using `kj::ArrayOutputStream`, like what we did in `can_list_to_can_capnp_cpp`:

https://github.com/commaai/openpilot/blob/070bdcdf89674e0086c93afbc9453cde42f1ba60/selfdrive/boardd/can_list_to_can_capnp.cc#L19-L22


adeebshihadeh:
Can you rebase?

deanlee:
rebased

PR #28709, opened 2023-06-26
car docs: use experimental long limits

sshane:
On some platforms (namely GM), we are able to allow lower enable speeds when using openpilot longitudinal. This passes experimental_long to the car interfaces which currently does not change any limits (until https://github.com/commaai/openpilot/pull/28708 is merged).


sshane:
@adeebshihadeh are you fine with showing the best limits possible/when using openpilot longitudinal? When using stock ACC, the camera still restricts to min 5 kph or stopped here

PR #26722, opened 2022-12-08
Toyota: LTA steering control

sshane:
Opened a PR to preserve history, old branch got deleted

New 2023 RAV4s only use LTA to steer and are unresponsive to LKA requests.

PR #28710, opened 2023-06-26
Toyota: simple Lane Tracing Assist (LTA) support

sshane:
This PR aims to support most of LTA (still behind dashcam), reducing diff of https://github.com/commaai/openpilot/pull/26722

There is still some exploration to do with the signals, they will not be changed with this PR


sshane:
this now doesn't do anything not experimental and not in toyota-lta branch, closing

PR #28711, opened 2023-06-27
Toyota: check Lane Tracing Assist (LTA) EPS faults

sshane:
Splitting from https://github.com/commaai/openpilot/pull/26722

If using LKA, no change (don't check LTA, probably fine but don't want to make a big change in this PR).

If using LTA, it checks both LKA and LTA for faults.

### DIFF
--- a/selfdrive/car/toyota/carstate.py
+++ b/selfdrive/car/toyota/carstate.py
@@ -12,7 +12,7 @@ from selfdrive.car.toyota.values import ToyotaFlags, CAR, DBC, STEER_THRESHOLD, 
 
 SteerControlType = car.CarParams.SteerControlType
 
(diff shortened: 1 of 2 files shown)


sshane:
It's quite rare, but in a few segments out of 140k in the last 3 months, sometimes LTA_STATE doesn't leave 0 (uninitialized). These are not LTA cars (Corolla, RAV4, Prius, etc.), and we don't know how it reacts when requesting LTA, so leaving this alone for now. Just a note.

image

sshane:
Ah I see what it is! LTA needs the more accurate angle sensor signal from STEER_TORQUE_SENSOR to work properly. I can no idea it could take this long to initialize. I'll add a lockout similar to Ford (only if LTA car).

image

sshane:
Code 21 is also possible

image

PR #28717, opened 2023-06-27
Toyota LTA: show alert when steering angle sensor is initializing

sshane:
Add lockout condition found in https://github.com/commaai/openpilot/pull/28711 for LTA. Basically, if we're trying to command with LTA and `STEER_TORQUE_SENSOR->STEER_ANGLE_INITIALIZING`=1, `EPS_STATUS->LTA_STATE` will be 0 (not available). Show an alert instead of a fault as this is expected.

Should check data to ensure this captures every case where LTA_STATE is 0

### DIFF
--- a/selfdrive/car/toyota/interface.py
+++ b/selfdrive/car/toyota/interface.py
@@ -8,6 +8,7 @@ from selfdrive.car import STD_CARGO_KG, scale_tire_stiffness, get_safety_config
 from selfdrive.car.interfaces import CarInterfaceBase
 
 EventName = car.CarEvent.EventName
+SteerControlType = car.CarParams.SteerControlType
 
 
 class CarInterface(CarInterfaceBase):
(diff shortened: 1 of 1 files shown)


sshane:
For some reason, there are cases where `EPS_STATUS->LTA_STATE` is unexpectedly 0 on startup, but it's only RAV4s of various years, and it's very rare.

There are only 5 segments out of 260k last 180 days from 2019 and 2022 RAV4 platforms that are triggered by this:

```python
for r in results:
  segment, platform, ... = r

  lta_not_ready_unexpected_cnt = 0
  lta_not_ready_expected_cnt = 0
  for lta, init in zip(lta_state, steer_angle_init):
    if lta == 0 and init == 0:
      lta_not_ready_unexpected_cnt += 1
    else:
      lta_not_ready_expected_cnt += 1

  if platform in TSS2_CAR:
    # LTA can take 10-20 frames to initialize, even with angle initialized
    # this usually happens before carState starts logging
    if lta_not_ready_unexpected_cnt > 20:
      print('unexpected lta=0', platform, segment, lta_not_ready_unexpected_cnt)
```

Typical startup:

image

Bad startup (one of the 5):

image

PR #28728, opened 2023-06-28
Tge fingerprint

chzollib:
Add FW for MY2023 MAN TGE


sshane:
again, this needs to be made against master

PR #28729, opened 2023-06-28
Update values.py with MY23 TGE/Crafter

chzollib:
Add new FW-Versions for MY23 Crafter/TGE


selfdrive/car/volkswagen/values.py
```diff
@@ -392,18 +392,24 @@ def init_make(self, CP: car.CarParams):
...
+      b'\xf1\x875QF959655AS\xf1\x890755\xf1\x82\x1315140015150011111100050200--1311120749',
     ],
     (Ecu.eps, 0x712, None): [
       b'\xf1\x872N0909143E \xf1\x897021\xf1\x82\x05163AZ306A2',
+      b'\xf1\x872N0909144K \xf1\x897045\xf1\x82\x05233AZ810A2',
     ],
     (Ecu.fwdRadar, 0x757, None): [
       b'\xf1\x872Q0907572M \xf1\x890233',
+      b'\xf1\x872Q0907572R \xf1\x890372',
+      b'\xf1\x872Q0907572AA \xf1\x890396',
```
jyoung8607, line 412: There's an extra space here that's probably a manual transcription error. Also, the AA one should be sorted at the top. It's chronologically wrong but alphabetically right, and what we're doing elsewhere for consistency.

```suggestion
      b'\xf1\x872Q0907572AA\xf1\x890396',
```

jyoung8607 (commented):
Can you share the routes you got these firmware versions from? Looks like you're working with at least two vehicles. comma will want those for verification and traceability.

Out of curiosity, are these the EV conversions that were discussed on Discord recently?

chzollib:
It was a stock vehicle with long wheelbase. 
The route is here: af1bf21a07172228|2023-06-28--21-00-49--0

Some FW-IDs were from converted vehicles, where I haven't gotten ACC/OP to work. Should I exclude those for now?

sshane:
you should make your pull request against master, not master-ci (which is auto-generated)

PR #28731, opened 2023-06-28
Toyota: refactor carcontroller for LTA

sshane:
prep


sshane:
cleaner/better in https://github.com/commaai/openpilot/pull/28735

PR #28734, opened 2023-06-28
Toyota: add steering limits and documentation

sshane:


sshane:
no need to do separately -> https://github.com/commaai/openpilot/pull/28735

PR #28735, opened 2023-06-29
Toyota: set and limit angle for LTA control

sshane:
v2 of https://github.com/commaai/openpilot/pull/28731

we always need to send LKA, and we always need to send LTA (if TSS2), so it doesn't make sense to have multiple statements to do this

does not apply fault work-arounds yet to reduce complexity (they operate at different rates)

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -70,7 +70,7 @@ class CarController:
       self.steer_rate_counter = 0
 
     # *** steer angle ***
-    if self.CP.steerControlType == car.CarParams.SteerControlType.angle:
+    if self.CP.steerControlType == SteerControlType.angle:
       # If using LTA control, disable LKA and set steering angle command
       apply_steer = 0
       apply_steer_req = 0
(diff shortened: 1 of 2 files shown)


selfdrive/car/toyota/carcontroller.py
```diff
@@ -1,6 +1,7 @@
 from cereal import car
 from common.numpy_fast import clip, interp
-from selfdrive.car import apply_meas_steer_torque_limits, create_gas_interceptor_command, make_can_msg
+from selfdrive.car import apply_std_steer_angle_limits, apply_meas_steer_torque_limits, \
```
sshane, line 3:
```suggestion
from selfdrive.car import apply_meas_steer_torque_limits, apply_std_steer_angle_limits, \
```

selfdrive/car/toyota/carcontroller.py
```diff
@@ -61,10 +69,22 @@ def update(self, CC, CS, now_nanos):
...
 
-    # Never actuate with LKA on cars that only support LTA
-    if self.CP.steerControlType == car.CarParams.SteerControlType.angle:
+    # *** steer angle ***
+    if self.CP.steerControlType == SteerControlType.angle:
+      # If using LTA control, disable LKA and set steering angle command
       apply_steer = 0
       apply_steer_req = 0
+      if self.frame % 2 == 0:
+        # EPS uses the torque sensor angle to control with, offset to compensate
```
sshane, line 78:
```suggestion
        # EPS uses the torque sensor angle to control against, offset to compensate
```

PR #28740, opened 2023-06-29
IsoTpParallelQuery: support request callbacks

### LINKED ISSUE #28672: FPv2: support request callbacks
You should be able to specify a function inside your request that takes in the previous responses as input and generates a new query.

For example, we query Toyota ECUs with `b'\x1a\x88\x01'` to get their FW versions, but the actual Toyota Techstream diagnostic software does things a little differently:

- It first queries `b'\x1a\x88\x00`
- Then gets a response like `b'\x1a\x88\x00\x01'`
- Then it uses that payload (`b'\x01'`) in the next response: `b'\x1a\x88\x01'` to get the FW version.

My guess is that some Toyota platforms have their FW version on a different data id and Techstream gets it by using this query back-and-forth.

Edit: this is a correct assumption, except the response can be any length and is a list describing the ids to get the full FW version. `b'\x1a\x88\x00\x81\x82` means you need to query both 0x81 and 0x82 suffixes to get the full thing

```python
def get_toyota_version_request_kwp(responses: List[bytes]) -> bytes:
  return b'\x1a\x88' + bytes([responses[-1][-1]])

def get_toyota_version_response_kwp(responses: List[bytes]) -> bytes:
  return b'\x5a\x88' + bytes([responses[-1][-1]])

FW_QUERY_CONFIG = FwQueryConfig(
  requests=[
    Request(
      [b'\x1a\x88\x00', get_toyota_version_request_kwp],
      [b'\x5a\x88\x00', get_toyota_version_response_kwp],
      bus=0,
    ),
)
```

sshane:
Configs can define requests with callbacks that take in the last response and output a list of new requests

edit: it works!

not replacing current query, since for some ECUs we will get more frames of info back (`\x01\x02` are the supported ids for fwdCamera) and thus exact matching won't work. can remove old query after data gathering

### DIFF
--- a/selfdrive/car/isotp_parallel_query.py
+++ b/selfdrive/car/isotp_parallel_query.py
@@ -126,8 +126,9 @@ class IsoTpParallelQuery:
           if counter + 1 < len(self.request):
             # If request callback defined, replace it with its results
             if callable(self.request[counter + 1]):  # and self.response[counter + 1]:
-              self.request[counter + 1:] = self.request[counter + 1](dat[len(expected_response):])
-              self.response[counter + 1:] = self.response[counter + 1](dat[len(expected_response):])
+              response = dat[len(expected_response):]
+              self.request[counter + 1:] = self.request[counter + 1](response)
+              self.response[counter + 1:] = self.response[counter + 1](response)
 
             response_timeouts[tx_addr] = time.monotonic() + timeout
             msg.send(self.request[counter + 1])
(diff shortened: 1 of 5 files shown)


sshane:
@gregjhogan what do you think of something like this? Should this be handled lower down, or is this not in any KWP2000 standard you've seen?

selfdrive/car/isotp_parallel_query.py
```diff
@@ -123,6 +123,14 @@ def get_data(self, timeout, total_timeout=60.):
         response_valid = dat.startswith(expected_response)
 
         if response_valid:
+          if counter + 1 < len(self.request):
+            # If request callback defined, replace it with its results
+            if callable(self.request[counter + 1]):  # and self.response[counter + 1]:
```
sshane, line 128:
```suggestion
            if callable(self.request[counter + 1]):
```

sshane:
Will re-open when ready

PR #28745, opened 2023-06-29
ui: remove Q_PROPERTY from `AnnotatedCameraWidget`

deanlee:
Remove Q_PROPERTY from `AnnotatedCameraWidget`, it just adds extra call overhead.


deanlee:
> @deanlee Hi, do you have plans on also removing Q_PROPERTY from `Sidebar`? I noticed that every time the sidebar is displayed via screen tap while onroad, the memory would bump up ~1% each time.

No such plan. Unlike `AnnotatedCameraWidget`, `Q_PROPERTY` in sidebar is used to ensure that the interface is redrawn only when the value changes. it will not cause an increase in memory.
If the memory is changed after the sidebar is shown or hidden. it could only be due to some cache/texture rebuild and invalidation caused by the resizing of onroad widgets

PR #28751, opened 2023-06-30
Revert "Nissan: add counter signal to forwarded message"

sshane:
Reverts commaai/openpilot#28336

Just wondering if fuzzy tests catch this!


sshane:
nice @maxime-desroches 🎉

maxime-desroches:
one weird trick to get rid of all your car interface bugs!

PR #28752, opened 2023-06-30
Lexus: add missing FW for UX Hybrid 2023

sshane:
waiting for a tester


sshane:
just to get the FW somewhere, closing for now

PR #28753, opened 2023-06-30
ui/settings: allow driving personality without CarParams

### LINKED ISSUE #28750: ui: allow driving personality without CarParams
Noticed that if you don't have a CarParams, we allow Experimental Mode but disable Driving Personality buttons. We should allow both if we don't have a CarParams saying we shouldn't

@deanlee want to clean up and de-duplicate the code to fix this at the same time?

deanlee:
resolve https://github.com/commaai/openpilot/issues/28750

### DIFF
--- a/selfdrive/ui/qt/util.cc
+++ b/selfdrive/ui/qt/util.cc
@@ -257,7 +257,7 @@ QPixmap bootstrapPixmap(const QString &id) {
   return pixmap;
 }
 
-bool hasExperimentalLongitudinal(const cereal::CarParams::Reader &car_params) {
+bool hasLongitudinalControl(const cereal::CarParams::Reader &car_params) {
   return car_params.getExperimentalLongitudinalAvailable()
              ? Params().getBool("ExperimentalLongitudinalEnabled")
              : car_params.getOpenpilotLongitudinalControl();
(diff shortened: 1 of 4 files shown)


selfdrive/ui/qt/util.cc
```diff
@@ -256,3 +256,9 @@ QPixmap bootstrapPixmap(const QString &id) {
   }
   return pixmap;
 }
+
+bool hasExperimentalLongitudinal(const cereal::CarParams::Reader &car_params) {
```
sshane, line 260:
```suggestion
bool hasLongitudinalControl(const cereal::CarParams::Reader &car_params) {
```

this function is to return if openpilot will control longitudinal without needing to restart to get a new carparams, think this name is more appropriate

selfdrive/ui/qt/util.cc
```diff
@@ -256,3 +256,9 @@ QPixmap bootstrapPixmap(const QString &id) {
   }
   return pixmap;
 }
+
+bool hasLongitudinalControl(const cereal::CarParams::Reader &car_params) {
+  return car_params.getExperimentalLongitudinalAvailable()
+             ? Params().getBool("ExperimentalLongitudinalEnabled")
+             : car_params.getOpenpilotLongitudinalControl();
+}
```
sshane, line 264:
```suggestion
bool hasLongitudinalControl(const cereal::CarParams::Reader &car_params) {
  // Using the experimental longitudinal toggle, returns whether longitudinal control
  // will be active without needing a restart of openpilot
  return car_params.getExperimentalLongitudinalAvailable()
             ? Params().getBool("ExperimentalLongitudinalEnabled")
             : car_params.getOpenpilotLongitudinalControl();
}
```

PR #28758, opened 2023-06-30
Added new 2023 Ram 1500 FW

vincentw56:
**Description** Added new 2023 Ram 1500

**Verification** dongle id: 960a6270ac410199


selfdrive/car/chrysler/values.py
```diff
@@ -228,6 +228,7 @@ class ChryslerCarInfo(CarInfo):
       b'68510283AG',
       b'68527346AE',
       b'68527375AD',
+      b'68510280AG',
```
sshane, line 231:
```suggestion
```

PR #28759, opened 2023-06-30
UI: fix memory leak in MapSettings

deanlee:
Resubmitted, fixed the issue : https://github.com/commaai/openpilot/commit/3e1ba3ff3522bb2eb8a8f41ecb1158314af6c3dc#commitcomment-120290398


selfdrive/ui/qt/maps/map_settings.h
```diff
@@ -1,5 +1,7 @@
 #pragma once
 
+#include<memory>
```
incognitojam, line 3:
```suggestion
#include <memory>
```

PR #28760, opened 2023-06-30
Hyundai CAN FD: remove undefined signal

sshane:
Related to https://github.com/commaai/openpilot/pull/28604 and https://github.com/commaai/opendbc/issues/869


sshane:
added test in https://github.com/commaai/openpilot/pull/28763

PR #28763, opened 2023-06-30
CAN Parser: assert no missing signals

sshane:
bumping opendbc for https://github.com/commaai/opendbc/pull/870


sshane:
catches the issue:

`RuntimeError: could not find signal 'CRUISE_INACTIVE' in 'SCC_CONTROL', DBC b'hyundai_canfd'`

PR #28764, opened 2023-06-30
IsoTpParallelQuery: skip finished requests

sshane:
no need to rx or tx to these addresses once done

pre-reqs:
- [ ] https://github.com/commaai/panda/pull/1487
- [ ] rx_in_progress -> any rx
- [ ] ensure exceptions from isotpmessages mark addr as done, and can't extend timeouts any more, even if we still get msgs


sshane:
Replaced by https://github.com/commaai/openpilot/pull/29654

PR #28765, opened 2023-06-30
UI: adjust opacity instead of visibility to indicate engageability

adeebshihadeh:


sshane:
@deanlee this PR is suspected to increase max draw time, do you have any ideas what could cause that?

sshane:
nvm

PR #28771, opened 2023-07-01
ui/MapSettings: remove white border around destinations

deanlee:
| Before  | After |
| Screenshot 2023-07-01 14:23:47 | Screenshot 2023-07-01 14:24:56  |


incognitojam:
Is this also visible for other scrollers, like in the network/wifi list? Maybe we could set it inside `ScrollView` instead

deanlee:
It will not affect other interfaces. Other places that use scrollview have set the stylesheet for the border.

We currently can't set it in the scrollview, which would cause the setting panel sto lose the border of the rounded rectangle.

Issue #28778, opened 2023-07-02
cabana segfaults when loading log files on macOS


visualage:
### Describe the bug

When loading log files (rlog or qlog) on macOS or any C++ compiler that does not have `<memory_resource>`, it may segfault. This is caused by a usage after free bug in `tools/replay/logreader.cc`.

### Provide a route where the issue occurs

bd059400a5cde994|2023-07-01--20-31-49--0

### openpilot version

6495a18cc9e8c1394ac7395d24bb0a661ce35fbf

PR #28779, opened 2023-07-02
replay: fix segfault caused by pointer usage after freeing

visualage:
In `replay/logreader.cc`, there is a usage after free bug, which caused cabana segfault loading certain routes on certain platforms (e.g. macOS). More descriptions of the issue can be found at https://github.com/commaai/openpilot/issues/28778

Loading the mentioned route won't segfault after the fix on macOS.


deanlee:
This bug was introduced by https://github.com/commaai/openpilot/pull/28779/files. Thanks for fixing it!

PR #28780, opened 2023-07-02
ui/map: draw ETA with rich formatted text

deanlee:
Draw ETA with rich formatted text to boost ETA rendering performance and simplify the code.
Screenshot from 2023-07-02 15-13-56

The original version of `MapETA::updateETA` takes 6~10ms or so to execute, sometimes even close to 20ms （adjustSize,repaint can be very time consuming), which may cause onroadWindow to drop frames. After refactoring, the average execution time of `MapETA::updateETA` + `MapETA::paint` is less than 1ms.


adeebshihadeh:
Nice! How did you benchmark? If just print statements, can we keep up with something that's easy to run?

adeebshihadeh:
Can you also make the style match the existing one? The bottom corners shouldn't be rounded and it should touch the border.

deanlee:
I just added some printf to do benchmark.

deanlee:
> Can you also make the style match the existing one? The bottom corners shouldn't be rounded and it should touch the border.

done

selfdrive/ui/qt/maps/map.h
```diff
@@ -50,20 +50,16 @@ public slots:
...
-
 public:
   MapETA(QWidget * parent=nullptr);
-
-public slots:
   void updateETA(float seconds, float seconds_typical, float distance);
+
+private:
+  void paintEvent(QPaintEvent *event) override;
+  void showEvent(QShowEvent *event) override { format_24h = Params().getBool("NavSettingTime24h"); }
```
adeebshihadeh, line 59: we shouldn't create the params every time

incognitojam:
The widget isn't expanding to the full width when there is more text, and I think the font is different for the numbers?
Screenshot 2023-07-06 125613
Screenshot 2023-07-06 125551

deanlee:
fixed the typo of the font name . the font should be "Inter", not "Inner"-_-

deanlee:
the spacing between strings may be a little different from the original after using RTF.

adeebshihadeh:
Why close? This seems like a nice cleanup :)

deanlee:
I'm not sure what happened, I didn't close this pr. maybe it's a bug in github?

deanlee:
I'm not sure what happened, I didn't close this pr. maybe it's a bug in github?

deanlee:
I see a prompt below that says it has been merged?

Screenshot from 2023-07-07 01-30-54

deanlee:
ah...I see, I accidentally reset it to master, and then github automatically closed it

incognitojam:
It seems to show the ETA in 24h format even if 24h toggle is off. When the toggle is on it says "17:04 eta", but when it is off it says "17:04 pm".

deanlee:
Sorry I didn't test enough to cause this problem. fixed:
Screenshot from 2023-07-09 00-28-12

sshane:
@deanlee alex noticed that this PR changes the spacing between the elements, can we revert that?

Before:
image

After:

image

deanlee:
@sshane I will change it to be basically the same as the original. 
It would be better if https://github.com/commaai/openpilot/pull/28976  could be merged before the change. otherwise there will be conflicts and have to rebase again.

sshane:
I think it would be better to fix it first, I don't have too much time to review that other PR at the moment.

incognitojam:
I just tested it out, merged!

PR #28782, opened 2023-07-02
Car Docs: (optionally) Display a list of tools

jnewb1:
**Description**

Adds a new "Tool" part, and splits the "View" dropdown into two sections. I always forget what size wrench you need, and new users might find that information hard to find too. Other car ports should also add their required tools.

image


adeebshihadeh:
Ah, I had forgotten about this. I actually wanted to include the tool with the Subaru harnesses but forgot about it.

jnewb1:
> Ah, I had forgotten about this. I actually wanted to include the tool with the Subaru harnesses but forgot about it. 

Ok to merge this?

selfdrive/car/docs_definitions.py
```diff
@@ -129,18 +129,28 @@ class Kit(EnumBase):
...
 
 
+class Tools:
+  @staticmethod
+  def socket(size, modifier=None):
+    return BasePart(f"Socket Wrench {size}" + f' ({modifier})' if modifier is not None else '')
+
+
+class Tool(EnumBase):
+  socket_8mm_deep = Tools.socket("8mm or 5/16", "deep")
```
sshane, line 139: need to say deep in a separate field?

jnewb1: ill just use basepart, tools don't seem to be very common anyway

PR #28783, opened 2023-07-02
ui/map: cache navigation images at startup

deanlee:
Cache images to avoid reloading, scaling, and flipping (rhd) images with each refresh.it can also prevents Qlable from redrawing the pixmap when its content has not changed.

### DIFF
--- a/selfdrive/ui/qt/maps/map.cc
+++ b/selfdrive/ui/qt/maps/map.cc
@@ -421,17 +421,20 @@ void MapInstructions::buildPixmapCache() {
   for (QString fn : dir.entryList({"*" + ICON_SUFFIX}, QDir::Files)) {
     QPixmap pm(dir.filePath(fn));
     QString key = fn.left(fn.size() - ICON_SUFFIX.length());
+    pm = pm.scaledToWidth(200, Qt::SmoothTransformation);
+
+    // Maneuver icons
+    pixmap_cache[key] = pm;
+    // lane direction icons
     if (key.contains("turn_")) {
-      pixmap_cache[key] = pm.scaled({125, 125}, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
-    } else {
-      pm = pm.scaledToWidth(200, Qt::SmoothTransformation);
-      pixmap_cache[key] = pm;
-      // for rhd, reflect direction and then flip
-      if (key.contains("_left")) {
-        pixmap_cache["rhd_" + key.replace("_left", "_right")] = pm.transformed(QTransform().scale(-1, 1));
-      } else if (key.contains("_right")) {
-        pixmap_cache["rhd_" + key.replace("_right", "_left")] = pm.transformed(QTransform().scale(-1, 1));
-      }
+      pixmap_cache["lane_" + key] = pm.scaled({125, 125}, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
+    }
+
+    // for rhd, reflect direction and then flip
+    if (key.contains("_left")) {
+      pixmap_cache["rhd_" + key.replace("_left", "_right")] = pm.transformed(QTransform().scale(-1, 1));
+    } else if (key.contains("_right")) {
+      pixmap_cache["rhd_" + key.replace("_right", "_left")] = pm.transformed(QTransform().scale(-1, 1));
     }
   }
 }
@@ -489,7 +492,7 @@ void MapInstructions::updateInstructions(cereal::NavInstruction::Reader instruct
     }
 
     // TODO: Make more images based on active direction and combined directions
-    QString fn = "direction_";
+    QString fn = "lane_direction_";
     if (left) {
       fn += "turn_left";
     } else if (right) {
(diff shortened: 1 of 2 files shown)


selfdrive/ui/qt/maps/map.cc
```diff
@@ -484,27 +506,14 @@ void MapInstructions::updateInstructions(cereal::NavInstruction::Reader instruct
...
-        fn.replace("right", "left");
-      }
-    }
-
-    QPixmap pix(fn);
-    if (is_rhd) {
-      pix = pix.transformed(QTransform().scale(-1, 1));
-    }
-    icon_01->setPixmap(pix.scaledToWidth(200, Qt::SmoothTransformation));
+    bool rhd  = is_rhd && fn.contains("_left") || fn.contains("_right");
```
incognitojam, line 515:
```suggestion
    bool rhd = is_rhd && (fn.contains("_left") || fn.contains("_right"));
```

incognitojam, line 515:
```suggestion
    bool rhd = is_rhd && fn.contains("_left") || fn.contains("_right");
```

selfdrive/ui/qt/maps/map.cc
```diff
@@ -460,27 +482,14 @@ void MapInstructions::updateInstructions(cereal::NavInstruction::Reader instruct
...
-        fn.replace("right", "left");
-      }
-    }
-
-    QPixmap pix(fn);
-    if (is_rhd) {
-      pix = pix.transformed(QTransform().scale(-1, 1));
-    }
-    icon_01->setPixmap(pix.scaledToWidth(200, Qt::SmoothTransformation));
+    bool rhd = is_rhd && fn.contains("_left") || fn.contains("_right");
```
adeebshihadeh, line 491: can you add some parentheses here to make this less ambiguous?

deanlee: added.this is also a bug if without parentheses

adeebshihadeh: is there a compiler flag we can set to make the build fail?

deanlee: Not sure, hope there is such a flag. i'll look up the documentation and add it if there is.

selfdrive/ui/qt/maps/map.cc
```diff
@@ -407,12 +408,33 @@ MapInstructions::MapInstructions(QWidget * parent) : QWidget(parent) {
...
+
   QPalette pal = palette();
   pal.setColor(QPalette::Background, QColor(0, 0, 0, 150));
   setAutoFillBackground(true);
   setPalette(pal);
 }
 
+void MapInstructions::buildPixmapCache() {
+  QDir dir("../assets/navigation");
+  for (QString fn : dir.entryList({"*.png"}, QDir::Files)) {
```
adeebshihadeh, line 421: use ICON_SUFFIX instead of hardcoding .png

selfdrive/ui/qt/maps/map.cc
```diff
@@ -516,7 +525,7 @@ void MapInstructions::updateInstructions(cereal::NavInstruction::Reader instruct
     if (!label->parentWidget()) {
       lane_layout->addWidget(label);
     }
-    label->setPixmap(loadPixmap(fn + ICON_SUFFIX, {125, 125}, Qt::IgnoreAspectRatio));
+    label->setPixmap(pixmap_cache[fn + ICON_SUFFIX]);
```
adeebshihadeh, line 528: remove ICON_SUFFIX from the key?

adeebshihadeh (changes requested):

selfdrive/ui/qt/maps/map.cc
```diff
@@ -411,6 +412,28 @@ MapInstructions::MapInstructions(QWidget *parent) : QWidget(parent) {
...
+    if (key.contains("turn_")) {
+      pixmap_cache[key] = pm.scaled({125, 125}, Qt::IgnoreAspectRatio, Qt::SmoothTransformation);
+    } else {
+      pm = pm.scaledToWidth(200, Qt::SmoothTransformation);
+      pixmap_cache[key] = pm;
+      // for rhd, reflect direction and then flip
+      if (key.contains("_left")) {
+        pixmap_cache["rhd_" + key.replace("_left", "_right")] = pm.transformed(QTransform().scale(-1, 1));
+      } else if (key.contains("_right")) {
+        pixmap_cache["rhd_" + key.replace("_right", "_left")] = pm.transformed(QTransform().scale(-1, 1));
```
sshane, line 433: would directions flip with rhd?

incognitojam:
The scaling appears to be wrong for some instructions:

This PR:
image

master:
image

The icons seem to be correct (or at least, same as master in replay) for RHD. Will test in the car just to make sure. The mapbox directions don't seem perfect for roundabouts here.

deanlee:
Fixed it, I didn't realize there were some icon files that were shared. It just scales differently when used.

deanlee:
I didn't realize there were some icon files that were shared. they  just scales differently when used.

incognitojam (approved):

PR #28784, opened 2023-07-02
Subaru: Global SnG with throttle

jnewb1:
**Description**

SnG support for global subaru models.

- Sending buttons doesn't work, because buttons are wired directly to Eyesight.
- Throttle message is only allowed when the car is stopped

This currently works about 70% of the time on gen2. I suspect that since throttle is at 100hz, it might be too fast for the COUNTER+1 strategy to work perfectly. Perhaps if we send two messages (COUNTER+1, COUNTER+2), it would work every time.


jnewb1:
throttle isn't ideal for gen2 as it's on the same bus as the pcm, we probably want to find a better solution

PR #28788, opened 2023-07-03
ui/map: reuse the Qlabels and remove call to `clearLayout`

deanlee:
clear layout and rebuild labels can impact performance. 
Reusing labels can also benefits from https://github.com/commaai/openpilot/pull/28783, as QLabel does not need to be redrawn when the pixmap has not changed.


adeebshihadeh:
Nice fix. We should probably remove all uses of clearLayout and delete it so it can't be used again.

PR #28795, opened 2023-07-03
Swap `belowSteerSpeed` and `resumeRequired` alert priorities

### LINKED ISSUE #28361: `resumeRequired` does not display in UI if `belowSteerSpeed`
On cars without steer to zero, the `belowSteerSpeed` message takes priority over `resumeRequired` when at a standstill. Displaying the latter message at a standstill makes it clear to the driver that their input is required, and its absence may lead to confusion about why the vehicle is not moving.

I have confirmed with several people that this is the case for GM, and am checking on Toyota (those are the only two makes that appear to set the `resumeRequired` message).

nworb-cire:
Should fix https://github.com/commaai/openpilot/issues/28361.

I also added the SnG criterion for sending the resume message so that we don't have to remember to add that if/when we get SnG implemented on GM.

### DIFF
--- a/selfdrive/car/gm/interface.py
+++ b/selfdrive/car/gm/interface.py
@@ -282,7 +282,7 @@ class CarInterface(CarInterfaceBase):
       events.add(EventName.belowEngageSpeed)
     if ret.cruiseState.standstill and not self.CP.autoResumeSng:
       events.add(EventName.resumeRequired)
-    elif ret.vEgo < self.CP.minSteerSpeed:
+    if ret.vEgo < self.CP.minSteerSpeed:
       events.add(EventName.belowSteerSpeed)
 
     ret.events = events.to_msg()
--- a/selfdrive/car/gm/interface.py
+++ b/selfdrive/car/gm/interface.py
@@ -280,9 +280,9 @@ class CarInterface(CarInterfaceBase):
     if below_min_enable_speed and not (ret.standstill and ret.brake >= 20 and
                                        self.CP.networkLocation == NetworkLocation.fwdCamera):
       events.add(EventName.belowEngageSpeed)
-    if ret.cruiseState.standstill:
+    if ret.cruiseState.standstill and not self.CP.autoResumeSng:
       events.add(EventName.resumeRequired)
-    if ret.vEgo < self.CP.minSteerSpeed:
+    elif ret.vEgo < self.CP.minSteerSpeed:
       events.add(EventName.belowSteerSpeed)
 
     ret.events = events.to_msg()
--- a/selfdrive/car/gm/interface.py
+++ b/selfdrive/car/gm/interface.py
@@ -280,7 +280,7 @@ class CarInterface(CarInterfaceBase):
     if below_min_enable_speed and not (ret.standstill and ret.brake >= 20 and
                                        self.CP.networkLocation == NetworkLocation.fwdCamera):
       events.add(EventName.belowEngageSpeed)
-    if ret.cruiseState.standstill and not self.CP.autoResumeSng:
+    if ret.cruiseState.standstill:
       events.add(EventName.resumeRequired)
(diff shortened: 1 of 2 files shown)


adeebshihadeh (changes requested):
This is a good fix, but it should use the alert priorities instead.

nworb-cire:
Currently `resumeRequired` is `LOW` and `belowSteerSpeed` is `MID`. Should we set `resumeRequired` to `HIGH` priority, or is it better to swap them? I don't want to clobber distracted driver alerts, although that's perhaps a non-issue while at standstill.

adeebshihadeh:
Unfortunately, the alert priorities are a bit of a mess. They should really be refactored to reason about them more easily, but in the meantime, want to make a simple debug script that prints out the alerts in order of their precedence? 

I'm thinking it would look something like this:
```
State.preEnabled
* HIGHEST: stockFcw, controlsInitializing, startup
* HIGH: stockAeb, steerTempUnavailableSilent
* MID:
* LOW: 
* LOWER: 
* LOWEST: 

State.softDisabling
...
```

Needs to be per-state, since we only [show a subset of alert types](https://github.com/commaai/openpilot/blob/e63e2dde18b9ae2925d1a38c527dccfd0c99d4c5/selfdrive/controls/controlsd.py#L485) in each state.

nworb-cire:
Done. Was a bit unsure of the best way of handling the dynamic alerts, so I just instantiated empty carParams/carState for the time being.

I can clean it up if the output is what you're looking for. Can also split into a separate PR if you'd like.

sshane:
@nworb-cire can you print the events that will now have a higher priority over the minSteerSpeed alert?

nworb-cire:
See #28898

Setting `belowSteerSpeed` to `Priority.LOW` would put it lower in priority than these other `WARNING`s:
```
Priority.HIGH:
       driverDistracted driverUnresponsive speedTooHigh

Priority.MID:
       promptDriverDistracted promptDriverUnresponsive resumeRequired
```

selfdrive/car/gm/interface.py
```diff
@@ -280,7 +280,7 @@ def _update(self, c):
     if below_min_enable_speed and not (ret.standstill and ret.brake >= 20 and
                                        self.CP.networkLocation == NetworkLocation.fwdCamera):
       events.add(EventName.belowEngageSpeed)
-    if ret.cruiseState.standstill:
+    if ret.cruiseState.standstill and not self.CP.autoResumeSng:
```
sshane, line 283: We only need one change, right? Either this or changing the priorities.

sshane: I worry that cruiseControl.resume may be noisy, so we will spam the audible min steering speed alert in stop and go.

I haven't personally tried this, so maybe it's a non-issue. Have you been driving on this branch, what is your experience? Is it annoying to hear the alert when you press resume?

sshane:
> Setting belowSteerSpeed to Priority.LOW would put it lower in priority than these other WARNINGs:

> Priority.HIGH:
 >       driverDistracted driverUnresponsive speedTooHigh

> Priority.MID:
 >       promptDriverDistracted promptDriverUnresponsive resumeRequired

Oh wow, we don't alert at all under 7 mph for DM?!

selfdrive/car/gm/interface.py
nworb-cire: This isn't doing anything with the CC resume signal, though you are correct that it is noisy. This change would just not show the "press resume" alert if the user is not in fact required to press resume. (Note that currently, no GM cars have working sng on stock; this is more of a preemptive change to the alert for when that becomes a reality.)

nworb-cire:
@adeebshihadeh anything holding this up? I believe it still requires your approval since you requested changes.

adeebshihadeh:
Looks good, thanks for writing the tool too! Also fyi, that's not how the reviews work.

PR #28796, opened 2023-07-03
nav ui: restore speed-adaptive map zoom level

jyoung8607:
Speed-adaptive map zoom has been mostly broken since #25784. The map does a one time zoom to an appropriate level for your speed at the moment you engage navigation, or after the timer expires on a pan or pinch action, but then remains stuck at that same zoom level no matter how your speed changes.

@deanlee was this an expensive operation? Is it okay to put back at 20Hz? Before proposing a full revert, I tried a version that called setZoom() at 1Hz and the visual effect was unpleasant compared to full rate.

### DIFF
--- a/selfdrive/ui/qt/maps/map.cc
+++ b/selfdrive/ui/qt/maps/map.cc
@@ -194,8 +194,7 @@ void MapWindow::updateState(const UIState &s) {
 
   if (zoom_counter == 0) {
     m_map->setZoom(util::map_val<float>(velocity_filter.x(), 0, 30, MAX_ZOOM, MIN_ZOOM));
-    zoom_counter = -1;
-  } else if (zoom_counter > 0) {
+  } else {
     zoom_counter--;
   }
 
--- a/selfdrive/ui/qt/maps/map.h
+++ b/selfdrive/ui/qt/maps/map.h
@@ -100,7 +100,7 @@ private:
   // Panning
   QPointF m_lastPos;
   int pan_counter = 0;
-  int zoom_counter = -1;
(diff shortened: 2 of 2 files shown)


sshane:
I had no idea we had this already! Explains the weird behavior when touching the screen and zoom changes 😅

deanlee:
> I had no idea we had this already! Explains the weird behavior when touching the screen and zoom changes sweat_smile

I was the one who didn't notice this feature first. You were misled by me.-_-

 I think we can restore this feature right away, and if there are performance issues, just find a way to optimize it.

jyoung8607:
> I had no idea we had this already! Explains the weird behavior when touching the screen and zoom changes 😅

I'd forgotten it was supposed to be there, till I started using nav regularly again with the advent of nav-on-openpilot, and driving mixed low/high speed routes started highlighting that zoom was wonky.

I feel it works quite well while in-motion; the velocity filter is nicely tuned for silky smooth reaction to speed changes. However, as a side effect, when you slow to a stop the map can take a few more seconds to settle. When you're at a stoplight and want to actually look at the map, it's not delightful for the map to still be in-motion. One thing to try would be adding a non-zero minimum to the velocity filter input, perhaps 2-3 m/s, so that we smoothly arrive and remain at a stable consistent zoom during very low or zero speed operation. But any such retuning of the algorithm is a matter for another PR.

sshane:
works on PC for me. @deanlee if you want to make some easy improvements, it looks like it continues to zoom when you pan manually, it probably should stop (like if you zoomed). Can also reset the filter when it resets the view/re-centers

PR #28800, opened 2023-07-04
ui/map: Use an overlay widget to simplify code and improve rendering performance.

deanlee:
Put all the widgets on top of map into an overlay widget. this will remove most of the moving/resizing-related code  and improve rendering performance.


incognitojam:
thanks!

PR #28802, opened 2023-07-04
ui/wifiManager: remove obsolete methods without definition.

deanlee:


adeebshihadeh:
Is there some kind of static analysis or compiler flag we can add for this? Would be nice if openpilot just didn't build in this case.

PR #28808, opened 2023-07-05
Type hint interfaces

nworb-cire:
This adds abstract base classes for `CarController` and `CarControllerParams`, as well as type hinting most things in `interfaces.py` and the make specific directories.

Combined with https://github.com/commaai/cereal/pull/493, this renders PyCharm (and I assume VS Code as well) capable of introspecting most of the codebase now, leading to a far friendlier development environment.


sshane:
I would avoid moving classes around when making larger changes like this, as it makes it harder to immediately review at a glance. I would also try to make some clean ups at the same time, like moving common variables in the base car controller class, which I think is a fine change on its own. In general, it's best to split things up where it makes sense to do so.

nworb-cire:
I did that to type hint the CarStateBase type in the interface constructor; I suppose we could put it in quotes instead, but that looks messy IMO.

I do see what you mean about making it harder to review the changes. I'll move it back and quote the type name, then maybe submit a follow-up PR to move things and unquote if I feel it necessary at that point.

selfdrive/car/body/carcontroller.py
```diff
@@ -14,8 +15,9 @@
 MAX_TURN_INTEGRATOR = 0.1  # meters
 
 
-class CarController:
+class CarController(CarControllerBase):
```
sshane, line 18: Perhaps you make this a base class in a new PR, bring out common stuff like setting self.CP, etc.

nworb-cire: I started on this, but quickly came to the realization that it would duplicate >75% of this PR if I were to do it separately and type-hint as I went

nworb-cire:
I'm unsure what's causing the CANPacker errors, but it seems unrelated to the changes made here as I don't see that I have touched anything relating to opendbc. scons builds this branch just fine on my device.

@sshane are there any further changes or additions you require before this can be merged?

PR #28809, opened 2023-07-05
ui: cleanup controls

deanlee:
1. remove `class LayoutWidget` . it's not used after https://github.com/commaai/openpilot/issues/28476
2. remove `horizontal_line`.  this control will not have any visible effect under the current stylesheet:
 

| Begin  | After |
| Screenshot 2023-07-05 12:31:42  | Screenshot 2023-07-05 12:32:31  |

### DIFF
--- a/selfdrive/ui/qt/widgets/controls.h
+++ b/selfdrive/ui/qt/widgets/controls.h
@@ -11,8 +11,6 @@
 #include "selfdrive/ui/qt/widgets/input.h"
 #include "selfdrive/ui/qt/widgets/toggle.h"
 
-QFrame *horizontal_line(QWidget *parent = nullptr);
-
 class ElidedLabel : public QLabel {
   Q_OBJECT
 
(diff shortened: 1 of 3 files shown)


selfdrive/ui/qt/widgets/controls.h
```diff
@@ -288,13 +286,3 @@ class ListWidget : public QWidget {
   QVBoxLayout outer_layout;
   QVBoxLayout inner_layout;
 };
-
-// convenience class for wrapping layouts
-class LayoutWidget : public QWidget {
```
adeebshihadeh, line 293: let's keep this, it's still useful

sshane:
@deanlee yesterday I briefly saw three very thin horizontal lines in the network pane before it loaded networks. could something you did recently cause that?

deanlee:
@sshane It may not be due to my modifications, the only style-related changes I can think of is this pr. 
 I haven't found the style problem you mentioned.  I will test again to see if I can reproduce it.

Issue #28810, opened 2023-07-05
UI: onroad alerts sometimes disappear for a frame


adeebshihadeh:
Noticed while "Reverse Gear" was showing in this segment: `a996a26002357498|2023-07-04--15-26-53--0`.

adeebshihadeh:
@deanlee want to look into this?

PR #28811, opened 2023-07-05
ui: fix alert disappear when no controlState is received in a frame.

deanlee:
issue: alert disappers when no controlState is received in a frame:

[Kazam_screencast_00122.webm](https://github.com/commaai/openpilot/assets/27770/6eb87e88-e920-4edb-8171-1788fbd88390)

@adeebshihadeh , I used a qlog to test it. It should fix the https://github.com/commaai/openpilot/issues/28810  as well.

### DIFF
--- a/selfdrive/ui/ui.h
+++ b/selfdrive/ui/ui.h
@@ -49,17 +49,22 @@ struct Alert {
 
   static Alert get(const SubMaster &sm, uint64_t started_frame) {
     const cereal::ControlsState::Reader &cs = sm["controlsState"].getControlsState();
-    Alert alert = {cs.getAlertText1().cStr(), cs.getAlertText2().cStr(),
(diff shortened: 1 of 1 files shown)


selfdrive/ui/ui.h
```diff
@@ -47,40 +47,40 @@ struct Alert {
     return text1 == a2.text1 && text2 == a2.text2 && type == a2.type && sound == a2.sound;
   }
 
-  static Alert get(const SubMaster &sm, uint64_t started_frame) {
+  static std::optional<Alert> get(const SubMaster &sm, uint64_t started_frame) {
```
adeebshihadeh, line 50: I think Alert::get should always return an alert to show, so the std::optional doesn't make sense. instead, i'd add some logic to handle this case inside of Alert::get

PR #28812, opened 2023-07-05
ui/map: fix layout disappear if reroute attempt fails

### LINKED ISSUE #28492: nav ui: overlays disappear if first reroute attempt fails
Easy to hit by going from home wifi -> LTE

deanlee:
resolve https://github.com/commaai/openpilot/issues/28492

### DIFF
--- a/selfdrive/ui/qt/maps/map.cc
+++ b/selfdrive/ui/qt/maps/map.cc
@@ -201,6 +201,7 @@ void MapWindow::updateState(const UIState &s) {
 
   if (sm.updated("navInstruction")) {
     if (sm.valid("navInstruction")) {
+      m_map->setLayoutProperty("navLayer", "visibility", "visible");
       auto i = sm["navInstruction"].getNavInstruction();
       emit ETAChanged(i.getTimeRemaining(), i.getTimeRemainingTypical(), i.getDistanceRemaining());
 


selfdrive/ui/qt/maps/map.cc
```diff
@@ -201,6 +201,7 @@ void MapWindow::updateState(const UIState &s) {
 
   if (sm.updated("navInstruction")) {
     if (sm.valid("navInstruction")) {
+      m_map->setLayoutProperty("navLayer", "visibility", "visible");
```
sshane, line 204: Does this work? `step_idx` is set to None in navd when it clears the route and hides the layout, so `navInstruction` won't be valid. Besides, navd recovers on its own since it's continually retrying until it has internet again. It then re-sends navRoute, showing the layout again.

I think we might want to just show a warning/error that it's waiting for internet, need to do this in navd probably.

sshane: In fact, I think any time we do `self.clear_route` in navd, we should give a reason. It can clear it because of an API error, no internet, user drives away after reaching destination, etc.

deanlee: Yes you are right.closed.

PR #28814, opened 2023-07-05
deleter: preserve flagged segments

### LINKED ISSUE #27857: deleter: simple preserve policy for bookmarked segments
Preserve last 5 segments on device?

incognitojam:
- [x] set xattr on segment dir when receive a userflag msg
- [x] don't delete last 5 flagged segments (for each, flagged segment and prior segment)

closes https://github.com/commaai/openpilot/issues/27857


adeebshihadeh:
This should also preserve the surrounding segments, e.g. a bookmark in segment 1 should result in preserving segments 0, 1, and 2.

system/loggerd/tests/test_loggerd.py
```diff
@@ -284,10 +284,10 @@ def test_userflag(self):
 
     # sleep enough for the first poll to time out
     # TODO: fix loggerd bug dropping the msgs from the first poll
-    # managed_processes["loggerd"].start()
-    # for s in services:
-    #   while not pm.all_readers_updated(s):
-    #     time.sleep(0.1)
+    managed_processes["loggerd"].start()
+    for s in services:
+      while not pm.all_readers_updated(s):
```
adeebshihadeh, line 289: I wrote a helper for this today https://github.com/commaai/cereal/commit/cf9c5cbb9196f80a25bd8421a9921ce4801a7561

incognitojam:
I think 5 (10 actually) segments would be 15% of the 60 minutes of storage on a 32GB edition?

We could consider preserving the route when it uploads to connect, so that we can free up the space on the device later, but:
1. connect don't have a list of preserved routes, it's only in useradmin right now
2. preserved routes per user is limited

adeebshihadeh:
I think it's fine to unmark them as preserved once they're uploaded, but we'll have to keep track of that per-file.

incognitojam:
I think you may want to upload more than just the qlog and qcam if you flag a segment, too.

system/loggerd/tests/test_deleter.py
```diff
@@ -98,6 +98,50 @@ def test_no_delete_with_lock_file(self):
...
+      time.sleep(2)
+    finally:
+      self.join_thread()
+
+    self.join_thread()
+
+    self.assertTrue(f_paths[0].exists() and f_paths[1].exists(), "File deleted when preserved")
+    self.assertFalse(f_paths[2].exists(), "File not deleted")
+
+  def test_delete_many_preserved(self):
```
incognitojam, line 120:
```suggestion
  def test_delete_oldest_preserved(self):
```

system/loggerd/tests/test_loggerd.py
```diff
@@ -276,6 +276,32 @@ def test_rlog(self):
       sent.clear_write_flag()
       self.assertEqual(sent.to_bytes(), m.as_builder().to_bytes())
 
+  def test_userflag(self):
```
adeebshihadeh, line 279:
```suggestion
  def test_preverving_flagged_segments(self):
```

system/loggerd/tests/test_loggerd.py
```diff
@@ -247,8 +248,7 @@ def test_rlog(self):
     # TODO: fix loggerd bug dropping the msgs from the first poll
     managed_processes["loggerd"].start()
     for s in services:
-      while not pm.all_readers_updated(s):
-        time.sleep(0.1)
+      pm.wait_for_readers_to_update(s, timeout=5)
```
adeebshihadeh, line 251: same here, assert

system/loggerd/tests/test_loggerd.py
```diff
@@ -276,6 +276,32 @@ def test_rlog(self):
...
+      for s in services:
+        try:
+          m = messaging.new_message(s)
+        except Exception:
+          m = messaging.new_message(s, random.randint(2, 10))
+        pm.send(s, m)
+        sent_msgs[s].append(m)
+
+    time.sleep(2)
+    managed_processes["loggerd"].stop()
```
adeebshihadeh, line 300: this whole sequence we do in a few places should be pulled out

system/loggerd/deleter.py
```diff
@@ -20,7 +29,19 @@ def deleter_thread(exit_event):
...
+      # skip deleting most recent N preserved segments (and their prior segment)
+      preserved_dirs = []
+      for n, d in enumerate(filter(has_preserve_xattr, reversed(dirs))):
+        if n == PRESERVE_COUNT:
+          break
+        date_str, _, seg_num = d.rpartition("--")
+
+        # preserve segment and its prior
+        preserved_dirs.append(d)
+        preserved_dirs.append(f"{date_str}--{int(seg_num) - 1}")
```
adeebshihadeh, line 42: this should be handled the same way we handle `DELETE_LAST`, just move them to the back in the sorting. we need a test case that catches the way this is written now; this breaks if you have less than 5 segments of storage

adeebshihadeh: sorting should go:
* boot/crash
* preserved segments, sorted by creation
* rest of segments, sorted by creation

system/loggerd/tests/test_deleter.py
```diff
@@ -98,6 +98,50 @@ def test_no_delete_with_lock_file(self):
...
+      f_paths_preserved.append(self.make_file_with_data(self.seg_dir, self.f_type, preserve_xattr=deleter.PRESERVE_ATTR_VALUE))
+
+    self.start_thread()
+
+    try:
+      with Timeout(2, "Timeout waiting for file to be deleted"):
+        while all([f_path.exists() for f_path in f_paths_preserved]):
+          time.sleep(0.01)
+    except TimeoutException:
+      pass
```
adeebshihadeh, line 138: what's the point of the timeout exception if it's just caught? just add to the while if you want that behavior

adeebshihadeh (changes requested):

system/loggerd/tests/test_loggerd.py
```diff
@@ -276,6 +263,16 @@ def test_rlog(self):
       sent.clear_write_flag()
       self.assertEqual(sent.to_bytes(), m.as_builder().to_bytes())
 
+  def test_preserving_flagged_segments(self):
+    services = set(random.sample(CEREAL_SERVICES, random.randint(5, 10))) | {"userFlag"}
+
+    managed_processes["loggerd"].start()
+    self._publish_random_messages(services)
+    managed_processes["loggerd"].stop()
```
adeebshihadeh, line 271: the loggerd start/stop and services can even be moved into _publish_random_messages

system/loggerd/tests/test_loggerd.py
```diff
@@ -276,6 +263,16 @@ def test_rlog(self):
...
+
+    segment_dir = self._get_latest_log_dir()
+    self.assertEqual(getxattr(segment_dir, PRESERVE_ATTR_NAME), PRESERVE_ATTR_VALUE)
```
adeebshihadeh, line 274: can we also add a case for non-preserved segs?

system/loggerd/tests/test_loggerd.py
```diff
@@ -71,6 +74,26 @@ def _check_sentinel(self, msgs, route):
...
+      for s in services:
+        try:
+          m = messaging.new_message(s)
+        except Exception:
+          m = messaging.new_message(s, random.randint(2, 10))
+        pm.send(s, m)
+        sent_msgs[s].append(m)
+      time.sleep(0.01)
+
+    time.sleep(1)
```
adeebshihadeh, line 94: can we replace the sleeps with pm.all_readers_updated?

system/loggerd/deleter.py
```diff
@@ -5,22 +5,43 @@
...
     out_of_percent = get_available_percent(default=MIN_PERCENT + 1) < MIN_PERCENT
 
     if out_of_percent or out_of_bytes:
+      dirs = listdir_by_creation(ROOT)
+
+      # skip deleting most recent N preserved segments (and their prior segment)
+      preserved_dirs = []
+      for n, d in enumerate(filter(has_preserve_xattr, reversed(dirs))):
+        if n == PRESERVE_COUNT:
+          break
```
adeebshihadeh, line 36: note for later: we can remove the explicit count once the uploader unpreserves uploaded segments

adeebshihadeh, line 36: i'd move this into a get_preserved_segments helper

system/loggerd/deleter.py
```diff
@@ -5,22 +5,43 @@
...
+        date_str, _, seg_num = d.rpartition("--")
+
+        # preserve segment and its prior
+        preserved_dirs.append(d)
+        preserved_dirs.append(f"{date_str}--{int(seg_num) - 1}")
```
adeebshihadeh, line 41: needs some exception handling. what if I set the preserve xattr on the boot dir?

system/loggerd/tests/loggerd_tests_common.py
```diff
@@ -105,8 +107,11 @@ def tearDown(self):
         raise
 
   def make_file_with_data(self, f_dir: str, fn: str, size_mb: float = .1, lock: bool = False,
-                          xattr: Optional[bytes] = None) -> Path:
+                          upload_xattr: Optional[bytes] = None, preserve_xattr: Optional[bytes] = None) -> Path:
```
adeebshihadeh, line 110: why don't the xattr args take bools instead?

incognitojam: We have tests to make sure that setting them to `b'0'` are treated as unset. https://github.com/commaai/openpilot/pull/28206

Could make it a union and accept both for this helper?

adeebshihadeh: nah, this is fine then

adeebshihadeh (approved):

PR #28815, opened 2023-07-05
nav model: fix flicker in nav enabled state

### LINKED ISSUE #28794: nav: enabled state bouncing
### Describe the bug

On a recent road trip, using navigate-on-openpilot, I saw occasional flickering in the UI. The borders were sometimes drawn in the nav-disabled state for just a moment before returning to the nav-enabled state.

It turns out the UI is behaving correctly, that modelv2.navEnabled is going False for single frames. It's not immediately obvious why; at a glance the [conditions for enabling model navigation](https://github.com/commaai/openpilot/blob/62ae247417c63b698b02e89462655891b68fe821/selfdrive/modeld/modeld.cc#L140C35-L150) seem okay.

c5cbf3352b0f03bb|2023-07-02--19-03-08, c5cbf3352b0f03bb|2023-07-02--22-07-16

### openpilot version

6495a18cc9e8c1394ac7395d24bb0a661ce35fbf plus a tweaked MPC constant

### Additional info

nav-enabled-flap-0

nav-enabled-flap-1

mitchellgoffpc:


adeebshihadeh:
Why don't we add the liveLocationKalman logMonoTime to the nav model packet instead? It's nice to have anyway, since it seems the only way to tie nav model packets to liveLocationKalmans is to use the mapRenderState to build a lookup table.

selfdrive/test/process_replay/model_replay.py
```diff
@@ -207,10 +207,12 @@ def do_GET(self):
         'logMonoTime',
         'modelV2.frameDropPerc',
         'modelV2.modelExecutionTime',
+        'modelV2.locationMonoTime',
         'driverStateV2.modelExecutionTime',
         'driverStateV2.dspExecutionTime',
         'navModel.dspExecutionTime',
         'navModel.modelExecutionTime',
+        'navModel.locationMonoTime',
```
adeebshihadeh, line 215: why isn't this deterministic? wouldn't the replay outputs differ if it wasn't?

selfdrive/modeld/modeld.cc
```diff
@@ -138,10 +138,9 @@ void run_model(ModelState &model, VisionIpcClient &vipc_client_main, VisionIpcCl
     }
 
     // Enable/disable nav features
-    double tsm = (float)(nanos_since_boot() - sm["mapRenderState"].getMapRenderState().getLocationMonoTime()) / 1e6;
-    bool route_valid = sm["navInstruction"].getValid() && (tsm < 1000);
-    bool render_valid = sm["mapRenderState"].getValid() && (sm["navModel"].getNavModel().getFrameId() == sm["mapRenderState"].getMapRenderState().getFrameId());
-    bool nav_valid = route_valid && render_valid;
+    uint64_t timestamp_llk = sm["navModel"].getNavModel().getLocationMonoTime();
+    double tsm = (float)(nanos_since_boot() - timestamp_llk) / 1e6;
+    bool nav_valid = sm["navInstruction"].getValid() && sm["navModel"].getValid() && (tsm < 1000);
```
adeebshihadeh, line 143: I'd expect navModel's valid to deal with navInstruction's valid

adeebshihadeh, line 143: also 1ms seems too strict?

mitchellgoffpc: I think the threshold is 1000ms here?

selfdrive/test/process_replay/model_replay.py
mitchellgoffpc: I believe they are deterministic, they just require the refs to be updated and I wanted to keep them the same. https://jenkins.comma.life/blue/organizations/jenkins/openpilot/detail/fix-nav-flicker/2/pipeline/70

selfdrive/modeld/modeld.cc
adeebshihadeh: hmm, that sounds like too much, but this is in nanoseconds. if it's not clear from the code, we should rename so it is

adeebshihadeh: I see 1s is two navModel packets

PR #28821, opened 2023-07-06
acados: update to 0.2.2

maxime-desroches:


fredyshox:
Mac binaries are not updated, I can build them if you don't have any mac at hand

maxime-desroches:
@fredyshox Yes please do

fredyshox:
Is the ref 28f4f22196632ec6ea145c09f635010622d99504 the one that you used? After building it on my machine I've got different files in `include/` and `acados_template/`

maxime-desroches:
Should be good now

fredyshox:
something is up with 0.2.1, because mpc tests are not passing on my mac too (same outputs), while on master they do

PR #28823, opened 2023-07-06
boardd: check fw version when connecting

### LINKED ISSUE #28301: boardd: add second independent fw version check
`pandad` does the flashing and version check, then launches boardd. boardd should also double check the fw version independently

deanlee:
resolve https://github.com/commaai/openpilot/issues/28301


selfdrive/boardd/panda.cc
```diff
@@ -153,6 +157,20 @@ std::optional<std::string> Panda::get_serial() {
   return err >= 0 ? std::make_optional(serial_buf) : std::nullopt;
 }
 
+bool Panda::verify_firmware_version() {
+  if (auto fw_sig = get_firmware_version()) {
+    typedef cereal::PandaState::PandaType Type;
+    bool panda_h7 = (hw_type == Type::RED_PANDA || hw_type == Type::RED_PANDA_V2 || hw_type == Type::TRES);
```
adeebshihadeh, line 163: don't like these hardcoded lists in multiple places. how about we just have a fw file list and return true if it matches any of them?

adeebshihadeh:
LGTM, any reason it's still a draft?

PR #28835, opened 2023-07-07
Revert "UI: Indicate when Nav on OP is enabled (#28665)"

sshane:
to run jenkins on


sshane:
I still got this run on the reverted commit, something else might've changed recently as well

```

------------------------------------------------

-------------- UI Draw Timing ------------------

------------------------------------------------

min  2.51ms

max  176.15ms

std  9.71ms

mean 6.68ms
```

sshane:
https://github.com/commaai/openpilot/commit/1690cf4af6c88c52eeec0c61eae10074a169d563

PR #28836, opened 2023-07-07
cabana: fix the issue that editing signal will segfault when there is a filter string

deanlee:
fixed the issue: https://github.com/commaai/openpilot/discussions/26091#discussioncomment-6381557


sshane:
do you think we could add some sort of fuzzing tests to Cabana? does QT have anything like that? I'm glad we're fixing bugs, but it seems like there's many that are still undiscovered

deanlee:
Yes. I agree with you. the basic features of cabana are now relatively stable. I will start writing test routines for the core modules.

Issue #27206, opened 2023-02-03
Hyundai Tucson 2023 poor tuning


adeebshihadeh:
Waiting on more stock routes from user. 

```
8426de28ebdc65a2|2023-01-27--20-07-53
```

- [ ] Fix tuning/steer scaling
- [ ] Look into other HKG models that are potentially like this

adeebshihadeh:
Looks like it's just one FW version that needs different tuning. Temporarily removed it in #27340 until we better understand the difference.

sshane:
@adeebshihadeh looks like this camera FW version was added again under `SANTA_CRUZ_1ST_GEN`. I wonder if it also has bad tuning.

sshane:
@nuwandavek by the way, it looks like we have 4 active Tucson users in the last 90 days with these different parts: `{'1966303b764e7142', '8426de28ebdc65a2', 'bf1a6b0ba293c37b', 'a0191364609ee493'}`

sunnyhaibin:
I've been running `LAT_ACCEL = 2.9 - 3.1` and `FRICTION = 0.095` for the past few days with the 2022 Hyundai Tucson Hybrid, and it is significantly better than the current values.
- https://github.com/commaai/openpilot/pull/30513

sshane:
I would like to combine the Santa Cruz and Tucson. According to the [wikipedia page](https://en.wikipedia.org/wiki/Hyundai_Santa_Cruz), it is "closely related" to the Tucson as it shares the platform code with an extension of OB meaning Open Bed. The mass and wheel base are a bit different though.

@nuwandavek can you look at these two vehicles in the fleet and determine if using a median value is safe, letting torqued learn the difference?

PR #28845, opened 2023-07-07
edit fw for 2022 hyundai tucson

zzhu415:
I installed the stock openpilot does not work for me even with master-ci.
I followed the instruction to get my fw and updated value.py
The following values does recognize my car. (2022 hyundai tucson limited)


sshane:
what is your dongle id?

zzhu415:
here is my dongle id
a0191364609ee493

sshane:
Unfortunately your car has a part number used in the Santa Cruz and therefore requires different tuning. Please refer to this issue: https://github.com/commaai/openpilot/issues/27206

PR #28847, opened 2023-07-08
Toyota LTA: lower angle rate limits

sshane:
lower limits, follow up to https://github.com/commaai/openpilot/pull/28735

matches Ford safety more closely with up/down jerks. plus the EPS on a TSS 2.5 vehicle limits jerk to 2.0 m/s^2 up so it doesn't make sense to try to request more.

injection test to find max jerk by the EPS with no limits:

image

297b4b460f361603|2023-07-06--22-11-36

### DIFF
--- a/selfdrive/car/toyota/values.py
+++ b/selfdrive/car/toyota/values.py
@@ -24,11 +24,11 @@ class CarControllerParams:
 
   # Lane Tracing Assist (LTA) control limits
   # Assuming a steering ratio of 13.7:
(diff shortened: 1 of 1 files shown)


sshane:
Varying SETME_X64 had the side effect of lowering the internal EPS rate limits along with max torque on my Camry which felt good and didn't have bad oscillations, but without that, did somewhat. With these lower up limits and full torque, it feels similar to what SETME_X64 did.

PR #28848, opened 2023-07-08
Toyota LTA: limit max angle with applied torque

sshane:
Uses EPS's applied torque signal (which includes total applied torque eg. driver, from LKAS/LTA, etc) to force the requested angle to lower if torque rises above our set max torque value.

Because we know the EPS is torque limited, setting this limit to anything below the max should cause no internal EPS windup, like it would if we tried to request more angle than the torque can reach.

Some injections with this code, EPS should be limited to 1500 units which it is.

image

image

image

image

image


sshane:
thinking of how this looks in panda safety before merging, down rate checks based on speed and now EPS torque will be super complicated to do cleanly in panda.

selfdrive/car/toyota/carcontroller.py
```diff
@@ -94,7 +95,11 @@ def update(self, CC, CS, now_nanos):
     can_sends.append(create_steer_command(self.packer, apply_steer, apply_steer_req))
     if self.frame % 2 == 0 and self.CP.carFingerprint in TSS2_CAR:
       lta_active = CC.latActive and self.CP.steerControlType == SteerControlType.angle
-      can_sends.append(create_lta_steer_command(self.packer, self.last_angle, lta_active, self.frame // 2))
+      need_to_winddown = False  # abs(CS.out.steeringAngleDeg) > abs(actuators.steeringAngleDeg) and abs(CS.out.steeringAngleDeg) - abs(actuators.steeringAngleDeg) > 4
+      enable_condition = abs(CS.out.steeringTorqueEps) < 1500 and abs(CS.out.steeringTorque) < 150 and not need_to_winddown
```
sshane, line 99: need to understand how specifically this works. just varying this bit does not seem to have the same affect as looking at the EPS torque. this limits the torque (when user not touching wheel) perfectly to whatever you set it to

sshane: Note: the enable bits completely drop torque immediately on the TSS 2.5 2023 RAV4s, not a viable way to limit torque. SETME_X64 at 0 drops torque slowly

PR #28854, opened 2023-07-09
ui/map: handle & display error in MapWindow

deanlee:
Currently the error message is displayed by reusing the distance label of `MapInstructions`. This will complicate the code of `MapInstructions`.
MapInstructions should only be responsible for displaying nav Instructions, and let `MapWindow` display the error message itself.This also makes it easier to use different colors or styles to display error in the future.

This PR also fixed an issue that "Map Loading"  or "Waiting for GPS" is always displayed if `navInstruction` is invalid on startup. (This was caused by the `MapInstructions::noError()`  in the original code, it simply setting error to false, and not hiding the distance label.)


incognitojam:
The instruction icon seems to take up half of the widget
Screenshot from 2023-07-11 17-16-31

deanlee:
fixed

deanlee:
Sorry for so many errors :(.  
To be honest, I'm currently testing it by modifying the code of map.cc to generate some nav data. It is both time-consuming and prone to errors.
 @sshane:  It would be great if the demo route could have navigation data.

sshane:
I will get that done soon!

selfdrive/ui/qt/maps/map.cc
```diff
@@ -378,46 +383,32 @@ void MapWindow::updateDestinationMarker() {
...
-    lane_layout = new QHBoxLayout(lane_widget);
-    layout->addWidget(lane_widget);
+  layout->addWidget(secondary = new QLabel);
+  secondary->setStyleSheet(R"(font-size: 50px;)");
+  secondary->setWordWrap(true);
 
-    main_layout->addLayout(layout);
-  }
+  layout->addLayout(lane_layout = new QHBoxLayout);
+  layout->addStretch();
```
deanlee, line 408: @incognitojam Not sure if we need this stretch, if there is only one message it will make the message always be displayed at the top. 
The  original version  didn't have such a stretch. it will centers labels vertically in the available space.
what do you think ? should we keep this stretch or remove it?

incognitojam: Should be fine without

Issue #28841, opened 2023-07-07
CI: test setup + build on latest Ubuntu


adeebshihadeh:

adeebshihadeh:
We'll just switch to 24.04 soon.

PR #28855, opened 2023-07-09
CI Update to Latest Ubuntu

gdifiore:
**Description**
Updated CI scripts to use latest Ubuntu according  to #28841


adeebshihadeh:
We don't want to update our default to 22.04 - we just want to add a new CI job to run `tools/ubuntu_setup.sh` and build openpilot. This new job should go in the tools_test file.

PR #28862, opened 2023-07-10
ui/map:  singleton navigation requests

deanlee:
There are multiple instances of `MapSettings` in home and map, each with its own `RequestRepeater` to query from commaapi for navigation requests.  these duplicate requests result in wasted bandwidth and resources, and may potentially lead to synchronization issues.

### DIFF
--- a/selfdrive/ui/qt/maps/map_settings.h
+++ b/selfdrive/ui/qt/maps/map_settings.h
@@ -27,7 +27,6 @@ class NavigationRequest : public QObject {
   Q_OBJECT
 
 public:
-  NavigationRequest(QObject *parent);
   static NavigationRequest *instance();
 
 signals:
@@ -35,6 +34,8 @@ signals:
   void nextDestinationUpdated(const QString &response, bool success);
 
 private:
+  NavigationRequest(QObject *parent);
+
   Params params;
 };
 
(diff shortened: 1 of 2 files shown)


selfdrive/ui/qt/maps/map_settings.h
```diff
@@ -23,6 +23,21 @@ const QString NAV_FAVORITE_LABEL_WORK = "work";
 class NavDestination;
 class DestinationWidget;
 
+class NavigationRequest : public QObject {
+  Q_OBJECT
+
+public:
+  NavigationRequest(QObject *parent);
```
incognitojam, line 30: should we make the constructor non-public to avoid doing this again accidentally?

incognitojam:
The destination list is empty onroad, first starting after this commit :confused:

incognitojam:
It does seem related as I can't reproduce the issue on the prior commit (d34138e2759e8067e9d7d005957e29a23684c71c)

deanlee:
I'm checking to see what the possible problem is. It's unlikely to be this pr's problem.

deanlee:
@incognitojam it's Indeed a related issue.  `NavigationRequest` only reads the locations from `ApiCache_NavDestinations` on initialization and then `emit locationsUpdated`. this results in only the first `mapSetting` window can gets an immediate update. the second `mapsetting` can only be updated when `NavigationRequest` gets the list from the server, which can take up to 30 seconds.

PR #28865, opened 2023-07-10
map settings: accept touch events

sshane:
Related: https://github.com/commaai/openpilot/pull/28476

### DIFF
--- a/selfdrive/ui/qt/maps/map_settings.cc
+++ b/selfdrive/ui/qt/maps/map_settings.cc
@@ -118,6 +118,11 @@ MapSettings::MapSettings(bool closeable, QWidget *parent)
   }
 }
 
+void MapSettings::mousePressEvent(QMouseEvent *ev) {
+//  m_lastPos = ev->localPos();
+  ev->accept();
+}
+
 void MapSettings::showEvent(QShowEvent *event) {
   updateCurrentRoute();
 }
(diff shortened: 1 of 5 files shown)


selfdrive/ui/qt/maps/map_settings.cc
```diff
@@ -244,11 +249,11 @@ DestinationWidget::DestinationWidget(QWidget *parent) : QPushButton(parent) {
   inner_frame->setSpacing(0);
   {
     title = new ElidedLabel(this);
-    title->setAttribute(Qt::WA_TransparentForMouseEvents);
+//    title->setAttribute(Qt::WA_TransparentForMouseEvents);
```
sshane: @incognitojam do you know why we need these?

sshane:
@incognitojam this if statement made the function a bit confusing to read, think it's simpler like this. It mimics what the map widget does. Any reason it was done this way, or is this good?

selfdrive/ui/qt/maps/map_settings.cc
incognitojam: if you tap on the text the click event doesn't get passed to the DestinationWidget without it

incognitojam:
If you test it and it seems to work fine then LGTM - can't think of any reason I did it this way

selfdrive/ui/qt/maps/map_settings.cc
sshane: https://github.com/commaai/openpilot/pull/28869/files

PR #28869, opened 2023-07-10
ui: fix ElidedLabel mouse event propagation

sshane:
`clicked` is only used here:

https://github.com/commaai/openpilot/blob/e63e2dde18b9ae2925d1a38c527dccfd0c99d4c5/selfdrive/ui/qt/offroad/networking.cc#L304-L306

and ElidedLabel is only used in BoxLayouts like the one in networking:

or QFrame, like in controls.h:

https://github.com/commaai/openpilot/blob/e63e2dde18b9ae2925d1a38c527dccfd0c99d4c5/selfdrive/ui/qt/widgets/controls.h#L67-L77

and so those receiving extra mouse release events shouldn't affect anything (no double taps or anything like that)


sshane:
not a clear improvement, we should treat this either as a button or not, not both

Issue #28857, opened 2023-07-10
memory leak when loggerd/encoderd get out of sync


adeebshihadeh:

adeebshihadeh:
@deanlee want to look into this?

deanlee:
looks like we need to rest the `LoggerdState` when get out of sync, including clearing the old packages that are in the queue.otherwise it will cause endless recursive calls.

deanlee:
already submitted two PRs to address the memory leak issues:

https://github.com/commaai/openpilot/pull/28873
https://github.com/commaai/openpilot/pull/28697

It may require a few more PRs to improve the stability of loggerd.

adeebshihadeh:
Do we need both? What's the smallest possible fix for the current issue?

deanlee:
#28873 is for the current issue.

the smallest possible quit and dirty fix could be something like this :https://github.com/commaai/openpilot/pull/28901

but I strongly recommend use #28873. the recursive call is very buggy.  we really should  refactoring loggerd instead of just keeping it working as it is.

adeebshihadeh:
I see, this is a big change to get in the upcoming release. I'll check out the fix you prefer after we ship this one.

PR #28873, opened 2023-07-11
loggerd: remove recursive call to handle_encoder_msg 

deanlee:
related issue: https://github.com/commaai/openpilot/issues/28857.  call `handle_encoder_msg` to write msgs in the queue will causes infinite recursive calls when out of sync with encoderd .


deanlee:
This PR just splits the code related to writing packet into function  `write_encode_data`. the rest of the code remains unchanged.

PR #28874, opened 2023-07-11
ui: remove unused MapPanel function

sshane:
added in https://github.com/commaai/openpilot/pull/28476, removed usage in https://github.com/commaai/openpilot/pull/28865


sshane:
@deanlee do you know if there is some unused function compiler error we can enable?

deanlee:
I haven't found any such flag so far.Currently only unused static functions can have compiler warnings, class member functions don't have similar flags.

sshane:
thanks for checking

PR #28883, opened 2023-07-11
experimental mode: gate nav

### LINKED ISSUE #28871: move nav on openpilot behind experimental mode

sshane:
Changes behavior of Experimental Mode: enabling end-to-end longitudinal control -> a toggle that enables all sorts of experimental features.

- [x] need to communicate which features are available when user turns on experimental mode (stock ACC cars won't have end-to-end longitudinal for example)


selfdrive/controls/controlsd.py
```diff
@@ -845,7 +845,7 @@ def step(self):
     self.prof.checkpoint("Ratekeeper", ignore=True)
 
     self.is_metric = self.params.get_bool("IsMetric")
-    self.experimental_mode = self.params.get_bool("ExperimentalMode") and self.CP.openpilotLongitudinalControl
+    self.experimental_mode = self.params.get_bool("ExperimentalMode")  # and self.CP.openpilotLongitudinalControl
```
sshane, line 848: Since we want experimental mode to be enabled if no longitudinal, I moved the logic for openpilotLongitudinalControl for end-to-end related features to the subscribers (ui path and planner). Decide if we want to publish something like `endToEndLongitudinal` here as well (need to wait for a CP packet in replay before it shows path

adeebshihadeh: I think this is fine

sshane: oh, one last place is VCruiseHelper for setting vCruise to 65 mph

PR #28893, opened 2023-07-11
loggerd: refactor remote encoder

### LINKED ISSUE #27512: PC: loggerd memory leak

### LINKED ISSUE #28857: memory leak when loggerd/encoderd get out of sync

deanlee:
- [x] refactor struct RemoteEncoder to class EncoderWriter. and split  to encoder_writer.cc/h.
- [x] fix memory leaks. 
resolve: https://github.com/commaai/openpilot/issues/28857
resolve: https://github.com/commaai/openpilot/issues/27512
- [x] fix sync issues & simplify sync strategy. 
- [x] unit tests for EoncoderWriter


deanlee:
the new sync strategy is very simple and straightforward:  rotate to the next segment if remote segment changes:https://github.com/commaai/openpilot/blob/cee606a95f080941df5e7aafe470ee65ca72fcb1/system/loggerd/encoder_writer.cc#L3-L27

If this strategy is fine, then I think this PR could greatly simplify the synchronization with `encoderd` and improve the robustness of the code. basically, most of the code is in encoder_writer.cc, excluding unit tests, which is about 100 lines or so. It shouldn't be too hard  to review. therefore, I have closed all other small fixes based on the original code.

PR #28896, opened 2023-07-12
Cloudlog: cleanup zmq cloudlog warnings

haraschax:

### DIFF
--- a/selfdrive/locationd/laikad.py
+++ b/selfdrive/locationd/laikad.py
@@ -3,6 +3,8 @@ import math
 import os
 import time
 import shutil
+import warnings
+
 from collections import defaultdict
 from concurrent.futures import Future, ProcessPoolExecutor
 from enum import IntEnum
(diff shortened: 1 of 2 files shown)


selfdrive/locationd/laikad.py
```diff
@@ -396,7 +398,9 @@ def fetch_navs(self, t: GPSTime, block):
 
 
 def get_orbit_data(t: GPSTime, valid_const, auto_update, valid_ephem_types, cache_dir):
-  astro_dog = AstroDog(valid_const=valid_const, auto_update=auto_update, valid_ephem_types=valid_ephem_types, cache_dir=cache_dir)
+  warnings.filterwarnings("ignore", category=ResourceWarning, message="unclosed.*<zmq.*>")
```
adeebshihadeh, line 401: hmm, we don't want to just ignore these. you know where it's coming from?

haraschax: Kacper was looking into it. AFAIK it just comes from having an open zmq socket when running forking a process. 

```
import warnings
warnings.filterwarnings("default")
import zmq
import os

zctx = zmq.Context()
sock = zctx.socket(zmq.PUSH)
#sock.close()

pid = os.fork()
```
Is there a solution except manually closing the socket before spawning a child, everywhere you imported cloudlog? Because that doesn't seem great.

jyoung8607: https://github.com/commaai/openpilot/issues/23666#issuecomment-1071002173

PR #28901, opened 2023-07-12
loggerd: quick and dirty fix for the memory leak when out of sync with encoder

deanlee:


deanlee:
Closed. There is something wrong with this fix. 
This part of the code may be called recursively while the previous level is looping in the re.q. This causes the queue to be changed upon return and the original queue loop may crash.

PR #28904, opened 2023-07-12
map panel: show map on interactive timeout

sshane:
shows map if interaction times out in map settings


sshane:
not sure if there's a standard way of signaling down to child widgets in QT, does this look good to you @incognitojam?

incognitojam:
I feel like just subscribing to the signal in `map_panel.cc` would be better than passing it down

sshane:
made `Device` a singleton. @deanlee is this how you would do this, or should we start putting this method inside the class (`Device::instance()`)

deanlee:
Yes, I think it's better to singleton it. otherwise you often run into the problem of not being able to access the device class.  I've had many headaches over this:D

deanlee:
Think we can remove the redundant `awake` from uistate now

sshane:
done!

PR #28907, opened 2023-07-12
ui: set max brightness when showing the QR code

### LINKED ISSUE #28906: ui: screen brightness at 100% while pairing QR code is up

deanlee:
resolve: [#28906](https://github.com/commaai/openpilot/issues/28906)

### DIFF
--- a/selfdrive/ui/qt/widgets/prime.cc
+++ b/selfdrive/ui/qt/widgets/prime.cc
@@ -26,12 +26,12 @@ PairingQRWidget::PairingQRWidget(QWidget* parent) : QWidget(parent) {
 void PairingQRWidget::showEvent(QShowEvent *event) {
   refresh();
   timer->start(5 * 60 * 1000);
-  uiState()->setOffroadBrightness(100);
+  device()->setOffroadBrightness(100);
 }
 
 void PairingQRWidget::hideEvent(QHideEvent *event) {
   timer->stop();
-  uiState()->setOffroadBrightness(BACKLIGHT_OFFROAD);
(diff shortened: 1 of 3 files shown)


selfdrive/ui/qt/widgets/prime.cc
```diff
@@ -26,10 +26,13 @@ PairingQRWidget::PairingQRWidget(QWidget* parent) : QWidget(parent) {
 void PairingQRWidget::showEvent(QShowEvent *event) {
   refresh();
   timer->start(5 * 60 * 1000);
+  // set max brightness
+  Hardware::set_brightness(100);
```
adeebshihadeh, line 30: think we want to only call this at one place in the ui. can you make some intermediate state that tells the Device class we want 100% brightness?

deanlee:
how about this one?

deanlee:
convert to draft. we can use device() to access the device class after https://github.com/commaai/openpilot/pull/28904

sshane:
it still goes through the filter, so it takes a while to ramp up and down from max brightness. we probably want it instantly here, right @adeebshihadeh?

adeebshihadeh:
I just remembered we won't even get full brightness anyway because of [brightnessd](https://github.com/commaai/agnos-builder/blob/master/userspace/usr/comma/brightnessd.py). I'll have to think about this one some more.

adeebshihadeh:
> it still goes through the filter, so it takes a while to ramp up and down from max brightness. we probably want it instantly here, right @adeebshihadeh?

This should be fine.

PR #28911, opened 2023-07-13
edit: Change Bolt required hardware

notAreYouScared:
Comma power & RJ45 cable's are not needed as the power is provided from the camera connector and harness box.


sshane:
You need to edit the car docs directly, not the generated file.

sshane:
@adeebshihadeh I don't think there's any reason to keep shipping these parts, fine to remove?

adeebshihadeh:
If we're not sure, let's keep it for now.

sshane:
Sounds good.

PR #28923, opened 2023-07-13
[GMC ACADIA DENALI 2018] Move to torque control, non-linear feed-forward

nuwandavek:


selfdrive/car/torque_data/override.yaml
```diff
@@ -49,6 +49,7 @@ KIA NIRO HYBRID 2ND GEN: [2.42, 2.5, 0.12]
 KIA NIRO EV 2ND GEN: [2.05, 2.5, 0.14]
 GENESIS GV80 2023: [2.5, 2.5, 0.1]
 KIA CARNIVAL 4TH GEN: [1.75, 1.75, 0.15]
+GMC ACADIA DENALI 2018: [1.6, 1.6, 0.2]
```
adeebshihadeh, line 52: why is this in the override? I thought things started here, then strictly move to params.yaml?

selfdrive/car/gm/interface.py
```diff
@@ -57,14 +48,17 @@ def sig(val):
...
+    elif self.CP.carFingerprint == CAR.ACADIA:
+      a, b, c, _ = [4.78003305, 1.0, 0.3122, 0.05591772]
 
     steer_torque = (sig(lateral_accel_value * a) * b) + (lateral_accel_value * c)
     return float(steer_torque) + friction
 
   def torque_from_lateral_accel(self) -> TorqueFromLateralAccelCallbackType:
-    if self.CP.carFingerprint == CAR.BOLT_EUV:
-      return self.torque_from_lateral_accel_bolt
+    if self.CP.carFingerprint in [CAR.BOLT_EUV, CAR.ACADIA]:
```
adeebshihadeh, line 60: this list is in two places. suggest moving the a, b, c weights to a dict at the top and using the dict keys here

selfdrive/car/torque_data/override.yaml
nuwandavek: This was not a torque control car (was pid) before. Also, these parameters for the non linear cars don't mean the same. It does not matter where we put it, but just wanted to note that the "factor" does not mean anything.

adeebshihadeh (approved):

PR #28925, opened 2023-07-14
torqued: always calculate estimates

sshane:
useful for debugging. changes behavior from:

- liveTorqueParams always broadcasting what the torque controller uses

to:

- liveTorqueParams always broadcasting its best estimate, even if we're not ready to use it. controlsd then just needs to check liveValid, which is what torqued used to check to update the estimates

todo next week:
- [ ] run on segments to catch all errors


adeebshihadeh:
Still relevant?

sshane:
Yes, we need to run it on a lot of segments to regression test though before merging.

adeebshihadeh:
@nuwandavek any update on this?

sshane:
I haven't had time to test it yet. This seems like a good improvement for porting cars, torqued almost never has fully learned by the time we merge a car, and quickly checking the qlog in useradmin would be a big step for people to set an accurate lat accel factor.

sshane:
As discussed, we will make a script that makes it easy to replay torqued on a route

PR #28926, opened 2023-07-14
Laikad cpu usage: enable test onroad

haraschax:


adeebshihadeh:
I don't think this is gonna be consistent enough. Until we have a better GPS setup in the testing closet, I think we can just check <80% and >1% as a sanity check.

PR #28928, opened 2023-07-14
Use features from navmodeld replay for modeld replay

mitchellgoffpc:


adeebshihadeh:
Shall we close this until after python modeld is in?

adeebshihadeh:
Is this still relevant?

mitchellgoffpc:
> Is this still relevant?

Yes, it's not all that urgent so it hasn't been a priority for me, but if we ever change the size of the nav feature vector we'd have to switch the route to get logs with the new feature vector size. This PR fixes this issue by using the features from the nav model instead of the ones from the logs. I'll try to clean it up and get it in next week if possible.

PR #28929, opened 2023-07-14
CANPacker: don't allow undefined signals or messages

sshane:
bumping opendbc to test with openpilot

### DIFF
--- a/selfdrive/car/__init__.py
+++ b/selfdrive/car/__init__.py
@@ -187,6 +187,8 @@ class CanBusBase:
     if CP is None:
       assert fingerprint is not None
       num = math.ceil(max([k for k, v in fingerprint.items() if len(v)], default=1) / 4)
+      print('here2', fingerprint, num)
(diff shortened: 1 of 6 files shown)


sshane:
added fuzzy car params, now catches bugs in non-default CP.flags function calls!

selfdrive/car/__init__.py
```diff
@@ -187,6 +187,7 @@ def __init__(self, CP, fingerprint: Optional[Dict[int, Dict[int, int]]]) -> None
     if CP is None:
       assert fingerprint is not None
       num = math.ceil(max([k for k, v in fingerprint.items() if len(v)], default=1) / 4)
+      print('here2', fingerprint, num)
```
sshane, line 190:
```suggestion
      print('here2', fingerprint, num)
```

sshane:
can't yet merge, since the interceptor stack relies on this behavior

PR #28935, opened 2023-07-14
encoderd: get avaiable streams from vipc server.

deanlee:


system/loggerd/encoderd.cc
```diff
@@ -123,12 +125,25 @@ void encoder_thread(EncoderdState *s, const LogCameraInfo &cam_info) {
...
+  }
+
+  if (!streams.empty()) {
+    std::vector<std::future<void>> encoder_threads;
+    for (auto stream : streams) {
+      auto it = std::find_if(std::begin(cameras_logged), std::end(cameras_logged),
+                             [stream](auto &cam) { return cam.stream_type == stream; });
+      assert(it != std::end(cameras_logged));
+      ++s.max_waiting;
+      encoder_threads.emplace_back(std::async(std::launch::async, encoder_thread, &s, *it));
```
adeebshihadeh, line 144: can we go back to the std::thread?

system/loggerd/encoderd.cc
```diff
@@ -62,7 +64,6 @@ void encoder_thread(EncoderdState *s, const LogCameraInfo &cam_info) {
         }
       } else {
         LOGE("not initting empty encoder");
-        s->max_waiting--;
```
adeebshihadeh, line 65: should we just remove this whole block and add an assert instead?

system/loggerd/encoderd.cc
```diff
@@ -58,16 +60,11 @@ void encoder_thread(EncoderdState *s, const LogCameraInfo &cam_info) {
     // init encoders
     if (encoders.empty()) {
       VisionBuf buf_info = vipc_client.buffers[0];
+      assert(buf_info.width > 0 && buf_info.height > 0);
       LOGW("encoder %s init %dx%d", cam_info.thread_name, buf_info.width, buf_info.height);
```
adeebshihadeh, line 64:
```suggestion
      LOGW("encoder %s init %dx%d", cam_info.thread_name, buf_info.width, buf_info.height);
      assert(buf_info.width > 0 && buf_info.height > 0);
```

system/loggerd/encoderd.cc
```diff
@@ -127,12 +124,27 @@ void encoder_thread(EncoderdState *s, const LogCameraInfo &cam_info) {
 void encoderd_thread() {
   EncoderdState s;
 
-  std::vector<std::thread> encoder_threads;
-  for (const auto &cam : cameras_logged) {
-    encoder_threads.push_back(std::thread(encoder_thread, &s, cam));
-    s.max_waiting++;
+  std::set<VisionStreamType> streams;
+  while (!do_exit) {
```
adeebshihadeh, line 128: we have this loop in so many places now, would be nice to make the blocking version handle this

PR #28931, opened 2023-07-14
test car interface: fuzzy CarParams

sshane:
Randomly generates car fw and fingerprints so each car interface's feature detection has a higher chance of exploring all possible states, like different flags. For example, we now explore the else branch here, catching any bugs in `create_buttons` where previously we didn't explore that at all:

https://github.com/commaai/openpilot/blob/719d14f039c7ef0558aefd697e843de107bba074/selfdrive/car/hyundai/carcontroller.py#L143-L149

PR #27903, opened 2023-04-13
Hyundai: factor out bus numbers

adeebshihadeh:
closes #26867

PR #28938, opened 2023-07-14
CanBusBase: fix offset calculation

sshane:
fixes crash in CanBusBase when there is only CAN traffic on bus 0 (no events in sentry, very unlikely to actually happen)

Added in: https://github.com/commaai/openpilot/pull/27903
Caught in: https://github.com/commaai/openpilot/pull/28931

### DIFF
--- a/selfdrive/car/__init__.py
+++ b/selfdrive/car/__init__.py
@@ -186,8 +186,7 @@ class CanBusBase:
   def __init__(self, CP, fingerprint: Optional[Dict[int, Dict[int, int]]]) -> None:
     if CP is None:
       assert fingerprint is not None
-      num = math.ceil(max([k for k, v in fingerprint.items() if len(v)], default=1) / 4)
+      num = max([k for k, v in fingerprint.items() if len(v)], default=0) // 4 + 1
     else:
       num = len(CP.safetyConfigs)
-
     self.offset = 4 * (num - 1)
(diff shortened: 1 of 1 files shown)


sshane:
```python
# test cases

fingerprint = {}
print(math.ceil(max([k for k, v in fingerprint.items() if len(v)], default=1) / 4))
print(max([k for k, v in fingerprint.items() if len(v)], default=0) // 4 + 1)
1
1

fingerprint = {0: {0:0}}
print(math.ceil(max([k for k, v in fingerprint.items() if len(v)], default=1) / 4))
print(max([k for k, v in fingerprint.items() if len(v)], default=0) // 4 + 1)
0
1

fingerprint = {0: {0:0}, 1: {0: 0}}
print(math.ceil(max([k for k, v in fingerprint.items() if len(v)], default=1) / 4))
print(max([k for k, v in fingerprint.items() if len(v)], default=0) // 4 + 1)
1
1

fingerprint = {0: {0:0}, 1: {0: 0}, 2: {1:1}, 3: {1:1}, 4: {1:2}}
print(math.ceil(max([k for k, v in fingerprint.items() if len(v)], default=1) / 4))
print(max([k for k, v in fingerprint.items() if len(v)], default=0) // 4 + 1)
1
2

fingerprint = {0: {0:0}, 1: {0: 0}, 2: {1:1}, 3: {1:1}, 4: {1:2}, 5: {1:2}}
print(math.ceil(max([k for k, v in fingerprint.items() if len(v)], default=1) / 4))
print(max([k for k, v in fingerprint.items() if len(v)], default=0) // 4 + 1)
2
2
```

incognitojam (approved):

PR #28941, opened 2023-07-14
Hyundai: fix alt button cancel message always blocked

### LINKED ISSUE #28822: Hyundai: alt button cancel message is always blocked
context in https://github.com/commaai/opendbc/issues/882

basically, since we don't set `aReqRaw` and `aReqValue` to 0, they get set to 0 as bytes, which is max decel parsed out in panda, thus these cancel messages get blocked.

image

sshane:


sshane:
We can either fix this by changing behavior of what happens when you don't set signals, or enforce that signals with non-zero offsets be explicitly set. I'm not sure which is best at the moment, so I'm going to merge this standalone. These are the only signals for Hyundai CANFD we send that have offsets.

PR #28948, opened 2023-07-14
Ford: radar + long control fixups

haraschax:


bjsmyth:
Line 140 has a typo for the ford radar_interface, there's a multiply next to the `cos` before the parameter parenthesis.

haraschax:
> Line 140 has a typo for the ford radar_interface, there's a multiply next to the `cos` before the parameter parenthesis.

oops, fixed

selfdrive/car/ford/carcontroller.py
```diff
@@ -89,9 +90,11 @@ def update(self, CC, CS, now_nanos):
       gas = accel
       if not CC.longActive or gas < CarControllerParams.MIN_GAS:
         gas = CarControllerParams.INACTIVE_GAS
+      else:
+        gas = 2.5
 
       stopping = CC.actuators.longControlState == LongCtrlState.stopping
-      can_sends.append(create_acc_msg(self.packer, self.CAN, CC.longActive, gas, accel, stopping))
+      can_sends.append(create_acc_msg(self.packer, self.CAN, CC.longActive, gas, accel, stopping, v_ego_kph=40 * CV.KPH_TO_MS))
```
incognitojam, line 97: is converting to m/s intentional? looking at logs I'm pretty sure this signal is kph with the stock system

selfdrive/car/ford/fordcan.py
```diff
@@ -133,7 +133,9 @@ def create_acc_msg(packer, CAN: CanBus, long_active: bool, gas: float, accel: fl
     "AccBrkTot_A_Rq": accel,                          # Brake total accel request: [-20|11.9449] m/s^2
     "Cmbb_B_Enbl": 1 if long_active else 0,           # Enabled: 0=No, 1=Yes
     "AccPrpl_A_Rq": gas,                              # Acceleration request: [-5|5.23] m/s^2
+    "AccPrpl_A_Pred": gas,                            # Acceleration request: [-5|5.23] m/s^2
     "AccResumEnbl_B_Rq": 1 if long_active else 0,
+    "AccVeh_V_Trg": v_ego_kph,                        # Target speed: [0|255] km/h
```
sshane, line 138: We can merge these separately, now, if they help control at all. @incognitojam have you tested on your car?

haraschax: Don't think I had any conclusive results on those on our bronco.

incognitojam: I haven't tested these on my car

sheaduncan:
Just a heads up, this doesn't work for the F150, we've reworked it and have a solution we've implemented. Once the F150 gets supported we can submit a PR for the changes.

adeebshihadeh:
Which part? There’s no reason not to open that PR now if it’s ready to go.

sheaduncan:
image

Some things are unrelated to this PR but within Ford_Radar that need to change for the F150 (signals, updated DBC, and reference the correct DBC).

Related to this code change directly, the radar points it was tracking kept throwing false positives.

Ford's CAN-FD platform has additional signals that parse the data and pick the lead. We pivoted to using these signals to get the lead and lead speed. This then feeds into radard.

It appears that Ford may be gaining traction now. Before, it wasn't apparent that the value was there to do a PR for the Ford CAN-FD platform since it was an unsupported vehicle.

adeebshihadeh:
We have to rent one anyways for the safety. A more complete port will make the best of that time.

FYI, this is how every car port works - we merge them in dashcam mode and iterate on them instead of leaving them up in a stale PR somewhere.

selfdrive/car/ford/fordcan.py
```diff
@@ -133,7 +133,9 @@ def create_acc_msg(packer, CAN: CanBus, long_active: bool, gas: float, accel: fl
     "AccBrkTot_A_Rq": accel,                          # Brake total accel request: [-20|11.9449] m/s^2
     "Cmbb_B_Enbl": 1 if long_active else 0,           # Enabled: 0=No, 1=Yes
     "AccPrpl_A_Rq": gas,                              # Acceleration request: [-5|5.23] m/s^2
+    "AccPrpl_A_Pred": gas,                            # Acceleration request: [-5|5.23] m/s^2
```
sshane, line 136: If it's like Hyundai's raw and filtered accel, we should probably check this in the safety, do you know what it does @haraschax? `AccPrpl_A_Rq` should be gas, `AccBrkTot_A_Rq` should be brake.

sshane: https://github.com/commaai/panda/pull/1776

PR #28949, opened 2023-07-14
test_models: panda safety tx test

sshane:
sanity checks on the panda safety for sending standard openpilot messages.

asserts:
- [x] We send all messages with empty CC message (nothing should be active)
- [x] Cancel message works (in correct state)
- [x] Resume message works


selfdrive/car/tests/test_models.py
```diff
@@ -201,6 +201,42 @@ def test_panda_safety_rx_valid(self):
           self.assertTrue(self.safety.addr_checks_valid())
     self.assertFalse(len(failed_addrs), f"panda safety RX check failed: {failed_addrs}")
 
+  def test_panda_safety_tx_cases(self, data=None):
+    """Asserts we can tx common messages"""
```
adeebshihadeh, line 205: why is notCar skipped?

selfdrive/car/tests/test_models.py
```diff
@@ -201,6 +201,42 @@ def test_panda_safety_rx_valid(self):
...
+    if self.CP.notCar:
+      self.skipTest("Skipping test for notCar")
```
adeebshihadeh, line 207: why are not cars skipped?

selfdrive/car/tests/test_models.py
```diff
@@ -201,6 +201,42 @@ def test_panda_safety_rx_valid(self):
...
+        _, sendcan = self.CI.apply(car_control, _now_nanos)
+        _now_nanos += 10e6
+        msgs_sent += len(sendcan)
+        for addr, _, dat, bus in sendcan:
+          to_send = libpanda_py.make_CANPacket(addr, bus % 4, dat)
+          self.assertTrue(self.safety.safety_tx_hook(to_send), (addr, dat, bus))
+
+      # Make sure we attempted to send messages
+      self.assertGreater(msgs_sent, 50)
+      return _now_nanos
```
adeebshihadeh, line 222: no need to return or pass this in, just use a nonlocal

selfdrive/car/tests/test_models.py
```diff
@@ -201,6 +201,42 @@ def test_panda_safety_rx_valid(self):
...
     self.assertFalse(len(failed_addrs), f"panda safety RX check failed: {failed_addrs}")
 
+  def test_panda_safety_tx_cases(self, data=None):
+    """Asserts we can tx common messages"""
+    if self.CP.notCar:
+      self.skipTest("Skipping test for notCar")
+
+    def test_car_controller(car_control, _now_nanos):
+      msgs_sent = 0
+      for _ in range(300):  # make sure we hit the slowest messages
```
adeebshihadeh, line 211: use DT_CTRL, would also probably make this run longer. maybe 10s?

selfdrive/car/tests/test_models.py
adeebshihadeh: or even better, make a new self.CI so that it's scope is just within the function

adeebshihadeh (approved):

adeebshihadeh:
Looks good, just needs some cleanup

selfdrive/car/tests/test_models.py
sshane: openpilot and safety both assume that `controls_allowed` will always effectively be true. if it's false, it blocks all messages, even zero torque ones, so we need to fix that first before enabling

PR #28956, opened 2023-07-15
Ford: add missing engine FW for 2021 Explorer

sshane:
From f166567f78fad32f|2023-07-14--19-31-27

We removed an ECU it was using for fuzzy FP in https://github.com/commaai/openpilot/pull/28493


selfdrive/car/ford/values.py
```diff
@@ -195,6 +195,7 @@ def init_make(self, CP: car.CarParams):
       b'MB5A-14C204-RC\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00',
       b'NB5A-14C204-AZD\x00\x00\x00\x00\x00\x00\x00\x00\x00',
       b'NB5A-14C204-HB\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00',
+      b'A714B7D487084AB7B15\x00\x00\x00\x00\x00',
```
sshane, line 198: @incognitojam know why the FW looks so different? It's a Floridian 2021 Explorer

sshane: Oh, is it the real camera? https://www.ebay.com/itm/203267707228

incognitojam: I don't think that's the LKAS camera. If it's park assist and "camera" it's probably to do with the IPMB or reversing camera..

incognitojam: It could be caused by an engine tune.

I did do a full scan for ECUs, and there are more things like the gateway module and infotainment system. My car doesn't have a lot of extras like a reversing camera so I wouldn't have seen that :/

incognitojam: do the extra_ecus get queried _after_ fingerprinting is complete? if they do then it's worth adding them I suppose

sshane: No, they're queried at the same time.

sshane: @incognitojam did you hear anything back from the user on Discord?

incognitojam: nothing beyond our initial conversation

incognitojam: User reports that it has an engine tune

incognitojam: The user used to have this engine FW: `b'MB5A-14C204-MD\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00'`

Maybe we just add this and ask the user to revert their tune to use openpilot

incognitojam: oh, it's already in there!

incognitojam:
User's original (un-tuned) engine FW is already in the code.

PR #28957, opened 2023-07-15
ui/map: update destination on param change

### LINKED ISSUE #28826: nav entry: destination not updated on param change

deanlee:
resolve: https://github.com/commaai/openpilot/issues/28826#issuecomment-1636520930

### DIFF
--- a/selfdrive/ui/qt/maps/map_settings.cc
+++ b/selfdrive/ui/qt/maps/map_settings.cc
@@ -2,7 +2,6 @@
 
 #include <QApplication>
 #include <QDebug>
-#include <QFileSystemWatcher>
 
 #include "common/util.h"
 #include "selfdrive/ui/qt/request_repeater.h"
--- a/selfdrive/ui/qt/maps/map_settings.cc
+++ b/selfdrive/ui/qt/maps/map_settings.cc
@@ -300,8 +300,8 @@ NavigationRequest::NavigationRequest(QObject *parent) : QObject(parent) {
       QObject::connect(repeater, &RequestRepeater::requestDone, this, &NavigationRequest::parseLocationsResponse);
     }
     {
-      auto file_watcher = new QFileSystemWatcher(this);
-      QObject::connect(file_watcher, &QFileSystemWatcher::fileChanged, this, &NavigationRequest::nextDestinationUpdated);
+      auto param_watcher = new ParamWatcher(this);
+      QObject::connect(param_watcher, &ParamWatcher::paramChanged, this, &NavigationRequest::nextDestinationUpdated);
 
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Ready for review?

deanlee:
Yeah. But QFileSystemWatcher::fileChanged will be emitted twice (one for content change, one for last modified time changed) when the param  changes. This causes the list to be cleared and refreshed twice.

selfdrive/ui/qt/maps/map_settings.cc
```diff
@@ -297,6 +298,9 @@ NavigationRequest::NavigationRequest(QObject *parent) : QObject(parent) {
       QObject::connect(repeater, &RequestRepeater::requestDone, this, &NavigationRequest::parseLocationsResponse);
     }
     {
+      auto file_watcher = new QFileSystemWatcher(this);
+      QObject::connect(file_watcher, &QFileSystemWatcher::fileChanged, this, &NavigationRequest::nextDestinationUpdated);
```
adeebshihadeh, line 302: can we make a custom param watcher so this is easier to reuse?

adeebshihadeh:
> y. But QFileSystemWatcher::fileChanged will be emitted twice (one for content change, one for last modified time changed) when the param changes. This causes the list to be cleared and refreshed twice.

Maybe we can a custom Param watcher class that only emits this once, but I agree that the clearing should be fixed anyway.

Also, maybe we should move the definition of `clearLayout` to the wifi settings so no one else is tempted to use it? then we can get rid of it entirely after that.

> I had planned to open this pr after #28961 was REVIEWED.

@incognitojam can you review and merge both of these tomorrow?

adeebshihadeh:
Thanks @deanlee!

adeebshihadeh:
Still doesn't seem to work completely. I suspect it's not adding back the watcher sometimes after the param is deleted?

deanlee:
`. I suspect it's not adding back the watcher sometimes after the param is deleted?`

y, how about not remove "NavDestination", just set it to empty? 
Otherwise we can only monitor the entire `params/d` directory and call `param_watcher->addParam("NavDestination");` after the NavDestination is recreated. this is very resource consuming.

adeebshihadeh:
We definitely don't want to monitor the whole directory. And I don't like just setting it to empty; it's effectively making a second implementation of Params::remove to enforce behavior elsewhere. 


Thinking about this more, the bug is when the destination is removed, so I don't think we need either of those anyway. Seems like we're just dropping the delete event sometimes?

deanlee:
created a pr to fix this sync issue: https://github.com/commaai/openpilot/pull/29103

PR #28958, opened 2023-07-15
nav: show green path when using navigate on openpilot

sshane:


selfdrive/ui/qt/maps/map.h
```diff
@@ -110,6 +110,9 @@ class MapWindow : public QOpenGLWidget {
   QPushButton *settings_btn;
   QPixmap directions_icon, settings_icon;
 
+  // Blue with normal nav, green when nav is input into the model
+  QColor getNavPathColor(bool nav_enabled) { return nav_enabled ? QColor("#31ee73") : QColor("#31a1ee"); }
```
adeebshihadeh, line 114: use multiple lines here, too much going on

adeebshihadeh (changes requested):

selfdrive/ui/qt/maps/map.cc
```diff
@@ -105,7 +105,11 @@ void MapWindow::initLayers() {
     nav["type"] = "line";
     nav["source"] = "navSource";
     m_map->addLayer(nav, "road-intersection");
-    m_map->setPaintProperty("navLayer", "line-color", QColor("#31a1ee"));
+
+    QVariantMap transition;
+    transition["duration"] = 400;  // 400 ms color transition for navigate on openpilot
```
sshane, line 110:
```suggestion
    transition["duration"] = 400;  // ms
```

PR #28961, opened 2023-07-15
ui/MapSettings: remove call to  `clearLayout`

deanlee:


incognitojam:
There seems to be a long delay after tapping a destination before the UI updates and it appears set

deanlee:
I forgot that there is a mapSetting widget in the home-_-b, so I deleted the updateCurrentRoute call from navigateTo.  restored  https://github.com/commaai/openpilot/pull/28961/commits/57f7455014f7a14020a986bde437fadc27a6bfaa

PR #28962, opened 2023-07-15
VW MQB: Add FW for 2023 Skoda Scala

Saber422:
route name：0bbe367c98fa1538|2023-07-15--11-42-25--0


sshane:
@jyoung8607 know if we can bump MY to 2023?

jyoung8607:
Yes, extending to 2020-23 should be fine.

PR #28968, opened 2023-07-16
loggerd: config remote encoders according to the received encoder data

deanlee:
this PR is split from https://github.com/commaai/openpilot/pull/28893: (I'm trying to split it into small PRs, the next pr will remove the recursive call to `handle_encoder_msg` if this one can be merged.)

1. remove the "hardcoded" `max_waiting`.
2. keep an encoder map in `LoggerdState` which is initialized from the received encoder data.
3. Add variable `encoder_info` to  `RemoteEncoder` to reduce function parameters.


adeebshihadeh:
FYI, I likely won't have time to review this until 0.9.4 is released. I recommend waiting to split up the PR further until this one is reviewed.

deanlee:
ok i see.

PR #28969, opened 2023-07-16
boardd: misc cleanup

adeebshihadeh:


deanlee:
There are also outdated includes in boardd. some may be left over from the c language version of boardd.

adeebshihadeh:
Want to open a PR to clean them up?

PR #28974, opened 2023-07-16
Ford: add F-150 2023 support (dashcam only)

incognitojam:
VIN: `1FTFW1E5XPFA32044`

**Checklist**
- [ ] ~~added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs~~
- [ ] ~~test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)~~
- [x] route with openpilot: `09136c309ba9461d|2023-07-16--11-18-53`
- [ ] route with stock system: ``

All FW segment: `09136c309ba9461d|2023-07-16--11-11-21--0`


selfdrive/car/ford/values.py
```diff
@@ -96,17 +106,57 @@ def init_make(self, CP: car.CarParams):
...
+    Request(
+      [StdQueries.TESTER_PRESENT_REQUEST, StdQueries.MANUFACTURER_SOFTWARE_VERSION_REQUEST],
+      [StdQueries.TESTER_PRESENT_RESPONSE, StdQueries.MANUFACTURER_SOFTWARE_VERSION_RESPONSE],
+      bus=0,
+      auxiliary=True,
     ),
   ],
   extra_ecus=[
-    (Ecu.shiftByWire, 0x732, None),
+    (Ecu.debug, 0x703, None),             # All Wheel Drive (AWD)
```
adeebshihadeh, line 123: looks good aside from these new requests. are they strictly needed ti get driving or just for debugging?

incognitojam: Just for debugging, I can remove them now

selfdrive/car/ford/values.py
```diff
@@ -96,13 +101,17 @@ def init_make(self, CP: car.CarParams):
...
       [StdQueries.TESTER_PRESENT_REQUEST, StdQueries.MANUFACTURER_SOFTWARE_VERSION_REQUEST],
       [StdQueries.TESTER_PRESENT_RESPONSE, StdQueries.MANUFACTURER_SOFTWARE_VERSION_RESPONSE],
       bus=0,
-      whitelist_ecus=[Ecu.eps, Ecu.abs, Ecu.fwdRadar, Ecu.fwdCamera, Ecu.shiftByWire],
+    ),
+    Request(
+      [StdQueries.TESTER_PRESENT_REQUEST, StdQueries.MANUFACTURER_SOFTWARE_VERSION_REQUEST],
+      [StdQueries.TESTER_PRESENT_RESPONSE, StdQueries.MANUFACTURER_SOFTWARE_VERSION_RESPONSE],
+      bus=0,
+      auxiliary=True,
```
sshane, line 114: Why remove the whitelisting?

sshane: Also I'm fine with adding more debugging/logging ECUs for data collection and improved fingerprinting, but that should be in a new PR.

incognitojam: The F-150 was only getting responses on bus 1, so I had to either add all the ECUs to the whitelist or just remove it. I see now though that it would've been worth investigating this further - `09136c309ba9461d|2023-07-15--18-24-31` in `errorLogMessage`:
- `iso-tp query bad response`
- `iso-tp query timeout after receiving partial response`

sshane: Adding all ECUs to the new query is the way to go, if you didn't want to affect CAN cars.

incognitojam: Oh, I misunderstood what auxiliary meant then. I thought it would query on bus 0 of the auxiliary panda, not "global bus 0" when an auxiliary panda is present?

selfdrive/car/ford/values.py
```diff
@@ -44,11 +44,12 @@ class CAR:
   BRONCO_SPORT_MK1 = "FORD BRONCO SPORT 1ST GEN"
   ESCAPE_MK4 = "FORD ESCAPE 4TH GEN"
   EXPLORER_MK6 = "FORD EXPLORER 6TH GEN"
+  F_150_MK14 = "FORD F-150 14TH GEN"
   FOCUS_MK4 = "FORD FOCUS 4TH GEN"
   MAVERICK_MK1 = "FORD MAVERICK 1ST GEN"
 
 
-CANFD_CARS: Set[str] = set()
+CANFD_CARS = {CAR.F_150_MK14}
```
sshane, line 52: can we rename to `CANFD_CAR`?

incognitojam: e7c858d590f84c4f476ec0a01eae67f49f7a114b

PR #28982, opened 2023-07-16
navd: publish navRoute periodically

adeebshihadeh:
so we get at least one per whole segment


sshane:
Any reason not to always publish it? We can clean up some code in map.cc if so

sshane:
Not final solution, track issue here: https://github.com/commaai/openpilot/issues/29026

PR #28986, opened 2023-07-17
Hyundai safety: unify bus definition

sunnyhaibin:
- Prerequisite of https://github.com/commaai/openpilot/issues/28046
- Requires https://github.com/commaai/panda/pull/1518


adeebshihadeh:
Closing this since there's no openpilot diff.

PR #28996, opened 2023-07-17
FPv2: don't multiplex auxiliary pandas

sshane:


selfdrive/car/ford/values.py
```diff
@@ -116,6 +116,12 @@ def init_make(self, CP: car.CarParams):
       bus=0,
       auxiliary=True,
     ),
+    Request(
+      [StdQueries.TESTER_PRESENT_REQUEST, StdQueries.MANUFACTURER_SOFTWARE_VERSION_REQUEST],
+      [StdQueries.TESTER_PRESENT_RESPONSE, StdQueries.MANUFACTURER_SOFTWARE_VERSION_RESPONSE],
+      whitelist_ecus=[Ecu.eps, Ecu.abs, Ecu.fwdRadar, Ecu.fwdCamera, Ecu.engine, Ecu.shiftByWire],
+      auxiliary=True,
+    ),
```
sshane, line 124: Added a test to catch this. This is adding a multiplexed query on bus 5 that would never return (since OBD port is connected to internal panda at bus 1)

sshane:
should be fixed in a better way

PR #28998, opened 2023-07-17
FPv2: speed up FW query with auxiliary pandas

sshane:
Idea is when you have an aux panda, we add queries for bus 0 and bus 4. And if there's an aux panda, there's no need to try bus 0, it's redundant and doesn't return anything.

TODO:
- [ ] handle this in the FwQueryConfig? It's created well before we know the available pandas. Maybe a class helper function that returns appropriate requests


sshane:
https://github.com/commaai/openpilot/issues/28999#issuecomment-1689346331

Issue #28777, opened 2023-07-02
Check for update bug


MJR1989:
### Describe the bug

Upon plugging in my comma 3 it automatically checks for update before wifi is connected and gives me the error “Immediately connect to the internet to check for updates. If you do not connect to the internet, openpilot won't engage in 1 day.” even though I just updated to 0.9.3 yesterday….device does not do this on old versions of openpilot only when updated to 0.9.3 tried system reset and reinstalling 0.9.3 still gave this error....then downloaded Erich's 092 branch and the error doesn't happen anymore 
![IMG_9997](https://github.com/commaai/openpilot/assets/96713527/75a37ddd-5b6b-41cd-a16c-2f3b5e122622)

### Provide a route where the issue occurs

b0a7a22dea55a1cd

### openpilot version

0.9.3

SikeItsMike:
I have the same error and so do a lot of other people from what I'm reading on discord. People seem to be attributing it to an error in the master rather than forks

adeebshihadeh:
This will be fixed in the upcoming release. You can try the `nightly` branch in the meantime if you want to give it a try; you can install with `installer.comma.ai/commaai/nightly`.

adeebshihadeh:
This should be fixed in the upcoming release. If you want to give it a try in the meantime, you can run the `nightly` branch with this URL: `installer.comma.ai/commaai/nightly`.

PR #29004, opened 2023-07-18
Updated: fix setting update time

jnewb1:
Addresses:
https://github.com/commaai/openpilot/issues/28777


jnewb1:
seems like some kind of race condition on accessing the panda

jnewb1:
can no longer reproduce the update issue with this change. still want to look into the panda set time issue.

jnewb1:
@adeebshihadeh this seems to fix the update issue, I'll look into the panda set time issue in another PR (causes boot logs to have wrong time)

adeebshihadeh:
What exactly is happening? Why does it show 1 day?

jnewb1:
> What exactly is happening? Why does it show 1 day?

failed_count is 0 on first loop. LastUpdateTime gets set to incorrect system time. Because failed_count is zero

```elif dt.days > DAYS_NO_CONNECTIVITY_MAX and failed_count > 1``` evaluates false

and ```max(DAYS_NO_CONNECTIVITY_MAX - dt.days, 1)``` when dt.days is >14 is 1

File: selfdrive/updated.py [hunk 1]
```diff
@@ -437,8 +437,8 @@ def main() -> None:
       init_overlay()
 
       # ensure we have some params written soon after startup
-      updater.set_params(update_failed_count, exception)
       update_failed_count += 1
+      updater.set_params(update_failed_count, exception)
```
adeebshihadeh, line 441: let's fix this in a more explicit way, it's super easy to write this bug again

selfdrive/updated.py
```diff
@@ -319,7 +319,7 @@ def get_description(basedir: str) -> str:
       set_offroad_alert("Offroad_UpdateFailed", True, extra_text=extra_text)
     elif dt.days > DAYS_NO_CONNECTIVITY_MAX and failed_count > 1:
       set_offroad_alert("Offroad_ConnectivityNeeded", True)
-    elif dt.days > DAYS_NO_CONNECTIVITY_PROMPT:
+    elif dt.days > DAYS_NO_CONNECTIVITY_PROMPT and failed_count > 1:
```
adeebshihadeh, line 322: same thing here, easy to write again. move both of these after a new "else if failed_count > 1"

adeebshihadeh (changes requested):
Ah, it's clipped. Some notes about this:
* needs a more descriptive title
* let's also add a check at the top of the main loop on [time validity](https://github.com/commaai/openpilot/blob/master/common/time.py#L3), no point in running if the time isn't valid

selfdrive/updated.py
```diff
@@ -426,49 +432,56 @@ def main() -> None:
...
-      init_overlay()
-
-      # ensure we have some params written soon after startup
-      updater.set_params(update_failed_count, exception)
-      update_failed_count += 1
-
-      # check for update
-      params.put("UpdaterState", "checking...")
-      updater.check_for_update()
+    if valid_datetime(datetime.datetime.now()): # only check for updates if we have a valid system time
```
adeebshihadeh, line 445: this PR is massive now. can you just do a continue here instead?

common/time.py
```diff
@@ -1,3 +1,6 @@
 import datetime
 
 MIN_DATE = datetime.datetime(year=2023, month=6, day=1)
+
+def valid_datetime(d: datetime.datetime):
+  return d > MIN_DATE
```
adeebshihadeh, line 6: this helper is about the same length as "d > MIN_DATE" but lacks the locality of that. if you want a helper, think it would be better to make one that just checks if the current time is valid

adeebshihadeh (commented):
Can you minimize the changes in this PR?

jnewb1:
I think it's more clear now.

- dont change LastUpdateTime unless we had a successful update
- dont show either type of warning unless we had a failed update (failed count > 0)
- dont actually run the updater if we dont yet have a valid system time

selfdrive/updated.py
```diff
@@ -437,9 +439,13 @@ def main() -> None:
       init_overlay()
 
       # ensure we have some params written soon after startup
-      updater.set_params(update_failed_count, exception)
+      updater.set_params(False, update_failed_count, exception)
       update_failed_count += 1
```
adeebshihadeh, line 443: shouldn't be incremented if time invalid

common/time.py
```diff
@@ -1,3 +1,6 @@
 import datetime
 
 MIN_DATE = datetime.datetime(year=2023, month=6, day=1)
+
+def valid_system_time():
+  return datetime.datetime.now() > MIN_DATE
```
adeebshihadeh, line 6: think "if system_time_valid" reads a little better

selfdrive/updated.py
```diff
@@ -437,9 +439,13 @@ def main() -> None:
...
 
+      if not valid_system_time(): # updated should wait for valid system time before trying to update
```
adeebshihadeh, line 445: comment doesn't really add anything here

jnewb1: removed

common/time.py
jnewb1: done

selfdrive/updated.py
jnewb1: done

selfdrive/updated.py
```diff
@@ -437,7 +439,12 @@ def main() -> None:
       init_overlay()
 
       # ensure we have some params written soon after startup
-      updater.set_params(update_failed_count, exception)
+      updater.set_params(False, update_failed_count, exception)
+
+      if not system_time_valid():
+        wait_helper.sleep(5)
```
adeebshihadeh, line 445: updater.set_params writes to disk, so let's not do this too often
```suggestion
        wait_helper.sleep(60)
```

adeebshihadeh (approved):

jnewb1:
agnos needs updating to test this further (need python 3.11), but i tested it quite a bit before the cleanup, I think its good to merge

PR #29005, opened 2023-07-18
Revert "experimental mode: gate navigate on openpilot (#28883)"

sshane:


sshane:
Going to re-do with no change regarding longitudinal control

PR #29011, opened 2023-07-18
non-pcmCruise: initialize to vEgoCluster

### LINKED ISSUE #28446: ui: fudge set speed to match fudged ego speed if not pcmCruise

sshane:
Should fix when you enable openpilot while traveling at x mph (as-shown on the device), and it sets the speed to x-y mph (y some small offset) when vEgoCluster is set to the car's dash and not pcmCruise

### DIFF
--- a/selfdrive/controls/lib/drive_helpers.py
+++ b/selfdrive/controls/lib/drive_helpers.py
@@ -138,7 +138,7 @@ class VCruiseHelper:
     if any(b.type in (ButtonType.accelCruise, ButtonType.resumeCruise) for b in CS.buttonEvents) and self.v_cruise_kph_last < 250:
       self.v_cruise_kph = self.v_cruise_kph_last
     else:
-      self.v_cruise_kph = int(round(clip(CS.vEgo * CV.MS_TO_KPH, initial, V_CRUISE_MAX)))
+      self.v_cruise_kph = int(round(clip(CS.vEgoCluster * CV.MS_TO_KPH, initial, V_CRUISE_MAX)))
 
     self.v_cruise_cluster_kph = self.v_cruise_kph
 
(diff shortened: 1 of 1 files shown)


pfeiferj:
@sshane I can confirm that this pr sets the max speed to the currently displayed speed. On master my car tends to set to about 1-2 mph off from the currently displayed speed when going around 60 mph, but with this change it does properly set to the shown speed. Hyundai Santa Fe 2020 (2019)

sshane:
I realized this will cause your car to speed up and show a vEgo higher than the set speed.

vEgo is 60 mph
vEgoCluster is 62 mph

Press set, cruise set to 62 mph.

Cruise is compared to real vEgo in long planner: 62 mph set speed vs 60 mph, so we speed up.

End result: 62 mph set speed, 62 mph real speed, maybe 64 mph cluster speed.

@pfeiferj did you see this?

pfeiferj:
> I realized this will cause your car to speed up and show a vEgo higher than the set speed.
> 
> vEgo is 60 mph vEgoCluster is 62 mph
> 
> Press set, cruise set to 62 mph.
> 
> Cruise is compared to real vEgo in long planner: 62 mph set speed vs 60 mph, so we speed up.
> 
> End result: 62 mph set speed, 62 mph real speed, maybe 64 mph cluster speed.
> 
> We almost need to use vEgoCluster to control against.
> 
> @pfeiferj did you see this?

Yes, mentioned it in my other PR. This would still be needed once you have adjusted for the car speed when limiting though. Basically my PR that I made for the issue did the limiting thing you're mentioning but didn't fix the initial set of max like this does

pfeiferj:
For the limiting part you could take a look at: https://github.com/commaai/openpilot/pull/28595

PR #29016, opened 2023-07-18
Ford: test FW query config and versions

incognitojam:
Rules on the FW version formatting from here: https://forscan.org/forum/viewtopic.php?p=70008#p70008


selfdrive/car/ford/tests/test_ford.py
```diff
@@ -0,0 +1,81 @@
...
+  ],
+}
+
+
+class TestFordFW(unittest.TestCase):
+  def test_fw_query_config(self):
+    for (ecu, addr, subaddr) in FW_QUERY_CONFIG.extra_ecus:
+      self.assertIn(ecu, ECU_ADDRESSES, "Unknown ECU")
+      self.assertEqual(addr, ECU_ADDRESSES[ecu], "ECU address mismatch")
+      self.assertIsNone(subaddr, "Unexpected ECU subaddress")
```
sshane, line 49: For this kind of test, I think it should be explicitly in the FwQueryConfig at some point, defining all your Ecus that you are able to use, we don't test this anywhere else right now, so don't know if we need it

sshane:
Can we merge https://github.com/commaai/openpilot/pull/28956 after this?

incognitojam:
> Does this block https://github.com/commaai/openpilot/pull/28956 after this?
Yeah it should! I'm pretty confident all the versions will follow this format.

sshane:
Oh, I think we should support those users, but what do you think @adeebshihadeh?

selfdrive/car/ford/tests/test_ford.py
```diff
@@ -0,0 +1,81 @@
...
+
+  @parameterized.expand(FW_VERSIONS.items())
+  def test_fw_versions(self, car_model: str, fw_versions: Dict[Tuple[capnp.lib.capnp._EnumModule, int, Optional[int]], Iterable[bytes]]):
+    self.assertIn(car_model, CAR.__dict__.values())
```
sshane, line 53:
```suggestion
```

this should be a general test, not ford specific

selfdrive/car/ford/tests/test_ford.py
```diff
@@ -0,0 +1,80 @@
...
+      # Prefix changes based on the family of part. It includes the model year
+      #   and likely the platform.
+      # Core identifies the type of the item (e.g. 14D003 = PSCM, 14C204 = PCM).
+      # Suffix specifies the version of the part. -AA would be followed by -AB.
+      #   Small increments in the suffix are usually compatible.
+      # Details: https://forscan.org/forum/viewtopic.php?p=70008#p70008
+      for fw in fws:
+        self.assertEqual(len(fw), 24, "Expected ECU response to be 24 bytes")
+
+        # TODO: parse with regex
```
sshane, line 68:
```suggestion
        # TODO: parse with regex, don't need detailed error message
```

PR #28864, opened 2023-07-10
athenad: test connection timeout

incognitojam:
- test athena disconnects in a reasonable time after connection is lost (due to network change)
- test recovery and receives a ping after connection resumes

Findings:
- websocket takes 60 seconds before notifying that the connection is closed. I'm not sure what the mechanism is for this.

Related:
- https://github.com/commaai/openpilot/pull/26141

PR #29017, opened 2023-07-18
athena: stricter socket timeout when onroad

incognitojam:
continuation of #26141, with tests #28864

When onroad, the time to socket timeout after a network change is reduced to ~25s. The config used for offroad is the same as the current defaults.

### DIFF
--- a/selfdrive/athena/athenad.py
+++ b/selfdrive/athena/athenad.py
@@ -771,7 +771,7 @@ def ws_manage(ws: WebSocket, end_event: threading.Event) -> None:
       onroad = onroad_prev
 
       sock.setsockopt(socket.IPPROTO_TCP, socket.TCP_KEEPIDLE, 5 if onroad else 30)      # s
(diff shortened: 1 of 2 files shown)


common/params.cc
```diff
@@ -129,6 +129,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"HasAcceptedTerms", PERSISTENT},
     {"IMEI", PERSISTENT},
     {"InstallDate", PERSISTENT},
+    {"IsDeviceAwake", CLEAR_ON_MANAGER_START},
```
adeebshihadeh, line 132: don't think we need a new param for this, IsOffroad/IsOnroad should be sufficient

adeebshihadeh (commented):
does this improve the reconnect time in the test? if the tests show this is good, let's try to get it in soon

incognitojam:
Tested athena functions (taking snapshot, uploading camera files) and seems to work well on Wi-Fi and LTE

PR #29020, opened 2023-07-18
nav: show green path when enabled

### LINKED ISSUE #29019: nav: show green path when enabled
@deanlee interested in this one? we want to make it only show the green path when controlsState is enabled

sshane:
`getPaintProperty` [does not exist in QT](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/qt/src/qmapboxgl.cpp), and `setPaintProperty` invokes some json. Takes about 0.05 to 0.1 ms to set a color on PC, but this still only sets on change

### DIFF
--- a/selfdrive/ui/qt/maps/map.cc
+++ b/selfdrive/ui/qt/maps/map.cc
@@ -108,7 +108,7 @@ void MapWindow::initLayers() {
 
     QVariantMap transition;
     transition["duration"] = 400;  // ms
-    m_map->setPaintProperty("navLayer", "line-color", getNavPathColor(nav_path_active));
+    m_map->setPaintProperty("navLayer", "line-color", getNavPathColor(uiState()->scene.navigate_on_openpilot));
     m_map->setPaintProperty("navLayer", "line-color-transition", transition);
     m_map->setPaintProperty("navLayer", "line-width", 7.5);
     m_map->setLayoutProperty("navLayer", "line-cap", "round");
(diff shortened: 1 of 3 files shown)


selfdrive/ui/qt/maps/map.cc
```diff
@@ -139,20 +139,25 @@ void MapWindow::updateState(const UIState &s) {
...
-    // set path color on change, and show map on rising edge of navigate on openpilot
     bool nav_enabled = sm["modelV2"].getModelV2().getNavEnabled();
-    if (nav_enabled != uiState()->scene.navigate_on_openpilot) {
-      if (loaded_once) {
-        m_map->setPaintProperty("navLayer", "line-color", getNavPathColor(nav_enabled));
-      }
-      if (nav_enabled) {
-        emit requestVisible(true);
-      }
+    if (nav_enabled && !uiState()->scene.navigate_on_openpilot) {
```
adeebshihadeh, line 145: shouldn't this be the same check as the color?

sshane: I think the interaction would be more confusing that way (ie. the map won't appear until you first engage, then you can hide it again when you disengage)

adeebshihadeh: yeah, I think this is confusing since you don’t care if the model is using the map when you’re not engaged

sshane: changed! it will let you hide map and change color to blue when disengaged, but nav features are being used

adeebshihadeh: yes

sshane: I still like always showing the map, but we can see how it feels in the car

PR #29021, opened 2023-07-18
Update pycapnp and Cython

maxime-desroches:


adeebshihadeh:
Looks like we'll need to re-lock to update one of the packages

adeebshihadeh (approved):
LGTM, let's merge once CI passes

adeebshihadeh:
Looks like model replay is failing because the CL CI image doesn't rebuild the base image. https://github.com/commaai/openpilot/blob/b2d0f74ec3cc48faa6b54c26267d53ac554e140f/.github/workflows/selfdrive_tests.yaml#L316

.github/workflows/selfdrive_tests.yaml
```diff
@@ -305,6 +305,8 @@ jobs:
       with:
         submodules: true
     - uses: ./.github/workflows/setup
+    - name: Build Base Docker image (temp for ci)
+      run: eval "$BUILD"
```
adeebshihadeh, line 309: this shouldn't be temporary. can you make  separate PR to do this after this one is merged?

PR #29022, opened 2023-07-18
Force build of openpilot-base for OpenCL tests

maxime-desroches:


adeebshihadeh:
also remove the pull in BUILD_CL?

PR #29023, opened 2023-07-18
Toyota: add hybrid ECU FW versions for 2023 RAV4 Hybrid

sshane:
Think it's a good idea to start doing this on new platforms, where we don't need to worry about FW dumping due to the small initial user base.


sshane:
This enables fuzzy fingerprinting for `459707e5f696e897` which shares or skips every ECU except the engine (with hybrid that makes two, the minimum).

PR #29027, opened 2023-07-19
process_replay: capture process output

fredyshox:


adeebshihadeh:
Ideally, we don't do this ourselves. [pytest](https://docs.pytest.org/en/7.4.x/) does this really well; perhaps we can integrate with it.

fredyshox:
@adeebshihadeh I could use `CaptureManager` (https://github.com/pytest-dev/pytest/blob/main/src/_pytest/capture.py#L710). Its in internal module `_pytest.capture`.

PR #29032, opened 2023-07-19
navd: fix hysteresis when reaching destination

sshane:
Noticed this when reading the code to start refactoring. Basically, the desired behavior (from the comments) is that you reach your destination, stop, but nothing is cleared until you start driving away:

https://github.com/commaai/openpilot/blob/5cfb79e35ffecf1a37759f22d084fa4aeb959ef4/selfdrive/navd/navd.py#L271-L277

However since we clear the parameter first, we clear the route anyway on the next loop:

https://github.com/commaai/openpilot/blob/5cfb79e35ffecf1a37759f22d084fa4aeb959ef4/selfdrive/navd/navd.py#L102-L105

We can avoid clearing the param until we clear the route and don't need to worry about re-routing due to this check here: 

https://github.com/commaai/openpilot/blob/5cfb79e35ffecf1a37759f22d084fa4aeb959ef4/selfdrive/navd/navd.py#L300-L302


sshane:
However this is odd behavior to show all zeros for the ETA, maybe we should remove this broken code altogether. @adeebshihadeh have any preference?

sshane:
This is the fixed behavior, before this, it would briefly show that you've arrived, then disappear. The nav instructions are aren't perfect (it gets hidden here), so might want to just remove this behavior?

https://github.com/commaai/openpilot/assets/25857203/80d42943-1c7d-4aac-bcc7-a014e72afff8

selfdrive/navd/navd.py
```diff
@@ -269,11 +269,11 @@ def send_instruction(self):
         self.recompute_countdown = 0
       else:
         cloudlog.warning("Destination reached")
-        Params().remove("NavDestination")
 
         # Clear route if driving away from destination
         dist = self.nav_destination.distance_to(self.last_position)
         if dist > REROUTE_DISTANCE:
+          self.params.remove("NavDestination")
```
adeebshihadeh, line 276: move this to clear route instead?

adeebshihadeh (commented):
Remove the distance in the top banner too?

selfdrive/navd/navd.py
sshane: Can't, it's used to temporary hide the route on errors as well. I want to move the param into a capnp field so the ui can use that to know why the route was cleared

sshane:
https://github.com/commaai/openpilot/assets/25857203/b84f49a2-2398-43c2-a088-5f484eb85c6a

PR #29035, opened 2023-07-19
nav: show warning with no internet

sshane:
draft, includes other PRs atm. will get it split out and cleaned up


sshane:
done in https://github.com/commaai/openpilot/pull/29055

PR #29037, opened 2023-07-19
ui/networking: remove call to clearLayout

deanlee:
the `cabana/OpenPandaWidget` is still using clearLayout. so I haven't removed function clearLayout in this pr yet, just added a `[[deprecated]] `attribute to it.


selfdrive/ui/qt/util.h
```diff
@@ -18,6 +18,8 @@ QString getBrand();
 QString getUserAgent();
 std::optional<QString> getDongleId();
 QMap<QString, QString> getSupportedLanguages();
+
+[[deprecated]]
 void clearLayout(QLayout* layout);
```
adeebshihadeh, line 23: let's move it to cabana instead, want to make sure nothing else in the openpilot ui uses it

adeebshihadeh (changes requested):

selfdrive/ui/qt/offroad/networking.cc
```diff
@@ -279,76 +277,88 @@ WifiUI::WifiUI(QWidget *parent, WifiManager* wifi) : QWidget(parent), wifi(wifi)
...
+}
+
+void WifiItem::setItem(const Network &n, const QPixmap &status_icon, bool show_forget_btn, const QPixmap &strength_icon) {
+  network = n;
+
+  ssidLabel->setText(n.ssid);
+  ssidLabel->setEnabled(n.security_type != SecurityType::UNSUPPORTED);
+  ssidLabel->setProperty("disconnected", network.connected == ConnectedType::DISCONNECTED);
+  ssidLabel->style()->unpolish(ssidLabel);
+  ssidLabel->style()->polish(ssidLabel);
```
adeebshihadeh, line 356: this seems like it would be slow. we should probably get rid of this at some point?

adeebshihadeh (commented):
Noticeable improvement! Want to take a stab at #29375 next?

sshane:
this breaks the consistent spacing when there's no icon

image

PR #29043, opened 2023-07-19
process_replay: fix radard replays when radarUnavailable

fredyshox:
Most (all?) radar interfaces, fake the desired radar frequency with `time.sleep` if `carParams.radarUnavailable` is true. Since process_replay sets `NO_RADAR_SLEEP`, it needs to control the frequency by buffering can msgs.

### DIFF
--- a/selfdrive/test/process_replay/process_replay.py
+++ b/selfdrive/test/process_replay/process_replay.py
@@ -337,6 +337,29 @@ def dmonitoringmodeld_rcv_callback(msg, cfg, frame):
   return msg.which() == "driverCameraState"
 
 
+class RadardRcvCallback:
+  def __init__(self):
+    self.radar_available = None
+    self.frequency = None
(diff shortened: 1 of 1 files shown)


haraschax:
Is this the only process like this? Maybe we can just have radard follow modeld? I think that probably makes more sense?

adeebshihadeh:
> Is this the only process like this? Maybe we can just have radard follow modeld? I think that probably makes more sense?

That would be nice to have regardless of process replay if we're ok with the extra latency.

fredyshox:
> If we change that, @fredyshox this will be a new pattern for process replay: poll on modelV2 and drain on can.

Gonna look into that, but I believe the current ReplayContext should be able to accommodate that, if it would set up events for `modelV2` and leave `can` be.

fredyshox:
Refactor of radard #29063 will make this unnecessary

PR #29049, opened 2023-07-20
Fix destination pin offset

### LINKED ISSUE #29029: ui map: center destination pin and fix overlap with path

mitchellgoffpc:


sshane:
New marker is a few pixels larger, but otherwise LGTM! Coord: `32.718869, -117.154729`


image

sshane:
Actually there's a way to do this with your other PR: https://github.com/commaai/openpilot/pull/29050

sshane:
Done by now setting icon anchor: https://github.com/commaai/openpilot/pull/29054

PR #29050, opened 2023-07-20
nav: draw destination pin on top of navigation path

### LINKED ISSUE #29029: ui map: center destination pin and fix overlap with path

mitchellgoffpc:


sshane:
With my ordering change, looks good! Need to hide it when the route clears though, I'll do that

image

sshane:
I'm not sure why, but the `visibility` layout property always has a fade in transition, even if we don't update updateSource *and* set a global transition duration of 0 ms (`m_map->setTransitionOptions(0, 0);`) *and* set the `icon-opacity-transition`, nothing removed the slow fade in with `visibility`. Must be a bug with symbol type layers and/or the visibility property as it doesn't happen with `navRoute` or `carPosLayer`. However, if you comment out the line that updates the car pos source continually, it fades just like the pin layer (`m_map->updateSource("carPosSource", carPosSource);`)

There's also no mention of a visibility transition on https://docs.mapbox.com/mapbox-gl-js/style-spec/layers/#paint-line-line-blur so my work around was to use the `icon-opacity` paint property with a 0 ms duration transition.

https://github.com/commaai/openpilot/assets/25857203/b4ab2bd4-77e4-46c0-bc28-104f8dfd9b6e

selfdrive/ui/qt/maps/map.cc
```diff
@@ -112,7 +112,27 @@ void MapWindow::initLayers() {
     m_map->setPaintProperty("navLayer", "line-color-transition", transition);
     m_map->setPaintProperty("navLayer", "line-width", 7.5);
     m_map->setLayoutProperty("navLayer", "line-cap", "round");
-    m_map->addAnnotationIcon("default_marker", QImage("../assets/navigation/default_marker.svg"));
+  }
+  if (!m_map->layerExists("pinLayer")) {
+    qDebug() << "Initializing pinLayer";
+    m_map->addImage("default_marker", QImage("../assets/navigation/default_marker.svg"));
+
+    QVariantMap pin;
```
sshane, line 120:
```suggestion
    QVariantMap pin;
```

sshane:
@deanlee let me know if you have any ideas why visibility has an animation here, I couldn't figure it out looking at the source code.

PR #29055, opened 2023-07-20
nav: show warning for routing problems

### LINKED ISSUE #28492: nav ui: overlays disappear if first reroute attempt fails

sshane:
v2 of https://github.com/commaai/openpilot/pull/29035

I don't want to show "Re-routing" or similar here, because when you leave the route and _have_ internet, it just silently updates the `navRoute` and we can't know to show an alert. Plus this covers API exceptions from no internet and empty route responses, so I'm catching-all with a "Waiting for internet" error for now.

We can make more specific messages in the future when we add more of the state to `navRoute`.

---

On receipt of a valid `navInstruction` packet, we reset the warning (it's only valid when step_size isn't None, and that's true when we get a valid response from the API)

https://github.com/commaai/openpilot/blob/d0ea8be95fd5900930070b55c8582bafd329639e/selfdrive/navd/navd.py#L211-L214

https://github.com/commaai/openpilot/blob/d0ea8be95fd5900930070b55c8582bafd329639e/selfdrive/navd/navd.py#L187-L196


sshane:
image

@cydia2020 do you know any alternatives here?

cydia2020:
ｲﾝﾀﾈｯﾄを待機中 should work.

sshane:
great! can you open a PR?

PR #29057, opened 2023-07-20
Toyota: Enable Resume OP w/ Brake Hold Active

jsegill:
Toyota Bug Fix- Allows OP to be resumed while the Brake Hold is Active. The changes prevent the integrator from continuing to ramp up. 

Tested on route 4ecfcac756318cbf|2023-07-03--09-32-23--0 and verified via log data in Cabana. 

Route: 4ecfcac756318cbf|2023-07-03--09-32-23--0

Attempt at fixing issue:  Unable to activate OP while brake hold active commaai/openpilot#33988


sshane:
I believe disallowing engagement when the auto brake hold feature is on is an intentional decision.

PR #29063, opened 2023-07-20
radard: tie radard frequency to modelV2

fredyshox:


selfdrive/car/toyota/radar_interface.py
```diff
@@ -42,17 +42,22 @@ def __init__(self, CP):
     self.rcp = None if CP.radarUnavailable else _create_radar_can_parser(CP.carFingerprint)
     self.trigger_msg = self.RADAR_B_MSGS[-1]
     self.updated_messages = set()
+    self.f = open("/tmp/radar_data.txt", "w")
```
haraschax, line 45: This intentional?

fredyshox: Oh no, that’s some leftover from debugging. Good catch!

morrislee:
Would you be able to implement Kalman Filter for the vision lead data?

selfdrive/controls/radard.py
```diff
@@ -218,26 +229,32 @@ def update(self, sm: messaging.SubMaster, rr: Optional[car.RadarData]):
     if sm.updated['modelV2']:
       self.ready = True
 
-    ar_pts = {}
-    for pt in radar_points:
-      ar_pts[pt.trackId] = [pt.dRel, pt.yRel, pt.vRel, pt.measured]
+    if not no_radar:
```
haraschax, line 232: not not always makes my had spin. Why not just give rr a proper name and do "If RadarSomething is not None:"

selfdrive/controls/radard.py
```diff
@@ -26,6 +26,8 @@
 RADAR_TO_CENTER = 2.7   # (deprecated) RADAR is ~ 2.7m ahead from center of car
 RADAR_TO_CAMERA = 1.52  # RADAR is ~ 1.5m ahead from center of mesh frame
 
+RADAR_STATE_FREQUENCY = 20.  # Hz
```
haraschax, line 29: Use DT_MDL from common/realtime for this

haraschax:
@morrislee  The kalman filter is mostly there to get accel from speed. This is possible with radar because radar speeds are very accurate. Not the case for vision

haraschax:
Besides comments LGTM if test results look good on 20k segments

PR #29068, opened 2023-07-20
ui: new nav settings button

### LINKED ISSUE #28830: nav: better settings button design
- [ ] the first time you view the map without a destination, it shows settings button then quickly changes to the nav icon (or vice-versa), with the positional changes and everything

_Originally posted by @sshane in https://github.com/commaai/openpilot/issues/28826#issuecomment-1624214004_

sshane:
- shows map settings on first click
- toggles between views subsequent clicks
- hides itself when alert it showing
- [x] DM icon moves


selfdrive/ui/qt/maps/map_panel.cc
```diff
@@ -33,3 +33,11 @@ MapPanel::MapPanel(const QMapboxGLSettings &mapboxSettings, QWidget *parent) : Q
   });
   content_stack->addWidget(settings);
 }
+
+void MapPanel::toggleMapSettings() {
+  // show settings if not visible, then toggle between map and settings
+  int new_index = isVisible() ? (1 - content_stack->currentIndex()) : 1;
+  content_stack->setCurrentIndex(new_index);
```
adeebshihadeh, line 40: always better to use setCurrentWidget if we can

adeebshihadeh (approved):

selfdrive/ui/qt/maps/map_panel.cc
sshane: Can make that change in another PR since the pattern is used in multiple places (and the widgets are different types so no ternaries without casting)

Issue #29056, opened 2023-07-20
ui: update last GPS location on offroad transition


sshane:
or onroad. we only read it on `MapWindow` creation, so if you park your car and come back before it restarts, it should show the previous stored location until you get GPS (nobody's noticed?)

https://github.com/commaai/openpilot/blob/d0ea8be95fd5900930070b55c8582bafd329639e/selfdrive/ui/qt/maps/map.cc#L75-L78

PR #29069, opened 2023-07-20
Nav: update last location on offroad/onroad transition

jnewb1:
https://github.com/commaai/openpilot/issues/29056


sshane:
oh now that I'm reading the code, we never reset last_position. it's continually updated from llk. so probably not an issue

PR #29078, opened 2023-07-21
ui/MapSettings: remove function `shorten`

deanlee:
Calling `shorten` is not necessary. the `ElidedLabel` will elide text too long to fit the label.


incognitojam:
Tested, works fine! Thanks!

Issue #29074, opened 2023-07-21
ui: sort non-favorite nav destinations by max of api time, device time


mitchellgoffpc:
Currently we show the non-favorite destinations in the map settings in the order they're returned from the API, which corresponds to how recently they were chosen in connect. However, this doesn't take into account how recently you chose each destination through the map settings, so common destinations won't float to the top until you select them again in connect. To fix this, I propose we do the following:

* [x] Keep a param with a list of recent destinations and their "last navigated" timestamps. When a destination is chosen in the map settings, that destination's entry in the list is updated with the current timestamp, or added to the list if it doesn't exist yet.
* [x] Update the sort order in map_settings.cc for non-favorite destinations to `max(location["timestamp"], dest_timestamps.get(location["place_name"], 0))`, where `dest_timestamps` is the param mapping recent destinations to the last time they were selected in the menu (defaulting to 0 for destinations not in the list)
* I'm not sure if it makes more sense to keep the param as a CSV or JSON or something else, JSON would be nice but it sounds like there are some locale issues with json11 and qt json is kind of a pain.
* [ ] We can update the API to return timestamps for each location if it doesn't already.

PR #29079, opened 2023-07-21
ui/MapSettings: sort destinations by recent activity

deanlee:
1. related https://github.com/commaai/openpilot/issues/29074 . sort by last used time descending.
2. refactor `NavigationRequest` to `NavManager`.  Move all params/nav related operations into `NavManager`. `MapSettings` is only responsible for UI stuffs. this will also reduce some IO operations.
3. fix syc issue with param `NavDestination`


common/params.cc
```diff
@@ -156,6 +156,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"LiveTorqueParameters", PERSISTENT | DONT_LOG},
     {"LongitudinalPersonality", PERSISTENT},
     {"NavDestination", CLEAR_ON_MANAGER_START | CLEAR_ON_OFFROAD_TRANSITION},
+    {"NavDestinations", PERSISTENT},
```
adeebshihadeh, line 159: let's call this `NavPastDestinations`, this one looks too similar to NavDestination

selfdrive/ui/qt/maps/map_settings.cc
```diff
@@ -290,6 +287,10 @@ NavigationRequest *NavigationRequest::instance() {
 }
 
 NavigationRequest::NavigationRequest(QObject *parent) : QObject(parent) {
+  if (auto v = params.get("NavDestinations"); !v.empty()) {
```
adeebshihadeh, line 290: use multiple lines here

incognitojam:
thanks!

PR #29080, opened 2023-07-21
dmonitoringmodeld: initialize model first

### LINKED ISSUE #28919: dmonitoringmodeld never started running
<s>Likely stuck in the VisionIPC connect loop</s> actually it could also be stuck in the model init, which was refactored recently. added some logging around that #28920
```
83b2b6de0cb09d7f|2023-07-13--10-10-54
```

sshane:
Prevents two processes from initializing and building their models at the same (ie. dmonitoringmodeld and navmodeld)

### DIFF
--- a/selfdrive/modeld/navmodeld.cc
+++ b/selfdrive/modeld/navmodeld.cc
@@ -45,7 +45,9 @@ int main(int argc, char **argv) {
   // there exists a race condition when two processes try to create a
   // SNPE model runner at the same time, wait for dmonitoringmodeld to finish
   LOGW("waiting for dmonitoringmodeld to initialize");
-  Params().getBool("DmModelInitialized", true);
+  if (!Params().getBool("DmModelInitialized", true)) {
+    return 0;
+  }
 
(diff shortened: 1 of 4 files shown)


selfdrive/modeld/navmodeld.cc
```diff
@@ -41,6 +42,11 @@ void run_model(NavModelState &model, VisionIpcClient &vipc_client) {
 int main(int argc, char **argv) {
   setpriority(PRIO_PROCESS, 0, -15);
 
+  // there exists a race condition when two processes try to create a
+  // SNPE model runner at the same time, wait for dmonitoringmodeld to finish
+  LOGW("waiting for dmonitoringmodeld to initialize");
+  Params().getBool("DmModelInitialized", true);
```
deanlee, line 48: getBool(param, block) will return false if SIGINT caught. 
it might be better to exit main if it return false.

sshane:
this PR:
image

master:
image

sshane:
More events:

image

PR #29084, opened 2023-07-21
ui/map: hide icon if no maneuver type

deanlee:


sshane:
It looks like there's one navInstruction packet that's valid with no text, type, but distance is still present when you reach your destination. Fixing that in navd sounds like the real fix (since it still shows the distance here, right?). Did you have any other cases where type is empty?

image

deanlee:
I'm not sure what else would cause maneuver to be empty. I didn't check the packages in the logs. just thought it would better/safer to  complete the if-else check.

sshane:
Is icon_01 reset somewhere? Does it look like that if there is no type, it will show the previous icon? Or does that realistically never happen?

sshane:
it's added conditionally in parse_banner_step, so it's possible it's empty based on the API response

PR #29087, opened 2023-07-21
[commabody] setup and run live stream encoders

nuwandavek:


selfdrive/manager/process_config.py
```diff
@@ -35,8 +38,8 @@ def qcomgps(started, params, CP: car.CarParams) -> bool:
   NativeProcess("clocksd", "system/clocksd", ["./clocksd"]),
   NativeProcess("logcatd", "system/logcatd", ["./logcatd"]),
   NativeProcess("proclogd", "system/proclogd", ["./proclogd"]),
-  PythonProcess("logmessaged", "system.logmessaged", offroad=True),
-  PythonProcess("micd", "system.micd"),
+  PythonProcess("logmessaged", "system.logmessaged", offroad=True, callback=iscar),
```
adeebshihadeh, line 41: what's this for?

selfdrive/manager/process_config.py
```diff
@@ -35,8 +38,8 @@ def qcomgps(started, params, CP: car.CarParams) -> bool:
...
+  PythonProcess("micd", "system.micd", callback=iscar),
```
adeebshihadeh, line 42: same here

nuwandavek: @haraschax Was this because we want microphone control outside of `micd`?

selfdrive/manager/process_config.py
nuwandavek: I think this was to stuff the yolo message in this. We don't need this anymore.

system/loggerd/loggerd.h
```diff
@@ -69,10 +71,32 @@ const EncoderInfo main_driver_encoder_info = {
...
 };
 
+const EncoderInfo stream_road_encoder_info = {
+  .publish_name = "livestreamRoadEncodeData",
+  .encode_type = cereal::EncodeIndex::Type::QCAMERA_H264,
+  .record = false,
+  .bitrate = LIVESTREAM_BITRATE,
+  INIT_ENCODE_FUNCTIONS(LivestreamRoadEncode),
+};
+const EncoderInfo stream_wide_road_encoder_info = {
```
adeebshihadeh, line 81: there should be new lines between these EncoderInfos

selfdrive/manager/process_config.py
haraschax: yes

haraschax:
Not sure we can have the streaming encoders running in regular openpilot, would need to test or have it run on only body

nuwandavek:
`stream_encoderd` needs a timing check (failed jenkins)

haraschax:
LGTM

haraschax:
Make sure jenkins passes before merging though

system/loggerd/loggerd.h
```diff
@@ -26,11 +28,10 @@ const int SEGMENT_LENGTH = LOGGERD_TEST ? atoi(getenv("LOGGERD_SEGMENT_LENGTH"))
 
 constexpr char PRESERVE_ATTR_NAME[] = "user.preserve";
 constexpr char PRESERVE_ATTR_VALUE = '1';
-
 class EncoderInfo {
 public:
   const char *publish_name;
-  const char *filename;
+  const char *filename = "";
```
adeebshihadeh, line 34:
```suggestion
  const char *filename = NULL;
```

system/loggerd/encoderd.cc
```diff
@@ -147,14 +148,21 @@ void encoderd_thread() {
...
     ret = util::set_core_affinity({3});
     assert(ret == 0);
   }
-  encoderd_thread();
-  return 0;
+  if (argc > 1) {
+    std::string arg1(argv[1]);
+    if (arg1 == "--stream") {
+      encoderd_thread(stream_cameras_logged);
+    }
```
adeebshihadeh, line 163: add an else that shows a good error message. currently will exit silently with a good exit code

selfdrive/test/test_onroad.py
```diff
@@ -28,6 +28,7 @@
   "selfdrive.controls.controlsd": 39.0,
   "./loggerd": 14.0,
   "./encoderd": 17.0,
+  "stream_encoderd": 17.0,
```
adeebshihadeh, line 31: this shouldn't be necessary since this test runs as a car, see https://github.com/commaai/openpilot/blob/8076107b4fa1b3c2b7c14a8464aa92e949e2e139/selfdrive/test/test_onroad.py#L261C114-L261C114

adeebshihadeh (changes requested):

adeebshihadeh:
I also changed the title to make it more specific. Can you make sure to do that next time?

selfdrive/manager/process_config.py
```diff
@@ -35,12 +38,13 @@ def qcomgps(started, params, CP: car.CarParams) -> bool:
...
   NativeProcess("proclogd", "system/proclogd", ["./proclogd"]),
   PythonProcess("logmessaged", "system.logmessaged", offroad=True),
-  PythonProcess("micd", "system.micd"),
+  PythonProcess("micd", "system.micd", callback=iscar),
   PythonProcess("timezoned", "system.timezoned", enabled=not PC, offroad=True),
 
   DaemonProcess("manage_athenad", "selfdrive.athena.manage_athenad", "AthenadPid"),
   NativeProcess("dmonitoringmodeld", "selfdrive/modeld", ["./dmonitoringmodeld"], enabled=(not PC or WEBCAM), callback=driverview),
   NativeProcess("encoderd", "system/loggerd", ["./encoderd"]),
+  NativeProcess("stream_encoderd", "system/loggerd", ["./encoderd", "--stream"], callback=notcar),
```
nuwandavek, line 47: @adeebshihadeh should `onroad=False` here, like it for the `webjoystick` process?

adeebshihadeh: it should match webjoystick

sshane:
@nuwandavek micd is now running offroad at all times because of this change

PR #29095, opened 2023-07-22
loggerd: check `userFlag` outside the while loop

deanlee:


system/loggerd/loggerd.cc
```diff
@@ -190,7 +190,7 @@ void loggerd_thread() {
   typedef struct QlogState {
     std::string name;
     int counter, freq;
-    bool encoder;
+    bool encoder, user_flag;
```
adeebshihadeh, line 193: this has nothing to do with qlogs. rename to ServiceState?

adeebshihadeh (changes requested):

PR #29096, opened 2023-07-22
swaglog: enable type-checked against format string.

deanlee:
enable type-check & correct format string.

added a macro `SWAG_LOG_CHECK_FMT` to causes the compiler to check the arguments in calls to `cloudlog` for consistency with the `printf` style format string argument:
```

system/ubloxd/ublox_msg.cc:393:57: error: format specifies type 'int' but the argument has type 'uint64_t' (aka 'unsigned long') [-Werror,-Wformat]
      LOGE("SV_ID != SLOT_NUMBER: %d %d", msg->sv_id(), data->n());
                                     ~~                 ^~~~~~~~~
                                     %lu
```


deanlee:
failed to pass the checks. the sub modules have wrong format string.

system/ubloxd/ublox_msg.cc
```diff
@@ -390,7 +390,7 @@ kj::Array<capnp::word> UbloxMsgParser::parse_glonass_ephemeris(ubx_t::rxm_sfrbx_
     eph.setP4(data->p4());
     eph.setSvURA(glonass_URA_lookup.at(data->f_t()));
     if (msg->sv_id() != data->n()) {
-      LOGE("SV_ID != SLOT_NUMBER: %d %d", msg->sv_id(), data->n());
+      LOGE("SV_ID != SLOT_NUMBER: %d %" PRIu64, msg->sv_id(), data->n());
```
adeebshihadeh, line 393: why not llu?

adeebshihadeh (commented):
can you open the matching rednose PR?

deanlee:
> can you open the matching rednose PR?

opened at : https://github.com/commaai/rednose/pull/31

system/ubloxd/ublox_msg.cc
deanlee: use llu will report error on my laptop.  the uint64_t is not the same as unsigned long long.  https://stackoverflow.com/questions/32198368/unsigned-long-long-conflict-with-uint64-t

the PRIu64 can correctly format uint64_t as lu or llu on different platforms

deanlee: another way is use llu, and convert uint64_t to (unsigned long long):

LOGE("SV_ID != SLOT_NUMBER: %d %llu", msg->sv_id(), (unsigned long long)data->n());

PR #29102, opened 2023-07-23
CI: run plotjuggler test on mac

adeebshihadeh:


adeebshihadeh:
Added this because I was hitting that pycurl issue again, but looks like it works in CI. @fredyshox any idea why it's happening?


```
macbookpro:plotjuggler adeeb$ ./juggle.py 
  File "/Users/adeeb/openpilot/tools/plotjuggler/./juggle.py", line 15, in <module>
  File "/Users/adeeb/openpilot/tools/lib/logreader.py", line 11, in <module>
    from tools.lib.filereader import FileReader
  File "/Users/adeeb/openpilot/tools/lib/filereader.py", line 2, in <module>
    from tools.lib.url_file import URLFile
  File "/Users/adeeb/openpilot/tools/lib/url_file.py", line 8, in <module>
    import pycurl
ImportError: pycurl: libcurl link-time version (7.85.0) is older than compile-time version (7.87.0)
```

PR #29103, opened 2023-07-23
ui/map: fix destination sync issue after removing `NavDestination`

deanlee:
1. always `emit nextDestinationUpdated()` after `addParam("NavDestination");`  to ensure that the current destination can be updated after param 'NavDestination' is recreated.
~2. set  `NavDestination` to empty instead of remove it  after user clicks "remove current destination" button.~

### DIFF
--- a/selfdrive/ui/qt/maps/map_settings.cc
+++ b/selfdrive/ui/qt/maps/map_settings.cc
@@ -60,7 +60,7 @@ MapSettings::MapSettings(bool closeable, QWidget *parent) : QFrame(parent) {
   current_widget = new DestinationWidget(this);
   QObject::connect(current_widget, &DestinationWidget::actionClicked, [=]() {
     if (current_destination.empty()) return;
-    params.remove("NavDestination");
+    params.put("NavDestination", "");
     updateCurrentRoute();
   });
   frame->addWidget(current_widget);
@@ -298,6 +298,7 @@ NavigationRequest::NavigationRequest(QObject *parent) : QObject(parent) {
     }
     {
       auto param_watcher = new ParamWatcher(this);
+      param_watcher->addParam("NavDestination");
       QObject::connect(param_watcher, &ParamWatcher::paramChanged, this, &NavigationRequest::nextDestinationUpdated);
 
       // Destination set while offline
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/maps/map_settings.cc
```diff
@@ -60,7 +60,7 @@ MapSettings::MapSettings(bool closeable, QWidget *parent) : QFrame(parent) {
   current_widget = new DestinationWidget(this);
   QObject::connect(current_widget, &DestinationWidget::actionClicked, [=]() {
     if (current_destination.empty()) return;
-    params.remove("NavDestination");
+    params.put("NavDestination", "");
```
adeebshihadeh, line 63: this is effectively a second implementation of Params::remove that just happens to make the file watcher work. can we just remove from this PR for now?

adeebshihadeh:
I don't understand how this fixes it, though I can see how it masks over the problem. Isn't the problem that we're not getting the `paramChanged` signal or not properly updating on it?

deanlee:
> I don't understand how this fixes it, though I can see how it masks over the problem. Isn't the problem that we're not getting the `paramChanged` signal or not properly updating on it?

It is the combination of these two problems. after NavDestination recreated, the next call to `addParam` will add watch on it again, but will not emit the `paramChanged` at this point. We need to manually `emit nextDestinationUpdated();` to update the mapSettings after re-watch on NavDestination

adeebshihadeh:
Why don't we do the fileChanged check when adding a param path?

deanlee:
did you mean check the param value after the very first call to addParam (line 301) ?  because it's called in constructor, mapSetting has not connected to the `nextDestinationUpdated` at this point.  but mapsetting will read param in `showEvent`. So it's not a big problem.
the original mapsetting did not consider these latter requirements. it need sto be refactorted  to something like https://github.com/commaai/openpilot/pull/29079 :

> refactor NavigationRequest to NavManager. Move all params/nav related operations into NavManager. MapSettings is only responsible for UI stuffs. this will also reduce some IO operations.

deanlee:
closed due to https://github.com/commaai/openpilot/pull/29120.

PR #29107, opened 2023-07-23
params: add new flag DONT_REMOVE

deanlee:
add  new flag `DONT_REMOVE` to clear param content instead of deleting it.

### DIFF
--- a/common/params.cc
+++ b/common/params.cc
@@ -156,7 +156,7 @@ std::unordered_map<std::string, uint32_t> keys = {
     {"LiveTorqueCarParams", PERSISTENT},
     {"LiveTorqueParameters", PERSISTENT | DONT_LOG},
     {"LongitudinalPersonality", PERSISTENT},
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
This is better, but still pretty subtle. 

Perhaps `Params::remove` should just have this behavior by default? Though unlink is a lot nicer in most cases. Let me think about this some more.

common/params.cc
```diff
@@ -316,7 +325,12 @@ void Params::clearAll(ParamKeyType key_type) {
       if (de->d_type != DT_DIR) {
         auto it = keys.find(de->d_name);
         if (it == keys.end() || (it->second & key_type)) {
-          unlink(getParamPath(de->d_name).c_str());
+          if (it != keys.end() && (it->second & DONT_REMOVE)) {
+            // clear the content
+            fclose(fopen(getParamPath(de->d_name).c_str(), "w"));
```
deanlee, line 330: this line needs to be refactored. it's not safe and atomic for reader.

deanlee:
This pr adds complexity. Closed for now.

PR #29116, opened 2023-07-24
Subaru: remove 2020+ temporary steering faults

### LINKED ISSUE #25311: Subaru: EPS faults on 2020+ models at low speeds
EPS seems to temporarily fault relatively frequently below 35mph.

jnewb1:
You often get very annoying steer warnings at low speeds for 2020+ cars. I've found that this is because of a high steering angle rate, which you often get a low speeds. Needs a bit more testing, but works pretty well for our outback. Haven't found a way around the 90 degree warning.

depends: https://github.com/commaai/openpilot/pull/29140


selfdrive/car/subaru/carstate.py
```diff
@@ -47,6 +50,16 @@ def update(self, cp, cp_cam, cp_body):
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(can_gear, None))
 
     ret.steeringAngleDeg = cp.vl["Steering_Torque"]["Steering_Angle"]
+
+    # have not found a steering rate message. calculate it manually
+    STEERING_FREQUENCY = 2
+    ret.steeringRateDeg = (ret.steeringAngleDeg - self.prev_angle) / (DT_CTRL * STEERING_FREQUENCY)
```
sshane, line 56: have you looked for any bits that change before a steering fault occurs? might want to look up to half a second back. I have this script which I modify, for example change the comparison to a variable like `bit_to_find` and set that appropriately above: https://gist.github.com/sshane/a7965aefd6462186edc6152ccf203eed#file-find_torque_sign_bit-py-L63

selfdrive/car/subaru/values.py
```diff
@@ -553,3 +553,5 @@ class SubaruCarInfo(CarInfo):
 
 GLOBAL_GEN2 = (CAR.OUTBACK, CAR.LEGACY)
 PREGLOBAL_CARS = (CAR.FORESTER_PREGLOBAL, CAR.LEGACY_PREGLOBAL, CAR.OUTBACK_PREGLOBAL, CAR.OUTBACK_PREGLOBAL_2018)
+
+STEER_LIMITED = (*GLOBAL_GEN2, CAR.IMPREZA_2020)
```
sshane, line 557: Can we make these into sets?

selfdrive/car/subaru/carcontroller.py
```diff
@@ -32,13 +37,31 @@ def update(self, CC, CS, now_nanos):
...
+        else:
+          self.steer_rate_counter = 0
+        
+        if self.steer_rate_counter > MAX_STEER_RATE_FRAMES:
+          apply_steer_req = 0
+          self.steer_rate_counter = 0
+
+        # Any steering past 90 appears to cause temp fault
+        if abs(CS.out.steeringAngleDeg) > MAX_STEER_ANGLE:
+          apply_steer_req = 0
```
sshane, line 59: we have a few safety modes which use this same logic, might be good to abstract it in another PR

jnewb1:
Would resolve: https://github.com/commaai/openpilot/issues/25311

selfdrive/car/subaru/carstate.py
jnewb1: I've done some searching in cabana but I'll look at making a script

selfdrive/car/subaru/values.py
jnewb1: done

selfdrive/car/subaru/carstate.py
jnewb1: I looked with a script and couldn't find anything. In techinfo, I can see that one ES cancel condition is "steering angle changed excessively", but in ssm4 I don't see a "steering rate" message on either the EPS or ES, so my assumption is that both the EPS and ES calculate these manually, ES typically cancels before the steer fault, and since we don't cancel, the EPS faults.

selfdrive/car/subaru/carcontroller.py
jnewb1: made one here: https://github.com/commaai/openpilot/pull/29140

jnewb1:
here's a route with a bunch of low speed driving, high rate of steering angle change, and a bunch of disengage / engages past 90. no steering faults:

c3d1ccb52f5f9d65|2023-07-28--15-05-15

selfdrive/car/subaru/carstate.py
```diff
@@ -47,6 +50,11 @@ def update(self, cp, cp_cam, cp_body):
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(can_gear, None))
 
     ret.steeringAngleDeg = cp.vl["Steering_Torque"]["Steering_Angle"]
+    
+    if self.car_fingerprint in STEER_LIMITED_2020:
+      # have not found a steering rate message. calculate it manually
+      ret.steeringRateDeg = self.angle_rate_calulator.update(ret.steeringAngleDeg, cp.vl["Steering_Torque"]["COUNTER"])
```
sshane, line 56: I would feel a bit more comfortable if we could detect a limit in the steering wheel angle delta over some time, since we haven't standardized calculating the steering wheel rate just yet, and any change this calculation would change fault workaround. Is there anything like > 10 deg in a second? we can use?

sshane:
Are there any routes where you've triggered the fault many times over and over?

jnewb1:
> Are there any routes where you've triggered the fault many times over and over?

@sshane c3d1ccb52f5f9d65|2023-07-22--01-23-20

around 20 instances on this route where I was doing longitudinal control

selfdrive/car/subaru/carcontroller.py
```diff
@@ -32,13 +37,25 @@ def update(self, CC, CS, now_nanos):
...
+        apply_steer_req = 0
+
+      if self.CP.carFingerprint in STEER_LIMITED_2020:
+        # Steering rate fault prevention
+        self.steer_rate_counter, apply_steer_req = \
+          common_fault_avoidance(CS.out.steeringRateDeg, MAX_STEER_RATE, apply_steer_req,
+                                 self.steer_rate_counter, MAX_STEER_RATE_FRAMES)
+        
+        # >90 degree steering fault prevention
+        _, apply_steer_req = common_fault_avoidance(CS.out.steeringAngleDeg, MAX_STEER_ANGLE, apply_steer_req, 0, 0, 0)
```
sshane, line 53: Let's only handle one fault at a time

sshane:
I think the 25 deg/s for 7 frames limit might have some false positives, it didn't fault here where it was 89 deg/s over a 0.21 sec time period (could the limit be the same as Toyota? 100 deg/s?):

image

I would use a LogReader (MultiLogIterator for multiple segments) on your route above to test some hypothesis, that's how I found the exact limits of Toyota (or as close as I could get them at the time).

selfdrive/car/subaru/carstate.py
```diff
@@ -47,6 +50,11 @@ def update(self, cp, cp_cam, cp_body):
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(can_gear, None))
 
     ret.steeringAngleDeg = cp.vl["Steering_Torque"]["Steering_Angle"]
+    
+    if self.car_fingerprint in STEER_LIMITED_2020:
```
adeebshihadeh, line 54: let's set this for all, no reason not to

selfdrive/car/subaru/carstate.py
```diff
@@ -47,6 +50,11 @@ def update(self, cp, cp_cam, cp_body):
...
+      # have not found a steering rate message. calculate it manually
```
adeebshihadeh, line 55:
```suggestion
      # ideally we get this from the car, but unclear if it exists. diagnostic software doesn't even have it
```

selfdrive/car/subaru/carstate.py
adeebshihadeh, line 56: should be in a separate PR

selfdrive/car/subaru/values.py
```diff
@@ -615,3 +615,4 @@ def init_make(self, CP: car.CarParams):
 LKAS_ANGLE = {CAR.FORESTER_2022, CAR.OUTBACK_2023}
 GLOBAL_GEN2 = {CAR.OUTBACK, CAR.LEGACY, CAR.OUTBACK_2023}
 PREGLOBAL_CARS = {CAR.FORESTER_PREGLOBAL, CAR.LEGACY_PREGLOBAL, CAR.OUTBACK_PREGLOBAL, CAR.OUTBACK_PREGLOBAL_2018}
+STEER_LIMITED_2020 = GLOBAL_GEN2 | {CAR.IMPREZA_2020}
```
adeebshihadeh, line 618: why is 2020 in the name?

jnewb1: its 2020+ cars that have this limit

selfdrive/car/subaru/carcontroller.py
```diff
@@ -32,13 +37,25 @@ def update(self, CC, CS, now_nanos):
...
+      apply_steer_req = 1
+
       if not CC.latActive:
         apply_steer = 0
+        apply_steer_req = 0
+
+      if self.CP.carFingerprint in STEER_LIMITED_2020:
+        # Steering rate fault prevention
+        self.steer_rate_counter, apply_steer_req = \
+          common_fault_avoidance(CS.out.steeringRateDeg, MAX_STEER_RATE, apply_steer_req,
```
sshane, line 49: i'd leave a todo saying that detection isn't perfect and to improve it

selfdrive/car/subaru/carstate.py
sshane: we decided to allow erroneous detection to fix most faults

selfdrive/car/subaru/values.py
sshane: the 90 degree limit or high angle rate, or both? a comment might be better

jnewb1: 90 deg fault it pretty rare in normal driving but I can check the data. I'll just name it STEER_RATE_LIMITED for now and we can come back to the 90 deg fault later

selfdrive/car/subaru/carcontroller.py
jnewb1: removed 90 deg prevention, we can investigate that later when it becomes more relevant when we get subaru long control

selfdrive/car/subaru/carcontroller.py
jnewb1: done

jnewb1:
image
Here is a normalized plot where x=0 is the frame where the fault occurred. Most of the faults will be removed by this steer rate fault avoidance. However we might want to re investigate it as some point to get closer to the actual limit.

jnewb1:
@sshane if I update the refs, anything else on this?

selfdrive/car/subaru/carcontroller.py
```diff
@@ -1,8 +1,12 @@
...
-from selfdrive.car import apply_driver_steer_torque_limits
+from selfdrive.car import apply_driver_steer_torque_limits, common_fault_avoidance
 from selfdrive.car.subaru import subarucan
-from selfdrive.car.subaru.values import DBC, GLOBAL_GEN2, PREGLOBAL_CARS, CanBus, CarControllerParams, SubaruFlags
+from selfdrive.car.subaru.values import DBC, GLOBAL_GEN2, PREGLOBAL_CARS, CanBus, STEER_RATE_LIMITED, CarControllerParams, SubaruFlags
 
+# These limits aren't exact, and the real limit it more than likely over a larger time period and
+# involves the total steering angle change rather than rate, but these limits work well for now
+MAX_STEER_RATE = 25 # deg/s
+MAX_STEER_RATE_FRAMES = 7  # tx control frames needed before torque can be cut
```
sshane, line 10:
```suggestion
# These limits aren't exact. The real limit is more than likely over a larger time period and
# involves the total steering angle change rather than rate, but these limits work well for now
MAX_STEER_RATE = 25  # deg/s
MAX_STEER_RATE_FRAMES = 7  # tx control frames needed before torque can be cut


```

selfdrive/car/subaru/carcontroller.py
```diff
@@ -32,13 +37,22 @@ def update(self, CC, CS, now_nanos):
...
+
       if not CC.latActive:
         apply_steer = 0
+        apply_steer_req = 0
+
+      if self.CP.carFingerprint in STEER_RATE_LIMITED:
+        # Steering rate fault prevention
+        self.steer_rate_counter, apply_steer_req = \
+          common_fault_avoidance(CS.out.steeringRateDeg, MAX_STEER_RATE, apply_steer_req,
+                                 self.steer_rate_counter, MAX_STEER_RATE_FRAMES)
```
sshane, line 50: it takes a bool now

sshane (changes requested):

sshane:
Make sure this doesn't regress control in normal situations, since 25 deg/s is rather low, but LGTM.

sshane:
Oh did you add the panda safety yet?

selfdrive/car/subaru/carcontroller.py
```diff
@@ -1,8 +1,13 @@
 from openpilot.common.numpy_fast import clip, interp
 from opendbc.can.packer import CANPacker
-from openpilot.selfdrive.car import apply_driver_steer_torque_limits
+from openpilot.selfdrive.car import apply_driver_steer_torque_limits, common_fault_avoidance
 from openpilot.selfdrive.car.subaru import subarucan
-from openpilot.selfdrive.car.subaru.values import DBC, GLOBAL_GEN2, PREGLOBAL_CARS, HYBRID_CARS, CanBus, CarControllerParams, SubaruFlags
+from openpilot.selfdrive.car.subaru.values import DBC, GLOBAL_GEN2, PREGLOBAL_CARS, HYBRID_CARS, STEER_RATE_LIMITED, CanBus, CarControllerParams, SubaruFlags
+
+# These limits aren't exact. The real limit is more than likely over a larger time period and
```
sshane, line 7:
```suggestion
# FIXME: These limits aren't exact. The real limit is more than likely over a larger time period and
```

selfdrive/car/subaru/carcontroller.py
```diff
@@ -32,13 +38,22 @@ def update(self, CC, CS, now_nanos):
...
+        # Steering rate fault prevention
+        self.steer_rate_counter, apply_steer_req = \
+          common_fault_avoidance(abs(CS.out.steeringRateDeg) > MAX_STEER_RATE, apply_steer_req,
+                                 self.steer_rate_counter, MAX_STEER_RATE_FRAMES)
 
       if self.CP.carFingerprint in PREGLOBAL_CARS:
         can_sends.append(subarucan.create_preglobal_steering_control(self.packer, self.frame // self.p.STEER_STEP, apply_steer, CC.latActive))
       else:
-        can_sends.append(subarucan.create_steering_control(self.packer, apply_steer, CC.latActive))
+        can_sends.append(subarucan.create_steering_control(self.packer, apply_steer, apply_steer_req))
```
sshane, line 56:
```suggestion
        apply_steer_req = CC.latActive
        if self.CP.carFingerprint in STEER_RATE_LIMITED:
          # Steering rate fault prevention
          self.steer_rate_counter, apply_steer_req = common_fault_avoidance(abs(CS.out.steeringRateDeg) > MAX_STEER_RATE,
                         apply_steer_req, self.steer_rate_counter, MAX_STEER_RATE_FRAMES)
        can_sends.append(subarucan.create_steering_control(self.packer, apply_steer, apply_steer_req))
```

Formatting is probably wrong

selfdrive/car/subaru/carcontroller.py
```diff
@@ -32,13 +38,22 @@ def update(self, CC, CS, now_nanos):
...
+
       if not CC.latActive:
         apply_steer = 0
+        apply_steer_req = 0
+
+      if self.CP.carFingerprint in STEER_RATE_LIMITED:
+        # Steering rate fault prevention
+        self.steer_rate_counter, apply_steer_req = \
+          common_fault_avoidance(abs(CS.out.steeringRateDeg) > MAX_STEER_RATE, apply_steer_req,
+                                 self.steer_rate_counter, MAX_STEER_RATE_FRAMES)
```
sshane, line 51:
```suggestion
```

not used except in the non-preglobal section

sshane (changes requested):

selfdrive/car/subaru/carcontroller.py
```diff
@@ -35,10 +41,18 @@ def update(self, CC, CS, now_nanos):
       if not CC.latActive:
         apply_steer = 0
 
+      apply_steer_req = CC.latActive
+
+      if self.CP.carFingerprint in STEER_RATE_LIMITED:
+        # Steering rate fault prevention
+        self.steer_rate_counter, apply_steer_req = \
+          common_fault_avoidance(abs(CS.out.steeringRateDeg) > MAX_STEER_RATE, apply_steer_req,
+                                 self.steer_rate_counter, MAX_STEER_RATE_FRAMES)
```
sshane, line 50: do any pre-globals use the steer fault prevention?

jnewb1: not current, I can move it under the non-preglobal section

selfdrive/car/subaru/values.py
```diff
@@ -707,3 +707,6 @@ def init_make(self, CP: car.CarParams):
 GLOBAL_GEN2 = {CAR.OUTBACK, CAR.LEGACY, CAR.OUTBACK_2023, CAR.ASCENT_2023}
 PREGLOBAL_CARS = {CAR.FORESTER_PREGLOBAL, CAR.LEGACY_PREGLOBAL, CAR.OUTBACK_PREGLOBAL, CAR.OUTBACK_PREGLOBAL_2018}
 HYBRID_CARS = {CAR.CROSSTREK_HYBRID, CAR.FORESTER_HYBRID}
+
+# cars that appear to be steering angle rate limited. appears to be all cars that started production after 2020
```
sshane, line 711:
```suggestion
# cars that temporarily fault when steering angle rate is greater than some threshold.
# appears to be all cars that started production after 2020
```

sshane (approved):

PR #29118, opened 2023-07-24
[commabody] Support both self-balancing bodies and the wheeled bodies

nuwandavek:
Till we ship out wheels, the self-balancing controller needs to be supported.


adeebshihadeh (approved):
I think this should probably be another fingerprint instead of a param, but still seems worth merging now though. Currently, there's no way to tell the difference in software.

nuwandavek:
SG!

PR #29126, opened 2023-07-24
boardd: handle false positive ignitions on red panda setups

adeebshihadeh:


sshane (approved):

sshane:
I would update the comment at the top however:

```c++
// Ignition:
// - If any of the ignition sources in any panda is high, ignition is high
```

adeebshihadeh:
Don't think that's necessary, since we don't want it to be true for aux pandas generally. I think @robbederks has an aux panda setup for Tesla long that should still work after this.

robbederks:
> Don't think that's necessary, since we don't want it to be true for aux pandas generally. I think @robbederks has an aux panda setup for Tesla long that should still work after this.

In that setup both pandas detect ignition, so I'd be fine with either way.

PR #29134, opened 2023-07-25
ui/OnroadWindow: do not update state when offroad

deanlee:


sshane:
Can you quickly check in OnroadWindow->updateState and nvg->updateState to see if there is anything that depends on running while offroad? And would now miss some some information?

deanlee:
checked. nothing is depends on running updateState while offroad.

PR #29136, opened 2023-07-25
ui/map: delete map after unsubscribing from prime

### LINKED ISSUE #29135: ui/map: handle losing prime
onroad doesn't subscribe to primeTypeChanged to disallow/delete the map and nav settings button. @deanlee want to check this out?

if you buy prime, we only create the map when you go onroad. but if you unsubscribe, not even an offroad transition will block the map

deanlee:
resolve https://github.com/commaai/openpilot/issues/29135


selfdrive/ui/qt/onroad.cc
```diff
@@ -107,6 +108,17 @@ void OnroadWindow::offroadTransition(bool offroad) {
   alerts->updateAlert({});
 }
 
+void OnroadWindow::primeTypeChanged(int prime_type) {
+#ifdef ENABLE_MAPS
+  if (map && !(prime_type || !MAPBOX_TOKEN.isEmpty())) {
```
sshane, line 113:
```suggestion
  if (map && !(prime_type || MAPBOX_TOKEN.isEmpty())) {
```

deanlee:
also need to switch MapSetting in HomeWindow to DriverView

incognitojam:
Tested and the prime -> no prime transition works even if you unsubscribe while onroad (if you set `RequestRepeater->while_onroad = true`)

PR #29138, opened 2023-07-25
ui/onroad: only draw data received after the started_frame

deanlee:


adeebshihadeh:
Perhaps we can refactor this state in such a way that's it's easy to just reset on the offroadTransition?

adeebshihadeh:
Closing until it's in a better state

deanlee:
opened a new pr #29666

PR #29140, opened 2023-07-25
Homogenize Steering Fault Avoidance

jnewb1:
Cleaned up the steering fault avoidance into a common function, since it will be common across a lot of cars.


selfdrive/car/__init__.py
```diff
@@ -132,6 +132,21 @@ def apply_std_steer_angle_limits(apply_angle, apply_angle_last, v_ego, LIMITS):
   return clip(apply_angle, apply_angle_last - angle_rate_lim, apply_angle_last + angle_rate_lim)
 
 
+def common_fault_avoidance(measured_value, max_value, request, current_request_frames=0, max_request_frames=1, cut_request_frames=1):
+  # Count up to max_request_frames, at which point we need to cut the request for cut_request_frames to avoid a fault
```
sshane, line 136: in a """ comment, can you explain how this workaround works? Roughly, the car will hold your torque for a short amount of time if you set the request bit low, and max angle and max angle rate limits are reset when the request bit is low, so we can pulse the bit when we detect we're about to fault

selfdrive/car/__init__.py
```diff
@@ -132,6 +132,21 @@ def apply_std_steer_angle_limits(apply_angle, apply_angle_last, v_ego, LIMITS):
   return clip(apply_angle, apply_angle_last - angle_rate_lim, apply_angle_last + angle_rate_lim)
 
 
+def common_fault_avoidance(measured_value, max_value, request, current_request_frames=0, max_request_frames=1, cut_request_frames=1):
```
sshane, line 135: good idea to type annotate when using generic names

selfdrive/car/hyundai/carcontroller.py
```diff
@@ -94,27 +101,13 @@ def update(self, CC, CS, now_nanos):
...
         can_sends.append([0x7b1, 0, b"\x02\x3E\x80\x00\x00\x00\x00\x00", self.CAN.ECAN])
 
-    # >90 degree steering fault prevention
-    # Count up to MAX_ANGLE_FRAMES, at which point we need to cut torque to avoid a steering fault
-    if CC.latActive and abs(CS.out.steeringAngleDeg) >= MAX_ANGLE:
-      self.angle_limit_counter += 1
-    else:
-      self.angle_limit_counter = 0
-
-    # Cut steer actuation bit for two frames and hold torque with induced temporary fault
```
sshane, line 104: good to keep make-specific details of how it works around in comments

selfdrive/car/tests/test_common.py
```diff
@@ -0,0 +1,77 @@
+import unittest
+
+from selfdrive.car import common_fault_avoidance
+
+class TestCommonFaultAvoidance(unittest.TestCase):
```
sshane, line 5: we already test all this specific behavior in panda, perhaps we should improve our fuzzy test cases in test_models for this

sshane: If there's any test case here that you noticed we don't test in panda, we should add it there @jnewb1

sshane:
I think this is fine as a function, as we have for the torque and angle limits, but I would like to eventually make base car control class, and we can abstract common code. For example, we could have `self.get_steer_request_bit()` that just returns the bit

selfdrive/car/tests/test_common.py
jnewb1: Seems like the panda safety already checks all these scenarios, I'll go ahead and remove the tests.

selfdrive/car/hyundai/carcontroller.py
jnewb1: done

selfdrive/car/__init__.py
jnewb1: done

selfdrive/car/__init__.py
jnewb1: done

PR #29148, opened 2023-07-26
Fixing the US HR-V 2023 docs to indicate that the 8 degree mount is required for the vehicle.

galegozi:


sshane:
It looks like your old routes without the mount are pointed down 8 degrees, but recent ones are pointed up 5 degrees. Likely due to mounting position. Asked in Discord for a mounting photo

PR #29149, opened 2023-07-26
ui/onroad: move duplicated code to `drawIcon()`

deanlee:
The UI is exactly the same as before
before
Screenshot 2023-07-26 14:06:38
after
Screenshot 2023-07-26 14:09:20


sshane:
maybe put it into a base class? or do you think this is cleaner?

deanlee:
think this is cleaner

sshane:
it looks like this breaks something about the circle background (straight line on top and left)

Old:
image


New:
image

deanlee:
fixed

sshane:
ty!

PR #29159, opened 2023-07-26
map: fix exit lane direction when multiple directions

sshane:
We only take the first direction that's true for left, right, then straight. But in cases where a lane can go in multiple directions, we should consider the active direction. (best is to just use an icon that takes into account all the directions, left that TODO)

Before:

After:

navInstruction here:

```
( logMonoTime = 246245457423,
  valid = true,
  navInstruction = (
    maneuverPrimaryText = "I 8 East",
    maneuverSecondaryText = "El Centro",
    maneuverDistance = 136.20595,
    maneuverType = "fork",
    maneuverModifier = "right",
    distanceRemaining = 9351.9414,
    timeRemaining = 576.86774,
    timeRemainingTypical = 646.55939,
    lanes = [
      (directions = [left], active = false, activeDirection = none),
      ( directions = [left, right],
        active = true,
        activeDirection = right ) ],
    showFull = true,
    speedLimit = 0,
    speedLimitSign = mutcd ) )
```

https://docs.mapbox.com/api/navigation/directions/

Still might be wrong for inactive lanes, will check later


sshane:
when not active, we should just make and use more granular icons

PR #29166, opened 2023-07-27
v4l_encoder: fix error handling for poll

deanlee:
The current code does not handle the error if  rc<0.  This causes subsequent code to continue executing when the poll returns an error and may result in undefined behavior.


system/loggerd/encoder/v4l_encoder.cc
```diff
@@ -75,17 +75,27 @@ void V4LEncoder::dequeue_handler(V4LEncoder *e) {
...
-
   // save the header
   kj::Array<capnp::byte> header;
 
   while (!exit) {
+    // POLLIN is capture, POLLOUT is frame
+    struct pollfd pfd = {
+      .events = POLLIN | POLLOUT,
+      .fd = e->fd,
+    };
```
adeebshihadeh, line 86: why move this into the loop?

PR #29171, opened 2023-07-27
aarch64: ubuntu build support

fredyshox:


SConstruct
```diff
@@ -64,13 +64,18 @@ AddOption('--no-test',
           default=os.path.islink(Dir('#laika/').abspath),
           help='skip building test files')
 
+## Architechture name breakdown (arch)
+## - larch64: linux tici aarch64
```
adeebshihadeh, line 68: we want to fix this properly, such that we don't need something called "real_arch". QCOM2 is also used for tici/agnos

fredyshox: This is also badly named, should be called `target` instead. That would leave `arch` for actual CPU architecture

adeebshihadeh: yeah, that's what I mean. feel free to refactor as much of this as you think is necessary; it's all pretty messy right now

fredyshox: This is going to be much bigger than I thought, and requires changes in most of the submodules. I'm going to make separate PR for this

PR #29174, opened 2023-07-28
deleter: improve performance

deanlee:
`listdir_by_creation`&`get_preserved_segments` is very resource consuming if there are a lot of segments on disk.
calling these two functions every time before deleting a directory will impact performance,  especially when `deleter` needs to delete old files continuously after `loggerd` is started when the disk is full.

This pr cache the sorted segments to variable `dirs`. only repopulate it after it becomes empty.

### DIFF
--- a/system/loggerd/deleter.py
+++ b/system/loggerd/deleter.py
@@ -4,6 +4,7 @@ import shutil
 import threading
 from typing import List
 
+from common.params import Params
 from system.swaglog import cloudlog
 from system.loggerd.config import ROOT, get_available_bytes, get_available_percent
 from system.loggerd.uploader import listdir_by_creation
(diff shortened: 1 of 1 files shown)


incognitojam:
It's possible that a segment currently being written could be marked as preserved after the initial check/cache. We could either:
- check again if the segment is preserved immediately before deleting it
- when listing files for deletion, always ignore the most recent? though this might not be the segment we are currently writing if the timestamp is wrong.
any other ideas?

incognitojam:
I think that works as long as the timestamps are correct, which they should be now since thermald will block going onroad with a bad date/time set

deanlee:
updated. 
Check if there is a *.lock file in the directory to exclude the folder that are being written by loggerd. the downside is that building the cache will be slower.

incognitojam:
want to add a test for this?

adeebshihadeh:
We need a test before we can merge this.

adeebshihadeh:
@deanlee thanks for the PR, but gonna close for now. This makes deleter more "bursty" and generally we want to keep the load of our services consistent.

PR #29179, opened 2023-07-29
Optimize Honda Accord stop and go resume too slow issue.

javenzy:
When the leader car is stopped and gone, Honda Accord will take a long time to start to go.Sync vEgoStarting and startAccel parameters from Hyundai to let the car resume quickly.


royjr:
I think `startAccel` is only used when `long_control_state == LongCtrlState.starting` which is only if `startingState` is set

sshane:
Can you post a route with the poor acceleration, then with this PR showing the differences?

javenzy:
> Can you post a route with the poor acceleration, then with this PR showing the differences?

OK, I will post the route to check soon.Thanks!

javenzy:
> I think `startAccel` is only used when `long_control_state == LongCtrlState.starting` which is only if `startingState` is set

yes, I think when my car stop and go following leader car, system will run into the LongCtrlState.starting state.

sshane:
Closing for inactivity.

PR #29183, opened 2023-07-31
Multilang: th translation update

Iamz:
Update Thai translations


sshane:
Btw if you want to get this language in release, look into fixing the Thai fonts in AGNOS. Or let me know if they work already

Iamz:
> Btw if you want to get this language in release, look into fixing the Thai fonts in AGNOS. Or let me know if they work already

It seems fine to me. Is there anything specific I should look for?

sshane:
just checked, it looks good to me now! when we first added thai, the font was blocky. I guess it was fixed in an AGNOS update since then

sshane:
https://github.com/commaai/openpilot/pull/29447

PR #29184, opened 2023-07-31
controls: use common car interface acceleration limits

crwusiz:
Accel param using car.interfaces

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -14,6 +14,9 @@ Ecu = car.CarParams.Ecu
 
 
 class CarControllerParams:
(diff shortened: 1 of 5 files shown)


selfdrive/car/hyundai/values.py
```diff
@@ -14,9 +14,6 @@
 
 
 class CarControllerParams:
-  ACCEL_MIN = -3.5 # m/s
-  ACCEL_MAX = 2.0 # m/s
```
sshane, line 18: This vary per brand (sometimes), and should probably remain inside the port. However, we can maybe make a common or base CarControllerParams to clean up some of the duplication

crwusiz: ok.. brand value revert...

PR #29185, opened 2023-07-31
CarControllerParams is CCP

crwusiz:
VW Same variable name


incognitojam:
Looks like there are a few typos https://github.com/commaai/openpilot/actions/runs/5715610880/job/15485331702?pr=29185

sshane:
think we will leave this as-is for now, at least until we put some work into making a base CarController class

PR #27740, opened 2023-03-30
Honda Bosch Radarless: update stopAccel to match actuation limits

royjr:
This should allow max braking quicker and help with lead collisions.. or so we thought.

PR #25364, opened 2022-08-04
Honda Bosch Radarless: experimental longitudinal support

royjr:
Possible pull for #25355. Works great except stops way too close to the lead car.
[See panda PR](https://github.com/commaai/panda/pull/1008)

PR #29187, opened 2023-07-31
Honda Bosch Radarless: less harsh stops

royjr:
The stopAccel exception, initially introduced in https://github.com/commaai/openpilot/pull/25364 and later modified in https://github.com/commaai/openpilot/pull/27740, was designed to prevent models from crashing into lead vehicles. However, with the current models' improved performance, they can effectively stay behind leads, rendering the stopAccel modification unnecessary. This PR reverts the stopAccel exception and adopts Toyota's stoppingDecelRate for a much smoother, less harsh braking (its currently really bad) experience when the vehicle's speed falls below the vEgoStopping threshold of 0.5.


sshane:
We chose `ret.stopAccel = -4.0` because it was used by the stock system, which we should generally try to match its behavior. This it because we don't have this car to test lower values (for rolling back on hills, or rolling forward in general). It's probably fine to lower, but I think it's better to match the stock system.

I also think you can improve the smoothness just with the stopping decel rate, doesn't it stop before it reaches -2 m/s^2 accel request?

royjr:
engaged_accel_actuator seems to be following the specified rate, but engaged_accel_actual is all over the place. Is this a controls issue?

stock `1a5d045d2c531a6d|2023-07-31--18-24-53`
Screenshot 2023-07-31 at 10 08 19 PM

stoppingDecelRate = 0.2 `1a5d045d2c531a6d|2023-07-31--18-28-00`
Screenshot 2023-07-31 at 10 08 11 PM

selfdrive/car/honda/interface.py
```diff
@@ -75,6 +75,7 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
       ret.longitudinalActuatorDelayUpperBound = 0.5 # s
       if candidate in HONDA_BOSCH_RADARLESS:
         ret.stopAccel = CarControllerParams.BOSCH_ACCEL_MIN  # stock uses -4.0 m/s^2 once stopped but limited by safety model
+        ret.stoppingDecelRate = 0.3  # reach stopping target smoothly
```
sshane, line 78: can also tune vEgoStopping, perhaps it's a bit too high at ~1 mph. can you look at some stock routes and see what it does coming to a stop? at what speed does it really start to ramp down to -4 m/s^2?

royjr: I was going to set it to 0.25 as with gm/toyota but according to these screenshots it looks like 0.3 is a more accurate speed.

Screenshot 2023-08-05 at 10 25 11 AM
Screenshot 2023-08-05 at 10 26 55 AM
Screenshot 2023-08-05 at 10 28 15 AM
Screenshot 2023-08-05 at 10 29 04 AM


```suggestion
        ret.vEgoStopping = 0.3
```



deceleration minimums around 2.5.

Screenshot 2023-08-05 at 10 27 25 AM
Screenshot 2023-08-05 at 10 31 28 AM
Screenshot 2023-08-05 at 10 32 28 AM

`5130484aa8069bad|2023-08-03--16-23-11`

sshane:
this good to review now?

selfdrive/car/honda/interface.py
royjr: Screenshot 2023-08-07 at 2 45 43 PM

`1a5d045d2c531a6d|2023-08-07--13-59-44`

wasn't any better, so currently using `ret.stoppingDecelRate = 0.15` with decent results tested on both downhills and uphills.

selfdrive/car/honda/interface.py
```diff
@@ -75,6 +75,7 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
       ret.longitudinalActuatorDelayUpperBound = 0.5 # s
       if candidate in HONDA_BOSCH_RADARLESS:
         ret.stopAccel = CarControllerParams.BOSCH_ACCEL_MIN  # stock uses -4.0 m/s^2 once stopped but limited by safety model
+        ret.stoppingDecelRate = 0.15  # reach stopping target smoothly
```
sshane, line 78: this could be too slow for hills, I would pick a value that largely improves thing -- it doesn't need to be perfect. then we can more easily merge without testing extremes

royjr: Sounds good, back to 0.3 for now, definitely an improvement without being perfect.

sshane:
did you post any routes with 0.3 on any hills?

TheOfficialJosh:
Hey @sshane while its not a super steep hill, I have two from today's drive. I'll try to find a steeper hill to test on later.
5130484aa8069bad|2023-08-08--09-31-21--11

TheOfficialJosh:
Still couldn't get a harder stop on a hill @sshane, but got a good enough one. It stopped fine and didn't have any issues.
5130484aa8069bad|2023-08-12--20-53-27--6

selfdrive/car/honda/interface.py
sshane: so the stock system sets it to 0.5 well before stopping, then transitions to -4.0 immediately:

image

sshane:
The longitudinal error gains for Honda Bosch are rather low, and I see some consistent error especially while stopping. I'm worried this change masks a tuning issue, so going to close for now. Feel free to try the other tune for Honda with a lower actuator delay (stops later), and higher gains (hopefully slower when it does stop, less jolt).

Here it was off by 1 mph:

image

PR #29194, opened 2023-08-01
ui/onroad: equal margins on imperial

royjr:
This PR rectifies an inconsistency in OpenPilot's top/left margins of the set speed box between imperial and metric, ensuring equal dimensions across both.. unless this was intentional.

After:
Screenshot 2023-07-31 at 10 53 18 PM
Screenshot 2023-07-31 at 10 53 28 PM

Before:
Screenshot 2023-07-31 at 10 54 15 PM
Screenshot 2023-07-31 at 10 54 06 PM

### DIFF
--- a/selfdrive/ui/qt/onroad.cc
+++ b/selfdrive/ui/qt/onroad.cc
@@ -355,7 +355,7 @@ void AnnotatedCameraWidget::drawHud(QPainter &p) {
   int top_radius = 32;
   int bottom_radius = has_eu_speed_limit ? 100 : 32;
 
-  QRect set_speed_rect(QPoint(60 + (default_size.width() - set_speed_size.width()) / 2, 45), set_speed_size);
+  QRect set_speed_rect(QPoint((is_metric ? 60 : 45) + (default_size.width() - set_speed_size.width()) / 2, 45), set_speed_size);
   p.setPen(QPen(whiteColor(75), 6));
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/onroad.cc
```diff
@@ -355,7 +355,7 @@ void AnnotatedCameraWidget::drawHud(QPainter &p) {
   int top_radius = 32;
   int bottom_radius = has_eu_speed_limit ? 100 : 32;
 
-  QRect set_speed_rect(QPoint(60 + (default_size.width() - set_speed_size.width()) / 2, 45), set_speed_size);
+  QRect set_speed_rect(QPoint((is_metric ? 60 : 45) + (default_size.width() - set_speed_size.width()) / 2, 45), set_speed_size);
```
royjr, line 358: We can also match the margins of the other corner elements (experimental, dm, maps) but not sure of they should since they're different shapes.
```suggestion
  QRect set_speed_rect(QPoint((is_metric ? 45 : 30) + (default_size.width() - set_speed_size.width()) / 2, 30), set_speed_size);
```
Screenshot 2023-07-31 at 11 11 17 PM

sshane:
I'm not sure of the original intention, so I think it's better to wait for a re-design or re-work of the spacing of all the widgets.

PR #29206, opened 2023-08-01
third_party: remove curl binaries and headers

fredyshox:
This was related to EON platform, and is currently no longer used


adeebshihadeh:
Nice find!

Issue #29173, opened 2023-07-28
Restarting bridge. Error: rpc::rpc_error during call in function `get_sensor_token`


nathanoy:
### Describe the bug

Hello,

when executing the `bridge.py` it spams this error multiple times a second:
```
WARNING: Version mismatch detected: You are trying to connect to a simulator that might be incompatible with this API 
WARNING: Client API version     = 43b5e70-dirty 
WARNING: Simulator API version  = 0.9.13 
Restarting bridge. Error: rpc::rpc_error during call in function get_sensor_token
```

this occurs when executing either `./start_openpilot_docker.sh` or `./bridge.py` and `./launch_openpilot.sh`

My efforts to make the Client API version match the Simulator API version were unsuccessful.

# Temporary fix:
### Using an [older docker image](https://github.com/commaai/openpilot/pkgs/container/openpilot-sim/110716233), seems to work as a workaround (from before the 21 July 2023 ! )
I have not tested which version exactly introduces the breaking changes, but it seems like the update to Python 3.11 broke the client api compatibility.

### OS Version

Ubuntu 20.04.6 LTS

### openpilot version or commit

f911e5f02869475acfb634a058771390fb2273c7

haormj:
same question, ubuntu 22.04, run sim

dankanti20:
Hello guys, 
Please, I need your help on how to run this simulation. I'm novice.
I have already downloaded the dataset.

nathanoy:
After some more fooling around with the situation, it seems to me like the version change to Python 3.11 broke the client API compatibility.

In this pr #28843 the carla client version is defined to be exactly this:

https://github.com/commaai/openpilot/blob/21802265bec5ede0eaf4093ba0cd2c4c5be3b1af/pyproject.toml#L68C1-L68C1
```toml
carla = { url = "https://github.com/commaai/carla/releases/download/3.11.4/carla-0.9.14-cp311-cp311-linux_x86_64.whl", platform = "linux", markers = "platform_machine == 'x86_64'" }
```
Which seems to be made for carla version 0.9.14, yet the file `tools/sim/start_carla.sh` specifically uses `0.9.13`.
After updating `tools/sim/start_carla.sh` to use carla `0.9.14` the error disappears, only the version warning remains (makes sense).

This seems like a great opportunity to try make my first contribution.

Dragoadri:
THANKS SO MUCH THIS FIXED MY PROBLEM!

PR #29210, opened 2023-08-02
Updated Carla to version 0.9.14

nathanoy:
### Bugfix:

**Description** 
When following [the tools readme](https://github.com/commaai/openpilot/blob/master/tools/README.md) and then [the sim readme](https://github.com/commaai/openpilot/blob/master/tools/sim/README.md), one ends up with the error described in this issue: #29173 (`Restarting bridge. Error: rpc::rpc_error during call in function get_sensor_token`)
The bridge will not start with this error, and just keeps spamming. This Pull Request fixes this issue.

This issue seems to be originating from the switch to Python 3.11 about two weeks ago. In #28843 the Python carla package was updated to be this specific one:
https://github.com/commaai/openpilot/blob/21802265bec5ede0eaf4093ba0cd2c4c5be3b1af/pyproject.toml#L68

This package seems to be there, because there is no carla python package for python 3.11.
The package is for the carla version 0.9.14, but the file `tools/sim/start_carla.sh` specifies 0.9.13.
Simply bumping this version up to 0.9.14 fixes the error mentioned in #29173.
Note: the warning remains, but can safely be ignored versions match after this PR is merged.

**Verification**
I tested to run the simulation two ways:
1. by starting the carla docker (`tools/sim/start_carla.sh`) and then the openpilot docker (`tools/sim/start_openpilot_docker.sh`)
2. by starting the carla docker (`tools/sim/start_carla.sh`) and then openpilot (`tools/sim/launch_openpilot.sh`) followed by the bridge (`tools/sim/bridge.py`).

Both methods executed successfully.

_This is my first contribution, I hope everything is okay. Please don't hesitate to mention anything I can do better next time. Have a lovely day!_

### DIFF
--- a/tools/sim/start_carla.sh
+++ b/tools/sim/start_carla.sh
@@ -15,7 +15,7 @@ if ! $(apt list --installed | grep -q nvidia-container-toolkit); then
   fi
 fi
 
-docker pull carlasim/carla:0.9.13
+docker pull carlasim/carla:0.9.14
 
 EXTRA_ARGS="-it"
 if [[ "$DETACH" ]]; then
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Thanks for the PR!

PR #29216, opened 2023-08-03
athenad: update comment about timeout constant

sshane:
mentioned it was exposed in [3.7 release notes](https://github.com/python/cpython/blob/46366ca0486d07fe94c70d00771482c8ef1546fc/Doc/whatsnew/3.7.rst#L1282), so not sure why we thought it wasn't available (checked a venv and it's there in 3.8.11)

related PR: https://github.com/commaai/openpilot/pull/28864

### DIFF
--- a/selfdrive/athena/athenad.py
+++ b/selfdrive/athena/athenad.py
@@ -43,6 +43,9 @@ from system.swaglog import SWAGLOG_DIR, cloudlog
 from system.version import get_commit, get_origin, get_short_branch, get_version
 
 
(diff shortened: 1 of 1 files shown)


incognitojam:
I think it's just a typing issue, the type is defined conditionally: https://github.com/python/typeshed/blob/9456f5501b564b333891fb474f44f7172c6d5174/stdlib/_socket.pyi#L514

sshane:
yeah it's only available on linux, made a mypy PR (https://github.com/python/mypy/pull/15811). we can ignore for now since we know what the issue is

PR #29226, opened 2023-08-03
Cabana: Fix >1 hour route time displaying

jnewb1:
For routes >1 hour, time is displayed incorrectly, since QT uses the timezone. If you instead specify UTC, it works correctly.
 
Example: c2585539cc5ebcb3|2023-07-29--07-37-37
Before:
image

After:
image


jnewb1:
fromTime_t is also deprecated
https://doc.qt.io/qt-5/qdatetime-obsolete.html

PR #29227, opened 2023-08-03
sensord: Avoid extra copies on sensor init

ianwal:
## Description:

Use emplace_back instead of push_back on sensor init.

This avoids an extra copy by constructing the std::pair in place.

If it were a primitive both would optimize to the same instructions, but std::pair makes it very unlikely to be optimized.

See https://godbolt.org/z/dheqYj611

## Verification:

Currently, I don't have the hardware to test the changes myself.


adeebshihadeh:
Thanks for the PR, but we tend to not merge PRs like this as it effectively makes no difference. Nice job with the linked example, though!

PR #29228, opened 2023-08-03
Nissan: allow B drive mode for Nissan Leaf

cellfreak:
Add B ( Battery save ) drive mode for Nissan Leaf - selfdrive/car/nissan/interface.py

Line 50 -  events = self.create_common_events(ret, extra_gears=[car.CarState.GearShifter.brake])

Road test successful - https://connect.comma.ai/0b17985326a570c4/1691100300431/1691100676848


cellfreak:
cd /data && rm -rf openpilot && git clone --recursive https://github.com/cellfreak/openpilot -b master && sudo reboot

Ran on my Comma Three with no issues.

selfdrive/car/nissan/interface.py
```diff
@@ -47,7 +47,7 @@ def _update(self, c):
     be.type = car.CarState.ButtonEvent.Type.accelCruise
     buttonEvents.append(be)
 
-    events = self.create_common_events(ret)
+    events = self.create_common_events(ret, extra_gears=[car.CarState.GearShifter.brake])
```
sshane, line 50: Checked and this mode is only in the leaf DBC, so we can be sure that it will never occur on other cars

PR #29229, opened 2023-08-03
athenad: comment for TCP user timeout

sshane:


sshane:
observed behavior doesn't really make sense, https://blog.cloudflare.com/when-tcp-sockets-refuse-to-die/ doesn't explain what I'm seeing. closing for now

PR #29230, opened 2023-08-04
Rewrite modeld in python

mitchellgoffpc:


selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,271 @@
...
+      'feature_buffer': np.zeros(HISTORY_BUFFER_LEN * FEATURE_LEN, dtype=np.float32),
+    }
+
+    self.model = ONNXModel("models/supercombo.onnx", self.output, Runtime.GPU, False, context)
+    self.model.addInput("input_imgs", None)
+    self.model.addInput("big_input_imgs", None)
+    for k,v in self.inputs.items():
+      self.model.addInput(k, v)
+
+  def eval(self, buf:VisionBuf, wbuf:VisionBuf, transform:np.ndarray, transform_wide:np.ndarray, inputs:Dict[str, np.ndarray], prepare_only:bool) -> Optional[np.ndarray]:
```
sshane, line 81:
```suggestion
  def eval(self, buf: VisionBuf, wbuf: VisionBuf, transform: np.ndarray, transform_wide: np.ndarray, inputs: Dict[str, np.ndarray], prepare_only: bool) -> Optional[np.ndarray]:
```

think there's always a space here, even though equals don't have spaces

selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,271 @@
...
+
+FEATURE_LEN = 128
+HISTORY_BUFFER_LEN = 99
+DESIRE_LEN = 8
+TRAFFIC_CONVENTION_LEN = 2
+DRIVING_STYLE_LEN = 12
+NAV_FEATURE_LEN = 256
+OUTPUT_SIZE = 5990
+MODEL_OUTPUT_SIZE = 6120
+MODEL_FREQ = 20
```
sshane, line 27: we have DT_MDL in realtime.py

selfdrive/test/test_onroad.py
```diff
@@ -36,7 +36,7 @@
   "selfdrive.locationd.paramsd": 9.0,
   "./_sensord": 12.0,
   "selfdrive.controls.radard": 4.5,
-  "./_modeld": 4.48,
+  "selfdrive.modeld.modeld": 4.48,
```
adeebshihadeh, line 39: make sure to bump this to the new CPU usage

selfdrive/modeld/SConscript
```diff
@@ -60,12 +48,37 @@ else:
     del libs[libs.index('SNPE')]
     del common_src[common_src.index('runners/snpemodel.cc')]
 
+if use_thneed and arch == "larch64" or GetOption('pc_thneed'):
```
adeebshihadeh, line 51: use parentheses here, precedence isn't clear

selfdrive/modeld/modeld
```diff
@@ -1,11 +1,8 @@
...
-cd $DIR
+cd "$DIR/../../"
 
-if [ -f /TICI ]; then
-  export LD_LIBRARY_PATH="/usr/lib/aarch64-linux-gnu:/data/pythonpath/third_party/snpe/larch64:$LD_LIBRARY_PATH"
-else
-  export LD_LIBRARY_PATH="$DIR/../../third_party/snpe/x86_64-linux-clang:$DIR/../../openpilot/third_party/snpe/x86_64:$LD_LIBRARY_PATH"
-fi
-exec ./_modeld
+PYTHON=$(pyenv which python)
```
adeebshihadeh, line 6: this shouldn't rely on pyenv, your environment should handle this

selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,307 @@
...
+    logging.info("Features enqueued")
+
+    return self.output
+
+
+def main():
+  setproctitle("selfdrive.modeld.modeld")
+  if not PC:
+    set_realtime_priority(54)
+    set_core_affinity([7])
```
adeebshihadeh, line 134: use config_realtime_process instead, it also disables the GC

selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,307 @@
...
+      main_wide_camera = VisionStreamType.VISION_STREAM_ROAD not in available_streams
+      break
+    time.sleep(.1)
+
+  vipc_client_main_stream = VisionStreamType.VISION_STREAM_WIDE_ROAD if main_wide_camera else VisionStreamType.VISION_STREAM_ROAD
+  vipc_client_main = VisionIpcClient("camerad", vipc_client_main_stream, True, cl_context)
+  vipc_client_extra = VisionIpcClient("camerad", VisionStreamType.VISION_STREAM_WIDE_ROAD, False, cl_context)
+  logging.warning(f"vision stream set up, main_wide_camera: {main_wide_camera}, use_extra_client: {use_extra_client}")
+
+  # TODO: Is it safe to use blocking=True here?
```
adeebshihadeh, line 154: don't think so because of msgq, but don't remember if this hits msgq yet. should be fine if you test it

selfdrive/modeld/models/commonmodel.pxd
```diff
@@ -0,0 +1,24 @@
+# distutils: language = c++
+#cython: language_level=3
```
adeebshihadeh, line 2: don't think it's necessary to specify `language_level` anymore, let's remove them

selfdrive/modeld/models/commonmodel_pyx.pyx
```diff
@@ -0,0 +1,45 @@
+# distutils: language = c++
+# cython: c_string_encoding=ascii, language_level=3
+
+import numpy as np
+cimport numpy as cnp
+from libc.string cimport memcpy
+from cereal.visionipc.visionipc cimport cl_mem
```
adeebshihadeh, line 7: split up stdlib imports from our stuff with a newline

selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,307 @@
...
+    #   ((mt2 - mt1)*1000, (mt1 - last)*1000, meta_extra.frame_id, frame_id, frame_drop_ratio))
+    # last = mt1
+    last_vipc_frame_id = meta_main.frame_id
+
+
+if __name__ == "__main__":
+  try:
+    main()
+  except KeyboardInterrupt:
+    sys.exit()
```
adeebshihadeh, line 307: what's the exception handling here for?

selfdrive/modeld/SConscript
```diff
@@ -60,12 +48,37 @@ else:
...
+  lenvCython['CFLAGS'].append("-DUSE_THNEED")
+  lenvCython['CXXFLAGS'].append("-DUSE_THNEED")
+
 onnxrunner_path = File("runners/onnx_runner.py").abspath
 transform_path = File("transforms/transform.cl").abspath
 loadyuv_path = File("transforms/loadyuv.cl").abspath
 lenv['CXXFLAGS'].append(f'-DTRANSFORM_PATH=\\"{transform_path}\\"')
 lenv['CXXFLAGS'].append(f'-DLOADYUV_PATH=\\"{loadyuv_path}\\"')
 lenv['CXXFLAGS'].append(f'-DONNXRUNNER_PATH=\\"{onnxrunner_path}\\"')
+lenvCython['CXXFLAGS'].append(f'-DTRANSFORM_PATH=\\"{transform_path}\\"')
```
adeebshihadeh, line 61: this shouldn't be copy/pasted

selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,307 @@
...
+from pathlib import Path
+from typing import Dict, Optional
+from setproctitle import setproctitle  # pylint: disable=no-name-in-module
+from cereal.messaging import PubMaster, SubMaster
+from cereal.visionipc import VisionIpcClient, VisionStreamType, VisionBuf
+from system.hardware import PC
+from common.params import Params
+from common.filter_simple import FirstOrderFilter
+from common.realtime import set_core_affinity, set_realtime_priority
+from selfdrive.modeld.models.commonmodel_pyx import ModelFrame, CLContext, Runtime # pylint: disable=import-error, no-name-in-module
```
adeebshihadeh, line 15: can probably get rid of these pylint suppressions, don't think ruff will error on this

selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,307 @@
...
+      self.model.addInput(k, v)
+
+  def run(self, buf: VisionBuf, wbuf: VisionBuf, transform: np.ndarray, transform_wide: np.ndarray,
+                inputs: Dict[str, np.ndarray], prepare_only: bool) -> Optional[np.ndarray]:
+    # Model decides when action is completed, so desire input is just a pulse triggered on rising edge
+    inputs['desire_pulse'][0] = 0
+    self.inputs['desire_pulse'][:-DESIRE_LEN] = self.inputs['desire_pulse'][DESIRE_LEN:]
+    self.inputs['desire_pulse'][-DESIRE_LEN:] = np.where(inputs['desire_pulse'] - self.prev_desire > .99, inputs['desire_pulse'], 0)
+    self.prev_desire[:] = inputs['desire_pulse']
+    logging.info("Desire enqueued")
```
adeebshihadeh, line 100: logging -> cloudlog

selfdrive/modeld/runners/thneedmodel.pxd
```diff
@@ -0,0 +1,9 @@
+# distutils: language = c++
+#cython: language_level=3
+
+from libcpp.string cimport string
+from cereal.visionipc.visionipc cimport cl_context
+
```
adeebshihadeh, line 6: a bunch of the diff here is just wrapper. can we reduce this at all? at least collapse into a single pxd/pyx pair?

selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,307 @@
...
+from selfdrive.modeld.models.commonmodel_pyx import ModelFrame, CLContext, Runtime # pylint: disable=import-error, no-name-in-module
+from selfdrive.modeld.models.driving_pyx import USE_THNEED, PublishState # pylint: disable=import-error, no-name-in-module
+from selfdrive.modeld.models.driving_pyx import create_model_msg, create_pose_msg, update_calibration # pylint: disable=import-error, no-name-in-module
+
+if USE_THNEED:
+  from selfdrive.modeld.runners.thneedmodel_pyx import ThneedModel as ModelRunner # pylint: disable=import-error, no-name-in-module
+else:
+  from selfdrive.modeld.runners.onnxmodel_pyx import ONNXModel as ModelRunner # pylint: disable=import-error, no-name-in-module
+
+FEATURE_LEN = 128
```
adeebshihadeh, line 24: these constants shouldn't be duplicated. if we can't have cython handle them, we can generate a header file at build time

selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,307 @@
...
+
+    vec_desire = np.zeros(DESIRE_LEN, dtype=np.float32)
+    if desire >= 0 and desire < DESIRE_LEN:
+      vec_desire[desire] = 1
+
+    # Enable/disable nav features
+    timestamp_llk = sm["navModel"].locationMonoTime
+    nav_valid = sm.valid["navModel"] # and (nanos_since_boot() - timestamp_llk < 1e9)
+    use_nav = nav_valid and params.get_bool("ExperimentalMode")
+    if not nav_enabled and use_nav:
```
haraschax, line 245: Why isn't this just:

if use_nav:
else:

selfdrive/modeld/modeld.py
```diff
@@ -0,0 +1,307 @@
...
+TRAFFIC_CONVENTION_LEN = 2
+DRIVING_STYLE_LEN = 12
+NAV_FEATURE_LEN = 256
+NAV_INSTRUCTION_LEN = 150
+OUTPUT_SIZE = 5990
+MODEL_OUTPUT_SIZE = 6120
+MODEL_FREQ = 20
+
+MODEL_WIDTH = 512
+MODEL_HEIGHT = 256
```
haraschax, line 36: This is potentially a different PR, but those constants can all be extracted from the model?

selfdrive/modeld/modeld.py
haraschax: I think it doesn't exit when you block. @mitchellgoffpc Can you check it out and make sure the keyboardinterrupt works if that's the issue in a separate PR? That's a good fix anyway.

haraschax:
There is still a lot of ugly stuff here. Like all the duplicated constants. Those should really all be extracted from the model. What is needed to do that properly? As in have multi-output ONNX and use those shapes? It should be seperate PR, but I don't really wanted master in a state for any amount of time with way more magic numbers, when there should be none.

It's good the input mess is fixed though.

mitchellgoffpc:
> There is still a lot of ugly stuff here. Like all the duplicated constants. Those should really all be extracted from the model. What is needed to do that properly? As in have multi-output ONNX and use those shapes? It should be seperate PR, but I don't really wanted master in a state for any amount of time with way more magic numbers, when there should be none.
> 
> It's good the input mess is fixed though.

I agree, although the constants aren't anything new, I copied them verbatim from the c++ code. So I don't think this is regressing the ugliness of the code in any way. We should be able to load the majority of the constants from the onnx metadata, but I think that belongs in a separate PR.

selfdrive/modeld/modeld.py
mitchellgoffpc: Most of them can, yes, but I think this should be a separate PR since these constants are literal translations of what was in the C++ code before.

adeebshihadeh:
> copied them verbatim

This is a regression.

selfdrive/modeld/modeld.py
mitchellgoffpc: `nav_features` and `nav_instructions` only need to be zero'ed out on the falling edge of `use_nav`. Zero-ing them out on every frame wastes a bit of CPU.

haraschax: If it's a negligible amount of cpu it seems better to just set it every time, especially since you need that cpu anyway when nav is running.

mitchellgoffpc:
> > copied them verbatim
> 
> This is a regression.

My bad, I thought I'd deleted those constants from the C++ code, but a bunch of them are still there. I changed it to import those constants through cython for now since they're needed for the struct definitions.

selfdrive/modeld/modeld.py
mitchellgoffpc: Without the exception handling, KeyboardInterrupt will cause the process to exit with code 130. This was causing some tests to fail.

selfdrive/modeld/modeld.py
mitchellgoffpc: Confirmed KeyboardInterrupt doesn't work with block=True, which makes sense since the blocking is handled in visionipc_client.cc. I think the only real way around this is to do the blocking in visionipc_pyx, but it's a bit ugly to have that logic duplicated...

selfdrive/modeld/runners/thneedmodel.pxd
mitchellgoffpc: Not easily :/ thneedmodel has to be its own pxd file since it's only compiled under certain conditions, onnxmodel and snpemodel could technically be combined but since we're going to get rid of onnxmodel as soon as navmodeld and dmonitoringmodeld are rewritten (which I'm going to do right after this), I don't think it really makes sense to combine them.

adeebshihadeh:
Can we also split out the wrappers/prep for modeld.py from this?

PR #29233, opened 2023-08-04
FPv2: fast fuzzy fingerprinting

### LINKED ISSUE #28615: FPv2: return as soon as we get a fuzzy match
Right now we need an exact match to break early and match as a car. We should be able to do it for fuzzy matches too, but should check the fleet to see if this would ever change any outcomes. Would be a big speed up for fuzzy cars (which is going to be more common with [brand-specific fuzzy FP](https://github.com/commaai/openpilot/issues/26865)).

sshane:
Now breaks from the FW query loop as soon as we get a match, exact or not. Need to check data that this is fine.

### DIFF
--- a/selfdrive/car/fw_versions.py
+++ b/selfdrive/car/fw_versions.py
@@ -243,7 +243,7 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
     car_fw = get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, num_pandas=num_pandas, debug=debug, progress=progress)
     all_car_fw.extend(car_fw)
 
-    # If there is a platform match using this brand's FW alone, finish querying early
+    # If there is a match using this brand's FW alone, finish querying early
     matches = match_fw_to_car(car_fw, log=False)
     if len(matches) == 1:
       break
--- a/selfdrive/car/fw_versions.py
+++ b/selfdrive/car/fw_versions.py
@@ -243,7 +243,7 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
     car_fw = get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, num_pandas=num_pandas, debug=debug, progress=progress)
     all_car_fw.extend(car_fw)
 
-    # If there is a platform match using this brand's FW, finish querying early
+    # If there is a platform match using this brand's FW alone, finish querying early
     matches = match_fw_to_car(car_fw, log=False)
     if len(matches) == 1:
       break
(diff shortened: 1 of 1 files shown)


sshane:
Checked 1,861,628 qlogs for the past 500 days. Any segment with non-empty carFw list was checked.

- In a segment, FW was grouped by its request brand (eg. `{'toyota': [engine, fwdRadar, etc...], 'hyundai': [engine]}`)
- Each group's FW was run through `match_fw_to_car` for exact and fuzzy matches
- If there was a match, its brand needed to equal the online brand
- TODO: keep track of all matches for each FW group and ensure it's a set of len <= 1 (doesn't switch inside a brand somehow)

selfdrive/car/fw_versions.py
```diff
@@ -242,8 +242,9 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
 
     car_fw = get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, num_pandas=num_pandas, debug=debug, progress=progress)
     all_car_fw.extend(car_fw)
-    # Try to match using FW returned from this brand only
-    matches = match_fw_to_car_exact(build_fw_dict(car_fw))
+
+    # If there is a platform match using this brand's FW alone, finish querying early
```
sshane, line 246:
```suggestion
    # If there is a match using this brand's FW alone, finish querying early
```

selfdrive/car/fw_versions.py
```diff
@@ -242,8 +242,9 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
 
     car_fw = get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, num_pandas=num_pandas, debug=debug, progress=progress)
     all_car_fw.extend(car_fw)
-    # Try to match using FW returned from this brand only
-    matches = match_fw_to_car_exact(build_fw_dict(car_fw))
+
+    # If there is a match using this brand's FW alone, finish querying early
+    matches = match_fw_to_car(car_fw, log=False)
     if len(matches) == 1:
```
gijskoning, line 248: match_fw_to_car always returns a tuple:
```
    ...
    if len(matches):
      return exact_match, matches

  return True, set()
```
So this if statement is irrelevant.

selfdrive/car/fw_versions.py
```diff
@@ -242,8 +242,9 @@ def get_fw_versions_ordered(logcan, sendcan, ecu_rx_addrs, timeout=0.1, num_pand
 
     car_fw = get_fw_versions(logcan, sendcan, query_brand=brand, timeout=timeout, num_pandas=num_pandas, debug=debug, progress=progress)
     all_car_fw.extend(car_fw)
-    # Try to match using FW returned from this brand only
-    matches = match_fw_to_car_exact(build_fw_dict(car_fw))
+
+    # If there is a match using this brand's FW alone, finish querying early
+    matches = match_fw_to_car(car_fw, log=False)
```
sshane, line 247:
```suggestion
    _, matches = match_fw_to_car(car_fw, log=False)
```

selfdrive/car/fw_versions.py
sshane: ah you're right, good catch!

PR #29240, opened 2023-08-04
radard: tie radard frequency to modelV2 vol. 2

fredyshox:


sshane:
@fredyshox this has some static analysis failures

fredyshox:
@sshane thanks for fixing them

PR #29259, opened 2023-08-06
Devcontainer

robbederks:
This makes it super easy to set up a dev environment with vscode.
It prepares the docker, and automatically mounts in the work dir.


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "openpilot devcontainer",
+  "initializeCommand": "docker build -t openpilot_base -f Dockerfile.openpilot_base .",
```
robbederks, line 3: This isn't great, since it needs the whole context of the openpilot folder to build (quite slow).
Ideally it just takes in the files it needs.

We could also base the dev container on the one in the image repo, but that wouldn't take into account local changes

Dockerfile.openpilot_dev
```diff
@@ -0,0 +1,5 @@
+FROM openpilot_base:latest
```
fredyshox, line 1: There's prebuilt image available at `ghcr.io/commaai/openpilot-base:latest`. It's x86 only, but we should be able to build and ship aarch64 ones on CI

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,18 @@
...
+  },
+  "privileged": true,
+  "containerEnv": {
+    "DISPLAY": "${localEnv:DISPLAY}",
+    "force_color_prompt": "1",
+    "PYTHONPATH": "/workspaces/openpilot"
+  },
+  "runArgs": [
+    "--volume=/tmp/.X11-unix:/tmp/.X11-unix:rw",
+    "--volume=${localEnv:XAUTHORITY}:/root/.Xauthority"
```
fredyshox, line 16: Additional local fs mountpoints, which are useful to persist:
* `$HOME/.comma` => `/root/.comma ` - stores auth tokens 
* `/tmp/comma_download_cache` => `/tmp/comma_download_cache` - `COMMA_CACHE`

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,18 @@
...
+    "context": "..",
+    "dockerfile": "../Dockerfile.openpilot_dev"
+  },
+  "privileged": true,
+  "containerEnv": {
+    "DISPLAY": "${localEnv:DISPLAY}",
+    "force_color_prompt": "1",
+    "PYTHONPATH": "/workspaces/openpilot"
+  },
+  "runArgs": [
```
fredyshox, line 14: add some more shared memory. `--shm-size 1G`?

Dockerfile.openpilot_dev
robbederks: Yes, the issue is that this one doesn't take into account local changes, but that might be fine

.devcontainer/devcontainer.json
robbederks: Yes, sounds good!

Dockerfile.openpilot_dev
robbederks: and yes, if we can build an aarch64 version in CI too that would be nice. Don't think there's github action runners for this yet, although we could use qemu?

fredyshox: > Yes, the issue is that this one doesn't take into account local changes, but that might be fine

That doesn't happen that often, we rarely change setup scripts or add packages. IMO it's fine, we can always provide additional instructions for people which will want to use custom builds.

> Don't think there's github action runners for this yet, although we could use qemu?

Unfortunately they don't have them yet. QEMU is viable solution, just have to examine how slow it's gonna be.

robbederks: Ok, let's just support x86 for now then, and we can fix this later when GH finally releases ARM builders.

robbederks:
@fredyshox Mind writing a small README for this? Think it's already in a pretty usable state

adeebshihadeh:
I imagine some contributors are going to be relying on this, so think we should add a small test before merging this.

Dockerfile.openpilot_dev
```diff
@@ -0,0 +1,5 @@
+FROM ghcr.io/commaai/openpilot-base:latest
```
adeebshihadeh, line 1: would put this file in .devcontainer/

tools/README.md
```diff
@@ -36,6 +36,14 @@ Build openpilot with this command:
 scons -u -j$(nproc)
 ```
 
+### Dev Container
+
+openpilot supports VS Code's Dev Containers. Dev containers provide customizable and consistent development environment wrapped inside a container. This means you can develop in a designated environment matching our primary development target, regardless of you local setup.
```
robbederks, line 41: It's not purely for vscode: https://containers.dev/supporting

fredyshox: Mentioned that. Still focused on VSCode in my short usage instructions since most people will be using that.

fredyshox:
Btw I've checked how slow would it be to build aarch64 images using QEMU and `docker buildx` on x64 machine. It's way to slow, after an hour it only made it to py3.11 installation (and crashed while doing so). 

That's very unfortunate, because this feature makes it super easy to have stable development experience on a mac.

fredyshox:
I've just found out that Azure pipelines seems to [support arm64](https://azure.microsoft.com/en-us/updates/azure-devops-pipelines-introduces-support-for-linuxarm64/). We're already using that in some places, maybe we could utilize that for building such images.

adeebshihadeh:
I don't want to introduce another CI system for this. QEMU should be fine, or BuildJet (#29277) if we end up merging that.

tools/README.md
```diff
@@ -44,6 +44,19 @@ Dev containers are supported in [multiple editors and IDEs](https://containers.d
...
+GUI apps like `ui` or `cabana` can also run inside the container by leveraging X11 forwarding. To make use of it on macOS, additional configuration steps must be taken. First of all, install [XQuartz](https://formulae.brew.sh/cask/xquartz#default):
+```
+brew install --cask xquartz
+```
+
+Open XQuartz, go to `XQuartz` > `Settings...` > `Security` and check "Authenticate connections" and "Allow connections from network clients" boxes. Then allow for connections from localhost with:
+```
+xhost +localhost
+```
+Note that this is temporary and only affects current XQuartz session. To make it pernament, add the above line to shell rc file.
```
robbederks, line 58: typo: permanent

fredyshox:
Added simple CI test for this which builds an image and runs the container using `@devcontainers/cli`, which should act very similar to vscode extension.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ghcr.io/commaai/openpilot-base:latest
+
+RUN printf "\nsource /workspaces/openpilot/tools/openpilot_env.sh\n" >> ~/.bashrc
+# remove gitconfig if exists, since its gonna be replaced by host one
+RUN rm -f /root/.gitconfig
```
adeebshihadeh, line 5: won't it be shadowed anyway when it's mounted in?

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,7 @@
...
+RUN apt update && apt install -y vim net-tools usbutils htop
+RUN pip install ipython jupyter jupyterlab
```
adeebshihadeh, line 7: we should consider adding this as a poetry group

.devcontainer/Dockerfile
fredyshox: We are not mounting it outselves, vscode/devcontainer does that for us, and it fails silently if it exists already.

tools/README.md
```diff
@@ -36,6 +36,34 @@ Build openpilot with this command:
 scons -u -j$(nproc)
 ```
 
+### Dev Container
+
+openpilot supports [Dev Containers](https://containers.dev/). Dev containers provide customizable and consistent development environment wrapped inside a container. This means you can develop in a designated environment matching our primary development target, regardless of you local setup.
+
+Dev containers are supported in [multiple editors and IDEs](https://containers.dev/supporting), including Visual Studio Code. To start using it in VS Code, make sure you have `docker` installed. Then install `Dev Containers` extension. More detailed installation instructions can be found [here](https://code.visualstudio.com/docs/devcontainers/containers#_installation).
```
adeebshihadeh, line 43: this new section is quite long. we should just link to the real instructions for each thing, i.e. https://code.visualstudio.com/docs/devcontainers/create-dev-container for this one

fredyshox:
@adeebshihadeh is this mergeable now?

.github/workflows/tools_tests.yaml
```diff
@@ -88,4 +88,19 @@ jobs:
...
         $DOCKER_LOGIN
         docker push $DOCKER_REGISTRY/openpilot-docs-base:latest
-        docker push $DOCKER_REGISTRY/openpilot-docs:latest
\ No newline at end of file
+        docker push $DOCKER_REGISTRY/openpilot-docs:latest
+
+  devcontainer:
+    name: devcontainer
+    runs-on: ubuntu-latest
+    if: github.repository == 'commaai/openpilot'
```
adeebshihadeh, line 96: what's this for?

.github/workflows/tools_tests.yaml
```diff
@@ -88,4 +88,19 @@ jobs:
...
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: true
+    - name: Setup Dev Container CLI
+      run: npm install -g @devcontainers/cli
+    - name: Build dev container image
+      run: devcontainer build --workspace-folder .
+    - name: Run dev container
+      run: devcontainer up --workspace-folder .
```
adeebshihadeh, line 106: can we also run something like "scons --dry-run" to get some coverage on the environment?

adeebshihadeh (approved):

adeebshihadeh:
Just got this while trying it out
image

.github/workflows/tools_tests.yaml
fredyshox: not needed in this case

fredyshox:
> Just got this while trying it out image

Seems something is wrong with permission to submodules. Gonna see what i can do

fredyshox:
Should be fine now, can you try again?

adeebshihadeh:
Same thing (this is not an efficient loop)

adeebshihadeh:
New issue
```
bash: /tmp/tools/openpilot_env.sh: No such file or directory
bash: /workspaces/openpilot/tools/openpilot_env.sh: No such file or directory
```

fredyshox:
> New issue
> 
> ```
> bash: /tmp/tools/openpilot_env.sh: No such file or directory
> bash: /workspaces/openpilot/tools/openpilot_env.sh: No such file or directory
> ```

Removed references to openpilot_env since it got removed

fredyshox:
> Seems it's missing the scons cache

I don't think that causes big inconvenience, as it's relatively quick to build. Including it would significantly inflate image size, which is already quite big.

adeebshihadeh:
It should be volume mounted, not copied in

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,22 @@
...
+    "DISPLAY": "${localEnv:DISPLAY}",
+    "PYTHONPATH": "${containerWorkspaceFolder}",
+    "force_color_prompt": "1"
+  },
+  "runArgs": [
+    "--volume=/tmp/.X11-unix:/tmp/.X11-unix",
+    "--volume=${localEnv:XAUTHORITY}:/root/.Xauthority",
+    "--volume=${localEnv:HOME}/.comma:/root/.comma",
+    "--volume=/tmp/comma_download_cache:/tmp/comma_download_cache",
+    "--volume=/tmp/devcontainer_scons_cache:/tmp/scons_cache",
```
adeebshihadeh, line 19: why is there a special devcontainer cache?

fredyshox: I didn't want to mess with hosts cache, which may be incompatible (like on macos)

PR #29263, opened 2023-08-07
remove xx dependencies

adeebshihadeh:


openpilot/dummy_pip_package.py
```diff
@@ -0,0 +1,3 @@
+
+
```
adeebshihadeh, line 2: @haraschax I'd rather just start doing this properly instead of making a dummy package. is #29338 enough to make this work?

sshane:
can't find poetry or pyenv after running ubuntu_setup now, I'll check it out.

PR #29264, opened 2023-08-07
VW MQB: Add FW for 2023 Skoda Karoq

Saber422:
route name：0bbe367c98fa1538|2023-08-08--00-12-15--0


jyoung8607:
If you have a fingerprint that expands the tested and supported model-year range, as this one does, go ahead and expand the range in CAR_INFO, for example, here:

https://github.com/commaai/openpilot/blob/ae8252c14a87181c9da10bab7e2df64a592987a6/selfdrive/car/volkswagen/values.py#L263

And then run `selfdrive/car/docs.py`, which will automatically re-generate docs/CARS.md to show the change.

Issue #29208, opened 2023-08-02
[magnetometer] Accurate reading procedure on mmc5603nj


nuwandavek:
- We currently read and store the uncalibrated values
- The SET/RESET function in the chip can be used to both eliminate temperature effects and strong magnetization effects at source (Procedure on [page 13](https://www.mouser.com/datasheet/2/821/Memsic_09102019_MMC5603NJ_Datasheet_Rev.B-1635324.pdf#page=13))
- We should also record the temperature of the sensor

PR #29265, opened 2023-08-07
sensord: fix LSM magnetometer thermal variation

adeebshihadeh:


nuwandavek:
Tackles https://github.com/commaai/openpilot/issues/29208

nuwandavek:
@adeebshihadeh I added magnetometer to a new thread at 25Hz. It is ready for review.

PR #29268, opened 2023-08-07
test_models: assert sent addresses on bus

sshane:
asserts all sent message addresses from carcontroller exist on the bus. opened to test (doesn't catch GV60 because it's not fuzzy)


sshane:
need lots of exceptions for messages we expect to not be on the bus (in the case of radar or ADAS ECU disable), closing for now until I can think of a more generic way

PR #29270, opened 2023-08-07
test_models: check relay malfunction

sshane:
just checked the RAV4 2017 failure, looks legit (not dashcamOnly, but we're not sending anything nor breaking the relay :thinking:)


adeebshihadeh:
If we suspect the routes are bad, let's just add new ones.

sshane:
Yep, will do!

sshane:
So these routes were all either dashcammed when they were made, or the openpilot toggle was off (we don't check that). Checking this, no failures.

selfdrive/car/tests/routes.py
```diff
@@ -165,7 +165,6 @@
   CarTestRoute("5f5afb36036506e4|2019-05-14--02-09-54", TOYOTA.COROLLA_TSS2),
   CarTestRoute("5ceff72287a5c86c|2019-10-19--10-59-02", TOYOTA.COROLLAH_TSS2),
   CarTestRoute("d2525c22173da58b|2021-04-25--16-47-04", TOYOTA.PRIUS),
-  CarTestRoute("b14c5b4742e6fc85|2020-07-28--19-50-11", TOYOTA.RAV4),
```
sshane, line 168: we have another route below to test a specific pedal safety mismatch issue

PR #29277, opened 2023-08-08
CI: use buildjet for process replay

adeebshihadeh:
- [ ] worth it?
- [ ] only move jobs over that are faster
- [x] do forks get charged to us? no
- [x] does fork CI just work? `runs-on` supports expressions, so we can just fallback to the normal GitHub runners
  - [ ] need a test that enforces this,  otherwise it'll break silently

### DIFF
--- a/.github/workflows/setup/action.yaml
+++ b/.github/workflows/setup/action.yaml
@@ -9,11 +9,6 @@ inputs:
     description: 'Whether or not to pull the git lfs'
     required: false
     default: 'true'
-  is_buildjet:
-    description: 'Whether or not this is a buildjet runner'
-    required: false
-    default: 'false'
-
 
 runs:
   using: "composite"
@@ -50,8 +45,8 @@ runs:
         find . -type f -executable -not -perm 755 -exec chmod 755 {} \;
         find . -type f -not -executable -not -perm 644 -exec chmod 644 {} \;
     - id: setup-buildx-action
-      if: ${{ inputs.is_buildjet == 'true' }}
-      name: Set up Docker Buildx
+      if: contains("${{runner.name}}", "buildjet")
+      name: Set up Docker Buildx on buildjet to ensure a consistent cache
       uses: docker/setup-buildx-action@v2
       with:
         driver: docker-container
--- a/.github/workflows/setup/action.yaml
+++ b/.github/workflows/setup/action.yaml
@@ -45,7 +45,7 @@ runs:
         find . -type f -executable -not -perm 755 -exec chmod 755 {} \;
         find . -type f -not -executable -not -perm 644 -exec chmod 644 {} \;
     - id: setup-buildx-action
-      if: contains("${{runner.name}}", "buildjet")
+      if: contains(runner.name, 'buildjet')
       name: Set up Docker Buildx on buildjet to ensure a consistent cache
       uses: docker/setup-buildx-action@v2
       with:
--- a/.github/workflows/setup/action.yaml
+++ b/.github/workflows/setup/action.yaml
@@ -9,6 +9,11 @@ inputs:
     description: 'Whether or not to pull the git lfs'
     required: false
     default: 'true'
+  is_buildjet:
+    description: 'Whether or not this is a buildjet runner'
+    required: false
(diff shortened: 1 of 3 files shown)


adeebshihadeh:
@jnewb1 can you finish this up and evaluate how much faster it is?

jnewb1:
ignoring setup time:

| Runner | Process Replay Time | Unit Tests Time |
| ---------- | ------ | ------ |
| Github (master) | 17:05 | 20:03 |
| buildjet-4cpu | 4:58 | 19:33 |
| buildjet-8cpu | 2:46 | 19:11 |

Perhaps if we split the unit tests across the 4 or 8 cpus it would be significantly faster for the unittests:
https://pytest-xdist.readthedocs.io/en/stable/

adeebshihadeh:
* Definitely worth it for process replay.
* Unit tests need to be properly parallelized before we'll realize much benefit from this
* Static analysis can stay on GH Actions or go, either seems fine
* how's test car models? hoping that one is decently faster (it already supports parallel running with `pytest -n 8`

jnewb1:
```how's test car models? hoping that one is decently faster```
added to table, about 6 mins compared to 11 for 8vcpu

**Comment by jnewb1** [comment]
My verdict would be:
- run cars models and process replay on 8vcpu builtjet
- run the rest on GH, until we can split up the unit tests to take advantage of multiple cpus

TODO:
- figure out the caching issues on buildjet
  - Docker cache is pulled but not used?
  - scons_cache always gives a cache miss

**Comment by jnewb1** [comment]
Buildjet running on my fork, <6 mins for car and replay tests 
https://github.com/jnewb1/openpilot/actions/runs/5973895185/job/16207030187?pr=5

**Comment by jnewb1** [comment]
@adeebshihadeh 
i found that only buildjet requires pinning the docker version, and since it's so much faster than github actions for those cases it shouldn't be an issue

**Comment by jnewb1** [comment]
buildjet has a concurrency limit of 64 cpus, meaning only 8 jobs could run at once. We can also pay 300$ per month to increase that to 164

we might want to run the car tests on 4cpu instead of 8, 8 might be overkill since 2/3 of the time is in setup

**Comment by adeebshihadeh** [comment]
From what I understand, process replay is still the only job that clearly needs a faster machine. The cars tests take ~7m on master, and I suspect you might even get a furhter speedup by using 3 or 4 jobs on the GitHub runners. As for the unit tests, it's not clear where all that time is going.

**Comment by jnewb1** [comment]
> From what I understand, process replay is still the only job that clearly needs a faster machine. The cars tests take ~7m on master, and I suspect you might even get a furhter speedup by using 3 or 4 jobs on the GitHub runners. As for the unit tests, it's not clear where all that time is going.

right, but we could combine the car tests into maybe 1 or 2 jobs if we use buildjet. Tried using 4 jobs for the car tests on github actions and it was about the same on average.

**Comment by adeebshihadeh** [comment]
Why would we want to do that? We want to use BuildJet as little as possible: it costs us money and it diverges our CI from external contributors'.

File: .github/workflows/setup/action.yaml [hunk 1]
```diff
@@ -44,6 +49,12 @@ runs:
       run: |
         find . -type f -executable -not -perm 755 -exec chmod 755 {} \;
         find . -type f -not -executable -not -perm 644 -exec chmod 644 {} \;
+    - id: setup-buildx-action
+      if: ${{ inputs.is_buildjet == 'true' }}
```
adeebshihadeh, line 53: why can't we detect this in the setup action?

adeebshihadeh (commented):
If this is ready, let's merge for process replay.

adeebshihadeh:
"Post Cache test routes" has a bunch of permission issues. Is that a problem?

jnewb1:
> "Post Cache test routes" has a bunch of permission issues. Is that a problem?
> 
> Also seems like it didn't find the cache for some reason?

issue on master: https://github.com/commaai/openpilot/pull/29636

.github/workflows/setup/action.yaml
jnewb1: done

PR #29278, opened 2023-08-08
Locationd: switch to laikad

haraschax:


adeebshihadeh:
Seems CI didn't run because of the merge conflict. This seems fine to me, but can try to deal with it if you think it's a problem.

haraschax:
No its fine

sshane:
@haraschax laikad is not on master

PR #29291, opened 2023-08-09
Big test models temp

sshane:
other branch broke again :(


sshane:
What it looks like in jenkins: https://jenkins.comma.life/blue/organizations/jenkins/openpilot/detail/big-test-models-temp/3/pipeline

PR #29292, opened 2023-08-09
refactor hyundai canfd car

crwusiz:
split ev / hev / ice
rename hybrid_car -> hev_car


sshane:
why split hybrids and EVs into a CAN FD list?

sshane:
I don't think is useful for us

selfdrive/car/hyundai/values.py
```diff
@@ -1930,7 +1943,7 @@ def match_fw_to_car_fuzzy(live_fw_versions) -> Set[str]:
 CAMERA_SCC_CAR = {CAR.KONA_EV_2022, }
 
 # these cars use a different gas signal
-HYBRID_CAR = {CAR.IONIQ_PHEV, CAR.ELANTRA_HEV_2021, CAR.KIA_NIRO_PHEV, CAR.KIA_NIRO_HEV_2021, CAR.SONATA_HYBRID, CAR.KONA_HEV, CAR.IONIQ,
+HEV_CAR = {CAR.IONIQ_PHEV, CAR.ELANTRA_HEV_2021, CAR.KIA_NIRO_PHEV, CAR.KIA_NIRO_HEV_2021, CAR.SONATA_HYBRID, CAR.KONA_HEV, CAR.IONIQ,
```
sshane, line 1946: all the information we need is these two lists and a CAN FD list

PR #29294, opened 2023-08-09
CI: Pylint to ruff

jnewb1:


tools/lib/tests/test_caching.py
```diff
@@ -4,7 +4,7 @@
 import unittest
 
 os.environ["COMMA_CACHE"] = "/tmp/__test_cache__"
-from tools.lib.url_file import URLFile, CACHE_DIR
+from tools.lib.url_file import URLFile, CACHE_DIR # noqa: E402
```
adeebshihadeh, line 7: we can disable E402

selfdrive/car/nissan/values.py
```diff
@@ -45,7 +46,6 @@ class NissanCarInfo(CarInfo):
 }
 
 FINGERPRINTS = {
-  # pylint: disable=C0301
```
adeebshihadeh, line 48: how is this handled now?

jnewb1: ruff doesn't allow disabling blocks of code like pylint. I made ruff exclude E501 by putting
```# ruff: noqa: E501``` to the top of the file

only other alternative I see is adding that line to the end of every fingerprint

tools/lib/tests/test_caching.py
jnewb1: done

selfdrive/car/nissan/values.py
jnewb1: don't see any regex comparing ability in ruff

PR #29295, opened 2023-08-09
Move FW versions and fingerprints into their own file

### LINKED ISSUE #29293: Move FW versions and fingerprints into their own file

sshane:


adeebshihadeh:
Make sure to remove the linter ignore in the values.py files for line length

selfdrive/car/body/fingerprints.py
```diff
@@ -0,0 +1,36 @@
...
+
+Ecu = car.CarParams.Ecu
+
+FINGERPRINTS = {
+  CAR.BODY: [{
+    513: 8, 516: 8, 514: 3, 515: 4,
+  }],
+}
+
+FW_QUERY_CONFIG = FwQueryConfig(
```
adeebshihadeh, line 13: think this should still live in values.py. the idea is that these files eventually become largely auto-generated by an offline job pulling in fuzzy fingerprints

sshane: As far as I know, only Hyundai will have the circular import problem (since it has a fuzzy fingerprinting function that imports from fingerprints.py). We can just do the importing at run time instead of import time (in a cached function) with `get_interface_attr`. Not sure how clean that is though

sshane: Though I eventually wanted to make the refactor to put the FW versions inside the config, and we could combine a bunch of logic by having the configs manage everything, which this makes harder.

sshane:
Done in https://github.com/commaai/openpilot/pull/30790

PR #29299, opened 2023-08-09
test implicit string concat

sshane:
opening to run CI


sshane:
@jnewb1 this used to fail with pylint, can you see if https://github.com/commaai/openpilot/pull/29294 regressed this?

PR #29300, opened 2023-08-09
Ruff: enable more code quality rules

jnewb1:
With ruff, some of the rules must be explicity enabled, even ones that were enabled by default in pylint. Is there any other rules that we want?

https://beta.ruff.rs/docs/rules/#flake8-implicit-str-concat-isc

https://github.com/commaai/openpilot/pull/29299

@sshane


selfdrive/car/docs_definitions.py
```diff
@@ -174,12 +174,12 @@ def all_parts(self):
...
   EXP_LONG_AVAIL = CarFootnote(
-    "Experimental openpilot longitudinal control is available behind a toggle; " +
+    "Experimental openpilot longitudinal control is available behind a toggle; "
     "the toggle is only available in non-release branches such as `devel` or `master-ci`. ",
     Column.LONGITUDINAL, docs_only=True)
   EXP_LONG_DSU = CarFootnote(
-    "By default, this car will use the stock Adaptive Cruise Control (ACC) for longitudinal control. " +
-    "If the Driver Support Unit (DSU) is disconnected, openpilot ACC will replace " +
+    "By default, this car will use the stock Adaptive Cruise Control (ACC) for longitudinal control. "
+    "If the Driver Support Unit (DSU) is disconnected, openpilot ACC will replace "
```
sshane, line 182: We don't want implicit concatenation

adeebshihadeh:
Let's try these:
* https://beta.ruff.rs/docs/rules/#flake8-builtins-a
* https://beta.ruff.rs/docs/rules/#flake8-comprehensions-c4
* https://beta.ruff.rs/docs/rules/#flake8-pie-pie


Any changes we make should also be propagated to the submodule configs too.

sshane:
Also noticed we don't catch mutable args in functions too (`dangerous-default-value`). Let's make sure we set up ruff to catch the most common bugs to write.

jnewb1:
dangerous-default-value is B006, from flake8-bugbear. flake8-implicit-str-concat.allow-multiline=false disabled implicit concat across multiple lines.

However, ISC003 disables explicit string concat
```
    "Experimental openpilot longitudinal control is available behind a toggle; " +
    "the toggle is only available in non-release branches such as `devel` or `master-ci`. ",
```

in favor of implicit with parens
```
 ("Experimental openpilot longitudinal control is available behind a toggle; "
    "the toggle is only available in non-release branches such as `devel` or `master-ci`. ")
```

So I disabled ISC003, unless we like the parens better?

jnewb1:
splitting into multiple PRS

PR #29304, opened 2023-08-10
Poetry lock: bump poetry and fix xx deps

haraschax:


adeebshihadeh:
Need to pin sentry due to a breaking change, and we can't update without updating AGNOS. See the version in #29263

PR #29306, opened 2023-08-10
test_models: match online fingerprinting behavior

sshane:
only adds first `FRAME_FINGERPRINT` CAN messages to fingerprint, as we do online and in the pipeline. `cls.can_msgs` is unaffected.

simply `m.src, m.address, len(m.dat)` is apparently expensive.

a lot of the time is still spent on decompressing the bz2 file, 0.5 to 1.0 seconds each segment.


sshane:
Ran 50 random internal test models segs sequentially, avg time for adding the CAN messages went from ~CALCULATING (> 1 second) to ~0.226 seconds.

PR #29308, opened 2023-08-10
carcontroller.py clean up

eFiniLan:
do not compute apply_steer if lat control is disabled


adeebshihadeh:
This isn't clearly better. We should probably just abstract out the repeated logic.

PR #29311, opened 2023-08-10
map ui: Fixed ETA text alignment issue in CJK

eFiniLan:
ETA section text seems out of alignment when text is translated into CJK.

English:
old_eng
CHT:
old_cht
Japanese:
old_jp
Korean:
old_kr

--------

Adding vertical-align: bottom fixed the issue a little (not perfect, but good enough for OCD):
English:
new_eng
CHT:
new_cht
Japanese:
new_jp
Korean:
new_kr


sshane:
LGTM, @deanlee let me know if you think there's a better way to do this

PR #29312, opened 2023-08-10
util: random_string is now more entropic

ghost:
sizeof(char*) is going to be 8 on a 64-bit system, not the length of the string. Before, only strings of integer values [0-6] were being generated. Also switched to C++ constructs for simplicity.

### DIFF
--- a/common/util.cc
+++ b/common/util.cc
@@ -214,9 +214,9 @@ std::string hexdump(const uint8_t* in, const size_t size) {
 }
 
 std::string random_string(std::string::size_type length) {
-  const char* chrs = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+  const std::string chrs {"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"};
   std::mt19937 rg{std::random_device{}()};
-  std::uniform_int_distribution<std::string::size_type> pick(0, sizeof(chrs) - 2);
+  std::uniform_int_distribution<std::string::size_type> pick(0, chrs.length());
   std::string s;
   s.reserve(length);
   while (length--) {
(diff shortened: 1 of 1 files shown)


common/util.cc
```diff
@@ -214,9 +214,9 @@ std::string hexdump(const uint8_t* in, const size_t size) {
 }
 
 std::string random_string(std::string::size_type length) {
-  const char* chrs = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+  const std::string chrs {"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"};
   std::mt19937 rg{std::random_device{}()};
-  std::uniform_int_distribution<std::string::size_type> pick(0, sizeof(chrs) - 2);
+  std::uniform_int_distribution<std::string::size_type> pick(0, chrs.length());
```
sshane, line 219:
```suggestion
  std::uniform_int_distribution<std::string::size_type> pick(0, chrs.length() - 1);
```

you're accessing out of bounds memory

Misophists: Good pickup! Confused by ASAN didn't throw an error on that when I ran it multiple times.

Misophists: Apparently str[str.size()] is valid in C++11, so presumably that isn't an out-of-bounds read to ASAN.

Super.

common/util.cc
```diff
@@ -214,9 +214,9 @@ std::string hexdump(const uint8_t* in, const size_t size) {
 }
 
 std::string random_string(std::string::size_type length) {
-  const char* chrs = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
+  const std::string chrs {"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ"};
```
sshane, line 217:
```suggestion
  const std::string chrs = "0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ";
```

PR #29313, opened 2023-08-10
Carcontroller clean up

eFiniLan:
* Implement a CarControllerBase class for abstraction
* refactor steer torque limiters

@adeebshihadeh  is this something you are thinking?


sshane:
This is a change I ultimately want to make, but don't have the time to verify right now. So closing for now.

PR #29315, opened 2023-08-10
Ruff: enable flake8-builtins

jnewb1:
mostly A003, for shadowing builtin python variables / functions ("id", "type", "any", etc)

### DIFF
--- a/common/logging_extra.py
+++ b/common/logging_extra.py
@@ -65,7 +65,7 @@ class SwagFormatter(logging.Formatter):
 
     return record_dict
 
-  def format_json(self, record):
+  def format(self, record): # noqa: A003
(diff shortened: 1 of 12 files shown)


common/logging_extra.py
```diff
@@ -65,7 +65,7 @@ def format_dict(self, record):
 
     return record_dict
 
-  def format(self, record):
+  def format(self, record): # noqa: A003
```
adeebshihadeh, line 68: why is this ignored?

jnewb1: format is a function of logging.formatter, and needs to be overridden

adeebshihadeh (approved):

PR #29317, opened 2023-08-10
Ruff: comprehensions conventions

jnewb1:


system/sensord/tests/test_sensord.py
```diff
@@ -170,7 +170,7 @@ def test_lsm6ds3_timing(self):
   def test_events_check(self):
     # verify if all sensors produce events
 
-    sensor_events = dict()
+    sensor_events = {}
```
adeebshihadeh, line 173: is there a reasoning for this one? rest of the changes look good

jnewb1: C408: rewrite dict() as {}
https://pypi.org/project/flake8-comprehensions/

adeebshihadeh: hmm, their reasoning is that it's slower. don't think it's enough of a difference to restrict the style, so let's ignore C408

jnewb1: done

adeebshihadeh (approved):

PR #29320, opened 2023-08-11
Ruff: enable most of bugbear

jnewb1:


pyproject.toml
```diff
@@ -197,7 +197,7 @@ build-backend = "poetry.core.masonry.api"
 
 # https://beta.ruff.rs/docs/configuration/#using-pyprojecttoml
 [tool.ruff]
-select = ["E", "F", "W", "PIE", "C4", "ISC", "RUF100"]
+select = ["E", "F", "W", "PIE", "C4", "ISC", "RUF100", "B006"]
```
adeebshihadeh, line 200: think we want all the B's

jnewb1: Added all the B's except a few I marked below

system/hardware/tici/hardware.py
```diff
@@ -83,6 +83,17 @@ def affine_irq(val, action):
   for i in irqs:
     sudo_write(str(val), f"/proc/irq/{i}/smp_affinity_list")
 
+@lru_cache
```
adeebshihadeh, line 86: what's the reason on this one? seems worse

selfdrive/athena/athenad.py
```diff
@@ -213,6 +213,16 @@ def retry_upload(tid: int, end_event: threading.Event, increase_count: bool = Tr
         break
 
 
+def cb(sm, item, tid, sz: int, cur: int) -> None:
```
adeebshihadeh, line 216: same here

system/hardware/tici/hardware.py
jnewb1: # lru_cache and cache can cause memory leaks when used in classes

jnewb1: https://beta.ruff.rs/docs/rules/cached-instance-method/

selfdrive/athena/athenad.py
jnewb1: https://beta.ruff.rs/docs/rules/function-uses-loop-variable/

system/hardware/tici/hardware.py
adeebshihadeh: oh, this happens to be fine here, but seems like a good rule to have

selfdrive/athena/athenad.py
jnewb1: I can also just move the function up before the loop rather than to the global scope

adeebshihadeh (approved):

PR #29327, opened 2023-08-11
ui/map: show map on destination set/change

### LINKED ISSUE #27710: ui: nav does not show when changing destination
nav does not show when you change destination without clearing it

deanlee:
resolve https://github.com/commaai/openpilot/issues/27710


selfdrive/ui/qt/maps/map.cc
```diff
@@ -229,7 +218,7 @@ void MapWindow::updateState(const UIState &s) {
     }
   }
 
-  if (sm.rcv_frame("navRoute") != route_rcv_frame) {
+  if (sm.updated("navRoute")) {
```
sshane, line 221: I forgot the exact reason, but the two code blocks for navRoute are currently required (something to do with the map not being ready, and this calls map-specific stuff). Is this fixed?

selfdrive/ui/qt/maps/map.cc
```diff
@@ -159,14 +159,14 @@ void MapWindow::updateState(const UIState &s) {
...
-    // Only open the map on setting destination the first time
-    if (allow_open) {
-      emit requestSettings(false);
-      emit requestVisible(true); // Show map on destination set/change
-      allow_open = false;
+  if (sm.updated("navRoute")) {
+    auto nav_dest = coordinate_from_param("NavDestination");
+    if (std::exchange(current_nav_dest, nav_dest) != nav_dest) {
+      // Show map on destination set/change
+      if (current_nav_dest && !isVisible()) {
```
sshane, line 166: can you combine to one check? just want to show if current_nav_dest and !nav_dest

selfdrive/ui/qt/maps/map.cc
```diff
@@ -155,14 +155,13 @@ void MapWindow::updateState(const UIState &s) {
     }
   }
 
-  if (sm.updated("navRoute") && sm["navRoute"].getNavRoute().getCoordinates().size()) {
```
sshane, line 158: There may be a reason this was here: https://github.com/commaai/openpilot/pull/24352/files

I believe if you tap a destination without internet, it will popup with the invalid navRoute, correct?

selfdrive/ui/qt/maps/map.cc
```diff
@@ -155,14 +155,17 @@ void MapWindow::updateState(const UIState &s) {
...
+    auto nav_dest = coordinate_from_param("NavDestination");
+    bool allow_open = std::exchange(last_valid_nav_dest, nav_dest) != nav_dest &&
+                      last_valid_nav_dest && !isVisible();
     qWarning() << "Got new navRoute from navd. Opening map:" << allow_open;
 
-    // Only open the map on setting destination the first time
+    // Show map on destination set/change
     if (allow_open) {
       emit requestSettings(false);
       emit requestVisible(true); // Show map on destination set/change
```
deanlee, line 168: Remove redundant comments?

selfdrive/ui/qt/maps/map.cc
```diff
@@ -375,16 +379,17 @@ void MapWindow::offroadTransition(bool offroad) {
   last_bearing = {};
 }
 
-void MapWindow::updateDestinationMarker(std::optional<QMapbox::Coordinate> nav_dest) {
-  if (nav_dest) {
+void MapWindow::updateDestinationMarker() {
+  m_map->setPaintProperty("pinLayer", "visibility", "none");
+
+  auto nav_dest = coordinate_from_param("NavDestination");
```
deanlee, line 385: should we use last_valid_nav_dest here, instead of read it  from param?

selfdrive/ui/qt/maps/map.cc
sshane, line 168:
```suggestion
      emit requestVisible(true);
```

selfdrive/ui/qt/maps/map.cc
sshane: if you get a navRoute with no coords, the value of last_valid_nav_dest will be different than reading from the param

selfdrive/ui/qt/maps/map.cc
```diff
@@ -155,14 +155,17 @@ void MapWindow::updateState(const UIState &s) {
     }
   }
 
+  // TODO: add a valid flag to navRoute
```
sshane, line 158:
```suggestion
```

sshane:
this makes replay on PC slightly more divergent from online - if you don't have a local NavDestination the map won't open when it gets a `navRoute`. I don't see any other way to do this at the moment on-device, so we'll have to move NavDestination to a service to better replay online behavior

PR #29331, opened 2023-08-11
ui/map: check NavDestination only if navInstruction is invalid

deanlee:

### DIFF
--- a/selfdrive/ui/qt/maps/map.cc
+++ b/selfdrive/ui/qt/maps/map.cc
@@ -213,8 +213,7 @@ void MapWindow::updateState(const UIState &s) {
     // - API exception/no internet
     // - route response is empty
     // - any time navd is waiting for recompute_countdown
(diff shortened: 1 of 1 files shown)


selfdrive/ui/qt/maps/map.cc
```diff
@@ -213,8 +213,7 @@ void MapWindow::updateState(const UIState &s) {
     // - API exception/no internet
     // - route response is empty
     // - any time navd is waiting for recompute_countdown
-    auto dest = coordinate_from_param("NavDestination");
-    routing_problem = !sm.valid("navInstruction") && dest.has_value();
+    routing_problem = !sm.valid("navInstruction") && coordinate_from_param("NavDestination");
```
sshane, line 216:
```suggestion
    routing_problem = !sm.valid("navInstruction") && coordinate_from_param("NavDestination").has_value();
```

I think this is clearer

PR #29332, opened 2023-08-11
ui/map: fix css error

deanlee:
fix css error message: 

> : QCssParser::parseColorValue: Specified color without alpha value but alpha given: 'rgb 0, 0, 0, 150'

Screenshot from 2023-08-11 17-44-44


adeebshihadeh:
Is there some strict mode we can enable to make this fail more obviously? Something like PYTHONWARNINGS=error or -Werror

deanlee:
no idea yet.

PR #29336, opened 2023-08-11
Ruff: b905 (strict zip)

jnewb1:
depends: https://github.com/commaai/openpilot/pull/29320


adeebshihadeh:
This one just seems like poor design in the standard library, but guess we should just enable it.

PR #29339, opened 2023-08-11
CI: use default caching action

jnewb1:
prerequisite of using buildjet's cache is to be on the latest version of actions/cache


jnewb1:
seems like we get a cache miss after upgrading, probably because we were on a two year old commit of actions/cache, that doesn't support the newest format?

I added something in to allow more time in building if there is no scons-cache, which will also be useful if for whatever reason the cache gets totally cleared in github

adeebshihadeh (approved):
LGTM, just make sure the cache works properly once we have new ones from master

PR #29341, opened 2023-08-11
ui/map: assign the return value to `last_position`

deanlee:


selfdrive/ui/qt/maps/map.cc
```diff
@@ -43,11 +43,7 @@ MapWindow::MapWindow(const QMapboxGLSettings &settings) : m_settings(settings),
   overlay_layout->addStretch(1);
   overlay_layout->addWidget(map_eta);
 
-  auto last_gps_position = coordinate_from_param("LastGPSPosition");
-  if (last_gps_position.has_value()) {
-    last_position = *last_gps_position;
-  }
-
+  last_position = coordinate_from_param("LastGPSPosition");
```
sshane, line 46: should be fine since last_position is an optional too

PR #29346, opened 2023-08-11
replay: increase file descriptor limit macOS

### LINKED ISSUE #29281: [macOS] unable to run replay: Assertion failed: (*fd >= 0), function malloc_with_fd, file visionbuf_cl.cc, line 24.
### Describe the bug

environment:
* macbook m1
* ventura 13.4.1

the command below caused the issue:
```python
./replay --data_dir /Users/ricklan/projects/realdata "0000000000000000|2023-08-05--20-19-31"
```

however, if I add --no-vipc then I can replay the route without video stream just fine:
```python
./replay --data_dir /Users/ricklan/projects/realdata "0000000000000000|2023-08-05--20-19-31" --no-vipc
```

Screenshot 2023-08-08 at 11 26 00 PM

### OS Version

macOS ventura 13.4.1

95289cb180ac53cbe0f522fcbd81a8ad0019a2cc

### Additional info

_No response_

fredyshox:
Each ZMQ socket takes 3 descriptors on macOS (2 on linux). 71 services = 213 descriptors. VIPC creates 40 buffers = 40 descriptors. Add some networking and open some files, and we easily hit 256 limit.

Seems like for a while replay was right on the edge of the fd limit on macOS.

Fixes: #29281

### DIFF
--- a/tools/replay/main.cc
+++ b/tools/replay/main.cc
@@ -4,12 +4,11 @@
 #include "common/prefix.h"
 #include "tools/replay/consoleui.h"
 #include "tools/replay/replay.h"
-#include "tools/replay/util.h"
 
 int main(int argc, char *argv[]) {
 #ifdef __APPLE__
   // With all sockets opened, we might hit the default limit of 256 on macOS
-  setFileDescriptorLimit(1024);
+  util::set_file_descriptor_limit(1024);
 #endif
 
   QCoreApplication app(argc, argv);
--- a/tools/replay/main.cc
+++ b/tools/replay/main.cc
@@ -8,7 +8,7 @@
 
 int main(int argc, char *argv[]) {
 #ifdef __APPLE__
-  // With all the sockets opened, we might hit the default limit of 256 on macOS
+  // With all sockets opened, we might hit the default limit of 256 on macOS
   setFileDescriptorLimit(1024);
 #endif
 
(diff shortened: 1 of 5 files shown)


tools/replay/main.cc
```diff
@@ -4,8 +4,14 @@
 #include "common/prefix.h"
 #include "tools/replay/consoleui.h"
 #include "tools/replay/replay.h"
+#include "tools/replay/util.h"
 
 int main(int argc, char *argv[]) {
+#ifdef __APPLE__
+  // With all sockets opened, we might hit the default limit of 256 on macOS
+  setFileDescriptorLimit(1024);
```
adeebshihadeh, line 12: this doesn't have anything to do with replay, should probably go in the messaging backend

adeebshihadeh (changes requested):

tools/replay/main.cc
fredyshox: But it only affects replay process, because of extensive of sockets. Not every process needs to increase that limit, since most processes use only a couple of sockets at most. So I'd argue that it's the process responsibility to handle that.

fredyshox: But it could be moved elsewhere, so each process would be able to set it up if needed. Maybe `common/util.cc`?

fredyshox: Moved `set_file_descriptor_limit` to common, so its accessible to other tools/processes if ever needed.

PR #29349, opened 2023-08-12
CI: speed up FW unit tests

sshane:
Takes half the time of the unit tests! 155s -> 30s on local PC, 517s -> 318s in CI

Only look at the DB FW for the current brand we're fingerprinting

Currently we try to fingerprint on every candidate, when only FW from toyota is passed in, for example. This should never happen, and was partially verified in https://github.com/commaai/openpilot/pull/29233

edit: hmm, not getting nearly as much speed up in CI than local, think I'll just make this a car-focused clean up


sshane:
oh, forgot I have https://github.com/commaai/openpilot/pull/28435 🤦‍♂️

PR #29359, opened 2023-08-12
Nav - Wrong direction check

### LINKED ISSUE #28496: nav: doesn't re-route if traveling opposite direction on same road
https://connect.comma.ai/297b4b460f361603/1686472834765/1686472941839

relevant TODO: https://github.com/commaai/openpilot/blob/master/selfdrive/navd/navd.py#L321

pfeiferj:
**Problem:**
If you somehow get turned around on a nav-route you will not be redirected while you are backtracking along already driven portions of the route. An example would be you came across a downed tree and needed to turn around. In that case you would not have gone off the route enough to cause a re-route because you likely turned around on the road itself.

**Solution:**
This introduces a simple solution that just tracks if your distance remaining is increasing and then triggers a re-route. This should be a simpler implementation than calculating your current heading and comparing that against the heading of the route. However, this may take longer to trigger the re-route than checking the headings.

### DIFF
--- a/selfdrive/navd/navd.py
+++ b/selfdrive/navd/navd.py
@@ -1,8 +1,10 @@
 #!/usr/bin/env python3
+from enum import StrEnum
 import json
 import math
 import os
 import threading
+from typing import Tuple, Optional
 
 import requests
 
@@ -20,6 +22,12 @@ from openpilot.common.swaglog import cloudlog
 REROUTE_DISTANCE = 25
 MANEUVER_TRANSITION_THRESHOLD = 10
 REROUTE_COUNTER_MIN = 3
+
+
+class RECOMPUTE_REASON(StrEnum):
+  WRONG_DIRECTION = "wrong direction"
+  DISTANCE_TO_ROUTE = "distance to route"
+  NEW_DESTINATION = "new destination"
 
 
 class RouteEngine:
@@ -46,7 +54,11 @@ class RouteEngine:
 
     self.ui_pid = None
 
-    self.reroute_counter = 0
+    self.distance_reroute_counter = 0
+    self.wrong_direction_reroute_counter = 0
+
+    self.smallest_total_distance_remaining = math.inf
+    self.distance_delta = 0
 
     if "MAPBOX_TOKEN" in os.environ:
       self.mapbox_token = os.environ["MAPBOX_TOKEN"]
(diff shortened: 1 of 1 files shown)


rchybicki:
Tested this today, rerouted beautifully.

sshane:
@pd0wm I'm not too familiar with the routing code, wanted to see what you thought of this. I can check it out if you are too busy though!

pd0wm:
When I originally wrote this code I had something in mind that looks at the heading between the previous and the next point. But I guess something like this works as well!

Maybe it's cleaner to keep track of `distance_difference` in `send_instruction()` so you only need to introduce one new variable.

pfeiferj:
Ok, refactored it some.
* Moved the difference calculation over to send_instruction and stored it instead of the total distance.
* Went ahead and renamed the variable to distance_delta as well because distance_difference just sounds weird.

I'm not sure there's a way to get down to a single new variable as we need to keep track of the closest we've been to the destination and we're still using the delta in a different part of the code. I also kept using the total distance as it seems easier to keep track of when we need to reset the smallest distance value, i.e. once upon receiving the route rather than at every instruction.

selfdrive/navd/navd.py
```diff
@@ -273,6 +276,11 @@ def send_instruction(self):
     msg.navInstruction.timeRemaining = total_time
     msg.navInstruction.timeRemainingTypical = total_time_typical
 
+    # Compute if we are going the wrong way
+    if self.smallest_distance_remaining is None or total_distance < self.smallest_distance_remaining:
```
pd0wm, line 280: If you use `math.inf` instead of None you can replace this by a simple ` self.smallest_distance_remaining = min( self.smallest_distance_remaining, total_distance)`

pd0wm:
Since `total_distance` is constant during the whole route it still feels like you should be able to rewrite the math so you only need to store a single variable. However, if it becomes hard to follow it might be better to store two.

pd0wm:
Probably also nice to add a reroute reason to the `cloudlog.warning` in `compute_route()` then the comma team can track this in their backend and see if there are any issues with the new logic. If I didn't miss anything, valid reasons can be `new destination`, `wrong direction` or `distance to route`

pfeiferj:
@pd0wm updated with your suggestions. To add a reason to the warning I ended up doing some re-organizing of the recompute conditions:
* Pulled the wrong direction check into its own function.
* Pulled the current should_recompute distance check logic into a new function.
* Pulled the no recompute while in a tunnel logic into should_recompute.
* Pulled the new destination check into should_recompute
* should_recompute calls the distance and wrong direction checks
* should_recompute now returns a reason

I also made some of the variables a little more verbose so it's more clear that they are _remaining_ values

selfdrive/navd/navd.py
```diff
@@ -327,7 +338,33 @@ def reset_recompute_limits(self):
...
+      cloudlog.warning(f"Got new destination from NavDestination param {new_destination}")
+      return True, RECOMPUTE_REASON.NEW_DESTINATION
+
+    if self.too_far_from_route():
+      return True, RECOMPUTE_REASON.DISTANCE_TO_ROUTE
+
+    if self.going_wrong_direction():
+      return True, RECOMPUTE_REASON.WRONG_DIRECTION
+
+    return False, None
```
sshane, line 356: Is this required for this PR?

pfeiferj: not required at all, was just for the suggestion of logging the reason for recompute. I'll work on pulling it out

pfeiferj:
closing, if i get back around to this i'll just open a new PR

PR #29362, opened 2023-08-12
Subaru: ES_Distance cruise cancel safety

jnewb1:
Current cancel request method (using simulated buttons) uses the same message that contains Cruise_Throttle meaning we could have sent any throttle we wanted

depends: https://github.com/commaai/panda/pull/1576


selfdrive/car/subaru/carcontroller.py
```diff
@@ -64,7 +65,7 @@ def update(self, CC, CS, now_nanos):
         can_sends.append(subarucan.create_preglobal_es_distance(self.packer, cruise_button, CS.es_distance_msg))
 
     else:
-      if pcm_cancel_cmd and (self.frame - self.last_cancel_frame) > 0.2:
+      if pcm_cancel_cmd and (self.frame - self.last_cancel_frame)*DT_CTRL > 0.2:
```
jnewb1, line 68: this check didn't do anything before, since frames are whole numbers

adeebshihadeh, line 68: this should be done in a separate PR, it changes behavior

adeebshihadeh (changes requested):

selfdrive/car/subaru/carcontroller.py

PR #29368, opened 2023-08-13
Subaru: remove unused backoff rate calculation

jnewb1:
~~this check didn't do anything before, since frames are whole numbers. Need to verify if we can still cancel if we aren't spamming at the full 50hz~~

since cancel it working fine anyway, I'll just remove the check entirely


selfdrive/car/subaru/carcontroller.py
```diff
@@ -64,7 +65,7 @@ def update(self, CC, CS, now_nanos):
         can_sends.append(subarucan.create_preglobal_es_distance(self.packer, cruise_button, CS.es_distance_msg))
 
     else:
-      if pcm_cancel_cmd and (self.frame - self.last_cancel_frame) > 0.2:
+      if pcm_cancel_cmd and (self.frame - self.last_cancel_frame)*DT_CTRL > 0.2:
```
sshane, line 68:
```suggestion
      if pcm_cancel_cmd and (self.frame - self.last_cancel_frame) * DT_CTRL > 0.2:
```

jnewb1:
cancel seems to work fine on our outback: 54827bf84c38b14f|2023-08-15--09-54-57

adeebshihadeh:
I'm only seeing a couple dozen cancels in that route. We need to check that the distribution of cancellation times hasn't (meaningfully) changed.

The other option here is to remove the check that it's never hitting anyway instead of "fixing" it since that has been working fine.

jnewb1:
> I'm only seeing a couple dozen cancels in that route. We need to check that the distribution of cancellation times hasn't (meaningfully) changed.
> 
> The other option here is to remove the check that it's never hitting anyway instead of "fixing" it since that has been working fine.

If there's no particular reason that other car ports have a backoff rate, then I'll just remove it

selfdrive/car/subaru/carcontroller.py
```diff
@@ -100,7 +100,7 @@ def update(self, CC, CS, now_nanos):
           can_sends.append(subarucan.create_es_distance(self.packer, CS.es_distance_msg, 0, pcm_cancel_cmd,
                                                         self.CP.openpilotLongitudinalControl, cruise_brake > 0, cruise_throttle))
       else:
-        if pcm_cancel_cmd and (self.frame - self.last_cancel_frame) > 0.2:
+        if pcm_cancel_cmd:
           bus = CanBus.alt if self.CP.carFingerprint in GLOBAL_GEN2 else CanBus.main
           can_sends.append(subarucan.create_es_distance(self.packer, CS.es_distance_msg, bus, pcm_cancel_cmd))
           self.last_cancel_frame = self.frame
```
sshane, line 106:
```suggestion
```

don't need this?

PR #29377, opened 2023-08-13
ui/networking: remove dynamic property and style sheet

deanlee:


adeebshihadeh:
Is this actually faster?

deanlee:
y, possibly 20 times faster.  the dynamic property&style takes ~0.07ms for one item, the setFont  takes only ~0.003ms.

adeebshihadeh:
Nice!

adeebshihadeh:
Btw, did you get my email (sent to hotmail)? We need some more info to ship your stuff

deanlee:
y, replied!

PR #29378, opened 2023-08-13
ui/networking: show "connecting" immediately after clicking on a network.

deanlee:
`emit refreshSignal()` when connecting to network to fix this old bug.


selfdrive/ui/qt/offroad/networking.h
```diff
@@ -12,8 +12,9 @@ class WifiItem : public QWidget {
   void setItem(const Network& n, const QPixmap &icon, bool show_forget_btn, const QPixmap &strength);
 
 signals:
-  void connectToNetwork(const Network &n);
-  void forgotNetwork(const Network &n);
+  // Cannot pass Network by reference. it may change after the signal is sent.
```
deanlee, line 15: Also fixes a bug caused by passing network by references. (introduced in  #29037)

sshane: Can you make that fix a new PR?

deanlee: split to https://github.com/commaai/openpilot/pull/29448

sshane:
does this ever happen? perhaps it was the bug fixed in https://github.com/commaai/openpilot/pull/29396 causing it not to show as connecting for a few seconds https://github.com/commaai/openpilot/pull/29396

deanlee:
think this still happen. although the connected state can be refreshed correctly after #29396, The UI still won't be refreshed before the refresh timer is activated.

sshane:
I think things should be better after https://github.com/commaai/openpilot/pull/29396, I believe we get a refresh signal as soon as network manager starts connecting, so it's fine to wait until then. I checked on my Pixel and there also is a delay before it shows that it's connecting.

adeebshihadeh:
This is a real improvement.

sshane:
we can merge this before https://github.com/commaai/openpilot/pull/29647, since it adds the infrastructure for the other PR

PR #29389, opened 2023-08-14
soundd: set volume on changes

### LINKED ISSUE #28992: soundd: baseline CPU is 20%

deanlee:
~async~ set volumn on changes. (same as async set_brightnes in ui.cc)
resolve https://github.com/commaai/openpilot/issues/28992  （not tested on the device at the moment, but it should be solved.）

`hardware::set_volume `invokes expensive system calls.  blocking call it at 10hz can impact performance, And it may cause `soundd` to be unable to run stably at 20hz, and in the worst case, may cause some alerts to be lost.


adeebshihadeh:
I think we should only do async when absolutely necessary since this has the potential for very hard to debug issues. What if we just set the volume independently of playing a sound and only on change?

deanlee:
reverted "async set volume".  though I still think it might be better to use asyc system call here.

> What if we just set the volume independently of playing a sound and only on change?

Think we can do this later if there is still  a problem after set volume on change.

PR #29394, opened 2023-08-14
jenkins: re-enable test_models

sshane:
Forgot to remove in https://github.com/commaai/openpilot/pull/29270. ASCM is on in the Volt route for some reason

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -158,7 +158,8 @@ pipeline {
             sh "git lfs pull"
             // tests that our build system's dependencies are configured properly, needs a machine with lots of cores
             sh "scons --clean && scons --no-cache --random -j42"
-            //sh "INTERNAL_SEG_LIST=selfdrive/car/tests/test_models_segs.txt FILEREADER_CACHE=1 pytest -n42 --dist=loadscope selfdrive/car/tests/test_models.py"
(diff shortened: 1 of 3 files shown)


Jenkinsfile
```diff
@@ -158,7 +158,7 @@ pipeline {
             sh "git lfs pull"
             // tests that our build system's dependencies are configured properly, needs a machine with lots of cores
             sh "scons --clean && scons --no-cache --random -j42"
-            //sh "INTERNAL_SEG_LIST=selfdrive/car/tests/test_models_segs.txt FILEREADER_CACHE=1 pytest -n42 --dist=loadscope selfdrive/car/tests/test_models.py"
+            sh "INTERNAL_SEG_LIST=selfdrive/car/tests/test_models_segs.txt FILEREADER_CACHE=1 pytest -n42 --dist=loadscope selfdrive/car/tests/test_models.py"
```
sshane, line 161:
```suggestion
            //sh "INTERNAL_SEG_LIST=selfdrive/car/tests/test_models_segs.txt FILEREADER_CACHE=1 pytest -n42 --dist=loadscope selfdrive/car/tests/test_models.py"
```

sshane:
to randomly select segments each time, the selection needs to be consistent for all the pytest processes, so we need to pass in a seed. we can do later if we really need, but the test should be fast enough we can test everything, everytime

PR #29397, opened 2023-08-14
CI: docs speedup by caching first stage

jnewb1:
the docs image is a multistage build meaning we weren't using any of the first stage as cache. @adeebshihadeh what do you think about using the docker/build-push-action@v4 for all the docker builds?


jnewb1:
I'll do the build-push-action in a seprate PR if we like that better

adeebshihadeh:
What does that action get us?

jnewb1:
> What does that action get us?

I think its easier to read and understand but more verbose

PR #29400, opened 2023-08-14
add PyQt5 package

adeebshihadeh:


fredyshox:
Seems like PyQt5 fails to install on Apple Silicon, more specifically its dependency: `PyQt5-Qt5`

adeebshihadeh:
It's only for debug stuff right now - shall we make it linux only?

PR #29401, opened 2023-08-14
CI: prevent moc from running in cache population step

fredyshox:


adeebshihadeh:
What's this for?

fredyshox:
QT moc sources are stored inside `/tmp/scons_cache/moc_files`. scons_cache is deleted in cache repopulation steps (`build_all` and `build_mac` jobs), so they are regenerated/recompiled again, which adds on avg. 1-2 minutes of build time.

adeebshihadeh:
Good catch, but we still need to manage the old ones somehow. This cleanup step is done so we're not caching old build products.

fredyshox:
Added cleanup script, which looks at scons task-trace to determine which moc files are currently in use. It's based on a simple regex path matching, but should be enough to delete outdated build products.

PR #29406, opened 2023-08-15
ui: fix world objects with zoom

sshane:
Wasn't very noticeable for fcam due to how close it was to 100% (`stream height/widget height`, yes I removed UI border here to test, normally it's 30px less):

```python
round(2*1208/1080, 2)
Out[77]: 2.24
round(1*1208/1080, 2)
Out[78]: 1.12
```

The problem is that in cameraview.cc, zoom is taken to mean: how much to scale video based on the original size. 1.0 zoom means it is not scaled and just placed in the UI. Since video's height (1208) is bigger than UI widget height (1080), the top and bottom of the video extends past the widget frame.

But ui.cc assumes that 1.0 scale means that the video is always scaled down (maintaining aspect ratio) to fit into the UI widget.

This mismatch causes ecams with high value calibs to not be drawn properly. This PR makes it so that `zoom` in cameraview.cc is in respect to the UI frame size, not the video size.

### DIFF
--- a/selfdrive/ui/qt/onroad.cc
+++ b/selfdrive/ui/qt/onroad.cc
@@ -454,7 +454,6 @@ void AnnotatedCameraWidget::updateFrameMat() {
   // 1) Put (0, 0) in the middle of the video
   // 2) Apply same scaling as video
   // 3) Put (0, 0) in top left corner of video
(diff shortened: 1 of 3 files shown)


sshane:
don't think this is right

PR #29409, opened 2023-08-15
Subaru: calculate steering angle rate

jnewb1:
angle rate has yet to be found in any can signal, and doesn't appear in the diagnostic software.

prereq to https://github.com/commaai/openpilot/pull/29116


selfdrive/controls/lib/drive_helpers.py
```diff
@@ -211,3 +211,27 @@ def get_speed_error(modelV2: log.ModelDataV2, v_ego: float) -> float:
     vel_err = clip(modelV2.temporalPose.trans[0] - v_ego, -MAX_VEL_ERR, MAX_VEL_ERR)
     return float(vel_err)
   return 0.0
+
+
+class CanSignalRateCalculator:
```
adeebshihadeh, line 216: this should live in selfdrive/car since that's where CAN signals are parsed

selfdrive/controls/lib/drive_helpers.py
```diff
@@ -211,3 +211,27 @@ def get_speed_error(modelV2: log.ModelDataV2, v_ego: float) -> float:
...
+  """
+  Calculates the instantaneous rate of a can signal by using the counter
```
adeebshihadeh, line 218: can -> CAN like it is on the next line. make sure to stay consistent

selfdrive/car/subaru/carstate.py
```diff
@@ -13,6 +14,8 @@ def __init__(self, CP):
     can_define = CANDefine(DBC[CP.carFingerprint]["pt"])
     self.shifter_values = can_define.dv["Transmission"]["Gear"]
 
+    self.angle_rate_calulator = CanSignalRateCalculator(2)
```
adeebshihadeh, line 17: think it makes more sense to input the frequency. you'd need to know what DT_CTRL is while reading this

selfdrive/controls/lib/drive_helpers.py
```diff
@@ -211,3 +211,27 @@ def get_speed_error(modelV2: log.ModelDataV2, v_ego: float) -> float:
...
+  variable and the known period of the CAN message that contains it.
+  """
+  def __init__(self, period):
+    self.period = period
+    self.previous_counter = 0
```
adeebshihadeh, line 223: reference and previous are confusing. can't this be done with only one pair?

selfdrive/controls/lib/drive_helpers.py
jnewb1: done

selfdrive/controls/lib/drive_helpers.py
jnewb1: done

selfdrive/car/subaru/carstate.py
jnewb1: done

selfdrive/controls/lib/drive_helpers.py
jnewb1: Yep is can be done with 2 if you store the rate

jnewb1: done

selfdrive/car/subaru/carstate.py
```diff
@@ -47,6 +50,11 @@ def update(self, cp, cp_cam, cp_body):
     ret.gearShifter = self.parse_gear_shifter(self.shifter_values.get(can_gear, None))
 
     ret.steeringAngleDeg = cp.vl["Steering_Torque"]["Steering_Angle"]
+
+    if self.car_fingerprint not in PREGLOBAL_CARS:
```
adeebshihadeh, line 54: no reason to gate this unless we had a better signal for PREGLOBAL

jnewb1: preglobal doesn't have a counter

adeebshihadeh (approved):

jnewb1:
going to try and rerun the replay test, but we might need this first:
 https://github.com/commaai/openpilot/pull/29411

selfdrive/car/__init__.py
```diff
@@ -213,3 +213,24 @@ def __init__(self, CP, fingerprint: Optional[Dict[int, Dict[int, int]]]) -> None
...
+  """
+  def __init__(self, frequency):
+    self.frequency = frequency
+    self.previous_counter = 0
+    self.previous_value = 0
+    self.rate = 0
+
+  def update(self, current_value, current_counter):
+    if current_counter != self.previous_counter:
+      self.rate = (current_value - self.previous_value) * self.frequency
```
sshane, line 231: You can also check vl_all to know if there is a new value, might be more generic

PR #29410, opened 2023-08-15
CI: ignore most of repo when building docker images

jnewb1:
Docker has to copy most of the repo into the docker context atm, but these two dockerfiles only use a few scripts. Should speed up the build and also might prevent cache misses on buildjet


adeebshihadeh:
Doesn't BuildKit skip the copy? I think this is only relevant for the legacy docker builder.

PR #29411, opened 2023-08-15
update azure-storage-blob

adeebshihadeh:
https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/storage/azure-storage-blob/migration_guide.md


adeebshihadeh:
Looks like the API changed. @incognitojam can you finish this up?

PR #29414, opened 2023-08-15
thermald: type annotate main thread

sshane:
function signatures are required for typing checks, unless we enable all. types inside untyped function is a warning: https://github.com/python/mypy/issues/3948

couple issues:

```python
selfdrive/thermald/thermald.py:276: error: Right operand of "or" is never evaluated  [unreachable]
selfdrive/thermald/thermald.py:360: error: Statement is unreachable  [unreachable]
```


sshane:
seems like a bug with mypy, it actually is reachable. the first loop it will always be none, and so it is treated as never reachable https://github.com/python/mypy/issues/11612

PR #29415, opened 2023-08-15
Subaru gen2: limit jerk by reducing steer delta

jnewb1:
We currently limit gen2 cars to about half of gen1, but the delta is not half, which causes the jerk to be too high.

We might want to redo the lateral acceleration tests with the outback after the steering alignment was fixed, because it probably isn't actually that extreme:
https://github.com/commaai/openpilot/pull/29097

Should also limit panda to the same


jnewb1:
Interestingly, using lower limits here causes steering faults, seems the be the panda is blocking the messages when there is some driver torque. I'll look into it more:

c726b2698bd24c5e|2023-08-19--13-57-25

Issue #29298, opened 2023-08-09
pyenv initialization is duplicated


gregjhogan:
### Describe the bug

I believe this code was added so that if you delete your openpilot repo it doesn't break your pyenv installation.
https://github.com/commaai/openpilot/blob/7fdd9fc37c1c5febf1cd3dc86441ebf0dec7f370/update_requirements.sh#L16-L24

We also have very similar code here:
https://github.com/commaai/openpilot/blob/c53545621e865dbee2661442f03f00152c48b92c/tools/openpilot_env.sh#L2-L19

So I think we want to remove the duplication from https://github.com/commaai/openpilot/blob/master/tools/openpilot_env.sh which potentially only leaves mac specific things in this script (`eval "$(pyenv virtualenv-init -)"` will work on macs once `pyenv-virtualenv` is installed).

### OS Version

all

gregjhogan:
https://github.com/commaai/openpilot/pull/23408 which added the `.pyenvrc` file mentions `eval "$(pyenv init --path)"` no longer appears to be needed.

PR #29419, opened 2023-08-16
setup cleanup

gregjhogan:
It seems like the following cleanup can now happen:
- we switched pyenv initialization from `openpilot_env.sh` to `~/.pyenv.rc` [long ago](https://github.com/commaai/openpilot/commit/f2f8e7b930ec9c8ca2d264076700933bee7fa7d0), so we can [eliminate this duplication](https://github.com/commaai/openpilot/issues/29298)
- It looks like we have committed to using `.env` for setting environment variables required for running openpilot, so we may as well move all environment variables there, which eliminates the `openpilot_env.sh` script as well as injecting anything openpilot specific into a shell rc/profile script.


gregjhogan:
@fredyshox can you test going through the initial setup (as close to a clean install as you can get) on your mac?  This should be good now but I would like someone else to verify because I found CI ignores things like environment variables from your shell profile script.  I tested this on a mac and and everything seems good, but it would be great if someone else can validate, too.

tools/install_python_dependencies.sh
```diff
@@ -14,21 +14,19 @@ if ! command -v "pyenv" > /dev/null 2>&1; then
...
   cat <<EOF > "${HOME}/.pyenvrc"
 if [ -z "\$PYENV_ROOT" ]; then
   export PATH=\$HOME/.pyenv/bin:\$HOME/.pyenv/shims:\$PATH
   export PYENV_ROOT="\$HOME/.pyenv"
+  eval "\$(pyenv init --path)"
   eval "\$(pyenv init -)"
   eval "\$(pyenv virtualenv-init -)"
 fi
 EOF
+  echo -e "\nsource ~/.pyenvrc" >> $RC_FILE
```
fredyshox, line 26: mac setup installs pyenv from `brew`, so .pyenvrc will never be created (nor this .bashrc line) - same on any other machine when pyenv is already installed

gregjhogan: Yeah that is fine as far as I can tell.  This part doesn't run if pyenv was installed by other means and everything still works (since the alternative installation sets up the same things).

adeebshihadeh (approved):

jnewb1:
i think this broke something on the mac CI build, probably after the macos cache was pushed since it looks like it passed the first time. Thought it was my CI change but I tried reverting it and it's still failing. https://github.com/commaai/openpilot/pull/29450

@fredyshox want to take a look?

fredyshox:
Seems like PYTHONPATH export was deleted

adeebshihadeh:
Sure, but before we merge this again, we have to fix the mac CI passing without it really passing.

fredyshox:
Ok, I think the issue here was, that `.env` file (which is created during setup) was not added to cache list. Therefore the subsequent CI jobs using new cache would not have it @adeebshihadeh this might be hard to enforce, any idea how to prevent that in the future?

PR #29421, opened 2023-08-16
Unpin numpy

fredyshox:
Acados was updated in https://github.com/commaai/openpilot/commit/4c1b8b71d096e7658fe56598944e3f8bd3e347f0, numpy does not need to be pinned anymore.


pyproject.toml
```diff
@@ -25,13 +25,13 @@ cffi = "*"
 crcmod = "*"
 cryptography = "*"
 Cython = "*"
-future-fstrings = "*"  # for acados
```
adeebshihadeh, line 28: this comment is still relevant

fredyshox:
Note for future: Acados does not need to be recompiled for that, the only thing from there which relies on numpy is `acados_template/acados_ocp_solver_pyx.pyx`, which is recompiled every time solver is generated

PR #29423, opened 2023-08-16
test_models: check panda cruise state

sshane:


sshane:
will test on route list and merge!

PR #29429, opened 2023-08-16
Fix PyQT5 on non-x64 platforms

fredyshox:


fredyshox:
Doesn't work, its not possible to install pyqt5 from source using pip

adeebshihadeh:
Make sure to delete the branch after closing a PR

PR #29430, opened 2023-08-16
Disable pyqt5 on aach64

fredyshox:
There're no wheels for aach64

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,7 +101,7 @@ types-pycurl = "*"
 types-PyYAML = "*"
 types-requests = "*"
 types-tabulate = "*"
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -101,7 +101,7 @@ types-pycurl = "*"
 types-PyYAML = "*"
 types-requests = "*"
 types-tabulate = "*"
-pyqt5 = "*"
+pyqt5 = { version = "*", markers = "platform_machine == 'x86_64'" }
```
adeebshihadeh, line 104: when we write exceptions like this, make sure to leave a good comment explaining why so it's easy to resolve later

PR #29435, opened 2023-08-16
CI: add option to build a scons cache docker image

jnewb1:
If we build a docker image that contains the scons cache, we can use --mount in buildkit to use this cache in other docker images, without it bloating the size of the image

### DIFF
--- a/Dockerfile.scons_cache
+++ b/Dockerfile.scons_cache
@@ -1,0 +1,3 @@
+FROM alpine:3
+
+COPY ./scons_cache /tmp/scons_cache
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
Might be nice to break out all those docker helper environment variables into scripts in order to make running locally easier.

Dockerfile.scons_cache
```diff
@@ -0,0 +1,3 @@
+FROM alpine:3
+
+COPY ./scons_cache /tmp/scons_cache
```
adeebshihadeh, line 3: I don't want to pollute the root with a file like this. let's either move or generate on the fly

adeebshihadeh:
Btw, this doesn't quite seem to make sense to split up. If I understand this right, this PR on its own doesn't do anything?

jnewb1:
> Btw, this doesn't quite seem to make sense to split up. If I understand this right, this PR on its own doesn't do anything?

yep it just adds the option to setup, but its disabled by default. I'll just close this one and move to https://github.com/commaai/openpilot/pull/29436

PR #29445, opened 2023-08-17
Added script for tracking mypy coverage

SheyarAgha:
Changes made in regards to "track mypy coverage" issue (https://github.com/commaai/openpilot/issues/24623)

Description : Created a mypy coverage tracking script that keeps dated records of each mypy coverage check. When run, it also compares the latest coverage check with the prior one, giving feedback on coverage fluctuations after commits. 

Verification: Tested the script on "common" directory and verified all is working as intended. 

# Script is currently still set up to only check "common" directory for simplicity of verification.

### DIFF
--- a/mypy-coverage.py
+++ b/mypy-coverage.py
@@ -1,0 +1,55 @@
+import subprocess
+import re
+import datetime
+import time
+import glob
+
+
+def get_coverage_from_report(file_dir):
+  """Extracts coverage percentage from mypy report"""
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
Thanks for the PR, but I ended up closing that issue (see #24623). The real thing to do here is enable stricter options in mypy (#29458). That issue's up for grabs if you wanna take a stab at it.

PR #29451, opened 2023-08-17
Revert "setup cleanup (#29419)"

jnewb1:
seems to have broken macos, we can merge it again after it's fixed


adeebshihadeh:
@fredyshox can you also look into how this had passed before in the PR and the merged commit on master? seems like some state is not being properly managed in the mac CI

jnewb1:
passes macos test now

PR #29416, opened 2023-08-15
interfaces: add common STD_CARGO_KG to mass

jyoung8607:
**Description**

Add STD_CARGO_KG to vehicle mass once, instead of every single time we set vehicle mass.

* ~~special handling for Civic in `selfdrive/car/__init__.py` to prevent circular references~~
* special handling for comma body in `selfdrive/car/interfaces.py` so it doesn't carry a human around

GitHub will tell you this PR is +5, but if you listen to your heart 👩‍🎤 it's -140.

**Verification**

```python
#!/usr/bin/env python3

from selfdrive.car import gen_empty_fingerprint
from selfdrive.car.car_helpers import interfaces
from selfdrive.car.fingerprints import all_known_cars

if __name__ == "__main__":
  for car in all_known_cars():
    CarInterface, CarController, CarState = interfaces[car]
    car_fw = []
    fingerprints = gen_empty_fingerprint()
    CP = CarInterface.get_params(car, fingerprints, car_fw, experimental_long=False, docs=False)
    print(f"{CP.carFingerprint}: {CP.mass=}")
```

[mass_before.txt](https://github.com/commaai/openpilot/files/12353665/mass_before.txt)
[mass_after.txt](https://github.com/commaai/openpilot/files/12353666/mass_after.txt)

**Differences**

Some cars with mass listed in pounds (and therefore multiplied by `CV.LB_TO_KG`) change very slightly because the cereal Float32 conversion is now done before adding `STD_CARGO_KG` instead of afterward.

```
GMC ACADIA DENALI 2018: CP.mass=2110.486083984375
GMC ACADIA DENALI 2018: CP.mass=2110.48583984375
LEXUS RX 2016: CP.mass=2168.545654296875
LEXUS RX 2016: CP.mass=2168.5458984375
LEXUS RX 2020: CP.mass=2168.545654296875
LEXUS RX 2020: CP.mass=2168.5458984375
LEXUS RX HYBRID 2017: CP.mass=2168.545654296875
LEXUS RX HYBRID 2017: CP.mass=2168.5458984375
LEXUS RX HYBRID 2020: CP.mass=2168.545654296875
LEXUS RX HYBRID 2020: CP.mass=2168.5458984375
TOYOTA ALPHARD 2020: CP.mass=2088.713623046875
TOYOTA ALPHARD 2020: CP.mass=2088.71337890625
TOYOTA ALPHARD HYBRID 2021: CP.mass=2088.713623046875
[...]

PR #29459, opened 2023-08-18
Rename CivicParams

sshane:
split from https://github.com/commaai/openpilot/pull/29416


selfdrive/car/__init__.py
```diff
@@ -233,4 +232,4 @@ def update(self, current_value, current_counter):
     self.previous_counter = current_counter
     self.previous_value = current_value
 
-    return self.rate
\ No newline at end of file
+    return self.rate
```
sshane, line 235:
```suggestion
    return self.rate```

PR #29463, opened 2023-08-18
ui/settings: Don't apply styles to more elements than needed.

deanlee:
Don't apply border-raduis, backgound-color to all child items.
| Before  | After |
| Screenshot from 2023-08-18 17-02-47  | Screenshot from 2023-08-18 17-08-46  |
| Screenshot from 2023-08-18 17-02-27  | Screenshot from 2023-08-18 17-08-37  |
| Screenshot from 2023-08-18 17-02-01  | Screenshot from 2023-08-18 17-08-28  |

### DIFF
--- a/selfdrive/ui/qt/offroad/settings.cc
+++ b/selfdrive/ui/qt/offroad/settings.cc
@@ -345,10 +345,6 @@ SettingsWindow::SettingsWindow(QWidget *parent) : QFrame(parent) {
   QVBoxLayout *sidebar_layout = new QVBoxLayout(sidebar_widget);
   sidebar_layout->setMargin(0);
   panel_widget = new QStackedWidget();
-  panel_widget->setStyleSheet(R"(
-    border-radius: 30px;
-    background-color: #292929;
-  )");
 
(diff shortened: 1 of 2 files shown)


adeebshihadeh:
What does this do? Improve performance? Is there a good way to debug this or prevent introducing it in the future?

deanlee:
> What does this do? Improve performance? 

Improve performance.  before this pr, all widgets in settings need to draw a filled rounded rect before draw the text.
After this PR, most widgets no longer need to fill the background when drawing. This also improves performance when scrolling.

> Is there a good way to debug this or prevent introducing it in the future?

this is a common css mistake and difficult to debug. I haven't thought of any good way yet.

adeebshihadeh:
The difference is noticeable even on PC!

sshane:
Feels relatively the same on a three. @deanlee want to see if there's anything else obviously wrong we're doing to improve scrolling? Even on PC it skips and jumps around when scrolling, even views that don't update their content like the toggles pane. Might be a setting we need to change

PR #29464, opened 2023-08-18
ui/driverview: hide on interactive timeout

deanlee:
fis issue https://github.com/commaai/openpilot/pull/27457#issuecomment-1681631826:

> after https://github.com/commaai/openpilot/pull/28656, when it does time out, the driver view stays up and tapping once doesn't close it. If there's a signal that gets emitted when the timeout happens, we should close the driver view then.


selfdrive/ui/qt/offroad/driverview.cc
```diff
@@ -19,11 +19,19 @@ DriverViewWindow::DriverViewWindow(QWidget* parent) : QWidget(parent) {
   connect(cameraView, &CameraWidget::vipcThreadFrameReceived, scene, &DriverViewScene::frameUpdated);
   layout->addWidget(scene);
   layout->setCurrentWidget(scene);
+
+  QObject::connect(device(), &Device::interactiveTimeout, this, &DriverViewWindow::requestDone);
+}
+
+void DriverViewWindow::requestDone() {
```
sshane, line 26:
```suggestion
void DriverViewWindow::closeView() {
```
?

sshane:
I assume `hideEvent` isn't fired when the screen turns off since it stays up, right?

deanlee:
> I assume `hideEvent` isn't fired when the screen turns off since it stays up, right?

yes

sshane:
when the 60s timer runs out, it sets the timeout to 15s, so the screen turns off after 60+30s

deanlee:
think we should not call resetInteractiveTimeout in hideEvent. just let MainWindow::eventFilter reset it on clicks.

PR #29466, opened 2023-08-18
CI: run build_mac twice, if setup script changed

fredyshox:
This would run the build_mac job twice, to test a cache, if any of the setup scripts changed. 

This would prevent a situation like this: https://github.com/commaai/openpilot/pull/29456, where the first run succeeds, but the one after that, fails due to missing files in build cache.

### DIFF
--- a/.github/workflows/selfdrive_tests.yaml
+++ b/.github/workflows/selfdrive_tests.yaml
@@ -112,10 +112,10 @@ jobs:
         run: |
           if [[ ${{ steps.filter.outputs.setup_scripts }} == true ]]; then
             # setup scripts have changed, run the build_mac job twice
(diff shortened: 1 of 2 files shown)


fredyshox:
@adeebshihadeh any thoughts on this?

I don't see any other option for testing macos build cache, other than building it twice - once to rebuild the cache, and the to test if it builds with it.

adeebshihadeh:
Is the issue with the cache itself or restoring a bad cache?

fredyshox:
Restoring cache works well, the issue is when cache is incomplete - usually caused by someone forgetting to add some paths to it. Then this step fails - because we skip any setup/installation step if we hit cache, to speed it up.

adeebshihadeh:
This seems pretty janky, and we don't have to do this for our other jobs' caches. Can we just make the bug harder to write?

fredyshox:
I agree it feels janky...

PR #29467, opened 2023-08-18
ui/OffroadHome: cleanup layout and style sheet

deanlee:
| Before  | After |
| Screenshot 2023-08-19 03:56:03  | Screenshot 2023-08-19 03:56:54  |

1. Remove unnecessary widget containers to reduce overhead.
2. cleanup style sheet, use same border-radius & background-color for all children.
3. fixed a stretch issue:


sshane:
I think the stretch issue is intentional design

PR #29468, opened 2023-08-18
agnos manifest: add alt system image

incognitojam:
add alternate image file for commaai/flash

### DIFF
--- a/system/hardware/tici/agnos.json
+++ b/system/hardware/tici/agnos.json
@@ -7,7 +7,11 @@
     "size": 15636480,
     "sparse": false,
     "full_check": true,
-    "has_ab": true
+    "has_ab": true,
+    "alt": {
+      "hash": "8d8d8620de8b2687f3a8fffdb81b2abd1fe2ead5bc831361a1a212e5589ac279",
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
We'll likely have to make the flasher point to master until this gets in with the next release.

system/hardware/tici/agnos.json
```diff
@@ -17,7 +21,11 @@
     "size": 274432,
     "sparse": false,
     "full_check": true,
-    "has_ab": true
+    "has_ab": true,
+    "alt": {
```
adeebshihadeh, line 25: why do these partitions also have alts?

incognitojam: Only because they are gzips. I'm happy to remove and continue changing the file extension like before if you want

adeebshihadeh: oh, think it's best to just do the replace with that. at some point, think we want to fix the JS xz extract and don't want to do a manifest update when we ship that

incognitojam: I'll remove the .gz from the manifest, including changing the alt system image to .xz, and flash can handle it

PR #29470, opened 2023-08-19
ui/map: interactive counter fix

### LINKED ISSUE #28816: nav: zooming improvements
- [x] it looks like it continues to zoom when you pan manually, it probably should stop (like it does if you manually zoom) - https://github.com/commaai/openpilot/pull/29470
- [ ] can also reset the filter when it resets the view/re-centers
- [x] filter too slow, or set a breakpoint above 0, as it continues to zoom when you come to a stop - https://github.com/commaai/openpilot/pull/29601
- [x] fix zoom starting from 0 m/s when you get a fix while driving - https://github.com/commaai/openpilot/pull/29605

_From https://github.com/commaai/openpilot/issues/28796#issuecomment-1622629543_

jakethesnake420:
When the user zooms out to pan a long distance, the zoom counter will cause the view to zoom back in even if the user is still panning around. The same thing occurs for zooming without panning, the center will reset to the last_position. This PR fixes that behavior by combining the counters into a single interaction_counter.

fix https://github.com/commaai/openpilot/issues/28816

### DIFF
--- a/selfdrive/ui/qt/maps/map.cc
+++ b/selfdrive/ui/qt/maps/map.cc
@@ -193,12 +193,12 @@ void MapWindow::updateState(const UIState &s) {
     m_map->updateSource("carPosSource", carPosSource);
   }
 
-  if (interaction_counter != 0){
-    interaction_counter--;
-  } else {
+  if (interaction_counter == 0) {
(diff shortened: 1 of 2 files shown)


sshane:
nice PR!

PR #28479, opened 2023-06-10
Hyundai longitudinal (CAN): don't send FCA11 on non-FCA cars

sunnyhaibin:
**Prerequisite of**
- https://github.com/commaai/openpilot/pull/26967

PR #26967, opened 2023-01-15
Hyundai longitudinal: enable for IONIQ, KONA_EV, IONIQ_PHEV, and KIA_CEED

sunnyhaibin:
**Checklist**
- [x] updated entry in CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route updated in [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x]  route with openpilot longitudinal:
  - [x] `CAR.IONIQ` - Hyundai Ioniq Hybrid 2017-19: `012c95f06918eca4|2023-01-15--11-19-36`
    - Note: Steering low speed lockout of 32 MPH still exists
  - [x] `CAR.KONA_EV` - Hyundai Kona Electric 2018-21: `f90d3cd06caeb6fa|2023-01-15--18-02-31`
  - [x] `CAR.IONIQ_PHEV` - Hyundai Ioniq Plug-in Hybrid 2020-21: `79c52770a148c54d|2023-02-02--23-04-40`
  - [x] `CAR.CEED` - Kia Ceed 2019: `99315f265be0b26c|2023-06-08--22-04-49`

**Requires**
- https://github.com/commaai/openpilot/pull/28479
- https://github.com/commaai/openpilot/pull/29484
- https://github.com/commaai/openpilot/pull/29489

After some trials and testing, some platforms in the above list can now use the same radar disable command on `0x7D0` to disable the SCC radar unit. They can now accept openpilot longitudinal commands, with the ability to slow down to 0, which removes the limitation of low speed cutout at 8 km/h with stock SCC.

Stock AEB and FCW are disabled with this method, so we show the amber indicator on the car's dashboard to match stock when we manually disable AEB/FCW.

openpilot only sends `SCC11, SCC12, SCC13, SCC14` and does not send `FCA11, FCA12` messages to cars that do not broadcast `FCA11, FCA12`. Sending `FCA11, FCA12` to those cars will cause a fault `TCS13 | ACCEnable = 3`.

Thanks to community 2019 Hyundai Ioniq owner andy741217, 2020 Hyundai Kona Electric owner `ivan iv#6513` (Discord), 2020 Hyundai Ioniq Plug-in Hybrid owner `albertcalderon1#8886` (Discord).

**Amber indicator of AEB/FCW disabled**
image

PR #29472, opened 2023-08-19
add Kia Niro Hybrid 2019 to legacy safety mode list

sshane:
going to make a new list, no point in regressing the safety coverage

~9 dongles last 500 days, 10k qlogs, no one uses openpilot longitudinal control. Since this car [never has FCA11](https://github.com/commaai/openpilot/pull/28479#issuecomment-1683423682), we should remove it from the ability to use it until the FCA11 issue is fixed, as it possibly faults like the other cars without FCA11 (related: https://github.com/commaai/openpilot/pull/26967)

so we can merge https://github.com/commaai/openpilot/pull/28479 without worrying about this car faulting and/or not showing AEB disabled


sshane:
@sunnyhaibin we can now remove cars from the legacy safety mode without verifying they work with longitudinal (test_models should be sufficient). feel free to try some and open PRs

PR #21108, opened 2021-06-02
Add Elantra to legacy safety model

sshane:
Getting @archlyg to test this

PR #29473, opened 2023-08-19
Hyundai: enable more checks for some CAN cars

sshane:
thanks to https://github.com/commaai/openpilot/pull/29472

these cars were grouped into a legacy safety mode which also doesn't allow longitudinal, but we can disallow longitudinal in openpilot while still checking the counters and checksums


sshane:
Not sure why we did this, I don't see any issues in the old preserved routes: https://github.com/commaai/openpilot/pull/21108

sshane:
good enough for now, other cars aren't too popular

PR #29477, opened 2023-08-19
Nissan: add alt request for engine on Rogue

sshane:
I believe this is the only regression for Nissan from https://github.com/commaai/openpilot/pull/28634, the engine only responded to a few diagnostic session types. Found 0xda by trying all values for this byte, this was the first that worked.

New route: b7287535e2a23992|2023-08-18--20-55-56
Route with 0x81 diagnostic type: b7287535e2a23992|2023-07-23--13-00-19
Route with original 0xc0 diagnostic type: b7287535e2a23992|2023-03-19--18-48-25

Thanks to icecube8781 on Discord for the testing!


selfdrive/car/nissan/values.py
```diff
@@ -99,6 +104,10 @@ class NissanCarInfo(CarInfo):
       [NISSAN_DIAGNOSTIC_RESPONSE_KWP, NISSAN_VERSION_RESPONSE_KWP],
       rx_offset=NISSAN_RX_OFFSET,
     ),
+    Request(
```
sshane, line 107:
```suggestion
    # Rogue's engine solely responds to this
    Request(
```

PR #29480, opened 2023-08-19
Toyota: more accurate pedal compensation

sshane:


selfdrive/car/toyota/carstate.py
```diff
@@ -134,6 +134,10 @@ def update(self, cp, cp_cam):
     ret.cruiseState.enabled = bool(cp.vl["PCM_CRUISE"]["CRUISE_ACTIVE"])
     ret.cruiseState.nonAdaptive = cp.vl["PCM_CRUISE"]["CRUISE_STATE"] in (1, 2, 3, 4, 5, 6)
 
+    self.pcm_acc_braking = cp.vl["PCM_CRUISE"]["ACC_BRAKING"]
+    self.pcm_accel_net = cp.vl["PCM_CRUISE"]["ACCEL_NET"]
```
sshane, line 138:
```suggestion
```

sshane:
@haraschax not tested in a car with a pedal yet, but what do you think?

adeebshihadeh:
We really shouldn't be spending time on pedals.

PR #29481, opened 2023-08-19
Toyota: compensate for PCM

sshane:
PCM sometimes does weird modifications to our accel request to convert it into gas & brake. Some are explainable (like the force to overcome rolling/air resistence), some are not (laggy to brake when we're clearly requesting it), and some are undesireable (pitch compensation is lagged and imprecise).


cydia2020:
I think I've this figured out, https://github.com/cydia2020/dodgypilot/pull/15
`ATACC2 = AT_RAW + ACCAVL / self.CP.mass`, this way the car knows what openpilot is doing and gives openpilot a feedback force to work with. Now both decel and accel are extremely smooth, and the car behaves like stock where it releases the brake slightly upon stopping to decrease the "jerk" when the vehicle stops.

PR #29482, opened 2023-08-19
Keep the car icon aligned to its bearing

jakethesnake420:
When user is rotating the map, the car icon should stay fixed on the cars bearing, not m_map->bearing().
The car icon is setup to align to the 'map' bearing but this is not updated when interactive_counter != 0. 
This PR fixes this issue by calculating the error between the cars bearing and the m_map->bearing() and rotates the icon.

Before: (In this picture the icon-pitch-alignment setting was set to 'viewport', normally it's 'map')
image

After:
image


sshane:
thanks!

PR #29484, opened 2023-08-19
Hyundai longitudinal (CAN): display AEB indicator on non-FCA cars

sunnyhaibin:
**Prerequisite of**
- https://github.com/commaai/openpilot/pull/26967


sshane:
The only car this currently affects is one dongle of `CAR.IONIQ_EV_2020` according to https://github.com/commaai/openpilot/pull/28479#issuecomment-1683423682, as all other non-FCA cars are in legacy safety mode or unsupported longitudinal list.

The Ioniq was removed from legacy safety mode in https://github.com/commaai/openpilot/pull/29531, because that user had the FCA message and didn't require any of this special logic to work correctly. The other dongle without FCA likely is in the same boat as these cars, so this should be fine to merge as-is.

sshane:
@sunnyhaibin let me know if I got any details in the comment wrong!

PR #29489, opened 2023-08-20
Hyundai longitudinal (CAN): detect and conditionally send `FCA12`

sunnyhaibin:
Per discussion with @sshane on Discord, we need to detect and conditionally send `FCA12` to HKG CAN platforms that have `FCA12`.

**Prerequisite of**
- https://github.com/commaai/openpilot/pull/26967

### DIFF
--- a/selfdrive/car/hyundai/carcontroller.py
+++ b/selfdrive/car/hyundai/carcontroller.py
@@ -182,7 +182,8 @@ class CarController:
 
       # 5 Hz ACC options
       if self.frame % 20 == 0 and self.CP.openpilotLongitudinalControl:
-        can_sends.extend(hyundaican.create_acc_opt(self.packer))
+        send_fca12 = self.CP.flags & HyundaiFlags.SEND_FCA12.value
+        can_sends.extend(hyundaican.create_acc_opt(self.packer, send_fca12))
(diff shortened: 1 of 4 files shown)


sshane:
Detecting a new signal usually requires some verification, like measuring its startup time, to ensure we don't not detect it (the toyota gas pedal signal took too long to start, so we had to hard code it by platform). I would consult the list I posted in the other PR to see if it's safe to use FCA11 now that we know which platforms it's on.

sunnyhaibin:
List for reference: https://github.com/commaai/openpilot/pull/28479#issuecomment-1683423682

sshane:
We only need to detect FCA12 if we plan to enable longitudinal on the cars where FCA12 != FCA11 existing. That is only one in your list in https://github.com/commaai/openpilot/pull/26967; `HYUNDAI KONA ELECTRIC 2019`. Can you test on that platform still sending `FCA12`?

sshane:
FCA12 is just a user setting mode message largely, perhaps we need to set/unset other bits to not fault on the cars where it does now.

sshane:
You can also omit the Kona from the PR and just merge individual cars (preferred), that already have FCA.

PR #29490, opened 2023-08-20
update clone instructions

adeebshihadeh:


haraschax:
You need to add the submodule commands for the full clone

PR #29491, opened 2023-08-20
ui: use `enum PrimeType` for prime_type

deanlee:


selfdrive/ui/ui.cc
```diff
@@ -264,7 +264,7 @@ void UIState::update() {
   emit uiUpdate(*this);
 }
 
-void UIState::setPrimeType(int type) {
+void UIState::setPrimeType(PrimeType type) {
   if (type != prime_type) {
```
incognitojam, line 268: I think behaviour has changed since `prime_type` defaults to `PrimeType::NONE` it might not get written to params/event emitted

deanlee: y, this might be a bug.but It looks like the original version also had this issue: the prime_type is set to none in UIState::UIState if PrimeType not existed:
`prime_type = std::atoi(params.get("PrimeType").c_str());`

deanlee: added a new enum value PrimeType::UNKNOWN to fix this issue, what do you think?

incognitojam: that seems fine

incognitojam: should we add a helper for `hasPrime` in `UIState`?

sshane:
is this good to go?

PR #29498, opened 2023-08-21
add openpilot prefix to imports

adeebshihadeh:


adeebshihadeh:
Seems the "PR comments" CI job won't pass until this is merged since the import path is part of the enum's value

PR #29502, opened 2023-08-21
Remove Kia Sorento from LEGACY_SAFETY_MODE_CAR

kalub92:
# Description

After seeing @sshane's Discord [message](https://discord.com/channels/469524606043160576/524611978208215070/1142340953412223067) and [open PR](https://github.com/commaai/openpilot/pull/29473) regarding disallowing longitudinal in Openpilot while still checking the counters and checksums for cars that have historically been in `LEGACY_SAFETY_MODE_CAR`.

I ran his branch (`hkg-remove-from-legacy`) in my 2019 Kia Sorento.

# Verification

I verified the following cases with alpha longitudinal enabled/disabled:

### Alpha long disabled ❌ 
**Route:** `88da9d94a8980684|2023-08-19--23-47-41--0`
> No issues observed while removed from `LEGACY_SAFETY_MODE_CAR`.

### Alpha long enabled ✅
**Route:** `88da9d94a8980684|2023-08-19--23-52-14--0`
> `Harness Relay Malfunction` error and repeated errors in cluster, but I expected this because `disable_radar` doesn't work for my car at the moment.

### DIFF
--- a/selfdrive/car/hyundai/values.py
+++ b/selfdrive/car/hyundai/values.py
@@ -1939,8 +1939,8 @@ EV_CAR = {CAR.IONIQ_EV_2020, CAR.IONIQ_EV_LTD, CAR.KONA_EV, CAR.KIA_NIRO_EV, CAR
 
 # these cars require a special panda safety mode due to missing counters and checksums in the messages
 LEGACY_SAFETY_MODE_CAR = {CAR.HYUNDAI_GENESIS, CAR.IONIQ_EV_2020, CAR.IONIQ_EV_LTD, CAR.IONIQ_PHEV, CAR.IONIQ, CAR.KONA_EV,
-                          CAR.KIA_SORENTO, CAR.SONATA_LF, CAR.KIA_OPTIMA_G4, CAR.KIA_OPTIMA_G4_FL, CAR.VELOSTER,
-                          CAR.GENESIS_G70, CAR.GENESIS_G80, CAR.KIA_CEED, CAR.ELANTRA, CAR.IONIQ_HEV_2022}
+                          CAR.SONATA_LF, CAR.KIA_OPTIMA_G4, CAR.KIA_OPTIMA_G4_FL, CAR.VELOSTER, CAR.GENESIS_G70, CAR.GENESIS_G80,
+                          CAR.KIA_CEED, CAR.ELANTRA, CAR.IONIQ_HEV_2022}
 
 # these cars have not been verified to work with longitudinal yet - radar disable, sending correct messages, etc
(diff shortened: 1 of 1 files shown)


selfdrive/car/hyundai/values.py
```diff
@@ -1939,7 +1939,7 @@ def match_fw_to_car_fuzzy(live_fw_versions) -> Set[str]:
 
 # these cars require a special panda safety mode due to missing counters and checksums in the messages
 LEGACY_SAFETY_MODE_CAR = {CAR.HYUNDAI_GENESIS, CAR.IONIQ_EV_2020, CAR.IONIQ_EV_LTD, CAR.IONIQ_PHEV, CAR.IONIQ, CAR.KONA_EV,
-                          CAR.KIA_SORENTO, CAR.SONATA_LF, CAR.KIA_OPTIMA_G4, CAR.KIA_OPTIMA_G4_FL, CAR.VELOSTER,
```
sshane, line 1942: Make sure to add to unsupported long cars, since the radar disable doesn't work yer

sshane (changes requested):

selfdrive/car/hyundai/values.py
kalub92: Right, thanks for calling that out  – [done](https://github.com/commaai/openpilot/pull/29502/commits/bbe123d1f9dd26a2ec7f94512c1629396a68b496).

sshane:
thanks! let us know if you try to work on disabling the radar! it probably would enable a lot of these cars to use longitudinal (our Optima is one of them)

kalub92:
> thanks! let us know if you try to work on disabling the radar! it probably would enable a lot of these cars to use longitudinal (our Optima is one of them)

I actually am working on it! I've been talking back-and-forth with @sunnyhaibin for a few months about this and we've been trying tons of different things.

It seems my car (and likely the others that don't currently work) are using some form of sensor fusion between the radar and the lane keep assist camera and that both SCC and AEB are somehow managed in this combination of ECUs and possibly some parent ECU that controls the two of those.

In terms of sending the `disable_ecu` UDS command, I've confirmed that sending it without suppressing the response (`02 28 03`) to 0x7d0 (my radar address) returns a positive response (`02 68 03`) and yet SCC messages are still present in Cabana.

Something strange I discovered is that my car has both an SCC fuse (under the hood) and an AEB fuse (beneath the cluster).

To test what happens if I were to unplug the radar, I unplugged the SCC fuse, but I still see all SCC messages on bus 0. However, if I pull the AEB fuse, those messages are silenced completely (even if I leave in the SCC fuse).

PR #29505, opened 2023-08-21
Add initial fingerprint 2.0 for Nissan Leaf 2018-23

gellis12:

**Description** 
- [x] So far, the Leaf 2018-23 only supports the fingerprint 1.0 (can) method. This PR adds support for the fingerprint 2.0 (fw) method using the firmware versions present on my car.

**Verification**
- [x] I started the car and confirmed openpilot is able to load and correctly fingerprint the car, and further confirmed that `fingerprintSource` now reports as `fw` in UserAdmin.

**Route**
Route: [289d318fded3221a|2023-08-21--01-05-27]

### DIFF
--- a/selfdrive/car/nissan/values.py
+++ b/selfdrive/car/nissan/values.py
@@ -132,6 +132,20 @@ FW_VERSIONS = {
       b'284U29HE0A',
     ],
   },
+    CAR.LEAF: {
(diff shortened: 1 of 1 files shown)


sshane:
thanks! we can start to combine the platforms soon and use the FW to detect the different seatbelt message, I already see a pattern between these two platforms

PR #29506, opened 2023-08-21
boardd: keepTime after failing to get `ignition_opt`

deanlee:
Think it's better to call `keepTime` after failing to get ignition_opt  than retrying immediately in loop.


selfdrive/boardd/boardd.cc
```diff
@@ -480,6 +480,8 @@ void panda_state_thread(std::vector<Panda *> pandas, bool spoofing_started) {
     auto ignition_opt = send_panda_states(&pm, pandas, spoofing_started);
 
     if (!ignition_opt) {
+      LOGW("Failed to get ignition_opt");
```
adeebshihadeh, line 483:
```suggestion
      LOGE("Failed to get ignition_opt");
```

PR #29510, opened 2023-08-21
script to install only ubuntu dependencies

gregjhogan:
Projects that use openpilot as a pip package need to install the required apt packages but do not want to set up a virtual environment for openpilot (they have their own virtual environment).


adeebshihadeh:
Might as well merge mac/ubuntu_setup if we do this, but guess that can be a separate PR.

gregjhogan:
Yeah we should also enhance it to throw a nice error if you are on an unsupported version of an operating system.

PR #29513, opened 2023-08-21
test_models: pull only

sshane:


sshane:
Without decompressing and looping through the CAN messages twice for each segment, we can't know which rlog it will actually use at the moment, so we will end up spending time pulling more rlogs than we need, which will be slower. Besides that, I can't think of a clean way to do this non-hackily, so closing for now

PR #29514, opened 2023-08-21
CI: use xdist for car model tests

jnewb1:
Ideas:
~~more matrix jobs: we have had 5 jobs since we had 152 routes, now we have 215, meaning we should increase to 7 to match that growth.~~ not ideal
parallelization: these tests appear to independent, there isn't any param writing that might cause issues multithreading, so we should be able to use pytest-xdist to split across cores
speedup test itself:

about 50% of the test time is in the car interface, seems to mostly be the can parsing. we could possibly get some speedup by making that faster
image


adeebshihadeh:
5 is already a lot of runners, we shouldn't need more beyond that

sshane:
The caching used in FileReader isn't multiprocess/thread safe, that's why we do `--dist=loadscope` on the jenkins runner so each process only executes tests for one segment.

jnewb1:
don't think we currently use pytest-xdist on CI at all. Splitting across 2 cpus halved the time of the car tests. (ignoring setup time, because I had to add pytest-cov since coverage.py doesn't support xdist)

jnewb1:
Ignoring setup times

| | Master | This Branch |
| -- | ------------- | ------------- |
| Cars(0) | 11:47 | 5:14  |
| Cars(1) | 9:14  | 4:41  |
| Cars(2) | 9:00  | 6:31  |
| Cars(3) | 8:50  | 4:36  |
| Cars(4) | 8:58  | 5:30  |

jenkins uses xdist, so I think we should be safe to use it here

jnewb1:
pytest-cov also uses coverage.py under the hood, so I think the outputs should be identical

poetry.lock
```diff
@@ -387,9 +387,7 @@ version = "0.9.14"
 description = "Python API for communicating with the CARLA server."
 optional = false
 python-versions = "*"
-files = [
```
jnewb1, line 390: is there a standard way to generate the openpilot lockfile? do we include --dev?

adeebshihadeh: `poetry lock --no-update` for stuff like this

adeebshihadeh (approved):
Just make sure the coverage report is still valid. Also give this/the merged commit a better title

poetry.lock
jnewb1: ok, thats what I did. not sure how carla was added, probably someone did --dev

PR #29516, opened 2023-08-22
manager: fix body callbacks

### LINKED ISSUE #29469: micd shouldn't run offroad
https://github.com/commaai/openpilot/pull/29087

sshane:


adeebshihadeh:
This should be refactored into something more sane.

sshane:
Refactored the started logic to use two callbacks for onroad and offroad conditions, rather than two bools and ORing it with a callback that needed to be aware of the onroad status. Now the callbacks are agnostic to that, and you have to be more intentional by modifying either onroad or offroad callbacks, or both.

selfdrive/manager/process.py
```diff
@@ -63,12 +63,19 @@ def join_process(process: Process, timeout: float) -> None:
...
+
+
 class ManagerProcess(ABC):
   daemon = False
   sigkill = False
-  onroad = True
-  offroad = False
-  callback: Optional[Callable[[bool, Params, car.CarParams], bool]] = None
+  onroad_callback: Callable[[Params, car.CarParams], bool] = enabled_callback
+  offroad_callback: Callable[[Params, car.CarParams], bool] = disabled_callback
```
adeebshihadeh, line 78: do we really need two different callbacks?

sshane: If we have one callback, then you need to always make sure to implement `started` which I thought we wanted to refactor out

sshane: I haven't found a cleaner way, since we use the callback as an OR to allow processes that normally don't run offroad to run (like driverview), but in the case of the body, we wanted it to be an AND while onroad.

If there's an onroad and offroad callback, we can clearly see the logic in the process_config.py file. Let me know if you have any other ideas.

adeebshihadeh:
Made a simpler version of this that's net -10 line diff: #29574

PR #29517, opened 2023-08-22
CI: simulator disable pull to save local space

jnewb1:
After adding the scons-cache to the simulator, are starting to reach the local storage limit of github actions (only 14gb is guarenteed) so the sim building is failing occasionally: https://github.com/commaai/openpilot/actions/runs/5932532426/job/16087435248

Pulling the sim image would only rarely give us any cache (only if we don't change anything in openpilot, selfdrive, panda, and others). So there is not much point in wasting the extra ~2gb of space and time that it requires to pull the full simulator image

### DIFF
--- a/.github/workflows/tools_tests.yaml
+++ b/.github/workflows/tools_tests.yaml
@@ -64,7 +64,6 @@ jobs:
       run: eval "$BUILD_CL"
     - name: Build simulator image
       run: |
-        docker pull $DOCKER_REGISTRY/$IMAGE_NAME:latest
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
I think docker supports remote cache: https://martinheinz.dev/blog/61

jnewb1:
> I think docker supports remote cache: https://martinheinz.dev/blog/61

Actually, looks like buildkit will still pull the image automatically if it can use it in it's build cache, so this pull isn't required anyway.

you can see here that it read the cache manifest from the remote repo:
https://github.com/commaai/openpilot/actions/runs/5933332923/job/16088611840?pr=29517#step:5:43

adeebshihadeh:
We should fix this for the rest of the CI docker builds too

jnewb1:
> We should fix this for the rest of the CI docker builds too

I'll merge this to fix the CI and try to standardize the docker building in another PR

PR #29518, opened 2023-08-22
Ui/map: Velocity filter fix

jakethesnake420:
When the vehicle comes to a stop, the map continues to zoom in due to the velocity filter alpha being too high. This fixes the behavior by dynamically adjusting the alpha based on the current speed. This may need to be adjusted based on aesthetic preference. I divide the speed by 3 here velocity_filter.updateAlpha(abs(locationd_velocity.getValue()[0])/3); to keep the alpha similar to the init value at at high speed.

Related to https://github.com/commaai/openpilot/issues/28816

### DIFF
--- a/common/util.h
+++ b/common/util.h
@@ -152,8 +152,12 @@ struct unique_fd {
 class FirstOrderFilter {
 public:
   FirstOrderFilter(float x0, float ts, float dt) {
-    k_ = (dt / ts) / (1.0 + dt / ts);
     x_ = x0;
+    dt_ = dt;
+    updateAlpha(ts);
+  }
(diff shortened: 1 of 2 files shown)


sshane:
You already had https://github.com/commaai/openpilot/pull/29488, no need to re-create the PR

jakethesnake420:
Oh I found a bug I didn't know how to update the PR but I realize now it does it automatically. Sorry for the mistake I don't use github.

sshane:
I'm not sure if this is a strict bugfix, it changes the zoom speed. Going to make a simpler change here: https://github.com/commaai/openpilot/pull/29601

PR #29520, opened 2023-08-22
Kia: add test route for Optima Hybrid

sshane:


sshane:
Keeping as dash cam only, see comment in: https://github.com/commaai/panda/pull/1609#issuecomment-1694132411

PR #29527, opened 2023-08-22
CI: unittest reporting with pytest

jnewb1:
so we can see the time for each test

depends: https://github.com/commaai/openpilot/pull/29538


.github/workflows/selfdrive_tests.yaml
```diff
@@ -284,10 +270,16 @@ jobs:
...
                         ./tools/cabana/tests/test_cabana && \
                         ./system/camerad/test/ae_gray_test && \
-                        ./selfdrive/test/process_replay/test_fuzzy.py && \
-                        coverage xml"
+                        ./selfdrive/test/process_replay/test_fuzzy.py"
     - name: "Upload coverage to Codecov"
       uses: codecov/codecov-action@v3
+    - name: Test Report
+      uses: dorny/test-reporter@v1
+      if: success() || failure()    # run this step even if previous step failed
```
sshane, line 278:
```suggestion
      if: true
```
does this work?

jnewb1:
This test reporter isn't great, doesn't show time for each test. Will keep looking

jnewb1:
Perhaps pytest-html

https://github.com/pytest-dev/pytest-html

adeebshihadeh:
This should be easy to do on PC first. Everything that happens in CI should be easy to run or test locally.

PR #29533, opened 2023-08-22
Subaru: don't copy eyesight counters, calculate manually

jnewb1:
we currently copy the counters from eyesight, which can get slightly out of sync with the OP loop and causes skipped counters


jnewb1:
I'll test this out on the outback when it's back, should perhaps get a gen1 user to test it too

sshane:
Need to set counter?

jnewb1:
> CANParser has the counter start bit and size at -1 for Subaru. If it's common across all messages we can set them

It's common across all the signals we send, but canparser seems to require all messages (even ones we don't send) to have the same counter. So this doesn't look possible till we fix that.

sshane:
Qt just need to set counter in openpilot then

sshane:
@jnewb1 Can you re-open, set the counters, and merge?

jnewb1:
@sshane just tried this on the outback, works fine. Updated refs, anything else?
c3d1ccb52f5f9d65|2023-09-01--14-23-14

selfdrive/car/subaru/carcontroller.py
```diff
@@ -80,29 +80,30 @@ def update(self, CC, CS, now_nanos):
...
 
-          can_sends.append(subarucan.create_es_distance(self.packer, CS.es_distance_msg, 0, pcm_cancel_cmd,
+          can_sends.append(subarucan.create_es_distance(self.packer, self.frame // 5, CS.es_distance_msg, 0, pcm_cancel_cmd,
                                                         self.CP.openpilotLongitudinalControl, cruise_brake > 0, cruise_throttle))
       else:
         if pcm_cancel_cmd:
           if self.CP.carFingerprint not in HYBRID_CARS:
             bus = CanBus.alt if self.CP.carFingerprint in GLOBAL_GEN2 else CanBus.main
-            can_sends.append(subarucan.create_es_distance(self.packer, CS.es_distance_msg, bus, pcm_cancel_cmd))
+            can_sends.append(subarucan.create_es_distance(self.packer, self.frame // 10, CS.es_distance_msg, bus, pcm_cancel_cmd))
```
sshane, line 106: this isn't behind a modulo, so is this correct?
```suggestion
            can_sends.append(subarucan.create_es_distance(self.packer, self.frame, CS.es_distance_msg, bus, pcm_cancel_cmd))
```

jnewb1: For ES_Distance, we send counter+1 to do the cruise_cancel, so that param isn't used, I made it "None" instead, which is more clear that it isn't used

selfdrive/car/subaru/carcontroller.py
```diff
@@ -80,29 +80,30 @@ def update(self, CC, CS, now_nanos):
...
 
-          can_sends.append(subarucan.create_es_distance(self.packer, CS.es_distance_msg, 0, pcm_cancel_cmd,
+          can_sends.append(subarucan.create_es_distance(self.packer, self.frame // 5, CS.es_distance_msg, 0, pcm_cancel_cmd,
                                                         self.CP.openpilotLongitudinalControl, cruise_brake > 0, cruise_throttle))
       else:
         if pcm_cancel_cmd:
           if self.CP.carFingerprint not in HYBRID_CARS:
             bus = CanBus.alt if self.CP.carFingerprint in GLOBAL_GEN2 else CanBus.main
-            can_sends.append(subarucan.create_es_distance(self.packer, CS.es_distance_msg, bus, pcm_cancel_cmd))
+            can_sends.append(subarucan.create_es_distance(self.packer, None, CS.es_distance_msg, bus, pcm_cancel_cmd))
```
sshane, line 106:
```suggestion
            can_sends.append(subarucan.create_es_distance(self.packer, CS.es_distance_msg["COUNTER"] + 1, CS.es_distance_msg, bus, pcm_cancel_cmd))
```

you can also do what hyundai does, pass in the actual counter.

jnewb1: done (pass in counter + 1 to frame)

sshane (approved):

Issue #26771, opened 2022-12-12
2020 Silverado Wanders straight, hugs curves


solardude2:
### Describe the bug

Wanders lane straight, hugs lines straight, hugs inside curves and crosses lines in curves. Has been doing this since being supported.  View of road and 360 camera showing lane placement can be seen here.  Unlisted video (https://www.youtube.com/watch?v=xU4Cuonmq_s)

### Which car does this affect?

2020 Silverado

### Provide a route where the issue occurs

38aa7da107d5d252|2022-12-11--11-05-23

### openpilot version

 openpilot v0.9.1

morrislee:
Same here from Sierra, the lane cutting on curves seems to be happening other cars too, there are a few reports even on small cars.

sshane:
@nuwandavek hopefully your Bolt tune should extend to all GM

Verylukyguy:
Has there been any progress on the Bolt/GM tuning?

morrislee:
I believe they are referring to https://github.com/commaai/openpilot/pull/26371

sshane:
@solardude2 does this still happen on latest master?

solardude2:
Last time I used it. Nobody with this type truck is using master. All on forks. I went back to my old tried n true and the other guys are on a fork of master trying to get it to work better. I will happily switch back for testing. Cuts curves and doesn't stay straight.

sshane:
@nuwandavek does the non-linear function we have work for this car?

morrislee:
NNFF from twilsonco is very much of an improvement for the oscillation in straights and over cutting in turns. But nothing so far fixes hugging in laneless profile. Lane planner still exist in sunnypilot has slightly better reliability in centering. I dont know if its a planner cost issue needs to be tuned or model is natually hugging when its still a taller vehicle

morrislee:
**Lane Planner**
![image](https://github.com/commaai/openpilot/assets/607941/031e0479-aa9f-423c-a725-161e41e294bd)

**Laneless Planner**
![image](https://github.com/commaai/openpilot/assets/607941/390ecd56-647e-4ac8-8a31-0306e9a6ce04)

Same route with different planner...

Doesnt look like a control issue, they look identical, this is with tim's NNFF, lane planner from sunny's fork to compare.

Lane positioning is very close to the Laneless from master that tim has added NNFF in, master just more oscillation and poor turn performances

nuwandavek:
@morrislee Can you post these route ids? 
[...]

PR #29540, opened 2023-08-22
CHEVROLET SILVERADO 1500 2020: Make Silverado lateral tune non-linear

nuwandavek:


sshane:
Fixes https://github.com/commaai/openpilot/issues/26771

PR #29094, opened 2023-07-22
Toyota TSS2: radar disable support

sshane:
No clue how I wasn't able to do this before, super easy. The ECUs automatically disable tx when you set them in programming diagnostic session type, but better to send disable tx (disable rx never worked).

POC to see if RAV4 2022-23 will work, worked on my TSS 2.5 Camry Hybrid.

panda PR: https://github.com/commaai/panda/pull/1611

PR #29541, opened 2023-08-22
Toyota: prepare to disable radar

sshane:
split from https://github.com/commaai/openpilot/pull/29094 to get something merged, this PR just adds the code and relevant comments about toyota radar disable.


selfdrive/car/toyota/interface.py
```diff
@@ -261,6 +261,15 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
 
     return ret
 
+  # uncomment to disable radar:
+  # @staticmethod
```
adeebshihadeh, line 265: this is better than leaving around in a PR, but I'd merge it as actual code behind an `if False` or a flag so it survives refactors

PR #29544, opened 2023-08-23
Toyota: handle `PCS_HUD`

cydia2020:
needs #29094

### DIFF
--- a/selfdrive/car/toyota/carcontroller.py
+++ b/selfdrive/car/toyota/carcontroller.py
@@ -161,13 +161,15 @@ class CarController:
         # forcing the pcm to disengage causes a bad fault sound so play a good sound instead
         send_ui = True
 
+      # *** lane keep assist ui ***
       if self.frame % 20 == 0 or send_ui:
         can_sends.append(create_ui_command(self.packer, steer_alert, pcm_cancel_cmd, hud_control.leftLaneVisible,
                                            hud_control.rightLaneVisible, hud_control.leftLaneDepart,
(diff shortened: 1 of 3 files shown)


selfdrive/car/toyota/carcontroller.py
```diff
@@ -161,13 +161,18 @@ def update(self, CC, CS, now_nanos):
...
                                            hud_control.rightLaneDepart, CC.enabled, CS.lkas_hud))
 
-      if (self.frame % 100 == 0 or send_ui) and (self.CP.enableDsu or self.CP.flags & ToyotaFlags.DISABLE_RADAR.value):
-        can_sends.append(create_fcw_command(self.packer, fcw_alert))
+      # *** pre-collision ui ***
+      # do not send on TSS-P cars with DSU
+      # uncomment "or self.CP.radarUnavailable)" when ready to disable
+      if (self.frame % 100 == 0 or send_ui) and (self.CP.enableDsu or self.CP.flags & ToyotaFlags.DISABLE_RADAR.value \
+                                                 or self.CP.carFingerprint in TSS2_CAR):
+        can_sends.append(create_fcw_command(self.packer, fcw_alert, CS.pcs_hud))
```
sshane, line 175: Hmm, the reference DBC states that the PCS_HUD is sent by the DSU and received by the FCM (camera). This is likely out of date for TSS2 cars, and the camera probably doesn't listen to its own messages. But since we don't have time to verify for now, let's close. Thanks for the PR! I'll re-open when we can verify.

PR #29465, opened 2023-08-18
ui/map_instructions: center lanes

deanlee:
| Before  | After |
| ------------- | ------------- |
| Screenshot 2023-08-18 22:17:02  | Screenshot 2023-08-18 22:17:58  |
| Screenshot 2023-08-18 22:15:57  | Screenshot 2023-08-18 22:14:20  |
| Screenshot 2023-08-18 22:16:32  | Screenshot 2023-08-18 22:20:05  |

Center lanes like google map & mapbox

google map:
Screenshot from 2023-08-18 22-30-44

mapbox (https://docs.mapbox.com/android/navigation/guides/ui-components/maneuver/):
Screenshot from 2023-08-18 22-29-47

PR #29548, opened 2023-08-23
ui/map_instructions: restore padding

deanlee:
restore padding before  #29465

Restore the 6 pixels of padding missing due to the removal of the widget container.

| Version  | padding |
| Before #29456 | 261631919-01d8a250-e990-4f7f-a164-27362e697a6b |
| #29456  | Screenshot 2023-08-23 14:52:38  |
| Now  | Screenshot 2023-08-23 15:12:19  |


sshane:
Still looks like there's less space from the arrows to the bottom.

sshane:
There seems to be an extra pixel of space between distance and primary only when secondary is hidden, but matches exactly besides that

PR #29554, opened 2023-08-23
cpplint: add `readability/braces` filter

deanlee:
this filter adds the following checks:

1. You don't need a ; after a } 
2. If an else has a brace on one side, it should have it on both 
3. If/else bodies with multiple statements require braces
4. Else clause should be indented at the same level as if
5. Did you mean "else if"? If not, start a new line for "if".


adeebshihadeh:
Nice!

PR #29555, opened 2023-08-23
Ford: change LatCtlPrecision_D_Rq to comfortable

### LINKED ISSUE #1121: Ford: steering oscillation
**Ford Focus**
Steering oscillations while driving straight on the highway:
- `e886087f430e7fe7|2023-07-16--21-08-19--11`
image
- `e886087f430e7fe7|2023-07-17--01-00-16--{22,23}`

**Ford Escape**
Oscillations during some turns:
- `2631ea836ece842e|2023-07-11--16-39-52--16` (and at other parts of this route)
- `2631ea836ece842e|2023-07-17--16-17-15--{11,12,13}`
- `2631ea836ece842e|2023-07-17--16-17-15--{18,19}`

incognitojam:
- Lane changes feel smoother (almost hard to tell whether you have triggered the lane change)
- Reduced steering wheel oscillation (still occurs after bumps in the road, but recovers)

I did not notice any large oscillations after driving for 6 hours on the motorway. Previously, I saw it more than once in 1hr at high speeds. Resolves https://github.com/commaai/opendbc/issues/1121

### DIFF
--- a/selfdrive/car/ford/fordcan.py
+++ b/selfdrive/car/ford/fordcan.py
@@ -76,8 +76,8 @@ def create_lat_ctl_msg(packer, CAN: CanBus, lat_active: bool, path_offset: float
                                                 #       3=InterventionRight, 4-7=NotUsed [0|7]
     "LatCtlRampType_D_Rq": 0,                   # Ramp speed: 0=Slow, 1=Medium, 2=Fast, 3=Immediate [0|3]
                                                 #             Makes no difference with curvature control
-    "LatCtlPrecision_D_Rq": 1,                  # Precision: 0=Comfortable, 1=Precise, 2/3=NotUsed [0|3]
-                                                #            The stock system always uses comfortable
+    "LatCtlPrecision_D_Rq": 0,                  # Precision: 0=Comfortable, 1=Precise, 2/3=NotUsed [0|3]
+                                                # We have not observed the stock system using precise
     "LatCtlPathOffst_L_Actl": path_offset,      # Path offset [-5.12|5.11] meter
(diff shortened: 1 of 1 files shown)


incognitojam:
I'm planning to test this once more on the same stretch of road where I reproduced the steering oscillations while driving straight

haraschax:
Does desired_lat_accel reflect what what was sent to the car? Or is there more logic after that?

incognitojam:
The only processing after that would be rate limits. `actuatorsOutput.curvature` is the final output

incognitojam:
`e886087f430e7fe7|2023-07-16--21-08-19--11` (before changes)
image

Maybe we can close the loop on curvature

adeebshihadeh:
How does the stock system use this signal?

incognitojam:
I have never seen it set to precise on the stock system.

haraschax:
Weird, it seems to just be broken on their end then.

incognitojam:
I don't believe the stock system uses "precise" at all from the logs I've seen (there's also a comment I wrote in `fordcan.py` saying that)

sshane:
any updates?

adeebshihadeh:
If the stock system never uses the "precise" value, this seems like a reasonable change, pending further investigation. Agree @haraschax?

haraschax:
Ok sounds good, but please add a comment mentioning the choice.

sshane:
I sent a route to @incognitojam from an old test with this signal at comfortable. I don't know if he checked it out yet but I recall this signal didn't change the absolute max jerk limits up or down. Neither did `LatCtlRampType_D_Rq`, but it *did* change how quickly it dropped torque after disabling with it wound it.

sshane:
@incognitojam deleting, lmk if you need the branch

PR #29556, opened 2023-08-23
cpplint:   `build/include_what_you_use`

deanlee:


adeebshihadeh:
Hmm, this adds a lot of unnecessary lines. What do you think about this one? I'm 50/50 on it.

deanlee:
I didn't think it made much sense at first. But halfway through, I think it's worth it. We don't have to guess where the relevant header files are, "Do not rely on transitive inclusions" is a good style (https://google.github.io/styleguide/cppguide.html#Include_What_You_Use)

I've also added some comments to make include a little clearer:
```
#include <algorithm>  // for sort
#include <utility>  // for pair<>
```

adeebshihadeh:
Sounds good, but let's skip the comments. There's nothing enforcing the comments stay up to date.

deanlee:
In addition, after adding this filter, we can safely delete many unused includes

PR #29557, opened 2023-08-23
v4l_encoder: improve `dequeue_buffer`

deanlee:
Improve `dequeue_buffer`'s error handling by referring to the v4l-related code in webtrc and chromium.

### DIFF
--- a/system/loggerd/encoder/v4l_encoder.cc
+++ b/system/loggerd/encoder/v4l_encoder.cc
@@ -27,21 +27,39 @@ static void checked_ioctl(int fd, unsigned long request, void *argp) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
What exactly does this fix or improve?

deanlee:
it can handle the EAGAIN error, and log the error message if V4L2_BUF_FLAG_ERROR is included in flags.

I read the source code related to v4l encoding/decoding in webcrt and chromium, they both handle the eagin error. I'm not sure if c3 returns this error. But if it appears and we don't handle it, the encoder will crash.

The changes in `dequeue_buffer` are basically copied from chromium. feel free to close if it doesn‘t make sense.

adeebshihadeh:
Gonna close for now since we haven't seen this in the field.

PR #29560, opened 2023-08-23
CI: use pytest for unittests

jnewb1:
pytest allows for creating reports, which will be useful in analyzing test times


jnewb1:
Tests aren't independent, the logger test has some output from the maneuver tests in it

PR #29562, opened 2023-08-23
cpplint: add filter `whitespace/operators`

deanlee:


adeebshihadeh:
This one's fairly aggressive, will have to think about it. Generally the # of errors is a good indicator of whether we want to enable a rule: low violation count and they all look like reasonable failures means we want to enable. "include what you use" is the exception to that rule.

deanlee:
y,closed.

PR #29571, opened 2023-08-23
cpplint: add filter `whitespace/ending_newline`

deanlee:


adeebshihadeh:
Don't think this one is necessary, we don't enforce for Python either

PR #29574, opened 2023-08-24
manager: refactor process starting logic

### LINKED ISSUE #29469: micd shouldn't run offroad

adeebshihadeh:


selfdrive/manager/process_config.py
```diff
@@ -8,7 +8,7 @@
 WEBCAM = os.getenv("USE_WEBCAM") is not None
 
 def driverview(started: bool, params: Params, CP: car.CarParams) -> bool:
-  return params.get_bool("IsDriverViewEnabled")  # type: ignore
+  return started or params.get_bool("IsDriverViewEnabled")
 
 def notcar(started: bool, params: Params, CP: car.CarParams) -> bool:
   return CP.notCar  # type: ignore
```
nuwandavek:
```suggestion
  return started and CP.notCar  # type: ignore
```

adeebshihadeh: outdated before you could even comment :)

PR #29579, opened 2023-08-24
Rewrite navmodeld in python

mitchellgoffpc:


selfdrive/modeld/navmodeld.py
```diff
@@ -0,0 +1,126 @@
...
+      'y': [p.y for p in model_result.plan.mean],
+      'xStd': [math.exp(p.x) for p in model_result.plan.std],
+      'yStd': [math.exp(p.y) for p in model_result.plan.std]}}
+
+  return msg
+
+
+def main():
+  gc.disable()
+  set_realtime_priority(53)
```
mitchellgoffpc, line 91: In navmodeld.cc we used setpriority(-15), but in python this seems to cause a permissions error on device and also doesn't appear to be used anywhere else in the codebase; all the other processes seem to prefer set_realtime_priority. so I've switched to using it here as well. It's not entirely clear what priority is appropriate here, I set it to 53 (same as camerad, one level below modeld) but maybe a lower priority would be better.

selfdrive/modeld/SConscript
```diff
@@ -44,6 +44,10 @@ for pathdef, fn in {'TRANSFORM': 'transforms/transform.cl', 'LOADYUV': 'transfor
     xenv['CXXFLAGS'].append(f'-D{pathdef}_PATH=\\"{File(fn).abspath}\\"')
 
 # Compile cython
+snpe_rpath_qcom = "/data/pythonpath/third_party/snpe/larch64"
+snpe_rpath_pc = f"{Dir('#').abspath}/third_party/snpe/x86_64-linux-clang"
+snpe_rpath = lenvCython['RPATH'] + [snpe_rpath_qcom if arch == "larch64" else snpe_rpath_pc]
```
mitchellgoffpc, line 49: We used to set LD_LIBRARY_PATH in the navmodeld launch script, but that's been removed since navmodeld is now a PythonProcess forked from manager.py. snperunner still needs to be linked to libSNPE.so, and after some research it _seems_ like the most straightforward way to do this is to set the RPATH in scons. However, I'm not super familiar with these linker flags and it's possible there are some downsides I'm not aware of. I also removed `/usr/lib/aarch64-linux-gnu` and `"#/openpilot/third_party/snpe/x86_64-linux-clang` for TICI/PC, respectively, since it doesn't seem like libSNPE ever exists in the former as far as I can tell, and the later path seems to not exist at all.

selfdrive/modeld/models/driving.h
```diff
@@ -6,13 +6,16 @@
...
-#include "selfdrive/modeld/models/nav.h"
+#include "selfdrive/modeld/models/commonmodel.h"
+#include "selfdrive/modeld/runners/run.h"
 
 constexpr int FEATURE_LEN = 128;
 constexpr int HISTORY_BUFFER_LEN = 99;
 constexpr int DESIRE_LEN = 8;
 constexpr int DESIRE_PRED_LEN = 4;
 constexpr int TRAFFIC_CONVENTION_LEN = 2;
+constexpr int NAV_FEATURE_LEN = 256;
```
mitchellgoffpc, line 17: It's not ideal that NAV_FEATURE_LEN is specified in both driving.h and nav.h, but this is sort of an artifact of the fact that the output size of navmodel.onnx and the size of the `nav_features` input on supercombo.onnx aren't guaranteed to be the same. These constants should probably be checked against each other in the modeld SConscript, but that seemed outside the scope of this PR to me.

selfdrive/modeld/navmodeld.py
```diff
@@ -0,0 +1,125 @@
...
+from openpilot.system.hardware import TICI
+from openpilot.system.swaglog import cloudlog
+from openpilot.common.params import Params
+from openpilot.common.realtime import set_realtime_priority
+from openpilot.selfdrive.modeld.constants import IDX_N
+from openpilot.selfdrive.modeld.models.commonmodel_pyx import Runtime
+
+USE_SNPE = int(os.getenv('USE_SNPE', str(int(TICI))))
+if USE_SNPE:
+  os.environ['ADSP_LIBRARY_PATH'] = "/data/pythonpath/third_party/snpe/dsp/"
```
mitchellgoffpc, line 22: Again, this used to be specified in the navmodeld launch script but now needs to be specified elsewhere. This seemed like the most appropriate place to me.

selfdrive/modeld/navmodeld.py
```diff
@@ -0,0 +1,125 @@
...
+  from selfdrive.modeld.runners.snpemodel_pyx import SNPEModel as ModelRunner
+else:
+  from selfdrive.modeld.runners.onnxmodel_pyx import ONNXModel as ModelRunner
+
+NAV_INPUT_SIZE = 256*256
+NAV_FEATURE_LEN = 256
+NAV_DESIRE_LEN = 32
+NAV_OUTPUT_SIZE = 2*2*IDX_N + NAV_DESIRE_LEN + NAV_FEATURE_LEN
```
mitchellgoffpc, line 30: Again, these constants (along with the ones in modeld and dmonitoringmodeld) should ideally be inferred by checking the input/output sizes of the onnx files, but that seemed outside the scope of this PR.

selfdrive/modeld/navmodeld.py
adeebshihadeh: let's set to 1

selfdrive/modeld/SConscript
adeebshihadeh: nope, RPATH is good (even preferable over the wrapper scripts). we didn't use to do it because android (bionic) didn't support it

selfdrive/modeld/navmodeld.py
adeebshihadeh: lgtm, though can we put the ADSP_LIBRARY_PATH setting in snpemodel_pyx and move this if statement into a helper since it'll be in all the model runners?

selfdrive/modeld/navmodeld.py
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env python3
+import gc
+import math
+import time
+import ctypes
+import numpy as np
+from pathlib import Path
+from typing import Tuple, Dict
+from cereal import messaging
```
adeebshihadeh, line 9: add a line between openpilot and the other imports

selfdrive/modeld/navmodeld.py
```diff
@@ -0,0 +1,118 @@
...
+NAV_OUTPUT_SIZE = 2*2*IDX_N + NAV_DESIRE_LEN + NAV_FEATURE_LEN
+MODEL_PATHS = {
+  ModelRunner.SNPE: Path(__file__).parent / 'models/navmodel_q.dlc',
+  ModelRunner.ONNX: Path(__file__).parent / 'models/navmodel.onnx'}
+
+class NavModelOutputXY(ctypes.Structure):
+  _fields_ = [
+    ("x", ctypes.c_float),
+    ("y", ctypes.c_float)]
+class NavModelOutputPlan(ctypes.Structure):
```
adeebshihadeh, line 30: break these up with a new line

selfdrive/modeld/navmodeld.py
```diff
@@ -0,0 +1,118 @@
...
+    'locationMonoTime': location_ts,
+    'modelExecutionTime': execution_time,
+    'dspExecutionTime': dsp_execution_time,
+    'features': model_result.features[:],
+    'desirePrediction': model_result.desire_pred[:],
+    'position': {
+      'x': [p.x for p in model_result.plan.mean],
+      'y': [p.y for p in model_result.plan.mean],
+      'xStd': [math.exp(p.x) for p in model_result.plan.std],
+      'yStd': [math.exp(p.y) for p in model_result.plan.std]}}
```
adeebshihadeh, line 76: can you put the closing bracket on its own line?

selfdrive/modeld/navmodeld.py
```diff
@@ -0,0 +1,118 @@
...
+    self.model.execute()
+    t2 = time.perf_counter()
+    return self.output, t2 - t1
+
+def get_navmodel_packet(model_output: np.ndarray, valid: bool, frame_id: int, location_ts: int, execution_time: float, dsp_execution_time: float):
+  model_result = model_output.ctypes.data_as(ctypes.POINTER(NavModelResult)).contents
+  msg = messaging.new_message('navModel')
+  msg.valid = valid
+  msg.navModel = {
+    'frameId': frame_id,
```
adeebshihadeh, line 66: `msg.navModel.frameId` is generally preferred. you don't get any static analysis with the dict

PR #29586, opened 2023-08-24
manager: test process exclusion

sshane:
split from https://github.com/commaai/openpilot/pull/29574


selfdrive/manager/test/test_manager.py
```diff
@@ -32,6 +33,10 @@ def test_manager_prepare(self):
     os.environ['PREPAREONLY'] = '1'
     manager.main()
 
+  def test_blacklisted_procs(self):
+    # TODO: ensure blacklisted procs until we have a dedicated test
```
sshane, line 37:
```suggestion
    # TODO: ensure there are blacklisted procs until we have a dedicated test
```

PR #29587, opened 2023-08-24
CI: specify buildx version

jnewb1:
seems like the ci caching issue might have to do with what version of runner we are running. the runner release 8/21 has a new version of docker which seems to break the cache, or vice versa.

8/14: https://github.com/actions/runner-images/releases/tag/ubuntu20%2F20230814.1
8/21: https://github.com/actions/runner-images/releases/tag/ubuntu20%2F20230821.1

For example, these are all cache misses, running 8/21
https://github.com/commaai/openpilot/actions/runs/5960825552/job/16168840072#step:1:9
https://github.com/commaai/openpilot/actions/runs/5960825552/job/16168840528#step:1:9
https://github.com/commaai/openpilot/actions/runs/5960825552/job/16168840690#step:1:9
https://github.com/commaai/openpilot/actions/runs/5960825552/job/16168841061#step:1:9

and these are all cache hits, running 8/14
https://github.com/commaai/openpilot/actions/runs/5960825552/job/16168840223#step:1:9
https://github.com/commaai/openpilot/actions/runs/5960825552/job/16168840390#step:1:9
https://github.com/commaai/openpilot/actions/runs/5960825552/job/16168840875#step:1:9
https://github.com/commaai/openpilot/actions/runs/5960825552/job/16168839367#step:1:9
https://github.com/commaai/openpilot/actions/runs/5960825552/job/16168837934#step:1:9

It will probably flip (8/14 becomes miss) if docker_push happened to run on 8/21. I think we should just avoid this problem by using buildx and pinning the version we use

doesn't seem like we can pin to a specific runner version either, its just whatever the host happens to have

### DIFF
--- a/.github/workflows/prebuilt.yaml
+++ b/.github/workflows/prebuilt.yaml
@@ -10,7 +10,7 @@ env:
 
   DOCKER_LOGIN: docker login ghcr.io -u ${{ github.actor }} -p ${{ secrets.GITHUB_TOKEN }}
   BUILD: |
-      DOCKER_BUILDKIT=1 docker build --pull --cache-to type=inline --cache-from $DOCKER_REGISTRY/$BASE_IMAGE:latest -t $DOCKER_REGISTRY/$BASE_IMAGE:latest -t $BASE_IMAGE:latest -f Dockerfile.openpilot_base .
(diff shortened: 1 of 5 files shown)


jnewb1:
> It will probably flip (8/14 becomes miss) if docker_push happened to run on 8/21

it did flip in this run: https://github.com/commaai/openpilot/actions/runs/5960909210/job/16169107062?pr=29587

jnewb1:
Pinning buildx doesn't do anything, seems we need to pin the underlying docker version

PR #29594, opened 2023-08-24
common/i2c: remove duplicate #include

deanlee:


deanlee:
cpplint can't check for issues like this. I found it with eyes

PR #29596, opened 2023-08-24
cpplint: add filter `build/include_alpha`

deanlee:
I use the following style to format(resort) the includes (vscode):

The reduced number of lines is due to the c++ formatter removing redundant blank lines.


adeebshihadeh:
We don't enforce this for Python etiher. One of the goals for the linter is to be able to work without it in the loop and it passes most of the time.

PR #29597, opened 2023-08-24
ui: add translation files to assets.qrc

deanlee:
split from https://github.com/commaai/openpilot/pull/25680 .

1. move translation files to assets.qrc.
2. build translation files before build assets.


selfdrive/assets/assets.qrc
```diff
@@ -14,5 +14,16 @@
...
+  <file alias="languages.json">../ui/translations/languages.json</file>
+  <file alias="main_de">../ui/translations/main_de.qm</file>
+  <file alias="main_fr">../ui/translations/main_fr.qm</file>
+  <file alias="main_pt-BR">../ui/translations/main_pt-BR.qm</file>
+  <file alias="main_tr">../ui/translations/main_tr.qm</file>
+  <file alias="main_th">../ui/translations/main_th.qm</file>
+  <file alias="main_zh-CHS">../ui/translations/main_zh-CHS.qm</file>
+  <file alias="main_zh-CHT">../ui/translations/main_zh-CHT.qm</file>
+  <file alias="main_ko">../ui/translations/main_ko.qm</file>
+  <file alias="main_ja">../ui/translations/main_ja.qm</file>
```
sshane, line 27: any way to keep the list in one spot?

deanlee: There should be a way, but that might complicate things.
adding new language is not a frequent operation. probably not a few times a year. I recommend keeping it as it is. we can Improve it later if necessary.what do you think?

sshane:
can you rebase?

sshane:
You forgot `main_en`, it's already mismatched 😢

deanlee:
ah!😢

selfdrive/ui/SConscript
```diff
@@ -37,11 +37,30 @@ widgets = qt_env.Library("qt_widgets", widgets_src, LIBS=base_libs)
...
 qt_libs = [widgets, qt_util] + base_libs
 
+qt_src = ["main.cc", "qt/sidebar.cc", "qt/onroad.cc", "qt/body.cc",
+          "qt/window.cc", "qt/home.cc", "qt/offroad/settings.cc",
+          "qt/offroad/software_settings.cc", "qt/offroad/onboarding.cc",
+          "qt/offroad/driverview.cc", "qt/offroad/experimental_mode.cc"]
+
+# build translation files
+with open(File("translations/languages.json").abspath) as f:
+  languages = json.loads(f.read())
```
sshane, line 47: Since we already read the json, can you make a temporary translations assets file and then build below?

selfdrive/assets/assets.qrc
sshane: done, not too complicated

deanlee:
the "Dynamic update" is so much better!

PR #29607, opened 2023-08-24
Added cython bindings for model runners and commonmodel

mitchellgoffpc:


selfdrive/modeld/SConscript
```diff
@@ -55,6 +56,27 @@ else:
 for pathdef, fn in {'TRANSFORM': 'transforms/transform.cl', 'LOADYUV': 'transforms/loadyuv.cl', 'ONNXRUNNER': 'runners/onnx_runner.py'}.items():
   path = File(fn).abspath
   lenv['CXXFLAGS'].append(f'-D{pathdef}_PATH=\\"{path}\\"')
+  lenvCython['CXXFLAGS'].append(f'-D{pathdef}_PATH=\\"{path}\\"')
+
+if (use_thneed and arch == "larch64") or GetOption('pc_thneed'):
+  lenvCython['CFLAGS'].append("-DUSE_THNEED")
+  lenvCython['CXXFLAGS'].append("-DUSE_THNEED")
```
adeebshihadeh, line 63: all the flags are duplicated everywhere. can you just update the cython ones after the main env config is done?

adeebshihadeh (approved):
LGTM, once the scons comment and the leftovers one from the other PR are addressed

selfdrive/modeld/SConscript
mitchellgoffpc: Fixed now for the path defines, unfortunately DUSE_THNEED can't easily be factored out because injecting into lenv breaks pc-thneed (have to inject into llenv instead). The pc-thneed section will be deleted in modeld-python, along with all the other instances of -DUSE_THNEED outside this block.

PR #29611, opened 2023-08-24
Linter: remove pylint

jnewb1:
remove old references to pylint


adeebshihadeh (approved):
What about the submodules?

jnewb1:
> What about the submodules?

will do those too

PR #29615, opened 2023-08-24
CI: increase plotjuggler timeout

jnewb1:
seems we are just barely missing the timeout

### DIFF
--- a/.github/workflows/tools_tests.yaml
+++ b/.github/workflows/tools_tests.yaml
@@ -36,7 +36,7 @@ jobs:
     - name: Build Docker image
       run: eval "$BUILD"
     - name: Unit test
-      timeout-minutes: 2
+      timeout-minutes: 3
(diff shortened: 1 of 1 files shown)


.github/workflows/tools_tests.yaml
```diff
@@ -36,7 +36,7 @@ jobs:
     - name: Build Docker image
       run: eval "$BUILD"
     - name: Unit test
-      timeout-minutes: 2
+      timeout-minutes: 3
       run: |
         ${{ env.RUN }} "scons -j$(nproc) --directory=/tmp/openpilot cereal && \
                         apt-get update && \
```
adeebshihadeh, line 42: can we just remove all this apt stuff instead? only missing package is `libdw-dev` - perhaps it should be in there

adeebshihadeh:
#29617

PR #29629, opened 2023-08-25
Tests: speedup match_fw_to_car

jnewb1:
build dict first instead of every loop


sshane:
How much did you see this speed up your test? Locally it takes 15s for `TestFwFingerprint` regardless with 2 jobs.

jnewb1:
> How much did you see this speed up your test? Locally it takes 15s for `TestFwFingerprint` regardless with 2 jobs.

basically nothing, but perhaps still worthwhile?

sshane:
I would speed up the lowest hanging fruit first.

PR #29631, opened 2023-08-25
cabana: improve frequency accuracy

deanlee:
Fix for inaccurate frequency.

### DIFF
--- a/tools/cabana/messageswidget.cc
+++ b/tools/cabana/messageswidget.cc
@@ -174,8 +174,7 @@ QVariant MessageListModel::data(const QModelIndex &index, int role) const {
 
   auto getFreq = [](const CanData &d) -> QString {
     if (d.freq > 0 && (can->currentSec() - d.ts - 1.0 / settings.fps) < (5.0 / d.freq)) {
-      auto nearby = std::nearbyint(d.freq);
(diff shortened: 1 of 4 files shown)


tools/cabana/messageswidget.cc
```diff
@@ -174,7 +174,8 @@ QVariant MessageListModel::data(const QModelIndex &index, int role) const {
 
   auto getFreq = [](const CanData &d) -> QString {
     if (d.freq > 0 && (can->currentSec() - d.ts - 1.0 / settings.fps) < (5.0 / d.freq)) {
-      return d.freq >= 1 ? QString::number(std::nearbyint(d.freq)) : QString::number(d.freq, 'f', 2);
+      auto nearby = std::nearbyint(d.freq);
+      return nearby >= 1 ? QString::number(nearby) : QString::number(d.freq, 'f', 2);
```
sshane, line 178:
```suggestion
      return d.freq >= 0.9 ? QString::number(std::nearbyint(d.freq)) : QString::number(d.freq, 'f', 2);
```

how about 0.9 instead of 0.5 as a threshold to round up to 1?

sshane:
as someone unfamiliar with the code, it would be useful if refactors/clean ups are split out. I think this looks good though, thanks!

deanlee:
got it.

PR #29636, opened 2023-08-25
CI: fix proc replay download cache permissions

jnewb1:
Appears to have been an issue for awhile, but the replay logs only take 10 second to download anyway.

(random commit from May, same issue) https://github.com/commaai/openpilot/actions/runs/5101950826/job/13812264100#step:24:3

### DIFF
--- a/.github/workflows/selfdrive_tests.yaml
+++ b/.github/workflows/selfdrive_tests.yaml
@@ -301,6 +301,7 @@ jobs:
       timeout-minutes: 30
       run: |
         ${{ env.RUN }} "CI=1 coverage run selfdrive/test/process_replay/test_processes.py -j$(nproc) && \
+                        chmod -R 777 /tmp/comma_download_cache && \
                         coverage xml"
     - name: Print diff
       id: print-diff


.github/workflows/selfdrive_tests.yaml
```diff
@@ -301,6 +301,7 @@ jobs:
       timeout-minutes: 30
       run: |
         ${{ env.RUN }} "CI=1 coverage run selfdrive/test/process_replay/test_processes.py -j$(nproc) && \
+                        chmod -R 777 /tmp/comma_download_cache && \
```
adeebshihadeh, line 304: can we do this at the end of our custom setup action?

jnewb1: that would require spinning up the docker image twice ($RUN has to be called twice), but probably isn't that big of a deal

adeebshihadeh: that should have almost no overhead. the goal is to make this bug harder to write in the future

jnewb1: it also needs to run at the end because the files are created by the root user while in docker

jnewb1: I can add it to the end of the docker_run script when I get around to making that

ElonVampire:
收到！

PR #29649, opened 2023-08-26
ui/network: fix inconsistent spacing with no status icon

sshane:
https://github.com/commaai/openpilot/pull/29037#issuecomment-1694241948

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -173,7 +173,6 @@ env = Environment(
     "-fPIC",
     "-O2",
     "-Wunused",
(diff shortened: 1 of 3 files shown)


deanlee:
didn't see this before submit #29650. should I close it?

sshane:
wasn't sure which which way was best, const is probably best since we don't need to fill a new QPixmap each time?

deanlee:
I think set fixed width is better than create and fill a transparent pixmap each time

PR #29650, opened 2023-08-26
ui/networking: fix icon label alignment issue

deanlee:
| Before  | After |
|  |  |


sshane:
thanks!

PR #29652, opened 2023-08-26
cabana: use a monotonic buffer to allocate CanEvent

deanlee:
Use a custom monotonic buffer to allocate CanEvent, and align all events by 16 bytes. this can improves overall performance, especially when running in live streaming mode.


sshane:
@deanlee cabana now crashes if there are no events (opened a route with just qlogs)

deanlee:
fixed #29795 . it's an old bug.

PR #29655, opened 2023-08-26
HKG: Add FW versions for EU Tucson Hybrid 2021

leZudiste:


incognitojam:
What is your dongle ID?

leZudiste:
c3231807b9f5b5f6

selfdrive/car/hyundai/values.py
```diff
@@ -206,7 +206,7 @@ def init_make(self, CP: car.CarParams):
     HyundaiCarInfo("Hyundai Tucson 2022", car_parts=CarParts.common([CarHarness.hyundai_n])),
     HyundaiCarInfo("Hyundai Tucson 2023", "All", car_parts=CarParts.common([CarHarness.hyundai_n])),
   ],
-  CAR.TUCSON_HYBRID_4TH_GEN: HyundaiCarInfo("Hyundai Tucson Hybrid 2022-23", "All", car_parts=CarParts.common([CarHarness.hyundai_n])),
+  CAR.TUCSON_HYBRID_4TH_GEN: HyundaiCarInfo("Hyundai Tucson Hybrid 2021-23", "All", car_parts=CarParts.common([CarHarness.hyundai_n])),
```
sshane, line 209: We usually target the US market with our docs, and it looks like it may be marketed different where you are. Here, 2022 was the first model year of the NX4 generation:

image

https://en.wikipedia.org/wiki/Hyundai_Tucson

PR #29657, opened 2023-08-26
ui: show driver view in popup dialog

deanlee:


deanlee:
Looks like a bug in CameraView

PR #29659, opened 2023-08-27
gasd

nworb-cire:
## TL;DR
This pull request creates a new daemon called `gasd` which improves the longitudinal controller on certain makes by directly learning the correct gas and brake signals to provide a specified acceleration. This learning is done in a similar way to `torqued`, and in fact much of the code is borrowed from there. The new controller scheme is entirely optional, and is implemented in a way that requires no code changes for makes and models that wish not to use it.

## Why this is important
As part of its operation on certain makes, openpilot must convert the `CC.actuators.accel` into gas and brake CAN signals sent to the car. I will hereafter refer to this as the "__controls map__". Currently, the controls map uses lookup tables and usually a univariate linear interpolation. These lookup tables are generally reused between several models, do not capture individual vehicle dynamics such as cargo weight or brake wear, and do not take into account externalities such as pitch. These discrepancies have historically been dealt with by the long PID controller and any issues are swept under the rug and often blamed on the PID tune, not on the controls map. It is also largely unclear where these lookup tables come from and how they can be improved.

The problem becomes immediately apparent when considering the relationship between the controls map zero point and long deadzones. If the `accel` command is zero, but the controls map sends a gas command that produces nonzero acceleration, the car will proceed to accelerate or (usually) decelerate instead of maintaining exact velocity. (One might call this a _controls mismatch_.) In the case of openpilot ACC long, this may not matter so much, as the growing speed difference will eventually cause the controller to desire sufficient acceleration to overcome the deadzone. However, with the end-to-end longitudinal controller, this is not guaranteed, and indeed the steady deceleration may well condition the model not to accelerate as it should. For reference, on master my car appears to require a `CC.actuators.accel` input of about 0.3 m/s/s in order to maintain speed at 40mph.

My subjective experience driving with the (offline) fit is that it produces noticeable improvements to experimental mode, especially in the braking department. Compared to these changes, master tends to brake late or insufficiently. My hope with this pull request is to bring such benefits to all vehicles.

### Past work
The only recent attempt I am aware of at improving the controls map has been by @twilsonco in https://github.com/commaai/openpilot/pull/25453. His work only updates the lookup tables to be velocity-dependent, but does not provide any sort of live parameter learning, and all results are derived from a single vehicle, so individual vehicle dynamics are ignored. It is also implemented in a way that makes it difficult to add additional input variables.

## Implementation details
`gasd` functions similarly to `torqued`. It collects three input variables (`CarState.aEgo`, `CarState.vEgo`, `LiveLocationKalman.calibratedOrientationNED.value[1]`) and two output variables (`CarControl.actuatorsOutput.gas`, `CarControl.actuatorsOutput.brake`). It buckets these and performs a corectilinear least squares fit (i.e. it ignores gas points where brake is nonzero and vice versa, and clips/rectifies each output appropriately). The outputs are coefficients that are used to compute the gas and brake signals.

In my analysis of my own car, I found that the gas and brake signals have a nearly affine relationship to the three input variables. Hence, using LLS here is a good fit. I assume the affine relationship holds for other models and makes as well.

In `controlsd`, after computing `CC.actuators.accel`, this is then fed into the long controller with `v_ego` and pitch, and the gas and brake signals are computed by taking dot products with the learned coefficients. These are then stored in the currently unused (deprecated?) `actuators.gas` and `actuators.brake` fields.

In the GM carcontroller, I check if both of these are not null, and if so, use them instead of the current lookup table controls map.

### Assumptions made
- The gas and brake signals have a roughly affine relationship with the three input variables, and the these input variables are sufficient to capture the dynamics of the car.
- The braking signal is always positive. This appears to be the case on GM, Honda and Subaru, which are the three makes that currently use controls maps.
- The OLS problem is sufficiently well-conditioned to perform a numerically stable fit with `np.linalg.lstsq`.

### Possible changes
Here are some possible changes that could be made to what is already here:
- Better bucketing would be boundlessly beneficial. The current setup does not always produce a good fit. Too few points and the fit is not ideal, too many buckets and it takes a long time to collect a sufficient amount of points (especially if the buckets are to include pitch information, it may be impossible for those who live in flat areas). If @nuwandavek could clarify how the bucket size was chosen for `torqued`, that would be helpful here. 
- The gas offset should be removed from CarParams or CarControllerParams so that it is not duplicated in two places
- Much of the code is duplicated from `torqued`. This could be refactored into a common library. (I did this a little bit, but wanted to touch `torqued` as little as possible.)
- We may consider using `velocityCalibrated` and `accelerationCalibrated` instead of `CarState.vEgo` and `CarState.aEgo`. It is unclear to me which would be preferred here.

## TODO:
- [ ] Identify better sizes and bounds of buckets as detailed above.
- [ ] Validate the linear relationship between the inputs and outputs for other makes and models
- [ ] Generate offline initial fits for other makes and models
- [ ] Refactor other carcontrollers on relevant makes:
  - [ ] to use new system
  - [ ] to set `new_actuators.gas` and `.brake`
- [ ] Possibly attempt some basic feature engineering. Analysis on my car shows the possibility that `log(vEgo)` may have a more linear relationship with gas and brake than `vEgo` itself, for instance


nuwandavek:
Hey @nworb-cire, thanks for the PR and the thoughtful notes. This is a good idea, but is not something we will be able to upstream currently. Here are a few reasons:
1. Things that affect multiple platforms (even if we start with only a few initially) across the board require a massive validation effort (especially the assumption wrt. linear relationships), which we cannot accommodate in our roadmap at the moment.
2. Using new inputs like pitch also fall under the above category (we spent multiple months making sure roll is as perfect as it is right now).
3. A lot of the controls issues exclusive to experimental mode are because the current models are pretty dumb. Something like `gasd` would be the wrong layer of abstraction to try and fix them. We're working on some very exciting ground-up model solutions to these problems, which will have marked improvements. We expect a lot of the planning/noise to be handled by the model better.
4. We started the lateral controls effort by having good metrics around openpilot's inadequacies. We're working on things to ship on this front in the near future. I expect longitudinal controls to follow a similar arc.

That said, I think forks are the best place for something like this. I can definitely answer any questions you may have about torqued (the bucket bounds were drawn from the distribution of data, and the size was determined from experimentation and expected convergence time). We may revisit this in the future, but I'm going to close this now.

PR #27478, opened 2023-03-02
HKG: Car Port for Hyundai Palisade and Kia Telluride 2023-24 (non-HDA2)

sunnyhaibin:
**Checklist**
- [x] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with stock system:
  - Hyundai Palisade 2023: `f2cf0de0119d2a16|2023-03-02--09-39-50`
  - Kia Telluride 2023:
- [x] route with openpilot:
  - Hyundai Palisade 2023 (non-HDA2): `f2cf0de0119d2a16|2023-03-03--21-13-05`
  - Kia Telluride 2023 (non-HDA2):
- [x] harness type:
  - Hyundai Palisade 2023:
    - [Hyundai A](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide) (non-HDA2) ✅
  - Kia Telluride 2023:
    - [Hyundai L](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide) (non-HDA2) ✅
- [x] architecture: CAN-FD

**Link to all physical parts during installation**
  - [comma 3X](https://comma.ai/shop/comma-3x)
  - [car harness kit](https://comma.ai/shop/car-harness)
    - Hyundai Palisade 2023:
      - [Hyundai A](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide) (non-HDA2)
    - Kia Telluride 2023:
      - [Hyundai L](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide) (non-HDA2)

**Software URL**
- https://bit.ly/palisade-2023-no-hda2

**Requires**
- https://github.com/commaai/panda/pull/1267
- https://github.com/commaai/opendbc/pull/784
- https://github.com/commaai/openpilot/pull/29695
- https://github.com/commaai/openpilot/pull/29733

**Resolves**
- https://github.com/commaai/openpilot/issues/26571

**Related**
- https://github.com/commaai/openpilot/pull/27392
- https://github.com/commaai/panda/pull/1243

Thanks to community Hyundai Palisade 2023 owner `SKILLED#7567` (Discord).

PR #27392, opened 2023-02-19
HKG: Car Port for Hyundai Palisade and Kia Telluride 2023-24 (HDA2)

sunnyhaibin:
**Checklist**
- [x] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with stock system:
  - Hyundai Palisade 2023: `28aa956828c3407d|2023-01-28--21-56-53`
  - Kia Telluride 2023: `caf6a54b6d467dbd|2023-02-18--20-59-58`
- [x] route with openpilot:
  - Hyundai Palisade 2023 (HDA2): `28aa956828c3407d|2023-08-27--03-22-18`
  - Kia Telluride 2023 (HDA2): `caf6a54b6d467dbd|2023-02-28--21-18-30`
- [x] harness type:
  - Hyundai Palisade 2023:
    - ~~[Hyundai N](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)~~ [Hyundai A (Modified)](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide) (HDA2), with CAN bus pairs flipped manually (see links below with instructions) ✅
      - https://cdn.discordapp.com/attachments/524611978208215070/1003351781251756182/unknown.png
      - [Video instruction on how to flip CAN bus pairs](https://github.com/commaai/openpilot/assets/47793918/c0a52d4f-f759-4323-939e-fd759a7e9b1e)
      - https://github.com/commaai/openpilot/issues/27580 is the direct replacement of the Hyundai A (Modified)
  - Kia Telluride 2023:
    - [Hyundai P](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide) (HDA2) ✅
- [x] architecture: CAN-FD

**Link to all physical parts during installation**
  - [comma 3X](https://comma.ai/shop/comma-3x)
  - [car harness kit](https://comma.ai/shop/car-harness)
    - Hyundai Palisade 2023:
      - [Hyundai R](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide) (HDA2)
    - Kia Telluride 2023:
[...]

PR #29663, opened 2023-08-27
Hyundai: consolidate button messages creation

sunnyhaibin:
**Prerequisite of**
- https://github.com/commaai/openpilot/pull/27392
- https://github.com/commaai/openpilot/pull/27478


selfdrive/car/hyundai/carcontroller.py
```diff
@@ -56,6 +56,41 @@ def __init__(self, dbc_name, CP, VM):
...
+          if self.CP.flags & HyundaiFlags.CANFD_ALT_BUTTONS:
+            # TODO: resume for alt button cars
+            pass
+          else:
+            for _ in range(20):
+              can_sends.append(hyundaicanfd.create_buttons(self.packer, self.CP, self.CAN, CS.buttons_counter+1, Buttons.RES_ACCEL))
+            self.last_button_frame = self.frame
+
+    return can_sends
+
```
sshane, line 93: If you just need to send one or the other with the Palisade, you just just have one bool

sunnyhaibin: Sounds good, updated!

PR #29664, opened 2023-08-27
Update French translation

Korben00:
Hello

I noticed that some translations were sticking out a bit in the new interface, so I'm proposing an update to correct this. I've arranged for the translations that cause problems to have the same number of letters including spaces as the original sentence so that they fit in the UI.

@maxime-desroches if you can give me your feedback, that would be very kind. Thank you.

### DIFF
--- a/selfdrive/ui/translations/main_fr.ts
+++ b/selfdrive/ui/translations/main_fr.ts
@@ -517,7 +517,7 @@
     <name>PrimeAdWidget</name>
     <message>
         <source>Upgrade Now</source>
-        <translation>Mettre à niveau maintenant</translation>
+        <translation>Mettre à niveau</translation>
     </message>
     <message>
         <source>Become a comma prime member at connect.comma.ai</source>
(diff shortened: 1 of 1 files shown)


selfdrive/ui/translations/main_fr.ts
```diff
@@ -860,7 +860,7 @@ Cela peut prendre jusqu&apos;à une minute.</translation>
     <name>SoftwarePanel</name>
     <message>
         <source>Updates are only downloaded while the car is off.</source>
-        <translation>Les mises à jour sont téléchargées uniquement lorsque la voiture est éteinte.</translation>
+        <translation>MàJ téléchargées si la voiture est éteinte.</translation>
```
maxime-desroches, line 863:
```suggestion
        <translation>Les MàJ sont téléchargées uniquement si la voiture est éteinte.</translation>
```

sshane: lgtm

image

sshane:
thanks!

PR #29665, opened 2023-08-27
ui/NavManager: parse location timestamp from API

### LINKED ISSUE #29074: ui: sort non-favorite nav destinations by max of api time, device time

incognitojam:
The API now includes "modified" timestamps for saved locations. This is the time the location was "last modified", and is useful for determining when the device navigated to a "recent" location.

https://api.comma.ai/#retrieve-saved-locations

Parse and combine this timestamp with the local "last activity" time to improve location sorting.

Related: #29079
Closes: #29074

### DIFF
--- a/selfdrive/ui/qt/maps/map_settings.cc
+++ b/selfdrive/ui/qt/maps/map_settings.cc
@@ -15,6 +15,11 @@ static bool locationEqual(const QJsonValue &v1, const QJsonValue &v2) {
   return v1["latitude"] == v2["latitude"] && v1["longitude"] == v2["longitude"];
 }
 
+static qint64 convertTimestampToEpoch(const QString &timestamp) {
+  QDateTime dt = QDateTime::fromString(timestamp, Qt::ISODate);
+  return dt.isValid() ? dt.toSecsSinceEpoch() : 0;
+}
+
 MapSettings::MapSettings(bool closeable, QWidget *parent) : QFrame(parent) {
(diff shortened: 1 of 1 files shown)


incognitojam:
Tested and seems to work well on device, any comments @deanlee ?

deanlee:
looks good! 
However, will there be synchronization problems? such as the time returned by the API may be less than the time saved locally?

PR #29672, opened 2023-08-28
radard: tie radard frequency to modelV2 vol. 3

fredyshox:


fredyshox:
This test route from Toyota Avalon Hybrid 2019 seems sketchy (https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py#L165). First 238 cans are reported as invalid. Previous implementation of radar_interface ignored `can_valid`, because it didn't spot msg at triggering address

sshane:
Re: Avalon route - in older versions of openpilot we always kept the safety param at 0, multiplexing bus 1 to the OBD port, so we didn't get radar messages during that time. We nee to either wait a few seconds more to check validity, or detect when it stops multiplexing with something like https://github.com/commaai/openpilot/pull/29759

sshane:
can you update the test to ensure it adds radar errors without the header msg?

selfdrive/controls/tests/test_leads.py
```diff
@@ -29,7 +29,7 @@ def single_iter_pkg():
     states = [m for m in out if m.which() == "radarState"]
     failures = [not state.valid and len(state.radarState.radarErrors) for state in states]
 
-    self.assertTrue(len(states) == 0 or all(failures))
+    self.assertTrue(all(failures))
```
sshane, line 32: if states was previously empty, this will still allow that (all on an empty list is true)

selfdrive/car/tests/test_car_interfaces.py
```diff
@@ -123,7 +123,10 @@ def test_car_interfaces(self, car_name, data):
     if not car_params.radarUnavailable and radar_interface.rcp is not None:
       cans = [messaging.new_message('can', 1).to_bytes() for _ in range(5)]
       rr = radar_interface.update(cans)
-      self.assertTrue(rr is None or len(rr.errors) > 0)
+      self.assertTrue(len(rr.errors) > 0)
+
+      rr = radar_interface.update([])
+      self.assertTrue(len(rr.errors) > 0)
```
sshane, line 129: I would assert it's false, then assert it's true after a single run

fredyshox: can you elaborate? `rr` is not defined before running update

sshane: ah I see, perhaps then just switch these two conditions so it first runs on an empty list

selfdrive/controls/tests/test_leads.py
```diff
@@ -29,7 +29,7 @@ def single_iter_pkg():
     states = [m for m in out if m.which() == "radarState"]
     failures = [not state.valid and len(state.radarState.radarErrors) for state in states]
 
-    self.assertTrue(len(states) == 0 or all(failures))
+    self.assertTrue(len(failures) != 0 and all(failures))
```
sshane, line 32:
```suggestion
    self.assertTrue(len(failures) > 0 and all(failures))
```

selfdrive/car/chrysler/radar_interface.py
```diff
@@ -51,16 +51,26 @@ def update(self, can_strings):
...
+    if ret.parseCompleted:
+      ret.points = self._update_radar_points(self.updated_messages)
+      self.updated_messages.clear()
+    else:
+      ret.points = []
+      ret.errors = self._radar_errors()
+
+    return ret
+
+  def _radar_errors(self):
```
sshane, line 65:
```suggestion
  def _get_errors(self) -> List[...]:
```

selfdrive/car/chrysler/radar_interface.py
```diff
@@ -51,16 +51,26 @@ def update(self, can_strings):
...
-      return None
-
     ret = car.RadarData.new_message()
+    ret.parseCompleted = self.trigger_msg in self.updated_messages
+    if ret.parseCompleted:
+      ret.points = self._update_radar_points(self.updated_messages)
+      self.updated_messages.clear()
+    else:
+      ret.points = []
+      ret.errors = self._radar_errors()
```
sshane, line 61: are you never adding any radar errors if parseCompleted?

selfdrive/car/toyota/radar_interface.py
```diff
@@ -36,46 +39,67 @@ def __init__(self, CP):
...
           self.valid_cnt[ii] = max(self.valid_cnt[ii] - 1, 0)
 
-        score = self.rcp.vl[ii+16]['SCORE']
-        # print ii, self.valid_cnt[ii], score, cpt['VALID'], cpt['LONG_DIST'], cpt['LAT_DIST']
+        n_b_scores = len(radar_b_msgs['SCORE'])
+        if n_b_scores > 0:
+          score_index = min(index, n_b_scores - 1)
+          score = radar_b_msgs['SCORE'][score_index]
+        else:
+          score = None
```
sshane, line 99: this is making this code more confusing to read

selfdrive/controls/radard.py
```diff
@@ -98,11 +103,12 @@ def get_RadarState(self, model_prob: float = 0.0):
       "vLead": float(self.vLead),
       "vLeadK": float(self.vLeadK),
       "aLeadK": float(self.aLeadK),
+      "aLeadTau": float(self.aLeadTau),
       "status": True,
       "fcw": self.is_potential_fcw(model_prob),
       "modelProb": model_prob,
       "radar": True,
-      "aLeadTau": float(self.aLeadTau)
+      "radarTrackId": self.identifier,
```
sshane, line 111: do we need this?

selfdrive/controls/radard.py
```diff
@@ -158,23 +164,25 @@ def get_RadarState_from_vision(lead_msg: capnp._DynamicStructReader, v_ego: floa
...
+    "radarTrackId": -1,
   }
 
 
 def get_lead(v_ego: float, ready: bool, tracks: Dict[int, Track], lead_msg: capnp._DynamicStructReader,
              model_v_ego: float, low_speed_override: bool = True) -> Dict[str, Any]:
   # Determine leads, this is where the essential logic happens
-  if len(tracks) > 0 and ready and lead_msg.prob > .5:
+  lead_msg_empty = any([lead_msg.x, lead_msg.y, lead_msg.v])
+  if len(tracks) > 0 and ready and not lead_msg_empty and lead_msg.prob > .5:
```
sshane, line 177: what does this do?

selfdrive/controls/radard.py
```diff
@@ -204,41 +212,44 @@ def __init__(self, radar_ts: float, delay: int = 0):
...
-      self.tracks[ids].update(rpt[0], rpt[1], rpt[2], v_lead, rpt[3])
+        # create the track if it doesn't exist or it's a new track
+        if ids not in self.tracks:
+          self.tracks[ids] = Track(ids, v_lead, self.kalman_params)
+        self.tracks[ids].update(rpt[0], rpt[1], rpt[2], v_lead, rpt[3])
+    else:
+      # *** no radar points, keep existing tracks, update v_lead
+      for track in self.tracks.values():
+        v_lead = track.vRel + self.v_ego_hist[0]
+        track.update_vlead(v_lead)
```
sshane, line 252: does this preserve behavior or change it? Previously when returning None, it just skipped running any updates until it got the next header message. Now it runs `update_vlead` while there's no header message

sshane (changes requested):

selfdrive/controls/radard.py
fredyshox: previously the tracks were always removed/added because it run only when radarData was not None. Now it's possible that update gets executed before RI seen trigger message, so in that case it only updates vLead based on last vLead value and new vEgo

selfdrive/controls/radard.py
fredyshox: i think this was for fuzzy tests to pass - fuzzy test would create leads with x/y/v of variable length

selfdrive/controls/radard.py
fredyshox: IMO it's useful for debugging - I've used that to track differences Tracks between two versions of radard

selfdrive/car/chrysler/radar_interface.py
fredyshox: bad indent

selfdrive/controls/radard.py
sshane: can be a new PR then

selfdrive/controls/radard.py
sshane: Right, but it would only publish that update once, never extrapolating the vLead data (correct?). Now we have two choices: publish the tracks with the same data, except updating vLead, or publish the tracks with the same data without modification (until we have an update).

sshane: Though isn't it intended we only do this once, since we expect delay in the radar measurement? This is only executed for Honda which sets `self.delay` in RadarInterface.

selfdrive/controls/radard.py
fredyshox: sure: https://github.com/commaai/openpilot/pull/29809

selfdrive/controls/radard.py
fredyshox: Not sure if i follow that hondas delay actually mean.

```
   modelV2        modelV2         modelV2

----- | ------------ | ------------- |
--- | ---------------- | ---------------- |
 trigger            trigger            trigger
                       
```
^This is the case that this covers. If radar is slow (some of them run at sub 20hz), the update may be called without any points from RadarInterface (second modelV2 in the graph above), since its tied to modelV2 now. carState updates at 100Hz, so we get new vEgo during that time, so we update existing lead speeds accordingly.

PR #29679, opened 2023-08-28
CI: try car tests with double the processor count

jnewb1:


jnewb1:
no difference for car or replay tests

PR #29681, opened 2023-08-28
Tests: test_caching reduce global state

jnewb1:
use mock and tempdir instead of the env variable to make the test independent of other tests

### DIFF
--- a/tools/lib/tests/test_caching.py
+++ b/tools/lib/tests/test_caching.py
@@ -1,21 +1,11 @@
 #!/usr/bin/env python3
 import os
 import unittest
-import tempfile
-from unittest import mock
-
 from openpilot.tools.lib.url_file import URLFile
+from tools.lib.tests import temporary_cache_dir
 
 
 class TestFileDownload(unittest.TestCase):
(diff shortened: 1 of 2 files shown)


tools/lib/tests/test_caching.py
```diff
@@ -1,18 +1,24 @@
...
-os.environ["COMMA_CACHE"] = "/tmp/__test_cache__"
-from openpilot.tools.lib.url_file import URLFile, CACHE_DIR
+from openpilot.tools.lib.url_file import URLFile
 
 
 class TestFileDownload(unittest.TestCase):
+  def setUp(self):
+    self.temp_dir = tempfile.TemporaryDirectory()
+    self.cache_dir_patch = mock.patch("openpilot.tools.lib.url_file.CACHE_DIR", self.temp_dir.name)
+    self.cache_dir_patch.start()
```
adeebshihadeh, line 14: seems like something that we'll want to use in a lot of places. I'd setup a decorator to do this

jnewb1: done

PR #29690, opened 2023-08-29
Subaru: cleanup and fix preglobal

jnewb1:
Preglobal (currently in dashcam mode) was broken on master:

- COUNTER has to be set in OP since we have to make the checksum in python
- brakepressed was mismatched from the panda safety
- added ability to specific the preglobal checksum byte, which is required for ES_Status and ES_Brake since the checksum is in a different place

Also some cleanup for subaru:
- carstate was also pretty messy and had lots of duplicates, cleaned that up

### DIFF
--- a/selfdrive/car/subaru/carcontroller.py
+++ b/selfdrive/car/subaru/carcontroller.py
@@ -36,7 +36,7 @@ class CarController:
         apply_steer = 0
 
       if self.CP.carFingerprint in PREGLOBAL_CARS:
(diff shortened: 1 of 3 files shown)


sshane (approved):

sshane:
If you want to catch more of these cases, you can write a quick PR to satisfy https://github.com/commaai/opendbc/issues/914

adeebshihadeh:
> If you want to catch more of these cases, you can write a quick PR to satisfy [commaai/opendbc#914](https://github.com/commaai/opendbc/issues/914)

This should be the standard workflow. Resolving the opendbc issue has 10x the value of this individual fix.

PR #29691, opened 2023-08-29
docker: multiarch images (amd64/aarch64)

fredyshox:
- [x] update build script to build arch specific images
- [x] script merge manifests
- [x] CI job for arm build using buildjet
- [x] CI job for merging manifests
- [x] check if it doesn't break existing CI jobs and caching


.github/workflows/selfdrive_tests.yaml
```diff
@@ -202,6 +202,39 @@ jobs:
       run: |
         eval "$BUILD_CL"
 
+  docker_push_arm64:
+    name: docker push (arm64)
+    runs-on: buildjet-4vcpu-ubuntu-2204-arm
```
jnewb1, line 207: is there a github actions runner for arm? we are trying to limit our buildjet usage. If not, make sure you put it behind the same flag to prevent forks from being forced to use buildjet

selfdrive/test/docker_tag_multiarch.sh
```diff
@@ -0,0 +1,49 @@
+#!/usr/bin/env bash
+set -e
+
+if [ $# -lt 2 ]; then
+  echo "Usage: $0 <base|docs|sim|prebuilt|cl> <arch1> <arch2> ..."
+  exit 1
+fi
+
+if [ $1 = "base" ]; then
```
jnewb1, line 9: can you make a common script for these checks and source it in both the scripts?

selfdrive/test/docker_tag_multiarch.sh
```diff
@@ -0,0 +1,49 @@
...
+  export DOCKER_FILE=Dockerfile.openpilot
+elif [ $1 = "cl" ]; then
+  export DOCKER_IMAGE=openpilot-base-cl
+  export DOCKER_FILE=Dockerfile.openpilot_base_cl
+else
+  echo "Invalid docker build image $1"
+  exit 1
+fi
+
+export DOCKER_REGISTRY=ghcr.io/commaai
```
jnewb1, line 29: same with these, maybe docker_common.sh?

.github/workflows/selfdrive_tests.yaml
jnewb1: maybe qemu https://github.com/docker/setup-qemu-action

selfdrive/test/docker_build.sh
```diff
@@ -28,9 +28,10 @@ fi
 export DOCKER_REGISTRY=ghcr.io/commaai
 export COMMIT_SHA=$(git rev-parse HEAD)
 
+ARCH=$(uname -m)
 LOCAL_TAG=$DOCKER_IMAGE
-REMOTE_TAG=$DOCKER_REGISTRY/$LOCAL_TAG
-REMOTE_SHA_TAG=$REMOTE_TAG:$COMMIT_SHA
+REMOTE_TAG=$DOCKER_REGISTRY/$LOCAL_TAG:$ARCH
```
jnewb1, line 33: should arch be a part of the tag? my thought was it would be part of the name, but not sure about that.
maybe openpilot-base becomes the multiarch one, and openpilot-base-arm64 and openpilot-base-amd64 are separate images

.github/workflows/selfdrive_tests.yaml
fredyshox: this would only be run on master, so the usage shouldn't be too high

selfdrive/test/docker_build.sh
fredyshox: thats a good point, it's less confusing that way

.github/workflows/selfdrive_tests.yaml
jnewb1: Perhaps we also want to test the arm build on non-master though? So no one breaks it without knowing until they push to master, like if they install an x86 only package

jnewb1:
Is the multiarch stuff just for the Mac devcontainer? Or is there other uses too

fredyshox:
yes, it's mainly for that.

jnewb1:
I wonder if it makes more sense to have the user just build the image locally then, especially if we have to make a non-root image too

fredyshox:
The problem is this ruins OOBE experience for devcontainers, since you have to build openpilot-base image manually before running a devcontainer (and probably modify parent image's tag to local one). This is what im currently doing when im working on my mac, and it's not great. 

We could write some scripts to help with that, but then again there're problems like:
- keeping local base image up to date
- setup would take much longer: 10-15 minutes instead of 1-2 minutes
- users experienced with devcontainers will run it directly after cloning, and get stuck with emulated x64 environment which is painfully slow

jnewb1:
@adeebshihadeh what do you think about using buildjet to build the arm image? There is no arm github runners and qemu will be quite slow

.github/workflows/selfdrive_tests.yaml
```diff
@@ -91,6 +91,34 @@ jobs:
         path: .ci_cache/scons_cache
         key: scons-${{ env.CACHE_COMMIT_DATE }}-${{ github.sha }}
 
+  build_arm64:
+    name: build all (arm64)
+    runs-on: buildjet-4vcpu-ubuntu-2204-arm
```
jnewb1, line 96: probably only need 2 vcpu, which is the same as what's on the github machines. Can you also make this a composite since it's repeated?

fredyshox: what do you mean by composite?

jnewb1: similar to .github/workflows/setup, like a macro to be used elsewhere

fredyshox:
@jnewb1 made build job a composite action, and created matrix strategy covering each arch. 2vcpu machine builds only 7 minutes slower than 4vcpu (clean build 22 min vs 15 min, so its not 2x and most of the time is consumed by setup/install process), which i think is acceptable. After we fill some caches, the total cost should be much lower than process_replay

jnewb1:
> @jnewb1 made build job a composite action, and created matrix strategy covering each arch. 2vcpu machine builds only 7 minutes slower than 4vcpu (clean build 22 min vs 15 min, so its not 2x and most of the time is consumed by setup/install process), which i think is acceptable. After we fill some caches, the total cost should be much lower than process_replay

yea it will probably only be a few minutes once the cache is filled. lgtm

.github/workflows/build/action.yaml
```diff
@@ -0,0 +1,27 @@
+name: 'openpilot build'
```
adeebshihadeh, line 1: I would call this action "compile-openpilot", build can be ambiguous (docker build, scons, etc.)

.github/workflows/selfdrive_tests.yaml
```diff
@@ -67,30 +67,20 @@ jobs:
...
-                          scons -j$(nproc) --cache-populate"
-    - name: Save scons cache
-      uses: actions/cache/save@v3
-      if: github.ref == 'refs/heads/master'
+    - uses: ./.github/workflows/build
+      timeout-minutes: ${{ ((steps.restore-scons-cache.outputs.cache-hit == 'true') && 15 || 30) }} # allow more time when we missed the scons cache
       with:
-        path: .ci_cache/scons_cache
-        key: scons-${{ env.CACHE_COMMIT_DATE }}-${{ github.sha }}
+        cache_key_prefix: ${{ (matrix.arch == 'aarch64') && 'scons_aarch64' || 'scons' }}
```
adeebshihadeh, line 83: why not always include the arch in the prefix?

.github/workflows/setup/action.yaml
```diff
@@ -5,6 +5,10 @@ inputs:
     description: 'Whether or not to pull the git lfs'
     required: false
     default: 'true'
+  cache_key_prefix:
+    description: 'Prefix for caching key'
+    required: false
+    default: 'scons'
   is_retried:
```
adeebshihadeh, line 12: why doesn't this use the new build action?

.github/workflows/selfdrive_tests.yaml
```diff
@@ -185,24 +175,46 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    if: github.ref == 'refs/heads/master' && github.event_name != 'pull_request' && github.repository == 'commaai/openpilot'
+    needs: [docker_push]
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: false
+    - name: Setup to push to repo
+      run: |
+        echo "PUSH_IMAGE=true" >> "$GITHUB_ENV"
```
adeebshihadeh, line 212: why is this here? it's not used in this step

fredyshox: docker_tag_multiarch.sh uses that to determine whether it should push merged manifest to registry - similarly to docker_build.sh

adeebshihadeh: i’m saying it should go there, ideally before the command

.github/workflows/setup/action.yaml
fredyshox: im not sure if i follow, this only builds base images and restores caches, while build actually compiles stuff

.github/workflows/selfdrive_tests.yaml
fredyshox: wanted to keep it compatible with existing cache entries, but i can change that if we are ok with breaking that

adeebshihadeh: that's fine, the caches are invalidated all the time

fredyshox: ok

PR #29693, opened 2023-08-29
Revert "IsoTpParallelQuery: extend timeout for every frame type"

sshane:
Reverts commaai/openpilot#29654

See panda PR for details: https://github.com/commaai/panda/pull/1616


sshane:
Will try to get https://github.com/commaai/panda/pull/1329 finally merged, testing IsoTpParallelQuery with it before re-merging

PR #29658, opened 2023-08-26
Hyundai CAN-FD HDA2: specify signals to suppress LFA

sunnyhaibin:
**Requires**
- https://github.com/commaai/opendbc/pull/927

image

PR #29696, opened 2023-08-29
Parser: assert no duplicate message checks

jnewb1:
depends:
matching opendbc: https://github.com/commaai/opendbc/pull/930
~~sunny's dbc change:  https://github.com/commaai/openpilot/pull/29658~~

### DIFF
--- a/selfdrive/car/hyundai/carstate.py
+++ b/selfdrive/car/hyundai/carstate.py
@@ -305,19 +305,19 @@ class CarState(CarStateBase):
   def get_can_parser_canfd(self, CP):
     messages = [
       (self.gear_msg_canfd, 100),
-      (self.cruise_btns_msg_canfd, 50),
       (self.accelerator_msg_canfd, 100),
       ("WHEEL_SPEEDS", 100),
       ("STEERING_SENSORS", 100),
       ("MDPS", 100),
       ("TCS", 50),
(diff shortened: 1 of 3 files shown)


selfdrive/car/hyundai/carstate.py
```diff
@@ -311,11 +311,15 @@ def get_can_parser_canfd(self, CP):
...
-      ("CRUISE_BUTTONS_ALT", 50),
       ("BLINKERS", 4),
       ("DOORS_SEATBELTS", 4),
     ]
 
+    if not (CP.flags & HyundaiFlags.CANFD_ALT_BUTTONS):
+      messages += [
+        ("CRUISE_BUTTONS_ALT", 50)
+      ]
+
```
sshane, line 322: is this right?

jnewb1: not super familiar with hyundai, but this matches the current setup, just without duplicate messages.

CRUISE_BUTTONS_ALT is used for the metric test, even if the buttons aren't used? https://github.com/commaai/openpilot/blob/ddddc7a8cbb9de28b24cc7d762c25b57ad43e620/selfdrive/car/hyundai/carstate.py#L169

jnewb1: I think its more clear if we just add CRUISE_BUTTONS if the flag is not set

PR #29702, opened 2023-08-29
rename GPIO pin with ublox in the name that is not ublox specific

gregjhogan:


common/gpio.h
```diff
@@ -5,7 +5,7 @@
   #define GPIO_HUB_RST_N        30
   #define GPIO_UBLOX_RST_N      32
   #define GPIO_UBLOX_SAFEBOOT_N 33
-  #define GPIO_UBLOX_PWR_EN     34
+  #define GPIO_GNSS_PWR_EN      34
```
adeebshihadeh, line 8: can you keep the pin name from the schematic as a comment?

gregjhogan: sure, I added it as an inline comment here and in pins.py

PR #29704, opened 2023-08-29
Devcontainer: use "batman" user instead of root

jnewb1:
if you use the root user, any files you modify will then be chmoded incorrectly and cause permission issues when you try to edit locally again


jnewb1:
did you also get this issue testing on mac @fredyshox

fredyshox:
it doesn't seem to be an issue on mac, i'm still owning all the files, regardless of who created them while in container

.devcontainer/devcontainer.json
```diff
@@ -4,7 +4,7 @@
     "dockerfile": "Dockerfile"
   },
   "postCreateCommand": "bash -c 'if [[ $DISPLAY == *xquartz* ]]; then echo \"export DISPLAY=host.docker.internal:0\" >> /root/.bashrc; fi'",
```
fredyshox, line 6: This doesn't work now, no permissions to edit roots bashrc

.devcontainer/setup_container.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Link root to new batman user
+rm /home/batman/.bashrc
+ln -s /root/.pyenv      /home/batman/.pyenv
```
fredyshox, line 5: with this batman doesn't have permissions to install new packages. needs to be chmoded/chowned

.devcontainer/Dockerfile
```diff
@@ -1,6 +1,6 @@
 FROM ghcr.io/commaai/openpilot-base:latest
 
 # remove gitconfig if exists, since its gonna be replaced by host one
-RUN rm -f /root/.gitconfig
+RUN rm -f /home/batman/.gitconfig
```
fredyshox, line 4: this line is not needed anymore in that case, since it doesn't exist in parent image

.devcontainer/setup_container.sh
```diff
@@ -0,0 +1,16 @@
...
+ln -s /root/.pyenvrc    /home/batman/.pyenvrc
+ln -s /root/.bashrc     /home/batman/.bashrc
+ln -s /root/.Xauthority /home/batman/.Xauthority
+
+# Setup permissions
+sudo chown batman: /root
+sudo chmod u+w /root
+sudo chown batman: /tmp/scons_cache
```
fredyshox, line 13: add `-R`, otherwise scons won't be able to interact with cache, if it was previously used with root

.devcontainer/setup_container.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Link root to new batman user
+rm /home/batman/.bashrc
+ln -s /root/.pyenv      /home/batman/.pyenv
+ln -s /root/.pyenvrc    /home/batman/.pyenvrc
+ln -s /root/.bashrc     /home/batman/.bashrc
+ln -s /root/.Xauthority /home/batman/.Xauthority
```
fredyshox, line 8: why not mounting it directly on /home/batman?

jnewb1:
image
example: ran scons in the container and it made files as the root use, that now can't be overwritten locally

.devcontainer/devcontainer.json
jnewb1: can we run the devcontainer tests on macos too?

fredyshox:
IMO that would be much nicer, because in current state setup_container.sh would need some additional maintenance in the future. Should be straightforward to do.

.devcontainer/devcontainer.json
```diff
@@ -14,12 +14,22 @@
...
-    "--volume=${localWorkspaceFolder}/.devcontainer/.Xauthority:/root/.Xauthority",
-    "--volume=${localEnv:HOME}/.comma:/root/.comma",
+    "--volume=${localWorkspaceFolder}/.devcontainer/.Xauthority:/home/batman/.Xauthority",
+    "--volume=${localEnv:HOME}/.comma:/home/batman/.comma",
     "--volume=/tmp/comma_download_cache:/tmp/comma_download_cache",
     "--volume=/tmp/devcontainer_scons_cache:/tmp/scons_cache",
     "--shm-size=1G"
   ],
+  "features": {
+    "ghcr.io/devcontainers/features/common-utils:2": {
```
adeebshihadeh, line 24: this features thing is pretty cool, perhaps we should use it instead of stuff like this https://github.com/commaai/openpilot/pull/29967

Dockerfile.openpilot_base
```diff
@@ -24,7 +24,7 @@ RUN cd /tmp && \
 
 ENV POETRY_VIRTUALENVS_CREATE=false
 ENV PYENV_VERSION=3.11.4
-ENV PYENV_ROOT="/root/.pyenv"
+ENV PYENV_ROOT="/opt/pyenv"
```
jnewb1, line 27: if we move the pyenv to /opt, it will still allow for the CI to work normally, but the devcontainer can chown it as batman and allow for writes

jnewb1:
devcontainer CI will fail until the upstream base image moves to /opt/pyenv

.devcontainer/devcontainer.json
jnewb1: I think they aren't as useful for one-off features like this, unless we wanted to have a devcontainer for the submodules too. They require creating docker artifacts, and need a whole CI setup. Doesn't look like they support local features.

https://github.com/devcontainers/feature-starter

adeebshihadeh:
> devcontainer CI will fail until the upstream base image moves to /opt/pyenv

We should fix that first. CI shouldn't really on any state from previous runs or master.

.devcontainer/host_setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # pull base image
-docker pull ghcr.io/commaai/openpilot-base:latest
```
jnewb1, line 4: it might make since to rebuild the base image here instead of pulling, so you can test local changes to the base image. It should be about the same amount of time, since it will use the remote image as cache

jnewb1:
> > devcontainer CI will fail until the upstream base image moves to /opt/pyenv
> 
> We should fix that first. CI shouldn't really on any state from previous runs or master.

fixed by building image first in CI

Dockerfile.openpilot_base
gregjhogan: There is no problem with user id 1000 always owning the files.  Create the user with id 1000 with `RUN useradd -m -s /bin/bash -u 1000 <username>` (use batman if you like, all that matters is the user id) and set it active with `USER <username>`, then when you install it will already be owned by that user (I would put `PYENV_ROOT` in `/home/<username>/.pyenv` to keep things simple).  This should allow you to delete what you added to `.devcontainer/container_post_create.sh`.  If we still want the container to run as root by default you can issue another USER command at the end to switch back to root.

gregjhogan:
> > > devcontainer CI will fail until the upstream base image moves to /opt/pyenv
> > 
> > 
> > We should fix that first. CI shouldn't really on any state from previous runs or master.
> 
> fixed by building image first in CI

None of this permission switching should be necessary, we can simplify, see [here](https://github.com/commaai/openpilot/pull/29704#discussion_r1333832079).

Dockerfile.openpilot_base
jnewb1: done

.devcontainer/container_post_create.sh
```diff
@@ -12,3 +12,6 @@ if [ -n "\$DISPLAY" ]; then
 fi
 EOF
 fi
+
+USER=batman
+sudo chown -R $USER: /tmp/scons_cache
```
gregjhogan, line 17: This is not necessary if you create the directories in their current location in the Dockerfile after switching users:
```
USER $USER
...
RUN mkdir /tmp/comma_download_cache
RUN mkdir /tmp/scons_cache
```
This is better because it has the benefit of things working if you use the docker container outside of using vscode dev containers.

Dockerfile.openpilot_base
```diff
@@ -22,18 +22,23 @@ RUN cd /tmp && \
     rm -rf arm/ && \
     rm -rf thumb/nofp thumb/v6* thumb/v8* thumb/v7+fp thumb/v7-r+fp.sp
 
+ARG USER=batman
+RUN useradd -m -s /bin/bash -u 1000 $USER
```
gregjhogan, line 26: I would recommend also adding an `ARG` for the user ID. It makes it easy to build the container correctly if you log into your physical system as a user ID other than 1000.  The username used inside the docker container is irrelevant, what is important is the user ID matches the desired user ID outside the docker container (they can have different names).

.devcontainer/container_post_create.sh
jnewb1: These are actually volume mounted in, so what happens to me here was my old dev container was running as root, it created files as root, and then I switched to batman and the volume still had files as root. Perhaps we should leave this so we don't break the devcontainer for existing users?

Dockerfile.openpilot_base
jnewb1: The dev container feature actually changes the UID of the "Batman" user to the UID of the host user

.devcontainer/container_post_create.sh
gregjhogan: I think that is just a question of when we should remove these lines from `.devcontainer/container_post_create.sh`.  I assume backward compatibility is only necessary for a user until this script runs once.  If we want to support backward compatibility I don't think we want to support it forever, and moving the directory creation to the `Dockerfile` is a prerequisite, so I think it should still be done now.  If we keep these lines in this file I would also add a comment explaining this is for backwards compatibility and that we should delete this in the future.

jnewb1: Added a comment about how it should be temporary. But adding it to the dockerfile won't do anything AFAIK because it's volume mounted in (to preserve the cache across devcontainer restarts), so it will just copy the chown of the host

Dockerfile.openpilot_base
gregjhogan: This is kind of wild, I guess they try to re-build the container for you (or build a new container inheriting from the specified container?) if they detect the user ID inside the container doesn't match the user ID outside the container?!
https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user#_specifying-a-user-for-vs-code
```
However, UID/GID updates are only applied when the container is created and requires a rebuild to change.
```

Another section seems to indicate their magic maybe doesn't always work and so they provide an alternative example where you parameterize the user ID inside your Dockerfile:
https://code.visualstudio.com/remote/advancedcontainers/add-nonroot-user#_change-the-uidgid-of-an-existing-container-user

.devcontainer/container_post_create.sh
gregjhogan: OK yeah I see now it is fine.  I thought they were declared in the Dockerfile using the `VOLUME` keyword (which always creates as root) and it wouldn't work if you run the container as a non-root user and didn't mount in anything, but I see  we don't declare these directories `VOLUME` in the Dockerfile and because /tmp is world writable everything does work.

PR #29707, opened 2023-08-30
HKG: Add FW Versions for Kia Sorento ICE 2023

sunnyhaibin:
Add firmware for the 2023 Kia Sorento ICE.

Dongle ID: `dcad9b1ccb991be4`

Thanks to the community 2023 Sorento ICE owner technodude (Discord).


sshane:
this would've fingerprinted if we whitelisted this car

PR #29708, opened 2023-08-30
ui/map: roundabout icon fix

### LINKED ISSUE #25480: nav: parse roundabout step properly
### Describe the bug

Instruction show first right, but it's 3rd right...

6111bd7df5043394|2022-08-18--17-58-07

### openpilot version

fix-lfa-warn

### Additional info

_No response_

jakethesnake420:
Should now show the correct roundabout/rotary entry/exit icon. This uses the existing rhd logic to mirror the icons. This is not a great solution if someone is driving a lhd car in rhd traffic and vise versa. In this case it would not display the icon.

Also its possible to have a 0 degree roundabout which requires another icon like this:
image

For now I have just used the existing 'direction_roundabout.png' icon renamed to 'direction_roundabout_right_0.png'.

resolves: https://github.com/commaai/openpilot/issues/25480

Also this needs to be tested better. I just tried a simulated lhd 90 and a rhd 270.


selfdrive/ui/qt/maps/map_instructions.cc
```diff
@@ -89,6 +89,10 @@ void MapInstructions::updateInstructions(cereal::NavInstruction::Reader instruct
     QString fn = "direction_" + type;
     if (!modifier.isEmpty()) {
       fn += "_" + modifier;
+      if (type == "roundabout" || type == "rotary"){
+        // Normalize and round to nearest 45deg
+        fn += "_" + QString::number(((instruction.getManeuverDegrees() + 382) % 360 / 45 * 45));
+      }
```
sshane, line 95: does it make more sense to fix in navd instead?

incognitojam:
I've created a new icon for 0 degrees which matches the current style

direction_roundabout_right_0

incognitojam:
Can you open a PR for the cereal changes too?

selfdrive/ui/qt/maps/map_instructions.cc
jakethesnake420: Hmm I would say it is simpler if its done here. I guess the issue is that the modification won't get captured in the logs. If that's important, then maybe its worth moving the math to navd.py.

jakethesnake420:
> Can you open a PR for the cereal addition too?

done

selfdrive/ui/qt/maps/map_instructions.cc
sshane: Shouldn't need to rename the icons as well, the maneuver direction should just be changed

jakethesnake420: > I think it's worth fixing at the source (navd), those instructions are used in the model as well

Good point however would it not be better if the model had the unmodified data from the response? The rounding is only needed to display the limited number of icons that are available.

> Shouldn't need to rename the icons as well, the maneuver direction should just be changed

The maneuver direction is for the entry angle and does not represent which exit should be taken. Mapbox could specify slight_right roundabout 180 which would not display anything since there is no icon for it.

image

if you use roundabout_exits=true in the request it will output a separate maneuver for the exit as well.  So technically you could show a different icon for every combination of entry/exit maneuvers and degrees.

> Let's fix in navd, a field only specific to one maneuver doesn't sound the cleanest

I agree it doesn't seem like great design but it keeps the code aligned to the mapbox spec and the current code style. The other way this can be done is if you map the degrees to a maneuver name which eliminates the need for the degree field at the cost of accuracy/clarity. This is how I thought of doing it initially but its a kludge imo that does pan out if you want more detailed icons/model data in the future.

I will make another PR that doesn't add a cereal field and does it in navd.py and you can choose what's better.

jakethesnake420: If I remove the degrees field in cereal and overwrite the maneuver modifier i gotta do something like this.

in helpers.py
```
if field_valid(p, 'degrees'):
  if p['type'] == 'roundabout' or p['type'] == 'rotary':
    directions = [
      'straight',     # 0 or 360
      'slight_right', # 45
      'right',        # 90
      'tight_right',  # 135
      'straight',     # 180
      'slight_left',  # 225
      'left',         # 270
      'tight_left'    # 315
    ]
    deg = (p['degrees'] + 360) % 360
    i = (deg + 22) // 45 % 8
    instruction['maneuverModifier'] = directions[i]
```

I don't like this method but if you think the model is better suited to this method then I can submit the PR for it.

I was also thinking that you can use the degree field for other maneuvers by filling it with the bearing_before - bearing_after.

jakethesnake420:
I don't think it can get much better than what I have here with 5 lines of easy to read code

selfdrive/ui/qt/maps/map_instructions.cc
mike8643: So I just round this conversation but this is what I have been using to resolve this issue in navd helpers.py
roundabout
``` 
  if field_valid(p, 'degrees'):
    maneuverDegrees = int(p['degrees'])
    drivingSide = p['driving_side']
    modifier1 = ""
    modifier2 = ""
    if (0 <= maneuverDegrees <= 67) or (292 < maneuverDegrees <= 360):
      modifier1 = "sharp "
    elif (112 < maneuverDegrees <= 157) or (202 < maneuverDegrees <= 247):
      modifier1 = "slight "
    if 0 <= maneuverDegrees <= 157:
      modifier2 = drivingSide
    elif 157 < maneuverDegrees <= 202:
      modifier2 = "straight"
    elif 202 < maneuverDegrees <= 360:
      modifier2 = "left" if drivingSide == "right" else "right"
    instruction['maneuverModifier'] = modifier1 + modifier2

jakethesnake420:
Gonna bump this since this is still a problem. I really don't see how my solution would interfere with the model since I am not affecting it at all. This is only adjusting how the manuevers are displayed.

adeebshihadeh:
So this PR is mostly blocked by a lack of a good way to test this. Can you come up with a good way to test this and other related changes?

jakethesnake420:
@adeebshihadeh might be good to make a UI controller for human in the loop testing for general development like this. The way I tested it before was replaying a route that had a roundabout which is inconvenient.

jakethesnake420:
All i am really willing to do at this point is prove that changing the degrees modifier displays the correct icon as the mapbox API specifies here:
image

I made a little test script to set the modifier and took some screenshots with left and right hand drive.

LHD:

Screenshot from 2023-12-23 02-17-29
Screenshot from 2023-12-23 02-17-34
Screenshot from 2023-12-23 02-17-39
Screenshot from 2023-12-23 02-17-47
Screenshot from 2023-12-23 02-17-51
Screenshot from 2023-12-23 02-17-55
Screenshot from 2023-12-23 02-17-58
Screenshot from 2023-12-23 02-18-02
Screenshot from 2023-12-23 02-18-07


RHD:

Screenshot from 2023-12-23 02-14-12
Screenshot from 2023-12-23 02-14-19
Screenshot from 2023-12-23 02-14-22
Screenshot from 2023-12-23 02-14-25
Screenshot from 2023-12-23 02-14-28
Screenshot from 2023-12-23 02-14-33
Screenshot from 2023-12-23 02-14-36
Screenshot from 2023-12-23 02-14-41
Screenshot from 2023-12-23 02-14-45

selfdrive/ui/qt/maps/map_instructions.cc
jakethesnake420: > So I just found this conversation but this is what I have been using to resolve this issue in navd helpers.py I am not using any additional icons than stock and it seems to work pretty well. roundabout
> 
> ```
>   if field_valid(p, 'degrees'):
>     maneuverDegrees = int(p['degrees'])
>     drivingSide = p['driving_side']
>     modifier1 = ""
>     modifier2 = ""
>     if (0 <= maneuverDegrees <= 67) or (292 < maneuverDegrees <= 360):
>       modifier1 = "sharp "
>     elif (112 < maneuverDegrees <= 157) or (202 < maneuverDegrees <= 247):
>       modifier1 = "slight "
>     if 0 <= maneuverDegrees <= 157:
>       modifier2 = drivingSide
>     elif 157 < maneuverDegrees <= 202:
>       modifier2 = "straight"
>     elif 202 < maneuverDegrees <= 360:
>       modifier2 = "left" if drivingSide == "right" else "right"
>     instruction['maneuverModifier'] = modifier1 + modifier2
> ```

Doing it this way will change the model inputs.

mike8643: If you aren’t trying to change model inputs and solely going after navigation icons for roundabouts, I suggest moving to what Waze does and use the simple roundabout icon with exit number in the center. That can be extracted very easily from the map box api provided. It also lines up with the human instructions they provide as well. 
After driving on my version of the degrees fix, map box does not always supply the correct number of degrees for roundabouts. Exit number is always accurate though. 
Just a thought.

jakethesnake420: Can you provide a route where mapbox doesn't supply the correct angle. If its an issue with mapbox then we can submit an edit. if its an issue with parsing then we need to fix it.

Can you provide an example roundabout where it provides the wrong degrees value.

PR #29709, opened 2023-08-30
CI: use pytest for unittesting

jnewb1:


jnewb1:
won't be any faster until I enable the parallel tests, just want to get the pytest part in first

selfdrive/car/tests/test_fw_fingerprint.py
```diff
@@ -175,7 +175,7 @@ def test_fw_requests(self):
 
 class TestFwFingerprintTiming(unittest.TestCase):
   N: int = 5
-  TOL: float = 0.1
+  TOL: float = 0.12
```
jnewb1, line 178: this test failed occasionally without this. also fails locally for me. Shane's looking to improve the tests: https://github.com/commaai/openpilot/pull/29712#issuecomment-1698732253

selfdrive/athena/tests/test_athenad.py
```diff
@@ -48,6 +48,7 @@ def setUp(self):
       else:
         os.unlink(p)
 
+    dispatcher["listUploadQueue"]() # ensure queue is empty at start
```
jnewb1, line 51: this test also failed occasionally for me without this cleanup

system/tests/test_logmessaged.py
```diff
@@ -46,6 +47,7 @@ def test_simple_log(self, temp_dir):
     assert len(self._get_log_files()) >= 1
 
   @temporary_swaglog_dir
+  @temporary_swaglog_ipc
```
adeebshihadeh, line 50: this common stuff should be done more generically. can we run all tests by default in an OpenpilotPrefix?

jnewb1: you mean a temporary directory that everything uses? maybe we should have a param for "TempWorkspace" and everything uses that as a base?

adeebshihadeh: we already have this (https://github.com/commaai/openpilot/blob/9719b227611c880c76a60df223ef6c2f14f1a6ab/selfdrive/test/process_replay/helpers.py#L9), but can we use it by default?

jnewb1: I think that will require a bunch of refactoring, okay if I move that to a separate PR? I'll make an issue for it

PR #29712, opened 2023-08-30
FW query timing test: mock get_data function for timeout

jnewb1:
rather than actually timing out, use a mock function to simulate the timeout to speed up the tests

@sshane any problem with using a mock function there? saves a bit of time in CI and will also be more consistent. it was failing in pytest sometimes: https://github.com/commaai/openpilot/pull/29709

takes about 30 seconds as is, which makes sense as we run 5x, each with a 6s timeout
```30.61s call     selfdrive/car/tests/test_fw_fingerprint.py::TestFwFingerprintTiming::test_fw_query_timing```

### CODE REVIEW

**Comment by sshane** [comment]
I want to start testing the full fingerprinting stack in openpilot, using https://github.com/commaai/panda/pull/1329, which probably requires the original timeouts. Once I get that merged we look into this?

I think the other test in this file also takes longer.

**Comment by jnewb1** [comment]
> I want to start testing the full fingerprinting stack in openpilot, using https://github.com/commaai/panda/pull/1329, which probably requires the original timeouts. Once I get that merged we look into this?
> 
> 
> 
> I think the other test in this file also takes longer.

Okay. Test also is inconsistent locally, okay if I bump up the tolerance?

**Comment by jnewb1** [comment]
Closing this until we can test full panda stack

**Comment by sshane** [comment]
The individual brands should be fairly consistent, it's the total time that is easy to deviate from the tolerance.

File: selfdrive/car/tests/test_fw_fingerprint.py [hunk 1]
```diff
@@ -183,20 +184,30 @@ def _run_thread(thread: threading.Thread) -> float:
...
     return time.perf_counter() - t
 
+  def fake_get_data(self, timeout):
+    self.fake_timeout_time += timeout
+    return {}
+
   def _benchmark_brand(self, brand, num_pandas):
+    query_patch = mock.patch("openpilot.selfdrive.car.isotp_parallel_query.IsoTpParallelQuery.get_data", self.fake_get_data)
+    query_patch.start()
+
```
sshane, line 201: I didn't realize it dynamically adds whatever timeout IsoTpParallelQuery, can we get this in now, if it will improve reliability and allow us to lower the threshold again

jnewb1: sure I'll reopen this

selfdrive/car/tests/test_fw_fingerprint.py
```diff
@@ -176,28 +177,38 @@ def test_fw_requests(self):
...
     t = time.perf_counter()
+
     while thread.is_alive():
-      time.sleep(0.02)
+      time.sleep(0.001)
       if not params.get_bool("ObdMultiplexingChanged"):
         params.put_bool("ObdMultiplexingChanged", True)
     return time.perf_counter() - t
 
+  def fake_get_data(self, timeout):
```
sshane, line 195: Can make local function and use nonlocal. `get_data_mock`

selfdrive/car/tests/test_fw_fingerprint.py
```diff
@@ -176,28 +177,38 @@ def test_fw_requests(self):
...
   @staticmethod
   def _run_thread(thread: threading.Thread) -> float:
     params = Params()
     params.put_bool("ObdMultiplexingEnabled", True)
     thread.start()
     t = time.perf_counter()
+
     while thread.is_alive():
-      time.sleep(0.02)
+      time.sleep(0.001)
```
sshane, line 190: Can you see if the change in timings for OBD multiplexing matches what it is when fingerprinting on device (think going onroad on a desk is enough)?

selfdrive/car/tests/test_fw_fingerprint.py
```diff
@@ -176,28 +177,38 @@ def test_fw_requests(self):
 
 class TestFwFingerprintTiming(unittest.TestCase):
   N: int = 5
```
sshane, line 179: Might be able to just run once now

PR #29714, opened 2023-08-30
Docker: split openpilot-base into two stages

jnewb1:
this will allow us to cache the ubuntu dependencies even if we change something in pypoetry, reducing build times

also will allow us to insert a "user" step, to run as non-root

### DIFF
--- a/Dockerfile.openpilot_base
+++ b/Dockerfile.openpilot_base
@@ -33,6 +33,7 @@ COPY tools/install_python_dependencies.sh /tmp/tools/
 RUN cd /tmp && \
     tools/install_python_dependencies.sh && \
     rm -rf /tmp/* && \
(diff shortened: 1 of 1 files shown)


adeebshihadeh:
How much does this increase the overall image size with the additional layers?

jnewb1:
> How much does this increase the overall image size with the additional layers?

same size. Apt and python cache are still removed so the only difference is which layer the files are stored at.

image

gregjhogan:
There is probably a large poetry cache directory that isn't getting cleaned up.  It is usually a good idea to run `rm -rf $HOME/.cache/pypoetry/cache $HOME/.cache/pypoetry/artifacts $HOME/.cache/pip` in the same command that installs the pip packages.  You can check with https://github.com/wagoodman/dive.

PR #29718, opened 2023-08-30
Revert "ui/driverview: inherit from CameraWidget"

sshane:
Reverts commaai/openpilot#29661

### DIFF
--- a/selfdrive/ui/qt/offroad/driverview.cc
+++ b/selfdrive/ui/qt/offroad/driverview.cc
@@ -3,33 +3,62 @@
 #include <algorithm>
 #include <QPainter>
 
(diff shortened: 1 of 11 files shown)


sshane:
@deanlee this emits done whenever the interactive timeout is reached, which calls `slayout->setCurrentWidget(home);` and puts the device permanently offroad

deanlee:
@sshane fixed and reopened: https://github.com/commaai/openpilot/pull/29769

emit done on visible to fix the bug: https://github.com/commaai/openpilot/pull/29769/commits/aa07723122e76ccc13ebb3a7f3e9a8e02784db66

PR #29725, opened 2023-08-31
timezoned: set headers

sshane:
Issue: for new devices, if API responds with 429 / Too Many Requests, timezoned waits an hour to retry, leaving all routes until then with incorrect local times (it uses the UTC time).

Found 3 cases of time jumping backward just looking at the last 9 dongles to first fingerprint.

User-Agent matters to the timezone API. This is running on a prime device over LTE:

https://github.com/commaai/openpilot/assets/25857203/ac9509de-125b-4edf-82ed-584d04d09f9e

### DIFF
--- a/system/timezoned.py
+++ b/system/timezoned.py
@@ -11,7 +11,9 @@ from timezonefinder import TimezoneFinder
 from openpilot.common.params import Params
 from openpilot.system.hardware import AGNOS
 from openpilot.system.swaglog import cloudlog
+from openpilot.system.version import get_version
 
+REQUEST_HEADERS = headers = {'User-Agent': "openpilot-" + get_version()}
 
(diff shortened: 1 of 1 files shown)


system/timezoned.py
```diff
@@ -11,7 +11,9 @@
 from openpilot.common.params import Params
 from openpilot.system.hardware import AGNOS
 from openpilot.system.swaglog import cloudlog
+from openpilot.system.version import get_version
 
+REQUEST_HEADERS = headers = {'User-Agent': "openpilot-" + get_version()}
```
sshane, line 16:
```suggestion
REQUEST_HEADERS = {'User-Agent': "openpilot-" + get_version()}
```

adeebshihadeh (approved):

PR #29727, opened 2023-08-31
locationd: Add camera-IMU cross-checks

nuwandavek:
Problem: Erroneous gyro measurements cause a degradation in the localizer, which adversely affects driving. We should detect the bad measurements, do not update them to the filter, and finally add an alert.
image

To Do:
- [x]  Crosscheck gyro and cam odo rot to each other
- [x] Fix the decays to ensure sane alert duration
- [x] Run offline on a large seglist and compute TP/FP


nuwandavek:
@adeebshihadeh I think we can have an alert (I checked on 50k segs, without any FP). But we could also maybe add without alert on `inputsOK`, track closely offline and then add it in the next release. Don't have strong opinions on it, but it should have the alert eventually for sure.

selfdrive/controls/controlsd.py
```diff
@@ -378,7 +378,7 @@ def update_events(self, CS):
       self.events.add(EventName.vehicleModelInvalid)
     if not self.sm['lateralPlan'].mpcSolutionValid:
       self.events.add(EventName.plannerError)
-    if not (self.sm['liveParameters'].sensorValid or self.sm['liveLocationKalman'].sensorsOK) and not NOSENSOR:
+    if not (self.sm['liveParameters'].sensorValid or self.sm['liveLocationKalman'].sensorsOK or self.sm['liveLocationKalman'].inputsOK) and not NOSENSOR:
```
nuwandavek, line 381: `and` instead of `or`? It should trigger on all checks.

adeebshihadeh, line 381: `inputsOK` captures a lot more than sensors (e.g. `carState` going invalid). we need to be very careful about alert text; this seems like it deserves its own alert

adeebshihadeh (changes requested):

selfdrive/locationd/locationd.cc
```diff
@@ -711,9 +720,9 @@ int Localizer::locationd_thread() {
     // 100Hz publish for notcars, 20Hz for cars
     const char* trigger_msg = sm["carParams"].getCarParams().getNotCar() ? "accelerometer" : "cameraOdometry";
     if (sm.updated(trigger_msg)) {
-      bool inputsOK = sm.allAliveAndValid() && this->are_inputs_ok();
+      bool inputsOK = sm.allAliveAndValid();
       bool gpsOK = this->is_gps_ok();
-      bool sensorsOK = sm.allAliveAndValid({"accelerometer", "gyroscope"});
+      bool sensorsOK = sm.allAliveAndValid({"accelerometer", "gyroscope"}) && this->are_inputs_ok();
```
adeebshihadeh, line 725: doesn't this include a bunch of non-sensor stuff?

nuwandavek: It has `{"cameraOdometry", "liveCalibration", "accelerometer", "gyroscope"}`. 
All are necessary to be checked, IMO. Split this to sensor only for now?

nuwandavek:
`excessiveResets` triggered `localizerMalfunction` alert, but did not do anything. In this PR, that alert has been repurposed to be the catch-all localizer alert, and `inputsOK` triggers it now.

selfdrive/controls/controlsd.py
```diff
@@ -378,9 +378,11 @@ def update_events(self, CS):
       self.events.add(EventName.vehicleModelInvalid)
     if not self.sm['lateralPlan'].mpcSolutionValid:
       self.events.add(EventName.plannerError)
-    if not (self.sm['liveParameters'].sensorValid or self.sm['liveLocationKalman'].sensorsOK) and not NOSENSOR:
+    if not (self.sm['liveParameters'].sensorValid and self.sm['liveLocationKalman'].sensorsOK) and not NOSENSOR:
```
adeebshihadeh, line 381: not relevant to this PR, should be split out

PR #29728, opened 2023-08-31
Revert "ui/map: delete map after unsubscribing from prime"

sshane:
Reverts commaai/openpilot#29136


sshane:
@incognitojam can you merge this if the fix is not easy? When you delete `MapPanel`, its signals still exist, so on interaction timeout, it tries to set a widget that doesn't exist:

https://github.com/commaai/openpilot/blob/6a68944c638029361a7776cff08816919ecac351/selfdrive/ui/qt/maps/map_panel.cc#L17-L19

incognitojam:
I think we can fix it by providing a context to Qt

PR #29732, opened 2023-08-31
timezoned: check other methods while waiting for IP lookup timeout

incognitojam:
Don't sleep after IP lookup attempt, as GPS position may become available while we are waiting


adeebshihadeh:
This doesn't really make a difference in this case anyway, since you need to go onroad to get your first GPS position.

PR #29733, opened 2023-08-31
Hyundai CAN-FD: Panda safety config assignments cleanup

sunnyhaibin:
- [x] Consolidate safety mode assignments for CAN and CAN-FD platforms
- [x] Unify HDA2-related flag and safety config assignments
- [x] Consolidate camera SCC safety config assignments

**Prerequisite of**
- https://github.com/commaai/openpilot/pull/27392
- https://github.com/commaai/openpilot/pull/27478


selfdrive/car/hyundai/interface.py
```diff
@@ -283,30 +294,20 @@ def _get_params(ret, candidate, fingerprint, car_fw, experimental_long, docs):
...
-        # these cars require a special panda safety mode due to missing counters and checksums in the messages
-        ret.safetyConfigs = [get_safety_config(car.CarParams.SafetyModel.hyundaiLegacy)]
-      else:
-        ret.safetyConfigs = [get_safety_config(car.CarParams.SafetyModel.hyundai, 0)]
-
-      if candidate in CAMERA_SCC_CAR:
-        ret.safetyConfigs[0].safetyParam |= Panda.FLAG_HYUNDAI_CAMERA_SCC
 
+    if ret.flags & HyundaiFlags.CANFD_CAMERA_SCC or candidate in CAMERA_SCC_CAR:
+      ret.safetyConfigs[-1].safetyParam |= Panda.FLAG_HYUNDAI_CAMERA_SCC
```
sshane, line 310: We should only have one source of truth for this, why two?

PR #29745, opened 2023-09-01
gnss raw measurements through nmea port

gregjhogan:


adeebshihadeh:
Let's merge with some comments?

PR #29749, opened 2023-09-01
CAR docs: add video for Hyundai Kona Electric 2023 (KOR) HDA2

haram-KONA:
Add Driving Video for Hyundai Kona Electric 2023 (KOR) HDA2 
update values.py


adeebshihadeh:
Thanks!

PR #29753, opened 2023-09-01
Subaru: add GEN1 longitudinal test route

### LINKED ISSUE #29511: Subaru: add longitudinal safety route
We have long behind a manual toggle, we should get add a test route with long enabled.

jnewb1:


sshane:
any blocker?

jnewb1:
> any blocker? let's also add a gen 2

seems like an issue similar to: https://github.com/commaai/openpilot/commit/5d1550de871b0ac036b1dd44545fd7ff17fb91cd

was going to look into it more tomorrow.

also, because the longitudinal is behind a comment, I don’t think this actually doesn’t anything as is, it’s just running with stock long. I was going to look to write something in test_models that allows you to temporarily change the get_params to enable long. I think that would also be useful for testing cars that are in dashcam mode

sshane:
I think I had this conversation with adeeb about process replay and a dashcam car, we didn't want to add more code to test the dashcam route. Adding as a comment is fine, we can uncomment once we merge gen 2 long then.

selfdrive/car/tests/routes.py
```diff
@@ -242,6 +242,7 @@ class CarTestRoute(NamedTuple):
   CarTestRoute("c321c6b697c5a5ff|2020-06-23--11-04-33", SUBARU.FORESTER),
   CarTestRoute("791340bc01ed993d|2019-03-10--16-28-08", SUBARU.IMPREZA),
   CarTestRoute("8bf7e79a3ce64055|2021-05-24--09-36-27", SUBARU.IMPREZA_2020),
+  CarTestRoute("8de015561e1ea4a0|2023-08-29--17-08-31", SUBARU.IMPREZA), # openpilot longitudinal
```
sshane, line 245:
```suggestion
  # CarTestRoute("8de015561e1ea4a0|2023-08-29--17-08-31", SUBARU.IMPREZA),  # openpilot longitudinal
```

sshane (approved):
make sure you preserve (not required but useful for now) and run update_ci_routes.py

PR #29756, opened 2023-09-01
CI: run unittests in parallel with xdist

jnewb1:
depends: https://github.com/commaai/openpilot/pull/29800

parallelize ~~some~~ all of the tests

### DIFF
--- a/.github/workflows/selfdrive_tests.yaml
+++ b/.github/workflows/selfdrive_tests.yaml
@@ -249,7 +249,7 @@ jobs:
       timeout-minutes: 40
       run: |
         ${{ env.RUN }} "export SKIP_LONG_TESTS=1 && \
-                        $PYTEST && \
+                        $PYTEST -n auto --dist=loadscope && \
                         selfdrive/locationd/test/_test_locationd_lib.py && \
                         ./system/ubloxd/tests/test_glonass_runner && \
                         ./selfdrive/ui/tests/create_test_translations.sh && \
--- a/selfdrive/car/tests/test_lateral_limits.py
+++ b/selfdrive/car/tests/test_lateral_limits.py
@@ -32,7 +32,7 @@ ABOVE_LIMITS_CARS = [
 car_model_jerks: DefaultDict[str, Dict[str, float]] = defaultdict(dict)
 
 
-@parameterized_class('car_model', [(c,) for c in CAR_MODELS])
+@parameterized_class('car_model', [(c,) for c in sorted(CAR_MODELS)])
 class TestLateralLimits(unittest.TestCase):
   car_model: str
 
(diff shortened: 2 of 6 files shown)


jnewb1:
> Which ones can't be parallel now that we have the prefix fixture setup? I'd rather have parallel by the default and mark tests as not parallelizable.

logmessaged and loggerd are two that are failing right now, starting to fix those:
https://github.com/commaai/openpilot/pull/29800

adeebshihadeh:
Let's fix those properly to respect the prefix and switch the marker to "not parallel"

selfdrive/car/tests/test_lateral_limits.py
```diff
@@ -32,7 +32,7 @@
 car_model_jerks: DefaultDict[str, Dict[str, float]] = defaultdict(dict)
 
 
-@parameterized_class('car_model', [(c,) for c in CAR_MODELS])
+@parameterized_class('car_model', [(c,) for c in sorted(CAR_MODELS)])
```
jnewb1, line 35: for xdist, the test ordering has to be deterministic

adeebshihadeh: good catch!

selfdrive/controls/tests/test_cruise_speed.py
```diff
@@ -31,21 +32,21 @@ def run_cruise_simulation(cruise, e2e, t_end=20.):
   return output[-1, 3]
 
 
+@parameterized_class(("e2e", "personality", "speed"), itertools.product(
+                      [True, False], # e2e
+                      [log.LongitudinalPersonality.relaxed, # personality
```
adeebshihadeh, line 37: use `log.LongitudinalPersonality.schema.enumerants` for this list

.github/workflows/selfdrive_tests.yaml
```diff
@@ -249,7 +249,7 @@ jobs:
       timeout-minutes: 40
       run: |
         ${{ env.RUN }} "export SKIP_LONG_TESTS=1 && \
-                        $PYTEST && \
+                        $PYTEST -n auto --dist=loadscope && \
```
adeebshihadeh, line 252: we should consider adding this as the default in addopts in pyproject.toml

adeebshihadeh (approved):

jnewb1:
going to run the tests a few times to make sure there isn't any issues. Looks like it's around 10-11 minutes now, would probably be 7-8 on buildjet

adeebshihadeh:
Make sure to test aggressively locally as well. I usually run new tests in a loop like this `time while true; do ./test.py || break; done`

jnewb1:
Fixed a couple issues and ran it locally 20+ times without issues.

adding it to pypoetry causes issue on jenkins, and we probably don't want to run those tests in parallel anyway:
https://jenkins.comma.life/blue/organizations/jenkins/openpilot/detail/parallel-tests/22/pipeline#step-327-log-90

Maybe we just make a local testing script? like selfdrive/test/run_unittests?

adeebshihadeh:
We should handle that with a marker, but it's fine to do later.

PR #29759, opened 2023-09-02
test_models: detect when relay opens

sshane:
can messages from radar on bus 1 will not always be present, if fingerprinting, since it can be multiplexed to the OBD port. skip these messages

Ignore the name `ACC_CONTROL`, had the wrong DBC. It's a message from the radar.

image


fredyshox:
Fixes my radar_tests in #29672

fredyshox:
Seems like  `test_panda_safety_rx_valid` fails on the same segment from AVALONH 2019 as my radar test

fredyshox (approved):

selfdrive/car/tests/test_models.py
```diff
@@ -130,6 +133,18 @@ def setUpClass(cls):
             if param.key == 'OpenpilotEnabledToggle':
               enabled_toggle = param.value.strip(b'\x00') == b'1'
 
+        # Log which can frame the panda safety mode left ELM327, for CAN validity and relay malfunction checks
```
sshane, line 136:
```suggestion
        # Log which can frame the panda safety mode left ELM327, for CAN validity checks
```

fredyshox:
LGTM! (still works with my radar case)

selfdrive/car/tests/test_models.py
```diff
@@ -238,9 +253,9 @@ def test_panda_safety_rx_valid(self):
         if t > 1e6:
           self.assertTrue(self.safety.addr_checks_valid())
 
-          # No need to check relay malfunction on disabled routes (relay closed) or for reasonable fingerprinting time
-          # TODO: detect when relay has flipped to properly check relay malfunction
-          if self.openpilot_enabled and t > 5e6:
+          # No need to check relay malfunction on disabled routes (relay closed),
+          # or before fingerprinting is done (1s of tolerance)
+          if self.openpilot_enabled and t / 1e4 > (self.elm_frame + 100):
```
sshane, line 258: The reason we need this extra second of tolerance is because the safety mode went to silent after elm, before finally switching to the car safety mode and opening the relay. We can fix that, but we'd need two frame variables. Tolerance is fine for now.

adeebshihadeh: this should be a comment in the code, not the PR

PR #29764, opened 2023-09-03
ui: Countdown timer for experimental features and device critical actions

sunnyhaibin:
**Description**

This PR introduces countdown timers to enhance the user experience during critical actions within the device menu, toggles menu, and software menu. This mechanism ensures that users pause and review the operation's description before confirming, reducing the likelihood of unintended or unanticipated consequences. The countdown timers are designed to encourage users, especially newcomers, to engage with and understand the implications of specific high-risk actions, thereby reducing user confusion with features and contributing to a more informed user base with reduced toggle confusion.

**Implemented Countdown Timers**

When a countdown timer is in progress, the `confirm` button is intentionally disabled. Users are unable to confirm their action until the countdown timer completes its countdown, at which point the `confirm` button is enabled.

- **Device Menu**
  - `Reboot`
    - Onroad: 3-second countdown
    - Offroad: No countdown

- **Toggles Menu**
  - `openpilot Longitudinal Control (Alpha)`: 5-second countdown
  - `Experimental Mode` (without `Confirmed` flag added to the param): 5-second countdown

- **Software Menu**
  - `Uninstall`: 3-second countdown

**Verifications**
<table>
  <tr>
    <th colspan="2">During Countdown</th>
    <th colspan="2">After Countdown</th>
  </tr>
  <tr>
    <td colspan="4" align="center"><strong>Device - Reboot</strong></td>
  </tr>
  <tr>
    <td colspan="2">Image</td>
    <td colspan="2">Image</td>
  </tr>
  <tr>
    <td colspan="4" align="center"><strong>Toggles - openpilot Longitudinal Control (Alpha)</strong></td>
  </tr>
  <tr>
    <td colspan="2">Image</td>
    <td colspan="2">Image</td>
  </tr>
  <tr>
    <td colspan="4" align="center"><strong>Toggles - Experimental Mode</strong></td>
  </tr>
  <tr>
    <td colspan="2">Image</td>
    <td colspan="2">Image</td>
  </tr>
  <tr>
    <td colspan="4" align="center"><strong>Software - Uninstall</strong></td>
  </tr>
  <tr>
    <td colspan="2">Image</td>
    <td colspan="2">Image</td>
  </tr>
</table>


sshane:
I don't think placing the timeout on power settings is a good idea

sunnyhaibin:
> I don't think placing the timeout on power settings is a good idea

I agree, will remove the timeout for Power Off.

sshane:
The only thing I see this being useful for is the experimental and alpha longitudinal pop-ups, but this is a lot of code for something that doesn't really enforce reading either any more than what we do now. @adeebshihadeh any thoughts?

sunnyhaibin:
Some mobile apps have this type of UX to somewhat enforce the pop up to pause and have the content of the toggle on the screen before confirming. This is more towards new users who have never enabled Experimental and alpha longitudinal, and we had seen a fair share of new users that enabled these toggles out of excitements without understanding the implications (FCW/AEB is disable, "Experimental Mode is an alpha feature, mistakes should be expected", etc.)

https://github.com/commaai/openpilot/assets/47793918/0c2231d4-a81a-4be9-a4f7-f231e3875d10

deanlee:
From a code perspective, there is indeed too much code for this. +10 lines would be enough. you can take https://github.com/commaai/openpilot/compare/master...deanlee:ui_countdown_confirm?expand=1 as an example. feel free to copy what you need if you think it's useful

[Kazam_screencast_00035.webm](https://github.com/commaai/openpilot/assets/27770/4bca05a8-00f5-4f25-bb6b-36b8a3627ade)

sunnyhaibin:
> From a code perspective, there is indeed too much code for this. +10 lines would be enough. you can take https://github.com/commaai/openpilot/compare/master...deanlee:ui_countdown_confirm?expand=1 as an example. feel free to copy what you need if you think it's useful
> 
>  Kazam_screencast_00035.webm

@deanlee This cleanup is great, I will include them here, thanks!

selfdrive/ui/qt/widgets/input.cc
```diff
@@ -220,6 +222,18 @@ ConfirmationDialog::ConfirmationDialog(const QString &prompt_text, const QString
     confirm_btn->setObjectName("confirm_btn");
     btn_layout->addWidget(confirm_btn);
     QObject::connect(confirm_btn, &QPushButton::clicked, this, &ConfirmationDialog::accept);
+    countdown = _count_down != 0 ? _count_down : 0;
```
deanlee, line 225: why not init it in ctor's member initializer lists?  `ConfirmationDialog(..., int count_down) : countdown(count_down), ... {`

sunnyhaibin: Fixed!

sunnyhaibin:
Do we want the `Reboot` button to have a timeout for both onroad/offroad?

Emalton:
I personally wouldn't like a timer for rebooting at all. There are times people need to reboot the C3 while driving, and taking their attention (especially waiting 3 seconds) isn't the safest in my opinion. I think the confirmation dialog that pops up when rebooting is good enough to prevent fumbling.

sshane:
Quickly discussed and we don't think this is an effective way to get users to read the popup, and degrades user experience.

PR #29771, opened 2023-09-04
Hyundai: Car Port for Nexo 2021

sunnyhaibin:
**Checklist**
- [x] added entry to CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route added to [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with stock system: `a74afe0cf708748f|2023-09-02--19-56-27`
- [x] route with openpilot: `a74afe0cf708748f|2023-09-10--11-18-49`
- [x] harness Type: [Hyundai H](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)
- [x] architecture: CAN

**Link to all physical parts during installation**
  - [comma 3X](https://comma.ai/shop/comma-3x)
  - [car harness kit](https://comma.ai/shop/car-harness)
    - [Hyundai H](https://github.com/commaai/openpilot/wiki/Hyundai-Kia-Genesis#harness-guide)

**Requires**
- https://github.com/commaai/panda/pull/1637
- https://github.com/commaai/opendbc/pull/933
- https://github.com/commaai/openpilot/pull/29772

Thanks to community 2021 Hyundai Nexo owner SimonYang402.

PR #29772, opened 2023-09-04
Hyundai CAN: accelerator pedal signals cleanup

sunnyhaibin:
- Consolidated common accelerator messages and signals for maintainability
- Prerequisite of https://github.com/commaai/openpilot/pull/29771 due to a new accelerator pedal signal being added


sshane:
the extra if statement seems cleaner to me

PR #29779, opened 2023-09-05
Hyundai longitudinal: enable for `KONA_EV`

sunnyhaibin:
Split from https://github.com/commaai/openpilot/pull/26967

**Checklist**
- [x] updated entry in CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route updated in [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot longitudinal: `f90d3cd06caeb6fa|2023-09-06--17-15-47`


sshane:
Plot with cars that have the FCA messages for reference:

https://github.com/commaai/openpilot/pull/28479#issuecomment-1683423682

sunnyhaibin:
> Plot with cars that have the FCA messages for reference:
> 
> [#28479 (comment)](https://github.com/commaai/openpilot/pull/28479#issuecomment-1683423682)
> 
> This doesn't have FCA12, but it doesn't fault when you send it, correct?

Yes, that's correct. Per the user, only the AEB amber lamp is displayed on the cluster (expected behavior).

Issue #25221, opened 2022-07-20
Kia Niro 2019 - "Check FCA" warning when engaging openpilot


DerGue69:
### Describe the bug

I have a Kia Niro EV 2019 (EU - Austria - 1st Edition). A "Hyundai H" Harness is my connector.
After installin stock OP my car was not recognized. I had to add values to values.py from my cars fingerprint (see additional info)
(I thought My Niro EV is already compatible with OpenPilot)

After that OP started working with my car. But everytime I activate OP I get a "Check FCA ( Forward Coll. Avoidance Assist)" warning. 

I recorded a route and made public.
https://cabana.comma.ai/?route=6111bd7df5043394%7C2022-07-20--14-47-58&exp=1689861829&sig=9eylZ8Csg1lvrMr6yneTzSZbSC7qN5ZjymwZ2jOTkVM%3D&max=-1&url=https%3A%2F%2Fchffrprivate.azureedge.net%2Fchffrprivate3%2Fv2%2F6111bd7df5043394%2Fc7e80e23a0c545d687808655907ee3ee_2022-07-20--14-47-58

Please help me to get rid of this error ;)

![image](https://user-images.githubusercontent.com/96002277/180004334-6e2aa636-f70d-49ec-90a5-5fc3a8c6dd47.png)

### Which car does this affect?

Kia Niro EV 2019 (EU - Austria - 1st Edition)

### Provide a route where the issue occurs

6111bd7df5043394%7C2022-07-20--14-47-58

### openpilot version

0.8.14

### Additional info

![image](https://user-images.githubusercontent.com/96002277/180004251-9bf3ca9c-be34-4e08-9088-809c5e9646a8.png)

'carFingerprint': 'KIA NIRO EV 2020',
               'carFw': [{'address': 2004,
                          'ecu': 'eps',
                          'fwVersion': b'\xf1\x00DE  MDPS C 1.00 1.04 56310Q4100\x00 4DEEC104',
                          'responseAddress': 0,
                          'subAddress': 0},
                         {'address': 2000,
                          'ecu': 'fwdRadar',
                          'fwVersion': b'\xf1\x00DEev SCC F-CUP      1.00 1.02 96400-Q4000         ',
                          'responseAddress': 0,
                          'subAddress': 0},
                         {'address': 1988,
                          'ecu': 'fwdCamera',
                          'fwVersion': b'\xf1\x00DEE MFC  AT EUR LHD 1.00 1.03 95740-Q4000 180821',
[...]

PR #29780, opened 2023-09-05
Hyundai longitudinal: enable for `KIA_CEED`

sunnyhaibin:
Split from https://github.com/commaai/openpilot/pull/26967

**Checklist**
- [x] updated entry in CarInfo in selfdrive/car/*/values.py and ran `selfdrive/car/docs.py` to generate new docs
- [x] test route updated in [routes.py](https://github.com/commaai/openpilot/blob/master/selfdrive/car/tests/routes.py)
- [x] route with openpilot longitudinal: `99315f265be0b26c|2023-09-07--14-32-24`


sshane:
Let's avoid letting small PRs like this sit around, opening only when ready to go.

sshane:
This has FCA11, but still faults, so we're probably sending unexpected values.

https://github.com/commaai/openpilot/pull/28479#issuecomment-1683423682

sshane:
@sunnyhaibin let me know when you do some investigation.

sunnyhaibin:
@sshane Probably similar to this
- https://github.com/commaai/openpilot/issues/25221

